PR #2188, opened 2022-09-15
Pattern matching syntax and semantics

zygoloid:
This paper proposes concrete syntax and semantic choices for Carbon patterns.

PR #2022, opened 2022-08-13
Unused Pattern Bindings (Unused Function Parameters)

matthew-russo:
This proposal specifies how unused pattern bindings are written in the Carbon programming language. This is a more general problem statement of "how do users specify unused function parameters" as function parameter declarations are a more specific form of pattern

Related issue #1996

PR #2853, opened 2023-05-25
Document pattern syntax and semantics

### LINKED ISSUE #2852: Document pattern syntax and semantics
### Description of the bug:

`docs/design/pattern_matching.md` hasn't been updated to reflect the adoption of #2188. In particular, it doesn't document the syntax or semantics of patterns themselves, only the constructs that use them (which it correctly marks as provisional).

### Link to documentation:

https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/pattern_matching.md

_No response_

_No response_

geoffromer:
Integrates the content of #2188 into `pattern_matching.md`

Closes #2852

### DIFF
--- a/docs/design/pattern_matching.md
+++ b/docs/design/pattern_matching.md
@@ -35,7 +35,6 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
     -   [Pattern matching in local variables](#pattern-matching-in-local-variables)
 -   [Open questions](#open-questions)
     -   [Slice or array nested value pattern matching](#slice-or-array-nested-value-pattern-matching)
-    -   [Generic/template pattern matching](#generictemplate-pattern-matching)
     -   [Pattern matching as function overload resolution](#pattern-matching-as-function-overload-resolution)
 -   [Alternatives considered](#alternatives-considered-5)
 -   [References](#references)
@@ -44,12 +43,12 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 ## Overview
 
-The most prominent mechanism to manipulate and work with types in Carbon is
-pattern matching. This may seem like a deviation from C++, but in fact this is
-largely about building a clear, coherent model for a fundamental part of C++:
-overload resolution.
-
-We refer to the value being matched by a pattern as the _scrutinee_.
+A _pattern_ is an expression-like syntax that describes the structure of some
+value. The pattern may contain unknowns, so it can potentially match multiple
+values, and those unknowns may have names, in which case they are called
+_bindings_. When a pattern is executed by giving it a value called the
+_scrutinee_, it determines whether the scrutinee matches the pattern, and if so,
+determines the values of the bindings.
 
 ## Pattern Syntax and Semantics
 
@@ -734,11 +733,6 @@ a local variable named `p` which is then returned.
 
 An open question is how to effectively fit a "slice" or "array" pattern into
 nested value pattern matching, or whether we shouldn't do so.
-
-### Generic/template pattern matching
-
-An open question is going beyond a simple "type" to things that support generics
-and/or templates.
 
 ### Pattern matching as function overload resolution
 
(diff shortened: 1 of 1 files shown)


geoffromer:
FYI: commit f0f454a0ac is almost a raw cut-and-paste from p2188.md, with some edits at the beginning and end, so you can use that as a diff base to see the edits I made to that content.

docs/design/pattern_matching.md
```diff
@@ -157,22 +168,15 @@ fn H(n: i32) {
 }
 ```
 
-As specified in
-[#2022](/proposals/p2022.md#the-behavior-of-unused-name-bindings), the `unused`
-keyword indicates that the binding is intended to not be used.
```
zygoloid: I think we should include some text like this from #2022 describing what `unused` does:

> Names that are qualified with the `unused` keyword are visible for name lookup
> but uses are invalid, including when they cause ambiguous name lookup errors. If
> attempted to be used, a compiler error will be shown to the user, instructing
> them to either remove the `unused` qualifier or remove the use.

docs/design/pattern_matching.md
```diff
@@ -186,7 +190,7 @@ fn F(n: i32) {
 As specified in [#1084](/proposals/p1084.md), function redeclarations may
 replace named bindings with wildcards but may not use different names.
```
zygoloid: Due to the change in section name, we no longer say what a "wildcard" is.

```suggestion
replace binding names with `_`s but may not use different names.
```

docs/design/pattern_matching.md
```diff
@@ -207,7 +219,7 @@ A `:!` can be used in place of `:` for a binding that is usable at compile time.
 -   _generic-pattern_ ::= `template`? `_` `:!` _expression_
```
zygoloid, line 223: Perhaps:

```suggestion
-   _generic-pattern_ ::= `template`? _identifier_ `:!` _expression_
-   _generic-pattern_ ::= `template`? `_` `:!` _expression_
-   _generic-pattern_ ::= `unused` `template`? _identifier_ `:!` _expression_
```

... for consistency with the changes to the previous section.

docs/design/pattern_matching.md
```diff
@@ -685,13 +690,32 @@ In order to match a value, whatever is specified in the pattern must match.
 Using `auto` for a type will always match, making `_: auto` the wildcard
 pattern.
 
+#### Guards
```
zygoloid, line 698: Do we have an approved design for `match` statements and their syntax? This would seem to fit better in a separate documentation file specifically describing `match` statements, `case`, `default`, etc. than here, if we're ready to write such a file.

docs/design/pattern_matching.md
```diff
@@ -718,3 +742,15 @@ and/or templates.
```
zygoloid, line 742: (Can't comment on the right line, sorry.)

Can we remove the "Generic/template pattern matching" section above?

docs/design/pattern_matching.md
```diff
@@ -53,7 +60,7 @@ not permitted as proper patterns, so cannot contain bindings.
```
zygoloid, line 59: Can we drop some text from / rewrite the overview? Once we have an overloading design it might make sense to add some of that back, but for now it seems out of place to talk about overload resolution here.

zygoloid (approved):
Looks good. Thanks for separating the proposal text from the edits; that made the review much easier :)

Some relatively minor comments, but I think this is fine to land as-is and iterate from there if needed.

docs/design/pattern_matching.md
geoffromer: IIRC there hasn't been an approved proposal yet, but I could split the existing discussion of `match` into a separate file if you like.

docs/design/pattern_matching.md
geoffromer: How's this?

docs/design/pattern_matching.md
zygoloid: Let's defer that for now.

docs/design/pattern_matching.md
zygoloid: I think this comment is still unresolved.

geoffromer: Whoops, missed that. Done.

PR #2855, opened 2023-05-26
Canonicalize struct types.

jonmeow:
This adds canonicalization of struct types based on their type fields. It obsoletes the current CanImplicitAsStruct because the type ids should now be identical when they're structurally identical; there's only a reason to implicit CanImplicitAsStruct to detect _compatible_ conversions.

The type fields themselves aren't canonicalized because it would need to be done during the first parse, and could yield name conflicts being associated with the wrong location. i.e.:

```
var x: {a: i32, a: i32};
var y: {a: i32, b: i32, a: i32};
```

This should yield two separate name conflict diagnostics pointing at the type fields for each respective line, but if struct type fields were canonicalized then both would point at the first `a: i32` field definition. This isn't expected to be an issue for types because I'm trying to print those, but we may also end up with a "first defined at" situation in some cases (still, less confusing because the type should match). Regardless, I think individual fields gets much more awkward.

### DIFF
--- a/toolchain/semantics/semantics_context.cpp
+++ b/toolchain/semantics/semantics_context.cpp
@@ -282,48 +282,12 @@ auto SemanticsContext::ImplicitAsImpl(SemanticsNodeId value_id,
     }
   }
 
-  if (value_type_id != SemanticsTypeId::TypeType &&
-      as_type_id != SemanticsTypeId::TypeType) {
-    auto value_type = semantics_->GetNode(semantics_->GetType(value_type_id));
-    auto as_type = semantics_->GetNode(semantics_->GetType(as_type_id));
-    if (CanImplicitAsStruct(value_type, as_type)) {
-      // Under the current implementation, struct types are only allowed to
-      // ImplicitAs when they're equivalent. What's really missing is type
-      // consolidation such that this would fall under the above `value_type_id
-      // == as_type_id` case. In the future, this will need to handle actual
-      // conversions.
-      return ImplicitAsKind::Identical;
-    }
-  }
+  // TODO: Handle ImplicitAs for compatible structs and tuples.
 
   if (output_value_id != nullptr) {
     *output_value_id = SemanticsNodeId::BuiltinInvalidType;
   }
   return ImplicitAsKind::Incompatible;
-}
-
-auto SemanticsContext::CanImplicitAsStruct(SemanticsNode value_type,
-                                           SemanticsNode as_type) -> bool {
-  if (value_type.kind() != SemanticsNodeKind::StructType ||
-      as_type.kind() != SemanticsNodeKind::StructType) {
-    return false;
-  }
-  auto value_type_refs = semantics_->GetNodeBlock(value_type.GetAsStructType());
-  auto as_type_refs = semantics_->GetNodeBlock(as_type.GetAsStructType());
-  if (value_type_refs.size() != as_type_refs.size()) {
-    return false;
-  }
-
-  for (int i = 0; i < static_cast<int>(value_type_refs.size()); ++i) {
-    auto value_type_field = semantics_->GetNode(value_type_refs[i]);
-    auto as_type_field = semantics_->GetNode(as_type_refs[i]);
-    if (value_type_field.type_id() != as_type_field.type_id() ||
-        value_type_field.GetAsStructTypeField() !=
-            as_type_field.GetAsStructTypeField()) {
-      return false;
-    }
-  }
-  return true;
 }
 
 auto SemanticsContext::ParamOrArgStart() -> void {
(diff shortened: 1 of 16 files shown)


toolchain/semantics/semantics_context.h
```diff
@@ -97,6 +97,16 @@ class SemanticsContext {
   // TODO: This should eventually return a type ID.
   auto CanonicalizeType(SemanticsNodeId node_id) -> SemanticsTypeId;
 
+  // Handles canonicalization of struct types. This may create a new struct type
+  // when it has a new structure, or reference an existing struct type when it
+  // duplicates a prior type.
+  //
+  // Individual struct type fields aren't canonicalized because they may have
+  // name conflicts or other diagnostics during creation, which can use the
+  // parse node.
```
chandlerc, line 107: Rather than avoiding canonicalizing struct type fields, what about refusing to canonicalize the type subtree that contains an invalid node?

(maybe that doesn't actually make sense...)

toolchain/semantics/semantics_context.h
```diff
@@ -161,6 +171,38 @@ class SemanticsContext {
...
+    static auto getHashValue(const KeyType& val) -> unsigned {
+      unsigned hash_value = 0;
+      for (const auto& [string_id, type_id] : val) {
+        hash_value = llvm::detail::combineHashValue(
+            hash_value, IntMapInfo::getHashValue(string_id.index));
+        hash_value = llvm::detail::combineHashValue(
+            hash_value, IntMapInfo::getHashValue(type_id.index));
+      }
+      return hash_value;
+    }
```
chandlerc: All of this may prove really inefficient, as in go quadratic or other sharp scaling cliffs that cause problems long before we get to benchmarking... DenseMap assumes its keys are reasonably small and easy to build.

LLVM actually has a utility for building exactly this kind of data structure that is optimized for the case you're handling: https://llvm.org/docs/ProgrammersManual.html#llvm-adt-foldingset-h

However, that type is *also* optimized for the case of a polymorphic object as the nodes, which you don't need. Not sure which will end up being dominant here.

toolchain/semantics/semantics_context.h
```diff
@@ -161,6 +171,38 @@ class SemanticsContext {
...
+
+    static auto getHashValue(const KeyType& val) -> unsigned {
+      unsigned hash_value = 0;
+      for (const auto& [string_id, type_id] : val) {
+        hash_value = llvm::detail::combineHashValue(
+            hash_value, IntMapInfo::getHashValue(string_id.index));
+        hash_value = llvm::detail::combineHashValue(
+            hash_value, IntMapInfo::getHashValue(type_id.index));
+      }
+      return hash_value;
```
chandlerc: If you do keep this approach, you should be able to use the range API to get something much more efficient w/o much effort: https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/ADT/Hashing.h#L484-L493

toolchain/semantics/semantics_context.cpp
```diff
@@ -380,6 +344,33 @@ auto SemanticsContext::CanonicalizeType(SemanticsNodeId node_id)
...
+  for (const auto& ref_id : refs) {
+    auto ref = semantics_->GetNode(ref_id);
+    fields.push_back({ref.GetAsStructTypeField(), ref.type_id()});
+  }
+
+  // If a struct with matching fields was already created, reuse it.
+  auto it = canonical_struct_types_.find(fields);
+  if (it != canonical_struct_types_.end()) {
+    return it->second;
+  }
```
chandlerc: If the `FoldingSet` approach doesn't work out (see below, or above... not sure which order my comments show up here), can you leave a TODO to revisit this for efficiency? Not worth doing it right away, but this seems likely to get into a loop doing heap allocation just to query the canonicalization. Pretty sure we can do better by incrementally computing the hash/key while walking the semantics rather than allocating it, or doing some other somewhat clever thing here. But seems much better to think about this in the future rather than now.

chandlerc (approved):
Really nice.

Bunch of comments about the implementation details / hashing. These aren't blocking or terribly urgent. Mostly wanted to record them.

Left more of a question around skipping canonicalization of fields you mention in the change decsription. Not sure I fully understood the implications of this, so sorry if my comment doesn't make sense. That's also not blocking, the change LG regardless. I mostly wanted to fully understand what's going on there.

toolchain/semantics/semantics_context.h
jonmeow: I think that whether structs with invalid nodes are canonicalized doesn't affect the issue with canonicalizing invalid type fields.

To benefit from canonicalizing struct type fields, it needs to be done during construction of each field -- before the type is finished. The second half of the conflict is only seen after canonicalization of the first half of the conflict has occurred, so whether or not that later node (or later yet, struct with invalid nodes) is canonicalized doesn't improve the ability to emit an error that correctly identifies the code location of the first half of the conflict.

toolchain/semantics/semantics_context.h
jonmeow: Switched to FoldingSet.

toolchain/semantics/semantics_context.h
jonmeow: Switched to FoldingSet.

toolchain/semantics/semantics_context.cpp
jonmeow: Switched to FoldingSet.

jonmeow:
Just to note, leaving the FoldingSet switch in a separate commit for (hopefully) easier review.

chandlerc (approved):
(still LG)

PR #2860, opened 2023-05-26
Add loads for data.

jonmeow:
Start inserting loads when getting a value. This is done conditionally, but in theory we could start tracking nodes which need loading differently if the `isa` is considered cumbersome.


zygoloid (approved):
Longer-term, I think we should be adding nodes to the semantics IR for loads, but this seems OK as a short-term workaround.

jonmeow:
> Longer-term, I think we should be adding nodes to the semantics IR for loads, but this seems OK as a short-term workaround.

Yeah, I tend towards agreeing -- probably more complex structures will cause the current approach to be overly complex.

PR #2862, opened 2023-05-31
Fix crash on use of uninitialized array element, and improve unformed checking

zygoloid:
Per #257, we should be treating unformedness as all-or-nothing, rather than being a per-field or per-array-element property. Previously we initialized an array with no explicit initializer as containing a sequence of uninitialized values, but that led to crashes when attempting to access those values, as the checks for reading an uninitialized value only expected values to be uninitialized at the top level.

Also, we had existing tests that attempt to store to an element of an uninitialized array. We now detect that and treat it as UB during evaluation, rather than crashing due to trying to perform field access into an uninitialized value.

Finally, many of these problems can be detected statically, but the resolve_unformed pass wasn't catching them because it missed a few expression and declaration forms. Support for those cases has been added too. This causes the pass to recurse more often, and in particular our existing recursion test started hitting a stack overflow after this, so resolve_unformed now uses `RunWithExtraStack`. In passing, remove the need to explicitly tell `RunWithExtraStack` the return type, and infer it as the return type of the callable instead.

### DIFF
--- a/explorer/interpreter/heap.cpp
+++ b/explorer/interpreter/heap.cpp
@@ -38,6 +38,11 @@ auto Heap::Write(const Address& a, Nonnull<const Value*> v,
                  SourceLocation source_loc) -> ErrorOr<Success> {
   CARBON_RETURN_IF_ERROR(this->CheckAlive(a.allocation_, source_loc));
   if (states_[a.allocation_.index_] == ValueState::Uninitialized) {
+    if (!a.element_path_.IsEmpty()) {
+      return ProgramError(source_loc)
+             << "undefined behavior: store to subobject of uninitialized value "
+             << *values_[a.allocation_.index_];
+    }
     states_[a.allocation_.index_] = ValueState::Alive;
   }
   CARBON_ASSIGN_OR_RETURN(values_[a.allocation_.index_],
--- a/explorer/interpreter/resolve_unformed.cpp
+++ b/explorer/interpreter/resolve_unformed.cpp
@@ -73,7 +73,7 @@ struct PendingResolveUnformedStep {
   Nonnull<const Expression*> expression;
   FlowFacts::ActionType action;
 };
-}
+}  // namespace
 
 // Resolve the formed/unformed state of an expression, and enqueue any
 // subexpressions to be checked, in reverse evaluation order.
@@ -114,6 +114,9 @@ static auto ResolveOneUnformedExpression(
         // When a variable is taken address of, defer the unformed check to
         // runtime. A more sound analysis can be implemented when a points-to
         // analysis is available.
+        // TODO: This isn't enough to permit &x.y or &x[i] when x is
+        // uninitialized, because x.y and x[i] both require x to be
+        // initialized.
         queue.push_back(
             {opt_exp.arguments().front(), FlowFacts::ActionType::Form});
       } else {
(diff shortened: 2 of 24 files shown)


explorer/interpreter/heap.cpp
```diff
@@ -38,6 +38,11 @@ auto Heap::Write(const Address& a, Nonnull<const Value*> v,
                  SourceLocation source_loc) -> ErrorOr<Success> {
   CARBON_RETURN_IF_ERROR(this->CheckAlive(a.allocation_, source_loc));
   if (states_[a.allocation_.index_] == ValueState::Uninitialized) {
+    if (!a.element_path_.IsEmpty()) {
+      return ProgramError(source_loc)
+             << "undefined behavior: store to subobject of uninitialized value "
```
jonmeow, line 43: I think this handles:

```
p.x = 0;
```

I worry a little that this leaves the door open for issues e.g. with:

```
// This line should be an error, right, since it's accessing the address of uninitialized?
var x: i32* = &p.x;
*x = 0;
```

Maybe that's handled in some other way? Might be worth testing?

explorer/interpreter/resolve_unformed.cpp
```diff
@@ -117,15 +123,35 @@ static auto ResolveUnformed(Nonnull<const Expression*> expression,
...
           flow_facts, FlowFacts::ActionType::Check));
       break;
+    case ExpressionKind::IndexExpression:
+      CARBON_RETURN_IF_ERROR(
+          ResolveUnformed(&cast<IndexExpression>(*expression).object(),
+                          flow_facts, FlowFacts::ActionType::Check));
+      CARBON_RETURN_IF_ERROR(
+          ResolveUnformed(&cast<IndexExpression>(*expression).offset(),
+                          flow_facts, FlowFacts::ActionType::Check));
+      break;
```
jonmeow, line 161:
```suggestion
    case ExpressionKind::IndexExpression: {
      const auto& index_exp = cast<IndexExpression>(*expression);
      CARBON_RETURN_IF_ERROR(
          ResolveUnformed(&index_exp.object(),
                          flow_facts, FlowFacts::ActionType::Check));
      CARBON_RETURN_IF_ERROR(
          ResolveUnformed(&index_exp.offset(),
                          flow_facts, FlowFacts::ActionType::Check));
      break;
    }
```

It took me a few looks to spot `object` vs `offset`, that this wasn't just a copy-paste mistake. Maybe factor out the cast?

(TBH, makes me think it'd also help if we cut out a step in `FlowFacts::ActionType::Check` to reduce the duplicate text, but that seems out of scope)

jonmeow (approved):
Thank you!

explorer/interpreter/heap.cpp
zygoloid: We reject that example in resolve_unformed -- we don't permit using `p.x` when `p` is definitely uninitialized, even when taking its address. I've added a TODO for that.

There is a corresponding case where the static check doesn't fire and this happens at runtime, though, and that is already handled here. The element path is part of the computed `Address` value, so in examples like yours the path will be non-empty. I've added a test for that.

explorer/interpreter/resolve_unformed.cpp
```diff
@@ -171,6 +176,22 @@ static auto ResolveUnformed(Nonnull<const Expression*> expression,
   return Success();
 }
 
+static auto ResolveUnformed(Nonnull<const Expression*> expression,
+                            FlowFacts& flow_facts, FlowFacts::ActionType action)
+    -> ErrorOr<Success> {
+  // We visit subexpressions in evaluation order, by performing a LIFO
+  // traversal here and enqueueing subexpressions in reverse order in
+  // ResolveOneUnformedExpression.
+  llvm::SmallVector<PendingResolveUnformedStep, 32> queue;
```
jonmeow: The explorer predominantly uses std types -- I think this would be the first use of SmallVector.  Would it make sense to switch?

jonmeow (commented):
First note: don't forget to update the PR description with substantial changes. :)

---

The changes for recursion look like a very narrow fix; e.g., if statements are still non-recursive and could blow the stack. We just happen to test one case of expressions, and your fix is limited to that.

WDYT of either:
- Using the stack space library to allow recursion without substantively changing the implementation.
- At least adding a TODO to fix the rest of the code to address it non-recursively, and taking that up as a task.

FWIW, the stack space library seems like the lower effort generic fix.

zygoloid:
> WDYT of either:
> 
> * Using the stack space library to allow recursion without substantively changing the implementation.

Sure, done.

jonmeow (approved):
LGTM.

Yeah I was mostly missing the right trick to elide the return type. Thank you for fixing that.

PR #2864, opened 2023-06-01
Refactor NodeStack APIs to better handle the increase of distinct ID types.

jonmeow:
I'm looking at adding CallableId, so figured I'd do this API refactoring which should reduce code duplication. This increases the amount of cross-calls between APIs because it may not be an issue for performance after optimizations, and should simplify reading of the API.

Also note the prior static_asserts on layout were missing a couple types, which is why I'm moving them into Entry where it's going to be more obvious when something's added.


toolchain/semantics/semantics_handle.cpp
```diff
@@ -70,10 +71,10 @@ auto SemanticsHandleDesignatedName(SemanticsContext& context,
 
 auto SemanticsHandleDesignatorExpression(SemanticsContext& context,
                                          ParseTree::Node parse_node) -> bool {
-  auto [_, name_id] = context.node_stack().PopForParseNodeAndNameId(
+  auto [_, name_id] = context.node_stack().PopWithParseNode<SemanticsStringId>(
       ParseNodeKind::DesignatedName);
```
zygoloid: It seems strange to pop with parse node then discard the parse node. Can you:
```suggestion
  auto name_id = context.node_stack().Pop<SemanticsStringId>(
      ParseNodeKind::DesignatedName);
```

toolchain/semantics/semantics_handle_struct.cpp
```diff
@@ -47,8 +49,8 @@ auto SemanticsHandleStructFieldUnknown(SemanticsContext& context,
 auto SemanticsHandleStructFieldValue(SemanticsContext& context,
                                      ParseTree::Node parse_node) -> bool {
   auto [value_parse_node, value_node_id] =
-      context.node_stack().PopForParseNodeAndNodeId();
-  auto [_, name_id] = context.node_stack().PopForParseNodeAndNameId(
+      context.node_stack().PopWithParseNode<SemanticsNodeId>();
+  auto [_, name_id] = context.node_stack().PopWithParseNode<SemanticsStringId>(
       ParseNodeKind::DesignatedName);
```
zygoloid: Same here:
```suggestion
  auto name_id = context.node_stack().Pop<SemanticsStringId>(
      ParseNodeKind::DesignatedName);
```

zygoloid (approved):

toolchain/semantics/semantics_handle.cpp
jonmeow: Done

toolchain/semantics/semantics_handle_struct.cpp
jonmeow: Done

PR #2760, opened 2023-04-12
Consistent `class` and `interface` syntax

josh11b:
Update syntax of `class` and `interface` definitions to be more consistent. Constructs that add names to the class or interface from another definition are always prefixed by the `extend` keyword.

Implements the decisions in:

-   [#995: Generics external impl versus extends](https://github.com/carbon-language/carbon-lang/issues/995),
-   [#1159: adaptor versus adapter may be harder to spell than we'd like](https://github.com/carbon-language/carbon-lang/issues/1159),
-   [#2580: How should Carbon handle conditionally implemented internal interfaces](https://github.com/carbon-language/carbon-lang/issues/2580), and
-   [#2770: Terminology for internal and external implementations](https://github.com/carbon-language/carbon-lang/issues/2770).

PR #2866, opened 2023-06-01
Update design docs with syntax changes from #2760

josh11b:
This includes:

- Syntax changes from this chart:

| Before                          | After                                   |
| ------------------------------- | --------------------------------------- |
| `class D extends B { ... }`     | `class D { extend base: B; ... }`       |
| `external impl C as Sub;`       | `impl C as Sub;`                        |
| `class C { impl as Sortable; }` | `class C { extend impl as Sortable; }`  |
| `adapter A for C { ... }`       | `class A { adapt C; ... }`              |
| `adapter A extends C { ... }`   | `class A { extend adapt C; ... }`       |
| `interface I { impl as J; }`    | `interface I { require Self impls J; }` |
| `interface I { extends J; }`    | `interface I { extend J; }`             |

- Dropping the syntax for conditionally implemented internal interfaces.

This does not include:

- terminology changes from #2760 ("internal" and "external")
- changes to code, such as explorer, toolchain, language grammars, or other tooling


docs/design/classes.md
```diff
@@ -2233,6 +2252,12 @@ the type of `U.x`."
 
     -   [No unqualified lookup when defining outside a scope](/proposals/p2287.md#no-unqualified-lookup-when-defining-outside-a-scope)
 
+-   [#2760: Consistent `class` and `interface` syntax](https://github.com/carbon-language/carbon-lang/pull/2760)
+<!-- FIXME: Include once 2760 is merged.
+    -   [Use `extends` instead of `extend`](/proposals/p2760.md#use-extends-instead-of-extend)
+    -   [List base class in class declaration](/proposals/p2760.md#list-base-class-in-class-declaration)
+-->
+
```
jonmeow:
```suggestion
    -   [Use `extends` instead of `extend`](/proposals/p2760.md#use-extends-instead-of-extend)
    -   [List base class in class declaration](/proposals/p2760.md#list-base-class-in-class-declaration)

```

docs/design/generics/details.md
```diff
@@ -304,27 +304,27 @@ To implement more than one interface when defining a type, simply include an
...
   }
 }
 ```
 
 In this case, all the functions `Add`, `Scale`, and `Draw` end up a part of the
 API for `Point`. This means you can't implement two interfaces that have a name
-in common (unless you use an `external impl` for one or both, as described
-below).
+in common (unless you use an `impl` without `extend` for one or both, as
+described below).
```
jonmeow:
```suggestion
in common (unless you use an `impl` without `extend` for one or both, for
an [external impl](#external-impl)).
```

A quirk of this change is that it's hard to search on now, so given it's not the immediately following text, maybe a link to be explicit?

docs/design/generics/details.md
```diff
@@ -523,8 +523,8 @@ unlike Swift and Rust.
 Given a value of type `Point3` and an interface `Vector` implemented for that
 type, you can access the methods from that interface using a
 [qualified member access expression](terminology.md#qualified-member-access-expression)
-whether or not the implementation is done externally with an `external impl`
-declaration. The qualified member access expression writes the member's
+whether or not the implementation is done externally with an `impl` declaration
+without `extend`. The qualified member access expression writes the member's
```
jonmeow:
```suggestion
whether or not the implementation is done with an [external `impl` declaration](#external-impl). The qualified member access expression writes the member's
```

Just a suggestion -- the "with ... without" could be difficult to parse, and linking provides context.

docs/design/generics/details.md
```diff
@@ -3707,7 +3728,8 @@ class Vector(T:! type) {
 }
 ```
 
-This is equivalent to naming the type between `impl` and `as`:
+This is equivalent to naming the type between `impl` and `as`, though this form
```
jonmeow: This is referred to as the "existing type", line 386. "type to the left of `as`" on 3788.

I think it'd be good to refer to it using a consistent phrase, and maybe there should be a name for it, such as "`impl` receiver"? (e.g., "This is equivalent to specifying the `impl` receiver, though this form ...")

docs/design/generics/details.md
```diff
@@ -3763,8 +3784,8 @@ interface when its element type satisfies the same interface:
     if the element type is comparable.
 -   A container is copyable if its elements are.
 
-To do this with an [out-of-line `impl`](#external-impl), specify a more-specific
-`Self` type to the left of the `as` in the declaration:
+This may be done with an [external `impl`](#external-impl) by specifying a
+more-specific `Self` type to the left of the `as` in the declaration:
```
jonmeow:
```suggestion
more specific `Self` type to the left of the `as` in the declaration:
```

I don't think this should be hyphenated.

docs/design/generics/details.md
```diff
@@ -3763,8 +3784,8 @@ interface when its element type satisfies the same interface:
     if the element type is comparable.
 -   A container is copyable if its elements are.
 
-To do this with an [out-of-line `impl`](#external-impl), specify a more-specific
-`Self` type to the left of the `as` in the declaration:
+This may be done with an [external `impl`](#external-impl) by specifying a
```
jonmeow: In theory, an external `impl` could be inline, and that's strongly implied by saying "more specific `Self`" (which is invalid before `as` when out-of-line). However, the example is out-of-line. If inline external `impl` is supported for this, perhaps the example should be changed? Otherwise, I think the prior text for "out-of-line" may have been more precise.

(an inline definition might not work in the `forall` because the `Vector(T)` would be incomplete)

docs/design/generics/details.md
```diff
@@ -3785,24 +3806,58 @@ external impl forall [T:! Printable] Vector(T) as Printable {
...
+  extend impl forall [P:! Printable] Array(P, N) as Printable { ... }
 }
 ```
 
-All internal `impl` declarations in the body of a `class` definition must be for
-the class being defined. It is an error to declare `impl i32 as Printable`
-inside `class Array`.
+**Note:** This was changed in
+[proposal #2760](https://github.com/carbon-language/carbon-lang/pull/2760).
+
```
jonmeow, line 3822:
```suggestion
```

I assume this was temporary while #2760 was in flight, since references are generally at the bottom.

docs/design/generics/details.md
```diff
@@ -3785,24 +3806,58 @@ external impl forall [T:! Printable] Vector(T) as Printable {
...
-both `P` and `T` refer to the same type, but `P` has the type-of-type of
-`Printable` and so has a `Print` member. The relationship between `T` and `P` is
-as if there was a `where P == T` clause.
+It is still legal to declare or define an external impl lexically with the class
+scope, as in:
+
+```
+class Array(T:! type, template N:! i64) {
+  // ✅ Allowed: external impl defined in class scope may use `forall`
+  // and/or specify a type.
```
jonmeow:
```suggestion
  // or specify a type.
```

Style encourages rewriting to avoid `and/or`. In this case, `impl X forall Y as Z` is invalid, so maybe we can drop `and`? ("may specify a type or use `forall`" might read more easily, too)
https://developers.google.com/style/word-list#and-or

docs/design/generics/details.md
```diff
@@ -3785,24 +3806,58 @@ external impl forall [T:! Printable] Vector(T) as Printable {
...
+// as long as `Unprintable` doesn't implement `Printable`.
+no_print.Print();
+```
 
-It is legal to add the keyword `external` before the `impl` keyword to switch to
-an external impl defined lexically within the class scope. Inside the scope,
-both `P` and `T` refer to the same type, but `P` has the type-of-type of
-`Printable` and so has a `Print` member. The relationship between `T` and `P` is
-as if there was a `where P == T` clause.
+It is still legal to declare or define an external impl lexically with the class
```
jonmeow:
```suggestion
It is still legal to declare or define an external impl lexically inside the class
```

docs/design/generics/details.md
```diff
@@ -3814,55 +3869,39 @@ equal.
 ```
 interface Foo(T:! type) { ... }
 class Pair(T:! type, U:! type) { ... }
-external impl forall [T:! type] Pair(T, T) as Foo(T) { ... }
+impl forall [T:! type] Pair(T, T) as Foo(T) { ... }
 ```
 
-You may also define the `impl` inline, in which case it can be internal:
+You may also define the `impl` inline, but it may not be combined with the
+`extend` keyword:
```
jonmeow: Could 3847-3856 and 3875-3882 be unified, since they are both explaining that an `impl` can be inline without `extend`?

docs/design/generics/overview.md
```diff
@@ -205,28 +205,28 @@ class Song {
...
 ```
 
 Implementations may be defined within the class definition itself or
-out-of-line. Implementations may optionally start with the `external` keyword to
-say the members of the interface are not members of the class. Out-of-line
-implementations must be external. External implementations may be defined in the
-library defining either the class or the interface.
+out-of-line. Implementations may optionally start with the `extend` keyword to
+say the members of the interface are also members of the class, which may only
+be used in class scope. Otherwise, implementations may be defined in the library
```
jonmeow:
```suggestion
be used in a class scope. Otherwise, implementations may be defined in the library
```

jonmeow (commented):
I didn't see any issues with the changes to examples, but do have a few questions about textual changes.

docs/design/generics/details.md
josh11b: `impl X forall Y as Z` is never legal, but that is because the `forall Y` comes first (and should include `[`...`]`). `impl forall [Y] X as Z` is definitely a thing, as show by this specific example. I've updated the text to replace "and/or."

docs/design/generics/details.md
jonmeow: As I'm glancing at this again -- a later example clarifies it's intended to work, so maybe just adjust the example to show inline?

josh11b: - I made the change to make it more accurate, inline external `impl` is supported for this.
- I will delete the word `Self` since it seems to be causing some confusion. It is true that you can't put `Self` before `as`, but I was referring to it as the `Self` type since `Self` inside the `impl` definition body is set to that type.
- External impls can be declared inline, but they will most commonly be out-of-line, and the example reflected that. There is an example further down to show it can also be inline.

docs/design/generics/details.md
josh11b: I think this comment is useful here in context, rather than somewhere else.

docs/design/generics/details.md
josh11b: I've added the words "as before" to make it clear this is just another example of the same thing.

docs/design/generics/details.md
josh11b: According to https://doc.rust-lang.org/reference/items/implementations.html , this is called the "implementing type" in Rust.

jonmeow: I don't really have a strong opinion here if others find that logical, although I think "implementing type" is confusing for external impls, where I wouldn't think of the type that's targeted by the external impl as the type providing implementation.

This is pre-existing though, so I want to be clear I don't think it should block merging.

docs/design/generics/details.md
jonmeow: Just to note, it still feels redundant to me as both examples seem like they're demonstrating an absence of `extend`, but maybe others will find the repetition useful.

docs/design/generics/details.md
jonmeow: Thanks! I agree removing `Self` clarifies.

jonmeow (approved):
Looks good to me. I'm approving, but I'm hesitant to merge since I'm not sure whether you're expecting feedback from others.

josh11b:
> Looks good to me. I'm approving, but I'm hesitant to merge since I'm not sure whether you're expecting feedback from others.

There is a bigger rewrite coming, so I expect this text will be reviewed again. So I'll submit this soon without waiting. If someone else wants to give feedback, I'm happy to incorporate it in another PR.

PR #2867, opened 2023-06-01
Refactor lowering of functions to be more up-front in action, and handle parameter lowering.

jonmeow:
Previously, lowering occurred purely on a block basis. This switches that structure to be more function based. The overall intent is that lowering will do, in order:

1. Types
2. Function declarations
3. Global variable declarations (TODO: determine whether this includes initializers or if that should be step 5)
4. Function definitions

This relies on the particular ordering so that, for example, a global variable initializer can refer to a function declaration. From a structural perspective, it's pretty much the same to SemanticsIR if it provides access to the storage of these directly instead of via the SemanticsNode intermediary.

We will need to long-term work on the handling of generics and similar. However, for now the focus is for basic functions. The rename from "callable" to "function" reflects this: although a call expression could still do something like define a specific form of generic, the Callable object right now should be more specific to functions. Maybe I'll end up undoing part of this, but structurally we need more awareness of things which are particular to functions -- for example, whether there's a function definition, which would have a dramatically different from for generics. Philosophically I'm trying to keep the implementation easily refactored, but without driving too closely at future use-cases that may end up better suited towards a different approach.

This removes the FunctionDefinition SemanticsNode type because it no longer seems necessary to emit. Unlike the declaration, the definition isn't directly referenced by calling code, so I think it's cruft.


zygoloid (approved):

zygoloid:
Accidentally landed this as part of #2869 => just closing without merging.

Issue #1077, opened 2022-02-10
find a way to permit impls of CommonTypeWith where the LHS and RHS type overlap


zygoloid:
Per the [current design](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/expressions/if.md#same-type), `CommonTypeWith` has an `impl` of this form:

```
final impl [T:! Type] T as CommonTypeWith(T) where .Result = T {}
```

Given this `final impl`, it's unclear how to write a parameterized `impl` of `CommonTypeWith` for two parameterized types that overlap. For example:

```
// Error: would specialize the above `final impl`.
impl [U:! Type, T:! CommonTypeWith(U)] Vec(T) as CommonTypeWith(Vec(U)) where .Result = Vec(T.Result) {}
```

The error here is unnecessary: there's no actual coherence problem here because in the case where `T == U`, the computed `Result` is `Vec(T)`.

It's not clear how to resolve this. Some options:

* Switch from `final impl` to a [final associated constant](https://github.com/carbon-language/carbon-lang/blob/trunk/proposals/p0983.md#final-associated-constants-instead-of-final-impls).
* Provide some way for the author of this `impl` to say that it's prioritized below the `final impl`.
* Provide some way for the author of the `final impl` to say that it is considered more specialized than any other `impl`, rather than disallowing `impl`s that overlap with it.
* Add support for [type inequality constraints](https://github.com/carbon-language/carbon-lang/blob/trunk/proposals/p0818.md#type-inequality-constraints) so that the new `impl` could exclude the `T == U` case.
* Allow a `final impl` to be specialized if the specialization is identical.

josh11b:
As far as I am aware, we would still like an answer to this question. An answer was proposed in https://github.com/carbon-language/carbon-lang/issues/1077#issuecomment-1093606977 but never accepted.

chandlerc:
Yep, leads are happy with the resolution here: https://github.com/carbon-language/carbon-lang/issues/1077#issuecomment-1093606977

PR #2868, opened 2023-06-01
Allow overlap with a `final impl` if identical

josh11b:
Allow an `impl` to overlap with a `final impl` if they agree on the overlap. Agreement is defined as all values comparing equal, and functions never comparing equal.  Implements the decision in question-for-leads issue #1077.

### DIFF
--- a/docs/design/generics/details.md
+++ b/docs/design/generics/details.md
@@ -4264,20 +4264,20 @@ generic function call, such as using an operator:
 // Interface defining the behavior of the prefix-* operator
 interface Deref {
   let Result:! type;
-  fn DoDeref[self: Self]() -> Result;
+  fn Op[self: Self]() -> Result;
 }
 
 // Types implementing `Deref`
 class Ptr(T:! type) {
   ...
   external impl as Deref where .Result = T {
-    fn DoDeref[self: Self]() -> Result { ... }
+    fn Op[self: Self]() -> Result { ... }
   }
 }
 class Optional(T:! type) {
   ...
   external impl as Deref where .Result = T {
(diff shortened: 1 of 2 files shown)


docs/design/generics/details.md
```diff
@@ -4323,9 +4323,13 @@ class Optional(T:! type) {
 ```
 
 This prevents any higher-priority impl that overlaps a final impl from being
-defined. Further, if the Carbon compiler sees a matching final impl, it can
-assume it won't be specialized so it can use the assignments of the associated
-types in that impl definition.
+defined unless it agrees with the `final` impl on the overlap. Since we do not
```
zygoloid: Is this well-defined? Given two arbitrary `impl` declarations, can we always compute what the overlap is, and whether the associated constants have the same values in the overlapping region? (I think the answer might be yes if all parameters are generic and no if there are any template parameters, but I'm not sure.)

What do we want to do for overlap between a templated impl and a final impl / between a templated final impl and a generic impl? That case seems problematic both because there can be template constraints that make the overlap non-computable, and the values of associated constants might depend on the template parameters, which perhaps might make them not symbolically comparable. Do we delay the diagnostic until the template is instantiated?

zygoloid (commented):
Broadly this looks good, but I worry there are details here we need to dig into a little more.

docs/design/generics/details.md
josh11b: I believe the overlap between generic and concrete `impl` declarations can be computed using unification. I've added a sentence about templated impls being delayed.

zygoloid (approved):
Approving; this proposal adopts the decision in the question-for-leads issue #1077.

PR #2873, opened 2023-06-04
Fix test env setup in macOS

vmingchen:
Some tests failed to run in my macOS (13.3.1) setup:

```
$bazel test //...:all
...
//toolchain/source:source_buffer_test                                    PASSED in 0.3s
//explorer/lit_testdata:hello.carbon.test                                FAILED in 2.2s
  /private/var/tmp/_bazel_ming.chen/308b02a1da3e82e017b979084b8d8d91/execroot/carbon/bazel-out/darwin_arm64-fastbuild/testlogs/explorer/lit_testdata/hello.carbon.test/test.log
//explorer/lit_testdata:trace.carbon.test                                FAILED in 1.3s
//explorer/lit_testdata:zero.carbon.test                                 FAILED in 1.6s
//toolchain/driver/testdata:errors_sorted_test.carbon.test               FAILED in 2.2s
//toolchain/driver/testdata:errors_streamed_test.carbon.test             FAILED in 1.2s
//toolchain/driver/testdata:semantics_builtin_nodes.carbon.test          FAILED in 0.7s
...
```

The log of the failed tests shows that `bash` is not found in path:

```
(trunk)$ cat /private/var/tmp/_bazel_ming.chen/308b02a1da3e82e017b979084b8d8d91/execroot/carbon/bazel-out/darwin_arm64-fastbuild/testlogs/explorer/lit_testdata/hello.carbon.test/test.log
Executing tests from //explorer/lit_testdata:hello.carbon.test
env: bash: No such file or directory
...
```

Some digging revealed that we hide the PATH in `lit_test.py` and the PATH does not include `bash` (zsh has replaced bash as the default shell in macOS). For example, `bash` is not at a common path on my machine:

```
(fixtest)$ type bash
bash is /opt/homebrew/bin/bash
```

Trying to add bash in the tool list lead to more problems:

```
-    for tool in ("env", "grep", "python3", "sh", "which"):
+    for tool in ("env", "bash", "grep", "python3", "sh", "which"):
```

```
INFO: From Testing //explorer/lit_testdata:hello.carbon.test:
==================== Test output for //explorer/lit_testdata:hello.carbon.test:
/opt/homebrew/Cellar/pyenv/2.3.18/libexec/pyenv-which: line 61: basename: command not found
```

Further adding `basename` didn't work either. So it seems to me that the way we hide PATH is indeed fragile as the comment hinted.

Removing the PATH hiding enables all tests to pass. So this PR does so to fix the tests.

### DIFF
--- a/testing/lit_test/lit_test.py
+++ b/testing/lit_test/lit_test.py
@@ -9,7 +9,6 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 import argparse
 import os
 from pathlib import Path
-import shutil
 import subprocess
 
 
@@ -34,20 +33,9 @@ def main():
         "-v",
     ]
 
-    # Force tests to be explicit about command paths. Some tools are required by
-    # bazel's py_binary output in order to find python3, so we add these. The
-    # list of tools may prove to be fragile: if so, we may need to change or
-    # remove the PATH hiding.
-    system_tools = Path(os.environ["TEST_TMPDIR"]).joinpath("system_tools")
-    system_tools.mkdir()
-    for tool in ("env", "grep", "python3", "sh", "which"):
-        system_tools.joinpath(tool).symlink_to(shutil.which(tool))
-    env = os.environ.copy()
-    env["PATH"] = str(system_tools)
-
     # Run lit.
     try:
-        subprocess.check_call(args=args + parsed_args.lit_args, env=env)
+        subprocess.check_call(args=args + parsed_args.lit_args)
     except subprocess.CalledProcessError as e:
         # Print without the stack trace.
         exit(e)


jonmeow:
Even though it's not the default shell, on my mac, `bash` is at `/bin/bash`. Note the use of `which` is deliberate to handle different setups.

That said, the functionality you're proposing to delete is really a trade-off. On the one hand, it means having to be more careful about what commands are available; on the other, it does also make it more fragile. I believe both ends are noted in the comments on line 37-40.

The problem is that if the commands aren't limited, it can create functionality that's _also_ fragile by allowing executions to depend on environment settings that don't work on other machines (e.g., `not` in the path). Although it's really a bit of a lose-lose situation, I think that per the comments, we prefer to list the commands we're using. If you're seeing some missing, please feel free to add more, but we won't remove the validation.

jonmeow:
For reference, it looks like the `bash` dependency is in `lit` itself: https://github.com/llvm/llvm-project/blob/main/llvm/utils/lit/lit/LitConfig.py#L139

vmingchen:
> Even though it's not the default shell, on my mac, `bash` is at `/bin/bash`. Note the use of `which` is deliberate to handle different setups.
> 
> That said, the functionality you're proposing to delete is really a trade-off. On the one hand, it means having to be more careful about what commands are available; on the other, it does also make it more fragile. I believe both ends are noted in the comments on line 37-40.
> 
> The problem is that if the commands aren't limited, it can create functionality that's _also_ fragile by allowing executions to depend on environment settings that don't work on other machines (e.g., `not` in the path). Although it's really a bit of a lose-lose situation, I think that per the comments, we prefer to list the commands we're using. If you're seeing some missing, please feel free to add more, but we won't remove the validation.

Interesting. I have `/bin/bash` as well. I will see if this is something wrong with my setup later (probably over the weekend). Closing this for now. Thanks for the feedback!

chandlerc:
FWIW, I suspect this is actually the the root of the issue...

> For example, `bash` is not at a common path on my machine:
> 
> ```
> (fixtest)$ type bash
> bash is /opt/homebrew/bin/bash
> ```

I suspect none have been testing things with a homebrew installed bash. I'm a bit surprised it causes the failures you're seeing, but I'd guess that's the big difference between your mac setup and the ones we've tested and found working. Hope it helps with tracking down the problem.

vmingchen:
FTR, I re-ordered my PATH so that `/usr/bin` comes before `/opt/homebrew/bin`. After that, every test is passing. Thanks for the help!

PR #2875, opened 2023-06-05
Functions, function types, and function calls

zygoloid:
Specify the behavior of function calls and the type and behavior of the entity
introduced by a function declaration.

### DIFF
--- a/proposals/p2875.md
+++ b/proposals/p2875.md
@@ -1,0 +1,70 @@
+# Functions, function types, and function calls
+
+<!--
+Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+Exceptions. See /LICENSE for license information.
+SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+-->
+
+[Pull request](https://github.com/carbon-language/carbon-lang/pull/2875)
+
+<!-- toc -->
+
+## Table of contents
+
+-   [Abstract](#abstract)
+-   [Problem](#problem)
+-   [Background](#background)
+-   [Proposal](#proposal)
+-   [Details](#details)
+-   [Rationale](#rationale)
+-   [Alternatives considered](#alternatives-considered)
+
+<!-- tocstop -->
+
+## Abstract
+
+TODO: Describe, in a succinct paragraph, the gist of this document. This
+paragraph should be reproduced verbatim in the PR summary.
+
+## Problem
+
+TODO: What problem are you trying to solve? How important is that problem? Who
+is impacted by it?
+
+## Background
+
+TODO: Is there any background that readers should consider to fully understand
+this problem and your approach to solving it?
+
+## Proposal
+
+TODO: Briefly and at a high level, how do you propose to solve the problem? Why
+will that in fact solve it?
+
+## Details
+
+TODO: Fully explain the details of the proposed solution.
+
+## Rationale
+
+TODO: How does this proposal effectively advance Carbon's goals? Rather than
+re-stating the full motivation, this should connect that motivation back to
+Carbon's stated goals and principles. This may evolve during review. Use links
+to appropriate sections of [`/docs/project/goals.md`](/docs/project/goals.md),
+and/or to documents in [`/docs/project/principles`](/docs/project/principles).
+For example:
+
+-   [Community and culture](/docs/project/goals.md#community-and-culture)
+-   [Language tools and ecosystem](/docs/project/goals.md#language-tools-and-ecosystem)
+-   [Performance-critical software](/docs/project/goals.md#performance-critical-software)
+-   [Software and language evolution](/docs/project/goals.md#software-and-language-evolution)
+-   [Code that is easy to read, understand, and write](/docs/project/goals.md#code-that-is-easy-to-read-understand-and-write)
+-   [Practical safety and testing mechanisms](/docs/project/goals.md#practical-safety-and-testing-mechanisms)
+-   [Fast and scalable development](/docs/project/goals.md#fast-and-scalable-development)
+-   [Modern OS platforms, hardware architectures, and environments](/docs/project/goals.md#modern-os-platforms-hardware-architectures-and-environments)
+-   [Interoperability with and migration from existing C++ code](/docs/project/goals.md#interoperability-with-and-migration-from-existing-c-code)
+
+## Alternatives considered
+
+TODO: What alternative solutions have you considered?
--- a/proposals/p2875.md
+++ b/proposals/p2875.md
@@ -100,7 +100,7 @@ other than asking for the type of the function value.
 fn TypeOf[T:! type](x: T) -> type { return T; }
 
 // ✅ `F` is a first-class value with a first-class type.
-alias FType = TypeOf(F);
+let template FType:! type = TypeOf(F);
 var my_f: FType = F;
 ```
 
--- a/proposals/p2875.md
+++ b/proposals/p2875.md
@@ -190,7 +190,7 @@ checking proceeds as follows:
     -   If the parameter is a `template :!` binding, the argument expression is
         converted to have the same type as the binding and template constant
         expression phase.
-    -   If the parameter is a checked `:!` binding, the argument expression is
+    -   If the parameter is a symbolic `:!` binding, the argument expression is
         converted to have the same type as the binding and symbolic constant
         expression phase.
     -   Otherwise, the parameter is pattern-matched against the argument.
--- a/proposals/p2875.md
+++ b/proposals/p2875.md
@@ -170,10 +170,10 @@ permitted.
 
 ### Direct calls
 
-A call expression is a _direct call_ when the callee is:
+A call expression is a _direct call_ when the callee:
 
--    the name of a parameterized entity, like a generic class or interface,
--    has a function type or bound method type.
+-   is the name of a parameterized entity, like a generic class or interface, or
+-   has a function type or bound method type.
 
 In a direct call, a call signature is available which is used to check the given
 arguments against the callee's declared implicit and explicit parameters. This
--- a/proposals/p2875.md
+++ b/proposals/p2875.md
@@ -242,6 +242,13 @@ A call expression that is not a direct call is translated into an invocation of
   auto ord: auto = cmp.(Call((T, T)).Op)((*v)[i], (*v)[j]);
 ```
 
+Note that the types of the call arguments are modeled as an interface
+parameter. This permits the call interface to model function overloading.
+However, the return type is an associated type, not a parameter -- we do not
+permit overloading on return types, and we don't want type information to
+propagate from the context in which a call expression appears inwards to the
+call.
+
 ### Function types and `Call`
 
 A function type or bound method type implements the `Call` interface for every
@@ -483,6 +490,24 @@ match_first {
 
 For a type that is `Unsigned & Floating`, the former `impl` will be selected,
 matching the behavior of overload selection.
+
+### Overloading on expression category
+
+We should consider whether calls can be overloaded on the expression category
+of the callee. Rust provides separate `Fn`, `FnMut`, and `FnOnce` to
+distinguish between different ways of passing the callable object into an
+overloaded function call, and C++ permits `operator()` to take `*this` as
+`const`, non-`const`, or even `&&`.
+
+This would most likely take the form of multiple `Call` interfaces to handle
+different kinds of `self` parameter, modeled in a similar way to the
+`IndexWith` and `IndirectIndexWith` interfaces used for subscripting.
+
+### Variadics
+
+Once Carbon supports variadics, the overloaded call mechanism should be revised
+to make use of them, instead of using a tuple type to approximate a variadic
+parameter list.
 
 ### Lambdas
 
--- a/proposals/p2875.md
+++ b/proposals/p2875.md
@@ -437,8 +437,8 @@ we would synthesize implementations that behave as follows:
 
 ```
 match_first {
-  impl forall [T:! Unsigned] Abs as Call(T) where .Result = T { ... }
-  impl forall [T:! Floating] Abs as Call(T) where .Result = T { ... }
+  impl forall [T:! Unsigned] Abs as Call((T,)) where .Result = T { ... }
+  impl forall [T:! Floating] Abs as Call((T,)) where .Result = T { ... }
 }
 ```
 
--- a/proposals/p2875.md
+++ b/proposals/p2875.md
@@ -30,6 +30,8 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
     -   [Make direct and indirect calls behave uniformly](#make-direct-and-indirect-calls-behave-uniformly)
 -   [Future work](#future-work)
     -   [Overloading](#overloading)
+    -   [Overloading on expression category](#overloading-on-expression-category)
+    -   [Variadics](#variadics)
     -   [Lambdas](#lambdas)
     -   [Function pointers](#function-pointers)
 
@@ -76,8 +78,8 @@ a syntax resembling a tuple of arguments. There are several kinds of callable:
     [overload function call syntax](#overloaded-call-operator).
 
 In the first four cases, argument deduction is used to determine the values of
-deduced arguments, and then pattern matching is used to bind parameter patterns to
-argument values.
+deduced arguments, and then pattern matching is used to bind parameter patterns
+to argument values.
 
 In the remaining cases, the built-in `Call` interface is used to determine the
 semantics of the call, and the call `F(args)` is translated into
@@ -180,8 +182,8 @@ arguments against the callee's declared implicit and explicit parameters. This
 checking proceeds as follows:
 
 -   Argument deduction is performed by comparing the declared parameter types
-    against the actual argument types and deducing values for implicit
-    arguments that make the types equal.
+    against the actual argument types and deducing values for implicit arguments
+    that make the types equal.
 -   Then, for each binding in the explicit parameter list in turn, all argument
     values that have been deduced are substituted into the parameter.
 
@@ -242,12 +244,11 @@ A call expression that is not a direct call is translated into an invocation of
   auto ord: auto = cmp.(Call((T, T)).Op)((*v)[i], (*v)[j]);
 ```
 
-Note that the types of the call arguments are modeled as an interface
-parameter. This permits the call interface to model function overloading.
-However, the return type is an associated type, not a parameter -- we do not
-permit overloading on return types, and we don't want type information to
-propagate from the context in which a call expression appears inwards to the
-call.
+Note that the types of the call arguments are modeled as an interface parameter.
+This permits the call interface to model function overloading. However, the
+return type is an associated type, not a parameter -- we do not permit
+overloading on return types, and we don't want type information to propagate
+from the context in which a call expression appears inwards to the call.
 
 ### Function types and `Call`
 
@@ -493,15 +494,15 @@ matching the behavior of overload selection.
 
 ### Overloading on expression category
 
-We should consider whether calls can be overloaded on the expression category
-of the callee. Rust provides separate `Fn`, `FnMut`, and `FnOnce` to
-distinguish between different ways of passing the callable object into an
-overloaded function call, and C++ permits `operator()` to take `*this` as
-`const`, non-`const`, or even `&&`.
+We should consider whether calls can be overloaded on the expression category of
+the callee. Rust provides separate `Fn`, `FnMut`, and `FnOnce` to distinguish
+between different ways of passing the callable object into an overloaded
+function call, and C++ permits `operator()` to take `*this` as `const`,
+non-`const`, or even `&&`.
 
 This would most likely take the form of multiple `Call` interfaces to handle
-different kinds of `self` parameter, modeled in a similar way to the
-`IndexWith` and `IndirectIndexWith` interfaces used for subscripting.
+different kinds of `self` parameter, modeled in a similar way to the `IndexWith`
+and `IndirectIndexWith` interfaces used for subscripting.
 
 ### Variadics
 
--- a/proposals/p2875.md
+++ b/proposals/p2875.md
@@ -180,7 +180,7 @@ arguments against the callee's declared implicit and explicit parameters. This
 checking proceeds as follows:
 
 -   Argument deduction is performed by comparing the declared parameter types
-    against the actual parameter types and deducing values for implicit
+    against the actual argument types and deducing values for implicit
     arguments that make the types equal.
 -   Then, for each binding in the explicit parameter list in turn, all argument
     values that have been deduced are substituted into the parameter.
--- a/proposals/p2875.md
+++ b/proposals/p2875.md
@@ -18,6 +18,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 -   [Proposal](#proposal)
 -   [Details](#details)
     -   [Functions and function types](#functions-and-function-types)
+    -   [Bound methods](#bound-methods)
     -   [Call syntax](#call-syntax)
     -   [Direct calls](#direct-calls)
     -   [Generic callable parameters](#generic-callable-parameters)
@@ -56,19 +57,23 @@ work in C or C++.
 
 ## Proposal
 
-A function declaration in Carbon introduces a new, unique type, called a
-_function type_. The function name is a stateless value of that function type.
+A function declaration in Carbon introduces a new, unique, stateless type,
+called a _function type_. The function name is bound to a value of that function
+type.
 
 Function calls use C-like syntax: an expression naming a callable is followed by
 a syntax resembling a tuple of arguments. There are several kinds of callable:
 
--   Functions, and more generally values of function types.
--   Bound member functions, such as `my_vector.Begin`.
+-   Functions, and more generally
+    [values of function types](#functions-and-function-types).
+-   [Bound methods](#bound-methods), such as `my_vector.Begin`.
 -   Lambdas, once they are introduced to Carbon.
 -   Parameterized entities, such as a generic class `Vector` or a generic
     interface `AddWith`.
--   Values of dependent types that are constrained to be callable.
--   User-defined class types that overload function call syntax.
+-   Values of dependent types that are
+    [constrained to be callable](#generic-callable-parameters).
+-   User-defined class types that
+    [overload function call syntax](#overloaded-call-operator).
 
 In the first four cases, argument deduction is used to determine the values of
 deduced arguments, and then pattern matching is used to bind parameter values to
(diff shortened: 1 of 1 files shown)


proposals/p2875.md
```diff
@@ -0,0 +1,467 @@
...
+```
+
+For the purpose of the orphan rule, a function type is considered to be declared
+by the function declaration that introduces the function value.
+
+### Call syntax
+
+Calls take the form `a(b, c, d)`. This is syntactically equivalent to a primary
+expression followed by a tuple literal, except that no trailing comma is
+required for the case of a single argument, `a(b)`.
```
jonmeow:
```suggestion
expression followed by a tuple literal, except that the trailing comma is
optional for the case of a single argument, `a(b)`.
```

I'm pretty sure this is the intent? Because what you're saying is "no trailing comma is required", which could be read as saying `a(b,)` is invalid because it has a trailing comma, while the lack thereof is required. However, explorer currently allows the optional trailing comma.

proposals/p2875.md
```diff
@@ -0,0 +1,467 @@
...
+    -   If the parameter is a `template :!` binding, the argument expression is
+        converted to a constant value phase expression of the type of the
+        binding.
+    -   If the parameter is a non-`template` `:!` binding, the argument
+        expression is converted to a symbolic value phase expression of the type
+        of the binding.
+    -   Otherwise, the parameter is pattern-matched against the argument.
+-   If a parameter is a `:!` binding, its corresponding converted argument
+    expression is evaluated, and its value is added to the list of deduced
+    argument values before any later parameters are processed.
```
jonmeow, line 200: My impression is that this actually happens as part of what's stated in 146 because that says "for each binding". Is that accurate? It's just a little confusing because that bullet seems to be saying it does something for each argument, then later you have "but this happens in the middle" without explicitly saying that it's actually during the previous bullet. Maybe it'd help to describe this more as a flow?

- For each binding:
  - Substitute:
    - the three cases
  - If a parameter is a `:!`

proposals/p2875.md
```diff
@@ -0,0 +1,467 @@
...
+because it is the name of a parameterized entity like a generic class or
+interface, or because it has a function type, the call is a _direct call_.
+
+In a direct call, a call signature is available which is used to check the given
+arguments against the callee's declared implicit and explicit parameters. This
+checking proceeds as follows:
+
+-   Argument deduction is performed by comparing the declared parameter types
+    against the actual parameter types and deducing values for implicit
+    arguments that make the types equal.
```
jonmeow: Isn't the deduction of `self` a little different, since the object isn't an actual parameter type? Or is that just not covered here since classes defines behavior?

proposals/p2875.md
```diff
@@ -0,0 +1,467 @@
...
+-   Then, for each binding in the explicit parameter list in turn, all argument
+    values that have been deduced are substituted into the parameter.
+    -   If the parameter is a `template :!` binding, the argument expression is
+        converted to a constant value phase expression of the type of the
+        binding.
+    -   If the parameter is a non-`template` `:!` binding, the argument
```
jonmeow:
```suggestion
    -   If the parameter is a `:!` binding that is not a `template`, the argument
```

I think this is saying the same as you meant, just suggesting the swap so that the difference of a space not being in backticks isn't significant. This seems particularly subtle for html rendering, but I'm particularly suggesting it because I want to confirm I'm interpreting correctly.

jonmeow (commented):
I'm trying to think about this, but I'll particularly note that comparing:

```
void Foo(std::function<void()> x)
fn Foo[F:! Call((,))] (x: F)
```

It does feel like separating the type and the variable name reduces readability.

proposals/p2875.md
zygoloid: Yeah, I suppose using "no X is required" to mean "X is not required" rather than "X is required to be absent" may be hard for non-native readers to parse. Though I'd prefer to keep "required" here because I think that better contrasts with the rule for tuple literals:

```suggestion
expression followed by a tuple literal, except that a trailing comma is
not required for the case of a single argument, `a(b)`.
```

proposals/p2875.md
zygoloid: The way I'm viewing this is that `self` is bound when forming the bound member function: in
```
class A { fn F[self: Self, T:! type](x: T); }
fn DoThing(a: A) { a.F(0); }
```
... `a.F` already has a value for `self`, leaving behind a function with signature `[T:! type](x: T)`, which is informally what's [described in the design for member access](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/expressions/member_access.md#instance-binding).

proposals/p2875.md
zygoloid: Done, with a different phrasing (I didn't like "that is not a `template`" for bindings because I don't think we want to say that a template parameter is a template).

proposals/p2875.md
zygoloid: Hopefully addressed by removing the bullet symbol from this paragraph. Is this clear enough that the "for each" scopes over the whole bullet, or would it still be better to explicitly indent this within the "for each binding"?

proposals/p2875.md
```diff
@@ -0,0 +1,469 @@
...
+```
+
+For the purpose of the orphan rule, a function type is considered to be declared
+by the function declaration that introduces the function value.
+
+### Call syntax
+
+Calls take the form `a(b, c, d)`. This is syntactically equivalent to a primary
+expression followed by a tuple literal, except that a trailing comma is not
+required for the case of a single argument, `a(b)`.
```
jonmeow:
```suggestion
required for the case of a single argument, either `a(b)` or `a(b,)`.
```

Just looking for a different way to emphasize the meaning here (to your point about tuple syntax, most of my concern is that this'll be copied into the design, and the direct comparison won't be obvious to a new reader).

proposals/p2875.md
jonmeow: This looks good, thanks!

jonmeow (commented):
Just resolving through my prior comments. I'm still hoping to take a more complete look, but I think I'm basically good with this.

jonmeow (commented):
Just resolving through my prior comments. I'm still hoping to take a more complete look, but I think I'm basically good with this.

proposals/p2875.md
zygoloid: I've taken another pass over this section; I think this new approach sufficiently emphasizes what I'm going for here and should still be clear when copied into the design.

proposals/p2875.md
```diff
@@ -0,0 +1,475 @@
...
+implicitly provides an `impl` of `Call` matching the function signature.
+
+```
+fn Select[T:! type](b: bool, x: T, y: T) -> T {
+  return if b then x else y;
+}
+
+// Generated:
+impl forall [T:! type, BType:! ImplicitAs(bool)]
+    Select as (BType, T, T) where .Result = T {
```
jonmeow:
```suggestion
    Select as Call((BType, T, T)) where .Result = T {
```

Shouldn't the interface name be here?

proposals/p2875.md
```diff
@@ -0,0 +1,475 @@
...
+overloaded fn Abs[T:! Unsigned](x: T) -> T { return x; }
+overloaded fn Abs[T:! Floating](x: T) -> T { return x < 0 ? -x : x }
+```
+
+we would synthesize implementations that behave as follows:
+
+```
+match_first {
+  impl forall [T:! Unsigned] Abs as Call(T) where .Result = T { ... }
+  impl forall [T:! Floating] Abs as Call(T) where .Result = T { ... }
```
jonmeow:
```suggestion
  impl forall [T:! Unsigned] Abs as Call((T,)) where .Result = T { ... }
  impl forall [T:! Floating] Abs as Call((T,)) where .Result = T { ... }
```

In the current design you've chosen to use tuples instead of variadics, so this should be passing a one-length tuple, right?

proposals/p2875.md
```diff
@@ -0,0 +1,475 @@
...
+  // ✅ `my_f` has function type `FType`. This is a direct call to `F`.
+  return my_f(1);
+}
+
+fn Sort[template F:! type, T:! type](v: Vector(T)*, f: F);
+fn Compare(a: i32, b: i32) -> Ordering { return a.(Ordered.Compare)(b); }
+fn SortInts(v: Vector(i32)*) { Sort(v, Compare); }
+```
+
+For the purpose of the orphan rule, a function type is considered to be declared
```
geoffromer:
```suggestion
For the purpose of the [orphan rule](/docs/design/generics/terminology.md), a function type is considered to be declared
```

proposals/p2875.md
```diff
@@ -0,0 +1,475 @@
...
+    trailing comma is permitted but not required after the final argument.
+
+Call syntax is syntactically equivalent to a primary expression followed by a
+tuple literal, except that a tuple literal requires a trailing comma to form a
+single-element tuple `(b,)`, whereas in call syntax both `a(b)` and `a(b,)` are
+permitted.
+
+### Direct calls
+
+When the callee in a call expression can be resolved to a declaration, either
```
geoffromer: According to the examples below, `cmp((*v)[i], (*v)[j])` isn't a direct call, but `cmp` can be resolved to a declaration. Should this say "function or parameterized entity declaration"?

proposals/p2875.md
```diff
@@ -0,0 +1,475 @@
...
+  // In Sort...
+  auto ord: auto = cmp((*v)[i], (*v)[j]);
+  // ... is translated into ...
+  auto ord: auto = cmp.(Call((T, T)).Op)((*v)[i], (*v)[j]);
+```
+
+### Function types and `Call`
+
+A `fn` declaration that declares a function with a runtime signature (see below)
+implicitly provides an `impl` of `Call` matching the function signature.
```
geoffromer:
```suggestion
A `fn` declaration implicitly provides an `impl` of `Call` matching the function signature
(unless the function it declares doesn't have a runtime signature; see below).
```

(Rephrasing to make it clearer that you don't need to understand "runtime signature" on a first read)

proposals/p2875.md
```diff
@@ -0,0 +1,475 @@
...
+  // ❌ Can't call `CompileTime` this way, it doesn't have a runtime signature.
+  CompileTime.(Call(type, i32).Op)(i32, 0);
+  // ❌ Can't call `RefutablePattern` this way, it can't implement `Call(i32)`
+  // for arbitrary i32 arguments.
+  RefutablePattern.(Call(i32).Op)(0);
+}
+```
+
+> **Note:** There is no support for passing compile-time parameters to a checked
+> generic function parameter.
```
geoffromer: Does this limitation apply only to indirect calls, or is even something like this invalid?

```
fn f(T:! type);

fn g(T:! type) {
  f(T);
}
```

proposals/p2875.md
```diff
@@ -0,0 +1,475 @@
...
+
+## Background
+
+This proposal assumes the reader understands how functions and function pointers
+work in C or C++.
+
+## Proposal
+
+A function declaration in Carbon introduces a new, unique type, called a
+_function type_. The function name is a stateless value of that function type.
```
geoffromer: This was confusing to me because I think of names as _referring to_ values, rather than _being_ values.

```suggestion
A function declaration in Carbon introduces a new, unique type, called a
_function type_. Uses of the function name refer to a stateless value of that function type.
```

geoffromer: In some ways this compromises our "types are values" philosophy, because we're introducing both a type and a singular value of that type, in order to represent the same information in the type and value domains. I don't see how we can avoid that, though, so this is just an observation; no action is needed.

josh11b: @mconst Happening again. Note the argument for this approach in the "alternatives considered" below.

proposals/p2875.md
```diff
@@ -0,0 +1,475 @@
...
+functions have distinct function types, even if they have the same signature. A
+function type is an empty, trivial type. There is no way to name a function type
+other than asking for the type of the function value.
+
+```
+// Compile-time function.
+fn TypeOf[T:! type](x: T) -> type { return T; }
+
+// ✅ `F` is a first-class value with a first-class type.
+alias FType = TypeOf(F);
```
josh11b: I think you can only have names on the right side of an alias, not an expression like this.
```suggestion
let template FType:! type = TypeOf(F);
```

proposals/p2875.md
josh11b: I'd link to either https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/generics/details.md#orphan-rule or https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/generics/terminology.md#coherence

proposals/p2875.md
```diff
@@ -0,0 +1,475 @@
...
+The result of the call expression depends on the callee:
+
+-   If the callee is a parameterized entity such as a generic class or a generic
+    interface, the result is the specific instance of that generic, such as a
+    class or interface, and the call is a value expression of type `type`.
+-   If the callee is a function value, the call is an initializing expression
+    whose type is the substituted return type of the function. When evaluated,
+    the call expression will invoke the function and produce whatever value it
+    returns.
+
```
josh11b, line 214: Do we want to say anything about functions called from a type position being evaluated at compile time, like `TypeOf` above?

proposals/p2875.md
```diff
@@ -0,0 +1,475 @@
...
+class Func(Arg:! type) {
+  impl as Call((Arg,)) where .Result = () {
+    fn Op[self: Self](arg: (Arg,)) { Print("hello, world"); }
+  }
+}
+
+fn Run() {
+  let f: Func(i32) = {};
+  // ✅ Prints "hello, world".
+  f();
```
josh11b: Shouldn't this take an `i32` argument?

proposals/p2875.md
```diff
@@ -0,0 +1,475 @@
...
+```
+class X { var n: i32; }
+// ✅ OK, but inadvisable.
+impl {.a: X} as Call(()) where .Result = i32 {
+  fn Op[self: Self](args: ()) -> i32 {
+    return self.a.n;
+  }
+}
+fn Run() -> i32 {
+  return {.a = {.n = 1} as X}();
```
josh11b, line 362:
```suggestion
  // Returns 1.
  return {.a = {.n = 1} as X}();
```

proposals/p2875.md
```diff
@@ -0,0 +1,475 @@
...
+stateless.
+
+### Function pointers
+
+For interoperability with C++, and for purposes of cheaply storing references to
+stateless functions, function pointers should be supported. A function pointer
+could be modeled as:
+
+```
+alias FunctionPtr(Args:! type, Result:! type) =
```
josh11b, line 548: I don't think `alias` can do this.

proposals/p2875.md
geoffromer: Oh, yeah, I meant to include `#coherence` in the suggestion.

proposals/p2875.md
zygoloid: We don't have a design for `alias` yet, so I don't think we know that one way or the other :) If `alias` can't do this  then we're missing functionality compared to C++ alias templates.

In any case, this is "future work".

proposals/p2875.md
zygoloid: We definitely want to allow `Vector(i32)` on the right hand side of an alias, so it's not just names. But without an actual design for aliases, it's hard to say -- and I like your suggestion for that reason :)

proposals/p2875.md
zygoloid: Rephrased.

proposals/p2875.md
zygoloid: Done.

zygoloid: I think any time a type is stateless, we're in some sense representing the same information in the type and value domains, simply because by definition there is no information in the value domain (and we view the type as being part of the value). But note that a bound member function is not stateless -- `x.F` holds `x` in the value domain -- and the same will be true for lambda expressions, so I think in this case function types being stateless is just a degenerate case of a more general pattern of potentially-stateful function-like objects.

proposals/p2875.md
zygoloid: Done.

proposals/p2875.md
zygoloid: Yes, but not here; I'm writing a separate proposal for that :)

proposals/p2875.md
zygoloid: Rephrased substantially, and removed the term "runtime signature" entirely.

proposals/p2875.md
zygoloid: What I meant was, there is no support for passing a compile-time argument to a compile-time parameter of a value whose type is a checked generic type parameter that is constrained to be `Call`able.

```
fn f(T:! type);
// Error, but would be OK with `template C:! type`.
fn g[C:! Call(type)](c:! C) { c(i32); }
```

But I think saying that is redundant and confusing, so I've just removed this entirely.

proposals/p2875.md
zygoloid: Done.

proposals/p2875.md
```diff
@@ -138,9 +170,9 @@ permitted.
 
 ### Direct calls
 
-When the callee in a call expression can be resolved to a declaration, either
-because it is the name of a parameterized entity like a generic class or
-interface, or because it has a function type, the call is a _direct call_.
+When the callee in a call expression is either the name of a parameterized
+entity like a generic class or interface or has a function type or bound method
+type, the call is a _direct call_.
```
josh11b: I found this a bit hard to parse since it has the word "or" three times, with different precedences. I think you mean:

> When the callee in a call expression is either (the name of a parameterized entity like a generic (class or interface)) or (has a (function type or bound method type)), the call is a _direct call_.

I think this may just be a bit too complicated of a sentence structure. Maybe use bullets to make this clearer?

```suggestion
A call expression is a _direct call_ when the callee is:

-    the name of a parameterized entity, like a generic class or interface,
-    has a function type, or
-    has a bound method type.
```

or even

```suggestion
A call expression is a _direct call_ when the callee is:

-    the name of a parameterized entity, like a generic class or interface,
-    has a function type or bound method type.
```

josh11b (approved):

proposals/p2875.md
```diff
@@ -0,0 +1,514 @@
...
+tuple literal, except that a tuple literal requires a trailing comma to form a
+single-element tuple `(b,)`, whereas in call syntax both `a(b)` and `a(b,)` are
+permitted.
+
+### Direct calls
+
+A call expression is a _direct call_ when the callee is:
+
+-    the name of a parameterized entity, like a generic class or interface,
+-    has a function type or bound method type.
```
zygoloid:
```suggestion
A call expression is a _direct call_ when the callee:

-   is the name of a parameterized entity, like a generic class or interface, or
-   has a function type or bound method type.
```

josh11b (approved):

proposals/p2875.md
```diff
@@ -0,0 +1,514 @@
...
+
+A generic parameter can be constrained to be a callable type using the `Call`
+interface:
+
+```
+interface Call(Args: type) {
+  let Result:! type;
+  fn Op[self: Self](args: Args) -> Result;
+}
+```
```
josh11b, line 225: I feel like we need 2-3 variations of this interface with different `self` parameters, with an `extend` relationship between them.

proposals/p2875.md
```diff
@@ -0,0 +1,514 @@
...
+-   Parameterized entities, such as a generic class `Vector` or a generic
+    interface `AddWith`.
+-   Values of dependent types that are
+    [constrained to be callable](#generic-callable-parameters).
+-   User-defined class types that
+    [overload function call syntax](#overloaded-call-operator).
+
+In the first four cases, argument deduction is used to determine the values of
+deduced arguments, and then pattern matching is used to bind parameter values to
+arguments.
```
chandlerc: Hmm, I would have expected this to read:

```suggestion
deduced arguments, and then pattern matching is used to bind parameter names to
argument values.
```

WDYT? Or am I missing something important about the prior phrasing?

One possible tweak is that not all parameters are named, so maybe:
```suggestion
deduced arguments, and then pattern matching is used to bind parameters to
argument values.
```

Or:
```suggestion
deduced arguments, and then pattern matching is used to bind parameter patterns to
argument values.
```

proposals/p2875.md
```diff
@@ -0,0 +1,514 @@
...
+
+-   is the name of a parameterized entity, like a generic class or interface, or
+-   has a function type or bound method type.
+
+In a direct call, a call signature is available which is used to check the given
+arguments against the callee's declared implicit and explicit parameters. This
+checking proceeds as follows:
+
+-   Argument deduction is performed by comparing the declared parameter types
+    against the actual parameter types and deducing values for implicit
```
chandlerc: Should this be:

```suggestion
    against the actual argument types and deducing values for implicit
```

Not sure though...

proposals/p2875.md
```diff
@@ -0,0 +1,514 @@
...
+    the `self` value in the bound method value.
+
+### Generic callable parameters
+
+A generic parameter can be constrained to be a callable type using the `Call`
+interface:
+
+```
+interface Call(Args: type) {
+  let Result:! type;
```
chandlerc, line 222: Why are the argument types parameters but the result type an associated type? I can imagine there being good reasons for this, but would be helpful to articulate them here I think.

proposals/p2875.md
```diff
@@ -0,0 +1,514 @@
...
+
+### Signature-based function types
+
+We could give each function signature a distinct type, as is done in C and C++.
+This would provide a story for [function pointers](#function-pointers) that does
+not rely on generics or type erasure or fancy representation optimizations.
+
+The main down side of this approach is one we see frequently in C++: passing a
+function to a template generic would be less efficient than passing a function
+object to the same template. For example, in C++, given
```
chandlerc, line 424: If this were solved in some other way, would that change the direction you'd want here?

I see other ways of avoiding this specific issue, so I'm curious to know how important this point of rationale ended up being.

zygoloid: I think the point mentioned at the end of this section would still be relevant: giving functions unique types means that functions, lambdas, and function-like objects are more similar and have more uniform behavior. If the approach we took to avoid this specific issue affected the other cases in the same way, I think that would fully address the concern.

proposals/p2875.md
zygoloid: Yeah, I think that's a good point, that we shouldn't forget about. I've added a section to "Future work" on this; I think we can decouple that from this proposal and add the extra interfaces separately.

proposals/p2875.md
zygoloid: Added an explanation.

josh11b (approved):

proposals/p2875.md
```diff
@@ -0,0 +1,540 @@
...
+match_first {
+  impl forall [T:! Unsigned] Abs as Call((T,)) where .Result = T { ... }
+  impl forall [T:! Floating] Abs as Call((T,)) where .Result = T { ... }
+}
+```
+
+For a type that is `Unsigned & Floating`, the former `impl` will be selected,
+matching the behavior of overload selection.
+
+### Overloading on expression category
```
josh11b: It would be good to add that we want to, as future work, possibly overload on expression category for other parameters than the receiver.

It would also be good to have a future work section saying we will consider overloading on expression phase.

zygoloid: Done.

proposals/p2875.md
```diff
@@ -0,0 +1,559 @@
...
+
+-   Argument deduction is performed by comparing the declared parameter types
+    against the actual argument types and deducing values for implicit arguments
+    that make the types equal.
+-   Then, for each binding in the explicit parameter list in turn, all argument
+    values that have been deduced are substituted into the parameter.
+
+    -   If the parameter is a `template :!` binding, the argument expression is
+        converted to a constant value phase expression of the type of the
+        binding.
```
chandlerc: I think missing a "template" here, and some expression phrase adjustments?

```suggestion
    -   If the parameter is a `template :!` binding, the argument expression is
        converted to a template constant expression phase of the type of the
        binding.
```

proposals/p2875.md
```diff
@@ -0,0 +1,559 @@
...
+    -   If the parameter is a checked `:!` binding, the argument expression is
+        converted to a symbolic value phase expression of the type of the
+        binding.
```
chandlerc: And here missing "constant", and maybe restructuring a bit to orient around expression phases?

I'm also happy with both omitting "constant" if you'd prefer, but good to be consistent with the recent terminology choices.

```suggestion
    -   If the parameter is a checked `:!` binding, the argument expression is
        converted to a symbolic constant expression phase of the type of the
        binding.
```

proposals/p2875.md
```diff
@@ -0,0 +1,559 @@
...
+A more general, ambitious and ergonomic approach that we have started exploring
+would be to allow the function call signature to be described as part of the
+call interface:
+
+```carbon
+impl MyCallable as call(T:! type, x: T) -> T;
+```
+
+This might be a shorthand syntax for some way of expressing the signature as an
+interface, or a new form of first-class language primitive.
```
chandlerc, line 524: Is there a good place to link to the issue where we're tracking the longer-term effort here?

chandlerc (commented):
Picking this back up with the idea that it is sufficiently useful incremental progress towards the more general area around overloading. Some fairly minor comments inline.

proposals/p2875.md
zygoloid: Done.

proposals/p2875.md
zygoloid: Done.

proposals/p2875.md
zygoloid: Done.

proposals/p2875.md
```diff
@@ -523,6 +523,9 @@ impl MyCallable as call(T:! type, x: T) -> T;
 This might be a shorthand syntax for some way of expressing the signature as an
 interface, or a new form of first-class language primitive.
 
+The options in this space are currently being explored in issue
+[#3154](https://github.com/carbon-language/carbon-lang/issues/3154)
```
josh11b:
```suggestion
[#3154](https://github.com/carbon-language/carbon-lang/issues/3154).
```

proposals/p2875.md
```diff
@@ -188,11 +188,11 @@ checking proceeds as follows:
     values that have been deduced are substituted into the parameter.
 
     -   If the parameter is a `template :!` binding, the argument expression is
-        converted to a constant value phase expression of the type of the
-        binding.
+        converted to have the same type as the binding and template constant
+        expression phase.
     -   If the parameter is a checked `:!` binding, the argument expression is
```
josh11b:
```suggestion
    -   If the parameter is a symbolic `:!` binding, the argument expression is
```

josh11b (approved):

chandlerc (approved):
LG and approving for leads -- nothing controversial here given the goal of addressing the remaining gaps in future work.

PR #2876, opened 2023-06-05
Add support for splitting a test file.

jonmeow:
I'm looking at this as I start thinking about handling `import`. Syntax is based on llvm's `split-file` tool.

The `std::vector` -> `llvm::SmallVector` switch is minor, I'm doing it here because I had to touch everything anyways and I think for tests I'll lean slightly more towards the toolchain's way of doing things versus explorer's.


testing/file_test/file_test_base.cpp
```diff
@@ -71,15 +68,86 @@ auto FileTestBase::TestBody() -> void {
...
+    llvm::SmallVector<testing::Matcher<std::string>>& expected_stderr) -> void {
   std::ifstream file_content(path());
+  std::ofstream split_file;
+  bool found_content_pre_split = false;
   int line_index = 0;
   std::string line_str;
   while (std::getline(file_content, line_str)) {
-    ++line_index;
     llvm::StringRef line = line_str;
+    if (line.consume_front("// ---")) {
```
chandlerc, line 129: WDYT about: `// FILE: a.carbon` as the syntax?

Mostly that seems like it might match the other special comment syntaxes and be fairly unsurprising.

chandlerc (approved):
Neat!

Mostly FYI -- LLVM and Clang use a different approach in their test suites. Instead of splitting files, they have subdirectories containing non-test inputs. And the test just use relative paths to access them.

That has some advantages:
- It's fairly easy to manually reproduce a test run, the files are laid out in the tree the way they're processed.
- You can easily reuse inputs across multiple tests.

But it also has disadvantages:
- Need to track which inputs are used by which test. In some of my implementations of `lit` with Bazel, this is done in a *very* wasteful way: all non-test files in a subdirectory of the test suite are inputs to all tests in that subdirectory. =[
- Much less easy to see the whole test in a single location

Overall, I like the splitting model as a way to have very focused tests between a set of files, so happy with this direction.

I do wonder if eventually we'll want a way to provide some shared input files to be available. But that seems fine to do if and when we need to.

Other future ideas (nothing urgent or short term):
- Eventually we'll want a virtual filesystem abstraction for the toolchain. When that exists, it'd be great to leverage it rather than all the filesystem traffic here.
- I think, especially once we have a VFS, we should directly build support into the lexer for doing this. Clang (and GCC) somewhat have this for preprocessing because you can simulate the *result* of `#include` with the internal interface of [line directives](https://gcc.gnu.org/onlinedocs/cpp/Line-Control.html). Because of this, it was easy to build single-file reproductions of problems with the compiler that reproduced the effects of multi-file textual inclusion. When adding fancier header behaviors and especially modules support to Clang, there was *huge* demand to retain this single-file-reproduction mode. Lots of effort went into a bunch of options here, which were somewhat sadly ineffective. I actually think we'll really benefit from building a *very* low level facility into the lexer to split source buffers exactly as if from separate files, and then use it to build reproducers for bugs as well as to build this kind of split-file testing. Anyways, no need to frontload any of this, just wanted to write it down while I was thinking about it. =D

I did leave one minor suggestion on spelling the split marker in a way that looks a bit more similar to the other directives we have. I'm not set on any specific spelling here, mostly trying to make it fit into other things that are specially recognized. Also happy for fixing forward if needed there, so LGTM in whichever state is convenient.

testing/file_test/file_test_base.cpp
zygoloid: I believe the current syntax is chosen to match the behavior of the LLVM `split-file` tool.

zygoloid:
> Mostly FYI -- LLVM and Clang use a different approach in their test suites. Instead of splitting files, they have subdirectories containing non-test inputs. And the test just use relative paths to access them.

FWIW, Clang has increasingly been using the LLVM `split-file` tool instead of the `Inputs/` directory, though `Inputs` is still more common (1492 tests vs 243 using `split-file`). The ability to use the same inputs across multiple tests turned out to mostly be an attractive nuisance in Clang's tests, at least in part because different tests end up wanting to put different `expected-diagnostic` annotations in the same file.

One place where shared input files is nice and has, I think, worked well is where there's some library functionality that the compiler doesn't provide but still wants to use as test support. In Clang, the main examples of that are things like `std::initializer_list` and the coroutine support in the standard library. I'm not sure how much that'll come up for Carbon, though, given that we make the prelude available to tests.

I too am +1 on using splitting instead of input files for the most part.

testing/file_test/file_test_base.cpp
chandlerc: Ah, hadn't seen that tool, and didn't realize it used this syntax. I'm happy with whichever approach others prefer (consistency with `split-file`, or consistency with other special markers).

If no one else cares, TBH, I'd do our own thing unless we ever expect to share the test suite materially. But it's basically a coin toss for me.

jonmeow: I'm going to keep with `// ---` syntax for now, for the aforementioned `split-file` compatibility. Similarly a coin toss for me, but I think it may help people coming from the LLVM side of things similar to how we're trying to be consistent with `lit` and `FileCheck` syntax where it makes sense. It's also something that's easy to change later. [although hopefully not too much later, to avoid churn]

jonmeow:
I can look at switching to streams, but of course the current APIs don't support that, so I'll investigate separately. I don't think it should affect the core functionality of splitting so I'm going to go ahead and merge.

Issue #2153, opened 2022-09-07
Checked generics calling templates


josh11b:
Question: Given a function `TemplateFunc` with a template parameter, defined either in Carbon or C++, when is it legal for a generic function `GenericFunc` to call it?
```
fn TemplateFunc[template TT:! TC](x: TT);
fn GenericFunc[GT:! GC](x: GT) {
  TF(x);
}
```
Since `TemplateFunc` has a template parameter, it may have a _monomorphization error_, where the error is not detected when `TemplateFunc` is defined, but when it is called with a concrete type value for `TT`. This is expected, since templates are specifically for the cases when the definition of `TemplateFunc` cannot be checked on its own (or C++ interop). Constraints (like `TC`) are not required on template functions, but having them, assuming they match the actual conditions for template instantiation to succeed, "shifts errors left / earlier in the call stack." This narrows down where the error could be, improving the user experience. Named constraints may be reused, avoiding repeating constraints and maybe getting out of sync.

Background: Previously this question was investigated in #136 . 

Note: I am assuming that there are two kinds of constraints:

- **symbolic constraints** that operate symbolically on types, like "type implements interface `C`", and
- **value constraints** that require knowing the actual passed value to be evaluated, like "integer less than 100." These include all C++20 concepts.

**Options** from most permissive to least:

1. Always allowed, but may generate a monomorphization error in `GenericFunc` if the concrete `GT` used to instantiate `GenericFunc` does not satisfy the constraints on `TT`. "Calling a template exposes you to monomorphization errors, even from generics."
2. Always allowed, but the constraint `GC` may optionally include value constraints, including `TC`. This would give `GenericFunc` the _option_ of including `TC` in the constraints on `GT`. Non-generic callers of `GenericFunc` would validate the value constraints like `TC`. Generic callers of `GenericFunc` would either ignore value constraints like `TC`, or evaluate them at monomorphization time, potentially triggering monomorphization errors in the caller. Value constraints don't give additional capabilities on `GT` in the body of `GenericFunc`. "Generic functions may optionally include template constraints to shift monomorphization errors left/earlier in the call stack."
[...]

PR #2878, opened 2023-06-06
Bring constraints into scope when a generic calls a template.

zygoloid:
When a template has an argument that involves a generic parameter, we're supposed to delay instantiation until we know the concrete value, but explorer is not set up to do that yet, so for now we instead instantiate the template with the symbolic argument. When that happens, bring the constraints on the generic parameter into scope so they can be used inside the template instantiation.

This requires adding a new search over a value for the generic parameters that appear within it; a `VisitNestedValues` visitor is added to visit all the `Value`s nested with a value, and also convert an existing place where we were doing the same thing in a way that was incorrect (but harmlessly incorrect for now) to use it.

This is needed by #2881, which needs implementations of `ImplicitAs` for nested types when instantiating a builtin impl of `ImplicitAs` for an aggregate type.


explorer/ast/value.h
```diff
@@ -122,6 +122,12 @@ auto ValueEqual(Nonnull<const Value*> v1, Nonnull<const Value*> v2,
                 std::optional<Nonnull<const EqualityContext*>> equality_ctx)
     -> bool;
 
+// Call the given `visitor` on all values nested within the given value,
+// including `value` itself. Aborts and returns `false` if `visitor` returns
+// `false`, otherwise returns `true`.
```
geoffromer: Do we want to guarantee that it's a preorder traversal?

explorer/ast/value.h
```diff
@@ -122,6 +122,12 @@ auto ValueEqual(Nonnull<const Value*> v1, Nonnull<const Value*> v2,
...
+auto VisitNestedValues(Nonnull<const Value*> value,
```
geoffromer, line 128: What would you think of calling this `Traverse` or something like that? I feel like it might be a little clearer about the fact that it includes the root value and all descendants, not just the direct children.

explorer/ast/value.cpp
```diff
@@ -27,6 +28,86 @@ using llvm::dyn_cast;
 using llvm::dyn_cast_or_null;
 using llvm::isa;
 
+namespace {
+// A visitor that walks the Value*s nested within a value.
+struct NestedValueVisitor {
```
geoffromer, line 33: Is there any particular reason to make this a struct rather than a collection of free functions that take a callback parameter?

explorer/interpreter/type_checker.cpp
```diff
@@ -6407,18 +6397,47 @@ auto TypeChecker::InstantiateImplDeclaration(
...
+  // arguments into scope.
+  //
+  // TODO: There shouldn't be any checked generic bindings in the template
+  // arguments by the time we come to perform an instantiation, but in order
+  // for that to work, we need to defer instantiating templates until we know
+  // the values of checked generic parameters, such as by performing
+  // monomorphization for checked generics, which explorer doesn't yet do. See
+  // #2153 for the plan here.
+  //
+  // As a workaround for the lack of support for #2153, we can instantiate
```
geoffromer, line 6423: What's the motivation for implementing this workaround now, rather than waiting for proper #2153 support?

explorer/interpreter/type_checker.cpp
```diff
@@ -6407,18 +6397,47 @@ auto TypeChecker::InstantiateImplDeclaration(
...
-  auto* type_checker = const_cast<TypeChecker*>(this);
+  // Bring all impls from any checked generic bindings in the template
+  // arguments into scope.
+  //
+  // TODO: There shouldn't be any checked generic bindings in the template
+  // arguments by the time we come to perform an instantiation, but in order
+  // for that to work, we need to defer instantiating templates until we know
+  // the values of checked generic parameters, such as by performing
+  // monomorphization for checked generics, which explorer doesn't yet do. See
+  // #2153 for the plan here.
```
geoffromer: This phrasing led me to expect that #2153 tracks the plan for implementing this functionality in the Explorer.

```suggestion
  // monomorphization for checked generics (see #2153 for details). However,
  // explorer doesn't yet support that.
```

explorer/testdata/template/impl_scope.carbon
```diff
@@ -0,0 +1,32 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
```
geoffromer: Even with the rest of this PR for context, I have a hard time working out what this test is supposed to validate. I think it's testing support for a template argument depending on a generic binding, but e.g. why attach the template parameter to an `impl forall` rather than a function?

It would probably help to have a more descriptive filename, and maybe some comments and/or more descriptive names in the body of the file.

explorer/ast/value.cpp
zygoloid: Yes, there are a few:
- This avoids needing to explicitly pass the callback parameter through the various recursive calls.
- This avoids the need to forward declare of `Visit` / `VisitParts`.
- We want a closed overload set for the `Visit` functions; using a class ensures that we don't pick up any functions with the same name declared elsewhere (eg, found by ADL).

explorer/ast/value.h
zygoloid: Code is going to rely on that anyway, so that seems like a good idea. Done.

explorer/ast/value.h
zygoloid: I picked this name to be consistent with `VisitNestedPatterns`. No strong opinions about `VisitNested` vs `Traverse` other than that we should be consistent. For me, "nested" indicates "all descendants" well enough, and I find it a little clearer than `Traverse` -- what structure are we traversing? For example, do we follow pointers to their pointees? -- but I'm fine with `Traverse` if you feel strongly that it's a better name.

explorer/interpreter/type_checker.cpp
zygoloid: This is needed by #2881; I've updated the summary to say so. I don't expect explorer to implement #2153 in the near future, because it's more or less a rip-out-and-reimplement-generics change for explorer and relatively low priority if we don't need C++ interoperability, and I think #2881 is higher priority.

explorer/testdata/template/impl_scope.carbon
zygoloid: Is this better?

explorer/ast/value.h
geoffromer: Yeah, consistency should win here, and renaming `VisitNestedPatterns` is out of scope for this PR.

geoffromer (approved):

PR #2881, opened 2023-06-07
Support builtin conversions that internally rely on user-defined conversions

zygoloid:
There are a few implicit conversions that are implemented by an `impl` of `ImplicitAs` that delegates to code in explorer:

- Converting between tuple types
- Converting from tuples of types to `type`
- Converting from tuples of values to an array type
- Converting between struct types
- Converting from a struct type to a class type

These conversions can all rely on performing more conversions for elements or subobjects, but previously those inner conversions could only be performed if they were built into explorer. This change instead uses the full implicit conversion machinery in explorer to perform these conversions, including searching for a user-defined `impl` of `ImplicitAs` when necessary.

For example, this permits a conversion from `{.a: T}` to `{.a: U}`, or from `(T, T)` to `(U, U)`, or from `(T, T)` to `[U; 2]` when there is a user-defined conversion from `T` to `U`.

Depends on #2878


explorer/interpreter/type_checker.cpp
```diff
@@ -766,6 +768,186 @@ auto TypeChecker::BuildSubtypeConversion(Nonnull<Expression*> source,
...
+                FieldInitializer(destination_field.name, converted));
+          }
+          auto* result = arena_->New<StructLiteral>(source->source_loc(),
+                                                    std::move(result_fields));
+          CARBON_RETURN_IF_ERROR(TypeCheckExp(result, impl_scope));
+          return result;
+        }
+        case Value::Kind::NominalClassType: {
+          CARBON_ASSIGN_OR_RETURN(
+              auto field_types,
```
geoffromer:
```suggestion
              std::vector<NamedValue> field_types,
```

explorer/interpreter/type_checker.cpp
```diff
@@ -766,6 +768,186 @@ auto TypeChecker::BuildSubtypeConversion(Nonnull<Expression*> source,
...
+                                arena_->New<StructType>(field_types)));
+          // Perform a builtin conversion from struct to class.
+          return make_builtin_conversion(result);
+        }
+        case Value::Kind::TypeType:
+          // A value of empty struct type implicitly converts to type `type`.
+          if (cast<StructType>(*source_type).fields().empty()) {
+            return make_builtin_conversion(source);
+          }
+          break;
```
geoffromer: I think this (and most if not all of the other uses of `break` below) eventually transfers control to the `return conversion_failed();` at the bottom of the function, but it takes some effort to figure that out. Could these `break`s be `return conversion_failed();` instead?

explorer/testdata/array/conversion.carbon
```diff
@@ -0,0 +1,33 @@
...
+}
+
+fn Main() -> i32 {
+  var arr1: [i32; 2] = (1, 2 as A);
+  Print("1: {0}", arr1[0]);
+  Print("2: {0}", arr1[1]);
+
+  var arr2: [A; 2] = (3, 4 as A);
+  Print("3: {0}", arr1[0]);
+  Print("4: {0}", arr1[1]);
```
geoffromer: Were these supposed to be `arr2`?

zygoloid: They were, thanks. Fixed.

explorer/interpreter/type_checker.cpp
zygoloid: Sounds good, done.

explorer/interpreter/type_checker.cpp
zygoloid: Done here and for the two pre-existing callers of this function :)

geoffromer (approved):

PR #2883, opened 2023-06-08
Parsing support for `if` expressions.

zygoloid:
We model `if a then b else` as a prefix operator for parsing precedence purposes. The rule that a statement starting with `if` is never an `if` expression is handled implicitly because the statement parser never invokes the expression parser for a statement starting with `if`.

This exposed a bug in our diagnosis of the whitespace rule for prefix operators, which was incorrectly being applied to non-symbolic operators in some cases, and was producing a bogus second diagnostic in some cases, which is also fixed here.

### DIFF
--- a/toolchain/parser/parser_state.def
+++ b/toolchain/parser/parser_state.def
@@ -242,6 +242,29 @@ CARBON_PARSER_STATE(ExpressionLoopForBinary)
 //   1. ExpressionLoop
 CARBON_PARSER_STATE(ExpressionLoopForPrefix)
 
+// Completes an IfExpression.
+//
+// Always:
+//   (state done)
+CARBON_PARSER_STATE(ExpressionIfFinish)
+
+// Handles the `then` token in an `if` expression.
+//
+// If `Then`:
+//   1. Expression
+//   2. ExpressionElse
+// Else:
+//  (state done)
+CARBON_PARSER_STATE(ExpressionThen)
+
+// Handles the `else` token in an `if` expression.
+//
+// If `Else`:
+//   1. Expression
+// Else:
+//  (state done)
+CARBON_PARSER_STATE(ExpressionElse)
+
 // Handles the `;` for an expression statement, which is different from most
 // keyword statements.
 //
--- a/toolchain/parser/parser_state.def
+++ b/toolchain/parser/parser_state.def
@@ -181,7 +181,11 @@ CARBON_PARSER_STATE_VARIANTS2(Designator, Expression, Struct)
 
 // Handles processing of an expression.
 //
-// If valid prefix operator:
+// If `If`:
+//   1. Expression
+//   2. ExpressionThen
+//   3. ExpressionIfFinish
+// Else if valid prefix operator:
 //   1. Expression
 //   2. ExpressionLoopForPrefix
 // Else:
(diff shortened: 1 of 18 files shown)


toolchain/parser/parser_state.def
```diff
@@ -242,6 +242,29 @@ CARBON_PARSER_STATE(ExpressionLoopForBinary)
 //   1. ExpressionLoop
 CARBON_PARSER_STATE(ExpressionLoopForPrefix)
 
+// Completes an IfExpression.
```
jonmeow, line 249: Please note the entry condition in `CARBON_PARSER_STATE(Expression)`, around line 187 (unfortunately can't comment there):

```
// If `If`:
//   1. Expression
//   2. ExpressionThen
//   3. ExpressionIfFinish
```

(at least as currently implemented)

toolchain/semantics/semantics_handle.cpp
```diff
@@ -398,6 +398,21 @@ auto SemanticsHandlePrefixOperator(SemanticsContext& context,
   return context.TODO(parse_node, "HandlePrefixOperator");
 }
 
+auto SemanticsHandleIfExpression(SemanticsContext& context,
```
jonmeow: FWIW, I've generally tried to keep these lexically ordered to make it easier to sift through (`IfExpression*` before `IfStatement*`)

jonmeow (approved):
LGTM

toolchain/parser/testdata/if_expression/basic.carbon
```diff
@@ -0,0 +1,39 @@
...
+// CHECK:STDOUT:       {kind: 'IfExpressionElse', text: 'else'},
+// CHECK:STDOUT:       {kind: 'NameReference', text: 'y'},
+// CHECK:STDOUT:     {kind: 'IfExpression', text: 'if', subtree_size: 6},
+// CHECK:STDOUT:   {kind: 'ReturnStatement', text: ';', subtree_size: 8},
+// CHECK:STDOUT: {kind: 'FunctionDefinition', text: '}', subtree_size: 27},
+// CHECK:STDOUT: {kind: 'FileEnd', text: ''},
+// CHECK:STDOUT: ]
+
+fn F(b: bool, x: i32, y: i32) -> bool {
+  return if b then x else y;
```
jonmeow, line 38: One minor thing, you might also test with a missing expression:

`return if;`

(right now, the testing of invalid syntax is limited though, throughout parser)

toolchain/parser/testdata/if_expression/fail_then_expr_missing.carbon
```diff
@@ -0,0 +1,32 @@
...
+// CHECK:STDOUT:       {kind: 'InvalidParse', text: ';', has_error: yes},
+// CHECK:STDOUT:     {kind: 'IfExpression', text: 'if', has_error: yes, subtree_size: 6},
+// CHECK:STDOUT:   {kind: 'VariableDeclaration', text: ';', subtree_size: 12},
+// CHECK:STDOUT: {kind: 'FunctionDefinition', text: '}', subtree_size: 18},
+// CHECK:STDOUT: {kind: 'FileEnd', text: ''},
+// CHECK:STDOUT: ]
+
+fn F() {
+  // CHECK:STDERR: fail_then_expr_missing.carbon:[[@LINE+2]]:28: Expected expression.
+  // CHECK:STDERR: fail_then_expr_missing.carbon:[[@LINE+1]]:28: Expected `else` after `if ... then ...`.
```
jonmeow: More of a TODO, but probably we will want to consolidate these errors.

jonmeow (approved):

jonmeow (approved):

PR #2885, opened 2023-06-08
Switch FileTest to use StringRefs instead of files.

jonmeow:
In explorer, we already support parsing a string_view, so use that. In toolchain, we need to build support, probably using vfs, so that's a todo.

bazel test //explorer:file_test --runs_per_test=5

- branch: Stats over 250 runs: max = 18.3s, min = 5.2s, avg = 11.2s, dev = 2.9s
- trunk: Stats over 250 runs: max = 22.3s, min = 5.9s, avg = 12.1s, dev = 2.8s

Not a dramatic improvement, but maybe more effective long-term, and this'd been requested on #2876

### DIFF
--- a/explorer/file_test.cpp
+++ b/explorer/file_test.cpp
@@ -32,7 +32,7 @@ class ParseAndExecuteTestFile : public FileTestBase {
     }
   }
 
-  auto RunWithFiles(const llvm::SmallVector<std::string>& test_files,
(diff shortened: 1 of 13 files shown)


chandlerc (commented):
Nice!

chandlerc (approved):
And actually hitting the LGTM button....

PR #2888, opened 2023-06-09
Add vfs support to toolchain.

jonmeow:
This adds vfs support to the toolchain, allowing Driver to take in-memory inputs in tests. As a consequence, I'm simplifying SourceBuffer: rather than allowing tests to pass in their own memory buffer, I'm using InMemoryFileSystem to push for greater consistency with production code. This does hit a quirk where I need to be careful about null terminator handling because fuzzer imports don't always have one, but that's probably more robust anyways.

### DIFF
--- a/toolchain/driver/driver_fuzzer.cpp
+++ b/toolchain/driver/driver_fuzzer.cpp
@@ -66,10 +66,10 @@ extern "C" auto LLVMFuzzerTestOneInput(const unsigned char* data, size_t size)
     size -= arg_length;
   }
 
-  auto fs = llvm::vfs::getRealFileSystem();
+  llvm::vfs::InMemoryFileSystem fs;
   std::string error_text;
   llvm::raw_string_ostream error_stream(error_text);
-  Driver d(*fs, llvm::nulls(), error_stream);
+  Driver d(fs, llvm::nulls(), error_stream);
   if (!d.RunFullCommand(args)) {
     error_stream.flush();
     if (error_text.find("ERROR:") == std::string::npos) {
--- a/toolchain/driver/driver_fuzzer.cpp
+++ b/toolchain/driver/driver_fuzzer.cpp
@@ -66,10 +66,10 @@ extern "C" auto LLVMFuzzerTestOneInput(const unsigned char* data, size_t size)
     size -= arg_length;
   }
 
+  auto fs = llvm::vfs::getRealFileSystem();
   std::string error_text;
   llvm::raw_string_ostream error_stream(error_text);
-  llvm::raw_null_ostream output_stream;
-  Driver d(output_stream, error_stream);
+  Driver d(*fs, llvm::nulls(), error_stream);
   if (!d.RunFullCommand(args)) {
     error_stream.flush();
     if (error_text.find("ERROR:") == std::string::npos) {
--- a/toolchain/driver/driver_test.cpp
+++ b/toolchain/driver/driver_test.cpp
@@ -21,7 +21,7 @@ using ::testing::ElementsAre;
 using ::testing::HasSubstr;
 using ::testing::StrEq;
 
-static constexpr llvm::StringLiteral TestFileName = "test_file.txt";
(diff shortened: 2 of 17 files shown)


toolchain/driver/driver_fuzzer.cpp
```diff
@@ -66,10 +66,10 @@ extern "C" auto LLVMFuzzerTestOneInput(const unsigned char* data, size_t size)
     size -= arg_length;
   }
 
+  auto fs = llvm::vfs::getRealFileSystem();
```
zygoloid: What do you think about creating an `InMemoryFileSystem` here? It seems a bit concerning to allow the fuzzer to read from arbitrary files on the file system. I mean, it's not likely to find `/dev/zero` or similar and run forever, but it certainly could. Eventually, filling that `InMemoryFileSystem` with a known set of example files would be nice too, but just getting to the point where the fuzzer isn't also fuzzing the file system itself would be a nice advantage of switching to a vfs :)

(Also, once the driver starts being able to write to files, we definitely don't want it to be able to overwrite arbitrary real files when fuzzing.)

I'm totally happy for you to say "not in this PR".

toolchain/driver/driver_test.cpp
```diff
@@ -21,86 +21,72 @@ using ::testing::ElementsAre;
 using ::testing::HasSubstr;
 using ::testing::StrEq;
 
-TEST(DriverTest, FullCommandErrors) {
-  TestRawOstream test_output_stream;
-  TestRawOstream test_error_stream;
-  Driver driver = Driver(test_output_stream, test_error_stream);
+static constexpr llvm::StringLiteral TestFileName = "test_file.txt";
```
zygoloid:
```suggestion
static constexpr llvm::StringLiteral TestFileName = "test_file.carbon";
```

toolchain/parser/parse_tree_fuzzer.cpp
```diff
@@ -22,9 +22,15 @@ extern "C" int LLVMFuzzerTestOneInput(const unsigned char* data,
...
-  auto source = SourceBuffer::CreateFromText(
-      llvm::StringRef(reinterpret_cast<const char*>(data), size));
+  static constexpr llvm::StringLiteral TestFileName = "test.carbon";
+  llvm::vfs::InMemoryFileSystem fs;
+  llvm::StringRef data_ref(reinterpret_cast<const char*>(data), size);
+  CARBON_CHECK(fs.addFile(
+      TestFileName, /*ModificationTime=*/0,
+      llvm::MemoryBuffer::getMemBuffer(data_ref, /*BufferName=*/TestFileName,
+                                       /*RequiresNullTerminator=*/false)));
+  llvm::errs() << "made\n";
```
zygoloid:
```suggestion
```

toolchain/source/source_buffer.cpp
```diff
@@ -4,139 +4,38 @@
...
-#endif
-                           file_descriptor, /*offset=*/0);
-  // The `MAP_FAILED` macro may expand to a cast to pointer that `clang-tidy`
-  // complains about.
-  // NOLINTNEXTLINE(performance-no-int-to-ptr)
-  if (mapped_text == MAP_FAILED) {
-    return ErrnoToError(errno);
+  auto size = status->getSize();
+  if (size >= std::numeric_limits<int32_t>::max()) {
+    return Error(llvm::formatv("`{0}` is over the 2MB input limit.", filename));
```
zygoloid:
```suggestion
    return Error(llvm::formatv("`{0}` is over the 2GB input limit.", filename));
```
or maybe
```suggestion
    return Error(llvm::formatv("`{0}` is over the 2GiB input limit.", filename));
```

zygoloid (changes requested):

toolchain/driver/driver_fuzzer.cpp
jonmeow: Done

zygoloid (approved):

PR #2889, opened 2023-06-09
Switch SemanticsIRTest to just use the driver.

jonmeow:
Now that the driver uses vfs, there's less reason for tests to do their own flow. Switch SemanticsIRTest to use the driver directly as an example simplification.


chandlerc (approved):
Slightly unfortunate layering change, but I suppose its OK for test code. LG.

PR #2890, opened 2023-06-09
Migrate //common test libraries to //testing/util.

jonmeow:
This is just a cleanup. Since we now have a testing directory, I think this is a better home for testonly libraries than //common. (I was thinking about this when I was considering adding more test_raw_ostream deps)


geoffromer (approved):

chandlerc (commented):
I'm hesitant to suggest changes given the amount of merge conflicts I see in my future...

That said, why a separate package below `//testing`?

I'd also somewhat rather not have a mixture of both `common` and `util` subdirectories as then we have to remember which ones are which, but maybe there is no need to cross that bridge.

jonmeow:
> That said, why a separate package below `//testing`?
> 
> I'd also somewhat rather not have a mixture of both `common` and `util` subdirectories as then we have to remember which ones are which, but maybe there is no need to cross that bridge.

It's my preference not to have significant code in a parent package. So for example since //testing has lots of child packages (e.g., //testing/file_test) I'd want to avoid a significant //testing package that also has lots of files. There are a couple reasons for this, but I will note I think //common is becoming a bit of a mess so I don't think it's a positive example. Personally I'd be in favor of an equivalent //common/util package that collects things not in their own directory [versus folding //testing/util to //testing]. Maybe it'd be good for others to weigh in with thoughts?

jonmeow:
Another option is //common/testing, //common/file_test, //common/test_scripts, etc.

I'm holding off on merging to see about addressing @chandlerc's concerns.

jonmeow:
Discussed this with Richard a little: I'd forgotten about //explorer/common and //toolchain/common. The way we're leaning is that /util directories should be used in those places, and //common should mean "common across top-level projects".

My impression is you're mainly objecting to the mix of these, so I'm going to merge and follow up with renames of those /common subdirectories (leaving the top-level directory alone).

chandlerc:
Definitely glad this merged, holding that up doesn't make sense. =D This is all *super* easy to fix forward.

But let's actually chat about what renaming will work best across all of us before actually doing more renaming? I've had really bad experience with naming things `util` and have heard the same from a bunch of library maintainers over the years, and so I'm very hesitant with that naming. For me, `common` had specific utility -- it isn't just *any* utility code, it is specifically code that can be *common* across different users (whether different parts of the toolchain, or the entire project). Anyways, I suspect we can find a good balance here, but probably best to discuss and find that before making more PRs.

PR #2891, opened 2023-06-09
Migrate compatible uses to TestRawOstream.

jonmeow:
Replacing direct raw_string_ostream uses. I figure the wrapper should be used more consistently.

There are still remaining raw_string_ostream uses that weren't compatible -- I'm continuing to look at those, but felt it was cleaner to have this on its own.


chandlerc (approved):
LG (whenever the prior PRs land)

PR #2892, opened 2023-06-09
Migrate some lexer tests to file tests.

jonmeow:
These tests are doing string comparisons on output that don't seem to be meaningfully different from a file_test.

I'm tempted to migrate lexer tests in general, but I'm not doing that here since others may find more value in the current approach.


toolchain/driver/driver_test.cpp
```diff
@@ -21,86 +21,72 @@ using ::testing::ElementsAre;
...
-  EXPECT_FALSE(driver.RunFullCommand({}));
-  EXPECT_THAT(test_error_stream.TakeStr(), HasSubstr("ERROR"));
+class DriverTest : public testing::Test {
+ protected:
+  DriverTest() : driver_(fs_, test_output_stream_, test_error_stream_) {}
 
-  EXPECT_FALSE(driver.RunFullCommand({"foo"}));
-  EXPECT_THAT(test_error_stream.TakeStr(), HasSubstr("ERROR"));
+  auto CreateTestFile(llvm::StringRef text) -> void {
+    fs_.addFile(TestFileName, /*ModificationTime=*/0,
```
geoffromer, line 30: Could the file name be a parameter to this function? Having it use a fixed name seems confusingly implicit to me, and could also limit the generality of the fixture.

toolchain/lexer/tokenized_buffer_test.cpp
```diff
@@ -31,19 +31,24 @@ using ::testing::StrEq;
...
+  auto Lex(llvm::StringRef text,
            DiagnosticConsumer& consumer = ConsoleDiagnosticConsumer())
       -> TokenizedBuffer {
     return TokenizedBuffer::Lex(GetSourceBuffer(text), consumer);
   }
 
-  llvm::SmallVector<SourceBuffer, 16> source_storage;
+  llvm::vfs::InMemoryFileSystem fs;
+  int file_index = 0;
+  std::forward_list<SourceBuffer> source_storage;
```
geoffromer: Shouldn't these have a `_` at the end of their names?

toolchain/lexer/testdata/digit_padding.carbon
```diff
@@ -0,0 +1,17 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
```
zygoloid: Maybe rename this test to `printing_digit_padding`? It took me a moment to realize that this was testing the YAML printing rather than the lexing behavior.

zygoloid (approved):
LGTM (not merging because there are other changes above this in the stack).

toolchain/lexer/testdata/digit_padding.carbon
jonmeow: Done

toolchain/lexer/tokenized_buffer_test.cpp
jonmeow: Done -- I'd been leaving source_storage alone initially and aiming for consistency, but now I suppose I'm making up the majority.

toolchain/driver/driver_test.cpp
jonmeow: Made a return. Note that generality isn't a goal; this is called twice and works for both callers.

geoffromer (approved):

PR #2893, opened 2023-06-10
Semantic analysis for `if` expressions

zygoloid:
Add semantic analysis and semantics IR building for `if` expressions, and add the first parts of control flow handling to semantics IR. After discussion with @chandlerc, use [block arguments](https://en.wikipedia.org/wiki/Static_single-assignment_form#Block_arguments) to convey values from the two arms of the `if` to the result. For now, only a single block argument is supported, but we should revisit this as we explore more of the requirements of the Semantics IR form.

Functions can now contain multiple code blocks, so grab the entry block up-front instead of assuming the entry block will be at the top of the block stack when we reach the end of function emission.

Add trivial support for `bool` type literal, because without it we can't write testcases.


toolchain/driver/testdata/semantics_verbose.carbon
```diff
@@ -6,7 +6,7 @@
 //
 // Only checks a couple statements in order to minimize manual update churn.
 // CHECK:STDERR: Node Push 0: FunctionIntroducer -> <none>
-// CHECK:STDERR: AddNode block{{[0-9]+}}: {kind: FunctionDeclaration, arg0: function{{[0-9]+}}}
+// CHECK:STDERR: AddNode block2: {kind: Return}
```
jonmeow:
```suggestion
// CHECK:STDERR: AddNode block{{[0-9]+}}: {kind: Return}
```

I'm fine switching the node checked for, but it should probably remain block agnostic in case numbering changes, minimizing brittleness.

toolchain/semantics/semantics_handle_function.cpp
```diff
@@ -18,17 +18,10 @@ auto SemanticsHandleFunctionDefinition(SemanticsContext& context,
          ParseNodeKind::FunctionDefinitionStart) {
     context.node_stack().PopAndIgnore();
   }
-  auto function_id = context.node_stack().Pop<SemanticsFunctionId>(
-      ParseNodeKind::FunctionDefinitionStart);
-
+  context.node_stack().PopAndIgnore();
```
jonmeow: Please strongly prefer PopAndDiscard over PopAndIgnore. The difference is that PopAndDiscard continues to validate structure, whereas PopAndIgnore doesn't -- allowing incorrect structure expectations to go on longer.
```suggestion
  context.node_stack().PopAndDiscardSoloParseNode(ParseNodeKind::FunctionDefinitionStart);
```

toolchain/semantics/semantics_handle_if_expression.cpp
```diff
@@ -0,0 +1,91 @@
...
+                                 ParseTree::Node parse_node) -> bool {
+  auto else_value = context.node_stack().Pop<SemanticsNodeId>();
+  auto [else_node, else_block] =
+      context.node_stack().PopWithParseNode<SemanticsNodeBlockId>();
+  auto then_end_block = context.node_stack().Pop<SemanticsNodeBlockId>();
+  auto then_value = context.node_stack().Pop<SemanticsNodeId>();
+  auto [then_node, then_block] =
+      context.node_stack().PopWithParseNode<SemanticsNodeBlockId>();
+  auto if_block = context.node_stack().Pop<SemanticsNodeBlockId>();
+  auto cond_value = context.node_stack().Pop<SemanticsNodeId>();
```
jonmeow: When popping, please pass in expected parse node kinds, for non-expressions (since expressions can have basically any kind). As before, this is mainly for verification that the structure is as expected: it's easy to slip up and have cascading errors making it hard to identify the root cause.

toolchain/semantics/semantics_handle_function.cpp
```diff
@@ -54,18 +47,21 @@ auto SemanticsHandleFunctionDefinitionStart(SemanticsContext& context,
   auto name_str = context.parse_tree().GetNodeText(name_node);
   auto name_id = context.semantics_ir().AddString(name_str);
 
-  // Add the callable, but with an invalid body for now. The body ID is only
-  // finalized on function completion.
+  // Create the entry block.
+  auto outer_block = context.node_block_stack().PeekForAdd();
```
jonmeow, line 51: FWIW I've been trying to do `_id` as a reminder for when it's a node block ID versus the actual node block (and similar).

toolchain/semantics/semantics_handle_function.cpp
```diff
@@ -54,18 +47,21 @@ auto SemanticsHandleFunctionDefinitionStart(SemanticsContext& context,
...
+  context.node_block_stack().Push();
+
+  // Add the callable.
   auto function_id = context.semantics_ir().AddFunction(
       {.name_id = name_id,
        .param_refs_id = param_refs_id,
        .return_type_id = return_type_id,
-       .body_id = SemanticsNodeBlockId::Invalid});
-  auto decl_id = context.AddNode(
+       .body_id = context.node_block_stack().PeekForAdd()});
```
jonmeow, line 59: WDYT of adding a `PushForAdd() -> SemanticsNodeBlockID` to NodeBlockStack?

In general, I've been optimizing towards avoiding creating empty node blocks. I will note, this change negates that because it always creates an empty node block. But the consequence is that `Push` adds an invalid ID, then `PeekForAdd` immediately pops and creates another (with a branch). A `PushForAdd` would reduce churn by unconditionally creating the node block, and could also return the ID directly.

Note, I'm mulling empty function bodies, and maybe they'll be exceptionally rare in Carbon and thus not worth optimizing for? In C++ you'd often see that for constructors, but given Carbon's going in the way of functions, a C++ `Foo() {}` might become in Carbon `fn MakeFoo() -> Foo() { return {}; }` which means there will be fewer empty function bodies. If you think this would be helpful, maybe add a TODO: it'd probably be as simple as peeking at the next parse node, and if it's `}` then just add a function with an empty body (but would still need to deal with parameters to catch name duplicates).

toolchain/semantics/semantics_handle_if_expression.cpp
```diff
@@ -0,0 +1,91 @@
...
+  // Create the `then` block.
+  context.node_block_stack().Push();
+  auto then_block = context.node_block_stack().PeekForAdd();
+  context.node_stack().Push(then_node, then_block);
+  return true;
+}
+
+auto SemanticsHandleIfExpressionElse(SemanticsContext& context,
+                                     ParseTree::Node else_node) -> bool {
+  context.node_stack().Push(else_node, context.node_block_stack().Pop());
```
jonmeow: Why are you pushing back the node that's just popped, with a different parse node associated?

toolchain/semantics/semantics_handle_if_expression.cpp
```diff
@@ -0,0 +1,91 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "toolchain/semantics/semantics_context.h"
+
+namespace Carbon {
+
+auto SemanticsHandleIfExpressionThen(SemanticsContext& context,
+                                     ParseTree::Node then_node) -> bool {
+  // TODO: Find the `if` node.
+  ParseTree::Node if_node = then_node;
```
jonmeow: Finding the future parse node is O(n) where n is the number of nodes in the parse tree because it requires traversing expressions. Why do you need it? Should the parse tree be redesigned in order to make it O(1)?

toolchain/semantics/semantics_handle_if_expression.cpp
```diff
@@ -0,0 +1,91 @@
...
+  // Convert the `else` value to the `then` value's type, and finish the `else`
+  // block.
+  // TODO: Find a common type, and convert both operands to it instead.
+  auto result_type = context.semantics_ir().GetNode(then_value).type_id();
+  else_value = context.ImplicitAsRequired(parse_node, else_value, result_type);
+  auto else_end_block = context.node_block_stack().Pop();
+
+  // Create a resumption block.
+  context.node_block_stack().Push();
+  auto resume_block = context.node_block_stack().PeekForAdd();
```
jonmeow: Uses like this are also a good case for a `PushForAdd` on `NodeBlockStack`. But do consider, you aren't making sure there's anything in resume_block.

Also, should this be created closer to where it's used, after the `if_block` logic?

jonmeow (commented):
I think we need to discuss the semantic handling versus the parse tree structure.

In the parse tree, you have 5 nodes flat beneath `if`: `if_expr`, `then`, `then_expr`, `else`, `else_expr`.

The way this is being handled though, you're manipulating the tree with pop+push instead of doing a single pass. The extra entries on node_stack feel weird to me; I've been using that to match the parse tree, and there's an extra charge for using it for single indices of data because there's the parse node overhead. Also note the O(n) complexity for finding `if` from `then`. This makes me think that there's some skew between the parse tree design and what you're trying to do.

Here's some suggestions:

1. Redesign the parse tree:

    ```
      +-- if_expr
      |
    +-+ if
    |
    | +-- then_expr
    | |
    +-+ then
    |
    +-- else_expr
    |
    + else
    ```

    In this model, you eliminate the find of `if` by having it immediately after `if_expr`. `else` is now the root node. `if` could alternately be a child of `then`, but it feels like it doesn't matter thus I suggest a child of `else`.

2. Behavior:
    1. On `if`:
        1. node_stack: pop `if_expr`
        2. Convert the value to a bool.
        3. node_block_stack: pop the `if_expr` block.
        4. node_block_stack: push the resume block, `else_expr` block, and `then_expr` block in that order.
        5. Add BranchIf and Branch to the `if_expr` block, since you have all the necessary blocks.
        6. node_stack: push `if` solo, no data [habit, mirroring the parse tree]
    2. On `then`:
        1. node_stack: pop `then_expr`
        2. node_block_stack: pop the `then_expr` block.
        3. node_stack: push `then` with the `then_expr` NodeBlockId.
    3. On `else`:
        1. node_stack: pop `else_expr`, `then` with NodeBlockId, and `if` solo (no value)
        2. Note the type of the `then_expr` is the last node on the `then` NodeBlockId.
        3. Cast `else_expr` to the `then_expr` type (conversion TODO still applies).
        4. Add the BranchWithArg for the resume in `then_expr` and `else_expr` blocks, noting that it's deterministic what the relative location is in node_block_stack. (requires new support on node_block_stack)
        5. node_block_stack: Pop the `else_expr` block.
        6. Add BlockArg (will go onto the resume block)

What I'm thinking about for this:

- By using `node_block_stack` to store the blocks up-front, it'll eliminate the `node_stack` overhead.
  - The blocks need to go there regardless, they're just being added earlier.
- `node_stack` gets entry counts mirroring parse tree structure.
  - [each node is pushed and popped precisely once]
- Branches are inserted as you go [somewhat]; there's [less] need to back up for them.
- When adding the implicit conversion to a common type, this leaves things open to add the cast to the end of the `then_expr` block without needing a StubReference.

toolchain/semantics/semantics_handle_if_expression.cpp
zygoloid: Done, though most of the remaining calls here are popping expressions.

toolchain/semantics/semantics_handle_function.cpp
zygoloid: Done.

toolchain/semantics/semantics_handle_if_expression.cpp
zygoloid: Note that this is popping a node block and pushing a node.

toolchain/semantics/semantics_handle_if_expression.cpp
zygoloid: Done

toolchain/semantics/semantics_handle_if_expression.cpp
zygoloid: The resume block always contains the block parameter node, so is never empty. Also, it's used as a branch target, so we can't elide it even if it were empty. Creation of this block has now been moved earlier.

jonmeow (approved):
Looks good, thanks! Maybe we'll need a standard solution for token propagation/state mutation issue in the future, but this seems good for now.

PR #2894, opened 2023-06-10
Update main.cpp

madhupashish:
I have made some changes that are Listed below :

1. Added some of the missing include directives.

2. Removed some of unnecessary headers.

3. Added   ' llvm::sys::fs::OpenFlags::OF_Text '   flag when opening the trace file stream to ensure it's opened in text mode.

4. Updated the usage of     'cl::list'    to specify  
    'cl::CommaSeparated'   to  allow comma-separated values.

adding, and why.**

Closes #ISSUE

### DIFF
--- a/explorer/main.cpp
+++ b/explorer/main.cpp
@@ -43,8 +43,7 @@ auto ExplorerMain(int argc, char** argv, void* static_for_main_addr,
   llvm::errs().tie(&llvm::outs());
 
   cl::opt<std::string> input_file_name(cl::Positional, cl::desc("<input file>"),
-                                       cl::Required, cl::value_desc("filename"),
-                                       cl::init("input.carbon"));
+                                       cl::Required);
   cl::opt<bool> parser_debug("parser_debug",
                              cl::desc("Enable debug output from the parser"));
 
--- a/explorer/main.cpp
+++ b/explorer/main.cpp
@@ -29,34 +29,73 @@ namespace Carbon {
 namespace cl = llvm::cl;
 namespace path = llvm::sys::path;
 
-int ExplorerMain(int argc, char** argv, void* static_for_main_addr, llvm::StringRef relative_prelude_path) {
-  llvm::setBugReportMsg("Please report issues to https://github.com/carbon-language/carbon-lang/issues and include the crash backtrace.\n");
+auto ExplorerMain(int argc, char** argv, void* static_for_main_addr,
+                  llvm::StringRef relative_prelude_path) -> int {
+  llvm::setBugReportMsg(
+
+      "Please report issues to "
+      "https://github.com/carbon-language/carbon-lang/issues and include the "
+      "crash backtrace.\n");
   llvm::InitLLVM init_llvm(argc, argv);
 
-  // Printing to stderr should flush stdout. This is most noticeable when stderr is piped to stdout.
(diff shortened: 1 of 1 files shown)


madhupashish:
Please Review it and Merge it.

jonmeow (changes requested):
All I'm seeing here seems to be removing the formatting we use (we use clang-format via pre-commit, see [here](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/project/contribution_tools.md#tools)), with none of the fixes mentioned in the PR comment. Did you push the right change?

madhupashish (commented):
Must Read My New Code

explorer/main.cpp
```diff
@@ -123,9 +128,8 @@ auto ExplorerMain(int argc, char** argv, void* static_for_main_addr,
     llvm::outs() << "result: " << *result << "\n";
     return EXIT_SUCCESS;
   } else {
-    llvm::errs() << result.error() << "\n";
+    llvm::errs() << "Execution error: " << result.error() << "\n";
```
jonmeow:
```suggestion
    llvm::errs() << result.error() << "\n";
```

The returned error will already be labeled, so this would be duplicative.

explorer/main.cpp
```diff
@@ -123,9 +128,8 @@ auto ExplorerMain(int argc, char** argv, void* static_for_main_addr,
...
     return EXIT_FAILURE;
   }
 }
```
jonmeow:
```suggestion
}

```

We keep blank lines in spots like this per [our style guide](https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace):

> Blank lines immediately inside a declaration of a namespace or block of namespaces may help readability by visually separating the load-bearing content from the (largely non-semantic) organizational wrapper. Especially when the first declaration inside the namespace(s) is preceded by a comment, this becomes a special case of the previous rule, helping the comment to "attach" to the subsequent declaration.

explorer/main.cpp
```diff
@@ -43,14 +43,18 @@ auto ExplorerMain(int argc, char** argv, void* static_for_main_addr,
   llvm::errs().tie(&llvm::outs());
 
   cl::opt<std::string> input_file_name(cl::Positional, cl::desc("<input file>"),
-                                       cl::Required);
+                                       cl::Required, cl::value_desc("filename"),
+                                       cl::init("input.carbon"));
```
jonmeow:
```suggestion
                                       cl::Required);
```

This flag is positional and required. I don't think either value_desc or init have an effect, but if I'm missing something, can you explain with a diff in help output?

jonmeow (commented):
Thanks! Looking through these changes, I would like to restore a few things.

Can you also check the main PR comment and update it to reflect the changes currently being made? It looks like things have fallen out of sync, e.g. there aren't any include changes.

madhupashish:
I have merged all your request and i think now your are also satisfied with my code .
Now you can Merge it.

jonmeow (approved):
Thank you for the edits. Following up on my previous message, I will rewrite your commit message, but we'd really appreciate if you provide accurate commit messages if you plan to continue contributing.

jonmeow:
I want to note, we removed this patch because it does not qualify as a Contribution under the terms of Google's contributor license agreement.

PR #2895, opened 2023-06-13
Update the LLVM version in use.

jonmeow:
The prior terminfo and zlib calls are obsolete. I'm adding the zstd library myself as a quick fix, although I want to investigate if we can make better use of [LLVM's workspace](https://github.com/llvm/llvm-project/blob/main/utils/bazel/WORKSPACE) (where the zstd dep comes from).

```
ERROR: .../external/llvm-project/llvm/BUILD.bazel:184:11: no such package '@llvm_zstd//': The repository '@llvm_zstd' could not be resolved: Repository '@llvm_zstd' is not defined and referenced by '@llvm-project//llvm:Support'
```

The particular commit in use fixes a macos build error. https://github.com/llvm/llvm-project/commit/c5f6a287499a816cba5585708999e2c8b134290f

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -5,6 +5,7 @@
 workspace(name = "carbon")
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
 
 skylib_version = "1.3.0"
 
@@ -112,7 +113,7 @@ http_archive(
 
 # We pin to specific upstream commits and try to track top-of-tree reasonably
 # closely rather than pinning to a specific release.
-llvm_version = "219ba2fb7b0ae89101f3c81a47fe4fc4aa80dea4"
+llvm_version = "c5f6a287499a816cba5585708999e2c8b134290f"
 
 http_archive(
     name = "llvm-raw",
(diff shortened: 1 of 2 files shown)


jonmeow:
Working on a patch to clang:

```
genrule(
    name = "builtin_headers_gen",
    srcs = builtin_headers,
    outs = [hdr.replace("lib/Headers/", "staging/include/") for hdr in builtin_headers],
    cmd = """
       srcs_array=($(SRCS))
       outs_array=($(OUTS))
       for i in $${!srcs_array[@]}; do
         cp $${srcs_array[$$i]} $${outs_array[$$i]}
       done""",
    output_to_bindir = 1,
    toolchains = [
        ":workspace_root",
    ],
)
```

What they do right now for cmd [here](https://github.com/llvm/llvm-project/blob/0258a53521cfedf5cb80c2b1d4a66c942615de74/utils/bazel/llvm-project-overlay/clang/BUILD.bazel#L1692-L1698) appears not to be cross-platform compatible.

zygoloid (approved):
Looks fine, but the test failure looks likely to be real.

jonmeow:
> Looks fine, but the test failure looks likely to be real.

Adding zstd handling similar to zlib (basically just ignoring it)

PR #2896, opened 2023-06-13
`true` and `false` support, and lowering for `bool` type.

zygoloid:


toolchain/semantics/semantics_node.h
```diff
@@ -68,6 +68,21 @@ struct SemanticsCrossReferenceIRId : public IndexBase {
...
+struct SemanticsBoolLiteralId : public IndexBase {
+  using IndexBase::IndexBase;
+  auto Print(llvm::raw_ostream& out) const -> void {
+    if (index == 0) {
+      out << "false";
+    } else if (index == 1) {
+      out << "true";
+    } else {
+      out << "int";
+      IndexBase::Print(out);
```
jonmeow:
```suggestion
      CARBON_FATAL() << "Invalid bool value: " << index;
```

Definitely shouldn't keep the `int` label, would be `bool`, but that seems only necessary if you create with Invalid or other indices. That seems to be unsupported behavior, so maybe nix it?

toolchain/semantics/semantics_node.h
```diff
@@ -68,6 +68,21 @@ struct SemanticsCrossReferenceIRId : public IndexBase {
   }
 };
 
+// The ID of a boolean value.
+struct SemanticsBoolLiteralId : public IndexBase {
+  using IndexBase::IndexBase;
+  auto Print(llvm::raw_ostream& out) const -> void {
+    if (index == 0) {
```
jonmeow: Maybe a switch? :)

toolchain/semantics/semantics_node.h
```diff
@@ -68,6 +68,21 @@ struct SemanticsCrossReferenceIRId : public IndexBase {
   }
 };
 
+// The ID of a boolean value.
+struct SemanticsBoolLiteralId : public IndexBase {
```
jonmeow: Maybe drop the "Id" from the name and comment, since this is the literal value. No concern sticking with IndexBase

jonmeow (approved):
LGTM with a couple few comments about SemanticsBoolLiteralId

jonmeow (approved):

PR #2897, opened 2023-06-14
Parsing, semantic analysis, and lowering for `and`, `or`, `not`.

zygoloid:
Lowering for `and` and `or` is not yet complete because `Branch` lowering isn't done yet.


toolchain/lexer/tokenized_buffer.h
```diff
@@ -284,6 +284,10 @@ class TokenizedBuffer {
 
   [[nodiscard]] auto size() const -> int { return token_infos_.size(); }
 
+  [[nodiscard]] auto expected_parse_tree_size() const -> int {
```
jonmeow, line 287: Should this be something more like "parse_tree_size_upper_bound"? As we'd discussed, `and` is used in constraint contexts that may not require virtual nodes, and as a consequence this is more of an estimate than a precise value. Or do you figure to just leave it as `expected` until we empirically have an example which differs?

toolchain/parser/parse_tree.cpp
```diff
@@ -310,8 +310,8 @@ auto ParseTree::Verify() const -> ErrorOr<Success> {
     prev_index = n.index;
   }
 
-  if (!has_errors_ &&
-      static_cast<int32_t>(node_impls_.size()) != tokens_->size()) {
+  if (!has_errors_ && static_cast<int32_t>(node_impls_.size()) !=
+                          tokens_->expected_parse_tree_size()) {
     return Error(
         llvm::formatv("ParseTree has {0} nodes and no errors, but "
                       "TokenizedBuffer has {1} tokens.",
```
jonmeow: Please include the expected parse tree size here; the intent of this message is to note the actual and expected sizes for debug.

toolchain/semantics/testdata/operators/and_or.carbon
```diff
@@ -0,0 +1,106 @@
...
+fn F() -> bool { return true; }
+fn G() -> bool { return true; }
+
+fn And() -> bool {
+  return F() and G();
+}
+
+fn Or() -> bool {
+  return F() or G();
+}
```
jonmeow: It might help to put `and` and `or` tests in separate files, in order to make it easier to determine what IR each generates. I'm not seeing anything in this test that improves they be together.

Also, you might consider for simplicity using parameters, as you do in unary_op (consistency too?):

```
fn And(a: bool, b: bool) -> bool {
  return a and b;
}
```

The amount of IR generated would probably be significantly lower, making it easier to see specific nodes by offset.

toolchain/semantics/semantics_handle.cpp
```diff
@@ -193,21 +193,37 @@ auto SemanticsHandleInfixOperator(SemanticsContext& context,
...
       context.AddNodeAndPush(
           parse_node,
           SemanticsNode::BinaryOperatorAdd::Make(
               parse_node, context.semantics_ir().GetNode(lhs_id).type_id(),
               lhs_id, rhs_id));
       break;
+
+    case TokenKind::And:
+    case TokenKind::Or: {
+      auto rhs_block_id = context.node_block_stack().PopForAdd();
```
jonmeow: Please add a brief comment explaining the branch behavior. I think the binary op was straightforward, but this is probably complex enough that it could use a little explanation.

toolchain/semantics/semantics_handle.cpp
```diff
@@ -492,6 +526,46 @@ auto SemanticsHandleSelfValueIdentifier(SemanticsContext& context,
...
+  SemanticsNodeId branch_value_id = SemanticsNodeId::Invalid;
+  switch (auto token_kind = context.tokens().GetKind(token)) {
+    case TokenKind::And:
+      branch_value_id = cond_value_id;
+      break;
+
+    case TokenKind::Or:
+      branch_value_id = context.AddNode(SemanticsNode::UnaryOperatorNot::Make(
+          parse_node, context.semantics_ir().GetNode(cond_value_id).type_id(),
+          cond_value_id));
```
jonmeow: When you have a moment, I'd like to discuss whether we should be hewing closer to `br` and `phi` rather than the current `BranchIf` + `BranchWithArg` + `BlockArg` approach. Seeing this code made me wonder a bit why the output paths aren't just swapped.

I consider this non-blocking as it'd change existing merged code too, and perhaps it was already considered.

toolchain/semantics/semantics_handle.cpp
```diff
@@ -492,6 +526,46 @@ auto SemanticsHandleSelfValueIdentifier(SemanticsContext& context,
...
+  // Create a block for the right-hand side and for the continuation.
+  auto lhs_block_id = context.node_block_stack().PopForAdd();
+  auto end_block_id = context.node_block_stack().PushForAdd();
+  auto rhs_block_id = context.node_block_stack().PushForAdd();
+  context.AddNodeToBlock(
+      lhs_block_id,
+      SemanticsNode::BranchIf::Make(parse_node, rhs_block_id, branch_value_id));
+  context.AddNodeToBlock(
+      lhs_block_id, SemanticsNode::BranchWithArg::Make(parse_node, end_block_id,
+                                                       cond_value_id));
```
jonmeow: Minor, but I wonder if this'd be better branching with a constant true/false (known from the operator) rather than a value.

jonmeow (approved):
LGTM with a few minor comments to address, although looking more closely at the IR makes me want to take a moment to discuss the lowering plan for and/or.

toolchain/semantics/semantics_handle.cpp
zygoloid: I discussed this at some length with @chandlerc in the context of the `if` expressions PR, and he proposed we use block arguments rather than `phi` nodes. Definitely open to reconsidering it, but I'd want @chandlerc involved in the disucssion, and I think we shouldn't block on it for now.

We could swap over the branch destinations instead of generating a `not`, but that would require something like a `BranchIfWithArg` (which awkwardly needs three operands and therefore needs separate storage) or generation of an additional block. I think those are reasonable options to consider, but maybe it'd make some sense to wait until we see what we need from other branching constructs?

toolchain/lexer/tokenized_buffer.h
zygoloid: I think it's simpler to leave it as an exact expected count until we have a case that differs, even though we know that `where` expressions also reuse `and` tokens for another purpose -- maybe they'll need a virtual node too, or maybe we'll switch to a different syntax before we implement them :-)

toolchain/parser/parse_tree.cpp
zygoloid: Done.

toolchain/semantics/semantics_handle.cpp
zygoloid: I've added some explanation.

toolchain/semantics/semantics_handle.cpp
zygoloid: Yeah, I think this will make it easier to lower to better IR, if nothing else.

Done.

toolchain/semantics/testdata/operators/and_or.carbon
zygoloid: I've split the test.

I don't want to use parameters because I want to be sure that the arbitrary side-effects of the function  call end up in the separate block. (For the case of naming a parameter, it'd be reasonable -- and Clang does this for example -- to unconditionally evaluate the second operand if it has no side-effects, and I wanted the test to be robust against that change in output.)

toolchain/semantics/semantics_handle.cpp
jonmeow: I'm considering this resolved for now; probably see how lowering goes and also discuss with chandlerc.

PR #2900, opened 2023-06-14
Semantics handling for grouping parentheses.

zygoloid:


toolchain/semantics/semantics_handle.cpp
```diff
@@ -396,12 +396,17 @@ auto SemanticsHandleParameterListStart(SemanticsContext& context,
...
+  context.node_stack().PopAndDiscardSoloParseNode(
+      ParseNodeKind::ParenExpressionOrTupleLiteralStart);
+  context.node_stack().Push(parse_node, value_id);
+  return true;
 }
 
 auto SemanticsHandleParenExpressionOrTupleLiteralStart(
     SemanticsContext& context, ParseTree::Node parse_node) -> bool {
-  return context.TODO(parse_node, "HandleParenExpressionOrTupleLiteralStart");
+  context.node_stack().Push(parse_node);
```
zygoloid, line 408: We don't currently need to push this, but I think we will need it once we start handling tuple literals, so I'm adding it now.

jonmeow: FWIW, I support this just on the philosophy "1:1 node_stack entry per parse_node"

jonmeow (approved):

PR #2902, opened 2023-06-14
Treesitter parser

maan2003:
A push towards better editor support.
See utils/treesitter/README.md


utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+
+function repeat_sep1(thing, sep) {
+  return seq(thing, repeat(seq(sep, thing)));
+}
+
+function repeat_sep(thing, sep) {
+  return optional(seq(thing, repeat(seq(sep, thing))));
+}
+
+// follows toolchain/parser/precedence.cpp
```
josh11b: Should have some explanation about how this grammar will accept more code than Carbon since it uses precedence more.

utils/treesitter/src/scanner.c
```diff
@@ -0,0 +1,81 @@
...
+  return 0;  // zero bytes used to serialize
+}
+
+void tree_sitter_carbon_external_scanner_deserialize(void* /* payload */,
+                                                     const char* /* buffer */,
+                                                     unsigned /* length */) {}
+
+void tree_sitter_carbon_external_scanner_destroy(void* /* payload */) {}
+
+static bool is_operand_start(char c) {
```
josh11b: Is this somehow from the rules in https://github.com/carbon-language/carbon-lang/blob/trunk/proposals/p0601.md#whitespace ? Please explain the connection.

utils/treesitter/src/scanner.c
```diff
@@ -0,0 +1,81 @@
...
+    if (is_whitespace(lexer->lookahead) && whitespace) {
+      lexer->result_symbol = BINARY_STAR;
+    } else if (!whitespace && is_operand_start(lexer->lookahead)) {
+      lexer->result_symbol = BINARY_STAR;
+    } else {
+      lexer->result_symbol = POSTFIX_STAR;
+    }
+    return true;
+  } else {
+    // should never happen
```
josh11b: I would expect this to happen. Have you tried making this exit with an error?

utils/treesitter/test-toolchain.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
```
josh11b: The rest of the project is using bazel for testing.

utils/treesitter/src/scanner.c
```diff
@@ -0,0 +1,81 @@
...
+
+static bool is_operand_start(char c) {
+  return c == '(' || c == '[' || c == '{' || c == '\"' || c == '_' ||
+         (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') ||
+         (c >= '0' && c <= '9');
+}
+
+static bool is_whitespace(char c) { return c == ' ' || c == '\n'; }
+
+bool tree_sitter_carbon_external_scanner_scan(void* /* payload */,
```
josh11b: Please add comments to explain what is trying to be accomplished here.

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+const PREC = {
+  TermPrefix: 12,
+  TermPostfix: 12, // not in toolchain
+  NumericPrefix: 11,
+  NumericPostfix: 11,
+  Modulo: 9,
```
josh11b: Doesn't seem to be used.

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+              field('left', $._expression),
+              field('operator', operator),
+              field('right', $._expression)
+            )
+          )
+        )
+      );
+    },
+
+    as_expression: ($) => prec.left(seq($._expression, 'as', $._expression)),
```
josh11b, line 253: Why specify left associative? https://github.com/carbon-language/carbon-lang/tree/trunk/docs/design/expressions says it is non-associative.

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+  LogicalAnd: 2,
+  LogicalOr: 2,
+  Where: 1,
+  If: 1,
+  SimpleAssignment: 0,
+  CompoundAssignment: 0,
+};
+
+module.exports = grammar({
+  name: 'carbon',
```
josh11b, line 42: I don't see `word:` here, is the problem from https://tree-sitter.github.io/tree-sitter/creating-parsers#keywords a concern?

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+
+  conflicts: ($) => [
+    [$.paren_pattern, $.paren_expression],
+    [$.struct_literal, $.struct_type_literal],
+  ],
+  extras: ($) => [/\s/, $.comment],
+
+  // NOTE: must match the order in src/scanner.c, the names don't matter
```
josh11b:
```suggestion
  // NOTE: must match the order in src/scanner.c, names are not used for matching
```

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+        optional($.package_directive),
+        repeat($.import_directive),
+        repeat($.declaration)
+      ),
+    comment: ($) => token(seq('//', /.*/)),
+    ident: ($) => /[A-Za-z_][A-Za-z0-9_]*/,
+    bool_literal: ($) => choice('true', 'false'),
+    integer_literal: ($) => /[0-9]+/,
+    float_literal: ($) => /[0-9\.]+/,
+    sized_type_literal: ($) => /[iuf][1-9][0-9]*/,
```
josh11b: What is this? I don't think this is part of Carbon.

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+      seq(
+        optional($.package_directive),
+        repeat($.import_directive),
+        repeat($.declaration)
+      ),
+    comment: ($) => token(seq('//', /.*/)),
+    ident: ($) => /[A-Za-z_][A-Za-z0-9_]*/,
+    bool_literal: ($) => choice('true', 'false'),
+    integer_literal: ($) => /[0-9]+/,
+    float_literal: ($) => /[0-9\.]+/,
```
josh11b: https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/lexical_conventions/numeric_literals.md#real-number-literals
- Must start with a digit
- Exactly one `.`
- May optionally be followed by `e[+-]?[0-9]+`
- Missing support for digit separators https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/lexical_conventions/numeric_literals.md#digit-separators

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+    source_file: ($) =>
+      seq(
+        optional($.package_directive),
+        repeat($.import_directive),
+        repeat($.declaration)
+      ),
+    comment: ($) => token(seq('//', /.*/)),
+    ident: ($) => /[A-Za-z_][A-Za-z0-9_]*/,
+    bool_literal: ($) => choice('true', 'false'),
+    integer_literal: ($) => /[0-9]+/,
```
josh11b: - Missing support for hex and binary integer literals https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/lexical_conventions/numeric_literals.md#integer-literals
- Missing support for digit separators https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/lexical_conventions/numeric_literals.md#digit-separators

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+      ),
+    struct_type_literal: ($) =>
+      seq(
+        '{',
+        repeat_sep(seq($.designator, ':', $._expression), ','),
+        optional(','),
+        '}'
+      ),
+
+    builtin_type: ($) => choice('Self', 'String', 'bool', 'type', 'i32'),
```
josh11b: More integer and floating point types than that:
https://github.com/carbon-language/carbon-lang/blob/trunk/proposals/p2015.md#syntax
or https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/expressions/literals.md#numeric-type-literals

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+        $.bool_literal,
+        $.integer_literal,
+        $.string_literal,
+        $.sized_type_literal,
+        $.float_literal,
+        $.struct_literal,
+        $.struct_type_literal
+      ),
+
+    _binding_lhs: ($) => seq(optional('addr'), choice($.ident, '_')),
```
josh11b: I believe `addr` may only precede `self`, so I think this would be better as a `choice` than a `seq`.

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+      );
+    },
+
+    as_expression: ($) => prec.left(seq($._expression, 'as', $._expression)),
+
+    ref_expression: ($) => prec.right(PREC.TermPrefix, seq('&', $._expression)),
+    deref_expression: ($) =>
+      prec.right(PREC.TermPrefix, seq('*', $._expression)),
+    fn_type_expression: ($) =>
+      prec.left(seq('fn', $.tuple, '->', $._expression)),
```
josh11b: I don't think this is part of Carbon, see https://github.com/carbon-language/carbon-lang/pull/2875

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+        )
+      ),
+    where_clause: ($) =>
+      prec(
+        PREC.Where,
+        choice(
+          seq($._expression, '==', $._expression),
+          seq($._expression, 'impls', $._expression),
+          seq($.designator, '=', $._expression),
+          prec.left(seq($.where_clause, choice('or', 'and'), $.where_clause))
```
josh11b: Only `and`, no `or` here.

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+    _assign_operator: ($) =>
+      choice('=', '+=', '/=', '*=', '%=', '-=', '&=', '|=', '^=', '<<=', '>>='),
+
+    match_statement: ($) =>
+      seq('match', '(', $._expression, ')', '{', $.clause_list, '}'),
+    returned_var_statement: ($) =>
+      seq(
+        'returned',
+        'var',
+        $._variable_declaration_inner,
```
josh11b: This should match regular `var` statements.

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+        $.ref_expression,
+        $.deref_expression,
+        $.where_expression
+      ),
+
+    _variable_declaration_inner: ($) => seq($.ident, ':', $._expression),
+    variable_declaration: ($) =>
+      choice(
+        seq('var', $.pattern, optional(seq('=', $._expression)), ';'),
+        seq('let', $.pattern, '=', $._expression, ';')
```
josh11b: I think these should both be `_non_expression_pattern` instead of `pattern`.

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+      ),
+
+    clause: ($) =>
+      seq(choice(seq('case', $.pattern), 'default'), '=>', $.block),
+    clause_list: ($) => seq($.clause_list, $.clause),
```
josh11b: Move closer to use, I think these are only used with `match_statement`

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+
+    assign_statement: ($) =>
+      seq($._expression, $._assign_operator, $._expression, ';'),
+
+    _assign_operator: ($) =>
+      choice('=', '+=', '/=', '*=', '%=', '-=', '&=', '|=', '^=', '<<=', '>>='),
+
+    match_statement: ($) =>
+      seq('match', '(', $._expression, ')', '{', $.clause_list, '}'),
+    returned_var_statement: ($) =>
```
josh11b: Blank lines are used inconsistently. Either always have blank lines between rules, or only omit them between closely related rules. `assign_statement` is more closely related to `_assign_operator` than `match_statement` is to `returned_var_statement`.

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+        $.as_expression,
+        $.paren_expression,
+        $.postfix_expression,
+        $.call_expression,
+        $.ref_expression,
+        $.deref_expression,
+        $.where_expression
+      ),
+
+    _variable_declaration_inner: ($) => seq($.ident, ':', $._expression),
```
josh11b: Move close to use; after my suggestions I think this is only used in `for` statements.

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+ * Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+ * Exceptions. See /LICENSE for license information.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+function repeat_sep1(thing, sep) {
+  return seq(thing, repeat(seq(sep, thing)));
+}
+
+function repeat_sep(thing, sep) {
```
josh11b: Looks like in all uses, sep is `,` and this is followed by an optional `,`. Perhaps a `comma_sep` function would be better?

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+    namespace_declaration: ($) => seq('namespace', $.declared_name, ';'),
+    alias_declaration: ($) =>
+      seq('alias', $.declared_name, '=', $._expression, ';'),
+    type_params: ($) => $.tuple_pattern,
+    interface_body_item: ($) =>
+      choice(
+        $.function_declaration,
+        seq('let', $.generic_binding, ';'),
+        seq('extends', $._expression, ';'),
+        seq('impl', $._expression, 'as', $._expression, ';')
```
josh11b: These rules have changed by #2760

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+    class_body_item: ($) => $.declaration,
+    class_body: ($) => seq('{', repeat($.class_body_item), '}'),
+    class_declaration: ($) =>
+      seq(
+        optional(choice('base', 'abstract')),
+        'class',
+        $.declared_name,
+        optional($.deduced_params),
+        optional($.type_params),
+        optional(seq('extends', $._expression)),
```
josh11b: This has also been changed by #2760

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+      ),
+    constraint_declaration: ($) =>
+      seq(
+        'constraint',
+        $.declared_name,
+        optional($.deduced_params),
+        optional($.type_params),
+        $.interface_body
+      ),
+    class_body_item: ($) => $.declaration,
```
josh11b: More choices here introduced in #2760

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,407 @@
...
+        $.class_declaration,
+        ';' // empty declaration
+      ),
+
+    api_or_impl: ($) => choice('api', 'impl'),
+    library_path: ($) => seq('library', $.string_literal),
+    package_directive: ($) =>
+      seq('package', $.ident, optional($.library_path), $.api_or_impl, ';'),
+    import_directive: ($) =>
+      seq('import', $.ident, optional($.library_path), ';'),
```
josh11b: Move these close to the `source_file` rule

utils/treesitter/queries/highlights.scm
```diff
@@ -0,0 +1,112 @@
...
+(function_declaration (declared_name (ident) @function))
+(call_expression (ident) @function)
+
+; upper case => type
+((ident) @type
+  (#match? @type "^[A-Z]"))
+
+; lower case => variable
+((ident) @variable
+  (#match? @variable "^[a-z_]"))
```
josh11b, line 28: The case is not a reliable indicator here. Would be much better to do what you do with functions: identifiers from patterns should be highlighted like variables, identifiers from class/interface/constraint/etc. declarations should be highlighted like types.

utils/treesitter/queries/highlights.scm
```diff
@@ -0,0 +1,112 @@
...
+  "fn"
+  "alias"
+  "interface"
+  "extends"
+  "constraint"
+  "class"
+  "package"
+  "import"
+  "api"
+  "library"
```
josh11b: Going to be hard to keep these up to date unless they are alphabetized.

utils/treesitter/grammar.js
maan2003: tried, this failed
https://github.com/carbon-language/carbon-lang/blob/b908c6e274a418952b0a8e9e5ce2dd44469e22b2/toolchain/parser/testdata/generics/generic_params/template_addr.carbon#L20

utils/treesitter/grammar.js
maan2003: I think this is explorer only. and did s/fn/__Fn/
https://github.com/carbon-language/carbon-lang/blob/b908c6e274a418952b0a8e9e5ce2dd44469e22b2/explorer/syntax/parser.ypp#L511-L516

utils/treesitter/grammar.js
josh11b: That is not legal Carbon, as far as I'm aware.

bazel/cc_toolchains/clang_cc_toolchain_config.bzl
```diff
@@ -125,7 +125,7 @@ def _impl(ctx):
                 flag_groups = ([
                     flag_group(
                         flags = [
-                            "-Werror",
+                            # "-Werror",
```
maan2003: treesitter build was failing due to unused variable in generate code.

utils/treesitter/queries/highlights.scm
maan2003: Good idea. I would like to do it in a follow up.

utils/treesitter/test-toolchain.sh
maan2003: I have updated this file to use bazel internally. But I could not convert this into bazel.

bazel/cc_toolchains/clang_cc_toolchain_config.bzl
josh11b: Can this be a comment in the file itself? Even better would be to disable just that specific warning for those specific files.

utils/treesitter/.gitignore
```diff
@@ -0,0 +1,11 @@
+# Part of the Carbon Language project, under the Apache License v2.0 with LLVM
```
josh11b, line 1: A goal of switching to bazel would be to drop the need for this file.

utils/treesitter/README.md
```diff
@@ -0,0 +1,25 @@
+<!--
+Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+Exceptions. See /LICENSE for license information.
+SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+-->
+
+# Tree-sitter grammar for Carbon
+
+## Installation
```
josh11b:
```suggestion
## Editor Installation
```

utils/treesitter/README.md
```diff
@@ -0,0 +1,25 @@
+<!--
+Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+Exceptions. See /LICENSE for license information.
+SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+-->
+
+# Tree-sitter grammar for Carbon
```
josh11b, line 7: Explain what this is and the things you can use it for.

utils/treesitter/build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
```
josh11b: Can we get rid of this file? If not, please include documentation saying what it is for.

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,507 @@
...
+ */
+
+// This grammar is more permissive than toolchain because it is geared towards editor use.
+
+function repeat_sep1(thing, sep) {
+  return seq(thing, repeat(seq(sep, thing)));
+}
+
+function comma_sep(thing) {
+  // Trailing comma is only allowed if there is atleast one element.
```
josh11b:
```suggestion
  // Trailing comma is only allowed if there is at least one element.
```

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,507 @@
+/*
+ * Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+ * Exceptions. See /LICENSE for license information.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ */
+
+// This grammar is more permissive than toolchain because it is geared towards editor use.
```
josh11b: Seems like there are lines >80 characters in this file. I guess our formatting checks don't include a test for that for js, even though we do have it for other languages?

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,507 @@
...
+        $.numeric_type_literal,
+        $.string_literal,
+        $.struct_literal,
+        $.struct_type_literal
+      ),
+
+    _binding_lhs: ($) => choice($.ident, '_', seq('addr', 'self')),
+
+    paren_pattern: ($) =>
+      seq('(', comma_sep($._non_expression_pattern, ','), ')'),
```
josh11b:
```suggestion
      seq('(', comma_sep($._non_expression_pattern), ')'),
```

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,507 @@
...
+      ),
+
+    pattern: ($) => choice($._non_expression_pattern, $._expression),
+
+    unary_prefix_expression: ($) => {
+      const table = [
+        [PREC.NumericPrefix, '-'],
+        [PREC.NumericPrefix, '--'],
+        [PREC.NumericPrefix, '++'],
+        [PREC.BitwisePrefix, '~'],
```
josh11b: `~` is not (yet) a unary operator. bitwise negation is spelled `^x`, see https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/expressions/bitwise.md

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,507 @@
...
+
+    if_expression: ($) =>
+      prec(
+        PREC.IfExpression,
+        seq('if', $._expression, 'then', $._expression, 'else', $._expression)
+      ),
+
+    paren_expression: ($) => seq('(', comma_sep($._expression), ')'),
+
+    tuple: ($) => $.paren_expression,
```
josh11b: Why introduce this rule instead of just using `paren_expression`?

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,507 @@
...
+      prec(PREC.TypePostfix, seq($._expression, $.postfix_star)),
+
+    _expression: ($) =>
+      choice(
+        $.ident,
+        $.pointer_expression,
+        $.index_expression,
+        $.fn_type_expression,
+        $.designator,
+        seq('.', 'Self'),
```
josh11b: FYI, I think this is only allowed in a where clause.

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,507 @@
...
+    pointer_expression: ($) =>
+      prec(PREC.TypePostfix, seq($._expression, $.postfix_star)),
+
+    _expression: ($) =>
+      choice(
+        $.ident,
+        $.pointer_expression,
+        $.index_expression,
+        $.fn_type_expression,
+        $.designator,
```
josh11b: Is this needed? I don't think designators are generally allowed in expressions at the moment.

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,507 @@
...
+        seq('.', 'Self'),
+        $.literal,
+        $.builtin_type,
+        $.paren_expression,
+        $.array_literal,
+        $.if_expression,
+        $.binary_expression,
+        $.unary_prefix_expression,
+        $.as_expression,
+        $.paren_expression,
```
josh11b, line 325: appears twice in this list. Maybe sort alphabetically so it is easier?

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,507 @@
...
+        $.while_statement,
+        $.break_statement,
+        $.continue_statement,
+        $.return_statement,
+        $.for_statement
+      ),
+
+    _statement_list: ($) => seq($.statement, repeat($.statement)),
+
+    block: ($) => seq('{', optional($._statement_list), '}'),
```
josh11b: Simpler?

```suggestion
    _statement_list: ($) => repeat($.statement),

    block: ($) => seq('{', $._statement_list, '}'),
```

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,507 @@
...
+    declared_name: ($) => repeat_sep1($.ident, '.'),
+
+    generic_binding: ($) =>
+      seq(optional('template'), $.ident, ':!', $._expression),
+
+    deduced_param: ($) =>
+      choice(
+        $.generic_binding,
+        seq(optional('addr'), $.ident, ':', $._expression)
+      ),
```
josh11b: Some redundancy with `_non_expression_pattern`

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,507 @@
...
+      choice(
+        $.bool_literal,
+        $.numeric_literal,
+        $.numeric_type_literal,
+        $.string_literal,
+        $.struct_literal,
+        $.struct_type_literal
+      ),
+
+    _binding_lhs: ($) => choice($.ident, '_', seq('addr', 'self')),
```
josh11b: `addr self` and `self` are only allowed in the deduced parameter list, which doesn't use this rule.

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,507 @@
...
+
+    deduced_param: ($) =>
+      choice(
+        $.generic_binding,
+        seq(optional('addr'), $.ident, ':', $._expression)
+      ),
+
+    deduced_params: ($) => seq('[', comma_sep($.deduced_param), ']'),
+
+    tuple_pattern: ($) => $.paren_pattern,
```
josh11b: Why define `tuple_pattern` and not just use `paren_pattern` directly?

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,507 @@
...
+
+    interface_body: ($) => seq('{', repeat($.interface_body_item), '}'),
+
+    interface_declaration: ($) =>
+      seq(
+        'interface',
+        $.declared_name,
+        optional($.deduced_params),
+        optional($.type_params),
+        $.interface_body
```
josh11b: Explorer only allows an interface body here, but the design doc also allows `;` here, for forward declarations.

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,507 @@
...
+      ),
+
+    constraint_declaration: ($) =>
+      seq(
+        'constraint',
+        $.declared_name,
+        optional($.deduced_params),
+        optional($.type_params),
+        $.interface_body
```
josh11b: `;` is allowed instead of an interface body here too.

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,507 @@
...
+      ),
+
+    impl_body_item: ($) => choice($.function_declaration, $.alias_declaration),
+
+    impl_body: ($) => seq('{', $.impl_body_item, '}'),
+
+    impl_declaration: ($) =>
+      seq(
+        'impl',
+        optional(seq('forall', $.deduced_params)),
```
josh11b, line 465: optional type expression between `forall` and `as`

utils/treesitter/queries/highlights.scm
josh11b: Please include a TODO comment to indicate what is expected to change.

utils/treesitter/queries/highlights.scm
```diff
@@ -0,0 +1,124 @@
+; Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+; Exceptions. See /LICENSE for license information.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
```
josh11b, line 4: I'm not sure how this file relates to the other files. Could you please add comments explaining?

utils/treesitter/src/scanner.c
```diff
@@ -0,0 +1,82 @@
...
+
+static bool is_whitespace(char c) { return c == ' ' || c == '\n'; }
+
+// https://tree-sitter.github.io/tree-sitter/creating-parsers#external-scanners
+//
+// > If a token in the externals array is valid at a given position in the
+// > parse, the external scanner will be called first before anything else is
+// > done.
+//
+// > But if the external scanner may return false and in this case Tree-sitter
```
josh11b:
```suggestion
// > But the external scanner may return false and in this case Tree-sitter
```

utils/treesitter/test.cpp
```diff
@@ -0,0 +1,51 @@
...
+    std::cerr << "Usage: treesitter_carbon_tester <file>\n";
+    return 2;
+  }
+
+  std::string str = ReadFile(std::string(argv[1]));
+
+  auto* parser = ts_parser_new();
+  ts_parser_set_language(parser, tree_sitter_carbon());
+
+  auto* tree = ts_parser_parse_string(parser, nullptr, str.c_str(), str.size());
```
josh11b: If you are passing in the size, you should not need the null terminator that `c_str` gives you.
```suggestion
  auto* tree = ts_parser_parse_string(parser, nullptr, str.data(), str.size());
```

utils/treesitter/.gitignore
maan2003: it is ok to add these if someone is using treesitter-cli to build (like me) and editors might also create these files during installation.

utils/treesitter/grammar.js
maan2003: rules are not allowed to match to empty string in treesitter. I think this is because of the error recovery mechanism.

utils/treesitter/grammar.js
maan2003: I think this is small enough to duplicate and make it more correct.

utils/treesitter/BUILD
```diff
@@ -0,0 +1,40 @@
...
+        "testdata/**/fail_*",
+        "testdata/function/declaration/addr.carbon",
+        "testdata/generics/generic_params/template_addr.carbon",
+        "testdata/operators/recover_infix_uneven_space_before.carbon",
+        "testdata/operators/recover_postfix_space_surrounding.carbon",
+    ],
+)
+
+cc_test(
+    name = "tests",
```
josh11b: Usually the name of the test would be expected to match the name of the source file, unless there are multiple tests sharing the same source file.

utils/treesitter/BUILD
```diff
@@ -0,0 +1,40 @@
...
+    srcs = ["test.cpp"],
```
josh11b: Could you rename this file to be more specific?

utils/treesitter/BUILD
```diff
@@ -0,0 +1,40 @@
...
+        ":treesitter",
+    ],
+)
+
+tests = glob(
+    ["testdata/**/*.carbon"],
+    exclude = [
+        "testdata/**/fail_*",
+        "testdata/function/declaration/addr.carbon",
+        "testdata/generics/generic_params/template_addr.carbon",
```
josh11b: This does not match the "current results" in the PR summary

utils/treesitter/BUILD
```diff
@@ -0,0 +1,40 @@
...
+        "testdata/operators/recover_infix_uneven_space_before.carbon",
+        "testdata/operators/recover_postfix_space_surrounding.carbon",
```
josh11b: FYI, If this list grows, it would be reasonable to exclude all the tests starting with `recover_`.

utils/treesitter/BUILD
```diff
@@ -0,0 +1,40 @@
...
+
+cc_binary(
+    name = "test_runner",
+    srcs = ["test.cpp"],
+    deps = [
+        ":treesitter",
+    ],
+)
+
+tests = glob(
```
josh11b: Looks like these are specifically tests from the toolchain. Would be good to (a) name this so that is clear, and (b) also test using the explorer tests.

I tried running this with the explorer test files, but it seemed to get a lot of errors. Is it future work? If so, would be good to have a comment about that.

utils/treesitter/testdata
```diff
@@ -0,0 +1 @@
+../../toolchain/parser/testdata
```
josh11b: Rather than creating a symlink, can you make https://github.com/carbon-language/carbon-lang/blob/trunk/toolchain/parser/BUILD export a target that this directory uses to get those test files?

utils/treesitter/.gitignore
josh11b: Then please include a comment explaining why this file is needed.

utils/treesitter/grammar.js
josh11b: Then how about this:

```suggestion
    block: ($) => seq('{', repeat($.statement), '}'),
```

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,508 @@
...
+
+    declared_name: ($) => repeat_sep1($.ident, '.'),
+
+    generic_binding: ($) =>
+      seq(optional('template'), $.ident, ':!', $._expression),
+
+    deduced_param: ($) =>
+      choice(
+        $.generic_binding,
+        seq(optional('addr'), $.ident, ':', $._expression)
```
josh11b:
```suggestion
        seq(optional('addr'), `self`, ':', $._expression)
```

utils/treesitter/queries/highlights.scm
```diff
@@ -0,0 +1,129 @@
...
+  "or"
+  ; "override"
+  "package"
+  ; "partial"
+  ; "private"
+  ; "protected"
+  "require"
+  "return"
+  "returned"
+  "Self"
```
josh11b, line 124:
```suggestion
  "self"
  "Self"
```

utils/treesitter/test.cpp
```diff
@@ -0,0 +1,64 @@
...
+// Reads a file to string.
+static auto ReadFile(std::filesystem::path path) -> std::string {
+  std::ifstream file(path);
+  std::stringstream buffer;
+  buffer << file.rdbuf();
+  file.close();
+  return buffer.str();
+}
+
+auto main(int argc, char** argv) -> int {
```
jonmeow: This should probably be using gtest and (in particular) //testing/file_test to make better use of standard infrastructure. Please at least add a TODO.

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,508 @@
...
+        $.deref_expression,
+        $.fn_type_expression,
+        $.ident,
+        $.if_expression,
+        $.index_expression,
+        $.literal,
+        $.paren_expression,
+        $.pointer_expression,
+        $.postfix_expression,
+        $.ref_expression,
```
josh11b, line 328:
```suggestion
        $.ref_expression,
        `self`,
```

utils/treesitter/testdata
jonmeow: Please don't include tests by symlink. Any tests should probably instead be exposed as a target instead, to make dependencies clearer.

However, the toolchain's parser is also not a great baseline to compare with -- we're going to make pragmatic choices about whether diagnostics belong in the lexer, parser, or semantics. Also, as the support expands and more invalid syntax use-cases are handled, it should be expected that the parser will skew further from tree-sitter behavior. If you're looking for better tests to compare with, the explorer might be a better choice because the passing tests there are all going to be semantically valid.

jonmeow (commented):
Sorry, a couple drop-in comments due to the questions raised by josh11b [on discord](https://discord.com/channels/655572317891461132/655578254970716160/1126264600338378782).

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,508 @@
...
+      seq('import', $.ident, optional($.library_path), ';'),
+
+    comment: ($) => token(seq('//', /.*/)),
+
+    ident: ($) => /[A-Za-z_][A-Za-z0-9_]*/,
+
+    bool_literal: ($) => choice('true', 'false'),
+
+    numeric_literal: ($) => {
+      // This is using variables because rules are not allowed in token.immediate and token.
```
josh11b: Still seem to have long lines in this file.

utils/treesitter/grammar.js
```diff
@@ -0,0 +1,554 @@
...
+        $.literal,
+        $.paren_expression,
+        $.pointer_expression,
+        $.postfix_expression,
+        $.ref_expression,
+        $.unary_prefix_expression,
+        $.where_expression,
+        'self',
+        '.Self',
+        $.designator
```
maan2003, line 334: These were added back again. These are used in where_clause's expression

utils/treesitter/grammar.js
maan2003: I think javascript formatter has greater column limit by default. I have manually wrapped these for now.

utils/treesitter/testdata
maan2003: added as TODO for a follow up in utils/treesitter/BUILD

josh11b: Looks like there is already a `filegroup` with the explorer's testdata. It is declared: https://github.com/carbon-language/carbon-lang/blob/trunk/explorer/BUILD#L78
and used: https://github.com/carbon-language/carbon-lang/blob/c0d18a62ebddfbfb97e64dc6b935a37abe885bf3/explorer/fuzzing/BUILD#L35

You should be able to copy that example, and just add another package to the visibility list

maan2003: I don't understand how to filter the `carbon_files` filegroup in treesitter/BUILD (to exclude strings and fail_ tests)  

closest result I found was https://groups.google.com/g/bazel-discuss/c/U98eiNCwxIA, which is unresolved

josh11b: One solution is to make a new filegroup in instead of `carbon_files` in the explorer/BUILD that does the exclusions.

maan2003: I didn't like that earlier because it felt like moving treesitter implementation detail into explorer BUILD.

Now, I think it is fine because explorer controls what testdata to give to treesitter.

docs/design/lexical_conventions/words.md
```diff
@@ -85,6 +89,7 @@ The following words are interpreted as keywords:
 -   `return`
 -   `returned`
 -   `Self`
+-   `self`
```
josh11b: Skip this until #2984 is resolved.

utils/treesitter/grammar.js
```diff
@@ -286,53 +294,55 @@ module.exports = grammar({
         choice(
           seq($._expression, '==', $._expression),
           seq($._expression, 'impls', $._expression),
-          seq($.designator, '=', $._expression),
+          seq($._expression, '=', $._expression),
+          // FIXME: logical and conflicts with this
```
josh11b: Is this a FIXME that will be fixed before submit, or a TODO that will be fixed later?

utils/treesitter/grammar.js
josh11b: Will they be removed once where clauses get their own rules instead of reusing expression?

utils/treesitter/grammar.js
```diff
@@ -401,13 +401,12 @@ module.exports = grammar({
     deduced_param: ($) =>
       choice(
         $.generic_binding,
-        seq(optional('addr'), $.ident, ':', $._expression)
+        seq($.ident, ':', $._expression),
+        seq('addr', 'self', ':', $._expression)
```
josh11b: I think at the moment we only allow generic bindings or `self`, so this could just be:

```suggestion
        seq(optional('addr'), 'self', ':', $._expression)
```

utils/treesitter/grammar.js
```diff
@@ -475,11 +475,23 @@ module.exports = grammar({
...
 
+    destructor_declaration: ($) =>
+      seq(
+        optional(choice('virtual', 'impl')),
+        'destructor',
+        optional($.deduced_params),
+        choice($.block, ';')
+      ),
+
+    mix_declaration: ($) => seq('__mix', $._expression, ';'),
```
josh11b: Mixins are not officially in carbon (hence the `__` before `mix`), so I'd be find dropping this rule and excluding the mixin tests (your choice).

utils/treesitter/test_runner.cpp
```diff
@@ -35,12 +36,19 @@ auto main(int argc, char** argv) -> int {
   ts_parser_set_language(parser, tree_sitter_carbon());
 
   std::vector<std::string> failed;
+  std::vector<std::string> skipped;
   for (int i = 1; i < argc; i++) {
     std::string file_path = argv[i];
     std::string source = ReadFile(file_path);
 
+    // `and` in where clauses is not parsed correctly.
```
josh11b, line 44: Please include a TODO to remove this temporary workaround.

utils/treesitter/grammar.js
maan2003: will be fixed later

utils/treesitter/grammar.js
maan2003: Not sure yet

utils/treesitter/grammar.js
maan2003: it was simple enough to implement. It is nice to highlight all explorer tests correctly.

Same is case for `__match_first`, but that is also used in prelude.

utils/treesitter/grammar.js
```diff
@@ -402,7 +402,7 @@ module.exports = grammar({
       choice(
         $.generic_binding,
         seq($.ident, ':', $._expression),
```
josh11b: This line may now be deleted -- we don't allow `:` in the deduced parameter list at the moment, except after `self`.

utils/treesitter/grammar.js
josh11b:
```suggestion
          // TODO: Fix conflict with logical and
```

utils/treesitter/grammar.js
josh11b:
```suggestion
        // TODO: Remove these two once `where` clauses don't use the expression rule
        '.Self',
        $.designator
```

utils/treesitter/grammar.js
josh11b: Okay, just know that the explorer syntax for mixins is not the planned mixin syntax for Carbon, which is described here: https://github.com/carbon-language/carbon-lang/blob/trunk/proposals/p2760.md#future-work-mixins
You should have some comment on your mixin rules explaining the situation.

josh11b (approved):

PR #2904, opened 2023-06-15
Lowering of `Branch` / `BranchIf` / `BranchWithArg` / `BlockArg`.

zygoloid:
This gives us complete lowering of `if` expressions plus `and` and `or`.


toolchain/lowering/lowering_block_worklist.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include "llvm/ADT/DenseSet.h"
+#include "llvm/ADT/SmallVector.h"
+#include "toolchain/semantics/semantics_node.h"
+
+namespace Carbon {
+
+// A worklist for blocks that need to be lowered. Blocks are popped from the
+// worklist in the order in which they are first pushed, in a depth-first
+// traversal.
```
jonmeow: I'm confused by this description. I read the description as FIFO ("in which they are first pushed"), and I'm just not sure what you mean by "in a depth-first traversal". But obviously with the `reverse` you intend something else here, and it's not clear to me that you're getting the output order that you might want.

Given the logic, I think it'd help if you add tests for nested control flow, such as `a and ((b and c) and c)`, to check the resulting block order. Right now you only test a single level, unless I'm missing something.

But looking at this, I'm just confused what the `reverse` buys:

```
// CHECK:STDOUT: define i1 @And(i1 %a, i1 %b, i1 %c, i1 %d) {
// CHECK:STDOUT:   br i1 %a, label %1, label %2
// CHECK:STDOUT:
// CHECK:STDOUT: 1:                                                ; preds = %0
// CHECK:STDOUT:   br i1 %b, label %4, label %5
// CHECK:STDOUT:
// CHECK:STDOUT: 2:                                                ; preds = %8, %0
// CHECK:STDOUT:   %3 = phi i1 [ false, %0 ], [ %9, %8 ]
// CHECK:STDOUT:   ret i1 %3
// CHECK:STDOUT:
// CHECK:STDOUT: 4:                                                ; preds = %1
// CHECK:STDOUT:   br label %5
// CHECK:STDOUT:
// CHECK:STDOUT: 5:                                                ; preds = %4, %1
// CHECK:STDOUT:   %6 = phi i1 [ false, %1 ], [ %c, %4 ]
// CHECK:STDOUT:   br i1 %6, label %7, label %8
// CHECK:STDOUT:
// CHECK:STDOUT: 7:                                                ; preds = %5
// CHECK:STDOUT:   br label %8
// CHECK:STDOUT:
// CHECK:STDOUT: 8:                                                ; preds = %7, %5
// CHECK:STDOUT:   %9 = phi i1 [ false, %5 ], [ %d, %7 ]
// CHECK:STDOUT:   br label %2
// CHECK:STDOUT: }
```

Functionally I think this works, but what's the benefit of the `reverse` if block order doesn't reflect code execution order? A simple LIFO stack would probably be poorly ordered, but avoiding the reverse.

Note, an approach that'd emit in a different order would be just going FIFO on a vector without popping, just a cursor (building up entries until the end of the function, which could be erased if needed in the future). But I'm not sure that's really better.

toolchain/lowering/lowering_block_worklist.h
```diff
@@ -0,0 +1,44 @@
...
+
+  // Pop the next block to lower.
+  auto Pop() -> SemanticsNodeBlockId {
+    std::reverse(worklist_.begin() + size_after_last_pop_, worklist_.end());
+    SemanticsNodeBlockId result = worklist_.pop_back_val();
+    size_after_last_pop_ = worklist_.size();
+    return result;
+  }
+
+  auto Empty() -> bool { return worklist_.empty(); }
```
jonmeow:
```suggestion
  auto empty() -> bool { return worklist_.empty(); }
```

I think lowercase would be more consistent with the property naming conventions.

toolchain/lowering/lowering_function_context.h
```diff
@@ -48,13 +71,26 @@ class LoweringFunctionContext {
     return lowering_context_->GetType(type_id);
   }
 
+  // Create a synthetic block that corresponds to no SemanticsNodeBlockId. Such
+  // a block should only ever have a single predecessor, and is used when we
+  // need multiple `llvm::BasicBlock`s to model the linear control flow in a
+  // single SemanticsIR block.
+  auto CreateSyntheticBlock() -> llvm::BasicBlock*;
+
+  // Determine whether block_ is the most-recently-created synthetic block.
```
jonmeow:
```suggestion
  // Determine whether block_ is the most recently created synthetic block.
```

I'm declaring war on unnecessary hyphens. :-P

toolchain/lowering/lowering_function_context.h
```diff
@@ -48,13 +71,26 @@ class LoweringFunctionContext {
...
+  auto IsCurrentSyntheticBlock(llvm::BasicBlock* block_) -> bool {
+    return synthetic_block_ == block_;
```
jonmeow:
```suggestion
  auto IsCurrentSyntheticBlock(llvm::BasicBlock* block) -> bool {
    return synthetic_block_ == block;
```

toolchain/lowering/lowering_function_context.h
```diff
@@ -68,6 +104,15 @@ class LoweringFunctionContext {
...
   llvm::IRBuilder<> builder_;
 
+  // A worklist for blocks we need to lower.
+  LoweringBlockWorklist block_worklist_;
+
+  // Maps a function's SemanticsIR blocks to lowered blocks.
+  llvm::DenseMap<SemanticsNodeBlockId, llvm::BasicBlock*> blocks_;
+
+  // The synthetic block we most recently created.
+  std::optional<llvm::BasicBlock*> synthetic_block_ = std::nullopt;
```
jonmeow: Is there any overhead when using optional? Should this be `nullptr`?

toolchain/lowering/lowering_function_context.h
```diff
@@ -21,6 +24,26 @@ class LoweringFunctionContext {
   explicit LoweringFunctionContext(LoweringContext& lowering_context,
                                    llvm::Function* function);
 
+  // Returns a basic block corresponding to the start of the given semantics
+  // block.
+  auto GetBlock(SemanticsNodeBlockId block_id) -> llvm::BasicBlock*;
```
jonmeow, line 29: WDYT if we used references to emphasize non-nullness, including things like GetValue?

toolchain/lowering/lowering_function_context.cpp
```diff
@@ -14,11 +14,39 @@ LoweringFunctionContext::LoweringFunctionContext(
...
+  llvm::BasicBlock* block = GetBlock(block_id);
+
+  // Find the existing phi, if any.
+  auto phis = block->phis();
+  if (!phis.empty()) {
+    return &*phis.begin();
+  }
+
+  // The number of predecessor slots to reserve.
+  unsigned NumReservedPredecessors = 2;
```
jonmeow:
```suggestion
  static constexpr unsigned NumReservedPredecessors = 2;
```

I'll bet this is a clang-tidy warning right now.

toolchain/lowering/lowering_function_context.cpp
```diff
@@ -14,11 +14,39 @@ LoweringFunctionContext::LoweringFunctionContext(
...
+  auto* phi = llvm::PHINode::Create(GetType(type_id), NumReservedPredecessors);
+  phi->insertInto(block, block->begin());
```
jonmeow, line 58: Shouldn't the block always be empty when this is inserted?

toolchain/lowering/lowering_function_context.cpp
```diff
@@ -14,11 +14,39 @@ LoweringFunctionContext::LoweringFunctionContext(
...
+
+auto LoweringFunctionContext::GetBlockArg(SemanticsNodeBlockId block_id,
+                                          SemanticsTypeId type_id)
+    -> llvm::PHINode* {
+  llvm::BasicBlock* block = GetBlock(block_id);
+
+  // Find the existing phi, if any.
+  auto phis = block->phis();
+  if (!phis.empty()) {
+    return &*phis.begin();
```
jonmeow, line 52:
```suggestion
    CARBON_CHECK(phis.size() == 1) << "Expected at most one phi, found " << phis.size();
    return &*phis.begin();
```

This feels like it's assumed, make it explicit?

toolchain/lowering/lowering_function_context.cpp
```diff
@@ -14,11 +14,39 @@ LoweringFunctionContext::LoweringFunctionContext(
...
+  unsigned NumReservedPredecessors = 2;
+  auto* phi = llvm::PHINode::Create(GetType(type_id), NumReservedPredecessors);
+  phi->insertInto(block, block->begin());
+  return phi;
 }
 
+auto LoweringFunctionContext::CreateSyntheticBlock() -> llvm::BasicBlock* {
+  synthetic_block_ = llvm::BasicBlock::Create(llvm_context(), "", function_);
+  return *synthetic_block_;
+}
```
jonmeow, line 65:
```suggestion
}

```

toolchain/lowering/lowering_block_worklist.h
```diff
@@ -0,0 +1,44 @@
...
+namespace Carbon {
+
+// A worklist for blocks that need to be lowered. Blocks are popped from the
+// worklist in the order in which they are first pushed, in a depth-first
+// traversal.
+class LoweringBlockWorklist {
+ public:
+  // Add a block to the work list, if it's not already present.
+  auto Push(SemanticsNodeBlockId id) -> void {
+    if (found_.insert(id).second) {
```
jonmeow: In LoweringFunctionContext, you have `blocks_` which I think maps all found blocks. Maybe you could only `Push` entries not found in `blocks_`, removing `found_`?

jonmeow (commented):
Thanks, this answers some questions I had on the `phi` approach plan.

toolchain/lowering/lowering_block_worklist.h
zygoloid: Perhaps a different way of phrasing this would help. The goal here is to emit IR such that the order in which blocks are first mentioned is the same as the order in which they're emitted. Note in the above, we emit block references in this order:

- `label %1`
- `label %2`
- `label %4`
- `label %5`
- `label %7`
- `label %8`

The point is, code is easier to read if it always "goes forward", that is, we only have branches to earlier code in the case where there is a back-edge and a cycle in the control flow. Given that we're doing a depth-first traversal, the `reverse` is necessary to provide that: otherwise, code that says `BranchIf(cond1, block1); BranchIf(cond2, block2); Branch(block3)` would push `block1` then `block2` then `block3`, so we'd emit `block3` then `block2` then `block1`.

Now, we could fix this by doing a breadth-first traversal instead, and doing a FIFO traversal. But that creates a different problem with the readability of the IR: we generally want locality in the IR, so that a branch and the blocks that it branches to are close together. So we want a depth-first traversal rather than a breadth-first one.

Another way of viewing what this data structure is doing is:

1) Emit a block, and gather up its new block references, in order of reference.
2) Recursively emit those newly-referenced blocks.

When viewed that way, the work done between `Pop`s is: emit a block, performing some set of `Push`es, which accumulate in order from first to last. Then `reverse` those to put them in order from last to first, so that we will `Pop` them in the order they are pushed. Then the recursive emission is made non-recursive by repeating this procedure on the same queue.

toolchain/lowering/lowering_function_context.h
zygoloid: Hm, https://developers.google.com/style/hyphens#compound-modifiers-before-a-noun says to hyphenate "if needed for clarity". That's not especially helpful here :) But I guess the fact that you're asking for this indicates these hyphens aren't needed for clarity.

toolchain/lowering/lowering_function_context.h
zygoloid: Switched to just a pointer.

toolchain/lowering/lowering_block_worklist.h
zygoloid: Sure, done.

toolchain/lowering/lowering_block_worklist.h
zygoloid: I've expanded the class comment to explain what I'm trying to achieve here.

toolchain/lowering/lowering_block_worklist.h
zygoloid: Done.

toolchain/lowering/lowering_function_context.cpp
zygoloid: For the Semantic IR we currently generate, yes. That will be true if the `BlockArg` is always the first semantic node in the block, or if the first reference to a block with an argument is always through a `BranchWithArg` node. The current implementation doesn't require those properties to hold, although both of them do hold for all the Semantics IR that we build so far. I don't think we get much of a simplification by assuming the block is empty here, but can do that if you like.

toolchain/lowering/lowering_function_context.h
zygoloid: It's unidiomatic in the LLVM API to use pointers to `llvm::BasicBlock`. Eg, in "include/llvm/IR" there are 18 `BasicBlock&`s, including the declaration of the copy constructor and copy assignment, versus nearly 500 `BasicBlock*`s. We can switch here, but it'll lead to a lot of callers taking the address.

toolchain/lowering/lowering_block_worklist.h
zygoloid: Per in-person discussion, will revisit with a linear list of all blocks in the function in a separate PR.

jonmeow (approved):

toolchain/lowering/lowering_function_context.h
```diff
@@ -48,13 +66,25 @@ class LoweringFunctionContext {
     return lowering_context_->GetType(type_id);
   }
 
+  // Create a synthetic block that corresponds to no SemanticsNodeBlockId. Such
+  // a block should only ever have a single predecessor, and is used when we
+  // need multiple `llvm::BasicBlock`s to model the linear control flow in a
+  // single SemanticsIR block.
+  auto CreateSyntheticBlock() -> llvm::BasicBlock*;
+
+  // Determine whether block_ is the most-recently-created synthetic block.
```
jonmeow:
```suggestion
  // Determine whether block is the most-recently-created synthetic block.
```

Oops, missed this, sorry.

PR #2906, opened 2023-06-15
Implement syntax changes from #2760 in Explorer

josh11b:


Pixep (approved):
Look great to me! No comments on the class inheritance & logic specifically, and the rest looks solid. Thanks for fixing the 4 space tabs in my tests :).

PR #2006, opened 2022-08-12
Values, variables, pointers, and references

chandlerc:
Introduce a concrete design for how Carbon values, objects, storage, variables,
and pointers will work. This includes fleshing out the design for:

-   The expression categories used in Carbon to represent values and objects,
    how they interact, and terminology that anchors on their expression nature.
-   An expression category model for readonly, abstract values that can
    efficiently support function inputs.
-   A customization system for value expression representations, especially as
    seen on function boundaries in the calling convention.
-   An expression category model for references instead of a type system model.
-   How patterns match different expression categories.
-   How initialization works in conjunction with function returns.
-   Specific pointer syntax, semantics, and library customization mechanisms.
-   A `const` type qualifier for use when the value expression category system
    is too abstracted from the underlying objects in storage.

PR #2907, opened 2023-06-15
Explorer: Add initial initializing expression support for variable declaration

Pixep:
Add partial support for initializing expressions for variable declaration. This is based on https://github.com/carbon-language/carbon-lang/pull/2006, which introduces expression categories, and how it is possible to convert to/from those different categories.

## Functional changes

* Initializing expressions initialize directly the provided storage when used to initialize a variable.
   * Allows initializing expressions to avoid a copy when using `[var|let] name: type = call_expression(...)` by initializing `name` in-place.
   * Support `returned var: ...` and `return <expr>`
   * Support nested initializing expressions

* Updated PatternMatch logic to handle expression categories
* Updated `VariableDefinition` interpreter statement to allocate and pass a location to initializing expressions
    * Update statement actions to allow passing an allocation, used by return expr or returned var
* Modified the RuntimeScope API to be one step closer to the memory model we want to have
    * Remove `GetAllocationId` and older `Bind` which don't apply
* New set of tests to highlight those different situations
    * Added a new intrinsic to print the allocation stack (and make sure we behave correctly, beyond visible side effects)

## Next work

* Dedicated `Action` to retrieve expression category information in the interpreter (https://github.com/carbon-language/carbon-lang/pull/2927)
* Avoid copies when initializing value expression from reference expression and prevent mutations for the duration of the "pinning" (https://github.com/carbon-language/carbon-lang/pull/2927)
* Avoid unnecessary copies from value expression to value expression, after ensuring that even value expression temporaries are registered for destruction.
* Avoid unnecessary copies when binding function arguments


Pixep:
Merging `trunk` changes the behavior of some allocations/destructions, will address that next.

explorer/ast/expression_category.h
```diff
@@ -18,6 +20,8 @@ enum class ExpressionCategory {
   Initializing,
 };
 
+auto ExprCategoryToString(ExpressionCategory cat) -> llvm::StringRef;
```
jonmeow:
```suggestion
auto ExpressionCategoryToString(ExpressionCategory cat) -> llvm::StringRef;
```

This is used in one place, let's just match names.

explorer/ast/value.h
```diff
@@ -95,6 +96,38 @@ class Value {
   const Kind kind_;
 };
 
+// Contains the result of the evaluation of an expression, including a value,
+// the original expression category, and an optional address if available.
+class ExpressionResult {
+ public:
+  ExpressionResult(Nonnull<const Value*> v, std::optional<Address> address,
```
jonmeow: Given the factory functions, does the constructor need to be public?

explorer/ast/value.h
```diff
@@ -95,6 +96,38 @@ class Value {
...
+  static auto Reference(Nonnull<const Carbon::Value*> v, Address address)
+      -> ExpressionResult {
+    return ExpressionResult(v, std::move(address),
+                            ExpressionCategory::Reference);
+  }
+  static auto Initializing(Nonnull<const Carbon::Value*> v, Address address)
+      -> ExpressionResult {
+    return ExpressionResult(v, std::move(address),
+                            ExpressionCategory::Initializing);
+  }
```
jonmeow:
```suggestion
  static auto Value(Nonnull<const Carbon::Value*> v) -> ExpressionResult {
    return ExpressionResult(v, std::nullopt, ExpressionCategory::Value);
  }
  static auto Reference(Nonnull<const Carbon::Value*> v, Address address)
      -> ExpressionResult {
    return ExpressionResult(v, std::move(address),
                            ExpressionCategory::Reference);
  }
  static auto Initializing(Nonnull<const Carbon::Value*> v, Address address)
      -> ExpressionResult {
    return ExpressionResult(v, std::move(address),
                            ExpressionCategory::Initializing);
  }

  ExpressionResult(Nonnull<const Value*> v, std::optional<Address> address,
                   ExpressionCategory cat)
      : value_(v), address_(std::move(address)), expr_cat_(cat) {}
```

Per style, factory functions come before constructors (4 and 5 in [Declaration Order](https://google.github.io/styleguide/cppguide.html#Declaration_Order))

explorer/ast/value.h
```diff
@@ -95,6 +96,38 @@ class Value {
...
+
+  auto value() const -> Nonnull<const Carbon::Value*> { return value_; }
+  auto address() const -> std::optional<Address> { return address_; }
```
jonmeow:
```suggestion
  auto address() const -> const std::optional<Address>& { return address_; }
```

Not sure if you just missed this here, but the Address is expensive to copy.

explorer/interpreter/action.cpp
```diff
@@ -44,32 +46,38 @@ void RuntimeScope::Print(llvm::raw_ostream& out) const {
   out << "}";
 }
 
-void RuntimeScope::Bind(ValueNodeView value_node, Nonnull<const Value*> value) {
+void RuntimeScope::Bind(ValueNodeView value_node, Address address) {
+  CARBON_CHECK(!value_node.constant_value().has_value());
+  auto [it, success] =
+      locals_.insert({value_node, heap_->arena().New<LocationValue>(address)});
```
jonmeow:
```suggestion
  bool success =
      locals_.insert({value_node, heap_->arena().New<LocationValue>(address)}).second;
```

`it` is unused, so no reason to keep it.

explorer/interpreter/action.cpp
```diff
@@ -44,32 +46,38 @@ void RuntimeScope::Print(llvm::raw_ostream& out) const {
...
-    auto [it, success] = locals_.insert(
-        {value_node, heap_->arena().New<LocationValue>(Address(id))});
-    CARBON_CHECK(success) << "Duplicate definition of " << value_node.base();
-  } else {
-    auto [it, success] = locals_.insert(
-        {value_node,
-         heap_->arena().New<LocationValue>(Address(*allocation_id))});
-    CARBON_CHECK(success) << "Duplicate definition of " << value_node.base();
-  }
+  auto [it, success] = locals_.insert({value_node, value});
```
jonmeow:
```suggestion
  bool success = locals_.insert({value_node, value}).second;
```

And I see you're just moving this around, but given you're modifying the lines I'm still going to suggest it. Otherwise I look for how `it` is used.

explorer/interpreter/action.cpp
```diff
@@ -44,32 +46,38 @@ void RuntimeScope::Print(llvm::raw_ostream& out) const {
...
+    -> Nonnull<const LocationValue*> {
   CARBON_CHECK(!value_node.constant_value().has_value());
   CARBON_CHECK(value->kind() != Value::Kind::LocationValue);
   allocations_.push_back(heap_->AllocateValue(value));
-  auto [it, success] = locals_.insert(
-      {value_node,
-       heap_->arena().New<LocationValue>(Address(allocations_.back()))});
+  const auto* location =
+      heap_->arena().New<LocationValue>(Address(allocations_.back()));
+  auto [it, success] = locals_.insert({value_node, location});
```
jonmeow:
```suggestion
  bool success = locals_.insert({value_node, location}).second;
```

explorer/interpreter/action.h
```diff
@@ -45,30 +48,43 @@ class RuntimeScope {
...
+  // - returns its `LocationValue*` if bound to a reference expression in this
+  // scope,
+  // - returns a `Value*` if bound to a value expression in this scope,
+  // - returns `nullptr` if not bound.
   auto Get(ValueNodeView value_node) const
-      -> std::optional<Nonnull<const LocationValue*>>;
+      -> std::optional<Nonnull<const Value*>>;
 
-  // Returns the local values in created order
+  // Returns the local values with allocation in created order
```
jonmeow:
```suggestion
  // Returns the local values with allocation in created order.
```

Well, if you're updating the comment anyways...

explorer/interpreter/action.h
```diff
@@ -45,30 +48,43 @@ class RuntimeScope {
...
   // scopes must not define the same name, and must be backed by the same Heap.
   void Merge(RuntimeScope other);
 
-  // Returns the local storage for value_node, if it has storage local to
-  // this scope.
+  // Given node `value_node`:
+  // - returns its `LocationValue*` if bound to a reference expression in this
+  // scope,
+  // - returns a `Value*` if bound to a value expression in this scope,
+  // - returns `nullptr` if not bound.
```
jonmeow:
```suggestion
  // Given node `value_node`, returns:
  // - its `LocationValue*` if bound to a reference expression in this scope,
  // - a `Value*` if bound to a value expression in this scope, or
  // - `nullptr` if not bound.
```

explorer/data/prelude.carbon
```diff
@@ -730,6 +730,9 @@ class Heap {
   fn Delete[T:! type, self: Self](p : T*) {
     __intrinsic_delete(p);
   }
+  fn PrintAllocations[self: Self]() {
+    __intrinsic_print_allocs();
```
jonmeow: Please rename either `PrintAllocations`->`PrintAllocs` or `print_allocs`->`print_allocations` so that both versions of the function either abbreviate or don't, rather than switching. I don't have a strong preference for either (if you wanted feedback, I'd probably write `allocs` myself).

explorer/interpreter/heap.h
```diff
@@ -52,6 +50,13 @@ class Heap : public HeapAllocationInterface {
   void Deallocate(AllocationId allocation) override;
   void Deallocate(const Address& a);
 
+  // Marks this allocation, and all its sub-objects, as discarded.
+  void Discard(AllocationId allocation);
+  // Returns whether the given allocation was unused and discarded.
+  auto IsDiscarded(AllocationId allocation) const -> bool;
+  // Returns whether the given allocation was initialized.
+  auto IsInitialized(AllocationId allocation) const -> bool;
```
jonmeow: Since this have trivial implementations, naming them `is_discarded`/`is_initialized` may be more consistent with the property naming style.

explorer/interpreter/interpreter.cpp
```diff
@@ -291,46 +293,89 @@ auto Interpreter::CreateStruct(const std::vector<FieldInitializer>& fields,
...
+      break;
+    case ExpressionCategory::Value:
+      if (v.expression_category() == ExpressionCategory::Value) {
+        // TODO: Extend value lifetime to encompass this value
+        (*bindings)->Initialize(value_node, v.value());
+      } else if (v.expression_category() == ExpressionCategory::Reference) {
+        // TODO: Prevent mutation, error on mutation, or copy
+        // Bind the reference expression value directly.
+        (*bindings)->BindValue(value_node, v.value());
+      } else /* ExpressionCategory::Initializing */ {
```
jonmeow:
```suggestion
      } else {
```

As above.

explorer/interpreter/interpreter.cpp
```diff
@@ -291,46 +293,89 @@ auto Interpreter::CreateStruct(const std::vector<FieldInitializer>& fields,
...
+static auto InitializePlaceholderValue(
+    const ValueNodeView& value_node, ExpressionResult v,
+    std::optional<Nonnull<RuntimeScope*>> bindings) {
+  switch (value_node.expression_category()) {
+    case ExpressionCategory::Reference:
+      if (v.expression_category() == ExpressionCategory::Value ||
+          v.expression_category() == ExpressionCategory::Reference) {
+        // Build by copying from value or reference expression
+        (*bindings)->Initialize(value_node, v.value());
+      } else /* ExpressionCategory::Initializing */ {
```
jonmeow:
```suggestion
      } else {
```

Since Carbon would not support this style of commenting, we are going without it, [per style](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/project/cpp_style_guide.md#syntax-and-formatting). If you want the case to be explicit, use a `switch`, since that would also validate that the enum is fully handled.

Similar for 321.

explorer/interpreter/interpreter.cpp
```diff
@@ -291,46 +293,89 @@ auto Interpreter::CreateStruct(const std::vector<FieldInitializer>& fields,
...
 
-auto PatternMatch(Nonnull<const Value*> p, Nonnull<const Value*> v,
+static auto InitializePlaceholderValue(
+    const ValueNodeView& value_node, ExpressionResult v,
+    std::optional<Nonnull<RuntimeScope*>> bindings) {
+  switch (value_node.expression_category()) {
+    case ExpressionCategory::Reference:
+      if (v.expression_category() == ExpressionCategory::Value ||
+          v.expression_category() == ExpressionCategory::Reference) {
+        // Build by copying from value or reference expression
```
jonmeow:
```suggestion
        // Build by copying from value or reference expression.
```

explorer/interpreter/interpreter.cpp
```diff
@@ -291,46 +293,89 @@ auto Interpreter::CreateStruct(const std::vector<FieldInitializer>& fields,
...
+        // Location initialized by initializing expression, bind node to
+        // address.
+        CARBON_CHECK(v.address())
+            << "Missing location from initializing expression";
+        (*bindings)->Bind(value_node, *v.address());
+      }
+      break;
+    case ExpressionCategory::Value:
+      if (v.expression_category() == ExpressionCategory::Value) {
+        // TODO: Extend value lifetime to encompass this value
```
jonmeow:
```suggestion
        // TODO: Extend value lifetime to encompass this value.
```

explorer/interpreter/interpreter.cpp
```diff
@@ -291,46 +293,89 @@ auto Interpreter::CreateStruct(const std::vector<FieldInitializer>& fields,
...
+}
+
+auto PatternMatch(Nonnull<const Value*> p, ExpressionResult v,
                   SourceLocation source_loc,
                   std::optional<Nonnull<RuntimeScope*>> bindings,
                   BindingMap& generic_args, Nonnull<TraceStream*> trace_stream,
                   Nonnull<Arena*> arena) -> bool {
   if (trace_stream->is_enabled()) {
-    *trace_stream << "match pattern " << *p << "\nwith value " << *v << "\n";
+    *trace_stream << "match pattern " << *p << "\nfrom a "
```
jonmeow:
```suggestion
    *trace_stream << "match pattern " << *p << "\nfrom "
```

Consider dropping "a" because it won't read right with "from a InitializingExpression", while "from InitializingExpression" reads okay.

explorer/ast/value.cpp
```diff
@@ -1237,4 +1237,15 @@ auto NominalClassType::InheritsClass(Nonnull<const Value*> other) const
   return false;
 }
 
+auto ExprCategoryToString(ExpressionCategory cat) -> llvm::StringRef {
+  switch (cat) {
+    case ExpressionCategory::Value:
+      return "ValueExpression";
```
jonmeow: Consider having the caller tag with "expression" so that you're just returning the enum value's name here. In part, this is because we have things like EnumBase that'd generate this kind of function as directly returning the enum value name.

explorer/interpreter/interpreter.cpp
```diff
@@ -966,21 +1020,28 @@ auto Interpreter::CallDestructor(Nonnull<const DestructorDeclaration*> fun,
...
            << "destructors currently don't support `addr self` bindings";
   }
-  if (placeholder->value_node().has_value()) {
-    method_scope.Bind(*placeholder->value_node(), receiver);
+  if (auto& value_node = placeholder->value_node()) {
+    if (value_node->expression_category() == ExpressionCategory::Value) {
+      method_scope.BindValue(*placeholder->value_node(), receiver);
+    } else {
+      // TODO: [self addr: Self*] destructors not implemented yet.
+      CARBON_FATAL() << "[self addr: Self*] destructors not implemented yet";
```
jonmeow:
```suggestion
      CARBON_FATAL() << "TODO: [self addr: Self*] destructors not implemented yet";
```

It feels odd to me to have a TODO that copies the error message, can just add the TODO to the message. :)

explorer/interpreter/interpreter.cpp
```diff
@@ -2110,37 +2188,86 @@ auto Interpreter::StepStmt() -> ErrorOr<Success> {
...
+          // TODO: Handle allocation forwarding for initializing expression.
+          const auto address = Address(*act.location_received());
+          scope.Bind(*value_node, address);
+          CARBON_RETURN_IF_ERROR(heap_.Write(address, v, stmt.source_loc()));
+        } else /* not a returned var */ {
+          BindingMap generic_args;
+          CARBON_CHECK(
+              PatternMatch(p, ExpressionResult(v, v_location, expr_category),
+                           stmt.source_loc(), &scope, generic_args,
+                           trace_stream_, this->arena_))
```
jonmeow:
```suggestion
           bool matched =   PatternMatch(p, ExpressionResult(v, v_location, expr_category),
                           stmt.source_loc(), &scope, generic_args,
                           trace_stream_, this->arena_))
          CARBON_CHECK(matched)
```

There's a preference for only putting debug logic inside a check. i.e. if all checks were removed, behavior shouldn't significantly change. I think this match is significant, so requesting it be moved out.

(I'm going to let clang-format reformat my suggestion)

explorer/testdata/let/value_expression_declaration.carbon
```diff
@@ -0,0 +1,79 @@
...
+//
+// AUTOUPDATE
+// CHECK:STDOUT: Setup
+// CHECK:STDOUT: 0: Heap{}, 1: C{.id = 1}, 2: C{.id = 2}
+// CHECK:STDOUT:
+// CHECK:STDOUT: Initialize c1 from reference expression
+// CHECK:STDOUT: 0: Heap{}, 1: C{.id = 1}, 2: C{.id = 2}, 3: C{.id = 1}
+// CHECK:STDOUT:
+// CHECK:STDOUT: Initialize c2 from value expression
+// CHECK:STDOUT: 0: Heap{}, 1: C{.id = 1}, 2: C{.id = 2}, 3: C{.id = 1}, 4: C{.id = 2}
```
jonmeow: c1 is assigned from a `var`, and so a copy is useful in order to prevent observation of the `var` mutation (unless you get really technical, like observing `c1_var` doesn't change). However, `c2` is a `let`, and so it could reference `c2_let` without creating a copy. Does it match your expectation that that's the end goal, even if the code doesn't do it right now?

explorer/testdata/pointer/print_allocs.carbon
```diff
@@ -0,0 +1,16 @@
...
+// AUTOUPDATE
+// CHECK:STDOUT: 0: Heap{}, 1: 1, 2: 2
+// CHECK:STDOUT: result: 0
+
+package ExplorerTest api;
+
+fn Main() -> i32 {
+  var a: i32 = 1;
+  var b: i32 = 2;
+  heap.PrintAllocations();
```
jonmeow: This test is in `/pointer`, but maybe `/basic_syntax` would be a better home because this doesn't deal with pointers? Similar with `print_allocs_empty`.

explorer/testdata/var/local/reference_expression_binding.carbon
```diff
@@ -0,0 +1,54 @@
...
+    return {.id = id};
+  }
+  destructor[self: Self] {
+    Print("c{0} destroyed", self.id);
+  }
+  var id: i32;
+}
+
+fn CallWithValueExpressionBinding(var c: C) {
+  return;
```
jonmeow:
```suggestion
```

explorer/testdata/var/local/reference_expression.carbon
```diff
@@ -0,0 +1,84 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+// CHECK:STDOUT: Initialize c1 from reference expression
+// CHECK:STDOUT: 0: Heap{}, 1: C{.id = 1}, 2: C{.id = 1}
+// CHECK:STDOUT: c1 destroyed
+// CHECK:STDOUT: c1 destroyed
+// CHECK:STDOUT:
+// CHECK:STDOUT: Initialize c2 from value expression
+// CHECK:STDOUT: 0: Heap{}, 1: !!C{.id = 1}, 2: !!C{.id = 1}, 3: C{.id = 2}, 4: C{.id = 2}
```
jonmeow: You might want to print the heap between each function, so that the scope destructions are easily ignored. You could do that either with a call in Main, or just write a class with a destructor that prints the allocations and allocate it before other local vars.

jonmeow (commented):
I think the core implementation looks good. I have a number of comments, but I think they trend more stylistic. Don't worry about me being concerned with the core implementation, basically. :)

explorer/ast/value.h
Pixep: It is used in at least one place, where this avoids branching based on expression category by having that constructor (`PatternMatch`). Might want to revisit in the dependent PR, where it might not be needed anymore though.

explorer/interpreter/action.cpp
Pixep: Thanks, had it fixed in the dependent, but not here. I have a slight preference for `auto [_, success]` vs appending a member access. Any objections?

jonmeow: In code, we have cases of `insert(...).second` in resolve_names.cpp and bindings.cpp in the explorer, plus some in the toolchain. I don't think we have any `[_, success]` uses. We seem not to use `_` in bindings in general: I see one spot I did `[param_name_id, _]` in lowering. I think this is actually partly because zygoloid had me change some similar cases.

Consider this a somewhat strong leaning towards `.first`/`.second` over `_` captures for somewhat idiomatic consistency, and because the `.second` still makes it clear there's part of a return value being discarded.

explorer/testdata/let/value_expression_declaration.carbon
Pixep: I added a todo in `InitializePlaceholderValue`, and an action in the "next work" section of the PR. Let expressions for temporaries are not bound to allocations, causing the destructor to never be called. The safe option for now is to copy to ensure that the destructor is called at least once. So this isn't the end goal, yes.

explorer/ast/expression_category.h
Pixep: Done

explorer/ast/value.h
Pixep: Done

explorer/testdata/pointer/print_allocs.carbon
Pixep: Ha good, was looking for a good location for that. Done

explorer/testdata/var/local/reference_expression_binding.carbon
Pixep: Done

explorer/interpreter/action.cpp
Pixep: Sounds good! Done

explorer/data/prelude.carbon
Pixep: Done

explorer/interpreter/heap.h
Pixep: Done

explorer/interpreter/action.cpp
Pixep: Done

explorer/interpreter/interpreter.cpp
Pixep: Done

explorer/ast/value.cpp
Pixep: Sounds good, done.

explorer/ast/value.h
Pixep: Definitely a miss, done.

explorer/interpreter/action.cpp
Pixep: Done

explorer/testdata/var/local/reference_expression.carbon
Pixep: Done with manual calls, to avoid polluting the Heap{}

explorer/interpreter/interpreter.cpp
```diff
@@ -2231,19 +2232,23 @@ auto Interpreter::StepStmt() -> ErrorOr<Success> {
...
-          // expressions.
-          // Allocate storage for initializing expression.
-          const auto allocation_id =
-              heap_.AllocateValue(arena_->New<UninitializedValue>(
-                  &definition.init().static_type()));
-          act.set_location_created(allocation_id);
-          RuntimeScope scope(&heap_);
-          scope.BindLifetimeToScope(Address(allocation_id));
-          todo_.MergeScope(std::move(scope));
+          if (!(definition.is_returned() && act.location_received())) {
```
jonmeow: Join the ifs?

i.e., if (kind && expression_category && !(is_returned && location_received))

explorer/interpreter/interpreter.cpp
```diff
@@ -2253,19 +2258,25 @@ auto Interpreter::StepStmt() -> ErrorOr<Success> {
...
         if (definition.has_init()) {
-          const auto init_location = act.location_created();
           if (expr_category == ExpressionCategory::Initializing &&
-              init_location && heap_.IsInitialized(*init_location)) {
+              init_location && heap_.is_initialized(*init_location)) {
             const auto address = Address(*init_location);
             CARBON_ASSIGN_OR_RETURN(
                 v, heap_.Read(address, definition.source_loc()));
-            CARBON_CHECK(v == act.results()[0]);
+            // CARBON_CHECK(v == act.results()[0]);
```
jonmeow: This should either be uncommented again or removed, I'm not sure what's right (maybe you commented it because it's no longer correct, given the cast)
```suggestion
```

explorer/interpreter/interpreter.cpp
```diff
@@ -2396,11 +2408,8 @@ auto Interpreter::StepStmt() -> ErrorOr<Success> {
             value, heap_.Read(location->address(), ret_var.source_loc()));
       }
       const CallableDeclaration& function = cast<Return>(stmt).function();
-      CARBON_ASSIGN_OR_RETURN(
-          Nonnull<const Value*> return_value,
-          Convert(value, &function.return_term().static_type(),
-                  stmt.source_loc()));
-      return todo_.UnwindPast(*function.body(), return_value);
+      // TODO: Not needed?
```
jonmeow: If I recall correct what unwind is doing, I expect this is still needed to deal with returning from a nested scope.

explorer/testdata/var/local/reference_expression_declaration.carbon
```diff
@@ -5,22 +5,30 @@
...
+// CHECK:STDOUT: c3 destroyed
+// CHECK:STDOUT: 0: Heap{}, 1: !!C{.id = 1}, 2: !!C{.id = 1}, 3: !!C{.id = 2}, 4: !!C{.id = 2}, 5: !!C{.id = 3}, 6: !!3, 7: Uninit<class C>, 8: !!C{.id = 3}
 // CHECK:STDOUT:
 // CHECK:STDOUT: Initialize c4 from initializing expression (returned var)
-// CHECK:STDOUT: 0: Heap{}, 1: !!C{.id = 1}, 2: !!C{.id = 1}, 3: !!C{.id = 2}, 4: !!C{.id = 2}, 5: C{.id = 3}, 6: !!3, 7: !!C{.id = 3}, 8: C{.id = 4}, 9: !!4
+// CHECK:STDOUT: 0: Heap{}, 1: !!C{.id = 1}, 2: !!C{.id = 1}, 3: !!C{.id = 2}, 4: !!C{.id = 2}, 5: !!C{.id = 3}, 6: !!3, 7: Uninit<class C>, 8: !!C{.id = 3}, 9: C{.id = 4}, 10: !!4
+// CHECK:STDOUT: End of scope
 // CHECK:STDOUT: c4 destroyed
-// CHECK:STDOUT: c3 destroyed
+// CHECK:STDOUT: 0: Heap{}, 1: !!C{.id = 1}, 2: !!C{.id = 1}, 3: !!C{.id = 2}, 4: !!C{.id = 2}, 5: !!C{.id = 3}, 6: !!3, 7: Uninit<class C>, 8: !!C{.id = 3}, 9: !!C{.id = 4}, 10: !!4
```
jonmeow: Consider splitting tests. :-P

jonmeow (approved):
Looks good! Just a last few comments, but they're straightforward so feel free to merge after addressing them.

Pixep:
> Looks good! Just a last few comments, but they're straightforward so feel free to merge after addressing them.

Thank you for this last round, that included some draft code, I should have let you know directly. All addressed.
Thanks @jonmeow !

PR #2825, opened 2023-05-14
Add array size deduction from tuple(fix#1590)

kshokhin:
Allow array size deduction from tuples and arrays on array declaration

Closes #1590

PR #2909, opened 2023-06-15
Add nested array size deduction from tuple

kshokhin:
As a follow up to #2825 this pr implements size deduction for nested arrays.
E.g. `var x: [[i32;];] = ((1,2), (3,4));`.

Also updated the pattern matching logic for arrays, now it also checks element types of the tuple size being deduced from. As a result code like `var x: [i32;] = ("foo", "bar");`(note, that it tries to init an array of i32 with a tuple of strings) fails to compile with a pattern match error instead of an implicit cast failed error.

Moved some common type-related logic used in type_checker.cpp and interpreter.cpp to the separate file.

### DIFF
--- a/explorer/interpreter/pattern_match.cpp
+++ b/explorer/interpreter/pattern_match.cpp
@@ -8,6 +8,7 @@
 #include "explorer/base/arena.h"
 #include "explorer/base/trace_stream.h"
 #include "explorer/interpreter/action.h"
+#include "explorer/interpreter/type_utils.h"
 #include "llvm/Support/Casting.h"
 
 using llvm::cast;
@@ -202,14 +203,37 @@ auto PatternMatch(Nonnull<const Value*> p, ExpressionResult v,
       // on the typechecker to ensure that `v.value()` is a type.
       return true;
     case Value::Kind::StaticArrayType: {
+      const auto& p_arr = cast<StaticArrayType>(*p);
       switch (v.value()->kind()) {
         case Value::Kind::TupleType:
         case Value::Kind::TupleValue: {
+          const auto& v_tup = cast<TupleValueBase>(*v.value());
+          if (v_tup.elements().empty()) {
+            return !TypeIsDeduceable(&p_arr.element_type());
+          }
+
+          const size_t deduced_size = GetSize(v_tup.elements()[0]);
+          for (const auto& tup_elem : v_tup.elements()) {
+            if (!PatternMatch(&p_arr.element_type(), make_expr_result(tup_elem),
+                              source_loc, bindings, generic_args, trace_stream,
+                              arena)) {
+              return false;
+            }
+
+            if (deduced_size != GetSize(tup_elem)) {
+              return false;
+            }
+          }  // for
           return true;
         }
         case Value::Kind::StaticArrayType: {
           const auto& v_arr = cast<StaticArrayType>(*v.value());
-          return v_arr.has_size();
+          if (!v_arr.has_size()) {
+            return false;
+          }
+          return PatternMatch(
+              &p_arr.element_type(), make_expr_result(&v_arr.element_type()),
+              source_loc, bindings, generic_args, trace_stream, arena);
         }
         default:
           return false;
--- a/explorer/interpreter/pattern_match.cpp
+++ b/explorer/interpreter/pattern_match.cpp
@@ -4,6 +4,8 @@
 
 #include "explorer/interpreter/pattern_match.h"
 
+#include <algorithm>
+
 #include "explorer/ast/value.h"
 #include "explorer/base/arena.h"
 #include "explorer/base/trace_stream.h"
@@ -212,7 +214,8 @@ auto PatternMatch(Nonnull<const Value*> p, ExpressionResult v,
             return !TypeIsDeduceable(&p_arr.element_type());
           }
 
-          const size_t deduced_size = GetSize(v_tup.elements()[0]);
+          std::vector<Nonnull<const Value*>> deduced_types;
+          deduced_types.reserve(v_tup.elements().size());
           for (const auto& tup_elem : v_tup.elements()) {
             if (!PatternMatch(&p_arr.element_type(), make_expr_result(tup_elem),
                               source_loc, bindings, generic_args, trace_stream,
@@ -220,11 +223,14 @@ auto PatternMatch(Nonnull<const Value*> p, ExpressionResult v,
               return false;
             }
 
-            if (deduced_size != GetSize(tup_elem)) {
-              return false;
-            }
+            deduced_types.emplace_back(
+                DeducePatternType(&p_arr.element_type(), tup_elem, arena));
           }  // for
-          return true;
+
+          return std::adjacent_find(deduced_types.begin(), deduced_types.end(),
+                                    [](const auto& lhs, const auto& rhs) {
+                                      return !TypeEqual(lhs, rhs, std::nullopt);
+                                    }) == deduced_types.end();
         }
         case Value::Kind::StaticArrayType: {
           const auto& v_arr = cast<StaticArrayType>(*v.value());
--- a/explorer/interpreter/type_checker.h
+++ b/explorer/interpreter/type_checker.h
@@ -565,6 +565,10 @@ class TypeChecker {
   auto InterpExp(Nonnull<const Expression*> e)
       -> ErrorOr<Nonnull<const Value*>>;
 
+  // Deduces concrete type for 'type' based on 'expected'
+  auto Deduce(Nonnull<const Value*> type, Nonnull<const Value*> expected)
+      -> Nonnull<const Value*>;
+
   Nonnull<Arena*> arena_;
   Builtins builtins_;
 
--- a/explorer/interpreter/type_checker.h
+++ b/explorer/interpreter/type_checker.h
@@ -565,10 +565,6 @@ class TypeChecker {
   auto InterpExp(Nonnull<const Expression*> e)
       -> ErrorOr<Nonnull<const Value*>>;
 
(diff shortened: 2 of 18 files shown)


geoffromer:
Is it possible to have a separate PR that just moves code to type_utils, either before or after this one? That would really make this change easier to review.

kshokhin:
Depends on #2926 that moves the code to type_utils.

geoffromer:
Now that #2926 is merged, can you re-sync with `trunk`?

Thanks for your patience with all this.

explorer/testdata/array/fail_to_deduce_nested_array_size_from_tuple_different_subtuples_size.carbon
```diff
@@ -0,0 +1,14 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+
+package ExplorerTest api;
+
+fn Main() -> i32 {
+  // CHECK:STDERR: COMPILATION ERROR: fail_to_deduce_nested_array_size_from_tuple_different_subtuples_size.carbon:[[@LINE+1]]: type pattern '{{\[\[}}i32;];]' does not match actual type '((i32, i32), (i32, i32, i32))'
```
geoffromer, line 10: It seems like something's gone wrong with the type pattern formatting.

explorer/interpreter/pattern_match.cpp
```diff
@@ -200,14 +201,37 @@ auto PatternMatch(Nonnull<const Value*> p, ExpressionResult v,
...
+
+          const size_t deduced_size = GetSize(v_tup.elements()[0]);
+          for (const auto& tup_elem : v_tup.elements()) {
+            if (!PatternMatch(&p_arr.element_type(), make_expr_result(tup_elem),
+                              source_loc, bindings, generic_args, trace_stream,
+                              arena)) {
+              return false;
+            }
+
+            if (deduced_size != GetSize(tup_elem)) {
```
geoffromer: I'm having trouble following this. If I understand correctly, this is trying to make sure that the tuple elements all have the same size, right? If so, that's confusing to me, because in this context it seems like we have no particular reason to expect the tuple elements to have a meaningful size, so this seems out of place.

Also, won't the type checker have already taken care of this? If not, don't we also need to check that `v_tup.elements().size()` matches the size of the array?

(Part of my confusion here is that we're using pattern matching to handle something that really should be type deduction, but that's a pre-existing problem.)

explorer/interpreter/type_checker.h
```diff
@@ -565,6 +565,10 @@ class TypeChecker {
   auto InterpExp(Nonnull<const Expression*> e)
       -> ErrorOr<Nonnull<const Value*>>;
 
+  // Deduces concrete type for 'type' based on 'expected'
+  auto Deduce(Nonnull<const Value*> type, Nonnull<const Value*> expected)
```
geoffromer: The name and comment make this function sound far more general than it seems like the implementation actually supports. It also makes it sound like this is redundant with `ArgumentDeduction`. Is it possible to use `ArgumentDeduction` for this purpose? If not, can the name/comment be changed to make it clearer how they're different?

explorer/interpreter/pattern_match.cpp
kshokhin: Sorry for the delay with this.

> If I understand correctly, this is trying to make sure that the tuple elements all have the same size, right?

Indeed, the idea is to check that the tuple has a homogenous structure, thus matches the array structure. Could you please elaborate a bit on 

> we have no particular reason to expect the tuple elements to have a meaningful size

?
Do you mean that, e.g. i32 has no meaningful size? If so, `GetSize` returns 0 for such types. Probably it'd be more explicit to return `std::optional<size_t>` holding an actual value only for "sizeable" types?

> Also, won't the type checker have already taken care of this?

Probably I'm missing some contexts, where this function is also used, but AFAIU `TypeChecker::TypeCheckPattern` uses it for type deduction, so no checks are performed yet. 

>  If not, don't we also need to check that v_tup.elements().size() matches the size of the array?

Good point, I guess currently this check is performed during the implicit conversion of the tuple to the array and pattern matching is actually used for type deduction. However, for cases like
```
var x: [[i32;];3] = ((1,2), (3,4,5));
```
it seems like pattern matching should fail. I'll add the check.

explorer/testdata/array/fail_to_deduce_nested_array_size_from_tuple_different_subtuples_size.carbon
kshokhin: Plain
```
type pattern '[[i32;];]'
```
doesn't work, test fails with an error
```
Unexpected [[, should be {{\[\[}} at `i32;]`
```
I believe it's because of the FileCheck's substitution blocks syntax, so regex used to avoid the confusion.
If I'm missing something and there's a better way to verify the output, would appreciate any guidance.

explorer/interpreter/pattern_match.cpp
geoffromer: Ohhhh, I think I understand what's going on now: we're deducing the element type of an array from the initializer, so we need to deduce from each element separately, and verify that all deductions produce the same result. But since we're limiting this to the case where the element type is already known to be an array type, we only need to verify that all deductions deduce the same size.

Assuming that's right, I think the code will be clearer (and more general) if we don't rely on that limitation here: instead of checking that the deduced types all _have the same size_, I think we should check that the deduced types _are all equal_. Does that make sense?

(It might even be better to drop that limitation altogether, and add support for things like `var x: [auto; 3] = (1, 2, 3);`, but that doesn't necessarily need to be in this PR.)

explorer/interpreter/pattern_match.cpp
```diff
@@ -200,14 +201,37 @@ auto PatternMatch(Nonnull<const Value*> p, ExpressionResult v,
       // on the typechecker to ensure that `v.value()` is a type.
       return true;
     case Value::Kind::StaticArrayType: {
+      const auto& p_arr = cast<StaticArrayType>(*p);
       switch (v.value()->kind()) {
         case Value::Kind::TupleType:
         case Value::Kind::TupleValue: {
+          const auto& v_tup = cast<TupleValueBase>(*v.value());
+          if (v_tup.elements().empty()) {
```
geoffromer, line 213: We should probably have tests covering this case.

explorer/interpreter/pattern_match.cpp
kshokhin: Yep, the general idea is correct. However we do not make any assumptions about array elements' type, we just know that the pattern we're trying to match with is an array.
So what's happening here is we check, that every tuple element matches array's element type.
However this is not enough, as if the array's elements are arrays themselves and their size should be deduced, then even if tuple elements have different sizes, they'll successfully match "outer" array's elements' type. Thus check for deduced size added. 

I agree, that code is not obvious and your suggestion absolutely makes sense.
To sum up:

1. Check that all tuple elements have equal types
2. Check that this type matches array's elements' type
3. If array's size explicitly set, check that it's equal to tuple's size

Does it look correct to you?

It also seems like it enables support for `auto` as an element type.

geoffromer: That sounds basically right, but I was envisioning something subtly different for steps 1 and 2: rather than check the tuple element types are equal, we use each tuple element type to deduce the array element type, and then check that the results of those deductions are equal. That way the same logic should work for the case where the tuple elements have different types, but the array element type is explicitly specified, and the tuple element types are convertible to it.

geoffromer:
It looks like you've pushed some new changes, but I can't see what's actually changed because your pushes "rewrote history". Is this ready for another round of review?

By the way, please don't force-push during code review; instead, apply changes in new commits on top of the existing ones. See [here](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/project/code_review.md#responding-to-review-comments) for more about how our review process works.

kshokhin:
Just noticed the comment, sorry for the delays.

> It looks like you've pushed some new changes

I've rebased to the actual trunk, nothing changed in that force-pushed commits. Not found anything about pulling actual trunk to the feature branch in the provided article, what is the recommended way of doing it?

>  Is this ready for another round of review?

I'm not actually sure, that it's ready.
I've made a "naive" implementation of what we were talking about - deduce type from every tuple element and check that they're all the same, also added a test with an empty tuple. If you have some free time - I would be happy to hear your feedback.

Currently I'm digging into `ArgumentDeduction`, trying to understand, whether it could be reused for this problem.

What I see now is that it depends on the `TypeChecker`, so it'd be difficult to use it from the `Interpreter`(and pattern matching also happens there).
Also I've tried to use it inside the `TypeChecker` for pattern matching and my first attempt was not very successful :)

You've also mentioned that 

>  we're using pattern matching to handle something that really should be type deduction, but that's a pre-existing problem

and at this point it seems to me that the "right" or more clear solution requires differentiating pattern matching part from type deduction part.

However, it seems that I'm not able to address it as of now.

To sum up: the current plan is to try to reuse `ArgumentDeduction`, as it's seemingly doing the same thing that I need during pattern matching(and even more).

geoffromer:
> Just noticed the comment, sorry for the delays.
> 
> > It looks like you've pushed some new changes
> 
> I've rebased to the actual trunk, nothing changed in that force-pushed commits. Not found anything about pulling actual trunk to the feature branch in the provided article, what is the recommended way of doing it?

The recommended way of doing it is with `git merge`, but it's usually easiest to not do that until the PR is approved.

geoffromer (approved):

geoffromer:
This looks good. Thanks for pushing it through!

PR #2851, opened 2023-05-25
Improved trace output selection using program phase filtering

prabhatexit0:
By implementing these improvements, users will have the ability to choose specific parts of the trace output.

Currently, when executing a file using the explorer with the --trace_file=- or --trace_file=filename.txt flag, the resulting output is an extensive and verbose log containing all the information.

In this PR, I have introduced the `ProgramPhase` enum class, which have distinct phases encountered during the compilation of a program in the explorer. Each member of this enum class corresponds to a specific phase, signifying the relevant information to be included in the trace output.

The phases covered by the `ProgramPhase` enum class are as follows:

1. Printing the source program
2. Name resolution
3. Control flow resolution
4. Type checking
5. Unformed variable resolution
6. Printing declarations
7. Printing the timings
8. Printing whole output.

These phases can be selected by passing the following compiler flags along with `--trace_file=-`.
`-trace_source_program`, `-trace_name_resolution`, `-trace_control_flow_resolution`, `-trace_type_checking`, `-trace_unformed_variables_resolution`, `-trace_declarations`, `-trace_execution`, `-trace_timing` and `-trace_all`. If none of these flags is passed only execution trace will be added to the output.

PR #2913, opened 2023-06-15
Clean up lit tests and config to reflect current uses.

### LINKED ISSUE #2912: Fix run-explorer lit uses broken by #2851
### Description of the bug:

#2851 modified run-explorer to remove the FileCheck, but didn't update pre-existing run-explorer uses to explicitly call FileCheck. As a consequence, these tests no longer do much.

Note, this may be a good chances to fix lit issues in general, but definitely this is a bug.

See https://github.com/carbon-language/carbon-lang/tree/trunk/explorer/lit_testdata

_No response_

### What actually happened?

_No response_

_No response_

jonmeow:
Overall, cleaning up remaining lit uses.

#2851 had removed FileCheck invocations from some of the explorer tests; this starts as just restoring that. But, now that we have far fewer `lit` tests, it seems best to refine `lit.cfg.py` to focus on providing fewer commands (not all were even used).

Also, adding testing of an error to explorer (which I noticed due to a change that would've broken that) made me notice that autoupdate_lit_test's for_lit logic didn't actually work, so I'm just cutting it and going to manual updates. Really, we might want to just remove lit autoupdate support altogether since it's only a couple tests using it, but I'm not ready to make that change right now.

Fixes #2912

### DIFF
--- a/explorer/autoupdate_lit_testdata.py
+++ b/explorer/autoupdate_lit_testdata.py
@@ -24,8 +24,9 @@ def main() -> None:
         # Flags to configure for explorer testing.
         "--tool=explorer",
         "--testdata=explorer/lit_testdata",
-        "--lit_run=%{explorer-run}",
-        "--lit_run=%{explorer-run-trace}",
+        "--lit_run=%{explorer} | %{FileCheck-strict}",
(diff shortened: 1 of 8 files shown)


chandlerc (approved):
LG

PR #2915, opened 2023-06-16
Add more information to trace state

prabhatexit0:
The `TraceState` has been invoked from two methods: `Interpreter::EvalRecursively` and `Interpreter::RunAllSteps`. 
With this enhancement, action kind and action details are added to the trace output before the `TraceState` is invoked.


explorer/interpreter/interpreter.cpp
```diff
@@ -576,8 +576,11 @@ auto Interpreter::StepLocation() -> ErrorOr<Success> {
 
 auto Interpreter::EvalRecursively(std::unique_ptr<Action> action)
     -> ErrorOr<Nonnull<const Value*>> {
+  auto action_kind = action->kind_string();
   if (trace_stream_->is_enabled()) {
-    *trace_stream_ << "--- recursive eval\n";
+    *trace_stream_ << "--- recursive eval for '" << action_kind << "'\n'''\n";
+    action->Print(trace_stream_->stream());
+    *trace_stream_ << "\n''' ---> \n";
```
zygoloid:
````suggestion
    *trace_stream_ << "--- recursive eval for '" << action_kind << "'\n```\n";
    action->Print(trace_stream_->stream());
    *trace_stream_ << "\n``` ---> \n";
````

May as well use markdown-like triple-quotes.

File: explorer/interpreter/interpreter.cpp [hunk 2]
```diff
@@ -2505,6 +2509,10 @@ auto Interpreter::Step() -> ErrorOr<Success> {
 auto Interpreter::RunAllSteps(std::unique_ptr<Action> action)
     -> ErrorOr<Success> {
   if (trace_stream_->is_enabled()) {
+    *trace_stream_ << "--- running all steps for '" << action->kind_string()
+                   << "'\n'''\n";
+    action->Print(trace_stream_->stream());
+    *trace_stream_ << "\n''' ---> \n";
```
zygoloid:
```suggestion
    *trace_stream_ << "--- running all steps for '" << action->kind_string()
                   << "'\n```\n";
    action->Print(trace_stream_->stream());
    *trace_stream_ << "\n``` ---> \n";
```

explorer/interpreter/interpreter.cpp
```diff
@@ -592,7 +595,8 @@ auto Interpreter::EvalRecursively(std::unique_ptr<Action> action)
     }
   }
   if (trace_stream_->is_enabled()) {
-    *trace_stream_ << "--- recursive eval done\n";
+    *trace_stream_ << "--- finished recursive eval for '" << action_kind
+                   << "'\n";
```
zygoloid, line 682: Maybe drop the `'`s here.

zygoloid (commented):
I think this is going to be helpful -- assuming the output isn't overwhelmingly verbose. Can you provide a before/after view of what this does do the trace output? Just putting it in a comment here on github would help.

I also think it'd be useful to have a test (with `AUTOUPDATE` turned on) where we produce and check a full trace of a simple sample program, so that we can easily see the effects of changes to tracing output such as this one.

prabhatexit0:
> Can you provide a before/after view of what this does do the trace output? Just putting it in a comment here on github would help.

**Before:**
````
********** calling main function **********
{
stack: 
memory: 0: {}, 1: Heap{}
}
--- step exp Main() .0. (<Main()>:0) --->
{
stack: Main .0. ## Main() .1.
memory: 0: {}, 1: Heap{}
}
--- step exp Main .0. (<Main()>:0) --->
{
stack: Main() .1. [[fun<Main>]]
memory: 0: {}, 1: Heap{}
}
...
````

**After:**
````
********** calling main function **********
--- running all steps for expression action
```
Main() .0.
``` ---> 
{
stack: 
memory: 0: {}, 1: Heap{}
}
--- step exp Main() .0. (<Main()>:0) --->
{
stack: Main .0. ## Main() .1.
memory: 0: {}, 1: Heap{}
}
--- step exp Main .0. (<Main()>:0) --->
{
stack: Main() .1. [[fun<Main>]]
memory: 0: {}, 1: Heap{}
}
...
````

File: explorer/interpreter/interpreter.cpp [hunk 4]
```diff
@@ -40,6 +40,33 @@ using llvm::isa;
...
+  if (!lines.empty()) {
+    res += lines.front();
+    res += lines.size() > 1 ? "\n ...\n" : "\n";
+  }
+
+  // Adds the last line
+  if (lines.size() > 1) {
+    res += lines.back().str() + "\n";
+  }
+  res += "\n```" + step.str();
```
zygoloid, line 66: Putting the last line outside the \`\`\` doesn't seem to be working very reliably. In the trace output below are things like:

```
// CHECK:STDOUT: ```
// CHECK:STDOUT:
// CHECK:STDOUT: ```i32 .0. --->{{ }}
```

and 

```
// CHECK:STDOUT: ```
// CHECK:STDOUT:
// CHECK:STDOUT: ```Main() .0. --->{{ }}
```

Maybe it'd be better to move this code to render a truncated snippet within \`\`\`s into the action printing logic, and wrap it around the printing of the declaration / expression.

Also, when printing a `DeclarationAction`, you could change the `Action::Print` function to use `Declaration::PrintID` instead of `Declaration::Print`, so we don't print out the body in the first place.

prabhatexit0: > Putting the last line outside the ``` doesn't seem to be working very reliably. In the trace output below are things like:

I'm sorry I missed that, thank you so much for pointing it out!
I'll do the suggested changes in the next commit.

PR #2916, opened 2023-06-16
Trace output filtering based on file context

prabhatexit0:
To filter based on file contexts, you can use `-trace_file_context=...` flag along with `--trace_file=...` flag. These are the following options you can pass to the `-trace_file_context=...` flag (you can also pass them as list by separating them with comma).
1. `main`: Include trace output for file containing the main function.
2. `prelude`: Include trace output for prelude.
3. `import`: Include trace output for imports.
4. `include`: Include trace output for all.

If the flag isn't used or non of the options are passed, by default only main file context is traced.

File contexts are distinguished based upon the source location by passing the source location to `TraceStream::is_enabled(..)` as arguments. If no arguments are passed, file context is `FileContext::Unknown` and for now it will add the trace from unknown file context.

Note:  `import` option currently doesn't work as imports aren't supported yet.


explorer/interpreter/action.h
```diff
@@ -121,6 +121,32 @@ class Action {
   // object.
   auto kind() const -> Kind { return kind_; }
 
+  auto kind_string() -> std::string_view {
```
zygoloid: It looks like you have some changes from the other PR mixed in here.

explorer/main.cpp
```diff
@@ -80,6 +80,20 @@ auto ExplorerMain(int argc, char** argv, void* static_for_main_addr,
...
+      cl::values(
+          clEnumValN(
+              FileContext::Main, "file_main",
+              "Include trace output for file containing the main function"),
+          clEnumValN(FileContext::Prelude, "file_prelude",
+                     "Include trace output for prelude"),
+          clEnumValN(FileContext::Import, "file_import",
+                     "Include trace output for imports"),
+          clEnumValN(FileContext::All, "file_all",
+                     "Include trace output for all files")));
```
zygoloid: These flag names should include `trace` -- `-trace_file_main`, `-trace_file_prelude`, etc maybe?

zygoloid (commented):
I think it would be better to set the current location on the `trace_stream` instead of passing it to each call to `is_enabled`. That would have a few benefits:

- We could add logging of the current location to the `trace_stream` infrastructure (for example, print the location before the first output in each location)
- All the trace output for a particular step would be enabled / disabled together, based on the location of that step, rather than having to pass the same location to each function that produces trace output within a step. At the moment, turning off trace output for a particular file will only disable the trace output that's been annotated with a location, and some of the output may still show up if it doesn't pass a location or passes a different location.
- We can do the work of figuring out whether trace output is currently enabled a little less often (only when we change source locations, not on each `is_enabled` call).

explorer/interpreter/action.h
prabhatexit0: Sorry, I missed removing it. Will do in the next commit.

explorer/main.cpp
prabhatexit0: Yes, it would make more sense. But I also think flag names are getting bigger, should we create short aliases as well?

prabhatexit0:
> I think it would be better to set the current location on the `trace_stream` instead of passing it to each call to `is_enabled`. That would have a few benefits:
> 
>     * We could add logging of the current location to the `trace_stream` infrastructure (for example, print the location before the first output in each location)
> 
>     * All the trace output for a particular step would be enabled / disabled together, based on the location of that step, rather than having to pass the same location to each function that produces trace output within a step. At the moment, turning off trace output for a particular file will only disable the trace output that's been annotated with a location, and some of the output may still show up if it doesn't pass a location or passes a different location.
> 
>     * We can do the work of figuring out whether trace output is currently enabled a little less often (only when we change source locations, not on each `is_enabled` call).

Thank you for this suggestion! I was particularly concerned about the second point you mentioned while working on this feature as well.

explorer/interpreter/action.h
```diff
@@ -249,8 +254,9 @@ class TypeInstantiationAction : public Action {
 // local context.
 class WitnessAction : public Action {
  public:
-  explicit WitnessAction(Nonnull<const Witness*> witness)
-      : Action(Kind::WitnessAction), witness_(witness) {}
+  explicit WitnessAction(Nonnull<const Witness*> witness,
+                         std::optional<SourceLocation> source_loc)
```
zygoloid: Can we make the `SourceLocation` mandatory for `WitnessAction` and for `InstantiateWitness`? Looking at the creations of `WitnessAction` below, it looks like we'll always have a source location.

explorer/interpreter/type_checker.cpp
```diff
@@ -2961,8 +2963,7 @@ auto TypeChecker::TypeCheckExp(Nonnull<Expression*> e,
                                const ImplScope& impl_scope)
     -> ErrorOr<Success> {
   return RunWithExtraStack([&]() { return TypeCheckExpImpl(e, impl_scope); });
-}
-
+};
```
zygoloid: Looks like this might be an accidental change.

explorer/main.cpp
```diff
@@ -80,6 +80,20 @@ auto ExplorerMain(int argc, char** argv, void* static_for_main_addr,
...
+          clEnumValN(
+              FileContext::Main, "trace_file_main",
+              "Include trace output for file containing the main function"),
+          clEnumValN(FileContext::Prelude, "trace_file_prelude",
+                     "Include trace output for prelude"),
+          clEnumValN(FileContext::Import, "trace_file_import",
+                     "Include trace output for imports"),
+          clEnumValN(FileContext::All, "trace_file_all",
+                     "Include trace output for all files")));
+
```
zygoloid: Can you switch to supporting `--trace_file=main` etc here?

zygoloid (changes requested):
Style nit: please can you go through the comments added in this patch and make sure they're all terminated by periods (`.`)?

explorer/main.cpp
prabhatexit0: Sure, pushing changes in the next commit.

explorer/interpreter/action.h
prabhatexit0: Yes, I made `SourceLocation` mandatory for `WitnessAction` and `InstantiateWitness` in one of the recent commits.
Thanks for pointing this out!

explorer/main.cpp
zygoloid: I think it'd be useful to add a "trace everything" flag that turns on tracing for all files and all steps, for debugging purposes. (But let's do that as a separate change.)

explorer/common/trace_stream.h
```diff
@@ -126,6 +176,30 @@ class SetProgramPhase {
...
+    trace_stream_.set_source_loc(source_loc);
+  }
+
+  // This can be used for cases when source location needs to be updated
+  // multiple times within the same scope.
+  auto update_source_loc(std::optional<SourceLocation> source_loc) {
+    trace_stream_.set_source_loc(source_loc);
+  }
+
+  ~SetFileContext() { trace_stream_.set_source_loc(initial_source_loc_); }
```
zygoloid: Please order the destructor after the constructor and before `update_source_loc` (and also for `SetProgramPhase`). Carbon's C++ style guide is based on the Google C++ style guide; see the [relevant section](https://google.github.io/styleguide/cppguide.html#Declaration_Order) on declaration order in classes.

explorer/common/trace_stream.h
```diff
@@ -45,12 +50,32 @@ enum class ProgramPhase {
 // contexts.
 class TraceStream {
  public:
+  explicit TraceStream() { set_allowed_file_contexts({FileContext::Unknown}); }
+
+  // This method gets the file context by using filename from source location
```
zygoloid:
```suggestion
  // This method gets the file context by using filename from source location.
```

zygoloid (approved):
Thanks, this looks great. Only a couple of minor stylistic comments left.

explorer/common/trace_stream.h
prabhatexit0: Okay

zygoloid (approved):

PR #2920, opened 2023-06-16
Semantics IR building for `if` statements.

zygoloid:
This follows the same structure as `if` expressions, except that no result value is needed.

Semantics IR building for code blocks is also added. Rather than popping all the node stack entries we push for statements within a code block, change statements and declarations to not push themselves onto the stack. We're not notionally performing work recursively within prior statements, and we don't need their value for anything, so it seems cleaner to not push them. This also allows statements and declarations to determine what syntactic context they're in by peeking at the top of the stack, though that's not used in this patch.


toolchain/semantics/semantics_handle_codeblock.cpp
```diff
@@ -0,0 +1,24 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "toolchain/semantics/semantics_context.h"
+#include "toolchain/semantics/semantics_node.h"
+
+namespace Carbon {
+
+auto SemanticsHandleCodeBlockStart(SemanticsContext& context,
+                                   ParseTree::Node parse_node) -> bool {
+  // TODO: Enter a new scope.
```
jonmeow: This should probably be calling PushScope/PopScope; was there a particular reason to leave that for a TODO?

toolchain/semantics/semantics_handle_if.cpp
```diff
@@ -0,0 +1,83 @@
...
+                                    ParseTree::Node parse_node) -> bool {
+  // Switch to emitting the else block.
+  auto then_block_id = context.node_block_stack().PopForAdd();
+  context.node_stack().Push(parse_node, then_block_id);
+  return true;
+}
+
+auto SemanticsHandleIfStatement(SemanticsContext& context,
+                                ParseTree::Node parse_node) -> bool {
+  auto sub_block_id = context.node_block_stack().PopForAdd();
```
jonmeow, line 53:
```suggestion
  // Either the `then` or `else` block; we need to look at the next node to find out.
  auto sub_block_id = context.node_block_stack().PopForAdd();
```

jonmeow (commented):
Nice change, but I'm wondering about node_stack push/pop behaviors being changed here (particularly noticed this because of the removed push around `return`).

So far semantics has been adding nodes to match the parse tree on node_stack on a 1:1 basis, to provide consistency. It means VLOG on node push/pop will pretty closely mirror parse tree structure, which is something I've used for debugging. Are you removing that guarantee because you don't think it offers useful consistency versus the overhead?

I think this gets a little weird to me in examples like IfCondition, where it leaves IfConditionStart on the stack rather than popping (taking care of all of its children) and pushing itself. As a consequence, the structure of the IfCondition subtree is required to be known by the parent IfStatement subtree, versus having some layering between. (although, that you could switch just by pushing IfCondition instead of IfConditionStart -- might want to consider that even if you think we should stop adding uninteresting node_stack entries)

toolchain/semantics/semantics_handle_codeblock.cpp
zygoloid: I just didn't realize those already existed :) Done, and test added.

toolchain/semantics/semantics_handle_if.cpp
zygoloid: Done, but rephrased (I found "the next node" unclear in this context).

zygoloid:
> So far semantics has been adding nodes to match the parse tree on node_stack on a 1:1 basis, to provide consistency. It means VLOG on node push/pop will pretty closely mirror parse tree structure, which is something I've used for debugging. Are you removing that guarantee because you don't think it offers useful consistency versus the overhead?

Yeah, I think that property is not worth the overhead, and would also get in the way of (for example) the handlers for function declarations / variable declarations determining whether they're declaring a member function by looking at what's at the top of the stack. I think we should only be pushing nodes onto the node stack if there's some later use of them, either to track a value, or as meaningful bracketing, or to track what kind of construct we're processing, or something like that.

If you want the VLOGs to line up, we could push and then immediately pop in these cases, or add some function to produce a VLOG. Or we could VLOG each parse tree node before we handle it:

```
FunctionDefinitionStart:
Node Pop 2: ParameterList -> block6
Node Pop 1: DeclaredName -> node<invalid>
Node Pop 0: FunctionIntroducer -> node<invalid>
node_block_stack_ Push 1
node_block_stack_ Add 1: block7
AddNode block1: {kind: FunctionDeclaration, arg0: function3}
Node Push 0: FunctionDefinitionStart -> function3

IfConditionStart:
Node Push 1: IfConditionStart -> <none>

NameReference:
Node Push 2: NameReference -> node+3

IfCondition:
Node Pop 2: NameReference -> node+3
node_block_stack_ Pop 1: block7
node_block_stack_ Push 1
node_block_stack_ Add 1: block8
node_block_stack_ Push 2
node_block_stack_ Add 2: block9
AddNode block7: {kind: BranchIf, arg0: block9, arg1: node+3}
AddNode block7: {kind: Branch, arg0: block8}

CodeBlockStart:
Node Push 2: CodeBlockStart -> <none>

NameReference:
Node Push 3: NameReference -> node+0

CallExpressionStart:
Node Pop 3: NameReference -> node+0
Node Push 3: CallExpressionStart -> node+0
params_or_args_stack_ Push 0

CallExpression:
params_or_args_stack_ Pop 0: block<invalid>
Node Pop 3: CallExpressionStart -> node+0
AddNode block9: {kind: Call, arg0: block0, arg1: function0}
Node Push 3: CallExpression -> node+8

ExpressionStatement:
Node Pop 3: CallExpression -> node+8

CodeBlock:
Node Pop 2: CodeBlockStart -> node<invalid>

IfStatementElse:
node_block_stack_ Pop 2: block9
Node Push 2: IfStatementElse -> block9

CodeBlockStart:
Node Push 3: CodeBlockStart -> <none>

NameReference:
Node Push 4: NameReference -> node+1

CallExpressionStart:
Node Pop 4: NameReference -> node+1
Node Push 4: CallExpressionStart -> node+1
params_or_args_stack_ Push 0

CallExpression:
params_or_args_stack_ Pop 0: block<invalid>
Node Pop 4: CallExpressionStart -> node+1
AddNode block8: {kind: Call, arg0: block0, arg1: function1}
Node Push 4: CallExpression -> node+9

ExpressionStatement:
Node Pop 4: CallExpression -> node+9

CodeBlock:
Node Pop 3: CodeBlockStart -> node<invalid>

IfStatement:
node_block_stack_ Pop 1: block8
Node Pop 2: IfStatementElse -> block9
node_block_stack_ Push 1
node_block_stack_ Add 1: block10
AddNode block9: {kind: Branch, arg0: block10}
AddNode block8: {kind: Branch, arg0: block10}
Node Pop 1: IfConditionStart -> node<invalid>
```

> I think this gets a little weird to me in examples like IfCondition, where it leaves IfConditionStart on the stack rather than popping (taking care of all of its children) and pushing itself. As a consequence, the structure of the IfCondition subtree is required to be known by the parent IfStatement subtree, versus having some layering between. (although, that you could switch just by pushing IfCondition instead of IfConditionStart -- might want to consider that even if you think we should stop adding uninteresting node_stack entries)

Switched to pushing `IfCondition` as the "no else" marker instead.

jonmeow (approved):
Yeah, maybe this is just reason to look closer at VLOG. Let's handle that separately, we can merge this change.

PR #2922, opened 2023-06-17
Introduce a new Conduct team

chandlerc:
The Carbon Project relied on the three leads to handle conduct concerns initially as it bootstrapped its community team and expertise. We now have an active and effective community lead and team of moderators. Our community lead has worked to train and ramp up a new and independent conduct team.

This proposal both provides an overview of the process and hands off conduct handling to the new team! Going forward, we expect routine updates to the conduct team to happen without full proposals as they allow trained folks to rotate in and out of this difficult but essential role on the project.

Last but not least, making these changes uncovered a restriction in the Code of
Conduct itself that we expect to be problematic to adhere to going forward.
While well intentioned, it has a bunch of unanticipated effects that made both
current and new conduct teams want to remove it. A related section has had its
wording strengthened to try and address the underlying motivation at least
partially.

Most of this content comes from @CelineausBerlin with some additions and edits from myself and the other leads as we worked toward building the new CoC team.


CelineausBerlin (approved):
Great, thank you Chandler!

CODE_OF_CONDUCT.md
```diff
@@ -145,19 +145,10 @@ any member of the conduct team directly.
...
 The conduct team can be emailed at conduct@carbon-lang.dev, and currently has
 the following members:
 
--   Chandler Carruth (@chandlerc on Discord and
-    [GitHub](https://github.com/chandlerc))
--   Kate Gregory (@KateGregory on Discord and
-    [GitHub](https://github.com/KateGregory))
--   Richard Smith (@zygoloid on Discord and
-    [GitHub](https://github.com/zygoloid))
+-   TBA
```
geoffromer: I assume this will be updated before merging?

proposals/p2922.md
```diff
@@ -0,0 +1,153 @@
...
+So, it is time to build a specialized team to deal with escalations which is not
+subordinate to the Carbon leads, and is able to handle diverse misconduct
+escalations in an adequate and timely manner. This means we can recruit and/or
+train people from diverse backgrounds to handle such escalations for the Carbon
+community, and provide competent support to the leads, the rest of the
+moderation team, and everyone else in the community.
+
+TODO: What problem are you trying to solve? How important is that problem? Who
+is impacted by it?
+
```
geoffromer:
```suggestion
```

CODE_OF_CONDUCT.md
```diff
@@ -145,19 +145,10 @@ any member of the conduct team directly.
 
 ## Conduct team
```
geoffromer, line 146: It would probably be worth documenting the process for changing the composition of the team (in particular, the fact that the conduct team can add/remove members without a proposal)

proposals/p2922.md
```diff
@@ -0,0 +1,153 @@
...
+### Initial team members
+
+-   TBA
+
+We will continue to look for more candidates in order to get to a point where we
+have a sustainable number, especially as the community grows.
+
+### Process
+
+Selection and sourcing:
```
geoffromer, line 165: Do we want to say anything about how this should happen going forward, e.g. any goals or criteria that the conduct team should consider in selecting new members? In particular, do we prefer the conduct team to be drawn from the Carbon community (so that they have on-the-ground knowledge and context), or do we prefer for them to be fully independent?

proposals/p2922.md
```diff
@@ -0,0 +1,153 @@
...
+    a senior moderation team member.
+-   We organize regular informal moderation meetups to learn with and from each
+    other, including a book club.
+
+Exit:
+
+-   We are aware that this work tends to be straining and that each person's
+    involvement may only be temporary.
+-   We will make sure that conduct team members feel safe to leave the team and
+    community when they need to step out by developing a bench of candidates.
```
geoffromer: Who's "we"? Do people outside the conduct team have an ongoing role in developing that bench?

zygoloid (approved):
This direction has my full support :)

CODE_OF_CONDUCT.md
chandlerc: Yep, and now they are in place!

proposals/p2922.md
chandlerc: Sure, I think there was more of that in the original draft documents I was pulling from. I've added a bunch of that content, and tried to restructure and reword stuff to flow a bit better. Hopefully I didn't mess any of it up in the process, @CelineausBerlin - please double check my work here!

That said, I'm not sure the original content talked much about the specific aspect you mentioned -- whether they should be "from" the Carbon community or not. Not sure if Céline or others have already thought about this?

CODE_OF_CONDUCT.md
chandlerc: Done. I've created a dedicated team page to cover this.

proposals/p2922.md
chandlerc: This specific case is talking about the proposers and the current conduct team. Proposal docs often use "we" I think? If there is a better wording, happy to update!

chandlerc (commented):
Thanks for reviews, many updates here.

Notably, this includes a change to the CoC itself to fix an issue uncovered in the process of doing this handover.

And also notably, this includes all the new conduct team members!!!

proposals/p2922.md
CelineausBerlin: Hi! Let me try to explain my thinking and see if there is something to be added to the latest version:
We _"typically" pick people from the Carbon moderation team_, because it allows us to see how a person is actually behaving in our spaces, and whether they are demonstrating the required skills and competence as mentioned in this as well (e.g. "Advanced moderation and mediation skills", "Intercultural competence", etc.), or not.
_"On-the-ground" knowledge is not a requirement_, but comes with interacting with our community, sooner or later, as a Carbon moderator.
The current state of affairs is that people can remain active moderators after joining the CoC team, or focus on conduct work only. It mostly depends on personal availability and interest.

With that said, how can we improve this version so those questions are answered more clearly, or maybe more concretely, @geoffromer?

CelineausBerlin (approved):
LGTM

zygoloid (approved):

proposals/p2922.md
geoffromer: This looks good. Thanks!

proposals/p2922.md
```diff
@@ -71,26 +78,87 @@ Our current CoC team:
...
+Team members, so at least 3 are available to discuss and act upon incident
+reports in a timely and adequate manner.
+
+The initial team was recruited by our Community lead and pulls from the current
+trained moderator pool. We propose that the Conduct team itself is responsible
+to continue recruiting both new Conduct team members and new moderators to make
+both efforts sustainable. They will work toward better representation, embracing
+both broad and intersectional diversity.
+
+### What we are looking for in Conduct team members
```
geoffromer, line 121: Should this maybe go in `conduct_team.md`, for greater discoverability and because it represents ongoing guidance rather than one-off context for this specific proposal?

proposals/p2922.md
```diff
@@ -0,0 +1,153 @@
...
+## Details
+
+### Initial team members
+
+-   TBA
+
+We will continue to look for more candidates in order to get to a point where we
+have a sustainable number, especially as the community grows.
+
+### Process
```
geoffromer, line 163: Some parts of this section seem like they're intended to remain true over time (e.g. "New members will have access..." and "We will make sure...") rather than being tied to this proposal -- should those go in `conduct_team.md`?

proposals/p2922.md
geoffromer: This is a forward-looking, ongoing commitment (unlike most uses of "we" in proposals), so it seems important to be clear about who is responsible for following through on it.

> This specific case is talking about the proposers and the current conduct team.

Since this is an ongoing responsibility, it seems like it would be better to vest it in an ongoing body like the leads or the conduct team, rather than ~4 specific people. It also seems like the conduct team ought to have a role in developing the bench, since they'll be the ones selecting people from it.

proposals/p2922.md
chandlerc: I don't think thot proposals are *strictly* one-off? They just aren't explicitly maintained *living* documents.

I can try to also create some maintained, living documentation for how the conduct team, but just putting it into `conduct_team.md` seems a bit awkward. And I'm hesitant to drive creating the living documentation for the conduct team as I'm not going to be *on* that team going forward.

For language features, we routinely create a proposal first, and then incorporate it as needed into the living design document. If, when, and where the conduct team wants to create a persistent document that incorporates this, I think that would be OK to do in follow-up PRs, and those wouldn't need to be proposals.

For now, rather than spending time figuring out the right structure and wording to make these effective as on-going living documentation, I'd prefer to start here.

proposals/p2922.md
chandlerc: See above -- basically not sure this needs to happen in the critical path, and not sure I'm the best person to create the long-term documentation there.

proposals/p2922.md
chandlerc: Ok, I'm not sure I fully understand the concern here, but I've reworded these sections to hopefully address it.

proposals/p2922.md
geoffromer: The rule of thumb that I prefer is that any information that has ongoing, non-historical relevance should live somewhere other than a proposal document, both because proposals aren't living documents and because the information in them isn't very discoverable.

However, you're right that we can (and often do) accomplish that with follow-up PRs, and you make a good point that the conduct team should probably take the lead on those documents, so I'm OK with deferring it for now.

proposals/p2922.md
geoffromer: This looks good, thanks.

geoffromer (approved):

proposals/p2922.md
chandlerc: Double checked with the leads to make sure they're comfortable with this plan, and they are so we should be good to land this as is and then any improvements to more long-term docs can be follow-up work led by the conduct team themselves.

chandlerc:
Thanks again all for the reviews!!

I've double checked and all three leads are happy so we're landing this and handing over CoC to the new team with a *huge* thank you and much appreciation!

PR #2924, opened 2023-06-17
Update BUILD

madhupashish:
The provided code, you can make the following changes:

1. Update the `py_binary` and `py_test` rules to use the `python_version` attribute instead of the deprecated `PY3` value. You can set it to the desired Python version, such as `PY37` for Python 3.7.

2. Verify and update the `data` and `deps` attributes if necessary. Ensure that the `template.md` file exists in the specified location and adjust the dependencies (`deps`) according to your project structure.

That's it! These changes will update the code to use the proper `python_version` attribute and ensure that the code is compatible with the specified Python version.

adding, and why.**

Closes #ISSUE

### DIFF
--- a/proposals/scripts/BUILD
+++ b/proposals/scripts/BUILD
@@ -5,13 +5,13 @@
 py_binary(
     name = "new_proposal",
     srcs = ["new_proposal.py"],
-    python_version = "PY3",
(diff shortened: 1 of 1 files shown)


madhupashish:
Must View This Code

PR #2925, opened 2023-06-17
README.md: Add links to C++Now 2023 talks slides

Pixep:


README.md
```diff
@@ -300,6 +300,14 @@ Past Carbon focused talks from the community:
...
 -   [Carbon Language: Syntax and trade-offs](https://www.youtube.com/watch?v=9Y2ivB8VaIs),
     Core C++ 2022
+-   [Carbon’s Successor Strategy: From C++ interop to memory safety (Slides)](https://chandlerc.blog/slides/2023-cppnow-carbon-strategy/index.html#/),
+    C++Now 2023
+-   [Definition-Checked Generics (Part 1) (Slides)](https://chandlerc.blog/slides/2023-cppnow-generics-1/#/),
+    C++Now 2023
+-   [Definition-Checked Generics (Part 2) (Slides)](https://chandlerc.blog/slides/2023-cppnow-generics-2/#/),
+    C++Now 2023
+-   [Modernizing Compiler Design for Carbon’s Toolchain (Slides)](https://chandlerc.blog/slides/2023-cppnow-compiler/index.html#/),
+    C++Now 2023
```
jonmeow:
```suggestion
-   C++Now 2023:
    -   Carbon’s Successor Strategy: From C++ interop to memory safety [(Slides)](https://chandlerc.blog/slides/2023-cppnow-carbon-strategy/index.html#/)
    -   Definition-Checked Generics: Part 1 [(Slides)](https://chandlerc.blog/slides/2023-cppnow-generics-1/#/) and Part 2 [(Slides)](https://chandlerc.blog/slides/2023-cppnow-generics-2/#/)
    -   Modernizing Compiler Design for Carbon’s Toolchain [(Slides)](https://chandlerc.blog/slides/2023-cppnow-compiler/index.html#/)
```

How does this format read to you? I'm wondering if it's worth it to switch a little so that we don't repeat as much. Also, maybe we should decide whether to link video/slides. But also, if you just want to merge as-is, I think that's reasonable too.

jonmeow (approved):

jonmeow:
Thanks for pulling these links together here!

README.md
Pixep: I thought about that, I think it creates an asymmetry in the content that I don't think is justified (all talks have the same importance). But maybe we could separate in 2022, and 2023? That way we see how recent those are more easily. Still some repetition with the event name I suppose, but I think that's acceptable?

jonmeow: Clustering by year might be beneficial in a different way, in that it makes it clear up front when content is older. It's fine if you leave the conference at the end.

I'd still merge the two parts of the generics talk. That's not about each talk having the same importance, it's about making it clear it's really one long talk. :)

Pixep: Good point for the two parts, merged. Added year too.
The reason for the slides is that only slides are available currently for C++Now. Otherwise video would be preferred (at least for consistency if anything).

jonmeow (approved):

PR #2926, opened 2023-06-18
Add type_utils file

kshokhin:
Prerequisite for #2909. Moves some type-related functions to the separate file.


explorer/interpreter/type_utils.h
```diff
@@ -0,0 +1,55 @@
...
+
+// Attempts to match `v` against the pattern `p`, returning whether matching
+// is successful. If it is, populates **bindings with the variables bound by
+// the match; `bindings` should only be nullopt in contexts where `p`
+// is not permitted to bind variables. **bindings may be modified even if the
+// match is unsuccessful, so it should typically be created for the
+// PatternMatch call and then merged into an existing scope on success.
+// The matches for generic variables in the pattern are output in
+// `generic_args`.
+[[nodiscard]] auto PatternMatch(
```
geoffromer: I wouldn't expect to find this in a header called `type_utils.h`. Maybe we should call it `common.h` or something like that instead?

kshokhin: Agree, `type_utils.h` is not the best name.

Wouldn't `common.h` be too generic? I feel like the file with the name `common.h` could absorb lots of unrelated stuff, which seems too small to put into a separate file and doesn't fit any other existing files. As a result the file could become rather big over time, and one has to include the whole file because of the one function/struct/class/enum/whatever needed.

Maybe it would be better to split it into multiple files?
E.g.
`PatternMatch` goes to `pattern_match.h`.
Other functions go to `type_properties.h` or `type_helpers.h` or maybe `type_common.h`, as all of them tell some properties of the type.

geoffromer: Splitting into `pattern_match.h` and a header of type-related stuff sounds good to me. I'd say stick with `type_utils.h` for the latter, mainly because that seems like the least surprising choice.

geoffromer:
Sorry for the long delay, I didn't realize this was ready for review.

This looks good now, but it looks like there are merge conflicts that need to be resolved.

geoffromer (approved):

PR #2930, opened 2023-06-20
Fix a fuzzer-found crash.

chandlerc:
The crash occurred because of an alias target of a mixin. While eventually, we should probably have some ability to alias mixins, this isn't yet setup in the explorer and doesn't seem like a current priority.

We got here because the mixin type checking made it far enough to not reject this within the type checker, but the next step wasn't prepared for this to come out of the type checker. The simplest fix seems to be to reflect that it *can* escape the type checker, but still isn't (yet) a valid alias target.

Test case added.


jonmeow (approved):
Thanks! This feels like a good workaround for now.

PR #2932, opened 2023-06-20
Implemented object code generation option in the tool chain

fasiddique:
Takes a carbon file that returns either 0 or 1 and outputs the assembly to the stdout.


WORKSPACE
```diff
@@ -113,7 +113,9 @@ http_archive(
 
 # We pin to specific upstream commits and try to track top-of-tree reasonably
 # closely rather than pinning to a specific release.
-llvm_version = "c5f6a287499a816cba5585708999e2c8b134290f"
+
+llvm_version = "0258a53521cfedf5cb80c2b1d4a66c942615de74"
```
jonmeow: This file shouldn't be getting changed -- https://github.com/carbon-language/carbon-lang/pull/2895 did the update on trunk, and you should be using that now. It looks like you might've grabbed the wrong versions when merging.

toolchain/codegen/BUILD
```diff
@@ -0,0 +1,22 @@
...
+    deps = [
+        "@llvm-project//llvm:AArch64AsmParser",
+        "@llvm-project//llvm:AArch64CodeGen",
+        "@llvm-project//llvm:Core",
+        "@llvm-project//llvm:MC",
+        "@llvm-project//llvm:Support",
+        "@llvm-project//llvm:Target",
+        "@llvm-project//llvm:TargetParser",
+        "@llvm-project//llvm:X86AsmParser",
+        "@llvm-project//llvm:X86CodeGen",
```
jonmeow:
```suggestion
        "@llvm-project//llvm:AllTargetsAsmParsers",
        "@llvm-project//llvm:AllTargetsCodeGens",
        "@llvm-project//llvm:Core",
        "@llvm-project//llvm:MC",
        "@llvm-project//llvm:Support",
        "@llvm-project//llvm:Target",
        "@llvm-project//llvm:TargetParser",
```

Digging again in the [BUILD file](https://github.com/llvm/llvm-project/blob/main/utils/bazel/llvm-project-overlay/llvm/BUILD.bazel#L2395), I think these `AllTargets` rules avoid the need to specify targets individually.

toolchain/codegen/codegen.h
```diff
@@ -0,0 +1,20 @@
...
+#include "llvm/IR/Module.h"
+
+namespace Carbon {
+
+class CodeGen {
+ public:
+  llvm::raw_ostream& output = llvm::outs();
+  void GenerateObjCodeFromModule(llvm::Module& module);
+};
+}  // namespace Carbon
```
jonmeow, line 17:
```suggestion

}  // namespace Carbon

```

Spacing for style

toolchain/codegen/codegen.cpp
```diff
@@ -0,0 +1,69 @@
...
+  llvm::InitializeAllAsmParsers();
+  llvm::InitializeAllAsmPrinters();
+
+  auto target_triple = llvm::sys::getDefaultTargetTriple();
+
+  std::string error;
+  const auto* target = llvm::TargetRegistry::lookupTarget(target_triple, error);
+
+  if (!target) {
+    output << error;
```
jonmeow: Errors should be send to stderr.

toolchain/codegen/codegen.cpp
```diff
@@ -0,0 +1,69 @@
...
+  // if (ec) {
+  //   output << "Could not open file: " << ec.message();
+  //   return;
+  // }
+
+  llvm::legacy::PassManager pass;
+  auto file_type = llvm::CGFT_AssemblyFile;  // llvm::CGFT_ObjectFile;
+
+  if (target_machine->addPassesToEmitFile(pass, dest, nullptr, file_type)) {
+    output << "Could not write to object file\n";
```
jonmeow: Looking at the code, the `false` return indicates that a stop option was set. Doesn't that mean there's just nothing to write?
https://llvm.org/doxygen/classllvm_1_1TargetPassConfig.html#ad0db8596710a5666b67e513da0d9b415

toolchain/codegen/codegen.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+  // if (ec) {
+  //   output << "Could not open file: " << ec.message();
+  //   return;
+  // }
+
+  llvm::legacy::PassManager pass;
+  auto file_type = llvm::CGFT_AssemblyFile;  // llvm::CGFT_ObjectFile;
+
+  if (target_machine->addPassesToEmitFile(pass, dest, nullptr, file_type)) {
```
jonmeow: What happens if you pass the `output` stream directly?

toolchain/codegen/codegen.cpp
```diff
@@ -0,0 +1,69 @@
...
+  // llvm::raw_fd_ostream dest(file_name, ec, llvm::sys::fs::OF_None);
+  llvm::raw_fd_ostream dest(fileno(stdout), false);
+
+  // if (ec) {
+  //   output << "Could not open file: " << ec.message();
+  //   return;
+  // }
+
+  llvm::legacy::PassManager pass;
+  auto file_type = llvm::CGFT_AssemblyFile;  // llvm::CGFT_ObjectFile;
```
jonmeow: Maybe this should be called `dump assembly` rather than `dump object-code`?

jonmeow:
```suggestion
  llvm::raw_fd_ostream dest(fileno(stdout), false);

  llvm::legacy::PassManager pass;
  auto file_type = llvm::CGFT_AssemblyFile;
```

Can the commented out code either be removed or consolidated? You're going to need a separate code path if you keep the current flow, so you wouldn't be able to just uncomment them. If you consolidate them, please add a TODO explaining your plan.

toolchain/codegen/codegen.cpp
```diff
@@ -0,0 +1,69 @@
...
+  // std::error_code ec;
+  // llvm::raw_fd_ostream dest(file_name, ec, llvm::sys::fs::OF_None);
+  llvm::raw_fd_ostream dest(fileno(stdout), false);
+
+  // if (ec) {
+  //   output << "Could not open file: " << ec.message();
+  //   return;
+  // }
+
+  llvm::legacy::PassManager pass;
```
jonmeow, line 57: Should this be using a `ModulePassManager` instead of the `legacy::PassManager`?

toolchain/codegen/codegen.cpp
```diff
@@ -0,0 +1,69 @@
...
+  if (!target) {
+    output << error;
+    return;
+  }
+
+  const auto* cpu = "generic";
+  const auto* features = "";
+
+  llvm::TargetOptions target_opts;
+  auto reloc_model = std::optional<llvm::Reloc::Model>();
```
jonmeow:
```suggestion
  std::optional<llvm::Reloc::Model> reloc_model;
```

toolchain/driver/driver.h
```diff
@@ -7,11 +7,21 @@
 
 #include <cstdint>
 
+#include "clang/Driver/Compilation.h"
+#include "clang/Driver/Driver.h"
+#include "clang/Frontend/TextDiagnosticPrinter.h"
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/StringRef.h"
+#include "llvm/IR/LegacyPassManager.h"
+#include "llvm/MC/TargetRegistry.h"
```
jonmeow: You shouldn't still be adding includes here.

toolchain/codegen/codegen.h
```diff
@@ -0,0 +1,20 @@
...
+
+#include "llvm/IR/Module.h"
+
+namespace Carbon {
+
+class CodeGen {
+ public:
+  llvm::raw_ostream& output = llvm::outs();
+  void GenerateObjCodeFromModule(llvm::Module& module);
+};
```
jonmeow:
```suggestion
void GenerateObjCodeFromModule(llvm::Module& module);
```

We prefer free functions when the class isn't used.

Also, please add comments explaining what this does.

toolchain/driver/driver.cpp
```diff
@@ -236,6 +241,12 @@ auto Driver::RunDumpSubcommand(DiagnosticConsumer& consumer,
                   /*IsForDebug=*/true);
   }
 
+  if (dump_mode == DumpMode::ObjectCode) {
+    CodeGen code_gen = CodeGen();
```
jonmeow: Other comments of mine suggest getting rid of this class, but note that explicitly calling the constructor isn't needed.

toolchain/driver/driver.cpp
```diff
@@ -236,6 +241,12 @@ auto Driver::RunDumpSubcommand(DiagnosticConsumer& consumer,
...
+    code_gen.GenerateObjCodeFromModule(*module);
```
jonmeow: What's your plan for how this'll look when actually generating, versus printing as you do here? Should this have a different function name?

jonmeow (commented):
Thanks! Doing a quick review to give you comments to look through as we head to the meeting.

toolchain/codegen/codegen.cpp
fasiddique: The function takes  `PassManagerBase` object and `ModulePassManager`  doesn't inherit this class.
https://llvm.org/doxygen/classllvm_1_1LLVMTargetMachine.html
https://llvm.org/doxygen/classllvm_1_1legacy_1_1PassManagerBase.html

jonmeow: Per discussion, it turns out the legacy pass manager is still used for this detail.
https://github.com/llvm/llvm-project/blob/main/clang/lib/CodeGen/BackendUtil.cpp#L1071-L1074
https://llvm.org/docs/NewPassManager.html#status-of-the-new-and-legacy-pass-managers

toolchain/driver/BUILD
```diff
@@ -14,15 +14,25 @@ cc_library(
...
+        "@llvm-project//clang:frontend",
+        "@llvm-project//llvm:AArch64AsmParser",
+        "@llvm-project//llvm:AArch64CodeGen",
         "@llvm-project//llvm:Core",
+        "@llvm-project//llvm:MC",
         "@llvm-project//llvm:Support",
+        "@llvm-project//llvm:Target",
+        "@llvm-project//llvm:TargetParser",
+        "@llvm-project//llvm:X86AsmParser",
+        "@llvm-project//llvm:X86CodeGen",
```
jonmeow:
```suggestion
        "@llvm-project//llvm:Core",
        "@llvm-project//llvm:Support",
```

[note: these are probably deps that you added while drafting the initial change, but should now be handled in codegen]

toolchain/driver/driver.h
```diff
@@ -7,11 +7,19 @@
...
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Support/Debug.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/TargetSelect.h"
 #include "llvm/Support/VirtualFileSystem.h"
 #include "llvm/Support/raw_ostream.h"
+#include "llvm/Target/TargetMachine.h"
+#include "llvm/Target/TargetOptions.h"
+#include "llvm/TargetParser/Host.h"
```
jonmeow:
```suggestion
#include "llvm/ADT/ArrayRef.h"
#include "llvm/ADT/StringRef.h"
#include "llvm/Support/Debug.h"
#include "llvm/Support/VirtualFileSystem.h"
#include "llvm/Support/raw_ostream.h"
```

toolchain/driver/driver.cpp
```diff
@@ -131,10 +133,11 @@ auto Driver::RunDumpSubcommand(DiagnosticConsumer& consumer,
                        .Case("parse-tree", DumpMode::ParseTree)
                        .Case("semantics-ir", DumpMode::SemanticsIR)
                        .Case("llvm-ir", DumpMode::LLVMIR)
+                       .Case("assembly", DumpMode::Assembly)
                        .Default(DumpMode::Unknown);
   if (dump_mode == DumpMode::Unknown) {
     error_stream_ << "ERROR: Dump mode should be one of tokens, parse-tree, "
-                     "semantics-ir, or llvm-ir.\n";
+                     "semantics-ir, llvm-ir or assembly.\n";
```
```suggestion
                     "semantics-ir, llvm-ir, or assembly.\n";
```

Missed this one before -- we prefer Oxford commas.

toolchain/driver/driver.cpp
```diff
@@ -222,7 +225,9 @@ auto Driver::RunDumpSubcommand(DiagnosticConsumer& consumer,
   llvm::LLVMContext llvm_context;
   const std::unique_ptr<llvm::Module> module =
       LowerToLLVM(llvm_context, input_file_name, semantics_ir, vlog_stream_);
+
   CARBON_VLOG() << "*** LowerToLLVM done ***\n";
+
```
jonmeow, line 237:
```suggestion
  CARBON_VLOG() << "*** LowerToLLVM done ***\n";
```

Minor, but the lack of spacing here is deliberately both for consistency with the above, and so that substantial steps are a block.

toolchain/codegen/codegen.cpp
```diff
@@ -0,0 +1,65 @@
...
+  std::string error;
+  const auto* target = llvm::TargetRegistry::lookupTarget(target_triple, error);
+
+  if (!target) {
+    error_stream << error;
+    return;
+  }
+
+  const auto* cpu = "generic";
+  const auto* features = "";
```
jonmeow:
```suggestion
  constexpr llvm::StringLiteral cpu = "generic";
  constexpr llvm::StringLiteral features = "";
```

Subtle thing about constant strings, StringLiteral will yield slightly better results in certain cases, and makes for a good habit.

toolchain/codegen/codegen.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+  llvm::TargetOptions target_opts;
+  std::optional<llvm::Reloc::Model> reloc_model =
+      std::optional<llvm::Reloc::Model>();
```
jonmeow:
```suggestion
  std::optional<llvm::Reloc::Model> reloc_model;
```

No need to have the constructor call.

toolchain/codegen/codegen.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/TargetSelect.h"
+#include "llvm/Target/TargetMachine.h"
+#include "llvm/Target/TargetOptions.h"
+#include "llvm/TargetParser/Host.h"
+
+namespace Carbon {
+
+void PrintAssemblyFromModule(llvm::Module& module) {
+  llvm::raw_ostream& error_stream = llvm::errs();
```
jonmeow: Rather than using llvm::errs and llvm::outs directly, these should be parameters passed from the Driver, which has `output_stream_` and `error_stream_` as members that you can use for this purpose (i.e., take a `llvm::raw_ostream&` parameter for each, and pass them along). That way, the stream they're going to can easily be changed (which we use in certain tests, and may eventually want to do here).

jonmeow (commented):
As discussed, make sure you reset the workspace changes (just use the version on trunk) because the update was done separately, and to a slightly later version. Switching should fix the test issues you're currently running into.

toolchain/codegen/codegen.cpp
jonmeow: Also llvm::outs

toolchain/driver/testdata/codegen.carbon
```diff
@@ -0,0 +1,8 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// RUN: %{carbon} dump assembly --target_triple=x86_64-unknown-linux-gnu %s | %{FileCheck-allow-unmatched}
```
jonmeow, line 5: Now that you've added --target_triple, can you add a test of an invalid target triple to check the error?

jonmeow (commented):
I think you need to rebase this PR onto trunk. You appear to have merged the changes from #2937 some other way, and as a consequence, they're showing in the delta even though #2937 is merged. (look at "File changes" in GitHub, "all commits" -- that should only show the relevant delta)

toolchain/driver/driver.cpp
```diff
@@ -236,6 +246,15 @@ auto Driver::RunDumpSubcommand(DiagnosticConsumer& consumer,
                   /*IsForDebug=*/true);
   }
 
+  if (dump_mode == DumpMode::Assembly) {
+    consumer.Flush();
+    has_errors |= PrintAssemblyFromModule(*module, target_triple, error_stream_,
+                                          output_stream_);
+    if (!has_errors) {
+      return !has_errors;
+    }
```
jonmeow: What happens if has_errors is true?

toolchain/codegen/codegen.h
```diff
@@ -0,0 +1,22 @@
...
+
+#ifndef CARBON_TOOLCHAIN_CODEGEN_CODEGEN_H_
+#define CARBON_TOOLCHAIN_CODEGEN_CODEGEN_H_
+
+#include <cstdint>
+
+#include "llvm/IR/Module.h"
+
+namespace Carbon {
+inline bool codegen_has_errors = false;
```
jonmeow: This shouldn't be here. Globals are generally undesirable (they prevent multi-threaded execution), but also it's only used locally inside PrintAssemblyFromModule.

toolchain/driver/testdata/failed_codegen_invalid_triple.carbon
```diff
@@ -0,0 +1,8 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// RUN:  %{not} %{carbon} dump assembly --target_triple=x86_687-unknown-linux-gnu %s | %{FileCheck-allow-unmatched}
```
jonmeow:
```suggestion
// RUN: %{not} %{carbon} dump assembly --target_triple=x86_687-unknown-linux-gnu %s | %{FileCheck-strict}
```

Consider making this strict, so that crash bugs are flagged.

jonmeow:
See #2944. @fasiddique Please close PRs if you're switching.

PR #2933, opened 2023-06-21
CI: run tests for GitHub merge groups

Pixep:
Experimentations on workflow file only, as I don't have other permissions.


jonmeow (approved):

jonmeow:
Seems fine to experiment, but I'm not merging in case there's something extra you wanted to do here.

PR #2934, opened 2023-06-21
Check explorer's full trace output 

prabhatexit0:
Created `explorer/trace_testdata/full_trace.carbon` to test the whole trace and some changes in `explorer/file_test.cpp` to treat `/trace_testdata/` tests differently.


explorer/autoupdate_testdata.py
```diff
@@ -25,6 +25,16 @@ def main() -> None:
...
+
+    args = [
+        str(autoupdate_py),
+        # Flags to configure for explorer testing.
+        "--tool=explorer",
+        "--testdata=explorer/testdata/trace",
+        "--autoupdate_arg=--trace_file=-",
+        "--autoupdate_arg=-trace_all",
+    ] + sys.argv[1:]
     exit(subprocess.call(args))
```
jonmeow: Make sure that you aren't just churning your trace files by updating them with two different commands. These updates should be non-overlapping, much like `autoupdate_lit_testdata`.

That means either a different directory, or handling the update some other way.

prabhatexit0: Thanks for letting me know about this!
I did some changes, please let me know if this works.

jonmeow: It doesn't. When you run the script, note it says you can pass arguments: make sure that behavior works.

Probably this script shouldn't be modified to execute autoupdate_py twice, it should be once. Either use two separate scripts, or add flags to the base.

(as this is a draft PR, I may not comment further on this thread; I typically don't look at drafts but I'd glanced)

jonmeow: I should add -- also, testdata is a str and not a list. Please make sure you test your changes before asking for review.

prabhatexit0: Oh I'm sorry about that, thanks for reviewing! Will do these suggested changes.

explorer/file_test.cpp
```diff
@@ -30,6 +35,11 @@ class ParseAndExecuteTestFile : public FileTestBase {
                      "testdata/linked_list/typed_linked_list.carbon")) {
         GTEST_SKIP() << "Expensive test to trace";
       }
+    } else {
+      if (is_trace_test) {
+        GTEST_SKIP() << "`trace` tests only check for trace output.";
+        is_trace_test = true;
```
zygoloid: This assignment looks redundant.

prabhatexit0: I will remove it in the next commit. Thanks!

testing/scripts/autoupdate_testdata_base.py
```diff
@@ -216,26 +220,38 @@ def format(
...
+                    r"Time elapsed in \1: {{[0-9]+}}ms", result, count=1
                 )
-            else:
-                result = self.line_number_pattern.sub(
-                    r"\g<prefix>{{.*}}\g<suffix>",
-                    result,
-                    count=1,
+            elif trailing_match:
+                result = self.trailing_whitespace_pattern.sub(
+                    r"{{[ \t]}}", result
```
zygoloid:
```suggestion
                    r"{{\1}}", result
```
Let's test for the exact whitespace sequence we're expecting.

prabhatexit0: After making this change, I'm getting this error locally when running the autoupdate script.
```
raise s.error("invalid group reference %d" % index, pos)
re.error: invalid group reference 1 at position 3
```
...
```
raise ValueError(f"Failed to update {test}") from e
ValueError: Failed to update explorer/trace_testdata/full_trace.carbon
```

prabhatexit0: > After making this change, I'm getting this error locally when running the autoupdate script.
> 
> ```
> raise s.error("invalid group reference %d" % index, pos)
> re.error: invalid group reference 1 at position 3
> ```
> 
> ...
> 
> ```
> raise ValueError(f"Failed to update {test}") from e
> ValueError: Failed to update explorer/trace_testdata/full_trace.carbon
> ```

I'm sorry, I missed adding the capture group in the pattern.

zygoloid (approved):

testing/scripts/autoupdate_testdata_base.py
```diff
@@ -198,9 +198,13 @@ def __init__(
         super().__init__()
         self.filename = Path(test).name
         self.indent = ""
-        self.out_line = out_line.rstrip()
+        self.out_line = out_line
         self.line_number_delta_prefix = line_number_delta_prefix
         self.line_number_pattern = line_number_pattern
+        self.time_elapsed_pattern = re.compile(
+            r"Time elapsed in (\S+): (\d+)ms"
+        )
```
jonmeow, line 206: Minor note, but the autoupdate script supports a --extra_check_replacement flag that can be used for this. If you run into more cases that need it for the explorer, please use that -- I may take a look at migrating this, too, since we should probably be cautious about adding explorer-specific code where it affects the toolchain.

PR #2935, opened 2023-06-22
Explorer: Include `self` pattern in method type

### LINKED ISSUE #2857: Fuzzer issue: invalid runtime cast in `Interpreter::EvalPrim()`
### Description of the bug:

I get this assertion failure

> decltype(auto) llvm::cast(const From &) [To = Carbon::IntValue, From = Carbon::Value]: Assertion `isa<To>(Val) && "cast<Ty>() argument of incompatible type!"' failed.

triggered from `explorer/interpreter/interpreter.cpp:217:27` with a certain input carbon program.

```
package Foo api;

interface Interface {
  fn G() -> Self;
}

impl i32 as Interface {
  fn G[self: i32]() -> i32 {
    return (self + 1);
  }
}

fn Main() -> i32 {
  var n: i32 = 1;
  if (n.(Interface.G)() == 2) {
    Print("Constraint OK", );
  }
  return 3;
}
```

File can be found [here](https://github.com/josh11b/carbon-lang/blob/fuzz/test.carbon).
Error can be reproduced by running: `bazel run -c dbg //explorer -- test.carbon`

No crash

### What actually happened?

Get this assertion failure:

```
explorer: external/llvm-project/llvm/include/llvm/Support/Casting.h:567: decltype(auto) llvm::cast(const From &) [To = Carbon::IntValue, From = Carbon::Value]: Assertion `isa<To>(Val) && "cast<Ty>() argument of incompatible type!"' failed.
...
#10 0x0000563086a36888 decltype(auto) llvm::cast<Carbon::IntValue, Carbon::Value>(Carbon::Value const&) /proc/self/cwd/external/llvm-project/llvm/include/llvm/Support/Casting.h:568:43
#11 0x0000563086b0204b Carbon::Interpreter::EvalPrim(Carbon::Operator, Carbon::Value const*, std::__1::vector<Carbon::Value const*, std::__1::allocator<Carbon::Value const*> > const&, Carbon::SourceLocation) /proc/self/cwd/explorer/interpreter/interpreter.cpp:217:27
#12 0x0000563086b175f0 Carbon::Interpreter::StepExp() /proc/self/cwd/explorer/interpreter/interpreter.cpp:1526:9
```

(full output [here](https://github.com/josh11b/carbon-lang/blob/fuzz/no-msan.output.txt)).

josh11b:
This fixes a bug where Explorer would not detect when:

- an implementation used a method to implement a class function in an interface
- an implementation used a class function to implement a method in an interface
- an implementation method used `addr self` when the interface method did not
- an interface method used `addr self` when the implementation method did not

at type checking time. This would then cause a crash at runtime.

Closes #2857


explorer/ast/value.cpp
```diff
@@ -585,13 +585,20 @@ void Value::Print(llvm::raw_ostream& out) const {
...
              fn_type.deduced_bindings()) {
           out << sep << *deduced;
         }
+        if (fn_type.is_unbound_method()) {
+          if (fn_type.addr_self()) {
+            out << sep << "addr self: " << fn_type.self_type() << "*";
+          } else {
+            out << sep << "self: " << fn_type.self_type();
+          }
+        }
```
zygoloid:
```suggestion
        if (auto self = fn_type.method_self()) {
          if (self->addr_self) {
            out << sep << "addr self: " << *self->self_type << "*";
          } else {
            out << sep << "self: " << *self->self_type;
          }
        }
```

explorer/ast/value.h
```diff
@@ -675,7 +697,15 @@ class FunctionType : public Value {
     return impl_bindings_;
   }
 
+  auto method_self() const -> std::optional<MethodSelf> { return method_self_; }
+  // Returns true if function has an unbound `self` receiver parameter.
+  auto is_unbound_method() const -> bool { return method_self_.has_value(); }
+  // Only valid if is_unbound_method() returns true.
+  auto addr_self() const -> bool { return method_self_->addr_self; }
+  auto self_type() const -> const Value& { return *method_self_->self_type; }
```
zygoloid: The preconditions here make worry that people will get this wrong. I wonder if we can get away without adding these. I've made some suggested edits in `value.cpp` to see what this would look like. I don't feel strongly about this; feel free to reject if you'd prefer this as-is.

explorer/ast/value.cpp
```diff
@@ -844,6 +851,17 @@ auto TypeEqual(Nonnull<const Value*> t1, Nonnull<const Value*> t2,
...
+      // Verify `self` parameters match
+      if (fn1.is_unbound_method() != fn2.is_unbound_method()) {
+        return false;
+      }
+      if (fn1.is_unbound_method()) {
+        if (fn1.addr_self() != fn2.addr_self() ||
+            !TypeEqual(&fn1.self_type(), &fn2.self_type(), equality_ctx)) {
+          return false;
+        }
+      }
```
zygoloid:
```suggestion
      // Verify `self` parameters match
      auto self1 = fn1.method_self();
      auto self2 = fn2.method_self();
      if (self1.has_value() != self2.has_value()) {
        return false;
      }
      if (self1) {
        if (self1->addr_self != self2->addr_self ||
            !TypeEqual(self1->self_type, self2->self_type, equality_ctx)) {
          return false;
        }
      }
```

zygoloid (approved):
Looks good. One optional suggestion.

explorer/ast/value.h
josh11b: Done.

zygoloid (approved):
Thanks, LG! (Not merging due to conflicts.)

PR #2936, opened 2023-06-22
Explicit 'one or more' for digits in non-decimal integers.

jonmeow:


chandlerc (commented):
I'm pretty confident this matches the intent of the design, but @zygoloid should maybe confirm and approve.

josh11b (approved):

zygoloid (approved):
Yes, that's what I meant, thanks :)

(This is also consistent with the phrasing we use for real-number literals below.)

jonmeow:
> (This is also consistent with the phrasing we use for real-number literals below.)

Yep, deliberately copied. :)

PR #2939, opened 2023-06-22
Unify DeclaredName and DesignatedName as just Identifier

jonmeow:
This is just a simplification: I think these different forms are getting in the way more than they're helping, particularly as I was looking into namespace functionality. The handling in semantics can be identical, providing a more uniform behavior.


zygoloid (approved):
As I recall, this separation was hoped to be helpful to name lookup, but that seems to not be the case, so I'm happy for it to go away.

I wonder if we can find a better name for the resulting merged parse node -- we now have `NameReference` (which is an identifier expression that's always immediately looked up) and `Identifier` (which is any other identifier -- either a declared name or part of a member access or a designator in a struct type or struct literal), which I think could be a bit confusing, given that a `NameReference` is also an identifier.

Maybe `Name` rather than `Identifier`, so that we can collectively refer to `Name` and `NameReference` as identifiers?

I don't see a particular need to get the name right when we first merge this, so I'm happy for this to land as-is and think about renaming separately, but not merging in case you have a good naming idea you want to switch to immediately.

jonmeow:
I'm kind of neutral, can go for `Name` for now for the consistency you mention. Will do that then merge.

jonmeow:
I should note -- `Identifier` was for consistency with lexer.

jonmeow:
Oh, and SelfTypeIdentifier... I take it back, I'm going to stick with Identifier. We should figure out a consistent naming pattern, but I think `Identifier` has greater consistency, and it might actually be `NameReference` that should be changed.

PR #2940, opened 2023-06-22
Add namespace support.

jonmeow:
This handles namespacing of functions. Parsing and semantics are changed significantly, while lowering works without changes. Variables can't be namespaced yet because they're dealing with patterns, and I didn't dig through that code.

Most of the logic is done through the new name declaration stack, which is necessary because semantics isn't quite sure where the declaration name ends. It'd be complex for parsing to send a signal about this, probably involving node variants and rewrites of the tree, and this solution seems to work well. Unfortunately this means a new stack, but that may be inevitable due to the extra information needing to be tracked.

Note this doesn't deal with scoped lookups of non-namespace things, which we'll need for generics. That'll probably involve pushing resolved scopes onto a stack (or maybe just setting a singleton value?) to affect contextual name lookup. But, I think the basics are there to make it work when we can test the behavior.

This renames "designator expression" to "qualified expression" and adds "qualified declaration" in order to use terminology more consistent with C++.

Namespaces will probably need to be considered for name mangling down the line, but this still uses the basic name.


toolchain/parser/parse_node_kind.def
```diff
@@ -230,11 +237,20 @@ CARBON_PARSE_NODE_KIND_CHILD_COUNT(CallExpressionStart, 1)
...
 //   _external_: Identifier
-// DesignatorExpression
-CARBON_PARSE_NODE_KIND_CHILD_COUNT(DesignatorExpression, 2)
+// QualifiedDeclaration
+//
+// TODO: This will eventually more general expressions, for example with
+// `GenericType(type_args).ChildType(child_type_args).Name`.
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(QualifiedDeclaration, 2)
+
+// A qualified expression, such as `a.b` or `GetObject().(Interface.member)`:
```
zygoloid: In the design, we call these [member access expression](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/expressions/member_access.md)s. It's probably best to use the same terminology here.

toolchain/parser/parser_handle_declaration_name_and_params.cpp
```diff
@@ -7,18 +7,70 @@
...
-                                                 bool params_required) -> void {
+static auto ParserHandleDeclarationNameAndParams(
+    ParserContext& context, std::optional<ParserState> after_name) -> void {
   auto state = context.PopState();
 
-  if (!context.ConsumeAndAddLeafNodeIf(TokenKind::Identifier,
-                                       ParseNodeKind::Identifier)) {
+  // TODO: Should handle designated names.
+  if (auto identifier = context.ConsumeIf(TokenKind::Identifier)) {
+    state.state = *after_name;
```
zygoloid: You're unconditionally dereferencing an `optional` here. Should the parameter be `optional`? It looks like all the callers provide it.

zygoloid (commented):
I've only reviewed the parser half of this so far.

toolchain/semantics/semantics_context.cpp
```diff
@@ -76,43 +78,93 @@ auto SemanticsContext::AddNodeAndPush(ParseTree::Node parse_node,
...
+    }
+    CARBON_CHECK(!it->second.empty())
+        << "Should have been erased: " << semantics_ir_->GetString(name_id);
 
-  // TODO: Check for ambiguous lookups.
-  return it->second.back();
+    // TODO: Check for ambiguous lookups.
+    return it->second.back();
+  } else {
+    auto scope = semantics_ir_->GetNameScope(scope_id);
```
zygoloid: Should this be `const auto&`? If not, please use a type here instead of `auto` -- when reading this code in isolation, I don't know what this `auto` represents, and I'm concerned that we might be making a copy of a name lookup table.

toolchain/semantics/semantics_context.cpp
```diff
@@ -76,43 +78,93 @@ auto SemanticsContext::AddNodeAndPush(ParseTree::Node parse_node,
...
+    auto it = scope.find(name_id);
+    if (it == scope.end()) {
+      if (print_diagnostics) {
+        emitter_->Emit(parse_node, NameNotFound,
+                       semantics_ir_->GetString(name_id));
+      }
+      return SemanticsNodeId::BuiltinInvalidType;
```
zygoloid, line 160: I was initially surprised by this -- returning a value that is specifically a type for any kind of failed lookup seems unexpected. But, if I remember correctly, the type of the invalid type is also the invalid type, so maybe it does make sense to use this builtin value as a more general invalid value, and not only as an invalid type value. If we're using this more generally, do you think it'd make sense to rename it to `BuiltinInvalidValue`? (That also seems more accurate: it's not formally a type given that its type is not `type`.)

None of this is new in this patch, so I'm not requesting any action here, just musing more generally.

toolchain/semantics/semantics_context.h
```diff
@@ -20,6 +20,81 @@ namespace Carbon {
 // Context and shared functionality for semantics handlers.
 class SemanticsContext {
  public:
+  // Context for declaration name construction, tracked in
+  // declaration_name_stack_.
+  //
+  // A qualified declaration name will consist of entries which are either
+  // Identifiers are full expressions. Expressions are expected to resolve to
```
zygoloid:
```suggestion
  // Identifiers or full expressions. Expressions are expected to resolve to
```

toolchain/semantics/semantics_context.h
```diff
@@ -20,6 +20,81 @@ namespace Carbon {
...
+  // types, such as how `fn Vector(i32).Clear() { ... }` uses the expression
+  // `Vector(i32)` to indicate the type whose member is being declared.
+  // Identifiers such as `Clear` will be resolved to a name if possible, for
+  // example when declaring things that are in a non-generic type or namespace,
+  // and are otherwise marked as an unresolved identifier.
+  //
+  // Unresolved identifiers are valid if and only if they are the last step of a
+  // qualified name. Resolved identifiers in the last step will occur for both
```
zygoloid: From discord discussion, we could strengthen this to: Unresolved identifiers are valid if and only if they are the last step of a qualified name and either the qualifier is empty or names a namespace.

I don't think we need to apply that rule at this level, but eventually we're going to need a rule that you can't add new members to a class / interface / other non-namespace construct with a qualified declaration, and this may be the easiest place to implement it.

toolchain/semantics/semantics_handle.cpp
```diff
@@ -43,13 +43,44 @@ auto SemanticsHandleDeducedParameterListStart(SemanticsContext& context,
...
+  // applied.
+  auto [parse_node2, node_or_name_id2] =
+      context.node_stack().PopWithParseNode<SemanticsNodeId>();
+  if (context.parse_tree().node_kind(context.node_stack().PeekParseNode()) ==
+      ParseNodeKind::QualifiedDeclaration) {
+    // QualifiedDeclarations have already been applied.
+    context.node_stack().PopAndDiscardSoloParseNode(
+        ParseNodeKind::QualifiedDeclaration);
+  } else {
+    auto [parse_node1, node_or_name_id1] =
```
zygoloid: Can we `CHECK` that the top of the declaration name stack is in the `New` state here? I think that'd make the behavior here a little easier for me to follow.

zygoloid (commented):
There's a fair amount of name lookup code in `SemanticsContext` now. Given that we don't want to end up with a very large class like Clang's `Sema`, I think it'd be a good idea to try to factor that stuff out into a separate class at some point -- though doing that as a follow-up change rather than in this PR would seem reasonable to me.

toolchain/parser/parse_node_kind.def
jonmeow: Done, MemberAccessExpression and handle_period as discussed.

toolchain/parser/parser_handle_declaration_name_and_params.cpp
jonmeow: Fixed, this is no longer optional.

toolchain/semantics/semantics_context.cpp
jonmeow: Fixed to be `const auto&`, copy was an error

toolchain/semantics/semantics_context.cpp
jonmeow: The InvalidType is being used as a type in semantics (it's specifically valid as a type_id). Given the directionality (types can be values, but values have a hard time becoming types), I prefer the type label over value.

We could also break out of the type/value mindset and go for something like "SemanticError" which may better capture the concept -- I've been mulling a shift in that direction mainly to make the distinction clearer, that this isn't _really_ a type.

toolchain/semantics/semantics_context.h
jonmeow: Done

toolchain/semantics/semantics_context.h
jonmeow: Going with: "qualified name; all resolved qualifiers must resolve to an entity with members, such as a namespace."

I agree other non-namespace constructs will likely be handled here, which is why I'm aiming for the "such as" phrasing.

toolchain/semantics/semantics_handle.cpp
jonmeow: This is a little awkward -- we'd need to either add a function or a parameter for a check, since the stack is an implementation detail of the context. Do you still want it?

I've added a comment to clarify behavior. Also, shifted the behavior a little to reduce pushes/pops.

jonmeow:
> There's a fair amount of name lookup code in `SemanticsContext` now. Given that we don't want to end up with a very large class like Clang's `Sema`, I think it'd be a good idea to try to factor that stuff out into a separate class at some point -- though doing that as a follow-up change rather than in this PR would seem reasonable to me.

Noted

toolchain/semantics/semantics_context.cpp
zygoloid: Perhaps a different viewpoint would be that this value is permitted basically anywhere as a placeholder for an error, which is why we allow it as a `type_id` in particular. We'd also presumably allow this value as a base class (despite it not being a class), or as an array bound (despite it not being an integer), so it seems consistent to allow it as a type (despite it not being a value of type `type`).

`SemanticError` seems fine to me. I guess that'd be broader than `InvalidValue`, in that it'd potentially be usable even in non-value contexts where some other kind of node is typically expected. That probably adds some complexity to consumers of those other kinds of node, though, if they otherwise only had one semantic node kind that they needed to handle.

toolchain/semantics/semantics_handle.cpp
```diff
@@ -372,6 +385,34 @@ auto SemanticsHandlePrefixOperator(SemanticsContext& context,
...
 
+auto SemanticsHandleQualifiedDeclaration(SemanticsContext& context,
+                                         ParseTree::Node parse_node) -> bool {
+  // The first two qualifiers in a chain will be a QualifiedDeclaration with two
+  // Identifier or expression children. Later qualifiers will have a
+  // QualifiedDeclaration as the first child, and an Identifier or expression as
+  // the second child.
+  //
+  // The child parse nodes are in reverse order versus how they need to be
+  // applied.
```
zygoloid: I don't think I'm following this comment: my understanding is that given
```
fn Foo.Bar.Baz()
```
... the parse tree nodes are in this order
```
1: Foo
2:     Bar
3:    .
4:         Baz
5:        .
```
which looks like the order in which we apply them:

New --(`Foo`)-> Resolved --(`Bar`)-> Resolved --(`Baz`)-> Unresolved

toolchain/semantics/semantics_context.cpp
jonmeow: To be sure, still as a builtin. The complexity of needing to handle errors is, I think, there regardless of the shape it takes. However, not using a node kind reduces a level of investigation needed to detect an error.

toolchain/semantics/semantics_handle.cpp
jonmeow: Rephrasing: "The nodes are applied in reverse order due to the postorder tree."

jonmeow: Actually, just removing, because this is probably just confusing because it's always that way.

toolchain/semantics/semantics_handle.cpp
zygoloid: Let's leave it as-is; I think the added comments help sufficiently.

zygoloid (approved):

PR #2941, opened 2023-06-23
Track the list of blocks that form the body of a function.

zygoloid:
Use that list for lowering in lexical order, instead of rediscovering the list based on which blocks are referenced as branch targets.


toolchain/lowering/lowering_context.cpp
```diff
@@ -103,15 +103,15 @@ auto LoweringContext::BuildFunctionDefinition(SemanticsFunctionId function_id)
...
-    function_lowering.builder().SetInsertPoint(
-        function_lowering.GetBlock(block));
-    for (const auto& node_id : semantics_ir().GetNodeBlock(block)) {
+  // Lower all blocks.
+  // TODO: Determine the set of reachable blocks, and only lower those ones.
+  for (auto block_id : body_block_ids) {
+    CARBON_VLOG() << "Lowering " << block_id << "\n";
+    auto* llvm_block = function_lowering.GetBlock(block_id);
+    // Keep the LLVM blocks in lexical order.
+    llvm_block->moveBefore(llvm_function->end());
```
jonmeow, line 112: Could you avoid this reordering if you did two passes:

1) Create all body blocks.
2) Lower all body blocks.

Then remove the ability for GetBlock to create blocks.

But maybe that'd leave synthetic blocks out of order?

jonmeow (approved):
Looks good! One comment, but I suspect the idea wouldn't work out so I'm merging.

toolchain/lowering/lowering_context.cpp
zygoloid: Yeah, we'd need some care to put synthetic blocks in the right place, but that'd be possible (we can find the current basic block from the insertion point in the IR builder, and insert the synthetic block after that one). The reason I didn't do that was that this would get in the way of our optimization to avoid creating empty blocks -- we'd either need to move that optimization into step 1, or delete some of the blocks we created in step 1 if we find they end up being empty -- but I think that is a surmountable problem.

PR #2942, opened 2023-06-23
Add implied `return;` at end of non-value-returning functions. 

zygoloid:
Add validation that every code block in a function is terminated by a sequence of terminating instructions, and that terminators don't appear anywhere else in code blocks.

This required tracking whether we're in a reachable code block. That's done on the fly when we create a new code block; the new `SemanticsNodeBlockId::Unreachable` is used to represent the case where we're not actually creating a code block because we're in unreachable code.


toolchain/semantics/semantics_node_kind.h
```diff
@@ -22,6 +22,11 @@ class SemanticsNodeKind : public CARBON_ENUM_BASE(SemanticsNodeKind) {
 #include "toolchain/semantics/semantics_node_kind.def"
 
   using EnumBase::Create;
+
+  // Determine whether this node kind is a code block terminator, such as a
```
jonmeow: "Returns true if"

This function is called to determine, but doesn't feel like it determines itself. (also, it'd be "Determines" per style)

toolchain/semantics/semantics_ir.cpp
```diff
@@ -92,9 +92,48 @@ auto SemanticsIR::MakeFromParseTree(const SemanticsIR& builtin_ir,
   context.VerifyOnFinish();
 
   semantics_ir.has_errors_ = err_tracker.seen_error();
+
+#ifndef NDEBUG
+  if (auto verify = semantics_ir.Verify(); !verify.ok()) {
+    CARBON_FATAL() << "Built invalid semantics IR: " << verify.error() << "\n"
+                   << semantics_ir;
```
jonmeow: Maybe swap this so that the error is last... the semantics_ir is pretty long, so it'll take time to locate the actual error otherwise (also bear in mind, the stack trace comes first).

jonmeow (approved):
Looks good to me.

This makes me wonder if we should stop emitting a block after the first terminator (e.g., `return; return;`). I actually think we might be able to do that in the SemanticsIR, by stopping additions to the block (but still validating by creating the nodes). Maybe something to think about, but definitely not part of this PR.

jonmeow:
Oh, I wonder how we should handle:

```
fn F() -> i32 {
  return 3;
  if (true) {
  }
}
```

Which would fail verification here. (maybe my above comment about stopping emitting IR is more important than I was thinking?)

toolchain/semantics/semantics_node_kind.h
zygoloid: Done.

toolchain/semantics/semantics_ir.cpp
zygoloid: Done.

zygoloid:
> This makes me wonder if we should stop emitting a block after the first terminator (e.g., `return; return;`). I actually think we might be able to do that in the SemanticsIR, by stopping additions to the block (but still validating by creating the nodes).

Yeah, I think we need to at least stop emitting the current block and start emitting a new block when we finish a `return` statement. But that then breaks the naïve tracking of whether we're in reachable code that this PR was doing. I've followed the approach you suggested: when we're visiting unreachable code, we still create the nodes but don't add them to the block. I'm tracking whether we're currently within an unreachable block by setting the top block on the node block stack to a new block ID, `SemanticsNodeBlockId::Unreachable`, that doesn't correspond to a real block, in that case.

In order to properly track whether we're in a reachable block as we go, I've also refactored all the code block creation code to go through a few new functions on `SemanticsContext` that determines whether the new code block is reachable or not.

toolchain/semantics/semantics_node_block_stack.h
```diff
@@ -23,9 +23,18 @@ class SemanticsNodeBlockStack {
       llvm::raw_ostream* vlog_stream)
       : name_(name), node_blocks_(&node_blocks), vlog_stream_(vlog_stream) {}
 
+  // Adds a node block, but does not push it.
+  auto Add() -> SemanticsNodeBlockId;
```
jonmeow: My leaning is this belongs on SemanticsIR which owns node_blocks_, rather than here, since it's not related to the stack. This could switch to taking a pointer to semantics_ir for calling Add.

SemanticsIR already has Add functions for most of its data, so this would be consistent there. To the extent that the stack was abstracting away the Add, that's no longer possible with this addition.

toolchain/semantics/semantics_node_block_stack.h
```diff
@@ -23,9 +23,18 @@ class SemanticsNodeBlockStack {
...
+
+  // Pushes an existing node block.
+  auto Push(SemanticsNodeBlockId) -> void;
```
jonmeow:
```suggestion
  auto Push(SemanticsNodeBlockId id) -> void;
```

toolchain/semantics/semantics_node_block_stack.h
```diff
@@ -51,6 +60,11 @@ class SemanticsNodeBlockStack {
     return Pop();
   }
 
+  // Returns whether the current block is statically reachable.
+  auto CurrentBlockIsReachable() -> bool {
```
jonmeow:
```suggestion
  auto is_current_block_reachable() -> bool {
```

Soft suggestion

toolchain/semantics/semantics_context.h
```diff
@@ -57,9 +57,48 @@ class SemanticsContext {
...
+  auto AddConvergenceBlockWithArgAndPush(
+      ParseTree::Node parse_node,
+      std::initializer_list<std::pair<SemanticsNodeBlockId, SemanticsNodeId>>
+          blocks_and_args) -> SemanticsNodeId;
+
   // Add the current code block to the enclosing function.
   auto AddCurrentCodeBlockToFunction() -> void;
 
+  // Returns whether the current position in the current block is reachable.
+  auto CurrentPositionIsReachable() -> bool;
```
jonmeow:
```suggestion
  auto IsCurrentPositionReachable() -> bool;
```

Mild, but suggesting starting the name with "Is" so that it's clearer it's a query instead of a statement.

toolchain/semantics/semantics_node_block_stack.h
```diff
@@ -51,6 +60,11 @@ class SemanticsNodeBlockStack {
...
+    return stack_.empty() || stack_.back() != SemanticsNodeBlockId::Unreachable;
```
jonmeow:
```suggestion
    CARBON_CHECK(!stack_.empty());
    return stack_.back() != SemanticsNodeBlockId::Unreachable;
```

I believe this query should only ever be run when non-empty. i.e., this could be a CHECK if desired, although `Peek()` doesn't check first.

toolchain/semantics/semantics_context.cpp
```diff
@@ -143,6 +224,20 @@ auto SemanticsContext::AddCurrentCodeBlockToFunction() -> void {
...
 }
 
+auto SemanticsContext::CurrentPositionIsReachable() -> bool {
+  if (!node_block_stack().CurrentBlockIsReachable()) {
+    return false;
+  }
+
+  // Our current position is at the end of the current block. That position is
+  // reachable unless the previous instruction is a terminator instruction.
+  auto& block_contents = semantics_ir().GetNodeBlock(node_block_stack().Peek());
```
jonmeow: Peek is not guaranteed to provide a valid return. This probably works today because it won't be called from a lazily created block. I'd suggest one of:

1) Document in the header that this can only be called when a valid block is on the top. Maybe explicitly CHECK, but GetNodeBlock will fail if it gets an invalid block id.
2) Handle invalid. Probably `auto block_id = node_block_stack().Peek(); if (!block_id.is_valid()) { return true; }`

My slight leaning is option 1 because it's less logic, and the guard rail can be switched if behavior changes.

jonmeow (approved):

toolchain/semantics/semantics_context.cpp
zygoloid: Nice catch. This is working today because code blocks are never created as invalid -- we eagerly create the entry block so we can add it to the `SemanticsFunction`, and we eagerly create all other code blocks so we can create a branch to them. But I think it'd probably be reasonable to call this function after a new code block is pushed and before any branches to it are added -- and at least it'd be surprising if it fails in that case -- so I've gone with option 2.

toolchain/semantics/semantics_context.h
zygoloid: Done, and lower-cased for consistency with `is_current_block_reachable`.

toolchain/semantics/semantics_node_block_stack.h
zygoloid: Done.

toolchain/semantics/semantics_node_block_stack.h
zygoloid: Done. I moved the check into `Peek` and switched this to call `Peek`.

zygoloid (commented):
I've fixed the handling of `BranchIf` to be more precise: we now validate that `BranchIf` only appears in the run up to the block terminator, and can't be used as the final instruction in a block, and that a block has exactly one proper terminator instruction. We no longer treat an instruction after a `BranchIf` as unreachable, though that doesn't make a difference in practice because we never call `is_current_position_reachable` in that situation currently.

toolchain/semantics/semantics_ir.h
```diff
@@ -82,6 +82,13 @@ class SemanticsIR {
   }
   auto Print(llvm::raw_ostream& out, bool include_builtins) const -> void;
 
+  // Adds an empty node block, returning an ID to reference it.
+  auto AddBlock() -> SemanticsNodeBlockId {
```
jonmeow: Naming AddNodeBlock would mirror GetNodeBlock; also, note how I've been placing all the Add/Get functions next to each other...

toolchain/semantics/semantics_node_kind.h
```diff
@@ -16,17 +16,32 @@ CARBON_DEFINE_RAW_ENUM_CLASS(SemanticsNodeKind, uint8_t) {
...
 
+// Whether a node is a terminator or part of the terminator sequence. The nodes
+// in a block appear in the order NotTerminator, then TerminatorSequence, then
+// Terminator, which is also the numerical order of these values.
+enum class SemanticsTerminatorKind {
+  // This node is not a terminator.
+  NotTerminator,
+  // This node is not itself a terminator, but forms part of a terminator
+  // sequence.
+  TerminatorSequence,
```
jonmeow, line 27: Looking at the code, couldn't this just be NotTerminator? I see code that compares with Terminator, but nothing comparing with NotTerminator or TerminatorSequence.

Partly I'm commenting because of the macros: terminators are nodes, so NODE -> TERMINATOR felt a little odd to me. But for one macro, I kind of shrugged. Now it's two, so I feel I should comment; maybe you could do something like `CARBON_SEMANTICS_NODE_KIND_WITH_TERMINATOR_KIND(BranchIf, TerminatorSequence)` and have one macro? It's borderline, but if more details keep getting added, I think it shouldn't be a new macro each time.

jonmeow (commented):
Continuing to comment, but if you want to merge and do an incremental change from there, please go ahead. I think focusing deltas may help regardless.

toolchain/semantics/semantics_ir.h
zygoloid: Done.

toolchain/semantics/semantics_node_kind.h
zygoloid: This is used in `SemanticsIR::Verify`, in `prior_kind > node_kind`, which is checking for the case of `Terminator` followed by `NotTerminator` or `TerminatorSequence`, and for the case of `TerminatorSequence` followed by `NotTerminator`.

I'll merge this and send a separate PR for the suggested macro change.

PR #2944, opened 2023-06-23
Codegen: Given a carbon file prints the assembly to the stdout

fasiddique:


toolchain/codegen/codegen.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include "llvm/Target/TargetOptions.h"
+#include "llvm/TargetParser/Host.h"
+
+namespace Carbon {
+
+auto PrintAssemblyFromModule(llvm::Module& module,
+                             llvm::StringRef target_triple,
+                             llvm::raw_pwrite_stream& error_stream,
+                             llvm::raw_pwrite_stream& output_stream) -> bool {
+  bool has_errors = false;
```
jonmeow: Why are you creating a `has_errors` variable which is never assigned, and returning `has_errors`/`!has_errors`? Why not instead return true/false?

toolchain/driver/testdata/error_codegen_target_triple.carbon
```diff
@@ -0,0 +1,8 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// RUN:  %{not} %{carbon} dump assembly --target_triple=x86_687-unknown-linux-gnu %s | %{FileCheck-strict}
```
jonmeow:
```suggestion
// RUN: %{not} %{carbon} dump assembly --target_triple=x86_687-unknown-linux-gnu %s | %{FileCheck-strict}
```

toolchain/driver/testdata/error_codegen_target_triple.carbon
```diff
@@ -0,0 +1,8 @@
...
+// CHECK:STDERR: {{ERROR:.*}}
```
jonmeow:
```suggestion
// CHECK:STDERR: ERROR: {{.*}}
```

Let's limit the regex to only be used where needed. Also, there should be a space after "ERROR:", so you can validate that.

toolchain/driver/driver.cpp
```diff
@@ -236,6 +246,17 @@ auto Driver::RunDumpSubcommand(DiagnosticConsumer& consumer,
...
   }
 
+  if (dump_mode == DumpMode::Assembly) {
+    consumer.Flush();
+    has_errors |= Carbon::PrintAssemblyFromModule(
+        *module, target_triple, error_stream_, output_stream_);
+    if (has_errors) {
+      CARBON_VLOG() << "Unable to dump assembly due to errors.";
+      return false;
+    }
```
```suggestion
```

The error should be printed, so I don't think there's a need to note an error.

toolchain/codegen/codegen.cpp
```diff
@@ -0,0 +1,72 @@
...
+  llvm::TargetOptions target_opts;
+  std::optional<llvm::Reloc::Model> reloc_model;
+  auto* target_machine = target->createTargetMachine(
+      target_triple, CPU, Features, target_opts, reloc_model);
+  module.setDataLayout(target_machine->createDataLayout());
+  module.setTargetTriple(target_triple);
+
+  // Using the legacy PM to generate the assembly since the new PM
+  // does not work with this yet.
+  // FIXME: make the new PM work with the codegen pipeline.
```
jonmeow:
```suggestion
  // TODO: make the new PM work with the codegen pipeline.
```

Prefer TODO instead of other possible markers.
https://google.github.io/styleguide/cppguide.html#TODO_Comments

toolchain/codegen/codegen.h
```diff
@@ -0,0 +1,20 @@
...
+
+#ifndef CARBON_TOOLCHAIN_CODEGEN_CODEGEN_H_
+#define CARBON_TOOLCHAIN_CODEGEN_CODEGEN_H_
+
+#include <cstdint>
+
+#include "llvm/IR/Module.h"
+
+namespace Carbon {
+// Prints the assembly to stdout for the given llvm module.
```
jonmeow, line 13: Idiomatically, I expect functions to return `true` on success and `false` on failure. Right now this does the reverse. Please make the behavior match the idiom, and document the return behavior for clarity.

jonmeow: Although you changed the behavior, please also document the return behavior here.

toolchain/driver/testdata/error_codegen_target_triple.carbon
jonmeow: I see you removed the space after ERROR: (versus my suggestion). Why shouldn't a space be there?

jonmeow (commented):
Thanks! I think we're almost done here, but I see a couple of my comments were only partially addressed, can you check them again?

jonmeow (approved):
Thanks! With these changes, this looks ready to merge.

PR #2945, opened 2023-06-24
Explorer: Prevent copies when initializing value expression from reference expression

Pixep:
Prevent copies when initializing value expression from reference expression. This is based on https://github.com/carbon-language/carbon-lang/pull/2006, which introduces expression categories, and how it is possible to convert to/from those different categories. Continuation of https://github.com/carbon-language/carbon-lang/pull/2907

## Functional changes

* Initializing a value expression from a reference expression takes its value without a copy
    * Reading from the value expression causes an error if the value changed from the time it was initialized
    * In this situation, prevents a copy both for variable definitions, and call parameter bindings

* Add new `ExpressionCategoryAction` which evaluates an expression and returns an `ExpressionValue` containing its category and address (if any), in addition to the resulting `Value*`
    * `ExpressionAction`s now invokes `ExpressionCategoryAction` and unwrap the returned `ExpressionValue`
* `RuntimeScope::BindAndPin` method, and corresponding when attempting to read a `value_node`.

## Next work

* Avoid unnecessary copies from value expression to value expression, after ensuring that even value expression temporaries are registered for destruction.


Pixep:
Wrong branch, oopsie.

PR #2946, opened 2023-06-25
Explorer: Prevent copies when initializing a let binding from reference expression

Pixep:
Prevent copies when initializing value expression from reference expression. This is based on https://github.com/carbon-language/carbon-lang/pull/2006, which introduces expression categories, and how it is possible to convert to/from those different categories. Continuation of https://github.com/carbon-language/carbon-lang/pull/2907

## Functional changes

* Initializing a value expression from a reference expression takes its value without a copy
    * Reading from the value expression causes an error if the value changed from the time it was initialized
    * In this situation, prevents a copy both for variable definitions, and call parameter bindings

* Add new `ExpressionCategoryAction`, which evaluates an expression and returns an `ExpressionValue` containing its category and address (if any), in addition to the resulting `Value*`
    * `ExpressionAction`s now invokes `ExpressionCategoryAction` and unwraps the returned `ExpressionValue`
* `RuntimeScope::BindAndPin` method, and corresponding when attempting to read a `value_node`.

## Next work

* Avoid unnecessary copies from value expression to value expression, after ensuring that even value expression temporaries are registered for destruction (https://github.com/Pixep/carbon-lang/pull/9)


explorer/interpreter/action.cpp
```diff
@@ -88,19 +102,36 @@ void RuntimeScope::Merge(RuntimeScope other) {
...
+    return {std::nullopt};
   }
+  auto pinned_it = pinned_values_.find(value_node);
+  if (pinned_it != pinned_values_.end()) {
+    // Check if pinned value was modified.
+    CARBON_CHECK(it->second->kind() == Value::Kind::LocationValue);
+    CARBON_ASSIGN_OR_RETURN(
+        Nonnull<const Value*> current,
+        heap_->Read(cast<LocationValue>(it->second)->address(), source_loc));
+    if (current != pinned_it->second) {
```
zygoloid: It looks like we're performing a `Value*` comparison here. Have we decided which of these we want:

1) If the value is mutated in any way between the binding and the read, that's bad.
2) If the value is actually different when it's read, that's bad, but it's OK to temporarily change it to something else and back again so long as no-one looks at it.

If we want (1), then I think this doesn't quite go far enough, because we could assign a different `Value*` and then assign back the original `Value*`. If we want (2), then I think we need to perform a `ValueEqual` comparison here rather than a pointer comparison, because it's possible for two different `Value*`s to point to equal values.

explorer/interpreter/heap.cpp
```diff
@@ -73,16 +76,31 @@ auto Heap::CheckInit(AllocationId allocation, SourceLocation source_loc) const
   return Success();
 }
 
-void Heap::Deallocate(AllocationId allocation) {
+auto Heap::CheckWritable(AllocationId allocation,
```
zygoloid: Is this function left behind from a different approach? I can't find anywhere that an allocation is put into the `AlivePinned` state. (I think the implementation in `RuntimeScope` doesn't need this machinery, and seems to properly handle the case where only a subobject of an allocation is pinned, which this approach presumably wouldn't.)

explorer/interpreter/interpreter.cpp
```diff
@@ -293,38 +297,43 @@ auto Interpreter::CreateStruct(const std::vector<FieldInitializer>& fields,
...
       }
       break;
     case ExpressionCategory::Value:
       if (v.expression_category() == ExpressionCategory::Value) {
         // TODO: Ensure value expressions of temporaries are registered as
         // allocation to allow us to reference it without the need for a copy.
-        (*bindings)->Initialize(value_node, v.value());
+        bindings->Initialize(value_node, v.value());
       } else if (v.expression_category() == ExpressionCategory::Reference) {
         // TODO: Prevent mutation, error on mutation, or copy
```
zygoloid: Can we remove this TODO now?

zygoloid (commented):
It looks like we end up creating a lot of `ExpressionValue`s for value expressions that are just wrapping a `Value*`, and then immediately unwrapping them when we get to the end of the `ExpressionAction` that encloses the `ExpressionCategoryAction`. I wonder if we can avoid most of that wrapping and unwrapping:

- Only create `ExpressionValue` for the `Reference` and `Initializing` cases, never for the `Value` case.
- In `ExpressionCategoryAction`, don't create the `ExpressionValue` wrappers for value expressions, and in the places where you're unwrapping an `ExpressionValue`, treat a `Value` that's not an `ExpressionValue` in the same way you currently treat an `ExpressionValue` for a value expression.
- Possibly consider splitting `ExpressionValue` up into two different `Value` subclasses for the two cases. Something like `ReferenceValue` and `InitializingExpresionValue` might make sense.
- Possibly consider renaming `ExpressionAction` to `ValueExpressionAction`, and renaming `ExpressionCategoryAction` to `ExpressionAction`, to reflect that `ValueExpressionAction` is computing a value, even when given an expression of a different category.

explorer/interpreter/heap.cpp
Pixep: Thanks for catching that... yes it was indeed the previous implementation, I suppose I just got used to seeing it. Thanks!

explorer/interpreter/action.cpp
Pixep: Relayed to #pointers on Discord. That's a good point... Going with 2. would be trivial to implement, 1. not so much, as we would need a revision counter, or notify observers of a mutation. But that would enable CoW scenarios.
Going to add `ValueEqual` for now at least.

explorer/interpreter/interpreter.cpp
Pixep: We can indeed! Removed.

explorer/interpreter/action.cpp
Pixep: Going with (1) as discussed https://discord.com/channels/655572317891461132/753021843459538996/1125137499937849395.
This was trivially implemented using a revision counter in `Heap`.

zygoloid: It looks like the revision counter is per-allocation, which I think means that we'd treat a case like this as invalid:

```
fn F() -> i32 {
  var v: {.a: i32, .b: i32} = {.a = 1, .b = 2};
  let a: i32 = v.a;
  v.b = 3;
  // Should be OK: `v.a` has not been modified,
  // even though another part of the same allocation has.
  return a;
}
```

Pixep:
> It looks like we end up creating a lot of `ExpressionValue`s for value expressions that are just wrapping a `Value*`, and then immediately unwrapping them when we get to the end of the `ExpressionAction` that encloses the `ExpressionCategoryAction`. I wonder if we can avoid most of that wrapping and unwrapping:
> 
> * Only create `ExpressionValue` for the `Reference` and `Initializing` cases, never for the `Value` case.
> * In `ExpressionCategoryAction`, don't create the `ExpressionValue` wrappers for value expressions, and in the places where you're unwrapping an `ExpressionValue`, treat a `Value` that's not an `ExpressionValue` in the same way you currently treat an `ExpressionValue` for a value expression.
> * Possibly consider splitting `ExpressionValue` up into two different `Value` subclasses for the two cases. Something like `ReferenceValue` and `InitializingExpresionValue` might make sense.
> * Possibly consider renaming `ExpressionAction` to `ValueExpressionAction`, and renaming `ExpressionCategoryAction` to `ExpressionAction`, to reflect that `ValueExpressionAction` is computing a value, even when given an expression of a different category.

Addressed all of those listed above. I was a bit concerned with having to branch too much, but this is manageable, avoids all that wrapping, and makes the intent clearer.

Small difference compared to what you mentioned: we don't have an `InitializingExpressionValue` for now, but this leaves room for that if we want to change the way we propagate locations into, or out of a call.

The "revision counter" point raised is still outstanding on my side.

explorer/interpreter/action.cpp
Pixep: Added that use case, and reworked it. I tried to keep the logic as simple as possible, not super happy with the fact we now have some `ValueNodeView`(`AstNode*` actually) in the Heap, but maybe that's good enough for this round.
Most of the logic is visible in `heap.cpp`.

Pixep:
https://github.com/carbon-language/carbon-lang/pull/2946/commits/c8c4547ce24a8ab6f56d3c8383aa9c701d8dbc2d updates the trace test file, I haven't dug yet to see if that diff is expected or not.

zygoloid (approved):

zygoloid:
Trace differences are due to running additional actions, each of which traces the stack and heap state. The extra output will be cleaned up by #2982, so I think it's fine for now.

PR #2952, opened 2023-06-27
LLVM should be included using quotes.

jonmeow:
This also results in correct dependency discovery (the missing LLVM dep on expression_category).


chandlerc (approved):
LG

PR #2953, opened 2023-06-27
update keywords in textmate syntax

maan2003:
keywords were taken from docs/design/lexical_conventions/words.md

### DIFF
--- a/utils/textmate/Syntaxes/Carbon.plist
+++ b/utils/textmate/Syntaxes/Carbon.plist
@@ -61,7 +61,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
                 <key>name</key>
                 <string>keyword.other.carbon</string>
                 <key>match</key>
-                <string>\b(abstract|adapt|addr|alias|and|api|as|auto|base|class|constraint|extend|final|fn|forall|friend|impl|impls|import|in|interface|let|library|like|namespace|not|observe|or|override|package|partial|private|protected|require|Self|type|var|virtual|where|_)\b</string>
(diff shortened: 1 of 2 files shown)


utils/textmate/Syntaxes/Carbon.plist
```diff
@@ -54,14 +54,14 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
...
                 <key>match</key>
-                <string>\b(break|case|continue|default|else|if|for|match|return|then|while)\b</string>
+                <string>\b(break|case|continue|default|else|if|for|match|return|returned|then|while)\b</string>
             </dict>
             <!-- Rule: other keywords -->
             <dict>
                 <key>name</key>
                 <string>keyword.other.carbon</string>
                 <key>match</key>
-                <string>\b(abstract|addr|alias|and|api|as|auto|base|choice|class|constraint|destructor|extends|external|final|forall|fn|friend|impl|import|interface|let|library|like|namespace|not|observe|or|override|package|partial|private|protected|returned|template|var|virtual|where|while|_)\b</string>
```
zygoloid, line 64: Most of these removed keywords should be added to `words.md` instead of removed here, because they were added to the language by an approved proposal:

- `choice` was added in p0157
- `destructor` was added in p1154
- `returned` was added in p0257
- `template` was added in p0676
-  `while` was added in p0340

I think removing `extends` and `external` here is appropriate.

zygoloid (changes requested):

utils/textmate/Syntaxes/Carbon.plist
maan2003: I have added `choice`, `destructor` , `template` to words.md. 
`returned` and `while` are in control flow keywords.

zygoloid (approved):

PR #2955, opened 2023-06-27
Creates object file from the module.

fasiddique:
This pr creates the object file for the carbon code that returns either 0 or 1. The command to convert the object code to binary is `clang <object_file_name> -o <binary_name>`


zygoloid (changes requested):
It looks like something weird has happened to your git branch (around commit 929633d4c222d186bc479d168689b5d211c774a5), and this PR has ended up with changes that undo several recent trunk commits. We'll need to fix that before we can review and merge this PR.

toolchain/codegen/codegen.cpp
```diff
@@ -10,16 +10,11 @@
...
 #include "llvm/Target/TargetOptions.h"
 #include "llvm/TargetParser/Host.h"
 
 namespace Carbon {
-
-auto PrintAssemblyFromModule(llvm::Module& module,
-                             llvm::StringRef target_triple,
-                             llvm::raw_pwrite_stream& error_stream,
-                             llvm::raw_pwrite_stream& output_stream) -> bool {
+auto CodeGen::CreateTargetMachine() -> llvm::TargetMachine* {
```
zygoloid:
```suggestion

auto CodeGen::CreateTargetMachine() -> llvm::TargetMachine* {
```

toolchain/codegen/codegen.cpp
```diff
@@ -48,24 +43,53 @@ auto PrintAssemblyFromModule(llvm::Module& module,
   std::optional<llvm::Reloc::Model> reloc_model;
   auto* target_machine = target->createTargetMachine(
       target_triple, CPU, Features, target_opts, reloc_model);
-  module.setDataLayout(target_machine->createDataLayout());
-  module.setTargetTriple(target_triple);
+  return target_machine;
+}
+auto CodeGen::EmitCode(llvm::raw_pwrite_stream& dest,
```
zygoloid, line 51:
```suggestion

auto CodeGen::EmitCode(llvm::raw_pwrite_stream& dest,
```

toolchain/codegen/codegen.h
```diff
@@ -8,15 +8,44 @@
...
 
 namespace Carbon {
-// Prints the assembly to stdout for the given llvm module.
-// If print fails, this returns false and any information about the failure is
-// printed to the error stream.
-auto PrintAssemblyFromModule(llvm::Module& module,
-                             llvm::StringRef target_triple,
-                             llvm::raw_pwrite_stream& error_stream,
-                             llvm::raw_pwrite_stream& output_stream) -> bool;
+class CodeGen {
```
zygoloid, line 15:
```suggestion

class CodeGen {
```

toolchain/codegen/codegen.h
```diff
@@ -8,15 +8,44 @@
...
+  // is printed to the error stream.
+  auto CreateTargetMachine() -> llvm::TargetMachine*;
+  // Using the llvm pass emits either assembly or object code to dest.
+  // Returns false in case of failure, and any information about the failure is
+  // printed to the error stream.
+  auto EmitCode(llvm::raw_pwrite_stream& dest,
+                llvm::TargetMachine* target_machine,
+                llvm::CodeGenFileType file_type) -> bool;
+};
 }  // namespace Carbon
```
zygoloid, line 54:
```suggestion

}  // namespace Carbon
```

toolchain/codegen/codegen.h
```diff
@@ -8,15 +8,44 @@
...
+class CodeGen {
+ public:
+  CodeGen(llvm::Module& module, llvm::StringRef triple,
+          llvm::raw_pwrite_stream& error_stream,
+          llvm::raw_pwrite_stream& output_stream)
+      : Module(module),
+        output_stream_(output_stream),
+        error_stream_(error_stream),
+        target_triple(triple){};
+  // Generates the object code file.
```
zygoloid, line 25:
```suggestion

  // Generates the object code file.
```

toolchain/codegen/codegen.h
```diff
@@ -8,15 +8,44 @@
...
+  // Creates the target machine for triple.
+  // Returns nullptr in case of failure, and any information about the failure
+  // is printed to the error stream.
+  auto CreateTargetMachine() -> llvm::TargetMachine*;
+  // Using the llvm pass emits either assembly or object code to dest.
+  // Returns false in case of failure, and any information about the failure is
+  // printed to the error stream.
+  auto EmitCode(llvm::raw_pwrite_stream& dest,
+                llvm::TargetMachine* target_machine,
+                llvm::CodeGenFileType file_type) -> bool;
```
zygoloid:
```suggestion
  // Creates the target machine for triple.
  // Returns nullptr in case of failure, and any information about the failure
  // is printed to the error stream.
  auto CreateTargetMachine() -> llvm::TargetMachine*;

  // Using the llvm pass emits either assembly or object code to dest.
  // Returns false in case of failure, and any information about the failure is
  // printed to the error stream.
  auto EmitCode(llvm::raw_pwrite_stream& dest,
                llvm::TargetMachine* target_machine,
                llvm::CodeGenFileType file_type) -> bool;

  llvm::Module& Module;
  llvm::raw_pwrite_stream& output_stream_;
  llvm::raw_pwrite_stream& error_stream_;
  llvm::StringRef target_triple;
```

toolchain/codegen/codegen.cpp
```diff
@@ -48,24 +43,53 @@ auto PrintAssemblyFromModule(llvm::Module& module,
...
+  auto* target_machine = CreateTargetMachine();
+  if (target_machine == nullptr) {
+    return false;
+  }
+  bool has_errors = EmitCode(output_stream_, target_machine,
+                             llvm::CodeGenFileType::CGFT_AssemblyFile);
+  delete target_machine;
+  return has_errors;
+}
+auto CodeGen::GenerateObjectCode(llvm::StringRef output_file) -> bool {
```
zygoloid:
```suggestion

auto CodeGen::GenerateObjectCode(llvm::StringRef output_file) -> bool {
```

toolchain/codegen/codegen.cpp
```diff
@@ -48,24 +43,53 @@ auto PrintAssemblyFromModule(llvm::Module& module,
   std::optional<llvm::Reloc::Model> reloc_model;
   auto* target_machine = target->createTargetMachine(
```
zygoloid: Prefer `std::unique_ptr<TargetMachine>` instead of manual memory management with `delete`.

toolchain/driver/driver.cpp
```diff
@@ -163,6 +165,22 @@ auto Driver::RunDumpSubcommand(DiagnosticConsumer& consumer,
     args = args.drop_front();
   }
 
+  llvm::StringRef output_file;
+  if (dump_mode == DumpMode::ObjectCode) {
+    if (!args.empty() && args.front().starts_with("--target_triple=")) {
```
zygoloid:
```suggestion
    if (!args.empty() && args.front().starts_with("--target=")) {
```
I think we can use a shorter flag name here. (Calling these "triple"s is a somewhat-LLVM-specific anachronism these days, so I'd prefer that we don't use that word in our UI.)

zygoloid: Hm, looks like we're already using this flag name for `Assembly` mode, and we should be consistent. You could change them both in this PR, or keep the `--target_triple` name for now and switch to `--target` in a separate PR.

toolchain/driver/driver.cpp
```diff
@@ -163,6 +165,22 @@ auto Driver::RunDumpSubcommand(DiagnosticConsumer& consumer,
...
+      args = args.drop_front();
+    }
+    if (!args.empty() && args.front().starts_with("--output_file=")) {
+      output_file = args.front().split("=").second;
+      args = args.drop_front();
+    }
+    if (output_file.empty()) {
+      error_stream_ << "ERROR: Must provide an output file.\n";
+      return false;
+    }
```
zygoloid: Would it make sense to support `output_file` for all commands, and make it optional for everything other than `objcode` mode? It seems useful for the other modes too.

toolchain/driver/driver.cpp
```diff
@@ -163,6 +165,22 @@ auto Driver::RunDumpSubcommand(DiagnosticConsumer& consumer,
     args = args.drop_front();
   }
 
+  llvm::StringRef output_file;
+  if (dump_mode == DumpMode::ObjectCode) {
+    if (!args.empty() && args.front().starts_with("--target_triple=")) {
+      target_triple = args.front().split("=").second;
+      args = args.drop_front();
+    }
+    if (!args.empty() && args.front().starts_with("--output_file=")) {
```
zygoloid: It'd be preferable to support flags appearing in any order. Maybe we could put a loop around this flag-parsing code? We could keep going until we run out of args or find an argument that doesn't start with `--`, and diagnose any `--` flag that we don't recognize.

toolchain/driver/driver.cpp
```diff
@@ -248,8 +266,16 @@ auto Driver::RunDumpSubcommand(DiagnosticConsumer& consumer,
...
+    CodeGen codegen =
+        CodeGen(*module, target_triple, error_stream_, output_stream_);
+    has_errors |= !codegen.PrintAssembly();
+    return !has_errors;
+  }
+
+  if (dump_mode == DumpMode::ObjectCode) {
+    CodeGen codegen =
+        CodeGen(*module, target_triple, error_stream_, output_stream_);
+    has_errors |= !codegen.GenerateObjectCode(output_file);
```
zygoloid: Our behavior would be more uniform if we set `output_stream_` to be the output file here, and don't pass a separate output file to `GenerateObjectCode`.

fasiddique: But if we don't provide an output file to write the object code into how do we generate the binary?

toolchain/driver/driver.cpp
fasiddique: Thank you for the suggestion! I will change it to` target` in a separate PR.

toolchain/driver/driver.cpp
fasiddique: will fix it in a separate PR.

toolchain/codegen/codegen.cpp
```diff
@@ -37,35 +34,65 @@ auto PrintAssemblyFromModule(llvm::Module& module,
...
 }
+
+auto CodeGen::PrintAssembly() -> bool {
+  auto target_machine = CreateTargetMachine();
+  if (target_machine == nullptr) {
+    return false;
+  }
+  bool has_errors = EmitCode(output_stream_, target_machine.get(),
+                             llvm::CodeGenFileType::CGFT_AssemblyFile);
+  return has_errors;
```
zygoloid:
```suggestion
  return EmitCode(output_stream_, target_machine.get(),
                  llvm::CodeGenFileType::CGFT_AssemblyFile);
```

toolchain/codegen/codegen.cpp
```diff
@@ -37,35 +34,65 @@ auto PrintAssemblyFromModule(llvm::Module& module,
...
+  }
+  std::error_code ec;
+  llvm::raw_fd_ostream dest(output_file, ec, llvm::sys::fs::OF_None);
+  if (ec) {
+    error_stream_ << "Error: Could not open file: " << ec.message() << "\n";
+    return false;
+  }
+  bool has_errors = EmitCode(dest, target_machine.get(),
+                             llvm::CodeGenFileType::CGFT_ObjectFile);
+  return has_errors;
```
zygoloid:
```suggestion
  return EmitCode(dest, target_machine.get(),
                  llvm::CodeGenFileType::CGFT_ObjectFile);
```

toolchain/driver/driver.cpp
```diff
@@ -248,8 +270,16 @@ auto Driver::RunDumpSubcommand(DiagnosticConsumer& consumer,
 
   if (dump_mode == DumpMode::Assembly) {
     consumer.Flush();
-    has_errors |= !Carbon::PrintAssemblyFromModule(
-        *module, target_triple, error_stream_, output_stream_);
+    CodeGen codegen =
+        CodeGen(*module, target_triple, error_stream_, output_stream_);
```
zygoloid:
```suggestion
    CodeGen codegen(*module, target_triple, error_stream_, output_stream_);
```

toolchain/driver/driver.cpp
```diff
@@ -248,8 +270,16 @@ auto Driver::RunDumpSubcommand(DiagnosticConsumer& consumer,
...
+    has_errors |= !codegen.PrintAssembly();
+    return !has_errors;
+  }
+
+  if (dump_mode == DumpMode::ObjectCode) {
+    CodeGen codegen =
+        CodeGen(*module, target_triple, error_stream_, output_stream_);
```
zygoloid:
```suggestion
    CodeGen codegen(*module, target_triple, error_stream_, output_stream_);
```

toolchain/driver/driver.cpp
zygoloid: I'm thinking of something like this:
```
    std::error_code ec;
    llvm::raw_fd_ostream dest(output_file, ec, llvm::sys::fs::OF_None);
    if (ec) {
      error_stream_ << "Error: Could not open file: " << ec.message() << "\n";
      return false;
    }
    CodeGen codegen(*module, target_triple, error_stream_, dest);
    has_errors |= !codegen.GenerateObjectCode();
```

toolchain/driver/testdata/error_codegen_output_file.carbon
```diff
@@ -0,0 +1,8 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// RUN: %{not} %{carbon} dump objcode --target_triple=x86_687-unknown-linux-gnu %s | %{FileCheck-strict}
+// CHECK:STDERR: ERROR: {{.*}}
```
zygoloid: Can we check the error message here and in the other test?

zygoloid (commented):
Please can you add a passing testcase too? I don't think we're ready to check the contents of the output file in any way, but we should at least have a test that verifies the command succeeds when given a valid set of arguments.

toolchain/driver/testdata/error_codegen_output_file.carbon
fasiddique: The reason I am not checking the error message is, we get the error message from llvm and if it's changed on their side, these tests will fail. 
If we want to check the entire error message, instead of using llvm's error message we can use our own generic error message. However, in this case the error message may not be meaningful.

zygoloid: I'm expecting the error here to be the "Must provide an output file." error from driver.cpp, not something from LLVM.

fasiddique: I am sorry for misunderstanding this. Added the error message check for this one.

toolchain/driver/testdata/error_codegen_output_file_without_triple.carbon
```diff
@@ -0,0 +1,8 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// RUN: %{not} %{carbon} dump objcode %s | %{FileCheck-strict}
+// CHECK:STDERR: ERROR: {{.*}}
```
zygoloid: This one is presumably also "Must provide an output file."

toolchain/driver/testdata/error_codegen_objcode_target_triple.carbon
```diff
@@ -0,0 +1,8 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// RUN: %{not} %{carbon} dump assembly --target_triple=x86_687-unknown-linux-gnu %s | %{FileCheck-strict}
```
zygoloid: This is testing `assembly` mode not `objcode` mode, but based on the filename I think you meant for this to be an `objcode` test. You'll also need to provide an output file name or this test will be the same as `error_codegen_objcode_output_file_with_triple.carbon`.

toolchain/driver/testdata/error_codegen_objcode_target_triple.carbon
```diff
@@ -0,0 +1,8 @@
...
+// CHECK:STDERR: ERROR: {{.*}}
```
zygoloid: I would like to see at least some of the error message here, so that we can have confidence that this test is testing the right thing. You should include something in the error message about the `--target_triple=` flag being invalid; that way users will know what they got wrong (that won't be in the LLVM error message) and also we can avoid test failures if the LLVM error string changes by keeping that part of the diagnostic inside `{{.*}}`.

toolchain/driver/testdata/error_codegen_objcode_output_file_without_triple.carbon
```diff
@@ -0,0 +1,8 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
```
zygoloid: The name of this test file doesn't match what it's doing: it says `output_file_without_triple`, but it has neither an output file nor a target specified.

toolchain/driver/testdata/error_codegen_objcode_output_file_with_triple.carbon
```diff
@@ -0,0 +1,8 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
```
zygoloid: Let's rename this test to `error_codegen_objcode_no_output_file.carbon` (the triple isn't relevant here, and I think it's useful for the test name to say "no output file" not just "output file").

zygoloid (approved):

Issue #1391, opened 2022-07-15
New name for "constant" value phase


josh11b:
(separate from the question of whether the value phases are actually just value categories)

The thing we want to name is "a value that is known at type-checking time" in contrast to "symbolic value", which is a value that will be known at code generation / monomorphization time but type checking is done without knowing its value, or "runtime value" which is not known until runtime.

Original suggestion for this name in [#typesystem on Discord](https://discord.com/channels/655572317891461132/708431657849585705/996547601451204630). Suggestion to find a new name in [PR #1378](https://github.com/carbon-language/carbon-lang/pull/1378/files/198c7bd152bc4e223f1fc2484455e17ba31f19f9#r921592116).

Turbine1991:
! is typically associated with negating a value or use within the inequality operator.

I'm not sure const comes to anyone's mind when they see !. Remember to make a language as readable and inviting as possible. There's a reason we don't all use Rust.

chandlerc:
Just confirming that the leads are happy with the tweaked names here:

https://github.com/carbon-language/carbon-lang/issues/1391#issuecomment-1185245169

Marking this as decided!

Turbine1991:
> Just confirming that the leads are happy with the tweaked names here:
> 
> [#1391 (comment)](https://github.com/carbon-language/carbon-lang/issues/1391#issuecomment-1185245169)
> 
> Marking this as decided!

To be quite honest, if it's too different to other languages. People may as well learn Rust. This language will likely be sidetracked to some existing projects which will get updates.

I've given up on Carbon and started learning Rust as it's nothing like C++/Javascript/C# syntactically, and Rust is almost just as different.

chandlerc:
Just a quick reminder to not use specific issues to post more general feedback -- that should happen on its own discussion thread. Also, please keep feedback constructive.

Turbine1991:
> Just a quick reminder to not use specific issues to post more general feedback -- that should happen on its own discussion thread. Also, please keep feedback constructive.

It was constructive criticism to not alienate the C++ folk.

chandlerc:
> > Just a quick reminder to not use specific issues to post more general feedback -- that should happen on its own discussion thread. Also, please keep feedback constructive.
> 
> It was constructive criticism to not alienate the C++ folk.

[...]

PR #2964, opened 2023-06-30
Expression phase terminology

josh11b:
Update terminology around expression properties:

-   "value phase" -> "expression phase"
-   "symbolic value" -> "symbolic constant"
-   "constant" -> "template constant"
-   "constant or symbolic value" -> "constant"

Implements the decision in #1391 .


chandlerc (commented):
This looks like a good, simple proposal to me. No comments really, thanks for documenting where we ended up here.

jonmeow:
NB: Added an empty commit to unjam the test action, since it looks like this may be approved without edits.

chandlerc (approved):
Approving for leads.

PR #2970, opened 2023-07-05
Narrow example addr pattern to self.

jonmeow:
Although discussed for other situations, `self` is the only concrete use-case of `addr`.


josh11b (approved):
Thank you!

PR #2971, opened 2023-07-06
Add a fallback `test` workflow.

chandlerc:
To make this cleaner, this switches the main `test` workflow to use exclusion path patterns as that was the clearly documented intent anyways.

This works around a problem with required GitHub status checks: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

Without something like this, we won't be able to use the merge queue (or require passing tests).


jonmeow (approved):
Good investigation! I may look at an alternate fix to unify the files, but this makes sense to get in.

PR #2972, opened 2023-07-06
Use paths-filter to fix the test action.

jonmeow:
Trying to fix merge queue support for PRs that don't need tests, #2971 didn't work because it lacks the matrix configuration. This takes a different approach from that, instead filtering each step based on affected paths. This way we're also less likely to see skew in behavior.


chandlerc (approved):
Ugh. This is ... Ugh.

I can't say I'm thrilled by the ever expanding set of dependencies of our workflows, and the cost / complexity of this.

TBH, I would probably have replicated the matrix rather than this. But I guess if others prefer this, it does seem like one of the two local "maxima" that are left to us.

jonmeow:
FWIW, for me, I prioritize the low duplication here: that is, the alternative approach is 2 copies of the matrix and 6 copies of the paths, which would need to be manually kept in sync for any changes.

PR #2977, opened 2023-07-06
Basic support for non-defining declarations of functions.

zygoloid:
This doesn't support redeclaration, so there's no way to provide a definition in Carbon for a forward-declared function yet.

### DIFF
--- a/toolchain/semantics/semantics_handle_function.cpp
+++ b/toolchain/semantics/semantics_handle_function.cpp
@@ -6,9 +6,43 @@
 
 namespace Carbon {
 
+static auto HandleFunctionDeclarationOrDefinition(
+    SemanticsContext& context, ParseTree::Node /*parse_node*/)
+    -> SemanticsNodeId {
+  SemanticsTypeId return_type_id = SemanticsTypeId::Invalid;
+  if (context.parse_tree().node_kind(context.node_stack().PeekParseNode()) ==
+      ParseNodeKind::ReturnType) {
+    return_type_id =
+        context.node_stack().Pop<SemanticsTypeId>(ParseNodeKind::ReturnType);
+  } else {
+    // Canonicalize the empty tuple for the implicit return.
+    context.CanonicalizeType(SemanticsNodeId::BuiltinEmptyTupleType);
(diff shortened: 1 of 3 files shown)


toolchain/semantics/semantics_handle_function.cpp
```diff
@@ -6,9 +6,43 @@
 
 namespace Carbon {
 
+static auto HandleFunctionDeclarationOrDefinition(
```
jonmeow: Can we call this something like "HandleFunctionSignature" or "BuildFunctionDeclaration"? The "OrDefinition" part makes me think it's actually doing definition logic.

Also, comments may be good -- in particular to emphasize that this isn't just a Handle function with an accidental `static`. :)

toolchain/semantics/semantics_handle_function.cpp
```diff
@@ -6,9 +6,43 @@
...
+    SemanticsContext& context, ParseTree::Node /*parse_node*/)
```
jonmeow: You don't need to pass parse_node if you aren't using it.

toolchain/semantics/semantics_handle_function.cpp
```diff
@@ -31,57 +65,36 @@ auto SemanticsHandleFunctionDefinition(SemanticsContext& context,
...
-  }
-  auto param_refs_id = context.node_stack().Pop<SemanticsNodeBlockId>(
-      ParseNodeKind::ParameterList);
-  auto name_context = context.PopDeclarationName();
-  auto fn_node = context.node_stack().PopForSoloParseNode(
-      ParseNodeKind::FunctionIntroducer);
+  // Process the declaration portion of the function.
+  auto decl_id = HandleFunctionDeclarationOrDefinition(context, parse_node);
+  auto function_id =
+      context.semantics_ir().GetNode(decl_id).GetAsFunctionDeclaration();
```
jonmeow: Maybe also return the function_id in order to avoid this series of calls?

jonmeow (approved):
Looks good! Minor comments, please address before merging.

PR #2978, opened 2023-07-09
Add a rich command-line argument parsing library.

chandlerc:
This library is designed around supporting the kinds of use cases we expect in the toolchain and other Carbon tools. It supports subcommands, options, and prints help.

There is still a decent chunk of work to be done to finish polishing this, but it should give us a solid starting point.

For details about the library and a brief roadmap, see the main comment in `command_line.h` which provides a comprehensive overview of the library and a roadmap of the remaining work.

Porting the driver to this went fairly well, but did require some changes. That port is separated into a follow-up PR #2979.


chandlerc:
@lnihlen was interested in helping with early reviews here, so tagging her.

Also @fasiddique is probably interested in this as well (sorry if I messed up anything in codegen!)

common/argparse.h
```diff
@@ -0,0 +1,870 @@
...
+#include "llvm/Support/Allocator.h"
+
+namespace Carbon {
+
+// # Command-line argument parsing library.
+//
+// This is a collection of tools to describe both simple and reasonably complex
+// command line interfaces, and parse arguments based on those descriptions. It
+// optimizes for command line tools that will parse arguments exactly once per
+// execution, and specifically tools that make heavy use a subcommand-style
```
lnihlen: "make heavy use *of* subcommand-style"

common/argparse.h
```diff
@@ -0,0 +1,870 @@
...
+// command line interfaces.
+//
+// ## Terminology used by this library
+//
+// _Argument_ or _arg_: One of the parsed components of the command line.
+//
+// _Option_: A _named_ argument starting with `--` to configure some aspect of
+// the tool. These often have both long spellings that start with `--` and
+// short, single character spellings that start with `-` and can be bundled with
+// other single character options. _Flag_: A boolean or binary option. These can
```
lnihlen: line break here with _Flag_?

toolchain/codegen/codegen.cpp
lnihlen, line 1: What's going on with these changes? It's not obvious to me how they relate to the command-line parser.

common/argparse.h
```diff
@@ -0,0 +1,870 @@
...
+  // configuration.
+  //
+  // The `out` stream is treated like `stdout` would be for a Unix-style command
+  // line tool, and `errors` like `stderr`: any errors or diagnostic information
+  // are printed to `errors`, but meta-actions like printing a command's help go
+  // to `out`.
+  static auto Parse(llvm::ArrayRef<llvm::StringRef> unparsed_args,
+                    llvm::raw_ostream& out, llvm::raw_ostream& errors,
+                    const CommandInfo& command_info,
+                    llvm::function_ref<void(CommandBuilder&)> build)
```
lnihlen: What's the advantage you intend to gain from putting the configuration logic in this `build` function? I know you get repeatability, but this library is optimized for one-shot configuration and parsing. What was the thinking there?

toolchain/codegen/codegen.cpp
lnihlen: Ah, reading ahead I see this PR adds the library and changes Carbon to use it. It might be worth splitting this into two PRs.

common/argparse.h
```diff
@@ -0,0 +1,870 @@
...
+    static auto OneOfValue(llvm::StringRef string, T value) -> OneOfValueT<T>;
+
+    // Configures the argument to store a parsed value in the provided storage.
+    // Each time the argument is parsed, it will write a new value to this
+    // storage.
+    //
+    // Exactly one of this method or `AppendOneOf` below must be configured for
+    // the argument.
+    //
+    // For one-of arguments, this method also provides an array of possible
```
lnihlen: This list of requirements is a lot to parse, but looking at the examples in the test code it makes a lot more sense. Perhaps providing a short example here and for `AppendOneOf` would help? Otherwise I feel like I'm reading the disclaimer text at the end of a pharmaceutical advertisement :)

toolchain/driver/driver.cpp
lnihlen, line 1: It's pretty nice to see the cleanup in this file afforded by argparse.

lnihlen (commented):
Just a first pass with some initial feedback. Overall very clean and I find this an obvious improvement over the old solutions. I had one question about the API and also wonder if this might be better off as 2 PRs, one adding the library and one making the changes to the driver/others to use it.

toolchain/driver/driver.cpp
```diff
@@ -22,286 +25,481 @@
...
+
+    b.AddFlag(
+        {
+            .name = "force-obj-output",
+            .help = R"""(
+Force writing a binary object file, even when writing to stdout.
+
+This flag is only used when the code generation output file is set to stdout. In
+that case, it will override the forced default of textual assembly and output a
+binary output.
```
fasiddique: Why should we emit the object code to stdout? It is a machine code and so it would print out some gibberish in the stdout which would be impossible to comprehend. Instead should we not print the assembly to stdout and force the user to provide an output file if they want the object code?

toolchain/driver/driver.cpp
```diff
@@ -22,286 +25,481 @@
...
+The output filename for codegen.
+
+When this is a file name, either textual assembly or a binary object will be
+written to it based on the flag `--asm-output`. The default is to write a binary
+object file.
+
+This can also be the string `-` which will write the output to stdout. In that
+case, the flag `--asm-output` is ignored and the output defaults to textual
+assembly. Binary object output can be forced by enabling `--force-obj-output`.
+)""",
```
fasiddique: What is the purpose of allowing '-' as an output file name?  If the user doesn't provide an output file name, based on the codegen flag we can emit the assembly to stdout. However, if the codegen flag is set to --force-obj-output and the output file is not given, we can either throw an error or emit the object code to a default output file.

toolchain/driver/driver.cpp
```diff
@@ -22,286 +25,481 @@
...
+  }
+  if (vlog_stream_) {
+    CARBON_VLOG() << "assembly:\n";
+    codegen->EmitAssembly(*vlog_stream_);
   }
 
-  if (dump_mode == DumpMode::ObjectCode) {
+  if (options.output_file_name == "-") {
+    if (options.force_obj_output) {
+      if (!codegen->EmitObject(output_stream_)) {
```
fasiddique: I don't think we should allow emitting object code to stdout for the reason mentioned in my earlier comment.

toolchain/driver/driver.cpp
```diff
@@ -22,286 +25,481 @@
...
+      if (!codegen->EmitAssembly(output_stream_)) {
+        return false;
+      }
+    }
+  } else {
+    llvm::SmallString<256> output_file_name = options.output_file_name;
+    if (output_file_name.empty()) {
+      output_file_name = options.input_file_name;
+      llvm::sys::path::replace_extension(output_file_name,
+                                         options.asm_output ? ".s" : ".o");
```
fasiddique: if the output file is empty and the codegen flag is set to assembly why are we creating an output file instead of printing the assembly directly to the stdout?

toolchain/driver/driver.cpp
chandlerc: This is a somewhat common convention for directing output to stdout instead of a file:

```
> clang -c -S -o - example.c
        .section        __TEXT,__text,regular,pure_instructions
        .build_version macos, 13, 0
        .globl  _main                           ; -- Begin function main
        .p2align        2
_main:                                  ; @main
        .cfi_startproc
; %bb.0:
        sub     sp, sp, #16
        .cfi_def_cfa_offset 16
        mov     w0, #0
        str     wzr, [sp, #12]
        add     sp, sp, #16
        ret
        .cfi_endproc
                                        ; -- End function
.subsections_via_symbols
```

The common convention in compilers (Clang and GCC at least) is to select an output name based on the input name when an output flag isn't provided:

```
> clang -c -S example.c
> cat example.s
        .section        __TEXT,__text,regular,pure_instructions
        .build_version macos, 13, 0
        .globl  _main                           ; -- Begin function main
        .p2align        2
_main:                                  ; @main
        .cfi_startproc
; %bb.0:
        sub     sp, sp, #16
        .cfi_def_cfa_offset 16
        mov     w0, #0
        str     wzr, [sp, #12]
        add     sp, sp, #16
        ret
        .cfi_endproc
                                        ; -- End function
.subsections_via_symbols
```

While for testing, it is nice to have this streamed to stdout, that's somewhat rarely the desired behavior.

My goal with all of the driver changes like this is to try to line up the behavior with what is more conventional for compilers on the command line.

I'll try to write up a summary of this in the up-coming split PR description as well.

toolchain/driver/driver.cpp
chandlerc: While unusual, there are times when this is useful for using Unix pipes to process the object code. And it is something that existing compilers do allow -- they don't even use a scary looking flag for it, it is the *default* for those compilers:

```
> clang -c -o - example.c | llvm-objdump --disassemble -

<stdin>:        file format mach-o arm64

Disassembly of section __TEXT,__text:

0000000000000000 <ltmp0>:
       0: d10043ff      sub     sp, sp, #0x10
       4: 52800000      mov     w0, #0x0
       8: b9000fff      str     wzr, [sp, #0xc]
       c: 910043ff      add     sp, sp, #0x10
      10: d65f03c0      ret
```

I think this isn't a good default, but it seemed reasonable to preserve a way of achieving this where it is desired, and just make sure it is explicitly requested by the user.

Does that make sense?

toolchain/driver/driver.cpp
chandlerc: (tried to answer above)

toolchain/driver/driver.cpp
chandlerc: I think I covered this somewhat with the answer above, but just to be clear -- this is a fairly strong convention in compilers and several other command line tools to synthesize a filename with the correct extension from the input filename.

I think outside of compiler test suites, this is also the likely desired behavior.

chandlerc (commented):
> [I] wonder if this might be better off as 2 PRs, one adding the library and one making the changes to the driver/others to use it.

I think this is a good suggestion. As the review already shows, there are pretty separate questions there. And because I didn't split it up, there isn't any real discussion in the commit message about *why* I made a bunch of the choices on the driver. Thanks for the suggestion!

I'll leave this PR for the library, and create a new one and move the driver updates there.

Some quick replies to the driver questions here while I make the move -- but maybe good to move most of those discussions over.

I'll come back with responses on the library itself once I've split things.

chandlerc:
The port of the toolchain to this library has been split into #2979 now.

To *just* see the changes in the toolchain and not the changes in this PR you can use this link:
[`046fd9a` (#2979)](https://github.com/carbon-language/carbon-lang/pull/2979/commits/046fd9a44a21e9b54a8acb8790e61ed6c9fa08f0)

common/argparse.h
chandlerc: Done.

Also, FYI if you're not familiar w/ this part of GitHub -- you can cilck the "+/-"-in-a-file looking button to insert a suggested edit into a comment. There is a whole system for recognizing suggested edits, and GH can directly turn them into commits that apply those edits. It's super nice. Docs are here FWIW: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request#:~:text=Optionally%2C%20to%20suggest%20a%20specific%20change%20to%20the%20line%20or%20lines%2C%20click%20%2C%20then%20edit%20the%20text%20within%20the%20suggestion%20block.

common/argparse.h
chandlerc: Doh, good catch, done!

common/argparse.h
chandlerc: Done. Let me know if other tweaks would help here.

common/argparse.h
chandlerc: I guess, where would the configuration logic go instead?

I need some builder-like logic here. A fluent-style API looks really nice, but because these can *recurse* in complex ways, a lambda-passed-a-builder seemed a bit more clean.

If there are other ways to get the builder / DSL of constructing these into the API, happy to explore them.

chandlerc (commented):
Think I replied to all the remaining comments now, thanks again for the early looks!

common/argparse.h
```diff
@@ -0,0 +1,895 @@
...
+#include "llvm/Support/Allocator.h"
+
+namespace Carbon {
+
+// # Command-line argument parsing library.
+//
+// This is a collection of tools to describe both simple and reasonably complex
+// command line interfaces, and parse arguments based on those descriptions. It
+// optimizes for command line tools that will parse arguments exactly once per
+// execution, and specifically tools that make heavy use of a subcommand-style
```
lnihlen:
```suggestion
// execution, and specifically tools that make heavy use of subcommand-style
```

common/argparse.h
```diff
@@ -0,0 +1,895 @@
...
+// removed.
+//
+// _Command_: The container of options and positional arguments to parse and an
+// action to take when successful.
+//
+// _Subcommand_: A command nested within another command and identified by a
+// specific name that ends the parsing of arguments based on the parent and
+// switches to parse based on the specific subcommand's options and positional
+// arguments. A command with subcommands cannot parse positional arguments, only
+// a leaf command can do that.
```
lnihlen: To improve clarity consider defining "leaf command" here or in the _Command_ entry directly above.

common/argparse.h
```diff
@@ -0,0 +1,895 @@
...
+//
+// _Action_: An open ended callback, typically reflecting a specific subcommand
```
lnihlen:
```suggestion
// _Action_: An open-ended callback, typically reflecting a specific subcommand
```

common/argparse.h
```diff
@@ -0,0 +1,895 @@
...
+//     my-diff-tool a.txt b.txt c.txt -- new-a.txt new-b.txt new-c.txt
+//
+// The `--` used in this way *both* ends option parsing and the positional
+// argument sequence.
+//
+// ## Help text blocks and rendering
+//
+// At many points in this library, a block of text is specified as a string.
+// These should be written using multi-line raw string literals that start on
+// there own line such as:
```
lnihlen:
```suggestion
// their own line such as:
```

common/argparse.h
lnihlen: Looks good!

common/argparse.h
lnihlen: I've thought about it a bit. I prefer this lambda design over the fluent style for a couple of reasons:

a) More control over return types on the config functions. Fluent-style APIs are obligated to return the mutable ref to themselves.

b) Better clarity in the use of `CommandBuilder`. Often the fluent-style objects end up with comments in the header indicating when it is OK to call the configuration functions. For example, it might be possible to call additional config functions after parsing, and the compiler does not detect this misuse without cleverness.

c) Cleaner code. Those long call chains can lead to difficult-to-read (and even harder-to-debug) code stanzas.

d) Good configuration API separation from the library's other parts.

The *only* downside I see is that this promotes centralization of the configuration code. Without call chaining, all library configuration must happen in one function in one file. This could be a feature, but I imagine over time Carbon will grow in complexity, and the `driver.cpp` file will grow along with it.

Again, not saying this is the wrong approach. I think the "Where would the config logic go instead?" question looms large here. Some flags will likely impact multiple subsystems, so putting the config logic with the subsystem feels like a poor separation of concerns.

So I think my feedback is "LGTM so long as we accept a really long client config file."

lnihlen (approved):
LGTM. Oh and I think "Args" is a fine name for it.

toolchain/driver/driver.cpp
fasiddique: Thank you for the detailed explanation and I know that compilers allow that and this can be kept as an option for the scenario you have explained above. However, I don't think that it should be a default only because other compilers are doing it. The user can always dump the object code to a file and disassemble the file using the same command.

common/BUILD
```diff
@@ -12,6 +12,29 @@ cc_library(
     ],
 )
 
+cc_library(
+    name = "argparse",
+    srcs = ["argparse.cpp"],
+    hdrs = ["argparse.h"],
```
jonmeow: I believe our style is to use underscores to separate words in filename (arg_parse.h, rather than argparse.h)

toolchain/driver/testdata/error_codegen_target_triple.carbon
```diff
@@ -0,0 +1,9 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
```
jonmeow: Looks like a merge mistake

common/ostream.h
```diff
@@ -62,6 +62,35 @@ void PrintTo(T* p, std::ostream* out) {
...
+// should simply provide a `Print` method which will give them all of these
+// while also being more usable from a debugger. But enums cannot add this so we
+// let them define a single operator and connect to it here.
+//
+// This overload is disabled for non-enum types, or when there is no
+// `raw_ostream` streaming support. And to make this overload be unusually low
+// priority, it is designed to take even the `std::ostream` parameter as a
+// template, and SFINAE disable itself unless that template parameter matches
+// `std::ostream`. This ensures that an *explicit* operator will be preferred
+// when provided.
```
jonmeow: I'm confused by this, and what it prints for an enum. Can you please add a unit test to show the result? Also, what's the specific use-case that's making you add it? Is this for gtest, and should that be mentioned (and maybe even testonly under //testing rather than throughout code)?

Note in part, I thought we'd been using the EnumBase to provide reliable printing.

common/ostream.h
```diff
@@ -62,6 +62,35 @@ void PrintTo(T* p, std::ostream* out) {
...
+template <typename S, typename T,
+          typename = std::enable_if_t<std::is_enum_v<T>>,
+          typename = std::enable_if_t<HasRawOstream<T>>,
+          typename = std::enable_if_t<std::is_base_of_v<
+              std::ostream, std::remove_reference_t<std::remove_cv_t<S>>>>,
```
jonmeow: I'm surprised here that S needs to be a template instead of `std::ostream`, is that right?

common/argparse.h
```diff
@@ -0,0 +1,895 @@
...
+    // the command summarized as briefly as possible. Ideally, each variation on
+    // a single line. The goal is to show the *structure* of different command
+    // invocations, not to be comprehensive.
+    //
+    // When omitted, the library will generate these based on the parsing logic.
+    // The generated usage is expected to be suitable for the vast majority of
+    // users, but can be customized in cases where necessary.
+    llvm::StringRef usage = "";
+
+    // An optional epilog multi-line block of text appended to the help display
```
jonmeow: epilog -> epilogue

From what I can find, [epilog is in the minority, even in the US](https://sapling.ai/usage/epilog-vs-epilogue).

jonmeow:
> One open question is whether `Args` is the best top-level name, or it should switch to `CLI` or something else. Suggestions welcome!

Elsewhere, for example in semantics and lowering, I've tried to adopt namespaces for similar reasons as to why you're using Args here. You've asked for it to be removed. I don't understand how this is different; maybe Args should just be removed, for consistency?

common/argparse.cpp
```diff
@@ -0,0 +1,1362 @@
...
+  void PrintArgLongValues(const Arg& arg, llvm::StringRef indent) const;
+  void PrintArgHelp(const Arg& arg, llvm::StringRef indent) const;
+
+  void PrintRawUsageCommandAndOptions(
+      const Command& command,
+      int max_option_width = MaxLeafOptionUsageWidth) const;
+  void PrintRawUsage(const Command& command, llvm::StringRef indent) const;
+  void PrintUsage(const Command& command) const;
+
+  void PrintHelpSubcommands(const Command& command) const;
```
jonmeow: It may be worth commenting more of your functions. For example, is this printing subcommands of help, which is my first thought from the name, or help for subcommands, which should maybe be something like "PrintAllSubcommandHelp" instead, or maybe something else?

common/argparse.h
```diff
@@ -0,0 +1,895 @@
...
+//
+// ## Help text blocks and rendering
+//
+// At many points in this library, a block of text is specified as a string.
+// These should be written using multi-line raw string literals that start on
+// there own line such as:
+//
+// ```cpp
+//     ...
+//     .help = R"""(
```
jonmeow: These help strings make it really hard for me to read files, since indentation is interrupted. What would you think of one of:

- Don't use `R"""`
- Indent the strings, and de-dent in the parsing code.
- Move flags to their own file.
- Use `constexpr` to define these structs outside of scopes.

common/argparse.h
```diff
@@ -0,0 +1,895 @@
...
+// Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
+// tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,
+// quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
+// consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse
+// cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non
+// proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
```
jonmeow:
```suggestion
// tempor incididunt ut labore et dolore magna aliqua.
```

Minor: consider giving less boilerplate example text so that more documentation fits on-screen.

chandlerc:
> > One open question is whether `Args` is the best top-level name, or it should switch to `CLI` or something else. Suggestions welcome!
> 
> Elsewhere, for example in semantics and lowering, I've tried to adopt namespaces for similar reasons as to why you're using Args here. You've asked for it to be removed. I don't understand how this is different; maybe Args should just be removed, for consistency?

Whether this is a good pattern is part of the open question, yes.

To be clear, I actually tried to use a namespace and found that it did not support the features I needed here -- specifically, the ability to both make things private and use a `friend` declaration to grant access to private members. If you had previously hit the same issue and I didn't understand that, definitely sorry about missing that. I don't particularly *like* using a class here, but the alternatives with a namespace were much worse due to the friendship issues I ran into, and so it seemed worth trying out (again maybe).

Definitely interested in other perspectives on this.

common/argparse.cpp
```diff
@@ -0,0 +1,1362 @@
...
+      << "Only one-of arguments have interesting value snippets to print.";
+  bool first = true;
+  for (llvm::StringRef value_string : arg.value_strings) {
+    if (first) {
+      first = false;
+    } else {
+      out_ << ", ";
+    }
+    out_ << value_string;
+  }
```
jonmeow:
```suggestion
  llvm::ListSeparator sep;
  for (llvm::StringRef value_string : arg.value_strings) {
    out_ << sep << value_string;
  }
```

common/argparse.cpp
```diff
@@ -0,0 +1,1362 @@
...
+    out_ << "(";
+    bool first = true;
+    for (llvm::StringRef value_string : arg.value_strings) {
+      if (first) {
+        first = false;
+      } else {
+        out_ << "|";
+      }
+      out_ << value_string;
+    }
```
jonmeow:
```suggestion
    llvm::ListSeparator sep("|");
    for (llvm::StringRef value_string : arg.value_strings) {
      out_ << sep << value_string;
    }
```

common/argparse.h
```diff
@@ -0,0 +1,895 @@
...
+  bool has_default = false;
+  bool is_required = false;
+  bool is_append = false;
+  bool is_help_hidden = false;
+
+  // Meta action storage, only populated if this argument causes a meta action.
+  ActionT meta_action = {};
+
+  // Storage options depending on the kind.
+  union {
```
jonmeow: Why a union instead of a variant? It seems like a variant could be used because I wouldn't expect this to be performance sensitive. The advantage of a variant is it simplifies the destructor.

(I'm hoping I'm not accidentally repeating this comment)

common/argparse.h
```diff
@@ -0,0 +1,895 @@
...
+  CommandInfo info;
+  Command* parent;
+  ActionT action = {};
+  Kind kind = Kind::Invalid;
+
+  bool is_help_hidden = false;
+
+  llvm::SmallVector<std::unique_ptr<Arg>, 8> options = {};
+  llvm::SmallVector<std::unique_ptr<Arg>, 2> positional_args = {};
+  llvm::SmallVector<std::unique_ptr<Command>, 2> subcommands = {};
```
jonmeow: Are the explicit default initializers here doing much?

Can you add some comments explaining how you chose the sizes (2 and 8), versus using the default?

common/argparse_test.cpp
```diff
@@ -0,0 +1,853 @@
...
+    return Args::Parse(
+        args, llvm::errs(), llvm::errs(), TestCommandInfo, [&](auto& b) {
+          b.AddFlag({.name = "flag"}, [&](auto& arg_b) { arg_b.Set(&flag); });
+          b.AddIntegerOption({.name = "option1"},
+                             [&](auto& arg_b) { arg_b.Set(&integer_option); });
+          b.AddStringOption({.name = "option2"},
+                            [&](auto& arg_b) { arg_b.Set(&string_option); });
+          b.Do([] {});
+        });
+  };
```
jonmeow: This may be personal preference, but the structuring of this with nested lambdas for callbacks instead of an object that has flags added then is run (which is how Python's argparse is set up) feels like it yields a more difficult to read end result.

jonmeow (commented):
I still need to finish reading through the argparse code, sorry this is taking me a lot of time to get through. But I thought the incremental comments may be helpful.

common/argparse.cpp
```diff
@@ -0,0 +1,1362 @@
...
+  arg_.has_default = true;
+  arg_.default_integer = integer_value;
+}
+
+void Args::IntegerArgBuilder::Set(ssize_t* integer) {
+  arg_.is_append = false;
+  arg_.integer_storage = integer;
+}
+
+void Args::IntegerArgBuilder::Append(llvm::SmallVectorImpl<ssize_t>* sequence) {
```
jonmeow: Here and elsewhere: why `SmallVectorImpl` instead of `SmallVector`?

common/argparse.cpp
```diff
@@ -0,0 +1,1362 @@
...
+Args::ArgBuilder::ArgBuilder(Arg& arg) : arg_(arg) {}
+
+void Args::FlagBuilder::Default(bool flag_value) {
+  arg_.has_default = true;
+  arg_.default_flag = flag_value;
+}
+
+void Args::FlagBuilder::Set(bool* flag) { arg_.flag_storage = flag; }
+
+void Args::IntegerArgBuilder::Default(ssize_t integer_value) {
```
jonmeow: Why `ssize_t` instead of `int64_t` or even just `int`? Style-wise, I think we lean in the latter direction.
https://google.github.io/styleguide/cppguide.html#Integer_Types

common/argparse.cpp
```diff
@@ -0,0 +1,1362 @@
...
+    case Arg::Kind::Invalid:
+      CARBON_FATAL() << "Option configured without any action or kind!";
+  }
+}
+
+auto Args::Parser::ParseNegatedFlag(const Arg& flag,
+                                    std::optional<llvm::StringRef> value)
+    -> bool {
+  if (flag.kind != Arg::Kind::Flag) {
+    errors_ << "ERROR: Cannot use a negated flag name by prefixing it with "
```
jonmeow: It looks like a fair amount of this code bails on error. Would it be fair to consider changing to `ErrorOr<Success>` instead of using errors_ directly? The Success versus MetaSuccess result could even be `ErrorOr<bool>` where the bool indicates whether the error has already been handled, potentially removing this class from streaming errors.

common/argparse.cpp
```diff
@@ -0,0 +1,1362 @@
...
+    case Arg::Kind::Flag:
+    case Arg::Kind::Invalid:
+      CARBON_FATAL() << "Invalid kind!";
+  }
+}
+
+auto Args::Parser::SplitValue(llvm::StringRef& unparsed_arg)
+    -> std::optional<llvm::StringRef> {
+  // Split out a value if present.
+  std::optional<llvm::StringRef> value = {};
```
jonmeow: Why the explicit initialization on things like this?

common/argparse.cpp
```diff
@@ -0,0 +1,1362 @@
...
+  if (command.parent) {
+    PrintRawUsageCommandAndOptions(*command.parent, MaxParentOptionUsageWidth);
+    out_ << " ";
+  }
+
+  out_ << command.info.name;
+
+  // Buffer the options rendering so we can limit its length.
+  std::string buffer_str;
+  llvm::raw_string_ostream buffer_out(buffer_str);
```
jonmeow: Here and the other spot you're using raw_string_ostream, consider instead using llvm::buffer_ostream. I think it's a mild simplification for your use-case, since you wouldn't need to separately declare the buffer, just use `buffer.str()` (since it inherits from raw_svector_ostream).

(I've spent a lot of time looking at the stream apis now)

common/argparse.h
jonmeow: I think I note this elsewhere, but I find the lambda approach very difficult to read. e.g., consider the test example:

```
  auto parse = [&](llvm::ArrayRef<llvm::StringRef> args) {
    return Args::Parse(
        args, llvm::errs(), llvm::errs(), TestCommandInfo, [&](auto& b) {
          b.AddFlag({.name = "flag"}, [&](auto& arg_b) { arg_b.Set(&flag); });
          b.AddIntegerOption({.name = "option1"},
                             [&](auto& arg_b) { arg_b.Set(&integer_option); });
          b.AddStringOption({.name = "option2"},
                            [&](auto& arg_b) { arg_b.Set(&string_option); });
          b.Do([] {});
        });
  };
  ```

From looking at the toolchain, I think that accurately reflects what real code is intended to look like.

I think that could have been something more like:

```
ArgParser parser(llvm::errs(), llvm::errs(), TestCommandInfo);
parser.AddFlag("flag", "desc", &flag);
parser.AddIntegerOption("option1", "desc", &integer_option);
parser.AddStringOption("option2", "desc", &string_option);
parser.Parse(args);
```

I might've commented on this elsewhere, as I'd been expecting an API closer to argparse in structure. I'm commenting here for context given lnihlen's related response, but I don't see why this is just lambda versus fluent style builder.

I understand that extra complexity would need to be added in spots where custom flag parsing is needed, but I think the latter approach would still be quite adaptable. To me, the difference is that it remains simple for simple use-cases, which I think will remain the majority.

**Review by jonmeow — COMMENTED** [review]
We should probably discuss the namespace options, it's not clear to me why the `friend Arg` is special versus other code we have written. (I think in general, when it gets complex, we tend to just rely on good practice over firm enforcement -- but that may also be a side-effect of namespace design decisions)

File: common/argparse_test.cpp [hunk 31]
```diff
@@ -0,0 +1,853 @@
...
+  auto parse = [&](llvm::ArrayRef<llvm::StringRef> args, llvm::raw_ostream& s) {
+    return Args::Parse(args, s, s, TestCommandInfo, [&](auto& b) {
+      b.AddOneOfOption({.name = "option"}, [&](auto& arg_b) {
+        arg_b.SetOneOf(
+            {
+                arg_b.OneOfValue("x", 1),
+                arg_b.OneOfValue("y", 2),
+                arg_b.OneOfValue("z", 3),
+            },
+            &value);
```
jonmeow: I've been grappling with the OneOf template support. What would you think of an approach that only supports enums?

common/BUILD
chandlerc: My goal was to reference a well-known name for doing argument parsing with 'argparse'. That predates most of the design though, and I'm happy to rename this to whatever makes sense now, especially with whatever top-level name makes sense, whether a namespace or a class or whatever else.

I'm leaving this alone until things are ready to merge just to avoid tracking too mayn renames across branches and such.

common/ostream.h
chandlerc: > I'm confused by this, and what it prints for an enum. Can you please add a unit test to show the result?

Hmmm, there currently aren't any unit tests for this entire file. Everything is tested indirectly in the unittests of the users of this file.

I'm totally happy to add unit tests, but I'd prefer to cover the whole file and to do that in a separate PR. Should I split this out into a pre-requisite PR that also adds testing? Or do a follow-up PR?

> Also, what's the specific use-case that's making you add it? Is this for gtest, and should that be mentioned (and maybe even testonly under //testing rather than throughout code)?

My geal is to support streaming `enum` types to `CHECK`, `VLOG`, and gtest with only a single function rather than multiple ones.

> Note in part, I thought we'd been using the EnumBase to provide reliable printing.

For large enums, this works very well. But I ended up with a bunch of enums with 3-5 values than are unlikely to grow larger. That seems to make the boilerplate larger than the entire enum.

I also don't need any of the other features provided by that class, all I want to do is log the values in a human readable fashion.

It seemed simpler to just directly provide a stream operator for them. But because of the mixture of libraries, there isn't any one stream operator that works. =/ Not really trying to endorse the end result as wonderful, just explaining why I ended up here despite trying to use both the gtest specific stuff and EnumBase first. This was just the least bad option I found at the end.

common/ostream.h
chandlerc: I kept this consistent with the other overload at the bottom of the file. It should allow defining an explicit overload that will be preferred, in case a type has some reason to want different overloads for `std::ostream` and `llvm::raw_ostream`. I'm particularly worried about that coming up for some type in another namespace that still satisfies all the other conditions here.

toolchain/driver/testdata/error_codegen_target_triple.carbon
chandlerc: Fixed.

common/argparse.h
chandlerc: I've added the term at the top level, and tried to cross-reference all of these a bit better. WDYT?

common/argparse.h
chandlerc: I'd like to avoid de-denting in the parsing code. I have specific concerns about supporting that and the complexity it would bring.

Beyond that, I'm open to what styles people prefer here. I find most of the options hard to read TBH. I'd love to have Carbon's block string literals here.

Most of the non-test structs are already separated from the interesting code. I think it's just the help structs, and I'm happy to make those global constexprs.

I think the unittests are the only place where it seems nice to group the structs within the test function bodies when they aren't shared across multiple tests. And I think only the few tests that are specifically testing help text have the giant text blocks there -- is that OK?

common/argparse.h
chandlerc: I chose the spelling to stay consistent with the `argparse` library in Python:
https://docs.python.org/3/library/argparse.html#epilog

That seems the more useful consistency to me, WDYT?

common/argparse.h
chandlerc: While it will simplify the destructor, I think it will make the rest of the code more complex.

In earlier iterations of the library, this had multiple entries with the same type but usefully different names, and a variant wouldn't allow doing that. We'd also have to use the cumbersome embedded discriminator rather than an enum above.

I can make this change, but I think it makes the code worse. And I generally find C++'s variant to be more complex to read and understand than code in this style, so I have a strong preference to avoid it when it isn't an *exact* use case match (a set of types where the type is the exact desired discriminator and documentation for which entry is in use).

common/argparse.h
chandlerc: Mentally, I use them to document the members that don't need initializing in the constructor.

I can add comments for the sizes, but I'm not sure it helps readability at all. It's an intuition that there are likely "lots" of options but very few positional arguments or subcommands. But its no more than an intuition, so it seems hard to meaningfully document.

Sorry if this is repeating a prior discussion, but I treat an undocumented size in a SSO container as a guess that should be changed at any time specific information arrives to suggest a specific size. The only sizes I attribute some specific meaning to are those documented to clarify that meaning. This pattern is just the one I learned working in LLVM with the containers to avoid having repetitive comments saying "This is probably large-ish, so give it N elements in the small size buffer" which doesn't really add any information beyond "N". When we can just omit the size entirely, that's of course nice, and maybe I can do that here so it won't matter.

common/argparse_test.cpp
chandlerc: I'm not sure how only supporting enums avoids templates. I tried a bunch of variations, including one that only worked on enums and tried to type erase them into an integer, and it didn't improve much.

I also have concrete use cases that aren't enums: `--help=(short|long)`. I could create an enum for those cases, but it seems unfortunate.

common/argparse.cpp
chandlerc: I'm skeptical about `ErrorOr` making this much better given that the error messages are always going to the stream, and we're likely to have the stream as a member anyways and want to create the error text here. I'd worry it'll just allocate lots of error strings just to stream them a bit later, but maybe there is a pattern of using `ErrorOr` that I'm not seeing here?

common/argparse.cpp
chandlerc: I find it more readable. I can remove it if others disagree though.

common/argparse.cpp
chandlerc: Sorry, I've picked up this habit from a different bit of coding that I've been doing. It's not important in this case, and if folks want, I'm happy to switch this PR back to `int`.

FWIW, I'm planning to suggest we change Carbon to have the default integer be `ssize_t` instead of `int` because of the other bit of coding -- the use of `int` is actually causing real performance costs in hot loops in several places where I've been doing optimization work. It has somewhat convinced me that if performance is the goal, `int` is a bad default integer, and `ssize_t` is a bunch better tool for the job. =/ 

But that's not relevant to *this* code as much, so if which integer type is used bothers you I can change. (My preference is to leave it just to avoid churn if i succeed in the argument above, but I don't feel strongly at all given how little it matters in this code.)

common/argparse.cpp
chandlerc: `SmallVector` is designed to be passed as `SmallVectorImpl` for pointer or reference parameters so that the small size isn't encoded. I try to do that pervasively.

common/argparse.cpp
chandlerc: Maybe I'm missing something, but I'm not seeing how to use this here...

I don't want to stream this to `out_` necessarily. So the unconditional flush of the buffer to the wrapped stream in `buffer_ostream` seems like it will do the wrong thing. The point is to buffer the output here to see if it stays under a specific size, and only in that case to write it. I don't see how to do that with `buffer_ostream`.

For the other use case, it also doesn't seem like it is right. There, I'm caching an escaped version of the value to print in two different places. I can just break apart the main stream though and call the weird `printEscapedString` API in the two places I need it.

common/argparse.cpp
chandlerc: Happy to add comments if useful, but if better names work, that seems even better.

I added some high level grouping comments. Is that enough? Or would a name change been helpful as well? If name change, I'd lean more toward something about the *section* and less about "all". WDYT?

common/argparse.h
chandlerc: (we also chatted a bit about this in person, but trying to relay the core of it here)

I'm happy to try to iterate on the DSL used to describe commands FWIW. But I did try something similar to what you outline, and wasn't able to arrive at a solution that really worked well. Not saying it's impossible, but the result seemed wasteful or difficult to read, or both (or didn't let me implement help .... there were a lot of experiments that I backed out).

The most pressing challenge I encountered for a method-setting API (as opposed to a fluent builder or a lambda builder) was similar to the problem with the fluent builder. I struggled to build a general way to nest and recurse to build subcommands in a way that was general and not baking in a specific subcommand structure. I think the subcommands are where the lambda building really shines. The other place where I find it nice is the `OneOf` because we actually again need to recurse a tiny bit to build the individual values.

Between the two valuable spots, it seemed marginally better to be consistent and get used to the lambda-builder pattern than to more pervasively switch back and forth between lambda-builder and some other pattern (whether fluent-builder or method-setting). But I can definitely explore having a bit more of a hybrid.

One option that @zygoloid had suggested earlier when chatting about this before I even had the PR ready was to specialize the case where we call exactly one method in the lambda-builder so we don't need a lambda at all there. I think I see an easy way to do that. WDYT? Also, would you like to see that in this PR or in a follow-up?

chandlerc (commented):
I think I've gotten to most of the comments here, but lemme know if anything missing. Still some interesting discussion therads I think.

common/ostream.h
jonmeow: Okay, but I'm still unclear what this does, I don't even see it in your argparse tests.

common/argparse.h
jonmeow: Before I read this code I was not familiar with this spelling of epilogue. The close link in documentation isn't valuable to me in part because of Carbon's behavior divergence -- even in review, I'm not referring to Python's argparse documentation, and divergent APIs would make that difficult regardless.

Maybe ask on Discord which spelling others would prefer?

common/argparse.h
jonmeow: We can iterate on the API.

common/ostream.h
jonmeow: I'll see separately if we can switch the other API so that both can change together.

common/argparse_test.cpp
jonmeow: Mainly I thought it might reduce the complexity. As noted elsewhere, I might want to iterate on the API so this seems fine for now.

jonmeow (approved):
As noted, we can move forward with the current approach although I may keep seeing if we can change the API a little.

chandlerc:
FWIW, the two API iterations that I'm planning as follow-ups based on review:

- Try out something to reduce the use of lambda-builders in common cases for arguments.
- Add unit tests for `ostream` to both make sure it's working well and help document what it's doing.

That sound right? Anything we should coordinate so we're not duplicating work there?

---

Last but not least, I'd like to find what approach we want around cases where it seems useful to have a common prefix on names.

As you rightly point out, we've oscillated a bit on this, and while I wrote this patch with one particular approach, I can't say it makes me deligted.

I see three things that I think I'd be happy with both here and elsewhere, but really interested in others' views on this:

1) Just add a common prefix to all the names. Simple, easy, etc.
2) Use a namespace, but disallow using declarations -- only code inside the namespace would use names w/o explicitly qualifying it. Would work a bit like `std` today.
3) Use a class to fake a namespace but avoid using declarations like this PR does, but only when all the names in question can reasonably live in a single header. Otherwise, do (1).

None of these really spark joy for me TBH. But I think I like (3) (the one in the current PR) the least of all. For this *specific* example, `Args` is an especially bad prefix to use with (1). And it *happens* to fit into a single header. But that doesn't seem like it scales well at all, so I feel like I really should change to either (1) or (2).

For (1), I'd love ideas for a better prefix. The best I've thought of is `CLI`, but it's very three-letter-acronym-y. =/

For (2), I'd love thoughts on having this rule as part of Carbon's style -- it would mimic the planned usage of namespaces in Carbon itself (outside of the nested regions). But not sure how others feel.

Unfortunately, *especially* in a compiler and language-y project, "args" really needs some disambiguating name component, so I think this PR needs some solution.

jonmeow:
> That sound right? Anything we should coordinate so we're not duplicating work there?

SGTM on the two

> Last but not least, I'd like to find what approach we want around cases where it seems useful to have a common prefix on names.

Talking on Discord

common/argparse.h
```diff
@@ -0,0 +1,893 @@
...
+// command line interfaces.
+//
+// ## Terminology used by this library
+//
+// _Argument_ or _arg_: One of the parsed components of the command line.
+//
+// _Option_: A _named_ argument starting with `--` to configure some aspect of
+// the tool. These often have both long spellings that start with `--` and
+// short, single character spellings that start with `-` and can be bundled with
+// other single character options.
```
zygoloid: A definition of _value_ would be useful somewhere around here.

common/argparse.h
```diff
@@ -0,0 +1,893 @@
...
+// syntax that overlaps with this library's parsed syntax.
+//
+// ## Options and flags
+//
+// The option syntax and behavior supported by this library is designed to be
+// strict and relatively simple while still supporting a wide range of expected
+// use cases:
+//
+// - All options must have a unique long name that is accessed with a `--`
+//   prefix. The name must consist of characters in the set [-a-zA-Z0-1], and it
```
zygoloid:
```suggestion
//   prefix. The name must consist of characters in the set [-a-zA-Z0-9], and it
```
I assume this is a typo?

common/argparse.h
```diff
@@ -0,0 +1,893 @@
...
+//   must not start with a `-` or `no-`.
+//
+// - Values are always attached using an `=` after the name. Only a few simple
+//   value formats are supported:
+//   - Arbitrary strings
+//   - Integers as parsed by `llvm::StringRef` and whose value fits in an
+//     `ssize_t`.
+//   - One of a fixed set of strings
+//
+// - Options may be parsed multiple times, and configure the behavior:
```
zygoloid:
```suggestion
// - Options may be parsed multiple times, and the behavior can be configured:
```

common/argparse.h
```diff
@@ -0,0 +1,893 @@
...
+//   - There is no distinction between the behavior of long and short names.
+//   - The short name can only specify the positive or `true` value for flags.
+//     There is no negative form of short names.
+//   - Short names are parsed after a single `-`, such as `-v`.
+//   - Any number of short names may be grouped after `-`, such as `-xyz`: this
+//     is three options, `x`, `y`, and `z`.
+//   - Short options may include a value after an `=`, but only the last short
+//     option of a group may receive a value. Other short options must have a
+//     default value that will be used. Typically this is used for a set of
+//     flags that are all set to `true`.
```
zygoloid: This conflicts with the POSIX conventions for command-line arguments, where `-a=b` would give `-a` the value `=b`, not `b`. (I think everything else here complies with the conventions where the syntaxes overlap.) It also diverges from the [behavior of python argparse](https://docs.python.org/3/library/argparse.html#option-value-syntax). That's not necessarily a problem, if we have good rationale for it. What's the motivation for this choice?

This choice to require `-a=b` rather than `-ab` or `-a b` also rules out syntaxes used very commonly with C++ compilers and linkers, such as `-Ipath`, `-lmylib`, `-Lpath`, `-o outfile`, `-x c++` (arguably `-O2`, `-Wwarning-flag`, and `-fflag-name` also follow the convention of a short option joined to its value, though that's a bit of a stretch).

common/argparse.h
```diff
@@ -0,0 +1,893 @@
...
+//
+// - Options may additionally have a short name of a single character [a-zA-Z].
+//   - There is no distinction between the behavior of long and short names.
+//   - The short name can only specify the positive or `true` value for flags.
+//     There is no negative form of short names.
+//   - Short names are parsed after a single `-`, such as `-v`.
+//   - Any number of short names may be grouped after `-`, such as `-xyz`: this
+//     is three options, `x`, `y`, and `z`.
+//   - Short options may include a value after an `=`, but only the last short
+//     option of a group may receive a value. Other short options must have a
```
zygoloid: Have you considered disallowing grouping of short options with arguments? I think it's pretty confusing to allow things like `-abc=123` as equivalent to `-a -b -c=123`.

common/argparse.h
```diff
@@ -0,0 +1,893 @@
...
+// Subcommands can be nested to any depth, and each subcommand gets its own
+// option space.
+//
+// ## Positional arguments
+//
+// Leaf commands (and subcommands) can accept positional arguments. These work
+// similar to options but consist *only* of the value. They have names, but the
+// name is only used in documentation and error messages. Positional argument
+// values cannot start with a `-` character until after option parsing is ended
+// with a `--` argument.
```
zygoloid: It's common to make a special exception for an argument that's exactly `-`, which is treated as a positional argument even before / without a `--`. I would personally appreciate being able to do `echo "some carbon code" | carbon build -` (or however we spell that).

common/argparse.h
```diff
@@ -0,0 +1,893 @@
...
+//
+// Multiple positional arguments to a single command are parsed in sequence. For
+// appending positional arguments, the sequence of values appended is ended with
+// a `--` argument. For example, a command that takes two sequences of
+// positional arguments might look like:
+//
+//     my-diff-tool a.txt b.txt c.txt -- new-a.txt new-b.txt new-c.txt
+//
+// The `--` used in this way *both* ends option parsing and the positional
+// argument sequence.
```
zygoloid: What happens if the `--` precedes all positional arguments? Does it still end the first positional argument sequence, or not? I think the answer I would prefer is that it does not end the first sequence unless we've already parsed at least one positional argument or `--`, so if you want the first argument sequence to be empty then you use `my-diff-tool -- -- second.txt file.txt`.

common/argparse.h
```diff
@@ -0,0 +1,893 @@
...
+// At many points in this library, a block of text is specified as a string.
+// These should be written using multi-line raw string literals that start on
+// their own line such as:
+//
+// ```cpp
+//     ...
+//     .help = R"""(
+// Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
+// tempor incididunt ut labore et dolore magna aliqua.
+// )""",
```
zygoloid: Huh. I've never seen triple-quotes used for raw string literals before (or more precisely, `""` used as the delimiter between the opening `R"` and the `(`), but... I like it, that's really neat -- it both lets you use `)"` in the string text and resembles Carbon and Python block strings, and it's much easier to see the delimiters than with just `R"( ... )"`.

Should we add this to the Carbon style guide as our preferred syntax for multi-line raw strings?

common/argparse.h
```diff
@@ -0,0 +1,893 @@
...
+    // and does not appear in the actual command line.
+    llvm::StringRef name;
+
+    // Optional short name for options. This must consist of a single character
+    // in the set [a-zA-Z].
+    llvm::StringRef short_name = "";
+
+    // For options, it is sometimes useful to render a distinct value
+    // placeholder name to help clarify what the value should contain. These are
+    // often in `All-CAPS` to make it clear they are placeholders. For example,
```
zygoloid:
```suggestion
    // often in `ALL-CAPS` to make it clear they are placeholders. For example,
```

common/argparse.h
```diff
@@ -0,0 +1,893 @@
...
+    // especially easy to type and reasonably easy to read.
+    //
+    // For positional arguments this is only used for help text, and is often
+    // spelled with `ALL-CAPS` to make it clear that it is a placeholder name
+    // and does not appear in the actual command line.
+    llvm::StringRef name;
+
+    // Optional short name for options. This must consist of a single character
+    // in the set [a-zA-Z].
+    llvm::StringRef short_name = "";
```
zygoloid: If we're going to be using aggregate initialization here, I wonder if `std::optional<char>` would be better? (`.short_name = 'x'` vs `.short_name = "x"`).

common/argparse.h
```diff
@@ -0,0 +1,893 @@
...
+    // can contain extra, custom guidance that is especially useful to have at
+    // the very end of the output.
+    llvm::StringRef help_epilog = "";
+  };
+
+  // Commands are classified based on the action they result in when run.
+  //
+  // Commands that require a subcommand have no action and are just a means to
+  // reach the subcommand actions. This also distinguishes meta actions from
+  // normal ones.
```
zygoloid:
```suggestion
  // reach the subcommand actions. Meta actions are also distinguished
  // normal ones.
```
"This" here has an ambiguous antecedent. To me, the original sentence reads as "Commands that require a subcommand have no action and are just a means to reach the subcommand actions. This means of reaching the subcommand actions also distinguishes meta actions from normal ones."

common/ostream.h
```diff
@@ -62,6 +62,35 @@ void PrintTo(T* p, std::ostream* out) {
   }
 }
 
+// True if T can be used with `<<` and an `llvm::raw_ostream`.
+template <typename T>
+static constexpr bool HasRawOstream = Requires<const T, llvm::raw_ostream>(
```
zygoloid: The name `HasRawOstream` isn't working well for me. How about something like `IsStreamableToRawOstream`? It's a bit long but seems hard to misinterpret.

common/ostream.h
```diff
@@ -62,6 +62,35 @@ void PrintTo(T* p, std::ostream* out) {
...
+// `std::ostream`. This ensures that an *explicit* operator will be preferred
+// when provided.
+template <typename S, typename T,
+          typename = std::enable_if_t<std::is_enum_v<T>>,
+          typename = std::enable_if_t<HasRawOstream<T>>,
+          typename = std::enable_if_t<std::is_base_of_v<
+              std::ostream, std::remove_reference_t<std::remove_cv_t<S>>>>,
+          typename = std::enable_if_t<!std::is_same_v<
+              std::remove_reference_t<std::remove_cv_t<T>>, llvm::raw_ostream>>>
+auto operator<<(S& standard_out, const T& value) -> S& {
```
zygoloid: It's pretty scary to be defining something that is (or at least appears to be) this broad, and to be defining what Carbon would consider to be an orphan. (I have similar concerns with the magic to stream anything with a `Print` method, for what it's worth.) The risk of an API hijack or ODR issues seems very real, and might occur in a client of the Carbon libraries rather than in our own code.

What stream types are we using this with? Can we use our own stream class instead of `std::ostream`, that picks up `operator<<`s defined for either `std::ostream` or for `llvm::raw_ostream`?

common/ostream.h
```diff
@@ -62,6 +62,35 @@ void PrintTo(T* p, std::ostream* out) {
...
+// template, and SFINAE disable itself unless that template parameter matches
+// `std::ostream`. This ensures that an *explicit* operator will be preferred
+// when provided.
+template <typename S, typename T,
+          typename = std::enable_if_t<std::is_enum_v<T>>,
+          typename = std::enable_if_t<HasRawOstream<T>>,
+          typename = std::enable_if_t<std::is_base_of_v<
+              std::ostream, std::remove_reference_t<std::remove_cv_t<S>>>>,
+          typename = std::enable_if_t<!std::is_same_v<
+              std::remove_reference_t<std::remove_cv_t<T>>, llvm::raw_ostream>>>
```
zygoloid: I don't understand this constraint: how could `T` be `llvm::raw_ostream` here? We already checked that `T` is both an enum type and is streamable to a `raw_ostream`, and `llvm::raw_ostream` is neither.

common/ostream.h
chandlerc: Sure!

common/ostream.h
chandlerc: The example FWIW is that without this, the `Eq(Args::ParseResult::Success)` matcher for example won't be able to print the value it fails to match or is trying to match as `ParseResult` is an enum with only an `operator<<` overload for `llvm::raw_ostream` and not for `std::ostream`.

The code will still compile, but debugging it is extra painful.

That said, this is clearly becoming a distraction from the rest of the PR. And I don't need to do more debugging right this moment, so I'm just going to remove the whole thing for now. Happy to discuss w/ you and @zygoloid how best to address this problem.

common/ostream.h
chandlerc: Yeah, you're right. I inherited this from the other injected overload where it could, and didn't remove it. Anyways, irrelevant as I'm just going to remove this to reduce distraction.

common/ostream.h
chandlerc: I don't disagree at all that it'd be great to have some other approach here, but that seems like it should be a separate change?

I see two places where these are useful:

1. Carbon code like `CHECK` and `VLOG`
2. GoogleTest code like `EXPECT_THAT(..., Eq(...))`

My understanding is that (1) needs `llvm::raw_ostream` and (2) needs `std::ostream`.

My goal is to be able to easily define a single output routine for a small enum without having to reach for the larger `EnumBase`. I'm wide open to whatever solutions let us do this best.

I've removed this for now, and can just live with difficult to debug test failure messages until we figure out something folks are happy with.

common/argparse.h
chandlerc: Good idea, done.

common/argparse.h
chandlerc: Do you mean grouping of short options with a *value*?

I'm not a huge fan either. I almost deleted the code and then didn't at the last minute. I'm super ambivalent here and happy to be persuaded away from it (or toward it).

Regarding grouping generally and the short argument syntax generally, see below.

common/argparse.h
chandlerc: The motivation was simplicity and explicitness of parsing, and the general readability of command lines.

I find the `-Ipath` and `-llib` flag syntaxes very hard to read and work with. They also are pretty hostile to key bindings to navigate command lines, etc. My impression is that this trend started with GCC and has been copied, but is almost entirely motivated by compatibility with GCC. I'm hoping we can avoid that specific requirement and have an easier to understand and more explicit command line syntax as a consequence.

Similarly, I find values as the next argument less easy to work with, and to add complexity with multiple ways to parse (or provide) values.

So I had wanted a much simpler rule of values are *always* provided by appending an `=` and the value. The current short behavior tries to be roughly similar to the POSIX guidelines with that constraint -- allowing a group of boolean flags (as those we can reasonably restrict to not take a value) and optionally a trailing one with an attached value. I at least find `-abc=123` -> `-a -b -c=123` easier to understand than `-abc123` -> `-a -b -c123` in the more POSIX style.

But FWIW, it's the bundle of non-value short flags that I find most helpful for making common patterns of commands easy to hammer out. So I'd also be game to restrict down to that as I mentioned above.

common/argparse.h
chandlerc: Good call, I hadn't even thought of that but agreed we should make that just work.

common/argparse.h
chandlerc: No, it only ends a positional argument sequence that has begun, so your example is correct. Added a mention of this.

common/argparse.h
chandlerc: I've ended up liking using it, so sure.

When indentation matters, I wasn't a fan at first because the un-indented text is a bit disruptive, I've not found any styles that *aren't* disruptive, and this one has grown on me.

common/argparse.h
chandlerc: I'm always nervous about using a `char` as a type in places like this because so many things can convert to a `char`. So in the past, whenever there won't be runtime overhead associated with it, I've tried to use more string-ish types in C++.

But I don't feel strongly about this if others prefer the other way around.

common/argparse.h
chandlerc: Yeah, original was confusing. I've gone with an even sharper distinction.

common/BUILD
chandlerc: Given the naming discussion and moving this to `CommandLine`, I'll rename the files and library to use `command_line` as the base instead of `argparse` before I land this. I'm not doing that just yet to keep comment threads in the code.

chandlerc (commented):
I think I've addressed many more of the comments now. A few things that I'm not sure have a comment thread specifically that I can find any more:

- It's using a namespace of `CommandLine`.
- Fewer small sizes, and those have comments.

toolchain/driver/driver.cpp
chandlerc: If it required lots of additional logic, I think I would agree it would best be omitted. But because it is easily supported, it seems better to make things somewhat well rounded here, and it seems useful to have a clear mapping from existing compiler behavior.

But regardless, this is really a discussion for the toolchain-side of this PR. Going to resolve this and other comment threads here, and once this lands and I update the toolchain PR, that's maybe a good place to continue any discussion here.

common/argparse.h
zygoloid: > Do you mean grouping of short options with a _value_?

Er, yes :)

> I'm not a huge fan either. I almost deleted the code and then didn't at the last minute. I'm super ambivalent here and happy to be persuaded away from it (or toward it).

How about we disallow it for now and see if anyone complains :)

common/argparse.h
zygoloid: Hm. I appreciate the self-consistency of this approach, and I don't think the POSIX incompatibility is necessarily a huge problem; there are plenty of programs that don't follow the POSIX conventions. If we wanted the Carbon driver to be a [POSIX-compatible `cc` program](https://pubs.opengroup.org/onlinepubs/7908799/xcu/cc.html) then the situation would be different, but I don't think that's a goal (and we can always ship a shell script for that purpose).

I think on reflection my only persistent concern is that the meaning of the options with values often isn't assignment of a value to some attribute, and using `=` for non-assignment cases can be confusing. Eg, while `-o=blah` or `--output-file=blah` as assignment to an output file name might be reasonable, `-l=lib1 -l=lib2` or `--link-library=lib1 --link-library=lib2` seems confusing to me compared to `-l lib1 -l lib2`, and similarly for cases like `-W=warning-name` versus `-W warning-name` and so on.

... but ... having the option and its value be part of a single argument is certainly valuable, eg when forwarding arguments from one tool to another. And consistently using `=` means it's easy to pick the positional arguments out of the command line.

My feeling is that I'd still prefer for both long and short options to permit both a joined-with-`=` syntax and a separated two-argument syntax, but I'm happy to run the experiment and see if that's a persistent concern or if that feeling goes away with usage of the always-`=` syntax.

So let's leave this as-is for now :)

common/argparse.h
```diff
@@ -0,0 +1,862 @@
...
+// other single character options.
+//
+// _Flag_: A boolean or binary option. These can only be enabled or disabled.
+//
+// _Positional argument_: An argument that is not named but is identified based
+// on the order in which it is encountered in the command line, with options
+// removed. Only a leaf command can contain positional arguments.
+//
+// _Value_: The value parameter provided to an argument. For options, this is
+// provided after an `=`s and the option name. For positional arguments, the
```
zygoloid:
```suggestion
// provided after an `=` and the option name. For positional arguments, the
```

common/argparse.h
zygoloid: I don't think my feelings are any stronger than yours, let's leave as-is unless someone else voices an opinion.

common/argparse.cpp
chandlerc: I've switched to `int` throughout.

common/argparse.h
chandlerc: My informal poll says to ignore argparse, so done.

common/argparse.h
chandlerc: (all of this is done now)

common/argparse.h
chandlerc: Done.

chandlerc (commented):
I think I've addressed the outstanding review comments, PTAL, and thanks again for all the review!

common/argparse.cpp
chandlerc: (I thought I had changed all of these, but now I have.)

common/argparse.h
```diff
@@ -209,506 +218,464 @@ namespace Carbon {
...
+  auto AddArgImpl(const ArgInfo& info, ArgKind kind) -> Arg&;
+  void AddPositionalArgImpl(const ArgInfo& info, ArgKind kind,
+                            llvm::function_ref<void(Arg&)> build);
+  void Finalize();
+
+  Command& command_;
+  MetaPrinter& meta_printer_;
+
+  llvm::SmallDenseSet<llvm::StringRef> arg_names;
+  llvm::SmallDenseSet<llvm::StringRef> subcommand_names;
```
jonmeow: Minor style, but looks like the suffix underscore is missing for these members?

common/argparse.cpp
```diff
@@ -664,24 +722,36 @@ class Args::Parser {
...
 
   const Command* command_;
 
   OptionMapT option_map_;
   ShortOptionTableT short_option_table_;
   SubcommandMapT subcommand_map_;
 
-  ssize_t positional_arg_index = -1;
+  int positional_arg_index = 0;
   bool appending_to_positional_arg = false;
```
jonmeow: Missing suffix underscore on members?

common/argparse.h
```diff
@@ -0,0 +1,861 @@
...
+  // Directly copy the value strings into a heap-allocated array in the
+  // argument.
+  new (&arg_.value_strings)
+      llvm::OwningArrayRef<llvm::StringRef>(value_strings);
+
+  // And build a type-erased action that maps a specific value string to a value
+  // by index.
+  new (&arg_.value_action) Arg::ValueActionT(
+      [values, match](const Arg& arg, llvm::StringRef value_string) -> bool {
+        for (int i = 0; i < static_cast<int>(N); ++i) {
```
jonmeow: Should this switch to llvm::seq for consistency with elsewhere that you've done that?

common/argparse.cpp
```diff
@@ -7,11 +7,59 @@
...
+    case ArgKind::String:
+      return output << "String";
+    case ArgKind::OneOf:
+      return output << "OneOf";
+    case ArgKind::MetaActionOnly:
+      return output << "MetaActionOnly";
+    case ArgKind::Invalid:
+      return output << "Invalid";
+  }
+  CARBON_FATAL() << "Corrupt argument kind!";
```
jonmeow: I'm a little confused, why these CHECKs but not for the enum on line 17-24? (I don't know that these checks are necessary, just commenting on inconsistency)

jonmeow (commented):
Just a few quick comments.

common/argparse.cpp
chandlerc: I just forgot above, added.

common/argparse.h
chandlerc: Good call, I had looked at the .cpp file to try to find these, but not the .h file, thanks!

common/argparse.h
chandlerc: Doh, yep. Fixed.

common/argparse.cpp
chandlerc: Fixed.

chandlerc (commented):
Updated, also with new namespace approach.

But don't auto-merge, I'll need to rename files.

jonmeow (approved):

jonmeow:
Just echoing that my approval still applies. It seems zygoloid didn't get around to another look, but I think you should probably just go ahead and merge when you're ready.

PR #2979, opened 2023-07-10
Switch the toolchain to the new CLI library. 

chandlerc:
This also tries to restructure the command line interface to the
toolchain a bit to make it start operating more like a compiler that
could be integrated into a build system rather than primarily as
a testing tool.

1) This switches form a `dump` subcommand to a `compile` subcommand
   which has "dump" actions that can be enabled within it.

2) A distinct set of compile _phases_ that match the toolchain
   structure:
   - `lex` to run the lexer
   - `parse` to run the parser
   - `check` to fully check that the code is valid
   - `lower` to lower to LLVM's IR
   - `codegen` to generate executable code

3) The codegen phase has two output formats: textual assembly and
   a binary object. These outputs can be configured, with a default for
   an object when writing to a file and more firm default for textual
   assembly when writing to stdout.

4) Select and expose the use of the LLVM host detection to compute
   a default code generation target in the driver so that the command
   line interface can reflect this. For example, the `help` output will
   include the default target.

5) The `//toolchain/codegen` library APIs have been restructured a bit
   to make the code flow a bit more naturally when implementing the new
   command line structure. No real changes to the logic though.

There are also some minor tweaks to the command line interface based on
trying to use the shortest names for things that still seem likely to be
learnable for users:

- Switched `target-triple` to just `target`: the "triple" component to
  this name is historical and can be confusing. For example, almost all
  "triple" strings have more than three components today.

- Switched to just `--output` as now the fact that it is a file can be
  configured in the documentation -- it will render as `--output=FILE`.

This also adds support for two custom output filename modes. First, when
no output is specified, we now compute one in the conventional way for
compilers by removing the file extension of the input file and replacing
it with `.o` for an object file output or `.s` for an assembly file
output. This matches the behavior of Clang and GCC for example.

Second, output to stdout is enabled with the special output file name of
`-` since it is no longer the default. This also follows the convention
of most compilers and many other command line tools to use `-` as a file
name to signify using standard in/out pipes.

There are still some rough edges here that I suspect could be improved,
but this seems like a good start of switching over to a complete
argument parser.


chandlerc:
Handy link to the changes in this PR separate from those in #2978: [`046fd9a` (#2979)](https://github.com/carbon-language/carbon-lang/pull/2979/commits/046fd9a44a21e9b54a8acb8790e61ed6c9fa08f0)

toolchain/driver/testdata/error_codegen_target_triple.carbon
```diff
@@ -2,7 +2,8 @@
 // Exceptions. See /LICENSE for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// RUN: %{not} %{carbon} dump objcode --target_triple=x86_64-unknown-linux-gnu %s | %{FileCheck-strict}
-// CHECK:STDERR: ERROR: Must provide an output file.
```
jonmeow: You're removing this test of a missing output argument, so how is it tested now?

toolchain/driver/testdata/error_codegen_target_triple.carbon
```diff
@@ -2,7 +2,8 @@
...
+// RUN: %{not} %{carbon} compile --output=- --target=x86_687-unknown-linux-gnu \
+// RUN:   %s | %{FileCheck-strict}
+// CHECK:STDERR: ERROR: {{.*}}
```
jonmeow: Is this supposed to be replacing the tests that were looking for "Invalid -target_triple"? I'm pretty sure the flag name was in the error in order to reduce the chance that unrelated errors would cause a bug to be missed; can you please re-add it?

toolchain/driver/driver.cpp
```diff
@@ -22,286 +25,481 @@
...
+so every phase before the one selected will also be run. The default is to
+compile, lower, and generate machine code.
+)""",
+        },
+        [&](auto& arg_b) {
+          arg_b.SetOneOf(
+              {
+                  arg_b.OneOfValue("tokenize", Phase::Tokenize),
+                  arg_b.OneOfValue("parse", Phase::Parse),
+                  arg_b.OneOfValue("syntax", Phase::Syntax),
```
jonmeow: Maybe these should match the directories more closely, or vice versa? e.g., "lex" for "lexer" and "semantics" for "semantic analysis", or the other direction, "tokenizer" for "tokenize" and "syntax" for "syntax".

For comparison, "parse"/"parser" and "lower"/"lowering" seem to have better correspondence. "codegen" is slightly out but maybe that's inevitable.

jonmeow:
> 2. A distinct set of compile _phases_ that match the toolchain
>    structure:
>    
>    * `tokenize` to run the lexer
>    * `parse` to run the parser
>    * `syntax` to fully syntax check with the semantics-IR
>    * `lower` to lower to LLVM's IR
>    * `codegen` to generate executable code

I'm concerned that having distinct phase names from both the directory names, class names, --dump flag names will make it more difficult for newcomers to figure out what the relationship between everything is. Essentially, it feels like everything has four different names that people need to learn if they want to get involved. Can we try to match names better?

For example, I thought "lex" and "semantic analysis" were typical terms, can you detail why you're using "tokenize" and "syntax" instead? Should the directories be renames instead after your preferred phase names?

With the "--dump" flags, maybe we can put those instead in the context of the phase, such as "--dump-llvm-ir" -> "--dump-lower", "--dump-lower-result", or "--dump-lower-ir"? It may also be good to take another look at the flag documentation to explain the association, as otherwise someone may think that "--dump-llvm-ir" might mean the compiler would stop there, without needing to specify --phase.

> * Switched to just `--output` as now the fact that it is a file can be
>   configured in the documentation -- it will render as `--output=FILE`.
> 
> This also adds support for two custom output filename modes. First, when no output is specified, we now compute one in the conventional way for compilers by removing the file extension of the input file and replacing it with `.o` for an object file output or `.s` for an assembly file output. This matches the behavior of Clang and GCC for example.

Where is this behavior tested? I'm looking at the driver and not validation that a default output file is created, only where an output file name is passed in.

toolchain/driver/testdata/success_codegen_objfile.carbon
```diff
@@ -2,7 +2,8 @@
 // Exceptions. See /LICENSE for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// RUN: %{carbon} dump objcode --target_triple=x86_64-unknown-linux-gnu --output_file=objfile.o %s | %{FileCheck-strict}
-// CHECK:STDOUT: Success: Object file is generated!
+// TODO: Currently just makes sure the command doesn't fail. We should add support for testing by disassembling the object file.
```
jonmeow:
```suggestion
// TODO: Currently just makes sure the command doesn't fail. We should add
// support for testing by disassembling the object file.
```

Sadly not automatic. Times like this, it's tempting to write a code formatter.

toolchain/driver/driver.h
```diff
@@ -37,38 +38,20 @@ class Driver {
...
 
  private:
-  auto ReportExtraArgs(llvm::StringRef subcommand_text,
-                       llvm::ArrayRef<llvm::StringRef> args) -> void;
+  struct Options;
+  struct CompileOptions;
+
+  auto ParseArgs(llvm::ArrayRef<llvm::StringRef> args, Options& options)
+      -> Args::ParseResult;
+  auto Compile(const CompileOptions& options) -> bool;
```
jonmeow: Function comments?

toolchain/driver/driver.cpp
```diff
@@ -22,286 +25,481 @@
...
+
+    b.AddStringOption(
+        {
+            .name = "output",
+            .value_name = "FILE",
+            .help = R"""(
+The output filename for codegen.
+
+When this is a file name, either textual assembly or a binary object will be
+written to it based on the flag `--asm-output`. The default is to write a binary
```
jonmeow, line 120: Old flag name? (I'm not finding it)

Maybe a good reason to have a TODO to not hardcode flag names, maybe using formatv or somesuch instead? But the StringRefs aren't amenable to that kind of approach, it'd lead to a use-after-free.

toolchain/driver/driver.cpp
```diff
@@ -22,286 +25,481 @@
...
+        [&](auto& arg_b) { arg_b.Set(&output_file_name); });
+
+    b.AddStringOption(
+        {
+            .name = "target",
+            .help = R"""(
+Select a target platform. Uses the LLVM target syntax, often called a "triple"
+(despite not actually being a triple in many common cases).
+
+This corresponds to the `target` flag to Clang and accepts the same strings
```
```suggestion
Select a target platform. This corresponds to the `target` flag to Clang and accepts the same strings
```

Seems to be saying the same thing both ways, I think the "triple" comments can be elided ("despite not actually being a triple" in particular feels like you're criticizing Clang, and won't help users).

toolchain/driver/driver.cpp
```diff
@@ -22,286 +25,481 @@
...
+  void Build(Args::CommandBuilder& b) {
+    b.AddFlag(
+        {
+            .name = "verbose",
+            .short_name = "v",
+            .help = "Enable verbose logging to the stderr stream.",
+        },
+        [&](auto& arg_b) { arg_b.Set(&verbose); });
+
+    b.AddSubcommand(CompileOptions::Info, [&](auto& sub_b) {
```
jonmeow: I feel odd about all your `auto` in lambdas. Why is `sub_bo.Do` a callback instead of directly set when this lambda is called? Might be clearer with an actual type.

toolchain/driver/driver.cpp
```diff
@@ -22,286 +25,481 @@
...
   bool has_errors = false;
 
+  using Phase = CompileOptions::Phase;
+  switch (options.phase) {
+    case Phase::Tokenize:
+      if (options.dump_parse_tree) {
+        error_stream_ << "ERROR: Requested dumping the parse tree but compile "
+                         "phase is limited to '"
+                      << options.phase << "'\n";
+        has_errors = true;
```
jonmeow: Should these exit more immediately, because they're user-caused flag errors versus compile-time errors?

That is, I'm fine if you want to collect all of them, but I'm not convince the system should even open the file when there are invalid flags.

Also, maybe this would be better done as part of post-flag parsing validation? In the Options `Build`, could that handle this and fail really early?

toolchain/driver/driver.cpp
```diff
@@ -22,286 +25,481 @@
...
+      if (!codegen->EmitAssembly(output_stream_)) {
+        return false;
+      }
+    }
+  } else {
+    llvm::SmallString<256> output_file_name = options.output_file_name;
+    if (output_file_name.empty()) {
+      output_file_name = options.input_file_name;
+      llvm::sys::path::replace_extension(output_file_name,
+                                         options.asm_output ? ".s" : ".o");
```
jonmeow, line 515: Maybe add a TODO that force_obj_output, asm_output, and output should be linked/conflicting flags?

toolchain/driver/driver.cpp
```diff
@@ -22,286 +25,481 @@
...
+
+    b.AddFlag(
+        {
+            .name = "force-obj-output",
+            .help = R"""(
+Force writing a binary object file, even when writing to stdout.
+
+This flag is only used when the code generation output file is set to stdout. In
+that case, it will override the forced default of textual assembly and output a
+binary output.
```
jonmeow: I'm a bit confused by this help structure. Why two paragraphs? Trying to think of other ways to phrase this, maybe:

```suggestion
When `--output=-` is set, the default is textual assembly; this forces printing of a binary object file instead. Ignored for other `--output` values.
```

The difference for me is that "even when writing to stdout" implies this does something when output is set to a file, and could even recognize /dev/stdout. But really it's specific to `--output=-`.

toolchain/driver/driver.cpp
```diff
@@ -22,286 +25,481 @@
...
+            .name = "output",
+            .value_name = "FILE",
+            .help = R"""(
+The output filename for codegen.
+
+When this is a file name, either textual assembly or a binary object will be
+written to it based on the flag `--asm-output`. The default is to write a binary
+object file.
+
+This can also be the string `-` which will write the output to stdout. In that
```
```suggestion
Passing `--output=-` will write the output to stdout. In that
```

Maybe a shorter/clearer way to say the same.

toolchain/driver/driver.cpp
```diff
@@ -22,286 +25,481 @@
...
+          arg_b.Set(&input_file_name);
+        });
+
+    b.AddOneOfOption(
+        {
+            .name = "phase",
+            .help = R"""(
+Selects the compilation phase to run. These phases are always run in sequence,
+so every phase before the one selected will also be run. The default is to
+compile, lower, and generate machine code.
```
jonmeow: The use of "compile" here is a little confusing to me since "lower" is a separate phase. Maybe say "compile to machine code."?

jonmeow (commented):
High level comment: this seems like a good start, although I think we'll probably want to keep looking at ways to simplify it. Part of what I'm looking at for that is all the bool AddFlags which seem fairly similar in structure, and feel like they could share more code. But I think getting something in would be helpful, with that kind of thing as more incremental change.

toolchain/driver/driver.cpp
```diff
@@ -22,286 +25,481 @@
...
+    if (output_file_name.empty()) {
+      output_file_name = options.input_file_name;
+      llvm::sys::path::replace_extension(output_file_name,
+                                         options.asm_output ? ".s" : ".o");
+    }
+    CARBON_VLOG() << "Writing output to: " << output_file_name << "\n";
+
     std::error_code ec;
-    llvm::raw_fd_ostream dest(output_file, ec, llvm::sys::fs::OF_None);
+    llvm::raw_fd_ostream output_file(options.output_file_name, ec,
```
jonmeow: Was this intended to be output_file_name? (no options.)

toolchain/semantics/testdata/basics/verbose.carbon
```diff
@@ -2,7 +2,7 @@
 // Exceptions. See /LICENSE for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// ARGS: -v dump raw-semantics-ir %s
+// ARGS: -v compile --phase=check --dump-raw-semantics-ir %s
```
jonmeow:
```suggestion
// ARGS: -v compile --phase=check %s
```

This is only looking at `-v` output, the `--dump` shouldn't be needed given `--phase`.

toolchain/codegen/codegen.h
```diff
@@ -14,41 +14,32 @@ namespace Carbon {
...
   // Returns false in case of failure, and any information about the failure is
   // printed to the error stream.
-  auto EmitCode(llvm::raw_pwrite_stream& dest,
-                llvm::TargetMachine* target_machine,
-                llvm::CodeGenFileType file_type) -> bool;
+  auto EmitCode(llvm::raw_pwrite_stream& out, llvm::CodeGenFileType file_type)
+      -> bool;
+
+  llvm::Module& module_;
+  llvm::raw_ostream& errors_;
```
jonmeow: Why are you switching errors from a raw_pwrite_stream to a raw_ostream? I understand you cannot change `out`, I'm just curious why you prefer to pass them differently.

toolchain/driver/testdata/fail_errors_streamed.carbon
```diff
@@ -2,9 +2,8 @@
 // Exceptions. See /LICENSE for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// ARGS: --print-errors=streamed dump tokens %s
+// ARGS: compile --phase=lex --stream-errors %s
```
jonmeow, line 5:
```suggestion
// ARGS: --stream-errors compile --phase=lex %s
```

Similar to the `-v` flag, should `--stream-errors` be independent of the `compile` subcommand? Suggestion to clarify what I mean here.

jonmeow (commented):
I was just glancing through again due to the latest changes, trying to figure out if this needed fresh review.

chandlerc:
> I was just glancing through again due to the latest changes, trying to figure out if this needed fresh review.

Not quite, but thanks for the comments! I should have this updated and rebased in a bit.

toolchain/driver/testdata/fail_errors_streamed.carbon
chandlerc: So, `-v` seems definitely ubiquitous, but not sure that diagnostics (and particularly the ordering aspect) is ...

Thinking about other possible tools we might want here, I think many of them wouldn't really have the same meaning of streaming vs. sorting diagnostics that compilation has, so it seemed like it was really a compile-focused thing. Happy to chat more about this, etc.

jonmeow: That's fine, we can always switch if things run differently.

toolchain/driver/driver.cpp
chandlerc: Happy to rephrase here to your suggestion or something else.

Just so we're on the same page, some context on what I was trying to accomplish here -- it may still not make sense though, so really am OK with the suggestion (but think this whole thread will vanish if I commit it).

I had wanted to write the docs for the driver more focused on LLVM than Clang given that it is somewhat the common denomenator -- whether C++, Rust, Swift, or some other language as background, it seems reasonable for developers to be thinking about the fact that they're passing essentially LLVM flags here. But maybe that's not the right approach, I can see the value of using Clang as the basis here, especially when we match it very closely. I mean, we're linking to Clang docs here, so yeah, maybe this idea doesn't pan out.

The discussion of triples isn't criticism of Clang really, the terminology (and IMO, the confusion) long predates Clang and comes from autotools AFAIK. Lots of documentation on these strings specifically calls them "triples" or "triplets" and I wanted to avoid folks thinking they had found the wrong docs because it used a term that didn't seem to apply. And that's extra true because I wanted to link them to the Clang docs that call them triples. Maybe there is a better way to achieve this though?

toolchain/driver/driver.cpp
chandlerc: The two paragraphs thing is because I'm hoping to add a "short" help mode, and I had thought it might be nice to have the default short help text to be the first paragraph of the main help. My hope was to avoid having to basically repeat the same first sentence in both the short help string and the full help string this way. It's a similar pattern to what Git uses to extract the short summary of the commit log. Does that seem reasonable?

I agree with your comment about the wording though -- adapting the two-paragraph form based on your suggestion, WDYT?

toolchain/driver/driver.cpp
chandlerc: Good suggestion, done (as much as I can). Pulled this into a validation function that we can add other early errors to.

Regarding post-parsing validation -- I'd love to add to the command line library a nice way to do more custom validation like this. Just hadn't thought of a good way to formulate that quite yet. If you have specific ideas, lemme know!

toolchain/driver/driver.cpp
chandlerc: Good idea, added TODO, and will add validation in a follow-up PR.

toolchain/codegen/codegen.h
chandlerc: I was generally trying to use the least constraining interface -- the error stream doesn't need to be `pwrite`, and so it seemed better to not require that more specific type.

Does that seem reasonable? Would commenting maybe on the place where we do need to use `pwrite` to explain that difference help?

That said, I don't feel super strongly here, so happy to change if the above doesn't make sense?

toolchain/driver/driver.cpp
chandlerc: Yes, good catch and fixed.

toolchain/driver/driver.cpp
chandlerc: Good suggestion, done.

toolchain/driver/driver.cpp
chandlerc: I think it's below?

I have some ideas for rearranging things to be able to not hard-code the names, left a TODO to try that.

toolchain/driver/driver.cpp
chandlerc: I've replaced this `auto` with a type. Let me know if it helps?

toolchain/driver/driver.h
chandlerc: Done.

toolchain/driver/testdata/success_codegen_objfile.carbon
chandlerc: This is now tested in the driver's unittests.

toolchain/driver/driver.cpp
chandlerc: All of these renamed according to #3070.

toolchain/driver/testdata/error_codegen_target_triple.carbon
chandlerc: This is now covered in the unittests.

toolchain/driver/testdata/error_codegen_target_triple.carbon
chandlerc: I think this is now covered with a pretty precise check in the codegen tests, but let me know if not.

chandlerc (commented):
Ok, PTAL -- I think I've responded to all the comments, but there was a decent amount of churn from getting the other PR landed and so let me know if you spot anything I've missed.

toolchain/codegen/codegen.h
jonmeow: Given the skew, it'd probably be good to add pwrite. While passing pwrite streams seemed okay because it was consistent (i.e., less likely to think about things passed as the same type), the difference seems like it'd be deserving of documentation.

toolchain/driver/driver.cpp
jonmeow:
```suggestion
Select a target platform. Uses the LLVM target syntax. Also known as a "triple" for historical reasons.

This corresponds to the `target` flag to Clang and accepts the same strings
```


Here's a different suggestion. Versus "despite not actually being a triple in many common cases", "for historical reasons" still implies it's not always a triple, and I think phrasing like that is easier to have interpreted as objective and informative.

Note, if it helps understand how I'm reading this, I think the use of "actually" probably has the strongest influence on my interpretation right now, partly because it's only there for emphasis.

toolchain/driver/driver.cpp
jonmeow: It does help me here, but I'd also note this on line 301, line 321, etc. These are cases where I think understanding `auto` and what's happening require sufficient knowledge to understand what type will be used there; I think my unfamiliarity with CommandLine prevents me from inferring types, and "b" is also a difficult abbreviation to unpack.

toolchain/driver/driver_test.cpp
```diff
@@ -17,74 +23,108 @@
...
+  auto CreateTestFile(llvm::StringRef text,
+                      llvm::StringRef file_name = "test_file.carbon")
+      -> llvm::StringRef {
+    fs_.addFile(file_name, /*ModificationTime=*/0,
                 llvm::MemoryBuffer::getMemBuffer(text));
-    return TestFileName;
+    return file_name;
+  }
+
+  auto ScopedTempWorkingDir() {
```
jonmeow, line 60:
```suggestion
  // Makes a temp dir and sets the working directory. Returns a scope_exit that will delete it and restore the working directory.
  auto ScopedTempWorkingDir() {
```

I admit I'm a little uncomfortable with the auto return, although I assume it's because scope_exit is in llvm::detail. So I'd mainly suggest commenting to highlight the behavior.

jonmeow (approved):
Thanks! A few small things remain, but approving so that you can merge after you've had a chance to go through them. I think this is looking really good now.

toolchain/driver/driver_test.cpp
chandlerc: In addition to being in `detail`, I think it is also a template type that is hard to name. But definitely a good idea to comment this to clarify. I'll edit / enhance the comment a bit to have even more detail I think.

toolchain/codegen/codegen.h
chandlerc: Switched back to `pwrite` -- that seemed the best option for *this* patch, but wasn't sure if that was actually what you were suggesting or if it's the right approach long term.

We use just `raw_ostream` so prevalently for stderr-y things, that it seems especially surprising to me to make it use `pwrite` here. My preference would be to use `pwrite` as narrowly as possible and comment why it is needed there. If that makes sense, I'm happy to do that as a follow-up.

toolchain/driver/driver.cpp
chandlerc: Done for the lines mentioned.

Happy to switch others, although I worry about the loss of brevity for most of the argument cases. That said, maybe it makes sense to try some of the argument API improvements first and then see if that or a better name or some other change make the most sense.

PR #2982, opened 2023-07-12
Trace stack and memory only when there is change in state

prabhatexit0:
Add state (stack and heap information) to the trace stream only when there is a change in state, this would prevent consecutive repetition of state information in the trace output.

Checks it based on change in size of the stack and heap, added `std::pair<int, int> stack_heap_sizes` in `Interpreter` to store size of stack and heap for comparison.


explorer/interpreter/interpreter.cpp
```diff
@@ -204,7 +208,11 @@ class Interpreter {
 //
 
 void Interpreter::TraceState() {
-  *trace_stream_ << "{\nstack: " << todo_ << "\nmemory: " << heap_ << "\n}\n";
+  if (stack_heap_sizes_.first != todo_.size() ||
+      stack_heap_sizes_.second != heap_.size()) {
```
zygoloid: An action might pop one state and then push another one, which would keep the size the same. Similarly, an action might write to a heap entry without changing the size of the heap. I think some more fine-grained checks would work better here -- let's trace each time the heap or stack changes, when it changes.

How about this:

- For the stack: add `ActionStack::Push` and `ActionStack::Pop` functions, and change all the calls to `todo_.Push` and `todo_.Pop` to call those instead. Trace the new stack after each change to it.
- For the heap: add tracing to `Heap::AllocateValue`, `Heap::Write`, and `Heap::Deallocaate`.

Then remove `Interpreter::TraceState` and all calls to it, as we'll already be tracing the heap and stack each time they change.

prabhatexit0: Thanks for this suggestion. I would implement it this way in the next commit.

prabhatexit0: Thanks for the suggestion! Will do the suggested changes in the next commit.

explorer/interpreter/action_stack.h
```diff
@@ -104,7 +109,19 @@ class ActionStack {
   auto UnwindPast(Nonnull<const Statement*> ast_node,
                   Nonnull<const Value*> result) -> ErrorOr<Success>;
 
-  void Pop() { todo_.Pop(); }
+  void Pop() {
+    todo_.Pop();
+    if (trace_stream_->is_enabled()) {
+      *trace_stream_ << "+++ stack: [ " << *this << " ]\n";
+    }
+  }
```
zygoloid: It looks like this function isn't being called -- in action_stack.cpp, `todo_.Pop()` is called directly instead of using this function.

zygoloid (commented):
Nice. Only one concern: it looks like we're only tracing stack pushes, not pops, right now.

As a follow-up, I think it'd be interesting to consider tracing only the thing that changed, rather than the entire stack / heap:

```
+++ push i32 .0.
+++ push i32 .1.
+++ pop i32 .1.
+++ pop i32 .0.
```
rather than
```
+++ stack: [i32 .0. ]
+++ stack: [i32 .0.  ## i32 .1. ]
+++ stack: [i32 .0. ]
+++ stack: []
```

prabhatexit0:
> Nice. Only one concern: it looks like we're only tracing stack pushes, not pops, right now.

Oh I missed that, fixing it and pushing changes in the next commit.

> 
> As a follow-up, I think it'd be interesting to consider tracing only the thing that changed, rather than the entire stack / heap:
> 
> ```
> +++ push i32 .0.
> +++ push i32 .1.
> +++ pop i32 .1.
> +++ pop i32 .0.
> ```
> 
> rather than
> 
> ```
> +++ stack: [i32 .0. ]
> +++ stack: [i32 .0.  ## i32 .1. ]
> +++ stack: [i32 .0. ]
> +++ stack: []
> ```

This seems like a really nice suggestion. Thanks! but tracing the whole heap shows which values are uninitialized or dead, should we trace the whole heap and just trace the push / pop for stack?

zygoloid:
> tracing the whole heap shows which values are uninitialized or dead, should we trace the whole heap and just trace the push / pop for stack?

Maybe for the heap tracing we can trace whenever values change state (eg, when they become initialized or dead) as well as when their values change? I think the transition to "dead" happens when the allocation is deallocated, so we don't need to trace that separately from the deallocation.

As an example, we could trace the operations somewhat like this (for `AllocationId(5)`):

```
+++ alloc #5 = uninitialized
+++ write #5 = 42
+++ read #5 = 42
+++ dealloc #5
```

though please feel free to come up with something else!

prabhatexit0:
> > tracing the whole heap shows which values are uninitialized or dead, should we trace the whole heap and just trace the push / pop for stack?
> 
> Maybe for the heap tracing we can trace whenever values change state (eg, when they become initialized or dead) as well as when their values change? I think the transition to "dead" happens when the allocation is deallocated, so we don't need to trace that separately from the deallocation.
> 
> As an example, we could trace the operations somewhat like this (for `AllocationId(5)`):
> 
> ```
> +++ alloc #5 = uninitialized
> +++ write #5 = 42
> +++ read #5 = 42
> +++ dealloc #5
> ```
> 
> though please feel free to come up with something else!

I really like this suggestion, it made the memory related trace more intuitive for me. Thanks!
also, I made some additional changes, mainly I prefixed `push` and `pop` with `stack-` and `alloc`, `write`, `read` and `dealloc` with `memory-`.
also I used `(+)` symbol before **push** and **alloc** and `(-)` symbol before **pop** and **dealloc** to make it more easily readable.
Please let me know what you think about these changes.
also I updated the `full_trace.carbon` test so that memory related trace is also added in the full trace.

zygoloid (approved):

PR #2985, opened 2023-07-12
Start node stack push/pop setting IdT based on ParseNodeKind.

jonmeow:
I think there's more we can do here, but this seemed like a good checkpoint to make sure the path I'm going down is roughly what you expected. There's one actual edit in if expression structure to match the increased enforcement.


zygoloid (approved):
I think this is a good direction.

PR #2986, opened 2023-07-13
Add name resolution information to trace stream

prabhatexit0:
Adds the following information about name resolution to the trace output:
* Name resolution process
```
** resolving stmt | decl `<stmt | decl>` (<source_location>)
   ...
** finished resolving stmt | decl `<stmt | decl>` (<source_location>)
```
* When a name is added to the static scope
```
--- declared `<name>` as `<entity>` in `<scope>` (<source_location>)
```
* When a name is resolved
```
--- resolved `<name>` as `<result>` in `<scope>` (<source_location>)
```
* Marking a name declared/usable
```
--- marked `<name>` <usable | declared but not usable> in `<scope>`
```


explorer/ast/static_scope.cpp
```diff
@@ -67,6 +79,14 @@ auto StaticScope::ResolveHere(std::optional<ValueNodeView> this_scope,
       return ProgramError(source_loc)
              << "name '" << name << "' has not been declared in this scope";
     }
+  } else {
+    if (this_scope) {
+      if (trace_stream_->is_enabled()) {
+        *trace_stream_ << "--- declared `" << name << "` in "
+                       << PrintAsID(this_scope->base()) << "` ("
+                       << this_scope->base().source_loc() << ")\n";
```
zygoloid: This function is used when looking / resolving up an already-declared name, not when declaring a new name. So I think we should produce a message that says something more like this:

```
--- resolved `foo` at myfile.carbon:5 in `fn F` as `var Foo: i32`
```

explorer/ast/static_scope.cpp
```diff
@@ -23,6 +23,12 @@ auto StaticScope::Add(std::string_view name, ValueNodeView entity,
     if (static_cast<int>(status) > static_cast<int>(it->second.status)) {
       it->second.status = status;
     }
+  } else {
+    if (trace_stream_->is_enabled()) {
+      *trace_stream_ << "--- declared `" << name << "` in `"
+                     << PrintAsID(entity.base()) << "` ("
```
zygoloid: `entity` here is the entity that we're binding the name to, not the name of the scope, so "in `entity`" isn't really the right thing to be saying here.

I think it would be useful to add an `AstNode*` member to `StaticScope` to track the AST node whose scope we're referring to, so we can print that information out here. Then perhaps we could report something like:

```
--- declared `foo` as `fn N.Foo` in `namespace N` (foo.carbon:4)
```

explorer/ast/static_scope.cpp
```diff
@@ -32,13 +38,19 @@ void StaticScope::MarkDeclared(std::string_view name) {
   CARBON_CHECK(it != declared_names_.end()) << name << " not found";
   if (it->second.status == NameStatus::KnownButNotDeclared) {
     it->second.status = NameStatus::DeclaredButNotUsable;
+    if (trace_stream_->is_enabled()) {
+      *trace_stream_ << "--- marked `" << name << "` declared but not usable\n";
```
zygoloid: For this and the trace below, it'd be great to also include the name of the scope, once we start tracking that.

explorer/interpreter/resolve_names.cpp
```diff
@@ -148,6 +153,11 @@ auto NameResolver::ResolveQualifier(DeclaredName name,
       return ProgramError(name.source_loc())
              << PrintAsID(node.base()) << " cannot be used as a name qualifier";
     }
+
+    if (trace_stream_->is_enabled()) {
+      *trace_stream_ << "--- resolved qualifier `" << qualifier << "` for "
+                     << "\n";
```
zygoloid: This message seems to be missing the name of the entity whose qualifier we resolved. Though perhaps we don't need this trace message at all with the tweak to `Resolve`'s output I suggested above.

explorer/interpreter/resolve_names.cpp
```diff
@@ -176,6 +187,11 @@ auto NameResolver::AddExposedNames(const Declaration& declaration,
                                    StaticScope& enclosing_scope,
                                    bool allow_qualified_names)
     -> ErrorOr<Success> {
+  if (trace_stream_->is_enabled()) {
+    *trace_stream_ << "--- add exposed name `";
+    declaration.PrintID(trace_stream_->stream());
+    *trace_stream_ << "` (" << declaration.source_loc() << ")\n";
+  }
```
zygoloid: I think we can skip this trace message. The message we get from actually introducing the name seems to be sufficient.

explorer/interpreter/resolve_names.cpp
```diff
@@ -277,13 +293,20 @@ auto NameResolver::AddExposedNames(const Declaration& declaration,
 auto NameResolver::ResolveNames(Expression& expression,
                                 const StaticScope& enclosing_scope)
     -> ErrorOr<std::optional<ValueNodeView>> {
+  if (trace_stream_->is_enabled()) {
+  }
```
zygoloid: Looks like this is left behind after some edits?

explorer/interpreter/resolve_names.cpp
```diff
@@ -277,13 +293,20 @@ auto NameResolver::AddExposedNames(const Declaration& declaration,
...
 }
 
 auto NameResolver::ResolveNamesImpl(Expression& expression,
                                     const StaticScope& enclosing_scope)
     -> ErrorOr<std::optional<ValueNodeView>> {
+  if (trace_stream_->is_enabled()) {
+    *trace_stream_ << "** resolving expr `";
+    expression.Print(trace_stream_->stream());
+    *trace_stream_ << "` (" << expression.source_loc() << ")\n";
+  }
```
zygoloid: Tracing every expression we enter and leave seems like it's going to be very verbose. Perhaps we can just trace statements and declarations, and not expressions and where clauses and patterns and the like?

explorer/trace_testdata/full_trace.carbon
```diff
@@ -13,6 +13,42 @@
...
+
+
+package ExplorerTest api;
+
+// CHECK:STDOUT: --- add exposed name `interface TestInterface` (full_trace.carbon:[[@LINE+2]])
+// CHECK:STDOUT: --- declared `TestInterface` in `interface TestInterface` (full_trace.carbon:[[@LINE+1]])
+interface TestInterface {}
+
+fn Main() -> i32 {
+  return 0;
```
zygoloid: This test doesn't actually resolve any names, so you're not getting good coverage of your name resolution additions here. Maybe we could add a little bit more code to this file so that we can see name resolution happening:

```suggestion
package ExplorerTest api;

// CHECK:STDOUT: --- add exposed name `interface TestInterface` (full_trace.carbon:[[@LINE+2]])
// CHECK:STDOUT: --- declared `TestInterface` in `interface TestInterface` (full_trace.carbon:[[@LINE+1]])
interface TestInterface {}

fn F(n: i32) -> i32 {
  return n + 1;
}

fn Main() -> i32 {
  return F(0);
```

... should do some semi-interesting name resolution.

zygoloid (commented):
This looks like a good start.

explorer/ast/static_scope.cpp
prabhatexit0: Okay, thanks for letting me know about that. Can we put the source location at end of line like this?
```
--- resolved `foo` in `fn F` as `var Foo: i32` at (.../myfile.carbon:5)
```
I was thinking of keeping source location at the end of line for consistency.

explorer/ast/static_scope.cpp
prabhatexit0: I'm sorry I missed that, thanks for mentioning it. 
I have one doubt, what kind of `AstNode*` should I use for **file scope**?

explorer/interpreter/resolve_names.cpp
prabhatexit0: Yes, I'll remove this in next commit.

explorer/interpreter/resolve_names.cpp
prabhatexit0: Sorry about that, will remove in the next commit.

explorer/interpreter/resolve_names.cpp
prabhatexit0: Yes, it is getting verbose when tracing enter/leave for every expression, where clause and pattern.
Tracing only statements and declarations looks like a better choice to me as well.

explorer/interpreter/resolve_names.cpp
```diff
@@ -568,6 +577,11 @@ auto NameResolver::ResolveNames(Statement& statement,
 auto NameResolver::ResolveNamesImpl(Statement& statement,
                                     StaticScope& enclosing_scope)
     -> ErrorOr<Success> {
+  if (trace_stream_->is_enabled()) {
+    *trace_stream_ << "** resolving stmt `";
+    statement.PrintID(trace_stream_->stream());
+    *trace_stream_ << "` (" << statement.source_loc() << ")\n";
```
zygoloid:
```suggestion
    *trace_stream_ << "** resolving stmt `"
                   << PrintAsID(statement)
                   << "` (" << statement.source_loc() << ")\n";
```
I think you can write this using `PrintAsID`.

explorer/trace_testdata/full_trace.carbon
```diff
@@ -6,13 +6,65 @@
...
+// CHECK:STDOUT: --- declared `TestInterface` as `interface TestInterface` in `root` (full_trace.carbon:[[@LINE+1]])
+interface TestInterface {}
+
+// CHECK:STDOUT: --- declared `N` as `namespace N` in `root` (full_trace.carbon:[[@LINE+1]])
+namespace N;
+
+// CHECK:STDOUT: --- resolved `N` as `namespace N` in `root` (full_trace.carbon:[[@LINE+1]])
+fn N.Foo(n: i32) -> i32 {
+  return n + 1;
+// CHECK:STDOUT: --- declared `Foo` as `fn N.Foo` in `root` (full_trace.carbon:[[@LINE+1]])
```
zygoloid: This should be in namespace `N`, but it says `root` here. I think the problem is that we're creating a `StaticScope` without a value node for namespaces here: https://github.com/carbon-language/carbon-lang/blob/3feb40fc5148012f44009b336f2823cf35719a69/explorer/interpreter/resolve_names.cpp#L192

... which I think should be

```c++
        namespace_scopes_.try_emplace(&namespace_decl, scope, &namespace_decl); 
```

explorer/ast/static_scope.cpp
```diff
@@ -23,22 +24,53 @@ auto StaticScope::Add(std::string_view name, ValueNodeView entity,
...
+    }
   }
   return Success();
 }
 
+void StaticScope::Print(llvm::raw_ostream& out) const {
+  if (ast_node_) {
+    ast_node_.value()->Print(out);
+  } else {
+    *trace_stream_ << "root";
```
zygoloid: The top level is package-scope, so maybe we can say "package" here?

explorer/trace_testdata/full_trace.carbon
```diff
@@ -6,13 +6,65 @@
...
+// CHECK:STDOUT: --- marked `N` usable in `root`
+// CHECK:STDOUT: ** finished resolving decl `namespace N` (full_trace.carbon:[[@LINE-18]])
+// CHECK:STDOUT: ** resolving decl `fn N.Foo` (full_trace.carbon:[[@LINE-13]])
+// CHECK:STDOUT: --- resolved `N` as `namespace N` in `root` (full_trace.carbon:[[@LINE-17]])
+// CHECK:STDOUT: --- marked `Foo` declared but not usable in `root`
+// CHECK:STDOUT: --- declared `n` as `n` in `fn N.Foo` (full_trace.carbon:[[@LINE-19]])
+// CHECK:STDOUT: --- marked `Foo` usable in `root`
+// CHECK:STDOUT: ** resolving stmt `{return (n + 1);}` (full_trace.carbon:[[@LINE-18]])
+// CHECK:STDOUT: ** resolving stmt `return (n + 1);` (full_trace.carbon:[[@LINE-21]])
+// CHECK:STDOUT: ** finished resolving stmt `return (n + 1);` (full_trace.carbon:[[@LINE-22]])
```
zygoloid: Between these two lines I was expecting to see
```
--- resolved `n` as `n: i32` in `fn N.Foo` (...)
```
or something like that. (More in the next comment.)

explorer/trace_testdata/full_trace.carbon
```diff
@@ -6,13 +6,65 @@
...
+// CHECK:STDOUT: ** finished resolving stmt `{return (n + 1);}` (full_trace.carbon:[[@LINE-21]])
+// CHECK:STDOUT: ** finished resolving decl `fn N.Foo` (full_trace.carbon:[[@LINE-22]])
+// CHECK:STDOUT: ** resolving decl `fn Main` (full_trace.carbon:[[@LINE+338]])
+// CHECK:STDOUT: --- marked `Main` declared but not usable in `root`
+// CHECK:STDOUT: --- marked `Main` usable in `root`
+// CHECK:STDOUT: ** resolving stmt `{return N.Foo(0);}` (full_trace.carbon:[[@LINE+335]])
+// CHECK:STDOUT: ** resolving stmt `return N.Foo(0);` (full_trace.carbon:[[@LINE-24]])
+// CHECK:STDOUT: --- resolved `Foo` as `fn N.Foo` in `root` (full_trace.carbon:[[@LINE-25]])
```
zygoloid: Similarly, I was expecting to see this here:
```
--- resolved `N` as `namespace N` in `root` (...)
```

Oh, I see what's wrong: `Resolve` calls `TryResolve`, which calls `TryResolveHere`, which doesn't do any tracing -- we never reach the trace line in `ResolveHere` when starting from `Resolve`. Perhaps you could move the trace output from `ResolveHere` to `TryResolveHere`.

explorer/ast/static_scope.cpp
prabhatexit0: Yes, that would be better.

explorer/trace_testdata/full_trace.carbon
prabhatexit0: Thanks for pointing this out!

explorer/ast/static_scope.h
```diff
@@ -35,11 +36,18 @@ class StaticScope {
...
-  StaticScope() = default;
+  explicit StaticScope(Nonnull<TraceStream*> trace_stream)
+      : ast_node_(std::nullopt), trace_stream_(trace_stream){};
 
   // Construct a scope that is nested within the given scope.
-  explicit StaticScope(Nonnull<const StaticScope*> parent)
-      : parent_scope_(parent) {}
+  explicit StaticScope(
+      Nonnull<const StaticScope*> parent,
+      std::optional<Nonnull<const AstNode*>> ast_node = std::nullopt)
```
zygoloid: Can we make the `ast_node` parameter mandatory?

explorer/ast/static_scope.h
```diff
@@ -35,11 +36,18 @@ class StaticScope {
   };
 
   // Construct a root scope.
-  StaticScope() = default;
+  explicit StaticScope(Nonnull<TraceStream*> trace_stream)
+      : ast_node_(std::nullopt), trace_stream_(trace_stream){};
```
zygoloid:
```suggestion
      : ast_node_(std::nullopt), trace_stream_(trace_stream) {}
```
I think the trailing semicolon is confusing clang-format here.

explorer/interpreter/resolve_names.cpp
```diff
@@ -166,7 +171,8 @@ auto NameResolver::AddExposedName(DeclaredName name, ValueNodeView value,
   // known but not declared namespace name.
   CARBON_ASSIGN_OR_RETURN(
       Nonnull<StaticScope*> scope,
-      ResolveQualifier(name, enclosing_scope, /*allow_undeclared=*/true));
+      ResolveQualifier(name, enclosing_scope,
+                       /*allow_undeclared=*/true));  // Qualifier Resolution
```
zygoloid: I don't think this comment is adding anything, can you remove it?

zygoloid (commented):
Thanks, only very minor comments left.

explorer/interpreter/resolve_names.cpp
zygoloid: There are a few more `PrintID` calls nearby that you can change in the same way.

explorer/ast/static_scope.h
prabhatexit0: Yes we can do that. I'll do it in the next commit.

zygoloid (approved):

PR #2988, opened 2023-07-14
Add a name to glob_lit_tests

rupprecht:
Adding `name` to `glob_lit_tests` will make it conform with other implementations of `glob_lit_tests` out there. If someone uses this repo while providing a different version of `glob_lit_tests` that requires a name, those build rules will become invalid.


jonmeow (approved):
Thank you!

PR #2992, opened 2023-07-17
Integrates tuple to semantic analysis

fasiddique:
Handles tuples (including nested tuples) in the semantic phase of the tool chain. Does not handle tuple element access yet.


toolchain/lowering/lowering_handle.cpp
```diff
@@ -215,6 +215,18 @@ auto LoweringHandleStructType(LoweringFunctionContext& /*context*/,
   // No action to take.
 }
 
+auto LoweringHandleTupleValue(LoweringFunctionContext& /*context*/,
+                              SemanticsNodeId /*node_id*/,
+                              SemanticsNode /*node*/) -> void {
+  // No Action to take
```
jonmeow: Are you sure tuple values don't need to be lowered? Maybe this should be a TODO instead?

toolchain/lowering/lowering_handle.cpp
```diff
@@ -215,6 +215,18 @@ auto LoweringHandleStructType(LoweringFunctionContext& /*context*/,
...
+}
+
+auto LoweringHandleTupleType(LoweringFunctionContext& /*context*/,
+                             SemanticsNodeId /*node_id*/,
+                             SemanticsNode /*node*/) -> void {
+  // No action to take
```
```suggestion
  // No action to take.
```

Please use sentence grammar and punctuation in comments. Note this also matches other similar comments.

toolchain/parser/parse_node_kind.def
```diff
@@ -208,12 +208,13 @@ CARBON_PARSE_NODE_KIND_CHILD_COUNT(WhileConditionStart, 0)
...
 
-// Parenthesized expressions:
+// Parenthesized expressions, such as `(2)`:
 //   ParenExpressionOrTupleLiteralStart
 //   _external_: expression
 // ParenExpression
 
-// Tuples:
+
+// Tuples, such as `(i32,) = (1,)`:
```
jonmeow:
```suggestion
// Tuples, such as `(1, 2)`:
```

The assignment operation isn't a tuple, so let's trim this example and make it a little more plain.

toolchain/parser/parse_node_kind.def
```diff
@@ -208,12 +208,13 @@ CARBON_PARSE_NODE_KIND_CHILD_COUNT(WhileConditionStart, 0)
...
 CARBON_PARSE_NODE_KIND_CHILD_COUNT(WhileStatement, 2)
 
-// Parenthesized expressions:
+// Parenthesized expressions, such as `(2)`:
 //   ParenExpressionOrTupleLiteralStart
 //   _external_: expression
 // ParenExpression
 
-// Tuples:
+
```
jonmeow: Please use an empty comment line (`//`) to connect these, since they both refer to the paren syntax. Adding more blank lines is undesirable.

toolchain/semantics/semantics_context.cpp
```diff
@@ -522,6 +540,7 @@ auto SemanticsContext::ImplicitAsImpl(SemanticsNodeId value_id,
   if (output_value_id != nullptr) {
     *output_value_id = SemanticsNodeId::BuiltinError;
   }
+
```
jonmeow:
```suggestion
```

You're adding a blank line here, but Incompatible is the matching return value for the above assignment and should not be separated.

toolchain/semantics/semantics_context.cpp
```diff
@@ -502,14 +504,30 @@ auto SemanticsContext::ImplicitAsImpl(SemanticsNodeId value_id,
...
+      if (output_value_id != nullptr) {
+        *output_value_id =
+            semantics_ir_->GetTypeAllowBuiltinTypes(tuple_type_id);
+      }
+      return ImplicitAsKind::Compatible;
+    }
     // When converting `{}` to a type, the result is `{} as Type`.
-    if (value.kind() == SemanticsNodeKind::StructValue &&
-        value.GetAsStructValue() == SemanticsNodeBlockId::Empty) {
+    else if (value.kind() == SemanticsNodeKind::StructValue &&
```
jonmeow:
```suggestion
   if (value.kind() == SemanticsNodeKind::StructValue &&
```

We're using a "return early" design so this doesn't change execution behavior. Note though that this change was breaking style for `else if` formatting, which should've had the `}` on the same line:
https://google.github.io/styleguide/cppguide.html#Formatting_Looping_Branching

Because of the comment in the middle, clang-format doesn't fix this mistake.

toolchain/semantics/semantics_context.h
```diff
@@ -233,6 +233,9 @@ class SemanticsContext {
   auto CanonicalizeStructType(ParseTree::Node parse_node,
                               SemanticsNodeBlockId refs_id) -> SemanticsTypeId;
 
+  auto CanonicalizeTupleType(ParseTree::Node parse_node,
+                             const llvm::SmallVector<SemanticsTypeId>& type_ids)
+      -> SemanticsTypeId;
```
jonmeow:
```suggestion
  // Handles canonicalization of tuple types. This may create a new tuple type
  // when it has a new structure, or reference an existing tuple type when it
  // duplicates a prior type.
  auto CanonicalizeTupleType(ParseTree::Node parse_node,
                             const llvm::SmallVector<SemanticsTypeId>& type_ids)
      -> SemanticsTypeId;

```

Should be commented.

toolchain/semantics/semantics_handle_name.cpp
```diff
@@ -49,6 +49,7 @@ auto SemanticsHandleMemberAccessExpression(SemanticsContext& context,
           context.semantics_ir().GetString(name_id));
       break;
     }
+
```
jonmeow: Why are you adding this?

toolchain/semantics/semantics_handle_paren.cpp
```diff
@@ -9,27 +9,50 @@ namespace Carbon {
...
 }
 
-auto SemanticsHandleTupleLiteralComma(SemanticsContext& context,
-                                      ParseTree::Node parse_node) -> bool {
-  return context.TODO(parse_node, "HandleTupleLiteralComma");
+auto SemanticsHandleParenExpressionOrTupleLiteralStart(
+    SemanticsContext& context, ParseTree::Node parse_node) -> bool {
+  context.node_stack().Push(parse_node);
+  context.ParamOrArgStart();
+  return true;
```
jonmeow: Why did you move this to the bottom of the file?

toolchain/semantics/semantics_handle_paren.cpp
```diff
@@ -9,27 +9,50 @@ namespace Carbon {
...
+      /*for_args=*/true, ParseNodeKind::ParenExpressionOrTupleLiteralStart);
+
+  context.node_stack()
+      .PopAndDiscardSoloParseNode<
+          ParseNodeKind::ParenExpressionOrTupleLiteralStart>();
+  auto node_blocks = context.semantics_ir().GetNodeBlock(refs_id);
+  llvm::SmallVector<SemanticsTypeId> type_ids;
+  for (auto node : node_blocks) {
+    auto type_id = context.semantics_ir().GetNode(node).type_id();
+    type_ids.push_back(type_id);
```
jonmeow:
```suggestion
    type_ids.push_back(context.semantics_ir().GetNode(node).type_id());
```

toolchain/semantics/semantics_handle_paren.cpp
```diff
@@ -9,27 +9,50 @@ namespace Carbon {
...
                                  ParseTree::Node parse_node) -> bool {
-  return context.TODO(parse_node, "HandleTupleLiteral");
+  auto refs_id = context.ParamOrArgEnd(
+      /*for_args=*/true, ParseNodeKind::ParenExpressionOrTupleLiteralStart);
+
+  context.node_stack()
+      .PopAndDiscardSoloParseNode<
+          ParseNodeKind::ParenExpressionOrTupleLiteralStart>();
+  auto node_blocks = context.semantics_ir().GetNodeBlock(refs_id);
+  llvm::SmallVector<SemanticsTypeId> type_ids;
```
jonmeow, line 47:
```suggestion
  llvm::SmallVector<SemanticsTypeId> type_ids;
  type_ids.reserve(node_blocks.size());
```

Reserving the size can reduce reallocations.

toolchain/semantics/testdata/return/fail_missing_return_empty_tuple.carbon
```diff
@@ -1,27 +0,0 @@
...
-// CHECK:STDOUT: types: [
-// CHECK:STDOUT: ]
-// CHECK:STDOUT: nodes: [
-// CHECK:STDOUT: ]
-// CHECK:STDOUT: node_blocks: [
-// CHECK:STDOUT:   [
-// CHECK:STDOUT:   ],
-// CHECK:STDOUT: ]
-
-// CHECK:STDERR: fail_missing_return_empty_tuple.carbon:[[@LINE+1]]:12: Semantics TODO: HandleTupleLiteral
```
jonmeow, line 25: Why are you removing this test? The message may change, but the code here is still invalid.

toolchain/semantics/testdata/return/tuple.carbon
```diff
@@ -0,0 +1,70 @@
...
+// CHECK:STDOUT:   ],
+// CHECK:STDOUT:   [
+// CHECK:STDOUT:     node+7,
+// CHECK:STDOUT:     node+9,
+// CHECK:STDOUT:   ],
+// CHECK:STDOUT: ]
+
+
+fn Main() -> (i32, i32) {
+	return (15, 35);
```
jonmeow:
```suggestion
  return (15, 35);
```

We've generally tried to format our test code similar to how we would c++ code, e.g. for indents here.

toolchain/semantics/testdata/tuples/fail_type_assign.carbon
```diff
@@ -0,0 +1,55 @@
...
+// CHECK:STDOUT:   ],
+// CHECK:STDOUT:   [
+// CHECK:STDOUT:     node+0,
+// CHECK:STDOUT:   ],
+// CHECK:STDOUT:   [
+// CHECK:STDOUT:     node+6,
+// CHECK:STDOUT:   ],
+// CHECK:STDOUT: ]
+
+// CHECK:STDERR: fail_type_assign.carbon:[[@LINE+1]]:25: Cannot implicitly convert from `(Type)` to `(i32)`.
```
jonmeow: In this error message, both of these are expressions (`(Type)`) instead of tuples (`(Type,)`). Can you please fix StringifyType for this?

jonmeow (commented):
Thank you! I'm glad this is working too well. I know we worked on this together, but stepping back and looking at the change I see a few things to clean up. I think it'd be good for @zygoloid to have a look too.

toolchain/semantics/testdata/return/fail_missing_return_empty_tuple.carbon
fasiddique: It is not needed anymore as the tuple is implemented hence it gets covered by the test `toolchain/semantics/testdata/return/fail_missing_return.carbon`

jonmeow: In Carbon, if no return type is specified, a function returns the empty tuple. Thus a function that explicitly returning the empty tuple is different from a function returning some other type. That's why this test was around. It should probably be restored.

toolchain/lowering/lowering_handle.cpp
```diff
@@ -215,6 +215,18 @@ auto LoweringHandleStructType(LoweringFunctionContext& /*context*/,
   // No action to take.
 }
 
+auto LoweringHandleTupleValue(LoweringFunctionContext& /*context*/,
+                              SemanticsNodeId /*node_id*/,
+                              SemanticsNode /*node*/) -> void {
+  // TODO.
```
zygoloid:
```suggestion
  CARBON_FATAL() << "TODO: Add support: " << node;
```
Let's diagnose this error condition, to ensure that lowering doesn't crash when looking for a local value that's not been set.

toolchain/semantics/semantics_context.cpp
```diff
@@ -502,14 +504,30 @@ auto SemanticsContext::ImplicitAsImpl(SemanticsNodeId value_id,
...
   if (as_type_id == SemanticsTypeId::TypeType) {
-    // TODO: When converting `()` to a type, the result is `() as Type`.
-    // Right now there is no tuple value support.
-
+    if (value.kind() == SemanticsNodeKind::TupleValue) {
+      auto tuple_block_id = value.GetAsTupleValue();
+      llvm::SmallVector<SemanticsTypeId> type_ids;
+      // If it is empty tuple type, we don't fetch anything
+      if (tuple_block_id != SemanticsNodeBlockId::Empty) {
+        auto tuple_block = semantics_ir_->GetNodeBlock(tuple_block_id);
```
zygoloid:
```suggestion
        const auto &tuple_block = semantics_ir_->GetNodeBlock(tuple_block_id);
```
Avoid making a copy of the vector here.

toolchain/semantics/semantics_context.cpp
```diff
@@ -502,14 +504,30 @@ auto SemanticsContext::ImplicitAsImpl(SemanticsNodeId value_id,
...
-
   if (as_type_id == SemanticsTypeId::TypeType) {
-    // TODO: When converting `()` to a type, the result is `() as Type`.
-    // Right now there is no tuple value support.
-
+    if (value.kind() == SemanticsNodeKind::TupleValue) {
+      auto tuple_block_id = value.GetAsTupleValue();
+      llvm::SmallVector<SemanticsTypeId> type_ids;
+      // If it is empty tuple type, we don't fetch anything
+      if (tuple_block_id != SemanticsNodeBlockId::Empty) {
```
zygoloid: Minor, but I don't think we need this `if` -- `GetNodeBlock` works for the empty block.

toolchain/semantics/semantics_context.cpp
```diff
@@ -569,6 +587,10 @@ auto SemanticsContext::ParamOrArgSave(bool for_args) -> void {
 
 auto SemanticsContext::CanonicalizeType(SemanticsNodeId node_id)
     -> SemanticsTypeId {
+  auto node = semantics_ir_->GetNode(node_id);
+  if (node.kind() == SemanticsNodeKind::StubReference) {
+    node_id = node.GetAsStubReference();
+  }
```
zygoloid: Should this be:
```suggestion
  while (node.kind() == SemanticsNodeKind::StubReference) {
    node_id = node.GetAsStubReference();
    node = semantics_ir_->GetNode(node_id);
  }
```
or is it guaranteed that we don't form a `StubReference` to a `StubReference`?

toolchain/semantics/semantics_context.cpp
```diff
@@ -604,13 +626,42 @@ auto SemanticsContext::CanonicalizeStructType(ParseTree::Node parse_node,
...
+  void* insert_pos;
+  auto* node =
+      canonical_tuple_types_.FindNodeOrInsertPos(canonical_id, insert_pos);
+  if (node != nullptr) {
+    return node->type_id();
+  }
+  // The tuple type doesn't already exist, so create and store it as canonical.
+  auto type_block_id = semantics_ir_->AddTypeBlock();
+  auto& type_block = semantics_ir_->GetTypeBlock(type_block_id);
+  type_block = type_ids;
```
zygoloid: If possible, it'd be great to `std::move` the vector in here to avoid an unnecessary memory allocation and copy.

toolchain/semantics/semantics_context.h
```diff
@@ -233,10 +233,17 @@ class SemanticsContext {
...
   // Converts an expression for use as a type.
   // TODO: This should eventually return a type ID.
   auto ExpressionAsType(ParseTree::Node parse_node, SemanticsNodeId value_id)
       -> SemanticsTypeId {
+    auto node = semantics_ir_->GetNode(value_id);
+    if (node.kind() == SemanticsNodeKind::StubReference) {
+      value_id = node.GetAsStubReference();
+    }
```
zygoloid: Same comment as before: if we can form a `StubReference` to a `StubReference` we'll need a loop here.

toolchain/semantics/semantics_handle_paren.cpp
```diff
@@ -9,27 +9,50 @@ namespace Carbon {
...
 auto SemanticsHandleTupleLiteral(SemanticsContext& context,
                                  ParseTree::Node parse_node) -> bool {
-  return context.TODO(parse_node, "HandleTupleLiteral");
+  auto refs_id = context.ParamOrArgEnd(
+      /*for_args=*/true, ParseNodeKind::ParenExpressionOrTupleLiteralStart);
+
+  context.node_stack()
+      .PopAndDiscardSoloParseNode<
+          ParseNodeKind::ParenExpressionOrTupleLiteralStart>();
+  auto node_blocks = context.semantics_ir().GetNodeBlock(refs_id);
```
zygoloid:
```suggestion
  const auto& node_blocks = context.semantics_ir().GetNodeBlock(refs_id);
```
Don't make a copy of the vector here. Also, I think the singular name `node_block` would be clearer here -- though ideally I'd like a variable name that reflects what's in this node block. `tuple_elements` maybe?

toolchain/semantics/semantics_ir.cpp
```diff
@@ -241,6 +242,24 @@ auto SemanticsIR::StringifyType(SemanticsTypeId type_id) -> std::string {
...
+        auto refs = GetTypeBlock(node.GetAsTupleType());
+        if (refs.empty()) {
+          out << "() as Type";
+          break;
+        } else if (step.index == 0) {
+          out << "(";
+        } else if (step.index < static_cast<int>(refs.size())) {
+          out << ", ";
+        } else {
+          out << ")";
```
zygoloid:
```suggestion
          out << "() as type";
          break;
        } else if (step.index == 0) {
          out << "(";
        } else if (step.index < static_cast<int>(refs.size())) {
          out << ", ";
        } else {
          out << ") as type";
```
`type` has a lowercase `t`. Also, let's include the `as type` part in all tuple types, not only empty ones.

toolchain/semantics/testdata/function/call/empty_tuple.carbon
```diff
@@ -0,0 +1,75 @@
...
+// CHECK:STDOUT:   Echo,
+// CHECK:STDOUT:   a,
+// CHECK:STDOUT:   Main,
+// CHECK:STDOUT: ]
+// CHECK:STDOUT: types: [
+// CHECK:STDOUT:   node+0,
+// CHECK:STDOUT:   nodeEmptyTupleType,
+// CHECK:STDOUT: ]
+// CHECK:STDOUT: nodes: [
+// CHECK:STDOUT:   {kind: TupleType, arg0: typeBlock0, type: typeTypeType},
```
zygoloid, line 25: This output doesn't include the contents of the `typeBlock`s. Can you update the print function to also print them out?

zygoloid (commented):
I'd like to see a test that the builtin implicit empty tuple type is treated as being the same type as an explicitly-created empty tuple type. (I don't see that being handled in this patch.) For example:

```
fn MakeImplicitEmptyTuple() {}
var v: () = MakeImplicitEmptyTuple();
```

If that doesn't work yet, I think it's fine to leave support for it to a future PR, but I'd like this PR to at least add a testcase with a TODO comment, to make sure we don't crash.

toolchain/semantics/semantics_context.cpp
fasiddique: Thank you! That is a great catch! I can't think of cases where `StubReference` is formed for a `StubReference` . But @jonmeow can answer this better than me.

toolchain/semantics/testdata/function/call/empty_tuple.carbon
fasiddique: I plan to add that in a separate PR as this would require auto updating all the tests and so this PR will get immensely large.

toolchain/semantics/semantics_context.cpp
fasiddique: It's added so that for empty tuple we don't need to call `GetNodeBlock` in order to save a fetch.

toolchain/semantics/semantics_context.cpp
fasiddique: If we use `std::move` here, `type_ids` will be unusable later on. As `type_ids` is passed by reference to this function, I don't think it will be safe to use `std::move` here. Please let me know if you think otherwise!

zygoloid: You can handle that by changing this function to take `type_ids` by value or by rvalue reference. In either case, the caller can pass in `std::move(their_type_ids)`, and that'll avoid the copy that's happening here. The difference is:

- If you use pass by value, then callers will implicitly make a copy if they don't use `std::move`.
- If you use pass by rvalue reference, then callers will get an error if they don't use `std::move` (and will need to manually force a copy if they want one).

Given that we never want to perform implicit and unnecessary memory allocations and copies in the toolchain, I think the second option is preferable here.

toolchain/semantics/testdata/function/call/empty_tuple.carbon
zygoloid: Sounds good to me, thanks.

toolchain/semantics/semantics_context.cpp
jonmeow: Right now I would think it's a bug if we have a recursive stub reference. While it may turn out to be an issue in the future, my preference would be either to not handle it or add a CARBON_CHECK that a while loop isn't needed -- that way if we start doing it we can consider the necessity or if there's a more efficient approach (part of my concern is accidentally adding extra stub unwinding when we could unwind once).

toolchain/semantics/semantics_context.cpp
```diff
@@ -604,13 +628,42 @@ auto SemanticsContext::CanonicalizeStructType(ParseTree::Node parse_node,
...
+  canonical_types_nodes_.push_back(
+      std::make_unique<TypeNode>(canonical_id, type_id));
+  canonical_struct_types_.InsertNode(canonical_types_nodes_.back().get(),
                                      insert_pos);
   return type_id;
 }
 
+auto SemanticsContext::CanonicalizeTupleType(
+    ParseTree::Node parse_node,
+    const llvm::SmallVector<SemanticsTypeId>&& type_ids) -> SemanticsTypeId {
```
zygoloid: This should be a non-const rvalue reference so that we can move from it.
```suggestion
    llvm::SmallVector<SemanticsTypeId>&& type_ids) -> SemanticsTypeId {
```

toolchain/semantics/semantics_context.cpp
```diff
@@ -604,13 +628,42 @@ auto SemanticsContext::CanonicalizeStructType(ParseTree::Node parse_node,
...
+  void* insert_pos;
+  auto* node =
+      canonical_tuple_types_.FindNodeOrInsertPos(canonical_id, insert_pos);
+  if (node != nullptr) {
+    return node->type_id();
+  }
+  // The tuple type doesn't already exist, so create and store it as canonical.
+  auto type_block_id = semantics_ir_->AddTypeBlock();
+  auto& type_block = semantics_ir_->GetTypeBlock(type_block_id);
+  type_block = type_ids;
```
zygoloid:
```suggestion
  type_block = std::move(type_ids);
```

toolchain/semantics/semantics_context.h
```diff
@@ -233,10 +233,17 @@ class SemanticsContext {
...
 
+  auto CanonicalizeTupleType(
+      ParseTree::Node parse_node,
+      const llvm::SmallVector<SemanticsTypeId>&& type_ids) -> SemanticsTypeId;
   // Converts an expression for use as a type.
   // TODO: This should eventually return a type ID.
   auto ExpressionAsType(ParseTree::Node parse_node, SemanticsNodeId value_id)
       -> SemanticsTypeId {
+    auto node = semantics_ir_->GetNode(value_id);
+    if (node.kind() == SemanticsNodeKind::StubReference) {
```
jonmeow, line 247: So you know, this can remain and will work for now, but as we discussed this morning: the point of having the StubReference there is to allow a conversion instruction to be added. By eliding the stub reference before the implicit cast, it removes that "slot" that a conversion might use.

jonmeow (commented):
Let me know when to take another full look; I'm seeing several comments marked resolved right now without edits reflected in the PR.

toolchain/semantics/semantics_context.cpp
jonmeow: Per question, I am saying I don't think this should have a `while`. Either:

1) Do nothing (just don't support it).
2) Explicitly CARBON_CHECK for the case that a stub reference points at another. This is the case where a `while` would have had an effect, but explicitly fails instead of supporting.

In both cases, the idea is to not support it, so that if it's needed we're more likely to evaluate support based on need at that time.

toolchain/semantics/semantics_context.h
fasiddique: In case of nested tuples,  `ImplicitAsImpl` calls `ExpressionAsType` with `StubReference` and then  `ExpressionAsType` calls the `ImplicitAsImpl` with same `StubReference` without canonizing the type. And so, it gets `typeError` even though it should not.  In order to fix that, I added this segment. Could you please let me know what would be the better way to handle this?

toolchain/semantics/semantics_context.cpp
fasiddique: Thank you! I added a carbon check so that we can detect if there's any such case.

fasiddique:
> Let me know when to take another full look; I'm seeing several comments marked resolved right now without edits reflected in the PR.

I addressed all the comments. You can take a look now. Thanks!

toolchain/parser/parse_node_kind.def
```diff
@@ -208,12 +208,12 @@ CARBON_PARSE_NODE_KIND_CHILD_COUNT(WhileConditionStart, 0)
 CARBON_PARSE_NODE_KIND_BRACKET(WhileCondition, WhileConditionStart)
 CARBON_PARSE_NODE_KIND_CHILD_COUNT(WhileStatement, 2)
 
-// Parenthesized expressions:
+// Parenthesized expressions, such as `(2)`:
 //   ParenExpressionOrTupleLiteralStart
 //   _external_: expression
 // ParenExpression
```
jonmeow:
```suggestion
//
```

As commented [here](https://github.com/carbon-language/carbon-lang/pull/2992#discussion_r1267113629), let's replace this blank line with an empty comment line to make it clear these comments are connected.

toolchain/semantics/semantics_context.cpp
jonmeow: This comment still needs to be addressed.

toolchain/semantics/semantics_handle_name.cpp
jonmeow: This comment still is awaiting a response.

toolchain/semantics/semantics_handle_paren.cpp
jonmeow: This comment still is awaiting a response.

toolchain/semantics/testdata/return/tuple.carbon
jonmeow: Unresolving; this is unchanged.

toolchain/semantics/semantics_ir.cpp
```diff
@@ -241,6 +242,27 @@ auto SemanticsIR::StringifyType(SemanticsTypeId type_id) -> std::string {
...
+        auto refs = GetTypeBlock(node.GetAsTupleType());
+        if (refs.empty()) {
+          out << "() as type";
+          break;
+        } else if (step.index == 0) {
+          out << "(";
+        } else if (step.index < static_cast<int>(refs.size())) {
+          out << ", ";
+        } else {
+          if (step.index == 1) {
```
jonmeow, line 257:
```suggestion
          // A tuple of one element has a comma to disambiguate from an expression.
          if (step.index == 1) {
```

Let's comment this behavior.

toolchain/semantics/semantics_context.h
jonmeow: We should be looking at how to handle it inside ImplicitAsImpl, not outside. Yes, ImplicitAsImpl calls ExpressionAsType: however, that same function is also called from outside ImplicitAsImpl in order to get an implicit conversion.

jonmeow (commented):
Thanks for clarifying, I've unresolved some comments. Also, I think you missed a couple: note it may help to check in the GitHub UI under the "Conversations" dropdown (on the file page, to the right of the commit and file filter dropdowns) where it'll show where you have and have not responded.

toolchain/semantics/semantics_context.h
fasiddique: Thank you! Yes, it occurred to me that `ExpressionAsType` can be called from other functions too. That's why I was thinking about adding a `bool` parameter to the `ExpressionAsType` function that would resolve if it is being called from the `ImplicitAsImpl` function. But then I realized it would need changing other codes not related to tuple implementation.

fasiddique:
> Thanks for clarifying, I've unresolved some comments. Also, I think you missed a couple: note it may help to check in the GitHub UI under the "Conversations" dropdown (on the file page, to the right of the commit and file filter dropdowns) where it'll show where you have and have not responded.

Thank you! I have handled all the comments now.

toolchain/semantics/semantics_context.h
fasiddique: Another approach could be resolving this `StubReference` type at the beginning of the `ImplicitAsImpl` function before [this](https://github.com/carbon-language/carbon-lang/blob/trunk/toolchain/semantics/semantics_context.cpp#L353) `if` condition.

toolchain/semantics/semantics_context.cpp
```diff
@@ -502,11 +504,29 @@ auto SemanticsContext::ImplicitAsImpl(SemanticsNodeId value_id,
...
   }
-
   if (as_type_id == SemanticsTypeId::TypeType) {
-    // TODO: When converting `()` to a type, the result is `() as Type`.
-    // Right now there is no tuple value support.
-
+    if (value.kind() == SemanticsNodeKind::TupleValue) {
+      auto tuple_block_id = value.GetAsTupleValue();
+      llvm::SmallVector<SemanticsTypeId> type_ids;
+      // If it is empty tuple type, we don't fetch anything
```
jonmeow:
```suggestion
      // If it is empty tuple type, we don't fetch anything.
```

Missing punctuation

toolchain/semantics/testdata/return/tuple.carbon
```diff
@@ -0,0 +1,70 @@
...
+// CHECK:STDOUT:   ],
+// CHECK:STDOUT:   [
+// CHECK:STDOUT:     node+7,
+// CHECK:STDOUT:     node+9,
+// CHECK:STDOUT:   ],
+// CHECK:STDOUT: ]
+
+
+fn Main() -> (i32, i32) {
+ return (15, 35);
```
jonmeow:
```suggestion
  return (15, 35);
```

Please make absolutely sure you have a two space indent this time. You removed a space in https://github.com/carbon-language/carbon-lang/pull/2992/commits/ffca5be5c834fb1f7ce70f46aee9c3c38d788b8e

toolchain/semantics/testdata/tuples/builtin_implicit_empty_tuple.carbon
```diff
@@ -0,0 +1,26 @@
...
+// CHECK:STDOUT: types: [
+// CHECK:STDOUT: ]
+// CHECK:STDOUT: nodes: [
+// CHECK:STDOUT: ]
+// CHECK:STDOUT: node_blocks: [
+// CHECK:STDOUT:   [
+// CHECK:STDOUT:   ],
+// CHECK:STDOUT: ]
+
+// TODO test if builtin implicit empty tuple type is treated as being the same type as an explicitly-created empty tuple type.
```
jonmeow: Previously I thought we'd discussed that the builtin empty tuple type should be removed. Is this the right test / TODO?

toolchain/semantics/semantics_context.cpp
```diff
@@ -569,6 +589,13 @@ auto SemanticsContext::ParamOrArgSave(bool for_args) -> void {
 
 auto SemanticsContext::CanonicalizeType(SemanticsNodeId node_id)
     -> SemanticsTypeId {
+  auto node = semantics_ir_->GetNode(node_id);
+  if (node.kind() == SemanticsNodeKind::StubReference) {
+    node_id = node.GetAsStubReference();
+  }
+  CARBON_CHECK(semantics_ir_->GetNode(node_id).kind() !=
+               SemanticsNodeKind::StubReference)
+      << "Stub reference should not point to another stub reference";
```
jonmeow: Why is this CHECK outside the if statement? Similar for the copy of it.

jonmeow (commented):
Thanks, I think this is really close now!

toolchain/semantics/testdata/tuples/builtin_implicit_empty_tuple.carbon
fasiddique: I added this test in response to [this](https://github.com/carbon-language/carbon-lang/pull/2992#pullrequestreview-1537849360). Please let me know if it is incorrect or if I should have handled it differently. Thanks!

jonmeow: @zygoloid requested there a test case that at least ensures the assignment doesn't crash. However, with the body commented out, it cannot be verifying a non-crash.

Since the implicit empty tuple is a feature of function calls, and we already have similar with function/call/i32.carbon, my suggest is to instead name this function/call/return_implicit.carbon. Test the assignment to an empty tuple, and do not comment out the code so that it [shows it] does not crash [even if it's an error].

The TODO should be that the implicit return should be a regular empty tuple.

toolchain/semantics/semantics_handle_paren.cpp
```diff
@@ -19,17 +25,36 @@ auto SemanticsHandleParenExpression(SemanticsContext& context,
...
-                                      ParseTree::Node parse_node) -> bool {
-  return context.TODO(parse_node, "HandleTupleLiteralComma");
+  context.node_stack()
+      .PopAndDiscardSoloParseNode<
+          ParseNodeKind::ParenExpressionOrTupleLiteralStart>();
+  const auto& node_block = context.semantics_ir().GetNodeBlock(refs_id);
+  llvm::SmallVector<SemanticsTypeId> type_ids;
+  type_ids.reserve(node_block.size());
+  for (auto node : node_block) {
+    type_ids.push_back(context.semantics_ir().GetNode(node).type_id());
```
fasiddique, line 50: I think types need to be canonicalized here otherwise it may add invalid type ids to the tuple type block [here](https://github.com/carbon-language/carbon-lang/pull/2992/files#diff-0aac17c90333793c50c1315abd8f2c66f6d3a6a4ca2c98a6bfee8083afce2afeR659) @jonmeow could you please let me know your thought regarding this.

toolchain/semantics/testdata/tuples/builtin_implicit_empty_tuple.carbon
fasiddique: @zygoloid currently this case crashes however it is not because of the changes made in this PR. It is because the type id is not used [here](https://github.com/carbon-language/carbon-lang/blob/trunk/toolchain/semantics/semantics_handle_function.cpp#L20)

toolchain/semantics/semantics_handle_paren.cpp
jonmeow: At type ID has, by definition, already been canonicalized.

jonmeow: Put differently, an invalid type id here indicates a bug in the code that led to it being here, not a need for canonicalization.

jonmeow (approved):
At this point I think the PR is in a reasonable state to merge, but I'm checking with Richard.

toolchain/semantics/semantics_context.cpp
zygoloid: For what it's worth, I'd expect the empty node block to be in cache anyway, so I would expect the cost of the fetch to be negligible. On the other hand, this adds a branch to the case of forming a non-empty tuple, but the cost of that seems likely to also be negligible. I wouldn't trust that this actually gives a performance improvement without measuring it.

But this is minor, and I think this PR is fine to merge as-is.

zygoloid (approved):

PR #2995, opened 2023-07-18
Temporary restriction that `extend base` is first in a class to avoid #2994

### LINKED ISSUE #2994: Fuzzer issue: extending a member class as a base class
### Description of the bug:

Proposal #2760 moved the declaration of a base class from the class declaration line into the body of the class definition. This change was implemented in Explorer in #2906 . However, this change did not account for the consequence of this change that the base class expression could now refer to something declared earlier in class scope. The implementation in #2906 of `TypeChecker::DeclareClassDeclaration(...)` in `explorer/interpreter/type_checker.cpp` type checks the base class expression before `DeclareDeclaration(...)` can be called on the members. If the base class expression contains a reference to a member (such as a member class), it fails when it tries to access that member's `static_type`, which hasn't been set yet.

Unfortunately, this code is difficult to reorder, since the members can refer to `Self` which is a `NominalClassType` created in this function, with `base_class` as a parameter, which we don't know until we've processed the `extend base` declaration.

This triggers the bug:
```
package P impl;

class C {
  class D { }
  extend base: D;
}

```

This does not:
```
package P impl;

class D { }
class C {
  extend base: D;
}

```

No crash. This code should instead produce some compilation error about trying to extend a class that is final.

### What actually happened?

Crash:
> ... optional:943: _LIBCPP_ASSERT 'this->has_value()' failed. optional operator* called on a disengaged value

_No response_

josh11b:
Closes #2994

### DIFF
--- a/explorer/interpreter/type_checker.cpp
+++ b/explorer/interpreter/type_checker.cpp
@@ -5291,57 +5291,54 @@ auto TypeChecker::DeclareClassDeclaration(Nonnull<ClassDeclaration*> class_decl,
     class_decl->set_static_type(&self->static_type());
   }
 
-  // Find base class declaration, if any. Verify that is before any data member
-  // declarations, and there is at most one.
-  std::optional<Nonnull<const NominalClassType*>> base_class;
-  bool after_data_member = false;
-  for (Nonnull<Declaration*> m : class_decl->members()) {
-    switch (m->kind()) {
-      case DeclarationKind::VariableDeclaration:
-      case DeclarationKind::MixDeclaration: {
-        after_data_member = true;
-        break;
-      }
-      case DeclarationKind::ExtendBaseDeclaration: {
-        if (after_data_member) {
-          return ProgramError(m->source_loc())
-                 << "`extend base:` declaration must not be after `var` or "
-                    "`mix` declarations in a class.";
-        }
-        if (base_class.has_value()) {
-          return ProgramError(m->source_loc())
-                 << "At most one `extend base:` declaration in a class.";
-        }
-        Nonnull<Expression*> base_class_expr =
-            cast<ExtendBaseDeclaration>(*m).base_class();
-        CARBON_ASSIGN_OR_RETURN(const auto base_type,
-                                TypeCheckTypeExp(base_class_expr, class_scope));
-        if (base_type->kind() != Value::Kind::NominalClassType) {
-          return ProgramError(m->source_loc())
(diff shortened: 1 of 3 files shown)


explorer/interpreter/type_checker.cpp
```diff
@@ -5291,57 +5291,54 @@ auto TypeChecker::DeclareClassDeclaration(Nonnull<ClassDeclaration*> class_decl,
...
+  // hasn't been typechecked yet and therefore doesn't have its `static_type`
+  // set.
+
+  // TODO: Verify just that is before any data member declarations, and there is
+  // at most one, and delay remaining work (type checking, base class
+  // evaluation, etc.) until the `extend base` declaration is processed in
+  // order.
   std::optional<Nonnull<const NominalClassType*>> base_class;
-  bool after_data_member = false;
+  bool first = true;
```
geoffromer: Would it work to process `class_decl->members()[0]` before the loop, and then loop over the remaining members to check that they're not `ExtendBaseDeclaration`s? That seems like it would be clearer to me, because the logic is expressed more in the static structure of the code rather than in a dynamic state variable. It might also be a little safer, because there's no risk of forgetting to update the state variable.

explorer/interpreter/type_checker.cpp
```diff
@@ -5291,57 +5291,54 @@ auto TypeChecker::DeclareClassDeclaration(Nonnull<ClassDeclaration*> class_decl,
...
-        class_decl->set_base_type(base_class);
-        break;
+    if (m->kind() != DeclarationKind::ExtendBaseDeclaration) {
+      first = false;
+    } else if (base_class.has_value()) {
+      return ProgramError(m->source_loc())
+             << "At most one `extend base:` declaration in a class.";
+    } else if (!first) {
+      return ProgramError(m->source_loc())
+             << "`extend base:` declaration must first in a class.";
```
geoffromer: It'd be nice to leave a breadcrumb that this is a limitation of the Explorer, not a language rule.

```suggestion
             << "`extend base:` declarations after the first declaration in the class are not yet supported";
```

josh11b: Agreed

explorer/interpreter/type_checker.cpp
josh11b: It isn't clearly better to me, but I made the change.

geoffromer (approved):

PR #2997, opened 2023-07-19
Fix crash if no input file is given to `carbon dump objcode`

zygoloid:
If the final argument was `--target_triple=...`, we tried to read off the end of the arguments list. Found by fuzzer.


jonmeow (approved):

jonmeow:
Looks good!

PR #2999, opened 2023-07-19
Track the file kind on SourceLocation rather than computing it from the file name.

zygoloid:
This will be more correct if a user file is named prelude.carbon, and reduces the explorer runtime by about 10% by removing a string comparison from the check for whether trace output is enabled.


explorer/parse_and_execute/parse_and_execute_test.cpp
```diff
@@ -36,9 +37,9 @@ TEST(ParseAndExecuteTest, Recursion) {
       ParseAndExecute("explorer/data/prelude.carbon", "test.carbon", source,
                       /*parser_debug=*/false, &trace_stream, &llvm::nulls());
   ASSERT_FALSE(err.ok());
-  EXPECT_THAT(err.error().message(),
-              Eq("RUNTIME ERROR: overflow:1: stack overflow: too many "
```
jonmeow, line 40: Shouldn't this still be tagged as a runtime error? Part of validating that was to ensure it's consistent with the typical error flow.

explorer/interpreter/interpreter.cpp
```diff
@@ -2706,21 +2706,28 @@ auto Interpreter::StepCleanUp() -> ErrorOr<Success> {
 
 // State transition.
 auto Interpreter::Step() -> ErrorOr<Success> {
+  Action& act = todo_.CurrentAction();
+
+  auto fatal_error_builder = [&] {
```
jonmeow: All the errors are "fatal", but not typically labeled as such -- this made me confused briefly versus CARBON_FATAL. Maybe just error_builder or choose_error_builder? It's too bad this needs to be a lambda instead of a variable...

explorer/interpreter/interpreter.cpp
```diff
@@ -2706,21 +2706,28 @@ auto Interpreter::StepCleanUp() -> ErrorOr<Success> {
...
+    if (auto loc = act.source_loc())
+      return ProgramError(*loc);
```
jonmeow:
```suggestion
    if (auto loc = act.source_loc()) {
      return ProgramError(*loc);
    }
```

Minor style

explorer/main.cpp
```diff
@@ -117,8 +118,34 @@ auto ExplorerMain(int argc, char** argv, void* static_for_main_addr,
 
   if (!trace_file_name.empty()) {
     // Adding allowed phases in the trace_stream
```
jonmeow:
```suggestion
    // Adding allowed phases in the trace_stream.
```

Not your comment, but I can't unsee it, and github lets me suggest. :)

explorer/common/source_location.h
```diff
@@ -13,32 +13,40 @@
...
 
   // The filename should be eternal or arena-allocated to eliminate copies.
-  constexpr SourceLocation(const char* filename, int line_num)
-      : filename_(filename), line_num_(line_num) {}
-  SourceLocation(Nonnull<const std::string*> filename, int line_num)
-      : filename_(*filename), line_num_(line_num) {}
+  constexpr SourceLocation(const char* filename, int line_num,
+                           FileKind file_kind)
+      : filename_(filename), line_num_(line_num), file_kind_(file_kind) {}
+  SourceLocation(Nonnull<const std::string*> filename, int line_num,
```
jonmeow:
```suggestion
  explicit constexpr SourceLocation(const char* filename, int line_num,
                           FileKind file_kind)
      : filename_(filename), line_num_(line_num), file_kind_(file_kind) {}
  explicit SourceLocation(Nonnull<const std::string*> filename, int line_num,
```

Per past explicit constructor style discussions, I think it'll work without affecting callers, and you're modifying them anyways.

explorer/common/source_location.h
```diff
@@ -13,32 +13,40 @@
...
 
   SourceLocation(const SourceLocation&) = default;
   SourceLocation(SourceLocation&&) = default;
   auto operator=(const SourceLocation&) -> SourceLocation& = default;
   auto operator=(SourceLocation&&) -> SourceLocation& = default;
 
   auto operator==(SourceLocation other) const -> bool {
-    return filename_ == other.filename_ && line_num_ == other.line_num_;
+    return filename_ == other.filename_ && line_num_ == other.line_num_ &&
+           file_kind_ == other.file_kind_;
```
jonmeow, line 43: This makes me wonder how we compare source locations. The filename being first seems like a poor ordering for performance, but I'm not sure that we really do this much.

(no action requested, just mulling)

jonmeow (approved):
I'm glad to see the performance improvement!

explorer/parse_and_execute/parse_and_execute_test.cpp
zygoloid: It is still tagged as a runtime error, but it also includes a real filename and line number now, which will be different depending on exactly what we set the stack limit to. I can check for a prefix of `RUNTIME ERROR:` as well if you like?

jonmeow: I'd prefer that, yes. You could switch to a regex match FWIW

zygoloid: Done.

explorer/interpreter/interpreter.cpp
zygoloid: Done.

explorer/common/source_location.h
zygoloid: For what it's worth, the `operator==` is currently only used in unit tests (eg, ast/expression_test.cpp).

jonmeow (approved):
Thanks!

PR #3006, opened 2023-07-20
Clean up uncommented SmallVector size parameters.

jonmeow:
The lack of comments makes it difficult to be sure, but if any of these are necessary, I'd like to make sure there are comments to explain why a specific value was shown (which I do add for diagnostics_ in this change).


zygoloid (approved):
Agreed, I can't really justify any of these other explicit sizes.

PR #3007, opened 2023-07-21
treesitter: implement strings completely

maan2003:
follow up to #2902

### DIFF
--- a/utils/treesitter/BUILD
+++ b/utils/treesitter/BUILD
@@ -29,15 +29,33 @@ cc_test(
 )
 
 filegroup(
-    name = "fail_testdata",
-    srcs = glob(["testdata/**/fail_*.carbon"]),
+    name = "string_testdata",
+    srcs = glob(
+        ["testdata/string/*.carbon"],
+        exclude = ["testdata/string/fail_*.carbon"],
+    ),
+)
+
+filegroup(
+    name = "string_fail_testdata",
+    srcs = glob(["testdata/string/fail_*.carbon"]),
 )
 
 cc_test(
-    name = "fail_tests",
+    name = "string_tests",
     srcs = ["test_runner.cpp"],
-    args = ["$(locations :fail_testdata)"],
-    data = [":fail_testdata"],
+    args = ["$(locations :string_testdata)"],
+    data = [":string_testdata"],
+    deps = [
+        ":treesitter",
+    ],
+)
+
+cc_test(
+    name = "string_fail_tests",
+    srcs = ["test_runner.cpp"],
+    args = ["$(locations :string_fail_testdata)"],
+    data = [":string_fail_testdata"],
     env = {
         "FAIL_TESTS": "1",
     },
--- a/utils/treesitter/BUILD
+++ b/utils/treesitter/BUILD
@@ -27,3 +27,21 @@ cc_test(
         ":treesitter",
     ],
 )
+
+filegroup(
+    name = "fail_testdata",
+    srcs = glob(["testdata/**/fail_*.carbon"]),
+)
+
+cc_test(
+    name = "fail_tests",
+    srcs = ["test_runner.cpp"],
+    args = ["$(locations :fail_testdata)"],
+    data = [":fail_testdata"],
+    env = {
+        "FAIL_TESTS": "1",
+    },
+    deps = [
+        ":treesitter",
+    ],
+)
--- a/utils/treesitter/src/scanner.c
+++ b/utils/treesitter/src/scanner.c
@@ -7,6 +7,7 @@
 enum TokenType {
   BINARY_STAR,
   POSTFIX_STAR,
+  STRING,
 };
 
 // our scanner is stateless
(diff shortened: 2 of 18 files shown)


utils/treesitter/src/scanner.c
```diff
@@ -54,34 +82,98 @@ static bool is_whitespace(char c) { return c == ' ' || c == '\n'; }
...
+             valid_symbols[STRING]) {
+    lexer->result_symbol = STRING;
+    int hash_count = 0;
+    while (lexer->lookahead == '#') {
+      advance(lexer);
+      hash_count++;
+    }
+    if (lexer->lookahead == '\'') {
+      if (!eat(lexer, "'''")) {
+        return false;
```
josh11b, line 112: This seems like it would be in a funny state here. Is `'` or `''` correctly handled? Would be good to have some tests for that.

utils/treesitter/src/scanner.c
```diff
@@ -44,6 +45,33 @@ static bool token_allowed_after_binary_operator(char c) {
...
+    string++;
+  }
+  return true;
+}
+
+static bool eat_count(struct TSLexer* lexer, char ch, int count) {
+  int matched = 0;
+  while (matched != count) {
+    if (lexer->lookahead != ch) {
+      break;
```
josh11b, line 56: Could just `return false;` here and `return true;` below

utils/treesitter/src/scanner.c
```diff
@@ -54,34 +82,98 @@ static bool is_whitespace(char c) { return c == ' ' || c == '\n'; }
...
+          advance(lexer);
+        }
+      }
+    } else if (lexer->lookahead == '"') {
+      advance(lexer);
+      while (!lexer->eof(lexer)) {
+        if (lexer->lookahead == '\\') {
+          advance(lexer);
+          if (eat_count(lexer, '#', hash_count)) {
+            // skip the next character
```
josh11b: I'd much rather the comment say why than say what the code does. It isn't surprising that `advance(lexer)` skips a character, it is more surprising that it is needed.

utils/treesitter/BUILD
```diff
@@ -27,3 +27,21 @@ cc_test(
...
+    srcs = ["test_runner.cpp"],
+    args = ["$(locations :fail_testdata)"],
+    data = [":fail_testdata"],
+    env = {
+        "FAIL_TESTS": "1",
+    },
+    deps = [
+        ":treesitter",
+    ],
+)
```
josh11b: This is acting like this is all about failures, when it seems to be all about strings.

utils/treesitter/src/scanner.c
```diff
@@ -44,6 +45,33 @@ static bool token_allowed_after_binary_operator(char c) {
 
 static bool is_whitespace(char c) { return c == ' ' || c == '\n'; }
 
+static void advance(struct TSLexer* lexer) {
+  lexer->advance(lexer, /* skip= */ false);
+}
+
+static bool eat(struct TSLexer* lexer, const char* string) {
```
josh11b: If this returns `false`, the lexer has advanced an unknown number of tokens , which seems hard to recover from.

utils/treesitter/src/scanner.c
```diff
@@ -54,34 +82,98 @@ static bool is_whitespace(char c) { return c == ' ' || c == '\n'; }
...
+              lexer->lookahead == '\"') &&
+             valid_symbols[STRING]) {
+    lexer->result_symbol = STRING;
+    int hash_count = 0;
+    while (lexer->lookahead == '#') {
+      advance(lexer);
+      hash_count++;
+    }
+    if (lexer->lookahead == '\'') {
+      if (!eat(lexer, "'''")) {
```
josh11b: Your calls to `eat` could be `eat_count(lexer, '\'', 3)` to only have one function, though maybe that is less clear. Advantage of only using `eat_count` is it could return a number (either the number matched or unmatched) instead of a bool, which would mean information wasn't lost.

utils/treesitter/src/scanner.c
maan2003: Thanks, it revealed incorrect case
```
var s: String = "\
";
```

utils/treesitter/src/scanner.c
maan2003: josh11b: I still don't see tests for this code path.

(copied from https://github.com/Maan2003/carbon-lang/commit/23987180df86bc36260cfdb72dccabfdaa06f993#r122813508)

utils/treesitter/BUILD
maan2003: renamed to string_testdata and string_fail_testdata

maan2003:
update: explorer string testdata is failing now.
EDIT: fixed now

utils/treesitter/src/scanner.c
maan2003: added fail tests for `'` and `''`

josh11b (approved):

PR #3010, opened 2023-07-21
Add line output to diagnostics to help identify error locations.

jonmeow:
This also makes the filename a reference to the buffer since the line seems better to have as a reference (versus copying a string per error). Most tests now have different line deltas due to the extra output, but the actual errors should overall stay the same.

Some of the error locations look like they could be improved, but this change is only making it clear where they were before.

### DIFF
--- a/toolchain/lexer/tokenized_buffer.cpp
+++ b/toolchain/lexer/tokenized_buffer.cpp
@@ -88,7 +88,7 @@ class TokenizedBuffer::Lexer {
         emitter_(translator_, consumer),
         token_translator_(&buffer, &current_column_),
         token_emitter_(token_translator_, consumer),
-        current_line_(buffer.AddLine({0, 0, 0})),
+        current_line_(buffer.AddLine(LineInfo(0))),
         current_line_info_(&buffer.GetLineInfo(current_line_)) {}
 
   // Perform the necessary bookkeeping to step past a newline at the current
@@ -97,7 +97,7 @@ class TokenizedBuffer::Lexer {
     current_line_info_->length = current_column_;
 
     current_line_ = buffer_->AddLine(
-        {current_line_info_->start + current_column_ + 1, 0, 0});
+        LineInfo(current_line_info_->start + current_column_ + 1));
     current_line_info_ = &buffer_->GetLineInfo(current_line_);
     current_column_ = 0;
     set_indent_ = false;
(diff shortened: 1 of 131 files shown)


jonmeow:
I do want to note, the handling of incomplete lines bothered me here, so the logic is refined in #3011. It builds on this change, but also more clearly shows no tests are affected by the logic change (which is intended to offer equivalent support).

toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -907,23 +907,37 @@ auto TokenizedBuffer::SourceBufferLocationTranslator::GetLocation(
...
+    line = buffer_->source_->text().substr(start).take_until(
+        [](char c) { return c == '\n'; });
+  } else if (line_it->length < 0) {
+    line =
+        buffer_->source_->text().substr(line_it->start).take_until([](char c) {
+          return c == '\n';
+        });
+  } else {
+    line = buffer_->source_->text().substr(line_it->start, line_it->length);
   }
```
zygoloid, line 937:
```suggestion
  // We might still be lexing the last line. If so, check to see if there are
  // any newline characters between the position we've finished lexing up to
  // and the given location.
  int64_t start = line_it->start;
  if (incomplete_line_info && column_number > *last_line_lexed_to_column_) {
    column_number = *last_line_lexed_to_column_;
    for (int64_t i = start + *last_line_lexed_to_column_; i != offset;
         ++i) {
      if (buffer_->source_->text()[i] == '\n') {
        start = i;
        ++line_number;
        column_number = 0;
      } else {
        ++column_number;
      }
    }
  }
  
  int64_t length = line_it->length;
  if (length < 0) {
    length = buffer_->source_->text().find_first_of('\n', start);
  }

  llvm::StringRef line = buffer_->source_->text().substr(start, length);
```

I think we can simplify this a little.

zygoloid (approved):
Very nice!

I think in the longer term we'll want the diagnostics layer to decide for itself what text it wants to quote from the source file(s), but tracking the line in the diagnostic seems fine for the time being.

toolchain/lexer/tokenized_buffer.cpp
jonmeow: I have a different change in 3011. I'm hesitant here to rely on how if no `\n` is found then npos results in substr going to the end, I think that can be a little confusing. 

(if you want to nudge more in this direction, let's use #3011 since I'm also rewriting the incomplete_line_info logic there)

jonmeow: Sidenote -- the length calculation on 930 is incorrect, that's the offset from the original line start, which can be changed on line 919

PR #3011, opened 2023-07-21
Clean up handling of incomplete line locations.

jonmeow:
Building on #3010, the handling of incomplete lines seems like it can be straightened out. Doing this separately because it seemed better to demonstrate tests aren't affected by the change.

### DIFF
--- a/toolchain/lexer/tokenized_buffer.cpp
+++ b/toolchain/lexer/tokenized_buffer.cpp
@@ -897,8 +897,6 @@ auto TokenizedBuffer::SourceBufferLocationTranslator::GetLocation(
   const auto* line_it = std::partition_point(
       buffer_->line_infos_.begin(), buffer_->line_infos_.end(),
       [offset](const LineInfo& line) { return line.start <= offset; });
-  bool incomplete_line_info = last_line_lexed_to_column_ != nullptr &&
-                              line_it == buffer_->line_infos_.end();
 
   // Step back one line to find the line containing the given position.
   CARBON_CHECK(line_it != buffer_->line_infos_.begin())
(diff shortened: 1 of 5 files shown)


toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -907,33 +905,31 @@ auto TokenizedBuffer::SourceBufferLocationTranslator::GetLocation(
...
+  llvm::StringRef line =
+      buffer_->source_->text().substr(line_it->start, line_it->length);
+  if (line_it->length == static_cast<int32_t>(llvm::StringRef::npos)) {
+    // The location of the diagnostic hasn't been lexed yet. Check to see if
+    // there are any newline characters between the position we've finished
+    // lexing up to and the given location.
+    if (column_number > *last_line_lexed_to_column_) {
+      auto start_newline_pos =
+          line.take_front(*last_line_lexed_to_column_).rfind('\n');
+      if (start_newline_pos != llvm::StringRef::npos) {
```
zygoloid: Can this happen? It looks like this would mean there is a newline in the first `*last_line_lexed_to_column_` bytes of the last line, which seems like it should be impossible (if there's a newline there, we should have a later last line).

jonmeow: I'd really thought there was something here, but I see my logic was off already. I've switched to a CHECK so that if this happens we can fix it.

toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -907,33 +905,26 @@ auto TokenizedBuffer::SourceBufferLocationTranslator::GetLocation(
...
+  // Start by grabbing the buffer. If the line isn't fully lexed, the length
+  // will be npos and the buffer will be grabbed from the start; we'll then
+  // adjust the length.
+  llvm::StringRef line =
+      buffer_->source_->text().substr(line_it->start, line_it->length);
+  if (line_it->length == static_cast<int32_t>(llvm::StringRef::npos)) {
+    CARBON_CHECK(column_number < *last_line_lexed_to_column_ ||
+                 line.substr(*last_line_lexed_to_column_,
+                             column_number - *last_line_lexed_to_column_)
+                         .count('\n') == 0)
```
zygoloid: I think there can be newlines between `*last_line_lexed_to_column_` and `column_number`, eg in:

```
var s: String = '''
  error here: '''
```

... our last line will be line 1, and it'll be lexed to the start of the `'''`, but the error column will be on a subsequent line. But there can't be newlines in the initial `*last_line_lexed_to_column_` portion (because if there were, we'd have created new lines for them already as part of lexing to that column).

toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -907,33 +905,26 @@ auto TokenizedBuffer::SourceBufferLocationTranslator::GetLocation(
...
-      } else {
-        ++column_number;
-      }
+  // Start by grabbing the buffer. If the line isn't fully lexed, the length
+  // will be npos and the buffer will be grabbed from the start; we'll then
+  // adjust the length.
+  llvm::StringRef line =
+      buffer_->source_->text().substr(line_it->start, line_it->length);
+  if (line_it->length == static_cast<int32_t>(llvm::StringRef::npos)) {
+    CARBON_CHECK(column_number < *last_line_lexed_to_column_ ||
```
zygoloid: It makes me a little uncomfortable to read from `last_line_lexed_to_column_` here without checking that it's not null. (It's only non-null while lexing.) We shouldn't have any lines with a length of `-1` once we finish lexing, so we could `CHECK` this, if we want to stop using it to determine whether we're in an incomplete line.

toolchain/lexer/tokenized_buffer.cpp
zygoloid: There can be newlines in `[*last_line_lexed_to_column_, column_number)`, but I think there can't be newlines in `[0, *last_line_lexed_to_column_)`.

toolchain/lexer/tokenized_buffer.cpp
jonmeow: I agree, I thought so too, I thought it's why this was added originally. However, #999 was just dealing with performance, and the regression case that was added there still does fine without the code.

Added an explicit test with your example to make it clear this is properly handled. Without a test to demonstrate it's needed, I do think this is removable.

toolchain/lexer/tokenized_buffer.cpp
jonmeow: Removed last_line_lexed_to_column_. It's a bit obtuse how it's passed in anyways.

toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -907,33 +905,23 @@ auto TokenizedBuffer::SourceBufferLocationTranslator::GetLocation(
...
-         ++i) {
-      if (buffer_->source_->text()[i] == '\n') {
-        start = i;
-        ++line_number;
-        column_number = 0;
-      } else {
-        ++column_number;
-      }
+  // Start by grabbing the buffer. If the line isn't fully lexed, the length
+  // will be npos and the buffer will be grabbed from the start; we'll then
```
zygoloid: I don't understand the "and the buffer will be grabbed from the start" here; should this say "and the buffer will be grabbed to the end"?

toolchain/lexer/tokenized_buffer.cpp
zygoloid: Hm. I see, we get away with this because the only place we lex a multi-line token (or more generally, do any lookahead past a newline) is in `LexStringLiteral`, and it happens to produce no errors that point into the source buffer until *after* it's scanned the literal and called `HandleNewline` for each newline in it.

This seems fragile, but I suppose the alternative is to add code here that we can't test to handle a case that doesn't currently happen.

zygoloid (approved):

toolchain/lexer/tokenized_buffer.cpp
jonmeow: Rephrased:

```
  // Start by grabbing the line from the buffer. If the line isn't fully lexed,
  // the length will be npos and the line will be grabbed from the known start
  // to the end of the buffer; we'll then adjust the length.
```

zygoloid (approved):

PR #3012, opened 2023-07-21
Change DiagnosticLocationTranslator to a function.

jonmeow:
There's a TODO in diagnostic_emitter to consider changing this class to just a std::function once more infrastructure is in. I think we're in a reasonable position right now to make this change. It feels like a helpful reduction in code.


jonmeow:
Note this PR is the 3rd commit.

toolchain/diagnostics/diagnostic_emitter.h
```diff
@@ -132,20 +132,6 @@ class DiagnosticConsumer {
   virtual auto Flush() -> void {}
 };
 
-// An interface that can translate some representation of a location into a
-// diagnostic location.
-//
-// TODO: Revisit this once the diagnostics machinery is more complete and see
-// if we can turn it into a `std::function`.
```
zygoloid, line 137: I don't think we're there yet. In particular, we don't support highlights or annotations on source ranges, which are the kinds of features that I was concerned may lead to us wanting additional functions on this interface.

jonmeow: Okay, I'll revert this change.

PR #3015, opened 2023-07-24
Explorer: add trace for unformed variable resolution

prabhatexit0:
Adds trace information for unformed variables resolution.


explorer/interpreter/resolve_unformed.cpp
```diff
@@ -52,39 +68,52 @@ auto FlowFacts::TakeAction(Nonnull<const AstNode*> node, ActionType action,
...
+  });
 }
 
-static auto ResolveUnformedImpl(Nonnull<const Expression*> expression,
+static auto ResolveUnformedImpl(Nonnull<TraceStream*> trace_stream,
+                                Nonnull<const Expression*> expression,
                                 FlowFacts& flow_facts,
                                 FlowFacts::ActionType action)
     -> ErrorOr<Success> {
+  if (trace_stream->is_enabled()) {
```
zygoloid: I worry that this is going to produce a lot of output; is it reasonable to do the same thing we did for name resolution and only log statements and declarations?

zygoloid:
> TODO: Add detailed description.

Can you clean up the initial comment on this PR? It'll end up being used as the commit description in the new commit-queue workflow.

prabhatexit0:
> > TODO: Add detailed description.
> 
> Can you clean up the initial comment on this PR? It'll end up being used as the commit description in the new commit-queue workflow.

Yes, I'll do that

explorer/interpreter/resolve_unformed.cpp
prabhatexit0: Yes, I think that sounds reasonable for this as well. I'll do the changes in the next commit.

zygoloid (approved):

PR #3016, opened 2023-07-24
Lowering for tuple

fasiddique:
Lowering for tuple with a workaround for typetype. I don't think this will generate tuple objects with correct values for the typetype issue.

### DIFF
--- a/toolchain/lowering/lowering_context.cpp
+++ b/toolchain/lowering/lowering_context.cpp
@@ -170,14 +170,7 @@ auto LoweringContext::BuildType(SemanticsNodeId node_id) -> llvm::Type* {
       llvm::SmallVector<llvm::Type*> subtypes;
       subtypes.reserve(refs.size());
       for (auto ref_id : refs) {
-        if (ref_id == SemanticsTypeId::TypeType) {
-          // TODO: Handle TypeType. Currently using a place holder
-          // but this is just a workaround not an actual solution.
-          subtypes.push_back(llvm::Type::getInt1Ty(*llvm_context_));
-        } else {
-          subtypes.push_back(GetType(ref_id));
-        }
-        CARBON_CHECK(ref_id.index < SemanticsBuiltinKind::ValidCount) << ref_id;
+        subtypes.push_back(GetType(ref_id));
       }
       return llvm::StructType::create(*llvm_context_, subtypes,
                                       "TupleLiteralType");
--- a/toolchain/lowering/lowering_context.cpp
+++ b/toolchain/lowering/lowering_context.cpp
@@ -165,6 +165,23 @@ auto LoweringContext::BuildType(SemanticsNodeId node_id) -> llvm::Type* {
       return llvm::StructType::create(*llvm_context_, subtypes,
                                       "StructLiteralType");
     }
+    case SemanticsNodeKind::TupleType: {
+      auto refs = semantics_ir_->GetTypeBlock(node.GetAsTupleType());
+      llvm::SmallVector<llvm::Type*> subtypes;
+      subtypes.reserve(refs.size());
+      for (auto ref_id : refs) {
+        if (ref_id == SemanticsTypeId::TypeType) {
+          // TODO: Handle TypeType. Currently using a place holder
+          // but this is just a workaround not an actual solution.
+          subtypes.push_back(llvm::Type::getInt1Ty(*llvm_context_));
+        } else {
+          subtypes.push_back(GetType(ref_id));
+        }
+        CARBON_CHECK(ref_id.index < SemanticsBuiltinKind::ValidCount) << ref_id;
+      }
+      return llvm::StructType::create(*llvm_context_, subtypes,
+                                      "TupleLiteralType");
+    }
     default: {
       CARBON_FATAL() << "Cannot use node as type: " << node_id;
     }
--- a/toolchain/lowering/lowering_function_context.h
+++ b/toolchain/lowering/lowering_function_context.h
@@ -45,12 +45,6 @@ class LoweringFunctionContext {
     return it->second;
   }
 
-  // This is required for handling typetype stub reference.
-  auto TryToReuseLocal(SemanticsNodeId node_id) -> bool {
-    auto it = locals_.find(node_id);
-    return it != locals_.end();
-  }
-
   // Returns a local (versus global) value for the given node in loaded state.
   // Loads will only be inserted on an as-needed basis.
   auto GetLocalLoaded(SemanticsNodeId node_id) -> llvm::Value*;
--- a/toolchain/lowering/lowering_function_context.h
+++ b/toolchain/lowering/lowering_function_context.h
@@ -50,6 +50,7 @@ class LoweringFunctionContext {
     auto it = locals_.find(node_id);
     return it != locals_.end();
   }
+
   // Returns a local (versus global) value for the given node in loaded state.
   // Loads will only be inserted on an as-needed basis.
   auto GetLocalLoaded(SemanticsNodeId node_id) -> llvm::Value*;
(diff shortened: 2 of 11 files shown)


jonmeow:
> @jonmeow could you please check it out and let me know if this could have been handled differently and more efficiently.

FYI -- the first comment in GitHub is [used for the PR description](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/project/pull_request_workflow.md#merging-a-pull-request). I'm removing the above text because it's not part of the description (it's not describing the change, and the use of `@name` in a commit message causes additional notifications on merge).

In the future, please use a separate comment for things like the above, though note you can also modify the reviewers if needed.

toolchain/lowering/lowering_function_context.h
```diff
@@ -45,6 +45,11 @@ class LoweringFunctionContext {
     return it->second;
   }
 
+  // This is required for handling typetype stub reference.
+  auto TryToReuseLocal(SemanticsNodeId node_id) -> bool {
+    auto it = locals_.find(node_id);
+    return it != locals_.end();
+  }
   // Returns a local (versus global) value for the given node in loaded state.
```
jonmeow: Please keep in mind we prefer to have a blank line between functions for style/readability.

https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace

toolchain/lowering/lowering_context.cpp
```diff
@@ -165,6 +165,23 @@ auto LoweringContext::BuildType(SemanticsNodeId node_id) -> llvm::Type* {
...
     }
+    case SemanticsNodeKind::TupleType: {
+      auto refs = semantics_ir_->GetTypeBlock(node.GetAsTupleType());
+      llvm::SmallVector<llvm::Type*> subtypes;
+      subtypes.reserve(refs.size());
+      for (auto ref_id : refs) {
+        if (ref_id == SemanticsTypeId::TypeType) {
+          // TODO: Handle TypeType. Currently using a place holder
+          // but this is just a workaround not an actual solution.
+          subtypes.push_back(llvm::Type::getInt1Ty(*llvm_context_));
```
jonmeow: What are your thoughts on having the TypeType logic shared in GetType versus here?

toolchain/lowering/lowering_handle.cpp
```diff
@@ -259,7 +273,17 @@ auto LoweringHandleStructValue(LoweringFunctionContext& context,
...
     -> void {
-  context.SetLocal(node_id, context.GetLocal(node.GetAsStubReference()));
+  if (node.type_id() == SemanticsTypeId::TypeType) {
+    if (!context.TryToReuseLocal(node_id)) {
+      llvm::Value* v = llvm::ConstantInt::get(context.builder().getInt1Ty(),
+                                              node.GetAsStubReference().index);
+      context.SetLocal(node_id, v);
+    } else {
+      context.SetLocal(node_id, context.GetLocal(node.GetAsStubReference()));
+    }
```
jonmeow: What's the purpose of the TryToReuseLocal call here, and which test covers the behavior?

toolchain/lowering/lowering_handle.cpp
```diff
@@ -259,7 +273,17 @@ auto LoweringHandleStructValue(LoweringFunctionContext& context,
 auto LoweringHandleStubReference(LoweringFunctionContext& context,
                                  SemanticsNodeId node_id, SemanticsNode node)
     -> void {
-  context.SetLocal(node_id, context.GetLocal(node.GetAsStubReference()));
+  if (node.type_id() == SemanticsTypeId::TypeType) {
```
jonmeow, line 278: Please add a TODO to revise TypeType support, similar to what you have in lowering_context.

toolchain/lowering/lowering_handle.cpp
```diff
@@ -259,7 +273,17 @@ auto LoweringHandleStructValue(LoweringFunctionContext& context,
...
+    if (!context.TryToReuseLocal(node_id)) {
+      llvm::Value* v = llvm::ConstantInt::get(context.builder().getInt1Ty(),
+                                              node.GetAsStubReference().index);
```
jonmeow: What are you expecting from passing this index, and how is it behaving in the tests?

jonmeow (commented):
This PR shows good progress, and the IR looks like what I'd expect. I have a few questions and maybe it's worth refactoring the code a little before merging.

toolchain/lowering/lowering_context.cpp
fasiddique: So far, I haven't come across any other cases that cause the TypeType to be handled except for the tuple. Hence, I added it here as this is tuple specific currently.

toolchain/lowering/lowering_handle.cpp
fasiddique: As you know this is just a workaround and as we are using `Int1Ty` as a placeholder type for TypeType until it is handled appropriately, I am passing the index here instead of 0 or 1 so that if the index is non negative it will get 1.  In the test, the store instruction is set to true if the index is non negative. Which means it will try to write to the memory location either true or false.

toolchain/lowering/lowering_handle.cpp
fasiddique: There were two different cases that made me add this function: 
When executing line#282 for typetype, it was getting carbon_check with the message "Index doesn't exist".  To remove this, I modified line#282 in the following manner `context.SetLocal(node_id, v);`. But then it gets the carbon check "Duplicate entry ".  And so I added this TryToReuseLocal function. 
The test named two_entries.carbon under the lowering/testdata/tuple folder covers this.

jonmeow: I've tried commenting out this code:

```
    // if (!context.TryToReuseLocal(node_id)) {
      llvm::Value* v = llvm::ConstantInt::get(context.builder().getInt1Ty(),
                                              node.GetAsStubReference().index);
      context.SetLocal(node_id, v);
    // } else {
    //   context.SetLocal(node_id, context.GetLocal(node.GetAsStubReference()));
    // }
```

All the tests still pass. I don't think this code is needed, because node_id is the stub reference here and the stub reference should only be in the IR once.

Can you please either remove this code, or add a test which fails without it?

toolchain/lowering/lowering_handle.cpp
jonmeow: In the tests, I see a bunch of `i1 true`, no `i1 false`. [edit: it may be that non-zero evaluates to true, rather than non-negative] Can you please either add a test which covers the behavior (and comment the code to explain the significance), or just pass 0?

toolchain/lowering/lowering_context.cpp
```diff
@@ -165,6 +165,23 @@ auto LoweringContext::BuildType(SemanticsNodeId node_id) -> llvm::Type* {
...
+        } else {
+          subtypes.push_back(GetType(ref_id));
+        }
+        CARBON_CHECK(ref_id.index < SemanticsBuiltinKind::ValidCount) << ref_id;
```
zygoloid: This `CHECK` looks incorrect to me: it's comparing a `SemanticsTypeId` to a `SemanticsBuiltinKind`, and seems to be ensuring that we only see tuple types containing builtin types, which won't be true in general. It might be reasonable to check that we don't see `SemanticsTypeId::Error` here, but it looks like that'll check-fail in `GetType`, so it seems like that case is already handled and we can just remove this `CHECK`.

toolchain/lowering/lowering_context.cpp
zygoloid: Here's a case where it happens for structs: `fn F() { {.b = {.a: i32} }; }`

I think this check should be done in `GetType`.

jonmeow (approved):
Thanks for the changes! This looks good now.

PR #3017, opened 2023-07-24
Relocate fuzzer files and the explorer_fuzzer rule.

jonmeow:
Putting fuzzer files under //testing to emphasize the testonly aspect (consolidates bazel and common subdirectories). The attributes on explorer_fuzzer are also a little skewed from what's desirable; it's been working okay, but this should still be a refinement.


chandlerc (approved):
LG

PR #3018, opened 2023-07-24
Add file_test support for specifying arguments and only checking a subset of output.

jonmeow:
This is sufficient to convert remaining toolchain tests to file_test. %s and %t are currently used. While #2978 might use %T with lit, I think that's resolving a lit-specific issue that's not necessary when TEST_TMPDIR is readily available for any necessary operations (i.e., setting the working directory). With ARGS, I think it's feasible to switch to file_test without %T.

Also adds documentation to file_test_base.h, which was starting to feel like a significant gap.

This is not yet handled by autoupdate, but I'm eyeing that next.


chandlerc (approved):
Nice!

PR #3019, opened 2023-07-24
Fix missing assignment to me_value found by msan.

jonmeow:
Looks like a small oversight in #2946. Probably missed because type access shouldn't really access the me_value.

### DIFF
--- a/explorer/interpreter/interpreter.cpp
+++ b/explorer/interpreter/interpreter.cpp
@@ -1486,26 +1486,23 @@ auto Interpreter::StepExp() -> ErrorOr<Success> {
             ElementPath::Component member(&access.member(), found_in_interface,
                                           witness);
             const Value* aggregate;
-            const Value* me_value;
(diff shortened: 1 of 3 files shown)


explorer/interpreter/interpreter.cpp
```diff
@@ -1486,26 +1486,23 @@ auto Interpreter::StepExp() -> ErrorOr<Success> {
             ElementPath::Component member(&access.member(), found_in_interface,
                                           witness);
             const Value* aggregate;
-            const Value* me_value;
+            Nonnull<const Value*> me_value = act.results()[0];
```
zygoloid: Maybe we should rename this to `lhs_value`, given that it's not the `me` value for a type access (there is no `me` value in that case)? Or make it `optional`.

jonmeow: I think it's only accessed for bound members, so switching to an optional seems better for your request.

zygoloid (approved):

PR #3021, opened 2023-07-25
Removed builtin empty tuple type

fasiddique:


toolchain/lowering/lowering_context.cpp
```diff
@@ -127,15 +127,6 @@ auto LoweringContext::BuildFunctionDefinition(SemanticsFunctionId function_id)
 
 auto LoweringContext::BuildType(SemanticsNodeId node_id) -> llvm::Type* {
   switch (node_id.index) {
-    case SemanticsBuiltinKind::EmptyTupleType.AsInt():
-      // Represent empty types as empty structs.
-      // TODO: Investigate special-casing handling of these so that they can be
-      // collectively replaced with LLVM's void, particularly around function
-      // returns. LLVM doesn't allow declaring variables with a void type, so
-      // that may require significant special casing.
```
jonmeow, line 135: Can you please move this TODO under TupleType, changing "of these" to specifically "of empty tuples"? The underlying interest in void remains.

toolchain/semantics/semantics_ir.h
```diff
@@ -217,10 +217,10 @@ class SemanticsIR {
...
     SemanticsTypeId type_id(types_.size());
     types_.push_back(node_id);
-    if (node_id == SemanticsNodeId::BuiltinEmptyTupleType) {
-      CARBON_CHECK(!empty_tuple_type_id_.is_valid());
-      empty_tuple_type_id_ = type_id;
-    }
+    // if (node_id == SemanticsNodeId::BuiltinEmptyTupleType) {
+    //   CARBON_CHECK(!empty_tuple_type_id_.is_valid());
+    //   empty_tuple_type_id_ = type_id;
+    // }
```
jonmeow: Don't forget to clean this up.

toolchain/semantics/semantics_ir.h
```diff
@@ -331,7 +331,7 @@ class SemanticsIR {
 
   // The type of the empty tuple. This is special-cased due to its use in
   // implicit function returns.
-  SemanticsTypeId empty_tuple_type_id_ = SemanticsTypeId::Invalid;
+  SemanticsTypeId empty_tuple_type_id_ = SemanticsTypeId::TypeType;
```
jonmeow: This looks incorrect; what's the type of the empty tuple? i.e., `() as type`

Note, you might want to just remove this, but it definitely needs to be correct if it remains.

jonmeow (commented):
In #2992, zygoloid had asked you to add the return_implicit.carbon test. As part of removing the builtin empty tuple, that test should now be expected to pass. Please use it to help fix correctness issues with this change.

fasiddique:
If there's no return type for functions, lowering currently assigns return type to void. However, that contradicts with the design that says that function with no return type should return an empty tuple. If we want to do that in lowering, the functions with no return type need to be changed to empty tuple type.  Or, we can do a type cast of CallInst to empty tuple. But it seems like it is not possible to typecast void CallInst to another type.  In this case, should I change the return type of function from void to typetype? 
@jonmeow @zygoloid

jonmeow:
Please note the current delta shows changes not related to this PR. You'll need to make sure you've merged or rebased trunk completely, and double-check your delta.

> If there's no return type for functions, lowering currently assigns return type to void. However, that contradicts with the design that says that function with no return type should return an empty tuple. If we want to do that in lowering, the functions with no return type need to be changed to empty tuple type. Or, we can do a type cast of CallInst to empty tuple. But it seems like it is not possible to typecast void CallInst to another type. In this case, should I change the return type of function from void to typetype?

That was necessary before @zygoloid did his `return;` addition, which you'd noticed yesterday. It should no longer be necessary. The lowered return type should be an empty tuple type, and `return;` should return an empty tuple value.

jonmeow:
Actually, I'll adjust my comment slightly -- the lowered result should probably be that an empty tuple value is created at the call site, and the call expression can just use that. The function can still return void, I think we can avoid passing a value around.

fasiddique:
> Actually, I'll adjust my comment slightly -- the lowered result should probably be that an empty tuple value is created at the call site, and the call expression can just use that. The function can still return void, I think we can avoid passing a value around.

Could you please elaborate? Are you saying that at the call site we should clone the function with a different (empty tuple) return type?  Won't this bloat the binary?

fasiddique:
> > Actually, I'll adjust my comment slightly -- the lowered result should probably be that an empty tuple value is created at the call site, and the call expression can just use that. The function can still return void, I think we can avoid passing a value around.
> 
> Could you please elaborate? Are you saying that at the call site we should clone the function with a different (empty tuple) return type? Won't this bloat the binary?

In case of assignment, I am currently checking if the call type is void, if it is void then instead of passing the void, I am passing empty tuple. Please let me know if you have any comments on this.

toolchain/lowering/testdata/function/call/implicit_empty_tuple_return.carbon
```diff
@@ -0,0 +1,29 @@
...
+// CHECK:STDOUT:   call void @MakeImplicitEmptyTuple()
+// CHECK:STDOUT:   store i1 false, ptr %var, align 1
+// CHECK:STDOUT:   ret void
+// CHECK:STDOUT: }
+
+fn MakeImplicitEmptyTuple() {
+}
+
+fn Main() {
+  var b: () = MakeImplicitEmptyTuple();
```
jonmeow: This test seems pretty similar to implicit_return.carbon in semantics. Would it be reasonable to align the test directory, name, and structure so that they match?

toolchain/lowering/lowering_context.h
```diff
@@ -38,6 +38,10 @@ class LoweringContext {
       // `type` is lowered to an empty LLVM StructType.
       return llvm::StructType::get(llvm_context());
     }
+    // Function with no return type gets void type.
+    if (type_id == SemanticsTypeId::Invalid) {
+      return llvm::Type::getVoidTy(llvm_context());
+    }
```
jonmeow: Previously this was being handled in BuildFunctionDeclaration. What are the trade-offs you're considering in moving the logic here?

toolchain/lowering/lowering_handle.cpp
```diff
@@ -22,6 +23,16 @@ auto LoweringHandleAssign(LoweringFunctionContext& context,
                           SemanticsNodeId /*node_id*/, SemanticsNode node)
     -> void {
   auto [storage_id, value_id] = node.GetAsAssign();
+  auto value_node = context.semantics_ir().GetNode(value_id);
+  if (value_node.kind() == SemanticsNodeKind::Call) {
+    auto [refs_id, function_id] = value_node.GetAsCall();
+    auto* function = context.GetFunction(function_id);
+    if (function->getReturnType()->isVoidTy()) {
+      llvm::Value* v = llvm::ConstantInt::get(context.builder().getInt1Ty(), 0);
+      context.builder().CreateStore(v, context.GetLocal(storage_id));
```
jonmeow: What's the value being created here? If you mean for this to be an empty tuple, does it structurally match an empty tuple?

toolchain/lowering/lowering_handle.cpp
```diff
@@ -22,6 +23,16 @@ auto LoweringHandleAssign(LoweringFunctionContext& context,
                           SemanticsNodeId /*node_id*/, SemanticsNode node)
     -> void {
   auto [storage_id, value_id] = node.GetAsAssign();
+  auto value_node = context.semantics_ir().GetNode(value_id);
+  if (value_node.kind() == SemanticsNodeKind::Call) {
```
jonmeow: It looks like you're special-casing the handling of:

```
fn Foo() {}
var x: () = Foo();
//        ^ the Assign
```

What happens if a call occurs inside an expression, for example:

```
fn Foo() {}
fn Bar() -> () { return Foo(); }
var x = Bar(Foo());
```

toolchain/lowering/testdata/function/call/empty_tuple.carbon
```diff
@@ -23,6 +22,7 @@
 // CHECK:STDOUT:   ret void
 // CHECK:STDOUT: }
 
+
```
jonmeow:
```suggestion
```

Please be careful about adding stray lines when editing tests.

toolchain/semantics/semantics_handle_call_expression.cpp
```diff
@@ -41,8 +42,13 @@ auto SemanticsHandleCallExpression(SemanticsContext& context,
                                          /*diagnostic=*/nullptr));
 
   // TODO: Propagate return types from callable.
+  SemanticsTypeId type_id = callable.return_type_id;
+  // For functions with no return type, setting return type to empty tuple type.
```
jonmeow:
```suggestion
  // For functions with an implicit return type, set the return type to empty tuple type.
```

Having an implicit return type is distinct from having no return type, so I want to be specific about that in comments.

jonmeow (commented):
I think this is heading in the right direction, but bear in mind that Call can happen in other places than direct variable assignment, so you might want to think about how you're placing the code in Assign.

toolchain/lowering/lowering_context.h
fasiddique: I added it here for consistency with how this function is called for getting the llvm::type for all the semantic types. However, this will cause some code to break if we are not assuming all invalid types should be void. In that case, it would be better to move it to BuildFunctionDeclaration. Please let me know your thoughts.

jonmeow: The use of invalid for void is limited to functions. We should not assume that will exist elsewhere in the future, so BuildFunctionDeclaration seems like the right place to me.

toolchain/lowering/lowering_handle.cpp
```diff
@@ -123,9 +122,16 @@ auto LoweringHandleCall(LoweringFunctionContext& context,
...
     args.push_back(context.GetLocalLoaded(ref_id));
   }
-  auto* value =
-      context.builder().CreateCall(function, args, function->getName());
-  context.SetLocal(node_id, value);
+  if (function->getReturnType()->isVoidTy()) {
+    context.builder().CreateCall(function, args);
+    // TODO call GetType with empty tuple type.
+    context.SetLocal(node_id, context.builder().CreateAlloca(
+                                  context.GetType(SemanticsTypeId::TypeType),
```
jonmeow: If you want to leave fixing this as a TODO, I think that's okay, but let's use an anonymous struct here instead of calling GetType for TypeType. This should not rely on the implementation of TypeType being equivalent to an empty tuple type.

toolchain/lowering/lowering_handle.cpp
```diff
@@ -123,9 +122,16 @@ auto LoweringHandleCall(LoweringFunctionContext& context,
   for (auto ref_id : context.semantics_ir().GetNodeBlock(refs_id)) {
     args.push_back(context.GetLocalLoaded(ref_id));
   }
-  auto* value =
-      context.builder().CreateCall(function, args, function->getName());
-  context.SetLocal(node_id, value);
+  if (function->getReturnType()->isVoidTy()) {
+    context.builder().CreateCall(function, args);
+    // TODO call GetType with empty tuple type.
```
```suggestion
    // TODO: Call GetType with an empty tuple type.
```

Just fixing towards standard TODO comment structure here.

toolchain/lowering/lowering_handle.cpp
jonmeow: Please restore the TODO. The current approach should be regarded as a workaround.

toolchain/lowering/testdata/function/call/implicit_empty_tuple_return.carbon
jonmeow: Pinging this, in case you've missed it.

toolchain/lowering/lowering_handle.cpp
jonmeow: Did you add a test to cover this? I'm looking again and not seeing one.

toolchain/lowering/lowering_handle.cpp
```diff
@@ -3,7 +3,6 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #include "toolchain/lowering/lowering_function_context.h"
-
```
jonmeow: Please restore this newline. We want to keep a blank line after includes.

toolchain/lowering/testdata/function/call/implicit_empty_tuple_return.carbon
fasiddique: Done! I am sorry that I missed that previously.

toolchain/lowering/testdata/function/call/implicit_empty_tuple_as_arg.carbon
```diff
@@ -0,0 +1,37 @@
...
+// CHECK:STDOUT:   ret void
+// CHECK:STDOUT: }
+
+
+fn Foo() {}
+
+fn Bar(a : ()) -> () { return a; }
+
+fn Main() {
+  var x : () = Bar(Foo());
```
jonmeow:
```suggestion
fn Bar(a: ()) -> () { return a; }

fn Main() {
  var x: () = Bar(Foo());
```

jonmeow (approved):

jonmeow:
I think this is about ready to submit, just one more small style issue.

PR #3022, opened 2023-07-25
Semantics and lowering for `type` keyword and type values.

zygoloid:
Also switch from modeling type values as `i1` to modeling them as an empty struct. We don't need any runtime representation for types, as there are no runtime operations on them, so an empty struct seems like a good representation.

### DIFF
--- a/toolchain/lowering/BUILD
+++ b/toolchain/lowering/BUILD
@@ -24,8 +24,11 @@ cc_library(
     srcs = [
         "lowering_context.cpp",
         "lowering_function_context.cpp",
-        "lowering_handle.cpp",
-    ],
+    ] +
+    # Glob handler files to avoid missing anyway.
+    glob([
+        "lowering_handle*.cpp",
+    ]),
     hdrs = [
         "lowering_context.h",
         "lowering_function_context.h",
--- a/toolchain/lowering/BUILD
+++ b/toolchain/lowering/BUILD
@@ -25,7 +25,7 @@ cc_library(
         "lowering_context.cpp",
         "lowering_function_context.cpp",
     ] +
-    # Glob handler files to avoid missing anyway.
+    # Glob handler files to avoid missing any.
     glob([
         "lowering_handle*.cpp",
     ]),
--- a/toolchain/lowering/lowering_context.h
+++ b/toolchain/lowering/lowering_context.h
@@ -5,6 +5,7 @@
 #ifndef CARBON_TOOLCHAIN_LOWERING_LOWERING_CONTEXT_H_
 #define CARBON_TOOLCHAIN_LOWERING_LOWERING_CONTEXT_H_
 
+#include "llvm/IR/Constants.h"
 #include "llvm/IR/LLVMContext.h"
 #include "llvm/IR/Module.h"
 #include "toolchain/semantics/semantics_ir.h"
@@ -34,12 +35,16 @@ class LoweringContext {
   auto GetType(SemanticsTypeId type_id) -> llvm::Type* {
     // InvalidType should not be passed in.
     if (type_id == SemanticsTypeId::TypeType) {
-      // TODO: Handle TypeType. Currently using a place holder
-      // as a workaround.
-      return llvm::Type::getInt1Ty(llvm_context());
+      // `type` is represented as `{}`.
+      return llvm::StructType::get(llvm_context());
     }
     CARBON_CHECK(type_id.index >= 0) << type_id;
     return types_[type_id.index];
+  }
+
+  // Returns a lowered value to use for a value of type `type`.
+  auto GetTypeValue() -> llvm::Value* {
+    return llvm::ConstantStruct::getAnon(llvm_context(), {});
   }
 
   auto llvm_context() -> llvm::LLVMContext& { return *llvm_context_; }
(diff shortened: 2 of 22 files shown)


toolchain/lowering/lowering_context.h
```diff
@@ -34,14 +35,18 @@ class LoweringContext {
...
-      return llvm::Type::getInt1Ty(llvm_context());
+      // `type` is represented as `{}`.
+      return llvm::StructType::get(llvm_context());
     }
     CARBON_CHECK(type_id.index >= 0) << type_id;
     return types_[type_id.index];
   }
 
+  // Returns a lowered value to use for a value of type `type`.
+  auto GetTypeValue() -> llvm::Value* {
```
jonmeow: Maybe GetTypeTypeAsValue or GetValueForTypeType? With GetType handling types generically, and values as a first-class concept, I think GetTypeValue is currently a little confusing.

toolchain/lowering/lowering_context.h
```diff
@@ -34,14 +35,18 @@ class LoweringContext {
   auto GetType(SemanticsTypeId type_id) -> llvm::Type* {
     // InvalidType should not be passed in.
     if (type_id == SemanticsTypeId::TypeType) {
-      // TODO: Handle TypeType. Currently using a place holder
-      // as a workaround.
-      return llvm::Type::getInt1Ty(llvm_context());
+      // `type` is represented as `{}`.
```
jonmeow:
```suggestion
      // `type` is represented as an empty StructType.
```

{} is meaningful in Carbon, and `type` is Carbon, so looking to disambiguate versus LLVM IR (because `type` _could_ be a Carbon `{}` if so chosen). Not that this phrasing is great either.

toolchain/lowering/lowering_function_context.h
```diff
@@ -40,6 +40,11 @@ class LoweringFunctionContext {
 
   // Returns a local (versus global) value for the given node.
   auto GetLocal(SemanticsNodeId node_id) -> llvm::Value* {
+    // All builtins are types, with the same empty lowered value.
+    if (node_id.index < SemanticsBuiltinKind::ValidCount) {
+      return GetTypeValue();
+    }
```
jonmeow: GetTypeValue is documented as returning the value for `type`. Should this only handle TypeType, or should the function documentation be fixed?

Consider a name like GetValueForBuiltinType if this needs to handle more than TypeType

toolchain/lowering/BUILD
```diff
@@ -24,8 +24,11 @@ cc_library(
     srcs = [
         "lowering_context.cpp",
         "lowering_function_context.cpp",
-        "lowering_handle.cpp",
-    ],
+    ] +
+    # Glob handler files to avoid missing anyway.
```
jonmeow:
```suggestion
    # Glob handler files to avoid missing any.
```

toolchain/lowering/testdata/function/call/empty_tuple.carbon
```diff
@@ -24,6 +23,7 @@
 // CHECK:STDOUT:   ret void
 // CHECK:STDOUT: }
 
+
```
jonmeow:
```suggestion
```

Looks like a mistaken newline rippling

toolchain/semantics/testdata/basics/builtin_types.carbon
```diff
@@ -55,9 +59,13 @@
...
+// CHECK:STDOUT:     node+12,
+// CHECK:STDOUT:     node+13,
+// CHECK:STDOUT:     node+14,
 // CHECK:STDOUT:   ],
 // CHECK:STDOUT: ]
 
 var test_i32: i32 = 0;
 var test_f64: f64 = 0.1;
 var test_str: String = "Test";
+var test_type: type = i32;
```
jonmeow, line 71: Maybe add a failing test, `var a: type = 1;` ?

jonmeow (commented):
Sorry! This seems like it's clearly following up on #3016, I thought your comments there had meant you were otherwise okay with the i1 approach. I hadn't realized myself that the anon struct was an option.

jonmeow:
FYI @fasiddique for visibility.

toolchain/lowering/BUILD
zygoloid: Fixed here and in the other two BUILD files I copied from :)

toolchain/lowering/lowering_context.h
zygoloid: Hopefully this phrasing is a bit clearer.

toolchain/lowering/lowering_context.h
zygoloid: Renamed to `GetTypeAsValue` as discussed.

toolchain/lowering/lowering_function_context.h
zygoloid: Per discussion, this is the value used for *any* type.

toolchain/semantics/testdata/basics/builtin_types.carbon
zygoloid: Done.

zygoloid:
> Sorry! This seems like it's clearly following up on #3016, I thought your comments there had meant you were otherwise okay with the i1 approach. I hadn't realized myself that the anon struct was an option.

No apologies needed :) I was fine with going ahead with #3016 as it was. I didn't want to block it on this concern, which seems separable from the tuple lowering.

jonmeow (approved):
Looks great!

PR #3023, opened 2023-07-25
Use precise line numbers in trace test.

zygoloid:
Don't replace line numbers with offsets from the CHECK line, because that's both hard to read and unstable when the trace output changes. Instead, include the actual line numbers. Move all the CHECKs to the end of the file so that they don't reference lines that appear after them.

This is intended to reduce the number of merge conflicts in full_trace.carbon and make the test CHECKs more readable.

### DIFF
--- a/explorer/trace_testdata/full_trace.carbon
+++ b/explorer/trace_testdata/full_trace.carbon
@@ -17,6 +17,7 @@ fn Main() -> i32 {
   return x;
 }
 
+// Place checks after code so that line numbers are stable, reducing merge conflicts.
 // AUTOUPDATE
 // CHECK:STDOUT: ********** source program **********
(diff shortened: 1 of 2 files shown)


explorer/trace_testdata/full_trace.carbon
```diff
@@ -1,7 +1,22 @@
...
+fn N.Foo(n: i32) -> i32 {
+  return n + 1;
+}
+
+fn Main() -> i32 {
+  var x: i32 = N.Foo(0);
+  return x;
+}
+
 // AUTOUPDATE
```
jonmeow, line 21:
```suggestion
// Place checks after code so that line numbers are stable, reducing merge conflicts.
// AUTOUPDATE
```

jonmeow (approved):

PR #3024, opened 2023-07-25
Implement canonicalization of Value and Element

### LINKED ISSUE #2951: Exponential run time with deeply nested tuple type
### Description of the bug:

Given the following input:

```
package EmptyIdentifier impl;

var EmptyIdentifier: ((((((((((((((((type, ), ), ), ), ), ), ), ), ), ), ), ), ), ), ), ) = (EmptyIdentifier, );
```

the Explorer takes 80 seconds (and several gigabytes of memory) to run on my machine, which seems excessive. That's with 16 pairs of parentheses in the type of `EmptyIdentifier`. Removing parentheses reveals a very clear exponential relationship: 15 parentheses takes 40 seconds, 14 parentheses takes 20 seconds, etc.

Curiously, with more than 16 parentheses the running time seems to be roughly constant at 100-105 seconds. The observable behavior changes as well: with 16 parentheses the output is:

```
RUNTIME ERROR: test16.carbon:4: could not find `EmptyIdentifier: ((((((((((((((((type))))))))))))))))`
```

But with 17 parentheses the output is:

```
COMPILATION ERROR: test17.carbon:4: type error in implicit conversion: '(((((((((((((((((type,),),),),),),),),),),),),),),),),)' is not implicitly convertible to '((((((((((((((((type,),),),),),),),),),),),),),),),)'
```

However, that may be a separate issue.

geoffromer:
See arena.h for discussion of what canonicalization means in this context. This is primarily intended to support implementing a memo table of template instantiations to resolve #2951, but could be useful for other purposes as well.

Additional changes:
- Pass `VTable` constructor parameters by pointer, to avoid the need to define `operator==` and `hash_value` for it.
- Clean up the recurring pattern of allocating identical `NamedElement`s on the stack and heap. Instead we always allocate it on the heap and pass it by pointer.
- Add `Print()` and `Dump()` to `Bindings` as a debugging convenience.

### DIFF
--- a/explorer/ast/element.h
+++ b/explorer/ast/element.h
@@ -29,6 +29,14 @@ struct NamedValue {
     return f(name, value);
   }
 
+  inline friend auto operator==(const NamedValue& lhs, const NamedValue& rhs)
+      -> bool {
+    return lhs.name == rhs.name && lhs.value == rhs.value;
+  }
+  inline friend auto hash_value(const NamedValue& value) -> llvm::hash_code {
+    return llvm::hash_combine(value.name, value.value);
+  }
+
   // The field name.
   std::string name;
 
@@ -36,14 +44,19 @@ struct NamedValue {
   Nonnull<const Value*> value;
 };
 
-// A generic member of a type.
+// A generic member of a type. This is can be a named, positional or other type
+// of member.
 //
-// This is can be a named, positional or other type of member.
+// Arena's canonicalization support is enabled for Element and all derived
+// types. As a result, all Elements must be immutable, and all their constructor
+// arguments must be copyable, equality-comparable, and hashable. See
+// Arena's documentation for details.
 class Element {
  protected:
   explicit Element(ElementKind kind) : kind_(kind) {}
 
  public:
+  using EnableCanonicalizedAllocation = void;
   virtual ~Element() = default;
 
   // Call `f` on this value, cast to its most-derived type. `R` specifies the
--- a/explorer/ast/element.h
+++ b/explorer/ast/element.h
@@ -29,6 +29,14 @@ struct NamedValue {
     return f(name, value);
   }
 
+  inline friend auto operator==(const NamedValue& lhs, const NamedValue& rhs)
+      -> bool {
+    return lhs.name == rhs.name && lhs.value == rhs.value;
+  }
+  inline friend auto hash_value(const NamedValue& value) -> llvm::hash_code {
+    return llvm::hash_combine(value.name, value.value);
+  }
+
   // The field name.
   std::string name;
 
--- a/explorer/ast/element.h
+++ b/explorer/ast/element.h
@@ -11,6 +11,7 @@
 
 #include "common/ostream.h"
 #include "explorer/ast/ast_rtti.h"
+#include "explorer/common/decompose.h"
 #include "explorer/common/nonnull.h"
 #include "llvm/ADT/PointerUnion.h"
 
@@ -20,21 +21,13 @@ class Declaration;
 class Value;
 
 // A NamedValue represents a value with a name, such as a single struct field.
-struct NamedValue {
+struct NamedValue : HashFromDecompose<NamedValue> {
   NamedValue(std::string name, Nonnull<const Value*> value)
       : name(std::move(name)), value(value) {}
 
   template <typename F>
   auto Decompose(F f) const {
     return f(name, value);
-  }
-
-  inline friend auto operator==(const NamedValue& lhs, const NamedValue& rhs)
-      -> bool {
-    return lhs.name == rhs.name && lhs.value == rhs.value;
-  }
-  inline friend auto hash_value(const NamedValue& value) -> llvm::hash_code {
-    return llvm::hash_combine(value.name, value.value);
   }
 
   // The field name.
--- a/explorer/ast/element.h
+++ b/explorer/ast/element.h
@@ -29,14 +29,6 @@ struct NamedValue {
     return f(name, value);
   }
 
-  inline friend auto operator==(const NamedValue& lhs, const NamedValue& rhs)
-      -> bool {
-    return lhs.name == rhs.name && lhs.value == rhs.value;
-  }
-  inline friend auto hash_value(const NamedValue& value) -> llvm::hash_code {
-    return llvm::hash_combine(value.name, value.value);
-  }
-
   // The field name.
   std::string name;
 
--- a/explorer/ast/value.h
+++ b/explorer/ast/value.h
@@ -739,11 +739,8 @@ class FunctionType : public Value {
         impl_bindings_(std::move(impl_bindings)),
         is_initializing_(is_initializing) {}
 
-  struct ExceptSelf {
-    friend auto operator==(ExceptSelf, ExceptSelf) -> bool { return true; }
-    friend auto hash_value(ExceptSelf) -> llvm::hash_code {
-      return llvm::hash_combine(0);
-    }
+  struct ExceptSelf : public HashFromDecompose<ExceptSelf> {
+    template <typename F> auto Decompose(F f) const { return f(); }
   };
   FunctionType(ExceptSelf, const FunctionType* clone)
       : FunctionType(std::nullopt, clone->parameters_,
--- a/explorer/ast/value.h
+++ b/explorer/ast/value.h
@@ -41,6 +41,16 @@ struct AllocateTrait {
 using VTable =
     llvm::StringMap<std::pair<Nonnull<const CallableDeclaration*>, int>>;
 
+// Returns a pointer to an empty VTable that will never be deallocated.
+//
+// Using this instead of `new VTable()` avoids unnecessary allocations, and
+// takes better advantage of Arena canonicalization when a VTable pointer is
+// used as a constructor argument.
+inline auto EmptyVTable() -> Nonnull<const VTable*> {
+  static Nonnull<const VTable*> result = new VTable();
+  return result;
+}
+
 // Abstract base class of all AST nodes representing values.
 //
 // Value and its derived classes support LLVM-style RTTI, including
@@ -49,8 +59,14 @@ using VTable =
 // every concrete derived class must have a corresponding enumerator
 // in `Kind`; see https://llvm.org/docs/HowToSetUpLLVMStyleRTTI.html for
 // details.
+//
+// Arena's canonicalization support is enabled for Value and all derived types.
+// As a result, all Values must be immutable, and all their constructor
+// arguments must be copyable, equality-comparable, and hashable. See
+// Arena's documentation for details.
 class Value {
  public:
+  using EnableCanonicalizedAllocation = void;
   enum class Kind {
 #define CARBON_VALUE_KIND(kind) kind,
 #include "explorer/ast/value_kinds.def"
@@ -684,6 +700,15 @@ class FunctionType : public Value {
       return f(index, binding);
     }
 
+    inline friend auto operator==(const GenericParameter& lhs,
+                                  const GenericParameter& rhs) -> bool {
+      return lhs.index == rhs.index && lhs.binding == rhs.binding;
+    }
+    inline friend auto hash_value(const GenericParameter& parameter)
+        -> llvm::hash_code {
+      return llvm::hash_combine(parameter.index, parameter.binding);
+    }
+
     size_t index;
     Nonnull<const GenericBinding*> binding;
   };
@@ -693,6 +718,14 @@ class FunctionType : public Value {
     template <typename F>
     auto Decompose(F f) const {
       return f(addr_self, self_type);
+    }
+
+    inline friend auto operator==(const MethodSelf& lhs, const MethodSelf& rhs)
+        -> bool {
+      return lhs.addr_self == rhs.addr_self && lhs.self_type == rhs.self_type;
+    }
+    inline friend auto hash_value(const MethodSelf& self) -> llvm::hash_code {
+      return llvm::hash_combine(self.addr_self, self.self_type);
     }
 
     // True if `self` parameter uses an `addr` pattern.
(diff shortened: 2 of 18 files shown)


explorer/common/arena.h
```diff
@@ -61,35 +119,162 @@ class Arena {
...
-    explicit ArenaEntryTyped(Args&&... args)
-        : instance_(std::forward<Args>(args)...) {}
+  // Manages allocations in an arena for destruction at shutdown.
+  std::vector<std::unique_ptr<ArenaEntry>> arena_;
+  int64_t allocated_ = 0;
 
-    template <typename... Args>
-    explicit ArenaEntryTyped(WriteAddressHelper, Args&&... args)
-        : ArenaEntryTyped(std::forward<Args>(args)...) {}
+  // Hash functor implemented in terms of hash_value (see llvm/ADT/Hashing.h)
```
jonmeow:
```suggestion
  // Hash functor implemented in terms of hash_value (see llvm/ADT/Hashing.h).
```

Minor, punctuation.

explorer/common/arena.h
```diff
@@ -61,35 +119,162 @@ class Arena {
...
+  struct LlvmHasher {
```
jonmeow, line 112: Shouldn't types be declared first in declaration order, before `arena_`?

explorer/common/arena.h
```diff
@@ -61,35 +119,162 @@ class Arena {
...
 
-    template <typename U, typename... Args>
-    explicit ArenaEntryTyped(WriteAddressTo<U> write_address, Args&&... args)
-        : ArenaEntryTyped(
-              (*write_address.target = &instance_, WriteAddressHelper{}),
-              std::forward<Args>(args)...) {}
+  // Factory metafunction for globally unique type IDs.
+  // &TypeId<T>::id == &TypeId<U>::id if and only if std::is_same_v<T,U>.
+  //
+  // Inspired by llvm::Any::TypeId.
```
jonmeow, line 123: Just as a possible alternative, had you considered using ast_rtti's enum to provide a unique index for the types?

explorer/common/arena.h
```diff
@@ -61,35 +119,162 @@ class Arena {
...
+  template <typename T>
+  struct TypeId {
+    static char id;
```
jonmeow, line 127:
```suggestion
    // This is only used for an address to compare; the value is unimportant.
    static char id;
```

This had been stumping me what was significant about `1`, can you add a brief explanatory comment?

explorer/common/arena.h
```diff
@@ -61,35 +119,162 @@ class Arena {
 
   // Templated destruction of a pointer.
   template <typename T>
-  class ArenaEntryTyped : public ArenaEntry {
-   public:
-    struct WriteAddressHelper {};
+  class ArenaEntryTyped;
```
jonmeow, line 109: Given the reference to llvm::Any for TypeId, I was looking and now wondering -- can this be replaced with llvm::Any?

explorer/common/arena.h
```diff
@@ -61,35 +119,162 @@ class Arena {
...
+  struct TypeId {
+    static char id;
+  };
+
+  // A canonicalization table maps a tuple of constructor argument values to
+  // a non-null pointer to a T object constructed with those arguments.
+  template <typename T, typename... Args>
+  using CanonicalizationTable =
+      std::unordered_map<std::tuple<ArgKeyType<Args>...>, Nonnull<const T*>,
+                         LlvmHasher>;
```
jonmeow, line 135: Did you consider using llvm::FoldingSet for this? We're using it in the toolchain now for type deduplication (in SemanticsContext), and I'm wondering if it'd be a reasonable choice here too, although the approach would be a bit different for how hashing is done versus if you were to build a FoldingSetNodeID.

explorer/common/arena.h
```diff
@@ -61,35 +119,162 @@ class Arena {
...
 
-    auto Instance() -> Nonnull<T*> { return Nonnull<T*>(&instance_); }
+  // Maps a CanonicalizationTable type to a unique instance of that type for
+  // this arena. For a key equal to &TypeId<T>::id for some T, the corresponding
+  // value contains a T*.
+  std::map<char*, std::any> canonical_tables_;
+};
+
+// Implementation details only below here -------------------------------------
```
jonmeow: Minor, but maybe match prelude.carbon's comment barriers (mainly niggling me for the missing punctuation)
```suggestion
// ---------------------------------------
// Implementation details only below here.
// ---------------------------------------
```

jonmeow (commented):
This seems fine to me, although I have a couple questions about the approach. I think you were discussing this with @zygoloid though, so maybe he's considered tradeoffs?

jonmeow:
Sorry, I meant to also ask -- what's the testing plan like for this, to ensure it's working as intended? I'm a little surprised this doesn't affect trace output already.

explorer/common/arena.h
geoffromer: I don't think so; `llvm::Any` can only hold values of types that are copy-constructible and copy-assignable.

explorer/common/arena.h
geoffromer: It's not a good tool for this job. This needs to support any type that can be used as a constructor argument for a canonicalized type. That includes a lot of types that we don't own (e.g. pointers, strings, ints,`std::vector<T>` for various `T`s), and I think it doesn't include any of the types in ast_rtti's enum (because those are all non-copyable).

explorer/common/arena.h
geoffromer: I wasn't aware of `FoldingSet`, but I did consider an approach along those lines. The main advantage I see is that it gives you fully precise deduplication, so that you never have two objects that are structurally equal but not identical (unless they were allocated in different Arenas). But I see two big drawbacks:

- It would be a much more invasive change: we'd have to substantially modify every subclass of `Value` and `Element`, and rewrite every line that allocates one.
- It seems like it has more room for false positive bugs (where two nodes are treated as duplicates even though they're observably different, e.g. if a field is left out of the `Profile` method). That can still happen with the approach in this PR if `operator==` or `hash_value` omits a field, but I'm less worried about that because the types that need those operations tend to be easier to reason about than the ones that would need `Profile` methods (and I think there's fewer of them).

explorer/common/arena.h
geoffromer: Done.

explorer/common/arena.h
```diff
@@ -26,32 +54,62 @@ class Arena {
...
+  void New(WriteAddressTo<U> addr, Args&&... args);
 
   auto allocated() -> int64_t { return allocated_; }
 
  private:
+  // Allocates an object in the arena. Unlike New, this will always allocate
+  // and construct a new object.
+  template <
+      typename T, typename... Args,
+      typename std::enable_if_t<std::is_constructible_v<T, Args...>>* = nullptr>
```
zygoloid: Do we need the constraint here? There's no other overload to call and we don't need this to be SFINAE-friendly.

explorer/ast/element.h
```diff
@@ -29,21 +29,34 @@ struct NamedValue {
     return f(name, value);
   }
 
+  inline friend auto operator==(const NamedValue& lhs, const NamedValue& rhs)
+      -> bool {
+    return lhs.name == rhs.name && lhs.value == rhs.value;
+  }
+  inline friend auto hash_value(const NamedValue& value) -> llvm::hash_code {
+    return llvm::hash_combine(value.name, value.value);
+  }
```
zygoloid: You're providing this pair of functions for various types that are `Decompose`able. Perhaps instead we should be providing blanket support for `==` and hashing for all `Decompose`able types, and adding `Decompose` support to the other cases in this patch where you are currently adding an explicit `==` / `hash_value`?

geoffromer (commented):
> Sorry, I meant to also ask -- what's the testing plan like for this, to ensure it's working as intended? I'm a little surprised this doesn't affect trace output already.

I think the testing plan is the same as it usually is for Explorer: lots of end-to-end tests. I've added some unit tests for `Arena`, but I don't think they'll be very interesting; at a unit level this feature is pretty straightforward.

How would you expect this to affect the trace output?

explorer/ast/element.h
geoffromer: Possibly. I see a few concerns with that approach:
- Where do we put that blanket support, so that it's available whenever any `Decompose`able type is?
- We'd need to use a wrapper type to work around the absence of a `hash_value` overload for `std::vector`, and it would need to be a slightly different wrapper than the one `Arena` uses (because here we want to wrap a reference rather than a copy).
- Those overloads seem like they will be an attractive nuisance in a lot of cases. If you're applying `==` to two `Value`s, you have either a performance bug (you should be comparing their addresses instead) or a correctness bug (you should be calling `ValueEqual` instead), so it seems good that such code currently doesn't compile. More broadly, `==` and hashing are typically not a good fit for types where identity is meaningful, so it would be kind of awkward to add those operations to `Value` as part of a change that makes `Value` identity more meaningful.

explorer/common/arena_test.cpp
```diff
@@ -0,0 +1,78 @@
...
+}
+
+struct CanonicalizedDummy {
+  explicit CanonicalizedDummy(int) {}
+  explicit CanonicalizedDummy(int*) {}
+  explicit CanonicalizedDummy(int, int*) {}
+  using EnableCanonicalizedAllocation = void;
+};
+
+TEST(ArenaTest, Canonicalization) {
```
jonmeow: One last thought on these tests, you might consider std::vector and std::nullopt variants to consider that functionality, since they're special-cased in Arena code.

jonmeow (approved):
This looks good to me. It would be good to make sure that @zygoloid is fine with the one comment, but also maybe further work there can be incremental since IIUC this is a step to more work on #2951.

> I think the testing plan is the same as it usually is for Explorer: lots of end-to-end tests. I've added some unit tests for `Arena`, but I don't think they'll be very interesting; at a unit level this feature is pretty straightforward.

Thanks, I like the unit test! I was partly also asking for #2951 -- but I'm expecting there that there'll be some test to cover the template stuff, even if just by regression testing the current exponential behavior.

> How would you expect this to affect the trace output?

I'd been thinking some of the canonicalization might result in different output, but I guess not.

explorer/ast/element.h
zygoloid: > * Where do we put that blanket support, so that it's available whenever any `Decompose`able type is?

Perhaps we can put it into `LlvmHasher` (and add an equality comparison function object for the `==` side of things). I think that would also address your other concerns: we can put the vector hashing support in `LlvmHasher`, and not add these `operator==` / `hash_value` functions that may be an attractive nuisance.

geoffromer: OK, I've done that, but it's not pretty. The code could probably be cleaned up to some extent, but a fair amount of the extra complexity seems inherent. In addition to the complexity of the metaprogramming itself, we have to reimplement hashing and equality for all the `std` templates that we want to support, in order to ensure that they recurse into our helpers rather than into `hash_value` and `operator==`, because the latter don't support `Decompose`able types.

I would prefer to revert 4362ff1bf6 (the commit implementing this suggestion) but if you strongly prefer it, I could clean it up for merging.

Alternatively, we could probably provide a CRTP helper that implements `operator==` and `hash_value` in terms of `Decompose`. That would require each type to opt in, but it would eliminate most of the per-type boilerplate while avoiding all the extra complexity in `Arena`.

geoffromer: I've switched to the CRTP approach, as discussed on Discord.

explorer/ast/value.h
```diff
@@ -723,11 +739,17 @@ class FunctionType : public Value {
...
-  FunctionType(ExceptSelf, const FunctionType& clone)
-      : FunctionType(std::nullopt, clone.parameters_, clone.generic_parameters_,
-                     clone.return_type_, clone.deduced_bindings_,
-                     clone.impl_bindings_, clone.is_initializing_) {}
+  struct ExceptSelf {
+    friend auto operator==(ExceptSelf, ExceptSelf) -> bool { return true; }
+    friend auto hash_value(ExceptSelf) -> llvm::hash_code {
+      return llvm::hash_combine(0);
+    }
+  };
```
zygoloid: Optionally: make this type `Decompose`able
```suggestion
  struct ExceptSelf : public HashFromDecompose<ExceptSelf> {
    template <typename F> auto Decompose(F f) const { return f(); }
  };
```

explorer/common/arena.h
```diff
@@ -61,35 +106,175 @@ class Arena {
...
+    friend auto hash_value(const VectorProxy& v) {
+      return llvm::hash_combine(
+          llvm::hash_combine_range(v.vec_.begin(), v.vec_.end()),
+          v.vec_.size());
+    }
+
+   private:
+    std::vector<T> vec_;
+  };
 };
```
zygoloid, line 190: Do we still need this?

zygoloid (approved):
Thanks, looks good to me.

We would probably benefit from making some more of these types decomposable -- notably, `Address` and `ElementPath` seem like good candidates, and I suspect that we're not properly doing template instantiation involving pointer values because we don't recursively decompose and rebuild them, and the `ElementPath` can recursively refer to things we have transformed. But I don't think we should do that as part of this PR.

Adding the base class had some complications I didn't anticipate with rebuilding values. I wonder if we could use a member `using` instead, like you do for `EnableCanonicalizedAllocation`? But I'm happy for you to land this as-is and potentially consider that as a follow-up.

explorer/common/arena.h
geoffromer: I'm afraid so. This is for handling cases where a `vector<T>` is directly used as a constructor argument, so the machinery in `HashFromDecompose` doesn't have a chance to kick in. Even if it did, we wouldn't want it to, because here we need to make a deep copy of the `vector` argument so that we can retain it in the cache.

PR #3025, opened 2023-07-25
Fix `class_value_ptr` invariant bug found by fuzzer

josh11b:
Invariant is `*class_value_ptr == this` when the static type matches the dynamic type.

### DIFF
--- a/explorer/ast/value.cpp
+++ b/explorer/ast/value.cpp
@@ -19,7 +19,6 @@
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/StringExtras.h"
 #include "llvm/Support/Casting.h"
-#include "llvm/Support/Error.h"
 
 namespace Carbon {
 
@@ -128,6 +127,7 @@ NominalClassValue::NominalClassValue(
       inits_(inits),
       base_(base),
       class_value_ptr_(class_value_ptr) {
+  CARBON_CHECK(!base || (*base)->class_value_ptr() == class_value_ptr);
   // Update ancestors's class value to point to latest child.
   *class_value_ptr_ = this;
 }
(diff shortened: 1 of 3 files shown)


explorer/ast/value.cpp
```diff
@@ -376,7 +376,8 @@ static auto SetFieldImpl(
                                     path_end, field_value, source_loc);
           inits.ok()) {
         return arena->New<NominalClassValue>(
-            &object.type(), *inits, object.base(), object.class_value_ptr());
+            &object.type(), *inits, object.base(),
```
zygoloid: To maintain the invariant, do we also need to recreate all the base classes with the new pointer?

josh11b: Maybe? Do you have a suggestion? I don't really know what is going on in this piece of code.

josh11b:
Added a comment to `NominalClassValue` in `value.h` to give more breadcrumbs to the next person looking at this.

explorer/ast/value.cpp
zygoloid: The overall task here is to build a new value that is like `value` but has the element described by the `path` replaced by `field_value`. In this case, we're changing a field of a derived class but not changing the  base class, but we still need to update the base class (and its base class and so on) to hold the new pointer to the new derivedmost object.

github doesn't let me provide a suggested edit here, but I think we want something like this:

```
auto* class_value_ptr = arena->New<const NominalClassValue*>();
std::vector<const NominalClassValue*> base_path;
for (auto base = object.base(); base; base = (*base)->base()) {
  base_path.push_back(*base);
}
std::optional<Nonnull<const NominalClassValue*>> base;
for (auto* base_path_elem : llvm::reversed(base_path)) {
  base = arena->New<NominalClassValue>(
      base_path_elem->type(), base_path_elem->inits(), base, class_value_ptr);
}
return arena->New<NominalClassValue>(
    &object.type(), *inits, base, class_value_ptr);
```

(Longer term I'd love to get rid of `class_value_ptr` entirely, but I think we need to finish the expression category work first.)

explorer/ast/value.h
```diff
@@ -400,7 +400,8 @@ class NominalClassValue : public Value {
   // NominalClassValue*, that must be common to all NominalClassValue of the
   // same object. The pointee is updated, when `NominalClassValue`s are
   // constructed, to point to the `NominalClassValue` corresponding to the
-  // child-most class type.
+  // child-most class type. Sets *class_value_ptr = this, which corresponds to
+  // the static type of the value matching its dynamic type.
   NominalClassValue(Nonnull<const Value*> type, Nonnull<const Value*> inits,
                     std::optional<Nonnull<const NominalClassValue*>> base,
                     Nonnull<const NominalClassValue** const> class_value_ptr);
```
zygoloid, line 407: Can we add an assert in this constructor that `!base || (*base)->class_value_ptr() == class_value_ptr`?

explorer/ast/value.cpp
josh11b: Great! With this fix, and the one other change from offline, the tests pass with the new assertion added.

explorer/ast/value.h
josh11b: With that change, I get a new failure. It is fixed by the changes I've included in the same commit.

zygoloid (approved):

PR #3026, opened 2023-07-26
Parsing for pointer types and pointer operators.

zygoloid:
This provides parsing support for the functionality added in #2006.


toolchain/lexer/testdata/keywords.carbon
```diff
@@ -34,100 +34,102 @@ case
 choice
 // CHECK:STDOUT: { index: 13, kind:               'Class', line: {{ *}}[[@LINE+1]], column:  1, indent: 1, spelling: 'class', has_trailing_space: true },
 class
-// CHECK:STDOUT: { index: 14, kind:          'Constraint', line: {{ *}}[[@LINE+1]], column:  1, indent: 1, spelling: 'constraint', has_trailing_space: true },
+// CHECK:STDOUT: { index: 14, kind:               'Const', line: {{ *}}[[@LINE+1]], column:  1, indent: 1, spelling: 'const', has_trailing_space: true },
+const
```
jonmeow, line 38: These are a lot of tests of CARBON_KEYWORD_TOKEN, aren't they overlapping?

toolchain/parser/parser_handle_expression.cpp
```diff
@@ -9,6 +9,8 @@ namespace Carbon {
 CARBON_DIAGNOSTIC(
     OperatorRequiresParentheses, Error,
     "Parentheses are required to disambiguate operator precedence.");
+CARBON_DIAGNOSTIC(UnaryOperatorRequiresParentheses, Error,
+                  "Parentheses are required around this unary operator.");
```
jonmeow: Should this mention what the operator is in the message, rather than just "this"? Also, why doesn't this explain why parentheses are required similar to the above diagnostic?

toolchain/parser/parser_handle_expression.cpp
```diff
@@ -21,7 +23,8 @@ auto ParserHandleExpression(ParserContext& context) -> void {
         OperatorPriority::RightFirst) {
       // The precedence rules don't permit this prefix operator in this
       // context. Diagnose this, but carry on and parse it anyway.
-      context.emitter().Emit(*context.position(), OperatorRequiresParentheses);
+      context.emitter().Emit(*context.position(),
+                             UnaryOperatorRequiresParentheses);
```
jonmeow: It looks like you're only using this diagnostic in one place, can it be next to the use?

toolchain/parser/parser_handle_period.cpp
```diff
@@ -39,15 +42,23 @@ auto ParserHandlePeriod(ParserContext& context, ParseNodeKind node_kind)
 }
 
 auto ParserHandlePeriodAsDeclaration(ParserContext& context) -> void {
-  ParserHandlePeriod(context, ParseNodeKind::QualifiedDeclaration);
+  ParserHandlePeriodOrArrow(context, ParseNodeKind::QualifiedDeclaration,
+                            false);
```
jonmeow:
```suggestion
                            /*is_arrow=*/false);
```

Maybe `/*is_arrow=*/`? I know people can look up to the function, but slight preference for marking bools just as a habit.

toolchain/parser/parser_handle_period.cpp
```diff
@@ -39,15 +42,23 @@ auto ParserHandlePeriod(ParserContext& context, ParseNodeKind node_kind)
...
 }
 
 auto ParserHandlePeriodAsExpression(ParserContext& context) -> void {
-  ParserHandlePeriod(context, ParseNodeKind::MemberAccessExpression);
+  ParserHandlePeriodOrArrow(context, ParseNodeKind::MemberAccessExpression,
+                            false);
```
jonmeow:
```suggestion
                            /*is_arrow=*/false);
```

toolchain/parser/parser_handle_period.cpp
```diff
@@ -39,15 +42,23 @@ auto ParserHandlePeriod(ParserContext& context, ParseNodeKind node_kind)
...
 }
 
 auto ParserHandlePeriodAsStruct(ParserContext& context) -> void {
-  ParserHandlePeriod(context, ParseNodeKind::StructFieldDesignator);
+  ParserHandlePeriodOrArrow(context, ParseNodeKind::StructFieldDesignator,
+                            false);
```
jonmeow:
```suggestion
                            /*is_arrow=*/false);
```

toolchain/parser/parser_handle_period.cpp
```diff
@@ -39,15 +42,23 @@ auto ParserHandlePeriod(ParserContext& context, ParseNodeKind node_kind)
...
+}
+
+auto ParserHandleArrowExpression(ParserContext& context) -> void {
+  ParserHandlePeriodOrArrow(context,
+                            ParseNodeKind::PointerMemberAccessExpression, true);
```
jonmeow:
```suggestion
                            ParseNodeKind::PointerMemberAccessExpression,
                            /*is_arrow=*/true);
```

toolchain/parser/parser_handle_period.cpp
```diff
@@ -6,21 +6,24 @@
 
 namespace Carbon {
 
-// Handles PeriodAs variants.
+// Handles PeriodAs variants and ArrowExpression.
 // TODO: This currently only supports identifiers on the rhs, but will in the
 // future need to handle things like `object.(Interface.member)` for qualifiers.
-auto ParserHandlePeriod(ParserContext& context, ParseNodeKind node_kind)
-    -> void {
+auto ParserHandlePeriodOrArrow(ParserContext& context, ParseNodeKind node_kind,
```
jonmeow:
```suggestion
static auto ParserHandlePeriodOrArrow(ParserContext& context, ParseNodeKind node_kind,
```

Sorry, pre-existing issue, but probably good to fix.

toolchain/parser/testdata/operators/prefix_repeat.carbon
```diff
@@ -9,14 +9,16 @@
...
+// CHECK:STDOUT:       {kind: 'PrefixOperator', text: '*', subtree_size: 4},
+// CHECK:STDOUT:     {kind: 'PrefixOperator', text: '*', subtree_size: 5},
+// CHECK:STDOUT:   {kind: 'ExpressionStatement', text: ';', subtree_size: 6},
+// CHECK:STDOUT: {kind: 'FunctionDefinition', text: '}', subtree_size: 12},
 // CHECK:STDOUT: {kind: 'FileEnd', text: ''},
 // CHECK:STDOUT: ]
 
 fn F() {
-  ++++n;
+  ***&n;
```
jonmeow, line 23: Why the switch in operators here? Just because they're shorter?

toolchain/parser/testdata/pointer/fail_pointer_type_in_expression.carbon
```diff
@@ -0,0 +1,75 @@
...
+// CHECK:STDOUT: {kind: 'FileEnd', text: ''},
+// CHECK:STDOUT: ]
+
+fn F() -> i32 {
+  // CHECK:STDERR: fail_pointer_type_in_expression.carbon:[[@LINE+6]]:17: Parentheses are required to disambiguate operator precedence.
+  // CHECK:STDERR:   return 3 * i32* * 4;
+  // CHECK:STDERR:                 ^
+  // CHECK:STDERR: fail_pointer_type_in_expression.carbon:[[@LINE+3]]:19: Parentheses are required to disambiguate operator precedence.
+  // CHECK:STDERR:   return 3 * i32* * 4;
+  // CHECK:STDERR:                   ^
```
jonmeow, line 63: This seems a little undesirable to repeat. Maybe we should have a TODO to better propagate errors in expression precedence?

jonmeow (approved):

toolchain/parser/parser_handle_expression.cpp
```diff
@@ -9,6 +9,8 @@ namespace Carbon {
 CARBON_DIAGNOSTIC(
     OperatorRequiresParentheses, Error,
     "Parentheses are required to disambiguate operator precedence.");
```
jonmeow: With this change, is this diagnostic now only used in one place such that it could be relocated there too?

toolchain/lexer/testdata/keywords.carbon
zygoloid: There's one test here for each keyword; what would you like to see instead?

toolchain/parser/parser_handle_expression.cpp
zygoloid: Done.

toolchain/parser/parser_handle_expression.cpp
zygoloid: Added operator name.

This doesn't mention why parentheses are required because the reason for the error is simply that we require parentheses. The "to disambiguate" text we used to produce in this case was wrong -- there's no ambiguity, we just require parentheses anyway.

toolchain/parser/parser_handle_expression.cpp
zygoloid: Done.

toolchain/parser/testdata/operators/prefix_repeat.carbon
zygoloid: The original testcase is not valid, per the design -- `++`, `--`, and assignment are only allowed as complete statements, not as subexpressions -- and I wanted to use an example here that will remain correct once we update toolchain to properly handle `++` and friends.

toolchain/parser/testdata/pointer/fail_pointer_type_in_expression.carbon
zygoloid: Done.

toolchain/lexer/testdata/keywords.carbon
jonmeow: Per discussion, more as a todo, just looking for something less churn-y. We could drop the index, we might not want to just ignore it in tests completely if it remains. Could also move this to a programmatic test that looped through all keywords. Could also just test a couple sample keywords, and leave the rest of keyword testing to downstream.

As you noted in person: it's not like this is exhaustive, it doesn't test all possible identifiers. :)

But more seriously, there's also no guarantee it tests all keywords, so might be a little misleading from that perspective.

PR #3028, opened 2023-07-26
Set `me_value` even when accessing a type

geoffromer:


geoffromer:
No new test coverage is needed; several of our existing tests would have caught the issue if we ran them under MSan.

jonmeow:
I'd considered this initially on #3019 but @zygoloid argued me out of it. Note that if me_value is unconditionally set, then the std::optional should be reverted.

jonmeow:
Also, can you just confirm that the msan issue is still seen after #3019?

geoffromer:
Yep, this is redundant with #3019, which fixes the msan issue.

PR #3030, opened 2023-07-26
Introduce flag handling into file_test.

jonmeow:
I'm integrating absl flag support with a few thoughts here...

1. It simplifies the handling of files in file_test.
    - Removes the need for a separate subset target.
3. Looking forward, I'm planning to add a flag to allow for autoupdate of golden files.
4. In explorer, there's been confusion about having file_test run tests "twice" so hopefully it's clearer when it's a separate target with a different flag on the target.

### DIFF
--- a/explorer/file_test.cpp
+++ b/explorer/file_test.cpp
@@ -2,45 +2,23 @@
 // Exceptions. See /LICENSE for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+#include "absl/flags/flag.h"
 #include "explorer/parse_and_execute/parse_and_execute.h"
 #include "testing/file_test/file_test_base.h"
 #include "testing/util/test_raw_ostream.h"
+
+ABSL_FLAG(bool, trace, false,
+          "Set to true to run tests with tracing enabled, even if they don't "
+          "otherwise specify it. This does not result in checking trace output "
+          "contents; it essentially only verifies there's not a crash bug.");
 
 namespace Carbon::Testing {
 namespace {
 
 class ParseAndExecuteTestFile : public FileTestBase {
  public:
-  explicit ParseAndExecuteTestFile(const std::filesystem::path& path,
-                                   bool trace)
-      : FileTestBase(path), trace_(trace) {}
-
-  auto SetUp() -> void override {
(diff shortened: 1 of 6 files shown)


explorer/file_test.cpp
```diff
@@ -2,45 +2,20 @@
 // Exceptions. See /LICENSE for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+#include "absl/flags/flag.h"
```
geoffromer, line 5: We seem to have a convention of using angle-bracket includes for third-party code other than llvm (see e.g. our usage of gtest). Should we follow that here?

explorer/file_test.cpp
```diff
@@ -55,11 +30,15 @@ class ParseAndExecuteTestFile : public FileTestBase {
       return false;
     }
 
+    bool check_trace_output =
+        path().string().find("/trace_testdata/") != std::string::npos;
```
geoffromer, line 38: What does `check_trace_output` mean, and why is it different from `FLAGS_trace`?

explorer/file_test.cpp
jonmeow: I think the convention is to follow documented include usage since that's what we should expect systems to support.

The intent is to follow system documentation on how to include, for example `#include "absl/strings/str_join.h"` per https://abseil.io/docs/cpp/quickstart#creating-and-running-a-binary

For bazel libraries, we have `#include "tools/cpp/runfiles/runfiles.h"` echoing the above style.

For protobuf libraries, we do angle brackets, but you can see this style in https://protobuf.dev/reference/cpp/arenas/ so we are following the documentation.

I see gtest also suggests quote includes like `#include "gtest/gtest.h"` in https://google.github.io/googletest/primer.html#writing-the-main-function. I don't know why we're using angle brackets, it may have been a mistake; I believe this should be migrated, not all the others.

explorer/file_test.cpp
jonmeow: Trace output is only checked for a few tests. By default, even with FLAGS_trace, it is not checked. Also, FLAGS_trace is used independent of /trace_testdata/.

geoffromer (approved):

PR #3032, opened 2023-07-27
Change explorer's file_test to support argument passing to main.

jonmeow:
This shifts explorer's file_test to use ExplorerMain in order to be able to pass arguments similar to how the command line would. It also means that various output wrapping done by the command line is shared, no longer copied by file_test.

In order to help make this work, I plugged vfs::FileSystem into explorer, similar to how we're doing this on the toolchain side (might try to share more code later). I was having trouble getting an overlay working, I think due to how paths are specified -- but due to the issues in fixing this, I'm just loading the prelude into the InMemoryFilesystem for now.

Under this approach, I'm unifying more of the file versus string handling. I think we should shift towards an approach where, like toolchain code, anyone trying to use Explorer should use a vfs implementation to supply code. This should reduce the number of distinct code paths which we're maintaining/testing.

Short-term, this allows the trace testdata to be merged into file_test with less special casing, using ARGS:. Long-term, it means that the file_test knows the ARGS to pass to generate checks to match against, which is the direction I'm planning for autoupdate.

### DIFF
--- a/explorer/fuzzing/ast_to_proto_main.cpp
+++ b/explorer/fuzzing/ast_to_proto_main.cpp
@@ -31,15 +31,10 @@ auto Main(int argc, char** argv) -> ErrorOr<Success> {
     return Error("Argument must be a file.");
   }
 
-  // Read the input file.
-  std::ifstream proto_file(argv[1]);
-  std::stringstream buffer;
-  buffer << proto_file.rdbuf();
-  proto_file.close();
-
   Arena arena;
-  const ErrorOr<AST> ast = Parse(&arena, argv[1], FileKind::Main,
-                                 /*parser_debug=*/false);
+  const ErrorOr<AST> ast =
+      Parse(*llvm::vfs::getRealFileSystem(), &arena, argv[1], FileKind::Main,
+            /*parser_debug=*/false);
   if (!ast.ok()) {
     return ErrorBuilder() << "Parsing failed: " << ast.error().message();
   }
--- a/explorer/main.cpp
+++ b/explorer/main.cpp
@@ -17,6 +17,7 @@
 #include "common/error.h"
 #include "explorer/common/trace_stream.h"
 #include "explorer/parse_and_execute/parse_and_execute.h"
+#include "llvm/ADT/ScopeExit.h"
 #include "llvm/ADT/SmallString.h"
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/FileSystem.h"
@@ -30,9 +31,10 @@ namespace cl = llvm::cl;
 namespace path = llvm::sys::path;
 
 auto ExplorerMain(int argc, char** argv, void* static_for_main_addr,
-                  llvm::StringRef relative_prelude_path) -> int {
+                  llvm::StringRef relative_prelude_path,
+                  llvm::raw_ostream& out_stream, llvm::raw_ostream& err_stream)
+    -> int {
   llvm::setBugReportMsg(
-
       "Please report issues to "
       "https://github.com/carbon-language/carbon-lang/issues and include the "
       "crash backtrace.\n");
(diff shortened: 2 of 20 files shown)


jonmeow:
Note this is a stacked changed, with https://github.com/carbon-language/carbon-lang/pull/3032/commits/f68f63f5cc4fbb73009f7b8d54fc024c11a60f79 being the commit specific to this PR.

explorer/fuzzing/ast_to_proto_main.cpp
```diff
@@ -38,8 +38,9 @@ auto Main(int argc, char** argv) -> ErrorOr<Success> {
   proto_file.close();
```
zygoloid: Pre-existing, but if you feel like fixing while you're here: this looks like it's trying to read the input file into a buffer before we call `Parse`, which seems unnecessary.

explorer/main.cpp
```diff
@@ -30,18 +31,33 @@ namespace cl = llvm::cl;
...
-
       "Please report issues to "
       "https://github.com/carbon-language/carbon-lang/issues and include the "
       "crash backtrace.\n");
   llvm::InitLLVM init_llvm(argc, argv);
 
   // Printing to stderr should flush stdout. This is most noticeable when stderr
   // is piped to stdout.
-  llvm::errs().tie(&llvm::outs());
+  err_stream.tie(&out_stream);
```
zygoloid: I think this probably belongs in `main`, not here.

If `out_stream` and `err_stream` are not going to the console (eg, if they're two different `raw_ostringstream`s, as I would imagine they are for `file_test`), then tying them like this doesn't seem necessary / appropriate, since they typically won't be interleaved anyway.

explorer/main.cpp
```diff
@@ -147,28 +164,28 @@ auto ExplorerMain(int argc, char** argv, void* static_for_main_addr,
     trace_stream.set_allowed_file_kinds(trace_file_kinds);
 
     if (trace_file_name == "-") {
-      trace_stream.set_stream(&llvm::outs());
+      trace_stream.set_stream(&raw_trace_stream);
     } else {
```
zygoloid: I find it a little surprising to change the meaning of `-` in this way. How would you feel about this alternative: pass a `std::optional<llvm::raw_ostream*> default_trace_stream` to this overload of `ExplorerMain` instead, and use that if no `trace_file_name` flag is passed.

(But this isn't so surprising to me that I'd block the patch on it; if you don't like my alternative, then please go ahead with this as-is.)

zygoloid (approved):

explorer/fuzzing/ast_to_proto_main.cpp
jonmeow: Removed

explorer/main.cpp
jonmeow: Per the header, this variant of ExplorerMain is intended for binary use. It's used by both //explorer and //installer:carbon, both of which need the tie behavior. So moving it out of this would only result in the behavior being copied into the two callers.

explorer/main.cpp
jonmeow: How about renaming to out_stream_for_trace with a little more documentation?

My concern with the behavior you described is that, right now, we can test that `--trace_file=-` causes sending trace output to the out stream. If we instead use default_trace_stream, it means that `--trace_file=-` is no longer necessary to produce trace output, leading to more untested code paths.

explorer/main.cpp
zygoloid: Hm, OK. Do we need to pass in the output and error streams? I think it's the fact that they're passed in and not hardcoded to `outs` / `errs` that's making me wonder whether this is the right place for this call. Both callers pass in `llvm::outs` and `llvm::errs`.

explorer/main.cpp
zygoloid: Sure, that works for me.

explorer/main.cpp
jonmeow: Will do.

PR #3033, opened 2023-07-27
Parser for index expression such as a[0]

fasiddique:


toolchain/parser/testdata/index/assign_to_var.carbon
```diff
@@ -0,0 +1,20 @@
...
+// CHECK:STDOUT:   {kind: 'VariableInitializer', text: '='},
+// CHECK:STDOUT:       {kind: 'NameExpression', text: 't'},
+// CHECK:STDOUT:     {kind: 'IndexExpressionStart', text: '[', subtree_size: 2},
+// CHECK:STDOUT:     {kind: 'Literal', text: '0'},
+// CHECK:STDOUT:   {kind: 'IndexExpression', text: ']', subtree_size: 4},
+// CHECK:STDOUT: {kind: 'VariableDeclaration', text: ';', subtree_size: 10},
+// CHECK:STDOUT: {kind: 'FileEnd', text: ''},
+// CHECK:STDOUT: ]
+
+var v : i32 = t[0];
```
jonmeow: Style-wise, we do not want a space between the name and the colon in a binding. I've suggested this style fix through GitHub before, but can you please fix these tests now?

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include "toolchain/semantics/semantics_context.h"
+
+namespace Carbon {
+
+auto SemanticsHandleIndexExpressionStart(SemanticsContext& /*context*/,
+                                         ParseTree::Node /*parse_node*/)
+    -> bool {
+  // TODO
+  return true;
```
jonmeow: Please look at other handling TODOs in semantics and copy how they call context.TODO. The return value is important.

toolchain/parser/parser_state.def
```diff
@@ -263,6 +279,8 @@ CARBON_PARSER_STATE(ExpressionInPostfix)
 // If `OpenParen`:
 //   1. CallExpression
 //   2. ExpressionInPostfixLoop
+// If `OpenSquareBracket`:
+//   1. IndexExpressionStart
```
jonmeow, line 280:
```suggestion
//   1. IndexExpressionStart
//   2. ExpressionInPostfixLoop
```

toolchain/parser/parser_handle_index_expression.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+#include "toolchain/parser/parser_context.h"
+
+namespace Carbon {
+
+auto ParserHandleIndexExpression(ParserContext& context) -> void {
+  auto state = context.PopState();
+  state.state = ParserState::IndexExpressionFinish;
+  context.PushState(state);
+  context.AddNode(ParseNodeKind::IndexExpressionStart, context.Consume(),
```
jonmeow:
```suggestion
  context.AddNode(ParseNodeKind::IndexExpressionStart, context.ConsumeChecked(TokenKind::OpenSquareBracket),
```

Please specify the token being consumed so that it can be validated. This is intended to help catch programming mistakes. Not all code is doing this yet, but it's the direction we want to head.

toolchain/parser/parser_handle_index_expression.cpp
```diff
@@ -0,0 +1,30 @@
...
+    context.emitter().Emit(*context.position(), InvalidIndexExpression);
+    context.ReturnErrorOnState();
+  } else {
+    context.PushState(ParserState::Expression);
+  }
+}
+
+auto ParserHandleIndexExpressionFinish(ParserContext& context) -> void {
+  auto state = context.PopState();
+  context.AddNode(ParseNodeKind::IndexExpression, context.Consume(),
```
jonmeow:
```suggestion
  context.AddNode(ParseNodeKind::IndexExpression, context.ConsumeChecked(TokenKind::CloseSquareBracket),
```

As above.

toolchain/parser/parser_state.def
```diff
@@ -39,6 +39,22 @@
   CARBON_PARSER_STATE_VARIANT(State, Variant1)                             \
   CARBON_PARSER_STATE_VARIANTS2(State, Variant2, Variant3)
 
+
```
jonmeow: Consistent with the previous state of this file, we would like to keep 1 blank line here, not two.

As much as possible, please try to maintain prior vertical whitespace formats. It's a good rule of thumb for following style.

toolchain/parser/parser_handle_index_expression.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+auto ParserHandleIndexExpression(ParserContext& context) -> void {
+  auto state = context.PopState();
+  state.state = ParserState::IndexExpressionFinish;
+  context.PushState(state);
+  context.AddNode(ParseNodeKind::IndexExpressionStart, context.Consume(),
+                  state.subtree_start, state.has_error);
+  if (context.PositionIs(TokenKind::CloseSquareBracket)) {
+    CARBON_DIAGNOSTIC(ExpectedExpressionAfter, Error,
+                      "Expected integer literal.");
```
jonmeow: While literals might be the only thing that you support right now for tuple literals, I think we should aim to have this message be more generic for index expressions in general, where expressions may generally be used. Something like "Expected an expression." would be better. Similarly I'd suggest just "ExpectedExpression" for the name since you don't have any "after" component to the diagnostic.

jonmeow (commented):
This looks like it'll work, although the amount to which this is a copy of CallExpression makes me wonder if we should be sharing code in the future. Mostly style fixes here, but please keep in mind that style is important!

toolchain/parser/parser_handle_index_expression.cpp
jonmeow: You've written "Expected expression inside the square bracket.", but note this is grammatically correct ("inside the square bracket" is singular, there are two brackets). Also, "ExpectedExpressionInside" is fairly generic relative to the specific nature of the message.

That said, I think this is something we can improve later. Change it to just "Expected an expression." and ExpectedExpression for the DiagnosticKind. Note here: (a) the message already provides the location where an expression was expected, and (b) although ExpectedExpression is generic, so is the message.

toolchain/parser/parser_handle_index_expression.cpp
```diff
@@ -0,0 +1,34 @@
...
+auto ParserHandleIndexExpression(ParserContext& context) -> void {
+  auto state = context.PopState();
+  state.state = ParserState::IndexExpressionFinish;
+  context.PushState(state);
+  context.AddNode(ParseNodeKind::IndexExpressionStart,
+                  context.ConsumeChecked(TokenKind::OpenSquareBracket),
+                  state.subtree_start, state.has_error);
+  if (context.PositionIs(TokenKind::CloseSquareBracket)) {
+    CARBON_DIAGNOSTIC(ExpectedExpressionInside, Error,
+                      "Expected expression inside the square bracket.", );
```
jonmeow:
```suggestion
                      "Expected expression inside the square bracket.");
```

toolchain/parser/testdata/index/fail_empty_expr.carbon
```diff
@@ -0,0 +1,23 @@
...
+// CHECK:STDOUT:   {kind: 'PatternBinding', text: ':', subtree_size: 3},
+// CHECK:STDOUT:   {kind: 'VariableInitializer', text: '='},
+// CHECK:STDOUT:       {kind: 'NameExpression', text: 't'},
+// CHECK:STDOUT:     {kind: 'IndexExpressionStart', text: '[', subtree_size: 2},
+// CHECK:STDOUT:   {kind: 'IndexExpression', text: ']', has_error: yes, subtree_size: 3},
+// CHECK:STDOUT: {kind: 'VariableDeclaration', text: ';', subtree_size: 9},
+// CHECK:STDOUT: {kind: 'FileEnd', text: ''},
+// CHECK:STDOUT: ]
+
+
```
jonmeow: Fix this vertical whitespace.

jonmeow (commented):
I think this is almost good to go, but the latest commit added a couple mistakes which should be fixed.

toolchain/parser/parser_handle_index_expression.cpp
```diff
@@ -0,0 +1,34 @@
...
+    context.emitter().Emit(*context.position(), ExpectedExpressionInside);
+    context.ReturnErrorOnState();
+  } else {
+    context.PushState(ParserState::Expression);
+  }
```
jonmeow:
```suggestion
  context.PushState(ParserState::Expression);
```

As you noticed when I said to use ExpectedExpression, the Expression state already handles this.  Conveniently, this allows us to just use Expression's handling!

toolchain/parser/parser_state.def
```diff
@@ -39,6 +39,21 @@
   CARBON_PARSER_STATE_VARIANT(State, Variant1)                             \
   CARBON_PARSER_STATE_VARIANTS2(State, Variant2, Variant3)
 
+// Handles an index expression `a[0]`.
+//
+// If `CloseSquareBracket`:
+//   1. IndexExpressionFinish
+// Else:
```
jonmeow:
```suggestion
// Always:
```

jonmeow (approved):
Thanks, glad that's working!

PR #3035, opened 2023-07-28
Refactor FileTest construction so that the test class is directly available.

jonmeow:
This is a simplification of the construction, although somewhat limiting (it means that the caller can't register the same file multiple times, though I stopped doing that anyways since it was causing confusion). What this more importantly _allows_ is logic on the FileTestBase child itself that's not test-specific -- in particular, autoupdate functionality which wouldn't use RUN_ALL_TESTS.

### DIFF
--- a/testing/file_test/file_test_base.h
+++ b/testing/file_test/file_test_base.h
@@ -147,6 +147,9 @@ struct FileTestFactory {
 // We can't use INSTANTIATE_TEST_CASE_P because of ordering issues between
 // container initialization and test instantiation by InitGoogleTest, but this
 // also allows us more flexibility in execution.
+//
+// We provide a `CARBON_FILE_TEST_FACTOR` macro below that provides a common,
+// convenient way to implement this function.
 extern auto GetFileTestFactory() -> FileTestFactory;
 
 // Provides a standard way to implement GetFileTestFactory.
--- a/testing/file_test/file_test_base.h
+++ b/testing/file_test/file_test_base.h
@@ -134,7 +134,7 @@ class FileTestBase : public testing::Test {
 
   const std::filesystem::path path_;
 };
-
+// A plain struct to aggregate a name and factory function for tests using this framework.
 struct FileTestFactory {
   // The test fixture name.
   const char* name;
--- a/testing/file_test/file_test_base.h
+++ b/testing/file_test/file_test_base.h
@@ -138,6 +138,7 @@ class FileTestBase : public testing::Test {
 struct FileTestFactory {
   // The test fixture name.
   const char* name;
+
   // A factory function for tests.
   std::function<FileTestBase*(const std::filesystem::path& path)> factory_fn;
 };
--- a/testing/file_test/file_test_base.h
+++ b/testing/file_test/file_test_base.h
@@ -19,8 +19,8 @@
 
 namespace Carbon::Testing {
 
-// A framework for testing files. Children implement `RegisterTestFiles` with
-// calls to `RegisterTests` using a factory that constructs the child.
+// A framework for testing files. Children write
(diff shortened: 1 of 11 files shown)


testing/file_test/file_test_base.h
```diff
@@ -148,12 +132,29 @@ class FileTestBase : public testing::Test {
...
+  std::function<FileTestBase*(const std::filesystem::path& path)> factory_fn;
 };
 
 // Must be implemented by the individual file_test to initialize tests.
-extern auto RegisterFileTests(
-    const llvm::SmallVector<std::filesystem::path>& paths) -> void;
+//
+// We can't use INSTANTIATE_TEST_CASE_P because of ordering issues between
+// container initialization and test instantiation by InitGoogleTest, but this
+// also allows us more flexibility in execution.
```
chandlerc, line 151: Maybe add a comment suggesting the use of the macro below?

```suggestion
//
// We can't use INSTANTIATE_TEST_CASE_P because of ordering issues between
// container initialization and test instantiation by InitGoogleTest, but this
// also allows us more flexibility in execution.
//
// We provide a `CARBON_FILE_TEST_FACTOR` macro below that provides a common,
// convenient way to implement this function.
```

testing/file_test/file_test_base.h
```diff
@@ -148,12 +132,29 @@ class FileTestBase : public testing::Test {
   static auto TransformExpectation(int line_index, llvm::StringRef in)
       -> testing::Matcher<std::string>;
 
-  const std::filesystem::path* path_;
+  const std::filesystem::path path_;
+};
+
```
chandlerc, line 137: Maybe a bit of documentation would be useful here?

```suggestion
// A plain struct to aggregate a name and factory function for tests using this framework.
```

testing/file_test/file_test_base.h
```diff
@@ -148,12 +132,29 @@ class FileTestBase : public testing::Test {
...
+struct FileTestFactory {
+  // The test fixture name.
+  const char* name;
+  // A factory function for tests.
```
chandlerc, line 143: Super small nit -- a blank line between these maybe?

```suggestion

  // A factory function for tests.
```

testing/file_test/file_test_base.h
```diff
@@ -148,12 +132,29 @@ class FileTestBase : public testing::Test {
   static auto TransformExpectation(int line_index, llvm::StringRef in)
       -> testing::Matcher<std::string>;
 
-  const std::filesystem::path* path_;
+  const std::filesystem::path path_;
+};
+
+struct FileTestFactory {
+  // The test fixture name.
+  const char* name;
```
chandlerc, line 141: I'm guessing `llvm::StringLiteral` doesn't work here? Unfortunate, but understandable if so.

chandlerc (approved):
Nice! A bunch of tiny / minor suggested edits. LG with these edits or with whatever re-phrasing makes more sense.

testing/file_test/file_test_base.h
jonmeow: It just gets awkward:

```
Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
testing/file_test/file_test_base.cpp:368:5: error: no matching function for call to 'RegisterTest'
    testing::RegisterTest(test_factory.name, test_name.c_str(), nullptr,
    ^~~~~~~~~~~~~~~~~~~~~
external/com_google_googletest/googletest/include/gtest/gtest.h:2259:11: note: candidate function template not viable: no known conversion from 'llvm::StringLiteral' to 'const char *' for 1st argument
TestInfo* RegisterTest(const char* test_suite_name, const char* test_name,
          ^
```

[using StringLiteral then falling back to `.data` feels odd to me]

PR #3038, opened 2023-07-28
Basic semantic checking for pointer types and const-qualified types.

zygoloid:
Semantic handling for use of `T*` and `const T` as types.

There's no way to form values of these types yet, and no conversions for them are supported.

Factor out the common code to canonicalize types using a folding set, and switch to using the same folding set for all kinds of type by adding the kind as part of the folding set key.

Improve type printing to not include the `as type` portion when the type is printed in a context within another type where a conversion to `type` is implied, as in `{}*` and pre-existing cases like `({}, {}) as type` (which we used to print as `({} as type, {} as type}) as type`.


toolchain/semantics/semantics_node_kind.h
```diff
@@ -42,6 +43,8 @@ class SemanticsNodeKind : public CARBON_ENUM_BASE(SemanticsNodeKind) {
   // code block appears after all other instructions, and ends with a
   // terminator instruction.
   [[nodiscard]] auto terminator_kind() const -> SemanticsTerminatorKind;
+
+  void Profile(llvm::FoldingSetNodeID& id) { id.AddInteger(AsInt()); }
```
jonmeow, line 49: Brief function comment for the unfamiliar?

toolchain/semantics/testdata/const/collapse.carbon
```diff
@@ -0,0 +1,67 @@
...
+// CHECK:STDOUT:     node+7,
+// CHECK:STDOUT:     node+8,
+// CHECK:STDOUT:     node+9,
+// CHECK:STDOUT:   ],
+// CHECK:STDOUT:   [
+// CHECK:STDOUT:     node+10,
+// CHECK:STDOUT:   ],
+// CHECK:STDOUT: ]
+
+// OK, `const (const i32)` is the same type as `const i32`.
```
jonmeow, line 64: Sort of feels like this should be a warning for excessive constness...

toolchain/semantics/semantics_ir.cpp
```diff
@@ -206,9 +254,17 @@ auto SemanticsIR::StringifyType(SemanticsTypeId type_id) -> std::string {
     SemanticsNodeId node_id;
     // The index into node_id to print. Not used by all types.
     int index = 0;
+    // Whether this step is printed in a context that will be coerced to `type`.
+    bool type_context = true;
```
jonmeow: It seems like you're using this to try to suppress `type` in things like `(i32) [as type]*` or `((i32,) [as type],) as type` which is more clearly a type. Is there an example (test) where you want to print this other than the outermost type?

Put differently, a test which shows why this can't just say after the loop "if outer_node_id is a tuple or {}, then print `as type`".

toolchain/semantics/semantics_handle_operator.cpp
```diff
@@ -80,6 +100,17 @@ auto SemanticsHandlePrefixOperator(SemanticsContext& context,
               value_id));
       break;
 
+    case TokenKind::Const: {
+      auto inner_type_id = context.ExpressionAsType(parse_node, value_id);
+      context.AddNodeAndPush(
+          parse_node,
+          SemanticsNode::ConstType::Make(
+              parse_node,
+              context.CanonicalizeType(SemanticsNodeId::BuiltinTypeType),
```
jonmeow: Can this SemanticsTypeId::TypeType? That is, I think BuiltinTypeType doesn't need canonicalization.
```suggestion
              SemanticsTypeId::TypeType,
```

toolchain/semantics/semantics_handle_operator.cpp
```diff
@@ -61,7 +61,27 @@ auto SemanticsHandleInfixOperator(SemanticsContext& context,
...
+  // Figure out the operator for the token.
+  auto token = context.parse_tree().node_token(parse_node);
+  switch (auto token_kind = context.tokens().GetKind(token)) {
+    case TokenKind::Star: {
+      auto inner_type_id = context.ExpressionAsType(parse_node, value_id);
+      context.AddNodeAndPush(
+          parse_node,
+          SemanticsNode::PointerType::Make(
+              parse_node,
+              context.CanonicalizeType(SemanticsNodeId::BuiltinTypeType),
```
jonmeow:
```suggestion
              SemanticsTypeId::TypeType,
```

See below comment

toolchain/semantics/semantics_handle_operator.cpp
```diff
@@ -61,7 +61,27 @@ auto SemanticsHandleInfixOperator(SemanticsContext& context,
 
 auto SemanticsHandlePostfixOperator(SemanticsContext& context,
                                     ParseTree::Node parse_node) -> bool {
-  return context.TODO(parse_node, "HandlePostfixOperator");
+  auto value_id = context.node_stack().PopExpression();
+
+  // Figure out the operator for the token.
+  auto token = context.parse_tree().node_token(parse_node);
+  switch (auto token_kind = context.tokens().GetKind(token)) {
+    case TokenKind::Star: {
```
jonmeow, line 67: The switch with one case seems odd to me -- how many postfix operators do you anticipate? My impression was we're dropping `++` and `--` from C++, so maybe `*` will be the only one?

Note, if you're not anticipating others, it might be more correct in the default to CARBON_FATAL instead of TODO. I want to get rid of the TODO once semantics is more complete, it's kind of a patchwork thing while we're building the toolchain so shouldn't be used where we don't anticipate more work.

toolchain/semantics/semantics_handle_operator.cpp
```diff
@@ -61,7 +61,27 @@ auto SemanticsHandleInfixOperator(SemanticsContext& context,
...
+              context.CanonicalizeType(SemanticsNodeId::BuiltinTypeType),
+              inner_type_id));
+      break;
+    }
+
+    default:
+      return context.TODO(parse_node, llvm::formatv("Handle {0}", token_kind));
+  }
+
+  return true;
```
jonmeow:
```suggestion
      return true;
    }

    default:
      return context.TODO(parse_node, llvm::formatv("Handle {0}", token_kind));
  }
```

Just for your consideration, trying to simplify control flow.

toolchain/semantics/semantics_context.h
```diff
@@ -262,6 +265,27 @@ class SemanticsContext {
...
+  // `profile_type(id)` is called to determine the `FoldingSetNodeID id` for
+  // this type. The ID should be distinct for all distinct type values with the
+  // same `kind`.
+  //
+  // `make_node()` is called to obtain a `SemanticsNodeId` that describes the
+  // type. It is only called if the type does not already exist, so can be used
+  // to lazily build the `SemanticsNode`.
+  template <typename ProfileType, typename MakeNode>
+  auto CanonicalizeTypeImpl(SemanticsNodeKind kind, ProfileType profile_type,
+                            MakeNode make_node) -> SemanticsTypeId;
```
jonmeow: What was the reason for the template approach for these callbacks? Can a std::function provide some structure?

toolchain/semantics/semantics_context.cpp
```diff
@@ -463,69 +488,87 @@ auto SemanticsContext::CanonicalizeType(SemanticsNodeId node_id)
...
+          });
+    }
+
+    case SemanticsNodeKind::StructType:
+    case SemanticsNodeKind::TupleType: {
+      CARBON_FATAL() << "Type should have been canonizalized when created: "
+                     << node;
+    }
+
+    default: {
```
jonmeow: Minor, but FWIW, these blank lines between cases that already have `{}` feel like they could be removed without harming readability.

jonmeow (commented):
Nice! A couple questions about the style of the approach, but I think the collapsing of folding sets is fine. To be sure I understand your logic, we're otherwise looking at multiple folding sets, and you think it's better to add the node kind to the FoldingSetNodeID than to store each kind separately?

toolchain/semantics/semantics_context.cpp
zygoloid: Removed.

toolchain/semantics/semantics_context.h
zygoloid: I wanted to avoid the indirect call overhead and optimization barrier from things like `std::function` or `llvm::function_ref`. But it's probably not a big difference and I don't mind going the other way.

toolchain/semantics/semantics_handle_operator.cpp
zygoloid: I think this is the only postfix operator token we currently have in the design. While we do have multi-token postfix operators like `.x`, they won't end up here. We've talked about adding a postfix `^` dereference operator but didn't get consensus on that so far at least.

Replaced the TODO with a `CARBON_FATAL`.

I thought about switching from a `switch` to an `if` too, but I'd like to keep the overall structure of this function as-is if we can, because there are three other functions adjacent to this one with the same structure (for prefix operators, infix operators, and short-circuiting operators) and I think the file as a whole is more consistent with this structure even though there are only two possibilities here.

toolchain/semantics/semantics_handle_operator.cpp
zygoloid: Sure, applied throughout this file for consistency.

toolchain/semantics/semantics_ir.cpp
zygoloid: I don't see a reason why we can't do the `as type` printing as a separate step after the loop.

Hm, a question, though: how do we want to format (eg) `const (i32, i32)` or `const {}`? Currently this PR will print them without the `as type` suffix, which is correct since `const` coerces its argument to type `type` anyway, but maybe it's inconsistent to use `const ()` versus `() as type` in our output.

toolchain/semantics/semantics_node_kind.h
zygoloid: Added.

toolchain/semantics/testdata/const/collapse.carbon
zygoloid: Added.

zygoloid:
> To be sure I understand your logic, we're otherwise looking at multiple folding sets, and you think it's better to add the node kind to the FoldingSetNodeID than to store each kind separately?

Yes. I'm trying to avoid the [explosion of `FoldingSet`s](https://github.com/llvm/llvm-project/blob/main/clang/include/clang/AST/ASTContext.h#L186) we saw in Clang as a result of having one set per kind of type. This should give us less code complexity and hopefully better locality, at the cost of every type having a slightly larger profile.

toolchain/semantics/semantics_context.cpp
```diff
@@ -448,6 +448,44 @@ auto SemanticsContext::ParamOrArgSave(bool for_args) -> void {
...
+#ifndef NDEBUG
+  // In a debug build, check that our insertion position is still valid. It
+  // could have been invalidated by a misbehaving `make_node`.
+  void* check_insert_pos;
+  auto* check_node =
+      canonical_type_nodes_.FindNodeOrInsertPos(canonical_id, check_insert_pos);
+  CARBON_CHECK(!check_node)
+      << "Type was created recursively during canonicalization";
+  CARBON_CHECK(insert_pos == check_insert_pos)
+      << "Insertion position changed during canonicalization";
```
jonmeow: Per discussion -- you might consider introducing DCHECK, particularly since it'd allow dropping NDEBUG (personally, I'd write that Find in the DCHECK simply because I don't like #ifdef inside code)

toolchain/semantics/semantics_context.h
jonmeow: I'd suggest a std::function or function_ref, partly because I'm not clear the overhead is an issue. I'd rather go a template (with its type-checking and readability implications) only if a function has too much of a performance impact (and I also wonder if benchmarks might push in a different direction too for code structure)

toolchain/semantics/semantics_ir.cpp
jonmeow: 🤷 

Error context offers a bit of "this is a type" hint, and `const` is also a type hint, so maybe the two hints together would generally be sufficient. It may even be that the error context is a sufficient hint and `as type` is more likely to confuse than help.

I mean, consider something like:

```
Cannot implicitly convert from `i32` to `()`.
```

In code, people write `fn F() -> ()`, not `fn F() -> () as type`. Maybe we don't need `as type` in error messages either?

Could also adjust error messages to do small things like "from" -> "from type" to hammer it in a little, without "as type" on the actual types.

toolchain/semantics/semantics_context.h
zygoloid: Done.

toolchain/semantics/semantics_context.cpp
zygoloid: Done.

toolchain/semantics/semantics_ir.cpp
zygoloid: We should probably have two modes here, a "printing a value, make sure it's unambiguous" mode, and a "known based on context to be printing a type" mode. But I'll leave it as-is for now, with the `as type` printing moved to the end as suggested, and we can consider what to do here in later PRs.

jonmeow (approved):
Thanks! I'm really happy with this.

PR #3039, opened 2023-07-30
Explorer: improve type checking trace output

prabhatexit0:
Updates the trace output for type checking to be more consistent like the rest of trace output.
This PR also includes few changes in the lit tests related to trace output.


explorer/ast/bindings.cpp
```diff
@@ -36,6 +36,19 @@ void Bindings::Add(Nonnull<const GenericBinding*> binding,
...
+  for (const auto& [binding, value] : args_) {
+    out << sep << "`" << *binding << "`: `" << *value << "`";
+  }
+  out << "]\n >  bindings witnesses: [";
+  for (const auto& [binding, value] : witnesses_) {
+    out << sep << "`" << *binding << "`: `" << *value << "`";
+  }
+  out << "]";
+};
+
```
zygoloid: See #3051: the output here will be non-deterministic, because `args_` and `witnesses_` are maps that use pointers as keys. That can result in trace tests being flaky.

We can address this by sorting the maps based on the `index()` of the bindings (for example, copy each map into a `vector`, sort by index, then print).

explorer/interpreter/impl_scope.cpp
```diff
@@ -415,10 +415,10 @@ auto ImplScope::TryResolveInterfaceHere(
 
 // TODO: Add indentation when printing the parents.
 void ImplScope::Print(llvm::raw_ostream& out) const {
-  out << "impl declarations: ";
   llvm::ListSeparator sep;
```
zygoloid: I wonder if it would be better to use a different separator here. The current lists of impl declarations are very hard to read and lead to extremely long lines. Maybe something like this:
```suggestion
  llvm::ListSeparator sep(",\n    ");
```

I'm thinking about output something like this:
```
--- impl declarations for `interface TestInterface` (full_trace.carbon:7)
 + []
 + [`bool` as `interface EqWith(U = bool)`,
    `i32` as `interface EqWith(U = i32)`,
    `String` as `interface EqWith(U = String)`,
    `i32` as `interface CompareWith(U = i32)`,
    `String` as `interface CompareWith(U = String)`,
    `i32` as `interface LessWith(U = i32)`,
    `String` as `interface LessWith(U = String)`,
    `i32` as `interface LessEqWith(U = i32)`,
    `String` as `interface LessEqWith(U = String)`,
    `i32` as `interface GreaterWith(U = i32)`,
    `String` as `interface GreaterWith(U = String)`,
    `i32` as `interface GreaterEqWith(U = i32)`,
    `String` as `interface GreaterEqWith(U = String)`,
    `i32` as `interface Negate`,
    `i32` as `interface AddWith(U = i32)`,
    `i32` as `interface SubWith(U = i32)`,
    `i32` as `interface MulWith(U = i32)`,
    `i32` as `interface DivWith(U = i32)`,
    `i32` as `interface ModWith(U = i32)`,
    `i32` as `interface BitComplement`,
    `i32` as `interface BitAndWith(U = i32)`,
    `i32` as `interface BitOrWith(U = i32)`,
    `i32` as `interface BitXorWith(U = i32)`,
    `i32` as `interface LeftShiftWith(U = i32)`,
    `i32` as `interface RightShiftWith(U = i32)`,
    `i32` as `interface Inc`,
    `i32` as `interface Dec`,
    `U` as `interface __EqualConverter` [0],
    `(T1,)` as `interface As(T = (U1,))` [0, 0, 0; 1, 1, 0, 0, 0],
    `(T1, T2)` as `interface As(T = (U1, U2))` [0, 0, 0; 0, 0, 1; 1, 1, 0, 0, 0; 1, 1, 0, 0, 1],
    `(T1, U1)` as `interface EqWith(U = (T2, U2))` [0, 0, 0; 0, 0, 1; 1, 1, 0, 0, 0; 1, 1, 0, 0, 1],
    `(T1, T2, T3)` as `interface As(T = (U1, U2, U3))` [0, 0, 0; 0, 0, 1; 0, 0, 2; 1, 1, 0, 0, 0; 1, 1, 0, 0, 1; 1, 1, 0, 0, 2],
    `T` as `interface ImplicitAs(T = U)` [0; 1, 1, 0],
    `T` as `interface As(T = U)` [0; 1, 1, 0],
    `T` as `interface ImplicitAs(T = U)` [0; 1, 1, 0],
    `T` as `interface As(T = U)` [0; 1, 1, 0],
    `T` as `interface AssignWith(U = U)` [0; 1, 1, 0],
    `T` as `interface AddAssignWith(U = U)` [0; 1, 1, 0],
    `T` as `interface SubAssignWith(U = U)` [0; 1, 1, 0],
    `T` as `interface MulAssignWith(U = U)` [0; 1, 1, 0],
    `T` as `interface DivAssignWith(U = U)` [0; 1, 1, 0],
    `T` as `interface ModAssignWith(U = U)` [0; 1, 1, 0],
    `T` as `interface BitAndAssignWith(U = U)` [0; 1, 1, 0],
    `T` as `interface BitOrAssignWith(U = U)` [0; 1, 1, 0],
    `T` as `interface BitXorAssignWith(U = U)` [0; 1, 1, 0],
    `T` as `interface LeftShiftAssignWith(U = U)` [0; 1, 1, 0],
    `T` as `interface RightShiftAssignWith(U = U)` [0; 1, 1, 0]]
```

... which I think is a lot more readable. (Though it's very long! I think we should be printing this out a lot less than we currently do. But we can leave any improvements there to a different PR.)

explorer/interpreter/type_checker.cpp
```diff
@@ -1029,9 +1029,10 @@ auto TypeChecker::ImplicitlyConvert(std::string_view context,
     destination = destination_constraint;
 
     if (trace_stream_->is_enabled()) {
-      *trace_stream_ << "converting type " << *converted_value
-                     << " to constraint " << *destination_constraint << " for "
-                     << context << " in scope " << impl_scope << "\n";
+      *trace_stream_ << "converting type `" << *converted_value
+                     << "` to constraint `" << *destination_constraint
+                     << "` for `" << context << "` in scope " << impl_scope
```
zygoloid: `context` here is a string that's intended to be interpolated into the trace message; I think we shouldn't put backticks around it.
```suggestion
                     << "` for " << context << " in scope " << impl_scope
```

explorer/interpreter/type_checker.cpp
```diff
@@ -5826,26 +5855,27 @@ auto TypeChecker::TypeCheckConstraintTypeDeclaration(
...
   if (constraint_decl->params().has_value()) {
     BringPatternImplBindingsIntoScope(*constraint_decl->params(),
                                       constraint_scope);
   }
   if (trace_stream_->is_enabled()) {
-    *trace_stream_ << constraint_scope;
+    *trace_stream_ << "--- impl declarations for `"
+                   << PrintAsID(*constraint_decl) << "` ("
+                   << constraint_decl->source_loc() << ")\n";
+    *trace_stream_ << " +  " << constraint_scope;
```
zygoloid: You have a `" +  "` here, and another `" +  "` in the print function for the constraint scope when you need a line continuation. I think it would be better to move all the `" +  "` printing to the same place, in `constraint_scope` itself, and do it before printing the scope rather than before printing the parent. That'd make it easier to keep the formatting consistent. What do you think?

zygoloid (commented):
Looking at this change, I'm wondering if we should change the way that we print types and values and such, so that the enclosing backticks are printed automatically by default. But let's go ahead with this as-is, and we can look at ways of automating the backticks later.

explorer/interpreter/impl_scope.cpp
prabhatexit0: I think it is much better this way.
I was thinking of adding source location information to `ImplFact` to deal with this issue and filter is based on file context but I'm not able to figure it out yet.

explorer/interpreter/type_checker.cpp
prabhatexit0: Yes it will make it more consistent. Thanks! Doing the change in next commit.

explorer/ast/bindings.cpp
prabhatexit0: I did it for `args_` using index but I am having issues with doing the same for `witness_` as there isn't index inside `ImplBinding`.
How should we handle that case?

also I was getting errors due to `index()` method as it was returning value of an optional int without checking, for now I've returned 0 in case the `index_` is `std::nullopt`, let me know if there is better way to handle it.

explorer/ast/bindings.cpp
```diff
@@ -37,21 +37,33 @@ void Bindings::Add(Nonnull<const GenericBinding*> binding,
...
-}
-
 void Bindings::Print(llvm::raw_ostream& out) const {
-  out << "{";
+  std::vector<std::pair<Nonnull<const GenericBinding*>, Nonnull<const Value*>>>
+      args;
+
+  for (const auto& [binding, value] : args_) {
+    args.push_back({binding, value});
+  }
```
zygoloid: You can construct the vector directly from the map because the elements are the same type:
```suggestion
  std::vector<std::pair<Nonnull<const GenericBinding*>, Nonnull<const Value*>>>
      args(args_.begin(), args_.end());
```

explorer/ast/pattern.h
```diff
@@ -299,7 +299,13 @@ class GenericBinding : public Pattern {
...
   // The index of this binding, which is the number of bindings that are in
   // scope at the point where this binding is declared.
-  auto index() const -> int { return *index_; }
+  auto index() const -> int {
+    if (index_) {
+      return *index_;
+    } else {
+      return 0;
+    }
+  }
```
zygoloid: If possible, I'd like to assert in `Bindings` that we never add a value for a `GenericBinding` whose index hasn't been set yet. The index should be set by type-checking, and I think we shouldn't ever have a value for a generic binding if we don't yet know its type.

explorer/interpreter/type_checker.cpp
```diff
@@ -1029,9 +1029,10 @@ auto TypeChecker::ImplicitlyConvert(std::string_view context,
     destination = destination_constraint;
 
     if (trace_stream_->is_enabled()) {
-      *trace_stream_ << "converting type " << *converted_value
-                     << " to constraint " << *destination_constraint << " for "
-                     << context << " in scope " << impl_scope << "\n";
+      *trace_stream_ << "converting type `" << *converted_value
+                     << "` to constraint `" << *destination_constraint
+                     << "` for " << context << " in scope " << impl_scope
```
zygoloid: `ImplScope` printing expects to be on a line of its own, so perhaps this would be better:
```suggestion
                     << "` for " << context << " in scope:\n" << impl_scope
```

explorer/ast/bindings.cpp
```diff
@@ -37,21 +37,33 @@ void Bindings::Add(Nonnull<const GenericBinding*> binding,
...
-  for (const auto& [name, value] : args_) {
-    out << sep << *name << " -> " << *value;
+  out << " >  bindings args: [";
+  for (const auto& [binding, value] : args) {
+    out << sep << "`" << *binding << "`: `" << *value << "`";
   }
-  for (const auto& [name, value] : witnesses_) {
-    out << sep << *name << " -> " << *value;
+  out << "]\n >  bindings witnesses: [";
+  for (const auto& [binding, value] : witnesses_) {
```
zygoloid: You can use `binding->type_var()->index()` as the sort key here.

prabhatexit0: Okay, I'll do that in next commit.

explorer/ast/pattern.h
prabhatexit0: Done. The assertion was failing previously, but I resolved it by making a change inside the `TypeChecker::DeclareImplDeclaration` function. Instead of directly using `generic_bindings.push_back(deduced)` to add deduced parameters to the scope, I replaced it with a call to `CollectAndNumberGenericBindingsInPattern(...)`. This adjustment appears to have resolved the issues.

zygoloid (approved):
Looks great! There's a merge conflict; if you message me when that's fixed I'll hit the merge button.

PR #3042, opened 2023-07-31
Update re2, absl, and googletest

jonmeow:
re2 has migrated from StringPiece to absl::string_view since we last updated. There are also some changes to how string_view is packaged by absl. As I do test work, I'm noticing these gaps and think an update would help.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -56,12 +56,12 @@ configure_clang_toolchain(name = "bazel_cc_toolchain")
 # Abseil libraries
 ###############################################################################
 
-# Head as of 2022-09-13.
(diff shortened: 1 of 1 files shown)


jonmeow:
Note the relevant commit here is https://github.com/carbon-language/carbon-lang/pull/3042/commits/373d8f2b6680b6b9037742251f54734f2542ebe5, but if this is approved before #3035 I'll probably rebase. Right now they're stacked because I'm working on changes needing both.

chandlerc (approved):
LG, leave it to you whether to rebase and merge or wait for the earlier commit and merge.

PR #3043, opened 2023-08-01
Implement autoupdate for file_test.

jonmeow:
I've migrated the toolchain autoupdate scripts here, I just need a little more time to do the explorer side (which I need to check performance, that may require multithreading as we do in the current script). However, this felt substantial enough to share and it means I can handle autoupdate in more of the toolchain, including preparatory work for autoupdate on multi-file tests.

Once explorer is done I'll remove the old script.


jonmeow:
[edit: no longer stacked] Note this PR is stacked, the last commit will apply; currently https://github.com/carbon-language/carbon-lang/commit/7f52e0fd54b41b067e5d952fa7981c2f4ca89226

testing/file_test/autoupdate.h
```diff
@@ -0,0 +1,70 @@
...
+#include <filesystem>
+
+#include "common/ostream.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/StringRef.h"
+
+namespace Carbon::Testing {
+
+// Interface for lines.
+class FileTestLineBase {
```
chandlerc: This type hierarchy being in `autoupdate.h` was a suprise to me. I understand the goal of trying to pull apart things from the main `file_test_base.h`, but I think these at least make much more sense there than here.

I'll try to update this comment if as I finish reviewing this PR I have other ideas for how to structure things that allows more fine grained headers.

scripts/fix_cc_deps.py
```diff
@@ -58,6 +58,10 @@ class ExternalRepo(NamedTuple):
     "@com_google_absl": ExternalRepo(
         lambda x: re.sub(":", "/", x), "...", None
     ),
+    # :re2/re2.h -> re2/re2.h
+    "@com_googlesource_code_re2": ExternalRepo(
+        lambda x: re.sub(":", "", x), ":re2", None
+    ),
```
chandlerc, line 64: Should this be in the RE 2 update commit? Or is this the first dependency that makes it needed?

testing/file_test/autoupdate.cpp
```diff
@@ -0,0 +1,241 @@
...
+  }
+
+  // Prepare to look for filenames in lines.
+  llvm::StringRef current_filename = filenames[0];
+  const auto* remaining_filenames = filenames.begin() + 1;
+
+  // %t substitution means we may see TEST_TMPDIR in output.
+  char* temp = getenv("TEST_TMPDIR");
+  CARBON_CHECK(temp != nullptr);
+  llvm::StringRef temp_ref = temp;
```
chandlerc: Just `temp` seems a bit confusing to me. WDYT about naming this after the variable? Maybe `test_tmpdir_env` here and the `StringRef` variable that we'll actually use as `test_tmpdir`?

testing/file_test/file_test_base.h
```diff
@@ -10,27 +10,34 @@
...
+// `bazel run :file_test -- --autoupdate` automatically constructs compatible
+// CHECK:STDOUT: and CHECK:STDERR: lines.
 //
 // Supported comment markers are:
 //
+// - // AUTOUDPATE
+//   // NOAUTOUPDATE
+//
+//   Controls whether the checks in the file will be autoupdated if --autoupdate
+//   is passed. Exactly one of these two markers must be present.
```
chandlerc: From reading the code, also governs where those checks are inserted -- maybe its worth a  brief bit of docs here about where to expect the `CHECK` lines to be placed?

toolchain/autoupdate_testdata.sh
```diff
@@ -0,0 +1,22 @@
...
+  //toolchain/lowering:lowering_file_test \
+  //toolchain/parser:parse_tree_file_test \
+  //toolchain/semantics:semantics_file_test
+
+$(dirname %~dp0)/codegen/autoupdate_testdata.sh
+$(dirname %~dp0)/driver/autoupdate_testdata.sh
+$(dirname %~dp0)/lexer/autoupdate_testdata.sh
+$(dirname %~dp0)/lowering/autoupdate_testdata.sh
+$(dirname %~dp0)/parser/autoupdate_testdata.sh
+$(dirname %~dp0)/semantics/autoupdate_testdata.sh
```
chandlerc: Rather than having all of these separate but largely redundant scripts, what about just having this one?

If its important to just autoupdate one part of the toolchain, could just accept a positional parameter and dispatch on that here and I think it'd be a bit cleaner than all the one-line test scripts that have to be kept in sync.

chandlerc (commented):
Feel like I didn't really finish getting the whole context, but wanted to quickly leave some comments. I'll dig a bit deeper on this PR tomorrow.

scripts/fix_cc_deps.py
jonmeow: This was the first dependency, so feels more appropriate here. Previously it was only used indirectly via googletest.

testing/file_test/autoupdate.cpp
jonmeow: Are you okay with this as tmpdir instead of test_tmpdir? Mainly I do things like target = TEST_TARGET elsewhere, so prefixing test here feels a bit inconsistent (if you want test_, I'd go adjust others).

Note, I had another temp and changed it to tmpdir too.

testing/file_test/autoupdate.h
jonmeow: How about this as a separate line.h?

testing/file_test/file_test_base.h
jonmeow: Added. Currently:

```
//   When autoupdating, CHECKs will be inserted starting below AUTOUPDATE. When
//   a CHECK has line information, autoupdate will try to insert the CHECK
//   immediately above the line it's associated with. When that happens, any
//   following CHECK lines without line information will immediately follow,
//   between the CHECK with line information and the associated line.
```

toolchain/autoupdate_testdata.sh
jonmeow: > Rather than having all of these separate but largely redundant scripts, what about just having this one?

I'd been thinking it was helpful to keep these scripts for people who are used to the current setup, but done.

> If its important to just autoupdate one part of the toolchain, could just accept a positional parameter and dispatch on that here and I think it'd be a bit cleaner than all the one-line test scripts that have to be kept in sync.

For reference, this is going to be the typical use-case, that someone is working on one part of the toolchain, and that changes there don't affect everything. From a development perspective this will help quick iteration.

jonmeow (commented):
Pushed an update.

testing/file_test/autoupdate.cpp
chandlerc: Yep, that makes sense.

testing/file_test/autoupdate.h
chandlerc: Works for me.

chandlerc (approved):
LGTM

PR #3044, opened 2023-08-01
Toolchain: Semantics for indexing

fasiddique:


toolchain/lowering/lowering_handle.cpp
```diff
@@ -156,6 +156,12 @@ auto LoweringHandleIntegerLiteral(LoweringFunctionContext& context,
   context.SetLocal(node_id, v);
 }
 
+auto LoweringHandleIndex(LoweringFunctionContext& /*context*/,
```
jonmeow: Please keep this lexically sorted, between FunctionDeclaration and IntegerLiteral.

I know this is a nit, but please fix this everywhere in this change.

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -5,17 +5,56 @@
...
 #include "toolchain/semantics/semantics_context.h"
+#include "toolchain/semantics/semantics_node.h"
+#include "toolchain/semantics/semantics_node_kind.h"
 
 namespace Carbon {
 
 auto SemanticsHandleIndexExpressionStart(SemanticsContext& context,
                                          ParseTree::Node parse_node) -> bool {
-  return context.TODO(parse_node, "HandleIndexExpressionStart");
+  context.node_stack().Push(parse_node, context.node_stack().PopExpression());
```
jonmeow:
```suggestion
  // Leave the expression on the stack for IndexExpression.
```

We've been trying an approach of minimizing push/pop. Since the IndexExpressionStart isn't significant, we don't need to add it.

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -5,17 +5,56 @@
...
+  return true;
 }
 
 auto SemanticsHandleIndexExpression(SemanticsContext& context,
                                     ParseTree::Node parse_node) -> bool {
-  return context.TODO(parse_node, "HandleIndexExpression");
+  auto ind_node = context.node_stack().PopExpression();
+  auto [ind_expr_parse_node, name_node_id] =
```
jonmeow: It looks like ind_expr_parse_node is unused, so you shouldn't need to fetch it. Note the "WithParseNode" is optional.

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -5,17 +5,56 @@
...
+    auto cast_ind_id = context.ImplicitAsRequired(
+        parse_node, ind_node,
+        context.CanonicalizeType(SemanticsNodeId::BuiltinIntegerType));
+    auto cast_ind_node = context.semantics_ir().GetNode(cast_ind_id);
+
+    if ((context.semantics_ir().GetTypeAllowBuiltinTypes(
+             cast_ind_node.type_id()) != SemanticsNodeId::BuiltinError) &&
+        (name_type.kind() == SemanticsNodeKind::TupleType)) {
+      auto ind_val = context.semantics_ir()
+                         .GetIntegerLiteral(cast_ind_node.GetAsIntegerLiteral())
```
jonmeow: Please add a unit test for an integer value:

```
var a: i32 = 1;
(0, 1)[a];
```

toolchain/semantics/semantics_handle_variable.cpp
```diff
@@ -4,6 +4,7 @@
 
 #include "toolchain/semantics/semantics_context.h"
 #include "toolchain/semantics/semantics_node.h"
+#include "toolchain/semantics/semantics_node_kind.h"
```
jonmeow: What's this being added for?

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -5,17 +5,56 @@
...
+                         .getSExtValue();
+      auto type_block_id = name_type.GetAsTupleType();
+      auto type_block = context.semantics_ir().GetTypeBlock(type_block_id);
+      if (ind_val >= static_cast<int>(type_block.size())) {
+        CARBON_DIAGNOSTIC(OutOfBoundAccess, Error, "Out of bound access.");
```
jonmeow: What do you think of adding the index and threshold to this error message?

Also, why "bound" instead of "bounds"?

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -5,17 +5,56 @@
...
+        context.emitter().Emit(parse_node, OutOfBoundAccess);
+      } else {
+        context.node_stack().Push(
```
jonmeow: We have an AddNodeAndPush helper that would simplify this a little.

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -5,17 +5,56 @@
...
                                     ParseTree::Node parse_node) -> bool {
-  return context.TODO(parse_node, "HandleIndexExpression");
+  auto ind_node = context.node_stack().PopExpression();
+  auto [ind_expr_parse_node, name_node_id] =
+      context.node_stack()
+          .PopWithParseNode<ParseNodeKind::IndexExpressionStart>();
+  auto name_node = context.semantics_ir().GetNode(name_node_id);
+
+  if (context.semantics_ir().GetTypeAllowBuiltinTypes(name_node.type_id()) !=
+      SemanticsNodeId::BuiltinError) {
```
jonmeow: Why do you call GetType for this, instead of comparing with SemanticsTypeId::Error directly?

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -5,17 +5,56 @@
...
+    auto name_type = context.semantics_ir().GetNode(
+        context.semantics_ir().GetType(name_node.type_id()));
```
jonmeow: Please add the unit test:

```
0[1];
```

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -5,17 +5,56 @@
...
+    auto cast_ind_id = context.ImplicitAsRequired(
+        parse_node, ind_node,
+        context.CanonicalizeType(SemanticsNodeId::BuiltinIntegerType));
+    auto cast_ind_node = context.semantics_ir().GetNode(cast_ind_id);
+
+    if ((context.semantics_ir().GetTypeAllowBuiltinTypes(
+             cast_ind_node.type_id()) != SemanticsNodeId::BuiltinError) &&
```
jonmeow: Why do you call GetType for this, instead of comparing with SemanticsTypeId::Error directly?

jonmeow (commented):
I think this is the right direction, but a couple more tests should be added. I'm not sure how much that'll change the flow of your code, but I think it's something worth thinking about assuming these tests are both crashing right now.

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -2,20 +2,65 @@
...
-  return context.TODO(parse_node, "HandleIndexExpression");
+  auto ind_node_id = context.node_stack().PopExpression();
+  auto name_node_id = context.node_stack().PopExpression();
+  auto name_node = context.semantics_ir().GetNode(name_node_id);
+
+  if (context.semantics_ir().GetNode(ind_node_id).kind() ==
+      SemanticsNodeKind::VarStorage) {
+    CARBON_DIAGNOSTIC(NondeterministicType, Error,
+                      "Type cannot be determined in compile time.");
+    context.emitter().Emit(parse_node, NondeterministicType);
```
jonmeow: Let's consider another case:

```
fn F() -> i32 {}

fn Run() {
  (1,)[F()];
}
```

The only thing actually allowed for ind_node_id is an integer literal. Consider checking for an integer literal, instead of checking for everything which is _not_ an integer literal but is still an int. Until more support is added, everything that is not an integer literal _must_ be an error.

This is perhaps the irony of the implicit cast call -- since ind_node_id must be an integer literal, the cast is actually not necessary to determine validity of the expression and can be removed without affecting the correctness.

toolchain/semantics/testdata/index/fail_non_tuple_access.carbon
```diff
@@ -0,0 +1,29 @@
...
+// CHECK:STDOUT: ]
+// CHECK:STDOUT: node_blocks: [
+// CHECK:STDOUT:   [
+// CHECK:STDOUT:   ],
+// CHECK:STDOUT: ]
+
+// CHECK:STDERR: fail_non_tuple_access.carbon:[[@LINE+3]]:1: Unrecognized declaration introducer.
+// CHECK:STDERR: 0[1];
+// CHECK:STDERR: ^
+0[1];
```
jonmeow:
```suggestion
fn Main() -> i32 {
  0[1];
}
```

Note here, make sure it's your error message being produced.

toolchain/lowering/lowering_handle.cpp
jonmeow: You seem to have only fixed this file, can you please fix other files too?

toolchain/semantics/testdata/index/fail_name_not_found.carbon
```diff
@@ -0,0 +1,46 @@
...
+// CHECK:STDOUT:     node+3,
+// CHECK:STDOUT:   ],
+// CHECK:STDOUT: ]
+
+// CHECK:STDERR: fail_name_not_found.carbon:[[@LINE+6]]:14: Name a not found
+// CHECK:STDERR: var b: i32 = a[0];
+// CHECK:STDERR:              ^
+// CHECK:STDERR: fail_name_not_found.carbon:[[@LINE+3]]:17: Invalid index expression.
+// CHECK:STDERR: var b: i32 = a[0];
+// CHECK:STDERR:                 ^
```
jonmeow: When you have an error coming in, the goal should be to not error again on that error.

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -2,20 +2,65 @@
...
+                                         ParseTree::Node /*parse_node*/)
+    -> bool {
+  // Leave the expression on the stack for IndexExpression.
+  return true;
 }
 
 auto SemanticsHandleIndexExpression(SemanticsContext& context,
                                     ParseTree::Node parse_node) -> bool {
-  return context.TODO(parse_node, "HandleIndexExpression");
+  auto ind_node_id = context.node_stack().PopExpression();
```
jonmeow: From a style perspective, please don't abbreviate "index" as "ind". I don't think it's a common abbreviation of index, and abbreviation is generally discouraged. You can find notes on abbreviation at: https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -2,20 +2,65 @@
...
+    if (cast_ind_node.type_id() != SemanticsTypeId::Error &&
+        name_type.kind() == SemanticsNodeKind::TupleType) {
+      auto ind_val = context.semantics_ir()
+                         .GetIntegerLiteral(cast_ind_node.GetAsIntegerLiteral())
+                         .getSExtValue();
+      auto type_block_id = name_type.GetAsTupleType();
+      auto type_block = context.semantics_ir().GetTypeBlock(type_block_id);
+      if (ind_val >= static_cast<int>(type_block.size())) {
+        CARBON_DIAGNOSTIC(OutOfBoundsAccess, Error,
+                          "Index: {0} is out of bound {1}.", int64_t, int64_t);
```
jonmeow: How about something more like:

```
"Index `{0}` is past the end of `{1}`.", int64_t, std::string);
```

... sending the type as {1}

"past the end" makes it clearer that this is a greater-than reason for out-of-bounds than just "out of bound". Providing the type lets people see how it's being evaluated.

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -2,20 +2,65 @@
...
+
+  if (context.semantics_ir().GetNode(ind_node_id).kind() ==
+      SemanticsNodeKind::VarStorage) {
+    CARBON_DIAGNOSTIC(NondeterministicType, Error,
+                      "Type cannot be determined in compile time.");
+    context.emitter().Emit(parse_node, NondeterministicType);
+  } else if (name_node.kind() != SemanticsNodeKind::VarStorage) {
+    CARBON_DIAGNOSTIC(InvalidIndexExpression, Error,
+                      "Invalid index expression.");
+    context.emitter().Emit(parse_node, InvalidIndexExpression);
```
jonmeow: Let's consider another case:

```
fn F() -> (i32,) { return (0,); }

fn Run() -> i32 {
  return F()[0];
}
```

You should probably be looking at the type of the node, not the kind.

jonmeow (commented):
I think it would be good to consider a little more what syntaxes indexing should work with, and not work with. Some tests noted.

toolchain/semantics/semantics_node_stack.h
```diff
@@ -256,6 +256,8 @@ class SemanticsNodeStack {
       case ParseNodeKind::StructLiteral:
       case ParseNodeKind::StructFieldType:
       case ParseNodeKind::StructTypeLiteral:
+      case ParseNodeKind::IndexExpressionStart:
+      case ParseNodeKind::IndexExpression:
```
jonmeow: Please fix the lexical ordering of these.

toolchain/semantics/semantics_ir.cpp
```diff
@@ -286,6 +286,7 @@ auto SemanticsIR::StringifyType(SemanticsTypeId type_id) -> std::string {
       case SemanticsNodeKind::CrossReference:
       case SemanticsNodeKind::FunctionDeclaration:
       case SemanticsNodeKind::IntegerLiteral:
+      case SemanticsNodeKind::Index:
```
jonmeow, line 369: Please fix the lexical ordering of this.

toolchain/semantics/semantics_node.h
```diff
@@ -379,6 +379,9 @@ class SemanticsNode {
   using IntegerLiteral = Factory<SemanticsNodeKind::IntegerLiteral,
                                  SemanticsIntegerLiteralId /*integer_id*/>;
 
+  using Index = Factory<SemanticsNodeKind::Index, SemanticsNodeId /*tuple_id*/,
+                        SemanticsNodeId /*tuple_index*/>;
```
jonmeow: Please fix the lexical ordering of this.

toolchain/semantics/semantics_node_stack.h
```diff
@@ -256,6 +256,8 @@ class SemanticsNodeStack {
       case ParseNodeKind::StructLiteral:
       case ParseNodeKind::StructFieldType:
       case ParseNodeKind::StructTypeLiteral:
+      case ParseNodeKind::IndexExpressionStart:
```
jonmeow: Can this be removed as unused now?

toolchain/lowering/lowering_handle.cpp
jonmeow: This was marked as resolved again, but I'm not seeing changes for this: I'm marking the other locations I'm seeing.

toolchain/semantics/semantics_handle_index.cpp
jonmeow: It looks like you're passing the SemanticsNodeKind instead of the type. What I meant was a std::string if the type, which you should be able to get with `semantics_ir().StringifyType(name_node.type_id())`

jonmeow:
This looks better now, just a few last small fixes.

jonmeow (approved):
Thanks! I think this is in a good state to merge now.

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -2,20 +2,63 @@
...
+  auto name_node = context.semantics_ir().GetNode(name_node_id);
+  auto name_type_id =
+      context.semantics_ir().GetTypeAllowBuiltinTypes(name_node.type_id());
+  auto name_type_node = context.semantics_ir().GetNode(name_type_id);
+
+  if (name_type_node.kind() == SemanticsNodeKind::TupleType &&
+      index_node.kind() == SemanticsNodeKind::IntegerLiteral) {
+    auto index_val = context.semantics_ir()
+                         .GetIntegerLiteral(index_node.GetAsIntegerLiteral())
+                         .getSExtValue();
```
zygoloid: This will assert if the index doesn't fit into a 64-bit signed integer. You should either check for that before calling `getSExtValue`, or remove the `getSExtValue` call so that we do a full-width comparison against the tuple size below; please also add a test for the case of a very large index.

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -2,20 +2,63 @@
...
+          context.semantics_ir().StringifyType(name_node.type_id()));
+    } else {
+      context.AddNodeAndPush(parse_node, SemanticsNode::Index::Make(
+                                             parse_node, type_block[index_val],
+                                             name_node_id, index_node_id));
+      return true;
+    }
+  } else if (index_node.kind() != SemanticsNodeKind::IntegerLiteral) {
+    CARBON_DIAGNOSTIC(NondeterministicType, Error,
+                      "Type cannot be determined in compile time.");
```
zygoloid:
```suggestion
                      "Type cannot be determined at compile time.");
```

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -2,20 +2,65 @@
...
+  auto name_node = context.semantics_ir().GetNode(name_node_id);
+  auto name_type_id =
+      context.semantics_ir().GetTypeAllowBuiltinTypes(name_node.type_id());
+  auto name_type_node = context.semantics_ir().GetNode(name_type_id);
+
+  if (name_type_node.kind() == SemanticsNodeKind::TupleType &&
+      index_node.kind() == SemanticsNodeKind::IntegerLiteral) {
+    auto index_val = *(context.semantics_ir()
+                           .GetIntegerLiteral(index_node.GetAsIntegerLiteral())
+                           .getRawData());
```
zygoloid: `getRawData` returns a pointer to the start of the `APInt` storage, which can in general be an array of multiple `uint64_t`s. This is now only looking at the first word of the `APInt` data, which means the comparison below won't do the right thing for (eg) `0x10000000000000000`.

Instead, you can compare the `APInt` value itself:
```suggestion
    const auto& index_val = *(context.semantics_ir()
                                  .GetIntegerLiteral(index_node.GetAsIntegerLiteral()));
```

(Additional suggestions below to adapt the code to work with `APInt`.)

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -2,20 +2,65 @@
...
+    auto type_block =
+        context.semantics_ir().GetTypeBlock(name_type_node.GetAsTupleType());
+
+    if (index_val >= static_cast<uint64_t>(type_block.size())) {
```
zygoloid:
```suggestion
    if (index_val.uge(type_block.size())) {
```
`APInt::uge` means "unsigned greater-than or equal".

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -2,20 +2,65 @@
...
+      CARBON_DIAGNOSTIC(OutOfBoundsAccess, Error,
+                        "Index `{0}` is past the end of `{1}`.", uint64_t,
```
zygoloid:
```suggestion
                        "Index `{0}` is past the end of `{1}`.", llvm::APSInt,
```

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -2,20 +2,65 @@
...
+                        std::string);
+      context.emitter().Emit(
+          parse_node, OutOfBoundsAccess, index_val,
```
zygoloid:
```suggestion
          parse_node, OutOfBoundsAccess, llvm::APSInt(index_val, /*isUnsigned=*/true),
```
We need to print as `APSInt` instead of as `APInt` so that we don't treat the high bit as a sign bit.

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -2,20 +2,65 @@
...
+          context.semantics_ir().StringifyType(name_node.type_id()));
+    } else {
+      context.AddNodeAndPush(parse_node, SemanticsNode::Index::Make(
+                                             parse_node, type_block[index_val],
```
zygoloid:
```suggestion
                                             parse_node, type_block[index_val.getZExtValue()],
```

zygoloid (approved):

PR #3045, opened 2023-08-01
Fix newline formatting in stale-issue message

geoffromer:
YAML string literals don't seem to support escape sequences like `\n`, so `\n\n\n` was showing up literally in the issue messages.


jonmeow (approved):
Thanks! Hopefully this works.

Issue #2770, opened 2023-04-14
Terminology for internal and external implementations


josh11b:
The decision #995 retires the keyword `external` to mark external implementations, and instead uses `extend` to mark internal implementations. Should we change the terminology in the documentation to match? If so, what terminology should we use instead?

chandlerc:
I think we should change to match.

"extended implementations" and "implementations" would be my initial suggestions.

There is a question of how important it is to have a dedicated or positive term for the non-extended case?

josh11b:
The non-extended case has different rules, for example:
- external implementations may be defined in different libraries
- blanket implementations must be external

chandlerc:
> The non-extended case has different rules.

Yes, but that alone doesn't make it clear to me _how important_ it is to have a dedicated term? I'm guessing you're trying to say that it is somewhat important, but I'm still not sure where that leaves things.

I also don't have any ideas for what the term should be here, but if you or others do, then maybe its a moot point.

Unrelated minor question:
 
> * blanket implementations must be external

Isn't this just *conditional* blanket implementations? I may be getting confused though.

josh11b:
> Isn't this just _conditional_ blanket implementations? I may be getting confused though.

We don't say "conditional blanket implementations" since it is redundant, all blanket implementations are of the form "for all types satisfying a criteria..."

mconst:
I'd avoid introducing a separate term for non-extended impls.  I think your examples both become clearer if we phrase them in terms of where you're allowed to write `extend`:

> * blanket implementations must be external

I'd say "you can't extend a blanket impl".

> * external implementations may be defined in different libraries

This one feels kind of obvious given #995, so we might not need to state it at all.  But if it comes up, I'd say something like "you can't use `extend` outside the class definition, but you can always impl your own interface for the class and then call those methods using qualified member access".

In general, to get the maxmium benefit from #995, I think we should try to treat `impl` and `extend` as separate concepts as much as possible.  If we run into a lot of cases where that doesn't work, then that would be a sign to maybe revisit #995, but these examples seem like they fit the model nicely.

josh11b:
> In general, to get the maxmium benefit from #995, I think we should try to treat `impl` and `extend` as separate concepts as much as possible. If we run into a lot of cases where that doesn't work, then that would be a sign to maybe revisit #995, but these examples seem like they fit the model nicely.

I find this argument compelling.

josh11b:
[...]

PR #2360, opened 2022-10-28
Types are values of type `type`

zygoloid:
Define a "type" to be a value of type `type`. Contexts expecting a type perform an implicit conversion to `type`. Values like `()` and `(i32, i32)` and `{}` are no longer types, but instead implicitly convert to `type`. Values of interface or constraint type (now called "facets") are similarly not formally types but implicitly convert to `type`.

PR #3048, opened 2023-08-01
Update Generics terminology document

josh11b:
This reflects changes from a number of approved proposals:
- #920 : concrete statements about orphan and overlap in Carbon
- #2138 : "generic" -> "checked generic", "template" -> "template generic"
- #2188 : binding patterns are forbidden in type position
- #2360 : "type", "facet type", "facet". Note: I am not using the term "generic type" from #2360 since that meaning conflicts with the generally accepted meaning of "generic type" of a type with a compile-time parameter.
- #2760 / #2770 : internal/external impl -> extending impl
- #2964 : "symbolic constant" and "template constant"

### DIFF
--- a/docs/design/generics/terminology.md
+++ b/docs/design/generics/terminology.md
@@ -465,7 +465,7 @@ requires.
 
 ### Extending an impl
 
-A type that _extends_ the implementation an interface has all the named members
+A type that _extends_ the implementation of an interface has all the named members
 of the interface as named members of the type. This means that the members of
 the interface are available by way of both
 [simple member access and qualified member access expressions](#member-access).
--- a/docs/design/generics/terminology.md
+++ b/docs/design/generics/terminology.md
@@ -122,7 +122,7 @@ Expected difference between checked and template parameters:
   <tr>
    <td>name lookup resolved for definitions in isolation ("early")
    </td>
-   <td>name lookup can use information from calls (name lookup may be "late")
+   <td>name lookup can use information from arguments (name lookup may be "late")
    </td>
   </tr>
   <tr>
@@ -325,6 +325,9 @@ time.
 
 ## Facet type
 
+> **TODO:** Documents using the obsolete term "type-of-type" should be
+> updated to say "facet type" instead.
+
 A _facet type_ is a [type](#types-and-type) whose values are some subset of the
 values of `type`, determined by a set of [type constraints](#type-constraints):
 
@@ -357,7 +360,7 @@ which is `i32`, in those contexts.
 
 ## Type expression
 
-A _type expression_ is an expression that is being used as a type. In some
+A _type expression_ is an expression that can be used as a type. In some
 cases, what is written in the source code is a value, like a [facet](#facet) or
 tuple of types, that is not a type but has an implicit conversion to `type`. In
 those cases, we are concerned with the type value after the implicit conversion.
@@ -366,7 +369,7 @@ those cases, we are concerned with the type value after the implicit conversion.
 
 We use the term _facet binding_ to refer to the name introduced by a
 [constant binding pattern](#bindings) (using `:!` with or without the `template`
-modifier) with a [facet type](#facet-type). In the binding pattern
+modifier) where the declared type is a [facet type](#facet-type). In the binding pattern
 `T:! Hashable`, `T` is a facet binding, and the value of `T` is a
 [facet](#facet).
 
--- a/docs/design/generics/terminology.md
+++ b/docs/design/generics/terminology.md
@@ -465,9 +465,9 @@ requires.
 
 ### Extending an impl
 
-A type that _extends_ the implementation of an interface has all the named members
-of the interface as named members of the type. This means that the members of
-the interface are available by way of both
+A type that _extends_ the implementation of an interface has all the named
+members of the interface as named members of the type. This means that the
+members of the interface are available by way of both
 [simple member access and qualified member access expressions](#member-access).
 
 If a type implements an interface without extending, the members of the
--- a/docs/design/generics/terminology.md
+++ b/docs/design/generics/terminology.md
@@ -293,7 +293,7 @@ are three kinds of binding patterns, corresponding to
     written using a `:`, as in `x: i32`.
 -   A _symbolic constant binding pattern_ or _symbolic binding pattern_ binds to
     a compile-time value that is not known when type checking, and is used to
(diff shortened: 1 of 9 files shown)


docs/design/generics/terminology.md
```diff
@@ -107,7 +122,7 @@ Expected difference between generics and templates:
   <tr>
    <td>name lookup resolved for definitions in isolation ("early")
    </td>
-   <td>some name lookup may require information from calls (name lookup may be "late")
+   <td>name lookup can use information from calls (name lookup may be "late")
```
geoffromer: Optional drive-by fix:

```suggestion
   <td>name lookup can use information from arguments (name lookup may be "late")
```

docs/design/generics/terminology.md
```diff
@@ -250,21 +265,111 @@ fail.
...
+A _type_ is a value of type `type`. Conversely, `type` is the type of all types.
+
+Expressions in type position, for example a [binding type expression](#bindings)
+or the return type of a function, are implicitly cast to type `type`. This means
+that it is legal to put a value that is not a type where a type is expected, as
+long as it has an implicit conversion to `type` that may be performed at compile
+time.
+
+## Facet type
+
```
geoffromer, line 327: After this PR, a lot of uses of the term "type-of-type" will link here. I assume you're planning to update them in a separate PR in order to keep this one small, but I'd prefer to minimize even temporary inconsistency. As an alternative to fixing all those usages in this PR, how about something like this?

```suggestion

> **TODO:** Documents using the obsolete term "type-of-type" should be
> updated to say "facet type" instead.

```

docs/design/generics/terminology.md
```diff
@@ -250,21 +265,111 @@ fail.
...
+`type`, so it is a facet that is not a type. However, in places where a type is
+expected, for example in a [binding type expression](#bindings) or after the
+`->` in a function declaration, there is an automatic implicit conversion to
+`type`. This means that a facet may be used in those positions. For example, the
+facet `i32 as Hashable` will implicitly convert to `(i32 as Hashable) as type`,
+which is `i32`, in those contexts.
+
+## Type expression
+
+A _type expression_ is an expression that is being used as a type. In some
```
geoffromer: I continue to be uncomfortable defining "type expression" to be a property of the context, not a property of the expression itself. Would this work instead?

```suggestion
A _type expression_ is an expression that can be used as a type. In some
```

docs/design/generics/terminology.md
```diff
@@ -250,21 +265,111 @@ fail.
...
+cases, what is written in the source code is a value, like a [facet](#facet) or
+tuple of types, that is not a type but has an implicit conversion to `type`. In
+those cases, we are concerned with the type value after the implicit conversion.
+
+## Facet binding
+
+We use the term _facet binding_ to refer to the name introduced by a
+[constant binding pattern](#bindings) (using `:!` with or without the `template`
+modifier) with a [facet type](#facet-type). In the binding pattern
```
geoffromer:
```suggestion
modifier) where the declared type is a [facet type](#facet-type). In the binding pattern
```

docs/design/generics/terminology.md
```diff
@@ -349,27 +452,24 @@ instead of associated entity.
...
+implement a named constraint as a way to implement all of the interfaces it
+requires.
 
-### Internal impl
+### Extending an impl
 
-A type that implements an interface _internally_ has all the named members of
-the interface as named members of the type. This means that the members of the
-interface are available by way of both
+A type that _extends_ the implementation an interface has all the named members
```
geoffromer: This terminology suggests that there's a point in time where the implementation exists, but hasn't yet been extended, which doesn't match my intuition (or the syntax, where `extend` comes first). Could we instead say the type "extends the interface", or "implements the interface as an extension"?

docs/design/generics/terminology.md
```diff
@@ -349,27 +455,24 @@ instead of associated entity.
...
```
geoffromer:
```suggestion
A type that _extends_ the implementation of an interface has all the named members
```

docs/design/generics/terminology.md
josh11b: The _syntax_ we decided on is `extend impl as`..., so I'm trying to make the terminology match. We are going to have other kinds of extending (base class, mixin), so whatever terminology we use should hopefully work consistently across those cases.

docs/design/generics/terminology.md
```diff
@@ -250,21 +265,114 @@ fail.
...
+declare function parameters, in `let` and `var` declarations, as well as to
+declare [generic parameters](#generic-means-compile-time-parameterized). There
+are three kinds of binding patterns, corresponding to
+[the three value phases](/docs/design/README.md#value-categories-and-value-phases):
+
+-   A _runtime binding pattern_ binds to a dynamic value at runtime, and is
+    written using a `:`, as in `x: i32`.
+-   A _symbolic constant binding pattern_ or _symbolic binding pattern_ binds to
+    a compile-time value that is not known when type checking, and is used to
+    declare [checked generics](#checked-versus-template-parameters) parameters.
```
zygoloid:
```suggestion
    declare [checked generic](#checked-versus-template-parameters) parameters.
```

docs/design/generics/terminology.md
```diff
@@ -736,40 +842,39 @@ types for which no implementation exists will be rejected at the call site:
 DoAdd(apple, orange);
 ```
 
+The type of an interface parameters and associated constants is commonly a
+[facet type](#facet-type), but not always. For example, one might have an
+integer type and be used to specify the size of an array type.
```
zygoloid: I misread "one might have an integer type" as "[a person] might have an integer type" at first. Maybe:

```suggestion
[facet type](#facet-type), but not always. For example, an interface parameter
that specifies an array bound might have an integer type.
```

zygoloid (approved):

docs/design/generics/terminology.md
josh11b: @zygoloid pointed out to me that it is in fact possible (though weird) to implement an interface for a type that is forward declared, and it only extends when the class is later defined.

zygoloid: I think you would be able to define the implementation first and extend it later:
```carbon
class A;
impl A as X { fn F() {} }
class A {
  extend impl A as X;
}
```
... though this is certainly not going to be the common case. But I think the implication that defining the implementation and extending it are separate steps is a correct implication.

PR #3049, opened 2023-08-02
Remove the type field from semantics nodes that don't produce values of that type.

zygoloid:
For `Assign` and `ReturnExpression`, this field wasn't used for anything. For `StructTypeField`, we stored the type of the field here, and now store it as an argument of the node instead.


toolchain/semantics/semantics_node.h
```diff
@@ -431,6 +433,8 @@ class SemanticsNode {
 
   auto parse_node() const -> ParseTree::Node { return parse_node_; }
   auto kind() const -> SemanticsNodeKind { return kind_; }
+
+  // Get the type of the value produced by evaluating this node.
```
jonmeow:
```suggestion
  // Gets the type of the value produced by evaluating this node.
```

> Function comments should be written with an implied subject of _This function_ and should start with the verb phrase; for example, "Opens the file", rather than "Open the file".
https://google.github.io/styleguide/cppguide.html#Function_Comments

jonmeow (approved):
Yes, agreed this makes sense to do.

PR #3050, opened 2023-08-02
Migrate explorer tests to file_test and remove lit support.

jonmeow:
This migrates explorer tests to file_test, using the new --autoupdate functionality. Per discussion, the trace tests that were using "not" output are mostly migrated to checking full output. The main exception is tests that were including trace output from the prelude: the prelude output is pretty long (multiple MB already) and I think it wasn't the intent to include, only trace output from the small program.

This is the remaining use of lit support, so the supporting libraries are also removed here.


jonmeow:
Relevant commit is https://github.com/carbon-language/carbon-lang/commit/ab4cc2fdb62af00fa8f0ebed44eaa239f0e08f46

explorer/file_test.cpp
```diff
@@ -17,7 +18,12 @@ namespace {
 
 class ExplorerFileTest : public FileTestBase {
  public:
-  using FileTestBase::FileTestBase;
+  explicit ExplorerFileTest(std::filesystem::path path)
+      : FileTestBase(std::move(path)),
+        prelude_line_re_(R"(prelude.carbon:(\d+))"),
+        timing_re_(R"((Time elapsed in \w+: )\d+(ms))") {
+    CARBON_CHECK(prelude_line_re_.ok()) << prelude_line_re_.error();
```
zygoloid, line 25: Why do we check one of these but not the other?

zygoloid (approved):

explorer/file_test.cpp
jonmeow: Added a check for the other. Just what I'd run into mistakes with. :)

PR #3052, opened 2023-08-02
Adds LLVM hardening flag for cc_toolchain libc++ for LLVM >= 17

samkhn:
LLVM 17 replaces ENABLE_ASSERTIONS with ENABLE_HARDEDNING. LLVM 18 is temporarily warning that ENABLE_ASSERTIONS is deprecated. Since we treat warnings as errors, this causes build failures for LLVM 18.

This commit can probably be safely removed/undone once LLVM hardening removes the check [here](https://github.com/llvm/llvmproject/blob/667602793bef72e64367ac46cbaf96e3afaea22e/libcxx/include/__config#L211).

### DIFF
--- a/bazel/cc_toolchains/clang_cc_toolchain_config.bzl
+++ b/bazel/cc_toolchains/clang_cc_toolchain_config.bzl
@@ -512,9 +512,6 @@ def _impl(ctx):
     # With clang 14 and lower, we expect it to be built with libc++ debug
     # support. In later LLVM versions, we expect the assertions define to work.
     # clang 17 deprecates LIBCPP_ENABLE_ASSERTIONS in favor of HARDENED_MODE.
-    # TODO(samkhn): Once LLVM makes ENABLE_ASSERTIONS equivalent to
-    # ENABLE_HARDENING, remove the HARDENED_MODE check
-    # https://github.com/llvm/llvm-project/blob/main/libcxx/include/__config
     if clang_version and clang_version <= 14:
         libcpp_debug_flags = ["-D_LIBCPP_DEBUG=1"]
     elif clang_version and clang_version <= 16:
--- a/bazel/cc_toolchains/clang_cc_toolchain_config.bzl
+++ b/bazel/cc_toolchains/clang_cc_toolchain_config.bzl
@@ -511,8 +511,14 @@ def _impl(ctx):
 
     # With clang 14 and lower, we expect it to be built with libc++ debug
     # support. In later LLVM versions, we expect the assertions define to work.
+    # clang 17 deprecates LIBCPP_ENABLE_ASSERTIONS in favor of HARDENED_MODE.
+    # TODO(samkhn): Once LLVM makes ENABLE_ASSERTIONS equivalent to
+    # ENABLE_HARDENING, remove the HARDENED_MODE check
+    # https://github.com/llvm/llvm-project/blob/main/libcxx/include/__config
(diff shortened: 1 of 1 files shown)


bazel/cc_toolchains/clang_cc_toolchain_config.bzl
```diff
@@ -511,8 +511,14 @@ def _impl(ctx):
...
+    # clang 17 deprecates LIBCPP_ENABLE_ASSERTIONS in favor of HARDENED_MODE.
+    # TODO(samkhn): Once LLVM makes ENABLE_ASSERTIONS equivalent to
+    # ENABLE_HARDENING, remove the HARDENED_MODE check
+    # https://github.com/llvm/llvm-project/blob/main/libcxx/include/__config
     if clang_version and clang_version <= 14:
         libcpp_debug_flags = ["-D_LIBCPP_DEBUG=1"]
+    elif clang_version and clang_version <= 17:
+        libcpp_debug_flags = ["-D_LIBCPP_ENABLE_HARDENED_MODE=1"]
     else:
         libcpp_debug_flags = ["-D_LIBCPP_ENABLE_ASSERTIONS=1"]
```
jonmeow:
```suggestion
    elif clang_version and clang_version <= 16:
        libcpp_debug_flags = ["-D_LIBCPP_ENABLE_ASSERTIONS=1"]
    else:
        libcpp_debug_flags = ["-D_LIBCPP_ENABLE_HARDENED_MODE=1"]
```

The code as written will set ENABLE_HARDENED_MODE in 17 and earlier versions; 18 and later will get ENABLE_ASSERTIONS. Did the comparison just get reversed when I was commenting about the ==? i.e., that it should be ASSERTIONS pre-17, then HARDENED_MODE in 17 and later, as in the suggested edit.

bazel/cc_toolchains/clang_cc_toolchain_config.bzl
```diff
@@ -511,8 +511,14 @@ def _impl(ctx):
 
     # With clang 14 and lower, we expect it to be built with libc++ debug
     # support. In later LLVM versions, we expect the assertions define to work.
+    # clang 17 deprecates LIBCPP_ENABLE_ASSERTIONS in favor of HARDENED_MODE.
+    # TODO(samkhn): Once LLVM makes ENABLE_ASSERTIONS equivalent to
+    # ENABLE_HARDENING, remove the HARDENED_MODE check
+    # https://github.com/llvm/llvm-project/blob/main/libcxx/include/__config
```
jonmeow: Reading https://reviews.llvm.org/D154997, I think the intent is to replace LIBCPP_ENABLE_ASSERTIONS with LIBCPP_ENABLE_HARDENED_MODE. So the change in v18 would be to drop support for LIBCPP_ENABLE_ASSERTIONS (or possibly make it an error). Wouldn't we then still need to keep setting LIBCPP_ENABLE_HARDENED_MODE to get the current behavior?

jonmeow (commented):
Thanks! Just a couple questions to make sure I understand the HARDENED_MODE direction correctly.

bazel/cc_toolchains/clang_cc_toolchain_config.bzl
samkhn: Ah thanks for catching that. Removed this with 4c685f4

jonmeow (approved):
Thank you!

PR #3053, opened 2023-08-02
Update contribution tools with bazel versioning details

samkhn:
You need bazel version 6.0.0 (with --host_per_file_copt) to build carbon. This adds versioning info and adds that Linux, Mac and Windows can use bazelisk.

### DIFF
--- a/docs/project/contribution_tools.md
+++ b/docs/project/contribution_tools.md
@@ -130,7 +130,7 @@ These tools are essential for work on Carbon.
             outdated, and not be upgraded.
 -   Main tools
     -   [Bazel](https://www.bazel.build/)
-        -   At least bazel version 6.0.0.
+        -   NOTE: See [the bazelisk config](/.bazeliskrc) for a supported version.
     -   [Bazelisk](https://docs.bazel.build/versions/master/install-bazelisk.html)
         (for macOS, Linux, Windows): Handles Bazel versions.
     -   [Clang](https://clang.llvm.org/) and [LLVM](https://llvm.org/)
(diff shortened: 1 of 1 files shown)


docs/project/contribution_tools.md
```diff
@@ -130,8 +130,9 @@ These tools are essential for work on Carbon.
             outdated, and not be upgraded.
 -   Main tools
     -   [Bazel](https://www.bazel.build/)
+        -   At least bazel version 6.0.0.
```
jonmeow:
```suggestion
        -   NOTE: See [the bazelisk config](/.bazeliskrc) for a supported version.
```

Thanks for pointing this out. We have this in a config file, how about pointing at that if someone needs a specific version? That way, we're less likely to forget to update a spot if the advice changes.

docs/project/contribution_tools.md
```diff
@@ -130,8 +130,9 @@ These tools are essential for work on Carbon.
...
     -   [Bazelisk](https://docs.bazel.build/versions/master/install-bazelisk.html)
-        (for macOS): Handles Bazel versions.
+        (for macOS, Linux, Windows): Handles Bazel versions.
```
jonmeow:
```suggestion
        (for macOS): Handles Bazel versions.
```

This documentation is explaining which tools are used according to the [install instructions above](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/project/contribution_tools.md#debian-or-ubuntu). While Bazelisk works on multiple platforms, and could be used instead of the instructions on this page, this list of tools is expressly what is "essential for work on Carbon", and Bazelisk is only essential on macOS.

jonmeow (approved):
Thanks!

PR #3055, opened 2023-08-03
Add support for `&` operator.

zygoloid:
Refactor type canonicalization so that we can reuse the same code for building a `T*` expression and for forming the type of an `&x` expression.

Add basic computation of expression category in order to check that we only take the address of durable reference expressions. This is currently computed on demand rather than being tracked as part of the semantics node, but in most cases can be determined by looking at only a single expression, so caching it in the node doesn't seem worthwhile yet. This decision should be revisited if we start doing more complex category calculations.

Also add trivial lowering support, but it doesn't work properly yet because lowering doesn't yet take the expression category into account.


toolchain/semantics/semantics_node.cpp
```diff
@@ -20,7 +20,7 @@ static auto PrintArgs(llvm::raw_ostream& out, std::pair<T0, T1> args) -> void {
   out << ", arg1: " << args.second;
 }
 
-void SemanticsNode::Print(llvm::raw_ostream& out) const {
+auto SemanticsNode::Print(llvm::raw_ostream& out) const -> void {
```
chandlerc, line 23: I thought `clang-tidy` canonicalized the other way? Or is this `clang-tidy` at work?

toolchain/semantics/semantics_ir.cpp
```diff
@@ -399,4 +403,80 @@ auto SemanticsIR::StringifyType(SemanticsTypeId type_id) -> std::string {
...
+      case SemanticsNodeKind::Return:
+      case SemanticsNodeKind::ReturnExpression:
+      case SemanticsNodeKind::StructTypeField:
+        return SemanticsExpressionCategory::NotExpression;
+
+      case SemanticsNodeKind::CrossReference: {
+        auto [xref_id, xref_node_id] = node.GetAsCrossReference();
+        ir = &semantics_ir.GetCrossReferenceIR(xref_id);
+        node_id = xref_node_id;
+        continue;
```
chandlerc, line 431: Fun implementation of recursion. ;]

This is just to help me learn -- what about cross references prevents this from being an infinite loop?

chandlerc (approved):
Would like to have a more complete change description if possible. =] This PR does quite a lot like adding the basic scaffolding for expression categories in the toolchain and reworking all of the type profiling. I think it's worth covering that and giving some of the motivation / background on the specific changes that ended up needed there (particularly around type profiling).

LGTM with the updated description, feel free to directly merge. Inline comments aren't blocking.

toolchain/semantics/semantics_node.cpp
zygoloid: `clang-tidy` doesn't canonicalize this one in either direction, but this is the prevailing style in the toolchain (6 occurrences with leading `void` versus at least 146 with trailing `void`).

The prevailing style in explorer is the exact reverse (269 leading `void` versus 15 trailing `void`). :shrug:

toolchain/semantics/semantics_ir.cpp
zygoloid: We only build cross references to the IRs for our dependencies, and our dependency graph is acyclic.

toolchain/semantics/semantics_node.cpp
chandlerc: 😭

PR #3056, opened 2023-08-03
Add formatted textual IR output

zygoloid:
Add a textual IR format to the toolchain.

The exact details of the format are somewhat arbitrary right now, and I expect them to change as we refine the semantics IR model, but at the moment they're somewhat directly following the current structure of the IR.

Semantics tests currently test both the "raw" format, which shows the details of the representation, and the textual format, which is somewhat higher level. We may want to revisit that decision once the textual format is a bit more stable, and test only the textual format in most of these tests, but for now it seems prudent to keep both sets of tests.


toolchain/semantics/semantics_ir.h
```diff
@@ -261,10 +262,12 @@ class SemanticsIR {
   }
 
   // Produces a string version of a type.
```
chandlerc: Add a comment to document how `in_type_context` impacts this?

toolchain/parser/testdata/index/fail_malformed_expr.carbon
```diff
@@ -3,6 +3,12 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
 // AUTOUPDATE
+
+// CHECK:STDERR: fail_malformed_expr.carbon:[[@LINE+3]]:17: Unexpected tokens before `]`.
+// CHECK:STDERR: var v: i32 = t[0,];
+// CHECK:STDERR:                 ^
+var v: i32 = t[0,];
```
chandlerc: Maybe a merge mis-fire here?

toolchain/semantics/semantics_ir_formatter.cpp
```diff
@@ -0,0 +1,584 @@
...
+#include "llvm/ADT/StringExtras.h"
+#include "llvm/ADT/StringSet.h"
+#include "llvm/Support/SaveAndRestore.h"
+#include "toolchain/lexer/tokenized_buffer.h"
+#include "toolchain/parser/parse_tree.h"
+
+namespace Carbon {
+
+namespace {
+// Assigns names to nodes, blocks, and scopes in the Semantics IR.
```
chandlerc, line 16: Once this stabilizes a bit, expand this with documentation of the naming scheme and some of the rationale behind it?

Alternatively, if creating a standalone documentation file for the textual format, could document it there and just have a reference here to those docs.

toolchain/semantics/semantics_ir_formatter.cpp
```diff
@@ -0,0 +1,584 @@
...
+
+    // Build each function scope.
+    for (int i = 0; i != semantics_ir.functions_size(); ++i) {
+      auto fn_id = SemanticsFunctionId(i);
+      auto fn_scope = GetScopeFor(fn_id);
+      const auto& fn = semantics_ir.GetFunction(fn_id);
+      GetScopeInfo(fn_scope).name =
+          "@" +
+          globals.AllocateName(*this,
+                               /*TODO*/ ParseTree::Node::Invalid,
```
chandlerc: Maybe add a bit of context for what this should be doing?

toolchain/semantics/semantics_ir_formatter.cpp
```diff
@@ -0,0 +1,584 @@
...
+
+    // Check for a builtin.
+    if (node_id.index < SemanticsBuiltinKind::ValidCount) {
+      return SemanticsBuiltinKind::FromInt(node_id.index).label().str();
+    }
+
+    auto& [node_scope, node_name] = nodes[node_id.index];
+    if (node_name.empty()) {
+      // This should not happen in valid IR.
+      return "<noderef " + llvm::itostr(node_id.index) + ">";
```
chandlerc: What do you think about here and elsewhere something like:

```suggestion
      // This should not happen in valid IR.
      return "<invalid noderef " + llvm::itostr(node_id.index) + ">";
```

toolchain/semantics/semantics_ir_formatter.cpp
```diff
@@ -0,0 +1,584 @@
...
+
+      // Append location information to try to disambiguate.
+      if (node.is_valid()) {
+        auto token = namer.parse_tree_.node_token(node);
+        name += ".L";
+        name += llvm::itostr(namer.tokenized_buffer_.GetLineNumber(token));
+        if (allocated.insert(name).second) {
+          return name;
+        }
+        name += ".";
```
chandlerc: I think it's useful to distinguish the column number from a non-location disambiguator.

I know it makes the name longer, but maybe:
```suggestion
        name += ".c";
```

I also find `c` a bit easier to separate from the number than `C`, and I'd probably use `l` for consistency, although there both `l` and `L` are hard to visually separate. Somewhat minor though.

toolchain/semantics/semantics_ir_formatter.cpp
```diff
@@ -0,0 +1,584 @@
...
+      }
+
+      // Append location information to try to disambiguate.
+      if (node.is_valid()) {
+        auto token = namer.parse_tree_.node_token(node);
+        name += ".L";
+        name += llvm::itostr(namer.tokenized_buffer_.GetLineNumber(token));
+        if (allocated.insert(name).second) {
+          return name;
+        }
```
chandlerc: Looking at the examples, I've found it a bit hard to use the line number alone when it is just the *first* entity on the line and subsequent ones have columns.

The reason is because I haven't seen the subsequent names yet when I see the first one without any column suffix, and the line given has a bunch of different constructs. I have to go find the next N things on that line, and eliminate those based on column number until one is remaining.

Ultimately, I think when there are multiple entries on a line, it's better to just put the column number in all of them.

The same thing comes up with multiple entries at the same column, but I didn't run into that as much reading examples -- not sure if that's just because it doesn't happen much or because it isn't as hard to disambiguate. But there also isn't anything to be done in that case.

chandlerc (commented):
Sorry I didn't finish a first pass here -- but I left a few comments on the naming scheme and on a few parts of the code. I'll work more on finishing the review as I can.

One thing I wanted to mention is potentially writing up a document in the tree as a reference for the textual format -- would that make sense? Doesn't need to be there to start or right now, I understand wanting to let the format evolve and settle a bit before investing lots in trying to document it. Mostly wanted to write down the thought.

toolchain/semantics/semantics_ir.h
zygoloid: Done.

toolchain/semantics/semantics_ir_formatter.cpp
zygoloid: I'll add documentation for the textual IR format as a follow-up.

toolchain/semantics/semantics_ir_formatter.cpp
zygoloid: Done.

toolchain/semantics/semantics_ir_formatter.cpp
zygoloid: Done, went with "unexpected" rather than "invalid" because the node reference itself is valid (references a valid node), but not one that we expected to be referenced by ID.

toolchain/semantics/semantics_ir_formatter.cpp
zygoloid: Format updated per in-person discussion.

toolchain/semantics/semantics_ir_formatter.cpp
zygoloid: Switched to using the shortest unambiguous name for an entity rather than giving a name to the first entity that wants it.

toolchain/semantics/semantics_ir_formatter.cpp
```diff
@@ -0,0 +1,654 @@
...
+#include "llvm/ADT/StringExtras.h"
+#include "llvm/ADT/StringSet.h"
+#include "llvm/Support/SaveAndRestore.h"
+#include "toolchain/lexer/tokenized_buffer.h"
+#include "toolchain/parser/parse_tree.h"
+
+namespace Carbon {
+
+namespace {
+// Assigns names to nodes, blocks, and scopes in the Semantics IR.
```
chandlerc, line 16: Adding a suggested edit adding TODOs here as a way for me to suggest things to consider for future work in a way that won't get lost in the sands of pull-request history, but definitely aren't interesting in this PR.

Feel free to reformat, move around, etc if you see a better place / way of documenting.

```suggestion
// Assigns names to nodes, blocks, and scopes in the Semantics IR.
//
// TODOs / future work ideas:
// - Add a documentation file for the textual format and link to the
//   naming section here.
// - Consider harmonizing `FooLiteral` vs. `foo_value` node names vs.
//   IR names.
//   - Also consider representing these as just `value` or `literal` 
//     in the IR and using the type to distinguish.
// - Add block names based on the control flow construct names (`for`,
//   `if`, `then`, `else`, ...). Either base this on the semantics or
//   just on the keywords used to build that control flow -- little
//   or no harm to getting it wrong. Tools will also want to be able
//   to query the construct that resulted in control flow.
```

toolchain/semantics/semantics_ir_formatter.cpp
```diff
@@ -0,0 +1,654 @@
...
+        CARBON_CHECK(value) << "cannot print a null name";
+        while (value->second.ambiguous && value->second.fallback) {
+          value = value->second.fallback.value_;
+        }
+        return value->first();
+      }
+
+      operator llvm::StringRef() const { return str(); }
+      operator llvm::Twine() const { return str(); }
+      operator std::string() const { return str().str(); }
```
chandlerc: I'm surprised all of these are needed rather than just `StringRef`? Extra surprised by making `std::string` construction be *implicit*.

(Happy with either learning why these are needed or narrowing to just `StringRef`, not really intended to be blocking.)

toolchain/semantics/semantics_ir_formatter.cpp
```diff
@@ -0,0 +1,654 @@
...
+          if (mark_ambiguous) {
+            // This name was allocated for a different node. Mark it as
+            // ambiguous and keep looking for a name for this node.
+            new_name.SetAmbiguous();
+          }
+        } else {
+          if (!best) {
+            best = new_name;
+          } else {
+            current.SetFallback(new_name);
```
chandlerc: To get a slightly easier to debug crash:

```suggestion
          } else {
            CARBON_CHECK(current);
            current.SetFallback(new_name);
```

toolchain/semantics/semantics_ir_formatter.cpp
```diff
@@ -0,0 +1,654 @@
...
+          }
+          current = new_name;
+        }
+        return added;
+      };
+
+      // All names start with the prefix.
+      name.insert(0, prefix.data(), prefix.size());
```
chandlerc: Does the `StringViewLike` `.insert` not work with `llvm::StringRef`? If so, that's... annoying.

toolchain/semantics/semantics_ir_formatter.cpp
```diff
@@ -0,0 +1,654 @@
...
+      // Append numbers until we find an available name.
+      name += ".";
+      auto name_size_without_counter = name.size();
+      for (int counter = 1;; ++counter) {
+        name.resize(name_size_without_counter);
+        name += llvm::itostr(counter);
+        if (add_name(/*mark_ambiguous=*/false)) {
+          return best;
+        }
+      }
```
chandlerc: Would using a string-stream be clearer? It seems a tiny bit better to me, but optional if you prefer as-is:

```suggestion
      // Append location information to try to disambiguate.
      llvm::raw_string_ostream name_os(name);
      if (node.is_valid()) {
        auto token = namer.parse_tree_.node_token(node);
        name_os << ".loc" << namer.tokenized_buffer_.GetLineNumber(token);
        add_name();

        name_os << "_" << namer.tokenized_buffer_.GetColumnNumber(token);
        add_name();
      }

      // Append numbers until we find an available name.
      name_os << ".";
      auto name_size_without_counter = name.size();
      for (int counter = 1;; ++counter) {
        name.resize(name_size_without_counter);
        name_os << counter;
        if (add_name(/*mark_ambiguous=*/false)) {
          return best;
        }
      }
```

toolchain/semantics/semantics_ir_formatter.cpp
```diff
@@ -0,0 +1,654 @@
...
+            scope.nodes.AllocateName(*this, node.parse_node(),
+                                     semantics_ir_.GetString(name_id).str())};
+      }
+    }
+
+    // Sequentially number all remaining values.
+    for (auto node_id : semantics_ir_.GetNodeBlock(block_id)) {
+      auto node = semantics_ir_.GetNode(node_id);
+      if (node.kind() != SemanticsNodeKind::BindName &&
+          node.kind().value_kind() != SemanticsNodeValueKind::None) {
```
chandlerc, line 290: Curious why this can't be part of the above loop over the nodes? Either adding that to the comment or merging are fine, mostly wasn't obvious to me when reading.

toolchain/semantics/semantics_ir_formatter.cpp
```diff
@@ -0,0 +1,654 @@
...
+  template <typename Kind>
+  auto FormatInstructionRHS(SemanticsNode node) -> void {
+    // By default, an instruction has a comma-separated argument list.
+    FormatArgs(Kind::Get(node));
+  }
+
+  // BindName is handled by the NodeNamer and doesn't appear in the output.
+  template <>
+  auto FormatInstruction<SemanticsNode::BindName>(SemanticsNodeId,
+                                                  SemanticsNode) -> void {}
```
chandlerc: I'm a bit surprised at this. If we have nodes to track the name binding, why not show them in the IR? I understand that we'll *also* actually use the bound name, just somewhat curious about the rationale here.

(Not blocking at all.)

chandlerc (approved):
LGTM with some non-blocking comments and some suggestions / alternatives below.

Mostly, I think this is a good initial cut. There is lots we can do to play with the format and other things to refine until it stabilizes a bit. But seems better to get an initial version into the tree and iterate / fix-forward.

toolchain/semantics/semantics_ir_formatter.cpp
zygoloid: These were an attempt to remove the ugly `.str().str()` calls, converting explicitly first to `StringRef` and then to `std::string`. But in the end, we only do that twice, which I think means these don't pull their weight. Removed and added the `.str().str()`s instead.

toolchain/semantics/semantics_ir_formatter.cpp
zygoloid: I didn't realize we were using a new enough standard library for that, and it does! Thanks.

toolchain/semantics/semantics_ir_formatter.cpp
zygoloid: I'm uncomfortable about modifying the `std::string` while it's being held by the ostream (even though it looks like `raw_string_ostream` doesn't cache anything about the string between output calls right now).

Done, creating the stream more frequently.

toolchain/semantics/semantics_ir_formatter.cpp
zygoloid: Added comment.

toolchain/semantics/semantics_ir_formatter.cpp
zygoloid: The only purpose of a `BindName` node in the IR currently is to give a name to some other node. We never reference `BindName` nodes or use them for any purpose other than as an annotation to use a specific name for another node, so including them in the IR would only add noise.

I've extended the comment here to explain.

PR #3058, opened 2023-08-03
Semantics & Lowering for simple assignment.

fasiddique:


toolchain/semantics/semantics_node_kind.def
```diff
@@ -28,6 +28,7 @@ CARBON_SEMANTICS_NODE_KIND(Invalid)
 CARBON_SEMANTICS_NODE_KIND(CrossReference)
 
 CARBON_SEMANTICS_NODE_KIND(Assign)
+CARBON_SEMANTICS_NODE_KIND(Assignment)
```
zygoloid: Can we use the same `Assign` node for both the variable declaration case and the assignment statement case? I think that was @jonmeow's intent here.

toolchain/semantics/semantics_handle_operator.cpp
```diff
@@ -51,7 +51,14 @@ auto SemanticsHandleInfixOperator(SemanticsContext& context,
               resume_block_id));
       return true;
     }
-
+    case TokenKind::Equal: {
+      // TODO: handle complex assignment expression such as `a = a + 1`.
+      context.ImplicitAsRequired(
+          parse_node, lhs_id, context.semantics_ir().GetNode(rhs_id).type_id());
```
zygoloid: I think we should do this the other way around: convert the RHS to the type of the LHS.

toolchain/semantics/semantics_handle_operator.cpp
```diff
@@ -51,7 +51,14 @@ auto SemanticsHandleInfixOperator(SemanticsContext& context,
...
+      context.AddNodeAndPush(parse_node, SemanticsNode::Assignment::Make(
+                                             parse_node, lhs_id, rhs_id));
```
zygoloid: We should check that the LHS is something that can be assigned to. I'm adding a mechanism to check that in #3055, so for now please just add a TODO here.

toolchain/lowering/testdata/operators/assignment.carbon
```diff
@@ -0,0 +1,37 @@
...
+// CHECK:STDOUT: ; ModuleID = 'assignment.carbon'
+// CHECK:STDOUT: source_filename = "assignment.carbon"
+// CHECK:STDOUT:
+// CHECK:STDOUT: %TupleLiteralType = type { {}, {} }
+// CHECK:STDOUT: %TupleLiteralType.0 = type { i32, i32 }
+// CHECK:STDOUT:
+// CHECK:STDOUT: define void @Main() {
+// CHECK:STDOUT:   %var = alloca i32, align 4
+// CHECK:STDOUT:   store i32 12, ptr %var, align 4
+// CHECK:STDOUT:   store i32 -7, ptr %var, align 4
```
zygoloid, line 15: Why is this `-7` and not `9`?

Actually... this seems to be a pre-existing issue with the lowering of integer values; don't worry about it.

toolchain/semantics/semantics_handle_operator.cpp
```diff
@@ -51,7 +51,15 @@ auto SemanticsHandleInfixOperator(SemanticsContext& context,
               resume_block_id));
       return true;
     }
-
+    case TokenKind::Equal: {
+      // TODO: handle complex assignment expression such as `a = a + 1`.
```
zygoloid: I would expect `a = a + 1;` already works; did you mean `a += 1;` here?

fasiddique: ah yes I meant to write a+=1. It's fixed now.

zygoloid (approved):

PR #2138, opened 2022-09-01
Checked and template generic terminology

josh11b:
Change terminology from "generics" and "templates" to "checked generics" and "template generics". Afterwards, "generics" will be an umbrella term that include templates.

PR #3060, opened 2023-08-03
Update generics goals doc

josh11b:
This reflects changes from #2138 , plus a number of other updates, including reflecting that we have decided to include templates in Carbon.


chandlerc (approved):
LGTM and merging, awesome!

PR #3061, opened 2023-08-03
Update generics overview

josh11b:
This reflects changes from a number of approved proposals:

- #2138 : "generic" -> "checked generic", "template" -> "template generic"
- #2360 : "type", "facet type", "facet". Note: I am not using the term "generic type" from #2360 since that meaning conflicts with the generally accepted meaning of "generic type" of a type with a compile-time parameter.
- #2760 / #2770 : internal/external impl -> extending impl
- #2964 : "symbolic constant" and "template constant"


docs/design/generics/overview.md
```diff
@@ -43,51 +43,79 @@ pointers to other design documents that dive deeper into individual topics.
...
+parameters, allowing
+[generic programming](https://en.wikipedia.org/wiki/Generic_programming). Carbon
+supports both
+[checked and template](terminology.md#checked-versus-template-parameters)
+generics.
+
+Template generics support a similar model as C++ templates, to help with interop
+and migration. They can be more convenient to write, and support some use cases,
+like [metaprogramming](https://en.wikipedia.org/wiki/Metaprogramming), that are
+difficult with checked generics.
```
zygoloid:
```suggestion
Template generics provide a similar model to C++ templates, to help with interop
and migration. They can be more convenient to write, and support some use cases,
like [metaprogramming](https://en.wikipedia.org/wiki/Metaprogramming), that are
difficult with checked generics.
```

I think a word other than "support" works better here in the context of the previous paragraph (I found the repeated use of "support" with different meanings hard to read).

docs/design/generics/overview.md
```diff
@@ -43,51 +43,79 @@ pointers to other design documents that dive deeper into individual topics.
...
-    [erased](terminology.md#type-erasure), except for the names defined in the
-    type-of-type.
--   _Deduced parameters_ are parameters whose values are determined by the
-    values and (most commonly) the types of the explicit arguments. Generic type
-    parameters are typically deduced.
--   A function with a generic type parameter can have the same function body as
-    an unparameterized one. Functions can freely mix generic, template, and
-    regular parameters.
+    their functionality. A given type may implement an interface at most once.
+-   Implementations may be declared inline in the body of a class definition, or
```
zygoloid: I read this as an exclusive "or", which doesn't seem right, given that the same implementation can be declared in both places. Maybe removing the comma would help:
```suggestion
-   Implementations may be declared inline in the body of a class definition or
```

docs/design/generics/overview.md
```diff
@@ -43,51 +43,79 @@ pointers to other design documents that dive deeper into individual topics.
...
+-   Types may _extend_ an implementation declared inline, in which case you can
+    directly call the interface's methods on those types.
+-   Out-of-line implementations may be defined in the library defining the
+    interface as an alternative to the type, or
+    [the library defining a type argument](#parameterized-impl-declarations).
+-   Interfaces may be used as the type of a generic parameter. Interfaces are
+    _facet types_, whose values are the subset of all types that implement the
+    interface. Facet types in general specify the capabilities and requirements
+    of the type. The value of a interface is called a _facet_, and are not
+    types, but are usable as types.
```
zygoloid:
```suggestion
    of the type. The value of a interface is called a _facet_. Facets are not
    types, but are usable as types.
```
(Plurality mismatch.)

docs/design/generics/overview.md
```diff
@@ -43,51 +43,79 @@ pointers to other design documents that dive deeper into individual topics.
...
+-   With a template generic, the concrete argument value used by the caller is
+    used for name lookup and typechecking. With checked generics, that is all
+    done with the declared restrictions expressed as the types of bindings in
+    the declaration. Inside the body of a checked generic with a facet
+    parameter, the API of the facet is just the names defined by the facet type.
+-   _Deduced parameters_ are parameters whose values are determined by the the
```
zygoloid:
```suggestion
-   _Deduced parameters_ are parameters whose values are determined by the
```

docs/design/generics/overview.md
```diff
@@ -329,20 +360,23 @@ fn PrintIt(p: Song*) {
...
+[qualified member access expression](#accessing-members-of-interfaces).
 
-A function can have a mix of generic, template, and regular parameters.
-Likewise, it's allowed to pass a template or generic value to a generic or
-regular parameter. _However, passing a generic value to a template parameter is
-future work._
+A function can have a mix of checked, template, and regular parameters. A
+checked parameter is defined using a symbolic binding pattern, a template
+parameter using a template binding pattern, and a regular parameter using a
+runtime binding pattern. Likewise, it's allowed to pass a symbolic or template
```
zygoloid:
```suggestion
A function can have a mix of checked, template, and regular parameters. A
checked parameter is defined using a symbolic binding pattern, a template
parameter is defined using a template binding pattern, and a regular parameter is defined using a
runtime binding pattern. Likewise, it's allowed to pass a symbolic or template
```

I originally read this as

> A checked parameter is defined using (a symbolic binding pattern, a template parameter using a template binding pattern, and a regular parameter using a runtime binding pattern).

Alternatively, I think something like this could work:

> Each kind of parameter is defined using a different syntax: a checked parameter is uses a symbolic binding pattern, a template parameter uses a template binding pattern, and a regular parameter uses a runtime binding pattern.

zygoloid (approved):

docs/design/generics/overview.md
josh11b: I used the alternative, but also noticed I needed to update the next sentence.

zygoloid (approved):

PR #3065, opened 2023-08-04
Move STDOUT CHECK lines for semantics tests to the end of the file.

zygoloid:
In preparation for making them reference line numbers.

Update the file_test infrastructure to allow tests to customize where their STDOUT / STDERR check lines are placed independently.


chandlerc (commented):
I like this output more, but I'm a bit hesitant to make this change w/o chatting w/ @jonmeow at least.

That said -- it *seems* like this should be really easy to revisit if anyone has concerns here, given that all of this is auto-updated. So as long as you don't see any likely challenges with iterating here if needed, I think it's fine to fix-forward if anything comes up.

It'd be good to update the autoupdate docs though: https://github.com/zygoloid/carbon-lang/tree/file_test_stdout_at_end/testing/file_test#:~:text=When%20autoupdating%2C%20CHECKs%20will%20be%20inserted%20starting%20below%20AUTOUPDATE

chandlerc:
Discussed with @zygoloid and it does seem like we can re-do this if when Jon is back we decide to go back to the original or go in yet another direction.

That said, I think the autoupdate docs still need to be updated, so that's I think the next step here.

zygoloid:
Having read through the doc inspired me to approach this differently, with separate in-file markers for where to put stdout and stderr. PR updated to do that, including a doc update this time :)

testing/file_test/README.md
```diff
@@ -75,18 +75,27 @@ Supported comment markers are:
...
+    When autoupdating, CHECKs will be inserted starting below the corresponding
+    AUTOUPDATE. When a CHECK has line information, autoupdate will try to insert
+    the CHECK immediately above the line it's associated with. When that
+    happens, any following CHECK lines without line information will immediately
+    follow, between the CHECK with line information and the associated line.
+
+    CHECK lines are not reordered or interleaved. If the STDOUT marker is first
+    in the file or the plain AUTOUPDATE marker is used, all STDOUT CHECKs will
+    precede all STDERR CHECKs, and if the STDERR marker is first in the file,
+    all STDERR CHECKs will precede all STDOUT CHECKs.
```
chandlerc, line 98: FWIW, this still seems a bit backwards to me.

I would generally expect all STDERR output and the associated checks to *precede* STDOUT output and the associated checks as a default.

And if we do that, does it cause most (if not all?) of the cases here to work automatically?

zygoloid: I tried that, but it does the wrong thing most of the time, because we put the STDOUT immediately after the STDERR, so instead of

```
// CHECK:STDOUT: lots
// CHECK:STDOUT: of
// CHECK:STDOUT: output

start of test
// CHECK:STDERR: something wrong: 'bad line'
bad line
```

we instead get

```
start of test
// CHECK:STDERR: something wrong: 'bad line'
// CHECK:STDOUT: lots
// CHECK:STDOUT: of
// CHECK:STDOUT: output
bad line
```

I also tried just forcing *all* STDOUT to the end of the file, which equally does the wrong thing in a lot of cases (when STDOUT references line numbers, it's no longer properly interleaved in the file).

zygoloid: See #3073 for an attempt to follow the direction of putting STDOUT as late as possible per in-person discussion.

chandlerc:
(assuming this is superseded by #3073)

zygoloid:
Closing in favor of #3073.

PR #3066, opened 2023-08-05
Add support for dereference operator.

zygoloid:
Also fix a couple of error-recovery issues exposed by the tests for this change.


toolchain/semantics/testdata/pointer/fail_dereference_not_pointer.carbon
```diff
@@ -0,0 +1,78 @@
...
+  // CHECK:STDERR:   ^
+  *n;
+  // CHECK:STDERR: fail_dereference_not_pointer.carbon:[[@LINE+3]]:3: Cannot dereference operand of non-pointer type () as type.
+  // CHECK:STDERR:   *();
+  // CHECK:STDERR:   ^
+  *();
+  // CHECK:STDERR: fail_dereference_not_pointer.carbon:[[@LINE+3]]:3: Cannot dereference operand of non-pointer type {} as type.
+  // CHECK:STDERR:   *{};
+  // CHECK:STDERR:   ^
+  *{};
```
chandlerc: Worth testing dereferencing a pointer *type*? `*(i32*);`

(Worth a custom error message?)

chandlerc (approved):
LG, a possible added test inline, but not-blocking.

toolchain/semantics/testdata/pointer/fail_dereference_not_pointer.carbon
zygoloid: Added a test and a custom note.

chandlerc: I do like the note. =]

FWIW, I was thinking of the confusion of an expression that _is_ a pointer type vs. an expression _of_ pointer type. Yes, the first is of type `type`, but wondering if we want to avoid users having to think about that.

Anyways, super minor point here.

PR #3073, opened 2023-08-08
Prefer to put STDOUT CHECK at the end of the file.

zygoloid:
Allow interleaving of STDOUT and STDERR check lines. Put STDOUT lines after the line they're attached to, and STDERR lines before. If no STDOUT check line is attached to any line, then put them all at the end of the file instead.

This is intended to better handle the case where stdout contains unreplaced mentions of line numbers, and also reflects that stdout is typically a consequence of the test rather than commentary on it, so placing it after the test seems likely to read better.


toolchain/lexer/testdata/basic_syntax.carbon
```diff
@@ -3,21 +3,22 @@
...
-// CHECK:STDOUT: { index:  3, kind: 'StringTypeLiteral', line: {{ *}}[[@LINE+4]], column:  8, indent: 1, spelling: 'String', has_trailing_space: true },
-// CHECK:STDOUT: { index:  4, kind:        'Identifier', line: {{ *}}[[@LINE+3]], column: 15, indent: 1, spelling: 'program', identifier: 1 },
-// CHECK:STDOUT: { index:  5, kind:        'CloseParen', line: {{ *}}[[@LINE+2]], column: 22, indent: 1, spelling: ')', opening_token: 2, has_trailing_space: true },
-// CHECK:STDOUT: { index:  6, kind:    'OpenCurlyBrace', line: {{ *}}[[@LINE+1]], column: 24, indent: 1, spelling: '{', closing_token: 10, has_trailing_space: true },
 fn run(String program) {
-  // CHECK:STDOUT: { index:  7, kind:            'Return', line: {{ *}}[[@LINE+3]], column:  3, indent: 3, spelling: 'return', has_trailing_space: true },
-  // CHECK:STDOUT: { index:  8, kind:        'Identifier', line: {{ *}}[[@LINE+2]], column: 10, indent: 3, spelling: 'True', identifier: 2 },
-  // CHECK:STDOUT: { index:  9, kind:              'Semi', line: {{ *}}[[@LINE+1]], column: 14, indent: 3, spelling: ';', has_trailing_space: true },
+// CHECK:STDOUT: [
```
chandlerc: This file is worth looking at to see a case where I think things aren't perfect with this scheme, but are generally fine.

The open `[` here is a bit odd, but I don't tihnk we can avoid that w/o something much fancier like `CHECK-LABEL` style things.

And the previous version had some oddities as well, such as the placement of the `]`. I actually prefer the way `EndOfFile` ends up handled here.

(no action needed here)

toolchain/parser/testdata/function/declaration/fail_missing_deduced_close.carbon
```diff
@@ -3,6 +3,19 @@
...
+// CHECK:STDERR: fn Div[();
+// CHECK:STDERR:       ^
+// CHECK:STDERR: fail_missing_deduced_close.carbon:[[@LINE+3]]:8: Expected parameter declaration.
+// CHECK:STDERR: fn Div[();
+// CHECK:STDERR:        ^
+fn Div[();
+
+// CHECK:STDERR: fail_missing_deduced_close.carbon:29:19: A `(` for parameters is required after deduced parameters.
+// CHECK:STDERR: // CHECK:STDOUT: ]
+// CHECK:STDERR:                   ^
```
chandlerc: While not the goal, this change surfaces I think a real problem with this diagnostic -- it is using the end-of-file location which results in a very strange diagnostic message here (and in a few other tests in this area).

But I don't think this is a bug with this change, rather an underlying issue with the parser's diagnostics here so we should fix it in a follow-up.

testing/file_test/README.md
```diff
@@ -88,6 +88,11 @@ Supported comment markers are:
     following CHECK lines without line information will immediately follow,
     between the CHECK with line information and the associated line.
 
+    STDERR checks are placed before the line they refer to if possible, and
+    otherwise as early in the file as possible. STDOUT check lines are placed
+    after the line they refer to if possible, and otherwise at the end of the
+    file.
```
chandlerc: I feel like it would read a bit better integrating this with the previous paragraph. But I can't make a suggested edit that covers both, so just suggesting here for you to use or rework if useful, but also feel free to replace with your own version:

```
    When autoupdating, CHECKs will be inserted starting below AUTOUPDATE.
    For CHECKs without line information, STDERR will be placed immediately
    below AUTOUPDATE, and STDOUT will be at the end of the file.

    When a CHECK has line information, autoupdate will try to insert the
    CHECK and any following CHECK lines without line information together
    and next to that line if possible. STDERR CHECKs will go immediately
    above the line, and STDOUT CHECKs immediately after. The goal is to
    reflect that STDERR usually reflects validation of a line, and STDOUT
    the result of the line.
```

chandlerc (approved):
Generally, this seems like a nice improvement. I've left a suggestion on the docs for `AUTOUPDATE` but likely not really blocking.

It's really hard to review this in GitHub because it won't load all the diffs and clicking on 100s of "Load diff" is no fun, but I went through with `git` on the command line. I've flagged a couple of files where the results are maybe slightly interesting so others can find them and look at them in the diff, but I think this is still a net-positive.

Maybe double check with @geoffromer that he's also OK with this? I'd like to get more thoughts, but lots of folks are busy / out. Fortunately, much like with #3065, if folks end up not liking this or it just doesn't work out well for whatever reason, it's super easy to change and revert or fix-forward given that we're *specifically* only changing things that are auto-updated. =D

Approving provided my docs suggestion makes sense, but happy to take another look if needed.

geoffromer (approved):

toolchain/lexer/testdata/basic_syntax.carbon
zygoloid: One interesting thing happening in this file (and a few others): we fail to replace the `line: 24` in the `EndOfFile` CHECK, because in general we can't cope with line replacements that refer to a CHECK line. (We can't track them as they get removed and recreated when generating the new test expectations, so we don't know what line is being referred to in the new output.)

Forcing the STDOUT to the end of the file means we run into this every time the output refers to EOF. Perhaps we should have a directive to say "write all remaining CHECK lines here" so that we can make the EOF get left alone in cases where it matters?

(We could force tests that refer to EOF to end in a blank line, but our pre-commit checks remove trailing blank lines so that'd require some additional changes elsewhere.)

zygoloid: I've pushed changes to fix the missing line replacements for references to the end of file, and to put the unattached STDOUT checks immediately after the AUTOUPDATE if there are no attached STDOUT checks at all. With that, I'm happy with how the lexer tests are looking.

toolchain/parser/testdata/function/declaration/fail_missing_deduced_close.carbon
zygoloid: Yes, I agree, we should definitely not be suggesting that the missing token ought to be placed in the middle of a comment :)

testing/file_test/README.md
zygoloid: Done and updated to match the new rules.

geoffromer (approved):

PR #3077, opened 2023-08-08
Only allow assignment to durable reference expressions.

zygoloid:
Now that we support expression categories, use them to determine whether the left-hand side of an assignment expression is valid.


toolchain/semantics/testdata/operators/fail_assigment_to_non_assignable.carbon
```diff
@@ -0,0 +1,164 @@
...
+  var n: i32 = 0;
+  // CHECK:STDERR: fail_assigment_to_non_assignable.carbon:[[@LINE+3]]:8: Expression is not assignable.
+  // CHECK:STDERR:   (n, n) = (1, 2);
+  // CHECK:STDERR:        ^
+  (n, n) = (1, 2);
+  // CHECK:STDERR: fail_assigment_to_non_assignable.carbon:[[@LINE+3]]:3: Expression is not assignable.
+  // CHECK:STDERR:   i32 = i32*;
+  // CHECK:STDERR:   ^
+  i32 = i32*;
+}
```
chandlerc: Potential other test cases:
```suggestion
  i32 = i32*;
  {.x = 1, .y = 2} = {.x = 3, .y = 4};
  (if true then 1 else 2) = 3;
}
```

toolchain/semantics/testdata/operators/assignment.carbon
```diff
@@ -7,6 +7,17 @@
...
+  b[0] = 3;
+  b[1] = 4;
+
+  var c: {.a: i32, .b: i32} = {.a = 1, .b = 2};
+  c.a = 3;
+  c.b = 4;
+
+  var p: i32* = &a;
+  *p = 5;
 }
```
chandlerc: Possible further test case:

```suggestion
  *p = 5;
  
  (if true then a else *p) = 10;
}
```

chandlerc (approved):
Looks good!

I've suggested some possible further test cases, but happy for you to merge and handle these in a follow-up, especially if they uncover much of anything.

toolchain/semantics/testdata/operators/assignment.carbon
zygoloid: See https://github.com/carbon-language/carbon-lang/blob/trunk/proposals/p0911.md#support-lvalue-conditionals: we decided so far that if-expressions are never reference expressions, so this is invalid. Added as a failing test and added the "workaround" of `*(if ...) = 10;` here.

PR #3078, opened 2023-08-09
Make LLVM IR emission a bit more concise and readable.

zygoloid:
Provide LLVM IR names for named types, and not for unnamed types.

Use shorter names for IR values.

### DIFF
--- a/toolchain/lowering/lowering_context.h
+++ b/toolchain/lowering/lowering_context.h
@@ -35,8 +35,7 @@ class LoweringContext {
   auto GetType(SemanticsTypeId type_id) -> llvm::Type* {
     // InvalidType should not be passed in.
     if (type_id == SemanticsTypeId::TypeType) {
-      // `type` is lowered to an empty LLVM StructType.
-      return llvm::StructType::get(llvm_context());
+      return GetTypeType();
     }
     CARBON_CHECK(type_id.index >= 0) << type_id;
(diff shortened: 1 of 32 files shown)


toolchain/lowering/lowering_context.h
```diff
@@ -65,6 +64,15 @@ class LoweringContext {
   // caller.
   auto BuildType(SemanticsNodeId node_id) -> llvm::Type*;
 
+  // Returns the empty LLVM struct type used to represent the type `type`.
+  auto GetTypeType() -> llvm::StructType* {
+    if (!type_type_) {
+      // `type` is lowered to an empty LLVM StructType.
+      type_type_ = llvm::StructType::create(*llvm_context_, {}, "type");
```
chandlerc, line 71: Naming this has the maybe-good and maybe-bad effect of causing each IR module to get a *different* type for `type`, and LLVM to not unique them... Really no idea whether this is good or bad. Anyways, probably not worth worrying about, and I can see why you prefer it from a readability perspective.

I think I actually argued for the prior behavior of using `{}` in LLVM IR for this, but the test case changes below seem pretty compelling so I think if so, I was wrong. =]

chandlerc (approved):
I do think this produces a bit more readable output. I feel a bit bad as I think I may have led to some of the old behavior, but good suggestion here. Merging.

PR #3081, opened 2023-08-09
Update precedence rules to match design.

zygoloid:
- Only allow assignment at the top level in an expression statement.
- Allow both negation and complement as subexpressions of both bitwise
  and numeric operators.
- Remove parsing support for postincrement and postdecrement.
- Add parsing support for `as` operator.
- Use the same ambient precedence for types and non-type expressions.


toolchain/parser/testdata/operators/precedence_assign.carbon
```diff
@@ -0,0 +1,77 @@
...
+//
+// AUTOUPDATE
+
+fn F(c: bool) {
+  var a: i32;
+  var b: i32;
+  var p: i32*;
+  *p = if c then 1 else 2;
+  (if c then a else b) += if c then 1 else 2;
+  ++if c then a else b;
```
chandlerc, line 14: Is it worth mentioning the oddity here?

```suggestion
  // These are valid to _parse_ even though rejected semantically.
  (if c then a else b) += if c then 1 else 2;
  ++if c then a else b;
```

chandlerc (approved):
LG

PR #3083, opened 2023-08-10
Add precedence rules for assignment operators to the precedence diagram.

### LINKED ISSUE #3079: ambient precedence of operand of `--` and `++`
We say that `++` and `--` can only be used as a statement, but what kinds of expression can appear within them? For example, is:

```
++x + y;
```

... parsed as `++(x + y);`, or is it a syntactic error?

There aren't many expression forms that can be reference expressions in Carbon's syntax -- only function calls, `a.b`, `a->b`, `*p`, `a[i]`, and `if` expressions can be reference expressions at the moment. All other operators always result in value expressions or initializing expressions. So we'll still always diagnose things like `++x + y;` due to an expression category error, but the diagnostic may not be clear about the nature of the problem.

On the other hand, if we impose a precendence restriction (eg, we could say the operand must have the precedence of `*p` or higher) then we would disallow

```
++if cond then x else y;
```

... which might be a desirable operation.

zygoloid:
Also indicate what can appear within parentheses.

This is intended to be a clarification, not a design change. Note that while we previously described the operand of `++` or `--` as being simply an expression, the operand can never be anything other than the kinds of expression the diagram now shows due to the expression category rules added in #2006.

Fixes #3079.

### DIFF
--- a/docs/design/expressions/README.md
+++ b/docs/design/expressions/README.md
@@ -85,6 +85,9 @@ graph BT
     complement["^x"]
     click complement "https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/expressions/bitwise.md"
 
+    incDec["++x;<br>--x;"]
+    click incDec "https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/assignment.md"
+
     unary((" "))
 
     as["x as T"]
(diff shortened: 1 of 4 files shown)


chandlerc (approved):
Nice!

PR #3084, opened 2023-08-10
Always create a StructType node for every syntactic struct type.

zygoloid:
Don't elide the StructType node if we've already created an equivalent type. Its spelling and location may be interesting to diagnostics, tooling, debug information, etc.


chandlerc (approved):
LG

PR #3086, opened 2023-08-10
Produce label names based on the reason for the branch.

zygoloid:
When producing textual semantics IR, map each branch instruction back to the construct that produced it and use that to determine a name for the corresponding block label.


toolchain/semantics/semantics_ir_formatter.cpp
```diff
@@ -246,21 +245,78 @@ class NodeNamer {
...
+      case ParseNodeKind::IfExpressionIf:
+        switch (node.kind()) {
+          case SemanticsNodeKind::BranchIf:
+            name = "if.expr.then";
+            break;
+          case SemanticsNodeKind::Branch:
+            name = "if.expr.else";
+            break;
+          case SemanticsNodeKind::BranchWithArg:
+            name = "if.expr.cont";
```
chandlerc: I know it's not 4 letters, but maybe spell this one out? Seems like we don't have too much pressure on length for these names.

```suggestion
            name = "if.expr.continue";
```

chandlerc: Actually, maybe we want a different name here so it doesn't resemble the naming we use with `continue;`?

```suggestion
            name = "if.expr.resume";
```

Or `after` or `merge`?

chandlerc: Nevermind, these *all* seem like they might collide, so we'll rely on the `if.expr.` context anyways... `continue` is probably better to stick with. Sorry for the noise.

toolchain/semantics/semantics_ir_formatter.cpp
```diff
@@ -246,21 +245,78 @@ class NodeNamer {
...
+          case SemanticsNodeKind::Branch:
+            name = "if.else";
+            break;
+          default:
+            break;
+        }
+        break;
+
+      case ParseNodeKind::IfStatement:
+        name = "if.cont";
```
chandlerc: Is the parse node the *predecessor* of the block? That wasn't obvious to me -- but not sure where to clarify it?

chandlerc (approved):
Wow, this made a bigger improvement than I even imagined. Wow.

Minor discussion of a wording tweak inline, but LG with whichever name we land on.

toolchain/semantics/semantics_ir_formatter.cpp
zygoloid: The parse tree node is the node associated with the branch instruction, so in some sense yes. I've updated the function comment to say what the given semantics node means.

toolchain/semantics/semantics_ir_formatter.cpp
zygoloid: Switched to `.done` for statement forms  (currently only `if`), and `.result` for expression forms.

chandlerc (approved):
(still LG)

PR #3087, opened 2023-08-10
Semantics for array type

fasiddique:


toolchain/semantics/semantics_context.cpp
```diff
@@ -379,6 +379,44 @@ auto SemanticsContext::ImplicitAsImpl(SemanticsNodeId value_id,
     // Type doesn't need to change.
     return ImplicitAsKind::Identical;
   }
+
+  auto array_type = semantics_ir_->GetTypeAllowBuiltinTypes(as_type_id);
+  auto array_node = semantics_ir_->GetNode(array_type);
+  if (array_node.kind() == SemanticsNodeKind::ArrayType) {
```
zygoloid:
```suggestion
  auto as_type = semantics_ir_->GetTypeAllowBuiltinTypes(as_type_id);
  auto as_type_node = semantics_ir_->GetNode(as_type);
  if (as_type_node.kind() == SemanticsNodeKind::ArrayType) {
```
I think these names would be clearer, because we don't yet know that `as_type_id` refers to an array type here. (We might also want to use these variables for other cases later as this function grows.)

toolchain/semantics/semantics_context.cpp
```diff
@@ -379,6 +379,44 @@ auto SemanticsContext::ImplicitAsImpl(SemanticsNodeId value_id,
...
+          *output_value_id =
+              semantics_ir_->GetTypeAllowBuiltinTypes(array_value_type);
+        }
+        return ImplicitAsKind::Compatible;
+      }
+    }
+    if (output_value_id != nullptr) {
+      *output_value_id = SemanticsNodeId::BuiltinError;
+    }
+    return ImplicitAsKind::Incompatible;
```
zygoloid:
```suggestion
```

We can just fall through to the code at the  bottom of this function to handle this case.

toolchain/semantics/semantics_context.cpp
```diff
@@ -379,6 +379,44 @@ auto SemanticsContext::ImplicitAsImpl(SemanticsNodeId value_id,
...
+  auto array_node = semantics_ir_->GetNode(array_type);
+  if (array_node.kind() == SemanticsNodeKind::ArrayType) {
+    auto [bound_node_id, element_type_id] = array_node.GetAsArrayType();
+    if (semantics_ir_
+            ->GetNode(semantics_ir_->GetTypeAllowBuiltinTypes(value_type_id))
+            .kind() == SemanticsNodeKind::TupleType) {
+      auto tuple_type_block_id =
+          semantics_ir_
+              ->GetNode(semantics_ir_->GetTypeAllowBuiltinTypes(value_type_id))
+              .GetAsTupleType();
```
zygoloid: There's some duplication here. Please factor out a variable to store the result of the `GetNode` call.

toolchain/semantics/semantics_context.cpp
```diff
@@ -379,6 +379,44 @@ auto SemanticsContext::ImplicitAsImpl(SemanticsNodeId value_id,
...
+      auto type_block = semantics_ir_->GetTypeBlock(tuple_type_block_id);
```
zygoloid:
```suggestion
      const auto& type_block = semantics_ir_->GetTypeBlock(tuple_type_block_id);
```
Avoid making a copy of the type block.

toolchain/semantics/semantics_context.cpp
```diff
@@ -379,6 +379,44 @@ auto SemanticsContext::ImplicitAsImpl(SemanticsNodeId value_id,
...
+        }
+        auto array_value_type = CanonicalizeType(array_type);
+        AddNode(SemanticsNode::ArrayValue::Make(value.parse_node(),
+                                                array_value_type, value_id));
+        if (output_value_id != nullptr) {
+          *output_value_id =
+              semantics_ir_->GetTypeAllowBuiltinTypes(array_value_type);
+        }
+        return ImplicitAsKind::Compatible;
+      }
```
zygoloid: This can be made a bit simpler with a standard library algorithm:

```suggestion
      if (type_block.size() == semantics_ir_->GetArrayBound(bound_node_id) &&
          std::all_of(type_block.begin(), type_block.end(),
                      [&] (auto type) { return type == element_type_id; })) {
        if (output_value_id != nullptr) {
          *output_value_id =
              AddNode(SemanticsNode::ArrayValue::Make(value.parse_node(),
                                                      as_type_id, value_id));
        }
        return ImplicitAsKind::Compatible;
      }
```

A couple of other things here:

- We don't need to `CanonicalizeType(array_type)`, because we already have the `SemanticsTypeId` of the array type -- that's `as_type_id`.
- We should be filling in the `*output_value_id` with the array value node itself, not the type of that node. And it looks like we're not supposed to actually build the node unless an `output_value_id` is provided.

toolchain/semantics/semantics_handle_array.cpp
```diff
@@ -3,22 +3,44 @@
...
+auto SemanticsHandleArrayExpressionSemi(SemanticsContext& /*context*/,
+                                        ParseTree::Node /*parse_node*/)
+    -> bool {
+  return true;
 }
 
 auto SemanticsHandleArrayExpression(SemanticsContext& context,
                                     ParseTree::Node parse_node) -> bool {
-  return context.TODO(parse_node, "HandleArrayExpression");
+  // TODO: Handle array type with undefined bound.
```
zygoloid, line 26: Please add a `context.TODO` call for that case, and add a testcase.

To tell whether you're in that case, you could push something onto the node stack from either `HandleArrayExpressionStart` or `HandleArrayExpressionSemi` and look for it here. Or you could use different parse node kinds for an array expression with a bound and an array expression without a bound.

toolchain/semantics/semantics_handle_array.cpp
```diff
@@ -3,22 +3,44 @@
...
+  auto element_type_node_id = context.node_stack().PopExpression();
+  auto bound_node = context.semantics_ir().GetNode(bound_node_id);
+  if (bound_node.kind() == SemanticsNodeKind::IntegerLiteral) {
+    auto bound_value = context.semantics_ir().GetIntegerLiteral(
+        bound_node.GetAsIntegerLiteral());
+    if (!bound_value.isNegative()) {
+      context.AddNodeAndPush(
+          parse_node, SemanticsNode::ArrayType::Make(
+                          parse_node, SemanticsTypeId::TypeType, bound_node_id,
+                          context.CanonicalizeType(element_type_node_id)));
```
zygoloid: You should use `ExpressionAsType` here instead of `CanonicalizeType`, for two reasons:

1) To reject things that aren't types. like [1; 2].
2) To allow things like `[(); 5]` where the element type expression needs to be converted to be usable as a type.

Please add tests for those cases too.

toolchain/semantics/semantics_ir.cpp
```diff
@@ -286,6 +288,12 @@ auto SemanticsIR::StringifyType(SemanticsTypeId type_id,
 
     auto node = GetNode(step.node_id);
     switch (node.kind()) {
+      case SemanticsNodeKind::ArrayType: {
+        auto [bound_id, type_id] = node.GetAsArrayType();
+        out << "array [" << StringifyType(type_id) << "; "
+            << GetArrayBound(bound_id) << "]";
+        break;
+      }
```
zygoloid: We want this function to be non-recursive, so that very complex types can't overflow our stack. Look at the implementation for `ConstType` to see how to print an inner type (the element type in this case) with some extra text wrapped around it (the `[` and `; bound]`).

toolchain/semantics/semantics_ir.cpp
```diff
@@ -286,6 +288,12 @@ auto SemanticsIR::StringifyType(SemanticsTypeId type_id,
 
     auto node = GetNode(step.node_id);
     switch (node.kind()) {
+      case SemanticsNodeKind::ArrayType: {
+        auto [bound_id, type_id] = node.GetAsArrayType();
+        out << "array [" << StringifyType(type_id) << "; "
```
zygoloid: This function is trying to print the type as Carbon syntax, so we shouldn't have the `array ` prefix here.

toolchain/semantics/semantics_ir.cpp
```diff
@@ -404,6 +413,7 @@ auto SemanticsIR::StringifyType(SemanticsTypeId type_id,
   if (!in_type_context) {
     auto outer_node = GetNode(outer_node_id);
     if (outer_node.kind() == SemanticsNodeKind::TupleType ||
+        outer_node.kind() == SemanticsNodeKind::ArrayType ||
```
zygoloid: I think we don't need this change: `[T; N]` already gives us a value of type `type`, so we don't need an `as type` to convert it.

toolchain/semantics/semantics_ir.h
```diff
@@ -82,6 +82,12 @@ class SemanticsIR {
   }
   auto Print(llvm::raw_ostream& out, bool include_builtins) const -> void;
 
+  // Returns array bound value from the bound node.
+  auto GetArrayBound(SemanticsNodeId bound_id) const -> uint64_t {
```
zygoloid: Maybe rename this to `GetArrayBoundValue`? Given the name of this function, I would expect it to take a description of the array type and return the array bound, not take a description of the array bound and return its value.

toolchain/semantics/semantics_handle_array.cpp
```diff
@@ -2,23 +2,55 @@
...
+  // TODO: Handle array type with undefined bound.
+  if (context.parse_tree().node_kind(context.node_stack().PeekParseNode()) ==
+      ParseNodeKind::ArrayExpressionSemi) {
+    context.node_stack().PopAndIgnore();
+    context.node_stack().PopAndIgnore();
+    return context.TODO(parse_node, "HandleArrayExpressionWithoutBounds");
+  }
+
+  auto bound_node_id = context.node_stack().PopExpression();
+  context.node_stack().PopAndIgnore();
```
zygoloid:
```suggestion
  context.node_stack().PopAndDiscardSoloParseNode<ParseNodeKind::ArrayExpressionSemi>();
```
We know what parse node kind we expect here, so it's better to explicitly name it. (This way we'll get an assertion failure if the expected node isn't at the top of the stack.)

zygoloid (approved):
Just one minor improvement then this LGTM. Thanks!

PR #3089, opened 2023-08-11
Model function calls as initializing expressions

zygoloid:
Start treating function calls as initializing expressions instead of as value expressions.

This required adding support for expression categories. Value bindings and temporary materialization conversions are created where necessary to transition between expression categories. For a function call with a return slot, we speculatively create a materialized temporary before the call and either commit to it or replace it with something else later, once we see how the function call expression is actually used.

This change follows the direction suggested in #3133 for initializing expressions: depending on the return type of a function, the return value will either be initialized in-place or returned directly. This is visible in the semantics IR, which is a little unfortunate but is probably necessary as this is part of the semantics of the program.


toolchain/lowering/testdata/function/call/var_param.carbon
```diff
@@ -22,7 +22,8 @@ fn Main() {
 // CHECK:STDOUT:   %var = alloca i32, align 4
 // CHECK:STDOUT:   store i32 0, ptr %var, align 4
 // CHECK:STDOUT:   %1 = load i32, ptr %var, align 4
-// CHECK:STDOUT:   call void @DoNothing(i32 %1)
+// CHECK:STDOUT:   %2 = load i32, ptr %var, align 4
```
chandlerc: Any idea why this results in a redundant / dead load here?

(not blocking regardless, happy to fix forward)

toolchain/semantics/semantics_node_kind.def
```diff
@@ -84,6 +88,8 @@ CARBON_SEMANTICS_NODE_KIND_IMPL(TupleIndex, "tuple_index", Typed, NotTerminator)
 CARBON_SEMANTICS_NODE_KIND_IMPL(TupleType, "tuple_type", Typed, NotTerminator)
 CARBON_SEMANTICS_NODE_KIND_IMPL(TupleValue, "tuple_value", Typed, NotTerminator)
 CARBON_SEMANTICS_NODE_KIND_IMPL(UnaryOperatorNot, "not", Typed, NotTerminator)
+CARBON_SEMANTICS_NODE_KIND_IMPL(ValueBinding, "value_binding", Typed,
+                                NotTerminator)
```
chandlerc: Thoughts on a more verb-like naming of `bind_value`? Or `bind_as_value`?

chandlerc (dismissed):
I'm not certain about the exact structure we have of semantic nodes with this, but I think it's fine to start here.

My guess is that we may want to look at what is needed to lower this really nicely into LLVM IR and tweak the exact node structure between things like `assign`, the `var` or temporary allocation, and the `initialize_from` nodes, but it's hard to say what would be "right" just yet, and seems better to get the infrastructure in place and then look at tweaks.

Overall LG, some minor questions inline that aren't really blocking either way.

toolchain/semantics/semantics_node_kind.def
zygoloid: Done.

toolchain/lowering/testdata/function/call/var_param.carbon
zygoloid: We weren't properly updating the call argument list when argument conversions were required. Fixed.

zygoloid (commented):
@chandlerc Sorry for the large influx of new changes since your last review; I wanted to explore this a bit more and see how the direction pans out before committing to it. In the end the changes here aren't exactly an incremental extension of the previous state of the PR, so I think it probably makes more sense to treat this as a from-scratch review rather than an incremental diff compared to the prior PR.

toolchain/semantics/semantics_context.cpp
```diff
@@ -263,6 +263,186 @@ auto Context::is_current_position_reachable() -> bool {
...
+                     << semantics_ir().GetNode(expr_id)
+                     << " to value expression";
+
+    case SemIR::ExpressionCategory::Initializing: {
+      // Commit to using a temporary for this initializing expression.
+      // TODO: Don't create a temporary if the initializing representation is
+      // already a value representation.
+      expr_id = FinalizeTemporary(expr_id, /*discarded=*/false);
+      [[fallthrough]];
+    }
```
chandlerc: I think braces aren't needed here?

```suggestion
    case SemIR::ExpressionCategory::Initializing:
      // Commit to using a temporary for this initializing expression.
      // TODO: Don't create a temporary if the initializing representation is
      // already a value representation.
      expr_id = FinalizeTemporary(expr_id, /*discarded=*/false);
      [[fallthrough]];
```

toolchain/semantics/semantics_context.cpp
```diff
@@ -263,6 +263,186 @@ auto Context::is_current_position_reachable() -> bool {
...
+        CARBON_FATAL() << "Initialization from unexpected node " << init;
+
+      case SemIR::NodeKind::StructValue:
+      case SemIR::NodeKind::TupleValue:
+        CARBON_FATAL() << init << " is not modeled as initializing yet";
+
+      case SemIR::NodeKind::StubReference: {
+        init_id = init.GetAsStubReference();
+        continue;
+      }
```
chandlerc:
```suggestion
      case SemIR::NodeKind::StubReference:
        init_id = init.GetAsStubReference();
        continue;
```

toolchain/semantics/testdata/array/assign_return_value.carbon
```diff
@@ -114,7 +123,10 @@ fn Run() {
 // CHECK:STDOUT:   %.loc10_17: type = array_type %.loc10_16, i32
 // CHECK:STDOUT:   %t: [i32; 1] = var "t"
 // CHECK:STDOUT:   %.loc10_22.1: (i32,) = call @F()
-// CHECK:STDOUT:   %.loc10_22.2: [i32; 1] = array_value %.loc10_22.1
-// CHECK:STDOUT:   assign %t, %.loc10_22.2
+// CHECK:STDOUT:   %.loc10_22.2: (i32,) = materialize_temporary
```
chandlerc, line 126: The materialize temporary coming *after* the call is a bit surprising.

To check my understanding, this is materializing a temporary in the *caller* not the return slot, and initializing it by copy?

chandlerc (approved):
I'm pretty tired and so worried I'm missing something, but I think this largely makes sense.

I have some boring style nits in suggested edits, and a question (but not really a concern).

I'm not super delighted by the extensive temporaries, but I understand why this is a reasonably, faithful lowering. And it makes sense to get this and the other semantic things sorted out first, and then revisit things to try to tighten up the emitted LLVM IR.

So I think this LG.

toolchain/semantics/testdata/array/assign_return_value.carbon
zygoloid: Yes, this happens because we're modeling the `(i32,)` return type as being initialized by copy instead of being initialized in place.

PR #3090, opened 2023-08-11
Explorer: Replace StringRef with string_view in trace stream

prabhatexit0:


zygoloid (approved):
Thanks!

chandlerc:
I was a bit surprised by this change -- @zygoloid, should we update https://github.com/carbon-language/carbon-lang/blob/trunk/docs/project/cpp_style_guide.md#foundational-libraries-and-data-types to prefer `std::string_view` at this point?

My hope is to avoid Carbon having both `std::string_view` and `llvm::StringRef` in heavy ues, not a preference specifically for one or the other.

zygoloid:
This is an interesting case where the convention in explorer is overwhelmingly one way (`string_view` rather than `StringRef`) and the convention in the toolchain is overwhelmingly the other way (`StringRef` rather than `string_view`). I think this makes a certain amount of sense -- the toolchain is closely coupled to LLVM, so reducing the impedance mismatch there makes sense, but explorer has no fundamental reason to depend on LLVM at all. But yes, we should at least be documenting this convention.

chandlerc:
Interesting, I didn't even remember the explorer having this, but I can see the reason and that it doesn't make sense to change that at this point certainly.

I do worry that eventually we'll want enough common code sharing between them that we'll be motivated to not have the divergence, but I guess we can cross that bridge later.

zygoloid:
#3092 has a documentation update to reflect the status quo.

PR #3092, opened 2023-08-11
Document the de facto conventions for library dependencies in explorer

zygoloid:
Unlike in the toolchain, we have not been preferring LLVM facilities over standard ones. Update the documentation to describe this and provide some rationale.


chandlerc (approved):
LG

PR #3095, opened 2023-08-14
Lowering for array

fasiddique:


toolchain/lowering/lowering_context.cpp
```diff
@@ -144,6 +144,12 @@ auto LoweringContext::BuildType(SemanticsNodeId node_id) -> llvm::Type* {
 
   auto node = semantics_ir_->GetNode(node_id);
   switch (node.kind()) {
+    case SemanticsNodeKind::ArrayType: {
+      auto [bound_node_id, type_id] = node.GetAsArrayType();
+      auto bound_value = semantics_ir_->GetArrayBoundValue(bound_node_id);
+      llvm::SmallVector<llvm::Type*> subtypes(bound_value, GetType(type_id));
+      return llvm::StructType::get(*llvm_context_, subtypes);
```
jonmeow: LLVM provides an ArrayType which should be suitable for use here.
https://llvm.org/doxygen/classllvm_1_1ArrayType.html

toolchain/lowering/testdata/array/assign_return_value.carbon
```diff
@@ -0,0 +1,31 @@
...
+// CHECK:STDOUT:   store i32 0, ptr %1, align 4
+// CHECK:STDOUT:   %2 = load { i32 }, ptr %tuple, align 4
+// CHECK:STDOUT:   ret { i32 } %2
+// CHECK:STDOUT: }
+// CHECK:STDOUT:
+// CHECK:STDOUT: define void @Run() {
+// CHECK:STDOUT:   %var = alloca { i32 }, align 8
+// CHECK:STDOUT:   %F = call { i32 } @F()
+// CHECK:STDOUT:   %array = alloca { i32 }, align 8
+// CHECK:STDOUT:   %1 = load { i32 }, ptr %array, align 4
```
jonmeow: Reading this code, it looks like what happens is:

1. %F has the result of the function call.
2. A new %array is created, and never initialized.
3. %array is loaded, and stored into %var (var t).

What are you expecting to happen here?

jonmeow (commented):
I spoke with zygoloid and I'll be picking this up. Sorry I'm still catching up with the array changes, so please let me know if I'm misunderstanding the ArrayValue behavior.

toolchain/lowering/lowering_context.cpp
fasiddique: As we are using the tuple value as the array value, I used StructType to match with the tuple type. Should I change it to ArrayType?

jonmeow: Yes, that's correct. ArrayType.

jonmeow (commented):
In assign_return_value.carbon, it looks like you now:

1. Copy %F to %as_tuple, where %F and %as_tuple have the same type.
2. Copy %as_tuple to the target value.

What's your intent in copying %F?

Please do also add comments to your HandleArrayValue function explaining the intent of significant code blocks.

fasiddique:
> In assign_return_value.carbon, it looks like you now:
> 
> 1. Copy %F to %as_tuple, where %F and %as_tuple have the same type.
> 2. Copy %as_tuple to the target value.
> 
> What's your intent in copying %F?
> 
> Please do also add comments to your HandleArrayValue function explaining the intent of significant code blocks.

Updated the comments. Please let me know if that makes sense!

toolchain/lowering/lowering_handle.cpp
```diff
@@ -24,16 +24,39 @@ auto LoweringHandleAddressOf(LoweringFunctionContext& context,
...
+  // it cannot be accessed directly to create the array value. Hence, we are
+  // creating a store in this block.
+  if (!tuple_value->getType()->isPointerTy()) {
+    tuple_value = context.builder().CreateAlloca(
+        tuple_type, /*ArraySize=*/nullptr, "as_tuple");
+    context.builder().CreateStore(context.GetLocalLoaded(tuple_node_id),
+                                  tuple_value);
+  }
+
+  for (uint64_t i = 0; i < array_size; i++) {
```
```suggestion
  for (int i = 0; i < llvm_type->getArrayNumElements(); ++i) {
```

There are a few stylistic things here:

- We tend to prefer int: https://google.github.io/styleguide/cppguide.html#Integer_Types
- We prefer variables to be close to their usage, but array_size is only used here: https://google.github.io/styleguide/cppguide.html#Local_Variables
- We prefer prefix increment: https://google.github.io/styleguide/cppguide.html#Preincrement_and_Predecrement

toolchain/lowering/lowering_handle.cpp
```diff
@@ -24,16 +24,39 @@ auto LoweringHandleAddressOf(LoweringFunctionContext& context,
...
+  auto* tuple_type =
+      context.GetType(context.semantics_ir().GetNode(tuple_node_id).type_id());
 
-auto LoweringHandleArrayValue(LoweringFunctionContext& /*context*/,
-                              SemanticsNodeId /*node_id*/, SemanticsNode node)
-    -> void {
-  CARBON_FATAL() << "TODO: Add support: " << node;
+  // For return values or call instructions, the tuple is not in the memory so
+  // it cannot be accessed directly to create the array value. Hence, we are
+  // creating a store in this block.
```
jonmeow: Thanks for explaining this, but I'm still uncertain about this issue. If the tuple weren't in memory, then I would've thought to see a load. Do you have a reference for how you reached this conclusion and solution?

I was looking at the similar C++ code, and I see `extractvalue` gets used: https://cpp.compiler-explorer.com/z/Wq31ajrv4 -- maybe we should investigate that? A TODO would be fine:

```
// TODO: Investigate if `extractvalue` can be used to avoid copying non-pointer data.
```

fasiddique: I discussed this with @zygoloid before implementing this.

fasiddique: Thank you for the suggestion! Updated using `extractvalue`.

toolchain/lowering/lowering_handle.cpp
```diff
@@ -31,31 +31,29 @@ auto LoweringHandleArrayValue(LoweringFunctionContext& context,
...
-    auto* load_inst = context.builder().CreateLoad(
-        llvm_type->getArrayElementType(), tuple_gep);
-    // Initializing the array with values.
-    auto* array_gep = context.builder().CreateStructGEP(llvm_type, alloca, i);
-    context.builder().CreateStore(load_inst, array_gep);
+  for (int i = 0; i < static_cast<int>(llvm_type->getArrayNumElements()); ++i) {
+    if (!tuple_value->getType()->isPointerTy()) {
+      // Non pointer types such as call or return, using extract value as gep
+      // cannot be used.
+      auto* gep = context.builder().CreateExtractValue(tuple_value, i);
```
jonmeow: Why "gep" for this variable name?

toolchain/lowering/lowering_handle.cpp
```diff
@@ -31,31 +31,29 @@ auto LoweringHandleArrayValue(LoweringFunctionContext& context,
...
+      // Extracting values from tuple and loading them to the memory.
+      auto* gep = context.builder().CreateStructGEP(
+          context.GetType(
+              context.semantics_ir().GetNode(tuple_node_id).type_id()),
+          tuple_value, i);
+      auto* load_gep =
+          context.builder().CreateLoad(llvm_type->getArrayElementType(), gep);
+      // Initializing the array with values.
+      context.builder().CreateStore(
+          load_gep, context.builder().CreateStructGEP(llvm_type, alloca, i));
```
jonmeow: These calls mirror lines 42-44, can you share the logic?

toolchain/lowering/lowering_handle.cpp
```diff
@@ -31,31 +31,29 @@ auto LoweringHandleArrayValue(LoweringFunctionContext& context,
...
-    // Extracting values from tuple and loading them to the memory.
-    auto* tuple_gep =
-        context.builder().CreateStructGEP(tuple_type, tuple_value, i);
-    auto* load_inst = context.builder().CreateLoad(
-        llvm_type->getArrayElementType(), tuple_gep);
-    // Initializing the array with values.
-    auto* array_gep = context.builder().CreateStructGEP(llvm_type, alloca, i);
-    context.builder().CreateStore(load_inst, array_gep);
+  for (int i = 0; i < static_cast<int>(llvm_type->getArrayNumElements()); ++i) {
+    if (!tuple_value->getType()->isPointerTy()) {
```
jonmeow: Given that the if/else needs to handle both true and false, why do you add `!` here? It reorders the branches, is there something you're gaining from that? I believe there's an advantage to having simpler conditions.

jonmeow (commented):
Thanks, I'm glad that worked! I think it's worth thinking about the style a little more, but then it should be ready to merge!

jonmeow (approved):
Looks good, thanks!

PR #3097, opened 2023-08-14
Small file_test refactorings

jonmeow:
Just some small refactorings stemming from #3073.

AddCheckLines -> BuildCheckLines because the two lists are now fully separate. Adding is_blank to be more direct about behavior than the Print call.


chandlerc (approved):
Nice!

PR #3099, opened 2023-08-14
Semantics for array index.

fasiddique:


toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -50,10 +68,9 @@ auto SemanticsHandleIndexExpression(SemanticsContext& context,
       return true;
     }
   } else if (index_node.kind() != SemanticsNodeKind::IntegerLiteral) {
-    CARBON_DIAGNOSTIC(NondeterministicType, Error,
-                      "Type cannot be determined at compile time.");
-    context.emitter().Emit(parse_node, NondeterministicType);
-  } else if (name_type_node.kind() != SemanticsNodeKind::TupleType &&
+    context.DiagnoseUndeterministicType(parse_node);
+  } else if ((name_type_node.kind() != SemanticsNodeKind::TupleType ||
+              name_type_node.kind() != SemanticsNodeKind::ArrayType) &&
```
jonmeow: This looks like it'll always evaluate to true: the kind will also be not equal to one of these. Was the check necessary before in a way that this should be fixed, or can it be removed?

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -27,21 +26,40 @@ auto SemanticsHandleIndexExpression(SemanticsContext& context,
...
 
-  if (name_type_node.kind() == SemanticsNodeKind::TupleType &&
-      index_node.kind() == SemanticsNodeKind::IntegerLiteral) {
+  if (name_type_node.kind() == SemanticsNodeKind::ArrayType) {
+    auto [bound_id, type_id] = name_type_node.GetAsArrayType();
+    if (index_node.kind() == SemanticsNodeKind::IntegerLiteral) {
+      const auto& index_val = context.semantics_ir().GetIntegerLiteral(
+          index_node.GetAsIntegerLiteral());
+      if (index_val.uge(context.semantics_ir().GetArrayBoundValue(bound_id))) {
+        context.DiagnoseOutOfBounds(parse_node, index_val, name_node);
```
jonmeow: These are only used from this function, so do not need to be moved onto SemanticsContext. It's better to keep them local to the function where they're used so that the error message can be easily viewed.

That is, you can declare the CARBON_DIAGNOSTIC once at the top of the function, then emit from two separate locations.

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -27,21 +26,40 @@ auto SemanticsHandleIndexExpression(SemanticsContext& context,
...
+                       .GetTypeAllowBuiltinTypes(index_node.type_id())
+                       .index) ==
+               Internal::SemanticsBuiltinKindRawEnum::IntegerType) {
+      context.AddNodeAndPush(
+          parse_node, SemanticsNode::ArrayIndex::Make(
+                          parse_node, type_id, name_node_id, index_node_id));
+      return true;
+    } else if (name_type_id != SemanticsNodeId::BuiltinError) {
+      context.DiagnoseUndeterministicType(parse_node);
+    }
```
jonmeow: For arrays, we generally want to allow any index that can be converted to an integer. So this should probably be calling ImplicitAsRequired to try converting index_node to IntegerType, versus the current requirement that index_node is already an IntegerType.

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -27,17 +30,40 @@ auto SemanticsHandleIndexExpression(SemanticsContext& context,
...
+                   context.CanonicalizeType(
+                       SemanticsNodeId::BuiltinIntegerType)) !=
+               SemanticsNodeId::BuiltinError) {
+      context.AddNodeAndPush(
+          parse_node, SemanticsNode::ArrayIndex::Make(
+                          parse_node, type_id, name_node_id, index_node_id));
+      return true;
+    }
+  } else if (name_type_node.kind() == SemanticsNodeKind::TupleType &&
+             index_node.kind() == SemanticsNodeKind::IntegerLiteral) {
```
jonmeow: As you add ArrayType support, it looks like line 78 (checking IntegerLiteral) is actually specific to TupleType. Can you restructure this `if` statement to group TupleType-specific logic similarly to how you do ArrayType-specific logic?

jonmeow (approved):
Thanks! We'll probably want to think about how to reduce the code duplication in bounds checking, but this seems good to merge.

Issue #3100, opened 2023-08-15
Restructuring for common and base


jonmeow:
We have:

- //common
- //testing
- //testing/util
- //explorer/common
- //toolchain/common

We've been discussing the naming scheme in use (some context [in docs](https://docs.google.com/document/d/1HoTc56jidLTPISlWwELt8f25byXPlqYzj4a3zlSUL04/edit), mentioned [on discord](https://discord.com/channels/655572317891461132/707150492370862090/1121207612675989525)), and I think where we agreed is:

- //common
- //testing
- //testing/base
- //explorer/base
- //toolchain/base

The principle here is that "common" indicates things which are really cross-component (explorer and toolchain), whereas "base" is the more shared stuff within a given component.

I'm be doing moves to effect this structure.

chandlerc:
What do you (or others) think about keeping `testing` unnested, and basically unnesting any meaningfully named common component? Mostly wondering if that would let us have somewhat fewer directory components.

jonmeow:
Per [discussion on Discord](https://discord.com/channels/655572317891461132/655578254970716160/1141065937173348422) -- I'd thought you'd preferred //testing move, but it sounds like not, so I'll just do //testing/util -> //testing/base.

PR #3102, opened 2023-08-15
Rename //testing to //common/testing

jonmeow:
Also //testing/util to //common/testing/base

Renaming per #3100


jonmeow:
Note dependency on #3101

PR #3105, opened 2023-08-15
Run clang-tidy over toolchain code.

jonmeow:
Amongst other changes, this removes the implicit StringRef cast in formatter code due to google-explicit-constructor. Only one line of code relied on the implicit cast, and style says "Do not define implicit conversions." (https://google.github.io/styleguide/cppguide.html#Implicit_Conversions).

### DIFF
--- a/toolchain/lexer/token_kind.cpp
+++ b/toolchain/lexer/token_kind.cpp
@@ -107,13 +107,7 @@ auto TokenKind::fixed_spelling() const -> llvm::StringRef {
 auto TokenKind::expected_parse_tree_size() const -> int {
   static constexpr int8_t Table[] = {
 #define CARBON_TOKEN(Name) 1,
-// CARBON_TOKEN_WITH_VIRTUAL_NODE wraps CARBON_TOKEN, so needs to be written
-// as-is.
-// TODO: Consider changing the macro setup to avoid the lint issue, e.g. so that
-// CARBON_KEYWORD_TOKEN would be called by a macro that could be replaced,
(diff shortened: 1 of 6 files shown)


toolchain/lexer/token_kind.cpp
```diff
@@ -107,6 +107,12 @@ auto TokenKind::fixed_spelling() const -> llvm::StringRef {
 auto TokenKind::expected_parse_tree_size() const -> int {
   static constexpr int8_t Table[] = {
 #define CARBON_TOKEN(Name) 1,
+// CARBON_TOKEN_WITH_VIRTUAL_NODE wraps CARBON_TOKEN, so needs to be written
+// as-is.
+// TODO: Consider changing the macro setup to avoid the lint issue, e.g. so that
+// CARBON_KEYWORD_TOKEN would be called by a macro that could be replaced,
+// instead of wrapped.
+// NOLINTNEXTLINE(bugprone-macro-parentheses)
 #define CARBON_TOKEN_WITH_VIRTUAL_NODE(size) 1 + size
```
zygoloid: Yeah, I think this is the wrong kind of clever. Perhaps we could do this for now:

```suggestion
#define CARBON_TOKEN_WITH_VIRTUAL_NODE(size) 2,
```

The existing code would support multiple virtual nodes through repeated wrapping with `CARBON_TOKEN_WITH_VIRTUAL_NODE`, but I'm not sure that's a feature that's worth preserving :)

zygoloid (approved):

PR #3106, opened 2023-08-16
Introduce git-based target selection.

chandlerc:
This adds access to the `target-determinator` tool that computes the possibly impacted set of targets between the current checkout and a specific `git` commit. The tool is downloaded (and cached) with a Python script that wraps it and allows us to easily run it in our CI environment.

And finally, switches the CI for pull requests and the merge queue to use this script and run a minimal set of impacted tests rather than all of them. In order to make this as simple as possible, this also merges the previously separate build and test steps of our CI.

Note that the CI testing post-submit (push events) continues to use a blanket target pattern so that we have a good backstop in case something outside of what is tracked here changes. The important paths, especially the ones that might be in an interactive critical path, are the PR and merge queue.

This has been tested on the `action-test` branch to try to make sure it works. Some example runs for folks to inspect:

- `push` event with blanket test: https://github.com/carbon-language/carbon-lang/actions/runs/6070060958/job/16465422569
- A "large" PR that impacts a bunch of toolchain tests:
  - ubuntu-opt: https://github.com/carbon-language/carbon-lang/actions/runs/6070063050/job/16465428107?pr=3188
  - macos-opt: https://github.com/carbon-language/carbon-lang/actions/runs/6070063050/job/16465428420?pr=3188
- The merge queue for this PR:
  - ubuntu-opt: https://github.com/carbon-language/carbon-lang/actions/runs/6070202665/job/16465825387
  - macos-opt: https://github.com/carbon-language/carbon-lang/actions/runs/6070202665/job/16465825740
- A "small" PR that impacts one test:
  - ubuntu-opt: https://github.com/carbon-language/carbon-lang/actions/runs/6070659668/job/16467147641?pr=3189
  - macos-opt: https://github.com/carbon-language/carbon-lang/actions/runs/6070659668/job/16467147933?pr=3189
- The merge queue for the small PR:
  - ubuntu-opt: https://github.com/carbon-language/carbon-lang/actions/runs/6070754584/job/16467427268
  - macos-opt: https://github.com/carbon-language/carbon-lang/actions/runs/6070754584/job/16467427736

I have observed one failure while testing with this PR, but I've not yet been able to reproduce it. Every other failure (including empty lists, etc.) I've tried to address. However, we may have to keep an eye on actions after this to make sure there isn't some scenario I've not been able to work through in a test branch.


WORKSPACE
```diff
@@ -300,3 +300,14 @@ http_archive(
...
 tree_sitter_register_toolchains()
+
+###############################################################################
+# bazel-diff
+###############################################################################
+
+http_jar(
+    name = "bazel_diff",
+    urls = [
+        "https://github.com/Tinder/bazel-diff/releases/download/4.3.0/bazel-diff_deploy.jar",
```
jonmeow:
```suggestion
        "https://github.com/Tinder/bazel-diff/releases/download/4.8.0/bazel-diff_deploy.jar",
```

I assume you were copying their instructions, but 4.8 is the latest version. https://github.com/Tinder/bazel-diff/releases

scripts/bazel_diff.py
```diff
@@ -0,0 +1,170 @@
...
+
+def git_checkout(commit: str) -> None:
+    subprocess.run(
+        [
+            "git",
+            "checkout",
+            "--quiet",
+            commit,
+        ],
+        check=True,
```
jonmeow: Why `check=True` instead of `subprocess.check_call`?
https://docs.python.org/3/library/subprocess.html#subprocess.check_output

scripts/bazel_diff.py
```diff
@@ -0,0 +1,170 @@
...
+        encoding="utf-8",
```
jonmeow: What's the effect of setting the encoding for these git calls?

scripts/bazel_diff.py
```diff
@@ -0,0 +1,170 @@
...
+        "get-impacted-targets",
+        f"-sh={baseline_hashes}",
+        f"-fh={current_hashes}",
+    ]
+    p = subprocess.run(
+        args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, encoding="utf-8"
+    )
+    if p.returncode != 0:
+        print(p.stderr)
+        exit(f"Bazel diff returned {p.returncode}")
```
jonmeow: Several functions have this same subprocess structure. I assume you're deliberately avoiding subprocess.check_output because it'd print stderr on success, but maybe you could still have a helper function for suppressing stderr this way?

scripts/bazel_diff.py
```diff
@@ -0,0 +1,170 @@
...
+    return p.stdout
+
+
+def filter_targets(bazel: str, targets: str) -> str:
+    args = [
+        bazel,
+        "query",
+        (
+            f"let t = set({targets}) in "
+            f"kind(rule, $t) except attr(tags, manual, $t)"
```
jonmeow: If a test is marked manual, would this prevent building it? Can we instead build affected targets, and just not run manual tests?

jonmeow (commented):
Cool, I think this could be helpful for choosing targets!

Given this modifies tests.yaml, which can block merges, what would you think of testing the action changes before merging them? Unfortunately I don't think that the new step is a noop if there's some error.

To do that, you could push to the `action-test` branch on the main repo, modifying `branches:` to `action-test` in `tests.yaml`.

Doing that with no changes would ensure the basic commands work. It might also be interesting to make a stub change to some part of the toolchain or explorer, and see how much build time changes.

scripts/bazel_diff.py
```diff
@@ -0,0 +1,170 @@
...
+        bazel,
+        "run",
+        f"--script_path={bazel_diff_path}",
+        "//bazel/diff:bazel-diff",
+    ]
+    p = subprocess.run(
+        args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, encoding="utf-8"
+    )
+    if p.returncode != 0:
+        print(p.stderr)
```
chrisphilip322: Should this be using the `log` function you defined?

scripts/bazel_diff.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+def git_checkout(commit: str) -> None:
+    subprocess.run(
+        [
+            "git",
+            "checkout",
+            "--quiet",
+            commit,
+        ],
```
chrisphilip322: Did you consider using git worktrees for this? `git checkout` affects local file state which can be surprising (especially if there is a failure or uncommitted changes). I think at least short-circuiting the script if there are uncommitted changes would prevent a lot of unexpected file modifications.

scripts/bazel_diff.py
```diff
@@ -0,0 +1,170 @@
...
+        bazel_diff = make_bazel_diff_script(bazel, tmpdir)
+        log(f"Wrote bazel-diff script: {bazel_diff}")
+
+        git_checkout(baseline)
+        log(f"Checked out commit: {baseline}")
+
+        baseline_hashes = compute_hashes(bazel, bazel_diff, tmpdir, "baseline")
+        log(f"Baseline hashes: {baseline_hashes}")
+
+        git_checkout(current)
```
chrisphilip322: This is the last time you call `git_checkout` so if `current` isn't actually the original git checked out commit, this script will leave the working directory in a different state. Should `current` be deduced from working directory state or should there be another call to `git_checkout` that returns the working directory to the actual original state?

scripts/bazel_diff.py
```diff
@@ -0,0 +1,170 @@
...
+    current = parsed_args.current
+    log(f"Diffing Bazel targets for {baseline}..{current}")
+    git_diff(baseline, current)
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        tmpdir = tempfile.mkdtemp()
+        bazel_diff = make_bazel_diff_script(bazel, tmpdir)
+        log(f"Wrote bazel-diff script: {bazel_diff}")
+
+        git_checkout(baseline)
```
chrisphilip322: It would be useful to wrap this in a try...finally block to undo the git_checkout call if there are any errors before git_checkout gets called again.

scripts/bazel_diff.py
```diff
@@ -0,0 +1,170 @@
...
+    scripts_utils.chdir_repo_root()
+    bazel = scripts_utils.locate_bazel()
+
+    baseline = parsed_args.baseline
+    current = parsed_args.current
+    log(f"Diffing Bazel targets for {baseline}..{current}")
+    git_diff(baseline, current)
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        tmpdir = tempfile.mkdtemp()
```
chrisphilip322: This call is redundant, you are already creating a temporary directory with the above call to TemporaryDirectory

scripts/bazel_diff.py
```diff
@@ -0,0 +1,170 @@
...
+
+def make_bazel_diff_script(bazel: str, tmpdir: str) -> str:
+    bazel_diff_path = os.path.join(tmpdir, "bazel_diff")
+    args = [
+        bazel,
+        "run",
+        f"--script_path={bazel_diff_path}",
+        "//bazel/diff:bazel-diff",
+    ]
+    p = subprocess.run(
```
chrisphilip322: Did you consider you the async version of subprocess? It would let you parallelize some of the subprocess calls.

https://docs.python.org/3/library/asyncio-subprocess.html

scripts/bazel_diff.py
```diff
@@ -0,0 +1,170 @@
...
+import sys
+
+import scripts_utils
+
+
+def log(s: str) -> None:
+    print(s, file=sys.stderr)
+
+
+def make_bazel_diff_script(bazel: str, tmpdir: str) -> str:
```
chrisphilip322: Consider using the pathlib.Path API instead of str and os.path.join. `Path` has a more modern interface, distinct type from `str`, and is generally accepted interchangeably by builtin APIs.

chrisphilip322 (commented):
Some python oriented comments; excited about Carbon!

scripts/bazel_diff.py
chandlerc: Done.

scripts/bazel_diff.py
chandlerc: Not really, but I don't think it's needed. None of these are terribly slow, and there is often a sequencing requirement that will limit the parallelism.

But good to know about, will keep it in mind.

scripts/bazel_diff.py
chandlerc: Yes, thanks.

scripts/bazel_diff.py
chandlerc: Yes, this prevents building manual tests.

But I think that's the current behavior of Bazel -- manual targets aren't built or tested by `bazel build //...` AFAIK.

I also think its useful to keep this -- it lets us avoid either expensive or non-portable builds. If there is something that should get built and tested in CI, we can add a non-manual build or test dependency on it?

I guess the issue you're worried about is something like a manual `cc_test` that includes code that we won't even compile. I agree that's not a great situation, but I don't think Bazel gives us lots of good options here. The best might be to have the test not be marked as manual, but to have whatever expensive test code within be disabled without a flag. Then we can add an additional manual test that uses the same binary but explicitly adds that test argument? I don't think we have any test today that need this, but it's something we should keep in mind.

scripts/bazel_diff.py
chandlerc: I hadn't really considered it because I wanted this to run in a GitHub action where we're using somewhat normal git checkouts. I also don't use worktree, and so don't have a lot of experience.

There are also some aspects that make this nice to target a normal `git checkout` -- it makes the Bazel server behavior generally "do what we want".

For now, I'll add some defensive checks to make sure this script isn't too risky to run (clean tree, etc).

I'll also think about whether there is some other approach that doesn't have to mutate the current checkout... But that'll probably be a future thing.

scripts/bazel_diff.py
chandlerc: No idea -- I copied it from some of our other Python scripts in this directory.

I'm a touch hesitant to remove it -- [Chesterton's fence](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence) and all -- but happy to if you think I should.

scripts/bazel_diff.py
chandlerc: Basically trying to follow the documentation for subprocess where at the top it says:

> The recommended approach to invoking subprocesses is to use the run() function for all use cases it can handle.

I understand it is *primarily* talking about avoiding the low-level `Popen` API, but it also seemed like the `check_call` and `check_output` versions were somewhat less preferred to the main `run` API.

Anyways, I see in other scripts we're using these APIs so I'll try to switch to them when they apply.

scripts/bazel_diff.py
chandlerc: This was for debugging to get a directory that wouldn't be deleted. Removed. =D

scripts/bazel_diff.py
chandlerc: Good idea, done.

Also made all the subprocess stuff reliably use exceptions so that we get back here and clean up things appropriately.

scripts/bazel_diff.py
chandlerc: When I first wrote this, I couldn't figure out how to compute a reliably "current" state from `git`, and so I just made it be passed in here. As you point out, it's super clunky.

I've figured this out I think and now just offer a baseline flag. The script seems to correctly detect the current commit or symbolic ref and return to it.

scripts/bazel_diff.py
chandlerc: Yeah, a helper function makes sense. And I can do a better job of this by actually catching the check exception and printing stderr from there I think. Take a look and let me know if this seems right?

chandlerc (commented):
Thanks for all the feedback!

I've got a major refresh of this PR now, and it seems to be working as desired on the `action-test` branch. You can even play with it there. I've enabled a merge queue for that branch so PRs can go through the whole process and I've sent several PRs. I've included links to interesting ones, and you can see the last one an bug fixes that got everything there.

I did look at using the checkout github action step rather than relying on the `bazel_diff.py` script, but it proved much harder to make work than making the python script work correctly. And this way it's a bit easier to at least replicate some of the results here locally.

Also, I have a few follow-up PRs in mind that should improve the speed of Bazel in CI even further.

scripts/bazel_diff.py
jonmeow: I was trying to figure out which you might have copied it from; I see one setting encoding:

```
    p = subprocess.run(
        args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, encoding="utf-8"
    )
```

This is piping output streams, which are: `A bytes sequence, or a string if [run()](https://docs.python.org/3/library/subprocess.html#subprocess.run) was called with an encoding, errors, or text=True.`

Note that describes the behavior of several flags. However, you're not setting any stream here, so there is no stream to change the encoding of.

scripts/bazel_diff.py
```diff
@@ -0,0 +1,205 @@
...
+            [
+                str(bazel),
+                "query",
+                f"--query_file={tmp.name}",
+            ]
+        )
+
+
+def git_is_dirty() -> bool:
+    output = subprocess.check_output(
```
jonmeow: You're mixing check_output with run (line 131). You'd pointed me to documentation saying run is preferred, should you use throughout?

scripts/bazel_diff.py
jonmeow: Makes sense to me, although now I see you're using subprocess.check_* calls, which way are you trying to switch?

scripts/bazel_diff.py
jonmeow: I'd missed that manual rules don't even get built, hmm. We only do that for fuzzing today, at least.

jonmeow:
GitHub's being quirky for me, I'm having trouble adding a comment to the build, so making it here:

Had you considered skipping the test command when there are no tests (you could generate two files, build targets and test targets), or handling [error code 4](https://docs.bazel.build/versions/0.21.0/guide.html#what-exit-code-will-i-get)? e.g.:

```
          # GitHub defaults to exit-on-error; this allows handling bazel's exit codes.
          set +e
          bazelisk test -c ${{ matrix.build_mode }} \
            --target_pattern_file=targets
          exitcode=$?
          # Suppress "no tests in targets" errors.
          if [ ${exitcode} -ne 0 && ${exitcode} -ne 4 ]; then
            exit ${exitcode}
          fi
```

Note, I think of //testing/base as the location for common libraries, and it isn't clear to me from reading the description of no_op_test how it's used -- I think the association should be clearer if kept.

scripts/bazel_diff.py
chandlerc: Sorry, I was trying to say -- I'll switch to the check_* calls since you suggested them and other scripts are using them.

But to be super clear -- I'm perfectly happy either way, really whatever works best for you (or other Python folks). Just let me know how these should look.

scripts/bazel_diff.py
chandlerc: Ah, thanks. OK, I'll remove it when not capturing either `stdout` or `stderr`. In fact, I think switching ta `check_call` has already removed it!

chandlerc:
> GitHub's being quirky for me, I'm having trouble adding a comment to the build, so making it here:
> 
> Had you considered skipping the test command when there are no tests (you could generate two files, build targets and test targets), or handling [error code 4](https://docs.bazel.build/versions/0.21.0/guide.html#what-exit-code-will-i-get)? e.g.:
> 
> ```
>           # GitHub defaults to exit-on-error; this allows handling bazel's exit codes.
>           set +e
>           bazelisk test -c ${{ matrix.build_mode }} \
>             --target_pattern_file=targets
>           exitcode=$?
>           # Suppress "no tests in targets" errors.
>           if [ ${exitcode} -ne 0 && ${exitcode} -ne 4 ]; then
>             exit ${exitcode}
>           fi
> ```

I did think about it, but it didn't seem as clean. I *think* this still builds everything and just produces the error at the end, but I'm not 100% certain. And I don't think the specific error code is an especially stable part of Bazel's command line API, so worried it could go away.

And I had wanted to avoid having two files and two bazel invocations -- it would also mean doing two queries and other extra work. It seemed cleanest to just inject a no-op test into the target list so that it always passes cleanly. It also seemed like the smallest amount of logic with the fewest ways for something to go wrong in the future.

I also poked around, and I think that's relatively conventional. There is a similar no-op test target inside Google that I believe is used in similar ways.

That said...

> Note, I think of //testing/base as the location for common libraries, and it isn't clear to me from reading the description of no_op_test how it's used -- I think the association should be clearer if kept.

Yeah, this was somewhat just a guess for where to locate it. The only purpose I think is to allow forming Bazel commands that reliably contain a test target. Any suggestions on where to place it? I could put it inside of the `//bazel` tree maybe?

jonmeow:
> I did think about it, but it didn't seem as clean. I _think_ this still builds everything and just produces the error at the end, but I'm not 100% certain. And I don't think the specific error code is an especially stable part of Bazel's command line API, so worried it could go away.

It's publicly documented as an API. update_roots and fix_cc_deps also special-case return codes. I think these non-1 return codes should probably be viewed as long-term stable APIs.

If you don't want to use it here, I won't keep pushing, but the workaround is taking more time for me to understand than using the exit code API.
 
> Yeah, this was somewhat just a guess for where to locate it. The only purpose I think is to allow forming Bazel commands that reliably contain a test target. Any suggestions on where to place it? I could put it inside of the `//bazel` tree maybe?

I think it should be placed alongside the file that's using it (so //scripts).

scripts/bazel_diff.py
jonmeow: Oh, I'm happy to switch to `run` if that's the actual advice. I'd probably not been reading closely. I've been using Python for way too long, so habits aren't always right.

jonmeow:
> I think it should be placed alongside the file that's using it (so //scripts).

Well, maybe not using it per se, but to the extent that the action's related script is in //scripts, it seems like that may be the best home for other related files.

chandlerc:
> > I did think about it, but it didn't seem as clean. I _think_ this still builds everything and just produces the error at the end, but I'm not 100% certain. And I don't think the specific error code is an especially stable part of Bazel's command line API, so worried it could go away.
> 
> It's publicly documented as an API. update_roots and fix_cc_deps also special-case return codes. I think these non-1 return codes should probably be viewed as long-term stable APIs.
> 
> If you don't want to use it here, I won't keep pushing, but the workaround is taking more time for me to understand than using the exit code API.

I'd prefer to stick with this approach specifically because it seems reasonable for Bazel to start exiting *early* rather than finishing building everything.

> > Yeah, this was somewhat just a guess for where to locate it. The only purpose I think is to allow forming Bazel commands that reliably contain a test target. Any suggestions on where to place it? I could put it inside of the `//bazel` tree maybe?
> 
> I think it should be placed alongside the file that's using it (so //scripts).

Just to note, it's the GitHub action workflow that needs this, not the `//scripts` part.

Putting this inside of `//.github/workflows` seems a bit surprising. Something in `//bazel` or `//testing` make more sense to me, er even `//scripts`. That said, where this goes is a very mild preference on my part. If you still think `//scripts` makes the most sense, happy to go with that, or any other location.

chandlerc:
> > I think it should be placed alongside the file that's using it (so //scripts).
> 
> Well, maybe not using it per se, but to the extent that the action's related script is in //scripts, it seems like that may be the best home for other related files.

Ah, just saw this, sorry.

Sure, I can move it to `//scripts`.

scripts/bazel_diff.py
chandlerc: Sure, done I think. I'll send a follow-up to update other scripts.

chandlerc (commented):
PTAL

jonmeow:
> Putting this inside of `//.github/workflows` seems a bit surprising. Something in `//bazel` or `//testing` make more sense to me, er even `//scripts`. That said, where this goes is a very mild preference on my part. If you still think `//scripts` makes the most sense, happy to go with that, or any other location.

To be sure, if you wanted to move the bazel-diff script that you've added along with the test to some `//testing/test_action`, I think that'd be fine too. But my sense is this won't see use outside the test action.

scripts/bazel_diff.py
```diff
@@ -25,145 +27,186 @@ def log(s: str) -> None:
...
-            "--stat",
-            f"{baseline}..{current}",
-        ],
+    p = subprocess.run(
+        ["git", "diff", "--stat", f"{baseline}..{current}"],
+        stdout=subprocess.PIPE,
         check=True,
         encoding="utf-8",
     )
+    log(p.stdout)
```
jonmeow: You could do: `stdout=sys.stderr.buffer`, removing the `stdout` and `encoding`. Note that would have the effect of printing output even on errors, when the current approach would suppress.

.github/workflows/tests.yaml
```diff
@@ -6,7 +6,7 @@ name: test
 
 on:
   push:
-    branches: [trunk]
+    branches: [trunk, action-test]
```
jonmeow: Don't forget to undo this.

.github/workflows/tests.yaml
```diff
@@ -174,23 +175,57 @@ jobs:
...
           BAZEL_USE_CPP_ONLY_TOOLCHAIN: 1
-        run: bazelisk build -c ${{ matrix.build_mode }} //...:all
+        run: |
+          bazelisk test -c ${{ matrix.build_mode }} //...
 
-      # Run all test targets.
+      # Compute the set of possible rules impacted by this change using
+      # Bazel-based diffing. This lets PRs and the merge queue have a much more
+      # efficient test CI action by avoiding even enumerating (and downloading)
+      # all of the unaffected Bazel targets
```
jonmeow:
```suggestion
      # all of the unaffected Bazel targets.
```

Sorry, punctuation style nit.

jonmeow (dismissed):
LGTM, feel free to merge -- just a couple last comments, including restoring the branches setting for tests.yaml.

scripts/bazel_diff.py
chandlerc: Brilliant, that's actually what I was trying to achieve, done.

chandlerc:
Well, this is annoying.

`bazel-diff` doesn't even handle the case of adding `bazel-diff` itself to your `WORKSPACE` file using its own instructions.

The failure is surprising -- it synthesizes as *workspace* rule rather than a *build* rule as the impacted dependency. So it can seem to work, unless you run it in exactly the precommit situation and include a `WORKSPACE` change.

I found some bugs about this closed indicating this is working-as-intended which makes me worry that this isn't actually going to be fixed.

However, in that bug, I found that there is another tool that has been built to solve this, and it even seems much (much) better: https://github.com/bazel-contrib/target-determinator

This actually bundles all the `git` manipulation out-of-the-box for us. I'm going to work on a fresh PR that switches to this tool. Unfortunately, it seems that the tool doesn't have obvious installation instructions, so I'll have to cobble something together to install the released binary. Stay tuned.

jonmeow:
> However, in that bug, I found that there is another tool that has been built to solve this, and it even seems much (much) better: https://github.com/bazel-contrib/target-determinator

Ouch, but hopefully this works well. Also promising that it's in bazel-contrib, so hopefully better support (more visible to the bazel team) long-term.

chandlerc:
Major update to a new tool, which dramatically simplifies everything. PTAL!

scripts/scripts_utils.py
```diff
@@ -85,69 +106,93 @@ def _download(url: str, local_path: Path) -> Optional[int]:
...
+    return _get_cached_binary(f"target-determinator{ext}", url, want_hash)
+
+
+def get_release(release: Release) -> str:
+    """Install a Bazel-released tool to carbon-lang's cache.
+
+    release: The release to cache.
+    """
+    # Translate platform information into Bazel's release form.
+    version = f"{platform.system().lower()}-{_get_machine()}"
```
jonmeow, line 187: This version label seems to be universal, maybe _get_machine() isn't quite the right helper granularity?

scripts/scripts_utils.py
```diff
@@ -85,69 +106,93 @@ def _download(url: str, local_path: Path) -> Optional[int]:
...
+    ext = _get_platform_ext()
+    url = f"{_BAZEL_TOOLS_URL}/{release.value}-{version}{ext}"
+
+    # Ensure the platform is supported, and grab its hash.
+    if version not in _BAZEL_TOOLS_VERSION_SHAS[release.value]:
+        # If this because a platform support issue, we may need to print errors.
+        exit(f"No {release.value} release available for platform: {version}")
+    want_hash = _BAZEL_TOOLS_VERSION_SHAS[release.value][version]
```
jonmeow: To the point of helpers, this seems like it could be a helper that takes in the sha list + version and returns the hash. You could also pass a label. The amount of code seems dense though to not have a helper, particularly in the context of something like _get_platform_ext()

scripts/scripts_utils.py
```diff
@@ -85,69 +106,93 @@ def _download(url: str, local_path: Path) -> Optional[int]:
...
 
+def _get_machine() -> str:
+    machine = platform.machine()
+    if machine == "x86_64":
+        machine = "amd64"
+    return machine
+
+
+def _get_platform_ext() -> str:
+    return ".exe" if platform.system() == "Windows" else ""
```
jonmeow: I'd really prefer to avoid Python's ternaries; while I understand the desire for brevity, the order of expressions really feels difficult to parse.

scripts/target_determinator.py
```diff
@@ -0,0 +1,109 @@
...
+from typing import List
+
+import scripts_utils
+
+
+def log(s: str) -> None:
+    print(s, file=sys.stderr)
+
+
+def quiet_run_output(args: List[str]) -> str:
```
jonmeow: This is only called once now, maybe it can be removed? Partly, I think the name makes less sense when quiet_run is gone.

scripts/target_determinator.py
```diff
@@ -0,0 +1,109 @@
...
+    with tempfile.NamedTemporaryFile(mode="w+") as tmp:
+        tmp.write(
+            f"let t = set({targets}) in "
+            "kind(rule, $t) except attr(tags, manual, $t)\n"
+        )
+        tmp.seek(0)
+        tmp_head = "".join(line for (line, _) in zip(tmp, range(10)))
+        if tmp.read(1) != "":
+            tmp_head += "...\n"
+        log(f"Bazel query file's first 10 lines:\n{tmp_head}---")
```
jonmeow: I was surprised to see the file immediately read back; if you want logging, can you log the query string directly?

jonmeow (commented):
Agreed, the new tool looks a lot simpler to use.

scripts/scripts_utils.py
chandlerc: Not quite, sadly.

The Bazel build tools use a `-` separator, and target-determinator uses `.` as the separator.

I can pass a separator into a helper function if you prefer, I just wasn't sure that the separator difference would be the only one going forward.

scripts/scripts_utils.py
chandlerc: Ah, sure. Done.

scripts/target_determinator.py
chandlerc: Sure, I'll switch to logging the string before writing it.

scripts/target_determinator.py
chandlerc: Sure, done.

chandlerc (commented):
Thanks for suggestions, PTAL

jonmeow (approved):
Note the ternary is still there, but approving.

scripts/target_determinator.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def filter_targets(bazel: Path, targets: str) -> str:
+    with tempfile.NamedTemporaryFile(mode="w+") as tmp:
+        query = (
+            f"let t = set({targets}) in "
+            "kind(rule, $t) except attr(tags, manual, $t)\n"
+        )
+        query_lines = query.splitlines()
+        query_snippet = "\n".join(query_lines[:5] + ["..."] + query_lines[-5:])
```
jonmeow:
```suggestion
        if len(query_lines) <= 10:
            query_snippet = "\n".join(query_lines)
        else:
            query_snippet = "\n".join(query_lines[:5] + ["..."] + query_lines[-5:])
```

chandlerc:
> Note the ternary is still there, but approving.

Thanks, and fixed!

Merging, hopefully this one sticks!

PR #3107, opened 2023-08-16
Fix missing APSInt include.

jonmeow:
Works by chance right now, but should be expected to break in future LLVM versions.

The builtin kind include is removed as unused.


chandlerc (approved):
LG

PR #3110, opened 2023-08-16
Lowering bug fix for tuple indexing.

fasiddique:
The existing tuple index lowering was crashing while accessing the element from return value. This PR fixes the bug.


jonmeow (approved):
Looks great! We should probably think about making the isPointerTy logic shared, but this seems okay as a spot fix.

PR #3111, opened 2023-08-17
Lowering for array index.

fasiddique:
This PR implements the lowering for array element access.  Besides, this creates a function for pointerTY to avoid code duplication.

### DIFF
--- a/toolchain/lowering/lowering_function_context.h
+++ b/toolchain/lowering/lowering_function_context.h
@@ -60,6 +60,14 @@ class LoweringFunctionContext {
     CARBON_CHECK(added) << "Duplicate local insert: " << node_id;
   }
 
+  auto GetValueForPoiterTY(llvm::Type* llvm_type, llvm::Value* val,
+                           unsigned idx, const llvm::Twine& name = "")
+      -> llvm::Value* {
+    return val->getType()->isPointerTy()
+               ? builder().CreateStructGEP(llvm_type, val, idx, name)
+               : builder().CreateExtractValue(val, idx);
+  }
+
   // Gets a callable's function.
   auto GetFunction(SemanticsFunctionId function_id) -> llvm::Function* {
     return lowering_context_->GetFunction(function_id);
--- a/toolchain/lowering/lowering_function_context.h
+++ b/toolchain/lowering/lowering_function_context.h
@@ -60,13 +60,14 @@ class LoweringFunctionContext {
     CARBON_CHECK(added) << "Duplicate local insert: " << node_id;
   }
 
-  // Checks if val is pointer type and returns the element value based on this.
-  auto GetValueForPoiterTY(llvm::Type* llvm_type, llvm::Value* val,
-                           unsigned idx, const llvm::Twine& name = "")
+  // Returns the requested index into val based on whether val is a pointer
+  // type.
+  auto GetIndexFromStructOrArray(llvm::Type* llvm_type, llvm::Value* val,
+                                 unsigned idx, const llvm::Twine& name)
       -> llvm::Value* {
     return val->getType()->isPointerTy()
                ? builder().CreateStructGEP(llvm_type, val, idx, name)
-               : builder().CreateExtractValue(val, idx);
+               : builder().CreateExtractValue(val, idx, name);
   }
 
   // Gets a callable's function.
--- a/toolchain/lowering/lowering_function_context.h
+++ b/toolchain/lowering/lowering_function_context.h
@@ -60,24 +60,13 @@ class LoweringFunctionContext {
     CARBON_CHECK(added) << "Duplicate local insert: " << node_id;
   }
 
-  // Check if val is pointer type and return the element value based on this.
+  // Checks if val is pointer type and returns the element value based on this.
   auto GetValueForPoiterTY(llvm::Type* llvm_type, llvm::Value* val,
                            unsigned idx, const llvm::Twine& name = "")
       -> llvm::Value* {
     return val->getType()->isPointerTy()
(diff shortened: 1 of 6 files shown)


toolchain/lowering/testdata/index/array_element_access.carbon
```diff
@@ -0,0 +1,50 @@
...
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+fn F() -> (i32, i32) { return (1, 2); }
+
+fn Run() {
+  var a: [i32; 2] = F();
+  var b: i32 = a[0];
+  var c: i32 = a[b];
```
jonmeow: Can you please also test direct indexing of a function returning an array, given that has different logic?

That is:

```
fn F() -> [i32; 2] { return (1, 2); }

fn Run() {
  var a: i32 = F()[0];
  var b: i32 = F()[a];
}
```

toolchain/lowering/lowering_function_context.h
```diff
@@ -60,6 +60,15 @@ class LoweringFunctionContext {
     CARBON_CHECK(added) << "Duplicate local insert: " << node_id;
   }
 
+  // Checks if val is pointer type and returns the element value based on this.
+  auto GetValueForPoiterTY(llvm::Type* llvm_type, llvm::Value* val,
+                           unsigned idx, const llvm::Twine& name = "")
+      -> llvm::Value* {
+    return val->getType()->isPointerTy()
+               ? builder().CreateStructGEP(llvm_type, val, idx, name)
+               : builder().CreateExtractValue(val, idx);
```
jonmeow: Why do you pass the name to gep, but not extractvalue?

toolchain/lowering/lowering_handle.cpp
```diff
@@ -24,10 +25,31 @@ auto LoweringHandleAddressOf(LoweringFunctionContext& context,
...
     -> void {
-  CARBON_FATAL() << "TODO: Add support: " << node;
+  auto [array_node_id, index_node_id] = node.GetAsArrayIndex();
+  auto* array_value = context.GetLocal(array_node_id);
+  auto* llvm_type =
+      context.GetType(context.semantics_ir().GetNode(array_node_id).type_id());
+  auto index_node = context.semantics_ir().GetNode(index_node_id);
+  llvm::Value* array_element_value;
+  if (index_node.kind() != SemanticsNodeKind::IntegerLiteral) {
+    auto* load_inst = context.builder().CreateLoad(
```
jonmeow: This is loading the index value, right? Maybe we can name the variable after what's contained instead of the function being called, with "index" instead of "load_inst"?

toolchain/lowering/lowering_handle.cpp
```diff
@@ -24,10 +25,31 @@ auto LoweringHandleAddressOf(LoweringFunctionContext& context,
...
+        llvm_type->getArrayElementType(), context.GetLocal(index_node_id));
+    array_element_value = context.builder().CreateInBoundsGEP(
+        llvm_type, array_value,
+        context.builder().CreateZExt(
```
jonmeow: Why does this need CreateZExt? Can you comment this code?

Also, please store the zext as a local, reducing function call nesting.

toolchain/lowering/lowering_handle.cpp
```diff
@@ -24,10 +25,31 @@ auto LoweringHandleAddressOf(LoweringFunctionContext& context,
...
+            load_inst, llvm::Type::getInt64Ty(context.llvm_context()),
+            "array.index"));
```
jonmeow: Below, you label the resulting value as "array.index". Here, you label the index as "array.index" and don't label the resulting value. Am I understanding that correctly? Can you look at this again?

toolchain/lowering/lowering_function_context.h
```diff
@@ -60,6 +60,15 @@ class LoweringFunctionContext {
     CARBON_CHECK(added) << "Duplicate local insert: " << node_id;
   }
 
+  // Checks if val is pointer type and returns the element value based on this.
+  auto GetValueForPoiterTY(llvm::Type* llvm_type, llvm::Value* val,
```
jonmeow: The "ForPointer" naming implies it only works for pointers. However, I think this function is intended for structs in general, regardless of whether it's a pointer. Maybe "GetIndexFromStructOrArray" would be a better name?

toolchain/lowering/lowering_function_context.h
```diff
@@ -60,6 +60,15 @@ class LoweringFunctionContext {
...
+                           unsigned idx, const llvm::Twine& name = "")
```
jonmeow: What would be the effect if you required a name be passed?

toolchain/lowering/lowering_function_context.h
```diff
@@ -60,6 +60,15 @@ class LoweringFunctionContext {
     CARBON_CHECK(added) << "Duplicate local insert: " << node_id;
   }
 
+  // Checks if val is pointer type and returns the element value based on this.
```
jonmeow:
```suggestion
  // Returns the requested index into val based on whether val is a pointer type.
```

I'm having trouble reading the current comment, what do you think of the above phrasing? Note how I avoid using "val" and "value" to mean different things, and avoid saying "this".

jonmeow (commented):
I think the `F()[a]` case might actually require a little extra work. If so, it might be appropriate for a todo, just marking that it doesn't lower correctly right now.

jonmeow (approved):
Thanks, looks good!

PR #3112, opened 2023-08-17
Language Server

maan2003:
Add a language server for carbon as part of GSoC.

This currently does code outline using toolchain parser.

See development steps in utils/vscode/README.md for running and using language server.


maan2003:
cc @josh11b for review

language_server/language_server.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include "toolchain/lexer/tokenized_buffer.h"
+#include "toolchain/parser/parse_node_kind.h"
+#include "toolchain/parser/parse_tree.h"
+#include "toolchain/source/source_buffer.h"
+
+namespace Carbon::LS {
+
+namespace {
+template <typename T>
+auto ParseJSON(llvm::json::Value& v, T& parsed) -> bool {
```
josh11b: Perhaps make these pointers if they are being modified as a side effect of this function, or `const &` if not?

language_server/language_server.cpp
```diff
@@ -0,0 +1,131 @@
...
+}  // namespace
+
+auto LanguageServer::onNotify(llvm::StringRef method, llvm::json::Value value)
+    -> bool {
+  if (method == "textDocument/didOpen") {
+    clang::clangd::DidOpenTextDocumentParams params;
+    if (!ParseJSON(value, params)) {
+      return false;
+    }
+    files_.emplace(params.textDocument.uri.file(), params.textDocument.text);
```
josh11b: I feel like this should `return true;` instead of falling through, since it is at this point that it is established that the function has succeeded.

language_server/language_server.cpp
```diff
@@ -0,0 +1,131 @@
...
+  }
+  if (method == "textDocument/didChange") {
```
josh11b: This is an alternative, not a case where both `if` branches could be taken.

```suggestion
  } else if (method == "textDocument/didChange") {
```

language_server/language_server.cpp
```diff
@@ -0,0 +1,131 @@
...
+    clang::clangd::DidChangeTextDocumentParams params;
+    if (!ParseJSON(value, params)) {
+      return false;
+    }
+    // full text is sent if full sync is specified in capabilities.
+    assert(params.contentChanges.size() == 1);
+    std::string file = params.textDocument.uri.file().str();
+    files_[file] = params.contentChanges[0].text;
+  }
+  return true;
```
josh11b, line 51: I'm surprised that an unrecognized `method` is considered a success. It would be clearer if this function had a comment explaining the contract in the `.h` file.

language_server/language_server.h
```diff
@@ -0,0 +1,40 @@
...
+#include <vector>
+
+#include "third_party/clangd/Protocol.h"
+#include "third_party/clangd/Transport.h"
+#include "toolchain/lexer/tokenized_buffer.h"
+#include "toolchain/parser/parse_tree.h"
+#include "toolchain/source/source_buffer.h"
+
+namespace Carbon::LS {
+class LanguageServer : public clang::clangd::Transport::MessageHandler {
```
josh11b: I'd expect to see comments documenting what this class does, as well as each method and static function.

language_server/language_server.h
```diff
@@ -0,0 +1,40 @@
...
+  auto onNotify(llvm::StringRef method, llvm::json::Value value)
+      -> bool override;
+  auto onCall(llvm::StringRef method, llvm::json::Value params,
+              llvm::json::Value id) -> bool override;
+  auto onReply(llvm::json::Value id, llvm::Expected<llvm::json::Value> result)
+      -> bool override;
+
+ private:
+  std::unique_ptr<clang::clangd::Transport> transport_;
+  std::unordered_map<std::string, std::string> files_;
```
josh11b, line 54: This is internal state that is mutated by the methods, and should be documented.

language_server/language_server.h
```diff
@@ -0,0 +1,40 @@
...
+
+  auto onNotify(llvm::StringRef method, llvm::json::Value value)
+      -> bool override;
+  auto onCall(llvm::StringRef method, llvm::json::Value params,
+              llvm::json::Value id) -> bool override;
+  auto onReply(llvm::json::Value id, llvm::Expected<llvm::json::Value> result)
+      -> bool override;
+
+ private:
+  std::unique_ptr<clang::clangd::Transport> transport_;
```
josh11b: Is this ever mutated, or can you add `const`?
```suggestion
  const std::unique_ptr<const clang::clangd::Transport> transport_;
```

third_party/clangd/Protocol.cpp
```diff
@@ -0,0 +1,251 @@
+//===--- Protocol.cpp - Language Server Protocol Implementation -----------===//
```
josh11b: Are these `clangd/` files copied directly? How should they be updated? My understanding was these files were already included in our LLVM dependency. It would be better to use that instead of having two copies.

language_server/language_server.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+auto LanguageServer::onCall(llvm::StringRef method, llvm::json::Value params,
+                            llvm::json::Value id) -> bool {
+  if (method == "initialize") {
+    llvm::json::Object capabilities{{"documentSymbolProvider", true},
+                                    {"textDocumentSync", /*Full=*/1}};
+    transport_->reply(
+        id, llvm::json::Object{{"capabilities", std::move(capabilities)}});
+  }
+  if (method == "textDocument/documentSymbol") {
```
josh11b:
```suggestion
  } else if (method == "textDocument/documentSymbol") {
```

language_server/language_server.cpp
```diff
@@ -0,0 +1,131 @@
...
+  return true;
+}
+
+auto LanguageServer::onReply(llvm::json::Value /*id*/,
+                             llvm::Expected<llvm::json::Value> /*result*/)
+    -> bool {
+  return true;
+}
+
+static auto getName(ParseTree& p, ParseTree::Node node)
```
josh11b, line 79: Comment saying what this function does so you don't have to read its implementation.

josh11b (commented):
I'll take another look after you've added some comments.

bazel/patches/llvm/0004_Add_library_for_clangd.patch
```diff
@@ -0,0 +1,276 @@
...
+index e44aee2d4..a031b877e 100644
+--- a/clang-tools-extra/clangd/Protocol.cpp
++++ b/clang-tools-extra/clangd/Protocol.cpp
+@@ -12,10 +12,8 @@
+
+ #include "Protocol.h"
+ #include "URI.h"
+ #include "support/Logger.h"
+-#include "clang/Basic/LLVM.h"
+-#include "clang/Index/IndexSymbol.h"
```
josh11b: Making this diff is small as possible is important -- what are the reasons for all of these changes? Can any be skipped? Can the reasoning be documented somewhere?

maan2003: remove SymbolID changes can be skipped, but at cost of compiling clang. 

these changes should be easy to redo when updating llvm.

language_server/language_server.h
```diff
@@ -0,0 +1,73 @@
...
+namespace Carbon::LS {
+class LanguageServer : public clang::clangd::Transport::MessageHandler,
+                       public clang::clangd::LSPBinder::RawOutgoing {
+ public:
+  // Start the language server.
+  static void Start();
+
+  // Transport::MessageHandler
+  // Handlers returns true to keep processing messages, or false to shut down.
+  // Handler called on notification by client.
```
josh11b, line 28:
```suggestion

  // Handler called on notification by client.
```

language_server/language_server.h
```diff
@@ -0,0 +1,73 @@
...
+  const std::unique_ptr<clang::clangd::Transport> transport_;
+  // content of files managed by the language client.
+  std::unordered_map<std::string, std::string> files_;
+  // handlers for client methods and notifications
+  clang::clangd::LSPBinder::RawHandlers handlers_;
+
+  explicit LanguageServer(std::unique_ptr<clang::clangd::Transport> transport)
+      : transport_(std::move(transport)) {}
+
+  // handlers
```
josh11b: Please be more specific. What are all of these handling, and what happens as a result of these calls?

utils/vscode/README.md
```diff
@@ -25,8 +25,12 @@ code --install-extension out/carbon.vsix
 
 ## Development
 
-1. Open utils/vscode folder in VS Code.
-2. Launch the extension using Run command (F5).
+1. `bazel build language_server` in project root.
```
josh11b, line 28: Maybe:
```suggestion
1. `bazel build -c opt language_server` in project root.
```

utils/vscode/src/extension.js
```diff
@@ -4,7 +4,32 @@
...
+  const serverOptions = {
+    run: { command },
+    debug: { command },
+  };
+
+  const clientOptions = {
+    documentSelector: [{ scheme: 'file', language: 'carbon' }],
+  };
+
+  let disposable = new LanguageClient(
```
josh11b: disposable? Perhaps there is a more descriptive name for this?

utils/vscode/README.md
maan2003: for now debug build is fine, build clang in release mode will be very slow. 

If we add more time sensitive things (like completion), we can this.

josh11b (approved):

bazel/patches/llvm/0004_Add_library_for_clangd.patch
```diff
@@ -0,0 +1,77 @@
...
+Subject: [PATCH] Add library for clangd
+
+This exports clangd language server protocol helpers publically.
+Feature.h needs Feature.inc which is generated during build process, so we
+disable all features for now.
+---
+ clang-tools-extra/clangd/BUILD.bazel  | 36 +++++++++++++++++++++++++++
+ clang-tools-extra/clangd/Protocol.cpp | 15 +++++++----
+ clang-tools-extra/clangd/Protocol.h   | 12 +++++++--
+ clang-tools-extra/clangd/Transport.h  |  1 -
```
maan2003, line 13: TODO: regenerate the patch file with git instead of manually modifying it.

PR #3113, opened 2023-08-17
Explorer: Remove `PrintDepth` and implement `PrintIndent`.

prabhatexit0:
This PR removes `PrintDepth` from statement and declaration.
Implements `PrintIndent` for better indented formatting along with various changes to make printing of statements and declarations better.


explorer/interpreter/exec_program.cpp
```diff
@@ -28,10 +28,13 @@ auto AnalyzeProgram(Nonnull<Arena*> arena, AST ast,
...
-    for (auto& declaration : ast.declarations) {
-      set_file_ctx.update_source_loc(declaration->source_loc());
+    for (size_t i = 0; i < ast.declarations.size(); ++i) {
+      set_file_ctx.update_source_loc(ast.declarations[i]->source_loc());
       if (trace_stream->is_enabled()) {
-        *trace_stream << *declaration;
+        *trace_stream << *ast.declarations[i];
+        if (i != ast.declarations.size() - 1) {
+          *trace_stream << "\n\n";
+        }
```
jonmeow: You can use llvm::Separator for cases like this. i.e.:

```
llvm::Separator sep("\n\n");
for (...) {
  *trace_stream << sep << declaration;
}
```

explorer/ast/declaration.h
```diff
@@ -50,6 +50,14 @@ class Declaration : public AstNode {
   void Print(llvm::raw_ostream& out) const override;
   void PrintID(llvm::raw_ostream& out) const override;
 
+  // The depth parameter signals how many levels of nesting to print, so
+  // recursive calls can pass `depth - 1`, and when depth is 0 it prints "..."
+  // and returns. when depth is -1 that means "no depth limit".
```
jonmeow:
```suggestion
  // and returns. When depth is -1 that means "no depth limit".
```

explorer/ast/statement.h
```diff
@@ -29,9 +29,16 @@ class Statement : public AstNode {
...
 
-  void Print(llvm::raw_ostream& out) const override { PrintDepth(-1, out); }
-  void PrintID(llvm::raw_ostream& out) const override { PrintDepth(1, out); }
-  void PrintDepth(int depth, llvm::raw_ostream& out) const;
+  void Print(llvm::raw_ostream& out) const override { PrintDepth(-1, 0, out); }
+  void PrintID(llvm::raw_ostream& out) const override { PrintDepth(1, 0, out); }
+
+  // The depth parameter signals how many levels of nesting to print, so
+  // recursive calls can pass `depth - 1`, and when depth is 0 it prints "..."
+  // and returns. when depth is -1 that means "no depth limit".
```
jonmeow:
```suggestion
  // and returns. When depth is -1 that means "no depth limit".
```

explorer/ast/statement.cpp
```diff
@@ -15,111 +15,136 @@ using llvm::cast;
...
     }
     case StatementKind::If: {
       const auto& if_stmt = cast<If>(*this);
-      out << "if (" << if_stmt.condition() << ")\n";
-      if_stmt.then_block().PrintDepth(depth - 1, out);
-      if (if_stmt.else_block()) {
-        out << "\nelse\n";
-        (*if_stmt.else_block())->PrintDepth(depth - 1, out);
+      out.indent(indent_num_spaces) << "if (" << if_stmt.condition() << ") ";
+      if (depth < 0 || depth > 1) {
```
jonmeow: On line 20, there's handling for depth == 0. So is this equivalent to "if (depth != 1)"?

In the prior code, when depth == 1, it would've called PrintDepth with depth == 0 which would've returned. Are these changes necessary? That is, before, wouldn't the code have printed `if (cond) ... else ...`?

explorer/ast/statement.cpp
```diff
@@ -15,111 +15,136 @@ using llvm::cast;
...
+        if_stmt.then_block().PrintDepth(depth - 1, indent_num_spaces, out);
+        if (if_stmt.else_block()) {
+          out << "\n";
+          out.indent(indent_num_spaces) << "else\n";
+          (*if_stmt.else_block())
+              ->PrintDepth(depth - 1, indent_num_spaces + 2, out);
+        }
+      } else {
+        if (if_stmt.else_block()) {
+          out << "} else { ... }";
```
jonmeow: I'm a little confused by this, it doesn't look like you print the if block, so won't this result in:

`if (cond) } else { ... }`

explorer/testdata/trace/phase_source_program.carbon
```diff
@@ -19,10 +19,7 @@ fn Main() -> i32 {
...
-// CHECK:STDOUT: fn Main ()-> i32 {
-// CHECK:STDOUT: {
-// CHECK:STDOUT: return 0;
-// CHECK:STDOUT: }
 // CHECK:STDOUT:
-// CHECK:STDOUT: }
-// CHECK:STDOUT: result: 0
+// CHECK:STDOUT: fn Main () -> i32 {
+// CHECK:STDOUT:   return 0;
+// CHECK:STDOUT: }result: 0
```
jonmeow: This looks like a newline is msising, can you please fix it?

explorer/ast/statement.cpp
```diff
@@ -15,111 +15,136 @@ using llvm::cast;
...
       const auto& while_stmt = cast<While>(*this);
-      out << "while (" << while_stmt.condition() << ")\n";
-      while_stmt.body().PrintDepth(depth - 1, out);
+      out.indent(indent_num_spaces)
+          << "while (" << while_stmt.condition() << ")\n";
+      if (depth < 0 || depth > 1) {
+        while_stmt.body().PrintDepth(depth - 1, indent_num_spaces + 2, out);
+      } else {
+        out.indent(indent_num_spaces) << "...";
+      }
```
jonmeow: Here and in some other places, why do you print at a different indent based on the depth value?

explorer/ast/statement.cpp
```diff
@@ -15,111 +15,136 @@ using llvm::cast;
...
       break;
     case StatementKind::VariableDefinition: {
       const auto& var = cast<VariableDefinition>(*this);
       if (var.is_returned()) {
-        out << "returned ";
+        out.indent(indent_num_spaces) << "returned ";
       }
-      out << "var " << var.pattern();
+      out.indent(var.is_returned() ? 0 : indent_num_spaces)
+          << "var " << var.pattern();
```
jonmeow: Can you avoid the multiple `out.indent` calls if you call it before the `if (var.is_returned())`?

explorer/ast/statement.cpp
```diff
@@ -15,111 +15,136 @@ using llvm::cast;
 
 Statement::~Statement() = default;
 
-void Statement::PrintDepth(int depth, llvm::raw_ostream& out) const {
+void Statement::PrintDepth(int depth, int indent_num_spaces,
+                           llvm::raw_ostream& out) const {
   if (depth == 0) {
-    out << " ... ";
+    out.indent(indent_num_spaces) << " ... ";
```
jonmeow: It looks like every form of statement starts with an indent, except for Block.

Second, what do you think of calling `out.indent` once at the start of this PrintDepth function, instead of at the start of each statement? Given `Block`, something like:

```
if (kind() != Statement::Block) {
  out.indent(...);
}
```

Note though, we'll probably support actual block statements down the line (doesn't seem to parse right now, surprisingly), for scoping reasons. e.g.:

```
fn F() {
  // Block statement:
  {
     var scoped: MyType = ...;
  }
  // Code that depends on `scoped` being destroyed goes here.
}
```

And then those should be indented too.

explorer/ast/declaration.cpp
```diff
@@ -15,104 +15,112 @@ using llvm::cast;
...
       break;
     }
     case DeclarationKind::ImplDeclaration: {
       const auto& impl_decl = cast<ImplDeclaration>(*this);
-      PrintID(out);
-      out << " {\n";
+      out.indent(indent_num_spaces) << PrintAsID(impl_decl) << " {\n";
       for (Nonnull<Declaration*> m : impl_decl.members()) {
-        out << *m;
+        m->PrintDepth(-1, indent_num_spaces + 2, out);
```
jonmeow: Shouldn't cases like this be passing `depth - 1` if you're adding `depth` as a parameter?

explorer/ast/declaration.cpp
```diff
@@ -15,104 +15,112 @@ using llvm::cast;
 Declaration::~Declaration() = default;
 
 void Declaration::Print(llvm::raw_ostream& out) const {
+  PrintDepth(-1, 0, out);
+}
+void Declaration::PrintDepth(int depth, int indent_num_spaces,
+                             llvm::raw_ostream& out) const {
+  if (depth == 0) {
+    out.indent(indent_num_spaces) << " ... ";
```
jonmeow: Can this indent once at the start of the function, for all declarations, avoiding the call at the start of each declaration?

jonmeow (commented):
zygoloid's out, so I'm taking a look at this.

I think the indentation changes are good, but I'm seeing some oddities with depth handling (I commented on examples). I've asked if there's a need to keep `depth`, maybe we can just get rid of it given that it's not used:
[#explorer question](https://discord.com/channels/655572317891461132/763516049710120960/1142124349386657923)

explorer/ast/statement.cpp
jonmeow: Note on indenting blocks, in some cases like `while` it's going to be after a newline, won't it? i.e., you might want to test, but I think you'd get:

```
fn F() {
  while (...)
{
    // while body is indented, open curly is not.
  }
}
```

This might speak to always indenting blocks, and always putting the open curly on the next line (with appropriate indent)

explorer/interpreter/exec_program.cpp
prabhatexit0: Oh okay but do you mean `llvm::ListSeparator`?

explorer/testdata/trace/phase_source_program.carbon
prabhatexit0: Yes, will do in the next commit.
Thanks for pointing it out.

explorer/ast/statement.cpp
prabhatexit0: Oh yes, these issues can arise with the current way of indentation. Thanks for pointing this out!
I was thinking can we have formatting like this?

```
fn Foo() -> type
{
  // body
  if(...)
  {
    ...
  }
}
```

instead of 
```
fn Foo() -> type {
  // body
  if(...) {
    ...
  }
}
```

It would be a bit simpler to implement.
What do you think?

explorer/ast/declaration.cpp
prabhatexit0: Sure, I'll do that

explorer/ast/declaration.cpp
prabhatexit0: I wasn't exactly sure how the `depth` thing was working by the time I wrote this code.
I'm waiting for your response in the **#explorer** on discord regarding if we need to keep `PrintDepth` or not before working on this change.

explorer/ast/statement.cpp
prabhatexit0: Sure, I'll do it in the next commit.

explorer/ast/statement.cpp
prabhatexit0: I wasn't exactly sure how the `depth` thing was working by the time I wrote this code.
I'm waiting for your response in the **#explorer** on discord regarding if we need to keep `PrintDepth` or not before working on this change.  (2)

explorer/ast/statement.cpp
prabhatexit0: Oh, I'll fix this in the next commit.

explorer/ast/statement.cpp
prabhatexit0: I wasn't exactly sure how the `depth` thing was working by the time I wrote this code.
I'm waiting for your response in the **#explorer** on discord regarding if we need to keep `PrintDepth` or not before working on this change. (3)

explorer/ast/declaration.cpp
jonmeow: Let's go ahead with that, maybe also renaming PrintDepth to PrintIndented or similar?

explorer/ast/statement.cpp
jonmeow: I think that'd be fine, yes.

jonmeow (commented):
As noted on #explorer, let's go ahead with removing depth. I'm betting against objections since it's been a couple days, and it's unused in code. :)

explorer/ast/declaration.cpp
prabhatexit0: Done, renamed it to `PrintIndent`.

explorer/ast/declaration.h
```diff
@@ -238,8 +240,8 @@ class CallableDeclaration : public Declaration {
         body_(context.Clone(other.body_)),
         virt_override_(other.virt_override_) {}
 
-  void PrintDepth(int depth, llvm::raw_ostream& out) const;
-
+  void PrintIndent(int indent_num_spaces,
+                   llvm::raw_ostream& out) const override;
```
jonmeow, line 244: So you know, I'm looking at this override and I think we should probably remove it (inlining the code) to reduce complexity and be more consistent (i.e., removing the indent special-casing for callables). It'd also mean PrintIndent wouldn't need to be virtual. I'm fine with it for now since the function was there before, I just want you to know the line of thought.

explorer/ast/statement.cpp
```diff
@@ -15,39 +15,86 @@ using llvm::cast;
 
 Statement::~Statement() = default;
 
-void Statement::PrintDepth(int depth, llvm::raw_ostream& out) const {
-  if (depth == 0) {
-    out << " ... ";
-    return;
+void Statement::PrintID(llvm::raw_ostream& out) const {
```
jonmeow, line 18: While this is replacing the PrintDepth(1, ...) call, will this work as intended? Every use of `Print`, including the implicit ones like `<< match.expression() <<`, corresponds to a fully recursive call.

Maybe it should just print a keyword or brief placeholder matching the statement? e.g., `match` for Match, `{ ... }` for Block, PrintID for ExpressionStatement, etc. This would be similar to how expressions print IDs.

explorer/ast/statement.cpp
```diff
@@ -15,39 +15,86 @@ using llvm::cast;
...
+    case StatementKind::ReturnExpression: {
+      Print(out);
+    }
+  }
+}
+
+void Statement::PrintIndent(int indent_num_spaces,
+                            llvm::raw_ostream& out) const {
+  if (kind() != StatementKind::Block) {
+    out.indent(indent_num_spaces);
```
jonmeow: Under the current approach, Block is also indented, so this conditional shouldn't be needed.

explorer/ast/statement.cpp
```diff
@@ -79,16 +126,19 @@ void Statement::PrintDepth(int depth, llvm::raw_ostream& out) const {
...
+      out << (inc_dec.is_increment() ? "++" : "--") << inc_dec.argument()
+          << ";";
       break;
     }
     case StatementKind::If: {
       const auto& if_stmt = cast<If>(*this);
-      out << "if (" << if_stmt.condition() << ")\n";
-      if_stmt.then_block().PrintDepth(depth - 1, out);
+      out << "if (" << if_stmt.condition() << ")";
+      out << "\n";
```
jonmeow:
```suggestion
      out << "if (" << if_stmt.condition() << ")\n";
```

explorer/ast/statement.cpp
```diff
@@ -79,16 +126,19 @@ void Statement::PrintDepth(int depth, llvm::raw_ostream& out) const {
...
+      if_stmt.then_block().PrintIndent(indent_num_spaces, out);
       if (if_stmt.else_block()) {
-        out << "\nelse\n";
-        (*if_stmt.else_block())->PrintDepth(depth - 1, out);
+        out << "\n";
+        out.indent(indent_num_spaces) << "else\n";
+        (*if_stmt.else_block())->PrintIndent(indent_num_spaces + 2, out);
```
jonmeow:
```suggestion
        (*if_stmt.else_block())->PrintIndent(indent_num_spaces, out);
```

explorer/ast/statement.cpp
```diff
@@ -15,39 +15,86 @@ using llvm::cast;
...
-      while_stmt.body().PrintDepth(depth - 1, out);
+      while_stmt.body().PrintIndent(indent_num_spaces, out);
       break;
     }
     case StatementKind::For: {
       const auto& for_stmt = cast<For>(*this);
       out << "for (" << for_stmt.variable_declaration() << " in "
           << for_stmt.loop_target() << ")\n";
-      for_stmt.body().PrintDepth(depth - 1, out);
+      for_stmt.body().PrintIndent(indent_num_spaces + 2, out);
```
jonmeow:
```suggestion
      for_stmt.body().PrintIndent(indent_num_spaces, out);
```

With the way blocks are printed, they shouldn't have an increased indent.

explorer/ast/statement.cpp
```diff
@@ -15,39 +15,86 @@ using llvm::cast;
...
   switch (kind()) {
     case StatementKind::Match: {
       const auto& match = cast<Match>(*this);
-      out << "match (" << match.expression() << ") {";
-      if (depth < 0 || depth > 1) {
+      out << "match (" << match.expression() << ") {\n";
+      for (const auto& clause : match.clauses()) {
+        out.indent(indent_num_spaces + 2)
+            << "case " << clause.pattern() << " => \n";
+        clause.statement().PrintIndent(indent_num_spaces + 2, out);
```
jonmeow, line 97:
```suggestion
        clause.statement().PrintIndent(indent_num_spaces, out);
```

This is actually always a block, so probably shouldn't increase indent.

For reference, from parser.ypp:
```
clause:
  CASE pattern DOUBLE_ARROW block
    { $$ = Match::Clause($[pattern], $[block]); }
| DEFAULT DOUBLE_ARROW block
    {
      $$ = Match::Clause(arena->New<BindingPattern>(
                             context.source_loc(), std::string(AnonymousName),
                             arena->New<AutoPattern>(context.source_loc()),
                             ExpressionCategory::Value),
                         $[block]);
    }
;
```

jonmeow (commented):
Thanks! I think this is the right direction, though we probably still need to simplify PrintID to avoid recursive printing.

explorer/ast/declaration.h
prabhatexit0: Thanks for letting me know about it! I can do something about it in some other PR.

explorer/ast/statement.cpp
prabhatexit0: Oh I'm sorry I missed that. I'll do that in the next commit. Thanks!

explorer/ast/statement.cpp
prabhatexit0: Okay. I will remove the conditional in the next commit.

explorer/ast/statement.cpp
prabhatexit0: Thanks for pointing this out!

explorer/ast/statement.cpp
prabhatexit0: Doing this is leading to this output"
```
    match (self.element) {
      case OptionalElement(T).None =>
    {
      return false;
    }
    }
```

Note: for clause the indent is `indent_num_spaces + 2` as it is nested inside the match block.

explorer/ast/statement.cpp
prabhatexit0: After trying it out I found out that printing some elements with very less detail was making it less readable in the trace output. 
So, for block I'm doing `PrintAsID` for the statements it contains, also doing `PrintAsID` for few other cases.
What do you think about this?

jonmeow: I think that PrintID should not do anything that is recursive or loops. If you are fine with recursion or looping, you should probably be using Print instead. [note, part of the issue with looping in a block is that you don't know how many statements there are, and there may be lots]

explorer/ast/statement.cpp
jonmeow: Ah, thanks, I'd missed it matching the clause indent.

explorer/ast/statement.cpp
prabhatexit0: Oh, to get rid of this looping issue with the block can we have the following cases?

if the number of statements inside the block are less than or equal to 2
```
{id stmt 1; id stmt 2}
```
if the statements in the block are greater than 2
```
{id stmt 1; ... id stmt2}
```

although in the trace output, source location is also added in the cases when `PrintID` is used but printing `{...}` for blocks is making it a bit hard to understand which part of code it belongs to.

jonmeow: That would be fine, but along similar lines you might consider:

0: `{}`
1: `{id stmt 1;}`
N: `{id stmt 1; ...[N-1] statements...}`

i.e., after the 1st statement, just print the # of statements that were elided.

prabhatexit0: Nice suggestion, I will implement that.

prabhatexit0: I also used `PrintID` at places where I felt like the information could be useful. Let me know what you think about it.
And I noticed an issue in expression's `PrintID` due to it there was an assertion error happening, I also tried to fix that.

jonmeow: Thanks for pointing out the changes. I'm still concerned about this: while a Block statement would just be `{ ... }` which could be confusing, there are more cues (keywords) for other pieces. I think too that not recursing is more in keeping with the prior PrintDepth implementation.

Note especially that your changes to For are back to recursively printing the entire for loop contents. This is returning to the issue I was trying to point out originally, that PrintID needs to not be fully recursive.

I would appreciate going back to a simple implementation so that we can finish review.

prabhatexit0: > I'm still concerned about this: while a Block statement would just be { ... } which could be confusing, there are more cues (keywords) for other pieces. I think too that not recursing is more in keeping with the prior PrintDepth implementation.

Oh okay. Thanks for letting me know about that!

> Note especially that your changes to For are back to recursively printing the entire for loop contents. This is returning to the issue I was trying to point out originally, that PrintID needs to not be fully recursive.

I actually forgot to use `PrintAsID(...)` there but going back to simple implementation.

> I would appreciate going back to a simple implementation so that we can finish review.

I'm reverting back to the simple implementation in next commit.

jonmeow: I do understand that non-recursive yields less information, but I think things like `{var x: T = {}.(interface ImplicitAs(T = class T).Convert)();return 0;}` (which had been in trace output before) are actually more indicative of an issue that needed to be fixed, that arbitrary amounts of output were getting printed.

I think this current setup is good, thanks!

jonmeow (approved):
Thanks for all the changes! I'll go ahead and merge this.

prabhatexit0:
> Thanks for all the changes! I'll go ahead and merge this.

Thanks for the review!

PR #3114, opened 2023-08-17
add documentation about debugging on macOS

lnihlen:
Adds some notes about the required build flags for lldb to successfully find the symbols on macOS debug builds. Also adds a recommended debugger configuration for interactive debugging in VSCode on macOS.

### DIFF
--- a/docs/project/contribution_tools.md
+++ b/docs/project/contribution_tools.md
@@ -13,18 +13,21 @@ contributions.
 
 ## Table of contents
 
--   [Setup commands](#setup-commands)
-    -   [Debian or Ubuntu](#debian-or-ubuntu)
-    -   [macOS](#macos)
--   [Tools](#tools)
-    -   [Main tools](#main-tools)
-        -   [Running pre-commit](#running-pre-commit)
-    -   [Optional tools](#optional-tools)
-    -   [Manually building Clang and LLVM (not recommended)](#manually-building-clang-and-llvm-not-recommended)
--   [Troubleshooting build issues](#troubleshooting-build-issues)
-    -   [Old LLVM versions](#old-llvm-versions)
-    -   [Asking for help](#asking-for-help)
--   [Troubleshooting debug issues](#troubleshooting-debug-issues)
+- [Contribution tools](#contribution-tools)
+  - [Table of contents](#table-of-contents)
+  - [Setup commands](#setup-commands)
+    - [Debian or Ubuntu](#debian-or-ubuntu)
+    - [macOS](#macos)
+  - [Tools](#tools)
+    - [Main tools](#main-tools)
+      - [Running pre-commit](#running-pre-commit)
+    - [Optional tools](#optional-tools)
+    - [Manually building Clang and LLVM (not recommended)](#manually-building-clang-and-llvm-not-recommended)
+  - [Troubleshooting build issues](#troubleshooting-build-issues)
+    - [Old LLVM versions](#old-llvm-versions)
+    - [Asking for help](#asking-for-help)
+  - [Troubleshooting debug issues](#troubleshooting-debug-issues)
+    - [Debugging on MacOS](#debugging-on-macos)
 
 <!-- tocstop -->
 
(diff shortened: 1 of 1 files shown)


docs/project/contribution_tools.md
```diff
@@ -268,3 +269,46 @@ Dwarf Error: DW_FORM_strx1 found in non-DWO CU
...
+
+If this build command doesn't seem to produce a debuggable binary you might need
+to both clear the build disk cache and clean the build. Running
+`scripts/clean_disk_cache.sh` may not be enough, you might try deleting all the
+files within the disk cache, typically located at
+`~/.cache/carbon-lang-build-cache`. Deleting the disk cache, followed by a
+`bazel clean` should allow your next rebuild, with the recommended options, to
+supply the symbols for debugging.
+
+For debugging on MacOS using VSCode, I've had good luck using the CodeLLDB
```
josh11b: Not clear who "I" is in this context. Perhaps rephrase assuming you don't expect people to track you down with questions? Maybe something like: "some people have had success using..."

lnihlen: Good call, reworded.

docs/project/contribution_tools.md
```diff
@@ -268,3 +269,46 @@ Dwarf Error: DW_FORM_strx1 found in non-DWO CU
...
 It means that the version of GDB used is too old, and does not support the DWARF
 v5 format.
+
+### Debugging on MacOS
+
+Getting Bazel to build debuggable binaries on MacOS requires passing a few extra
+options while building:
+
+```shell
+bazel build --spawn_strategy=local -c dbg //explorer --apple_generate_dsym
```
chandlerc: So, first off, super excited to have these steps. I don't think I would have ever figured out what steps to try here. Hat's off to you. All of the instructions worked perfectly for me.

I was really curious how this is so very broken on macOS, so I did a bunch of digging. I think the best explanation is scattered in some (but not all) of the comments on a Bazel bug and summarized pretty well here:
https://github.com/bazelbuild/bazel/issues/2537#issuecomment-449089673

Might be worth linking to that bug from the docs?

After reading that and trying a few things, I don't think you need `--apple_generate_dsym` here. As far as I can tell, it's not generating dsyms in the build, and that flag influences the Apple build rules (objc_binary and such) rather than the C/C++ rules.

The local spawn strategy is the key thing, because *without* dsyms, the linker encodes the paths to the `.o` files during the link, and the `.o` files to the source files during the compile. When sandboxed, none of these paths work after the build. But by disabling the sandbox, the paths computed are all stable and work.

And I think the issue with the build cache is because which path is in the `.o` file isn't a function of anything that keys the build cache.

I'm not sure it's realistic to get the dsyms built until Bazel team fix that bug. But I'm pondering whether we could (a) force macOS builds to use the local strategy somehow so folks don't have to discover this, and (b) cause the relevant paths to be *inputs* to the compiles so that the cache is more reliable. Right now, I think you could get "exciting" situations where two builds in two checkouts could share `.o` files in the cache and end up with LLDB showing source code from the wrong client. Anyways, none of this is blocking -- having these instructions is already a *huge* step forward, so thanks again.

docs/project/contribution_tools.md
```diff
@@ -268,3 +269,46 @@ Dwarf Error: DW_FORM_strx1 found in non-DWO CU
...
+`scripts/clean_disk_cache.sh` may not be enough, you might try deleting all the
+files within the disk cache, typically located at
+`~/.cache/carbon-lang-build-cache`. Deleting the disk cache, followed by a
+`bazel clean` should allow your next rebuild, with the recommended options, to
+supply the symbols for debugging.
+
+For debugging on MacOS using VSCode, some people have had success using the
+CodeLLDB extension. In order for LLDB to connect the project source files with
+the symbols you will need to add a `"sourceMap": { ".": "${workspaceRoot}" }`
+line to the `launch.json` configuration, for example:
```
chandlerc: Just to make this more obvious for folks not yet familiar with CodeLLDB, maybe:
```suggestion
line to the CodeLLDB `launch.json` configuration, for example:
```

chandlerc (approved):
LGTM (with the suggestions inline)

docs/project/contribution_tools.md
lnihlen: Reworked, tested without `--apple_generate_dsym` and added the link to the bug. PTAL

josh11b (approved):

PR #3117, opened 2023-08-18
Add `addr self` binding support in destructors

### LINKED ISSUE #2802: Add `addr self` binding support for destructors
### Description of the bug:

`addr self` is currently unsupported by the destructor support. It's worth adding support, ideally by merging CallDestructor logic more with CallFunction.

```
package ExplorerTest impl;

class A {
  destructor[addr self: Self*] {
    self->n += 1;
    Print("DESTRUCTOR A {0}", self->n);
  }

  var n: i32;
}

fn Main() -> i32 {
  var a: A = {.n = 2};
  return 0;
}
```

Should work, printing 3.

### What actually happened?

Currently a crash, making it an error though.

_No response_

lnihlen:
Refactors `CallDestructor` and `CallFunction` to both call a new method `BindSelfIfPresent`, which includes support for binding `addr self` if specified. Fixes the associated unit test.

Closes #2802.


explorer/interpreter/interpreter.cpp
```diff
@@ -182,7 +184,15 @@ class Interpreter {
       -> ErrorOr<Success>;
 
   auto CallDestructor(Nonnull<const DestructorDeclaration*> fun,
-                      Nonnull<const Value*> receiver) -> ErrorOr<Success>;
+                      Nonnull<const Value*> receiver,
+                      const LocationValue* class_addr) -> ErrorOr<Success>;
```
geoffromer: This method should have a comment. I'm particularly looking to understand the meaning of `class_addr`.

explorer/interpreter/interpreter.cpp
```diff
@@ -182,7 +184,15 @@ class Interpreter {
...
+
+  // If the given `method` has a self argument, bind it to `method_scope`.
```
geoffromer: What do the other parameters mean? I'm particularly looking to understand the relationship between `receiver` and `receiver_addr`. It's especially puzzling to me that they seem to be identical except in the destructor case -- if we could arrange for them to always be identical, we could drop one of the parameters. That would be really nice, because 6 is kind of a lot.

explorer/interpreter/interpreter.cpp
```diff
@@ -182,7 +184,15 @@ class Interpreter {
...
+  void BindSelfIfPresent(Nonnull<const CallableDeclaration*> decl,
+                         Nonnull<const Value*> receiver,
+                         Nonnull<const Value*> receiver_addr,
+                         const SourceLocation& source_location,
```
geoffromer: By convention, `SourceLocation` parameters tend to go last.

explorer/interpreter/interpreter.cpp
lnihlen: I will happily write a comment and fix the comment on the other method, and was hoping for some feedback about this approach.

The problem is within `BindSelfIfPresent`. If the source code is using the immutable `[self: Self]` pattern, which is the only binding destructors used to support, the `method_scope.BindValue(` call would happily bind to the `NominalClassValue` provided by calling code in `receiver`. However, in the `[addr self: Self*]` case, `PatternMatch` will assert on a binding to any value that is not a `LocationValue` type. So I now ask the caller to provide both values.

I could refactor the code in the interpreter that calls `CallDestructor` to only provide one, but it would then need to know the arguments to the destructor to understand which one to provide, thus duplicating the logic in `BindSelfIfPresent`. So I've gone with the 2 args. I'd love if there's a more elegant way to do this, do you have anything in mind?

explorer/interpreter/interpreter.cpp
lnihlen: Moved, thanks.

explorer/interpreter/interpreter.cpp
lnihlen: Added a comment with explanation.

explorer/interpreter/interpreter.cpp
geoffromer: I think that, rather than construct an `ExpressionResult` on the fly for the `PatternMatch` call, we should pass the receiver into `CallFunction` and `CallDestructor` as an `ExpressionResult`. This seems like exactly the kind of problem that `ExpressionResult` is designed to solve: representing a value that might or might not have an address, in a context that might want to operate on the value or the address. As a bonus, I think this could also help resolve the TODO in `BindSelfIfPresent`.

(I'm a little concerned about the way `ExpressionResult` blurs the line between compile-time and run-time, but that's a pre-existing problem, and a much bigger one. I have a terrible feeling the answer there will wind up being that we have to split `PatternMatch` up into `Action`s, so that it can do things like evaluate the implicit address-of expression in an `addr` pattern.)

explorer/interpreter/interpreter.cpp
```diff
@@ -1016,6 +976,60 @@ auto Interpreter::CallFunction(const CallExpression& call,
...
+    ExpressionResult v = receiver;
+    // See if we need to make a LocationValue from the address provided in the
+    // ExpressionResult
+    if (receiver.value()->kind() != Value::Kind::LocationValue) {
+      CARBON_CHECK(receiver.expression_category() ==
+                   ExpressionCategory::Reference);
+      CARBON_CHECK(receiver.address().has_value());
+      v = ExpressionResult::Value(
+          arena_->New<LocationValue>(receiver.address().value()));
+    }
```
lnihlen, line 1026: Note about this stanza - I could also refactor `PatternMatch` to make a `LocationValue` from a provided ExpressionResult if it is a `Reference` category. But that seemed a bigger change so I've kept this check here.

explorer/interpreter/interpreter.cpp
lnihlen: Done, PTAL

geoffromer (approved):

PR #3119, opened 2023-08-18
Adjust handling of cases where conditions are missing.

jonmeow:
In #3064, code was changed to look at a future token. This is an issue because the parser is set up to enforce that tokens aren't used without being consumed. That's part of #3118; related validation fails. Also, since it's not necessarily the open paren that was consumed, it could be a different opening symbol, which the closing symbol handling doesn't check.

Under this approach, it's tracked whether an open paren was consumed, and the open paren is associated with the state. That's more aligned with how the parser expects to be fed information.

In paren condition handling for if and while, I'm also adding some special casing for `if {` in particular to not assume the `{` is a struct. I just think that this will come up somewhat often and the resulting output is better this way (an error either way).  I'm not doing similar with `for` because there's already some `var` handling there, and I'd need a little more time to think about structure -- whereas right now I'm just trying to fix the crashes (`if {}`, `if []`, etc).

Fixes #3118

Issue #3118, opened 2023-08-18
Fuzzer: Check failure in toolchain parser "Tried to skip backwards"


josh11b:
### Description of the bug:

Input triggers Check failure in `toolchain/parser/parser_context.cpp`

### What did you do, or what's a simple way to reproduce the bug?

This input triggers the bug:
```
fn{if[]}
```

To reproduce:
```
echo 'fn{if[]}' > /tmp/testcase
bazel run -c dbg //toolchain/driver:carbon -- compile /tmp/testcase
```

### What did you expect to happen?

Diagnostic error report

### What actually happened?

Check failure on line https://github.com/carbon-language/carbon-lang/blob/trunk/toolchain/parser/parser_context.cpp#L232 :

> CHECK failure at toolchain/parser/parser_context.cpp:232: t >= *position_: Tried to skip backwards from 5 to 4

### Any other information, logs, or outputs that you want to share?

Here is the actual output:

```
Stack dump:
0.      Program arguments: /usr/local/google/home/joshl/.cache/bazel/_bazel_joshl/3df0398fc47b656dd56c4ce9f07c7299/execroot/carbon/bazel-out/k8-dbg/bin/toolchain/driver/carbon compile /tmp/testcase-6561792656474112
1.      Parser stack:
        0.      DeclarationScopeLoop @ 1:1: token 0 : Fn
        1.      FunctionDefinitionFinish @ 1:1: token 0 : Fn
        2.      StatementScopeLoop @ 1:4: token 2 : If
        3.      StatementIfConditionFinish @ 1:4: token 2 : If
        cursor  position_ @ 1:8: token 5 : CloseCurlyBrace
 #0 0x00005589895e378d llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) /proc/self/cwd/external/llvm-project/llvm/lib/Support/Unix/Signals.inc:602:11
 #1 0x00005589895e3c7b PrintStackTraceSignalHandler(void*) /proc/self/cwd/external/llvm-project/llvm/lib/Support/Unix/Signals.inc:675:1
 #2 0x00005589895e14b6 llvm::sys::RunSignalHandlers() /proc/self/cwd/external/llvm-project/llvm/lib/Support/Signals.cpp:104:5
 #3 0x00005589895e4975 SignalHandler(int) /proc/self/cwd/external/llvm-project/llvm/lib/Support/Unix/Signals.inc:413:1
[...]

PR #3120, opened 2023-08-18
Print token array when CHECK references indices

geoffromer:


jonmeow:
This is going to print out the entire token buffer, right? We're likely to see files with thousands of tokens as being common; won't that make it hard to find the error message?

What do you think of dumping tokens as a way to investigate when helpful? With chandlerc's changes too, that could look something like `compile --phase=check --dump-tokens`

geoffromer:
Yes, this is going to print the entire token buffer, albeit only on this particular `CHECK` failure. And yes, I suppose you could say this is "dumping tokens as a way to investigate when helpful", but AFAICT the token dump will essentially _always_ be helpful in the context of this `CHECK`: without the token dump, the token indices in the error message are meaningless.

I suppose an alternative would be to add something like "Run `carbon compile --phase=check --dump-tokens` to see the array those indices refer to." to the error message, but it really seems better for the error message to capture all the relevant information without extra steps. Debugging often requires doing a lot of iteration while maintaining a lot of mental state, so it seems highly desirable to remove steps from that loop where possible.

jonmeow:
> Yes, this is going to print the entire token buffer, albeit only on this particular `CHECK` failure. And yes, I suppose you could say this is "dumping tokens as a way to investigate when helpful", but AFAICT the token dump will essentially _always_ be helpful in the context of this `CHECK`: without the token dump, the token indices in the error message are meaningless.

The idea that the tokens will always be helpful does not match my experience; I did not use the token dump in my investigation of #3118 or #3119. In general, I think we should also be really careful before printing lots of information, because I'm also concerned about discouraging end users from copy-pasting messages into bug reports; but I want to emphasize that the stack trace is actually the most useful detail for me.

---

First, regarding the utility of token dumps, maybe it'd be useful to print the two tokens in question, but I don't think that analyzing a full dump is generally going to be helpful -- I'd suggest starting with the stack trace, and only looking at tokens if more information is helpful. I view the message as suggesting the issue to investigate (an N+1 token lookahead, versus some other form of token lookahead) rather than saying to look at the list of tokens.

Let me to describe the diagnostic process I used for #3119, as it resolved the related crash:

- For my initial bug fixing, I was investigating a timeout. To diagnose, I ran parsing with -v. This allowed me to see the loop of parser states. From there, I noticed the `*(TokenizedBuffer::TokenIterator(state.token) + 1)` operation on tokens. Using future tokens indicates that tokens might not be getting consumed (otherwise the token should have been provided), so I identified that as a bug likely to result in an infinite loop.

- For #3118, I find it easiest to identify the issue from the stack trace, particularly in the parser:

    ```
    #10 0x00005589890ae4a8 Carbon::ParserContext::SkipTo(Carbon::TokenizedBuffer::Token) /proc/self/cwd/toolchain/parser/parser_context.cpp:232:3
    #11 0x00005589890ae19b Carbon::ParserContext::ConsumeAndAddCloseSymbol(Carbon::TokenizedBuffer::Token, Carbon::ParserContext::StateStackEntry, Carbon::ParseNodeKind) /proc/self/cwd/toolchain/parser/parser_context.cpp:109:5
    #12 0x00005589890c1096 Carbon::ParserHandleParenConditionFinishAsIf(Carbon::ParserContext&) /proc/self/cwd/toolchain/parser/parser_handle_paren_condition.cpp:38:1
    #13 0x00005589890a39ff Carbon::ParseTree::Parse(Carbon::TokenizedBuffer&, Carbon::DiagnosticConsumer&, llvm::raw_ostream*) /proc/self/cwd/./toolchain/parser/parser_state.def:480:1
    ```

    Looking at ParserHandleParenConditionFinishAsIf will immediately show the `*(TokenizedBuffer::TokenIterator(state.token) + 1)` which is at the root of the issue, and is how I identified it was likely the same issue.

---

> I suppose an alternative would be to add something like "Run `carbon compile --phase=check --dump-tokens` to see the array those indices refer to." to the error message, but it really seems better for the error message to capture all the relevant information without extra steps. Debugging often requires doing a lot of iteration while maintaining a lot of mental state, so it seems highly desirable to remove steps from that loop where possible.

Second, regarding printing high volumes of information in CHECK messages. We structure error messages as:

1. Bug reporting information.
2. Contextual information, like the parser stack during parsing.
3. The stack dump.
4. The CHECK message.

A lot of information could be relevant to a crash, depending on the stage: the generated semantic IR, the parse tree, the tokens, the original source code, all of that for imports, etc. There are trade-offs to consider: when we print more information, it can become more difficult to look through errors.

I understand that may come as some trade-off for debugging, but I think this is the trade-off that I would prefer: trying to provide some basic information, but requiring toolchain developers reach a little if more is needed. This is because some issues will require a lot of information, and I think getting end users to even copy-paste the stack dump may provide challenging at times.

Note I'm not suggesting that we recommend adding dump instructions to the error message -- I think we can have notes on how to debug issues with Carbon that cover that, avoiding messages which may confuse end users.

zygoloid:
Perhaps we could use `PrintTokenForStackDump` to print the destination token we're skipping back to -- that would indicate the location of the destination token, without printing out the token buffer or other large quantities of information.

PR #3121, opened 2023-08-21
Introduce a full lexer benchmark.

chandlerc:
Currently this is focused on benchmarking the identifier and token lexing paths, but should expand in the future to cover other parts of the lexer.

In order to effectively benchmark tokens, this adds support to the `token_kind` library to produce a list of all the tokens in Carbon that the benchmark can use to create random inputs.

For identifiers, the benchmark has support for benchmarking different distributions of identifier sizes so it is easy to zoom into the performance specifically of short or long identifiers.

This benchmarking is motivated by profiling overall toolchain performance and noticing that an unreasonable amount of time is spent in the lexer. In turn, the identifier lexing was surprisingly hot. I have performance improvements in the works following this, but wanted to separately introduce the benchmarking framework as the review focus will be completely different.


toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -0,0 +1,136 @@
...
+    return TokenizedBuffer::Lex(source_, consumer);
+  }
+
+  auto DiagnoseErrors() -> llvm::StringRef {
+    auto buffer = TokenizedBuffer::Lex(source_, ConsoleDiagnosticConsumer());
+    CARBON_CHECK(buffer.has_errors())
+        << "Asked to diagnose errors but none found!";
+    // Return an empty string so this can be streamed into a check failure.
+    return "";
+  }
```
zygoloid: I wonder if putting an expression in the streaming portion of a `CHECK`, just for its side-effects, is a bit too subtle? Could we do something like this instead:
```suggestion
  auto DiagnoseErrors() -> std::string {
    std::string result;
    llvm::raw_string_ostream out(result);
    auto buffer = TokenizedBuffer::Lex(source_, StreamDiagnosticConsumer(out));
    CARBON_CHECK(buffer.has_errors())
        << "Asked to diagnose errors but none found!";
    return result;
  }
```
... or do we want the output to be written directly rather than passed through the `CHECK` infrastructure?

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -0,0 +1,136 @@
...
+ private:
+  llvm::vfs::InMemoryFileSystem fs_;
+  std::string filename_ = "test.carbon";
+  SourceBuffer source_;
+
+  auto MakeSourceBuffer(llvm::StringRef text) -> SourceBuffer {
+    CARBON_CHECK(fs_.addFile(filename_, /*ModificationTime=*/0,
+                             llvm::MemoryBuffer::getMemBuffer(text)));
+    return std::move(*SourceBuffer::CreateFromFile(fs_, filename_));
+  }
```
jonmeow: C++ style, declaration order -- members should be declared after functions.

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -0,0 +1,136 @@
...
+};
+
+constexpr int NumTokens = 100000;
```
jonmeow, line 52: What's the source of this constant? Is this being chosen based on typical values for source files, or something else? Trying to remove noise?

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -0,0 +1,136 @@
...
+// clang-format on
+
+void BM_ValidIdentifiers(benchmark::State& state) {
+  int min_length = state.range(0);
+  int max_length = state.range(1);
+  absl::BitGen gen;
+  std::string source;
+  llvm::raw_string_ostream os(source);
+  llvm::ListSeparator sep(" ");
+  for (int i = 0; i < NumTokens; ++i) {
```
jonmeow: Noting llvm::seq on line 102 and the loop on 57, do you have a rule of thumb for when to use one versus the other?

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+void BM_ValidKeywords(benchmark::State& state) {
+  absl::BitGen gen;
+  std::string source;
+  llvm::raw_string_ostream os(source);
+  llvm::ListSeparator sep(" ");
+  for (int i = 0; i < NumTokens; ++i) {
+    int token = absl::Uniform<int>(gen, 0, TokenKind::KeywordTokens.size());
+    os << sep << TokenKind::KeywordTokens[token].fixed_spelling();
+  }
```
zygoloid: I think this approach is going to add a bit of noise to your measurements.

By the Central Limit Theorem, the string we generate here will have a standard deviation √n times larger than that of the keyword tokens. Our current keyword list has a mean length of 5.356 with a standard deviation of 2.097, so the length of `source` should have a mean length of 635592 and a standard deviation of 663. I think that gives a ~5% chance of two runs of this test having input lengths (and presumably runtimes) that differ by 0.3% or more.

Perhaps instead we can generate N copies of the complete token list and then shuffle them? That way we'll be feeding the same amount of data into the benchmark each time we run it.

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+void BM_ValidIdentifiers(benchmark::State& state) {
+  int min_length = state.range(0);
+  int max_length = state.range(1);
+  absl::BitGen gen;
+  std::string source;
+  llvm::raw_string_ostream os(source);
+  llvm::ListSeparator sep(" ");
+  for (int i = 0; i < NumTokens; ++i) {
+    int length = absl::Uniform<int>(gen, min_length, max_length);
```
zygoloid: Same concern applies here: some runs of this might get lucky and generate more short identifiers and other runs might get unlucky and generate more long identifiers. We could maybe mitigate that by forming a fixed list of desired lengths (ideally matching the real-world length distribution for identifiers) and shuffling it.

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -0,0 +1,136 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <benchmark/benchmark.h>
+
+#include <forward_list>
```
jonmeow: Unused header?

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -0,0 +1,136 @@
...
+        continue;
+      }
+      if (id.consume_front("i") || id.consume_front("u") ||
+          id.consume_front("f")) {
+        if (llvm::all_of(id, [](const char c) { return llvm::isDigit(c); })) {
+          continue;
+        }
+      }
+      break;
+    }
```
jonmeow:
```suggestion
    llvm::StringRef id;
    do {
      // Erase any prior attempts to find an identifier.
      source.resize(id_start);
      for (int j : llvm::seq<int>(0, length)) {
        os << IdentifierChars[absl::Uniform<int>(gen, j == 0 ? 10 : 0,
                                                 sizeof(IdentifierChars))];
      }
      // Check if we ended up forming an integer type literal or a keyword, and
      // try again.
      id = llvm::StringRef(source).substr(id_start);
    } while (
      llvm::any_of(TokenKind::KeywordTokens, [id](auto token) {
            return id == token.fixed_spelling();
          }) ||
      (id.consume_front("i") || id.consume_front("u") ||
          id.consume_front("f")) &&
        llvm::all_of(id, [](const char c) { return llvm::isDigit(c); })));
```

This was taking me a bit to understand the continue/break flow. If I've understood correctly, WDYT of a do/while? Personally, I find this easier to read since there's just one condition.

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -0,0 +1,136 @@
...
+  LexerBenchHelper helper(source);
+  for (auto _ : state) {
+    TokenizedBuffer buffer = helper.Lex();
+    CARBON_CHECK(!buffer.has_errors()) << helper.DiagnoseErrors();
+  }
+
+  state.counters["TokenRate"] = benchmark::Counter(
+      NumTokens, benchmark::Counter::kIsIterationInvariantRate);
+}
+BENCHMARK(BM_ValidIdentifiers)->Args({3, 5})->Args({3, 16})->Args({12, 64});
```
jonmeow: Can you comment with any background on why these ranges? (maybe also noting that these are ranges of identifier sizes)

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+// clang-format off
+constexpr char IdentifierChars[] = {
+    // Digits:
+    '0', '1', '2', '3', '4', '5', '6', '7', '8', '9',
+    // Upper case:
+    'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O',
+    'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z',
+    // Lower case:
+    'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'M', 'n', 'o',
```
jonmeow:
```suggestion
    'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o',
```

Might be worth generating by code to reduce visual validation?

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -0,0 +1,136 @@
...
+  llvm::ListSeparator sep(" ");
+  for (int i = 0; i < NumTokens; ++i) {
+    int token = absl::Uniform<int>(gen, 0, TokenKind::KeywordTokens.size());
+    os << sep << TokenKind::KeywordTokens[token].fixed_spelling();
+  }
+
+  LexerBenchHelper helper(source);
+  for (auto _ : state) {
+    TokenizedBuffer buffer = helper.Lex();
+    benchmark::DoNotOptimize(buffer.has_errors());
```
jonmeow: Sorry, I'm trying to figure this out, code comments appreciated.

Why is has_errors called, and why the explicit DoNotOptimize? The other path does `CARBON_CHECK(!buffer.has_errors()) << helper.DiagnoseErrors();`, was there an issue with that here? (the check also ensures the parsing isn't faster for some error-related reason)

Alternately, if you only call `helper.Lex()`, was that an issue?

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -0,0 +1,136 @@
...
+  llvm::ListSeparator sep(" ");
+  for (int i = 0; i < NumTokens; ++i) {
+    int length = absl::Uniform<int>(gen, min_length, max_length);
+    os << sep;
+    int id_start = source.size();
+    for (;;) {
+      // Erase any prior attempts to find an identifier.
+      source.resize(id_start);
+      for (int j : llvm::seq<int>(0, length)) {
+        os << IdentifierChars[absl::Uniform<int>(gen, j == 0 ? 10 : 0,
```
jonmeow: Can you please document what 10 is here?

Also, as an alternative to documenting the constant, have you considered breaking out start chars to their own list? i.e.:

```
os << IdentifierStartChars[absl::Uniform<int>(gen, 0, sizeof(IdentifierStartChars))];
for (int j : llvm::seq<int>(1, length)) {
  os << IdentifierChars[absl::Uniform<int>(gen, 0, sizeof(IdentifierChars))];
}
```

toolchain/lexer/token_kind.h
```diff
@@ -74,12 +78,22 @@ class TokenKind : public CARBON_ENUM_BASE(TokenKind) {
...
 #define CARBON_TOKEN(TokenName) \
   CARBON_ENUM_CONSTANT_DEFINITION(TokenKind, TokenName)
 #include "toolchain/lexer/token_kind.def"
 
+constexpr TokenKind TokenKind::KeywordTokensStorage[] = {
+#define CARBON_KEYWORD_TOKEN(TokenName, Spelling) TokenKind::TokenName,
+#include "toolchain/lexer/token_kind.def"
+};
+constexpr llvm::ArrayRef<TokenKind> TokenKind::KeywordTokens =
+    KeywordTokensStorage;
```
jonmeow, line 95: Why the indirection of storage?

toolchain/lexer/tokenized_buffer_benchmark.cpp
jonmeow: I was looking at this, and had a related suggestion... The diagnostic consumer should only get called when has_errors is true, maybe you can pass ConsoleDiagnosticConsumer instead of NullDiagnosticConsumer and only check has_errors?

jonmeow (commented):
zygoloid commented faster than me, I was trying to figure out what this was doing while he commented. :)

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Stale from an earlier iteration, removed.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: To Jon's point -- that would result with basically the code I started with.

But I do think Richard's suggestion makes sense, so let's try that?

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Doh, right, thanks.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Applied the fix.

Note that it is *very* hard to make a `constexpr` variable that is generated by code. I tried, hard, and after a bunch of effort gave up.

I guess I'll change this to not be `constexpr`, and instead be a lazily computed global with a bunch of heap allocations behind it. Not seeing better options here, which is why I initially went with just a `constexpr` array.

Done, but made all the code a bit longer.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Using `llvm::seq` here resulted in an unused variable that I had to cast to void.

My rough rule of thumb is -- whichever makes the code shorter.

I'm happy to switch to some other heuristic for when to write which form, no strong opinion here.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: I had considered this, but the chained conditions seemed harder to read than the for loop.

That said, i don't feel strongly, so happy to switch to this (and suggested edit accepted).

toolchain/lexer/token_kind.h
chandlerc: The `ArrayRef` API is much nicer than the C-array API -- things like `.size()` and such. But it doesn't work without durable storage, so I did a two-stage wrapping.

I tried a bunch of other patterns like a `std::array` but couldn't get any of them to be a better fit and compile to `constexpr` defininitions other than this pattern.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Wanted it as large as I could for measurement stability without making benchmarking unreasonably slow. There isn't a real motivation other than trying out values while profiling things though, not sure how to write a coment for it.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: FWIW, that level of noise isn't a big deal for the changes I've been making.

Not sure about all of the math, I can't really follow those steps, but I'll take your conclusion.

Changing this to shuffling things I think wouldn't be a great solution. Fundamentally, Abseil's RNG doesn't support seeding to get a fixed PRNG behavior. And given that, we'd just be trading call-for-call noise for process run-for-run noise which seems a strictly worse situation. I'd rather be able to use things like `--benchmark_repititions` to understand the noise than have things swing from one execution to another.

That said, I'm happy to iterate on better ways to construct pseudo-random sequences for this and other benchmarks. But maybe that can be a follow-up?

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Sorry, at one point I was getting failures from `CARBON_CHECK` here and removed it, but they were surrious. Switched back to `CARBON_CHECK`.

A pattern that I find emerges in benchmark code is ensuring that the work done has some consequence so the optimizer doesn't remove it entirely or remove it from the loop breaking the benchmark.

One way of doing that is something like `CARBON_CHECK` that we expect to be very fast when it succeeds and only slow when it fails, but to have a real consequence that keeps code around.

But when I thought that wouldn't work (I was remembering when we lexed keywords that were then rejected; an old state of the lexer), I used `DoNotOptimize` to achieve a similar result for benchmarking without producing an error.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Same comment as above -- maybe a follow-up patch that works on addressing this? I think it is somewhat tricky as I don't want to regress to execution-level noise instead of noise the framework can measure and account for easily.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Either way. I picked this formulation because it was shorter, but I'll switch to the one you suggest if you like.

(Also happy to switch back if you like, I don't have strong opinions about this code, mostly want it in.)

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: I don't have any real details here -- I picked some ranges that seemed good to me because I needed to.

I can write that in a comment, but it didn't seem helpful. Sorry, I really want to try to address this, but I'm not sure what you'd like in the comment here. Happy to chat and figure out what comment would help.

chandlerc (commented):
Hopefully feedback addressed....

toolchain/lexer/tokenized_buffer_benchmark.cpp
jonmeow: Maybe:

```suggestion
// A large value for measurement stability without making benchmarking too slow.
constexpr int NumTokens = 100000;
```

toolchain/lexer/tokenized_buffer_benchmark.cpp
jonmeow: Note, I'd be happy with a habit of "prefer llvm::seq because it's easier to read", and just make that a de facto standard. I think if we use "whichever makes the code shorter" I think the old-style for loop is always going to win by a couple characters (note my suggestion on the 'A' - 'Z' loop is a couple chars shorter, although that's actually a correctness suggestion)

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -0,0 +1,148 @@
...
+auto IdentifierStartChars() -> llvm::ArrayRef<char> {
+  static llvm::SmallVector<char> chars = [] {
+    llvm::SmallVector<char> chars;
+    chars.push_back('_');
+    for (char c : llvm::seq('A', 'Z')) {
+      chars.push_back(c);
+    }
+    for (char c : llvm::seq('a', 'z')) {
+      chars.push_back(c);
+    }
```
jonmeow, line 83:
```suggestion
    for (char c = 'A'; c <= 'Z'; ++c) {
      chars.push_back(c);
    }
    for (char c = 'a'; c <= 'z'; ++c) {
      chars.push_back(c);
    }
```

llvm::seq won't include the end of the range, so you're dropping Z right now.

toolchain/lexer/tokenized_buffer_benchmark.cpp
jonmeow: I'm good with that!

toolchain/lexer/tokenized_buffer_benchmark.cpp
jonmeow: Note, I think even a comment of "// Ranges trying to test a few different potential cases." would be helpful, just so that if someone gets something they want to test more specifically in the future, it's clearer what the standard is to change.

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -0,0 +1,148 @@
...
+  state.counters["TokenRate"] = benchmark::Counter(
+      NumTokens, benchmark::Counter::kIsIterationInvariantRate);
+}
+BENCHMARK(BM_ValidKeywords);
+
+auto IdentifierStartChars() -> llvm::ArrayRef<char> {
+  static llvm::SmallVector<char> chars = [] {
+    llvm::SmallVector<char> chars;
+    chars.push_back('_');
+    for (char c : llvm::seq('A', 'Z')) {
```
jonmeow:
```suggestion
    for (char c = 'A'; c <= 'Z'; ++c) {
```

Note that `llvm::seq` omits the end, so Z and 9 are omitted in the current implementation. Probably a mistake?

Framing this as a suggestion mainly so that the difference in length is easy to see.

jonmeow: Sidenote -- reading docs, `seq_inclusive` would also work.

jonmeow (approved):
To note for my part, I'm fine with remaining comments being resolved and merging, I don't feel like I'll need another look. Might want to confirm zygoloid is good with that plan regarding generation of identifiers.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Sure, happy to just switch to `seq` everywhere I can.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Done.

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -0,0 +1,158 @@
...
+void BM_ValidIdentifiers(benchmark::State& state) {
+  int min_length = state.range(0);
+  int max_length = state.range(1);
+  absl::BitGen gen;
+  std::string source;
+  llvm::raw_string_ostream os(source);
+  llvm::ListSeparator sep(" ");
+  for (int i : llvm::seq(0, NumTokens)) {
+    static_cast<void>(i);
+    int length = absl::Uniform<int>(gen, min_length, max_length);
```
zygoloid: Should this be:
```suggestion
    int length = absl::Uniform<int>(absl::IntervalClosedClosed, gen, min_length, max_length);
```
... to include the `length == max_length` case, or did you intend for a semi-open interval here? The naming (`max_length`) to me suggests a closed interval.

toolchain/lexer/tokenized_buffer_benchmark.cpp
zygoloid: Sure, happy to leave this for a possible follow-up.

toolchain/lexer/tokenized_buffer_benchmark.cpp
zygoloid: Let's leave this for a follow-up. I didn't understand what you meant about process-run to process-run noise; maybe we should discuss that off-line, I think we might be talking past each other here :)

zygoloid (approved):
Happy for this to land either as-is or with some change for the `max_length` handling.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: I was happy with semi-open, but I also don't feel strongly so happy to make it closed.

PR #3122, opened 2023-08-21
Introduce SIMD optimized scanning for identifiers. 

chandlerc:
Currently, for long identifiers, a huge (>30%) fraction of time is spent
finding the end of the identifier. We can speed this up with a fun
application of SIMD and in-register lookup tables.

With this, the BM_ValidIdentifiers/12/64 benchmark goes from around 4 million
tokens/second to around 6 mt/s, so roughly 1.5x improvement. However,
there was a decent amount of noise in the measurement and I didn't study
it too closely as I was very happy with the overall result. The profile
shifted from >30% of the time in this loop to <10% of the time, so the
scan itself is 3x or more faster with this.

One concern with optimizing the lexer right now is that we don't have
full Unicode support from the design. This PR takes some steps to at least
try and avoid this pitfall -- the new routine works to classify UTF-8
code units, and has a fallback in that case that can grow the needed
logic.

### DIFF
--- a/toolchain/lexer/tokenized_buffer.cpp
+++ b/toolchain/lexer/tokenized_buffer.cpp
@@ -24,6 +24,10 @@
 #include "toolchain/lexer/numeric_literal.h"
 #include "toolchain/lexer/string_literal.h"
 
+#if __x86_64__
+#include <x86intrin.h>
+#endif
+
 namespace Carbon {
 
 // TODO: Move Overload and VariantMatch somewhere more central.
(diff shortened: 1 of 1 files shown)


toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -48,6 +52,164 @@ auto VariantMatch(V&& v, Fs&&... fs) -> decltype(auto) {
...
+      /*0x6:*/ 0b1011'1110,
+      /*0x7:*/ 0b1011'1110,
+      /*0x8:*/ 0b1011'1110,
+      /*0x9:*/ 0b1011'1110,
+      /*0xA:*/ 0b1011'1100,
+      /*0xB:*/ 0b1001'0100,
+      /*0xC:*/ 0b1001'0100,
+      /*0xD:*/ 0b1001'0100,
+      /*0xE:*/ 0b1001'0100,
+      /*0xF:*/ 0b1001'0101);
```
jonmeow: What do you think of switching to line comments for style? We'd been going that way in C++ because Carbon wouldn't allow `/* */` comments like this.

toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -48,6 +52,164 @@ auto VariantMatch(V&& v, Fs&&... fs) -> decltype(auto) {
...
+  //
+  // The goal is to get a bit mask classifying different sets of bytes. For each
+  // input byte, we want the mask bits to be set with the following logic
+  // derived by inspecting the high nibble and low nibble of the input:
+  // bit0 = 1 for `_` needs 1 for high `0x5` and low `0xF`
+  // bit1 = 1 for `0-9` needs 1 for high `0x3` and low `0x0` - `0x9`
+  // bit2 = 1 for `A-O` needs 1 for high `0x4` and low `0x1` - `0xF`
+  // bit3 = 1 for `P-Z` needs 1 for high `0x5` and low `0x0` - `0xA`
+  // bit4 = 1 for `a-o` needs 1 for high `0x6` and low `0x1` - `0xF`
+  // bit5 = 1 for `p-z` needs 1 for high `0x7` and low `0x0` - `0xA`
```
zygoloid:
```suggestion
  // bit0 = 1 for `_` needs high `0x5` and low `0xF`
  // bit1 = 1 for `0-9` needs high `0x3` and low `0x0` - `0x9`
  // bit2 = 1 for `A-O` needs high `0x4` and low `0x1` - `0xF`
  // bit3 = 1 for `P-Z` needs high `0x5` and low `0x0` - `0xA`
  // bit4 = 1 for `a-o` needs high `0x6` and low `0x1` - `0xF`
  // bit5 = 1 for `p-z` needs high `0x7` and low `0x0` - `0xA`
```
If I'm understanding these comments correctly, the `= 1 for` and `needs 1 for` parts are saying the same thing?

toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -48,6 +52,164 @@ auto VariantMatch(V&& v, Fs&&... fs) -> decltype(auto) {
...
+  // bit6 = unused
+  // bit7 = 1 for possible non-ASCII character in UTF-8, for all high nibbles
+  // with the high bit set and all low-nibbles.
+  //
+  // No bits set means definitively non-ID ASCII character.
+  //
+  // High bit set means contains the start of a non-ASCII character, need to
+  // fall back to a UTF-8 decoding path (not yet fully implemented).
+  //
+  // bit6 remains unused if we need to classify more characters.
```
zygoloid: Incidentally, we can save two more bits if we find a need, by combining `bit2` with `bit4` and `bit3` with `bit5`:

```
bit2 = 1 for `A-Oa-o` needs high `0x4` or `0x6` and low `0x1` - `0xF`.
bit3 = 1 for `P-Zp-z` needs high `0x5` or `0x7` and low `0x0` - `0xA`.
bit4 = unused
bit5 = unused
```

toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -48,6 +52,164 @@ auto VariantMatch(V&& v, Fs&&... fs) -> decltype(auto) {
...
+  int tail_utf8_mask;
+  int tail_ascii_mask;
+  int tail_mask;
+  // Use `ssize_t` for performance here as we index memory in a tight loop.
+  ssize_t i = 0;
+  const ssize_t size = text.size();
+  while ((i + 16) <= size) {
+    __m128i input =
+        _mm_loadu_si128(reinterpret_cast<const __m128i*>(text.data() + i));
+    __m128i low_mask = _mm_shuffle_epi8(low_lut, input);
```
zygoloid, line 172: I think this doesn't quite work for the UTF-8 case: `_mm_shuffle_epi8` always gives `0` for a lane where the corresponding byte of `input` is negative, so we'll produce a `low_mask` of `0` for all non-ASCII bytes.

Fortunately, I think we can address this without extra masking here, by changing the computation of `tail_utf8_mask` below:

```diff
-    tail_utf8_mask = _mm_movemask_epi8(mask);
+    tail_utf8_mask = _mm_movemask_epi8(high_mask);
```

The point being, we don't need to look at the low-order nibble at all to classify UTF-8, so it doesn't matter that this won't compute it properly due to the behavior of `pshufb` for negative bytes.

toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -48,6 +52,164 @@ auto VariantMatch(V&& v, Fs&&... fs) -> decltype(auto) {
...
+    __m128i input_high =
+        _mm_and_si128(_mm_srli_epi32(input, 4), _mm_set1_epi8(0x0f));
```
zygoloid, line 176: I think it'd be useful to comment why we need the mask with `0x0f0f0f0f...` here and not in the other case.

toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -48,6 +52,164 @@ auto VariantMatch(V&& v, Fs&&... fs) -> decltype(auto) {
...
+    // mask on more recent CPUs, but we generally expect the median length we
+    // encounter to be <16 characters and so we avoid the extra instruction in
+    // that case and predict this branch to succeed so it is laid out in a
+    // reasonable way.
+    if (LLVM_LIKELY(tail_mask != 0)) {
+      break;
+    }
+    i += 16;
+  }
+  if (LLVM_LIKELY((i + 16) <= size)) {
```
zygoloid: Rather than repeating the branch condition, I think it might be a bit clearer to move the `__builtin_ctz` above the `break` a few lines above.

zygoloid (commented):
This is neat!

For UTF-8 handling, we're going to need to deal with things like [Unicode whitespace characters](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/lexical_conventions/whitespace.md), so we can't get away with classifying everything non-ASCII as continuing an identifier, and then (presumably) doing some later checking -- we'll need to actually decode the characters during tokenization.

As a result, I wonder if maybe the best thing to do would be to fall out of the fast vectorized identifier loop whenever we see any non-ASCII character, and have the slower scalar loop also deal with UTF-8 characters (and ideally jump back into vectorized code if an identifier contains a mixture of ASCII and non-ASCII characters, but that's a rare enough case that not optimizing for it is probably fine). I think by doing that we could simplify the vectorized loop slightly, and maybe remove one of the `_mm_movemask` calls.

jonmeow (commented):
I'm happy to let zygoloid handle review here since it seems he understands the instructions better, but I did have one comment thing I'd noticed.

toolchain/lexer/tokenized_buffer.cpp
chandlerc: We can do better I think and grab the mask directly off the input as it just extracts the high bits. But good call, and I've updated comments to document this.

toolchain/lexer/tokenized_buffer.cpp
chandlerc: These are already long and hard to read, and I think line comments would make that even worse.

What I'd prefer is something more like a designated initializer. I think we'll have good options for this in Carbon, but in C++ it didn't seem worth trying to find one, and instead to just use a comment.

toolchain/lexer/tokenized_buffer.cpp
chandlerc: Doh, of course. Good idea, done!

toolchain/lexer/tokenized_buffer.cpp
chandlerc: Done.

toolchain/lexer/tokenized_buffer.cpp
chandlerc: Sure, good idea. and with a much simpler duplication, the fallback path is much cleaner I think, and we now have a single place to introduce unicode handling eventually.

chandlerc (commented):
PTAL!

zygoloid (approved):
Feels like we might be able to do ... something ... better now we're only using 4 bits of mask, but I don't see it. (We could fold the two lookup tables into one, which would save a register but add a shift and mask into the loop, which I suspect isn't a win.)

toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -48,6 +52,167 @@ auto VariantMatch(V&& v, Fs&&... fs) -> decltype(auto) {
...
+      /*0x6:*/ 0b0000'1110,
+      /*0x7:*/ 0b0000'1110,
+      /*0x8:*/ 0b0000'1110,
+      /*0x9:*/ 0b0000'1110,
+      /*0xA:*/ 0b0000'1100,
+      /*0xB:*/ 0b0000'0100,
+      /*0xC:*/ 0b0000'0100,
+      /*0xD:*/ 0b0000'0100,
+      /*0xE:*/ 0b0000'0100,
+      /*0xF:*/ 0b0000'0101);
```
zygoloid, line 146: I looked into whether we could use a similar approach as for `IsIdentifierByteTable` here, to compute this table rather than hard-coding it:

```suggestion
  constexpr auto luts = []() constexpr {
    std::array<std::array<std::uint8_t, 16>, 2> luts = {};
    auto AddRange = [&](int bit, unsigned char min, unsigned char max) {
      CARBON_CHECK((min >> 4) == (max >> 4));
      luts[1][min >> 4] |= 1 << bit;
      for (unsigned char c = min; c <= max; ++c) {
        luts[0][c & 0xF] |= 1 << bit;
      }
    };
    // This code uses a scheme derived from the techniques in Geoff Langdale
    // and Daniel Lemire's work on parsing JSON[1]. Specifically, that paper
    // outlines a technique of using two 4-bit indexed in-register look-up
    // tables (LUTs) to classify bytes in a branchless SIMD code sequence.
    //
    // [1]: https://arxiv.org/pdf/1902.08318.pdf
    //
    // The goal is to get a bit mask classifying different sets of bytes.
    // For each input byte, we first test for a high bit indicating a UTF-8
    // encoded Unicode character. Otherwise, we want the mask bits to be set
    // with the following logic derived by inspecting the high nibble and
    // low nibble of the input:

    // bit0 = 1 for `_`: high `0x5` and low `0xF`
    AddRange(0, '_', '_');

    // bit1 = 1 for `0-9`: high `0x3` and low `0x0` - `0x9`
    AddRange(1, '0', '9');

    // bit2 = 1 for `A-O` and `a-o`: high `0x4` or `0x6` and low `0x1` - `0xF`
    AddRange(2, 'A', 'O');
    AddRange(2, 'a', 'o');

    // bit3 = 1 for `P-Z` and 'p-z': high `0x5` or `0x7` and low `0x0` - `0xA`
    AddRange(3, 'P', 'Z');
    AddRange(3, 'p', 'z');

    // bit4 = unused
    // bit5 = unused
    // bit6 = unused
    // bit7 = unused
    //
    // No bits set means definitively non-ID ASCII character.
    //
    // bits 4-7 remain unused if we need to classify more characters.
    return luts;
  }();
  auto MakeMask = [&]<std::size_t... I>(int which,
                                        std::index_sequence<I...>) {
    return _mm_setr_epi8(luts[which][I]...);
  };
  const auto high_lut = MakeMask(1, std::make_index_sequence<16>());
  const auto low_lut = MakeMask(0, std::make_index_sequence<16>());
```

But in the end I think it's actually a lot easier to understand as just a table of data :)

zygoloid:
Not merging: the initial comment is probably not quite the commit message we want. (At least the reference to #3121 should be removed.) But this looks good to merge to me.

toolchain/lexer/tokenized_buffer.cpp
chandlerc: Yeah, the table is really helpful for visualizing the result of all of this.

toolchain/lexer/tokenized_buffer.cpp
jonmeow: Maybe instead put this in that kind of form, then? It looks like parameters are named b0 through b15:

```
_mm_setr_epi8(char __b0, char __b1, char __b2, char __b3, char __b4, char __b5,
              char __b6, char __b7, char __b8, char __b9, char __b10,
              char __b11, char __b12, char __b13, char __b14, char __b15) {
```

So wouldn't style be `/*__b0=*/`, etc?

Note, if you don't want to do this, it's probably worth thinking about what it means for Carbon if a function in a library that we don't own has parameter names you don't find useful.

chandlerc: I'll try this and see if I can get the parameter name approach to work well. Even though they're not the *most* readable, this would have actually caught a bug I wrote the first time (I used `_mm_set_epi8` instead).

That said, not sure it makes sense to generalize too much to Carbon here -- I feel like there are a bunch of overlapping challenges from C++ here and we might better off just working with Carbon code that ends up with bit tables and similar constructs in its API design, and see what the best tools are or what changes would give better results there.

PR #3123, opened 2023-08-21
Expand lexing benchmarks a little bit.

chandlerc:
This adds a benchmark that tries to synthesize mixtures of symbols,
keywords, and identifiers. It tweaks the distribution of identifier
lengths based on some empirical measurements of, for example LLVM's
codebase.

Also establish a framework for skewing the symbol distribution, although
that one is based entirely on intuition and not measurements. It should
be adjusted as we have measurements.

The ratios between symbols, keywords, and identifiers is also
unmeasured, but several different ratios are covered.

Neither literals nor grouping symbols are included yet, as both present
some additional challenges in forming them, and this seemed like
a plausible increment in expanding the benchmark coverage.

### DIFF
--- a/toolchain/lexer/tokenized_buffer_benchmark.cpp
+++ b/toolchain/lexer/tokenized_buffer_benchmark.cpp
@@ -15,6 +15,117 @@
 namespace Carbon::Testing {
 namespace {
 
+auto IdentifierStartChars() -> llvm::ArrayRef<char> {
+  static llvm::SmallVector<char> chars = [] {
+    llvm::SmallVector<char> chars;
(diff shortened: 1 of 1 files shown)


chandlerc:
All the preceding PRs have merged and so I cleaned this up into a clean commit on top of trunk now, and it should be ready for review.

This still fundamentally has some noise problems, I've not yet worked on that aspect. But I think the improvements are generally useful.

jonmeow (approved):
Looks good! Might be interesting to discuss the length distribution, but I think the edit is great to merge as-is.

PR #3124, opened 2023-08-21
[wip] [explorer] Support aliasing an interface method as class member

### LINKED ISSUE #2798: Support aliasing an interface method as a class member
### Description of the bug:

This is the feature version of #2583 (now that the crash is fixed), to support aliasing an interface method as a class member.

```
package sample api;

interface Printable {
  fn PrintIt[self: Self]();
}

impl String as Printable {
  fn PrintIt[self: String]() {
    Print(self);
  }
}

class Vector(T:! type) {
  var x: T;

  alias PrintIt = Printable.PrintIt;
}

// Conditionally implement the API for certain `T`s.
impl forall [U:! Printable] Vector(U) as Printable {
  fn PrintIt[self: Self]() {
    Print("{{");
    self.x.PrintIt();
    Print("}");
  }
}

fn Main() -> i32 {
  var v: Vector(String) = {.x = "test"};
  v.(Printable.PrintIt)();
  v.PrintIt();
  return 42;
}
```

That the alias makes using the member name equivalent to using a *qualified* member name in a call. Basically, lines 30 and 31 should be equivalent:

```
  v.(Printable.PrintIt)();
  v.PrintIt();
```

### What actually happened?

Currently unsupported.

_No response_

lnihlen:
Not ready for review. Early wip on adding support. Uploaded because I have some questions.

Closes #2798

### DIFF
--- a/explorer/interpreter/type_checker.cpp
+++ b/explorer/interpreter/type_checker.cpp
@@ -2937,9 +2937,37 @@ auto TypeChecker::TypeCheckExpImpl(Nonnull<Expression*> e,
                 access.set_expression_category(ExpressionCategory::Value);
                 break;
               }
-              case DeclarationKind::AliasDeclaration:
-                return ProgramError(access.source_loc())
-                       << "Member access to aliases is not yet supported.";
+              case DeclarationKind::AliasDeclaration: {
+                const auto* alias_decl = cast<AliasDeclaration>(member);
+                const auto& target = alias_decl->target();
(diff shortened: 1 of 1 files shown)


jonmeow:
> Not ready for review. Early wip on adding support. Uploaded because I have some questions

Converting to draft, since it sounds like review wasn't intended. I think the question is at [#explorer here](https://discord.com/channels/655572317891461132/763516049710120960/1143208586059718786)

PR #3125, opened 2023-08-21
Use std::array in command_line definitions

kehrazy:
This commit replaces c-style array usage in `OneOfArgBuilder` class member function definitions.

### DIFF
--- a/common/command_line.h
+++ b/common/command_line.h
@@ -472,7 +472,7 @@ class OneOfArgBuilder : public ArgBuilder {
   // optional, and also will cause that value to be stored into the result
   // even if the argument is not parsed explicitly.
   template <typename T, typename U, size_t N>
-  void SetOneOf(const OneOfValueT<U> (&values)[N], T* result);
+  void SetOneOf(const std::array<OneOfValueT<U>, N> &values, T* result);
 
   // Configures the argument to append a parsed value to the provided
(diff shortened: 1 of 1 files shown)


zygoloid:
Unfortunately, I don't think we can do this: we're relying on the ability to deduce the array element type and the array bound from a braced initializer, and that requires a real array rather than a `std::array`. (This PR is causing a build failure for this reason.)

We could potentially switch to a `std::initializer_list<T>` parameter if we don't actually need to deduce the array bound, but I've not looked deeply into this code so I don't know if that's a possibility.

zygoloid (changes requested):
Unfortunately, I don't think we can do this: we're relying on the ability to deduce the array element type and the array bound from a braced initializer, and that requires a real array rather than a `std::array`. (This PR is causing a build failure for this reason.)

We could potentially switch to a `std::initializer_list<T>` parameter if we don't actually need to deduce the array bound, but I've not looked deeply into this code so I don't know if that's a possibility.

kehrazy:
> we're relying on the ability to deduce the array element type and the array bound from a braced initializer, and that requires a real array rather than a `std::array`

ah, quite unfortunate.

this should probably be closed until further notice - I will look into making this a `std::array` compliant array instead =)

PR #3126, opened 2023-08-21
Fix a stack use after scope and a heap use after free found by fuzzing.

zygoloid:
There were two issues contributing to this crash:

- Primarily, the issue is that we queue up diagnostics and don't format them into a string until we reach the end of compilation. In some code paths in the driver, we destroyed the Semantics IR object before this happened. But diagnostics can contain references to Semantics IR objects, such as strings stored in the string table, which can lead to a use after destruction bug.

  This is fixed by ensuring the diagnotics consumer is flushed before destroying any of the objects that it can refer to. The current approach to this is not especially clean, unfortunately, but this requires fighting C++ as this isn't the order in which it wants to destroy things.

- This issue was obscured by the Semantics IR's string table holding a reference to whatever underlying storage it was given rather than its own string storage, so sometimes it would hold a reference to a string from the source file, and sometimes a string from the tokenized buffer's string table. The diagnostics were always flushed before the source file was destroyed, but not before the tokenized buffer was destroyed. So to see the issue, you'd need to have a string literal with certain contents followed by an identifier with a name that matched those contents.

  The crash is made more reliable by holding references to the Semantics IR's string map in its string table, rather than references to someone else's strings. This also fixes a latent bug where passing a string temporary to SemanticsIR::AddString would store a dangling reference in the string table. Incidentally, AddString is also changed to perform only one hash table lookup rather than two for each added string.

### DIFF
--- a/toolchain/driver/driver.cpp
+++ b/toolchain/driver/driver.cpp
@@ -400,7 +400,7 @@ auto Driver::Compile(const CompileOptions& options) -> bool {
   CARBON_VLOG() << "*** SourceBuffer::CreateFromFile done ***\n";
   // Require flushing the consumer before the source buffer is destroyed,
   // because diagnostics may reference the buffer.
-  auto flush = llvm::make_scope_exit([&]() { consumer->Flush(); });
+  auto flush_for_source = llvm::make_scope_exit([&]() { consumer->Flush(); });
   if (!source.ok()) {
     error_stream_ << "ERROR: Unable to open input source file: "
                   << source.error();
@@ -410,6 +410,8 @@ auto Driver::Compile(const CompileOptions& options) -> bool {
 
   CARBON_VLOG() << "*** TokenizedBuffer::Lex ***\n";
   auto tokenized_source = TokenizedBuffer::Lex(*source, *consumer);
+  // Diagnostics may reference the tokenized buffer.
+  auto flush_for_tokens = llvm::make_scope_exit([&]() { consumer->Flush(); });
   bool has_errors = tokenized_source.has_errors();
   CARBON_VLOG() << "*** TokenizedBuffer::Lex done ***\n";
   if (options.dump_tokens) {
@@ -424,6 +426,8 @@ auto Driver::Compile(const CompileOptions& options) -> bool {
 
   CARBON_VLOG() << "*** ParseTree::Parse ***\n";
   auto parse_tree = ParseTree::Parse(tokenized_source, *consumer, vlog_stream_);
+  // Diagnostics may reference the parse tree.
+  auto flush_for_parse = llvm::make_scope_exit([&]() { consumer->Flush(); });
   has_errors |= parse_tree.has_errors();
   CARBON_VLOG() << "*** ParseTree::Parse done ***\n";
   if (options.dump_parse_tree) {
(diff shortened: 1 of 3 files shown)


toolchain/driver/driver.cpp
```diff
@@ -400,7 +400,7 @@ auto Driver::Compile(const CompileOptions& options) -> bool {
   CARBON_VLOG() << "*** SourceBuffer::CreateFromFile done ***\n";
   // Require flushing the consumer before the source buffer is destroyed,
   // because diagnostics may reference the buffer.
-  auto flush = llvm::make_scope_exit([&]() { consumer->Flush(); });
+  auto flush_for_source = llvm::make_scope_exit([&]() { consumer->Flush(); });
```
jonmeow, line 403: Since this doesn't work as intended, maybe we should just explicitly flush before each return?

By my count, that'd be lines 407, 424, 439, and 461 (or maybe 463 and 467 if you want to be consistent). Either way, removing the 476 flush.

That versus a scope exit on 403, 414, 430, and 447 plus the 476 explicit flush.

Just observing as I scrutinize flushes, the flushes on 419, 434, and 451 actually seem like they should maybe be removed, as a flawed translation when the prior "dump" setup was split (i.e., dumping these shouldn't force diagnostics to print early). I'd be fine changing that in its own PR though.

toolchain/semantics/semantics_ir.h
```diff
@@ -200,16 +200,17 @@ class SemanticsIR {
 
   // Adds an string, returning an ID to reference it.
   auto AddString(llvm::StringRef str) -> SemanticsStringId {
-    // If the string has already been stored, return the corresponding ID.
-    if (auto existing_id = GetStringID(str)) {
```
jonmeow, line 204: Can GetStringID be removed now? I think it might've only been called here.

jonmeow (approved):
Sorry! I think both of these are my fault.

toolchain/driver/driver.cpp
zygoloid: Sure, we can do that. I have a lingering concern that any of these approaches seems too error-prone. I think it'll be easy to forget to add a flush, and the bugs you get as a consequence are subtle. Maybe we should also make the consumer CHECK-fail from its destructor if it's not been flushed, rather than auto-flushing?

I think it makes sense to flush before producing dump output: in the common case of running an action from the command line, I think I'd want to see error / warning messages before the dump output, rather than afterwards. It is strange that enabling dump output will produce diagnostics in a different order, but I guess we can't do much about that unless we want to buffer all the dump output and print it after we finish the action.

toolchain/semantics/semantics_ir.h
zygoloid: It can! Thanks, done.

toolchain/driver/driver.cpp
zygoloid: Let's decide what to do with the flush-before-dump cases in #3128.

jonmeow: > Sure, we can do that. I have a lingering concern that any of these approaches seems too error-prone. I think it'll be easy to forget to add a flush, and the bugs you get as a consequence are subtle. Maybe we should also make the consumer CHECK-fail from its destructor if it's not been flushed, rather than auto-flushing?

SGTM

> I think it makes sense to flush before producing dump output: in the common case of running an action from the command line, I think I'd want to see error / warning messages before the dump output, rather than afterwards. It is strange that enabling dump output will produce diagnostics in a different order, but I guess we can't do much about that unless we want to buffer all the dump output and print it after we finish the action.

Okay... I mean, pragmatically I don't think the choice here matters much since end users probably won't notice.

jonmeow (approved):

jonmeow:
Approving but not merging since I'm not sure whether you wanted to make the flush change as part of this.

zygoloid:
I'll send a separate PR for the auto-flush change.

PR #3128, opened 2023-08-21
Flush diagnostics before printing semantics IR rather than after.

zygoloid:
This matches what we do for all other dump output.

No test: this is just changing the behavior of a dump mode, and is really awkward to exercise without adding back in something like a lit test to observe the behavior when stdout and stderr go to the same place.


zygoloid:
Adding @jonmeow as he had suggested we go in the opposite direction in #3126, and never flush diagnostics before producing dump output.

jonmeow (approved):
Yeah, I mean, I prefer the other direction but the consistency here makes sense.

PR #3134, opened 2023-08-22
Remove automatic flushing in the SortingDiagnosticConsumer destructor.

zygoloid:
We already needed to manually flush diagnostics along every path out of the driver, so switch to a CHECK-failure if we get this wrong rather than a potential use-after-lifetime bug.

Switch the driver to flush explicitly rather than using a bunch of cleanup lambdas, now that we have checking that we get this right.

This is a follow-up after #3126.


jonmeow (approved):
Thanks!

PR #3135, opened 2023-08-22
Make more use of llvm::zip and enumerate in explorer

jonmeow:
I was initially looking at llvm::seq, but then looking through uses it seemed like enumerate/zip would be best for these, allowing less direct indexing and more range-based looping.

### DIFF
--- a/explorer/interpreter/interpreter.cpp
+++ b/explorer/interpreter/interpreter.cpp
@@ -11,7 +11,6 @@
 #include <optional>
 #include <random>
 #include <utility>
-#include <variant>
 #include <vector>
 
 #include "common/check.h"
@@ -28,13 +27,14 @@
 #include "explorer/base/trace_stream.h"
 #include "explorer/interpreter/action.h"
 #include "explorer/interpreter/action_stack.h"
+#include "explorer/interpreter/heap.h"
 #include "explorer/interpreter/pattern_match.h"
-#include "explorer/interpreter/stack.h"
 #include "explorer/interpreter/type_utils.h"
 #include "llvm/ADT/APInt.h"
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/StringExtras.h"
 #include "llvm/Support/Casting.h"
-#include "llvm/Support/Error.h"
 #include "llvm/Support/FormatVariadic.h"
 #include "llvm/Support/raw_ostream.h"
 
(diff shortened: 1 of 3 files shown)


explorer/interpreter/interpreter.cpp
```diff
@@ -1534,17 +1530,18 @@ auto Interpreter::StepExp() -> ErrorOr<Success> {
...
-          int i = 1 + num_args;
-          for (const auto& [impl_bind, impl_exp] : call.witnesses()) {
-            witnesses[impl_bind] = act.results()[i];
-            ++i;
+          for (const auto& [witness, result] : llvm::zip(
+                   call.witnesses(),
+                   llvm::ArrayRef(act.results()).drop_front(1 + num_args))) {
+            witnesses[witness.first] = result;
           }
         }
```
zygoloid: Maybe we should consider adding a `Split` utility so we can do this:

```suggestion
        std::vector<Nonnull<const Value*>> param_values;
        auto [arg_results, witness_results] =
            Split(llvm::ArrayRef(act.results()).drop_front(), num_args);
        for (const auto& arg_result : arg_results) {
          param_values.push_back(arg_result);
        }
        const auto* param_tuple = arena_->New<TupleValue>(param_values);
        // Prepare witnesses.
        ImplWitnessMap witnesses;
        if (num_witnesses > 0) {
          for (const auto& [witness, result] : llvm::zip(call.witnesses(), witness_results)) {
            witnesses[witness.first] = result;
          }
        }
```

(No action expected in this PR though.)

explorer/interpreter/interpreter.cpp
```diff
@@ -290,10 +290,9 @@ auto Interpreter::EvalPrim(Operator op, Nonnull<const Value*> /*static_type*/,
 auto Interpreter::CreateStruct(const std::vector<FieldInitializer>& fields,
                                const std::vector<Nonnull<const Value*>>& values)
     -> Nonnull<const Value*> {
-  CARBON_CHECK(fields.size() == values.size());
   std::vector<NamedValue> elements;
-  for (size_t i = 0; i < fields.size(); ++i) {
-    elements.push_back({fields[i].name(), values[i]});
+  for (const auto& [field, value] : llvm::zip_equal(fields, values)) {
```
zygoloid: Throughout this patch:

Using `const auto&` here seems unnecessary to me: the iterator will be generating new values, so a non-reference would do the same thing. Maybe remove the `&`s throughout?

zygoloid (approved):

explorer/interpreter/interpreter.cpp
jonmeow: Done

explorer/interpreter/interpreter.cpp
jonmeow: Acknowledged, although it seems not as worth it if we don't get more use-cases... (though, I can use `slice`, might go and fix that)

PR #3140, opened 2023-08-23
Optimize the outer lexer loop.

chandlerc:
Previously, the code would try each form of lexing and let that
sub-lexing routine reject the code. This was very branch heavy and also
hard to optimize -- lots of hard to inline function calls, etc.

However, it's really nice to keep the different categories of lexing
broken out into their own functions rather than flattening this into
a huge state machine.

So this creates a miniature explicit state machine by building a table
of function pointers that wrap the methods on the lexer. The main lexing
loop simply indexes this table with the first byte of the source code,
and calls the dispatch function pointer returned.

The result is that the main lex loop is *incredibly* tight code, and the
only time spent appears to be stalls waiting on memory for the next
byte. =]

As part of this, optimize symbol lexing specifically by recognizing all
the symbols that are exactly one character -- IE, we don't even need to
look at the *next* character, there is no max-munch or anything else.
For these, we pre-detect the exact token kind and hand that into the
symbol lexing routine to avoid re-computing it. The symbols in this
category are really frequent symbols in practice like `,` and `;`, so
this seems likely worthwhile in practice.

The one-byte-dispatch should also be reasonably extendable in the
future. For example, I suspect this is the likely hot-path for non-ASCII
lexing, where we see the UTF-8 marker at the start of a token and most
(if not all) of the token is non-ASCII. We can use this table to
dispatch immediately to an optimized routine dedicated to UTF-8
processing, without any slowdown for other inputs.

The benchmark results are best for keyword lexing because that is the
fastest thing to lex -- it goes form 25 mt/s to 30 mt/s. Other
improvements are less dramatic, but I think this is still worthwhile
because it gives a really strong basis for both expanded functionality
without performance hit and further optimizations.

### DIFF
--- a/toolchain/lexer/tokenized_buffer.cpp
+++ b/toolchain/lexer/tokenized_buffer.cpp
@@ -248,8 +248,8 @@ class TokenizedBuffer::Lexer {
     bool formed_token_;
   };
 
-  using DispatchFunctionT = LexResult(Lexer& lexer,
-                                      llvm::StringRef& source_text);
+  using DispatchFunctionT = auto (Lexer& lexer,
+                                  llvm::StringRef& source_text) -> LexResult;
   using DispatchTableT = std::array<DispatchFunctionT*, 256>;
 
   Lexer(TokenizedBuffer& buffer, DiagnosticConsumer& consumer)
@@ -473,17 +473,18 @@ class TokenizedBuffer::Lexer {
         return LexError(source_text);
       }
     } else {
-#ifndef NDEBUG
-      // Verify in a debug build that the incoming token kind is correct.
-      TokenKind debug_kind = llvm::StringSwitch<TokenKind>(source_text)
+      auto compute_kind = [] (llvm::StringRef source_text) {
+        return llvm::StringSwitch<TokenKind>(source_text)
 #define CARBON_SYMBOL_TOKEN(Name, Spelling) \
-  .StartsWith(Spelling, TokenKind::Name)
+    .StartsWith(Spelling, TokenKind::Name)
 #include "toolchain/lexer/token_kind.def"
-                                 .Default(TokenKind::Error);
-      CARBON_CHECK(kind == debug_kind)
+                                   .Default(TokenKind::Error);
+      };
+
+      // Verify in a debug build that the incoming token kind is correct.
+      CARBON_DCHECK(kind == compute_kind(source_text))
           << "Incoming token kind '" << kind
-          << "' does not match computed kind '" << debug_kind << "'!";
-#endif
+          << "' does not match computed kind '" << compute_kind(source_text) << "'!";
     }
 
     // In debug builds, re-check the kind after our optimizations to make sure
@@ -768,7 +769,7 @@ class TokenizedBuffer::Lexer {
       return lexer.LexKeywordOrIdentifier(source_text);
     };
     table['_'] = dispatch_lex_word;
-    // Note that we use raw loops because this needs to be `constexpr`
+    // Note that we don't use `llvm::seq` because this needs to be `constexpr`
     // evaluated.
     for (unsigned char c = 'a'; c <= 'z'; ++c) {
       table[c] = dispatch_lex_word;
--- a/toolchain/lexer/tokenized_buffer.cpp
+++ b/toolchain/lexer/tokenized_buffer.cpp
@@ -248,8 +248,8 @@ class TokenizedBuffer::Lexer {
     bool formed_token_;
   };
 
-  using DispatchFunctionT = auto (Lexer& lexer,
-                                  llvm::StringRef& source_text) -> LexResult;
+  using DispatchFunctionT = auto(Lexer& lexer, llvm::StringRef& source_text)
+      -> LexResult;
   using DispatchTableT = std::array<DispatchFunctionT*, 256>;
 
   Lexer(TokenizedBuffer& buffer, DiagnosticConsumer& consumer)
@@ -463,7 +463,7 @@ class TokenizedBuffer::Lexer {
     auto compute_symbol_kind = [](llvm::StringRef source_text) {
       return llvm::StringSwitch<TokenKind>(source_text)
 #define CARBON_SYMBOL_TOKEN(Name, Spelling) \
-    .StartsWith(Spelling, TokenKind::Name)
+  .StartsWith(Spelling, TokenKind::Name)
 #include "toolchain/lexer/token_kind.def"
           .Default(TokenKind::Error);
     };
@@ -811,27 +811,27 @@ auto TokenizedBuffer::Lex(SourceBuffer& source, DiagnosticConsumer& consumer)
 
   // Build a table of function pointers that we can use to dispatch to the
   // correct lexer routine based on the first byte of source text.
-  // 
+  //
   // While it is tempting to simply use a `switch` on the first byte and
   // dispatch with cases into this, in practice that doesn't produce great code.
   // There seem to be two issues that are the root cause.
-  // 
+  //
   // First, there are lots of different values of bytes that dispatch to a
   // fairly small set of routines, and then some byte values that dispatch
   // differently for each byte. This pattern isn't one that the compiler-based
   // lowering of switches works well with -- it tries to balance all the cases,
   // and in doing so emits several compares and other control flow rather than a
   // simple jump table.
-  // 
+  //
   // Second, with a `case`, it isn't as obvious how to create a single, uniform
   // interface that is effective for *every* byte value, and thus makes for a
   // single consistent table-based dispatch. By forcing these to be function
   // pointers, we also coerce the code to use a strictly homogenous structure
   // that can form a single dispatch table.
-  // 
+  //
   // These two actually interact -- the second issue is part of what makes the
   // non-table lowering in the first one desirable for many switches and cases.
-  // 
+  //
   // Ultimately, when table-based dispatch is such an important technique, we
   // get better results by taking full control and manually creating the
   // dispatch structures.
--- a/toolchain/lexer/tokenized_buffer.cpp
+++ b/toolchain/lexer/tokenized_buffer.cpp
@@ -13,6 +13,7 @@
 #include "common/check.h"
 #include "common/string_helpers.h"
 #include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/Sequence.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/ADT/StringSwitch.h"
 #include "llvm/Support/ErrorHandling.h"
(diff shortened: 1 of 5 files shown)


chandlerc:
Note that with this change, there is some sneaky performance hiding in these dispatch lambdas.

For ... various reasons ... LLVM chooses not to inline the functions into them, despite there being no other callers of these functions.

The result is that each of these lambdas turns into an actual function that tail calls into the relevant member function. This is still an improvement, but it isn't free or optimal.

I'm curious how folks would prefer I address this... I see several options...

1) Force inlining -- simple. But would have code size impact for symbols where we generate more meaningful thunks. We could try to leave just those.

2) Restructure to homogeneous member functions and use a table of pointers-to-member.

3) Restructure to homogeneous free functions.

#2 and #3 would both also involve passing the token, even when it is just a placeholder, and likely get some performance without code duplication that seems hard to get in #1.

Also interested in other ideas here. But to be clear, I'm mostly thinking of these as fix-forwards. This PR is still a strict improvement, I just finally dug through the profile enough to see some of the costs embedded here.

toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -247,6 +248,10 @@ class TokenizedBuffer::Lexer {
     bool formed_token_;
   };
 
+  using DispatchFunctionT = LexResult(Lexer& lexer,
+                                      llvm::StringRef& source_text);
```
zygoloid:
```suggestion
  using DispatchFunctionT = auto (Lexer& lexer,
                                  llvm::StringRef& source_text) -> LexResult;
```
I think our preference for trailing return types should apply here too.

toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -453,16 +458,47 @@ class TokenizedBuffer::Lexer {
     }
   }
 
-  auto LexSymbolToken(llvm::StringRef& source_text) -> LexResult {
-    TokenKind kind = llvm::StringSwitch<TokenKind>(source_text)
+  auto LexSymbolToken(llvm::StringRef& source_text, TokenKind kind)
+      -> LexResult {
+    // We use the `period` token as a place-holder for cases where one
+    // character isn't enough to pick a definitive symbol token. Recompute the
+    // kind using the full symbol set.
```
zygoloid: I found this choice a little surprising -- I think with the current lexical rules, `.` is one of the characters that is sufficient to pick a definitive symbol token. We don't have `...` as a token (yet), and numeric tokens can't start with `.`.

Could we use `TokenKind::Error` as the placeholder instead of a valid token kind?

toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -453,16 +458,47 @@ class TokenizedBuffer::Lexer {
...
+      // Verify in a debug build that the incoming token kind is correct.
+      TokenKind debug_kind = llvm::StringSwitch<TokenKind>(source_text)
+#define CARBON_SYMBOL_TOKEN(Name, Spelling) \
+  .StartsWith(Spelling, TokenKind::Name)
+#include "toolchain/lexer/token_kind.def"
+                                 .Default(TokenKind::Error);
+      CARBON_CHECK(kind == debug_kind)
+          << "Incoming token kind '" << kind
+          << "' does not match computed kind '" << debug_kind << "'!";
+#endif
```
zygoloid: @jonmeow expressed a distaste for using `#ifndef NDEBUG` in prior review, and we came up with a different pattern of using `CARBON_DCHECK`, containing a lambda if necessary:

```suggestion
      // Verify in a debug build that the incoming token kind is correct.
      CARBON_DCHECK([&] {
        TokenKind debug_kind = llvm::StringSwitch<TokenKind>(source_text)
#define CARBON_SYMBOL_TOKEN(Name, Spelling) \
    .StartsWith(Spelling, TokenKind::Name)
#include "toolchain/lexer/token_kind.def"
                                   .Default(TokenKind::Error);
        return kind == debug_kind;
      }) << "Incoming token kind '" << kind << "' does not match computed kind";
```

This unfortunately loses the `debug_kind` from the log, but we could get it back with (eg):

```suggestion
      auto compute_kind = [] (llvm::StringRef source_text) {
        return llvm::StringSwitch<TokenKind>(source_text)
#define CARBON_SYMBOL_TOKEN(Name, Spelling) \
    .StartsWith(Spelling, TokenKind::Name)
#include "toolchain/lexer/token_kind.def"
                                   .Default(TokenKind::Error);
      };

      // Verify in a debug build that the incoming token kind is correct.
      CARBON_DCHECK(kind == compute_kind(source_text))
          << "Incoming token kind '" << kind
          << "' does not match computed kind '" << compute_kind(source_text) << "'!";
```

... and there are other places in this function where we do the same computation and could use the `compute_kind` lambda.

toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -453,16 +458,47 @@ class TokenizedBuffer::Lexer {
...
+#ifndef NDEBUG
+    TokenKind debug_kind = llvm::StringSwitch<TokenKind>(source_text)
+#define CARBON_SYMBOL_TOKEN(Name, Spelling) \
+  .StartsWith(Spelling, TokenKind::Name)
+#include "toolchain/lexer/token_kind.def"
+                               .Default(TokenKind::Error);
+    CARBON_CHECK(debug_kind == kind)
+        << "Optimized code computed kind '" << kind
+        << "' but it should have been '" << debug_kind << "'";
+#endif
```
zygoloid: This check looks redundant, given that we either computed `kind` using identical code to the computation of `debug_kind`, or already did an identical check.

toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -692,6 +730,82 @@ class TokenizedBuffer::Lexer {
...
+    // orthogonal to any other punctuation. We do this separately because this
+    // needs to override soe of the generic handling above.
+#define CARBON_ONE_CHAR_SYMBOL_TOKEN(TokenName, Spelling)                  \
+  table[(Spelling)[0]] = +[](Lexer& lexer, llvm::StringRef& source_text) { \
+    return lexer.LexSymbolToken(source_text, TokenKind::TokenName);        \
+  };
+#include "toolchain/lexer/token_kind.def"
+
+    // For identifiers, keywords, and numeric type literals, we use the
+    // `Identifier` placeholder token kind.
```
zygoloid: Is this comment stale? We don't seem to be using a placeholder token kind here.

toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -692,6 +730,82 @@ class TokenizedBuffer::Lexer {
...
+    auto dispatch_lex_word = +[](Lexer& lexer, llvm::StringRef& source_text) {
+      return lexer.LexKeywordOrIdentifier(source_text);
+    };
+    table['_'] = dispatch_lex_word;
+    // Note that we use raw loops because this needs to be `constexpr`
+    // evaluated.
```
zygoloid: I assume this means "as opposed to `llvm::seq`" -- my first thought was that you were contrasting to
```c++
for (unsigned char c : "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ_") {
  table[c] = dispatch_lex_word;
}
```
or something like that.
```suggestion
    // Note that we don't use `llvm::seq` because this needs to be `constexpr`
    // evaluated.
```

toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -692,6 +730,82 @@ class TokenizedBuffer::Lexer {
...
+    // We dispatch all non-ASCII UTF-8 characters to the identifier lexing
+    // as whitespace characters should already have been skipped and the
+    // only remaining valid Unicode characters would be part of an
+    // identifier. That code can either accept or reject.
+    for (int i = 0x80; i < 0x100; ++i) {
+      table[i] = dispatch_lex_word;
+    }
+
+    // For numeric literals of all kinds, we use the `IntegerLiteral`
+    // placeholder token kind.
```
zygoloid: I think this comment is also stale.

toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -716,24 +830,16 @@ auto TokenizedBuffer::Lex(SourceBuffer& source, DiagnosticConsumer& consumer)
   ErrorTrackingDiagnosticConsumer error_tracking_consumer(consumer);
   Lexer lexer(buffer, error_tracking_consumer);
 
+  // Build a table of function pointers that we can use to dispatch to the
+  // correct lexer routine based on the first byte of source text.
```
zygoloid, line 813: Overall, as a reader of this I'm left wondering why we use a table of function pointers rather than a `switch` on the first character. I think it'd be useful to explain that here.

toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -716,24 +830,16 @@ auto TokenizedBuffer::Lex(SourceBuffer& source, DiagnosticConsumer& consumer)
...
+  constexpr Lexer::DispatchTableT DispatchTable = Lexer::MakeDispatchTable();
+
   llvm::StringRef source_text = source.text();
   while (lexer.SkipWhitespace(source_text)) {
```
zygoloid, line 841: Should we do whitespace skipping as part of the dispatch loop too? It seems unnecessary to first check whether we're at a whitespace character and then check again what kind of token we're forming, in the case where there isn't whitespace between tokens.

toolchain/lexer/tokenized_buffer.cpp
chandlerc: Didn't realize we had a `DCHECK`, nice!

toolchain/lexer/tokenized_buffer.cpp
chandlerc: Doh, meant to remove this when I hoisted the code up, sorry.

toolchain/lexer/tokenized_buffer.cpp
chandlerc: Yes, sorry. I had originally used placeholder tokens for *everything*, and then realized I could do better by directly building the dispatch table. Updated comment, thanks!

toolchain/lexer/tokenized_buffer.cpp
chandlerc: Yep!

toolchain/lexer/tokenized_buffer.cpp
chandlerc: Ah, good point. Added a comment about this.

toolchain/lexer/tokenized_buffer.cpp
chandlerc: Possibly, I'm not sure.

I want to vectorize the whitespace skipping first, and then see if it still makes sense... My *guess* is that we'll not check the first character, but immediately do a full vector scan here, and so not benefit much from the character approach. But definitely worth revisiting to see.

toolchain/lexer/tokenized_buffer.cpp
chandlerc: Done, also switched to a default argument so it is even more clear at this layer what is going on.

FWIW, I think we may want to chain things onto `.` in the future, but not super relevant at this point. =D

chandlerc (commented):
Thanks, PTAL!

zygoloid (approved):

PR #3141, opened 2023-08-23
Restructure benchmark for stability.

chandlerc:
It was pointed out in review that the approach to building random inputs
for the benchmark would produce significantly varying input lengths and
cause unavoidable noise of as much as 3% between runs.

While initially, these were only used for very coarse measurements, this
is a definite problem, and so this PR restructures things along the
lines suggested. Rather than generating random lengths or random
selections according to distributions, instead the lengths and coverage
of sets are done deterministically. Then the sequences are randomized in
order to prevent getting stuck in a silent local minima or maxima.

The technique works to ensure that the randomness is done on every run
of the benchmark, not just every run of the program, so that we don't
have noise hidden from run-to-run. That means fresh memory allocation
and some wasted time computed freshly shuffled inputs, but makes sure
the unavoidable noise from things like ASLR show up (as much as
possible) even with simple repetitions of the benchmark runs.

As part of this, this PR moves towards building an entirely custom
distribution of identifier lengths based on the direct measurements of
the LLVM codebase. These measurements are provided by a script that will
be in a subsequent PR, but the exact distribution doesn't matter as much
as our ability to control it and build on it in a fully deterministic
way.

To help with analyzing all of these, we also start tracking the bytes
processed in addition to the tokens processed as rates. The naming is
changed to be similar, and this produces the following nice output from
the benchmark now:
```
---------------------------------------------------------------------------------------------------------------
Benchmark                                  Time             CPU   Iterations bytes_per_second tokens_per_second
---------------------------------------------------------------------------------------------------------------
BM_ValidKeywords                     3365571 ns      3365577 ns          206       177.449M/s        29.7126M/s
BM_ValidIdentifiers<1, 64, false>   12646151 ns     12645834 ns           52       116.466M/s        7.90774M/s
BM_ValidIdentifiers<1, 1, true>      4388390 ns      4388141 ns          161       65.1988M/s        22.7887M/s
BM_ValidIdentifiers<3, 5, true>     16286551 ns     16286572 ns           43       35.1334M/s        6.14003M/s
BM_ValidIdentifiers<3, 16, true>    15798770 ns     15797567 ns           44       69.4229M/s        6.33009M/s
BM_ValidIdentifiers<12, 64, true>   15670499 ns     15670257 ns           38       243.421M/s        6.38152M/s
BM_ValidMix/10/40                    7415940 ns      7415564 ns           94       134.257M/s        13.4852M/s
BM_ValidMix/25/30                    7349454 ns      7349402 ns           95       121.762M/s        13.6065M/s
BM_ValidMix/50/20                    6691223 ns      6690969 ns          105         100.3M/s        14.9455M/s
BM_ValidMix/75/10                    5131517 ns      5131263 ns          135       86.4556M/s        19.4884M/s
```


toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -42,19 +46,19 @@ auto IdentifierChars() -> llvm::ArrayRef<char> {
...
 // (the default) the routine tries to generate a distribution that roughly
 // matches what we observe in C++ code.
-auto GenerateRandomIdentifier(absl::BitGen& gen, int min_length = 1,
-                              int max_length = 64, bool uniform_lengths = false)
-    -> std::string {
+auto GenerateRandomIdentifier(absl::BitGen& gen, int length) -> std::string {
   llvm::ArrayRef<char> start_chars = IdentifierStartChars();
   llvm::ArrayRef<char> chars = IdentifierChars();
 
+#if 0
```
jonmeow: Should this be removed?

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -89,41 +94,196 @@ auto GenerateRandomIdentifier(absl::BitGen& gen, int min_length = 1,
...
+    std::array<int, 64> id_length_counts;
+    // For non-uniform distribution, we simulate a distribution roughly based on
+    // the observed histogram of identifier lengths in LLVM, but smoothed a bit
+    // and reduced to small counts so that we cycle through all the lengths
+    // reasonably quickly. We want the sampling even 10% of NumTokens from this
+    // in a round-robin form to not be skewed overly much. This still inherently
+    // compresses the long tail as we'd rather have coverage even though it
+    // distorts the distribution a bit.
+    //
+    // 1 characters   [3976]  ███████████████████████████████▊
```
jonmeow, line 104: Are there stray characters at the end of the line here? Note I'm seeing similar for each of these lines.

As I've been looking at this, maybe you were trying to use some special character for a distribution chart? I'm not seeing an explanation of that though; maybe it'd be helpful to just state the percent?

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -89,41 +94,196 @@ auto GenerateRandomIdentifier(absl::BitGen& gen, int min_length = 1,
...
+    id_length_counts[0] = 40;
```
jonmeow, line 105: Had you considered doing something like "3976" or "3976 / 100" here instead of "40"? Particularly with both 3976 -> 40 and 3724 -> 40 (or 1568 -> 20, 2380 -> 20), the relation seems slightly ambiguous. But if you did something like "3976 / 100", you'd get similar values while also maybe being able to trim the value from the comment.

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -89,41 +94,196 @@ auto GenerateRandomIdentifier(absl::BitGen& gen, int min_length = 1,
...
+    for (int i : llvm::seq(18, 64)) {
+      id_length_counts[i] = 1;
+    }
+
+    // Used to track the different count buckets when in a non-uniform
+    // distribution.
+    int length_bucket_index = 0;
+    int length_count = 0;
+
+    for (int i : llvm::seq(0, NumTokens)) {
```
jonmeow: I think `llvm::seq(NumTokens)` should now work (with the recent LLVM update, it was added). Also comes up at least one other spot I see. Note, I'm not sure you'll like it without the `0`, but I'm a bit used to it from python's `range` api.

But also, where you're using `llvm::seq` right now, would `llvm::enumerate` work? I'm not sure whether that'd give you the mutable reference you'd want for assignment. (note here, I had to scroll up to verify that `ids` had a matching length, you might consider placing that closer to here -- although enumerate would also solve that)

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -179,66 +340,40 @@ void BM_ValidKeywords(benchmark::State& state) {
...
 
 // Also benchmark a few uniform distribution ranges of identifier widths to
 // cover different patterns that emerge with small, medium, and longer
 // identifiers.
-BENCHMARK_CAPTURE(BM_ValidIdentifiers, Uniform,
-                  /*uniform_lengths=*/true)
-    ->Args({3, 5})
-    ->Args({3, 16})
-    ->Args({12, 64});
+BENCHMARK(BM_ValidIdentifiers<3, 5, /*Uniform=*/true>);
```
jonmeow, line 373: Given the prevalence of single-char identifiers, would that be worth explicitly checking, particularly as it may be an edge case?

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -89,41 +94,196 @@ auto GenerateRandomIdentifier(absl::BitGen& gen, int min_length = 1,
...
-// symbols for simplicity.
-auto GenerateRandomSymbol(absl::BitGen& gen) -> llvm::StringRef {
-  llvm::ArrayRef<TokenKind> table = GetSymbolTokenTable();
-  auto index = absl::Uniform<int>(gen, 0, table.size());
-  return table[index].fixed_spelling();
+// Compute a random sequence of mixed symbols, keywords, and identifiers, with
+// percentages of each according to the parameters.
+auto RandomMixedSeq(int symbol_percent, int keyword_percent) -> std::string {
+  std::string result;
+  absl::BitGen gen;
```
jonmeow: Can these be moved closer to their first use? I find the style of declaring close to the first use helpful when reading code (https://google.github.io/styleguide/cppguide.html#Local_Variables). And actually, could BitGen even be inlined as a temporary?

Note this comes up in various functions, as you seem to be declaring some at the top, separate from use.

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -179,66 +340,40 @@ void BM_ValidKeywords(benchmark::State& state) {
...
-  state.counters["TokenRate"] = benchmark::Counter(
+  state.SetBytesProcessed(state.iterations() * source.size());
+  state.counters["tokens_per_second"] = benchmark::Counter(
       NumTokens, benchmark::Counter::kIsIterationInvariantRate);
 }
 // Benchmark the non-uniform distribution we observe in C++ code.
-BENCHMARK_CAPTURE(BM_ValidIdentifiers, Representative,
-                  /*uniform_lengths=*/false)
-    ->Args({1, 64});
+BENCHMARK(BM_ValidIdentifiers<>);
```
jonmeow: WDYT of removing the default values, and passing them explicitly here? They seem to only be used here, and it'd mean they'd be printed in benchmark output.

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -89,41 +94,196 @@ auto GenerateRandomIdentifier(absl::BitGen& gen, int min_length = 1,
...
+    // 14 characters  [ 284]  ██▎
+    id_length_counts[13] = 3;
+    // 15 characters  [ 172]  █▍
+    // 16 characters  [ 278]  ██▎
+    // 17 characters  [ 191]  █▌
+    // 18 characters  [ 207]  █▋
+    for (int i : llvm::seq(14, 18)) {
+      id_length_counts[i] = 2;
+    }
+    // 19 - 63 characters are all <100 in LLVM but non-zero., and we map them to
```
jonmeow: `non-zero.,` has extra punctuation.

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -89,41 +94,196 @@ auto GenerateRandomIdentifier(absl::BitGen& gen, int min_length = 1,
...
+}
+
+// Compute a random sequence of just identifiers.
+template <int MinLength = 1, int MaxLength = 64, bool Uniform = false>
+auto RandomIdentifierSeq() -> std::string {
+  std::string result;
+  absl::BitGen gen;
+
+  // Get a static pool of identifiers with the desired distribution.
+  llvm::ArrayRef<std::string> ids =
```
jonmeow: I had to double-check the length here... Could this be a `const std::array<std::string, NumTokens>&` so that it's an API commitment that len(ids) == len(indices)?

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -89,41 +94,196 @@ auto GenerateRandomIdentifier(absl::BitGen& gen, int min_length = 1,
...
+  };
+  std::array<std::pair<ElementKind, int>, NumTokens> indices;
+  for (int i : llvm::seq(0, NumTokens)) {
+    int i_percent = i % 100;
+    if (i_percent < symbol_percent) {
+      indices[i] = {Symbol, i % symbols.size()};
+    } else if (i_percent < keyword_percent) {
+      indices[i] = {Keyword, i % keywords.size()};
+    } else {
+      indices[i] = {Identifier, i % ids.size()};
```
jonmeow: Won't this produce a biased sample because ids.size() == NumTokens, so this is only ever using a random subset of `ids`? To the point of sample stability, should GetRandomIdentifiers be taking the number of `ids` to generate as an argument, or otherwise generating some pool that'd be repeated instead of subsampled? This seems particularly relevant for Uniform where the order is deterministic, and given symbol_percent/keyword_percent result in a rotation every 100 entries, that's why I think it's a specifically a biased subsample.

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -89,41 +94,196 @@ auto GenerateRandomIdentifier(absl::BitGen& gen, int min_length = 1,
...
+  // the desired distribution, then shuffle that list.
+  enum ElementKind {
+    Symbol,
+    Keyword,
+    Identifier,
+  };
+  std::array<std::pair<ElementKind, int>, NumTokens> indices;
+  for (int i : llvm::seq(0, NumTokens)) {
+    int i_percent = i % 100;
+    if (i_percent < symbol_percent) {
```
jonmeow: 1. I think you might get an odd distribution here because of how you split by i_percent, then do the i %. You get non-sequential `i` values that you then mod.

2. Also, you do `i_percent < symbol_percent` and `i_percent < keyword_percent`. With symbol_percent = 50 and keyword_percent = 20, doesn't that result in no keywords?

Had you considered something like loops for each:

```
CARBON_CHECK(symbol_percent + keyword_percent < 100) << "These are percents, and to avoid handling rounding errors we expect at least some identifiers.";
int symbol_count = NumTokens * 100 / symbol_percent;
for (int i : llvm::seq(0, symbol_count)) {
  indices[i] = {Symbol, i % symbols.size()};
}
int keyword_count = NumTokens * 100 / keyword_percent;
for (int i : llvm::seq(symbol_count, symbol_count + keyword_count)) {
  indices[i] = {Keyword, i % keywords.size()};
}
for (int i : llvm::seq(symbol_count + keyword_count, NumTokens)) {
  // Note as currently implemented this doesn't actually need the mod, since ids.size() == NumTokens.
  // But I'm wondering if that should be fixed.
  indices[i] = {Identifier, i % ids.size()};
}
```

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Doh, yes. Sorry.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: It's a bar-chart of the histogram that the script I used to collect the measurements renders. If you look down, you can see the relative size of each of these from the length of the bar.

I need to tidy up the python script and send it out for review, but seemed like that should be separate.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Yeah, but as I described in the comment, I tried to smooth out the distribution quite a bit. The goal isn't to try to *exactly* match the observed distribution, but get something fairly representative.

Does that make sense? Is there another structure that would seem better?

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Fixed, thanks.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Will try to use the shorter form of `seq`, but good idea about enumerate! Done.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Yeah, good idea, done.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Yeah, I kept thinking I'd need them in more places and then ended up not.

I've tried to move them down systematically, and make them temporaries where I spotted an easy place to do so.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Hmm, yeah, I didn't end up with a great scheme...

I'm not sure its biased, but it's definitely not working the intended way. My goal was to take the first N identifiers for the percentage -- with the identifier table trying to rapidly cycle through the range of lengths as rapidly as possible so this should get as much of a distribution of the lengths as possible.

I'll change how this works as part of the above fix.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Oof, yeah, all of this logic was just busted. Reworking it along the lines of your suggestion.

That said, I think its fine to support 0% identifiers here, so I'll let these sum right up to 100. That seem OK?

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Sure, done.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Yeah, good call. Also tweaked code to support that.

chandlerc (commented):
Thanks, PTAL!

toolchain/lexer/tokenized_buffer_benchmark.cpp
jonmeow: Thanks for explaining; given I had trouble understanding it, maybe it'd help to add an explanatory comment?

You might also consider changing using basic ASCII art like `#`, these block characters are something I could only make in copy-paste and might have some rendering quirks in Chrome.

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -89,41 +76,214 @@ auto GenerateRandomIdentifier(absl::BitGen& gen, int min_length = 1,
...
+    std::array<std::string, NumTokens> ids;
+    absl::BitGen gen;
+    for (auto [i, id] : llvm::enumerate(ids)) {
+      if (Uniform) {
+        // Rather than using randomness, for a uniform distribution rotate
+        // lengths in round-robin to get a deterministic and exact size on every
+        // run. We will then shuffle them at the end to produce a random
+        // ordering.
+        int length = MinLength +
+                     (MaxLength > MinLength ? i % (MaxLength - MinLength) : 0);
```
jonmeow: If MinLength == 3 and MaxLength == 5, you get `MinLength + i % (MaxLength - MinLength)` or `3 + i % 2`. This will result in values of 3 or 4. If MaxLength is intended to be non-inclusive, maybe you should just require MaxLength > MinLength and the 1-char case should be `1, 2` instead of `1, 1`? Otherwise, I think this could be just `MinLength + (i % (MaxLength - MinLength + 1))`, no ternary.

toolchain/lexer/tokenized_buffer_benchmark.cpp
```diff
@@ -89,41 +76,214 @@ auto GenerateRandomIdentifier(absl::BitGen& gen, int min_length = 1,
...
+    Symbol,
+    Keyword,
+    Identifier,
+  };
+  std::array<std::pair<ElementKind, int>, NumTokens> indices;
+
+  int num_symbols = (NumTokens / 100) * symbol_percent;
+  int num_keywords = (NumTokens / 100) * keyword_percent;
+  int num_identifiers =
+      (NumTokens / 100) * (100 - symbol_percent - keyword_percent);
```
jonmeow: As a possible alternative, maybe `NumTokens - num_symbols - num_keywords`? I think that'd also moot the next CHECK, but maybe you prefer this route.

jonmeow (approved):
Thanks! This looks good to me, though I still had a couple comments. Feel free to merge either way though.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: As these are copy-pasted from a script, I'd rather not manually adjust it. So far, I think it renders reasonably well everywhere I've seen.

If it's super important to change this, I would want to figure out how to do that in the script that generates these -- right now, it's just the output of `termplotlib` without any customization.

For now, going to just add a comment.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Richard had wanted this to be inclusive, and I just didn't do the math right here, and then did the wrong fix. Moving with the second suggestion, thanks for catching this.

toolchain/lexer/tokenized_buffer_benchmark.cpp
chandlerc: Sure, that's a nicer framing.

chandlerc (commented):
Thanks, responded and merging!

PR #3143, opened 2023-08-23
Switch to parameter name based comments.

chandlerc:


jonmeow (approved):
Thanks! Please make sure you meant to change the values on 116-119 though, I'm not seeing any comment about it.

jonmeow:
Sorry, one more comment -- you might consider changing the comments like:

```
// bit0 = 1 for `_`: high `0x5` and low `0xF`
```

to (IIUC):

```
// bit0 = 1 for `_`: high `b5` and low `b15`
```

toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -111,39 +111,39 @@ static auto ScanForIdentifierPrefix(llvm::StringRef text) -> llvm::StringRef {
...
-      /*0x7:*/ 0b0000'1000,
-      /*0x8:*/ 0b0000'0000,
-      /*0x9:*/ 0b0000'0000,
-      /*0xA:*/ 0b0000'0000,
-      /*0xB:*/ 0b0000'0000,
-      /*0xC:*/ 0b0000'0000,
-      /*0xD:*/ 0b0000'0000,
-      /*0xE:*/ 0b0000'0000,
-      /*0xF:*/ 0b0000'0000);
+      /* __b0=*/0b0000'0000,
```
zygoloid, line 114: I'm a bit hesitant to add comments listing the names these parameters happen to have in one "random" compiler's `<emmintrin.h>`.

These names don't match the ones from [intel's intrinsics guide](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_setr_epi8) where they're named `e15`, `e14`, ..., `e0`, nor the one in intel's Software Development Manual, volume 2C (page C-17, table C-2), where they're named `w15`, `w14`, ..., `w0`. And they don't match the names used by GCC's `<emmintrin.h>`, which are `__q00`, ..., `__q15`.

I suppose the point here is that these match the names used specifically in the header that `clang-tidy` is looking at, so maybe they're the best we can do if we want any kind of checking?

chandlerc: Sadly, we don't get clang-tidy checking for these either. I tried a bunch of spacing and other things, none got a clang-tidy issue when I wrote an actual bug.

chandlerc:
> Sorry, one more comment -- you might consider changing the comments like:
> 
> ```
> // bit0 = 1 for `_`: high `0x5` and low `0xF`
> ```
> 
> to (IIUC):
> 
> ```
> // bit0 = 1 for `_`: high `b5` and low `b15`
> ```

I don't think this is quite right.

The high `0x5` here is a high nibble of the byte `0x5F`. And that nibble *selects* the 5'th byte, which is passed to the `setr` intrinsic in `__b5`. But this comment is really talking about about what the nibbles of th ebyte are when "bit0" is set to 1, so I'd expect them to be written as nibbles-of-a-byte, and not get into the index-of-the-LUT side of the implementation technique.

chandlerc:
> Please make sure you meant to change the values on 116-119 though, I'm not seeing any comment about it.

No, hit the wrong buttons in the editor.

Anyways, real issue is finding something that both you and Richard are happy with. I don't have a super strong opinion there beyond desiring to write these as binary literals and line up the bits in columns so they're a bit easier to see the structural aspects. I'm happy with whatever comment pattern you all like.

toolchain/lexer/tokenized_buffer.cpp
zygoloid: The check intentionally ignores parameter names from system headers because they're definitely not part of the API contract and likely to be reserved identifiers: https://github.com/llvm/llvm-project/blob/2b664d678dcd6b23eda992b45d009788e34ea8f1/clang-tools-extra/clang-tidy/bugprone/ArgumentCommentCheck.cpp#L67

jonmeow:
So here's my understanding. In Carbon, we (a) will not support block comments, but (b) might support syntax similar to designated initializers for parameter names (basically, similar to a comment, but somewhat more syntactic -- and closer to a matching expectation).

Unsupporting block comments means that `/*0xE:*/` would not work. Requiring designated initializers for parameter names would mean knowing the parameter name.

Designated initializers for parameter names means a better understanding of what that parameter name is, particularly for refactoring. When it comes to certain headers like this one, the parameter name may be platform-dependent. If Carbon were wrapping this, maybe that means a named parameter would stop compiling?

I don't know that there's a perfect answer here. Maybe this indicates a flaw in Carbon's design for argument comments?

zygoloid:
These are logically more like an array than named parameters. But I don't think that observation helps, because we don't have a great syntax (in either C++ or Carbon) for initializing an array while specifying element indexes, nor a great way to pass the elements of an array as the arguments in a function call. I suppose something like this is possible:

```c++
  alignas(__m128i)
  constexpr unsigned char low_lut_data[] = {
      0b0000'1010,
      0b0000'1110,
      0b0000'1110,
      0b0000'1110,
      0b0000'1110,
      0b0000'1110,
      0b0000'1110,
      0b0000'1110,
      0b0000'1110,
      0b0000'1110,
      0b0000'1100,
      0b0000'0100,
      0b0000'0100,
      0b0000'0100,
      0b0000'0100,
      0b0000'0101};
  const auto low_lut = _mm_loadu_epi8(low_lut_data);
```

... but I don't think that really helps anything, unless we use the C99 / GNU designator extension:

```c++
  alignas(__m128i)
  constexpr unsigned char low_lut_data[] = {
      [0] = 0b0000'1010,
      [1] = 0b0000'1110,
      [2] = 0b0000'1110,
      // ...
```

... but it seems better to avoid language extensions here, even though we're using the intel intrinsic extensions.

chandlerc:
Going with unchecked parameter names as they're at least used on one platform. 🤷🏻 We can revisit if a better way of writing the code emerges.

PR #3144, opened 2023-08-23
Model parameters as value expressions rather than reference expressions.

zygoloid:
Add a new node kind to distinguish parameters from local variables so that we can give the two kinds of binding different expression categories.


jonmeow (approved):
Looks great!

PR #3145, opened 2023-08-23
Put the IR for the return type in the same block as the IR for the parameters.

zygoloid:
Previously, IR generated for the computation of the return type ended up in the package's node block. Put it with the IR for the parameters for consistency and cleanliness.


jonmeow (approved):
Sure, seems reasonable.

PR #3146, opened 2023-08-24
Don't try to infer where to put loads in lowering.

zygoloid:
Trust semantics to have put them in the right places.

Many parts of lowering still need to be updated to use the value
representation chosen at the semantics layer, but this is an incremental
step towards that.

### DIFF
--- a/toolchain/check/handle_index.cpp
+++ b/toolchain/check/handle_index.cpp
@@ -63,17 +63,24 @@ auto HandleIndexExpression(Context& context, Parse::Node parse_node) -> bool {
       auto cast_index_id = ConvertToValueOfType(
           context, index_node.parse_node(), index_node_id,
           context.CanonicalizeType(SemIR::NodeId::BuiltinIntegerType));
-      if (SemIR::GetExpressionCategory(context.semantics_ir(),
-                                       operand_node_id) ==
-          SemIR::ExpressionCategory::Value) {
+      auto array_cat =
+          SemIR::GetExpressionCategory(context.semantics_ir(), operand_node_id);
+      if (array_cat == SemIR::ExpressionCategory::Value) {
         // If the operand is an array value, convert it to an ephemeral
-        // reference to an array so we can index into it.
+        // reference to an array so we can perform a primitive indexing into it.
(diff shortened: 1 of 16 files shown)


jonmeow (dismissed):

jonmeow:
Not merging due to dependencies, but this PR looks straightforward on its own.

toolchain/check/handle_index.cpp
```diff
@@ -63,6 +63,14 @@ auto HandleIndexExpression(Context& context, Parse::Node parse_node) -> bool {
       auto cast_index_id = ConvertToValueOfType(
           context, index_node.parse_node(), index_node_id,
           context.CanonicalizeType(SemIR::NodeId::BuiltinIntegerType));
+      if (SemIR::GetExpressionCategory(context.semantics_ir(),
+                                       operand_node_id) ==
+          SemIR::ExpressionCategory::Value) {
+        // If the operand is an array value, convert it to an ephemeral
+        // reference to an array so we can index into it.
+        operand_node_id = context.AddNode(SemIR::Node::ValueAsReference::Make(
+            parse_node, operand_type_id, operand_node_id));
```
chandlerc: Maybe a TODO here to revisit when indexing works via interfaces? My guess is that will result in something different here where we'll have a way to do array indexing entirely in the value space without forming reference that we know we'll just re-bind to values later?

Hmmm, I see below -- this is actually modeling a value-rep that's a pointer? That's not really the same as turning a value into an ephemeral reference though as it doesn't materialize a temporary from the value and build a reference to the temporary?

Unsure if I'm misunderstanding this or the comment just needs to be updated (or this is just part of what is TODO).

chandlerc (approved):
Some questions about the `value_as_reference` bit, but I think this is fine to land regardless -- easy to fix forward here and seems like a really nice step of leveraging the semir improvements.

toolchain/check/handle_index.cpp
zygoloid: This comment made me realize that this isn't quite handling expression categories for array indexing properly. We were treating an array access of an ephemeral reference expression or a value expression as producing an ephemeral reference expression, but the design says that both operations produce a value expression.

Fixed that, updated the comments, added a TODO, and added a test.

PR #3147, opened 2023-08-24
Use `memcpy` for memory-to-memory assignment.

zygoloid:
For types using a pointer value representation, emit an `llvm.memcpy` call to perform assignment rather performing a first-class aggregate load and store.

### DIFF
--- a/toolchain/lowering/lowering_handle.cpp
+++ b/toolchain/lowering/lowering_handle.cpp
@@ -79,8 +79,38 @@ auto LoweringHandleArrayValue(LoweringFunctionContext& context,
 auto LoweringHandleAssign(LoweringFunctionContext& context,
                           SemIR::NodeId /*node_id*/, SemIR::Node node) -> void {
   auto [storage_id, value_id] = node.GetAsAssign();
-  context.builder().CreateStore(context.GetLocalLoaded(value_id),
-                                context.GetLocal(storage_id));
+  auto storage_type_id = context.semantics_ir().GetNode(storage_id).type_id();
+
+  // We can assign from either a value expression or a by-copy initializing
+  // expression. In either case, the value of the source is the value
+  // representation of the target.
+  // TODO: Should we use different semantic nodes for those operations?
+  switch (auto rep = SemIR::GetValueRepresentation(context.semantics_ir(),
+                                                   storage_type_id);
+          rep.kind) {
+    case SemIR::ValueRepresentation::None:
+      break;
+    case SemIR::ValueRepresentation::Copy:
+      context.builder().CreateStore(context.GetLocalLoaded(value_id),
+                                    context.GetLocal(storage_id));
+      break;
+    case SemIR::ValueRepresentation::Pointer: {
+      auto& layout = context.llvm_module().getDataLayout();
+      auto* type = context.GetType(storage_type_id);
+      auto size = layout.getTypeAllocSize(type);
+      // TODO: Compute known alignment of the source and destination, which may
+      // be greater than the alignment computed by LLVM.
+      auto align = layout.getABITypeAlign(type);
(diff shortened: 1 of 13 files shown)


toolchain/lowering/lowering_handle.cpp
```diff
@@ -79,14 +79,38 @@ auto LoweringHandleArrayValue(LoweringFunctionContext& context,
 auto LoweringHandleAssign(LoweringFunctionContext& context,
                           SemIR::NodeId /*node_id*/, SemIR::Node node) -> void {
   auto [storage_id, value_id] = node.GetAsAssign();
+  auto storage_type_id = context.semantics_ir().GetNode(storage_id).type_id();
+
   // We can assign from either a value expression or a by-copy initializing
   // expression. In either case, the value of the source is the value
   // representation of the target.
   // TODO: Should we use different semantic nodes for those operations?
```
jonmeow, line 87: FWIW, I could see this as being useful from a simple efficiency perspective, removing a common switch. But maybe it should be a benchmark-oriented decision? i.e., it may not make a difference for performance, and having a single node gives us more flexibility to decide lowering-specific optimizations.

toolchain/lowering/lowering_handle.cpp
```diff
@@ -79,14 +79,38 @@ auto LoweringHandleArrayValue(LoweringFunctionContext& context,
...
+    case SemIR::ValueRepresentation::None:
+      break;
+    case SemIR::ValueRepresentation::Copy:
+      context.builder().CreateStore(context.GetLocal(value_id),
+                                    context.GetLocal(storage_id));
+      break;
+    case SemIR::ValueRepresentation::Pointer: {
+      auto& layout = context.llvm_module().getDataLayout();
+      auto* type = context.GetType(storage_type_id);
+      auto size = layout.getTypeAllocSize(type);
```
jonmeow: You might consider inlining this since it's only used once.

toolchain/lowering/lowering_handle.cpp
```diff
@@ -79,14 +79,38 @@ auto LoweringHandleArrayValue(LoweringFunctionContext& context,
...
   // TODO: Should we use different semantic nodes for those operations?
-  // TODO: Handle the case where the value representation is not the object
-  // representation.
-  context.builder().CreateStore(context.GetLocal(value_id),
-                                context.GetLocal(storage_id));
+  switch (auto rep = SemIR::GetValueRepresentation(context.semantics_ir(),
+                                                   storage_type_id);
+          rep.kind) {
+    case SemIR::ValueRepresentation::None:
+      break;
```
jonmeow, line 92: I *think* we might want to generate the assign node in this case, but equally, maybe we don't need to? It seems like it might be unnecessary, and then it wouldn't reach this point -- a minor efficiency. And semantically, no assignment actually is expected, right?

jonmeow (approved):
Makes sense to me. A few comments, but really this can be merged as is. I'm only not merging due to the PR stack.

toolchain/lowering/lowering_handle.cpp
zygoloid: Yeah, I've been on the fence about how to represent cases like this in SemIR; I think we should discuss that offline. @chandlerc has been in favor of preserving information into SemIR for semantic analyses and I think this is probably a case where we want to do that.

toolchain/lowering/lowering_handle.cpp
zygoloid: Done.

toolchain/lowering/lowering_handle.cpp
jonmeow: Okay, makes sense to me.

PR #3148, opened 2023-08-24
Take the value representation into account in BindValue lowering.

zygoloid:
For types using a pointer value representation, emit the address of the source rather than loading the value.

### DIFF
--- a/toolchain/lowering/lowering_handle.cpp
+++ b/toolchain/lowering/lowering_handle.cpp
@@ -190,10 +190,10 @@ auto LoweringHandleCall(LoweringFunctionContext& context, SemIR::NodeId node_id,
 
   if (llvm_function->getReturnType()->isVoidTy()) {
     context.builder().CreateCall(llvm_function, args);
-    // TODO: A function with a void return type shouldn't be referred to by
-    // other nodes.
+    // The value of a function call with a void return type shouldn't used, but
+    // StubReference needs a value to propagate.
     context.SetLocal(node_id,
-                     llvm::UndefValue::get(context.GetType(node.type_id())));
+                     llvm::PoisonValue::get(context.GetType(node.type_id())));
   } else {
     context.SetLocal(node_id,
                      context.builder().CreateCall(llvm_function, args,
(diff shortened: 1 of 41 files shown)


toolchain/lowering/lowering_handle_expression_category.cpp
```diff
@@ -8,10 +8,27 @@ namespace Carbon {
 
 auto LoweringHandleBindValue(LoweringFunctionContext& context,
                              SemIR::NodeId node_id, SemIR::Node node) -> void {
-  // TODO: Take the value representation into account here.
-  context.SetLocal(node_id, context.builder().CreateLoad(
-                                context.GetType(node.type_id()),
-                                context.GetLocal(node.GetAsBindValue())));
+  auto source_id = node.GetAsBindValue();
```
jonmeow: I'll note, this feels a little odd to have out here since it's not used on all the paths, and is short (could be inlined).

toolchain/lowering/lowering_handle.cpp
```diff
@@ -220,10 +220,10 @@ auto LoweringHandleCall(LoweringFunctionContext& context, SemIR::NodeId node_id,
 
   if (llvm_function->getReturnType()->isVoidTy()) {
     context.builder().CreateCall(llvm_function, args);
-    // TODO: A function with a void return type shouldn't be referred to by
-    // other nodes.
+    // The value of a function call with a void return type shouldn't used, but
+    // StubReference needs a value to propagate.
```
jonmeow, line 236: Maybe StubReference of `()` should be special-cased not to propagate?

jonmeow (approved):
This change might be okay to merge as-is, but maybe we should still think about how StubReference of a void is being used?

toolchain/lowering/lowering_handle_expression_category.cpp
zygoloid: Done.

toolchain/lowering/lowering_handle.cpp
zygoloid: Yeah, I'll look into this in a follow-up.

PR #2511, opened 2023-01-04
Assignment statements

zygoloid:
Assignment is permitted only as a complete statement, not as a subexpression.
Assignment and compound assignment syntax follow C++ in all other respects.
Pre-increment is provided. Post-increment is not. Uses of all of these operators
are translated into calls to interface members.

PR #3149, opened 2023-08-24
Take value representation into account when lowering functions and calls.

zygoloid:
Pass function parameters directly or indirectly, depending on their value representation. For types such as empty tuples with an empty value representation, don't pass them at all.

### DIFF
--- a/toolchain/lowering/lowering_context.cpp
+++ b/toolchain/lowering/lowering_context.cpp
@@ -59,6 +59,7 @@ auto LoweringContext::BuildFunctionDeclaration(SemIR::FunctionId function_id)
     -> llvm::Function* {
   const auto& function = semantics_ir().GetFunction(function_id);
   const bool has_return_slot = function.return_slot_id.is_valid();
+  auto& param_refs = semantics_ir().GetNodeBlock(function.param_refs_id);
 
   SemIR::InitializingRepresentation return_rep =
       function.return_type_id.is_valid()
@@ -68,18 +69,20 @@ auto LoweringContext::BuildFunctionDeclaration(SemIR::FunctionId function_id)
                 .kind = SemIR::InitializingRepresentation::None};
   CARBON_CHECK(return_rep.has_return_slot() == has_return_slot);
 
-  // TODO: Lower type information for the arguments prior to building args.
-  auto param_refs = semantics_ir().GetNodeBlock(function.param_refs_id);
-  llvm::SmallVector<llvm::Type*> args;
-  llvm::SmallVector<SemIR::NodeId> param_nodes;
-  args.reserve(has_return_slot + param_refs.size());
-  param_nodes.reserve(has_return_slot + param_refs.size());
+  llvm::SmallVector<llvm::Type*> param_types;
+  // TODO: Consider either storing `param_node_ids` somewhere so that we can
+  // reuse it from `BuildFunctionDefinition` and when building calls, or factor
+  // out a mechanism to compute the mapping between parameters and arguments on
+  // demand.
+  llvm::SmallVector<SemIR::NodeId> param_node_ids;
+  param_types.reserve(has_return_slot + param_refs.size());
+  param_node_ids.reserve(has_return_slot + param_refs.size());
   if (has_return_slot) {
-    args.push_back(GetType(function.return_type_id)->getPointerTo());
-    param_nodes.push_back(function.return_slot_id);
+    param_types.push_back(GetType(function.return_type_id)->getPointerTo());
+    param_node_ids.push_back(function.return_slot_id);
   }
-  for (auto [i, param_ref] : llvm::enumerate(param_refs)) {
-    auto param_type_id = semantics_ir().GetNode(param_ref).type_id();
+  for (auto param_ref_id : param_refs) {
+    auto param_type_id = semantics_ir().GetNode(param_ref_id).type_id();
     switch (auto value_rep =
                 SemIR::GetValueRepresentation(semantics_ir(), param_type_id);
             value_rep.kind) {
@@ -87,12 +90,12 @@ auto LoweringContext::BuildFunctionDeclaration(SemIR::FunctionId function_id)
         break;
       case SemIR::ValueRepresentation::Copy:
       case SemIR::ValueRepresentation::Custom:
-        args.push_back(GetType(value_rep.type));
-        param_nodes.push_back(param_ref);
+        param_types.push_back(GetType(value_rep.type));
+        param_node_ids.push_back(param_ref_id);
         break;
       case SemIR::ValueRepresentation::Pointer:
-        args.push_back(GetType(value_rep.type)->getPointerTo());
-        param_nodes.push_back(param_ref);
+        param_types.push_back(GetType(value_rep.type)->getPointerTo());
+        param_node_ids.push_back(param_ref_id);
         break;
     }
   }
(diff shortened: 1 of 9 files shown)


toolchain/lowering/lowering_context.cpp
```diff
@@ -72,13 +71,30 @@ auto LoweringContext::BuildFunctionDeclaration(SemIR::FunctionId function_id)
...
   auto param_refs = semantics_ir().GetNodeBlock(function.param_refs_id);
   llvm::SmallVector<llvm::Type*> args;
-  args.resize_for_overwrite(first_param + param_refs.size());
+  llvm::SmallVector<SemIR::NodeId> param_nodes;
+  args.reserve(has_return_slot + param_refs.size());
+  param_nodes.reserve(has_return_slot + param_refs.size());
   if (has_return_slot) {
-    args[0] = GetType(function.return_type_id)->getPointerTo();
+    args.push_back(GetType(function.return_type_id)->getPointerTo());
+    param_nodes.push_back(function.return_slot_id);
```
jonmeow: Is there a test case that handles where there is a return slot and some parameters with no value representation? Would that be worth having?

toolchain/lowering/lowering_context.cpp
```diff
@@ -94,18 +110,18 @@ auto LoweringContext::BuildFunctionDeclaration(SemIR::FunctionId function_id)
...
-
-  // Set parameter names.
-  for (auto [i, param_ref] : llvm::enumerate(param_refs)) {
-    auto name_id = semantics_ir().GetNode(param_ref).GetAsParameter();
-    llvm_function->getArg(first_param + i)
-        ->setName(semantics_ir().GetString(name_id));
+  // Set up parameters.
+  for (auto [i, node_id, arg] : llvm::enumerate(
+           param_nodes,
+           map_range(llvm_function->args(), [](auto& arg) { return &arg; }))) {
```
jonmeow: I'm surprised, why does this need the map_range + lambda? I'm wondering if, given you have `i` anyways, continuing with `getArg(i)` might be a better choice for performance if it's needed. But, maybe it's something we can deal with if it's visible on benchmarks, which it might not be anyways.

Either way, I think you should choose one of:

1. `llvm::enumerate(param_nodes)`, use `getArg(i)`.
2. `llvm::zip_equal(param_nodes, llvm_function->args())`: isn't `i` unused in this approach, so no need to enumerate?

toolchain/lowering/lowering_context.cpp
```diff
@@ -72,13 +71,30 @@ auto LoweringContext::BuildFunctionDeclaration(SemIR::FunctionId function_id)
   // TODO: Lower type information for the arguments prior to building args.
   auto param_refs = semantics_ir().GetNodeBlock(function.param_refs_id);
   llvm::SmallVector<llvm::Type*> args;
-  args.resize_for_overwrite(first_param + param_refs.size());
+  llvm::SmallVector<SemIR::NodeId> param_nodes;
```
jonmeow: Is `args` a mistaken choice of names here? Maybe `params` or `param_types` would've been better? It's particularly the `args` versus `param_nodes` that makes it feel like a mismatch to me, though I'm sure `llvm::Function` using `Arguments` influenced the original naming...

toolchain/lowering/lowering_context.cpp
```diff
@@ -94,18 +110,18 @@ auto LoweringContext::BuildFunctionDeclaration(SemIR::FunctionId function_id)
...
+           param_nodes,
+           map_range(llvm_function->args(), [](auto& arg) { return &arg; }))) {
+    auto node = semantics_ir().GetNode(node_id);
+    if (node.kind() == SemIR::NodeKind::Parameter) {
+      arg->setName(semantics_ir().GetString(node.GetAsParameter()));
+    } else if (node_id == function.return_slot_id) {
+      arg->setName("return");
+      arg->addAttr(llvm::Attribute::getWithStructRetType(
+          llvm_context(), GetType(function.return_type_id)));
+    }
```
jonmeow: For the `else if`, what's the remaining case?

Could this be refactored to:

```
if (node_id == function.return_slot_id) {
  ...
} else {
  // Removes the GetNode for the return slot.
  auto node = semantics_ir().GetNode(node_id);
  ...
}
```

toolchain/lowering/lowering_context.cpp
```diff
@@ -124,17 +140,27 @@ auto LoweringContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
...
 
   const bool has_return_slot = function.return_slot_id.is_valid();
-  const int first_param = has_return_slot ? 1 : 0;
 
   // Add parameters to locals.
+  // TODO: This duplicates the mapping between semantics nodes and LLVM
+  // function parameters that was already computed in BuildFunctionDeclaration.
+  // We should only do that once.
   auto param_refs = semantics_ir().GetNodeBlock(function.param_refs_id);
+  int param_index = 0;
```
jonmeow, line 151: The purpose of the enumerate on 154 was presumably to handle this, is the enumerate still needed under this approach?

toolchain/lowering/lowering_context.cpp
```diff
@@ -124,17 +140,27 @@ auto LoweringContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
...
   if (has_return_slot) {
     function_lowering.SetLocal(function.return_slot_id,
-                               llvm_function->getArg(0));
+                               llvm_function->getArg(param_index++));
```
jonmeow: Per proposal #2511, Carbon won't have post-increment. WDYT of writing with Carbon's limitations, which wouldn't allow this?

toolchain/lowering/lowering_context.cpp
```diff
@@ -72,13 +71,30 @@ auto LoweringContext::BuildFunctionDeclaration(SemIR::FunctionId function_id)
...
-  args.resize_for_overwrite(first_param + param_refs.size());
+  llvm::SmallVector<SemIR::NodeId> param_nodes;
+  args.reserve(has_return_slot + param_refs.size());
+  param_nodes.reserve(has_return_slot + param_refs.size());
   if (has_return_slot) {
-    args[0] = GetType(function.return_type_id)->getPointerTo();
+    args.push_back(GetType(function.return_type_id)->getPointerTo());
+    param_nodes.push_back(function.return_slot_id);
   }
   for (auto [i, param_ref] : llvm::enumerate(param_refs)) {
```
jonmeow: Is this enumerate still needed?

jonmeow (commented):
It looks like you're changing loop contents in ways that affect what's being iterated over, I think there are still some updates that should be done here as a result.

toolchain/lowering/lowering_context.cpp
jonmeow: It might be worth calling this `param_node_ids` to emphasize that it's an id rather than a node.

toolchain/lowering/lowering_context.cpp
```diff
@@ -94,18 +110,18 @@ auto LoweringContext::BuildFunctionDeclaration(SemIR::FunctionId function_id)
...
-    return_slot->setName("return");
-  }
-
-  // Set parameter names.
-  for (auto [i, param_ref] : llvm::enumerate(param_refs)) {
-    auto name_id = semantics_ir().GetNode(param_ref).GetAsParameter();
-    llvm_function->getArg(first_param + i)
-        ->setName(semantics_ir().GetString(name_id));
+  // Set up parameters.
+  for (auto [i, node_id, arg] : llvm::enumerate(
```
jonmeow: I was thinking after the first round, is param_nodes efficient here?

IIUC, you're using it to rapidly iterate through the params a second time, versus mirroring the logic on 77-98 to again only take actions conditionally. TBH, I'm not sure what's better: repeating those branches, or having a vector allocation (but also still branches in this loop).

I'm not sure if you have a strong instinct about the best approach here, but maybe it's worth a TODO to see if removing param_nodes and doing a similar param_index as is done in BuildFunctionDefinition would have an impact on performance? Or maybe similarly, storing param_nodes with the Function for use by BuildFunctionDefinition?

toolchain/lowering/lowering_context.cpp
zygoloid: Hm, I'd convinced myself that the `zip` functions wouldn't work here because they'd make a copy of the `Argument`s, but it looks like that's only the case for `zip_longest`. Simplified, thanks!

toolchain/lowering/lowering_context.cpp
zygoloid: Added a TODO. As the mapping between Carbon parameters and LLVM parameters evolves, storing the mapping rather than computing it on the fly should be simpler and less error-prone, but yeah, this is potentially allocating memory we don't need if the `SmallVector` gets too big.

toolchain/lowering/lowering_context.cpp
zygoloid: Done.

toolchain/lowering/lowering_context.cpp
zygoloid: It's not. Removed, thanks.

toolchain/lowering/lowering_context.cpp
zygoloid: Seems reasonable, done.

toolchain/lowering/lowering_context.cpp
zygoloid: No, removed, thanks.

toolchain/lowering/lowering_context.cpp
zygoloid: Yeah, I'm not sure there's an ideal solution here, but renaming to `param_types` seems better. Done.

It's unfortunate that there are two different prevailing sets of terminology here -- "parameter" versus "argument", and "formal argument" versus "actual argument" -- and that LLVM picks the latter convention and shortens the terms to `Argument` and `Arg` respectively (!!), whereas language specifications (including Carbon's) usually use the former convention for clarity.

zygoloid: Done.

toolchain/lowering/lowering_context.cpp
zygoloid: Sure, added as `toolchain/lowering/testdata/function/call/tuple_param_with_return_slot.carbon`.

jonmeow (approved):
Looks good! Feel free to merge when you're ready.

PR #3150, opened 2023-08-24
Add a script to scan source code for basic stats.

chandlerc:
This is a rough script that uses regexes to do a simple scan of source code and extract some basic source code statistics. Things like column width, comment line density, identifier lengths and densities.

After scanning, it prints out both raw stats and in a few cases renders a quick histogram to the terminal to help visualize a relevant distribution.

I threw this together pretty quickly, and this is an area of Python I have very limited familiarity with, so happy to have any suggestions for how to better approach this.

### DIFF
--- a/scripts/source_stats.py
+++ b/scripts/source_stats.py
@@ -1,0 +1,187 @@
+#!/usr/bin/env python3
+
+"""Script to compute statistics about source code."""
+
+from __future__ import annotations
+
+__copyright__ = """
+Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+Exceptions. See /LICENSE for license information.
+SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+"""
+
+import argparse
+from alive_progress import alive_bar  # type:ignore
+from multiprocessing import Pool
+import re
+import termplotlib as tpl  # type:ignore
+from pathlib import Path
+from typing import Dict, List, Optional
+from dataclasses import dataclass, field, asdict
+
+BLANK_RE = re.compile(r"\s*")
+COMMENT_RE = re.compile(r"\s*///*\s*")
+LINE_RE = re.compile(
+    r"""
+    (?P<class_intro>\b(class|struct)\s+(?P<class_name>\w+)\b)|
+    (?P<end_open_curly>{\s*(?P<open_curly_trailing_comment>//.*)?)|
+    (?P<trailing_comment>//.*)|
+    (?P<id>\b\w+\b)
+""",
+    re.X,
+)
+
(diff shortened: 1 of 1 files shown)


scripts/source_stats.py
```diff
@@ -0,0 +1,187 @@
...
+        if width in self.identifier_widths:
+            self.identifier_widths[width] += 1
+        else:
+            self.identifier_widths[width] = 1
+
+    def add_ids_per_line(self, num_ids: int) -> None:
+        if num_ids in self.ids_per_line:
+            self.ids_per_line[num_ids] += 1
+        else:
+            self.ids_per_line[num_ids] = 1
```
josh11b: Usual way to do this is using https://docs.python.org/3/library/stdtypes.html#dict.get , https://docs.python.org/3/library/collections.html#collections.Counter , or `setdefault` like you do below.

scripts/source_stats.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def accumulate(self, other: _Stats) -> None:
+        self.lines += other.lines
+        self.blank_lines += other.blank_lines
+        self.empty_comment_lines += other.empty_comment_lines
+        self.comment_lines += other.comment_lines
+        for width, count in other.comment_line_widths.items():
+            self.comment_line_widths.setdefault(width, 0)
+            self.comment_line_widths[width] += count
```
josh11b: Could call `self.add_comment_line_width(width, count)` if you gave that method an optional `count` parameter that defaults to `1`.

scripts/source_stats.py
```diff
@@ -0,0 +1,187 @@
...
+    (?P<end_open_curly>{\s*(?P<open_curly_trailing_comment>//.*)?)|
+    (?P<trailing_comment>//.*)|
+    (?P<id>\b\w+\b)
+""",
+    re.X,
+)
+
+
+@dataclass
+class _Stats:
```
josh11b: Using `_` prefix seems like extra noise for a script like this.

chandlerc: Done.

scripts/source_stats.py
chandlerc: Ah, `Counter` is perfect. I somehow didn't find that when trying to figure out how to do this in Python, and ended up making a really bad version of `Counter` here, so thanks.

scripts/source_stats.py
chandlerc: Obviated now that I'm using `Counter`.

chandlerc (commented):
PTAL!

josh11b (approved):

PR #3152, opened 2023-08-24
Refactor index logic to share more code.

jonmeow:
Trying to improve diagnostic names and code flow. I was thinking "operand" might be a better name than "name" in the context of what `a` is in `a[b]`, where we already call `b` the `index`.

Flow-wise, note this replaces pushing BuiltinError with instead creating Index nodes that may contain an error. This is intended to aim towards producing a more consistent SemIR in the face of errors. There are pros and cons of both approaches, but I'm aiming for simpler control flow.


toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -15,79 +15,89 @@ auto HandleIndexExpressionStart(Context& /*context*/,
...
   auto index_node_id = context.node_stack().PopExpression();
   auto index_node = context.semantics_ir().GetNode(index_node_id);
-  auto name_node_id = context.node_stack().PopExpression();
-  name_node_id = context.MaterializeIfInitializing(name_node_id);
-  auto name_node = context.semantics_ir().GetNode(name_node_id);
-  auto name_type_id =
-      context.semantics_ir().GetTypeAllowBuiltinTypes(name_node.type_id());
-  auto name_type_node = context.semantics_ir().GetNode(name_type_id);
+  auto operand_node_id = context.node_stack().PopExpression();
+  operand_node_id = context.MaterializeIfInitializing(operand_node_id);
```
chandlerc, line 45: I'm OK with `operand`, but another candidate would be `indexed`, looking like:

```suggestion
  auto indexed_node_id = context.node_stack().PopExpression();
  indexed_node_id = context.MaterializeIfInitializing(indexed_node_id);
```

While I might prefer `indexed`/`index` vs `operand`/`index` by a slight margin, its really slight, so go with whichever works best for you.

toolchain/semantics/semantics_handle_index.cpp
```diff
@@ -15,79 +15,89 @@ auto HandleIndexExpressionStart(Context& /*context*/,
...
+  switch (operand_type_node.kind()) {
+    case SemIR::NodeKind::ArrayType: {
+      auto [bound_id, element_type_id] = operand_type_node.GetAsArrayType();
+      // We can check whether integers are in-bounds, although it doesn't affect
+      // the IR for an array.
+      if (index_node.kind() == SemIR::NodeKind::IntegerLiteral) {
+        ValidateIntegerLiteralBound(
+            context, parse_node, operand_node, index_node,
+            context.semantics_ir().GetArrayBoundValue(bound_id));
+      }
```
chandlerc: Curious why not produce an error in the SemIR for this?

One thing I could imagine wanting is to add some way of modelling errors that doesn't *remove* the SemIR information so that we can find the `-1` index operation. If that's the direction you're imagining, maybe leave a TODO?

chandlerc (approved):
Nice diagnostic improvements. A question inline, but may just be a TODO. Feel free to land if so (or if I've just misunderstood something).

toolchain/semantics/semantics_handle_index.cpp
jonmeow: I'd started with `indexed`, but it's visually very close to `index` and was hard for me to read. `operand` was my choice to make it easy to see one versus the other.

toolchain/semantics/semantics_handle_index.cpp
jonmeow: The diagnostic is sufficient to mark the IR as invalid.

It sounds though like you really want the IR to specifically include an error, so I've changed code to add logic that replaces the index node id with an error.

toolchain/semantics/semantics_handle_index.cpp
chandlerc: Yeah. Long-term, I think we want somehow to model both of these: that it *is* an error, and the IR that would have been there anyways. Right now, putting in the error breaks the connection in the IR.

The reason I think we'll want this is for tools that are trying to reason about the SemIR in the face of errors (for example to suggest fixes, etc) -- I think they'll want a bit more information.

But either way (with the error none in your latest push, or the original form) is fine for now, this isn't blocking at all.

chandlerc (approved):
(still LG)

toolchain/semantics/semantics_handle_index.cpp
jonmeow: SGTM, although I think you meant to put this comment on the other thread. ;)

chandlerc: Doh, yes, sorry.

PR #3153, opened 2023-08-24
Switch index arrays to StringRef arrays

jonmeow:
I was thinking this might be simpler than the index approach without creating undue overhead, and it nicely allows use of llvm::join.


chandlerc (approved):
:sniff: at my beautiful use of `std::iota`. ;]

Nice though, especially where I was awkwardly switching between the kinds. Better to just render first.

Issue #3070, opened 2023-08-06
Restructuring & namespacing plan for the toolchain


chandlerc:
I wanted to capture a discussion that several of us had around a restructuring of the toolchain to give ourselves a bit of a better directory layout and nomenclature.

Overall, the plan is to move towards a structure of major components where there is a reasonable namespace we can use to have a regular name component of code within that component. That is, rather than `SemanticsNode`, we'll have something structurally more like `Semantics::Node`.

As part of that, we noted a couple of things:
- There are really two layers inside of the `semantics` directory currently: the semantics-IR, and the code to check a parse tree and build that IR from it.
- The names we're using aren't awesome, and that showed up when trying to give command line flags that reflect the various stages.

One of the challenges we have is that the command line wants verb-shaped names as they're about the _action_ of that phase. But forming nouns out of the verbs for the phases doesn't follow a consistent structure. For example, "parser" is great, but "lowerer" is really awkward. So our thought was to switch to naming things after the verb everywhere.

The outcome of this (if I remember correctly -- please correct if I've misremembered) was to move to the following rough name / structure.

- `lex` directory, and if needed a `Carbon::Lex` namespace
- `parse` directory, and if needed a `Carbon::Parse` namespace
- `check` directory for the semantic & type checking logic that diagnoses invalid Carbon code and produces a valid semantics IR, and `Carbon::Check` namespace
- `lower` directory and `Carbon::Lower` namespace
- `codegen` directory and `Carbon::Codegen` namespace, treating "codegen" as a word

We talked about what to name the component that provides the semanitics IR definition and APIs, but didn't come up with anything that I think folks were delighted by. Some of the candidates:
- `SIR` -- not great if pronounced "sir".
- `SemIR` -- maybe pronounced "simmer"?
- `SemaIR` -- the abbreviation "sema" in Clang hasn't been enjoyable, especially for non-native speakers
- `CarbonIR` -- unclear if this is an IR written *in* Carbon or *for* Carbon. Also, we might end up with more IRs if we want to do high level optimizations.

We didn't really talk about how we want to handle the components that aren't named after a phase of compilation. Personally, I think these are maybe fine to have subdirectories but no namespace as they will tend to be widely used. And I'm also OK with the current names we have there. But flagging it in case others disagree.

[...]

PR #3155, opened 2023-08-24
Move lowering into a 'Lower' namespace

jonmeow:
Splitting namespace changes from file renames with the thought that it'll be more likely to work with git's merge logic.

Renamed LoweringContext to FileContext to disambiguate more clearly from FunctionContext.

This is part of #3070


chandlerc (approved):
LG!

PR #3156, opened 2023-08-24
Modify EnumBase to better support the namespacing of toolchain

jonmeow:
The different approach to Names avoids the issues with trying to define a static member (or also member function) of the templated instance of Carbon::Internal::EnumBase from a non-enclosing namespace such as Carbon::SemIR.

Note, I'm trying to do this from the cpp file. An alternative might be to do `inline constexpr llvm::StringLiteral Names[]` in the .h file, but I think concerns had been raised about that needing deduplication.

### DIFF
--- a/common/enum_base.h
+++ b/common/enum_base.h
@@ -51,8 +51,7 @@ namespace Carbon::Internal {
 //   #include ".../my_kind.def"
 //   };
 //   ```
-using EnumBaseNamesT = const llvm::StringLiteral[];
-template <typename DerivedT, typename EnumT, EnumBaseNamesT Names>
+template <typename DerivedT, typename EnumT, const llvm::StringLiteral Names[]>
 class EnumBase {
  public:
   // An alias for the raw enum type. This is an implementation detail and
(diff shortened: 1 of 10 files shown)


common/enum_base.h
```diff
@@ -51,7 +51,8 @@ namespace Carbon::Internal {
 //   #include ".../my_kind.def"
 //   };
 //   ```
-template <typename DerivedT, typename EnumT>
+using EnumBaseNamesT = const llvm::StringLiteral[];
+template <typename DerivedT, typename EnumT, EnumBaseNamesT Names>
```
chandlerc: Having the `using` first makes the big comment block seem to pertain to the `using` rather than the class template. And I don't think we really use it -- what about just inlining the type?

```suggestion
template <typename DerivedT, typename EnumT, const llvm::StringLiteral Names[]>
```

You could also make it a pointer if that seems more clear (I believe that is how this syntax works in C++ anyways):

```suggestion
template <typename DerivedT, typename EnumT, const llvm::StringLiteral* Names>
```

chandlerc (approved):
Nice, I like just passing the names array in! Minor stylistic suggestion inline, LG to submit with either.

common/enum_base.h
jonmeow: Fixed (1st suggestion). I'd done this because I'd been getting a compile error, but I think I was just messing up my syntax and failed to realize.

PR #3157, opened 2023-08-25
Manual cleanup of toolchain clang-tidy/clangd warnings.

jonmeow:
A lot of this is more boring "remove unused header", plus some other minor cleanups. I think the most significant changes were:

- yaml_test_helpers.cpp is doing a switch on an unsigned int, comparing to enum values.
  - [bugprone-switch-missing-default-case](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/switch-missing-default-case.html) is unhappy with EnumBase, but correctly identified yaml_test_helpers.cpp, so I'm opting to address it rather than disabling it even though it needs NOLINT in several locations as a result, in addition to what I think are some low-value `default` cases. I'd be fine going the other way with this too and disabling it globally (I could see it being noisier in the explorer).
- MarkInitializerFor swaps the argument names between the .h and .cpp. I think the .cpp had the order as intended.
- There's a new-ish [performance-enum-size](https://clang.llvm.org/extra/clang-tidy/checks/performance/enum-size.html) which I'm basically treating as "add int8_t to enums".

My main motivation here is to just clean up as many of these as I can so that I stop seeing them in vscode.

### DIFF
--- a/toolchain/lexer/tokenized_buffer.cpp
+++ b/toolchain/lexer/tokenized_buffer.cpp
@@ -536,7 +536,7 @@ class TokenizedBuffer::Lexer {
       // Too short to form one of these tokens.
       return LexResult::NoMatch();
     }
-    if (word[1] < '1' || '9' < word[1]) {
+    if (word[1] < '1' || word[1] > '9') {
       // Doesn't start with a valid initial digit.
       return LexResult::NoMatch();
     }
--- a/toolchain/lexer/tokenized_buffer.cpp
+++ b/toolchain/lexer/tokenized_buffer.cpp
@@ -7,13 +7,9 @@
 #include <algorithm>
 #include <array>
 #include <cmath>
-#include <iterator>
-#include <string>
 
 #include "common/check.h"
 #include "common/string_helpers.h"
-#include "llvm/ADT/STLExtras.h"
-#include "llvm/ADT/Sequence.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/ADT/StringSwitch.h"
 #include "llvm/Support/ErrorHandling.h"
@@ -540,7 +536,7 @@ class TokenizedBuffer::Lexer {
       // Too short to form one of these tokens.
       return LexResult::NoMatch();
     }
-    if (!('1' <= word[1] && word[1] <= '9')) {
+    if (word[1] < '1' || '9' < word[1]) {
       // Doesn't start with a valid initial digit.
       return LexResult::NoMatch();
     }
@@ -683,6 +679,8 @@ class TokenizedBuffer::Lexer {
         case '\t':
         case '\n':
           return false;
+        default:
+          break;
       }
       return llvm::StringSwitch<bool>(llvm::StringRef(&c, 1))
 #define CARBON_SYMBOL_TOKEN(Name, Spelling) .StartsWith(Spelling, false)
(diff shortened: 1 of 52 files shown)


toolchain/lexer/tokenized_buffer.cpp
```diff
@@ -540,7 +536,7 @@ class TokenizedBuffer::Lexer {
       // Too short to form one of these tokens.
       return LexResult::NoMatch();
     }
-    if (!('1' <= word[1] && word[1] <= '9')) {
+    if (word[1] < '1' || '9' < word[1]) {
```
zygoloid: I find this hard to read, how about:

```suggestion
    if (word[1] < '1' || word[1] > '9') {
```

toolchain/lowering/lowering_context.cpp
```diff
@@ -147,6 +147,8 @@ auto LoweringContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
     for (const auto& node_id : semantics_ir().GetNodeBlock(block_id)) {
       auto node = semantics_ir().GetNode(node_id);
       CARBON_VLOG() << "Lowering " << node_id << ": " << node << "\n";
+      // clang warns on unhandled enum values; clang-tidy is incorrect here.
+      // NOLINTNEXTLINE(bugprone-switch-missing-default-case)
```
zygoloid, line 181: What's the issue here? Is this something we could fix upstream in the `clang-tidy` checker? (No objection to the `NOLINT` in the mean time.)

toolchain/semantics/semantics_handle_if_expression.cpp
```diff
@@ -25,20 +26,20 @@ auto HandleIfExpressionIf(Context& context, ParseTree::Node if_node) -> bool {
...
+auto HandleIfExpressionThen(Context& context, ParseTree::Node parse_node)
     -> bool {
   // Convert the first operand to a value.
   auto [then_value_node, then_value_id] =
       context.node_stack().PopExpressionWithParseNode();
   context.node_stack().Push(then_value_node,
                             context.ConvertToValueExpression(then_value_id));
 
-  context.node_stack().Push(then_node, context.node_block_stack().Pop());
+  context.node_stack().Push(parse_node, context.node_block_stack().Pop());
```
zygoloid: This change is understandable but unfortunate -- there are multiple different parse nodes in scope here, so having descriptive names for the `then_node` versus the `then_value_node` seems useful. Same thing happens in the next function, where we have an `if_node`, a `then_node`, and an `else_node` in scope before this change, which seems easier to understand than calling one of the three `parse_node`.

Is there some way we can turn off this checking for the `Handle` functions? If so, do you think we should?

zygoloid (approved):

toolchain/lowering/lowering_context.cpp
jonmeow: In case you missed this in the PR description:

> [bugprone-switch-missing-default-case](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/switch-missing-default-case.html) is unhappy with EnumBase, but correctly identified yaml_test_helpers.cpp, so I'm opting to address it rather than disabling it even though it needs NOLINT in several locations as a result, in addition to what I think are some low-value default cases. I'd be fine going the other way with this too and disabling it globally (I could see it being noisier in the explorer).

I don't know what the root cause is; I'm not familiar enough with Clang AST to quickly investigate and fix it.

toolchain/semantics/semantics_handle_if_expression.cpp
jonmeow: Sounds like you have a stronger preference than me. Trying to figure out a solution, NOLINTNEXTLINE and NOLINTBEGIN didn't seem to suppress this warning the way I'd expect, so how about this "alias" approach?

PR #3158, opened 2023-08-25
Update C++Now talks with video links

jonmeow:
The successor strategy talk isn't up yet, but will be on Sep 8 -- just trying to get them all in one pass.

Slides were previously linked, which I could keep, although it's weird to have a long link with just `(Slides)` at the end because it implies the whole linked text is slides (rather than a video). So if that's desired, I might switch instead to something like:

```
- Talk name, con ([video](youtube link), [slides](slides link))
```

But, this current format (video-only full-name link) felt consistent with the 2022 links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,15 +305,12 @@ Past Carbon focused talks from the community:
 
 ### 2023
 
-Press `S` when viewing slides to see speaker notes.
-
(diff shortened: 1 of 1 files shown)


chandlerc (approved):
LGTM!

We can also use a playlist or any other approach we like if we get one in the future.

PR #2550, opened 2023-01-24
Simplified package declaration for the `Main` package

zygoloid:
Make the preamble of simple programs more ergonomic, by removing the
`package Main` from the main package and removing the `package` declaration
entirely from the main source file. Imports within a single package no longer
need to, and are not permitted to, specify the package name.

Partially covers #2001 / #1136.
Covers #1869.
Supersedes #2265.
Addresses design idea #2323.

PR #3160, opened 2023-08-25
Update `member_access.md` to reflect accepted proposals

josh11b:
Most changes are due to proposal #2360, but this also includes changes to reflect: #1136, #2138, #2006, #2550, and #2964.


docs/design/expressions/member_access.md
```diff
@@ -136,57 +151,79 @@ fn CallMyFunction2() {
...
--   The first operand is a value, and lookup should consider members of the
-    value's type.
--   The first operand is a type, and lookup should consider members of that
-    type. For example, `i32.Least` should find the member constant `Least` of
-    the type `i32`.
--   The first operand is a type-of-type, and lookup should consider members of
-    that type-of-type. For example, `Addable.Add` should find the member
-    function `Add` of the interface `Addable`. Because a type-of-type is a type,
-    this is a special case of the previous bullet.
+### Types and facets
```
zygoloid, line 179: Should we explicitly say here something like: "if the type of the first operand is a facet type, including the case where it is `type`, the first operand is required to be a symbolic constant"? (This was in the old rules, in a somewhat implicit way -- "If the first operand can be evaluated and evaluates to a type" -- and I think we still need this in the new rules.)

docs/design/expressions/member_access.md
```diff
@@ -229,48 +287,98 @@ fn F[T:! type](x: GenericWrapper(T)) -> T {
...
 
-> **TODO:** The behavior of `G` above is not yet fully decided. If class
-> templates can be specialized, then we cannot know the members of
-> `TemplateWrapper(T)` without knowing `T`, so this first lookup will find
-> nothing. In any case, as described below, the lookup will be performed again
-> when `T` is known.
+If the value or type depends on any template binding, the lookup is redone from
+a context where the values of those binding are known, but where the values of
+any symbolic bindings are still unknown. The lookup results from these two
+contexts are [combined](#lookup-ambiguity).
```
zygoloid:
```suggestion
If the value or type depends on any template bindings, the lookup is redone from
a context where the values of those bindings are known, but where the values of
any symbolic bindings are still unknown. The lookup results from these two
contexts are [combined](#lookup-ambiguity).
```

docs/design/expressions/member_access.md
```diff
@@ -229,48 +287,98 @@ fn F[T:! type](x: GenericWrapper(T)) -> T {
...
+bindings that are in scope are unknown. Unlike for a template binding, the
+actual value of a symbolic binding never affects the result of member
 resolution.
 
+##### Lookup ambiguity
+
+Multiple lookups can be performed when resolving a member access expression with
+a [template binding](#constant-bindings). We resolve this the same way as when
+looking in multiple interfaces that are
+[combined with `&`](/docs/design/generics/details.md#combining-interfaces-by-anding-type-of-types):
```
zygoloid: I feel like this dependency edge is backwards. I think when we define member name lookup (here), we should say that if we look in more than one place, we combine the results in the way you describe below, and that should apply in all such cases -- combining template and instantiation lookups, combining lookup results from multiple `extend`s, and combining lookups with `&` (which I think should behave as if it desugars to `extend`).

Perhaps we should have a general rule somewhere at the top level in this document, and have this section on template lookup ambiguity and the generics details doc both refer to it. Having a dedicated section here with examples of the template lookup ambiguity issue seems valuable, though.

docs/design/expressions/member_access.md
```diff
@@ -110,6 +116,15 @@ A member access expression is processed using the following steps:
 The process of _member resolution_ determines which member `M` a member access
 expression is referring to.
 
+For a simple member access, the second operand is a word, which must name a
+member of the first operand.
```
zygoloid: For the "Values" case, the second operand can name a member of the type of the first operand instead. We could also take the opportunity to define what we're "searching" here, which we use below. Maybe:

```suggestion
For a simple member access, the second operand is a word. If the first operand is
a type, facet, package, or namespace, a search for the word is performed in the first
operand. Otherwise, a search for the word is performed in the type of the first
operand. In either case, the search must succeed.
```

docs/design/expressions/member_access.md
```diff
@@ -136,57 +151,79 @@ fn CallMyFunction2() {
...
-For example:
+Simple member access `(Avatar as Cowboy).Draw` finds the `Cowboy.Draw`
+implementation for `Avatar`, ignoring `Renderable.Draw`.
 
-```
+### Values
+
+If the first operand is not a type, package, namespace, or facet it does not
+have member names, and member lookup is performed into the type of the first
+operand instead.
```
zygoloid:
```suggestion
have member names, and a search is performed into the type of the first
operand instead.
```

Harmonizing on the "search" terminology used below.

docs/design/expressions/member_access.md
```diff
@@ -200,27 +237,48 @@ fn PrintPointTwice() {
...
   p.(Printable.Print)();
 }
+```
+
+### Facet binding
+
+If any of the above lookups ever looks for members of a
+[facet binding](/docs/design/generics/terminology.md#facet-binding), it should
+consider members of the facet type, treating the facet binding as an
+[archetype](/docs/design/generics/terminology.md#archetype).
```
zygoloid: It's not clear to me whether `impl` binding applies for such a lookup, or whether the "treating the facet binding as an archetype" means that we're looking up members of the eventual `impl`.

If we want `impl` binding to apply here, I think this would help:

```suggestion
If any of the above lookups would perform a search of a
[facet binding](/docs/design/generics/terminology.md#facet-binding), it
performs a search of the facet type instead, treating the facet binding as an
[archetype](/docs/design/generics/terminology.md#archetype).
```

docs/design/expressions/member_access.md
```diff
@@ -136,57 +151,79 @@ fn CallMyFunction2() {
...
-**Note:** If lookup is performed into a type that involves a template parameter,
-the lookup will be performed both in the context of the template definition and
-in the context of the template instantiation, as described in
-[templates and generics](#templates-and-generics).
+Unlike the previous case, both simple and compound member access is allowed.
 
-For a simple member access, the word is looked up in the following types:
+Facets, such as `T as Cowboy`, also have members. Specifically, the members of
+the `impl`, `T as Cowboy`. Being part of the `impl` rather than the interface,
+no further [`impl` lookup](#impl-lookup) is needed.
```
zygoloid: Hm, there isn't necessarily a single `impl` here -- `Cowboy` could be implemented by multiple different `impl` definitions if it's for example something like `Cow & Boy`, or the `impl` could be something like `impl T as (Cowboy & BalletDancer)`. Presumably we don't actually want to perform lookup into the `impl`, and find members of `BalletDancer`.

I think something like this might help, but I'm sure there's a better phrasing:

```suggestion
Facets, such as `T as Cowboy`, also have members. Specifically, the members of
the `impl` or `impl`s that form the implementation of `T as Cowboy`. Being part
of the `impl` rather than the interface, no further [`impl` lookup](#impl-lookup)
is needed.
```

docs/design/expressions/member_access.md
```diff
@@ -375,32 +491,58 @@ the member access expression, `V`:
...
+-   `Integer.AliasForAdd` finds \#3, the `Add` member of the facet type
+    `Integer as Addable`, not \#1, the interface member `Addable.Add`.
+-   `my_int.AliasForAdd`, with `my_int: Integer`, finds \#3, the `Add` member of
+    the facet type `Integer as Addable`, and performs
+    [instance binding](#instance-binding) since the member is an instance
+    member.
+-   `Addable.AliasForSum` finds \#2, the member in the interface `Addable`, and
+    does not perform `impl` lookup.
+
+FIXME: Is the following example still needed?
```
zygoloid: I think we don't need it.

docs/design/expressions/member_access.md
```diff
@@ -334,39 +465,24 @@ class Integer {
...
 }
 ```
 
 The type `T` that is expected to implement `I` depends on the first operand of
 the member access expression, `V`:
 
--   If `V` can be evaluated and evaluates to a type, then `T` is `V`.
+-   If `V` can be evaluated and evaluates to a
+    [type](/docs/design/generics/terminology.md#types-and-type), then `T` is
+    `V`.
```
zygoloid: I think we need some more edits here -- this doesn't appear to be up to date with #2360, and if it were correct, would be repeating what we said above. In terms of the flow of this section, I think it might be best to delete all the text from "For the following examples," inclusive down to "The appropriate `impl T as I` implementation is located" exclusive, and move the examples to after "`M` is replaced by the member of the `impl`".

docs/design/expressions/member_access.md
```diff
@@ -509,21 +642,110 @@ fn DrawTriangle(t: TriangleWidget) {
...
+  // converted to `Interface`, and so uses `MyClass as Interface`.
+  MyClass.(Interface.NonInstanceInterfaceMember)();
+
+  // ❌ This is an error unless `my_value` implicitly converts to
+  // a type.
+  my_value.(Interface.NonInstanceInterfaceMember)();
+}
+```
+
+FIXME: Maybe instead continue the previous example?
```
zygoloid: Seems like a good idea to me.

docs/design/expressions/member_access.md
```diff
@@ -509,21 +642,110 @@ fn DrawTriangle(t: TriangleWidget) {
...
+of `x` and instance binding is performed if an instance member is found.
+
+If instance binding is performed:
 
 -   For a field member in class `C`, `V` is required to be of type `C` or of a
     type derived from `C`. The result is the corresponding subobject within `V`.
-    The result is an lvalue if `V` is an lvalue.
+    The result is a
+    [reference expression](/docs/design/values.md#reference-expressions) if `V`
+    is a reference expression.
```
zygoloid: I think the rule we want here post-#2006 is: if `v` is an initializing expression, then a temporary is materialized. The result has the same expression category as the possibly-materialized `v`.

That is:
- value -> value
- durable reference -> durable reference
- ephemeral reference -> ephemeral reference
- initializing -> ephemeral reference

I don't know that #2006 actually specifies this, though.

docs/design/expressions/member_access.md
```diff
@@ -551,6 +773,36 @@ value other than a type, then _instance binding_ is performed, as follows:
...
+  fn Print[self:Self]();
+}
+impl i32 as DebugPrint;
+impl type as DebugPrint;
+
+fn Debug() {
+  var i: i32 = 1;
+
+  // Prints `1` using `(i32 as DebugPrint).Print` bound to `i`.
+  i.(DebugPrintable.Print)();
```
zygoloid: What is `DebugPrintable`? Is that intended to be the interface name?

docs/design/expressions/member_access.md
josh11b: Meaning it can't be a template constant? Otherwise, aren't all facet types compile-time constants?

docs/design/expressions/member_access.md
josh11b: "perform a search of" isn't super clear to me.  Maybe "perform a search into"?

docs/design/expressions/member_access.md
josh11b: When I wrote this, I noticed there was a difference between the cases. This is trying to say treat "multiple lookups finding the same name" similarly to "one lookup finds the same name in multiple places." Maybe that isn't an important distinction? The general question of how we resolve finding the same name in multiple places isn't fully resolved, see for example #2745 . It also isn't clear if this is the right document for the broader question, which might better be addressed by https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/name_lookup.md .

docs/design/expressions/member_access.md
josh11b: I think #2360 mostly changed the behavior of compound member access, and left alone. It still seems to me that for simple member access we need to say how to figure out the type used for impl lookup, which I don't think is really covered in this section. Or are you saying it is implied by figuring out the type in an earlier section?

docs/design/expressions/member_access.md
josh11b: Deleted.

docs/design/expressions/member_access.md
josh11b: Was supposed to be `DebugPrint`. Changed.

docs/design/expressions/member_access.md
josh11b: I updated the text.

docs/design/expressions/member_access.md
josh11b: Switched.

docs/design/expressions/member_access.md
```diff
@@ -110,6 +116,17 @@ A member access expression is processed using the following steps:
 The process of _member resolution_ determines which member `M` a member access
 expression is referring to.
 
+For a simple member access, the second operand is a word. If the first operand
+is a type, facet, package, or namespace, a search for the word is performed in
+the first operand. Otherwise, a search for the word is performed in the type of
+the first operand. In either case, the search must succeed.
```
zygoloid: Perhaps add something about instance binding here:
```suggestion
the first operand. In either case, the search must succeed. In the latter case,
if the result is an instance member, then [instance binding](#instance-binding)
is performed on the first operand.
```
This is redundant given the description later in this file, but makes this paragraph better match the parallel description of compound member accesses in the next paragraph.

docs/design/expressions/member_access.md
```diff
@@ -110,6 +116,17 @@ A member access expression is processed using the following steps:
...
+
+For a compound member access, the second operand is evaluated as a compile-time
+constant to determine the member being accessed. The evaluation is required to
+succeed and to result in a member of a facet, type, or interface. If the result
```
zygoloid: Perhaps:
```suggestion
succeed and to result in a member of a type, interface, or non-type facet. If the result
```
... to reduce the overlap between cases, as described in person. (That still leaves overlap between "type" and "interface" though.)

docs/design/expressions/member_access.md
zygoloid: Oops, I meant "required to be a compile-time constant" (of either kind). The case I'm concerned about is:
```
fn F(t: type) {
  t.F();
}
```
... where `t` is not a compile-time constant. I don't think we disallow this currently, and we seemed to be leaning away from compile-time-only types, which I think means we instead want member access into a type to be a compile-time-only operation.

docs/design/expressions/member_access.md
zygoloid: Or maybe just "searches", eg:
```suggestion
If any of the above lookups would search for members of a
[facet binding](/docs/design/generics/terminology.md#facet-binding), it
searches the facet type instead, treating the facet binding as an
[archetype](/docs/design/generics/terminology.md#archetype).
```

Though I do wonder: should we instead say that a search into a facet binding `T:! C` performs a search into the facet `T as C`? Is that different, except in terms of when we perform an `impl` lookup? Notionally I think we want to reuse the `impl`(s) that we discovered for the facet binding rather than doing a fresh `impl` lookup in this case, even though the outcome should be the same.

docs/design/expressions/member_access.md
zygoloid: Lines 440-452 define `T` ; this is the old rule that that new text replaces.

docs/design/expressions/member_access.md
zygoloid: OK, I'm fine with deferring any change here until the rules are more resolved.

docs/design/expressions/member_access.md
josh11b: I think I still need the example since it will still be referenced, so I just moved it lower.

docs/design/expressions/member_access.md
josh11b: Done.

docs/design/expressions/member_access.md
josh11b: Done

zygoloid (approved):
Looks great, thanks!

PR #3161, opened 2023-08-28
Set up a 'Parse' namespace.

jonmeow:
Continuing on #3070.

I moved ParseTree::Node to just Parse::Node, versus Parse::Tree::Node. Other name changes are just removing "Parse" or "Parser" prefixes.

In EnumBase, I'm directly defining operator<< because the ostream.h approach just isn't working, not for either of Parse::State nor Parse::NodeKind. Errors look like:

```toolchain/parser/parser_context.cpp:449:34: error: invalid operands to binary expression ('llvm::raw_ostream' and 'const Carbon::Parse::State')
    output << "\t" << i << ".\t" << entry.state;
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~
```

The expected template in `Carbon::` is not in the error list; I only see the:

```
./common/ostream.h:112:6: note: candidate template ignored: requirement 'std::is_base_of_v<std::ostream, llvm::raw_ostream>' was not satisfied [with S = llvm::raw_ostream, T = Carbon::Parse::State]
auto operator<<(S& standard_out, const T& value) -> S& {
     ^
```

I'm still prodding at this, but not seeing an obvious fix.


common/enum_base.h
```diff
@@ -83,6 +83,14 @@ class EnumBase {
   // Prints this value using its name.
   auto Print(llvm::raw_ostream& out) const -> void { out << name(); }
 
+  // TODO: common/ostream.h tries to provide this, but does not reliably work on
+  // EnumBase children. Try to find a more generic solution.
```
zygoloid, line 87: FWIW, I think what's happening here is that, after this change, `Carbon` is no longer an associated namespace for types like `Parse::NodeKind`, because neither that type nor its base class are in namespace `Carbon` any more.

toolchain/driver/driver.cpp
```diff
@@ -436,7 +437,7 @@ auto Driver::Compile(const CompileOptions& options) -> bool {
   }
 
   const SemIR::File builtin_ir = SemIR::File::MakeBuiltinIR();
-  CARBON_VLOG() << "*** SemanticsIR::MakeFromParseTree ***\n";
+  CARBON_VLOG() << "*** SemanticsIR::MakeFromParse::Tree ***\n";
```
zygoloid:
```suggestion
  CARBON_VLOG() << "*** SemanticsIR::MakeFromParseTree ***\n";
```

toolchain/driver/driver.cpp
```diff
@@ -446,7 +447,7 @@ auto Driver::Compile(const CompileOptions& options) -> bool {
   consumer->Flush();
 
   has_errors |= semantics_ir.has_errors();
-  CARBON_VLOG() << "*** SemIR::File::MakeFromParseTree done ***\n";
+  CARBON_VLOG() << "*** SemIR::File::MakeFromParse::Tree done ***\n";
```
zygoloid:
```suggestion
  CARBON_VLOG() << "*** SemIR::File::MakeFromParseTree done ***\n";
```

zygoloid (approved):

common/enum_base.h
jonmeow: Yeah, I was thinking it might be something like that, but then why does it work for something like SemIR::Function? I think I just don't understand the lookup well enough...

jonmeow: (or why don't other SemIR enums hit the exact same issue?)

zygoloid: It also matters where the call to `operator<<` is -- if the call is happening from within namespace `Carbon` or something nested within it, at a point where the templated `operator<<` is visible to unqualified lookup and not hidden by any other `operator<<` in an inner namespace, then that outer operator will be found. So this will depend on whether there are other `operator<<`s in the inner namespace, whether the use of `<<` lives in a template, and if the `<<` is in a template, whether `common/ostream.h` was included before the template was defined.

Our approach here was very fragile, and we're sadly seeing the consequences of that. :(

PR #3162, opened 2023-08-28
Reduce ambiguity in terminology

josh11b:
Change terminology away from terms that are ambiguous:

-   Reserve "generic type" for types with (compile-time) parameters, like `Vector` in `Vector(T:! type)`. Don't use that term to refer to `T`, as it would with [#2360](https://github.com/carbon-language/carbon-lang/blob/trunk/proposals/p2360.md#terminology).
-   Use the term "compile-time" instead of "constant" to mean "template or symbolic." Expand the term "constant" to include values, such as from `let` bindings.


chandlerc (approved):
LG!

I'm going ahead and approving as this seems especially uncontroversial and something we can also easily adjust in the future if something ends up not working. The core change makes lots of sense and has been discussed without any real concerns being raised.

PR #3166, opened 2023-08-29
Provide a Printable CRTP parent to replace HasPrintable templates.

jonmeow:
With the toolchain splitting namespaces, ostream.h's `operator<<` templates aren't reliably found with name lookup, likely due to the loss of associated namespaces (zygoloid commented on this at https://github.com/carbon-language/carbon-lang/pull/3161#discussion_r1307941999). This is especially a barrier to moving the lex files into `Carbon::Lex`; versus other parts of the toolchain, they contain more printable types which are used cross-namespace, including `Carbon::Testing`. As a consequence, I'm looking at migrating ostream.h to a more reliable approach that doesn't rely as much on everything being in the `Carbon` namespace.


explorer/ast/value.h
```diff
@@ -1466,7 +1467,7 @@ class ParameterizedEntityName : public Value {
 // These values are used to represent the second operand of a compound member
 // access expression: `x.(A.B)`, and can also be the value of an alias
 // declaration, but cannot be used in most other contexts.
-class MemberName : public Value {
+class MemberName : public Value, Printable<MemberName> {
```
zygoloid: Do you need the base class here? I'd expect the printing functionality to get inherited from the base class `Value`.

explorer/interpreter/heap.h
```diff
@@ -19,7 +19,7 @@
 namespace Carbon {
 
 // A Heap represents the abstract machine's dynamically allocated memory.
-class Heap : public HeapAllocationInterface {
+class Heap : public Printable<Heap, HeapAllocationInterface> {
```
zygoloid:
```suggestion
class Heap : public HeapAllocationInterface, public Printable<Heap> {
```
Can we directly inherit instead of chaining through `Printable` here? That'd be easier to reason about, despite introducing multiple inheritance. (I think this doesn't violate the ideal of doing things like we would in Carbon, because in Carbon we'd model `Printable` as a mixin.)

explorer/ast/value.h
```diff
@@ -1062,7 +1061,9 @@ struct ImplsConstraint : public HashFromDecompose<ImplsConstraint> {
 };
 
 // A constraint that requires an intrinsic property of a type.
-struct IntrinsicConstraint : public HashFromDecompose<IntrinsicConstraint> {
+struct IntrinsicConstraint
+    : public Printable<IntrinsicConstraint,
+                       HashFromDecompose<IntrinsicConstraint>> {
```
zygoloid: I believe this is necessary to support `ConstructFromComponents` in `value_transform.h`, which was made to support aggregate initialization of a type with a single CRTP base class, but not a type with multiple of them.

In this case, I think we could add a constructor to `IntrinsicConstraint` instead of modeling it as an aggregate to avoid this chaining of CRTP bases.

explorer/ast/value.h
jonmeow: Note it's not virtual:
```
void Print(llvm::raw_ostream& out) const { member_->Print(out); }
```

explorer/interpreter/heap.h
jonmeow: Yes, I just hadn't undone this case. For me, there's a question of how valuable consistency would be.

(note there's a typo in your suggestion, extra `>`)

explorer/ast/value.h
jonmeow: Done

explorer/ast/value.h
zygoloid: Oh no, and the base class version does something different from the derived class version too! Yuck!

zygoloid: #3168

common/ostream.h
```diff
@@ -7,60 +7,46 @@
...
-
-// Support std::ostream << for types which implement:
-//   void Print(llvm::raw_ostream& out) const;
-template <typename T, typename = std::enable_if_t<HasPrintMethod<T>>>
-auto operator<<(std::ostream& out, const T& obj) -> std::ostream& {
-  llvm::raw_os_ostream raw_os(out);
-  obj.Print(raw_os);
-  return out;
-}
+template <typename DerivedT>
```
geoffromer, line 19:
```suggestion
// CRTP base class for printable types. `DerivedT` must define a
// `Print` method that takes an `llvm::raw_ostream` argument, and
// must inherit from `Printable<DerivedT>`.
template <typename DerivedT>
```

zygoloid (commented):
LGTM, leaving to @geoffromer to approve.

common/ostream.h
jonmeow: Thanks, going with a slightly edited comment:

```
// CRTP base class for printable types. Children (DerivedT) must implement:
// - auto Print(llvm::raw_ostream& out) -> void
```

(the dash is mainly for clang-format, to avoid wrapping the signature)

geoffromer (approved):

PR #3170, opened 2023-08-30
Moving lexer into a 'Lex' namespace.

jonmeow:
Continuing with #3070, this creates a `Lex` namespace for lexer. This is probably the last namespace addition right now, and will be followed by file moves, although I'll also share a PR separately moving member classes out of TokenizedBuffer.

### DIFF
--- a/toolchain/lexer/numeric_literal.h
+++ b/toolchain/lexer/numeric_literal.h
@@ -12,7 +12,7 @@
 #include "llvm/ADT/StringRef.h"
 #include "toolchain/diagnostics/diagnostic_emitter.h"
 
-namespace Carbon {
+namespace Carbon::Lex {
 
 // A numeric literal token that has been extracted from a source buffer.
 class LexedNumericLiteral {
@@ -71,6 +71,6 @@ class LexedNumericLiteral {
   int exponent_;
 };
 
-}  // namespace Carbon
+}  // namespace Carbon::Lex
 
 #endif  // CARBON_TOOLCHAIN_LEXER_NUMERIC_LITERAL_H_
--- a/toolchain/lexer/numeric_literal_test.cpp
+++ b/toolchain/lexer/numeric_literal_test.cpp
@@ -14,6 +14,7 @@
 namespace Carbon::Testing {
 namespace {
 
+using Lex::LexedNumericLiteral;
 using ::testing::_;
 using ::testing::Field;
 using ::testing::Matcher;
--- a/toolchain/parser/parser_context.cpp
+++ b/toolchain/parser/parser_context.cpp
@@ -49,10 +49,10 @@ Context::Context(Tree& tree, Lex::TokenizedBuffer& tokens,
       vlog_stream_(vlog_stream),
       position_(tokens_->tokens().begin()),
       end_(tokens_->tokens().end()) {
-  CARBON_CHECK(position_ != end_) << "Empty Lex::TokenizedBuffer";
+  CARBON_CHECK(position_ != end_) << "Empty TokenizedBuffer";
   --end_;
   CARBON_CHECK(tokens_->GetKind(*end_) == Lex::TokenKind::EndOfFile)
-      << "Lex::TokenizedBuffer should end with EndOfFile, ended with "
+      << "TokenizedBuffer should end with EndOfFile, ended with "
       << tokens_->GetKind(*end_);
 }
 
--- a/toolchain/parser/parser_context.cpp
+++ b/toolchain/parser/parser_context.cpp
@@ -40,8 +40,8 @@ static auto operator<<(llvm::raw_ostream& out, RelativeLocation loc)
   return out;
 }
 
-Context::Context(Tree& tree, TokenizedBuffer& tokens,
-                 TokenDiagnosticEmitter& emitter,
+Context::Context(Tree& tree, Lex::TokenizedBuffer& tokens,
+                 Lex::TokenDiagnosticEmitter& emitter,
                  llvm::raw_ostream* vlog_stream)
     : tree_(&tree),
       tokens_(&tokens),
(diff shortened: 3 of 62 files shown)


jonmeow:
Note, this is a stacked PR and only the final commit needs to be reviewed here.

toolchain/lexer/numeric_literal_test.cpp
```diff
@@ -14,6 +14,7 @@
 namespace Carbon::Testing {
 namespace {
 
+using Lex::LexedNumericLiteral;
```
geoffromer, line 17: Could we put the lexer tests in a `Carbon::Lex::Testing` namespace (or just `Carbon::Lex`; I'm not clear on what the `Testing` namespace adds), and avoid these `using` declarations? It can sometimes make sense to put the tests in a different namespace than the code under test, if the client is expected to be in a different namespace, but that doesn't seem to be the case here.

toolchain/lexer/numeric_literal.h
```diff
@@ -12,7 +12,7 @@
 #include "llvm/ADT/StringRef.h"
 #include "toolchain/diagnostics/diagnostic_emitter.h"
 
-namespace Carbon {
+namespace Carbon::Lex {
 
 // A numeric literal token that has been extracted from a source buffer.
 class LexedNumericLiteral {
```
geoffromer, line 18: Should we drop the `Lexed` prefix now that the namespace effectively says the same thing?

toolchain/parser/parser_context.cpp
```diff
@@ -40,23 +40,23 @@ static auto operator<<(llvm::raw_ostream& out, RelativeLocation loc)
...
+                 Lex::TokenDiagnosticEmitter& emitter,
                  llvm::raw_ostream* vlog_stream)
     : tree_(&tree),
       tokens_(&tokens),
       emitter_(&emitter),
       vlog_stream_(vlog_stream),
       position_(tokens_->tokens().begin()),
       end_(tokens_->tokens().end()) {
-  CARBON_CHECK(position_ != end_) << "Empty TokenizedBuffer";
+  CARBON_CHECK(position_ != end_) << "Empty Lex::TokenizedBuffer";
```
geoffromer: `Lex::` seems unnecessary here (sorry, GH won't let me express that as a suggestion)

toolchain/parser/parser_context.cpp
```diff
@@ -40,23 +40,23 @@ static auto operator<<(llvm::raw_ostream& out, RelativeLocation loc)
...
   --end_;
-  CARBON_CHECK(tokens_->GetKind(*end_) == TokenKind::EndOfFile)
-      << "TokenizedBuffer should end with EndOfFile, ended with "
+  CARBON_CHECK(tokens_->GetKind(*end_) == Lex::TokenKind::EndOfFile)
+      << "Lex::TokenizedBuffer should end with EndOfFile, ended with "
```
geoffromer: Here also

zygoloid (approved):

toolchain/parser/parser_context.cpp
```diff
@@ -230,37 +231,41 @@ auto Context::SkipPastLikelyEnd(TokenizedBuffer::Token skip_root)
...
 }
 
 // Determines whether the given token is considered to be the start of an
 // operand according to the rules for infix operator parsing.
-static auto IsAssumedStartOfOperand(TokenKind kind) -> bool {
-  return kind.IsOneOf({TokenKind::OpenParen, TokenKind::Identifier,
-                       TokenKind::IntegerLiteral, TokenKind::RealLiteral,
-                       TokenKind::StringLiteral});
+static auto IsAssumedStartOfOperand(Lex::TokenKind kind) -> bool {
+  return kind.IsOneOf({Lex::TokenKind::OpenParen, Lex::TokenKind::Identifier,
```
geoffromer, line 245: Have you considered using `using` declarations instead of `Lex::` qualifying everything (not in .h files, of course)? That seems like it would be less noisy (not to mention making the diff more manageable).

toolchain/lexer/numeric_literal_test.cpp
jonmeow: We'd discussed this and decided on the approach as part of CommandLine discussion, and the resolution was to make tests look a bit like calling code. Here's some context:

https://discord.com/channels/655572317891461132/655578254970716160/1141155947901882500

toolchain/lexer/numeric_literal.h
jonmeow: I'm trying to keep this PR just for the namespace. See #3171 which is already doing this.

toolchain/parser/parser_context.cpp
jonmeow: In discussion around SemIR in particular, the leaning had been that having checking and lowering refer to members with SemIR:: (without using) was going to be clearer, as a trade-off to introducing namespaces which make it shorter when working _within_ the namespace. Things like `Parse::Node` and `SemIR::Node` would also make `using` problematic as a pattern, which was previously addressed with classes-as-namespace (`ParseTree::Node`) or prefixes (`SemanticsNode`).

It's that higher level direction which I'm trying to echo here with `Lex::`, eschewing `using` statements when called from other parts of the toolchain.

toolchain/lexer/numeric_literal_test.cpp
geoffromer: But that's my point: this makes the tests look _unlike_ the calling code, because `LexedNumericLiteral` is only ever used inside the `Lex` namespace, except in the tests.

chandlerc: As came up in that thread -- using `Carbon::Lex::Testing` creates a bunch of problems because then any common Carbon testing utilities won't actually be in the namespace at all.

The `Testing` isn't to isolate the test (the anonymous namespace does that), it is to access testing utilities, and those have to live in a single namespace.

Given the tradeoffs, I'm not sure there is a significantly better namespace to use here, and the using declarations seem reasonably minimal?

jonmeow: Another trade-offs I might add on top of chandlerc's comments is that we would definitely have `Carbon::Testing` regardless -- for testing utilities, as noted -- and so one of the stated leanings was to have only that.

@geoffromer I'm blocking on your comment, but noting the trade-offs and discussion on #toolchain (and existing code in the style), would you consider it okay to merge? (Changes to other tests are beyond the scope of this PR, but could be cleaned up pursuant to further discussion of the style)

geoffromer: I still don't really understand the tradeoffs being made here, but yeah, I'll take this to Discord.

geoffromer (approved):
Looks like there are some merge conflicts to resolve

PR #3171, opened 2023-08-30
Rename some classes in Lex namespace

jonmeow:
Versus the namespace change in #3170, these felt like they may get more nuanced review, so I'm splitting them out:

- Removing "Lexed" from LexedNumericLiteral and LexedStringLiteral
- Moving classes out of the TokenizedBuffer class, so that they're now things like Lex::Token instead of Lex::TokenizedBuffer::Token (i.e., much shorter to type, more consistent with things like Parse::Node).

### DIFF
--- a/toolchain/lexer/tokenized_buffer.h
+++ b/toolchain/lexer/tokenized_buffer.h
@@ -123,39 +123,24 @@ class TokenIterator
 // This is either a dyadic fraction (mantissa * 2^exponent) or a decadic
 // fraction (mantissa * 10^exponent).
 //
-// The `TokenizedBuffer` must outlive any `RealLiteralValue`s referring to
-// its tokens.
+// `RealLiteralValue` carries a reference back to `TokenizedBuffer` which can be
+// invalidated if the buffer is edited or destroyed.
 class RealLiteralValue : public Printable<RealLiteralValue> {
  public:
+  auto Print(llvm::raw_ostream& output_stream) const -> void {
+    output_stream << mantissa << "*" << (is_decimal ? "10" : "2") << "^"
+                  << exponent;
+  }
+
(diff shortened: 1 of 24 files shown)


jonmeow:
Note, this is a stacked PR and only the final commit needs to be reviewed here.

toolchain/lexer/tokenized_buffer.h
```diff
@@ -23,161 +23,165 @@
...
-        : buffer_(buffer) {}
+  RealLiteralValue(const llvm::SmallVector<llvm::APInt>* literal_int_storage,
+                   int32_t literal_index, bool is_decimal)
+      : literal_int_storage_(literal_int_storage),
+        literal_index_(literal_index),
+        is_decimal_(is_decimal) {}
+
+  const llvm::SmallVector<llvm::APInt>* literal_int_storage_;
+  int32_t literal_index_;
+  bool is_decimal_;
```
zygoloid: Can we store an `ArrayRef<llvm::APInt>` or maybe a pair of `llvm::APInt*`s here, pointing directly to the storage for this value, rather than a vector pointer plus an index?

I think invalidation through reallocation of the `storage` array is not a problem here, because `RealLiteralValue`s are only created after we've already finished forming the `TokenizedBuffer`, at which point it's immutable.

zygoloid (commented):
These changes all look like improvements to me, thank you for your attention to detail!

toolchain/lexer/tokenized_buffer.h
jonmeow: Done. Switched to a struct since we're looking at this, hoping that's okay based on discussion (it falls out if it's just a pair of llvm::APInt*)

zygoloid (approved):

PR #3173, opened 2023-08-30
Retain information about how an initializing expression is used in semantics IR

zygoloid:
Switch from a fully lowering-oriented approach to initializing expressions in semantics IR to an approach that retains more information about the high-level semantics of the program. Specifically:

- Always form an `Assign` node for a variable initialization or assignment, even for empty types and when the actual initialization is performed in-place by evaluating the right-hand side.
- Always form a `ReturnExpression` node for a return statement with an expression, even when the actual initialization is performed in-place into the return slot and the function doesn't return a value.

This is intended to permit us to more easily write semantic checks over the IR, and to preserve a use edge from the use of an initializing expression and its initializer in all cases.

To better support this, `MaterializeTemporary` is also split into two nodes:

- A `TemporaryStorage` node allocates and provides the storage for a temporary, and is emitted prior to emitting the initialization of the temporary.
- A `Temporary` node connects the temporary storage to the initializer, and is emitted after emitting the initialization.

Example testcase:

```carbon
fn F() -> (i32, i32);

fn G() -> (i32, i32) {
  var v: (i32, i32) = F();
  v = F();
  return F();
}

fn H() -> i32 {
  return G()[0];
}
```

Before:

```carbon-semir
fn @G() -> %return: (i32, i32) {
  // ...
  %v: (i32, i32) = var "v"
  %.loc4_24: (i32, i32) = call @F() to %v
  %.loc5: (i32, i32) = call @F() to %v
  %.loc6: (i32, i32) = call @F() to %return
  return
}

fn @H() -> i32 {
!entry:
  %.loc10_11.1: (i32, i32) = materialize_temporary
  %.loc10_11.2: (i32, i32) = call @G() to %.loc10_11.1
  %.loc10_14: i32 = int_literal 0
  %.loc10_15.1: i32 = tuple_index %.loc10_11.1, %.loc10_14
  %.loc10_15.2: i32 = bind_value %.loc10_15.1
  return %.loc10_15.2
}
```

After:

```carbon-semir
fn @G() -> %return: (i32, i32) {
  // ...
  %v: (i32, i32) = var "v"
  %.loc4_24: (i32, i32) = call @F() to %v
  assign %v, %.loc4_24
  %.loc5: (i32, i32) = call @F() to %v
  assign %v, %.loc5
  %.loc6: (i32, i32) = call @F() to %return
  return %.loc6
}

fn @H() -> i32 {
!entry:
  %.loc10_11.1: (i32, i32) = temporary_storage
  %.loc10_11.2: (i32, i32) = call @G() to %.loc10_11.1
  %.loc10_14: i32 = int_literal 0
  %.loc10_11.3: (i32, i32) = temporary %.loc10_11.1, %.loc10_11.2
  %.loc10_15.1: i32 = tuple_index %.loc10_11.3, %.loc10_14
  %.loc10_15.2: i32 = bind_value %.loc10_15.1
  return %.loc10_15.2
}
```

Note that in the "Before" code, there is no indication in `G()` that we assigned to `v`, and no path through "use" edges from the tuple indexing in `H()` to the call to `G()`.

This change results in our not performing `memcpy`s into the destination for tuple and struct initializers. This is a consequence of those initializers not yet performing in-place initialization, and will be fixed in a subsequent change.

### DIFF
--- a/toolchain/lowering/lowering_function_context.h
+++ b/toolchain/lowering/lowering_function_context.h
@@ -100,12 +100,6 @@ class FunctionContext {
   auto FinishInitialization(SemIR::TypeId type_id, SemIR::NodeId dest_id,
                             SemIR::NodeId init_id) -> void;
 
-  // Emit a value copy for type `type_id` from `source_id` to `dest_id`.
-  // `source_id` must produce a value representation for `type_id`, and
-  // `dest_id` must be a pointer to a `type_id` object.
-  auto CopyValue(SemIR::TypeId type_id, SemIR::NodeId source_id,
-                 SemIR::NodeId dest_id) -> void;
-
   auto llvm_context() -> llvm::LLVMContext& {
     return file_context_->llvm_context();
   }
@@ -116,6 +110,12 @@ class FunctionContext {
   }
 
  private:
+  // Emits a value copy for type `type_id` from `source_id` to `dest_id`.
+  // `source_id` must produce a value representation for `type_id`, and
+  // `dest_id` must be a pointer to a `type_id` object.
+  auto CopyValue(SemIR::TypeId type_id, SemIR::NodeId source_id,
+                 SemIR::NodeId dest_id) -> void;
+
   // Context for the overall lowering process.
   FileContext* file_context_;
 
--- a/toolchain/lowering/lowering_function_context.h
+++ b/toolchain/lowering/lowering_function_context.h
@@ -93,6 +93,19 @@ class FunctionContext {
     return synthetic_block_ == block;
   }
 
+  // After emitting an initializer `init_id`, finishes performing the
+  // initialization of `dest_id` from that initializer. This is a no-op if the
+  // initialization was performed in-place, and otherwise performs a store or a
+  // copy.
+  auto FinishInitialization(SemIR::TypeId type_id, SemIR::NodeId dest_id,
+                            SemIR::NodeId init_id) -> void;
+
+  // Emit a value copy for type `type_id` from `source_id` to `dest_id`.
+  // `source_id` must produce a value representation for `type_id`, and
+  // `dest_id` must be a pointer to a `type_id` object.
+  auto CopyValue(SemIR::TypeId type_id, SemIR::NodeId source_id,
+                 SemIR::NodeId dest_id) -> void;
+
   auto llvm_context() -> llvm::LLVMContext& {
     return file_context_->llvm_context();
   }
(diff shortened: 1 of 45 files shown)


toolchain/lowering/lowering_function_context.h
```diff
@@ -93,6 +93,19 @@ class FunctionContext {
...
+  // initialization was performed in-place, and otherwise performs a store or a
+  // copy.
+  auto FinishInitialization(SemIR::TypeId type_id, SemIR::NodeId dest_id,
+                            SemIR::NodeId init_id) -> void;
+
+  // Emit a value copy for type `type_id` from `source_id` to `dest_id`.
+  // `source_id` must produce a value representation for `type_id`, and
+  // `dest_id` must be a pointer to a `type_id` object.
+  auto CopyValue(SemIR::TypeId type_id, SemIR::NodeId source_id,
+                 SemIR::NodeId dest_id) -> void;
```
jonmeow: Should this be private? It looks like it might just be an implementation detail of FinishInit, although maybe that's temporary.

toolchain/lowering/lowering_handle_expression_category.cpp
```diff
@@ -30,11 +30,18 @@ auto HandleBindValue(FunctionContext& context, SemIR::NodeId node_id,
...
                                               nullptr, "temp"));
 }
 
+auto HandleTemporary(FunctionContext& context, SemIR::NodeId node_id,
+                     SemIR::Node node) -> void {
+  auto [temporary_id, init_id] = node.GetAsTemporary();
+  context.FinishInitialization(node.type_id(), temporary_id, init_id);
+  context.SetLocal(node_id, context.GetLocal(temporary_id));
+}
+
```
jonmeow:
```suggestion
auto HandleTemporary(FunctionContext& context, SemIR::NodeId node_id,
                     SemIR::Node node) -> void {
  auto [temporary_id, init_id] = node.GetAsTemporary();
  context.FinishInitialization(node.type_id(), temporary_id, init_id);
  context.SetLocal(node_id, context.GetLocal(temporary_id));
}

auto HandleTemporaryStorage(FunctionContext& context, SemIR::NodeId node_id,
                            SemIR::Node node) -> void {
  context.SetLocal(
      node_id, context.builder().CreateAlloca(context.GetType(node.type_id()),
                                              nullptr, "temp"));
}

```

Sorry, I'm nitpicking ordering. :-P

jonmeow (approved):

jonmeow:
LGTM, just a couple quick little comments.

toolchain/lowering/lowering_function_context.h
zygoloid: Done. (If we end up wanting to use it more generally it's trivial to make it public again.)

PR #3175, opened 2023-08-30
Include the expression category in the semantics IR dump.

zygoloid:
Describe the broad expression category (value / reference / initializing) in the semantics IR textual format. This is redundant information that can be calculated from the text that's already in the output, but seems useful for understanding the structure of the IR.


jonmeow (approved):
Looks great!

PR #2200, opened 2022-09-20
Template generics

josh11b:
Add template generics, with optional constraints but no [SFINAE](https://en.cppreference.com/w/cpp/language/sfinae), to Carbon. Template generics allows the compiler postpone type checking of expressions dependent on a template parameter until the function is called and the value of that parameter is known.

Example usage:

```carbon
fn Identity[template T:! Type](x: T) -> T {
  return x;
}
```

PR #3177, opened 2023-08-31
Clarify some points in Member Access design

josh11b:
Follow-on to #3160. Incorporates decision about template specialization from #2200.

### DIFF
--- a/docs/design/expressions/member_access.md
+++ b/docs/design/expressions/member_access.md
@@ -17,7 +17,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
     -   [Values](#values)
     -   [Facet binding](#facet-binding)
         -   [Compile-time bindings](#compile-time-bindings)
-            -   [Lookup ambiguity](#lookup-ambiguity)
+        -   [Lookup ambiguity](#lookup-ambiguity)
 -   [`impl` lookup](#impl-lookup)
     -   [`impl` lookup for simple member access](#impl-lookup-for-simple-member-access)
     -   [`impl` lookup for compound member access](#impl-lookup-for-compound-member-access)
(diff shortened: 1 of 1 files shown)


jonmeow (approved):

jonmeow (approved):
Still looks good.

jonmeow:
Noticed you explicitly requested zygoloid's review, so not merging.

docs/design/expressions/member_access.md
```diff
@@ -336,29 +336,29 @@ fn CallsDrawTemplate(c: Cowboy) {
...
+class TemplateWrapper(template T:! type) {
+  var field: T;
+}
+fn G[template T:! type](x: TemplateWrapper(T)) -> T {
+  // ✅ Allowed, finds `TemplateWrapper(T).field`.
+  return x.field;
+}
+```
+
+In addition, the lookup will be performed again when `T` is known.
```
zygoloid: I think it'd be useful to also include an example where this second lookup matters when performing a lookup into an instantiation of a class template, eg:

```carbon
class DerivingWrapper(template T:! type) {
  extend base: T;
}
fn H[template T:! type](x: DerivingWrapper(T), U:! type) -> U {
  // ✅ Allowed, but no name `field` found in template definition.
  return x.field;
}
fn CallH(a: DerivingWrapper(TemplateWrapper(i32)),
         b: DerivingWrapper(i32)) {
  // ✅ Member `field` in base class found in instantiation.
  var x: i32 = H(a, i32);
  // ❌ Error, no member `field` in type of `b`.
  var y: i32 = H(b, i32);
}
```

josh11b: Done.

zygoloid (approved):

PR #3178, opened 2023-08-31
Rename parser dir to parse

jonmeow:
Continuing with #3070. Just a dir and file rename (mostly removing prefixes, although for parse_tree_fuzzer and parse_tree_file_test I'm dropping "tree" instead of "parse"). Everything in the parse dir should be marked as a move.


chandlerc (approved):
LG

PR #3179, opened 2023-08-31
Rename lexer dir to lex

### LINKED ISSUE #3070: Restructuring & namespacing plan for the toolchain

jonmeow:
Continuing with #3070. Just a dir and file rename (only prefix change is lexer_file_test). Everything in the lex dir should be marked as a move.

Note, I think this closes #3070. There may still be further cleanup later, but the organizational changes suggested there are being completed.


toolchain/lex/lex_helpers.h
```diff
@@ -2,8 +2,8 @@
 // Exceptions. See /LICENSE for license information.
```
zygoloid: Would it make sense to rename this file to just "helpers.h"?

I'm happy for that to be done in a separate PR (if at all).

zygoloid (approved):

toolchain/lex/lex_helpers.h
jonmeow: Double-checked that "helpers.h" is still preferred with "test_helpers.h" here too. Renamed.

chandlerc:
The prior commit is in the merge queue. Trying out adding this one to see if it DTRT and resolves the merge.

chandlerc:
That would be "no, it won't".

Going to see if I can add it when the other commit lands.

chandlerc:
Not auto-resolved. I'm going to try to resolve the conflicts and merge this so that all the renames are landed and done without more waiting and/or churn tomorrow.

PR #1885, opened 2022-08-03
ranged-based `for` for user-defined types

Pixep:
The goal of this proposal is to provide a way for user-defined types to support range-based iteration with `for`.
The current proposed solution exposes 3 interfaces that can be implemented by user types to enable support for
ranged-for loops.

Co-authored-by: josh11b <josh11b@users.noreply.github.com>

PR #3180, opened 2023-08-31
Update design overview to reflect submitted proposals

josh11b:
Incorporates proposals: #1885, #2138, #2188, #2200, #2360, #2760, #2964, and #3162.


docs/design/README.md
```diff
@@ -395,14 +397,43 @@ has a number of consequences:
...
+>     [#2360: Types are values of type `type`](https://github.com/carbon-language/carbon-lang/pull/2360)
+
+### Values usable as types
+
+A value used in a type position, like after a `:` in a variable declaration or
+the return type after a `->` in a function declaration, must be:
+
+-   [a compile-time constant](#expression-phases), so the compiler can evaluate
+    it at compile time, and
+-   have a defined implicit conversion to type `type`.
```
zygoloid:
```suggestion
the return type after a `->` in a function declaration, must:

-   be [a compile-time constant](#expression-phases), so the compiler can evaluate
    it at compile time, and
-   have a defined implicit conversion to type `type`.
```

(Avoid "must be [...] have a" for the second bullet.)

docs/design/README.md
```diff
@@ -2707,33 +2753,20 @@ binding.
...
 constant whereas the expression phase of a template parameter is template
-constant.
+constant. A binding pattern using `:!` is a _compile-time binding pattern_; more
+specifically a _template binding pattern_ if it uses `template`, and a _symbolic
+binding pattern_ if it does not.
 
 Although checked generics are generally preferred, templates enable translation
 of code between C++ and Carbon, and address some cases where the type checking
-rigor of generics are problematic.
+rigor of checked generics are problematic.
```
zygoloid:
```suggestion
rigor of checked generics is problematic.
```
(Minor grammar fix since we're in the area.)

docs/design/README.md
```diff
@@ -2812,16 +2883,22 @@ by replacing the definition scope in curly braces (`{`...`}`) with a semicolon.
...
+> -   Proposal
+>     [#2360: Types are values of type `type`](https://github.com/carbon-language/carbon-lang/pull/2360)
+> -   Proposal
+>     [#2760: Consistent `class` and `interface` syntax](https://github.com/carbon-language/carbon-lang/pull/2760)
 
 ### Combining constraints
 
-A function can require calling types to implement multiple interfaces by
-combining them using an ampersand (`&`):
+A function can require calling types to implement multiple interfaces (or other
```
zygoloid: This is pre-existing, but I had some trouble understanding what a "calling type" is. Would "parameter types" or "type arguments" (or maybe more verbosely, "deduced or explicit type arguments") work here?

docs/design/README.md
```diff
@@ -2851,35 +2928,84 @@ fn DrawTies[T:! Renderable & GameResult](x: T) {
...
+  x.Draw();
+}
+```
+
+This allows a safe transition from template to checked generics. Constraints can
+be added incrementally, with the compiler verifying that the semantics stay the
+same. If adding the constraint would change which function gets called, an error
+is triggered, as in `ConstrainedTemplateDraw` from the example. Once all
+constraints have been added, removing the word `template` to switch to a checked
+parameter is safe.
```
zygoloid: Optional, might make this easier to read:
```suggestion
constraints have been added, it is safe to remove the word `template` to switch
to a checked parameter.
```

docs/design/README.md
```diff
@@ -3053,18 +3180,18 @@ Carbon generics have a number of other features, including:
     same data representation as an existing type, so you may cast between the
     two types, but can implement different interfaces or implement interfaces
     differently.
--   Additional requirements can be placed on the associated types of an
+-   Additional requirements can be placed on the associated constants of an
     interface using
     [`where` constraints](generics/details.md#where-constraints).
```
zygoloid: I think this is currently only possible for associated facets, because `where` produces a facet type. Maybe:
```suggestion
-   Additional requirements can be placed on the associated facets of an
    interface using
    [`where` constraints](generics/details.md#where-constraints).
```

docs/design/README.md
```diff
@@ -3053,18 +3180,18 @@ Carbon generics have a number of other features, including:
...
 -   [Implied constraints](generics/details.md#implied-constraints) allows some
```
zygoloid:
```suggestion
-   [Implied constraints](generics/details.md#implied-constraints) allow some
```
Not sure about this: is "implied constraints" a singular noun, short for "the implied constraints feature" or "the use of implied constraints" or similar, or the plural of "implied constraint"?

docs/design/README.md
```diff
@@ -3053,18 +3180,18 @@ Carbon generics have a number of other features, including:
...
     constraints to be deduced and omitted from a function signature.
--   [Dynamic erased types](generics/details.md#runtime-type-fields) can hold any
-    value with a type implementing an interface, and allows the functions in
-    that interface to be called using
+-   _Planned_ [dynamic erased types](generics/details.md#runtime-type-fields)
+    can hold any value with a type implementing an interface, and allows the
```
zygoloid:
```suggestion
    can hold any value with a type implementing an interface, and allow the
```

zygoloid (approved):

docs/design/README.md
josh11b: Went with "type arguments"

PR #3181, opened 2023-09-01
Fix toolchain glob_sh_run uses.

jonmeow:
These were missed when changing the command line.

Also brings -v output to parity with formatted IR printing, and removes a redundant flush.


chandlerc (approved):
Sorry these got missed, but LG!

PR #3182, opened 2023-09-01
Add support for compiling multiple files at once.

jonmeow:
Rearranges driver logic into CompilationUnits in order to associate artifacts from the various stages of compilation.

Note, I'm not totally sure what the right thing to do is for lower/codegen, so I'm just doing a rote change there for now that mirrors prior phases (this is all the code supports anyways, so is probably right for now regardless).

SourceBuffer error output is moved local for consistency with other steps, and so that it's less ambiguous whether the error should be expected to already include a filename.

### DIFF
--- a/toolchain/driver/driver.cpp
+++ b/toolchain/driver/driver.cpp
@@ -4,9 +4,13 @@
 
 #include "toolchain/driver/driver.h"
 
+#include <memory>
+#include <optional>
+
 #include "common/command_line.h"
 #include "common/vlog.h"
 #include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/ScopeExit.h"
 #include "llvm/ADT/StringExtras.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/IR/LLVMContext.h"
@@ -80,7 +84,7 @@ The input Carbon source file to compile.
         },
         [&](auto& arg_b) {
           arg_b.Required(true);
-          arg_b.Set(&input_file_name);
+          arg_b.Append(&input_file_names);
         });
 
     b.AddOneOfOption(
@@ -249,7 +253,7 @@ Dump the generated assembly to stdout after codegen.
   llvm::StringRef target;
 
   llvm::StringRef output_file_name;
-  llvm::StringRef input_file_name;
+  llvm::SmallVector<llvm::StringRef> input_file_names;
 
   bool asm_output = false;
   bool force_obj_output = false;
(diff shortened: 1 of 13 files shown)


toolchain/driver/driver.cpp
```diff
@@ -376,179 +380,268 @@ auto Driver::ValidateCompileOptions(const CompileOptions& options) const
...
-  CARBON_VLOG() << "tokenized_buffer: " << tokenized_source;
-  if (options.phase == Phase::Lex) {
-    consumer->Flush();
-    return !has_errors;
+
+  // Check the parse tree and produce SemIR. Returns true on success.
+  auto RunCheck(const SemIR::File& builtins) -> bool {
+    LogCall("Check::CheckParseTree", [&] {
+      sem_ir_ = Check::CheckParseTree(builtins, *tokens_, *parse_tree_,
+                                      *consumer_, vlog_stream_);
```
zygoloid, line 440: We're now performing move assignments of various types (well, inside the `optional<T>` it's probably a move construction). Is that sufficiently cheap that we don't need to worry about it? (There is a technique to avoid the move operation here if we need it.)

`SemIR::File` in particular seems like it might be quite a large class, due to the inline storage of its various vectors.

toolchain/driver/driver.cpp
```diff
@@ -376,179 +380,268 @@ auto Driver::ValidateCompileOptions(const CompileOptions& options) const
...
-      output_stream_ << "\n";
+  llvm::SmallVector<std::unique_ptr<CompilationUnit>> units;
+  auto flush = llvm::make_scope_exit([&]() {
+    // The diagnostics consumer must be flushed before compilation artifacts are
+    // destructed, because diagnostics can refer to their state. This ensures
+    // they're flushed in order of arguments, rather than order of destruction.
+    for (auto& unit : units) {
+      unit->Flush();
     }
+  });
```
zygoloid, line 590: Can we make the `CompilationUnit` destructor do this now? It looks like that would make things happen in the right order. And it seems to have the right logic, of flushing right before we destroy the various compilation artifacts.

Hm, but that would issue diagnostics in reverse file order, because the vector elements are destroyed right-to-left.

toolchain/driver/driver.cpp
```diff
@@ -376,179 +380,268 @@ auto Driver::ValidateCompileOptions(const CompileOptions& options) const
...
-  // function, as diagnostics can refer to state that lives on our stack.
-  std::unique_ptr<SortingDiagnosticConsumer> sorting_consumer;
-  if (vlog_stream_ == nullptr && !options.stream_errors) {
-    sorting_consumer = std::make_unique<SortingDiagnosticConsumer>(*consumer);
-    consumer = sorting_consumer.get();
+  // Loads source and lexes it. Returns true on success.
+  auto RunLex() -> bool {
+    LogCall("SourceBuffer::CreateFromFile", [&] {
+      source_ = SourceBuffer::CreateFromFile(
+          driver_->fs_, driver_->error_stream_, input_file_name_);
```
zygoloid, line 405: Perhaps we should be passing a diagnostic consumer to this, and producing a properly-formatted diagnostic on failures.

zygoloid (approved):
Some thoughts, but nothing blocking.

LGTM, feel free to submit as-is (not merging in case you want to respond to any of the comments first).

toolchain/driver/driver.cpp
```diff
@@ -376,179 +380,268 @@ auto Driver::ValidateCompileOptions(const CompileOptions& options) const
...
-  CARBON_VLOG() << "*** Lex::TokenizedBuffer::Lex done ***\n";
-  if (options.dump_tokens) {
-    CARBON_VLOG() << "Finishing output.";
-    consumer->Flush();
-    output_stream_ << tokenized_source;
+  // Parses tokens. Returns true on success.
+  auto RunParse() -> bool {
+    LogCall("Parse::Tree::Parse", [&] {
+      parse_tree_ = Parse::Tree::Parse(*tokens_, *consumer_, vlog_stream_);
+    });
```
chandlerc, line 427: Not sure how strongly you prefer this form, but I find it more work to read than the prior direct use of `CARBON_VLOG`. It's about the same number of lines (although one is *much* shorter), but this one I find harder to read because it uses so much language machinery like lambdas, etc.

toolchain/driver/driver.cpp
```diff
@@ -376,179 +380,268 @@ auto Driver::ValidateCompileOptions(const CompileOptions& options) const
...
 
-  if (vlog_stream_) {
-    CARBON_VLOG() << "*** semantics_ir ***\n";
-    SemIR::FormatFile(tokenized_source, parse_tree, semantics_ir,
-                      *vlog_stream_);
+  // Lex.
+  bool success_before_lower = true;
+  for (auto& unit : units) {
+    success_before_lower &= unit->RunLex();
   }
```
chandlerc, line 600: Rather than doing each phase for all files, I would find processing each file fully from left-to-right less surprising. Among other things, this would avoid interleaving errors from two phases of two files (at least in the streaming case -- not sure how the sorting diagnostic layer works with all of this).

Was there anything that made that not work well?

toolchain/driver/driver.cpp
chandlerc: FWIW, I mildly prefer either the API using `ErrorOr` as before and handling the streaming here, or as suggested leveraging our diagnostic engine here as well.

But not important to converge either way in this commit.

chandlerc (commented):
FWIW, similarly don't see anything as really strictly blocking. Happy for anything to be addressed in a subsequent change wherever easier.

toolchain/driver/driver.cpp
jonmeow: Rather than brevity, what got me to do this was actually seeing the mistakes in:

```
  CARBON_VLOG() << "*** SemanticsIR::MakeFromParseTree ***\n";
  const SemIR::File semantics_ir = Check::CheckParseTree(...);
  (various logic, including a flush)
  CARBON_VLOG() << "*** SemIR::File::MakeFromParseTree done ***\n";
```

Mechanically, this at least prevents _one_ of the name skews in the snippet. Visually, I hope it'd be more likely to catch the other, and might discourage putting the Flush inside. Also, it allows to insert a timer at very low cost (not in this PR, but I think it'd be good to add).

toolchain/driver/driver.cpp
jonmeow: I tried switching to diagnostics, but fell down a rathole of opaque runtime problems. I've added a TODO while trying to figure this out (I've spent a day on it already, and may want to restructure the diagnostic object model to make it function better).

toolchain/driver/driver.cpp
jonmeow: To be sure I understand correctly, "processing each file fully" means everything -- lexing through codegen.

Semantics needs to reorder processing based on output from the parsing layer (package and import instructions). I was expecting that lowering should not occur if there were any semantic issues, which we can't determine until all files are semantically processed. Similar for generating files: I would think it odd if files are created for _some_ files when some inputs don't pass checking. So I would think each of these phases cannot be joined with other phases.

We could in theory have lexing and parsing joined. However, since checking needs to be separated, it actually felt more _inconsistent_ to provide that. Both in terms of general functionality, but also just consistency in the control flow.

That should be somewhat transparent for users, though: the sorting consumer allows for output to be reordered at the file level. I believe users of the streaming consumer should expect the interspersed errors as it's consistent with normal output.

toolchain/driver/driver.cpp
jonmeow: Per discussion -- this may be something we want to address because SemIR::File could be a few kb, but it's not clear it's a performance bottleneck. I also brought up that we could use std::unique_ptr instead of optional, which would create heap allocations but would be consistent with the llvm::Module stuff and avoid the peculiar nature of the std::optional workaround. I think zygoloid decided though that this wasn't worth changing right now, easy enough to fix later if it's an issue.

toolchain/driver/driver.cpp
jonmeow: Yeah, it's the reverse file order that's the issue ("rather than order of destruction" in the comments)

PR #3183, opened 2023-09-01
Merge filesystem logic into FileTest and update test files.

jonmeow:
two_files.carbon has odd STDOUT placement, but will pursue that separately.


chandlerc (approved):
LG

PR #3184, opened 2023-09-02
autoupdate: improve handling of CHECK lines in multi-file input.

zygoloid:
Write unattached CHECK:STDOUT lines at the end of the complete test file, not at the end of the first split file.

Also, perform line number remappings for the current file even if we see a check line for an earlier file first. We used to stop performing remapping after the first check line that referred to a previous file.

To facilitate this, instead of splitting the check lines up by output file prior to forming the output, we instead form a single list of check lines and have the check lines track which file they refer to.


testing/file_test/file_test_base_test.cpp
```diff
@@ -62,32 +62,34 @@ class FileTestBaseTest : public FileTestBase {
...
+             << "unattached message 2\n";
+      stderr << "unattached message 3\n"
+             << "unattached message 4\n";
+      return true;
     } else {
       return ErrorBuilder() << "Unexpected file: " << filename;
     }
   }
 
-  auto ValidateRun(const llvm::SmallVector<TestFile>& test_files)
```
zygoloid: @jonmeow It wasn't clear to me how much value this extra validation was adding, and it seemed to be getting a little unmanageable as I added more multi-file tests. Is this OK to remove?

zygoloid:
I've split this change up so that you can see the effects of the code change on the newly-added tests by looking at only the second commit in the PR.

testing/file_test/autoupdate.cpp
```diff
@@ -110,16 +128,17 @@ static auto BuildCheckLines(
...
+    -> llvm::SmallVector<CheckLine> {
+  llvm::SmallVector<CheckLine> check_lines;
   if (output.empty()) {
     return check_lines;
   }
 
   // Prepare to look for filenames in lines.
-  llvm::StringRef current_filename = filenames[0];
-  const auto* remaining_filenames = filenames.begin() + 1;
+  llvm::StringMap<int> file_to_number_map;
```
chandlerc: FYI: `StringMap` isn't more efficient unless you need to allocate memory for the strings. Any time the backing string already exists and you could reasonably use `DenseMap<StringRef, int>`, I would expect it to be similar or faster than `StringMap`. The primary benefit of `StringMap` AFAICT is building a single shared string pool instead of separate heap allocations for each string.

testing/file_test/file_test_base_test.cpp
chandlerc: Just a note that PR #3183 moves its logic but continues to preserve the behavior.

testing/file_test/testdata/alternating_files.carbon
```diff
@@ -0,0 +1,24 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+// AUTOUPDATE
+// CHECK:STDOUT: 3 args: `default_args`, `a.carbon`, `b.carbon`
+// CHECK:STDOUT: unattached message 1
+// CHECK:STDERR: unattached message 1
```
chandlerc: The relative order of `STDOUT` and `STDERR` seems surprising to me here? But also maybe not as important compared to other changes...

chandlerc (commented):
While this makes sense to me, seems important for @jonmeow to review this so just leaving a few FYI comments.

testing/file_test/file_test_base_test.cpp
jonmeow: When I was adding split file logic (and as I've been modifying it), I need some way to validate that the contents were what I expected, since it's more nuanced than the default form (which is just "copy in the full file"). It's hard to examine the contents manually since they're only in memory, and indirect testing by library consumers could miss unintentional changes that might have subtle effects on test results. I think testing such as this needs to remain in order to validate that the intended splits are produced.

testing/file_test/autoupdate.cpp
```diff
@@ -222,30 +236,30 @@ auto AutoupdateFileTest(
...
     int output_line_number = 0;
-    auto* stdout_check_line = stdout_check_file.begin();
-    auto* stderr_check_line = stderr_check_file.begin();
 
     // Add all check lines from the given vector until we reach a check line
     // attached to a line later than `to_line_number`.
-    auto add_check_lines = [&](const llvm::SmallVector<CheckLine>& lines,
+    // Explicit capture of file_number is a workaround for the inability for
+    // lambdas to implicitly capture a structured binding.
+    auto add_check_lines = [&, file_number = static_cast<int>(file_number)](
```
jonmeow:
```suggestion
    // Work around the inability to implicitly capture a structured binding.
    auto file_number = file_number;
    auto add_check_lines = [&](
```

Just mulling, does this work (and do you find it any easier to read)?

testing/file_test/autoupdate.cpp
```diff
@@ -44,10 +47,14 @@ class CheckLine : public FileTestLineBase {
     out << indent_ << line_;
   }
 
+  int file_number() const { return file_number_; }
```
jonmeow: What do you think of moving this next to `is_blank`, grouping accessors?

testing/file_test/autoupdate.cpp
```diff
@@ -222,30 +236,30 @@ auto AutoupdateFileTest(
...
+    // Explicit capture of file_number is a workaround for the inability for
+    // lambdas to implicitly capture a structured binding.
+    auto add_check_lines = [&, file_number = static_cast<int>(file_number)](
+                               const llvm::SmallVector<CheckLine>& lines,
                                CheckLine*& line, int to_line_number,
                                llvm::StringRef indent) {
-      for (; line != lines.end() && line->line_number() <= to_line_number;
+      for (; line != lines.end() && line->file_number() <= file_number &&
+             (line->file_number() < file_number ||
+              line->line_number() <= to_line_number);
```
jonmeow:
```suggestion
      for (; line != lines.end() && (line->file_number() < file_number ||
             (line->file_number() == file_number &&
              line->line_number() <= to_line_number));
```

Sorry, this feels a little nitpicky, but what do you think of shifting the comparisons like so? I found myself trying to figure out the overlap of `<= file_number ... < file_number`.

testing/file_test/testdata/fail_example.carbon
```diff
@@ -3,5 +3,6 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 // AUTOUPDATE
-// CHECK:STDOUT: 2 args: `default_args`, `fail_example.carbon`
 // CHECK:STDERR: Oops
+
+// CHECK:STDOUT: 2 args: `default_args`, `fail_example.carbon`
```
jonmeow: Just to note, the blank line between stderr and stdout here still seems odd to me. Neither have a line association so I would've expected them to be grouped together. Maybe a TODO?

testing/file_test/autoupdate.cpp
```diff
@@ -278,53 +292,52 @@ auto AutoupdateFileTest(
...
     // This should always be true after the first file is processed.
     CARBON_CHECK(reached_autoupdate);
 
-    // Print remaining check lines which -- for whatever reason -- come after
-    // all original lines.
-    if (stderr_check_line != stderr_check_file.end() ||
-        stdout_check_line != stdout_check_file.end()) {
+    // At the end of the last file, print remaining check lines which -- for
+    // whatever reason -- come after all original lines.
+    if (file_number == filenames.size() - 1 &&
```
jonmeow: Just to note, with the shift in structure, I'm starting to think some edits could shift this and the RemapLineNumbers out of the loop.

testing/file_test/autoupdate.cpp
```diff
@@ -206,14 +216,18 @@ auto AutoupdateFileTest(
...
+  llvm::SmallVector<CheckLine> stdout_check_lines = BuildCheckLines(
+      stdout, "STDOUT", filenames, line_number_replacement.has_file,
+      line_number_re, do_extra_check_replacements);
+  llvm::SmallVector<CheckLine> stderr_check_lines = BuildCheckLines(
+      stderr, "STDERR", filenames, line_number_replacement.has_file,
+      line_number_re, do_extra_check_replacements);
+  auto* stdout_check_line = stdout_check_lines.begin();
+  auto* stderr_check_line = stderr_check_lines.begin();
+
+  bool any_attached_stdout_lines = std::any_of(
```
jonmeow, line 227: Some tests (lexer tests) will set line_number without file_number. As written, those get associated with file 0, but should probably be treated as unattached. Maybe a TODO? I'm fine fixing it, I think the TODO for handling multi-file remapping is also going to affect multi-file testing so I would probably be picking that up too.

jonmeow (commented):
Thanks! I'm happy to help address the multi-file TODOs.

testing/file_test/testdata/alternating_files.carbon
```diff
@@ -0,0 +1,24 @@
...
+// CHECK:STDERR: a.carbon:[[@LINE+1]]: message 2
+aaa
+// CHECK:STDOUT: a.carbon:[[@LINE-1]]: message 2
+
+// --- b.carbon
+// CHECK:STDERR: b.carbon:[[@LINE+4]]: message 3
+// CHECK:STDERR: a.carbon:2: message 4
+// CHECK:STDERR: b.carbon:[[@LINE+2]]: message 5
+// CHECK:STDERR: unattached message 6
+bbb
```
jonmeow, line 20:
```suggestion
bbb
bbbbbb
```

Sorry, I was thinking about this looking at the multifile testing, and wanted to suggest the edit as a really small nuance that should change how stdout is printed in what I think is a significant way.

It might be good too if you print something like `// CHECK:STDOUT: a.carbon:2: message 5.5`, since that presumably gets placed differently than the line sandwiched between two `@LINE` lines (but does have a line associated, even if not the current file).

testing/file_test/autoupdate.cpp
zygoloid: Done.

testing/file_test/autoupdate.cpp
zygoloid: Done.

(I wasn't worrying about efficiency here; for autoupdate I don't think it matters. This was an arbitrary choice, but `DenseMap` does seem like a better arbitrary choice...)

testing/file_test/autoupdate.cpp
zygoloid: My thinking is that the cases where we don't get a file name from the line number regex essentially don't support multi-file input (there's no way we could tell whether the CHECK is attached to the right file or not), and assuming that we're always referring to file 0 seems like the right behavior for tests that don't support multi-file input. See also the comment on line 191-192.

If the particular test fixture doesn't want line number remapping in that case, it could get that effect by not specifying a line number regex at all, and then all the lines would be treated as unattached and line numbers won't get remapped in the CHECK lines.

testing/file_test/autoupdate.cpp
zygoloid: No, that doesn't work, the variable is in scope in its own initializer. We would need to give it a different name, at least.

I've added a couple of suggestions, take  your pick :-)

testing/file_test/autoupdate.cpp
zygoloid: Moving `static_cast` to uses:

```suggestion
auto add_check_lines = [&, file_number = file_number](
                               const llvm::SmallVector<CheckLine>& lines,
                               CheckLine*& line, int to_line_number,
                               llvm::StringRef indent) {
      for (; line != lines.end() &&
             line->file_number() <= static_cast<int>(file_number) &&
             (line->file_number() < static_cast<int>(file_number) ||
              line->line_number() <= to_line_number);
```

testing/file_test/autoupdate.cpp
zygoloid:
```suggestion
  for (auto [file_number_as_size_t, filename, non_check_file] :
       llvm::enumerate(filenames, non_check_lines)) {
    auto file_number = static_cast<int>(file_number_as_size_t);
    llvm::DenseMap<int, int> output_line_remap;
    llvm::SmallVector<CheckLine*> check_lines_this_file;
    int output_line_number = 0;

    // Add all check lines from the given vector until we reach a check line
    // attached to a line later than `to_line_number`.
    auto add_check_lines = [&](
```

Potentially a bit sneaky, but we can side-step the lambda capture issue and the wrong-typed-index issue at the same time.

testing/file_test/autoupdate.cpp
zygoloid: Sure, done.

testing/file_test/autoupdate.cpp
zygoloid: Yes, I looked into that, but didn't want to change too much at once. (We'd need to track remappings for (file, line) pairs rather than just for lines, and be able to detect the past-the-end-of-file special case for each file.) I think that will eventually be necessary in order to handle the case where a single CHECK line refers to lines in multiple files.

In the interest of keeping this change smaller, I can send a follow-up PR for that if you like?

testing/file_test/testdata/alternating_files.carbon
zygoloid: Mm, yeah. What's happening here is that `STDOUT` is always attached after a line, and `STDERR` is always attached before a line, so we attach the early `STDOUT` to the `AUTOUPDATE` line itself and attach the `STDERR` to the line after the `AUTOUPDATE`.

As it happens, that's probably what we want in this test, because the early `STDOUT` is describing the input to the program, so logically precedes the `STDERR`, but it does violate the normal ordering. So, while this is surprising, is it bad? Do we want the other order? (Or maybe just a documentation update to say this is expected?)

testing/file_test/testdata/alternating_files.carbon
zygoloid: Added, but it doesn't push the rest of the STDOUT lines to the end of the file -- STDOUT lines following an attached line get placed next to that line, until we see a new, later line that we can use as an anchor.

I don't know that there's a best choice here.-We don't know whether "message 4" is more information attached to "message 3", in which case placing them together makes sense, or whether "message 4" is a new, unrelated diagnostic, in which case placing it at the end of the file might be preferable. But putting "message 4" right after "message 3" works better if there is a later diagnostic that can be put next to its corresponding line.

testing/file_test/testdata/fail_example.carbon
zygoloid: I *think* this is the output that I would want in this case: STDERR is at the top of the file, before the (blank line) content and (because there are no attached STDOUT lines) STDOUT is at the bottom of the file, after the (blank line) content.

But I don't feel strongly about it at all :) I think this oddity only shows up for a completely empty input -- or when AUTOUPDATE appears after all the content, which is a configuration that we probably don't need any more now that we put STDOUT at the end. (The explorer trace tests don't need the AUTOUPDATE at the end any more, at least.)

testing/file_test/file_test_base_test.cpp
zygoloid: Thanks, I think I understand what you're looking for here. I've put the split results into the stdout for this test so that we can still validate the splitting.

jonmeow: Agreed, thanks!

testing/file_test/autoupdate.cpp
jonmeow: To be sure, lines 191-192 only affect the lexer, and regardless, it requires changes to both the lexer and autoupdate to make that logic work for the lexer in multi-file tests.

testing/file_test/autoupdate.cpp
jonmeow: Sure, that's fine, didn't really think it made sense here since as you note the file mappings would be needed.

testing/file_test/testdata/fail_example.carbon
jonmeow: Maybe STDERR should just go at the end when there's no attached lines, same as STDOUT?

Anyways, not necessary to address here.

testing/file_test/testdata/alternating_files.carbon
jonmeow: I mean, this behavior is consistent with what we'd want for diagnostics, so maybe it's right. But I do think it's worth being explicit on it. :)

testing/file_test/autoupdate.cpp
jonmeow: Slight leaning for this one if it seems better for you. But really, I think if you prefer the current setup, I don't mean to push hard.

testing/file_test/testdata/alternating_files.carbon
jonmeow: FWIW my leaning would be to swap this if it's easy, since these tests aren't representative. Typically in the toolchain I'd be expecting stderr to be prioritized, although diagnostics would be associating lines with stderr so it'd be largely moot (and thus not worth too much time).

jonmeow (approved):
Looks good from my end.

testing/file_test/testdata/alternating_files.carbon
zygoloid: Done.

testing/file_test/testdata/alternating_files.carbon
zygoloid: I've made the documentation explicit about this.

testing/file_test/file_test_base_test.cpp
zygoloid: @jonmeow FYI, the code here changed a fair bit when merging in #3183, you might want to take another look to make sure you're happy before we merge.

PR #3192, opened 2023-09-04
Use high concurrency on CI build & test.

chandlerc:
The expensive local actions will be separately gated to not overwhelm the machine, and currently highly asynchronous actions are a dominant part of our builds due to downloading cached artifacts. Without a high concurrency, these are downloaded roughly 2-at-a-time currently.

I've verified that on a small machine without a good cache this doesn't seem to generate huge amounts of work and local build and test actions are successfully gated on the local flags.

There is already a Bazel issue tracking this limitation: https://github.com/bazelbuild/bazel/issues/6394

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -163,6 +163,24 @@ jobs:
           build --remote_cache=https://storage.googleapis.com/carbon-builds-github-v${CACHE_VERSION}-${{ env.os_for_cache }}
           build --google_credentials=$HOME/gcp-builds-service-account.json
 
+          # Set an artificially high jobs count. This flag controls the number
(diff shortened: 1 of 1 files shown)


chandlerc:
Just a note, this PR by itself seems to drop the actual build+test step in our CI from 4-10 minutes to 1.5-2 minutes, or something like 4x improvement.

I think #3106 and #3191 are likely still worth landing to make our CI do a minimal amount of work, but just wanted to be transparent that at least right now, the specific improvements those provide are likely minimal compared to this very simple change. Those changes may be more valuable as the project gets larger though.

jonmeow:
Interesting, good find!

jonmeow (approved):

PR #3197, opened 2023-09-06
Switch SourceBuffer to diagnostics.

jonmeow:
This updates SourceBuffer to diagnostics. Some additional edits to diagnostics were necessary due to issues moving arguments around, which seems to stem from a compile error with clang 14 (fixed in later versions).

### DIFF
--- a/toolchain/diagnostics/diagnostic_emitter.h
+++ b/toolchain/diagnostics/diagnostic_emitter.h
@@ -314,9 +314,9 @@ class StreamDiagnosticConsumer : public DiagnosticConsumer {
     *stream_ << message.location.file_name;
     if (message.location.line_number > 0) {
       *stream_ << ":" << message.location.line_number;
-    }
-    if (message.location.column_number > 0) {
-      *stream_ << ":" << message.location.column_number;
+      if (message.location.column_number > 0) {
+        *stream_ << ":" << message.location.column_number;
+      }
     }
     *stream_ << ": " << message.format_fn(message) << "\n";
     if (message.location.column_number > 0) {
--- a/toolchain/diagnostics/diagnostic_emitter.h
+++ b/toolchain/diagnostics/diagnostic_emitter.h
@@ -54,9 +54,9 @@ struct DiagnosticLocation {
   // A reference to the line of the error.
   llvm::StringRef line;
   // 1-based line number.
-  int32_t line_number;
+  int32_t line_number = -1;
   // 1-based column number.
-  int32_t column_number;
+  int32_t column_number = -1;
 };
 
 // A message composing a diagnostic. This may be the main message, but can also
@@ -215,8 +215,8 @@ class DiagnosticEmitter {
               Internal::NoTypeDeduction<Args>... args) -> DiagnosticBuilder& {
       CARBON_CHECK(diagnostic_base.Level == DiagnosticLevel::Note)
           << static_cast<int>(diagnostic_base.Level);
-      diagnostic_.notes.push_back(
-          MakeMessage(location, diagnostic_base, std::move(args)...));
+      diagnostic_.notes.push_back(MakeMessage(
+          emitter_, location, diagnostic_base, {llvm::Any(args)...}));
       return *this;
     }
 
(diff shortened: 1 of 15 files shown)


toolchain/diagnostics/diagnostic_emitter.h
```diff
@@ -310,13 +311,19 @@ class StreamDiagnosticConsumer : public DiagnosticConsumer {
...
-             << message.location.line << "\n";
-    stream_->indent(message.location.column_number - 1);
-    *stream_ << "^\n";
+    *stream_ << message.location.file_name;
+    if (message.location.line_number > 0) {
+      *stream_ << ":" << message.location.line_number;
+    }
+    if (message.location.column_number > 0) {
+      *stream_ << ":" << message.location.column_number;
+    }
```
zygoloid:
```suggestion
    if (message.location.line_number > 0) {
      *stream_ << ":" << message.location.line_number;
      if (message.location.column_number > 0) {
        *stream_ << ":" << message.location.column_number;
      }
    }
```
To avoid ambiguity and tool confusion, I think we shouldn't print out the column number unless we print out a line number.

toolchain/diagnostics/diagnostic_emitter.h
```diff
@@ -234,22 +234,23 @@ class DiagnosticEmitter {
...
 
     template <typename... Args>
-    auto MakeMessage(LocationT location,
-                     const Internal::DiagnosticBase<Args...>& diagnostic_base,
-                     Internal::NoTypeDeduction<Args>... args)
-        -> DiagnosticMessage {
+    static auto MakeMessage(
+        DiagnosticEmitter<LocationT>* emitter, LocationT location,
+        const Internal::DiagnosticBase<Args...>& diagnostic_base,
+        llvm::SmallVector<llvm::Any> args) -> DiagnosticMessage {
```
zygoloid, line 250: Do you know why the old approach here stopped working? As far as I can see, the diagnostic argument types used by the driver diagnostics are `std::string` and `int64_t`, and we have other diagnostics that take a single `std::string` argument. Is the problem here the `int64_t`? What broke?

Moving the `SmallVector` when constructing a diagnostic might not be especially cheap, and in Clang it's not uncommon to produce thousands of warnings (that get filtered out at some later stage) in a single compilation, so it'd be nice to avoid these extra `SmallVector` moves if we can. But the whole diagnostics machinery is a little inefficient, so maybe it's not the low-hanging fruit here.

zygoloid: The code still builds and the tests pass with the changes in this file (other than the changes to `DiagnosticLocation`) reverted, so I guess whatever issue you were encountering was solved a different way?

jonmeow:
```
╚╡git checkout trunk diagnostics/diagnostic_emitter.h
Updated 1 path from c7d54344d
╚╡bazel run //toolchain/source:source_buffer_test
INFO: Invocation ID: 02f46eec-7bc7-4d31-a6fa-eb7bdd300889
INFO: Build option --compilation_mode has changed, discarding analysis cache.
INFO: Analyzed target //toolchain/source:source_buffer_test (1 packages loaded, 1542 targets configured).
INFO: Found 1 target...
Target //toolchain/source:source_buffer_test up-to-date:
  bazel-bin/toolchain/source/source_buffer_test
INFO: Elapsed time: 5.368s, Critical Path: 4.60s
INFO: 4 processes: 1 internal, 3 linux-sandbox.
INFO: Build completed successfully, 4 total actions
INFO: Running command line: external/bazel_tools/tools/test/test-setup.sh toolchain/source/source_buffer_test
Executing tests from //toolchain/source:source_buffer_test
[==========] Running 4 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 4 tests from SourceBufferTest
[ RUN      ] SourceBufferTest.MissingFile
Stack dump:
0.	Program arguments: /usr/local/google/home/jperkins/.cache/bazel/_bazel_jperkins/85deb7d9d96f7e0e80b42618a55969d7/execroot/carbon/bazel-out/k8-fastbuild/bin/toolchain/source/source_buffer_test.runfiles/carbon/toolchain/source/source_buffer_test
```

It hangs there, I don't know why, but I think there's an issue with arg passing.

zygoloid: (From offline discussion:) Looks like this is probably a miscompile bug in Clang 14.0, though we don't know the exact nature of the bug yet.

jonmeow:
```
╚╡clang --version
Debian clang version 14.0.6
```

As it turns out, although 14 is default, I also have 15 easy to install:

```
╚╡export CC=/usr/lib/llvm-15/bin/clang
╚╡$CC --version
Debian clang version 15.0.7
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/lib/llvm-15/bin
```

(for reference, 16 is unavailable)

It looks like the issue does not repro with clang-15.

Note GitHub only has 14: https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md

This is important for tests.

jonmeow: To be clear on the above -- I think making this work well on GitHub is going to be the limiting factor. Right now we do not do any side-installs of software, it's something we've avoided due to latency and fragility.

jonmeow: Also for reference, I've tried writing a test that I thought would be equivalent, but did not reproduce the issue:

```
struct FilenameFakeDiagnosticLocationTranslator
    : DiagnosticLocationTranslator<llvm::StringRef> {
  auto GetLocation(llvm::StringRef s) -> DiagnosticLocation override {
    return {.file_name = s};
  }
};

TEST_F(DiagnosticEmitterTest, EmitLLVMErrorMessage) {
  EXPECT_CALL(consumer_, HandleDiagnostic(
                             IsDiagnostic(DiagnosticKind::TestDiagnostic,
                                          DiagnosticLevel::Error, -1, -1, _)));
  std::error_code ec(3, std::generic_category());
  llvm::ErrorOr<bool> err(ec);
  FilenameFakeDiagnosticLocationTranslator translator;
  DiagnosticEmitter<llvm::StringRef> emitter(translator, consumer_);
  llvm::StringRef filename = "x";
  if (err.getError()) {
    CARBON_DIAGNOSTIC(TestDiagnostic, Error, "Error: {0}", std::string);
    emitter.Emit(filename, TestDiagnostic, err.getError().message());
  }
}
```

Small variations on the above also did not reproduce.

zygoloid (approved):
Looks good. Please can you update the commit message to say that the changes to diagnostic_emitter are working around a miscompile?

jonmeow:
> Looks good. Please can you update the commit message to say that the changes to diagnostic_emitter are working around a miscompile?

Done:

> Some additional edits to diagnostics were necessary due to issues moving arguments around, which seems to stem from a compile error with clang 14 (fixed in later versions).

PR #3200, opened 2023-09-06
Fix lowering of 8 and 9.

zygoloid:
IntegerLiterals are not signed, so get the zero-extended value rather than the sign-extended value. Sometimes we use the high bit of the `APInt`, though currently this only happens for the literals 8 and 9 due to the way we convert decimal integers to binary.


jonmeow (approved):
Oops

PR #3201, opened 2023-09-06
Fix printing of integer literals with the high bit set.

zygoloid:
Integer literals are not signed, so don't print them as signed.


jonmeow (approved):
Maybe we just need a wrapper for APInt that prevents these API misuses?

PR #3202, opened 2023-09-07
Improve testing and handling of unsigned APInt values

jonmeow:


jonmeow:
Note, numeric_literal lexing also checks getBitWidth for mantissa, you might want to double-check that logic (I figured you'd be able to tell at a glance if it's correct, it'd take me some more time)

zygoloid (approved):

zygoloid:
I've checked the `numeric_literal.cpp` code's use of `getBitWidth`, and it appears to be correct.

PR #3204, opened 2023-09-07
Disable raw IR printing in tests by default.

zygoloid:
Add a couple more tests to check/testdata/basic/ to explicitly test the raw dump.


jonmeow (approved):
LGTM, I'm really happy with the IR printing.

jonmeow:
Well, happy with the IR printing except for how the line updates can require two runs.

PR #3205, opened 2023-09-07
Switch to modeling array initialization analogously to a function call.

zygoloid:
Instead of modeling array initialization as a thin wrapper around  tuple initialization, handle it like a function call, with a return slot as part of its input. This better matches how initialization via a call to `ImplicitAs::Convert` will eventually work, and in particular lets us do in-place initialization of arrays rather than always creating a temporary.


toolchain/check/context.h
```diff
@@ -128,6 +128,24 @@ class Context {
...
+  // from the initializer to the target if the initialization is not in-place.
+  // That final store is often undesirable as it is performed by the consumer
+  // of the initializer, such as an `Assign` or `ReturnExpression` node. The
+  // resulting node describes the initialization operation that was performed.
+  auto InitializeAndFinalize(Parse::Node parse_node, SemIR::NodeId target_id,
+                             SemIR::NodeId value_id) -> SemIR::NodeId {
+    auto init_id = Initialize(parse_node, target_id, value_id);
+    if (SemIR::GetInitializingRepresentation(
+            semantics_ir(), semantics_ir().GetNode(target_id).type_id())
+            .kind == SemIR::InitializingRepresentation::ByCopy) {
```
```suggestion
    auto init_rep = SemIR::GetInitializingRepresentation(
            semantics_ir(), semantics_ir().GetNode(target_id).type_id());
    if (init_rep.kind == SemIR::InitializingRepresentation::ByCopy) {
```

Just with how the `.kind` wraps, WDYT of this?

toolchain/check/context.cpp
```diff
@@ -538,32 +661,27 @@ auto Context::ImplicitAsImpl(SemIR::NodeId value_id, SemIR::TypeId as_type_id,
...
-        return ImplicitAsKind::Compatible;
+      if (output_value_id == nullptr) {
+        return CanConvertTupleToArray(*this, value_type_node, as_type_node)
+                   ? ImplicitAsKind::Compatible
+                   : ImplicitAsKind::Incompatible;
       }
+      *output_value_id = ConvertTupleToArray(
+          *this, value_type_node, as_type_node, as_type_id, value_id);
+      // Return `Compatible` because we've already diagnosed any error as part
+      // of the attempted conversion.
```
jonmeow: Previously, the code looked like:

```
if (value kind == TupleType) {
  if (size == tuple size && elements match) {
    ...
    return Compatible;
  }
}
```

Now this looks like:

```
if (value kind == TupleType) {
  if (ignoring output) {
    Definitely return
  }
  ...
  return Compatible;
}
```

If I understand correctly, you're changing the meaning of "Compatible" to mean either compatible or "incompatible but you've already printed an error" -- what do you think of renaming "Compatible" to better reflect that, or more generally changing the API to suit this kind of flow?

However, part of the previous path was that by not immediately returning an error when a given conversion didn't work, it could still consider user-defined conversions. What's your plan for that? Are you assuming that user-defined conversions should be checked first, allowing them to override builtin conversions? If so, that might suggest towards a different flow of this function.

(note, I'm also a little concerned that this could lead to divergent casting errors depending upon whether CanConvertTupleToArray or ConvertTupleToArray is used -- see top-level review comment for a test that might reveal this)

toolchain/check/context.cpp
```diff
@@ -414,6 +427,20 @@ auto Context::MarkInitializerFor(SemIR::NodeId init_id, SemIR::NodeId target_id)
...
       }
+
+      case SemIR::NodeKind::ArrayInit: {
+        // Rewrite the return slot as a reference to our target. We can't just
+        // update the index in `refs_id`, like we do for a Call, because there
+        // will be other references to the return slot for the individual array
+        // element initializers.
+        auto [src_id, refs_id] = init.GetAsArrayInit();
+        semantics_ir().ReplaceNode(
+            semantics_ir().GetNodeBlock(refs_id).back(),
```
jonmeow: Why do you check the type of temporary_id in FinalizeTemporary but not here? (I'd have thought this case would be interesting because it's being replaced, so ensuring the right thing is erased)

toolchain/check/context.cpp
```diff
@@ -505,6 +532,102 @@ auto Context::ImplicitAsRequired(Parse::Node parse_node, SemIR::NodeId value_id,
   return output_value_id;
 }
 
+static auto CanConvertTupleToArray(Context& context, SemIR::Node tuple_type,
+                                   SemIR::Node array_type) -> bool {
```
jonmeow: Maybe have this return the enum directly, avoiding the ternary at the caller?

toolchain/check/context.cpp
```diff
@@ -505,6 +532,102 @@ auto Context::ImplicitAsRequired(Parse::Node parse_node, SemIR::NodeId value_id,
   return output_value_id;
 }
 
+static auto CanConvertTupleToArray(Context& context, SemIR::Node tuple_type,
```
```suggestion
// Returns true if the tuple could be converted to the array type.
static auto CanConvertTupleToArray(Context& context, SemIR::Node tuple_type,
```

Cursory comment request.

toolchain/check/context.cpp
```diff
@@ -505,6 +532,102 @@ auto Context::ImplicitAsRequired(Parse::Node parse_node, SemIR::NodeId value_id,
...
+      context.semantics_ir().GetArrayBoundValue(array_bound_id)) {
+    return false;
+  }
+
+  // TODO: We should do a proper convertibility check here for each
+  // tuple element.
+  return llvm::all_of(tuple_elem_types,
+                      [element_type_id = element_type_id](auto type_id) {
+                        return type_id == element_type_id;
+                      });
```
jonmeow: I see this in the old code, although it strikes me it should be simple to call `ImplicitAsImpl` for this (pass `tuple_value` and iterate values, instead of `tuple_type`). Though, that may suggest this function should be a member instead of a static function.

Maybe not worth it at the moment since it was like this before, but would particularly help code that includes builtin errors (e.g., `var x: [i32; 1] = (ProduceError(),);` would be "compatible").

toolchain/check/context.cpp
```diff
@@ -505,6 +532,102 @@ auto Context::ImplicitAsRequired(Parse::Node parse_node, SemIR::NodeId value_id,
...
+  auto [array_bound_id, element_type_id] = array_type.GetAsArrayType();
+  auto tuple_elem_types_id = tuple_type.GetAsTupleType();
+  const auto& tuple_elem_types =
+      context.semantics_ir().GetTypeBlock(tuple_elem_types_id);
+
+  auto array_bound = context.semantics_ir().GetArrayBoundValue(array_bound_id);
+  if (tuple_elem_types.size() != array_bound) {
+    CARBON_DIAGNOSTIC(
+        ArrayInitArgCountMismatch, Error,
+        "Cannot initialize array of {0} element(s) from {1} initializer(s).",
```
jonmeow, line 582: Would this lead to:

```
var a: (i32,) = (0, 1);
var b: [i32; 1] = (1,);
// Cannot initialize array of 1 element(s) from 2 initializer(s).
b = a;
```

Does the "initialize" phrasing still align for you? (partly I comment because referring to the tuples as "N initializers" wasn't matching how I think of the code)

toolchain/check/context.cpp
```diff
@@ -505,6 +532,102 @@ auto Context::ImplicitAsRequired(Parse::Node parse_node, SemIR::NodeId value_id,
...
+  // as the destination for the array initialization.
+  auto return_slot_id = context.AddNode(
+      SemIR::Node::TemporaryStorage::Make(value.parse_node(), array_type_id));
+
+  // Initialize each element of the array from the corresponding element of the
+  // tuple.
+  for (auto [i, src_type_id] : llvm::enumerate(tuple_elem_types)) {
+    // TODO: Add a new node kind for indexing an array at a constant index
+    // so that we don't need an integer literal node here.
+    auto int_id = context.AddNode(SemIR::Node::IntegerLiteral::Make(
```
jonmeow: Can you rename this to index_id or something? Having int_id and init_id next to each other feels close to a typo...

jonmeow (commented):
I see this is the ArrayInit approach we'd discussed. I do have a question though about the changes to ImplicitAsImpl flow, trying to make sure we're consistent in plans.

I'd like to highlight that calls currently do slightly different implicit cast logic, so you might want to add a failure test such as:

```
fn Run(x: [i32; 1]) {}

fn Main() {
  Run((0, 1));
}
```

In testdata/function/call/fail_param_type.carbon, you'll find:

```
No matching callable was found.
Function cannot be used: Cannot implicitly convert argument 0 from `f64` to `i32`.
```

I'm noting this case in particular due to the way you're changing ImplicitAsImpl flow to error directly from ConvertTupleToArray; I think there may be some indirect results you're not expecting.

toolchain/check/context.cpp
zygoloid: No good reason not to check it here; added a check.

toolchain/check/context.cpp
zygoloid: That doesn't seem great; added a different diagnostic for the case where we're not initializing from a tuple literal.

toolchain/check/context.h
zygoloid: Done, also moved the variable into the `if`.

toolchain/check/context.cpp
zygoloid: Obsolete after #3208.

toolchain/check/context.cpp
zygoloid: Obsolete after #3208.

toolchain/check/context.cpp
zygoloid: Obsolete after #3208.

toolchain/check/context.cpp

toolchain/check/context.cpp
zygoloid: Obsolete after #3208.

zygoloid (commented):
Rebased on #3208, waiting for merge queue to catch up.

toolchain/check/context.cpp
```diff
@@ -492,6 +542,96 @@ auto Context::ImplicitAsForArgs(Parse::Node call_parse_node,
...
+        context.semantics_ir().GetNodeBlock(value.GetAsTupleValue());
+  }
+
+  // Check that the tuple is the right size.
+  auto array_bound = context.semantics_ir().GetArrayBoundValue(array_bound_id);
+  if (tuple_elem_types.size() != array_bound) {
+    CARBON_DIAGNOSTIC(
+        ArrayInitFromLiteralArgCountMismatch, Error,
+        "Cannot initialize array of {0} element(s) from {1} initializer(s).",
+        decltype(array_bound), size_t);
```
jonmeow: This decltype seems odd to me, is it too difficult to provide a type?

jonmeow (approved):
LGTM

toolchain/check/context.cpp

arm666 (approved):

PR #3206, opened 2023-09-08
Merge toolchain file_test children in order to improve linking.

jonmeow:
Specifically this should improve linking by producing one large binary instead of one-per-directory. The inclusion of the driver hits the size issue. Separating out things which have more llvm deps has been discussed, but I'm not doing that here because I think the semantics layer will need to depend on clang for interop, and we'd lose a lot of the benefits that way. Also, having just one place to look seems simpler.

Includes supporting changes to file_test infrastructure, the most significant of which is probably passing tests via file instead of a really large args thing, using a custom rule to do that. That's because dealing with the layered filegroups that allow the toolchain setup is more complicated, and this approach scales well.

Combined test time is ~9s, so not sharding right now.

I wasn't sure if people would prefer having the autoupdate script under testing, so I left it alone for now.


testing/file_test/file_test_base.cpp
```diff
@@ -331,7 +335,7 @@ auto FileTestBase::ProcessTestFile(TestContext& context) -> ErrorOr<Success> {
   } else {
     // If no file splitting happened, use the main file as the test file.
     context.test_files.push_back(
-        TestFile(path().filename().string(), file_content));
+        TestFile(test_name_.rsplit("/").second.str(), file_content));
```
zygoloid: According to the documentation, `rsplit("/").second` is empty if the delimiter doesn't appear in the string. Is that what you want here? (Maybe we don't care because there's always a `testdata/`?)

testing/file_test/file_test_base.cpp
```diff
@@ -431,16 +435,38 @@ auto FileTestBase::TransformExpectation(int line_index, llvm::StringRef in)
...
+      continue;
+    }
+    all_tests.push_back(file_ref.str());
+  }
+  return all_tests;
 }
 
-}  // namespace Carbon::Testing
+// Uses --file_tests, verifying entries are valid.
+static auto FilterFileTests(const llvm::SmallVector<std::string>& tests)
```
zygoloid: The name of this function doesn't seem to match what it's doing. Maybe `GetAndValidateFileTests`? Though see my next comment.

testing/file_test/file_test_base.cpp
```diff
@@ -454,46 +480,33 @@ auto main(int argc, char** argv) -> int {
...
-    CARBON_CHECK(!ec) << file_test << ": " << ec.message();
-    CARBON_CHECK(llvm::StringRef(path.string()).starts_with(base_dir))
-        << "\n  " << path << "\n  should start with\n  " << base_dir;
-    if (absl::GetFlag(FLAGS_autoupdate)) {
-      std::unique_ptr<Carbon::Testing::FileTestBase> test(
-          test_factory.factory_fn(path));
+  llvm::SmallVector<std::string> tests = GetAllTests();
+  if (!absl::GetFlag(FLAGS_file_tests).empty()) {
+    tests = FilterFileTests(tests);
+  }
```
zygoloid: I wonder if it's worth doing a filtering / validation step here at all -- if I have a test that I'm working on that's not in the "all tests" list, maybe I should be able to run it manually by invoking the `file_test` binary? Maybe we should require that either `--file_tests` is specified or `--test_targets_file` is specified (and possibly error if both are)?

testing/file_test/file_test_base.h
```diff
@@ -166,10 +162,9 @@ struct FileTestFactory {
 extern auto GetFileTestFactory() -> FileTestFactory;
 
 // Provides a standard GetFileTestFactory implementation.
-#define CARBON_FILE_TEST_FACTORY(Name)                                         \
-  auto GetFileTestFactory()->FileTestFactory {                                 \
-    return {(#Name),                                                           \
-            [](const std::filesystem::path& path) { return new Name(path); }}; \
+#define CARBON_FILE_TEST_FACTORY(Name)                                     \
+  auto GetFileTestFactory()->FileTestFactory {                             \
+    return {(#Name), [](llvm::StringRef path) { return new Name(path); }}; \
```
zygoloid:
```suggestion
#define CARBON_FILE_TEST_FACTORY(Name)                                   \
  auto GetFileTestFactory()->FileTestFactory {                           \
    return {#Name, [](llvm::StringRef path) { return new Name(path); }}; \
```
(Trivial nit, no need for parens around a string literal.)

testing/file_test/file_test_base_test.cpp
```diff
@@ -31,7 +31,8 @@ class FileTestBaseTest : public FileTestBase {
     }
     stdout << "\n";
 
-    auto filename = path().filename();
+    auto filename =
+        std::filesystem::path(test_name().str()).filename().string();
```
zygoloid: Why is the `.string()` call added here?

zygoloid (commented):
I think it would be useful to still have targets to run the individual tests from the specific toolchain components. I sometimes want to run just the `check` tests when I've made some `check` changes and not made corresponding `lower` changes yet (beyond getting it to build). Would it be reasonable to keep around the per-directory `FileTest` implementations, and have the `toolchain/testing` code delegate to the appropriate `FileTest` implementation based on the component?

jonmeow:
> I think it would be useful to still have targets to run the individual tests from the specific toolchain components. I sometimes want to run just the `check` tests when I've made some `check` changes and not made corresponding `lower` changes yet (beyond getting it to build). Would it be reasonable to keep around the per-directory `FileTest` implementations, and have the `toolchain/testing` code delegate to the appropriate `FileTest` implementation based on the component?

I don't think that we can have per-directory `FileTest` implementations without bringing back the binaries, or some other significant complexity. Instead, if we were going this route, I'd probably do something like trying to bulk up the argument passing to just pass in all the relevant test files.

However, are you sure this will continue to be useful to you? In the old setup, the bazel bootstrapping was a significant cost. In the new setup, the bazel bootstrapping is still the most significant cost, but I'm seeing sub-second execution time.

I think this may not be necessary right now -- that is, thinking about the directory may not be saving time.

zygoloid:
> However, are you sure this will continue to be useful to you? In the old setup, the bazel bootstrapping was a significant cost. In the new setup, the bazel bootstrapping is still the most significant cost, but I'm seeing sub-second execution time.

Hm, thinking more about my recent interactions with the `file_test` machinery, I think I'd be OK with losing `bazel test` for subdirectories, but losing `autoupdate` for subdirectories I think would be more of an issue. It's frequently the case during `check` development work that `autoupdate` starts crashing in `lower` tests, and it would be an impediment if there's no good way to autoupdate just the `check` tests.

testing/file_test/file_test_base.cpp
jonmeow: More than that, this would only be an issue if we put a test in the root of the repo. Adding a comment.

But also switching to drop_front because I feel better about that, it's not discarding half the split.

testing/file_test/file_test_base.cpp
jonmeow: Ack

testing/file_test/file_test_base.cpp
jonmeow: Done. A part of me likes the safety check, but I don't feel strongly.

> Maybe we should require that either `--file_tests` is specified or `--test_targets_file` is specified (and possibly error if both are)?

`--test_targets_file` is provided as part of the target, so I'd rather keep requiring it.

testing/file_test/file_test_base.h
jonmeow: Done

testing/file_test/file_test_base_test.cpp
jonmeow: Incorrectly assumed it was required; removed

jonmeow (commented):
Added support along the lines of `./autoupdate_testdata.py lex/**/*`, does that work for you?

jonmeow:
> I think I'd be OK with losing `bazel test` for subdirectories,

Similar to the autoupdate filter, want me to add an equivalent script for `bazel test`?

(but then I may more strongly want to put these into the testing subdir)

toolchain/driver/testdata/fail_missing_file.carbon
```diff
@@ -5,5 +5,5 @@
 // ARGS: compile --phase=lex nonexistent.carbon
 //
 // AUTOUPDATE
-
 // CHECK:STDERR: nonexistent.carbon: Error opening file for read: No such file or directory
+
```
zygoloid: Sorry, I think I introduced this bug in autoupdate; this fails pre-commit. Sent #3209 to fix.

zygoloid (approved):
Looks good (other than pre-commit failure)

zygoloid:
> Similar to the autoupdate filter, want me to add an equivalent script for `bazel test`?

Let's try without and see how we go.

toolchain/driver/testdata/fail_missing_file.carbon
jonmeow: Thanks for the note, I'm going to revert this file from the change.

PR #3207, opened 2023-09-08
Try to fix a surprising error from the new target selection.

chandlerc:
Also add some extra debug logging to try to help diagnose the error.

### DIFF
--- a/scripts/target_determinator.py
+++ b/scripts/target_determinator.py
@@ -48,6 +48,7 @@ def filter_targets(bazel: Path, targets: str) -> str:
             )
         log(f"Bazel query snippet:\n```\n{query_snippet}\n```")
         tmp.write(query)
+        tmp.flush()
         try:
             p = subprocess.run(
                 [str(bazel), "query", f"--query_file={tmp.name}"],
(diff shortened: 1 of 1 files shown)


scripts/target_determinator.py
```diff
@@ -59,6 +60,10 @@ def filter_targets(bazel: Path, targets: str) -> str:
             return p.stdout
         except subprocess.CalledProcessError as err:
             log(err.stderr)
+            log("Query file:\n```")
+            with open(tmp.name) as f:
+                log(f.read())
+            log("```")
```
jonmeow:
```suggestion
            log("Full query file:\n```\n")
            with open(tmp.name) as f:
                log(f.read())
            log("```")
```

jonmeow (approved):
You're getting a small suggestion mainly due to the missing newline. :)

(you might also consider one on the close)

scripts/target_determinator.py
chandlerc: Log uses `print` which I thought already adds a newline at the end?

chandlerc: I tweaked the log string, but the docs for `print` indicate it includes a newline at the end.

PR #3208, opened 2023-09-09
Remove two-phase implicit conversion of function calls.

zygoloid:
In order to maintain diagnostic quality, add a mechanism to add notes to any diagnostics that are produced as part of initialization of function parameters. As suggested in review of #3205.

In passing, fix the only caller of `ImplicitAsRequired` outside of the implementation of `Check::Context` to instead use
`ConvertToValueExpressionOfType`. This causes some missing `value_binding` nodes to be added to the produced SemIR. Also fixed a matching bug in lowering where a bogus load was being added, that resulted in assertion failures when the checker bug was fixed.

The warning `-Wnon-virtual-dtor` starts producing false-positive warnings after this change. Replace it with the fixed version, `-Wdelete-non-virtual-dtor`.


toolchain/check/context.cpp
```diff
@@ -445,95 +449,61 @@ auto Context::ImplicitAsForArgs(
...
-                        "Function cannot be used: Cannot implicitly convert "
-                        "argument {0} from `{1}` to `{2}`.",
-                        size_t, std::string, std::string);
-      diagnostic->Note(param_parse_node, CallArgTypeMismatch, i,
-                       semantics_ir_->StringifyType(
-                           semantics_ir_->GetNode(value_id).type_id()),
-                       semantics_ir_->StringifyType(as_type_id));
-      return false;
-    }
+    DiagnosticAnnotationScope annotate_diagnostics(
```
jonmeow: Would it be better to hook the consumer outside of the loop, so that overhead of the scope/lambda construction doesn't scale with the number of args?

Maybe this is incremental cost from a performance perspective, although versus other diagnostic logic, it's going to occur regardless of whether a diagnostic is printed. It does make me wonder though if there's a better approach, like maybe the note printing should be passed instead of hooking the diagnostic. But I'm guessing you're building on experience from clang here, that this kind of approach is better at scale.

toolchain/check/context.cpp
```diff
@@ -445,95 +449,61 @@ auto Context::ImplicitAsForArgs(
 
   // If sizes mismatch, fail early.
   if (arg_refs.size() != param_refs.size()) {
-    CARBON_CHECK(diagnostic != nullptr) << "Should have validated first";
-    CARBON_DIAGNOSTIC(CallArgCountMismatch, Note,
+    CARBON_DIAGNOSTIC(CallArgCountMismatch, Error,
                       "Function cannot be used: Received {0} argument(s), but "
                       "require {1} argument(s).",
                       int, int);
```
jonmeow: Since you're changing the structure of the resulting diagnostic, should the text be changed as well? "Function cannot be used:" sounds odd when it's heading up the error; maybe something like "Caller passed {0} argument(s), but {1} argument(s) are required."?

As a sidenote -- although "function" is in a lot of these already, in my mind ImplicitAsForArgs is a more general caller/callee structure.

toolchain/diagnostics/diagnostic_emitter.h
```diff
@@ -358,6 +372,55 @@ class ErrorTrackingDiagnosticConsumer : public DiagnosticConsumer {
...
+// `DiagnosticBuilder& builder` for any diagnostic that is emitted through the
+// given emitter. That function can annotate the diagnostic by calling
+// `builder.Note` to add notes.
+template <typename LocationT, typename MakeNoteFn>
+class DiagnosticAnnotationScope {
+ public:
+  DiagnosticAnnotationScope(DiagnosticEmitter<LocationT>* emitter,
+                            MakeNoteFn make_note)
+      : consumer_(emitter, emitter->consumer_, std::move(make_note)),
+        saved_consumer_(emitter->consumer_, &consumer_) {}
```
jonmeow: For example with SortingDiagnosticConsumer, HandleDiagnostic is not guaranteed to be called immediately. By allowing references back to a local scope, this is implicitly adding a constraint that no delayed custom consumers should be added later, something that'd be worth commenting.

Had you considered making this a stack of scoped functions on the DiagnosticEmitter? i.e., versus saving a consumer, push/pop a scope. The DiagnosticBuilder could pass itself to the scopes, versus constructing secondary builders within HandleDiagnostic, guaranteeing any scope's use doesn't outlive the scope.

toolchain/diagnostics/diagnostic_emitter.h
```diff
@@ -358,6 +372,55 @@ class ErrorTrackingDiagnosticConsumer : public DiagnosticConsumer {
...
+    Consumer(DiagnosticEmitter<LocationT>* emitter,
+             DiagnosticConsumer* next_consumer, MakeNoteFn make_note)
+        : emitter_(emitter),
+          next_consumer_(next_consumer),
+          make_note_(std::move(make_note)) {}
+
+    auto HandleDiagnostic(Diagnostic diagnostic) -> void override {
+      typename DiagnosticEmitter<LocationT>::DiagnosticBuilder builder(
+          emitter_, std::move(diagnostic));
+      make_note_(builder);
```
jonmeow: Minor note, `make_note` implies that it's restricted to a single note. Perhaps a more generic `on_diagnostic`?

jonmeow (commented):
I'm assuming this is a response to my comments about ImplicitAs flow on #3205. What you're doing here seems good to me, just some comments about how the scope is being managed.

toolchain/check/context.cpp
zygoloid: Moved the scope outside the loop as suggested.

Regarding this scope versus passing information into the callee: I'm concerned that the complexity of passing around a note-printing mechanism would be substantial, as the volume of code that's reachable through parameter initialization (and other situations where we have context notes) grows -- we'd end up passing it to probably most or all of the code in `check` that can produce diagnostics, so having the scope change some state in the emitter seems better. Forwarding around an extra parameter is also likely to add up to a fair bit of overhead even if no diagnostic is emitted.

I'm expecting we'll want to change the implementation on the diagnostics side a bit (or perhaps a lot) in the future. For example, I think it would be a good idea for the diagnostics mechanism to hold a separate list of context callbacks rather than swapping out the diagnostic consumer. In Clang, we also only print the context note for the first diagnostic that's emitted within the scope, and we might want to consider making that change in the future as well. But I think the general model of (cheaply) registering some kind of callback, somewhere in the diagnostics infrastructure, is probably the right way to go.

toolchain/check/context.cpp
zygoloid: Rephrased diagnostic.

I think once we start using `ImplicitAsForArgs` more generally we can think about the best way to improve the diagnostic -- perhaps we want to pass in an argument to say what kind of call we're performing, to decide whether to include "function" or "generic class" etc. in the diagnostic text.

toolchain/diagnostics/diagnostic_emitter.h
zygoloid: Renamed to `annotate` to match the naming of the class.

toolchain/diagnostics/diagnostic_emitter.h
zygoloid: Switched to a stack of scoped functions on the emitter. That was the direction I was thinking we'd go in the long term, and on reflection I can't think of a single reason not to do that right now :-)

Thanks for the suggestion!

toolchain/check/context.cpp
zygoloid: > I think it would be a good idea for the diagnostics mechanism to hold a separate list of context callbacks rather than swapping out the diagnostic consumer.

I see you had the same suggestion in a later comment. Done :-)

toolchain/diagnostics/diagnostic_emitter.h
```diff
@@ -295,8 +306,37 @@ class DiagnosticEmitter {
   }
 
  private:
+  // Base class for scopes in which we perform diagnostic annotation, such as
+  // adding notes with contextual information.
+  class DiagnosticAnnotationScopeBase {
```
jonmeow, line 311: Why the separation of logic here, versus something like `llvm::SmallVector<std::function<void(DiagnosticBuilder)>>`?

Maybe this thought is a little too quirky/limiting, but:

```
// Returns a scope_exit, so we let the return type be inferred.
auto AddAnnotationScope(std::function<void(DiagnosticBuilder)> annotator) {
  annotators_.push_back(annotator);
  return llvm::make_scope_exit([annotator]() {
    CARBON_CHECK(annotators_.back() == annotator);
    annotators_.pop_back();
  });
}
```

toolchain/diagnostics/diagnostic_emitter.h
```diff
@@ -358,6 +398,39 @@ class ErrorTrackingDiagnosticConsumer : public DiagnosticConsumer {
...
+      typename DiagnosticEmitter<LocationT>::DiagnosticBuilder& builder)
+      -> void override {
+    annotate_(builder);
+  }
+
+  AnnotateFn annotate_;
+};
+
+template <typename LocationT, typename AnnotateFn>
+DiagnosticAnnotationScope(DiagnosticEmitter<LocationT>* emitter,
```
jonmeow, line 430: Am I missing something here, or is this undefined / unused?

jonmeow (approved):
I think this is in good state to merge, although two last comments.

toolchain/diagnostics/diagnostic_emitter.h
zygoloid: This is a [deduction guide](https://en.cppreference.com/w/cpp/language/class_template_argument_deduction). We're using it in `context.cpp`:

```c++
  DiagnosticAnnotationScope annotate_diagnostics(emitter_, [&](auto& builder) {
```

Note that we don't pass in template arguments here; this guide allows them to be deduced. (It's also what would happen implicitly / by default, but our style guide says to not rely on implicit deduction guides.)

toolchain/diagnostics/diagnostic_emitter.h
zygoloid: This is efficiency-driven; `std::function` will perform a heap allocation if the lambda doesn't fit into the small-function buffer, and often performs two indirect function calls to invoke its callee rather than one. (Also, minor, `std::function` doesn't have an `operator==` so we can't perform the `CHECK`.)

The `AddAnnotationScope` approach is definitely interesting. I'm going to land this first and then try that out.

jonmeow: Bah, I missed that https://en.cppreference.com/w/cpp/utility/functional/function says `==` was removed in c++20

PR #3209, opened 2023-09-09
Fix oscillating autoupdate output for empty source files.

zygoloid:
toolchain/driver/testdata/fail_missing_file.carbon has no content after its AUTOUPDATE line. This caused it to oscillate between three states:

1) If there were no content lines after AUTOUPDATE, autoupdate would add
   a blank line and then its STDERR CHECK line.
2) If there was a (blank) content line after AUTOUPDATE, autoupdate
   would move the STDERR CHECK line to immediately after AUTOUPDATE,
   leaving a trailing blank line.
3) pre-commit would remove the trailing blank line, leaving no content
   lines after AUTOUPDATE.

Handle the special case of no content lines after AUTOUPDATE by producing "early" STDERR check lines immediately after the AUTOUPDATE line, rather than before the next line, which might not exist. We already did this in the case where there were early STDOUT check lines.


jonmeow (approved):

jonmeow:
Looks good, but you'll need to get past the pre-commit issue.

PR #3214, opened 2023-09-12
Support multi-file lex printing and testing.

jonmeow:
Lex now prints its yaml as:
```
- filename: name
  tokens: [ ... ]
```

New support in file_test allows the `filename` marker at the top to define the default file number for later lines, meaning multi-file output from lexing is now associated with the appropriate file. Similar support will probably also apply to lowering, semir, and other places that print a filename once for the full dump.

This hammers a bit at how line number replacements work in file_test, allowing stacking them so that lex errors and stdout can both be line-associated properly. I've tried to make the autoupdate more frequently work in one pass, now also taking into account the file index when doing line replacements.

There are still some issues with EndOfFile that it may be good to discuss: because CHECK lines are appended to the end of the file now, and the EndOfFile token points at the last line including comments, new lex tests now take two runs to autoupdate (because without CHECK lines, the EndOfFile points at a content line, which content is then inserted after). Note that removing CHECK lines from the test is not a solution: autoupdate also started inserting blank lines, which breaks this for a similar reason. One solution here might be to not have EndOfFile associate with a line or column, which has been a bit of an issue regardless.

Also fixes a small issue with toolchain's autoupdate script.


zygoloid (approved):
Please see comments on #3217 which I accidentally left on the wrong PR :)

As noted there, if you prefer, please feel free to merge this and fix-forward for any of the comments.

jonmeow:
> Please see comments on #3217 which I accidentally left on the wrong PR :)
> 
> As noted there, if you prefer, please feel free to merge this and fix-forward for any of the comments.

Thanks! I believe the small things are fixed, big things I'd like to separate out.

PR #3215, opened 2023-09-12
Update parse tree yaml for multi-file.

jonmeow:
Building on #3214, updates parse tree yaml to be:

```
- filename: name
  parse_tree: [ ... ]
```


toolchain/parse/tree_test.cpp
```diff
@@ -68,7 +68,14 @@ TEST_F(TreeTest, PrintPostorderAsYAML) {
       Yaml::MappingValue{{"kind", "FileEnd"}, {"text", ""}},
   };
 
-  EXPECT_THAT(Yaml::Value::FromText(print_stream.TakeStr()), ElementsAre(file));
+  auto root = Yaml::SequenceValue{
```
geoffromer, line 71: Optional: consider naming this something like `expected_root` - as a reader it helps me quickly pick up the structure of the test. I also think it would be reasonable to embed this in the `EXPECT_THAT` line, rather than pull it out as a variable.

geoffromer (approved):
I didn't review all the `.carbon` files; just spot-checked enough to be satisfied the rest are right. I'm leaving the merge to you in case you want chandler to review as well.

toolchain/parse/tree_test.cpp
jonmeow: All the tests for yaml across the toolchain don't say "expected_", so I think this is consistent.

PR #3217, opened 2023-09-12
Update SemIR yaml and text for multi-file.

jonmeow:
Building on #3214 and #3215, updates sem_ir yaml to be:

```
- filename: name
  sem_ir: [ ... ]
```

Also, changes the textual format from `package { ... }` to `file <filename> { ... }`. My thought on packages there is:

```
file "foo.carbon" {
  package MyPackage

  ...
}
```

The reason for putting the file first is that it's easier if we put what we're grouping on first, whereas the package is an "annotation" on the file.

### DIFF
--- a/toolchain/base/yaml_test_helpers.h
+++ b/toolchain/base/yaml_test_helpers.h
@@ -132,7 +132,7 @@ MATCHER_P(Sequence, contents,
 // Same as testing::VariantWith<ScalarValue>(contents).
 // NOLINTNEXTLINE: Expands from GoogleTest.
 MATCHER_P(Scalar, value,
-          "has scalar value " + ::testing::PrintToString(value)) {
+          "has scalar value " + DescribeMatcher<std::string>(value)) {
   ::testing::Matcher<ScalarValue> value_matcher = value;
 
   if (auto* map = std::get_if<ScalarValue>(&arg)) {
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -15,7 +15,7 @@ auto CheckParseTree(const SemIR::File& builtin_ir,
                     const Lex::TokenizedBuffer& tokens,
                     const Parse::Tree& parse_tree, DiagnosticConsumer& consumer,
                     llvm::raw_ostream* vlog_stream) -> SemIR::File {
-  auto semantics_ir = SemIR::File(&builtin_ir);
+  auto semantics_ir = SemIR::File(tokens.filename().str(), &builtin_ir);
 
   Parse::NodeLocationTranslator translator(&tokens, &parse_tree);
   ErrorTrackingDiagnosticConsumer err_tracker(consumer);
--- a/toolchain/check/testdata/array/assign_return_value.carbon
+++ b/toolchain/check/testdata/array/assign_return_value.carbon
@@ -10,7 +10,7 @@ fn Run() {
   var t: [i32; 1] = F();
 }
 
-// CHECK:STDOUT: package {
+// CHECK:STDOUT: file "assign_return_value.carbon" {
 // CHECK:STDOUT:   %.loc7 = fn_decl @F
 // CHECK:STDOUT:   %.loc9 = fn_decl @Run
 // CHECK:STDOUT: }
--- a/toolchain/check/testdata/array/assign_var.carbon
+++ b/toolchain/check/testdata/array/assign_var.carbon
@@ -7,7 +7,7 @@
 var a: (i32, i32, i32) = (1, 2, 3);
 var b: [i32; 3] = a;
 
-// CHECK:STDOUT: package {
+// CHECK:STDOUT: file "assign_var.carbon" {
 // CHECK:STDOUT:   %.loc7_9: type = stub_reference i32
 // CHECK:STDOUT:   %.loc7_14: type = stub_reference i32
 // CHECK:STDOUT:   %.loc7_19: type = stub_reference i32
--- a/toolchain/check/testdata/array/base.carbon
+++ b/toolchain/check/testdata/array/base.carbon
@@ -8,7 +8,7 @@ var a: [i32; 1] = (1,);
 var b: [f64; 2] = (11.1, 2.2,);
 var c: [(); 5] = ((), (), (), (), (),);
 
-// CHECK:STDOUT: package {
+// CHECK:STDOUT: file "base.carbon" {
 // CHECK:STDOUT:   %.loc7_14: i32 = int_literal 1
 // CHECK:STDOUT:   %.loc7_15: type = array_type %.loc7_14, i32
 // CHECK:STDOUT:   %a: ref [i32; 1] = var "a"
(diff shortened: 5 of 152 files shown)


testing/file_test/autoupdate.cpp
```diff
@@ -26,20 +25,33 @@ static auto ParseLineNumber(absl::string_view matched_line_number) -> int {
   trimmed = trimmed.trim();
   // NOLINTNEXTLINE(google-runtime-int): API requirement.
   long long val;
-  CARBON_CHECK(!llvm::getAsSignedInteger(trimmed, 10, val));
+  CARBON_CHECK(!llvm::getAsSignedInteger(trimmed, 10, val))
+      << matched_line_number;
   return val;
 }
 
+// A replacement that provides the file and line numbers.
```
zygoloid: This comment isn't quite enough for me to figure out what this models. Having read the usage below, perhaps:

```suggestion
// The file and line number that a CHECK line refers to, and the
// replacement from which they were determined, if any.
```

testing/file_test/autoupdate.cpp
```diff
@@ -121,11 +153,45 @@ class CheckLine : public FileTestLineBase {
...
+        if (auto it = file_to_number_map.find(filename);
+            it != file_to_number_map.end()) {
+          return FileAndLineNumber(&replacement, it->second, line_number);
+        } else {
+          return FileAndLineNumber(default_file_number);
+        }
+      }
+    } else {
+      // There's no file association, so we only look at the line, and assume
+      // it refers to the main file.
```
zygoloid:
```suggestion
      // it refers to the default file.
```

testing/file_test/autoupdate.cpp
```diff
@@ -265,6 +333,8 @@ auto AutoupdateFileTest(
       // output_line_number.
       if (non_check_line.line_number() < 1) {
         new_lines.push_back(&non_check_line);
+        // These are ignored when calculated line counts.
```
zygoloid:
```suggestion
        // These are ignored when calculating line counts.
```

testing/file_test/autoupdate.cpp
```diff
@@ -265,6 +333,8 @@ auto AutoupdateFileTest(
...
+        ++file_offset_in_new_lines;
```
zygoloid: I'd find this easier to reason about with a `CHECK(file_offset_in_new_lines == new_lines.size());` after this line -- I think we expect all of these lines to be at the start of the output for the file, and they aren't part of the file contents.

testing/file_test/testdata/file_only_re_multi_file.carbon
```diff
@@ -0,0 +1,25 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+// AUTOUPDATE
+// CHECK:STDOUT: 3 args: `default_args`, `a.carbon`, `b.carbon`
+// CHECK:STDOUT: unattached message 1
+
+// --- a.carbon
+aaa
+// CHECK:STDOUT: file: a.carbon
+// CHECK:STDOUT: unattached message 2
```
zygoloid: Why do these messages appear here rather than at the start of the file?

Oh, presumably it's because we only print STDOUT lines immediately after a non-CHECK line, and that works in the main file because we can attach them to the AUTOUPDATE line. Perhaps we need to also print STDOUT CHECK lines at the start of the file in some cases?

testing/file_test/testdata/unattached_multi_file.carbon
```diff
@@ -6,12 +6,11 @@
 // CHECK:STDERR: unattached message 3
 // CHECK:STDERR: unattached message 4
 
+// CHECK:STDOUT: 3 args: `default_args`, `a.carbon`, `b.carbon`
+// CHECK:STDOUT: unattached message 1
+// CHECK:STDOUT: unattached message 2
```
zygoloid: I think we previously wanted to put these lines at the end of the overall test file, because they would be expected to describe the output of the overall test. Does that not work out with the current plan for multi-file output checking? (I don't have a strong opinion, if it's too hard or problematic to include them later.)

toolchain/driver/driver_test.cpp
```diff
@@ -126,32 +124,8 @@ TEST_F(DriverTest, DumpTokens) {
...
-                                     {"identifier", "1"},
-                                     {"has_trailing_space", "true"}},
-                  Yaml::MappingValue{{"index", "2"},
-                                     {"kind", "EndOfFile"},
-                                     {"line", "1"},
-                                     {"column", "12"},
-                                     {"indent", "1"},
-                                     {"spelling", ""}}}));
+  // Verify there is output without examining it.
+  EXPECT_FALSE(test_output_stream_.TakeStr().empty());
```
zygoloid: I'm fine with dropping the test of the contents, but maybe we should pass this through `Yaml::Value::FromText` to make sure it's at least valid YAML?

zygoloid (approved):
Looks good. Some minor things, but I don't think any of them is blocking if you'd like to land this sooner to avoid backsliding.

jonmeow:
Note, several comments refer to changes from #3214. I can backport them, but I want to flag that as this change is dependent upon that one (and it's awaiting review).

testing/file_test/autoupdate.cpp
jonmeow: Adding to #3214

testing/file_test/autoupdate.cpp
jonmeow: Adding to #3214

testing/file_test/autoupdate.cpp
jonmeow: Adding to #3214

testing/file_test/autoupdate.cpp
jonmeow: Adding to #3214

testing/file_test/testdata/file_only_re_multi_file.carbon
jonmeow: I tend to agree, but would like to consider it separately since this is how autoupdate is designed to work right now. I'm already asking a structural question on #3214 triggered by STDOUT issues, and #3219 is proposing a significant change for semantics too.

testing/file_test/testdata/unattached_multi_file.carbon
jonmeow: This is what I was trying to point out at https://github.com/carbon-language/carbon-lang/pull/3184#discussion_r1316265323, that starting with file_number=0 may not yield the right behavior. But really, we should consider #3219 and whether change this would/should have an impact on tests.

toolchain/driver/driver_test.cpp
jonmeow: Maybe `Yaml::Value::FromText` should be changed to validate that it's yaml, but it doesn't do that right now, so calling it wouldn't actually have an effect. I'll keep this in mind though for a future change.

jonmeow (commented):
Making changes #3214

PR #3219, opened 2023-09-12
Group by file in more cases, and autoupdate SemIR locs.

jonmeow:
At present, this starts replacing locations with offsets. If you prefer not to have that, I could add another configuration option to the line replacement to maintain the absolute line information at the end of the file. My actual goal here is to just reduce need to re-run autoupdate.


zygoloid (commented):
I think this is making the output less readable; the partial interleaving isn't working well for me. (If we properly interleaved for tests with multiple functions, that might be a different story.) Can we retain the absolute line numbers and rewrite them to new absolute line numbers? Or perhaps we could make autoupdate iterate until it converges like the old python script did?

Another option that might be interesting to try would be to move the printout of the `file` stanza to the end of the SemIR output. Maybe that'd make the interleaving work better?

jonmeow:
Closing this for now since I won't have time to work on it for a month.

PR #3220, opened 2023-09-12
Switch `GetNodeBlock` and `GetTypeBlock` to return an `ArrayRef`

zygoloid:
This removes the risk of accidentally performing a vector copy when calling these functions, and is a preparation step towards the new node block allocation design.

This required changing how we build call expressions. Instead of finishing the argument block and then later adding a return slot, we now delay finishing the argument block and checking for conversions to parameter types until after we've added the return slot to it.


toolchain/sem_ir/file.h
```diff
@@ -173,15 +178,22 @@ class File : public Printable<File> {
     return id;
   }
 
+  // Adds a node block with the given content, returning an ID to reference it.
+  auto AddNodeBlock(llvm::ArrayRef<NodeId> content) -> NodeBlockId {
```
jonmeow, line 182: If I understand correctly, you're adding this for later work, but you're not calling it in this PR. Is that right?

toolchain/check/context.cpp
```diff
@@ -69,6 +69,11 @@ auto Context::AddNodeToBlock(SemIR::NodeBlockId block, SemIR::Node node)
   return semantics_ir_->AddNode(block, node);
 }
 
+auto Context::AddNodeIdToBlock(SemIR::NodeBlockId block, SemIR::NodeId node_id)
+    -> void {
+  semantics_ir_->AddNodeId(block, node_id);
+}
```
jonmeow, line 75: This Context doesn't have wrappers for most of the Add/Get functions; maybe this one isn't needed? Contrasting with AddNodeToBlock, the CARBON_VLOG cannot by done by semantics_ir (which lacks vlog_stream). That added functionality means a wrapper is needed, whereas here there's no logging. Conceptually, the AddNode log is logging the initial creation, so I'm not arguing for it here either.

For reference, it looks like all but one of the AddNode calls goes through AddNodeToBlock, meaning basically all node adds are logged that way. The exception in handle_struct.cpp might be treated as a bug which should be using the wrapper to get logging -- but that's an observation unrelated to this PR.

toolchain/check/handle_call_expression.cpp
```diff
@@ -8,7 +8,7 @@
 namespace Carbon::Check {
 
 auto HandleCallExpression(Context& context, Parse::Node parse_node) -> bool {
-  auto refs_id = context.ParamOrArgEnd(
+  context.ParamOrArgEndNoPop(
```
jonmeow, line 13: I have some concerns about this flow, but maybe some comments here explaining it could help. Are you expecting other ParamOrArg constructs to use this?

What happens when this returns before ParamOrArgPop is called (e.g., line 19)? I think there may be a bug forgetting to call ParamOrArgPop.

I assume your goal here is to delay saving of the block; is that correct? Had you considered popping but returning the array and letting the caller deal with saving it, so that error paths don't need to do cleanup? i.e., versus ParamOrArgEnd combining ParamOrArgEndNoPop + ParamOrArgPop (dropping the second call is a bug), instead being ParamOrArgEndDelaySave + AddNodeBlock (dropping the second call is fine). (if I understand your direction correctly)

toolchain/sem_ir/file.h
```diff
@@ -173,15 +178,22 @@ class File : public Printable<File> {
...
+    NodeBlockId id(node_blocks_.size());
+    node_blocks_.push_back({});
+    node_blocks_.back().assign(content.begin(), content.end());
```
jonmeow, line 185: Why assign instead of append (also in AddTypeBlock)? It looks like assign might do a little extra work which wouldn't be relevant here.

jonmeow (commented):
I do prefer the readability result of AddTypeBlock taking a parameter, I just wish we could avoid the copies....

toolchain/check/context.cpp
jonmeow: I see you're removing this in #3221, so this is fine as-is.

toolchain/sem_ir/file.h
zygoloid: Yes, I thought I'd be calling it in this PR but it looks like the first call is added in #3223.

toolchain/check/handle_call_expression.cpp
zygoloid: Other `ParamOrArg` constructs would use this if they want the argument list to contain more elements after the final syntactic element. I'm not really expecting that to happen in other cases, so this might end up being specific to call expressions.

The early return was a bug; fixed, and test added.

The bigger picture goal is that blocks are finalized / saved as part of popping them from the node block stack, so delaying the pop is what we need to do in that model in order to delay saving the node block to SemIR.

toolchain/sem_ir/file.h
zygoloid: `node_blocks_.append(SmallVector<NodeId>(content.begin(), content.end()))` would create and initialize a temporary `SmallVector<NodeId>` and end up performing two copies of the content rather than one, in the case where the small vector is in its "small" state. We could use `emplace_back` to do this in one line, but ... this is being removed again in #3224 anyway so that'd just be churn.

jonmeow: I'm not suggesting to create a `SmallVector`, just:

`node_blocks_.back().append(content.begin(), content.end());`

For reference, the `assign` implementation seems to delegate directly to `append` (after the extra work):

template <typename ItTy, typename = EnableIfConvertibleToInputIterator<ItTy>>
  void [assign](https://llvm.org/doxygen/classllvm_1_1SmallVectorImpl.html#aaf8c4a3f62d6231ae323b77e21303e1c)([ItTy](https://llvm.org/doxygen/classItTy.html) in_start, [ItTy](https://llvm.org/doxygen/classItTy.html) in_end) {
    this->[assertSafeToReferenceAfterClear](https://llvm.org/doxygen/classllvm_1_1SmallVectorTemplateCommon.html#a709b8b7b7dd6a5ea1f9e8255b1dd5b13)(in_start, in_end);
    [clear](https://llvm.org/doxygen/classllvm_1_1SmallVectorImpl.html#aac0ea55010b7b1a301e65a0baea057aa)();
    [append](https://llvm.org/doxygen/classllvm_1_1SmallVectorImpl.html#a7efd1f0c1206d95e4fe01a9b49a57b82)(in_start, in_end);
  }

If this is removed, that's fine, I think it'd just be helpful to understand the choice.

jonmeow (approved):

PR #3221, opened 2023-09-13
Remove `Check::Context::AddNodeToBlock` and `SemIR::File::AddNode`.

zygoloid:
Nodes are now only ever added to the current block in the node block stack.

This is a preparation step towards the new node block allocation design.

### DIFF
--- a/toolchain/check/context.cpp
+++ b/toolchain/check/context.cpp
@@ -61,7 +61,7 @@ auto Context::VerifyOnFinish() -> void {
 }
 
 auto Context::AddNode(SemIR::Node node) -> SemIR::NodeId {
-  auto node_id = semantics_ir_->AddNodeNoBlock(node);
+  auto node_id = semantics_ir_->AddNodeInNoBlock(node);
   node_block_stack_.AddNodeId(node_id);
   CARBON_VLOG() << "AddNode " << node_block_stack_.Peek() << ": " << node
                 << "\n";
--- a/toolchain/check/context.cpp
+++ b/toolchain/check/context.cpp
@@ -9,6 +9,7 @@
 #include "common/check.h"
 #include "common/vlog.h"
 #include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/Sequence.h"
 #include "toolchain/check/declaration_name_stack.h"
 #include "toolchain/check/node_block_stack.h"
 #include "toolchain/diagnostics/diagnostic_kind.h"
@@ -60,18 +61,11 @@ auto Context::VerifyOnFinish() -> void {
 }
 
 auto Context::AddNode(SemIR::Node node) -> SemIR::NodeId {
-  return AddNodeToBlock(node_block_stack_.PeekForAdd(), node);
-}
-
-auto Context::AddNodeToBlock(SemIR::NodeBlockId block, SemIR::Node node)
-    -> SemIR::NodeId {
-  CARBON_VLOG() << "AddNode " << block << ": " << node << "\n";
-  return semantics_ir_->AddNode(block, node);
-}
-
-auto Context::AddNodeIdToBlock(SemIR::NodeBlockId block, SemIR::NodeId node_id)
-    -> void {
-  semantics_ir_->AddNodeId(block, node_id);
+  auto node_id = semantics_ir_->AddNodeNoBlock(node);
+  node_block_stack_.AddNodeId(node_id);
+  CARBON_VLOG() << "AddNode " << node_block_stack_.Peek() << ": " << node
+                << "\n";
+  return node_id;
 }
 
 auto Context::AddNodeAndPush(Parse::Node parse_node, SemIR::Node node) -> void {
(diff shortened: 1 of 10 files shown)


toolchain/sem_ir/file.h
```diff
@@ -148,16 +148,17 @@ class File : public Printable<File> {
     return name_scopes_[scope_id.index];
   }
 
-  // Adds a node to a specified block, returning an ID to reference the node.
-  auto AddNode(NodeBlockId block_id, Node node) -> NodeId {
+  // Adds a node to the node list, returning an ID to reference the node.
+  auto AddNodeNoBlock(Node node) -> NodeId {
```
jonmeow: Why "NoBlock" when there's no variant that adds it to a block?

toolchain/check/node_block_stack.h
```diff
@@ -44,7 +44,7 @@ class NodeBlockStack {
   }
 
   // Returns the top node block, allocating one if it's still invalid.
-  auto PeekForAdd() -> SemIR::NodeBlockId;
+  auto PeekForAdd(int depth = 0) -> SemIR::NodeBlockId;
```
jonmeow: With the edits, the comment should be updated since it doesn't always return the top node block.

toolchain/check/handle_if_statement.cpp
```diff
@@ -39,36 +40,31 @@ auto HandleIfStatementElse(Context& context, Parse::Node parse_node) -> bool {
...
+      // Branch from then block to else block, and start emitting the else
+      // block.
       context.node_stack()
           .PopAndDiscardSoloParseNode<Parse::NodeKind::IfCondition>();
-      context.AddNodeToBlock(
-          sub_block_id,
-          SemIR::Node::Branch::Make(parse_node,
-                                    context.node_block_stack().PeekForAdd()));
+      context.AddNode(SemIR::Node::Branch::Make(
+          parse_node, context.node_block_stack().PeekForAdd(/*depth=*/1)));
```
jonmeow: Where you call PeekForAdd with a depth, it seems to be to add a node. However, the PR description says "Nodes are now only ever added to the current block in the node block stack."

Is this going to go away, or could the PR description be clearer about how the adding flow works for Branch/BranchWithArg?

toolchain/check/context.cpp
```diff
@@ -178,45 +177,44 @@ auto Context::AddDominatedBlockAndBranchIf(Parse::Node parse_node,
...
       if (new_block_id == SemIR::NodeBlockId::Unreachable) {
         new_block_id = semantics_ir().AddNodeBlock();
       }
-      AddNodeToBlock(block_id,
-                     SemIR::Node::Branch::Make(parse_node, new_block_id));
+      AddNode(SemIR::Node::Branch::Make(parse_node, new_block_id));
     }
+    node_block_stack().Pop();
   }
   node_block_stack().Push(new_block_id);
```
jonmeow, line 194: I presume the purpose of the SwapTopBlocks approach is to avoid pushing the top block onto the node stack. I was going to comment to remove this Push variant since you seem to be avoiding using it in branch logic, but if it remains for this use-case, is SwapTopBlocks still a significant improvement?

toolchain/check/context.cpp
```diff
@@ -178,45 +177,44 @@ auto Context::AddDominatedBlockAndBranchIf(Parse::Node parse_node,
...
-  for (SemIR::NodeBlockId block_id : blocks) {
-    if (block_id != SemIR::NodeBlockId::Unreachable) {
+  for ([[maybe_unused]] auto _ : llvm::seq(num_blocks)) {
+    if (node_block_stack().is_current_block_reachable()) {
       if (new_block_id == SemIR::NodeBlockId::Unreachable) {
         new_block_id = semantics_ir().AddNodeBlock();
       }
-      AddNodeToBlock(block_id,
-                     SemIR::Node::Branch::Make(parse_node, new_block_id));
+      AddNode(SemIR::Node::Branch::Make(parse_node, new_block_id));
```
jonmeow, line 190: So this will make orphan blocks if there are non-final unreachable blocks? Is that necessary?

toolchain/sem_ir/file.h
zygoloid: There is a variant that adds the node to a block in `Check::Context` rather than `SemIR::File`. I think calling both functions `AddNode` would be error-prone -- it seems easy to accidentally use `context.semantics_ir().AddNode` instead of `context.AddNode`. Given that this function is the rare / "strange" operation, I thought it best to call the `Check::Context` one `AddNode` and make this name more verbose to draw attention to it doing something unusual.

Extended the function comment to point the reader at the Check::Context::AddNode function.

toolchain/check/node_block_stack.h
zygoloid: Done.

toolchain/check/handle_if_statement.cpp
zygoloid: We don't add nodes to the block returned by `PeekForAdd`, but we do add nodes that refer to that block ID.

I'm intending to rename `PeekForAdd` to something that better fits its purpose as a follow-up change.

toolchain/check/context.cpp
zygoloid: Hm, no, we can avoid `SwapTopBlocks` by deferring pushing the node until we're ready to start emitting it, and I think that makes conversion of `if` expressions and statements simpler and cleaner. Done, thanks!

toolchain/check/context.cpp
zygoloid: It shouldn't: we only create a new node block here if any of the predecessor blocks is reachable.

toolchain/sem_ir/file.h
jonmeow: Maybe it should be context.AddNodeToCurrentBlock instead of just AddNode? But really it's not part of this change.

Here, I think maybe "Without" instead of "No" would feel better for me, at least I'm not going to skim and read "AddNodeNoBlock" as "AddNodeToBlock".

Anyways, just my thoughts, not trying to block here.

toolchain/check/context.cpp
jonmeow: Ah, sorry, I misread the code.

toolchain/check/context.h
```diff
@@ -91,21 +83,21 @@ class Context {
                                     SemIR::NodeId cond_id)
       -> SemIR::NodeBlockId;
 
-  // Adds branches from the given list of blocks to a new block, for
-  // reconvergence of control flow, and pushes the new block onto the node
-  // block stack.
-  auto AddConvergenceBlockAndPush(
-      Parse::Node parse_node, std::initializer_list<SemIR::NodeBlockId> blocks)
+  // Adds branches from the top `num_blocks` on the node block stack to a new
+  // block, for reconvergence of control flow, pops the existing blocks, and
```
jonmeow: As you add "pops the existing blocks", the commas had previously indicated a subclause ("for reconvergence of control flow"), but now you're using them as a list (list item, subclause, list item, and list item). The subclause is now harder to distinguish from a list item. Maybe reformat this a bit further, either moving the subclause into parentheses or use bullets for the steps?

Same issue on AddConvergenceBlockWithArgAndPush

jonmeow (approved):
Looks good, thanks!

toolchain/sem_ir/file.h
zygoloid: Renamed to `AddNodeInNoBlock`.

toolchain/check/context.h
zygoloid: Yeah, that was pretty mangled. Rephrased.

PR #3222, opened 2023-09-13
Include information in the crash backtrace about which test we were running.

zygoloid:
Include information in the test crash output to identify which test we were running, and the corresponding resolved argument list, if either `bazel test` or autoupdate crashes.

### DIFF
--- a/testing/file_test/file_test_base.cpp
+++ b/testing/file_test/file_test_base.cpp
@@ -154,15 +154,6 @@ auto FileTestBase::ProcessTestFileAndRun(TestContext& context)
   CARBON_RETURN_IF_ERROR(
       DoArgReplacements(context.test_args, context.test_files));
 
-  // Pass arguments as StringRef.
-  llvm::SmallVector<llvm::StringRef> test_args_ref;
-  llvm::SmallVector<const char*> test_argv_ref;
-  test_args_ref.reserve(context.test_args.size());
-  for (const auto& arg : context.test_args) {
-    test_args_ref.push_back(arg);
-    test_argv_ref.push_back(arg.c_str());
-  }
-
   // Create the files in-memory.
   llvm::vfs::InMemoryFileSystem fs;
   for (const auto& test_file : context.test_files) {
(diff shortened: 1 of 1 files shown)


testing/file_test/file_test_base.cpp
```diff
@@ -163,6 +174,10 @@ auto FileTestBase::ProcessTestFileAndRun(TestContext& context)
     }
   }
 
+  // Add a stack trace entry for the test invocation.
+  llvm::PrettyStackTraceProgram stack_trace_entry(test_argv_ref.size(),
+                                                  test_argv_ref.data());
```
jonmeow: Can this be closer to where the arguments for it are initialized, either by moving this up or them down?

testing/file_test/file_test_base.cpp
```diff
@@ -147,9 +156,11 @@ auto FileTestBase::ProcessTestFileAndRun(TestContext& context)
 
   // Pass arguments as StringRef.
```
jonmeow: This is no longer accurate, can you please update it?

testing/file_test/file_test_base.cpp
```diff
@@ -147,9 +156,11 @@ auto FileTestBase::ProcessTestFileAndRun(TestContext& context)
...
   llvm::SmallVector<llvm::StringRef> test_args_ref;
+  llvm::SmallVector<const char*> test_argv_ref;
   test_args_ref.reserve(context.test_args.size());
```
jonmeow: Please either also reserve for your new list, or remove this for symmetry (it's a test so either way seems fine)

testing/file_test/file_test_base.cpp
```diff
@@ -147,9 +156,11 @@ auto FileTestBase::ProcessTestFileAndRun(TestContext& context)
 
   // Pass arguments as StringRef.
   llvm::SmallVector<llvm::StringRef> test_args_ref;
+  llvm::SmallVector<const char*> test_argv_ref;
```
jonmeow: I think it may be a little subtle with how this name differs from the above by one char in the middle. But, maybe "ref" isn't as appropriate given that it's not a StringRef. Maybe "test_args_for_trace"?

jonmeow (approved):
Good idea, I've had autoupdates crash and this seems like it'd be helpful.

testing/file_test/file_test_base.cpp
zygoloid: Done.

testing/file_test/file_test_base.cpp
zygoloid: Done.

testing/file_test/file_test_base.cpp
zygoloid: Done.

testing/file_test/file_test_base.cpp
zygoloid: Done.

PR #3223, opened 2023-09-13
Store contents of incomplete node blocks on the node block stack.

zygoloid:
`SemIR` now only stores complete blocks, and placeholders for blocks where we need a block ID before we know the block contents.

This is a preparation step towards the new node block allocation design.

### DIFF
--- a/toolchain/check/node_block_stack.cpp
+++ b/toolchain/check/node_block_stack.cpp
@@ -6,43 +6,64 @@
 
 #include "common/vlog.h"
 #include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/StringExtras.h"
 #include "toolchain/sem_ir/node.h"
 
 namespace Carbon::Check {
 
 auto NodeBlockStack::Push(SemIR::NodeBlockId id) -> void {
   CARBON_VLOG() << name_ << " Push " << stack_.size() << "\n";
-  CARBON_CHECK(stack_.size() < (1 << 20))
+  CARBON_CHECK(size() < (1 << 20))
       << "Excessive stack size: likely infinite loop";
-  stack_.push_back(id);
+  if (size_ == stack_.size()) {
+    stack_.emplace_back();
+  }
+  stack_[size_].Reset(id);
+  ++size_;
 }
 
 auto NodeBlockStack::PeekForAdd(int depth) -> SemIR::NodeBlockId {
-  CARBON_CHECK(static_cast<int>(stack_.size()) > depth) << "no such block";
-  int index = stack_.size() - depth - 1;
+  CARBON_CHECK(static_cast<int>(size()) > depth) << "no such block";
+  int index = size() - depth - 1;
   auto& slot = stack_[index];
-  if (!slot.is_valid()) {
-    slot = semantics_ir_->AddNodeBlock();
-    CARBON_VLOG() << name_ << " Add " << index << ": " << slot << "\n";
+  if (!slot.id.is_valid()) {
+    slot.id = semantics_ir_->AddNodeBlock();
+    CARBON_VLOG() << name_ << " Add " << index << ": " << slot.id << "\n";
   }
-  return slot;
+  return slot.id;
 }
 
 auto NodeBlockStack::Pop() -> SemIR::NodeBlockId {
-  CARBON_CHECK(!stack_.empty()) << "no current block";
-  auto back = stack_.pop_back_val();
-  CARBON_VLOG() << name_ << " Pop " << stack_.size() << ": " << back << "\n";
-  if (!back.is_valid()) {
+  CARBON_CHECK(!empty()) << "no current block";
+  --size_;
+  auto& back = stack_[size_];
+
+  // Finalize the block.
+  if (!back.content.empty() && back.id != SemIR::NodeBlockId::Unreachable) {
+    if (back.id.is_valid()) {
+      semantics_ir_->SetNodeBlock(back.id, back.content);
+    } else {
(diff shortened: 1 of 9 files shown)


toolchain/sem_ir/file.h
```diff
@@ -179,6 +170,15 @@ class File : public Printable<File> {
...
   }
 
+  // Sets the contents of an empty node block to the given content.
+  auto SetNodeBlock(NodeBlockId block_id, llvm::ArrayRef<NodeId> content)
+      -> void {
+    CARBON_CHECK(block_id != NodeBlockId::Unreachable);
+    CARBON_CHECK(node_blocks_[block_id.index].empty())
+        << "node block content set more than once";
+    node_blocks_[block_id.index].assign(content.begin(), content.end());
+  }
```
jonmeow, line 184: Maybe put keep the two AddNodeBlock variants adjacent?

toolchain/sem_ir/file.h
```diff
@@ -179,6 +170,15 @@ class File : public Printable<File> {
     return id;
   }
 
+  // Sets the contents of an empty node block to the given content.
+  auto SetNodeBlock(NodeBlockId block_id, llvm::ArrayRef<NodeId> content)
+      -> void {
+    CARBON_CHECK(block_id != NodeBlockId::Unreachable);
+    CARBON_CHECK(node_blocks_[block_id.index].empty())
+        << "node block content set more than once";
+    node_blocks_[block_id.index].assign(content.begin(), content.end());
```
jonmeow, line 183: I commented about assign vs append on earlier PRs, so just noting it here too.

toolchain/check/node_block_stack.cpp
```diff
@@ -6,43 +6,64 @@
...
 
 namespace Carbon::Check {
 
 auto NodeBlockStack::Push(SemIR::NodeBlockId id) -> void {
   CARBON_VLOG() << name_ << " Push " << stack_.size() << "\n";
-  CARBON_CHECK(stack_.size() < (1 << 20))
+  CARBON_CHECK(size() < (1 << 20))
       << "Excessive stack size: likely infinite loop";
-  stack_.push_back(id);
+  if (size_ == stack_.size()) {
```
jonmeow: Why size() versus size_? It might be helpful to prefer one or the other within a given function.

toolchain/check/node_block_stack.cpp
```diff
@@ -6,43 +6,64 @@
...
 
 auto NodeBlockStack::PrintForStackDump(llvm::raw_ostream& output) const
     -> void {
   output << name_ << ":\n";
-  for (auto [i, entry] : llvm::enumerate(stack_)) {
-    output << "\t" << i << ".\t" << entry << "\n";
+  for (const auto& [i, entry] : llvm::enumerate(stack_)) {
+    output << "\t" << i << ".\t" << entry.id << "\t{";
+    llvm::ListSeparator sep;
+    for (auto id : entry.content) {
```
jonmeow, line 63: Just to note, this might not fit on one line (typically would not, I think). We should probably consider in general how much we're printing in these, but noting the stack would also be very large, that's nothing new to this PR so fine as-is.

toolchain/check/node_block_stack.h
```diff
@@ -76,10 +70,27 @@ class NodeBlockStack {
...
   auto PrintForStackDump(llvm::raw_ostream& output) const -> void;
 
-  auto empty() const -> bool { return stack_.empty(); }
-  auto size() const -> size_t { return stack_.size(); }
+  auto empty() const -> bool { return size() == 0; }
+  auto size() const -> size_t { return size_; }
 
  private:
+  struct StackEntry {
+    StackEntry() { content.reserve(32); }
```
jonmeow, line 77: Please add comments here about how this constant was chosen, at least for future understanding, even if it's arbitrary.

I'm used to 60-line functions being common, so this seems small -- though maybe with flow control it breaks down closer to this size, I'm not really digging into details.

I assume the management means any growth from a large function is permanent and deliberate, but could have some unintended consequences...

toolchain/check/node_block_stack.h
```diff
@@ -90,9 +101,11 @@ class NodeBlockStack {
...
 
   // The actual stack.
-  // PushEntry and PopEntry control modification in order to centralize
-  // vlogging.
-  llvm::SmallVector<SemIR::NodeBlockId> stack_;
+  llvm::SmallVector<StackEntry> stack_;
+
+  // The size of the stack. Entries after this in `stack_` are kept around so
+  // that we can reuse the allocated buffer for their content.
+  size_t size_ = 0;
```
jonmeow: Given style's preference for int, why size_t?

jonmeow (approved):
Looks good, I think my comments are minor style things.

toolchain/sem_ir/file.h
zygoloid: Instead, renamed the first one to `AddNodeBlockId` to indicate that it is just allocating a node block ID but not actually storing anything in it.

toolchain/check/node_block_stack.h
zygoloid: Only to preserve the existing type for `size()`. But it looks like nothing really cares about that, so changed.

toolchain/check/node_block_stack.h
zygoloid: Added a comment saying this is arbitrary and should be replaced by something based on an actual measurement.

toolchain/check/node_block_stack.cpp
zygoloid: Done.

PR #3226, opened 2023-09-13
Fix handling of missing files.

jonmeow:
Parse/Check may still be called, so need to check for source. The test missed this because it unnecessarily specified a phase.

### DIFF
--- a/toolchain/driver/driver.cpp
+++ b/toolchain/driver/driver.cpp
@@ -422,6 +422,11 @@ class Driver::CompilationUnit {
 
   // Parses tokens. Returns true on success.
   auto RunParse() -> bool {
+    // Can be called when the file fails to load, so ensure there's source.
+    if (!source_) {
+      return false;
+    }
+
     LogCall("Parse::Tree::Parse", [&] {
       parse_tree_ = Parse::Tree::Parse(*tokens_, *consumer_, vlog_stream_);
     });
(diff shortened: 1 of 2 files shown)


toolchain/driver/driver.cpp
```diff
@@ -435,6 +440,11 @@ class Driver::CompilationUnit {
 
   // Check the parse tree and produce SemIR. Returns true on success.
   auto RunCheck(const SemIR::File& builtins) -> bool {
+    // Can be called when the file fails to load, so ensure there's source.
+    if (!source_) {
+      return false;
+    }
+
```
zygoloid: I don't feel strongly about this, but it might be more consistent and general for this to check for `parse_tree_` (and for `RunLower` to check for `sem_ir_`, and for `RunCodeGen` to check for `module_`).

zygoloid (approved):
I'm fine with this approach, but suggesting another alternative to consider.

toolchain/driver/driver.cpp
jonmeow: The reason I went this path is because it's valid to call these if there's no source_, but invalid to call RunLower if there's no sem_ir (if anything, that should CARBON_CHECK)

jonmeow: Added some CHECKs

PR #3229, opened 2023-09-14
Clean up interface of node block stack after recent changes.

zygoloid:
- `Peek()` no longer returns a useful value, because we often don't allocate a `NodeBlockId` until we finish building the block. Remove it and update both its callers.
- Rename `PeekForAdd()` to `PeekOrAdd()` since it's no longer used to get a block ID to add elements into.
- `PushForAdd()` was unused. Remove it.
- Add `NodeBlockStack::AddNode` to combine the operations of adding a node and inserting it into a block.

### DIFF
--- a/toolchain/check/context.cpp
+++ b/toolchain/check/context.cpp
@@ -61,10 +61,8 @@ auto Context::VerifyOnFinish() -> void {
 }
 
 auto Context::AddNode(SemIR::Node node) -> SemIR::NodeId {
-  auto node_id = semantics_ir_->AddNodeInNoBlock(node);
-  node_block_stack_.AddNodeId(node_id);
-  CARBON_VLOG() << "AddNode " << node_block_stack_.Peek() << ": " << node
-                << "\n";
+  auto node_id = node_block_stack_.AddNode(node);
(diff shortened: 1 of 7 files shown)


toolchain/check/context.cpp
```diff
@@ -61,10 +61,8 @@ auto Context::VerifyOnFinish() -> void {
 }
 
 auto Context::AddNode(SemIR::Node node) -> SemIR::NodeId {
-  auto node_id = semantics_ir_->AddNodeInNoBlock(node);
-  node_block_stack_.AddNodeId(node_id);
-  CARBON_VLOG() << "AddNode " << node_block_stack_.Peek() << ": " << node
-                << "\n";
+  auto node_id = node_block_stack_.AddNode(node);
+  CARBON_VLOG() << "AddNode: " << node << "\n";
```
jonmeow, line 65: You might consider moving this VLOG to the NodeBlockStack::AddNode, and removing this wrapper. Since node_block_stack is visible for callers, and NodeBlockStack can VLOG, nothing would really be lost and it'd remove some indirection.

But I'm going to go ahead and approve since I don't think this suggestion needs to be done in this PR.

jonmeow (approved):
Looks great!

PR #3231, opened 2023-09-14
Updates to generics design details, part 1

josh11b:
First step in updating `docs/design/generics/details.md`. It incorporates changes from proposals: #989 #2138 #2173 #2200 #2360 #2964 #3162 , but there are still more changes from those proposals to be made.

It also switches away from suggesting static-dispatch witness tables, and creates an appendix to describe that decision.

### DIFF
--- a/docs/design/generics/appendix-witness.md
+++ b/docs/design/generics/appendix-witness.md
@@ -188,7 +188,7 @@ It also could contain constants, to store the values of
 
 ### Example
 
-For example, given this `Vector` interface:
+For example, this `Vector` interface:
 
 ```carbon
 interface Vector {
@@ -213,7 +213,7 @@ class Vector {
 }
 ```
 
-The [impl of Vector for Point_Inline](details.md#inline-impl) would be a value
+The [impl of `Vector` for `Point_Inline`](details.md#inline-impl) would be a value
 of this type:
 
 ```
@@ -250,21 +250,22 @@ interface Container {
 }
 ```
 
-is represented by:
+could be represented by:
 
 ```
-class Iterator(Self:! type) {
+class Iterator {
+  var Self:! type;
   var Advance: fnty(this: Self*);
   ...
 }
-class Container(Self:! type) {
-  // Representation type for the iterator.
-  let IteratorType:! type;
+class Container {
+  var Self:! type;
+
   // Witness that IteratorType implements Iterator.
-  var iterator_impl: Iterator(IteratorType)*;
+  var IteratorType:! Iterator*;
 
   // Method
-  var Begin: fnty (this: Self*) -> IteratorType;
+  var Begin: fnty (this: Self*) -> IteratorType->Self;
   ...
 }
 ```
--- a/docs/design/generics/appendix-witness.md
+++ b/docs/design/generics/appendix-witness.md
@@ -213,8 +213,8 @@ class Vector {
 }
 ```
 
-The [impl of `Vector` for `Point_Inline`](details.md#inline-impl) would be a value
-of this type:
+The [impl of `Vector` for `Point_Inline`](details.md#inline-impl) would be a
+value of this type:
 
 ```
 var VectorForPoint_Inline: Vector  = {
--- a/docs/design/generics/appendix-witness.md
+++ b/docs/design/generics/appendix-witness.md
@@ -67,9 +67,9 @@ per method in the interface. However, in practice, it's more complex because it
 must model things like associated facets and interfaces.
 
 Witness tables are called "dictionary passing" in Haskell. Outside of generics,
-a [vtable](https://en.wikipedia.org/wiki/Virtual_method_table) is a witness
-table that witnesses that a class is a descendant of an abstract base class, and
-is passed as part of the object instead of separately.
+a [vtable](https://en.wikipedia.org/wiki/Virtual_method_table) is very similar
+to a witness table, "witnessing" the specific descendant of a base class.
+Vtables, however, are passed as part of the object instead of separately.
 
 ### Dynamic-dispatch witness table
 
--- a/docs/design/generics/appendix-witness.md
+++ b/docs/design/generics/appendix-witness.md
@@ -1,0 +1,270 @@
+# Generics appendix: Witness tables
+
+<!--
+Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+Exceptions. See /LICENSE for license information.
+SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+-->
+
+<!-- toc -->
+
+## Table of contents
+
+-   [Overview](#overview)
+-   [Terminology](#terminology)
+    -   [Witness tables](#witness-tables)
+    -   [Dynamic-dispatch witness table](#dynamic-dispatch-witness-table)
+    -   [Static-dispatch witness table](#static-dispatch-witness-table)
+-   [Limitations of witness tables](#limitations-of-witness-tables)
+    -   [Associated constants](#associated-constants)
+    -   [Blanket implementations](#blanket-implementations)
+    -   [Specialization](#specialization)
+    -   [Calling templated functions](#calling-templated-functions)
+-   [Implementing some Carbon generic features with witness tables](#implementing-some-carbon-generic-features-with-witness-tables)
+    -   [Overview](#overview-1)
+    -   [Example](#example)
+    -   [Associated facets example](#associated-facets-example)
+
+<!-- tocstop -->
+
+## Overview
+
+Witness tables are a strategy for implementing generics, specifically for
+allowing the behavior of a generic function to vary with the values of generic
+parameters. They have some nice properties:
+
+-   They can be used both for runtime and compile-time dispatch.
+-   They can support separate compilation even with compile-time dispatch.
+
+However, it can be a challenge to implement some features of a generic system
+with witness tables. This leads to limitations on the generic system, additional
+runtime overhead, or both.
+
+Swift uses witness tables for both static and dynamic dispatch, accepting both
+limitations and overhead. Carbon and Rust only use witness tables for dynamic
+dispatch, and apply limitations to control the runtime overhead when using that
+feature. As an implementation detail, Carbon compilers might also use witness
+tables for static dispatch, for example when the code conforms to the
+limitations of what witness tables support. However, part of the point of this
+document is to state the limitations and obstacles of doing that.
+
+## Terminology
+
+### Witness tables
+
+[Witness tables](https://forums.swift.org/t/where-does-the-term-witness-table-come-from/54334/4)
+are an implementation strategy where values passed to a compile-time type
+binding are compiled into a table of required functionality. That table is then
+filled in for a given passed-in type with references to the implementation on
+the original type. The generic is implemented using calls into entries in the
+witness table, which turn into calls to the original type. This doesn't
+necessarily imply a runtime indirection: it may be a purely compile-time
+separation of concerns. However, it insists on a full abstraction boundary
+between the generic user of a type and the concrete implementation.
+
+A simple way to imagine a witness table is as a struct of function pointers, one
+per method in the interface. However, in practice, it's more complex because it
+must model things like associated facets and interfaces.
+
+Witness tables are called "dictionary passing" in Haskell. Outside of generics,
+a [vtable](https://en.wikipedia.org/wiki/Virtual_method_table) is a witness
+table that witnesses that a class is a descendant of an abstract base class, and
+is passed as part of the object instead of separately.
+
+### Dynamic-dispatch witness table
+
+For dynamic-dispatch witness tables, actual function pointers are formed and
+used as a dynamic, runtime indirection. As a result, the generic code **will
+not** be duplicated for different witness tables.
+
+### Static-dispatch witness table
+
+For static-dispatch witness tables, the implementation is required to collapse
+the table indirections at compile time. As a result, the generic code **will**
+be duplicated for different witness tables.
+
+Static-dispatch may be implemented as a performance optimization for
+dynamic-dispatch that increases generated code size. The final compiled output
+may not retain the witness table.
+
+## Limitations of witness tables
+
+### Associated constants
+
+An interface with associated constants can use that to allow the signature of a
+function to vary. A similar issue arises with argument and return values
+involving `Self`. This adds to the cost of calling such functions, for example
+if they are not passed by pointer, then the generated code must support
+arguments and return values with a size only known at runtime.
+
+For this reason, Rust's dynamic trait dispatch system, trait objects, only works
+with traits that are
+["object safe,"](https://doc.rust-lang.org/reference/items/traits.html#object-safety)
+which includes a requirement that
+[all the associated types have specified values](https://github.com/rust-lang/rfcs/blob/master/text/0195-associated-items.md#trait-objects).
+This reduces the expressivity of Rust traits to the subset that could be
+supported by a C++ abstract base class.
+
+Swift instead supports types with size only known at runtime for its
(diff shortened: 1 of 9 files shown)


docs/design/generics/appendix-witness.md
```diff
@@ -0,0 +1,270 @@
...
+-   [associated facets](details.md#associated-facets)
+-   [type parameters](details.md#parameterized-interfaces)
+-   [interface requirements](details.md#interface-requiring-other-interfaces)
+
+It also could contain constants, to store the values of
+[associated constants](details.md#associated-constants), or the type's size.
+
+### Example
+
+For example, given this `Vector` interface:
```
zygoloid:
```suggestion
For example, this `Vector` interface:
```
(to match the sentence fragment after the example).

docs/design/generics/appendix-witness.md
```diff
@@ -0,0 +1,270 @@
...
+  fn Advance[addr self: Self*]();
+}
+
+interface Container {
+  let IteratorType:! Iterator;
+  fn Begin[addr self: Self*]() -> IteratorType;
+}
+```
+
+is represented by:
```
zygoloid:
```suggestion
could be represented by:
```

docs/design/generics/appendix-witness.md
```diff
@@ -0,0 +1,270 @@
...
+  // Representation type for the iterator.
+  let IteratorType:! type;
+  // Witness that IteratorType implements Iterator.
+  var iterator_impl: Iterator(IteratorType)*;
+
+  // Method
+  var Begin: fnty (this: Self*) -> IteratorType;
+  ...
+}
+```
```
zygoloid: Perhaps:

````suggestion
```
class Iterator {
  var Self:! type;
  var Advance: fnty(this: Self*);
  ...
}
class Container {
  var Self:! type;

  // Witness that IteratorType implements Iterator.
  var IteratorType:! Iterator*;

  // Method
  var Begin: fnty (this: Self*) -> IteratorType->Self;
  ...
}
```
````

to match the translation used in the previous section?

File: docs/design/generics/details.md [hunk 4]
```diff
@@ -252,57 +262,199 @@ have different definitions for `Add` and `Scale`, so we say their definitions
...
+entity affect a class' API, then that is mentioned with an `extend` declaration
+in the `class` definition.
 
-**Comparison with other languages:** Rust defines implementations lexically
-outside of the `class` definition. This Carbon approach means that a type's API
-is described by declarations inside the `class` definition and doesn't change
+**Comparison with other languages:** Rust only defines implementations lexically
+outside of the `class` definition. Carbon's approach results in every type's API
+is described by declarations inside its `class` definition and doesn't change
 afterwards.
```
zygoloid: I think this needs a grammar tweak.
```suggestion
outside of the `class` definition. Carbon's approach results in every type's API
being described by declarations inside its `class` definition and not changing
afterwards.
```
or
```suggestion
outside of the `class` definition. Carbon's approach results in the property that every type's API
is described by declarations inside its `class` definition and doesn't change
afterwards.
```

docs/design/generics/details.md
```diff
@@ -252,57 +262,199 @@ have different definitions for `Add` and `Scale`, so we say their definitions
...
+  }
+  fn Scale[self: Self](v: f64) -> Self {
+    return {.x = self.x * v, .y = self.y * v};
+  }
+}
+```
+
+Since `extend impl` may only be used inside the class definition, out-of-line
+definitions do not contribute to the class' API unless there is a corresponding
+[forward declaration in the class definition using `extend](#forward-impl-declaration).
```
zygoloid:
```suggestion
[forward declaration in the class definition using `extend`](#forward-impl-declaration).
```

docs/design/generics/details.md
```diff
@@ -252,57 +262,199 @@ have different definitions for `Add` and `Scale`, so we say their definitions
...
+    return {.x = self.x + b.x, .y = self.y + b.y};
+  }
+  fn Scale[self: Self](v: f64) -> Self {
+    return {.x = self.x * v, .y = self.y * v};
+  }
+}
+```
+
+Since `extend impl` may only be used inside the class definition, out-of-line
+definitions do not contribute to the class' API unless there is a corresponding
```
zygoloid:
```suggestion
definitions do not contribute to the class's API unless there is a corresponding
```
Our [style guide](https://developers.google.com/style/possessives) wants an extra "s" for a possessive singular that ends in "s".

docs/design/generics/details.md
```diff
@@ -252,57 +262,199 @@ have different definitions for `Add` and `Scale`, so we say their definitions
...
+An out-of-line `impl` declaration is allowed to be defined in a different
+library from `Point_OutOfLine`, restricted by
+[the coherence/orphan rules](#orphan-rule) that ensure that the implementation
+of an interface can't change based on imports. In particular, the `impl`
+declaration is allowed in the library defining the interface (`Vector` in this
+case) in addition to the library that defines the type (`Point_OutOfLine` here).
+This (at least partially) addresses
+[the expression problem](https://eli.thegreenplace.net/2016/the-expression-problem-and-its-solutions).
+
+You can't use `extend` outside the class definition, so an `impl` declarations
```
zygoloid:
```suggestion
You can't use `extend` outside the class definition, so an `impl` declaration
```

docs/design/generics/details.md
```diff
@@ -252,57 +262,199 @@ have different definitions for `Add` and `Scale`, so we say their definitions
...
+in a different library will never affect the class' API. This means that the API
```
zygoloid:
```suggestion
in a different library will never affect the class's API. This means that the API
```

docs/design/generics/details.md
```diff
@@ -252,57 +262,199 @@ have different definitions for `Add` and `Scale`, so we say their definitions
...
+of a class such as `Point_OutOfLine` doesn't change based on what is imported.
+It would be particularly bad if two different libraries implemented interfaces
+with conflicting names that both affected the API of a single type. As a
+consequence of this restriction, you can find all the names of direct members
+(those available by [simple member access](terminology.md#simple-member-access))
+of a type in the definition of that type. The only thing that may be in another
```
zygoloid: I think there's still the possibility that a type `extend`s a type that depends on a template parameter:
```carbon
class A(template T:! type) {
  extend base: T;
}
```
In that case, I think the direct members of `A` (for arbitrary `T`) could be written in any library in the program.

I think `T` must be a template parameter in this case, because if it's a checked generic parameter, we wouldn't know that it's an extensible class.

docs/design/generics/details.md
```diff
@@ -628,32 +719,36 @@ acts like a [supertype](https://en.wikipedia.org/wiki/Subtyping) of any `T`
 implementing `Vector`.
 
 For name lookup purposes, an archetype is considered to have
```
zygoloid:
```suggestion
For name lookup purposes, an archetype is considered to
```

docs/design/generics/details.md
```diff
@@ -1099,46 +1138,37 @@ constraint RenderableAndEndOfGame {
...
 }
 ```
 
-Reserving the name when there is a conflict is part of resolving what happens
-when you combine more than two type-of-types. If `x` is forbidden in `A`, it is
-forbidden in `A & B`, whether or not `B` defines the name `x`. This makes `&`
-associative and commutative, and so it is well defined on sets of interfaces, or
-other type-of-types, independent of order.
+Note that `&` is associative and commutative, and so it is well defined on sets
+of interfaces, or other facet types, independent of order.
```
zygoloid, line 1150: "Union minus conflicts" isn't associative if taken literally, because `(A & B) & C` will have the name `N` mean `C.N` but `A & (B & C)` will have the name `N` mean `A.N`, assuming that `A`, `B`, and `C` all have an `N`. For that reason, I prefer phrasing the rule as simply taking the union (with an ambiguity rule in name lookup to deal with conflicts).

docs/design/generics/details.md
```diff
@@ -1014,12 +1059,12 @@ fn PrintDrawPrint[T1:! PrintAndRender](x1: T1) {
...
-## Combining interfaces by anding type-of-types
+## Combining interfaces by anding facet types
 
 In order to support functions that require more than one interface to be
-implemented, we provide a combination operator on type-of-types, written `&`.
-This operator gives the type-of-type with the union of all the requirements and
-the union of the names minus any conflicts.
+implemented, we provide a combination operator on facet types, written `&`. This
+operator gives the facet type with the union of all the requirements and the
+union of the names minus any conflicts.
```
zygoloid:
```suggestion
union of the names.
```
(See next comment for rationale.)

docs/design/generics/details.md
```diff
@@ -1099,46 +1138,37 @@ constraint RenderableAndEndOfGame {
...
-See [#531](https://github.com/carbon-language/carbon-lang/issues/531) for the
-discussion.
-
-**Future work:** We may want to define another operator on type-of-types for
-adding requirements to a type-of-type without affecting the names, and so avoid
+**Future work:** We may want to define another operator on facet types for
+adding requirements to a facet type without affecting the names, and so avoid
 the possibility of name conflicts. Note this means the operation is not
 commutative. If we call this operator `[&]`, then `A [&] B` has the names of `A`
 and `B [&] A` has the names of `B`.
```
zygoloid: Is this still future work? We can spell `A [&] B` as `A where .Self impls B` today.

docs/design/generics/details.md
```diff
@@ -1512,43 +1551,18 @@ eventually be provided.
...
-}
-// Will prefer the following overload when it matches
-// since it is more specific.
-fn SearchInSortedList[IterT:! RandomAccessIntIterator]
-    (begin: IterT, end: IterT, needle: i32) -> bool {
-  ... // does binary search
-}
-```
+If one interface extends another, that gives the information to the compiler
+that the extending interface is implemented for a subset of types as the
```
zygoloid: "A is implemented for a subset of types as B" reads a bit strangely to me here. Not sure what to suggest, though: something like "A is implemented for a subset of the types for which B is implemented" seems very wordy. Maybe "every type implementing B also implements A", but that makes the subset relation a bit less clear.

docs/design/generics/details.md
```diff
@@ -1512,43 +1551,18 @@ eventually be provided.
...
+extended interface. This can be used to detect unreachable code.
 
-This would be an example of the more general rule that an interface `A`
-requiring an implementation of interface `B` means `A` is more specific than
-`B`.
+For example, the [`impl` prioritization rule](#prioritization-rule) is used to
+pick between `impl` declarations based on an explicit priority ordering given by
+the user. If the broader interface is prioritized over the more specific
```
zygoloid:
```suggestion
the developer. If the broader interface is prioritized over the more specific
```
If I remember correctly, this is our preferred term.

docs/design/generics/details.md
```diff
@@ -1923,9 +1947,8 @@ interface DrawingContext {
 impl Window as DrawingContext { ... }
 ```
 
-An adapter can make that much more convenient by making a compatible type where
-the interface is [implemented internally](terminology.md#extending-an-impl).
-This avoids having to
+An adapter can make that more convenient by making a compatible type that does
+extends the implementation of the interface. This avoids having to
```
zygoloid:
```suggestion
extend the implementation of the interface. This avoids having to
```

docs/design/generics/details.md
```diff
@@ -1943,7 +1966,20 @@ fn Render(w: Window) {
 }
 ```
 
-### Adapter with stricter invariants
+**Note:** Another way to achieve this is to use a local symbolic facet constant:
+
+```carbon
+fn Render(w: Window) {
+  let DrawInWindow:! Draw = Window;
+  let d: DrawInWindow = w as DrawInWindow;
```
zygoloid: Possibly:
```suggestion
  let d: DrawInWindow = w;
```
but I'm also happy with the `as` if you'd prefer to be explicit in this example.

docs/design/generics/appendix-witness.md
```diff
@@ -0,0 +1,270 @@
...
+  var Self:! type;
+  // `fnty` is placeholder syntax for a "function type",
+  // so `Add` is a function that takes two `Self` parameters
+  // and returns a value of type `Self`.
+  var Add: fnty(a: Self, b: Self) -> Self;
+  var Scale: fnty(a: Self, v: f64) -> Self;
+}
+```
+
+The [impl of Vector for Point_Inline](details.md#inline-impl) would be a value
```
zygoloid:
```suggestion
The [impl of `Vector` for `Point_Inline`](details.md#inline-impl) would be a value
```

docs/design/generics/terminology.md
```diff
@@ -651,42 +649,31 @@ of another interface, plus some additional API. Types implementing the extended
...
+and
+[trait objects in Rust](https://doc.rust-lang.org/book/ch17-02-trait-objects.html).
+Note that this often comes with limitations, since for example it is much more
+difficult to support when the associated constants of the interface are not
+known.
+
+Typically a reference to the witness table will be passed separately from the
+object, unlike a
+[virtual method table](https://en.wikipedia.org/wiki/Virtual_method_table),
+which otherwise is very similar to a witness table, "witnessing" the specific
```
zygoloid, line 675: In the witness tables appendix, you say that a vtable is a witness table. I don't think it really matters how we describe the relationship here but I think it makes sense to be consistent.

docs/design/generics/terminology.md
```diff
@@ -651,42 +649,31 @@ of another interface, plus some additional API. Types implementing the extended
...
-### Static-dispatch witness table
-
-For static-dispatch witness tables, the implementation is required to collapse
-the table indirections at compile time. As a result, the generic code **will**
-be duplicated for different witness tables.
-
-Static-dispatch may be implemented as a performance optimization for
-dynamic-dispatch that increases generated code size. The final compiled output
-may not retain the witness table.
+## Dynamic-dispatch witness table
```
zygoloid, line 652: I think this section should link to the new appendix.

zygoloid (commented):
Thanks!

docs/design/generics/details.md
josh11b: Reworded to include entities referenced by `extend` declarations.

docs/design/generics/details.md
josh11b: I made the change from the earlier suggestion. Is there something more I should say or is the text here now clear and accurate?

docs/design/generics/details.md
josh11b: Good point. I've updated the text. It will get a link to a relevant section in the next PR that adds that section.

docs/design/generics/details.md
josh11b: How about this new text?

docs/design/generics/details.md
josh11b: Changed but added a comment.

docs/design/generics/terminology.md
josh11b: Added.

docs/design/generics/terminology.md
josh11b: Updated the appendix.

docs/design/generics/terminology.md
```diff
@@ -675,6 +675,9 @@ object, unlike a
 which otherwise is very similar to a witness table, "witnessing" the specific
 descendant of a base class.
 
+Carbon approach to using witness tables is detailed in an
```
zygoloid: This is a bit bullet-pointy; maybe
```suggestion
The Carbon approach to using witness tables is detailed in an
```
or
```suggestion
Carbon's approach to using witness tables is detailed in an
```

zygoloid (approved):
Looks great!

PR #3233, opened 2023-09-14
When starting a split file, try adding stdout lines.

jonmeow:
Addresses https://github.com/carbon-language/carbon-lang/pull/3217#discussion_r1323581989


chandlerc (approved):
Nice!

PR #3234, opened 2023-09-14
When there are split files, put unattached stdout at the end of the last file.

jonmeow:
This is consistent with the goal of trying to push unattached stdout to the bottom of the test file. The current logic had a single file in mind, and was not adapted for split file logic.

Addresses https://github.com/carbon-language/carbon-lang/pull/3217#discussion_r1323584624

Note this depends on #3233


chandlerc (approved):
Also very nice!

PR #3235, opened 2023-09-14
Make toolchain/testing/file_test more robust

josh11b:
Support `toolchain/` being a subdirectory rather than the project root.

### DIFF
--- a/toolchain/testing/file_test.cpp
+++ b/toolchain/testing/file_test.cpp
@@ -88,9 +88,10 @@ class ToolchainFileTest : public FileTestBase {
  private:
   // Returns the toolchain subdirectory being tested.
   static auto GetComponent(llvm::StringRef test_name) -> llvm::StringRef {
+    // This handles cases where the toolchain directory may be copied into a repository that doesn't put it at the root.
     auto pos = test_name.find("toolchain/");
     CARBON_CHECK(pos != llvm::StringRef::npos) << test_name;
-    test_name = test_name.drop_front(pos + sizeof("toolchain/") - 1);
+    test_name = test_name.drop_front(pos + strlen("toolchain/"));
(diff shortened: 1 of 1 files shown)


toolchain/testing/file_test.cpp
```diff
@@ -88,7 +88,9 @@ class ToolchainFileTest : public FileTestBase {
  private:
   // Returns the toolchain subdirectory being tested.
   static auto GetComponent(llvm::StringRef test_name) -> llvm::StringRef {
-    CARBON_CHECK(test_name.consume_front("toolchain/"));
+    auto pos = test_name.find("toolchain/");
+    CARBON_CHECK(pos != llvm::StringRef::npos) << test_name;
+    test_name = test_name.drop_front(pos + sizeof("toolchain/") - 1);
```
jonmeow:
```suggestion
    test_name = test_name.drop_front(pos + strlen("toolchain/"));
```

Is this easier for you to read? I think the compiler-optimized result is equivalent for a string literal like this.

toolchain/testing/file_test.cpp
```diff
@@ -88,7 +88,9 @@ class ToolchainFileTest : public FileTestBase {
  private:
   // Returns the toolchain subdirectory being tested.
   static auto GetComponent(llvm::StringRef test_name) -> llvm::StringRef {
-    CARBON_CHECK(test_name.consume_front("toolchain/"));
+    auto pos = test_name.find("toolchain/");
```
jonmeow, line 93:
```suggestion
    // This handles cases where the toolchain directory may be copied into a repository that doesn't put it at the root.
    auto pos = test_name.find("toolchain/");
```

What do you think of commenting why this isn't the more trivial consume? That way it's less likely to get "simplified". :)

jonmeow (approved):
Thanks!

PR #3236, opened 2023-09-14
Instead of passing lines to AddCheckLines, pass a bool to_file_end

jonmeow:
This is just aiming for an API simplification. Instead of the typical `non_check_line_->line_number(), non_check_line_->indent()` with an `INT_MAX, ""` special-case, instead pass `to_file_end` and let the functions infer whether `non_check_line_` should be examined. (I'm also hoping removing `INT_MAX` improves understandability)

Note this depends on #3234


chandlerc (approved):
I'm happy with either of these approaches.

I could imagine a third one where we have something that more explicitly models the "file end" concept, and maybe *wraps* `INT_MAX` so that some of the code doesn't have to check the bool explicitly.

But I think if this is an improvement for you, let's go with it for now rather than making anything more complicated.

PR #3239, opened 2023-09-14
Detect invalid yaml.

jonmeow:
Previous code could fail silently, only caught with yaml output mismatches. This uses ErrorOr to be explicit about the error detection; recovery isn't supported because we should only print valid yaml (in tests, at least).

Also moves the test helpers from //toolchain/base to //toolchain/testing.

Switches tree_test to use the matchers so that, on mismatch, gtest prints a path to the mismatch (the straight value means it was just printing "not equal").

Addresses https://github.com/carbon-language/carbon-lang/pull/3217#discussion_r1323586836


chandlerc (approved):
I have a mild preference to just add this to //testing rather than having two layers of testing libraries. But not a huge deal, and fine to fix forward.

PR #3241, opened 2023-09-15
Fix use of backtick in shell code.

jonmeow:
The way this is written, it's executed, not a comment.

`/home/runner/work/_temp/1007824c-f2f7-49dd-ab5f-f5f5364a1864.sh: line 1: --local_*_resources: command not found`
https://github.com/carbon-language/carbon-lang/actions/runs/6203167016/job/16843338592


chandlerc (approved):
😭

But also, good catch!

PR #3243, opened 2023-09-15
Add cache of template instantiations.

### LINKED ISSUE #2951: Exponential run time with deeply nested tuple type

geoffromer:
Also some drive-by comment and trace improvements.

Closes #2951

### DIFF
--- a/explorer/interpreter/interpreter.cpp
+++ b/explorer/interpreter/interpreter.cpp
@@ -516,49 +516,62 @@ auto Interpreter::InstantiateType(Nonnull<const Value*> type,
                            << source_loc << ")\n";
   }
 
+  const Value* value = nullptr;
   switch (type->kind()) {
     case Value::Kind::VariableType: {
       CARBON_ASSIGN_OR_RETURN(
-          Nonnull<const Value*> value,
+          value,
           todo_.ValueOfNode(&cast<VariableType>(*type).binding(), source_loc));
       if (const auto* location = dyn_cast<LocationValue>(value)) {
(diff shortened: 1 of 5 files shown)


zygoloid (approved):
Awesome, this is very clean. Is it possible to add a testcase that demonstrates this is working -- perhaps one that results in an exponential number of instantiations without caching and a linear number with, instantiated with a suitably large N?

PR #3246, opened 2023-09-20
Perform in-place initialization for tuples and structs

zygoloid:
This implements initializing expression semantics for structs and tuples, following #2006 and discussions since.

Tuple and (and analogously, struct) literals are treated as having a mixed expression category that is later resolved based on how the literal is used, as either a tuple initializer or a tuple value, at which point we create a `TupleInit` or `TupleValue` that represents the formation of the tuple initializer or tuple value from the tuple literal.

There's quite a lot of TODOs here, and the SemIR representation is still not quite right, but this seems like a good place to checkpoint some incremental progress.


toolchain/check/context.cpp
```diff
@@ -251,6 +251,31 @@ auto Context::is_current_position_reachable() -> bool {
          SemIR::TerminatorKind::Terminator;
 }
 
+namespace {
+class CopyOnWriteBlock {
```
chandlerc, line 264: Some comments to document how this is supposed to work?

Seems especially important as C-o-W semantics can be a bit subtle.

chandlerc (approved):
Generally makes sense as a check-point. Kinda nice how much the Lowered LLVM improves already.

One place where I think some comments are worth adding now.

I'll also mention that some of the switches have enough code in their cases that I suspect it's time to factor out functions for each case and just dispatch to those functions. But not expecting that in this PR or necessarily the follow-up, fine to do that when you get to a good point.

PR #3248, opened 2023-09-20
Stop creating fake integer literals as indexes for tuple initialization.

zygoloid:
This makes struct and tuple initialization basically identical; unify them.


toolchain/check/context.cpp
```diff
@@ -371,18 +344,24 @@ auto Context::Initialize(Parse::Node parse_node, SemIR::NodeId target_id,
...
-                parse_node, inner_target_type, target_id,
-                SemIR::MemberIndex(i)));
-
+            auto inner_target_id =
+                AddNode(is_tuple ? SemIR::Node::TupleAccess::Make(
+                                       parse_node, inner_target_type, target_id,
+                                       SemIR::MemberIndex(i))
+                                 : SemIR::Node::StructAccess::Make(
+                                       parse_node, inner_target_type, target_id,
+                                       SemIR::MemberIndex(i)));
```
chandlerc, line 351: If you factor this into a helper function, you could use a function template and take `*Access` and `*Init` as type parameters with two calls.

Anyways, fine for fixing forward or to not if this doesn't end up working out.

chandlerc (approved):

toolchain/check/context.cpp
zygoloid: Nice idea. I want to refactor this in a different direction first but will return to this :)

PR #3249, opened 2023-09-20
Remove nearly all uses of StubReferences.

zygoloid:
The speculative insertion of StubReferences after elements in an argument list turned out to not be necessary, because we decided we want to insert per-argument initialization steps after all arguments are evaluated, rather than interleaving them. The StubReferences we insert are causing some minor code complexity, so remove them.

We still create StubReferences when performing patch-ups of already-emitted code, but we no longer ever need to look through them when determining whether an initializer was a literal or when evaluating a type expression.


toolchain/check/context.h
```diff
@@ -211,30 +200,28 @@ class Context {
...
   // list. This is the same as `ParamOrArgEndNoPop` followed by `ParamOrArgPop`.
-  auto ParamOrArgEnd(bool for_args, Parse::NodeKind start_kind)
-      -> SemIR::NodeBlockId;
+  auto ParamOrArgEnd(Parse::NodeKind start_kind) -> SemIR::NodeBlockId;
 
   // Saves a parameter from the top block in node_stack_ to the top block in
-  // params_or_args_stack_. If for_args, adds a StubReference of the previous
-  // node's result to the IR.
-  //
-  // This should only be called by other ParamOrArg functions, not directly.
```
chandlerc, line 236: Is this part also no longer true? Somewhat surprised.

chandlerc (approved):
Nice!

toolchain/check/context.h
zygoloid: This is a drive-by fix: this comment was not updated when `handle_call_expression.cpp` started calling this function directly to register the return slot argument.

PR #3250, opened 2023-09-21
Refactor finalization of initializers.

zygoloid:
Factor out the common code to find the return slot for an initializing expression, and use it to simplify the two different ways we finalize initializers, as either initializing a temporary or initializing some specific object.

This slightly changes the SemIR we create for function calls: instead of rewriting the Call node to have a different destination and replacing its temporary with a `no_op`, we now replace its temporary with a `stub_reference` to the new destination. This results in the same amount of SemIR being produced, but allows calls and other kinds of initializers to be handled uniformly.


toolchain/check/context.cpp
```diff
@@ -493,121 +493,92 @@ auto Context::ConvertToValueOrReferenceExpression(SemIR::NodeId expr_id,
...
-        auto temporary_id = std::exchange(
-            semantics_ir().GetNodeBlock(refs_id).back(), target_id);
-        auto temporary = semantics_ir().GetNode(temporary_id);
-        CARBON_CHECK(temporary.kind() == SemIR::NodeKind::TemporaryStorage)
-            << "Return slot for function call does not contain a temporary; "
-            << "initialized multiple times? Have " << temporary;
-        semantics_ir().ReplaceNode(
-            temporary_id, SemIR::Node::NoOp::Make(temporary.parse_node()));
+      if (!semantics_ir.GetFunction(callee_id).return_slot_id.is_valid()) {
+        return SemIR::NodeId::Invalid;
```
chandlerc, line 522: Hmm, is this going to cause the caller to restructure how the return is managed (making it not in-place)? For an invalid function, that seems like it might result in cascading errors. Should this instead return a return slot that is marked invalid? Or in some other way model that distinction?

chandlerc (approved):
LG!

I've left a comment about a recovery case, but its not blocking.

toolchain/check/context.cpp
zygoloid: The situation being handled here is that the function doesn't have a return slot (which is modeled by the return slot ID being marked as invalid). Unfortunately, we fairly pervasively use `SemIr::NodeId::Invalid` to mean "none" (not a valid ID) rather than "the program is invalid". Perhaps we should rename it.

PR #3251, opened 2023-09-22
Add "ERROR" to all error diagnostics

geoffromer:
This makes the difference between errors and lower-level diagnostics visible to users, and aligns the toolchain's behavior with the expectations in `driver_fuzzer.cpp`.


toolchain/check/testdata/array/fail_bound_overflow.carbon
```diff
@@ -4,7 +4,7 @@
 //
 // AUTOUPDATE
 
-// CHECK:STDERR: fail_bound_overflow.carbon:[[@LINE+3]]:32: Invalid array expression.
+// CHECK:STDERR: ERROR: fail_bound_overflow.carbon:[[@LINE+3]]:32: Invalid array expression.
```
zygoloid: Putting the `ERROR: ` at the start will break the ability for existing tools built around C and C++ compilers to parse these diagnostics. (Eg, vim and emacs have parsers that expect lines to start with `file:line:column: `.) Can we put it between the location and the message instead?

zygoloid (approved):

PR #3252, opened 2023-09-22
Place the computation of the destination of an initialization into the return slot.

zygoloid:
Fix a bug where we would perform the computation of the return location in SemIR after we have already used it in some cases, leading to assertion failures during lowering. Instead, accumulate a sequence of instructions to compute the return location in a temporary block, and overwrite the return slot with those instructions when we perform initialization.

StubReference is replaced by a more general SpliceBlock node, that takes a code block and a result value, executes the instructions in the block, and produces the result. This is used in the uncommon case where more than one instruction is required to compute the return slot, which can happen if we need to first emit a temporary and then index into it, or if we need to perform multiple levels of indexing before we reach an entity to initialize.


toolchain/check/context.cpp
```diff
@@ -681,39 +763,43 @@ static auto ConvertTupleToArray(Context& context, SemIR::Node tuple_type,
...
-    auto src_id =
-        !literal_elems.empty()
-            ? literal_elems[i]
-            : context.AddNode(SemIR::Node::TupleIndex::Make(
-                  value.parse_node(), src_type_id, value_id, index_id));
-    auto init_id =
-        context.InitializeAndFinalize(value.parse_node(), target_id, src_id);
+    auto src_id = !literal_elems.empty()
+                      ? literal_elems[i]
+                      : context.AddNode(SemIR::Node::TupleAccess::Make(
```
chandlerc, line 792: Should this be in `target_block` as well?

```suggestion
                      : target_block.AddNode(SemIR::Node::TupleAccess::Make(
```

And if so -- this seems awfully brittle. Is there anything to check / ensure we get this right?

chandlerc (approved):
See one comment in-line. I'm a bit nervous about getting this right consistently, but overall it makes sense.

That said, LG with the fix below if correct (provided test coverage), and also LG if I just missed the reason why that case is different.

toolchain/check/context.cpp
zygoloid: Added a comment to explain.

PR #2687, opened 2023-03-16
Termination algorithm for impl selection

josh11b:
This proposal replaces the termination algorithm for `impl` selection. The previous algorithm relied on a recursion limit, which is counter to [our goal for predictability](/docs/design/generics/goals.md#predictability).

The replacement is to terminate if any `impl` declaration depends transitively on itself with a "strict superset" type as one of its parameters.

Issue #2495, opened 2022-12-22
Keyword to use in place of `is` in `where`...`is` constraints


josh11b:
### Summary of issue:

A question arose in proposal #2483 about what keyword should be used to replace `is`. The particular syntax in question is when expressing a constraint that a type variable implements an interface or satisfies a named constraint, for example `where .ElementType is Ordered`. (It should be noted that this case is not expected to be particularly frequent.)

Proposal #2483 suggests replacing `is` with `impls`, but @jonmeow raised concerns that would be ambiguous with the plural of `impl`. My take is that there is a relationship between these two words that I want to preserve, that an `impls` constraint can be satisfied by writing an `impl` declaration, and in the examples I've looked at I had no problem disambiguating. I expect this is because this ambiguity commonly arises in English for words (like "pile") that can be used as both nouns and verbs, and so I have a lot of practice disambiguating.

I've also raised this question [in the #naming channel on Discord](https://discord.com/channels/655572317891461132/963846118964350976/10556030391484457570).

zygoloid:
Another possibility would be to use the same syntax that we use in an interface or named constraint:
- `where T is C` becomes `where impl T as C`
- `where .Self is C` becomes `where impl as C`

This approach would also give syntax for something that's hard to write at the moment:
- `AddWith(i32) where extends ImplicitAs(.Self.(AddWith(i32).Result))`, meaning the same thing as `AddWith(i32) & ImplicitAs(.Self.(AddWith(i32).Result))`, except the latter doesn't type-check because we don't know that `.Self` implements `AddWith(i32)` early enough.

However, `where T impls C` seems more readable to me, and better parallels the syntax of other `where` constraints.

jonmeow:
> I expect this is because this ambiguity commonly arises in English for words (like "pile") that can be used as both nouns and verbs, and so I have a lot of practice disambiguating.

I want to directly address this, because I think this is in conflict with Carbon's "Code that is easy to read, understand, and write" goal. We're going to have a lot of non-native, even non-fluent English speakers using Carbon. I find myself having trouble, on a re-reading, understanding whether "Parameterized impls" means "Parameterized `impl`s" or "Parameterized `impls`" (for example, what is the context in the table of contents that lets you disambiguate? Should I just know which it means based on experience?). I worry people who are both less familiar with English and less familiar with Carbon will find this to be a pretty steep learning curve, and an ongoing cost to require more context.

[...]

PR #2483, opened 2022-12-21
Replace keyword `is` with `impls`

josh11b:
Use the keyword `impls` instead of `is` when writing a `where` constraint that a type variable needs to implement an interface or named constraint.

What was previously (provisionally) written:
```
fn Sort[T:! Container where .ElementType is Ordered](x: T*);
```
will now be written:
```
fn Sort[T:! Container where .ElementType impls Ordered](x: T*);
```

PR #2173, opened 2022-09-13
Associated constant assignment versus equality

zygoloid:
Split the `where A == B` constraint in two: `where .A = B` produces a new
constraint from an old one, where the value of `.A` in the new constraint is
known and eagerly rewritten to `B`, and `where A == B`, which does not cause
`A` and `B` to be considered as identical by language rules but does permit
implicit (no-op) conversion between them.
 
This aims to provide an efficiently-computable and human-understandable type
equality rule, with type canonicalization and therefore transitive type
equality, without sacrificing too much in the way of ergonomics and without
sacrificing determinism, while still providing the full power of a general type
constraint system in a less ergonomic form.

PR #3253, opened 2023-09-22
Updates to generics design details, part 2

josh11b:
Continued from part 1: #3231. Second step updating `docs/design/generics/details.md`. There remains some work to incorporate proposal #2200.

- The biggest changes are incorporating much of the text of proposals:
  - #2173
  - #2687
- It incorporates changes from proposals:
  - #989
  - #1178
  - #2138
  - #2200
  - #2360
  - #2964
  - #3162
- It also updates the text to reflect the latest thinking from leads issues:
  - #996
  - #2153 -- most notably deleting the section on `TypeId`.
- Update to rule for prioritization blocks with mixed type structures from [discussion on 2023-07-18](https://docs.google.com/document/d/1gnJBTfY81fZYvI_QXjwKk1uQHYBNHGqRLI2BS_cYYNQ/edit?resourcekey=0-ql1Q1WvTcDvhycf8LbA9DQ#heading=h.7jxges9ojgy3)
- Adds reference links to proposals, issues, and discussions relevant to the text.
- Also tries to use more precise language when talking about implementations, to avoid confusing `impl` declaration and definitions with the `impls` operator used in `where` clauses, an issue brought up in
  - #2495 
  - #2483


docs/design/generics/details.md
```diff
@@ -1193,13 +1204,12 @@ Some interfaces will depend on other interfaces being implemented for the same
 type. For example, in C++,
 [the `Container` concept](https://en.cppreference.com/w/cpp/named_req/Container#Other_requirements)
 requires all containers to also satisfy the requirements of
-`DefaultConstructible`, `CopyConstructible`, `EqualityComparable`, and
-`Swappable`. This is already a capability for
-[facet types in general](#facet-types). For consistency we will use the same
-semantics and `require Self impls` syntax as we do for
+`DefaultConstructible`, `CopyConstructible`, `Eq`, and `Swappable`. This is
+already a capability for [facet types in general](#facet-types). For consistency
+we will use the same semantics and `require Self impls` syntax as we do for
```
zygoloid:
```suggestion
we use the same semantics and `require Self impls` syntax as we do for
```
I think this is describing the current state rather than a future state. (The "Some interfaces will depend" at the start of this section also doesn't seem to need the "will", but I think that's fine either way.)

docs/design/generics/details.md
```diff
@@ -1253,7 +1263,7 @@ When implementing an interface, we should allow implementing the aliased names
 as well. In the case of `Hashable` above, this includes all the members of
```
zygoloid: (Line above this one, can't add a comment there:) the "should" seems out of place here.

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
-that is being constrained has been given a name and use that name to express the
+described in the ["kinds of `where` constraints"](#kinds-of-where-constraints)
+section, but generally look like boolean expressions that should evaluate to
+`true`.
+
+The result of applying a `where` operator to a facet type is another facet type.
+Note that this expands the kinds of requirements that facet types can have from
+just interface requirements to also include the various kinds of constraints
+discussed later in this section. In addition, it can introduce relationships
+between different type variables, such as that a member of one is equal to the
```
zygoloid:
```suggestion
between different type variables, such as that a member of one is equal to a
```

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
+`C`.
 
-```
-fn PrintPoint2D[PointT:! NSpacePoint where .N = 2](p: PointT) {
-  Print(p.Get(0), ", ", p.Get(1));
-}
-```
+**References:** The definition of rewrite and same-type constraints were
+[proposal #2173](https://github.com/carbon-language/carbon-lang/pull/2173).
+Implements constraints switched using the `impls` keyword in
```
zygoloid:
```suggestion
Implements constraints switched to using the `impls` keyword in
```

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
+An implements constraint is written `where T impls C`, where `T` is a facet and
+`C` is a facet type. The constraint is that `T` satisfies the requirements of
+`C`.
 
-```
-fn PrintPoint2D[PointT:! NSpacePoint where .N = 2](p: PointT) {
-  Print(p.Get(0), ", ", p.Get(1));
-}
-```
+**References:** The definition of rewrite and same-type constraints were
```
zygoloid:
```suggestion
**References:** The definition of rewrite and same-type constraints were in
```
maybe?

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
 }
 ```
 
-This syntax is also used to specify the values of
-[associated constants](#associated-constants) when implementing an interface for
-a type.
+For types representing subsets of the real numbers, such as `i32` or `f32`, the
+`MagnitudeType` will match `Self`, the type implementing an interface. For types
+representing complex numbers, the types will be different. For example, the
+`Abs()` applied to a `Complex64` value would produce a `f32` result. The goal is
```
zygoloid: "the `Abs()` applied to" reads a bit strangely to me. How about:
```suggestion
representing complex numbers, the types will be different. For example,
`Abs()` applied to a `Complex64` value would produce a `f32` result. The goal is
```
or
```suggestion
representing complex numbers, the types will be different. For example, the
`Abs()` function applied to a `Complex64` value would produce a `f32` result. The goal is
```

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
-fn SumIntStack[T:! Stack where .ElementType = i32](s: T*) -> i32 {
-  var sum: i32 = 0;
-  while (!s->IsEmpty()) {
-    // s->Pop() has type `T.ElementType` == i32:
-    sum += s->Pop();
-  }
-  return sum;
+  // `Self` means the type implementing `Container`.
+  fn GetSlice[addr self: Self*]
+      (start: IteratorType, end: IteratorType) -> SliceType;
```
zygoloid: `IteratorType` is used here without being declared. GitHub doesn't let me make a suggested edit here, but maybe add
```
let IteratorType:! Iterator
    where .ElementType = ElementType;
```

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
 
-##### Set an associated type to a specific value
+```carbon
+interface Container {
+  let ElementType:! type;
 
-Functions accepting a generic type might also want to constrain one of its
-associated types to be a specific, concrete type. For example, we might want to
-have a function only accept stacks containing integers:
+  let SliceType:! Container
```
zygoloid: Did we decide to allow this? I remember at some point we weren't going to allow use of incomplete types like `Container` from within their own definitions, or at least we wouldn't allow name lookup into them (eg, the `.ElementType =` and `.SliceType =` below).

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
 
-```
-let IntStack:! auto = Stack where .ElementType = i32;
-constraint IntStack {
-  extend Stack where .ElementType = i32;
+```carbon
+constraint RealAbs {
+  extend HasAbs where .MagnitudeType = Self;
+  // Equivalent to:
+  extend HasAbs where .MagnitudeType = .Self;
```
zygoloid, line 2625: Do `Self` and `.Self` here have the same type? I think I'd expect `Self:! RealAbs` and `.Self:! HasAbs`.

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
 }
 ```
 
-This syntax is also used to specify the values of
-[associated constants](#associated-constants) when implementing an interface for
-a type.
+The `.Self` construct follows these rules:
 
-##### Equal generic types
+-   `X :!` introduces `.Self:! type`, where references to `.Self` are resolved.
```
zygoloid:
```suggestion
-   `X :!` introduces `.Self:! type`, where references to `.Self` are resolved
```

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
+    to `X`. This allows you to use `.Self` as an interface parameter as in
+    `X:! I(.Self)`.
+-   `A where` introduces `.Self:! A` and `.Foo` _designator_ for each member
```
zygoloid:
```suggestion
-   `A where` introduces `.Self:! A` and a `.Foo` _designator_ for each member
```

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
-`ElementType` of a `Container` interface as follows:
+So in `X:! A where ...`, `.Self` is introduced twice, after the `:!` and the
+`where`. This is allowed since both times it means `X`. After the `:!`, `.Self`
+has the type `type`, which gets refined to `A` after the `where`. In contrast,
+it is an error if `.Self` could mean two different things, as in:
 
+```carbon
+// ❌ Illegal: `.Self` could mean `T` or `T.A`.
+fn F[T:! InterfaceA where .A impls
+           (InterfaceB where .B = .Self)](x: T);
```
zygoloid, line 2664: Is it interesting to say how to write the two meanings? I think they're:
```
constraint InterfaceBWithSelf {
  extend InterfaceB where .B = Self;
}
constraint InterfaceBWith(T:! InterfaceA) {
  extend InterfaceB where .B = T;
}
fn F[T:! InterfaceA where .A impls InterfaceBWithSelf](x: T);
fn F[T:! InterfaceA where .A impls InterfaceBWith(.Self)](x: T);
```

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
 ```
 
-we can constrain them to be equal in a function signature:
+##### Combining constraints with `extends`
 
-```
-fn F[MatchedPairType:! PairInterface where .Left == .Right]
-    (x: MatchedPairType*);
-```
+Within an interface or named constraint, `extends` can be used to extend a
```
zygoloid: Is this still `extends` or is it now `extend`?

zygoloid (commented):
(Checkpoint of incomplete review)

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
+  impl as C;
 }
-```
 
-This kind of constraint can be named:
+var n: i32;
 
+// ❌ No implicit conversion from `i32` to `T.(A.T)`.
+// Resolved constraint on `T` is
+// `B where T.(A.T) == T.(A.U) and T.(A.U) = i32`.
```
zygoloid:
```suggestion
// `B where T.(A.T) == T.(A.U) and T.(A.U) == i32`.
```

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
+
+A purely syntactic check is used to determine if an `=` is specified directly in
+an expression:
+
+-   An `=` constraint is specified directly in its enclosing `where` expression.
+-   If an `=` constraint is specified directly in an operand of an `&` or
+    `(`...`)`, then it is specified directly in that enclosing expression.
+
+In an `impl as C` or `impl T as C` declaration in an interface or named
+constraint, `C` is not allowed to directly specify any `=` constraints:
```
zygoloid: This seems to repeat what we said at the start of this section (other than also mentioning `impl as C`).

Perhaps we could adjust this a little, so we start by describing just that `=` constraints get turned into `==` constraints, and then down here say that `=` constraints can't be specified directly (so that it'll be clear that's just a guard rail for the otherwise-surprising behavior).

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
-  let P:! PairInterface where .Left == .Right;
+
+##### Combining constraints with `impl as` and `impls`
+
+Within an interface or named constraint, the `impl T as C` syntax does not
+permit `=` constraints to be specified directly. However, such constraints can
+be specified indirectly, for example if `C` is a named constraint that contains
+rewrites. Because these constraints don't change the type of `T`, rewrite
+constraints in this context will never result in a rewrite being performed, and
+instead are equivalent to `==` constraints.
```
zygoloid: Suggestion to implement a later comment:
```suggestion
Within an interface or named constraint, the `impl T as C` and `impl as C` syntaxes do not change the type of `T` or `Self`, respectively, so any `=` constraints that they specify do not result in rewrites being performed when the type `T` or `Self` is later used. Such `=` constraints are equivalent to `==` constraints:
```

docs/design/generics/details.md
zygoloid: Continuation of suggestion to implement prior comment:
```suggestion
Because `=` constraints are effectively treated as `==` constraints in
an `impl as C` or `impl T as C` declaration in an interface or named
constraint, it is an error to specify such a `=` constraint directly in `C`.
A purely syntactic check is used to determine if an `=` is specified directly in
an expression:

-   An `=` constraint is specified directly in its enclosing `where` expression.
-   If an `=` constraint is specified directly in an operand of an `&` or
    `(`...`)`, then it is specified directly in that enclosing expression.

For example:
```

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
-    (haystack: SC, needles: CT) -> bool;
-```
+When a facet type is used as the declared type of a facet `T`, the constraints
+that were specified within that facet type are _resolved_ to determine the
+constraints that apply to `T`. This happens:
 
-#### Type bound for associated type
+-   When the constraint is used explicitly, when declaring symbolic binding,
+    like a generic parameter or associated constant, of the form
+    `T:! Constraint`.
```
zygoloid:
```suggestion
-   When the constraint is used explicitly when declaring a symbolic binding,
    like a generic parameter or associated constant, of the form
    `T:! Constraint`.
```

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
-between two type variables does not modify the set of member names of either
-type. (If you write `where .ElementType = String` with a `=` and a concrete
-type, then `.ElementType` is actually set to `String` including the complete
-`String` API.)
+##### Rewrite constraint resolution
 
-Note that `==` constraints are symmetric, so the previous declaration of
-`Contains` is equivalent to an alternative declaration where `CT` is declared
-first and the `where` clause is attached to `SortedContainer`:
+**FIXME: Should the precise rules for constraints be moved into an appendix?**
```
zygoloid: I think that's a good idea -- maybe it'd be  good to move all the rules from here down to `Same type constraints` into an appendix? And generally I think we should split this file up into a bunch of pieces.

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
 
+interface Commute {
+  let X:! Commute;
+  let Y:! Helper(X);
+
+  fn GetX[self: Self]() -> X;
+  fn GetY[self: Self]() -> Y;
+  // **FIXME: Don't think it is legal to write `X.X.Y` here.**
+  fn TakesXXY[self: Self](xxy: X.X.Y);
+}
```
zygoloid: We should decide what to do about the FIXMEs here, I guess. Maybe this:
```
interface CommuteBase {
  let X:! CommuteBase;
  let Y:! CommuteBase;
}
constraint CommuteWithX(X:! CommuteBase) {
  extend CommuteBase where .X == X;
}
interface Commute {
  extend CommuteBase where .X impls Commute and .Y impls (Commute & CommuteWithX(.X.Y));
  fn GetX[self: Self]() -> X;
  fn GetY[self: Self]() -> Y;
  fn TakesXXY[self: Self](xxy: X.X.Y);
}
```
That gets a bit further in explorer: https://carbon.godbolt.org/z/3xYhjr58r (with a bit more massaging for unqualified lookup not looking through `extend`) but still isn't accepted.

The above isn't really right, though -- `c.GetX().GetX()` below won't work because `c.GetX()` is constrained to be `CommuteBase`, not `Commute`. I think this *should* be valid under our current completeness rules, but goes a bit further beyond what's implemented in explorer:
```
interface Commute;
interface CommuteBase {
  let X:! Commute;
  let Y:! Commute;
}
constraint CommuteWithX(X:! Commute) {
  extend CommuteBase where .X == X;
}
interface Commute {
  extend CommuteBase where .Y impls CommuteWithX(.X.Y);
  fn GetX[self: Self]() -> X;
  fn GetY[self: Self]() -> Y;
  fn TakesXXY[self: Self](xxy: X.X.Y);
}
```

docs/design/generics/details.md
josh11b: I believe this design documents that this declaration is allowed, but not name lookup. See [Forward declarations and cyclic references](#forward-declarations-and-cyclic-references)

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
+same-type `where` constraints. These `observe` declarations may be included in
+an `interface` definition or a function body, as in:
+
+```carbon
+// **FIXME: Not clear how to fix this example.**
+interface Commute {
+  let X:! Commute;
+  let Y:! Commute where .X == X.Y;
+  ...
+  observe X.X.Y == X.Y.X == Y.X.X;
```
zygoloid: Does this work:
```suggestion
interface Commute;
interface CommuteBase {
  let X:! Commute;
  let Y:! Commute;
}
interface Commute {
  alias X = CommuteBase.X;
  alias Y = CommuteBase.Y;
  extend CommuteBase where .Y.(X) == .X.(Y);
  ...
  observe X.(X).(Y) == X.(Y).(X) == Y.(X).(X);
```
The idea is to use a compound member access to avoid performing a lookup into `Commute` while it's incomplete.

Though this is making me think we should revisit our decision that type completeness is all-or-nothing, and allow lookups into types that are in the middle of being defined, and reject if a member is named that was not declared at the point of the member access (eg, if we see a name being introduced and we've already done a lookup for that name, reject).

docs/design/generics/details.md
josh11b: Removed "will" as well.

docs/design/generics/details.md
josh11b: Done.

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
+  (a as T.C).InR();
+}
 ```
 
-must be replaced by:
+Only the current type is searched for interface implementations, so the call to
+`InR()` would be illegal without the cast. However, an
+[`observe`...`==`...`impls` declaration](#observing-equal-to-a-type-implementing-an-interface)
+can be used to identify interfaces that must be implemented through some equal
+type. This does not [extending](terminology.md#extending-an-impl) the API of the
```
zygoloid:
```suggestion
type. This does not [extend](terminology.md#extending-an-impl) the API of the
```

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
+    [ContainerType:! Container where .ElementType impls Ordered]
+    (container_to_sort: ContainerType*);
 ```
-fn LookUp[KeyType:! type](hm: HashMap(KeyType, i32)*,
-                          k: KeyType) -> i32;
 
-fn PrintValueOrDefault[KeyType:! Printable,
+In contrast to a [rewrite constraint](#rewrite-constraints) or a
+[same-type constraint](#same-type-constraints), this does not say what type
+`ElementType` exactly is, just that it must satisfy requirements of some facet
```
zygoloid:
```suggestion
`ElementType` exactly is, just that it must satisfy the requirements of some facet
```

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
+type.
+
+> **Note:** `Container` defines `ElementType` as having type `type`, but
+> `ContainerType.ElementType` has type `Ordered`. This is because
+> `ContainerType` has type `Container where .ElementType impls Ordered`, not
+> `Container`. This means we need to be a bit careful when talking about the
+> type of `ContainerType` when there is a `where` clause modifying it.
```
zygoloid: I found this surprising; I hadn't been expecting that `impls` would change the type. I don't think that can always happen; a `where` clause shouldn't change the type of some distant value:

```
[T:! type, U:! type where T impls Ordered](x: T, y: U) {
  // Shouldn't be valid: `where` in type of `U` shouldn't change the type of `T`.
  x.Compare(x);
}
```

But I think we can make this happen in the case where we allow an `=` constraint -- that is, when the left-hand side of the `impls` is `.AssociatedType`. Is that what we want to do?

If not, I think we could get the desired effect by allowing deduction to deduce through rewrites in associated constants:

```
// With this declaration, `ContainerType.ElementType` is of type `type`
fn SortContainer
    [ContainerType:! Container where .ElementType impls Ordered]
    (container_to_sort: ContainerType*);

// With this declaration, `ContainerType.ElementType` is rewritten to `Elem`,
// which is of type `Ordered`.
// A call to `SortContainer` deduces `ContainerType` from the argument,
// then deduces `Elem` as `(ContainerType as Container).ElementType`.
fn SortContainer
    [Elem:! Ordered, ContainerType:! Container where .ElementType = Elem]
    (container_to_sort: ContainerType*);
```

docs/design/generics/details.md
josh11b: That is an interesting point, not currently addressed by the text. I've added some text to make it match your expectations.

docs/design/generics/details.md
josh11b: Added.

docs/design/generics/details.md
```diff
@@ -3051,306 +3792,407 @@ and
...
+In `Contains_Rewrite`, `CT.ElementType` is rewritten to `SC.ElementType` and
+uses the facet type of `SC.ElementType`.
 
+In `Contains_SameType`, the `where` clause does not affect the API of
+`CT.ElementType`, and it would not even be considered to implement `Ordered`
+unless there is some declaration like
+`observe CT.ElementType == SC.ElementType impls Ordered`. Even then, the items
+from the `needles` container won't directly have a `Compare` method member.
+
+The rule is that an same-type `where` constraint between two type variables does
```
zygoloid:
```suggestion
The rule is that a same-type `where` constraint between two type variables does
```

docs/design/generics/details.md
```diff
@@ -3051,306 +3792,407 @@ and
...
 
--   manual work for the source code author to prove to the compiler that types
-    are equal; and
--   verbosity.
+// Constraint: `T` such that `Vector(T)` implements `Printable`
+fn PrintThree
+    [T:! type where Vector(.Self) impls Printable]
+    (a: T, b: T, c: T) {
+  var v: Vector(T) = (a, b, c);
+  Print(v);
```
zygoloid: This example seems incomplete: `Print` and print support for tuples is not described, though it's still pretty clear what this means. I wonder if it'd be useful to add
```
interface Printable { fn Print[self: Self](); }
impl forall [T:! Printable, U:! Printable, V:! Printable] (T, U, V) as Printable { ... }
```
and use `v.(Printable.Print)()` here?

docs/design/generics/details.md
```diff
@@ -2463,549 +2473,1280 @@ interface A(T:! B where ...) {
...
+#### Same-type constraints
+
+A same-type constraint describes that two type expressions are known to evaluate
+to the same value. Unlike a [rewrite constraint](#rewrite-constraints), however,
+the two type expressions are treated as distinct types when type-checking a
+symbolic expression that refers to them.
+
+Same-type constraints are brought into scope, looked up, and resolved exactly as
+if there were a `SameAs(U:! type)` interface and a `T == U` impl corresponded to
+`T is SameAs(U)`, except that `==` is commutative. As such, it's not possible to
```
zygoloid: We also wanted to allow equalities to apply to type components, so that `X(A, B, C) is SameType(X(D, E, F))` if we know that `A == D and B == E and C == F`. Eg, if we know that `T == i32` then we should also have `Vector(T)` be single-step equal to `Vector(i32)`.

See [#2173, "Moreover ..."](https://github.com/carbon-language/carbon-lang/blob/trunk/proposals/p2173.md#proposal:~:text=Moreover%2C%20if%20F%20is%20any%20pure%20type%20function%2C%20T%20is%20SameAs(U)%20implies%20F(T)%20is%20SameAs(F(U)).), which on reflection maybe doesn't quite say enough (I think it only permits us to use a single equality, so `A == B` and `C == D` would give `(A, i32) == (B, i32)` but not `(A, C) == (B, D)`).

docs/design/generics/details.md
josh11b: Fixed. Looks like there was some other old syntax used, like `impl as` instead of `require Self impls`.

docs/design/generics/details.md
```diff
@@ -3668,37 +4528,52 @@ fn F(...) {
...
+<!-- prettier-ignore-end -->
+
+This may also be called a _generic `impl` declaration_.
+
 ### Impl for a parameterized type
 
-Interfaces may be implemented for a parameterized type. This can be done
-lexically in the class' scope:
+Interfaces may be implemented for a [parameterized type](#parameterized-types).
+This can be done lexically in the class' scope:
```
zygoloid:
```suggestion
This can be done lexically in the class's scope:
```

docs/design/generics/details.md
```diff
@@ -4991,7 +5976,11 @@ interface TotalOrder {
 The workaround for this restriction is to use a
 [blanket impl declaration](#blanket-impl-declarations) instead:
 
-```
+**FIXME: Is it sensible to have both a `require` and a blanket implementation?
+Does the blanket implementation satisfy the requirement so you only need to
+implement `TotalOrder`?**
```
zygoloid: I think the `require` here has no effect, because the constraint is always satisfied by the blanket impl.

docs/design/generics/details.md
```diff
@@ -5110,8 +6099,9 @@ says that if `Self` implements `CommonTypeWith(T)`, then `T` must implement
 `CommonTypeWith(Self)`.
 
 An `require`...`impls` constraint in an `interface`, or `constraint`, definition
```
zygoloid:
```suggestion
A `require`...`impls` constraint in an `interface`, or `constraint`, definition
```

docs/design/generics/details.md
josh11b: Updated the text.

docs/design/generics/details.md
josh11b: Inserted a paragraph.

docs/design/generics/details.md
josh11b: It isn't that tuples are printable, it is that you can initialize a `Vector` with a tuple. I've added the `Printable` interface and clarified the code.

docs/design/generics/details.md
josh11b: That was what was written before https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/generics/details.md#type-bounds-on-associated-types-in-declarations
and is indeed something I think we still want since it is going to be important for usability. I'm happy to limit it to the single designator case that we also support for rewrite constraints.

docs/design/generics/details.md
josh11b: I tried switching to the Node/Edge example from the cyclic reference section, as discussed offline. You can see it in https://github.com/carbon-language/carbon-lang/pull/3253/commits/9a6a483c11d1de5c493a1bf1a6a39fce396d2a69 . Unfortunately, I dont' think it works for the `observe` examples in this section.

docs/design/generics/details.md
josh11b: Made that change, after replacing `impl T as` with `require T impls`

docs/design/generics/details.md
josh11b: Applied with `require`...`impls` instead of `impl`...`as`.

docs/design/generics/details.md
josh11b: Done.

docs/design/generics/details.md
josh11b: I've switched to the examples away from `Commute` to either `Node`/`Edge` or `Graph`.

docs/design/generics/details.md
```diff
@@ -3446,55 +2880,64 @@ fn F[T:! Transitive](t: T) {
...
-  // **FIXME: Don't think it is legal to write `X.X.Y` here.**
-  fn TakesXXY[self: Self](xxy: X.X.Y);
+interface Edge {
+  let N:! NodeFor(Self);
+  fn GetN[self: Self]() -> N;
+}
+interface Node {
+  let E:! EdgeFor(Self);
+  fn GetE[self: Self]() -> EdgeT;
+  fn AddE[addr self: Self*](e: EdgeT);
```
zygoloid:
```suggestion
  fn GetE[self: Self]() -> E;
  fn AddE[addr self: Self*](e: E);
```

docs/design/generics/details.md
```diff
@@ -3446,55 +2880,64 @@ fn F[T:! Transitive](t: T) {
...
+  fn NearN[self: Self](n: N) -> bool;
```
zygoloid:
```suggestion
  fn NearN[self: Self](n: Self) -> bool;
```

docs/design/generics/details.md
```diff
@@ -3567,19 +3015,15 @@ expression in the sequence by a single `where` equality constraint. In this
...
+  observe G.E.N == G.N.E.N == G.N == G.E.N.E.N;
+  // ...
 }
 ```
 
-the expression `X.Y.Y.X` is one equality away from `X.Y.X.Y` and so it is
-allowed. This is even though `X.Y.X.Y` isn't the type expression immediately
-prior to `X.Y.Y.X`.
+the expression `G.E.N.E.N` is one equality away from `G.N.E.N` and so it is
+allowed. This is even though `G.N.E.N` isn't the type expression immediately
```
zygoloid:
```suggestion
allowed. This is true even though `G.N.E.N` isn't the type expression immediately
```

docs/design/generics/details.md
```diff
@@ -3688,14 +3132,21 @@ fn SortContainer
...
+direct associated facet member of the facet type being constrained (similar to
+the restriction on [rewrite constraints](#rewrite-constraints)), gets special
+treatment. In this case, the type of the associated facet is
+[combined](#combining-interfaces-by-anding-facet-types) with the constraint. In
+the above example, `Container` defines `ElementType` as having type `type`, but
+`ContainerType.ElementType` has type `type & Ordered` (which is equivalent to
+`Ordered`). This is because `ContainerType` has type
+`Container where .ElementType impls Ordered`, not `Container`. This means we
+need to be a bit careful when talking about the type of `ContainerType` when
+there is a `where` clause modifying it.
```
zygoloid, line 3155: Maybe we should discuss further when @chandlerc gets back from vacation. It seems a little inconsistent to me to say that we need different syntax (`=` vs `==`) for type-changing equality constraint, but for `impls` we will just look at whether the left-hand side is written as a designator to determine whether it affects the type of the member. I think it's fine for now, though, and the new text seems like a more precise form of the old note, so I don't think we should block this PR on it.

docs/design/generics/details.md
zygoloid: This is doing name lookup on the next line: `where .ElementType =` is looking up `ElementType` in the incomplete interface `Container` (after the declaration of `ElementType`), and `.SliceType =` is looking up the member that we're in the middle of declaring. I think this example needs to be reworked a bit.

zygoloid (approved):
Thanks, I think we're basically converged here. I'm happy to take another look if you like after the `Container` example is fixed, but I don't feel it's necessary.

docs/design/generics/details.md
josh11b: Added "future work"

docs/design/generics/details.md
josh11b: I rewrote the example to make it legal. I expect there is a better example of the point, but I haven't thought of it yet.

docs/design/generics/details.md
```diff
@@ -2592,19 +2592,25 @@ this interface while `.Self` refers to the associated facet currently being
...
 interface Container {
   let ElementType:! type;
   let IteratorType:! Iterator where .ElementType = ElementType;
 
-  let SliceType:! Container
-      where .ElementType = ElementType and
-            // `.Self` means `SliceType`.
-            .SliceType = .Self;
+  // `.Self` means `SliceType`.
+  let SliceType:! SliceConstraint(ElementType, .Self);
```
zygoloid: Hm, how do we type-check this call to `SliceConstraint`? We don't know that `.Self impls Container` here. And
```
  let SliceType:! Container & SliceConstraint(ElementType, .Self);
```
doesn't help, because while type-checking the type expression we still have `.Self:! type`. This would work:
```suggestion
  let SliceType:! Container where .Self is SliceConstraint(ElementType, .Self);
```
and I don't think that's a bad workaround. In fact, I think you can drop the `.Self` parameter in this case and define `SliceConstraint` as 
```
constraint SliceConstraint(E:! type) {
  // Note, might need to use `.Self:! Container`, not `Self:! type` for the better type,
  // even though they would have the same value here.
  extend Container where .ElementType = E and
                         .SliceType = .Self;
}
```

Should this be a context where we have implied constraints? That is, should we pick up an implicit `.Self impls Container` constraint from the use of `.Self` as the argument to `S:! Container` here? I think that would make sense, but currently we only get implied constraints on a generic function. (I'm not sure whether we should only allow implied constraints on `.Self` or on other associated constants for the interface too.)

Example:

```carbon
interface Hashable {
  let KeyType:! type;
  // or `let ValueType:! type; require HashMap(KeyType, ValueType) impls type;`
  let ValueType:! type where HashMap(KeyType, .Self) impls type;
}
```
would be a useful shorthand for
```carbon
interface Hashable {
  let KeyType:! type;
  let ValueType:! type;
  // (whatever constraints are implied by `HashMap`)
  require KeyType impls HashMapKey;
  require ValueType impls HashMapValue;
}
```

josh11b: >  In fact, I think you can drop the `.Self` parameter in this case and define `SliceConstraint` as
> 
> ```
> constraint SliceConstraint(E:! type) {
>   // Note, might need to use `.Self:! Container`, not `Self:! type` for the better type,
>   // even though they would have the same value here.
>   extend Container where .ElementType = E and
>                          .SliceType = .Self;
> }
> ```

I realized I could have included the `.Self` inside `SliceConstraint`, but I was trying to make an example of an interface using `.Self` and `Self` to highlight how they are different.

zygoloid: Filed #3256 for the question about implied constraints.

PR #3255, opened 2023-09-29
Unify initialization and conversion logic

zygoloid:
Combine the initialization, implicit conversion, and value category conversion functions into a single function.

This substantially reduces the duplication between these steps, and ensures that we support the same set of conversions in all these contexts. This also fixes some issues where we would not use the proper value representation for tuples and structs after performing implicit conversions.


toolchain/sem_ir/file.h
```diff
@@ -305,16 +312,23 @@ class File : public Printable<File> {
...
-    T* storage = static_cast<T*>(
-        allocator_.Allocate(data.size() * sizeof(T), alignof(T)));
-    std::uninitialized_copy(data.begin(), data.end(), storage);
-    return llvm::MutableArrayRef<T>(storage, data.size());
+    T* storage =
+        static_cast<T*>(allocator_.Allocate(size * sizeof(T), alignof(T)));
+    return llvm::MutableArrayRef<T>(storage, size);
+  }
+
+  // Allocates a copy of the given data using our slab allocator.
```
chandlerc, line 326:
```suggestion
  // Allocates a copy of the given data using our allocator.
```

(or add "slab" back to the comment above)

toolchain/check/context.h
```diff
@@ -24,6 +24,40 @@ class Context {
   // A block of code that has not yet been inserted into SemIR.
   class PendingBlock;
 
+  // Description of the target of a conversion.
+  struct ConversionTarget {
```
chandlerc: I think at this point, it's important to factor `Convert` and all of this out into a separate class with its own header and source file.

Are there barriers to that? FWIW, if reasonable, I would prefer to do that here before reviewing the full implementation as I think it'll make it easier for me to review as well.

One key thing that I think this will clarify is to introduce a layering between the conversion and the rest of checking, and to understand how that layering works. It may even require some callbacks or something to handle complex aspects of the layering, but that would actually make the code easier to read for me because it would highlight where things can re-enter if it makes sense.

toolchain/check/context.cpp
```diff
@@ -362,200 +371,552 @@ class Context::PendingBlock {
...
+  //    definition of `ImplicitAs` itself or implementations of it.
+  //
+  // We also expect to see better performance by avoiding an `impl` lookup for
+  // common conversions.
+  //
+  // TODO: We should provide a debugging flag to turn off as many of these
+  // builtin conversions as we can so that we can test that they do the same
+  // thing as the library implementations.
+  //
+  // These builtin conversions all correspond to `final impl`s in the library,
```
chandlerc: This sentence seems to contradict some of the above comments about not being able to implement some of these in the library?

chandlerc (commented):
Some tiny nits I saw in glancing, and one really big comment that I'd like to chat about before I work on more of the review.

toolchain/check/context.h
zygoloid: Done.

toolchain/sem_ir/file.h
zygoloid: Done.

toolchain/check/context.cpp
zygoloid: Hm, I see your point. It's not that these conversions won't exist in the library -- they (eventually) will do -- it's that calling `ImplicitAs` won't do the right thing because it doesn't support taking (eg) a tuple literal with mixed expression categories for its elements.

I've updated the text above to try to not suggest these conversions won't exist in the library.

toolchain/check/initializing_expression.cpp
```diff
@@ -0,0 +1,109 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "toolchain/check/initializing_expression.h"
+
+#include "toolchain/sem_ir/file.h"
+
+namespace Carbon::Check {
+
+// Materialize a temporary to hold the result of the given expression if it is
+// an initializing expression.
```
chandlerc: My preference is to not duplicate the comments, even though it makes it a little bit farther from the definition.

```suggestion
```

toolchain/check/convert.h
```diff
@@ -0,0 +1,93 @@
...
+// Convert a value to another type and expression category.
+auto Convert(Context& context, Parse::Node parse_node, SemIR::NodeId value_id,
+             ConversionTarget target) -> SemIR::NodeId;
+
+// Performs initialization of `target_id` from `value_id`. Returns the
+// possibly-converted initialization expression, which should be assigned to
+// the target using a suitable node for the kind of initialization.
+auto Initialize(Context& context, Parse::Node parse_node,
+                SemIR::NodeId target_id, SemIR::NodeId value_id)
+    -> SemIR::NodeId;
```
chandlerc: WDYT about this living in `initializing_expression.h`? (Potentially renaming it to `initialize.h` if needed, but given that this returns the initializing expression, the current name seems fine.

Or if that would create cycles, maybe there should just be one header/source for all of this called `initialize.h` or the current `convert.h`?

Also, minor tweak I think:
```suggestion
// Performs initialization of `target_id` from `value_id`. Returns the
// possibly-converted initializing expression, which should be assigned to
// the target using a suitable node for the kind of initialization.
auto Initialize(Context& context, Parse::Node parse_node,
                SemIR::NodeId target_id, SemIR::NodeId value_id)
    -> SemIR::NodeId;
```

toolchain/check/convert.cpp
```diff
@@ -0,0 +1,729 @@
...
+#include "toolchain/check/context.h"
+#include "toolchain/check/declaration_name_stack.h"
+#include "toolchain/check/initializing_expression.h"
+#include "toolchain/check/node_block_stack.h"
+#include "toolchain/diagnostics/diagnostic_kind.h"
+#include "toolchain/lex/tokenized_buffer.h"
+#include "toolchain/parse/node_kind.h"
+#include "toolchain/sem_ir/file.h"
+#include "toolchain/sem_ir/node.h"
+#include "toolchain/sem_ir/node_kind.h"
```
zygoloid, line 17: TODO: Minimize #include lists in convert.cpp and context.cpp

toolchain/check/convert.cpp
```diff
@@ -0,0 +1,729 @@
...
+    }
+  }
+
+  // No builtin conversion applies.
+  return value_id;
+}
+
+auto Convert(Context& context, Parse::Node parse_node, SemIR::NodeId expr_id,
+             ConversionTarget target) -> SemIR::NodeId {
+  auto orig_expr_id = expr_id;
```
chandlerc: Consider:

```suggestion
             ConversionTarget target) -> SemIR::NodeId {
  auto semir = conetxt.semantic_ir();
  auto orig_expr_id = expr_id;
```

And using `semir` (or whatever name you prefer) below to make the code a bit shorter?

chandlerc: Also happy to use this pattern for any of the functions where there are more than 3-ish calls.

toolchain/check/convert.cpp
```diff
@@ -0,0 +1,729 @@
...
+  // We also expect to see better performance by avoiding an `impl` lookup for
+  // common conversions.
+  //
+  // TODO: We should provide a debugging flag to turn off as many of these
+  // builtin conversions as we can so that we can test that they do the same
+  // thing as the library implementations.
+  //
+  // These builtin conversions all correspond to `final impl`s in the library,
+  // so we don't need to worry about `ImplicitAs` being specialized in any of
+  // these cases.
```
chandlerc: If I understand correctly, this is what you're trying to say? Hoping to further remove confusion with the prior section of comments:

```suggestion
  // Even when these could be provided in Carbon code in the prelude and we do so
  // here for performance reasons, the prelude form would use `final impl`s. This
  // means we don't need to worry about `ImplicitAs` being specialized in any of
  // these cases.
```

toolchain/check/convert.cpp
```diff
@@ -0,0 +1,729 @@
...
+    // TODO: This call recurses back into conversion. Switch to an iterative
+    // approach.
+    auto init_id = Convert(context, value.parse_node(), src_id,
+                           {.kind = inner_kind,
+                            .type_id = dest_type_id,
+                            .init_id = target_id,
+                            .init_block = target.init_block});
+    if (init_id == SemIR::NodeId::BuiltinError) {
+      return SemIR::NodeId::BuiltinError;
+    }
```
chandlerc: Maybe extract this to a function?

toolchain/check/convert.cpp
```diff
@@ -0,0 +1,729 @@
...
```
chandlerc: Consider extracting to a function?

toolchain/check/convert.cpp
```diff
@@ -0,0 +1,729 @@
...
+    // TODO: This call recurses back into conversion. Switch to an iterative
+    // approach.
+    auto init_id = Convert(context, value.parse_node(), src_id,
+                           {.kind = ConversionTarget::FullInitializer,
+                            .type_id = element_type_id,
+                            .init_id = target_id,
+                            .init_block = target_block});
+    if (init_id == SemIR::NodeId::BuiltinError) {
+      return SemIR::NodeId::BuiltinError;
+    }
```
chandlerc: Consider extracting to a function?

chandlerc (approved):
Really nice. =D

I think some more refactorings are available. I've suggested a few places where I saw something obvious, but there may be more. Feel free to take as many of these in this PR as make sense, and happy for others to be follow-ups, this is already a huge improvement.

Also some more comment tweaking suggested.

None of this really blocking, LGTM.

toolchain/check/convert.h
zygoloid: Applied the comment fix and folded `initializing_expression.cpp` into `convert.cpp`.

toolchain/check/convert.cpp

toolchain/check/convert.cpp

toolchain/check/convert.cpp

toolchain/check/convert.cpp

toolchain/check/convert.cpp
zygoloid: I don't think that's quite what I mean: there *will* be `final impl`s in the library for all of these cases. (`T impls ImplicitAs(U)` still needs to work, etc.) But we won't use them. I've taken another pass over this comment; I think it's a lot clearer now.

PR #3257, opened 2023-10-03
Support for `let` declarations.

zygoloid:
A `let` declaration is represented by a `bind_name` node in SemIR:

```carbon-semir
  %b: i32 = bind_name "b", %a
```

Because `Check` encounters the pattern before it sees the value, we first create the `bind_name` node with an unset value and don't add it to the block. Then, once we've seen and converted the initializer, we update the `bind_name` to have the value and add it to the current block, after the initializer code.


toolchain/check/testdata/let/convert.carbon
```diff
@@ -0,0 +1,48 @@
...
+// CHECK:STDOUT:   %.loc8_36.8: init (i32, i32, i32) = tuple_init %.loc8_36.1, (%.loc8_36.3, %.loc8_36.5, %.loc8_36.7)
+// CHECK:STDOUT:   assign %v, %.loc8_36.8
+// CHECK:STDOUT:   %.loc10: (type, type, type) = tuple_literal (i32, i32, i32)
+// CHECK:STDOUT:   %.loc8_7.1: ref i32 = tuple_access %v, member0
+// CHECK:STDOUT:   %.loc8_7.2: i32 = bind_value %.loc8_7.1
+// CHECK:STDOUT:   %.loc8_7.3: ref i32 = tuple_access %v, member1
+// CHECK:STDOUT:   %.loc8_7.4: i32 = bind_value %.loc8_7.3
+// CHECK:STDOUT:   %.loc8_7.5: ref i32 = tuple_access %v, member2
+// CHECK:STDOUT:   %.loc8_7.6: i32 = bind_value %.loc8_7.5
+// CHECK:STDOUT:   %.loc8_7.7: (i32, i32, i32) = tuple_value %v, (%.loc8_7.2, %.loc8_7.4, %.loc8_7.6)
```
josh11b: Labels here seem weird, shouldn't we be doing stuff for `loc10` not `loc8`?

toolchain/lower/testdata/let/tuple.carbon
```diff
@@ -0,0 +1,61 @@
...
+// CHECK:STDOUT:   %tuple.elem8 = getelementptr inbounds { i32, i32 }, ptr %b, i32 0, i32 0
+// CHECK:STDOUT:   %7 = load i32, ptr %tuple.elem8, align 4
+// CHECK:STDOUT:   %tuple.elem9 = getelementptr inbounds { i32, i32 }, ptr %b, i32 0, i32 1
+// CHECK:STDOUT:   %8 = load i32, ptr %tuple.elem9, align 4
+// CHECK:STDOUT:   %tuple10 = alloca { i32, i32 }, align 8
+// CHECK:STDOUT:   %9 = getelementptr inbounds { i32, i32 }, ptr %tuple10, i32 0, i32 0
+// CHECK:STDOUT:   store i32 %7, ptr %9, align 4
+// CHECK:STDOUT:   %10 = getelementptr inbounds { i32, i32 }, ptr %tuple10, i32 0, i32 1
+// CHECK:STDOUT:   store i32 %8, ptr %10, align 4
+// CHECK:STDOUT:   %tuple11 = alloca { { i32, i32, i32 }, { i32, i32 } }, align 8
```
josh11b, line 52: Independent of this change: That's a bug in the type it is allocating, should have created a struct containing two pointers.

toolchain/check/testdata/let/convert.carbon
zygoloid: Fixed by the `name_reference` change.

toolchain/lower/testdata/let/tuple.carbon
zygoloid: Acknowledged, will send a separate PR.

josh11b (approved):

PR #3258, opened 2023-10-03
Support for `while` statements.

zygoloid:
Basic support for type-checking and lowering of `while` statements, as well as `break` and `continue`.


toolchain/check/handle_loop_statement.cpp
```diff
@@ -40,17 +67,57 @@ auto HandleForStatement(Context& context, Parse::Node parse_node) -> bool {
...
+  // Branch to the loop entry block.
+  auto loop_entry_id = context.AddDominatedBlockAndBranch(parse_node);
+  context.node_block_stack().Pop();
+
+  // Start emitting the loop entry block.
+  context.node_block_stack().Push(loop_entry_id);
+  context.AddCurrentCodeBlockToFunction();
+
+  context.node_stack().Push(parse_node, loop_entry_id);
+  return true;
```
chandlerc: What do you think about calling this the loop `header` block? It at least seems accurate given the LLVM terminology here:
https://llvm.org/docs/LoopTerminology.html

Notably, `entering` and `entry` are very similar but importantly different here -- the block you're adding here is inside the loop (evaluating the loop condition). Calling it the loop `header` would at least for me make this more unambiguous. TBF, that's likely because of how familiar I am with the LLVM loop terminology. But not sure there is any better terminology to anchor on here.

toolchain/check/handle_loop_statement.cpp
```diff
@@ -40,17 +67,57 @@ auto HandleForStatement(Context& context, Parse::Node parse_node) -> bool {
...
+  auto loop_body_id =
+      context.AddDominatedBlockAndBranchIf(parse_node, cond_value_id);
+  auto loop_exit_id = context.AddDominatedBlockAndBranch(parse_node);
+  context.node_block_stack().Pop();
+
+  // Start emitting the loop body.
+  context.node_block_stack().Push(loop_body_id);
+  context.AddCurrentCodeBlockToFunction();
+  context.break_continue_stack().push_back(
+      {.break_target = loop_exit_id, .continue_target = loop_entry_id});
```
chandlerc: FWIW, `loop_body` might be reasonably called `loop_latch` if you like, but there isn't the confusion here for me so really whatever folks like.

toolchain/check/handle_loop_statement.cpp
```diff
@@ -40,17 +67,57 @@ auto HandleForStatement(Context& context, Parse::Node parse_node) -> bool {
...
 
-auto HandleWhileCondition(Context& context, Parse::Node parse_node) -> bool {
-  return context.TODO(parse_node, "HandleWhileCondition");
-}
-
 auto HandleWhileConditionStart(Context& context, Parse::Node parse_node)
     -> bool {
-  return context.TODO(parse_node, "HandleWhileConditionStart");
+  // Branch to the loop entry block.
+  auto loop_entry_id = context.AddDominatedBlockAndBranch(parse_node);
```
chandlerc: Would it be worth seeing if the current block is empty, and if so, simply re-using it?

You can see the effect of this in the test below where we end up with blocks that contain only a single instruction to branch to the block created here.

Actually, thinking about that, this is good. It means that the current block becomes the loop preheader and dominates the entire loop. Otherwise, we'll just create an empty block during loop canonicalization.

Maybe leave a comment indicating that this is desirable to ensure a loop preheader?

toolchain/check/handle_loop_statement.cpp
chandlerc: Looking at the test case and how things get split subsequently for conditions, this doesn't end up the latch, so maybe body is a better term after all.

chandlerc (approved):
Realized I should definitely have taken a look at this given how much this really interacts with LLVM and lowering eventually. Sorry for the delay.

Overall, really nice. Some minor comments, mostly around naming and maybe some added comments to clarify things.

For context (especially for folks less familiar with LLVM internals), a lot of these fall out of the [LLVM loop terminology][1], and the canonical form for loops, called in that document "loop simplified form". I suspect it's worth making sure SemIR produces something like loop simplified form out of the gate -- in particular, it will simplify any analyses we might want to build on top of SemIR significantly, which seems important as it is immutable. I think the current is close if not 100% there -- I'm not sure about the loop exit blocks being dedicated (IE, not having any non-loop predecessors) -- might be worth checking that. And likely worth linking to the LLVM definitions for readers. FWIW, the conceptual structure here isn't unique to LLVM at all, so it seems a reasonably broadly useful set of terminology and structural maxims to follow.

[1]: https://llvm.org/docs/LoopTerminology.html

LG with the comment adjustments. If more changes would be needed to fully achieve loop simplified form, happy for those to be follow-ups.

toolchain/check/handle_loop_statement.cpp
zygoloid: Yeah, this is first block in the body, not necessarily the block with the backedge.

toolchain/check/handle_loop_statement.cpp
zygoloid: Added a comment.

Generally, we've shied away from doing these kinds of peephole optimizations while building SemIR. But... we do do this optimization already, in general, while lowering unconditional branches. I've added a test demonstrating that the preheader (unfortunately) disappears when we produce LLVM IR, and a FIXME -- it's not completely trivial to fix that because it requires looking at the parse node to tell if we're building a preheader block, and lowering currently doesn't have access to the parse tree. (The test also exposed a case where we were producing invalid LLVM IR due to the entry block having predecessors; I've fixed that.)

toolchain/check/handle_loop_statement.cpp
zygoloid: Hah, I started off calling this "header" and then renamed it to "entry" to better parallel the "exit" block name. From a quick search, it looks like "header" is a substantially more popular term than "entry" outside LLVM too (though both seem to be used), and in any case, anchoring on LLVM terminology seems very reasonable here. Done :)

PR #3260, opened 2023-10-03
Add `name_reference` SemIR node for references to names.

zygoloid:
Also add `name_reference_untyped` for references to non-first-class names without types, which currently covers namespaces and functions.

This improves the fidelity of the SemIR representation, and fixes some issues where we would use the wrong location for nodes and diagnostics downstream of a name reference.

We're still missing a representation for dotted name expressions, such as `Namespace.Function`, and we don't use the `untyped` node as an operand of any other node yet.


toolchain/check/context.h
```diff
@@ -64,6 +64,9 @@ class Context {
   // Pops the top scope from scope_stack_, cleaning up names from name_lookup_.
   auto PopScope() -> void;
 
+  // Skip over NameReference nodes to find the value named by a given node.
+  auto SkipNameReferences(SemIR::NodeId node_id) -> SemIR::NodeId;
```
josh11b: Maybe: `FollowNameReferences`? I would expect `SkipNameReferences` to return the next `NodeId` if `node_id` was a name reference.

toolchain/check/context.cpp
```diff
@@ -144,6 +144,26 @@ auto Context::PopScope() -> void {
   }
 }
 
+auto Context::SkipNameReferences(SemIR::NodeId node_id) -> SemIR::NodeId {
+  while (true) {
```
chandlerc, line 148: (super trivial comment)

Slightly shorter?

```suggestion
  for (;;) {
```

Or do we have an established idiom for infinite loops? I don't feel strongly.

toolchain/check/testdata/array/array_in_place.carbon
```diff
@@ -23,11 +23,13 @@ fn G() {
 // CHECK:STDOUT:   %.loc10_28: i32 = int_literal 2
 // CHECK:STDOUT:   %.loc10_29: type = array_type %.loc10_28, (i32, i32, i32)
 // CHECK:STDOUT:   %v: ref [(i32, i32, i32); 2] = var "v"
+// CHECK:STDOUT:   %F.ref.loc10_34 = name_reference_untyped "F", package.%.loc7
```
chandlerc: For package references, this doesn't really seem to be working the way I would expect.

The `package.%.loc7` is much less friendly than `@F`, and we don't use the `F.ref.loc10_34` below in the `call` instruction.

Are either of these worth fixing in this iteration?

chandlerc (commented):
One main question here.

toolchain/check/context.cpp
zygoloid: We have about a dozen `while (true)` and zero `for (;;)` in C++ code, one of each in `docs/`, and one of each in `proposals/` (the `for (;;)` in docs and proposals are both pointing out that that's a C++ thing that doesn't exist in that form in Carbon).

toolchain/check/testdata/array/array_in_place.carbon
zygoloid: Yeah, I think there's still work to be done here:

- We should name the package-scope values after the functions they represent.
- Calls should reference the node they're using to name the function, rather than referring to the function directly.
- Name qualifiers like `N.M.F` should reference the various name components in some way. This PR only adds SemIR for `N`. This is a prerequisite for the previous bullet because we currently have no node representing the callee in `N.M.F()`.

I think the first and third changes are best done in separate PRs, and the second one depends on the third so it makes most sense to defer that too. Updated the PR description to mention this future work.

toolchain/check/context.cpp
chandlerc: 🤷🏻

chandlerc (approved):

toolchain/check/context.cpp
josh11b: I think we should do the thing that is available in Carbon, and if it seems painful, use that as a signal that we should address that in Carbon's design.

toolchain/check/context.h
zygoloid: Sure, works for me. Done.

toolchain/check/testdata/array/array_in_place.carbon
zygoloid: PR for item 1 landed before this PR :)

josh11b (approved):

PR #3266, opened 2023-10-05
Add signature enforcement for `Main.Run` and give it the symbol name `main`.

zygoloid:
For #2550.


chandlerc (approved):
LG, seems like a reasonable first step.

PR #3272, opened 2023-10-06
Remove unnecessory constexpr for lambdas in tests.

hanickadot:
After reading Chandler's code I asked him why he has these. Lambdas will implicitly constexpr anyway.


chandlerc (approved):
Nice catch!

PR #3273, opened 2023-10-06
Switch lexer to fully table-driven design.

chandlerc:
This uses the musttail dispatched table approach to drive the entire lexing. The result is that there is no main lexer loop at all in a traditional sense, now everything is driven through tail recursive dispatch on the next byte of the source text.

This should be easy to extend still -- the design pattern is to add lexer methods for handling specific cases, and then add a dispatch function to dispatch to them from the table. For example, we can add a method that handles decoding UTF-8 outside of the ASCII subset and set the table entries used by non-ASCII initial bytes to dispatch to it.

The performance is already surprisingly good, benchmarks show a modest improvement across the board. That's despite there still being some *serious* performance issues that I'll fix in a separate patch. There are also opportunities to leverage this structure more heavily as needed by putting more specialized dispatch targets in for specific bytes.

A follow-up PR will re-organize the functions here, as almost all of the methods on the `Lexer` should become private, but I wanted to keep that a separate change since it will probably render the diff even more hard to read than it already is.

### DIFF
--- a/toolchain/lex/tokenized_buffer.cpp
+++ b/toolchain/lex/tokenized_buffer.cpp
@@ -267,7 +267,7 @@ class TokenizedBuffer::Lexer {
     buffer_->token_infos_.back().has_trailing_space = true;
   }
 
-  auto LexHorizWhitespace(llvm::StringRef& source_text) -> void {
+  auto LexHorizontalWhitespace(llvm::StringRef& source_text) -> void {
     CARBON_DCHECK(source_text.front() == ' ' || source_text.front() == '\t');
     NoteWhitespace();
     ++current_column_;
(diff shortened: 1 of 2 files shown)


toolchain/lex/tokenized_buffer.cpp
```diff
@@ -271,77 +267,72 @@ class TokenizedBuffer::Lexer {
     buffer_->token_infos_.back().has_trailing_space = true;
   }
 
-  auto SkipWhitespace(llvm::StringRef& source_text) -> bool {
-    const char* const whitespace_start = source_text.begin();
+  auto LexHorizWhitespace(llvm::StringRef& source_text) -> void {
```
zygoloid: I don't think abbreviating "Horizontal" here is worthwhile, and is a bit jarring given that you don't abbreviate "Vertical" below.

zygoloid (approved):
Looks very nice, thanks!

toolchain/lex/tokenized_buffer.cpp
chandlerc: Agreed, will fix before landing. Had meant to switch this but forgot, thanks for catching it.

PR #3274, opened 2023-10-06
Unblock more lexer inlining.

chandlerc:
The big change is to make the lexer helpers have internal linkage,
making all of them easy to inline into single call sites.

Looking at the profile showed several other cases of unfortunate
out-of-line functions. Two were due to the code size produced for checks
-- those are switched to `DCHECK`s to remove that code from optimized
builds. The loss of coverage seems minor.

A last one was closing open groups. This was a surprising routine to be
hot, but it the paths to discover "nothing to do here" were intertwined
into the code. This PR extracts this common trace into a separate
function that delegates to the looping recovery path. This lets the hot
path inline easily.

At this point, for a large lexing benchmark I'm using, 50% of the time
is in the identifier hash table at this point. The remaining
improvements are to actually make some of the hot routines like symbol
lexing and comment lexing faster.

### DIFF
--- a/toolchain/lex/token_kind.h
+++ b/toolchain/lex/token_kind.h
@@ -52,7 +52,7 @@ class TokenKind : public CARBON_ENUM_BASE(TokenKind) {
   // The token kind must be an opening symbol.
   [[nodiscard]] auto closing_symbol() const -> TokenKind {
     auto result = ClosingSymbol[AsInt()];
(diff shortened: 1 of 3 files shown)


chandlerc:
(dependent PR merged in)

zygoloid (approved):
Looks good. This reminded me that I was working on #3282 a while back, and I think I forgot about it over a vacation.

PR #3275, opened 2023-10-07
Enhance the main lexer benchmark.

chandlerc:
This generalizes the main lexer benchmark's source generation to be
a bit more comprehensive, specifically including whitespace and
comments. Without these, we're missing a key part of the lexer's
performance.

This also tidies up a bit of the code and adds a more specific
distribution of the different factors in lexing based on analysis of
LLVM's source code. The enhancements to the source statistics script
that helped collect the data here will be in a separate PR.

With this, the output on my AMD cloud instance is:
```
-------------------------------------------------------------------------------------------------------------------------
Benchmark                                            Time             CPU   Iterations bytes_per_second tokens_per_second
-------------------------------------------------------------------------------------------------------------------------
BM_ValidKeywords                               2809011 ns      2808912 ns          243       212.616M/s         35.601M/s
BM_ValidIdentifiers<1, 64, false>             11461783 ns     11461652 ns           61       128.499M/s        8.72475M/s
BM_ValidIdentifiers<1, 1, true>                3397293 ns      3397240 ns          208       84.2158M/s        29.4357M/s
BM_ValidIdentifiers<3, 5, true>               14165143 ns     14164962 ns           51       40.3956M/s        7.05967M/s
BM_ValidIdentifiers<3, 16, true>              15283128 ns     15282583 ns           47       71.7623M/s         6.5434M/s
BM_ValidIdentifiers<12, 64, true>             17417323 ns     17417109 ns           41       219.007M/s        5.74148M/s
------------------------------------------------------------------------------------------------------------------------------------------
Benchmark                                            Time             CPU   Iterations bytes_per_second lines_per_second tokens_per_second
------------------------------------------------------------------------------------------------------------------------------------------
BM_RandomSource                                8132211 ns      8132227 ns           84       137.223M/s        3.9036M/s        12.2968M/s
BM_SpeedOfLightStrCpy                            29610 ns        29608 ns        24631       36.8065G/s       1072.17M/s        3.37745G/s
BM_SpeedOfLightDispatch<1>                     2144418 ns      2144421 ns          327       520.387M/s       14.8035M/s        46.6326M/s
BM_SpeedOfLightDispatch<2>                     1945954 ns      1945827 ns          351       573.498M/s       16.3144M/s         51.392M/s
BM_SpeedOfLightDispatch<4>                     2519565 ns      2519467 ns          292       442.923M/s       12.5999M/s        39.6909M/s
[...]

PR #3276, opened 2023-10-07
Add whitespace- and comment-specific benchmarking.

chandlerc:
These benchmarks zero-in and stress test horizontal and vertical
whitespace as well as comment lexing performance. They set up
essentially a worst-case scenario of ramping up whitespace between very
sparse tokens to show how the lexer copes with this.

The horizontal whitespace benchmark is perhaps less important as
frequent runs of 50-characters of horizontal whitespace are relatively
rare already, and likely to be exceedingly rare without trailing
comments. But its good to include for completeness and it shows
reasonably strong performance with the current table-dispatch approach.

The blank line and comment line benchmarks are much more important. Lots
of code is relatively line-sparse, especially API files that are perhaps
the most useful to parse quickly. And many of these are a mixture of
sparse with blank lines and sparse with large comment blocks. The
benchmarks show that there are some serious limits here, even falling
below 100k tokens per second throughput on some of the stress tests
here.


zygoloid (approved):

chandlerc:
Just noting, not merging until #3275 lands as this is stacked.

PR #3279, opened 2023-10-10
Workaround issues with sanitizers on macOS.

chandlerc:
The latest versions of Clang, at least on an ARM mac, enforce that we not use static sanitizer runtimes. The flag is also a bit frustrating: it has to be split out of other flags in order to remove it, you can't just disable it or ignore a warning about it being unused.

This gets things to be build and run again. However, the runtimes (I'm guessing ones that ship with the OS?) its using dynamically don't actually work -- at least one check was hitting pretty obvious false positives. So I've added a set of sanitizer flag workarounds we can expand as needed to continue to work around the limitations of sanitizers on this platform.

Together, this restores fastbuild on my ARM macOS with the latest Clang installed.

### DIFF
--- a/bazel/cc_toolchains/clang_cc_toolchain_config.bzl
+++ b/bazel/cc_toolchains/clang_cc_toolchain_config.bzl
@@ -459,6 +459,15 @@ def _impl(ctx):
         name = "sanitizer_common_flags",
         requires = [feature_set(["nonhost"])],
         implies = ["minimal_optimization_flags", "minimal_debug_info_flags", "preserve_call_stacks"],
+    )
+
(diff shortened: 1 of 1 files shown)


chandlerc:
Note that our incremental "impacted" test running doesn't work for this PR at all -- it thinks no tests are impacted, when all are.

Not sure how important that is to fix. When it lands, the non-PR workflow will run all tests and get us a fresh baseline. But wanted to call out that the checks here (and I suspect in the merge queue as well) will fail to cover this PR meaningfully.

zygoloid (approved):

PR #3280, opened 2023-10-10
Add typed nodes to SemIR.

zygoloid:
Replace `SemIR::Node::GetAsFoo` and `SemIR::Node::Foo::Make` with `SemIR::Foo` class that represents a particular kind of node, with named fields.

Rename `SemIR::IntegerLiteral` and `SemIR::RealLiteral` to `IntegerValue` / `RealValue` to better reflect their purpose and avoid a name collision with the corresponding `SemIR` node kinds.

Remove `NodeKind::Invalid` and the `SemIR::Node` default constructor entirely, as they were not used for anything.

### DIFF
--- a/toolchain/base/BUILD
+++ b/toolchain/base/BUILD
@@ -2,7 +2,7 @@
 # Exceptions. See /LICENSE for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-load("@rules_cc//cc:defs.bzl", "cc_library")
+load("@rules_cc//cc:defs.bzl", "cc_library", "cc_test")
 
 package(default_visibility = ["//visibility:public"])
 
@@ -20,5 +20,21 @@ cc_library(
     hdrs = ["pretty_stack_trace_function.h"],
     deps = [
         "@llvm-project//llvm:Support",
+    ],
+)
+
+cc_library(
+    name = "struct_reflection",
+    hdrs = ["struct_reflection.h"],
+)
+
+cc_test(
+    name = "struct_reflection_test",
+    size = "small",
+    srcs = ["struct_reflection_test.cpp"],
+    deps = [
+        ":struct_reflection",
+        "//testing/base:gtest_main",
+        "@com_google_googletest//:gtest",
     ],
 )
--- a/toolchain/base/BUILD
+++ b/toolchain/base/BUILD
@@ -2,7 +2,7 @@
 # Exceptions. See /LICENSE for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-load("@rules_cc//cc:defs.bzl", "cc_library", "cc_test")
+load("@rules_cc//cc:defs.bzl", "cc_library")
 
 package(default_visibility = ["//visibility:public"])
 
@@ -20,21 +20,5 @@ cc_library(
     hdrs = ["pretty_stack_trace_function.h"],
     deps = [
         "@llvm-project//llvm:Support",
-    ],
-)
-
-cc_library(
-    name = "struct_reflection",
-    hdrs = ["struct_reflection.h"],
-)
-
-cc_test(
-    name = "struct_reflection_test",
-    size = "small",
-    srcs = ["struct_reflection_test.cpp"],
-    deps = [
-        ":struct_reflection",
-        "//testing/base:gtest_main",
-        "@com_google_googletest//:gtest",
     ],
 )
--- a/toolchain/check/context.cpp
+++ b/toolchain/check/context.cpp
@@ -149,13 +149,11 @@ auto Context::FollowNameReferences(SemIR::NodeId node_id) -> SemIR::NodeId {
     auto node = semantics_ir().GetNode(node_id);
     switch (node.kind()) {
       case SemIR::NodeKind::NameReference: {
-        auto [name_id, value_id] = node.GetAsNameReference();
-        node_id = value_id;
+        node_id = node.As<SemIR::NameReference>().value_id;
         break;
       }
       case SemIR::NodeKind::NameReferenceUntyped: {
-        auto [name_id, value_id] = node.GetAsNameReferenceUntyped();
-        node_id = value_id;
+        node_id = node.As<SemIR::NameReferenceUntyped>().value_id;
         break;
       }
       default:
@@ -247,9 +245,10 @@ auto Context::AddCurrentCodeBlockToFunction() -> void {
     return;
   }
 
-  auto function_id = semantics_ir()
-                         .GetNode(return_scope_stack().back())
-                         .GetAsFunctionDeclaration();
+  auto function_id =
+      semantics_ir()
+          .GetNodeAs<SemIR::FunctionDeclaration>(return_scope_stack().back())
+          .function_id;
   semantics_ir()
       .GetFunction(function_id)
       .body_block_ids.push_back(node_block_stack().PeekOrAdd());
@@ -339,45 +338,48 @@ static auto ProfileType(Context& semantics_context, SemIR::Node node,
                         llvm::FoldingSetNodeID& canonical_id) -> void {
   switch (node.kind()) {
     case SemIR::NodeKind::ArrayType: {
-      auto [bound_id, element_type_id] = node.GetAsArrayType();
+      auto array_type = node.As<SemIR::ArrayType>();
       canonical_id.AddInteger(
-          semantics_context.semantics_ir().GetArrayBoundValue(bound_id));
-      canonical_id.AddInteger(element_type_id.index);
+          semantics_context.semantics_ir().GetArrayBoundValue(
+              array_type.bound_id));
+      canonical_id.AddInteger(array_type.element_type_id.index);
       break;
     }
     case SemIR::NodeKind::Builtin:
-      canonical_id.AddInteger(node.GetAsBuiltin().AsInt());
+      canonical_id.AddInteger(node.As<SemIR::Builtin>().builtin_kind.AsInt());
       break;
     case SemIR::NodeKind::CrossReference: {
       // TODO: Cross-references should be canonicalized by looking at their
       // target rather than treating them as new unique types.
-      auto [xref_id, node_id] = node.GetAsCrossReference();
-      canonical_id.AddInteger(xref_id.index);
-      canonical_id.AddInteger(node_id.index);
+      auto xref = node.As<SemIR::CrossReference>();
+      canonical_id.AddInteger(xref.ir_id.index);
+      canonical_id.AddInteger(xref.node_id.index);
       break;
     }
     case SemIR::NodeKind::ConstType:
       canonical_id.AddInteger(
-          semantics_context.GetUnqualifiedType(node.GetAsConstType()).index);
+          semantics_context.GetUnqualifiedType(node.As<SemIR::ConstType>().inner_id)
+              .index);
       break;
     case SemIR::NodeKind::PointerType:
-      canonical_id.AddInteger(node.GetAsPointerType().index);
+      canonical_id.AddInteger(node.As<SemIR::PointerType>().pointee_id.index);
       break;
     case SemIR::NodeKind::StructType: {
-      auto refs =
-          semantics_context.semantics_ir().GetNodeBlock(node.GetAsStructType());
-      for (const auto& ref_id : refs) {
-        auto ref = semantics_context.semantics_ir().GetNode(ref_id);
-        auto [name_id, type_id] = ref.GetAsStructTypeField();
-        canonical_id.AddInteger(name_id.index);
-        canonical_id.AddInteger(type_id.index);
+      auto fields = semantics_context.semantics_ir().GetNodeBlock(
+          node.As<SemIR::StructType>().fields_id);
+      for (const auto& field_id : fields) {
+        auto field =
+            semantics_context.semantics_ir().GetNodeAs<SemIR::StructTypeField>(
+                field_id);
+        canonical_id.AddInteger(field.name_id.index);
+        canonical_id.AddInteger(field.type_id.index);
       }
       break;
     }
     case SemIR::NodeKind::TupleType:
-      ProfileTupleType(
-          semantics_context.semantics_ir().GetTypeBlock(node.GetAsTupleType()),
-          canonical_id);
+      ProfileTupleType(semantics_context.semantics_ir().GetTypeBlock(
+                           node.As<SemIR::TupleType>().elements_id),
+                       canonical_id);
       break;
     default:
       CARBON_FATAL() << "Unexpected type node " << node;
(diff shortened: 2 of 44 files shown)


toolchain/base/BUILD
```diff
@@ -22,3 +22,19 @@ cc_library(
...
+
+cc_test(
+    name = "struct_reflection_test",
+    size = "small",
+    srcs = ["struct_reflection_test.cpp"],
+    deps = [
+        ":struct_reflection",
+        "//testing/base:gtest_main",
+        "@com_google_googletest//:gtest",
+    ],
```
chandlerc: Nothing really toolchain-specific here, how about putting this in //common?

(happy for that to be a separate PR or something if it makes merges or other stuff hard)

toolchain/check/context.cpp
```diff
@@ -339,45 +336,49 @@ static auto ProfileType(Context& semantics_context, SemIR::Node node,
...
   switch (node.kind()) {
     case SemIR::NodeKind::ArrayType: {
-      auto [bound_id, element_type_id] = node.GetAsArrayType();
+      auto array_type = node.As<SemIR::ArrayType>();
       canonical_id.AddInteger(
-          semantics_context.semantics_ir().GetArrayBoundValue(bound_id));
-      canonical_id.AddInteger(element_type_id.index);
+          semantics_context.semantics_ir().GetArrayBoundValue(
+              array_type.bound_id));
+      canonical_id.AddInteger(array_type.element_type_id.index);
```
chandlerc, line 343: In general for these cases where there is only one node really at play over a short scope with the explicit type already visible, I don't find losing the structured binding to be a net readability improvement... I guess the concern is keeping the destructured names in sync with the member names?

I somewhat feel like the destructuring brevity is maybe reasonable (and can be kept even w/ the new API) in super brief contexts like this?

Not arguing to keep it always though, especially across multiple scopes or with multiple different nodes where it's useful to make their accesses distinct and type checked.

toolchain/check/context.cpp
```diff
@@ -422,25 +423,24 @@ auto Context::CanonicalizeTupleType(Parse::Node parse_node,
...
+      SemIR::PointerType(parse_node, SemIR::TypeId::TypeType, pointee_type_id));
 }
 
 auto Context::GetUnqualifiedType(SemIR::TypeId type_id) -> SemIR::TypeId {
   SemIR::Node type_node =
       semantics_ir_->GetNode(semantics_ir_->GetTypeAllowBuiltinTypes(type_id));
-  if (type_node.kind() == SemIR::NodeKind::ConstType) {
-    return type_node.GetAsConstType();
+  if (auto const_type = type_node.TryAs<SemIR::ConstType>()) {
+    return const_type->inner_id;
```
chandlerc, line 443: This is super nice though...

toolchain/sem_ir/node.h
```diff
@@ -92,7 +93,7 @@ constexpr BoolValue BoolValue::False = BoolValue(0);
 constexpr BoolValue BoolValue::True = BoolValue(1);
 
 // The ID of an integer literal.
-struct IntegerLiteralId : public IndexBase, public Printable<IntegerLiteralId> {
+struct IntegerValueId : public IndexBase, public Printable<IntegerValueId> {
```
chandlerc: The comment and name are out of sync now.

I also feel like these aren't *quite* values -- they're closer to literals, no? They do the whole type-encodes-the-value thing?

Not sure what best to call them. They're not *lexical* literals, but "value" is getting fairly distinct meaning in the context of SemIR.

On the flip side, we use `BoolValue`, so there is also precedent.

toolchain/sem_ir/node.h
```diff
@@ -207,272 +208,329 @@ struct MemberIndex : public IndexBase, public Printable<MemberIndex> {
...
-//   `std::pair` of values.
-//   - Using the wrong `node.GetAsKind()` is a programming error, and should
-//     CHECK-fail in debug modes (opt may too, but it's not an API guarantee).
-//
-// Internally, each Kind uses the `Factory*` types to provide a boilerplate
-// `Make` and `Get` methods.
-class Node : public Printable<Node> {
- public:
-  struct NoArgs {};
+// Data storage for the operands of each kind of node.
```
chandlerc, line 211: Does it make sense to declare (and maybe use) these with the `.def` file to ensure that every one is covered, even if with a single line to make it an explicit alias?

Or will the failure of a missing node be obvious and caught already?

Could maybe expand this comment to hint a bit about how these are used or checked for consistency?

chandlerc (commented):
Just some initial thoughts while reading as you'd indicated it was worth looking through things even though its a WIP.

toolchain/check/context.cpp
josh11b: I think "keeping the destructured names in sync with the member names" is a significant issue. I'd be particularly concerned about the order getting swapped in the case where the two return values have the same type.

toolchain/sem_ir/node.h
```diff
@@ -207,272 +208,329 @@ struct MemberIndex : public IndexBase, public Printable<MemberIndex> {
...
+                                        << " to wrong kind " << Typed::Kind;
+    return Typed::FromRawData(parse_node_, type_id_, arg0_, arg1_);
+  }
+
+  // If this node is the given kind, returns a typed node, otherwise returns
+  // nullopt.
+  template <typename Typed>
+  auto TryAs() const -> std::optional<Typed> {
+    if (kind() == Typed::Kind) {
+      return As<Typed>();
```
josh11b, line 523: Would it be worth creating an `AsUnchecked` to call from here (and if we determine that some other call site is performance sensitive)? Or is the `CARBON_CHECK` in `As` going to get completely optimized away?

toolchain/base/BUILD
zygoloid: Done.

toolchain/check/context.cpp
zygoloid: Generally my concern with destructuring is that the names could get out of sync, or the field order could be accidentally reversed. The same problem exists when creating the nodes, though, and this patch isn't addressing it, so maybe that's a somewhat weak argument?  (On the other hand, we could support providing designators when creating nodes, if we wanted to make that side more explicit too.)

There's also a future-looking aspect here: @josh11b had mentioned a desire to switch to something more visitor-ish for these switches over node kinds, replacing

```c++
 switch (node.kind()) {
   case SemIR::NodeKind::ArrayType: {
      auto array_type = node.As<SemIR::ArrayType>();
      canonical_id.AddInteger(
          semantics_context.semantics_ir().GetArrayBoundValue(
              array_type.bound_id));
      canonical_id.AddInteger(array_type.element_type_id.index);
      break;
    }
    case SemIR::NodeKind::Builtin:
      canonical_id.AddInteger(node.As<SemIR::Builtin>().builtin_kind.AsInt());
      break;
    // ...
```

with something like

```c++
 node.Visit(
   [&] (SemIR::ArrayType array_type) {
      canonical_id.AddInteger(
          semantics_context.semantics_ir().GetArrayBoundValue(
              array_type.bound_id));
      canonical_id.AddInteger(array_type.element_type_id.index);
    },
    [&] (SemIR::Builtin builtin) {
      canonical_id.AddInteger(builtin.builtin_kind.AsInt());
    },
    // ...
```

and switching away from structured bindings towards the typed nodes prepares us for that.

toolchain/sem_ir/node.h
zygoloid: My expectation is that we'll use these IDs to represent both the type-encodes-the-value cases, and also the result of constant evaluation of integer arithmetic, so I think eventually they will be general compile-time integer values, and not specific to literals in the way they are now.

Comments updated.

toolchain/sem_ir/node.h
zygoloid: Comment expanded.

After the end of this namespace, we use the `.def` file to inject typed node wrappers for each node declared in the `.def` file, so missing nodes here will get caught, with an error saying that type `NodeData::YourNewNode` doesn't exist, which seems like it should be reasonably obvious. Missing nodes in the `.def` file will get caught because the typed node type won't exist (which might be a somewhat less obvious error).

toolchain/check/context.cpp
chandlerc: The absence of even a clang-tidy check to keep destructured names consistent, and the relative difficulty of having *internal* type safety within them is persuading me we should just use members as you have here.

However, I think the visitor thing needs (much) more discussion. I don't think the language support for this in C++ is good enough to be usable -- the lambda pattern I think will make compile time and other things scale really badly over time. Happy to chat about options here though, and that seems like a separate change, if and where we want to make it.

toolchain/sem_ir/node.h
chandlerc: My point is more that they aren't going to represent all integer "value"s (as opposed to references etc) in the Carbon file, at least in the broad sense -- they're not the value model for an `i32`. And while we likely wouldn't call that just an "integer value" it seems like this could be confused with the representation for the value expression category of integers rather than the representation for a *specific* compile time integer's computed value.

Not sure what a better name for this construct is though. `CompTimeIntegerId` seems much too verbose if clear. I wish we had a word for the category of things between literals and (runtime) values...

toolchain/check/context.cpp
```diff
@@ -233,8 +230,7 @@ auto Context::AddConvergenceBlockWithArgAndPush(
   // Acquire the result value.
   SemIR::TypeId result_type_id =
       semantics_ir().GetNode(*block_args.begin()).type_id();
-  return AddNode(
-      SemIR::Node::BlockArg::Make(parse_node, result_type_id, new_block_id));
+  return AddNode(SemIR::BlockArg(parse_node, result_type_id, new_block_id));
```
josh11b, line 233: I think if we are worried about getting the order wrong when reading out of a SemIR Node, we arguably should also be worried about getting the order wrong when constructing one. Can we talks about some factory function options that associate names with arguments?

toolchain/sem_ir/node.h
zygoloid: Mm, yeah, I think you've persuaded me. And in fact, I think my expectation was wrong:

While it's tempting to use this as the value model for compile-time `i32`s, we might want denser storage than an `APInt` for that case, and perhaps we shouldn't be using an `Id` mechanism for integer values that fit in 32 bits already. Even for `i128` compile-time values, it's not obvious to me that we'd want `APInt` storage rather than a dedicated table of `int128`s. So using this only for the type-encodes-the-value case might be best.

There's an even sharper distinction for `RealValueId`, where the type-encodes-the-value case wants an arbitrary rational number as the representation per the design, but once we go to `fN`, we presumably would want to represent the value as the bits of an `fN`.

So for now, perhaps the best thing to do is think of these as covering only the type-encodes-the-value cases. I'm not sure what that means for naming them, though.

toolchain/sem_ir/node.h
zygoloid: I tried skipping the check here, which does change some things in the generated code, but the difference appears to be negligible. (Overall this change makes the binary 376 bytes larger, I think because we end up with more copies of the code for `Typed::FromRawData`.)

toolchain/check/context.cpp
zygoloid: It'd be easy to make this work:

```suggestion
  return AddNode(SemIR::BlockArg(parse_node, result_type_id, {.block_id = new_block_id}));
```

Or this:

```suggestion
  return AddNode(SemIR::BlockArg({.parse_node = parse_node}, {.type_id = result_type_id},
                                 {.block_id = new_block_id}));
```

If we want a single braced list with the `parse_node`, `type_id`, and node operands, that's going to require more reworking of how we construct the node types.

chandlerc: FWIW, I'm much less concerned. I think we'll have many more places inspecting fields than constructing nodes, and I think the code querying fields is more at risk of silently "working" in ways that are confusing.

If this is a major issue, I'd hope to try making the parameters more type-safe first.

zygoloid: Here's what that might look like: https://github.com/carbon-language/carbon-lang/commit/5a2e244270c7416d32af3e1c42c198bb05d50037 (just a sample: not all uses of the constructor have been updated).

Interestingly, this immediately found some cases where nodes had inappropriate names for their operands.

chandlerc: (my comment above still holds for me at least after seeing the result)

josh11b: In discussion, I agreed to postpone doing this until we observe there to be problems it would avoid.

toolchain/check/context.cpp
josh11b: I think I was suggesting a visitor for parse nodes. Anyway, I don't think there is any changes for this PR.

chandlerc (approved):
just saying LG overall after discussion and w/ the rational / integer name change

toolchain/sem_ir/node.h
zygoloid: Renamed to `Real`, `Integer`, `RealId`, `IntegerId`. (Not renaming to `Rational` yet because we're not yet using a rational representation here, though "real" is even less accurate.)

toolchain/sem_ir/node.h
```diff
@@ -144,9 +144,8 @@ constexpr NodeBlockId NodeBlockId::Invalid =
 constexpr NodeBlockId NodeBlockId::Unreachable =
     NodeBlockId(NodeBlockId::InvalidIndex - 1);
 
-// The ID of a value of a real number type. Despite the name, this type only
-// represents rational numbers.
-struct RealValueId : public IndexBase, public Printable<RealValueId> {
+// The ID of a real number value.
+struct RealId : public IndexBase, public Printable<RealId> {
```
josh11b, line 148: https://en.wikipedia.org/wiki/Real_ID_Act

toolchain/lower/handle.cpp
```diff
@@ -399,112 +381,108 @@ auto EmitStructOrTupleValueRepresentation(FunctionContext& context,
...
   // TODO: Eventually this name will be optional, and we'll want to provide
   // something like `var` as a default. However, that's not possible right now
   // so cannot be tested.
-  auto name = context.semantics_ir().GetString(node.GetAsVarStorage());
-  auto* alloca = context.builder().CreateAlloca(context.GetType(node.type_id()),
+  auto name = context.semantics_ir().GetString(node.name_id);
+  auto* alloca = context.builder().CreateAlloca(context.GetType(node.type_id),
                                                 /*ArraySize=*/nullptr, name);
   context.SetLocal(node_id, alloca);
 }
```
josh11b, line 488: This file seems much improved.

toolchain/sem_ir/node_kind.def
```diff
@@ -40,8 +40,6 @@
 #error "Must define the x-macro to use this file."
 #endif
 
-CARBON_SEMANTICS_NODE_KIND_IMPL(Invalid, "invalid", None, NotTerminator)
```
josh11b, line 43: Reminder: consider moving the field information from the NodeData namespace node.h to this file in a future PR. This would keep all the information together and remove redundancy.

toolchain/sem_ir/node.h
```diff
@@ -510,6 +559,237 @@ class Node : public Printable<Node> {
...
+// Convert a field to its raw representation.
+constexpr auto ToRaw(IndexBase base) -> int32_t { return base.index; }
+constexpr auto ToRaw(BuiltinKind kind) -> int32_t { return kind.AsInt(); }
+
+template <typename T>
+using FieldTypes = decltype(StructReflection::AsTuple(std::declval<T>()));
+
+// Base class for nodes that contains the node data.
+template <typename T, typename = FieldTypes<T>>
+struct DataBase;
```
josh11b, line 719: https://en.wikipedia.org/wiki/Database

josh11b (approved):

PR #3282, opened 2023-10-10
Improve diagnosis of mismatched brackets.

zygoloid:
Move handling of mismatched brackets out of the main lexing loop into a separate pass that is only run if there are mismatched brackets This is done in preparation for using both lookahead and lookbehind to work out how to match brackets, and to get this code far away from the hot lexing loop.

Fix bracket insertion location to be immediately after the token that we're inserting the bracket after, rather than potentially at the end of a comment. When there are open brackets at the end of the file, say that there are open brackets, not that there's a closing bracket without a matching opening bracket.


zygoloid:
@chandlerc Before I rebase this on the latest lexer changes, I'd appreciate your thoughts on whether this is a good direction.

chandlerc:
> @chandlerc Before I rebase this on the latest lexer changes, I'd appreciate your thoughts on whether this is a good direction.

JYSK, I have .... a few more changes than I've even posted. =/ Sorry I'm creating a review backlog. Anyways, may not yet be time to rebase this.

I'll look at this approach though...

chandlerc:
This approach should work really well. Some of my later changes will actually I think make it even nicer. I like just doing a post-processing step when we find mismatched grouping symbols, it should be much nicer. =]

zygoloid:
Finally circled back and merged this :) Should be good for review now.

I have a bunch of ideas for how we could make this do much better error recovery, but those can be done in later PRs.

toolchain/lex/lex.cpp
```diff
@@ -997,46 +991,22 @@ auto Lexer::LexOpeningSymbolToken(llvm::StringRef source_text, TokenKind kind,
...
-    TokenIndex token = buffer_.AddToken({.kind = TokenKind::Error,
-                                         .token_line = current_line(),
-                                         .column = ComputeColumn(position),
-                                         .error_length = 1});
-    ++position;
-    return token;
-  };
+  TokenIndex token = LexOneCharSymbolToken(source_text, kind, position);
 
-  // If we have no open groups, this is an error.
```
chandlerc, line 1012: Seems good to keep some comment, and maybe add to it to explain that the error here will be handled later?

toolchain/lex/lex.cpp
```diff
@@ -997,46 +991,22 @@ auto Lexer::LexOpeningSymbolToken(llvm::StringRef source_text, TokenKind kind,
...
   }
-  open_groups_.pop_back();
 
-  // Now that the groups are all matched up, lex the actual token.
-  TokenIndex token = LexOneCharSymbolToken(source_text, kind, position);
-
-  // Note that it is important to get fresh token infos here as lexing the
-  // open token would invalidate any pointers.
+  TokenIndex opening_token = open_groups_.pop_back_val();
   buffer_.GetTokenInfo(opening_token).closing_token = token;
```
chandlerc: Can do this once above the `if` now and keep the token info alive for the duration.

Similarly, it's probably best to get the token info for the just-lexed token immediately after lexing it.

toolchain/lex/lex.cpp
```diff
@@ -1297,15 +1231,158 @@ auto Lexer::LexFileEnd(llvm::StringRef source_text, ssize_t position) -> void {
...
+// If brackets didn't pair or nest properly, find a set of places to insert
+// brackets to fix the nesting, issue suitable diagnostics, and update the
+// token list to describe the fixes.
+auto Lexer::DiagnoseAndFixMismatchedBrackets() -> void {
+  llvm::SmallVector<std::pair<int, TokenizedBuffer::TokenInfo>> new_tokens;
+
+  bool fixed_anything = false;
+
+  // Handle an opening token with no matching closing token.
+  auto handle_missing_close = [&](TokenIndex opening_token,
```
chandlerc: Extract this to its own method?

toolchain/lex/lex.cpp
```diff
@@ -1297,15 +1231,158 @@ auto Lexer::LexFileEnd(llvm::StringRef source_text, ssize_t position) -> void {
...
+  if (has_mismatched_brackets_ || !open_groups_.empty()) {
+    DiagnoseAndFixMismatchedBrackets();
+  }
+}
+
+// If brackets didn't pair or nest properly, find a set of places to insert
+// brackets to fix the nesting, issue suitable diagnostics, and update the
+// token list to describe the fixes.
+auto Lexer::DiagnoseAndFixMismatchedBrackets() -> void {
+  llvm::SmallVector<std::pair<int, TokenizedBuffer::TokenInfo>> new_tokens;
```
chandlerc: A comment on this to explain how it is used would seem helpful.

toolchain/lex/lex.cpp
```diff
@@ -1297,15 +1231,158 @@ auto Lexer::LexFileEnd(llvm::StringRef source_text, ssize_t position) -> void {
...
+
+      CARBON_CHECK(kind ==
+                   buffer_.GetTokenInfo(opening_token).kind.closing_symbol())
+          << "Failed to balance brackets";
+      auto& opening_token_info = buffer_.GetTokenInfo(opening_token);
+      auto& closing_token_info = buffer_.GetTokenInfo(token);
+      opening_token_info.closing_token = token;
+      closing_token_info.opening_token = opening_token;
+    }
+  }
```
chandlerc: Maybe extract this into a method to apply a list of token updates?

toolchain/lex/lex.cpp
```diff
@@ -1297,15 +1231,158 @@ auto Lexer::LexFileEnd(llvm::StringRef source_text, ssize_t position) -> void {
...
+  // Look for mismatched brackets and decide where to add tokens to fix them.
+  //
+  // TODO: For now, we use a greedy algorithm for this. We could do better by
+  // taking indentation into account. For example:
+  //
+  //     1  fn F() {
+  //     2    if (thing1)
+  //     3      thing2;
+  //     4    }
+  //     5  }
```
chandlerc, line 1370: Make sense to include this in a test?

chandlerc (commented):
Nice!! =D Some comments inline.

toolchain/lex/lex.cpp
zygoloid: Done.

toolchain/lex/lex.cpp
zygoloid: Done.

toolchain/lex/lex.cpp
zygoloid: Done.

toolchain/lex/lex.cpp
zygoloid: Done.

toolchain/lex/lex.cpp
zygoloid: Done.

toolchain/lex/lex.cpp
zygoloid: Done.

chandlerc (approved):
Really nice!

PR #3284, opened 2023-10-11
Add `SemIR::Node::Is<TypedNode>()`.

zygoloid:
This allows node kind checks to be written a little more concisely.


zygoloid:
Depends on #3283.

josh11b (approved):

PR #3285, opened 2023-10-11
Made function and namespace nodes typed to remove a crash

josh11b:
Bug found by fuzzing. Problem was untyped SemIR nodes had an invalid type id, which was retrieved by `HandlePrefixOperator` and then passed to `context.GetUnqualifiedType`, ultimately performing an invalid access in `semantics_ir_->GetNode`.

We prefer to make a placeholder type for functions and namespaces to remove the need for checking for the untyped case everywhere. Eventually functions will have their own types, but this approach will be needed for namespaces (and perhaps other non-first-class entities like unbound methods and interface members) long term.

Co-authored-by: Richard Smith <richard@metafoo.co.uk>

### DIFF
--- a/toolchain/check/context.cpp
+++ b/toolchain/check/context.cpp
@@ -145,21 +145,11 @@ auto Context::PopScope() -> void {
 }
 
 auto Context::FollowNameReferences(SemIR::NodeId node_id) -> SemIR::NodeId {
-  while (true) {
-    auto node = semantics_ir().GetNode(node_id);
-    switch (node.kind()) {
-      case SemIR::NodeKind::NameReference: {
-        node_id = node.As<SemIR::NameReference>().value_id;
-        break;
-      }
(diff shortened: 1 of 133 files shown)


chandlerc:
Happy to let @zygoloid handle the review, but did want to say this looks really nice, and I like the approach.

One thing that might be good is to expand the commit message to record some of the thinking that motivated this specific fix and/or why other approaches weren't appealing.

josh11b:
> Happy to let @zygoloid handle the review, but did want to say this looks really nice, and I like the approach.

@zygoloid was the one who figured this out, I was just the one typing.

> One thing that might be good is to expand the commit message to record some of the thinking that motivated this specific fix and/or why other approaches weren't appealing.

I took a stab at it.

zygoloid (approved):

PR #3290, opened 2023-10-12
Fix incorrect `argc` calculation.

zygoloid:
Fix off-by-one error: the terminating `nullptr` in `argv` is not included in `argc`. This is currently causing crashing tests to crash again in their crash handler, meaning we don't get a backtrace or CHECK-failure message.


chandlerc (approved):
LG!

PR #3291, opened 2023-10-12
Provide the callee expression to the Call node.

zygoloid:
Track the callee expression in full, instead of only tracking the callee's FunctionId. This results in the `name_reference` denoting the function actually being used.

Lowering now propagates a `llvm::Function*` as the value associated with expressions of type `<function>`.

We were not creating `NameReference` node for names produced by member access into a namespace, such as the second name in `Namespace.Function`, which caused lowering of calls to such names to fail. This is now fixed, but the resulting `NameReference` node only refers to the name and the lookup result, not to the `Namespace.` qualifier. We'll need to decide how to fit a third operand into that node (perhaps we can stop storing the `name_id`, since it can be derived from the lookup result) but for now the qualifier is not tracked.

### DIFF
--- a/toolchain/lower/handle.cpp
+++ b/toolchain/lower/handle.cpp
@@ -129,14 +129,14 @@ auto HandleBuiltin(FunctionContext& /*context*/, SemIR::NodeId /*node_id*/,
 
 auto HandleCall(FunctionContext& context, SemIR::NodeId node_id,
                 SemIR::Call node) -> void {
-  auto* llvm_function = context.GetFunction(node.function_id);
-  const auto& function = context.semantics_ir().GetFunction(node.function_id);
+  auto* callee = llvm::cast<llvm::Function>(context.GetLocal(node.callee_id));
 
   std::vector<llvm::Value*> args;
   llvm::ArrayRef<SemIR::NodeId> arg_ids =
       context.semantics_ir().GetNodeBlock(node.args_id);
 
-  if (function.return_slot_id.is_valid()) {
+  if (SemIR::GetInitializingRepresentation(context.semantics_ir(), node.type_id)
+          .has_return_slot()) {
     args.push_back(context.GetLocal(arg_ids.back()));
     arg_ids = arg_ids.drop_back();
   }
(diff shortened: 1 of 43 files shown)


toolchain/lower/handle.cpp
```diff
@@ -149,17 +149,13 @@ auto HandleCall(FunctionContext& context, SemIR::NodeId node_id,
...
-                     llvm::PoisonValue::get(context.GetType(node.type_id)));
-  } else {
-    context.SetLocal(node_id,
-                     context.builder().CreateCall(llvm_function, args,
-                                                  llvm_function->getName()));
+  auto* call = context.builder().CreateCall(callee, args);
+  context.SetLocal(node_id, call);
+
+  // Name the call's result the same as the callee.
+  // TODO: Is this a helpful name?
```
josh11b, line 156: Can we answer this question yet?

toolchain/lower/handle.cpp
```diff
@@ -196,12 +192,17 @@ auto HandleNameReference(FunctionContext& context, SemIR::NodeId node_id,
...
+  if (type_node_id == SemIR::NodeId::BuiltinNamespaceType) {
     return;
   }
-  // TODO: Handle name references to globals
-  context.SetLocal(node_id, context.GetLocal(node.value_id));
+
+  auto target = context.semantics_ir().GetNode(node.value_id);
+  if (auto function_decl = target.TryAs<SemIR::FunctionDeclaration>()) {
+    context.SetLocal(node_id, context.GetFunction(function_decl->function_id));
+  } else {
```
josh11b, line 202: What about
```
  auto target = context.semantics_ir().GetNode(node.value_id);
  if (target.Is<SemIR::NamespaceDeclaration>()) {
    return;
  } else if (auto function_decl = target.TryAs<SemIR::FunctionDeclaration>()) {
    context.SetLocal(node_id, context.GetFunction(function_decl->function_id));
  } else {
```
Would that work instead of the `type_node_id` business?

zygoloid: I think that'll work for now but will stop working when we add aliases -- we could have a name reference to an alias to a name reference to a namespace.

josh11b: Is that going to be a problem for aliases to functions as well?

zygoloid: I don't think there's a problem for aliases to functions: an alias to a function should propagate the value produced by the `NameReference` that directly names the function, and a `NameReference` to the alias should likewise propagate the value.

(I think there is a problem for `NameReference`s to global aliases, in general, because we don't handle references to globals properly in lowering. But that's not specific to functions or namespaces.)

toolchain/lower/handle.cpp
zygoloid: I'll ask for opinions on Discord.

josh11b (approved):

PR #3294, opened 2023-10-13
Bare-bones support for forward-declared classes.

zygoloid:
This is mostly scaffolding, but is just about enough for pointers to classes to work properly as types.


chandlerc (approved):
Super exciting!!!

PR #3295, opened 2023-10-13
Validate parse nodes correspond to expected tokens

josh11b:
Can specify which tokens are allowed generally, and any additional tokens that only occur when the parse node has an error.


toolchain/parse/node_kind.cpp
```diff
@@ -38,12 +39,53 @@ auto NodeKind::bracket() const -> NodeKind {
...
+    return;                                     \
+  }                                             \
+  break;
+
+#define CARBON_TOKEN_EITHER(Expected1, Expected2)               \
+  if (token_kind == Lex::TokenKind::Expected1 ||                \
+      (has_error && token_kind == Lex::TokenKind::Expected2)) { \
+    return;                                                     \
+  }                                                             \
+  break;
```
zygoloid: I find this macro name a bit confusing; I'd expect it to simply mean "either `Expected1` or `Expected2`", not "`Expected1`, but on error then `Expected2`".

toolchain/parse/node_kind.cpp
```diff
@@ -38,12 +39,53 @@ auto NodeKind::bracket() const -> NodeKind {
...
+  case NodeKind::Name:                 \
+    MatchAction
+
+#define CARBON_PARSE_NODE_KIND_BRACKET(Name, BracketName, MatchAction) \
+  CARBON_CASE(Name, MatchAction)
+
+#define CARBON_PARSE_NODE_KIND_CHILD_COUNT(Name, Size, MatchAction) \
+  CARBON_CASE(Name, MatchAction)
+
+#include "toolchain/parse/node_kind.def"
```
zygoloid: I wonder if it'd be clearer to allow `MatchAction` to have a sequence of actions, and separate out the error handling into a separate action:
```suggestion
#define CARBON_ANY_TOKEN return;

#define CARBON_TOKEN(Expected)                  \
  if (token_kind == Lex::TokenKind::Expected) { \
    return;                                     \
  }

#define CARBON_IF_ERROR(MatchActions) \
  if (has_error) {                    \
    MatchActions                      \
  }

#define CARBON_CASE(Name, MatchActions) \
  case NodeKind::Name:                  \
    MatchActions                        \
    break;

#define CARBON_PARSE_NODE_KIND_BRACKET(Name, BracketName, MatchActions) \
  CARBON_CASE(Name, MatchActions)

#define CARBON_PARSE_NODE_KIND_CHILD_COUNT(Name, Size, MatchActions) \
  CARBON_CASE(Name, MatchActions)

#include "toolchain/parse/node_kind.def"
```

(Suggestion continued in `.def` file.)

toolchain/parse/node_kind.def
```diff
@@ -41,22 +52,24 @@
...
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(FileEnd, 0, CARBON_TOKEN(EndOfFile))
 
 // An invalid parse. Used to balance the parse tree. Always has an error.
-CARBON_PARSE_NODE_KIND_CHILD_COUNT(InvalidParse, 0)
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(InvalidParse, 0, CARBON_ANY_TOKEN)
 
 // An empty declaration, such as `;`.
-CARBON_PARSE_NODE_KIND_CHILD_COUNT(EmptyDeclaration, 0)
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(EmptyDeclaration, 0,
+                                   CARBON_TOKEN_UNLESS_ERROR(Semi))
```
zygoloid:
```suggestion
CARBON_PARSE_NODE_KIND_CHILD_COUNT(EmptyDeclaration, 0,
                                   CARBON_TOKEN(Semi)
                                       CARBON_IF_ERROR(CARBON_ANY_TOKEN))
```

toolchain/parse/node_kind.def
```diff
@@ -65,25 +78,28 @@ CARBON_PARSE_NODE_KIND_CHILD_COUNT(NameExpression, 0)
...
-CARBON_PARSE_NODE_KIND_CHILD_COUNT(PackageApi, 0)
-CARBON_PARSE_NODE_KIND_CHILD_COUNT(PackageImpl, 0)
-CARBON_PARSE_NODE_KIND_CHILD_COUNT(PackageLibrary, 1)
-CARBON_PARSE_NODE_KIND_BRACKET(PackageDirective, PackageIntroducer)
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(PackageIntroducer, 0, CARBON_TOKEN(Package))
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(PackageApi, 0, CARBON_TOKEN(Api))
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(PackageImpl, 0, CARBON_TOKEN(Impl))
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(PackageLibrary, 1, CARBON_TOKEN(Library))
+CARBON_PARSE_NODE_KIND_BRACKET(PackageDirective, PackageIntroducer,
+                               CARBON_TOKEN_EITHER(Semi, Package))
```
zygoloid:
```suggestion
                               CARBON_TOKEN(Semi)
                                   CARBON_IF_ERROR(CARBON_TOKEN(Package)))
```

toolchain/parse/node_kind.def
```diff
@@ -11,13 +11,24 @@
...
+//   In both cases, LexTokenKind says which Lex::TokenKind values that this
+//   parse node can correspond to:
+//   - CARBON_TOKEN(kind): This node always corresponds to one kind of token.
+//   - CARBON_TOKEN_EITHER(kind1, kind2): This node corresponds to one of two
+//     kinds of tokens. Generally this will be the first kind of token, but in
+//     error cases where that token is not available in the source, it will
+//     fall back to using the second kind of token.
+//   - CARBON_TOKEN_UNLESS_ERROR(kind): this node corresponds to one kind of
+//     token unless there is an error, or
+//   - CARBON_ANY_TOKEN: this node can correspond to many kinds of tokens.
```
zygoloid:
```suggestion
//   In both cases, LexTokenKinds says which Lex::TokenKind values that this
//   parse node can correspond to, and is a sequence of:
//   - CARBON_TOKEN(kind): This node can correspond to this kind of token.
//   - CARBON_ANY_TOKEN: This node can correspond to many kinds of tokens.
//   - CARBON_IF_ERROR(LexTokenKinds): This node can additionally correspond
//     to the given kinds of tokens if its `has_error` flag is set.
```

toolchain/parse/node_kind.def
```diff
@@ -41,22 +52,24 @@
...
-CARBON_PARSE_NODE_KIND_CHILD_COUNT(FileStart, 0)
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(FileStart, 0, CARBON_TOKEN(StartOfFile))
 
 // The end of the file.
-CARBON_PARSE_NODE_KIND_CHILD_COUNT(FileEnd, 0)
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(FileEnd, 0, CARBON_TOKEN(EndOfFile))
 
 // An invalid parse. Used to balance the parse tree. Always has an error.
-CARBON_PARSE_NODE_KIND_CHILD_COUNT(InvalidParse, 0)
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(InvalidParse, 0, CARBON_ANY_TOKEN)
```
zygoloid:
```suggestion
CARBON_PARSE_NODE_KIND_CHILD_COUNT(InvalidParse, 0,
                                   CARBON_IF_ERROR(CARBON_ANY_TOKEN))
```
(This now encodes "always has an error" as part of the token kind checking.)

toolchain/parse/node_kind.def
```diff
@@ -280,47 +331,50 @@ CARBON_PARSE_NODE_KIND_BRACKET(CallExpression, CallExpressionStart)
...
 //   _external_: lhs expression
 //   _external_: rhs expression
 // QualifiedExpression
-CARBON_PARSE_NODE_KIND_CHILD_COUNT(PointerMemberAccessExpression, 2)
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(PointerMemberAccessExpression, 2,
+                                   CARBON_TOKEN(MinusGreater))
 
 // A literal.
-CARBON_PARSE_NODE_KIND_CHILD_COUNT(Literal, 0)
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(Literal, 0, CARBON_ANY_TOKEN)
```
zygoloid: We could list
```suggestion
CARBON_PARSE_NODE_KIND_CHILD_COUNT(
    Literal, 0,
    CARBON_TOKEN(False)
        CARBON_TOKEN(True)
        CARBON_TOKEN(IntegerLiteral)
        CARBON_TOKEN(RealLiteral)
        CARBON_TOKEN(StringLiteral)
        CARBON_TOKEN(Bool)
        CARBON_TOKEN(IntegerTypeLiteral)
        CARBON_TOKEN(UnsignedIntegerTypeLiteral)
        CARBON_TOKEN(FloatingPointTypeLiteral)
        CARBON_TOKEN(StringTypeLiteral)
        CARBON_TOKEN(Type))
```
here.

zygoloid (commented):
Looking through the use of the macros, I wonder if it's worth splitting them up slightly differently:

- Allow multiple token kinds / actions for a parse tree node.
- Add a wrapper to say "only allowed if error" rather than having special case macros for "this, or that on error" and "this, or anything on error".

That'd address a couple of concerns I have with this patch:

- `CARBON_TOKEN_EITHER` seems like a confusing name for what it's doing, because to me it doesn't suggest that the second token is only allowed if the node has an error.
- There's no support yet for the case where a parse node has a short(ish) finite list of possible tokens, which I think is a case where the list of possible tokens seems useful to document.

Some suggestions for that approach added.

toolchain/parse/node_kind.def
josh11b: Turns out that this wasn't actually being created with an error: https://github.com/carbon-language/carbon-lang/blob/trunk/toolchain/parse/handle_declaration_name_and_params.cpp#L36
I've fixed the code to make this true.

toolchain/parse/node_kind.cpp
```diff
@@ -38,12 +39,54 @@ auto NodeKind::bracket() const -> NodeKind {
...
+
+#define CARBON_PARSE_NODE_KIND_BRACKET(Name, BracketName, MatchActions) \
+  CARBON_CASE(Name, MatchActions)
+
+#define CARBON_PARSE_NODE_KIND_CHILD_COUNT(Name, Size, MatchActions) \
+  CARBON_CASE(Name, MatchActions)
+
+#include "toolchain/parse/node_kind.def"
+
+#undef CARBON_LOG
```
zygoloid, line 85:
```suggestion
#undef CARBON_LOG
#undef CARBON_CASE
```

zygoloid (approved):
Thanks!

PR #3299, opened 2023-10-16
Skip blocks of comments with identical prefixes.

chandlerc:
Specifically, after lexing a comment line, look at the next line and see
if it starts with an identical sequence of indent, comment '/'s and
character after the '/'s. If so, skip it as part of a block of comments.
This skips repeatedly diagnosing the same erroneous comment introducer
after the first one in a block, but that seems like a feature rather
than a bug.

The big motivation is to make sure the lexer is minimally impacted by
the length of comment blocks and skips them as efficiently as possible.
While they aren't exactly common, large block comments do come up and
it'd be unfortunate for those to actually slow down the toolchain.

It also happens that this is particularly easy to do because we're just
looking to see if we see the same prefix byte sequence. With SIMD we can
typically handle the most common indents with just a few instructions.

Because of the diagnostic differences, I've included a scalar fallback
that replicates the functionality but has no limit on indent size or CPU
features. I've also added testing to cover this behavior.

The only non-noise benchmark changes are as expected the comment ones,
with a nice improvement across the board:

```
BM_CommentLines/1/0/0                          15.8ms ± 2%  15.6ms ± 2%   -0.87%  (p=0.004 n=19+19)
BM_CommentLines/4/0/0                          20.2ms ± 1%  18.8ms ± 1%   -6.75%  (p=0.000 n=18+18)
BM_CommentLines/128/0/0                         221ms ± 1%   167ms ± 1%  -24.44%  (p=0.000 n=20+19)
BM_CommentLines/1/30/0                         16.6ms ± 3%  16.5ms ± 3%     ~     (p=0.175 n=19+20)
BM_CommentLines/4/30/0                         26.1ms ± 1%  24.8ms ± 2%   -5.05%  (p=0.000 n=18+19)
BM_CommentLines/128/30/0                        233ms ± 1%   185ms ± 1%  -20.38%  (p=0.000 n=19+20)
BM_CommentLines/1/70/0                         19.2ms ± 1%  19.0ms ± 2%   -0.66%  (p=0.016 n=19+20)
BM_CommentLines/4/70/0                         27.9ms ± 1%  26.6ms ± 1%   -4.63%  (p=0.000 n=19+19)
BM_CommentLines/128/70/0                        251ms ± 1%   213ms ± 1%  -15.18%  (p=0.000 n=20+18)
BM_CommentLines/1/0/2                          15.9ms ± 1%  15.8ms ± 2%     ~     (p=0.061 n=19+19)
BM_CommentLines/4/0/2                          20.5ms ± 2%  19.0ms ± 2%   -7.53%  (p=0.000 n=20+20)
BM_CommentLines/128/0/2                         213ms ± 1%   153ms ± 1%  -28.18%  (p=0.000 n=19+20)
BM_CommentLines/1/30/2                         16.8ms ± 2%  16.7ms ± 3%     ~     (p=0.134 n=20+20)
BM_CommentLines/4/30/2                         26.6ms ± 1%  25.2ms ± 3%   -5.50%  (p=0.000 n=20+20)
BM_CommentLines/128/30/2                        238ms ± 1%   187ms ± 2%  -21.49%  (p=0.000 n=17+19)
BM_CommentLines/1/70/2                         19.3ms ± 1%  19.4ms ± 3%     ~     (p=0.407 n=17+20)
BM_CommentLines/4/70/2                         28.2ms ± 1%  26.9ms ± 2%   -4.70%  (p=0.000 n=19+19)
BM_CommentLines/128/70/2                        257ms ± 2%   214ms ± 1%  -16.52%  (p=0.000 n=20+18)
BM_CommentLines/1/0/8                          16.3ms ± 2%  16.1ms ± 2%   -1.22%  (p=0.001 n=20+20)
BM_CommentLines/4/0/8                          22.7ms ± 2%  20.4ms ± 2%  -10.20%  (p=0.000 n=20+20)
BM_CommentLines/128/0/8                         244ms ± 1%   153ms ± 1%  -37.26%  (p=0.000 n=20+18)
BM_CommentLines/1/30/8                         17.3ms ± 2%  17.2ms ± 3%     ~     (p=0.192 n=20+20)
BM_CommentLines/4/30/8                         28.0ms ± 2%  25.6ms ± 3%   -8.46%  (p=0.000 n=19+18)
BM_CommentLines/128/30/8                        272ms ± 1%   196ms ± 2%  -27.90%  (p=0.000 n=18+20)
BM_CommentLines/1/70/8                         19.9ms ± 2%  19.9ms ± 2%     ~     (p=0.531 n=20+19)
BM_CommentLines/4/70/8                         29.3ms ± 1%  27.3ms ± 1%   -6.87%  (p=0.000 n=19+19)
BM_CommentLines/128/70/8                        292ms ± 1%   228ms ± 1%  -21.97%  (p=0.000 n=20+19)
```

### DIFF
--- a/toolchain/lex/testdata/fail_bad_comment_introducers.carbon
+++ b/toolchain/lex/testdata/fail_bad_comment_introducers.carbon
@@ -1,0 +1,82 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Most of the CHECK lines in this file are produced by switching this to
+// auto-update. However, the very last set of tests is specifically designed to
+// check behavior during the *transition* between comment block indents, and so
+// we don't want CHECK comments to be placed in-between them. After generating
+// the checks, this should have auto-update disabled and the last set of lines
+// relocated to not intermingle with the final set of test comments.
+// NOAUTOUPDATE
+//
+// CHECK:STDOUT: - filename: fail_bad_comment_introducers.carbon
+// CHECK:STDOUT:   tokens: [
+// CHECK:STDOUT:     { index: 0, kind: 'StartOfFile', line: {{ *\d+}}, column:  1, indent: 1, spelling: '', has_trailing_space: true },
+
+// Comments have to have whitespace after `//` currently.
+// CHECK:STDERR: fail_bad_comment_introducers.carbon:[[@LINE+3]]:3: ERROR: Whitespace is required after '//'.
+// CHECK:STDERR: //abc
+// CHECK:STDERR:   ^
+//abc
+
+    // CHECK:STDERR: fail_bad_comment_introducers.carbon:[[@LINE+3]]:7: ERROR: Whitespace is required after '//'.
+    // CHECK:STDERR:     //indented
+    // CHECK:STDERR:       ^
+    //indented
+
+// We only want to diagnose these on the first line of a block of similarly
+// indented comments.
+
+// CHECK:STDERR: fail_bad_comment_introducers.carbon:[[@LINE+3]]:3: ERROR: Whitespace is required after '//'.
+// CHECK:STDERR: //abc
+// CHECK:STDERR:   ^
+//abc
+//abc
+//abc
+
+        // CHECK:STDERR: fail_bad_comment_introducers.carbon:[[@LINE+3]]:11: ERROR: Whitespace is required after '//'.
+        // CHECK:STDERR:         //indented block
+        // CHECK:STDERR:           ^
+        //indented block
+        //indented block
+        //indented block
+        //indented block
+
+// As we have specialized code to scan short indented comment blocks, make sure
+// the behavior is preserved even for (absurdly) large indents.
+
+                                                                                // CHECK:STDERR: fail_bad_comment_introducers.carbon:[[@LINE+3]]:83: ERROR: Whitespace is required after '//'.
+                                                                                // CHECK:STDERR:                                                                                 //very indented block
+                                                                                // CHECK:STDERR:                                                                                   ^
+                                                                                //very indented block
+                                                                                //very indented block
+                                                                                //very indented block
+                                                                                //very indented block
+
+// Last but not least, we currently only provide this behavior when the comment
+// blocks have the *exact* same indent. If the indent changes, we should get a
+// new diagnostic for the new "block" comment.
+//
+// CHECK:STDERR: fail_bad_comment_introducers.carbon:[[@LINE+10]]:3: ERROR: Whitespace is required after '//'.
+// CHECK:STDERR: //abc
+// CHECK:STDERR:   ^
+// CHECK:STDERR: fail_bad_comment_introducers.carbon:[[@LINE+10]]:5: ERROR: Whitespace is required after '//'.
+// CHECK:STDERR:   //abc
+// CHECK:STDERR:     ^
+// CHECK:STDERR: fail_bad_comment_introducers.carbon:[[@LINE+10]]:3: ERROR: Whitespace is required after '//'.
+// CHECK:STDERR: //abc
+// CHECK:STDERR:   ^
+
+//abc
+//abc
+//abc
+  //abc
+  //abc
+  //abc
+//abc
+//abc
+//abc
+
+// CHECK:STDOUT:     { index: 1, kind:   'EndOfFile', line: {{ *}}[[@LINE+1]], column: {{ *\d+}}, indent: 1, spelling: '' },
+// CHECK:STDOUT:   ]
--- a/toolchain/lex/testdata/fail_bad_comment_introducers.carbon
+++ b/toolchain/lex/testdata/fail_bad_comment_introducers.carbon
@@ -26,8 +26,8 @@
 // CHECK:STDERR: //abc
 // CHECK:STDERR:   ^
 //abc
-//abc
-//abc
+//123
+//dce
 
         // CHECK:STDERR: fail_bad_comment_introducers.carbon:[[@LINE+3]]:11: ERROR: Whitespace is required after '//'.
         // CHECK:STDERR:         //indented block
--- a/toolchain/lex/testdata/fail_bad_comment_introducers.carbon
+++ b/toolchain/lex/testdata/fail_bad_comment_introducers.carbon
@@ -2,13 +2,7 @@
 // Exceptions. See /LICENSE for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// Most of the CHECK lines in this file are produced by switching this to
-// auto-update. However, the very last set of tests is specifically designed to
-// check behavior during the *transition* between comment block indents, and so
-// we don't want CHECK comments to be placed in-between them. After generating
-// the checks, this should have auto-update disabled and the last set of lines
-// relocated to not intermingle with the final set of test comments.
-// NOAUTOUPDATE
+// AUTOUPDATE
 //
 // CHECK:STDOUT: - filename: fail_bad_comment_introducers.carbon
 // CHECK:STDOUT:   tokens: [
@@ -53,30 +47,3 @@
                                                                                 //very indented block
                                                                                 //very indented block
                                                                                 //very indented block
-
-// Last but not least, we currently only provide this behavior when the comment
-// blocks have the *exact* same indent. If the indent changes, we should get a
-// new diagnostic for the new "block" comment.
-//
-// CHECK:STDERR: fail_bad_comment_introducers.carbon:[[@LINE+10]]:3: ERROR: Whitespace is required after '//'.
-// CHECK:STDERR: //abc
-// CHECK:STDERR:   ^
-// CHECK:STDERR: fail_bad_comment_introducers.carbon:[[@LINE+10]]:5: ERROR: Whitespace is required after '//'.
-// CHECK:STDERR:   //abc
-// CHECK:STDERR:     ^
-// CHECK:STDERR: fail_bad_comment_introducers.carbon:[[@LINE+10]]:3: ERROR: Whitespace is required after '//'.
-// CHECK:STDERR: //abc
-// CHECK:STDERR:   ^
-
-//abc
-//abc
-//abc
-  //abc
-  //abc
-  //abc
-//abc
-//abc
-//abc
-
-// CHECK:STDOUT:     { index: 1, kind:   'EndOfFile', line: {{ *}}[[@LINE+1]], column: {{ *\d+}}, indent: 1, spelling: '' },
-// CHECK:STDOUT:   ]
--- a/toolchain/lex/testdata/fail_bad_comment_introducers.carbon
+++ b/toolchain/lex/testdata/fail_bad_comment_introducers.carbon
@@ -5,7 +5,7 @@
 // AUTOUPDATE
 // CHECK:STDOUT: - filename: fail_bad_comment_introducers.carbon
 // CHECK:STDOUT:   tokens: [
-// CHECK:STDOUT:     { index: 0, kind: 'StartOfFile', line: {{ *\d+}}, column:   1, indent: 1, spelling: '', has_trailing_space: true },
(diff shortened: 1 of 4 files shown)


toolchain/lex/testdata/fail_bad_comment_introducers.carbon
```diff
@@ -0,0 +1,82 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Most of the CHECK lines in this file are produced by switching this to
+// auto-update. However, the very last set of tests is specifically designed to
+// check behavior during the *transition* between comment block indents, and so
+// we don't want CHECK comments to be placed in-between them. After generating
+// the checks, this should have auto-update disabled and the last set of lines
+// relocated to not intermingle with the final set of test comments.
```
jonmeow: You might want to split this into multiple, more targeted test files. (which would allow for autoupdate-able comments to be autoupdated without file edits)

That said, for the comments that you don't want to autoupdate inside, have you tried putting the AUTOUPDATE line below? If I'm understanding correctly, that should result in CHECK lines being inserted at the end (after the autoupdate line), rather than the normal interleaved which seems to be the concern here.

```
// comments you don't want things injected into

// AUTOUPDATE
// CHECK inserted here
```

jonmeow (commented):
I'm assuming zygoloid to be the better reviewer for this, just a quick comment on autoupdate.

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -49,6 +49,21 @@ auto VariantMatch(V&& v, Fs&&... fs) -> decltype(auto) {
...
+#define CARBON_USE_SIMD 1
+
+// A table of masks to include 0-16 bytes of an SSE register.
+static const std::array<__m128i, sizeof(__m128i) + 1> prefix_masks = [] {
+  std::array<__m128i, sizeof(__m128i) + 1> masks = {};
+  for (auto [i, mask] : llvm::enumerate(masks)) {
+    memset(&mask, 0xFF, i);
+  }
+  return masks;
+}();
```
zygoloid: Can this be made `constexpr`? A quick experiment on godbolt.org suggests we'd get a dynamic initializer here. The `memset` seems problematic but I think this would work:

```suggestion
// A table of masks to include 0-16 bytes of an SSE register.
static constexpr std::array<__m128i, sizeof(__m128i) + 1> prefix_masks = [] {
  std::array<__m128i, sizeof(__m128i) + 1> masks = {};
  for (auto [i, mask] : llvm::enumerate(masks)) {
    if (i > 0) {
      mask = (__m128i)((static_cast<unsigned __int128>(0) - 1) >> ((16 - i) * 8));
    }
  }
  return masks;
}();
```

or

```suggestion
static constexpr std::array<__m128i, sizeof(__m128i) + 1> prefix_masks = [] {
  std::array<__m128i, sizeof(__m128i) + 1> masks = {};
  for (auto [i, mask] : llvm::enumerate(masks)) {
    uint8_t byte_masks[16] = {};
    for (auto [j, byte_mask] : llvm::enumerate(byte_masks)) {
        byte_mask = j < i ? -1 : 0;
    }
    mask = (__m128i)__builtin_bit_cast(unsigned __int128, byte_masks);
  }
  return masks;
}();
```

(We can't `bit_cast` directly to `__m128i` at the moment, sadly, because Clang doesn't support vectors in `std::bit_cast` yet. But we can go via `__int128`.)

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -352,32 +376,92 @@ class [[clang::internal_linkage]] TokenizedBuffer::Lexer {
...
 
     // Any comment must be the only non-whitespace on the line.
-    if (set_indent_) {
+    const auto* line_info = current_line_info();
+    if (LLVM_UNLIKELY(position != line_info->start + line_info->indent)) {
       CARBON_DIAGNOSTIC(TrailingComment, Error,
                         "Trailing comments are not permitted.");
 
       emitter_.Emit(source_text.begin() + position, TrailingComment);
     }
```
zygoloid, line 393: I think we shouldn't fall through into the logic below in this case, and should skip just a single comment. Currently, I think that lexing:

```carbon
  var a: i32 = 1; // comment
  var b: i32 = 2;
  various_things(a, b);
  other_things(a, b);
  // ... more stuff ...
```
will skip lines 2 and 3 (but not 4) because the first three characters after the indent match, and that seems to not be ideal for error recovery purposes =)

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -352,32 +376,92 @@ class [[clang::internal_linkage]] TokenizedBuffer::Lexer {
...
 
     // The introducer '//' must be followed by whitespace or EOF.
-    if (position + 2 < static_cast<ssize_t>(source_text.size()) &&
+    if (LLVM_UNLIKELY(position + 2 <
+                      static_cast<ssize_t>(source_text.size())) &&
         !IsSpace(source_text[position + 2])) {
```
zygoloid: This expectation looks reversed to me -- I'd expect the `//` is almost always not at the very end of the file. Should this perhaps be either:
```suggestion
    if (LLVM_LIKELY(position + 2 <
                    static_cast<ssize_t>(source_text.size())) &&
        !IsSpace(source_text[position + 2])) {
```
or
```suggestion
    if (LLVM_UNLIKELY(position + 2 <
                          static_cast<ssize_t>(source_text.size()) &&
                      !IsSpace(source_text[position + 2]))) {
```
? If it's correct as-is, a comment explaining why would be helpful :)

chandlerc:
> I'm assuming zygoloid to be the better reviewer for this, just a quick comment on autoupdate.

I had thought you weren't back until Tuesday, sorry for that. Happy for you to take over whichever reviews you feel comfortable with! Also happy to chat on Discord about any of them.

toolchain/lex/testdata/fail_bad_comment_introducers.carbon
chandlerc: Ah, great suggestions!

I like the checks being next to most of these, but I split out the one problematic case and used your suggestion to keep it auto-updating too!

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -352,32 +376,92 @@ class [[clang::internal_linkage]] TokenizedBuffer::Lexer {
...
+        auto prefix_diff = _mm_xor_si128(prefix, next_prefix);
+        // If we have any differences (non-zero bits) within the mask, we can't
+        // skip the next line too.
+        if (!_mm_test_all_zeros(mask, prefix_diff)) {
+          break;
+        }
+
+        skip_to_next_line();
+      } while (position + 16 < static_cast<ssize_t>(source_text.size()));
+#endif
```
zygoloid, line 468:
```suggestion
#elif CARBON_USE_SIMD
#error Unknown target for SIMD comment skipping.
#endif
```
Just in case this `#if` and the one earlier in the file get out of sync.

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -352,32 +376,92 @@ class [[clang::internal_linkage]] TokenizedBuffer::Lexer {
...
+      auto* next_line_info = &buffer_.line_infos_[line_index];
+      next_line_info->indent = indent;
+      position = next_line_info->start;
+    };
+    if (CARBON_USE_SIMD &&
+        position + 16 < static_cast<ssize_t>(source_text.size()) &&
+        indent <= MaxIndent) {
+#if __x86_64__
+      // Load a mask based on the amount of text we want to compare.
+      auto mask = prefix_masks[indent + 3];
```
zygoloid: I think this should be `+ 2` if the character after `//` was not a space. As-is, I think we would get one diagnostic for
```carbon
//hello
//hello
```
but two for
```carbon
//hello
//world
```

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -352,32 +376,92 @@ class [[clang::internal_linkage]] TokenizedBuffer::Lexer {
...
+        // masked.
+        auto prefix_diff = _mm_xor_si128(prefix, next_prefix);
+        // If we have any differences (non-zero bits) within the mask, we can't
+        // skip the next line too.
+        if (!_mm_test_all_zeros(mask, prefix_diff)) {
+          break;
+        }
+
+        skip_to_next_line();
+      } while (position + 16 < static_cast<ssize_t>(source_text.size()));
```
zygoloid, line 460: If we fall out of this loop this way (rather than via the `break`;), we should go to the `else` block below so that we get consistent diagnostics for an invalid comment block at the very end of the file.

toolchain/lex/tokenized_buffer.cpp
chandlerc: Could I do this in a follow-up?

I've commoned this code between ARM and x86 and I'd like to experiment and come up with an approach that works for both.

toolchain/lex/tokenized_buffer.cpp
chandlerc: I meant for it to be on the entire condition (your second suggestion), but I'd rather it actually only apply to the is-space, so I'll do that instead.

toolchain/lex/tokenized_buffer.cpp
chandlerc: Good catch. Test added and fixed.

toolchain/lex/tokenized_buffer.cpp
chandlerc: Enhanced test and fixed.

toolchain/lex/tokenized_buffer.cpp
chandlerc: Yeah, you're right.

Fixing this with the code in its current structure is horrible though. I need to refactor this entire routine to address this.

Would you rather that happen in this PR or in a follow-up? Happy to do either one, especially as the code as-is has a subtle edge case.

jonmeow:
Looks good from my end, I'm going to let Richard finish his review though. :)

toolchain/lex/tokenized_buffer.cpp
zygoloid: Sure, that's fine by me.

toolchain/lex/tokenized_buffer.cpp
zygoloid: I'm happy with whichever option works best for you. We're not doing any differential fuzzing of the vectorized vs non-vectorized forms, so the fact that they give different diagnostics in an ill-formed edge case doesn't seem like something that should block short-term progress :)

If it really doesn't matter to you either way, I'd say land this change as-is, in the interest of making incremental progress.

toolchain/lex/testdata/fail_trailing_comments.carbon
```diff
@@ -0,0 +1,47 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Trailing comments should produce errors, but we should continue to parse (and diagnose) code errors beyond them.
+//
+// AUTOUPDATE
```
zygoloid: Autoupdate has caused us to no longer have a line with a matching prefix following a line with a trailing comment. Maybe move the `AUTOUPDATE` to the end of this file?

zygoloid (approved):
LG, please add TODOs for the things being left to follow-ups.

toolchain/lex/tokenized_buffer.cpp
chandlerc: K, will do. Left a TODO.

toolchain/lex/tokenized_buffer.cpp
chandlerc: K, landing as-is, but with a TODO.

toolchain/lex/testdata/fail_trailing_comments.carbon
chandlerc: Fixed.

chandlerc (commented):
Thanks -- merging with TODOs and will start working on those in follow-ups!

Appreciate all the suggested fixes both for this PR and followups!

jonmeow:
Minor note, looks like macos tests failed due to bazel download flakiness, so I triggered a rerun.

https://github.com/carbon-language/carbon-lang/actions/runs/6546052489/job/17775735278?pr=3299

PR #3300, opened 2023-10-16
Port the comment block scanning SIMD to Arm.

chandlerc:
This adds an Arm Neon code path for comment block scanning. It also
tries to tightened up the exact architecture-specific coding pattern for
SIMD code here. Notably, moving to always having architecture-specific
code inside a macro that can be used to globally disable SIMD, but any
place where *all* architectures will require some custom code, an `else`
branch with an error.

Overall, this makes large-block lexing >5% faster on an ARM server
I have access to, but that's a bit misleading. The Neon performance
there doesn't seem very good. On my M1 laptop the difference is *much*
larger. There, even 4-line comments show a noticable improvement and the
block speed looks well over 20%. Sadly, I don't have the same nice
scripts to generate good statistical data.

Raw benchmark data from an ARM sever for reference:

```
BM_CommentLines/1/0/0                          19.4ms ± 7%  19.7ms ± 6%    ~     (p=0.121 n=20+20)
BM_CommentLines/4/0/0                          24.8ms ± 7%  24.8ms ± 6%    ~     (p=0.904 n=20+20)
BM_CommentLines/128/0/0                         251ms ± 2%   234ms ± 3%  -6.70%  (p=0.000 n=20+20)
BM_CommentLines/1/30/0                         21.6ms ±10%  21.9ms ±10%    ~     (p=0.157 n=20+20)
BM_CommentLines/4/30/0                         28.8ms ± 9%  28.8ms ±11%    ~     (p=0.779 n=20+20)
BM_CommentLines/128/30/0                        248ms ± 2%   233ms ± 2%  -5.72%  (p=0.000 n=20+20)
BM_CommentLines/1/70/0                         23.4ms ±12%  23.7ms ±12%    ~     (p=0.341 n=20+20)
BM_CommentLines/4/70/0                         30.8ms ± 9%  31.1ms ±11%    ~     (p=0.602 n=20+20)
BM_CommentLines/128/70/0                        302ms ± 4%   292ms ± 4%  -3.46%  (p=0.000 n=18+20)
BM_CommentLines/1/0/2                          19.8ms ± 7%  20.0ms ± 6%    ~     (p=0.149 n=20+20)
BM_CommentLines/4/0/2                          25.1ms ± 7%  25.3ms ± 8%    ~     (p=0.659 n=20+20)
BM_CommentLines/128/0/2                         225ms ± 2%   212ms ± 2%  -5.88%  (p=0.000 n=20+20)
BM_CommentLines/1/30/2                         22.0ms ± 9%  22.2ms ±10%    ~     (p=0.289 n=20+20)
BM_CommentLines/4/30/2                         29.0ms ±11%  29.1ms ±12%    ~     (p=0.738 n=20+20)
BM_CommentLines/128/30/2                        261ms ±10%   243ms ± 3%  -6.85%  (p=0.000 n=20+20)
BM_CommentLines/1/70/2                         23.5ms ±11%  23.8ms ±15%    ~     (p=0.429 n=20+20)
BM_CommentLines/4/70/2                         31.3ms ±10%  31.5ms ±11%    ~     (p=0.478 n=20+20)
BM_CommentLines/128/70/2                        306ms ± 4%   292ms ± 4%  -4.52%  (p=0.000 n=18+19)
BM_CommentLines/1/0/8                          20.9ms ± 8%  21.2ms ± 7%    ~     (p=0.127 n=20+20)
BM_CommentLines/4/0/8                          27.3ms ± 9%  27.5ms ±12%    ~     (p=0.678 n=20+20)
BM_CommentLines/128/0/8                         227ms ± 2%   210ms ± 2%  -7.35%  (p=0.000 n=19+20)
BM_CommentLines/1/30/8                         22.6ms ±11%  23.0ms ±10%    ~     (p=0.114 n=20+20)
BM_CommentLines/4/30/8                         29.4ms ±10%  29.4ms ±12%    ~     (p=0.947 n=20+20)
BM_CommentLines/128/30/8                        275ms ± 4%   257ms ± 7%  -6.59%  (p=0.000 n=19+20)
BM_CommentLines/1/70/8                         23.9ms ±13%  24.3ms ±14%    ~     (p=0.265 n=20+20)
BM_CommentLines/4/70/8                         32.3ms ±11%  32.4ms ± 9%    ~     (p=0.478 n=20+20)
BM_CommentLines/128/70/8                        319ms ± 4%   307ms ± 4%  -3.83%  (p=0.000 n=18+19)
```


toolchain/lex/tokenized_buffer.cpp
```diff
@@ -49,20 +55,35 @@ auto VariantMatch(V&& v, Fs&&... fs) -> decltype(auto) {
...
+#elif __x86_64__
+    std::array<__m128i, sizeof(__m128i) + 1>
 #else
-#define CARBON_USE_SIMD 0
+#error "Unsupported SIMD architecture!"
+#endif
+        prefix_masks = [] {
+          using MaskArrayT = std::remove_const_t<decltype(prefix_masks)>;
+          using MaskT = MaskArrayT::value_type;
+          MaskArrayT masks = {};
```
jonmeow: Had you considered moving the `using` into the platform `#if`, removing the `remove_const_t`, e.g.:

```
#if _ARM_NEON
using MaskT = uint8x16_t;
#elif __x86_64__
using MaskT = __m128i;
#else
#error "Unsupported SIMD architecture!"
#endif
using MaskArrayT = std::array<MaskT, sizeof(MaskT) + 1>;
static const MaskArrayT prefix_masks = [] {
  MaskArrayT masks = {};
```

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -49,20 +55,35 @@ auto VariantMatch(V&& v, Fs&&... fs) -> decltype(auto) {
...
+          for (auto [i, mask] : llvm::enumerate(masks)) {
+            std::array<uint8_t, sizeof(MaskT)> mask_bytes = {};
+            memset(&mask_bytes, 0xFF, i);
+#if __ARM_NEON
+            mask = vld1q_u8(mask_bytes.data());
+#elif __x86_64__
+            mask = _mm_loadu_si128(
+                reinterpret_cast<const __m128i*>(mask_bytes.data()));
```
jonmeow: Given `MaskT` on line 73, would it help readability if this used the same even though it's in a platform `#if`?

jonmeow (commented):
Just a minor style comment since I was glancing.

toolchain/lex/tokenized_buffer.cpp
chandlerc: Sure, done. Had to add an anonymous namespace and change the names to make more sense in the broader scope, but seems fine and especially with a portable way of building the mask seems pretty clean.

toolchain/lex/tokenized_buffer.cpp
chandlerc: I generally like having the platform specific types when using the platform specific intrinsics as it makes how things are working more clear for me.

But this is also somewhat obviated -- the suggestion to make this `constexpr` not just works, but works portably and so this is now portable code using the typedef. =]

chandlerc (commented):
All updated to merge in the changes (including bug fixes) from #3299 -- should be good for a last round of review.

This also now addresses the TODO around `constexpr`-ing the array, as it ended up working nicely in both architectures.

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -49,21 +55,29 @@ auto VariantMatch(V&& v, Fs&&... fs) -> decltype(auto) {
...
+namespace {
+#if __ARM_NEON
+using SIMDMaskT = uint8x16_t;
+#elif __x86_64__
+using SIMDMaskT = __m128i;
+#else
+#error "Unsupported SIMD architecture!"
+#endif
+using SIMDMaskArrayT = std::array<SIMDMaskT, sizeof(SIMDMaskT) + 1>;
+}  // namespace
```
jonmeow, line 68: What's the effect of this namespace? I thought `using` wouldn't manifest anything?

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -49,21 +55,29 @@ auto VariantMatch(V&& v, Fs&&... fs) -> decltype(auto) {
...
-static const std::array<__m128i, sizeof(__m128i) + 1> prefix_masks = [] {
-  std::array<__m128i, sizeof(__m128i) + 1> masks = {};
-  for (auto [i, mask] : llvm::enumerate(masks)) {
-    memset(&mask, 0xFF, i);
+static constexpr SIMDMaskArrayT PrefixMasks = []() constexpr {
+  SIMDMaskArrayT masks = {};
+  for (int i = 1; i < static_cast<int>(masks.size()); ++i) {
+    // The SIMD types and constexpr require a C-style cast.
+    // NOLINTNEXTLINE(google-readability-casting)
+    masks[i] = (SIMDMaskT)((static_cast<unsigned __int128>(0) - 1) >>
```
jonmeow:
```suggestion
    masks[i] = (SIMDMaskT)(std::numeric_limits<unsigned __int128>::max() >>
```

Is numeric limits going to give you what you want?

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -435,10 +449,28 @@ class [[clang::internal_linkage]] TokenizedBuffer::Lexer {
     if (CARBON_USE_SIMD &&
         position + 16 < static_cast<ssize_t>(source_text.size()) &&
         indent <= MaxIndent) {
-#if __x86_64__
       // Load a mask based on the amount of text we want to compare.
-      auto mask = prefix_masks[prefix_size];
-      // And use the current line's prefix as the exemplar to compare against.
+      auto mask = PrefixMasks[prefix_size];
+#if __ARM_NEON
+      // Load and mask the prefix if the current line.
```
jonmeow:
```suggestion
      // Load and mask the prefix of the current line.
```

jonmeow (approved):
Looks great! Just a couple minor comments.

toolchain/lex/tokenized_buffer.cpp
chandlerc: Eh, I don't find this as easy to read, but sure.

toolchain/lex/tokenized_buffer.cpp
chandlerc: I dunno, I have a simple rule that I work to never define a name in a `.cpp` file without either `static` or an anonymous namespace. It seems awfully subtle to rely on this being a using declaration (as opposed to a struct wrapping the type or something else) to avoid potential ODR conflicts with some other `SIMDMaskT` type in the wild.

jonmeow: Isn't `using` similar to `#define` with how it introduces file-scoped names though? I'm not sure what the subtlety is, is there a case where `using` can cause ODR violations?

For me, to be clear -- the namespace has an obfuscating effect, it's making me look for what it's fixing.

jonmeow:
To be sure, feel free to merge either way on the namespace comment.

toolchain/lex/tokenized_buffer.cpp
chandlerc: Ok, for me, I really can't keep track of what in C++ is or isn't an ODR violation. I think it is much better to just consistently either use an anonymous namespace or the `static` keyword. We do this for tests already for example.

That said, if other folks would prefer to only use an anonymous namespace when it is strictly needed, I'll switch in a follow-up. @zygoloid as maybe a third party here.

That said, you indicate this isn't blocking (thanks! I was confused), so landing the PR as-is for now.

PR #3302, opened 2023-10-16
Basic support for incomplete types.

zygoloid:
Incomplete types may be nested within other types; for example, a tuple type might have an incomplete type as an element. Handle such cases by walking through nested incomplete types when completing a type. This is done non-recursively in case a very complex type is formed.

Types are generally no longer completed at the point where they're formed. Instead, we attempt to complete a type when it is used in a context that requires a complete type, and diagnose if the type cannot be completed at that point. This will be necessary for classes, which can become complete after their first use, and helps tease out bugs where a type completeness check is missing.


toolchain/check/testdata/class/fail_incomplete.carbon
```diff
@@ -0,0 +1,169 @@
...
+fn MemberAccess(p: Class*) -> i32 {
+  // CHECK:STDERR: fail_incomplete.carbon:[[@LINE+9]]:11: ERROR: Invalid use of incomplete type `Class`.
+  // CHECK:STDERR:   return (*p).n;
+  // CHECK:STDERR:           ^
+  // CHECK:STDERR: fail_incomplete.carbon:[[@LINE-27]]:1: Class was forward declared here.
+  // CHECK:STDERR: class Class;
+  // CHECK:STDERR: ^
+  // CHECK:STDERR: fail_incomplete.carbon:[[@LINE+3]]:14: ERROR: Type `Class` does not support qualified expressions.
+  // CHECK:STDERR:   return (*p).n;
+  // CHECK:STDERR:              ^
```
jonmeow, line 39: Should this diagnostic be suppressed for the incomplete type? i.e., `(*p)` is an invalid expression, should it produce an invalid type in order to prevent cascading diagnostics, rather than `Class`?

Conversely, `ReturnIncomplete().n` might be desirable to make behave the same way as this. Right now, I think it won't produce a diagnostic because `ReturnIncomplete()` is producing an error node instead of `Class`. (which might be good to test)

toolchain/sem_ir/file.h
```diff
@@ -115,7 +114,7 @@ struct TypeInfo : public Printable<TypeInfo> {
 
   // The node that defines this type.
   NodeId node_id;
-  // The value representation for this type.
+  // The value representation for this type. Only set when the type is complete.
```
jonmeow: If it's only set for complete types, shouldn't it have an explicit default to "Unknown"? Or should the comment instead say something like "Set to unknown for incomplete types"?

toolchain/check/handle_pattern_binding.cpp
```diff
@@ -20,6 +20,7 @@ auto HandleGenericPatternBinding(Context& context, Parse::Node parse_node)
 auto HandlePatternBinding(Context& context, Parse::Node parse_node) -> bool {
   auto [type_node, parsed_type_id] =
       context.node_stack().PopExpressionWithParseNode();
+  auto type_node_copy = type_node;
```
jonmeow, line 23: Sorry, this code took a moment for me to figure out why you would make a copy here. Auto bindings and lambdas have a disappointing interaction, I'm just still wishing there were a better way to handle this. I wonder if we should come up with an idiom that isn't just `_copy`, but `_for_lambda` feels slightly long, and always using the copied value seems likely to pass up optimizations that we might care about.

This comment doesn't need changes, just thinking.

toolchain/check/convert.cpp
```diff
@@ -634,6 +634,27 @@ auto Convert(Context& context, Parse::Node parse_node, SemIR::NodeId expr_id,
     return SemIR::NodeId::BuiltinError;
   }
 
+  // We can only perform initialization for complete types.
+  if (!context.TryToCompleteType(target.type_id, [&] {
```
jonmeow, line 638: Looking through calls to TryToCompleteType, could it instead be taking a diagnostic to print that has a single std::string argument, rather than a lambda? This looks like the most complex case, but could still pass in a single diagnostic.

toolchain/check/context.h
```diff
@@ -21,6 +21,9 @@ namespace Carbon::Check {
 // Context and shared functionality for semantics handlers.
 class Context {
  public:
+  using DiagnosticEmitter = Carbon::DiagnosticEmitter<Parse::Node>;
```
jonmeow, line 24: I worry that shadowing names like this will hinder readability because it misleads about the type in use, how necessary is it? Had you considered using a different name, or only aliasing DiagnosticBuilder? (in particular, it looks like you did this specifically for DiagnosticBuilder?)

toolchain/check/context.cpp
```diff
@@ -279,233 +278,375 @@ auto Context::ParamOrArgEnd(Parse::NodeKind start_kind) -> SemIR::NodeBlockId {
   return ParamOrArgPop();
 }
 
-// Attempts to complete the given type.
-auto Context::TryToCompleteType(SemIR::TypeId type_id) -> bool {
-  auto node_id = semantics_ir().GetTypeAllowBuiltinTypes(type_id);
-  auto node = semantics_ir().GetNode(node_id);
+namespace {
+// State used when recursively completing a type.
+struct TypeCompleter {
```
jonmeow: How much complexity would be added in giving this its own file? It looks like 360 lines, which feels long enough to make it an option at least, maybe more important if it were to grow.

toolchain/check/context.cpp
```diff
@@ -279,233 +278,375 @@ auto Context::ParamOrArgEnd(Parse::NodeKind start_kind) -> SemIR::NodeBlockId {
   return ParamOrArgPop();
 }
 
-// Attempts to complete the given type.
-auto Context::TryToCompleteType(SemIR::TypeId type_id) -> bool {
-  auto node_id = semantics_ir().GetTypeAllowBuiltinTypes(type_id);
-  auto node = semantics_ir().GetNode(node_id);
+namespace {
+// State used when recursively completing a type.
```
jonmeow: Should this comment include a high-level overview of the workflow? Or maybe Push/ProcessAll/Process should have comments that explain it?

toolchain/check/context.cpp
```diff
@@ -279,233 +278,375 @@ auto Context::ParamOrArgEnd(Parse::NodeKind start_kind) -> SemIR::NodeBlockId {
...
     return true;
-  };
+  }
 
-  auto set_copy_representation = [&](SemIR::TypeId rep_id) {
-    semantics_ir().CompleteType(
-        type_id, {.kind = SemIR::ValueRepresentation::Copy, .type_id = rep_id});
-    return true;
-  };
+  auto Process() -> bool {
```
jonmeow: Maybe rename to `PopAndProcess` or `ProcessNext` to clarify that it just processes one item?

Going a little further, from an API perspective, maybe a public `Process(SemIR::TypeId type_id) -> bool` combining `ProcessAll` with a `Push`? That is, will `ProcessAll` be called without calling `Push` precisely once before it?

toolchain/check/context.cpp
jonmeow: By style, should this be a class? That is, structs are for passive data classes, everything else should be a class. This seems to have a lot of functionality, beyond the data it holds. It also has enough functions that private/public boundaries might help clarify what's an API.

https://google.github.io/styleguide/cppguide.html#Structs_vs._Classes

toolchain/check/context.cpp
```diff
@@ -279,233 +278,375 @@ auto Context::ParamOrArgEnd(Parse::NodeKind start_kind) -> SemIR::NodeBlockId {
...
+    auto value_rep =
+        context.semantics_ir().GetValueRepresentation(nested_type_id);
+    CARBON_CHECK(value_rep.kind != SemIR::ValueRepresentation::Unknown)
+        << "Complete type should have a value representation";
+    return value_rep;
+  };
+
+  // Builds and returns the value representation for the given type. All nested
+  // types, as found by AddNestedIncompleteTypes, are known to be complete.
+  auto BuildValueRepresentation(SemIR::TypeId type_id, SemIR::Node node) const
```
jonmeow, line 572: This function is about 190 lines long, what would you think of breaking it up a little per short function style? In particular, CrossReference, StructType, and TupleType are longer blocks of logic, maybe they could be split out in a helpful way?
https://google.github.io/styleguide/cppguide.html#Write_Short_Functions

toolchain/check/context.cpp
```diff
@@ -665,6 +796,15 @@ auto Context::CanonicalizeTupleType(Parse::Node parse_node,
                               make_tuple_node);
 }
 
+auto Context::GetBuiltinType(SemIR::BuiltinKind kind) -> SemIR::TypeId {
+  CARBON_CHECK(kind != SemIR::BuiltinKind::Invalid);
+  auto type_id = CanonicalizeType(SemIR::NodeId::ForBuiltin(kind));
+  // To keep client code simpler, complete builtin types before returning them.
+  bool complete = TryToCompleteType(type_id);
+  CARBON_CHECK(complete) << "Failed to complete builtin type";
```
jonmeow, line 840: Thinking about type-completion of builtin types, something like `bool` may not need to really do type completion (because it's likely truly built-in), instead doing some shorter logic (which might be faster, and for a frequently used type -- possibly skipping TryToTypeComplete). However, something like `i32` might need to do type-completion with lookup in the prelude, and could consequently fail (for example, if the prelude is disabled, or replaced).

Maybe this CHECK should be a diagnostic, to handle the incomplete `i32` case?

toolchain/check/context.cpp
```diff
@@ -279,233 +278,375 @@ auto Context::ParamOrArgEnd(Parse::NodeKind start_kind) -> SemIR::NodeBlockId {
...
+      case SemIR::ConstType::Kind:
+        // The value representation of `const T` is the same as that of `T`.
+        // Objects are not modifiable through their value representations.
+        return GetNestedValueRepresentation(
+            node.As<SemIR::ConstType>().inner_id);
     }
   }
 
-  llvm_unreachable("All node kinds were handled above");
+  enum class Phase {
```
```suggestion
  enum class Phase : int8_t {
```

https://clang.llvm.org/extra/clang-tidy/checks/performance/enum-size.html

jonmeow (commented):
Looks good overall, just some comments on details of implementation.

toolchain/check/testdata/class/fail_incomplete.carbon
zygoloid: Yes, we should suppress the follow-on diagnostic here. I'll send a separate PR for that.

toolchain/check/handle_pattern_binding.cpp
zygoloid: Let's switch to building the toolchain in C++20 mode. I'll send a PR.

toolchain/check/convert.cpp
zygoloid: We'd need to pass in a parse node as well, and that approach wouldn't support diagnostics that have additional arguments, cases where the diagnostic should have notes attached, etc. My experience from Clang is that passing in a diagnostic and having other code emit it is inflexible and should be avoided if we can.

toolchain/check/context.h
zygoloid: Perhaps a better solution would be to put the diagnostic mechanism in a namespace, so this would be `DiagnosticEmitter` versus `Diagnostic::Emitter<...>`?

toolchain/check/context.cpp
zygoloid: I want to move all the type handling out of `Check::Context` into its own file and class. But I'd prefer to not do that as part of this PR, to keep things simpler.

toolchain/check/context.cpp
zygoloid: Sure, done.

toolchain/check/context.cpp

toolchain/check/context.cpp
zygoloid: Once we start treating `i32` as a class type, I would expect we won't go through this function to form it, so the diagnostic would never be reachable. `TryToCompleteType` returns pretty quickly if the builtin type is already complete (and that codepath is being fully inlined into `GetBuiltinType`), so the cost here should be low except perhaps for the first time we get a particular builtin type.

toolchain/check/context.cpp
zygoloid: Done.

toolchain/sem_ir/file.h

toolchain/check/context.cpp
zygoloid: I'm sad about this change -- I think this makes the code communicate its intent less well, by suggesting that these enumerators have some meaning as integer values. But I suppose we need to do this in order to allow this clang-tidy check to deliver value elsewhere.

Done.

toolchain/check/context.h
jonmeow: Sure, that's fair, and out-of-scope here.

toolchain/check/context.cpp
jonmeow: SGTM

toolchain/check/convert.cpp
jonmeow: FWIW, part of why I'm suggesting this is I worry about the overhead with lambdas, but probably okay for now. I mean, I may misunderstand the lambda overhead too.

toolchain/check/context.cpp
jonmeow: Eh, I'm maybe expecting less to change, but hopefully it's not an issue.

jonmeow (approved):
Thanks for the changes!

toolchain/check/testdata/class/fail_incomplete.carbon
zygoloid: Sent #3303.

toolchain/check/convert.cpp
zygoloid: The cost of calling a lambda through an `llvm::function_ref` should be a single indirect function call, which I'd expect to be negligible compared to other costs of producing the diagnostic (eg, allocating memory, formatting a type, performing IO). The cost of forming the lambda and passing it in should also be small, and I'd expect it to be comparable to passing in a parse node and diagnostic.

toolchain/check/handle_pattern_binding.cpp
zygoloid: Sent #3304, but we can't do this yet. :(

PR #3303, opened 2023-10-17
Fix follow-on error for member access on an invalid expression.

zygoloid:
We were looking at the type of the base expression prior to conversion instead of the type after conversion.

As noted in review of #3302.


toolchain/check/handle_name.cpp
```diff
@@ -15,9 +15,9 @@ auto HandleMemberAccessExpression(Context& context, Parse::Node parse_node)
 
   auto base_id = context.node_stack().PopExpression();
 
-  auto base =
-      context.semantics_ir().GetNode(context.FollowNameReferences(base_id));
-  if (auto namespc = base.TryAs<SemIR::Namespace>()) {
+  if (auto namespc = context.semantics_ir()
```
jonmeow:
```suggestion
  if (auto namespace_node = context.semantics_ir()
```

Per style: "Do not abbreviate by deleting letters within a word." (but I do come from not being familiar with "namespc" as an abbreviation)

https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

jonmeow (approved):
Looks great!

PR #3304, opened 2023-10-17
Switch to building in C++20 mode.

zygoloid:
Remove a couple of workarounds for C++17 deficiencies.


jonmeow (approved):
Looks good, just trying to remember if there was anything else to update, I thought there was something...

zygoloid:
It looks like we'll need a minimum of Clang 16 to support lambda captures of structured bindings, and we aren't ready to require Clang 16 yet for other reasons. Closing for now, we can revisit this later if things change.

PR #3305, opened 2023-10-18
Support for class definitions with static member functions.

zygoloid:
This includes being able to define a class that was previously forward-declared, and being able to define a member function out-of-line that was previously declared inside a class.

No support for fields or methods yet, and a class definition doesn't yet cause the class to be treated as a complete type.


toolchain/check/handle_name.cpp
```diff
@@ -9,19 +9,43 @@
 
 namespace Carbon::Check {
 
+static auto GetAsNameScope(Context& context, SemIR::NodeId base_id)
```
jonmeow, line 14: Please add a brief comment to explain what this function is doing, e.g.:

```suggestion
// Returns the name scope corresponding to base_id, or nullopt if not a scope.
// On invalid scopes, prints a diagnostic and still returns the scope.
static auto GetAsNameScope(Context& context, SemIR::NodeId base_id)
```

toolchain/sem_ir/formatter.cpp
```diff
@@ -572,6 +582,26 @@ class Formatter {
     }
   }
 
+  auto FormatNameScope(NameScopeId id, llvm::StringRef separator,
+                       llvm::StringRef prefix) -> void {
+    // Name scopes aren't kept in any particular order. Sort the entries before
+    // we print them for stability and consistency.
+    std::vector<std::pair<NodeId, StringId>> entries;
```
```suggestion
    llvm::SmallVector<std::pair<NodeId, StringId>> entries;
```

Just suggesting for consistency, the rule of thumb to use SmallVector and not vector in toolchain code.

toolchain/check/testdata/class/basic.carbon
```diff
@@ -0,0 +1,63 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+
+class Class {
```
jonmeow, line 7: Should there be a test for something which requires delayed function definition processing? e.g.:

```
class Class {
  fn F() -> i32 {
    return G();
  }
  fn G() -> i32 {
    return F();
  }
}
```

(even if this test is currently a failure)

jonmeow (approved):
Looks great!

toolchain/sem_ir/formatter.cpp
zygoloid: (This is pre-existing, just moved from below.) Done.

toolchain/check/testdata/class/basic.carbon
zygoloid: Sure, added.

PR #3306, opened 2023-10-18
Directly convert from an initializer to a value where possible.

zygoloid:
If the initializing representation is the same as the value representation, don't materialize a temporary and perform a value binding. Instead, directly extract the value, using a new `value_of_initializer` node.

This removes a lot of redundant `alloca`s from our generated LLVM IR.


chandlerc (approved):
Really nice, love how this tidies up.

PR #3307, opened 2023-10-18
Add & update comments in `parse/node_kind.def`

josh11b:


toolchain/parse/node_kind.def
```diff
@@ -70,6 +70,30 @@ CARBON_PARSE_NODE_KIND_CHILD_COUNT(Name, 0,
...
 
+// ----------------------------------------------------------------------------
+
+// The comments below follow this pattern:
+
+// Descriptive heading:
+//   Child1
+//   Child2
+// Parent
+
```
jonmeow:
```suggestion
//
// // Descriptive heading:
// //   Child1
// //   Child2
// // Parent
//
```

I think using a contiguous block comment makes context clearer; the colon at the end of line 77 is a little confusing otherwise, matching the colon at the end of line 75.

toolchain/parse/node_kind.def
```diff
@@ -70,6 +70,30 @@ CARBON_PARSE_NODE_KIND_CHILD_COUNT(Name, 0,
...
+// In this, `Child1`, `Child2`, and `Parent` are all kinds of parse nodes, which
+// are then defined using the `CARBON_PARSE_NODE_KIND_*` macros. They are
+// written in the order they are emitted, and the indentation showing the tree
+// structure. Note that nodes are omitted in postorder (see
```
jonmeow:
```suggestion
// structure. Note that nodes are emitted in postorder (see
```

toolchain/parse/node_kind.def
```diff
@@ -70,6 +70,30 @@ CARBON_PARSE_NODE_KIND_CHILD_COUNT(Name, 0,
...
+// https://en.wikipedia.org/wiki/Tree_traversal#Depth-first_search), with the
```
jonmeow: What do you think of removing this link? I think postorder traversals are a common term, and I think adding links everywhere it's mentioned stands to become noisy.

That said, this code isn't handling the emission, so you might consider pointing to a more authoritative source for how the code works, such as:

```
They are written in postorder, with the indentation showing the tree structure. See tree.h for more information.
```

Note too, the tree supports both postorder and preorder traversal. While the default is to emit in postorder, you can pass `--preorder-parse-tree` and get preorder.

toolchain/parse/node_kind.def
```diff
@@ -226,10 +246,8 @@ CARBON_PARSE_NODE_KIND_CHILD_COUNT(ContinueStatement, 1,
 
 // `return`:
 //   ReturnStatementStart
-//   _external_: expression
+//   _optional_ _external_: expression
```
jonmeow, line 247: The way this reads feels a little odd to me, maybe remove the underscores instead?

jonmeow (approved):
Approving, though you might consider rephrasing the comments, per notes.

toolchain/parse/node_kind.def
```diff
@@ -70,6 +70,30 @@ CARBON_PARSE_NODE_KIND_CHILD_COUNT(Name, 0,
...
+// The comments below follow this pattern:
+
+// Descriptive heading:
+//   Child1
+//   Child2
+// Parent
+
+// In this, `Child1`, `Child2`, and `Parent` are all kinds of parse nodes, which
+// are then defined using the `CARBON_PARSE_NODE_KIND_*` macros. They are
+// written in the order they are emitted, and the indentation showing the tree
```
jonmeow:
```suggestion
// written in the order they are emitted, with indentation showing the tree
```

"and the indentation shows" or "with the indentation showing" would both work, but I think "and the indentation showing" is grammatically incorrect.

toolchain/parse/node_kind.def
josh11b: Used your text.

toolchain/parse/node_kind.def
josh11b: It was helpful for me to have something to set this apart from normal parse-node-kind names.

jonmeow: SGTM, maybe something for me to think about but I started this with `_external_`

PR #3309, opened 2023-10-18
SEMANTICS->SEM_IR in macro names. Update comments in `sem_ir/node.h`.

josh11b:
* Shift to "sem_ir" continued from #3176.
* Comments in `node.h` changed to reflect #3280.


jonmeow (approved):
Thanks!

PR #3310, opened 2023-10-18
Remove most of the metaprogramming in node.h in favor of listing all the members in the typed node structs.

zygoloid:
Split `node.h` into separate files for ID types (`id.h`) and for typed nodes (`typed_nodes.h`). The per-node-kind data is now specified as part of declaring the typed nodes, and is removed from the node kinds x-macros, which now simply enumerate the node kinds.


toolchain/sem_ir/node_kind.def
```diff
@@ -7,101 +7,64 @@
...
-#define CARBON_SEM_IR_NODE_KIND_IMPL(Name, IRName, ValueKind, TerminatorKind) \
-  CARBON_SEM_IR_NODE_KIND_WITH_IR_NAME(Name, IRName)
-#else
+#ifndef CARBON_SEM_IR_NODE_KIND
 #error "Must define the x-macro to use this file."
 #endif
 
 // A cross-reference between IRs.
-CARBON_SEM_IR_NODE_KIND_IMPL(CrossReference, "xref", Typed, NotTerminator)
+CARBON_SEM_IR_NODE_KIND(CrossReference)
```
jonmeow: Maybe merge this into the main list? It doesn't feel special anymore, I don't think it needed to be first.

toolchain/sem_ir/node_kind.def
```diff
@@ -7,101 +7,64 @@
...
-CARBON_SEM_IR_NODE_KIND_IMPL(TupleInit, "tuple_init", Typed, NotTerminator)
-CARBON_SEM_IR_NODE_KIND_IMPL(TupleLiteral, "tuple_literal", Typed,
-                             NotTerminator)
-CARBON_SEM_IR_NODE_KIND_IMPL(TupleType, "tuple_type", Typed, NotTerminator)
-CARBON_SEM_IR_NODE_KIND_IMPL(TupleValue, "tuple_value", Typed, NotTerminator)
-CARBON_SEM_IR_NODE_KIND_IMPL(UnaryOperatorNot, "not", Typed, NotTerminator)
-CARBON_SEM_IR_NODE_KIND_IMPL(ValueAsReference, "value_as_reference", Typed,
-                             NotTerminator)
-CARBON_SEM_IR_NODE_KIND_IMPL(VarStorage, "var", Typed, NotTerminator)
+CARBON_SEM_IR_NODE_KIND(AddressOf)
```
jonmeow, line 20:
```suggestion
// These have a 1:1 correspondence to entries in typed_nodes.h.
CARBON_SEM_IR_NODE_KIND(AddressOf)
```

Maybe add a comment in this file noting where to look for the rest?

toolchain/sem_ir/node_kind.h
```diff
@@ -72,6 +84,36 @@ class NodeKind : public CARBON_ENUM_BASE(NodeKind) {
 // We expect the node kind to fit compactly into 8 bits.
 static_assert(sizeof(NodeKind) == 1, "Kind objects include padding!");
 
+class NodeKind::Definition : public NodeKind {
```
jonmeow, line 92: Should probably have a comment explaining how this class is used, noting @josh11b's questions about that.

toolchain/sem_ir/typed_nodes.h
```diff
@@ -0,0 +1,482 @@
...
+template <typename T, typename ParseNodeType = Parse::Node T::*>
+constexpr bool HasParseNode = false;
+template <typename T>
+constexpr bool HasParseNode<T, decltype(&T::parse_node)> = true;
+
+// HasTypeId<T> is true if T has a `TypeId type_id` field.
+template <typename T, typename TypeIdType = TypeId T::*>
+constexpr bool HasTypeId = false;
+template <typename T>
+constexpr bool HasTypeId<T, decltype(&T::type_id)> = true;
```
jonmeow, line 487: It might be worth moving these to the top, since otherwise they might be easily missed beneath all the actual typed nodes. I can understand though if you'd rather not; it only really makes sense if we don't have too much extra code like this.

toolchain/sem_ir/typed_nodes.h
```diff
@@ -0,0 +1,482 @@
...
+//   for which a placeholder type should be used.
+// - Up to two `[...]Id` members describing the contents of the struct.
+//
+// The field names here matter -- the first two fields must have the names
+// specified above. When converting to a `SemIR::Node`, they will become the
+// parse node and type associated with the type-erased node.
+namespace Carbon::SemIR {
+
+struct AddressOf {
+  static constexpr auto Kind = NodeKind::AddressOf.Define("address_of");
```
josh11b, line 40: Please have a test that validates that you picked up the right enum value here.

toolchain/sem_ir/typed_nodes.h
```diff
@@ -0,0 +1,482 @@
...
+
+  Parse::Node parse_node;
+  TypeId type_id;
```
josh11b, line 43: Please include a test that these two fields, when present, are present first and in the right order.

toolchain/sem_ir/typed_nodes.h
```diff
@@ -0,0 +1,482 @@
...
+  NodeId lvalue_id;
+};
```
josh11b, line 45: Please have a test that each typed node can be converted from a Node and back.

toolchain/sem_ir/typed_nodes.h
josh11b, line 487: Document this technique in the toolchain doc.

toolchain/sem_ir/typed_nodes.h
```diff
@@ -0,0 +1,482 @@
...
+
+#ifndef CARBON_TOOLCHAIN_SEM_IR_TYPED_NODES_H_
+#define CARBON_TOOLCHAIN_SEM_IR_TYPED_NODES_H_
+
+#include "toolchain/parse/tree.h"
+#include "toolchain/sem_ir/builtin_kind.h"
+#include "toolchain/sem_ir/ids.h"
+#include "toolchain/sem_ir/node_kind.h"
+
+// Representations for specific kinds of nodes.
```
josh11b, line 13: Document the static `Kind` member as well. It would be good to include a summary of what happens when you write that line. This includes associating the enum value of NodeKind::<Name> with this struct type, the ir_name string constant, and the terminator kind. Perhaps it would be good to have some breadcrumbs here that this file is included by node_kind.cpp, which uses this data to populate its tables?

toolchain/sem_ir/typed_nodes.h
```diff
@@ -0,0 +1,482 @@
...
+//   associated location. Almost all nodes should have this, with exceptions
+//   being things that are generated internally, without any relation to source
+//   syntax, such as predeclared builtins.
+// - Optionally, a `TypeId type_id;` member, for nodes that produce a value.
+//   This includes nodes that produce an abstract value, such as a `Namespace`,
+//   for which a placeholder type should be used.
+// - Up to two `[...]Id` members describing the contents of the struct.
+//
+// The field names here matter -- the first two fields must have the names
+// specified above. When converting to a `SemIR::Node`, they will become the
```
josh11b:
```suggestion
// specified above when present. When converting to a `SemIR::Node`, they will become the
```

jonmeow (commented):
A few minor comments, but generally I think this seems fine. I know @josh11b had mentioned some tests which would help. If it sticks like this I'm good with the look-and-feel, I do think it's an improvement. You don't need to wait for my approval -- I'm only really not approving because I'm sure others want to review.

toolchain/sem_ir/formatter.cpp
```diff
@@ -665,7 +665,17 @@ class Formatter {
   template <typename NodeT>
   auto FormatInstructionRHS(NodeT node) -> void {
     // By default, an instruction has a comma-separated argument list.
-    std::apply([&](auto... args) { FormatArgs(args...); }, node.args_tuple());
+    using Info = TypedNodeArgsInfo<NodeT>;
+    if constexpr (Info::NumArgs == 2) {
+      auto tuple = StructReflection::AsTuple(node);
+      FormatArgs(std::get<Info::Arg0Field>(tuple),
+                 std::get<Info::Arg0Field + 1>(tuple));
```
josh11b: Would it be worth defining `Info::Arg1Field`?

toolchain/sem_ir/formatter.cpp
```diff
@@ -665,7 +665,17 @@ class Formatter {
   template <typename NodeT>
   auto FormatInstructionRHS(NodeT node) -> void {
     // By default, an instruction has a comma-separated argument list.
-    std::apply([&](auto... args) { FormatArgs(args...); }, node.args_tuple());
+    using Info = TypedNodeArgsInfo<NodeT>;
```
josh11b, line 670: Perhaps we can put more functionality into `TypedNodeArgsInfo` so this code doesn't need to deal with `StructReflection` and tuples? Could we make it so you write `Info::GetField0(node)` instead of
```
auto tuple = StructReflection::AsTuple(node);
std::get<Info::Arg0Field>(tuple)
```

toolchain/sem_ir/node.h
```diff
@@ -14,542 +14,118 @@
...
+// the specific kinds of node defined in `typed_nodes.h`. This provides access
+// to common fields present on most or all kinds of nodes:
 //
 // - `parse_node` for error placement.
 // - `kind` for run-time logic when the input Kind is unknown.
 // - `type_id` for quick type checking.
-// - Up to two Kind-specific members.
 //
-// To create a specific kind of `Node`, use the appropriate `TypedNode`
-// constructor. A `TypedNode` implicitly converts to a `Node`.
```
josh11b, line 482: Perhaps revive this comment with a pointer to `typed_nodes.h`?

toolchain/sem_ir/node.h
```diff
@@ -14,542 +14,118 @@
...
+// In addition, any kind-specific data is stored and can be accessed by casting
```
josh11b:
```suggestion
// In addition, any kind-specific data can be accessed by casting
```

toolchain/sem_ir/node.h
```diff
@@ -14,542 +14,118 @@
...
+    }
+    if constexpr (HasTypeId<TypedNode>) {
+      type_id_ = typed_node.type_id;
+    }
+    if constexpr (Info::NumArgs > 0) {
+      auto tuple = StructReflection::AsTuple(typed_node);
+      arg0_ = ToRaw(std::get<Info::Arg0Field>(tuple));
+      if constexpr (Info::NumArgs > 1) {
+        arg1_ = ToRaw(std::get<Info::Arg0Field + 1>(tuple));
+      }
```
josh11b: Additional `Info` methods for getting the fields would be useful here too.

toolchain/sem_ir/node.h
```diff
@@ -588,269 +171,13 @@ class Node : public Printable<Node> {
...
-
-// Convert a field from its raw representation.
-template <typename T>
-constexpr auto FromRaw(int32_t raw) -> T {
-  return T(raw);
-}
-template <>
-constexpr auto FromRaw<BuiltinKind>(int32_t raw) -> BuiltinKind {
-  return BuiltinKind::FromInt(raw);
+// Typed nodes can be printed by converting them to nodes.
```
josh11b, line 189: Is the indentation weird here?

toolchain/sem_ir/typed_nodes.h
zygoloid: These to me feel like implementation details that most readers of this file won't be interested in, so I think I'd prefer to keep them here, or maybe move them to a different file?

toolchain/sem_ir/typed_nodes.h
zygoloid: Done.

toolchain/sem_ir/typed_nodes.h
zygoloid: Done.

toolchain/sem_ir/typed_nodes.h
zygoloid: Done.

toolchain/sem_ir/typed_nodes.h
zygoloid: Done.

toolchain/sem_ir/typed_nodes.h
zygoloid: Done.

toolchain/sem_ir/formatter.cpp
zygoloid: Sure, done.

toolchain/sem_ir/formatter.cpp
zygoloid: It feels a bit wasteful to me to convert the struct to a tuple once for each field you read, but I suppose that'll all get optimized away. Done.

toolchain/sem_ir/node.h
zygoloid: The class comment already starts with: "that may be constructed from the specific kinds of node defined in `typed_nodes.h`", do you want more than that?

toolchain/sem_ir/node.h
zygoloid: Done.

toolchain/sem_ir/node.h
zygoloid: Done.

toolchain/sem_ir/node.h
zygoloid: It looks OK to me, but github seems to not deal very well with the large deletions here in the side-by-side view.

toolchain/sem_ir/node_kind.def
zygoloid: Done.

toolchain/sem_ir/node_kind.def
zygoloid: Done.

toolchain/sem_ir/node_kind.h
zygoloid: Done.

toolchain/sem_ir/typed_nodes.h
zygoloid: Done.

josh11b: FYI, I agree this feels secondary to the typed nodes themselves. I'm not sure if I would consider it an implementation detail, because they are named by consumers of the header, but I would be fine with them at the end or in a different file. If it was in a different file, that file would naturally also contain `TypedNodeArgsInfo`.

toolchain/sem_ir/node_kind.h
```diff
@@ -84,6 +84,11 @@ class NodeKind : public CARBON_ENUM_BASE(NodeKind) {
 // We expect the node kind to fit compactly into 8 bits.
 static_assert(sizeof(NodeKind) == 1, "Kind objects include padding!");
 
+// A definition of a node kind, including ancillary data such as the name to use
```
josh11b: How about we make this into a sentence?

```suggestion
// A definition of a node kind is a value of the NodeKind itself plus ancillary data such as the name to use
```

toolchain/sem_ir/typed_nodes.h
```diff
@@ -24,8 +24,15 @@
 // - Up to two `[...]Id` members describing the contents of the struct.
 //
 // The field names here matter -- the first two fields must have the names
-// specified above. When converting to a `SemIR::Node`, they will become the
-// parse node and type associated with the type-erased node.
+// specified above, when present. When converting to a `SemIR::Node`, they will
+// become the parse node and type associated with the type-erased node.
+//
+// In addition, each type defines a constant `Kind` that describes the node
+// kind for that node, including the name used in textual IR, and any other
```
josh11b: and something about how it associates this type with a particular value of the `NodeKind` enumerant?

toolchain/sem_ir/node_kind.h
zygoloid: Rephrased to include the "is a value of NodeKind" part.

(I found the suggested change as-is to make the comment harder for me to read -- generally I expect the first sentence in a class comment to be a noun phrase that could be preceded by "This class is...", in the same way that the first sentence in a class comment is missing its preceding "This function ...".)

toolchain/sem_ir/typed_nodes.h
zygoloid: Done.

josh11b (approved):

PR #3311, opened 2023-10-18
Add value store to be shared across compile stages.

jonmeow:
This updates lexing to use the data. I'll do checking separately, just to split changes.

Note the ValueStore structure is also set up such that SemIR::File can use it for other fields.


toolchain/base/value_store.h
```diff
@@ -0,0 +1,138 @@
...
+    return values_[id.index];
+  }
+
+ private:
+  llvm::SmallVector<ValueT> values_;
+};
+
+// Storage for StringRefs. The caller is responsible for ensuring storage is
+// allocated.
+class StringStore {
```
zygoloid: Have you considered making this a specialization of `ValueStore`?
```suggestion
template <typename IdT>
class ValueStore<std::string> {
```
I'm not sure one way or the other -- in particular, using `std::string` as the `ValueT` here is a bit odd because the strings aren't actually stored that way.

toolchain/base/value_store.h
```diff
@@ -0,0 +1,138 @@
...
+    CARBON_CHECK(id.is_valid());
+    return values_[id.index];
+  }
+
+ private:
+  llvm::SmallVector<ValueT> values_;
+};
+
+// Storage for StringRefs. The caller is responsible for ensuring storage is
+// allocated.
```
zygoloid, line 100: This comment doesn't seem to match the code -- a `StringMap` owns storage for its strings.

toolchain/base/value_store.h
```diff
@@ -0,0 +1,138 @@
...
+  static const IntegerId Invalid;
+  using IndexBase::IndexBase;
+  auto Print(llvm::raw_ostream& out) const -> void {
+    out << "int";
+    IndexBase::Print(out);
+  }
+};
+constexpr IntegerId IntegerId::Invalid(IntegerId::InvalidIndex);
+
+// Corresponds to a RealValue.
```
zygoloid:
```suggestion
// Corresponds to a Real value.
```

toolchain/base/value_store.h
```diff
@@ -0,0 +1,138 @@
...
+  llvm::SmallVector<llvm::StringRef> values_;
+};
+
+// The value of a real literal.
+//
+// This is either a dyadic fraction (mantissa * 2^exponent) or a decadic
+// fraction (mantissa * 10^exponent).
+//
+// `RealLiteralValue` carries a reference back to `TokenizedBuffer` which can be
+// invalidated if the buffer is edited or destroyed.
```
zygoloid:
```suggestion
```

toolchain/lex/testdata/multifile.carbon
```diff
@@ -18,9 +18,12 @@ a;
 // CHECK:STDOUT: - filename: b.carbon
 // CHECK:STDOUT:   tokens: [
 // CHECK:STDOUT:     { index: 0, kind: 'StartOfFile', line: {{ *\d+}}, column:  1, indent: 1, spelling: '', has_trailing_space: true },
-b;
-// CHECK:STDOUT:     { index: 1, kind:  'Identifier', line: {{ *}}[[@LINE-1]], column:  1, indent: 1, spelling: 'b', identifier: 0 },
+a;
+// CHECK:STDOUT:     { index: 1, kind:  'Identifier', line: {{ *}}[[@LINE-1]], column:  1, indent: 1, spelling: 'a', identifier: 0 },
```
zygoloid: We'd get this same output regardless of whether we reuse the value store across files. Maybe it'd be better to test `b;` and then `a;`?

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -556,14 +557,14 @@ class [[clang::internal_linkage]] TokenizedBuffer::Lexer {
...
-      auto token =
-          buffer_.AddToken({.kind = TokenKind::StringLiteral,
-                            .token_line = string_line,
-                            .column = string_column,
-                            .literal_index = static_cast<int32_t>(
-                                buffer_.literal_string_storage_.size())});
-      buffer_.literal_string_storage_.push_back(
-          literal->ComputeValue(emitter_));
+      buffer_.computed_strings_.push_back(
+          std::make_unique<std::string>(literal->ComputeValue(emitter_)));
```
zygoloid, line 563: This allocation seems to be redundant.

toolchain/base/value_store.h
```diff
@@ -0,0 +1,138 @@
...
+  llvm::APInt exponent;
+
+  // If false, the value is mantissa * 2^exponent.
+  // If true, the value is mantissa * 10^exponent.
+  bool is_decimal;
+};
+
+// Stores that will be used across compiler steps. This is provided mainly so
+// that they don't need to be passed separately.
+class CompileValueStores {
```
zygoloid: I find the word "Compile" here isn't helping me at all. If you'd prefer to keep it, I won't object, but I'd prefer to remove it, or replace it with something like "Shared" or "Common".

toolchain/base/value_store.h
```diff
@@ -0,0 +1,138 @@
...
+
+  // The mantissa, represented as an unsigned integer.
+  llvm::APInt mantissa;
+
+  // The exponent, represented as a signed integer.
+  llvm::APInt exponent;
+
+  // If false, the value is mantissa * 2^exponent.
+  // If true, the value is mantissa * 10^exponent.
+  bool is_decimal;
```
zygoloid: There's a lot of space here lost to padding -- do we care about that?

zygoloid (commented):
I'm not sure it's going to be all that useful to share `Real`s across layers, because the representation we want from lexing (mantissa, exponent, `is_decimal`) is quite different from the representation we want in checking for literals (a rational number) and different again from the representation we want in checking for `fN` (an `APFloat`). That said, factoring out the storage for these seems fine, in case we do want to create more of the lexing representation during checking, and it seems like a good thing for now because we are sharing the representation currently.

toolchain/base/value_store.h
jonmeow: sizeof(APInt) == 16, so this is minimally a 32-byte structure, which is pretty big compared to other data. The is_decimal adds 8 bytes for me, which definitely isn't ideal, but versus the alternative of loading up source code to recalculate is_decimal, my instinct is that it's acceptable (SemIR's Real previously had is_decimal).

Also, I was thinking that if metaprogramming is allowed to generate reals, or edge cases are added to lexing, the is_decimal recalculation may not be an option. To your point about SemIR maybe wanting a different representation than this, if lexing continues to produce the same representation that SemIR wants (even if that representation changes), then it may be better to keep independent storage even if the cost is undesirable.

I suspect there's some approach like embedding the bool into one of the APInts, but I'm similarly hesitant to do that; it seems like it'd be difficult to understand. This might be something to consider again once when redesigning the type.

toolchain/base/value_store.h
jonmeow: I'm going with "Shared". "Common" overlaps with me mentally for the "common" versus "base" directory discussion.

I'm hesitant of ValueStore versus ValueStores, with a single character difference (although there's a template argument for ValueStore, but the naming might also be read by some as having stronger implications that ValueStore shouldn't be reused outside ValueStores).

toolchain/base/value_store.h
jonmeow: Switched to DenseMap. This was my intent and I think I copied too much from SemIR without thinking it through. Note though, to use StringMap's storage, the values_.push_back would need to use the StringMap's copy of value, not value directly.

That said, do you think we _should_ use StringMap, and have this be responsible for storing strings? The downside is that for builtin identifiers, we'd be duplicating storage. That may be mitigated because identifiers are expected to be reused. The upside is it's easier for generated strings, both for literals and metaprogramming.

I'm on the fence about what's best here, I was defaulting to this because it felt more minimal, but could switch.

toolchain/base/value_store.h
jonmeow: Sure, done with llvm::StringRef.

toolchain/lex/tokenized_buffer.cpp
jonmeow: Is this about the StringMap? Please see my comments there, I think this isn't redundant if we stop using StringMap.

toolchain/lex/testdata/multifile.carbon
jonmeow: Done

toolchain/base/value_store.h
zygoloid: @chandlerc has been doing some serious digging into how to most efficiently store our name lookup table in the lexer, so I'd defer to him on whether the map should own (copies of) the strings, but I believe the plan is for this to be some custom hash table that's neither a `DenseMap` nor a `StringMap`, as the remaining big piece of the lexer speedup work.

As a result, I'd expect whatever decision we make here to be a short-term one, so I'm happy either way.

toolchain/lex/tokenized_buffer.cpp
zygoloid: This was primarily about the `StringMap`, but even with the `DenseMap` this is performing an extra allocation for longer strings, and will overallocate for shorter strings that fit easily into the small-string buffer (which I expect to be a common case.) Maybe consider something like:
```suggestion
      std::string value = literal->ComputeValue(emitter_);
      auto storage = std::make_unique<char[]>(value.size());
      std::copy(value.begin(), value.end(), storage.get());
      buffer_.computed_strings_.push_back(std::move(storage));
```
... or just add a TODO so we don't forget about this in the longer term.

(Having said all of that, I think the most common case will be where we can reuse the source buffer to provide the contents of the string -- if it contains no escapes and (if it's multiline) has no indentation -- and we're not optimizing for that case yet, so perhaps that's the more important TODO here!)

zygoloid (approved):
Thanks. As noted, I'm happy with either approach to string storage for now, given I expect this to get revisited soon anyway.

toolchain/base/value_store.h
zygoloid: OK, sounds like leaving this alone for now is best. Maybe add a TODO to look into making this smaller? I agree that even 32 bytes seems like an unnecessarily large representation for the common case.

toolchain/base/value_store.h
jonmeow: That's a good point. I'm going to stick with `DenseMap`: my suspicion is that a `DenseMap` implementation will be easier to transition since it doesn't keep ownership of the strings, and a custom hash table I think was intended to be generic across types so would need special-casing similar to `StringMap` to handle that use-case.

toolchain/lex/tokenized_buffer.cpp
jonmeow: I'm going to make a TODO here, mainly to break this out. I thought though that this would be moved, without resulting in an extra allocation. My thought regardless is that we could have ComputeValue return the type we want though, rather than doing logic here (avoiding the `std::copy`), but I'd like to do that separately.

toolchain/base/value_store.h
jonmeow: Done

toolchain/base/value_store.h
chandlerc: (Sorry I missed this comment thread in all the churn)

FWIW, my plan is to make a general hash table that is fast for string-ish keys. The ownership I think should be orthogonal.

TBH, I'm doubtful about the utility of bundled ownership the way `StringMap` works in general, but I haven't studied it in depth. My suspicion is that this is mostly an optimization in the face of a very poor system allocator. That makes some sense, but I still suspect that something which simply wraps a bump pointer allocator and a map of `StringRef`s would be preferable. But when you *can* avoid taking ownership, it seems very likely beneficial to do so and to just have a map of `StringRef` or `string_view`.

PR #3312, opened 2023-10-19
Use the full name declaration logic when adding `var` names to scope.

zygoloid:
This causes the names of class members to get added to the class scope.

### DIFF
--- a/toolchain/check/declaration_name_stack.cpp
+++ b/toolchain/check/declaration_name_stack.cpp
@@ -8,11 +8,22 @@
 
 namespace Carbon::Check {
 
+auto DeclarationNameStack::NewNameContext() -> NameContext {
+  return NameContext{.state = NameContext::State::New,
+                     .target_scope_id = context_->current_scope_id(),
+                     .resolved_node_id = SemIR::NodeId::Invalid};
+}
+
+auto DeclarationNameStack::MakeUnqualifiedName(Parse::Node parse_node,
+                                               SemIR::StringId name_id)
+    -> NameContext {
+  NameContext context = NewNameContext();
+  ApplyNameQualifierTo(context, parse_node, name_id);
+  return context;
+}
+
 auto DeclarationNameStack::Push() -> void {
-  declaration_name_stack_.push_back(
-      {.state = NameContext::State::New,
-       .target_scope_id = context_->current_scope_id(),
-       .resolved_node_id = SemIR::NodeId::Invalid});
+  declaration_name_stack_.push_back(NewNameContext());
 }
 
 auto DeclarationNameStack::Pop() -> NameContext {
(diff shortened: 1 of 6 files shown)


toolchain/check/testdata/var/fail_shadowing.carbon
```diff
@@ -0,0 +1,48 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+
+
```
jonmeow:
```suggestion
```

Extra blank line?

toolchain/check/declaration_name_stack.cpp
```diff
@@ -8,11 +8,22 @@
 
 namespace Carbon::Check {
 
+auto DeclarationNameStack::NewNameContext() -> NameContext {
+  return NameContext{.state = NameContext::State::New,
+                     .target_scope_id = context_->current_scope_id(),
+                     .resolved_node_id = SemIR::NodeId::Invalid};
+}
```
jonmeow: What if you were to add a constructor to NameContext that took the scope as an argument, and just pass that rather than this function?

Otherwise, "New" makes me think `new` allocation; I think Carbon's design is heading towards "Make" so that might be fair, or so would "Build". Alternatively, maybe you're trying to match the enum name: rename it and this to "Start"? (State::Start, StartNameContext)

toolchain/check/declaration_name_stack.h
```diff
@@ -78,7 +78,7 @@ class DeclarationNameStack {
     // The scope which qualified names are added to. For unqualified names in
     // an unnamed scope, this will be Invalid to indicate the current scope
     // should be used.
-    SemIR::NameScopeId target_scope_id = SemIR::NameScopeId::Invalid;
+    SemIR::NameScopeId target_scope_id;
```
jonmeow, line 81: I think this works partly because NameContext is never default constructed right now, it'd all gone through Push(), but the defaults should probably be removed from other fields too.

jonmeow (approved):
Looks good, with some small readability comments.

toolchain/check/declaration_name_stack.cpp
zygoloid: How about `State::Empty` and `MakeEmptyNameContext`?

toolchain/check/declaration_name_stack.h
zygoloid: Instead changed `MakeEmptyNameContext` to make use of the default member initializers. If you'd prefer I remove them as suggested, I'm happy to; let me know.

toolchain/check/declaration_name_stack.cpp
jonmeow: Sure

toolchain/check/declaration_name_stack.h
jonmeow: LGTM

PR #3313, opened 2023-10-19
Extend SharedValueStores to SemIR

jonmeow:
Building on #3311, change SemIR to use the SharedValueStore. Since this removes hermeticity, raw output no longer prints ints, reals, and strings. TokenizedBuffer accessors are modified to return IDs because values are often passed through in semantics without needing to read them.

I would've put SharedValueStores on Context, except for the GetArrayBoundValue convenience method. I felt awkward removing that, so it's on File, at least for now. That's then used by the formatter and Lower too. The flipside of this is that TokenizedBuffer has a SharedValueStores only for printing, so maybe that's similar enough to what File is doing. 

This doesn't start shifting other SemIR members to ValueStore, but that seems like a next step.

### DIFF
--- a/toolchain/check/convert.cpp
+++ b/toolchain/check/convert.cpp
@@ -134,7 +134,7 @@ static auto MakeElemAccessNode(Context& context, Parse::Node parse_node,
     // special case.
     auto index_id = block.AddNode(SemIR::IntegerLiteral(
         parse_node, context.GetBuiltinType(SemIR::BuiltinKind::IntegerType),
-        context.semantics_ir().AddInteger(llvm::APInt(32, i))));
+        context.semantics_ir().integers().Add(llvm::APInt(32, i))));
     return block.AddNode(
         AccessNodeT(parse_node, elem_type_id, aggregate_id, index_id));
   } else {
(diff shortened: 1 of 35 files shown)


toolchain/check/convert.cpp
```diff
@@ -134,7 +134,7 @@ static auto MakeElemAccessNode(Context& context, Parse::Node parse_node,
     // special case.
     auto index_id = block.AddNode(SemIR::IntegerLiteral(
         parse_node, context.GetBuiltinType(SemIR::BuiltinKind::IntegerType),
-        context.semantics_ir().AddInteger(llvm::APInt(32, i))));
+        context.semantics_ir().integers().Add(llvm::APInt(32, i))));
```
zygoloid, line 137: I wonder whether it'd be a useful convenience to have `Context` also expose access to the value stores?

zygoloid (approved):

toolchain/check/convert.cpp
jonmeow: Let's discuss this once the other things (functions, classes, etc) are also switched over (coming soon). I think it might be a little odd to have one and not the other, but also might be a fair amount to add.

PR #3314, opened 2023-10-19
Change StringLiteral to less frequently allocate a new string.

jonmeow:
Building on #3311, which started moving the result string into a `unique_ptr`, instead have `StringLiteral` use a `BumpPtrAllocator` to manage memory. But also, detect when a string is really trivial during `Lex` and, if so, return `contents_` directly.

### DIFF
--- a/toolchain/lex/string_literal.cpp
+++ b/toolchain/lex/string_literal.cpp
@@ -134,9 +134,9 @@ auto StringLiteral::Lex(llvm::StringRef source_text)
         content_needs_validation = true;
         break;
       case '\\':
-        content_needs_validation = true;
         if (escape.size() == 1 ||
             source_text.substr(cursor + 1).startswith(escape.substr(1))) {
+          content_needs_validation = true;
           cursor += escape.size();
           // If there's either not a character following the escape, or it's a
           // single-line string and the escaped character is a newline, we
@@ -265,6 +265,16 @@ static auto AppendChar(char*& buffer_cursor, char append_char) -> void {
   ++buffer_cursor;
 }
 
+// Appends the front of contents to the buffer and advances the cursor.
+static auto AppendFrontOfContents(char*& buffer_cursor,
+                                  llvm::StringRef contents, size_t len_or_npos)
+    -> void {
+  auto len =
+      len_or_npos == llvm::StringRef::npos ? contents.size() : len_or_npos;
+  memcpy(buffer_cursor, contents.data(), len);
+  buffer_cursor += len;
+}
+
 // Expand an escape sequence, appending the expanded value to the given
 // `result` string. `content` is the string content, starting from the first
 // character after the escape sequence introducer (for example, the `n` in
@@ -389,15 +399,11 @@ static auto ExpandEscapeSequencesAndRemoveIndent(
         return c == '\n' || c == '\\' ||
                (IsHorizontalWhitespace(c) && c != ' ');
       });
-      int regular_text_len = end_of_regular_text == llvm::StringRef::npos
-                                 ? contents.size()
-                                 : end_of_regular_text;
-      memcpy(buffer_cursor, contents.data(), regular_text_len);
-      buffer_cursor += regular_text_len;
+      AppendFrontOfContents(buffer_cursor, contents, end_of_regular_text);
       if (end_of_regular_text == llvm::StringRef::npos) {
         return llvm::StringRef(buffer, buffer_cursor - buffer);
       }
-      contents = contents.drop_front(regular_text_len);
+      contents = contents.drop_front(end_of_regular_text);
 
       if (contents.consume_front("\n")) {
         // Trailing whitespace in the source before a newline doesn't contribute
@@ -431,10 +437,7 @@ static auto ExpandEscapeSequencesAndRemoveIndent(
               "escape sequence in a string literal.");
           emitter.Emit(contents.begin(), InvalidHorizontalWhitespaceInString);
           // Include the whitespace in the string contents for error recovery.
-          int len = after_space == llvm::StringRef::npos ? contents.size()
-                                                         : after_space;
-          memcpy(buffer_cursor, contents.data(), len);
-          buffer_cursor += len;
+          AppendFrontOfContents(buffer_cursor, contents, after_space);
         }
         contents = contents.substr(after_space);
         continue;
@@ -478,6 +481,9 @@ auto StringLiteral::ComputeValue(llvm::BumpPtrAllocator& allocator,
     return content_;
   }
 
+  // "Expanding" escape sequences should only ever shorten content. As a
+  // consequence, the output string should allows fit within this allocation.
+  // Although this may waste some space, it avoids a reallocation.
   auto result = ExpandEscapeSequencesAndRemoveIndent(
       emitter, content_, hash_level_, indent,
       allocator.Allocate<char>(content_.size()));
--- a/toolchain/lex/string_literal.cpp
+++ b/toolchain/lex/string_literal.cpp
@@ -114,11 +114,13 @@ auto StringLiteral::Lex(llvm::StringRef source_text)
   terminator.resize(terminator.size() + hash_level, '#');
   escape.resize(escape.size() + hash_level, '#');
 
+  bool content_needs_validation = false;
+
   // TODO: Detect indent / dedent for multi-line string literals in order to
   // stop parsing on dedent before a terminator is found.
   for (; cursor < source_text_size; ++cursor) {
     // Use a lookup table to allow us to quickly skip uninteresting characters.
-    static constexpr CharSet InterestingChars = {'\\', '\n', '"', '\''};
+    static constexpr CharSet InterestingChars = {'\\', '\n', '"', '\'', '\t'};
     if (!InterestingChars[source_text[cursor]]) {
       continue;
     }
@@ -127,7 +129,12 @@ auto StringLiteral::Lex(llvm::StringRef source_text)
     // escape sequences starting with a predictable character and not containing
     // embedded and unescaped terminators or newlines.
     switch (source_text[cursor]) {
+      case '\t':
+        // Tabs have extra validation.
+        content_needs_validation = true;
+        break;
       case '\\':
+        content_needs_validation = true;
         if (escape.size() == 1 ||
             source_text.substr(cursor + 1).startswith(escape.substr(1))) {
           cursor += escape.size();
@@ -137,7 +144,8 @@ auto StringLiteral::Lex(llvm::StringRef source_text)
           if (cursor >= source_text_size || (introducer->kind == NotMultiLine &&
                                              source_text[cursor] == '\n')) {
             llvm::StringRef text = source_text.take_front(cursor);
-            return StringLiteral(text, text.drop_front(prefix_len), hash_level,
+            return StringLiteral(text, text.drop_front(prefix_len),
+                                 content_needs_validation, hash_level,
                                  introducer->kind,
                                  /*is_terminated=*/false);
           }
@@ -146,7 +154,8 @@ auto StringLiteral::Lex(llvm::StringRef source_text)
       case '\n':
         if (introducer->kind == NotMultiLine) {
           llvm::StringRef text = source_text.take_front(cursor);
-          return StringLiteral(text, text.drop_front(prefix_len), hash_level,
+          return StringLiteral(text, text.drop_front(prefix_len),
+                               content_needs_validation, hash_level,
                                introducer->kind,
                                /*is_terminated=*/false);
         }
@@ -158,7 +167,8 @@ auto StringLiteral::Lex(llvm::StringRef source_text)
               source_text.substr(0, cursor + terminator.size());
           llvm::StringRef content =
               source_text.substr(prefix_len, cursor - prefix_len);
-          return StringLiteral(text, content, hash_level, introducer->kind,
+          return StringLiteral(text, content, content_needs_validation,
+                               hash_level, introducer->kind,
                                /*is_terminated=*/true);
         }
         break;
@@ -169,7 +179,7 @@ auto StringLiteral::Lex(llvm::StringRef source_text)
   }
   // No terminator was found.
   return StringLiteral(source_text, source_text.drop_front(prefix_len),
-                       hash_level, introducer->kind,
+                       content_needs_validation, hash_level, introducer->kind,
                        /*is_terminated=*/false);
 }
 
@@ -341,9 +351,9 @@ static auto ExpandAndConsumeEscapeSequence(LexerDiagnosticEmitter& emitter,
 // Expand any escape sequences in the given string literal.
 static auto ExpandEscapeSequencesAndRemoveIndent(
     LexerDiagnosticEmitter& emitter, llvm::StringRef contents, int hash_level,
-    llvm::StringRef indent) -> std::string {
-  std::string result;
-  result.reserve(contents.size());
+    llvm::StringRef indent) -> StringLiteral::ComputedValue {
+  auto result = std::make_unique<std::string>();
+  result->reserve(contents.size());
 
   llvm::SmallString<16> escape("\\");
   escape.resize(1 + hash_level, '#');
@@ -376,22 +386,23 @@ static auto ExpandEscapeSequencesAndRemoveIndent(
         return c == '\n' || c == '\\' ||
                (IsHorizontalWhitespace(c) && c != ' ');
       });
-      result += contents.substr(0, end_of_regular_text);
+      *result += contents.substr(0, end_of_regular_text);
       contents = contents.substr(end_of_regular_text);
 
       if (contents.empty()) {
-        return result;
+        return {*result, std::move(result)};
       }
 
       if (contents.consume_front("\n")) {
         // Trailing whitespace in the source before a newline doesn't contribute
         // to the string literal value. However, escaped whitespace (like `\t`)
         // and any whitespace just before that does contribute.
-        while (!result.empty() && result.back() != '\n' &&
-               IsSpace(result.back()) && result.length() > last_escape_length) {
-          result.pop_back();
+        while (!result->empty() && result->back() != '\n' &&
+               IsSpace(result->back()) &&
+               result->length() > last_escape_length) {
+          result->pop_back();
         }
-        result += '\n';
+        *result += '\n';
         // Move onto to the next line.
         break;
       }
@@ -412,7 +423,7 @@ static auto ExpandEscapeSequencesAndRemoveIndent(
               "escape sequence in a string literal.");
           emitter.Emit(contents.begin(), InvalidHorizontalWhitespaceInString);
           // Include the whitespace in the string contents for error recovery.
-          result += contents.substr(0, after_space);
+          *result += contents.substr(0, after_space);
         }
         contents = contents.substr(after_space);
         continue;
@@ -420,7 +431,7 @@ static auto ExpandEscapeSequencesAndRemoveIndent(
 
       if (!contents.consume_front(escape)) {
         // This is not an escape sequence, just a raw `\`.
-        result += contents.front();
+        *result += contents.front();
         contents = contents.drop_front(1);
         continue;
       }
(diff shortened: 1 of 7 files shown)


chandlerc:
FYI, I ran this on an AMD server to collect some performance stats:
```
name                                           old cpu/op   new cpu/op   delta
BM_ValidString_Simple                          42.2µs ± 5%  42.2µs ± 2%     ~     (p=0.223 n=16+20)
BM_ValidString_Multiline                       41.7µs ± 4%  42.2µs ± 3%   +1.16%  (p=0.002 n=17+16)
BM_ValidString_MultilineDoubleQuote            41.3µs ± 2%  45.2µs ±17%   +9.39%  (p=0.000 n=17+19)
BM_ValidString_Raw                             42.2µs ± 1%  41.6µs ± 2%   -1.41%  (p=0.002 n=18+20)
BM_IncompleteWithEscapes_Simple                82.0µs ± 6%  71.1µs ±15%  -13.32%  (p=0.000 n=18+20)
BM_IncompleteWithEscapes_Multiline             93.1µs ± 8%  70.9µs ±13%  -23.81%  (p=0.000 n=17+20)
BM_IncompleteWithEscapes_MultilineDoubleQuote  93.7µs ± 3%  68.9µs ±12%  -26.40%  (p=0.000 n=18+20)
BM_IncompleteWithEscapes_Raw                    110µs ± 5%   110µs ± 4%     ~     (p=0.506 n=19+19)
BM_SimpleStringValue_Simple                    92.4µs ± 1%  41.9µs ± 1%  -54.70%  (p=0.000 n=19+19)
BM_SimpleStringValue_Multiline                 92.3µs ± 1%  41.2µs ± 2%  -55.33%  (p=0.000 n=18+16)
BM_SimpleStringValue_MultilineDoubleQuote      92.5µs ± 1%  41.0µs ± 1%  -55.70%  (p=0.000 n=18+20)
BM_SimpleStringValue_Raw                       92.4µs ± 1%  41.2µs ± 2%  -55.44%  (p=0.000 n=19+18)
```

jonmeow:
Looking at the benchmark, ComputeValue is only used with SimpleString. With the others, I think it's fair to say that this doesn't cause significant slowdown for Lex. But since it makes the other cases more boring, I've added NoGeneratedString that preserves the aforementioned before, and added an escape to other SimpleString so that they're forced to generate a string (restoring the slower behavior). i.e.:

```
----------------------------------------------------------------------------------------
Benchmark                                              Time             CPU   Iterations
----------------------------------------------------------------------------------------
BM_ValidString_Simple                              62313 ns        62312 ns        11265
BM_ValidString_Multiline                           62515 ns        62514 ns        11226
BM_ValidString_MultilineDoubleQuote                62174 ns        62173 ns        11162
BM_ValidString_Raw                                 32345 ns        32344 ns        21476
BM_IncompleteWithEscapes_Simple                    62402 ns        62386 ns        10944
BM_IncompleteWithEscapes_Multiline                 55096 ns        55089 ns        12799
BM_IncompleteWithEscapes_MultilineDoubleQuote      54893 ns        54885 ns        12838
BM_IncompleteWithEscapes_Raw                      105418 ns       105399 ns         6627
BM_SimpleStringValue_NoGeneratedString             62425 ns        62416 ns        11276
BM_SimpleStringValue_Simple                       126765 ns       126745 ns         5520
BM_SimpleStringValue_Multiline                    127194 ns       127175 ns         5542
BM_SimpleStringValue_MultilineDoubleQuote         126454 ns       126424 ns         5601
BM_SimpleStringValue_Raw                           96614 ns        96602 ns         7227
```

toolchain/lex/string_literal.cpp
```diff
@@ -341,9 +351,9 @@ static auto ExpandAndConsumeEscapeSequence(LexerDiagnosticEmitter& emitter,
 // Expand any escape sequences in the given string literal.
 static auto ExpandEscapeSequencesAndRemoveIndent(
     LexerDiagnosticEmitter& emitter, llvm::StringRef contents, int hash_level,
-    llvm::StringRef indent) -> std::string {
-  std::string result;
-  result.reserve(contents.size());
+    llvm::StringRef indent) -> StringLiteral::ComputedValue {
+  auto result = std::make_unique<std::string>();
```
zygoloid: `unique_ptr<string>` is a space-inefficient representation for immutable string data; for large strings, this requires two heap representations rather than one, and for small strings that fit comfortably in the small string buffer, it will overallocate.

If we want to guarantee to put the data into a heap allocation (so we can form stable pointers into it), perhaps we could use a `llvm::SmallString` here, and copy into a `std::unique_ptr<char[]>` when producing the result.

jonmeow: How much of an issue is space efficiency for strings? My rough understanding for Clang had been that most of the memory consumption was due to the AST, rather than string literals. I'm hesitant to favor additional string copies over the space efficiency, but are string literals actually a big space consumer? Put differently, while I think the heap allocation occurs either way, moving it from `std::string` (or `llvm::SmallString` if you prefer) to `char[]` removes the overhead of the string type but adds the cost of a copy.

I wouldn't expect cache density to be affected unless we get into a more esoteric storage, such as storing strings sequentially in a single vector (which, could get into more copy overhead, though could avoid most of it by preallocating a buffer the size of the source file). Which we could do: and it actually wouldn't be that bad because it could take the buffer as input, and the StringRef could just point at subsections of a bigger SmallString. Which, hmmmmm. Is that tempting to you?

Possibly relevant note, chandlerc had said he's not focused on optimizing string lexing because they're too rare in code (about 1% of lines IIRC), and thus improvements wouldn't be significant to his lexer performance goals. That said, that works both ways: slightly costlier lexing is more okay because they're rare, if memory efficiency is important.

zygoloid: My suggestion would reduce the number of heap allocations in most cases (because in the case where the string fits in the `SmallString` the only allocation would be the final one), and remove a level of indirection, in addition to reducing the wasted space. Maybe it's not a huge deal, but it seems needless and wasteful to store a pointer to a pointer to an overallocated buffer here.

jonmeow: I'd like to discuss this with you a little. To be sure I'm understanding correctly:

- SmallString will not heap allocate for strings under the chosen size, likely 64 bytes.
- The current logic only runs for "complex" strings that are either multi-line or have escapes.
- Thus, one heap allocation would be avoided for short strings which are also complex.

In turn, string copies would always occur, but it sounds like you're saying it's not worth considering as part of trade-offs. Maybe we can discuss this further?

jonmeow: Per discussion, here's an implementation using BumpPtrAllocator and SmallString.

I also worked on the benchmarks, trying to break them down a little further to see the performance impact of the different approach.

BumpPtr and SmallString (memcpy):

```
--------------------------------------------------------------------------------------------
Benchmark                                                  Time             CPU   Iterations
--------------------------------------------------------------------------------------------
BM_ValidString_Simple                                  31218 ns        31217 ns        22792
BM_ValidString_Multiline                               31077 ns        31076 ns        21898
BM_ValidString_MultilineDoubleQuote                    30713 ns        30708 ns        22088
BM_ValidString_Raw                                     30667 ns        30667 ns        22109
BM_IncompleteWithEscapes_Simple                        45669 ns        45668 ns        15482
BM_IncompleteWithEscapes_Multiline                     37472 ns        37469 ns        17954
BM_IncompleteWithEscapes_MultilineDoubleQuote          37836 ns        37835 ns        18471
BM_IncompleteWithEscapes_Raw                           92009 ns        91993 ns         7389
BM_ComputeValue_NoGenerate_Short                        20.2 ns         20.2 ns     34602757
BM_ComputeValue_NoGenerate_Long                         3055 ns         3054 ns       226271
BM_ComputeValue_WillGenerate_Short                      53.5 ns         53.5 ns     13017676
BM_ComputeValue_WillGenerate_Long                      15560 ns        15559 ns        44032
BM_ComputeValue_WillGenerate_Multiline                 15226 ns        15226 ns        43788
BM_ComputeValue_WillGenerate_MultilineDoubleQuote      15406 ns        15406 ns        45770
BM_ComputeValue_WillGenerate_Raw                       15400 ns        15400 ns        46383
```

Versus unique_ptr and SmallString (no memcpy):

```
--------------------------------------------------------------------------------------------
Benchmark                                                  Time             CPU   Iterations
--------------------------------------------------------------------------------------------
BM_ValidString_Simple                                  30177 ns        30176 ns        23630
BM_ValidString_Multiline                               30897 ns        30897 ns        22199
BM_ValidString_MultilineDoubleQuote                    30905 ns        30903 ns        22953
BM_ValidString_Raw                                     30965 ns        30962 ns        22308
BM_IncompleteWithEscapes_Simple                        45130 ns        45122 ns        15771
BM_IncompleteWithEscapes_Multiline                     40566 ns        40563 ns        17345
BM_IncompleteWithEscapes_MultilineDoubleQuote          41453 ns        41445 ns        17584
BM_IncompleteWithEscapes_Raw                           96260 ns        96250 ns         6956
BM_ComputeValue_NoGenerate_Short                        21.1 ns         21.1 ns     34254556
BM_ComputeValue_NoGenerate_Long                         3149 ns         3149 ns       225301
BM_ComputeValue_WillGenerate_Short                      50.0 ns         50.0 ns     14118668
BM_ComputeValue_WillGenerate_Long                       9304 ns         9304 ns        75569
BM_ComputeValue_WillGenerate_Multiline                  9272 ns         9270 ns        75587
BM_ComputeValue_WillGenerate_MultilineDoubleQuote       9333 ns         9333 ns        76496
BM_ComputeValue_WillGenerate_Raw                        9340 ns         9339 ns        74959
```

I understand that the benchmark may not be truly representative of runtime (the string size tested of 10KB should not be considered typical). For a short string it's similar performance.

toolchain/lex/string_literal.cpp
```diff
@@ -341,8 +353,8 @@ static auto ExpandAndConsumeEscapeSequence(LexerDiagnosticEmitter& emitter,
 // Expand any escape sequences in the given string literal.
 static auto ExpandEscapeSequencesAndRemoveIndent(
     LexerDiagnosticEmitter& emitter, llvm::StringRef contents, int hash_level,
-    llvm::StringRef indent) -> std::string {
-  std::string result;
+    llvm::StringRef indent) -> GeneratedString {
+  GeneratedString result;
   result.reserve(contents.size());
```
zygoloid: Hm! It strikes me that the result will *always* fit into this reserved storage, because escape sequences always expand to something no longer than themselves (in fact, always at least one character shorter currently). So instead of using a `SmallString`, we could allocate `contents.size()` from the `BumpPtrAllocator`, and fill it directly. That should get us to a single allocation and no copies.

(We may have a little unused space at the end of the allocation, which we could return to the `BumpPtrAllocator` if it had an API to do that, but currently no such API exists.)

jonmeow: Done. Let me know if you see an issue, but I've done my best to translate the code and the performance still isn't good. Still, maybe it's easier to improve like this.

```
--------------------------------------------------------------------------------------------
Benchmark                                                  Time             CPU   Iterations
--------------------------------------------------------------------------------------------
BM_ValidString_Simple                                  31166 ns        31166 ns        22384
BM_ValidString_Multiline                               32121 ns        32119 ns        21954
BM_ValidString_MultilineDoubleQuote                    31869 ns        31863 ns        21758
BM_ValidString_Raw                                     31928 ns        31928 ns        22025
BM_IncompleteWithEscapes_Simple                        47904 ns        47893 ns        13754
BM_IncompleteWithEscapes_Multiline                     41636 ns        41636 ns        16672
BM_IncompleteWithEscapes_MultilineDoubleQuote          42071 ns        42058 ns        16703
BM_IncompleteWithEscapes_Raw                          104998 ns       104988 ns         6697
BM_ComputeValue_NoGenerate_Short                        20.7 ns         20.7 ns     33566715
BM_ComputeValue_NoGenerate_Long                         3164 ns         3163 ns       220592
BM_ComputeValue_WillGenerate_Short                      52.0 ns         52.0 ns     13129991
BM_ComputeValue_WillGenerate_Long                      16368 ns        16366 ns        44319
BM_ComputeValue_WillGenerate_Multiline                 16194 ns        16191 ns        42727
BM_ComputeValue_WillGenerate_MultilineDoubleQuote      16438 ns        16435 ns        43282
BM_ComputeValue_WillGenerate_Raw                       16286 ns        16279 ns        43080
```

toolchain/lex/string_literal.cpp
```diff
@@ -412,15 +431,18 @@ static auto ExpandEscapeSequencesAndRemoveIndent(
               "escape sequence in a string literal.");
           emitter.Emit(contents.begin(), InvalidHorizontalWhitespaceInString);
           // Include the whitespace in the string contents for error recovery.
-          result += contents.substr(0, after_space);
+          int len = after_space == llvm::StringRef::npos ? contents.size()
+                                                         : after_space;
+          memcpy(buffer_cursor, contents.data(), len);
+          buffer_cursor += len;
```
zygoloid: This looks very similar to the code above (L392-396), including the handling of the `npos` case; consider factoring out a function to append a `StringRef` to the buffer.

toolchain/lex/string_literal.cpp
```diff
@@ -451,8 +474,17 @@ auto StringLiteral::ComputeValue(LexerDiagnosticEmitter& emitter) const
...
       multi_line_ ? CheckIndent(emitter, text_, content_) : llvm::StringRef();
-  return ExpandEscapeSequencesAndRemoveIndent(emitter, content_, hash_level_,
-                                              indent);
+  if (!content_needs_validation_ && (!multi_line_ || indent.empty())) {
+    return content_;
+  }
+
+  auto result = ExpandEscapeSequencesAndRemoveIndent(
+      emitter, content_, hash_level_, indent,
+      allocator.Allocate<char>(content_.size()));
```
zygoloid, line 489: Even with the `CHECK` below, I think we need a comment here to say why this is a sufficiently large allocation.

toolchain/lex/string_literal.cpp
```diff
@@ -114,11 +114,13 @@ auto StringLiteral::Lex(llvm::StringRef source_text)
   terminator.resize(terminator.size() + hash_level, '#');
   escape.resize(escape.size() + hash_level, '#');
 
+  bool content_needs_validation = false;
```
zygoloid, line 117: Declaring this `volatile` makes this code significantly faster, in the `BM_ValidString` and `BM_IncompleteWithEscapes` benchmarks. Faster than removing both of the stores and declaring it `const`. And for the non-raw cases, faster than the baseline -- running in about two thirds of the time. (The raw benchmark still gets slower, but by less -- 10% rather than 20%.)

@chandlerc pls hlp

toolchain/lex/string_literal.cpp
zygoloid: I can reproduce your results. Here's what I think:

- The slowdown in `IncompleteWithEscapes` is due to computing the `content_needs_validation` flag. I think something weird is happening when optimizing `StringLiteral::Lex`. Making the flag `volatile` helps substantially.
- The slowdown in `ComputeValue` is a measuring artifact. You're hitting the best case for the old code (repeatedly deallocating and reallocating a heap region of the exact same size over and over again) and hitting a bad / typical case for the new code (performing a long series of allocations into the same bump pointer allocator). Most of the difference goes away if you call `Reset` on the `BumpPtrAllocator` each loop iteration, but some difference remains.

zygoloid: Having spent a while looking at this with @chandlerc, the differences in at least `ValidString` and `IncompleteWithEscapes` seem to be performance ghosts caused by CPU weirdnesses like function alignment rather than an actual performance difference in the code. And the slowdown on the `ComputeValue` benchmark seems to be a measuring artifact. So I think the conclusion is that we're fine from a performance perspective. :shrug:

toolchain/lex/string_literal.cpp
zygoloid: This appears to be "noise", caused by something like loop alignment.

toolchain/lex/string_literal.cpp
```diff
@@ -127,7 +129,12 @@ auto StringLiteral::Lex(llvm::StringRef source_text)
     // escape sequences starting with a predictable character and not containing
     // embedded and unescaped terminators or newlines.
     switch (source_text[cursor]) {
+      case '\t':
+        // Tabs have extra validation.
+        content_needs_validation = true;
+        break;
       case '\\':
+        content_needs_validation = true;
```
zygoloid: May as well move this inside the `if` below; a plain `\\` that's not part of an escape doesn't need validation.

zygoloid (approved):

toolchain/lex/string_literal.cpp
jonmeow: Done. I was thinking about this wrong, thinking it would suppress a diagnostic when this is actually relevant to raw string handling.

toolchain/lex/string_literal.cpp
jonmeow: Added:

```
  // "Expanding" escape sequences should only ever shorten content. As a
  // consequence, the output string should allows fit within this allocation.
  // Although this may waste some space, it avoids a reallocation.
```

toolchain/lex/string_literal.cpp
jonmeow: Done

toolchain/lex/string_literal.cpp
zygoloid: For posterity, I think a fairer way to compare the results here is to add to `BM_ComputeValue` a `std::vector<std::string>` to hold the results in the cases without a `BumpPtrAllocator`, so that in all cases we store all allocations until the end of the benchmark. With that, I get:

Baseline (+ `std::vector<std::string>` non-stable storage):
```
BM_ComputeValue_NoGenerate_Short                        67.7 ns         67.7 ns     11734809
BM_ComputeValue_NoGenerate_Long                        15299 ns        15298 ns        46581
BM_ComputeValue_WillGenerate_Short                      89.5 ns         89.5 ns      9173553
BM_ComputeValue_WillGenerate_Long                      15293 ns        15292 ns        46916
BM_ComputeValue_WillGenerate_Multiline                 16082 ns        16082 ns        43286
BM_ComputeValue_WillGenerate_MultilineDoubleQuote      15793 ns        15790 ns        45479
BM_ComputeValue_WillGenerate_Raw                       15544 ns        15541 ns        45996
```

With the `std::unique_ptr<std::string>` version (+ `std::vector<StringLiteral::ComputedValue>` stable storage):
```
BM_ComputeValue_NoGenerate_Short                        56.0 ns         56.0 ns     10000000
BM_ComputeValue_NoGenerate_Long                         4770 ns         4770 ns       145437
BM_ComputeValue_WillGenerate_Short                       116 ns          116 ns      6843959
BM_ComputeValue_WillGenerate_Long                      15492 ns        15492 ns        44583
BM_ComputeValue_WillGenerate_Multiline                 15527 ns        15526 ns        43523
BM_ComputeValue_WillGenerate_MultilineDoubleQuote      15569 ns        15568 ns        43896
BM_ComputeValue_WillGenerate_Raw                       16037 ns        16036 ns        42307
```

This PR (with `llvm::BumpPtrAllocator` for stable storage):
```
BM_ComputeValue_NoGenerate_Short                        24.2 ns         24.1 ns     29077659
BM_ComputeValue_NoGenerate_Long                         4725 ns         4724 ns       147751
BM_ComputeValue_WillGenerate_Short                      62.9 ns         62.9 ns     11202462
BM_ComputeValue_WillGenerate_Long                      17900 ns        17898 ns        39221
BM_ComputeValue_WillGenerate_Multiline                 17984 ns        17981 ns        39195
BM_ComputeValue_WillGenerate_MultilineDoubleQuote      17886 ns        17885 ns        39069
BM_ComputeValue_WillGenerate_Raw                       17952 ns        17949 ns        38389
```

I think the first three of these are probably the most common and most important, and this approach seems to win out there. The bottom four cases exceed the `BumpPtrAllocator`'s default slab size, so they're allocating custom slabs, which is a little less efficient than doing a direct heap allocation here because the `BumpPtrAllocator` adds tracking overhead. So I think these numbers make sense.

PR #3315, opened 2023-10-19
Expand comments in `enum_base.h`

josh11b:


common/enum_base.h
```diff
@@ -51,6 +58,44 @@ namespace Carbon::Internal {
 //   #include ".../my_kind.def"
 //   };
 //   ```
+//
+// The result of the above:
+// - An enum class ("RawEnumType") defined in an `Internal` namespace with one
+//   enumerator per call to CARBON_MY_KIND(Name) in ".../my_kind.def", with name
+//   `Name`.
+// - A type `MyKind` with a RawEnumType field that includes:
+//   - all the public members of `EnumBase`, like `name` and `Print`
```
jonmeow: I believe each bullet should use sentence casing. (all -> All, etc)

Also, should this have a period at the end?

Relevant style: https://developers.google.com/style/lists#types-of-lists

common/enum_base.h
```diff
@@ -51,6 +58,44 @@ namespace Carbon::Internal {
 //   #include ".../my_kind.def"
 //   };
 //   ```
+//
+// The result of the above:
+// - An enum class ("RawEnumType") defined in an `Internal` namespace with one
+//   enumerator per call to CARBON_MY_KIND(Name) in ".../my_kind.def", with name
+//   `Name`.
+// - A type `MyKind` with a RawEnumType field that includes:
```
jonmeow: This may be a phrasing issue, but isn't it `MyKind` that contains the below list, versus `MyKind` containing `RawEnumType` and `RawEnumType` containing the below list? Why is `RawEnumType` listed as being here instead of as part of the list?

Also, you might want to include a warning: RawEnumType isn't intended to be a frequently used API, it's more to address niche use-cases that need it -- particularly as a template argument, since you can template on an `enum` but not the `MyKind` wrapper.

common/enum_base.h
```diff
@@ -51,6 +58,44 @@ namespace Carbon::Internal {
...
+//       switch (k) {
+//         case MyKind::Name1:
+//           // ...
+//           break;
+//
+//         case MyKind::Name2:
+//           // ...
+//           break;
+//
+//         // No `default` case needed if the above cases are exhaustive.
```
jonmeow, line 102: While this says that `default` is optional, doesn't adding it suppress a compiler warning? It may be helpful to be detailed if you're talking about this point, for example noting that the compiler is checking that all cases of RawEnumType are handled.

common/enum_base.h
```diff
@@ -51,6 +58,44 @@ namespace Carbon::Internal {
 //   #include ".../my_kind.def"
 //   };
 //   ```
+//
+// The result of the above:
+// - An enum class ("RawEnumType") defined in an `Internal` namespace with one
```
jonmeow: What's different about `RawEnumType` that you use quotes for it versus backticks for other identifiers? (e.g., `Internal`, `MyKind`)

Note you do neither in later mentions of `RawEnumType`, it's just plain text.

common/enum_base.h
```diff
@@ -51,6 +58,44 @@ namespace Carbon::Internal {
...
+//   enumerator per call to CARBON_MY_KIND(Name) in ".../my_kind.def", with name
```
jonmeow: On lines 26 and 54, files in comments use backticks. On lines 64 and 76, you're using quotes. Any reason for the switch?

common/enum_base.h
```diff
@@ -51,6 +58,44 @@ namespace Carbon::Internal {
...
+//   `Name`.
+// - A type `MyKind` with a RawEnumType field that includes:
+//   - all the public members of `EnumBase`, like `name` and `Print`
+//     ```
+//     auto ErrorMessage(MyKind k) -> std::string {
+//       return k.name() + " not found";
+//     }
+//     ```
+//   - any protected members of `EnumBase` that were included with `using`, like
+//     `AsInt`, `FromInt`.
```
jonmeow: Line 66 is talking about `MyKind` in general. Given you talk about access control details here, maybe line 66 should clarify that it's talking about the public API of `MyKind` rather than the type as a whole? (because the protected members are accessible to user-defined functions on `MyKind` regardless of whether they're public)

common/enum_base.h
```diff
@@ -51,6 +58,44 @@ namespace Carbon::Internal {
...
 //   ```
+//
+// The result of the above:
+// - An enum class ("RawEnumType") defined in an `Internal` namespace with one
+//   enumerator per call to CARBON_MY_KIND(Name) in ".../my_kind.def", with name
+//   `Name`.
+// - A type `MyKind` with a RawEnumType field that includes:
+//   - all the public members of `EnumBase`, like `name` and `Print`
+//     ```
+//     auto ErrorMessage(MyKind k) -> std::string {
```
jonmeow, line 72: Where does `ErrorMessage` come from? I'm not aware of it as a member of `EnumBase`. If it's supposed to be an example, can you please label and explain what it's demonstrating?

common/enum_base.h
```diff
@@ -51,6 +58,44 @@ namespace Carbon::Internal {
...
+//     }
+//     ```
+//   - any protected members of `EnumBase` that were included with `using`, like
+//     `AsInt`, `FromInt`.
+//   - one static constant member of type `MyKind` per call to
+//     CARBON_MY_KIND(Name) in ".../my_kind.def" with name `Name` and the
+//     corresponding value from RawEnumType.
+//     ```
+//     ErrorMessage(MyKind::Name1);
+//     ```
```
jonmeow: Had you considered trying to phrase this as code, such as:

```
A member `static const MyKind Name;` per call to `CARBON_MY_KIND(Name)` in ".../my_kind.def".
```

Also, can you please label and explain the `ErrorMessage` example? It took me several reads to understand it's probably an example rather than an entry in the list.

jonmeow (commented):
Thanks! A few comments to try to make it easier to understand.

common/enum_base.h
josh11b: I added backticks.

common/enum_base.h
josh11b: I've switched to quotes.

common/enum_base.h
josh11b: Reorganized and rephrased.

common/enum_base.h
josh11b: These were not sentences before, since they generally didn't have subject or verbs, but a list of things included in `MyKind`. I've rearranged to make these sentences.

common/enum_base.h
josh11b: Rephrased.

common/enum_base.h
josh11b: Labelled.

common/enum_base.h
josh11b: Done.

common/enum_base.h
josh11b: I added a sentence. Does that address what you were looking for?

common/enum_base.h
jonmeow: We're already using backticks for filenames, so it may be better not to switch if there's not a specific reason (and maybe I should've also noted line 21?). I think this reflects that filenames are closer to code.

For example, while it's hard to filter through `#include` in code, here are some md searches:

- [double quotes](https://github.com/search?q=repo%3Acarbon-language%2Fcarbon-lang+%2F%5C.%28h%7Ccpp%7Ccarbon%29%22%2F+path%3A**%2F*.md&type=code)
- [backticks](https://github.com/search?q=repo%3Acarbon-language%2Fcarbon-lang+%2F%5C.%28h%7Ccpp%7Ccarbon%29%60%2F+path%3A**%2F*.md&type=code)

common/enum_base.h
```diff
@@ -38,19 +34,78 @@ namespace Carbon::Internal {
...
+//   - `MyKind` includes a member `static const MyKind Name;` per call to
+//     `CARBON_MY_KIND(Name)` in ".../my_kind.def". It will have the
+//     corresponding value from `RawEnumType`. This is the primary way to create
+//     an instance of `MyKind`. For example, it might be used like:
+//     ```
+//     ErrorMessage(MyKind::Name1);
+//     ```
+//   - `MyKind` includes an implicit conversion to the `RawEnumType`, returning
+//     the value of a private field in `EnumBase`. This is used when writing a
+//      `switch` statement, as in this example:
```
jonmeow:
```suggestion
//     `switch` statement, as in this example:
```

jonmeow (approved):
Thanks!

PR #3316, opened 2023-10-19
Switch File functions, classes, and types to ValueStores

jonmeow:
Building on #3313, start using ValueStore on File. Functions and classes are straightforward. Types here I present as a borderline case where maybe we want a more bespoke API, but maybe this is okay? Most other things probably need a slightly different API, which although I might do that for a consistent interface, felt more out-of-scope for this change.

### DIFF
--- a/toolchain/sem_ir/file.h
+++ b/toolchain/sem_ir/file.h
@@ -153,42 +153,6 @@ class File : public Printable<File> {
     return *cross_reference_irs_[xref_id.index];
   }
 
-  // Adds a callable, returning an ID to reference it.
-  auto AddFunction(Function function) -> FunctionId {
-    FunctionId id(functions_.size());
-    // TODO: Return failure on overflow instead of crashing.
-    CARBON_CHECK(id.index >= 0);
-    functions_.push_back(function);
-    return id;
-  }
-
-  // Returns the requested callable.
-  auto GetFunction(FunctionId function_id) const -> const Function& {
-    return functions_[function_id.index];
-  }
-
-  // Returns the requested callable.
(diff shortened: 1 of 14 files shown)


toolchain/sem_ir/file.h
```diff
@@ -399,13 +344,18 @@ class File : public Printable<File> {
...
-  auto functions_size() const -> int { return functions_.size(); }
-  auto classes_size() const -> int { return classes_.size(); }
+  auto functions() -> ValueStore<FunctionId, Function>& { return functions_; }
+  auto functions() const -> const ValueStore<FunctionId, Function>& {
+    return functions_;
+  }
+  auto classes() -> ValueStore<ClassId, Class>& { return classes_; }
+  auto classes() const -> const ValueStore<ClassId, Class>& { return classes_; }
+  auto types() -> ValueStore<TypeId, TypeInfo>& { return types_; }
+  auto types() const -> const ValueStore<TypeId, TypeInfo>& { return types_; }
```
zygoloid, line 354: You mention this in the commit message, but just reiterating the change here: previously, `TypeInfo` had been an encapsulated implementation detail here, and the management of the value representation and type completeness wasn't exposed and could only be modified by directly calling `CompleteType`. I think we're more consistent after this change -- we don't have encapsulation for any of the representations of other things like members of `Class` or `Function`, for which we also might want invariants of various kinds.

I think we should have a more general discussion at some point of what we want to do regarding encapsulation of `SemIR` data. Right now, having essentially everything be public seems to be working well, but I do worry that we may have problems down the road when higher-level invariants start becoming more important. But not as a prerequisite for this change :)

zygoloid (approved):

PR #3317, opened 2023-10-20
Add more value store types to File.

jonmeow:
Finishing what #3316 started, add more bespoke ValueStore-like structures to File. With this, the things which previously had somewhat boilerplate Add/Get functions are now all on side classes, giving a uniform style of API for calling.

Note, I was on the fence about making things public on ValueStore. If it's preferred that I make some things there protected I certainly can, there's just a trade-off that may mean more distinct child/wrapper types.

### DIFF
--- a/toolchain/base/value_store.h
+++ b/toolchain/base/value_store.h
@@ -91,6 +91,13 @@ class ValueStore : public Printable<ValueStore<IdT, ValueT>> {
     return id;
   }
 
+  // Adds a default constructed value and returns an ID to reference it.
+  auto AddDefaultValue() -> IdT {
+    auto id = IdT(values_.size());
+    values_.resize(id.index + 1);
+    return id;
+  }
+
   // Returns a mutable value for an ID.
   auto Get(IdT id) -> ValueT& {
     CARBON_CHECK(id.index >= 0) << id.index;
@@ -102,6 +109,9 @@ class ValueStore : public Printable<ValueStore<IdT, ValueT>> {
     CARBON_CHECK(id.index >= 0) << id.index;
     return values_[id.index];
   }
+
+  // Reserves space.
+  auto Reserve(size_t size) -> void { values_.reserve(size); }
 
   auto Print(llvm::raw_ostream& out) const -> void { Print(out, 0); }
   auto Print(llvm::raw_ostream& out, int indent) const -> void {
--- a/toolchain/base/value_store.h
+++ b/toolchain/base/value_store.h
@@ -5,6 +5,8 @@
 #ifndef CARBON_TOOLCHAIN_BASE_VALUE_STORE_H_
 #define CARBON_TOOLCHAIN_BASE_VALUE_STORE_H_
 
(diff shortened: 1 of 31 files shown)


jonmeow:
@zygoloid With this, I'd also like your feedback on whether you think convenience wrappers on Context would be helpful. Note I'd probably just copy them all over (the mutable ones, no reason for const) and replace `context.semantics_ir().foo()` with just `context.foo()`. But to your earlier request, I think this change makes it clearest what that would look like, since it finishes the set of objects being moved over. (and if you want this, I'd prefer a separate PR, just to keep things incremental)

toolchain/base/value_store.h
```diff
@@ -79,24 +79,43 @@ constexpr StringId StringId::Invalid(StringId::InvalidIndex);
...
+    auto id = IdT(values_.size());
+    values_.resize(id.index + 1);
+    return id;
+  }
+
+  // Returns a mutable value for an ID.
+  auto Get(IdT id) -> ValueT& {
+    CARBON_CHECK(id.index >= 0) << id.index;
+    return values_[id.index];
+  }
```
zygoloid: I think we should distinguish between a `ValueStore` and a `MutableValueStore`. When storing (eg) `Integer`s, I think the guarantee of immutability is important, and we shouldn't provide any mutable access.

We could do this either with two different `ValueStore` types or by making the immutable ones have a `const`-qualified `ValueT` type. I think it's probably clearer to readers of the context / file APIs to have distinct type names for the `ValueStore`s, but I'd be happy either way.

toolchain/sem_ir/value_stores.h
```diff
@@ -0,0 +1,151 @@
...
+};
+
+// Provides a block-based ValueStore, which uses slab allocation of added
+// blocks. This allows references to values to outlast vector resizes that might
+// otherwise invalidate references.
+//
+// BlockValueStore is used as-is, but there are also children that expose the
+// protected members for type-specific functionality.
+template <typename IdT, typename ValueT>
+class BlockValueStore {
```
zygoloid, line 75: I really like how this turned out. Very clean, thanks!

toolchain/base/value_store.h
zygoloid: Having looked at the custom value stores in SemIR, many of which provide some kind of mutable access, I'm no longer so sure that putting the `const`ness in the name of the store will help readers very much. Just changing the `IndexedType`s for some of the `Id`s to be `const` seems fine to me.

jonmeow: Done with `const`. I think it's mainly the shared ValueStores that can use this; the semir types have become more mutable.

jonmeow:
FYI, I put in a change to deal with printability, mainly to solve problems I'm having with the array types. I may update that further (I wanted to use it more with file.cpp, but here I think it makes the most sense to just get things working with minimal additional deltas).

zygoloid (approved):
Looks good, hope you don't have too much merge pain.

PR #3319, opened 2023-10-20
Test for id overflow and string deduplication in `SharedValueStores`

josh11b:
Follow-on to #3311

### DIFF
--- a/toolchain/base/value_store.h
+++ b/toolchain/base/value_store.h
@@ -82,7 +82,7 @@ class ValueStore {
   // Stores the value and returns an ID to reference it.
   auto Add(typename IdT::IndexedType value) -> IdT {
     IdT id = IdT(values_.size());
-    CARBON_CHECK(id.index >= 0) << "Too many numeric literals";
+    CARBON_CHECK(id.index >= 0) << "Id overflow";
     values_.push_back(std::move(value));
     return id;
   }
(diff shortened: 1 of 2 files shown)


toolchain/base/value_store.h
```diff
@@ -81,7 +81,8 @@ class ValueStore {
  public:
   // Stores the value and returns an ID to reference it.
   auto Add(typename IdT::IndexedType value) -> IdT {
-    auto id = IdT(values_.size());
+    IdT id = IdT(values_.size());
+    CARBON_CHECK(id.index >= 0) << "Too many numeric literals";
```
jonmeow: Please fix the error message to be generic (this is not exclusive to numeric literals). Just saying "overflow" or nothing would, I think, be fine.

toolchain/base/value_store.h
```diff
@@ -106,6 +107,7 @@ class ValueStore<StringId> {
   auto Add(llvm::StringRef value) -> StringId {
     auto [it, inserted] = map_.insert({value, StringId(values_.size())});
     if (inserted) {
+      CARBON_CHECK(it->second.index >= 0) << "Too many unique string literals";
```
jonmeow: As above.

jonmeow (commented):
Bear in mind these are generic types not for a singular purpose.

toolchain/base/value_store.h
josh11b: Isn't this the specialization for strings? It seems valuable to have this additional information.

jonmeow: It's for strings, not necessarily string literals. Also, std::string, not Carbon strings

josh11b: Or is this not specific to literals?

toolchain/base/value_store.h
josh11b: Ah, I see this is used for more types in later PRs.

jonmeow (approved):

PR #3320, opened 2023-10-20
Adding ValueStore printing and --dump-shared-values

jonmeow:
This replaces the printing that was removed from SemIR's raw dump. It's separate because (for example) lexing generates shared values, and so reviewing them is not specific to any particular phase.

### DIFF
--- a/toolchain/base/value_store.h
+++ b/toolchain/base/value_store.h
@@ -10,6 +10,7 @@
 #include "llvm/ADT/APInt.h"
 #include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/YAMLParser.h"
 #include "toolchain/base/index_base.h"
 
 namespace Carbon {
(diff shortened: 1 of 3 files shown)


toolchain/base/value_store.h
```diff
@@ -120,14 +128,23 @@ class ValueStore<StringId> {
     return values_[id.index];
   }
 
+  auto Print(llvm::raw_ostream& out) const -> void { Print(out, 0); }
+  auto Print(llvm::raw_ostream& out, int indent) const -> void {
+    for (auto value : values_) {
+      out.indent(indent);
+      // TODO: Consider making this more robust for strings that need escaping.
+      out << "- \"" << value << "\"\n";
```
zygoloid: Here's one we made earlier:
```suggestion
      out << "- \"" << llvm::yaml::escape(value) << "\"\n";
```
You'll need to `#include "llvm/Support/YAMLParser.h"`.

zygoloid (approved):

toolchain/base/value_store.h
jonmeow: Thanks, done

PR #3324, opened 2023-10-21
Lower types in the order they were completed.

zygoloid:
This is a prerequisite for class support, where a class can be referenced as a type before it becomes complete. For example, given:

```carbon
class A {
  fn F(a: A);

  class B {}
  var b: B;
}

fn A.F(a: A) {}
```

we need to lower `B` before we lower `A`, even though `A` is used as a type first.

This will also start catching some cases where we don't require a type to be complete despite using it, as we now only lower types that are required to be complete.

Remove the poison values for struct and tuple literals. We don't need those any more, because we never generate references to those literals as values, and we don't have a type to use for them because we never require the type of a literal to be complete, only the type of the entity initialized by the literal, which can be different, for example when initializing an array from a tuple literal or a class from a struct literal.

This doesn't affect the output: `llvm::Type` objects that are not referenced by an LLVM module don't affect the IR for that module, and the order in which `llvm::Type`s are created doesn't affect anything either.

### DIFF
--- a/toolchain/check/context.cpp
+++ b/toolchain/check/context.cpp
@@ -370,6 +370,15 @@ class TypeCompleter {
           work_list_.push_back(
               {value_rep.type_id, Phase::BuildValueRepresentation});
         }
+        // For a pointer representation, the pointee also needs to be complete.
+        if (value_rep.kind == SemIR::ValueRepresentation::Pointer) {
+          auto pointee_type_id =
+              context_.semantics_ir().GetPointeeType(value_rep.type_id);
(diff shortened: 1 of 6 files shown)


jonmeow (commented):
This seems like a good change, but why doesn't it affect any tests, even to result in a different output order?

zygoloid:
> This seems like a good change, but why doesn't it affect any tests, even to result in a different output order?

I've updated the change description to explain.

jonmeow (approved):

jonmeow:
Thanks!

PR #3325, opened 2023-10-21
Cleanups of SIMD code and document no Arm port.

chandlerc:
I spent (a lot) of time working to see if there was any profitable way to port the SIMD code that scans for identifier length to Arm. There isn't really. =/ While working on these, I made some cleanups to the SIMD code that seemed worth landing, and added some benchmarks. All this PR does is the cleanups, benchmarks, and documents that Arm isn't just waiting to get attention but doesn't really have good options (so far).

For posterity, here are the core techniques I tried:

1) Direct 32-byte SIMD scanning using pair-wise add trees to build
   a 32-bit mask of valid identifier and then `clz` to compute the
   distance. This is a very good analog to the 16-byte SIMD structure
   used on x86-64. The pair-wise summing technique is the one used in
   simdjson for similar purposes.
2) A 16-byte SIMD scanning similar to the x86 version but using `shrn`
   to produce a 64-bit scalar bitmask with 4 bits per byte, and then
   scaling the bit-count distance.
3) Various hybrid versions of (1) and (2) with short scalar scans to
   identify short identifiers before paying the SIMD start-up cost.
4) A much fancier version of (1) that scanned 64-bytes at a time, but
   cached the resulting 64-bit mask and re-used it until exhausted.

Some good background on these techniques on Arm CPUs is in this blog post: https://community.arm.com/arm-community-blogs/b/infrastructure-solutions-blog/posts/porting-x86-vector-bitmask-optimizations-to-arm-neon

Sadly, both (1) and (2) were significantly slower than a scalar loop over the bytes. Even (3) was consistently slower.

The only approach that came close was (4) and it was very *slightly* slower in typical examples and very *slightly* faster in extremely difficult cases like huge identifiers.

Ultimately, the only path I see (suggested by Dougall on a Mastodon discussion of this whole problem space) is to take (4) to the limit of computing an identifier-or-not bitmask *for the entire source file* using a deeply throughput optimized routine (maybe as part of the line scanning). That should be able to manage the high latency you end up with when handling these patterns in SIMD on Arm.

The good news is that at least the M1 is *so* fast in the byte-scanning loop that this isn't hurting nearly as much as I feared.

### DIFF
--- a/toolchain/lex/tokenized_buffer.cpp
+++ b/toolchain/lex/tokenized_buffer.cpp
@@ -80,98 +80,128 @@ static constexpr SIMDMaskArrayT PrefixMasks = []() constexpr {
 }();
 #endif  // CARBON_USE_SIMD
 
-// Scans the provided text and returns the prefix `StringRef` of contiguous
-// identifier characters.
+// A table of booleans that we can use to classify bytes as being valid
+// identifier (or keyword) characters. This is used in the generic,
+// non-vectorized fallback code to scan for length of an identifier.
+constexpr std::array<bool, 256> IsIdByteTable = [] {
+  std::array<bool, 256> table = {};
+  for (char c = '0'; c <= '9'; ++c) {
+    table[c] = true;
+  }
+  for (char c = 'A'; c <= 'Z'; ++c) {
+    table[c] = true;
+  }
+  for (char c = 'a'; c <= 'z'; ++c) {
+    table[c] = true;
+  }
+  table['_'] = true;
+  return table;
+}();
+
+// Baseline scalar version, also available for scalar-fallback in SIMD code.
+// Uses `ssize_t` for performance when indexing in the loop.
 //
-// This is a performance sensitive function and so uses vectorized code
-// sequences to optimize its scanning. When modifying, the identifier lexing
-// benchmarks should be checked for regressions.
-//
-// Identifier characters here are currently the ASCII characters `[0-9A-Za-z_]`.
-//
-// TODO: Currently, this code does not implement Carbon's design for Unicode
-// characters in identifiers. It does work on UTF-8 code unit sequences, but
-// currently considers non-ASCII characters to be non-identifier characters.
-// Some work has been done to ensure the hot loop, while optimized, retains
-// enough information to add Unicode handling without completely destroying the
-// relevant optimizations.
-static auto ScanForIdentifierPrefix(llvm::StringRef text) -> llvm::StringRef {
-  // A table of booleans that we can use to classify bytes as being valid
-  // identifier (or keyword) characters. This is used in the generic,
-  // non-vectorized fallback code to scan for length of an identifier.
-  static constexpr std::array<bool, 256> IsIdByteTable = ([]() constexpr {
-    std::array<bool, 256> table = {};
-    for (char c = '0'; c <= '9'; ++c) {
-      table[c] = true;
-    }
-    for (char c = 'A'; c <= 'Z'; ++c) {
-      table[c] = true;
-    }
-    for (char c = 'a'; c <= 'z'; ++c) {
-      table[c] = true;
-    }
-    table['_'] = true;
-    return table;
-  })();
+// TODO: This assumes all Unicode characters are non-identifiers.
+static auto ScanForIdentifierPrefixScalar(llvm::StringRef text, ssize_t i)
+    -> llvm::StringRef {
+  const ssize_t size = text.size();
+  while (i < size && IsIdByteTable[static_cast<unsigned char>(text[i])]) {
+    ++i;
+  }
+
+  return text.substr(0, i);
+}
 
 #if CARBON_USE_SIMD && __x86_64__
-  // This code uses a scheme derived from the techniques in Geoff Langdale and
-  // Daniel Lemire's work on parsing JSON[1]. Specifically, that paper outlines
-  // a technique of using two 4-bit indexed in-register look-up tables (LUTs) to
-  // classify bytes in a branchless SIMD code sequence.
-  //
-  // [1]: https://arxiv.org/pdf/1902.08318.pdf
-  //
-  // The goal is to get a bit mask classifying different sets of bytes. For each
-  // input byte, we first test for a high bit indicating a UTF-8 encoded Unicode
-  // character. Otherwise, we want the mask bits to be set with the following
-  // logic derived by inspecting the high nibble and low nibble of the input:
-  // bit0 = 1 for `_`: high `0x5` and low `0xF`
-  // bit1 = 1 for `0-9`: high `0x3` and low `0x0` - `0x9`
-  // bit2 = 1 for `A-O` and `a-o`: high `0x4` or `0x6` and low `0x1` - `0xF`
-  // bit3 = 1 for `P-Z` and 'p-z': high `0x5` or `0x7` and low `0x0` - `0xA`
-  // bit4 = unused
-  // bit5 = unused
-  // bit6 = unused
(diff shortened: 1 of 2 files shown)


toolchain/lex/tokenized_buffer.cpp
```diff
@@ -80,98 +80,128 @@ static constexpr SIMDMaskArrayT PrefixMasks = []() constexpr {
...
+// bit2 = 1 for `A-O` and `a-o`: high `0x4` or `0x6` and low `0x1` - `0xF`
+// bit3 = 1 for `P-Z` and 'p-z': high `0x5` or `0x7` and low `0x0` - `0xA`
+// bit4 = unused
+// bit5 = unused
+// bit6 = unused
+// bit7 = unused
+//
+// No bits set means definitively non-ID ASCII character.
+//
+// bits 4-7 remain unused if we need to classify more characters.
```
jonmeow:
```suggestion
// Bits 4-7 remain unused if we need to classify more characters.
```

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -80,98 +80,128 @@ static constexpr SIMDMaskArrayT PrefixMasks = []() constexpr {
...
+  uint8_t nibble_a;
+  uint8_t nibble_b;
+  uint8_t nibble_c;
+  uint8_t nibble_d;
+  uint8_t nibble_e;
+  uint8_t nibble_f;
+
+  auto Load() const -> __m128i {
+    return _mm_load_si128(reinterpret_cast<const __m128i*>(this));
+  }
```
jonmeow: Style: Shouldn't functions be before members?

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -80,98 +80,128 @@ static constexpr SIMDMaskArrayT PrefixMasks = []() constexpr {
...
+// bit4 = unused
+// bit5 = unused
+// bit6 = unused
+// bit7 = unused
+//
+// No bits set means definitively non-ID ASCII character.
+//
+// bits 4-7 remain unused if we need to classify more characters.
+namespace {
+struct alignas(16) NibbleLUT {
```
jonmeow, line 143: Should the `alignas` use be commented? I suspect this is to get it to have consistent alignment with `__m128i`, although would it make sense to have a `static_assert` demonstrate that?

FWIW I do find these structs a little easier to read than the prior code; an interesting compromise after the previous discussions.

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -224,19 +254,49 @@ static auto ScanForIdentifierPrefix(llvm::StringRef text) -> llvm::StringRef {
...
-  // TODO: Optimize this with SIMD for other architectures.
-  return text.take_while(
-      [](char c) { return IsIdByteTable[static_cast<unsigned char>(c)]; });
+#endif  // CARBON_USE_SIMD && __x86_64__
+
+// Scans the provided text and returns the prefix `StringRef` of contiguous
+// identifier characters.
+//
+// This is a performance sensitive function and so uses vectorized code
+// sequences to optimize its scanning. When modifying, the identifier lexing
```
jonmeow:
```suggestion
// This is a performance sensitive function and uses vectorized code
// sequences to optimize its scanning on supported platforms. When modifying, the identifier lexing
```

I think?

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -224,19 +254,49 @@ static auto ScanForIdentifierPrefix(llvm::StringRef text) -> llvm::StringRef {
...
+// Some work has been done to ensure the hot loop, while optimized, retains
+// enough information to add Unicode handling without completely destroying the
+// relevant optimizations.
+static auto ScanForIdentifierPrefix(llvm::StringRef text) -> llvm::StringRef {
+  // Dispatch to an optimized architecture optimized routine.
+#if CARBON_USE_SIMD && __x86_64__
+  return ScanForIdentifierPrefixX86(text);
+#elif CARBON_USE_SIMD && __ARM_NEON
+  // Somewhat surprisingly, there is basically nothing worth doing in SIMD on
+  // Arm to optimize this scan. The Neon SIMD operations end up requiring you to
```
jonmeow, line 286:
```suggestion
  // ARM to optimize this scan. The Neon SIMD operations end up requiring you to
```

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -224,19 +254,49 @@ static auto ScanForIdentifierPrefix(llvm::StringRef text) -> llvm::StringRef {
...
+  // move from the SIMD unit to the scalar unit in the critical path of finding
+  // the offset of the end of an identifier. Current ARM cores make the code
+  // sequences here (quite) unpleasant. For example, on Apple M1 and similar
```
jonmeow, line 289:
```suggestion
  // sequences here (quite) unpleasant. For example, on Apple M2 and similar
```

Is it safe to assume M2 is the same? I wonder if it'd be better to refer to the more recent chip design here in order to reflect that this isn't addressed by the more modern design. I think in seeing M2 I'd be less likely to consider M1: both that an older chip design is likely to also not have support, and we'd also be less likely to produce optimizations for older chip designs.

jonmeow (approved):
Looks good, I think this is an improvement. Just some minor typo and style comments. Feel free to merge, regardless of resolution.

toolchain/lex/tokenized_buffer.cpp
chandlerc: Good call on adding an alignment comment.

Yeah, trying Arm as well really forced me to re-visit this holistically, and on Arm this is somewhat forced to be a load rather than the `set` style of intrinsic. And that in turn got me thinking about the original issue, and reformulating it in terms of a load opens up the exact kind of improvement you were looking for. Sorry this didn't occur more easily w/o having been forced with Arm.

Anyways, glad this finally gets to something more readable!

toolchain/lex/tokenized_buffer.cpp
chandlerc: FWIW, this changed recently for structs specifically to allow this ordering:
https://google.github.io/styleguide/cppguide.html#Declaration_Order:~:text=(Optionally%2C%20for%20structs%20only)%20non%2Dstatic%20data%20members

My understanding is that this is based on the idea that the public data members are actually part of the public API just like functions. 🤷🏻 

That said, happy to change as I don't have any strong feeling.

toolchain/lex/tokenized_buffer.cpp
chandlerc: While there used to be an acronym here (Acorn RISC Machine and then Advanced RISC Machine), this changed back in 2017. It's now always just the word "Arm". I understand there are a *lot* of confusing spellings, but trying to follow what folks from the company have directly told me. I think the authoritative citation is basically here: https://www.arm.com/company/policies/trademarks

toolchain/lex/tokenized_buffer.cpp
chandlerc: I don't have data on M2 at all (or hardware to test with), so I'm not ready to assume that yet. =/

toolchain/lex/tokenized_buffer.cpp
chandlerc: Roughly. I'm going to further tweak to say where profitable, as we clearly don't on a fairly supported platform in one case.

chandlerc (commented):
Thanks for the review!! Most fixed. Left some replies and merging. Happy to do follow-ups as needed!

PR #3328, opened 2023-10-23
Fix missing include for ids.h

jonmeow:
I think this is only noticeable in a more modular build, but we try to keep that working.


zygoloid (approved):
Oops, thanks!

PR #3329, opened 2023-10-23
Separate `ClassType` from `ClassDeclaration`.

zygoloid:
Retain the `ClassDeclaration` node to represent a syntactic declaration of a class (including possibly a declaration of a generic class), but use a separate SemIR node to represent the class type itself. This allows us to give the two separate treatment.

The `ClassDeclaration` is still entered into the name lookup table for its enclosing scope, but when it is named in an expression, the class type is produced instead. When the class declaration is named in a declaration name, it can be used to define members of the class, but an expression that resolves to the class type cannot be used to define members of the class.

In order to distinguish these cases, use `Name` rather than `NameExpression` for the left-hand side of a `QualifiedName` parse node. This removes the only use of the `Expression` form of a declaration name, so that is also removed.

In the future, `ClassType` will also be used to describe types such as `Vector(T)`, for which there is no corresponding `ClassDeclaration`.


toolchain/check/declaration_name_stack.h
```diff
@@ -15,10 +15,9 @@ class Context;
 
 // Provides support and stacking for qualified declaration name handling.
 //
-// A qualified declaration name will consist of entries which are either
-// Identifiers or full expressions. Expressions are expected to resolve to
-// types, such as how `fn Vector(i32).Clear() { ... }` uses the expression
-// `Vector(i32)` to indicate the type whose member is being declared.
+// A qualified declaration name will consist of entries, which are `Name`s
+// optionally followed by generic parameter lists, such as `Vector(T:! type)`
+// in `fn Vector(T:! type).Clear();`, but parameter lists aren't supported yet.
```
jonmeow, line 20: Given "aren't supported yet", should there be a TODO?

toolchain/check/context.cpp
```diff
@@ -89,11 +89,13 @@ auto Context::DiagnoseNameNotFound(Parse::Node parse_node, StringId name_id)
                  semantics_ir_->strings().Get(name_id));
 }
 
-auto Context::NoteIncompleteClass(SemIR::ClassDeclaration class_decl,
+auto Context::NoteIncompleteClass(SemIR::ClassId class_id,
                                   DiagnosticBuilder& builder) -> void {
   CARBON_DIAGNOSTIC(ClassForwardDeclaredHere, Note,
                     "Class was forward declared here.");
-  builder.Note(class_decl.parse_node, ClassForwardDeclaredHere);
+  auto& class_info = semantics_ir().classes().Get(class_id);
```
```suggestion
  const auto& class_info = semantics_ir().classes().Get(class_id);
```

Even though it's a short function, maybe explicit const?

toolchain/check/handle_name.cpp
```diff
@@ -136,18 +146,28 @@ auto HandleNameExpression(Context& context, Parse::Node parse_node) -> bool {
...
-      auto [parse_node1, node_id1] =
-          context.node_stack().PopExpressionWithParseNode();
-      context.declaration_name_stack().ApplyExpressionQualifier(
-          parse_node1, context.FollowNameReferences(node_id1));
-      // Add the QualifiedDeclaration so that it can be used for bracketing.
-      context.node_stack().Push(parse_node);
-    } else {
-      // Nothing to do: the QualifiedDeclaration remains as a bracketing node
-      // for later QualifiedDeclarations.
+    Parse::Node parse_node1 = context.node_stack().PeekParseNode();
```
jonmeow: With your changes, it looks like this lambda is only called in one place. Does this need to remain a lambda?

toolchain/parse/handle_declaration_name_and_params.cpp
```diff
@@ -17,10 +17,7 @@ static auto HandleDeclarationNameAndParams(Context& context, State after_name)
     context.PushState(state);
 
     if (context.PositionIs(Lex::TokenKind::Period)) {
-      // Because there's a qualifier, we process the first segment as an
-      // expression for simplicity. This just means semantics has one less thing
-      // to handle here.
-      context.AddLeafNode(NodeKind::NameExpression, *identifier);
+      context.AddLeafNode(NodeKind::Name, *identifier);
```
jonmeow, line 20: Hmm, thanks for the cleanup. I don't remember really why it was the other way, maybe it's a design choice that predated the name stack.

toolchain/check/handle_name.cpp
```diff
@@ -157,10 +177,8 @@ auto HandleQualifiedDeclaration(Context& context, Parse::Node parse_node)
     pop_and_apply_first_child();
     context.declaration_name_stack().ApplyNameQualifier(parse_node2, name_id2);
   } else {
-    SemIR::NodeId node_id2 = context.node_stack().PopExpression();
-    pop_and_apply_first_child();
-    context.declaration_name_stack().ApplyExpressionQualifier(parse_node2,
-                                                              node_id2);
+    CARBON_FATAL() << "Unexpected node kind on right side of qualified "
```
jonmeow: The `Pop` on line 176 includes a CHECK. As a consequence, you should be able to remove the `if` and this `FATAL` and get equivalent behavior without running `PeekParseNode` (replacing `Pop` with `PopWithParseNode`).

jonmeow (approved):
Looks great. I'm sorry I missed the NameExpression improvement before. Can you make sure to restructure HandleQualifiedDeclaration before merging? I think it's straightforward given your changes, just the flow could be more linear.

toolchain/check/declaration_name_stack.h
zygoloid: Added a TODO to the relevant place in the parser.

toolchain/check/context.cpp

toolchain/check/handle_name.cpp
zygoloid: Thanks, inlined and simplified.

toolchain/check/handle_name.cpp
zygoloid: Yes, thank you :) That's much nicer.

PR #3330, opened 2023-10-24
Refactor SemIR YAML printing to use dashed lists.

jonmeow:
This standardizes on having ValueStore and related structures provide printing, removing the handlers in file.cpp.

The `[]` is provided for empty sequences versus if there was simply nothing, in which case it would be a sequence when non-empty, and a null value when empty. Consistently (and explicitly) providing sequences feels easier to understand.

The changes to the output yaml are overall more terse. My hope is that this is an improvement for most readers.

Also fixes printing of APInt, defaulting to unsigned for consistency with Carbon's use.

### DIFF
--- a/toolchain/base/value_store.h
+++ b/toolchain/base/value_store.h
@@ -12,6 +12,7 @@
 #include "llvm/ADT/APInt.h"
 #include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/StringExtras.h"
 #include "llvm/Support/YAMLParser.h"
 #include "toolchain/base/index_base.h"
 
@@ -84,7 +85,61 @@ namespace Internal {
 // Used as a parent class for non-printable types. This is just for
 // std::conditional, not as an API.
 class ValueStoreNotPrintable {};
+
+// Provides YAML printing of a value.
+template <typename ValueT>
+inline auto PrintValue(llvm::raw_ostream& out, const ValueT& val) {
+  out << val;
+};
+template <>
+inline auto PrintValue(llvm::raw_ostream& out, const llvm::APInt& val) {
+  val.print(out, /*isSigned=*/false);
+};
+template <>
+inline auto PrintValue(llvm::raw_ostream& out, const llvm::StringRef& val) {
+  out << "\"" << llvm::yaml::escape(val) << "\"";
+};
+
 }  // namespace Internal
+
+// Provides YAML printing of a list. The first line indent applies even if there
+// is no label, in which case it applies to the first element.
+template <typename ValueT>
+inline auto PrintValueRange(
+    llvm::raw_ostream& out, llvm::iterator_range<const ValueT*> range,
+    std::optional<llvm::StringRef> label, int first_line_indent,
+    int later_indent, bool trailing_newline,
+    std::function<void(llvm::raw_ostream&, const ValueT& val)> print =
+        Internal::PrintValue<ValueT>) {
+  out.indent(first_line_indent);
+  if (label) {
+    out << *label << ":";
+    if (range.empty()) {
+      // Add a space between the `:` and the `[]` printed below.
+      out << " ";
+    } else {
+      out << "\n";
+      out.indent(later_indent);
+    }
+  }
+  if (range.empty()) {
+    out << "[]";
+    if (trailing_newline) {
+      out << "\n";
+    }
+    return;
+  }
+  std::string sep_str = "\n";
+  sep_str.append(later_indent, ' ');
+  llvm::ListSeparator sep(sep_str);
+  for (const auto& val : range) {
+    out << sep << "- ";
+    print(out, val);
+  }
+  if (trailing_newline) {
+    out << "\n";
+  }
+}
 
 // A simple wrapper for accumulating values, providing IDs to later retrieve the
 // value. This does not do deduplication.
@@ -94,6 +149,8 @@ class ValueStore
                               Printable<ValueStore<IdT, ValueT>>,
                               Internal::ValueStoreNotPrintable> {
  public:
+  using PrintFn = std::function<void(llvm::raw_ostream&, const ValueT& val)>;
+
   // Stores the value and returns an ID to reference it.
   auto Add(ValueT value) -> IdT {
     IdT id = IdT(values_.size());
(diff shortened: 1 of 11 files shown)


toolchain/base/value_store.h
```diff
@@ -84,8 +85,62 @@ namespace Internal {
...
+  out << val;
+};
+template <>
+inline auto PrintValue(llvm::raw_ostream& out, const llvm::APInt& val) {
+  val.print(out, /*isSigned=*/false);
+};
+template <>
+inline auto PrintValue(llvm::raw_ostream& out, const llvm::StringRef& val) {
+  out << "\"" << llvm::yaml::escape(val) << "\"";
+};
```
zygoloid: Do these need to be template specializations? It's generally advisable to prefer overloads instead of specializations wherever possible. (But perhaps this is supposed to be emulating an interface and impls in Carbon?)

toolchain/base/value_store.h
```diff
@@ -84,8 +85,62 @@ namespace Internal {
...
 
+// Provides YAML printing of a list. The first line indent applies even if there
+// is no label, in which case it applies to the first element.
+template <typename ValueT>
+inline auto PrintValueRange(
+    llvm::raw_ostream& out, llvm::iterator_range<const ValueT*> range,
+    std::optional<llvm::StringRef> label, int first_line_indent,
+    int later_indent, bool trailing_newline,
+    std::function<void(llvm::raw_ostream&, const ValueT& val)> print =
+        Internal::PrintValue<ValueT>) {
```
zygoloid: Using a `std::function` here feels a bit heavy / inefficient. We probably don't care much about efficiency for our YAML printing in general, but it looks like it'd be straightforward to switch to `llvm::function_ref` here, so I think we should do so as a general good practice, in the same way that we're passing in a `llvm::StringRef` instead of a `std::string`.

toolchain/base/value_store.h
zygoloid: OK, I see below what you're using this for. I think I would like to suggest:

- Turn these into overloads.
- Add a `DefaultPrinter` class:

```c++
struct DefaultPrinter {
  template<typename ValueT>
  void operator()(llvm::raw_ostream& out, const ValueT& value) {
    PrintValue(out, value);
  }
};
```

or if you prefer not to name the type:

```c++
inline constexpr auto default_printer =
    [](llvm::raw_ostream& out, const auto& value) {
      PrintValue(out, value);
    };
```

- Below, either pass in `DefaultPrinter()` (or the lambda) as the default:

```c++
template <typename ValueT>
inline auto PrintValueRange(
    // ...
    llvm::function_ref<void(llvm::raw_ostream&, const ValueT& val)> print =
        Internal::DefaultPrinter()) {
```

or allow a function object to be passed in and called directly (this is a bit more awkward if you use the lambda approach):

```c++
template <typename ValueT, typename PrintT = Internal::DefaultPrinter>
inline auto PrintValueRange(
    // ...
    PrintT print = {}) {
```

This would remove the need for the `std::decay_t` below, and allow use of overloads rather than function template specializations here, as well as permitting `PrintValue` to be an ADL extension point if we want.

toolchain/base/value_store.h
```diff
@@ -125,12 +182,18 @@ class ValueStore
...
+  auto Print(llvm::raw_ostream& out, std::optional<llvm::StringRef> label,
+             int first_line_indent, int later_indent,
+             // This decays so that `const llvm::APInt` printing catches the
+             // specialization.
+             PrintFn print = Internal::PrintValue<std::decay_t<ValueT>>) const
+      -> void {
+    PrintValueRange(out, llvm::iterator_range(values_), label,
+                    first_line_indent, later_indent, /*trailing_newline=*/true,
+                    print);
   }
```
zygoloid: Is it worth factoring this out to a CRTP base class? We seem to do the same thing three times so far, and I expect there'll be more as we add more value stores. I suppose we'll need to add an accessor to get the value range, but given that you're only supporting an `iterator_range<const ValueT*>` right now, it seems like we could use the `array_ref` accessor for now at least.

toolchain/base/value_store.h
```diff
@@ -188,12 +258,10 @@ class SharedValueStores : public Printable<SharedValueStores> {
...
-        << "  - integers:\n";
-    integers_.Print(out, 6);
-    out << "  - reals:\n";
-    reals_.Print(out, 6);
-    out << "  - strings:\n";
-    strings_.Print(out, 6);
+        << "  - ";
+    integers_.Print(out, "integers", 0, 6);
+    reals_.Print(out, "reals", 4, 6);
+    strings_.Print(out, "strings", 4, 6);
```
zygoloid, line 269: These magic numbers make me think that we could model this better -- maybe the entry should not be responsible for printing out its own first-line prefix, and maybe we could track the indent width automatically somehow, ...

Actually, maybe we should switch over to using the `llvm::yaml` machinery to do YAML printing?

toolchain/check/testdata/basics/raw_and_textual_ir.carbon
```diff
@@ -15,92 +15,67 @@ fn Foo(n: i32) -> (i32, f64) {
...
+// CHECK:STDOUT:         - node+14
+// CHECK:STDOUT:         - node+15
+// CHECK:STDOUT:         - node+16
+// CHECK:STDOUT:         - node+17
+// CHECK:STDOUT:       - - node+9
+// CHECK:STDOUT:         - node+10
+// CHECK:STDOUT:       - - node+13
+// CHECK:STDOUT:         - node+15
+// CHECK:STDOUT:       - - node+5
+// CHECK:STDOUT:         - node+6
```
zygoloid, line 78: Looking at this I wonder if a mapping of sequences would be clearer than a sequence of sequences when printing a value store. Eg:

```yaml
nodes:
  node+0: {kind: Parameter, arg0: str1, type: type0}
  node+1: {kind: TupleType, arg0: typeBlock0, type: typeTypeType}
  ...
node_blocks:
  block0: []
  block1: - node+0
  block2: - nodeIntegerTyype
          - nodeFloatingPointType
  ...
```

zygoloid (commented):
I think the big question for me here is, is it time to move over to using a mechanism designed to print YAML? We seem to be slowly building up an ad-hoc YAML printing infrastructure, but LLVM already has one, so maybe we can just use that instead.

toolchain/base/value_store.h
jonmeow: The implementations diverge in a way that would, IIUC, require a CRTP to be parameterized if it were to be used (they're close but not identical). I'm maybe more hesitant to adopt CRTP, the function calls feel easy to understand.

toolchain/base/value_store.h
jonmeow: Done, thanks.

toolchain/base/value_store.h
jonmeow: Done. I think I just don't use `std::function` enough to internalize using `llvm::function_ref` instead.

toolchain/check/testdata/basics/raw_and_textual_ir.carbon
jonmeow: Separating out this change for the YAML work.

zygoloid (approved):
I'm fine with this as-is; the remaining comments all seem like things we can defer. I know you're looking into switching to use a different interface for YAML formatting, so I'm not hitting the merge button in case this PR will cause you conflicts.

jonmeow:
> I'm fine with this as-is; the remaining comments all seem like things we can defer. I know you're looking into switching to use a different interface for YAML formatting, so I'm not hitting the merge button in case this PR will cause you conflicts.

I'll take this to merge, and use a separate PR for YAML. It's a little bit of churn, but I think the output will be similar enough that maybe it's easiest to review that way since you're okay with this as-is (and the YAML is taking a bit of time to figure out approaches).

toolchain/base/value_store.h
jonmeow: Separating out this change.

PR #3331, opened 2023-10-24
Make SemIR::File access more terse.

jonmeow:
1. In general, `semantics_ir` -> `sem_ir`, to match the directory name.
2. For the list of `ValueStore`-related accessors on `SemIR::File`, add them to `check`'s `Context` object, shortening access.

### DIFF
--- a/toolchain/check/context.cpp
+++ b/toolchain/check/context.cpp
@@ -21,17 +21,17 @@
 namespace Carbon::Check {
 
 Context::Context(const Lex::TokenizedBuffer& tokens, DiagnosticEmitter& emitter,
-                 const Parse::Tree& parse_tree, SemIR::File& semantics_ir,
+                 const Parse::Tree& parse_tree, SemIR::File& sem_ir,
                  llvm::raw_ostream* vlog_stream)
     : tokens_(&tokens),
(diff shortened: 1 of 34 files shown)


toolchain/check/context.cpp
```diff
@@ -75,7 +75,7 @@ auto Context::DiagnoseDuplicateName(Parse::Node parse_node,
                     "Duplicate name being declared in the same scope.");
   CARBON_DIAGNOSTIC(NameDeclarationPrevious, Note,
                     "Name is previously declared here.");
-  auto prev_def = semantics_ir_->nodes().Get(prev_def_id);
+  auto prev_def = sem_ir_->nodes().Get(prev_def_id);
```
zygoloid:
```suggestion
  auto prev_def = nodes().Get(prev_def_id);
```

There are a bunch more of these; looks like the ones of the form `sem_ir_->` didn't get caught, only the ones of the form `sem_ir().`?

zygoloid (approved):

toolchain/check/context.cpp
jonmeow: Fixed `sem_ir_` cases.

PR #3332, opened 2023-10-24
Add a `Field` node for fields in a class.

zygoloid:
This replaces the use of `VarStorage` in this case.

Add an `UnboundFieldType` type as the type of a field, in cases where it's referenced without an accompanying object.

Add a `BindName` node to describe the name binding performed for both variables and fields so that we can handle them more uniformly.


toolchain/sem_ir/formatter.cpp
```diff
@@ -386,11 +386,12 @@ class NodeNamer {
...
       };
-      auto add_node_name_id = [&](StringId name_id) {
+      auto add_node_name_id = [&](StringId name_id,
+                                  llvm::StringRef suffix = "") {
         if (name_id.is_valid()) {
-          add_node_name(sem_ir_.strings().Get(name_id).str());
+          add_node_name((sem_ir_.strings().Get(name_id).str() + suffix).str());
         } else {
-          add_node_name("");
+          add_node_name(suffix.str());
```
jonmeow, line 393: I'm trying to find this in case run in a test (where only a suffix is printed), can you point me in the right direction?

toolchain/check/handle_file.cpp
```diff
@@ -0,0 +1,21 @@
...
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "toolchain/check/context.h"
+
+namespace Carbon::Check {
+
+auto HandleFileStart(Context& context, Parse::Node parse_node) -> bool {
+  // Push the file as a sentinel so that it's always safe to peek at the
+  // enclosing node on the node stack to determine what context we're in.
```
jonmeow: I worry about how `Peek` takes arbitrary values, but you mainly use `Peek(1)`, and so you add a sentinel value for `Peek(1)`. But what if someone starts calling `Peek(2)`, should another sentinel value get added?

toolchain/check/handle_pattern_binding.cpp
```diff
@@ -29,29 +29,52 @@ auto HandlePatternBinding(Context& context, Parse::Node parse_node) -> bool {
...
+          SemIR::BindName{name_node, value_type_id, name_id, value_id});
       break;
+    }
 
     case Parse::NodeKind::ParameterListStart:
       // Parameters can have incomplete types in a function declaration, but not
       // in a function definition. We don't know which kind we have here.
       context.AddNodeAndPush(
           parse_node, SemIR::Parameter{name_node, cast_type_id, name_id});
+      // TODO: Create a `BindName` node.
```
jonmeow, line 75: Parameters seem a bit peculiar versus variables; doesn't the Parameter have the same information that a BindName would?

jonmeow (commented):
Comparing with `GetDeclarationContext`, is there overlap in what you're trying to do with `Peek`s?

Put differently, I'm not sure that `Peek(1)` is guaranteed to work the way you're expecting: for example, a `VariableIntroducer` might be wrapped with access control. Is `GetDeclarationContext` an option for what you're trying to do, given it's supposed to resolve to the `class` context?

toolchain/check/handle_file.cpp
zygoloid: We should only need a single sentinel. Think of this like null-termination: if we know that level N is not `FileStart`, then it's safe to peek at level N+1. (And if you don't know what level N is, you shouldn't be looking at level N+1, because it wouldn't in general be correct to skip past level N without knowing what it is.)

toolchain/check/handle_pattern_binding.cpp
zygoloid: If we support any kind of non-trivial pattern matching in parameters, then they're different, eg:
```carbon
fn F(a: i32, (b: i32, c: i32));
```
would have two `Parameter`s but three `BindName`s. `var` parameters and possibly variadics might also be interesting cases here. Also, it seems more consistent to always create `BindName` nodes for all pattern name bindings, and that may help us avoid special cases, but that's speculation on my part.

But really I'm just moving this `TODO` to here from above because it now only applies to this particular case and not to variables any more.

toolchain/sem_ir/formatter.cpp
zygoloid: I've tried to create such a test case, but I don't think the case where the name is invalid and a suffix is present is currently reachable -- we happen to never create the nodes in question with an invalid `NameId` -- so we'll always end up using the name `""` here.

zygoloid:
> Comparing with `GetDeclarationContext`, is there overlap in what you're trying to do with `Peek`s?
> Put differently, I'm not sure that `Peek(1)` is guaranteed to work the way you're expecting: for example, a `VariableIntroducer` might be wrapped with access control. Is `GetDeclarationContext` an option for what you're trying to do, given it's supposed to resolve to the `class` context?

Mm, yeah. I think access control might be OK, because it'll be a parent of the `var` in the parse tree and hence be visited after the `var` parse tree nodes, but something like an attribute would push something onto the node stack between the class and the `VariableIntroducer`.

I don't think `GetDeclarationContext` helps much -- I could use `GetDeclarationContext` during parsing to create a different parse node (or just have different parse rules for parsing a class body), and create different parse nodes for variables and fields. But during checking, I want to be able to access data associated with the enclosing class scope -- in particular, I need to be able to get to the `ClassId` -- and knowing I have a field isn't enough for that.

My new plan is to track an associated node for the scope on the scope stack, and use that to determine if we're declaring a field, and if so, what the associated class is.

toolchain/check/handle_file.cpp
zygoloid: I'm not using `Peek(n)` any more. I think the sentinel still makes sense though, to ensure that our `Peek()` calls are safe. Let me know if you'd prefer that I remove it again; I don't need it in this PR any more.

zygoloid:
> My new plan is to track an associated node for the scope on the scope stack, and use that to determine if we're declaring a field, and if so, what the associated class is.

Done.

toolchain/sem_ir/formatter.cpp
jonmeow: Is that changing, or expected behavior? Could a CHECK note the behavior?

toolchain/check/handle_file.cpp
jonmeow: My leaning would be to revert it as unused, since we may not need it. Note for null termination we typically use `StringRef` 😉 .

But it may be that directly checking `N+1` would be incrementally faster than bounds-checking, and if needed, maybe it makes sense on that end. My leaning would still be wait until it's helpful, I'm generally hesitant about the `Peek(N)` model, if it's the best/flexible fit.

jonmeow:
> I don't think `GetDeclarationContext` helps much -- I could use `GetDeclarationContext` during parsing to create a different parse node (or just have different parse rules for parsing a class body), and create different parse nodes for variables and fields. But during checking, I want to be able to access data associated with the enclosing class scope -- in particular, I need to be able to get to the `ClassId` -- and knowing I have a field isn't enough for that.

Oops, I had it in my head that `GetDeclarationContext` was available on semantics when it's currently just parsing.

> My new plan is to track an associated node for the scope on the scope stack, and use that to determine if we're declaring a field, and if so, what the associated class is.

This looks good, I think it's offering similar functionality which is probably what we'll want.

jonmeow (approved):
Generally looks good, just a couple last things for you to consider before merging.

toolchain/sem_ir/formatter.cpp
zygoloid: I would expect that we will have cases in the future where the name is invalid. Node naming is best-effort, so falling back to an empty string / just the suffix in the case where we don't find a good name seems reasonable to me. Also, generally I've tried to make the formatter be robust against finding invalid IDs even in places where we wouldn't usually expect them, so that the output can still be used for inspecting the result of possibly broken checker behavior.

toolchain/check/handle_file.cpp
zygoloid: The sentinel could also be useful for our current uses of `Peek`, which can fail if they happen at the top level. But I've looked through the things that we can handle as a top level declaration, and all of them push something before they do any `Peek`s, so I think all our current `Peek`s are fine.

Removed this push / pop, at least for now. Should be easy to add it back if we find we have a need later.

PR #3333, opened 2023-10-24
sem-ir: Write references to the `file` block as `file.` not `package.`

zygoloid:
This matches the renaming of the `package { ... }` block to `file { ... }`.


jonmeow (approved):
Looks good, yeah that should probably match.

PR #3334, opened 2023-10-24
Basic lowering support for classes.

zygoloid:
Track the fields in a class, and generate a corresponding struct type as the object representation for the class. For now, we always use a pointer as the value representation for a class.


jonmeow:
Just to be sure, is the appropriate commit range to review here d7e5c22 ("Build an object representation when a class becomes complete.") through e45394c ("Add test coverage of class value and initializing representation.")?

It'd be helpful if the eponymous commit were a starting point for review. I'm having to examine commit logs here and on the other PR just to figure out the scope of review.

jonmeow:
Hmmm, when I try viewing a commit range, I get "We went looking everywhere, but couldn’t find those commits." (I'm going to pause here, since d7e5c22 seems to be having issues)

zygoloid:
> It'd be helpful if the eponymous commit were a starting point for review. I'm having to examine commit logs here and on the other PR just to figure out the scope of review.

Sorry about that, I'll try to follow that pattern in future. Dependent commit is now merged and this has been rebased.

toolchain/lower/file_context.cpp
```diff
@@ -225,6 +225,13 @@ auto FileContext::BuildType(SemIR::NodeId node_id) -> llvm::Type* {
           GetType(array_type.element_type_id),
           sem_ir_->GetArrayBoundValue(array_type.bound_id));
     }
+    case SemIR::ClassType::Kind: {
+      auto object_representation_id =
+          sem_ir_->classes()
+              .Get(node.As<SemIR::ClassType>().class_id)
+              .object_representation_id;
```
jonmeow, line 232: Stray thought, not specific to this change. If you're writing code like this frequently, you could add to the classes value store (needing a wrapper class) something like:

```
auto Get(Node node) -> Class { return Get(node.As<SemIR::ClassType>().class_id); }
```

toolchain/sem_ir/file.h
```diff
@@ -62,6 +62,9 @@ struct Class : public Printable<Class> {
     out << "}";
   }
 
+  // Determine whether this class has been fully defined.
```
jonmeow: You might comment here that this is false when definition_id is valid while the definition is being processed (maybe holistically document the transition states for members that start as invalid?)

jonmeow (approved):

toolchain/sem_ir/file.h
zygoloid: Expanded the comments to say when each member is set, and to clarify that this determines whether we've reached the `}`.

toolchain/lower/file_context.cpp
zygoloid: Mm, something to think about. Sometimes we have a `ClassDeclaration` instead of a `ClassType`, though, so maybe we'd want `Get(Node)` to handle both cases. But then that means adding a branch where we don't have one today.

PR #3335, opened 2023-10-25
Support for member access into classes. 

zygoloid:
Add support for member access into classes, for both non-instance
members and for fields.


toolchain/lower/handle.cpp
```diff
@@ -166,6 +166,102 @@ auto HandleClassDeclaration(FunctionContext& /*context*/,
   // No action to perform.
 }
 
+// Extracts an element of either a struct or a tuple by index. Depending on the
+// expression category of the aggregate input, this will either produce a value
+// or a reference.
+static auto GetStructOrTupleElement(FunctionContext& context,
+                                    SemIR::NodeId aggr_node_id, unsigned idx,
```
jonmeow: `unsigned` -> `SemIR::MemberIndex`?

Note `unsigned` is slightly awkward here in part because the value is actually `int32_t`.

toolchain/check/handle_name.cpp
```diff
@@ -55,14 +55,72 @@ auto HandleMemberAccessExpression(Context& context, Parse::Node parse_node)
     return true;
   }
 
+  // If the base isn't a scope, it must have a complete type.
+  auto base_type_id = context.nodes().Get(base_id).type_id();
+  if (!context.TryToCompleteType(base_type_id, [&] {
+        CARBON_DIAGNOSTIC(IncompleteTypeInMemberAccess, Error,
+                          "Base of member access has incomplete type `{0}`.",
```
jonmeow: I'm not quite sure what "base" means here. Maybe "Member access into incomplete type" or "Cannot access member of incomplete type"?

toolchain/check/context.cpp
```diff
@@ -620,6 +644,7 @@ class TypeCompleter {
       case SemIR::BranchIf::Kind:
       case SemIR::BranchWithArg::Kind:
       case SemIR::Call::Kind:
+      case SemIR::ClassFieldAccess::Kind:
       case SemIR::ClassDeclaration::Kind:
```
```suggestion
      case SemIR::ClassDeclaration::Kind:
      case SemIR::ClassFieldAccess::Kind:
```

toolchain/check/context.h
```diff
@@ -96,6 +96,10 @@ class Context {
   // Follows NameReference nodes to find the value named by a given node.
   auto FollowNameReferences(SemIR::NodeId node_id) -> SemIR::NodeId;
 
+  // Gets the constant value of the given node, if it has one.
+  // TODO: This currently handles very few cases.
```
```suggestion
```

Maybe leave this for the cpp file, which has a similar TODO?

jonmeow (approved):
Looks good.

toolchain/check/handle_name.cpp
zygoloid: Done.

toolchain/lower/handle.cpp
zygoloid: Done.

zygoloid:
I realized that this was incorrectly treating class fields as having their value representation when performing field access on a class value. That's fixed now, but the changes are substantial enough that I'd appreciate another round of review.

toolchain/sem_ir/file.h
```diff
@@ -107,10 +107,33 @@ struct ValueRepresentation : public Printable<ValueRepresentation> {
...
+  // The kind of aggregate representation used by this type.
+  AggregateKind aggregate_kind = AggregateKind::NotAggregate;
   // The type used to model the value representation.
   TypeId type_id = TypeId::Invalid;
+
+  // Returns whether this is an aggregate that holds its elements by value.
+  auto elements_are_values() const {
+    return aggregate_kind == ValueAggregate ||
+           aggregate_kind == ValueAndObjectAggregate;
+  }
```
jonmeow: Sorry, stray comment -- declaration order should have this above members.

zygoloid: Fixed.

PR #3336, opened 2023-10-25
Rename Deduced parameters to Implicit parameters.

zygoloid:
In preparation for supporting `self`, which is implicit but not deduced.


jonmeow (approved):
Approving, but per discussion, it'd be nice if the design documentation named these explicitly as implicit parameters...

PR #3337, opened 2023-10-25
Refactor YAML handling to use the llvm::yaml API.

jonmeow:
Provides an adapter for the llvm::yaml API because it otherwise needs a bunch of const/non-const definitions, and the traits are difficult to diagnose issues with. The current approach is pretty simple to use, even if it's not super efficient (which, yaml output is more of a debugging thing so I'm not really expecting it to be an issue).

Changes the format of yaml output to provide more index information, just as reminders when seeing something like `node+0`. Note this would create more churn in deltas if we were reliant on the output yaml in tests, but we aren't so it should be okay.

### DIFF
--- a/toolchain/base/value_store.h
+++ b/toolchain/base/value_store.h
@@ -88,19 +88,6 @@ namespace Internal {
 // Used as a parent class for non-printable types. This is just for
 // std::conditional, not as an API.
 class ValueStoreNotPrintable {};
-
-// Provides YAML printing of a value.
-template <typename ValueT>
-inline auto PrintValue(llvm::raw_ostream& out, const ValueT& val) {
-  out << val;
-};
-inline auto PrintValue(llvm::raw_ostream& out, const llvm::APInt& val) {
-  val.print(out, /*isSigned=*/false);
-};
-inline auto PrintValue(llvm::raw_ostream& out, const llvm::StringRef& val) {
-  out << "\"" << llvm::yaml::escape(val) << "\"";
-};
-
 }  // namespace Internal
 
 // A simple wrapper for accumulating values, providing IDs to later retrieve the
@@ -149,10 +136,7 @@ class ValueStore
     return Yaml::OutputMapping([&](Yaml::OutputMapping::Map map) {
       for (auto i : llvm::seq(values_.size())) {
         auto id = IdT(i);
-        map.Add(PrintToString(id),
-                Yaml::OutputScalar([&](llvm::raw_ostream& out) {
-                  Internal::PrintValue(out, Get(id));
-                }));
+        map.Add(PrintToString(id), Yaml::OutputScalar(Get(id)));
       }
     });
   }
--- a/toolchain/base/value_store.h
+++ b/toolchain/base/value_store.h
@@ -12,10 +12,13 @@
 #include "llvm/ADT/APInt.h"
 #include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/STLFunctionalExtras.h"
+#include "llvm/ADT/Sequence.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/StringExtras.h"
 #include "llvm/Support/YAMLParser.h"
+#include "llvm/Support/YAMLTraits.h"
 #include "toolchain/base/index_base.h"
+#include "toolchain/base/yaml.h"
 
 namespace Carbon {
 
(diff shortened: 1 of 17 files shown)


toolchain/base/value_store.h
```diff
@@ -189,17 +146,17 @@ class ValueStore
...
+    return Yaml::OutputMapping([&](llvm::yaml::IO& io) {
+      for (auto i : llvm::seq(values_.size())) {
+        auto id = IdT(i);
+        Yaml::OutputMapping::Map(
+            io, PrintToString(id),
+            Yaml::OutputScalar([&](llvm::raw_ostream& out) {
+              Internal::PrintValue<ValueT>(out, Get(id));
+            }));
+      }
+    });
```
zygoloid: Taking inspiration from the command-line arguments library, I wonder if instead of passing an `IO&` in here, we could pass in an object describing the mapping, so we could instead write this as:

```suggestion
    return Yaml::OutputMapping([&](auto& map) {
      for (auto i : llvm::seq(values_.size())) {
        auto id = IdT(i);
        map.Add(
            PrintToString(id),
            Yaml::OutputScalar([&](llvm::raw_ostream& out) {
              Internal::PrintValue<ValueT>(out, Get(id));
            }));
      }
    });
```

toolchain/base/value_store.h
```diff
@@ -189,17 +146,17 @@ class ValueStore
...
-                    first_line_indent, later_indent, /*trailing_newline=*/true,
-                    print);
+  auto OutputYaml() const -> Yaml::OutputMapping {
+    return Yaml::OutputMapping([&](llvm::yaml::IO& io) {
+      for (auto i : llvm::seq(values_.size())) {
+        auto id = IdT(i);
+        Yaml::OutputMapping::Map(
+            io, PrintToString(id),
+            Yaml::OutputScalar([&](llvm::raw_ostream& out) {
+              Internal::PrintValue<ValueT>(out, Get(id));
```
zygoloid:
```suggestion
              Internal::PrintValue(out, Get(id));
```

toolchain/base/value_store.h
```diff
@@ -189,17 +146,17 @@ class ValueStore
...
-             int first_line_indent, int later_indent,
-             // This decays so that `const llvm::APInt` printing catches the
-             // specialization.
-             PrintFn print = Internal::DefaultPrinter()) const -> void {
-    PrintValueRange(out, llvm::iterator_range(values_), label,
-                    first_line_indent, later_indent, /*trailing_newline=*/true,
-                    print);
+  auto OutputYaml() const -> Yaml::OutputMapping {
+    return Yaml::OutputMapping([&](llvm::yaml::IO& io) {
+      for (auto i : llvm::seq(values_.size())) {
```
zygoloid, line 137: Could we use
```suggestion
      for (auto [i, value] : llvm::enumerate(values_.size())) {
```
instead of the `Get(id)` call below?

toolchain/base/yaml.h
```diff
@@ -0,0 +1,102 @@
...
+  // Maps a value. This mainly takes responsibility for copying the value,
+  // letting mapRequired take `&value`.
+  template <typename T>
+  static auto Map(llvm::yaml::IO& io, llvm::StringRef key, T value) -> void {
+    io.mapRequired(key.data(), value);
+  }
+  static auto Map(llvm::yaml::IO& io, llvm::StringRef key,
+                  const llvm::StringRef value) -> void {
+    llvm::StringRef v = value;
+    io.mapRequired(key.data(), v);
```
zygoloid:
```suggestion
                  llvm::StringRef value) -> void {
    io.mapRequired(key.data(), value);
```

jonmeow: Done, this must've been a leftover somehow.

toolchain/base/value_store.h
jonmeow: Done

toolchain/base/value_store.h
jonmeow: Sticking with `Get(id)` due to the "auto capture + lambda" not-C++20 problem.

toolchain/base/value_store.h
jonmeow: Done (though I'm hesitant about `auto` there)

zygoloid (approved):

jonmeow:
It looked like the mac build was failing because llvm::yaml's support for size_t is platform dependent (works on linux, not on maos). This sent me down a different path on printing scalars, though I'm happier with the new state which removes `Internal::PrintValue` and reduces explicit lambdas.

zygoloid (approved):

PR #3339, opened 2023-10-26
Disable field-initializers when building.

guille2718:
This allows building carbon with newer versions of clang. This makes the build aligned with the `compile_flags.txt` file, which also has this warning disabled.


jonmeow:
Thank you for reporting this to us!

Digging into this a little, there's a warning on anonymous union initialization, which @zygoloid reported to llvm as a false positive: https://github.com/llvm/llvm-project/issues/70384. Because of that issue, we're going to disable it for now, although just for clang-18, per #3342.

If the regression gets fixed we'll probably want to fix forward and re-enable the flag. I'm not sure how much work there'll be once false-positives are fixed.

Regarding `compile_flags.txt`, I'm actually not sure why it omits this warning, since we were using it in builds.

If #3342 isn't silencing this for you, please let me know.

guille2718:
#3342 works for me, thanks!

PR #3341, opened 2023-10-26
Distinguish between name IDs and string IDs in the type system.

zygoloid:
Add a `NameId` that is effectively just a wrapper around a `StringId`, with
some additional predefined values for names that don't correspond to
strings, such as the name of `self` or the function's return slot.


zygoloid:
For reviewing convenience, I've put the main changes and the removal of `"` around names in SemIR in separate commits, because the latter introduced a lot of noise into the tests.

toolchain/check/context.h
```diff
@@ -241,29 +242,17 @@ class Context {
...
+  auto reals() -> decltype(auto) { return sem_ir().reals(); }
+  auto strings() -> decltype(auto) { return sem_ir().strings(); }
+  auto functions() -> decltype(auto) { return sem_ir().functions(); }
+  auto classes() -> decltype(auto) { return sem_ir().classes(); }
+  auto names() -> decltype(auto) { return sem_ir().names(); }
+  auto name_scopes() -> decltype(auto) { return sem_ir().name_scopes(); }
+  auto types() -> decltype(auto) { return sem_ir().types(); }
+  auto type_blocks() -> decltype(auto) { return sem_ir().type_blocks(); }
+  auto nodes() -> decltype(auto) { return sem_ir().nodes(); }
+  auto node_blocks() -> decltype(auto) { return sem_ir().node_blocks(); }
```
jonmeow: Does this meet the style expectation of `auto` being used only to make the code clearer, not just to avoid writing the type? I'm concerned it may be a bit too indirect here.

Also, why is `decltype` needed?

Style reference: https://google.github.io/styleguide/cppguide.html#Type_deduction

toolchain/sem_ir/value_stores.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+auto NameStore::GetFormatted(NameId name_id) const -> llvm::StringRef {
+  // If the name is a string name with a keyword spelling, format it with an
+  // `r#` prefix. Return any other string name as-is.
+  if (auto string_name = GetAsString(name_id)) {
+    return llvm::StringSwitch<llvm::StringRef>(*string_name)
+#define CARBON_KEYWORD_TOKEN(Name, Spelling) .Case(Spelling, "r#" Spelling)
+#include "toolchain/lex/token_kind.def"
+        .Default(*string_name);
+  }
```
jonmeow: This means each diagnostic mention of an identifier will be compared to each keyword, right? How much should we care about the string comparisons?

toolchain/sem_ir/value_stores.h
```diff
@@ -43,6 +43,48 @@ class NodeStore {
...
+  auto GetAsStringId(NameId name_id) const -> std::optional<StringId> {
+    if (name_id.index >= 0) {
+      return StringId(name_id.index);
+    }
+    return std::nullopt;
+  }
+
+  // Returns the requested name as a string, if it is a string name. This
+  // returns std::nullopt for special names.
+  auto GetAsString(NameId name_id) const -> std::optional<llvm::StringRef> {
```
jonmeow: "GetAsString" is something I expect to return `std::string`. Maybe `GetIfInStrings`, `GetIfNotSpecial`, `GetFromStrings`, `Get`, or similar?

toolchain/sem_ir/value_stores.h
```diff
@@ -43,6 +43,48 @@ class NodeStore {
...
+class NameStore {
+ public:
+  NameStore(ValueStore<StringId>* strings) : strings_(strings) {}
+
+  // Adds a string name, returning an ID to reference it.
+  auto Add(StringId string_id) -> NameId { return NameId(string_id.index); }
+
+  // Returns the requested name as a string ID, if it is a string name. This
+  // returns std::nullopt for special names.
+  auto GetAsStringId(NameId name_id) const -> std::optional<StringId> {
```
jonmeow: I'm seeing this called from `GetAsString`, but would it be used otherwise? The extra `std::optional` wrapper versus using `StringId::Invalid` might be a bit of overhead that could be avoided.

toolchain/sem_ir/value_stores.h
```diff
@@ -43,6 +43,48 @@ class NodeStore {
...
 };
 
+// Provides a ValueStore-like interface for names. This is currently a wrapper
+// around a string store that has no state of its own.
+class NameStore {
+ public:
+  NameStore(ValueStore<StringId>* strings) : strings_(strings) {}
+
+  // Adds a string name, returning an ID to reference it.
+  auto Add(StringId string_id) -> NameId { return NameId(string_id.index); }
```
jonmeow: This feels a bit odd: it's not really adding things, it's just wrapping the type. Had you considered a constructor on NameId?

toolchain/sem_ir/value_stores.h
```diff
@@ -43,6 +43,48 @@ class NodeStore {
   ValueStore<NodeId, Node> values_;
 };
 
+// Provides a ValueStore-like interface for names. This is currently a wrapper
+// around a string store that has no state of its own.
+class NameStore {
```
jonmeow: Can you add comments here explaining the difference between this and the string store?

jonmeow (commented):
Can you share your thoughts on why you implemented this in SemIR versus making it more visible at the lexing level?

How are you expecting lexing to work? e.g., is lexing keeping the `r#` so that parsing can print the `r#` if diagnostics arise, or dropping them so that StringIds are unchanged for checking?

Note, I think that integrating `r#` earlier avoids that kind of choice arising, but also may remove the need for a separate `NameId` from `StringId` (may mainly mean the shared value stores might be better defined under `lex`, versus `base`).

Also, are the unusual cases tested? In particular, looking for tests that mention `<invalid>`, `<return slot>`, a `r#` name, etc.

toolchain/check/context.h
zygoloid: `decltype` is needed because `auto` doesn't deduce a reference type.

The use of `decltype(auto)` here is not to make the code clearer, it's to make the code safer, which the style guide also allows. `sem_ir().names()` currently returns by value, and we want `Context` to do the same thing, to avoid it returning a dangling reference. But if `sem_ir().names()` changes to return by reference, we want `Context` to also change, to avoid making an unintended copy. A `decltype(auto)`  return type means "I return exactly the same thing that the function I'm calling returns", which is what we mean here, and is safe against changes to `SemIR` in a way that an explicit type would not be.

toolchain/sem_ir/value_stores.cpp
zygoloid: `StringSwitch` is supposed to optimize relatively well, and it's what we use to identify keywords and operators in the lexer, where it doesn't seem to be a performance bottleneck (I'm assuming, based on Chandler having not changed it!). So I think this is probably OK, especially in a function that's used to generate diagnostics / user-visible output.

zygoloid:
> Can you share your thoughts on why you implemented this in SemIR versus making it more visible at the lexing level?

My thinking is: names are the keys in our name lookup table, so they're a semantic notion, not a lexical one. After lexing, we have tokens, such as identifiers, which have a spelling, but only once we've parsed them and determined what grammatical function they're serving would they be mapped into names. Having a notion of names at the lexing level feels like a layering mistake to me.

> How are you expecting lexing to work? e.g., is lexing keeping the `r#` so that parsing can print the `r#` if diagnostics arise, or dropping them so that StringIds are unchanged for checking?

My plan is for `TokenizedBuffer::GetIdentifier` to return a `StringId` that does not contain the `r#` prefix, because `foo` and `r#foo` are the same identifier (written in different ways), just like `"x"` and `#"x"#` are the same string literal token. `TokenizedBuffer::GetTokenText` would return the original spelling, with the `r#`.

> Note, I think that integrating `r#` earlier avoids that kind of choice arising, but also may remove the need for a separate `NameId` from `StringId` (may mainly mean the shared value stores might be better defined under `lex`, versus `base`).

I think the type system distinction between strings and names is valuable independent of how we handle `r#`. That's my primary motivation here.

> Also, are the unusual cases tested? In particular, looking for tests that mention `<invalid>`, `<return slot>`, a `r#` name, etc.

We don't support lexing `r#` identifiers yet, so I think that case is not reachable yet through a file_test. For the other cases you mention, I think the only time they would show up would be during error recovery, and I don't know off-hand of any situation that would result in either of those appearing.

I can add gunit tests if you think it's important that we cover those cases.

toolchain/sem_ir/value_stores.h
zygoloid: The intended abstraction here is that `NameStore` is a `ValueStore` for names, that happens to not need to actually store any state at all. I don't mind adding a `NameId` constructor, but it does break that abstraction a little bit by encoding into the API the fact that we can convert between `NameId`s and `StringId`s without a `NameStore`.

Would a different name for this function resolve your feeling of oddness, or do you think we should be more directly exposing the conversion between `NameId` and `StringId`?

toolchain/sem_ir/value_stores.h
zygoloid: Done, `GetIfIdentifier`, plus extra comments to explain identifier names versus special names.

toolchain/sem_ir/value_stores.h
zygoloid: Thanks, I ended up not needing this for anything. Removed.

toolchain/sem_ir/value_stores.h

jonmeow:
I think comments converge on two points; regarding [Add](https://github.com/carbon-language/carbon-lang/pull/3341#discussion_r1373912207) and [the PR-level comment](https://github.com/carbon-language/carbon-lang/pull/3341#issuecomment-1782100587).

What I'm concerned with is how cross-references between IRs work. In merging the `StringStore` as shared across compilation units, I was planning approach wherein equal `StringId` values could be used (directly) cross-IR for name lookup. The per-SemIR string storage was deliberately removed towards this purpose. I think this offers useful efficiencies, since it means one IR's name lookup table could be directly accessed.

In introducing `NameStore` as a `ValueStore`-equivalent per-IR storage of names, we're going back to the pre-`StringStore` model. Lookups of `NameId` -> `StringId` -> `NameId` may remain cheaper than full string comparisons. [But the `ValueStore` model encourages divergence, which could be costly for cross-IR name lookup.]

> "After lexing, we have tokens, such as identifiers, which have a spelling, but only once we've parsed them and determined what grammatical function they're serving would they be mapped into names."

I believe that since lexing has already decided they're identifiers, they are a name. What causes lexed identifiers and checked names to diverge? Maybe I'm missing something here which would make me understand the divergence between lexing/parsing and checking/lowering models.

Maybe a quick chat would help me understand your approach here?

> We don't support lexing `r#` identifiers yet, so I think that case is not reachable yet through a file_test. For the other cases you mention, I think the only time they would show up would be during error recovery, and I don't know off-hand of any situation that would result in either of those appearing.
> 
> I can add gunit tests if you think it's important that we cover those cases.

I think there should at least be TODOs, although it's making me wonder how much time `r#` support in lexing would take. Maybe I'll try whipping something together quickly.

zygoloid:
> What I'm concerned with is how cross-references between IRs work. In merging the `StringStore` as shared across compilation units, I was planning approach wherein equal `StringId` values could be used (directly) cross-IR for name lookup. The per-SemIR string storage was deliberately removed towards this purpose. I think this offers useful efficiencies, since it means one IR's name lookup table could be directly accessed.

I think it'd be fine to move the `NameStore` into the shared value stores. While I think the lexer shouldn't be generating `NameId`s, because it's operating at a different level of abstraction, having them shared between IRs produced by the same toolchain invocation seems fine to me.

(Mostly unrelated to this discussion: I think we'll still need a mechanism for mapping between `NameId`s or `StringId`s in different IRs produced by other compilations and imported.)

> Maybe a quick chat would help me understand your approach here?

Sure, let's talk in person next week. I don't think there's any urgency in this PR; it's not blocking the other work on classes.

jonmeow:
> > What I'm concerned with is how cross-references between IRs work. In merging the `StringStore` as shared across compilation units, I was planning approach wherein equal `StringId` values could be used (directly) cross-IR for name lookup. The per-SemIR string storage was deliberately removed towards this purpose. I think this offers useful efficiencies, since it means one IR's name lookup table could be directly accessed.
> 
> I think it'd be fine to move the `NameStore` into the shared value stores.

👍 

> While I think the lexer shouldn't be generating `NameId`s, because it's operating at a different level of abstraction, having them shared between IRs produced by the same toolchain invocation seems fine to me.

FWIW, I was viewing `NameId` as mapping directly to `Identifier` tokens, separating out from string literals. Happy to discuss this further though. 

> (Mostly unrelated to this discussion: I think we'll still need a mechanism for mapping between `NameId`s or `StringId`s in different IRs produced by other compilations and imported.)

Just to sketch out my thoughts here... The shared store isn't really "stored" in whole, meaning all shared ids would need to be remapped a bit. They probably need to be remapped on save in order to avoid gaps, for example. On load, they could probably store in the singleton shared store and either (a) update ids again or (b) have an adapter that handles the remapping on demand (likely by remapping a swathe of ids, not just 1-by-1, since we're likely to need other related info and probably don't want to remap each time). I think this is mostly an efficiency question, though.

jonmeow:
> > While I think the lexer shouldn't be generating `NameId`s, because it's operating at a different level of abstraction, having them shared between IRs produced by the same toolchain invocation seems fine to me.
> 
> FWIW, I was viewing `NameId` as mapping directly to `Identifier` tokens, separating out from string literals. Happy to discuss this further though.

I think this is resolved as a separate NameId.

> > (Mostly unrelated to this discussion: I think we'll still need a mechanism for mapping between `NameId`s or `StringId`s in different IRs produced by other compilations and imported.)
> 
> Just to sketch out my thoughts here... The shared store isn't really "stored" in whole, meaning all shared ids would need to be remapped a bit. They probably need to be remapped on save in order to avoid gaps, for example. On load, they could probably store in the singleton shared store and either (a) update ids again or (b) have an adapter that handles the remapping on demand (likely by remapping a swathe of ids, not just 1-by-1, since we're likely to need other related info and probably don't want to remap each time). I think this is mostly an efficiency question, though.

To note, I think this was mostly resolved in yesterday's discussion: the priority is a clean storage, and so we'll store per-IR and remap cross-IR.

toolchain/sem_ir/ids.h
```diff
@@ -112,6 +112,36 @@ struct BoolValue : public IndexBase, public Printable<BoolValue> {
...
 
+// The ID of a name. A name is either a string or a special name such as
+// `self`, or eventually `Self` or `base`.
+struct NameId : public IndexBase, public Printable<NameId> {
+  // An explicitly invalid ID.
+  static const NameId Invalid;
+  // The name of `self`.
+  static const NameId SelfValue;
+  // The name of the return slot in a function.
+  static const NameId ReturnSlot;
```
jonmeow: Should the constraints that this creates for potential negative values of StringId (or IdentifierId) be documented [on that type]?

toolchain/check/context.h
jonmeow: I'm not totally sold, given the style guide explicitly discourages deduced return types for public functions; if the copy constructor is a concern, it should possibly be deleted.

But, please change the matching ValueStore functions in file.h: I believe the rationale for or against using `auto` there should be equivalent, and directly related to your changes here.

toolchain/sem_ir/value_stores.h
jonmeow: A different name ("Wrap"?) would help, although maybe also renaming `NameStore` -> `Names` or similar may help too. As mentioned, it's actually the "ValueStore-like" aspect that I'm concerned about this being treated as a `ValueStore`.

If you're okay with #3352, you could also consider merging this with the `StringStore` there, just having `NameId` one more thing that it can toss around. Then, versus not doing "any additional explicit storage", it's merged with the rest in practice.

The converse of the above -- as I'm thinking more -- is if you don't want to merge with `StringStore` from #3352, and given the planned separation of `SharedValueStore` to be per-compilation-unit, then this probably belongs in `SemIR`. But I'd still then prefer this to be split a little more from `Store` objects which actually store their data, so that we don't end up with a new string store by accident.

toolchain/check/context.h
zygoloid: Switched back to an explicit return type.

toolchain/sem_ir/value_stores.h
zygoloid: Renamed to `NameStoreWrapper` to align with `StringStoreWrapper`.

toolchain/sem_ir/ids.h
zygoloid: Done.

toolchain/sem_ir/ids.h
```diff
@@ -112,6 +112,55 @@ struct BoolValue : public IndexBase, public Printable<BoolValue> {
...
+  // IdentifierId if this is a special name.
+  auto AsIdentifierId() -> IdentifierId {
+    return index >= 0 ? IdentifierId(index) : IdentifierId::Invalid;
+  }
+
+  using IndexBase::IndexBase;
+  auto Print(llvm::raw_ostream& out) const -> void {
+    out << "name";
+    if (*this == Invalid) {
+      out << "Invalid";
```
jonmeow: Why is `Invalid` printed this way versus the default `<invalid>` that `IndexBase::Print` would give?

toolchain/sem_ir/ids.h
```diff
@@ -112,6 +112,55 @@ struct BoolValue : public IndexBase, public Printable<BoolValue> {
...
+    return NameId(id.index);
+  }
+
+  // Returns the IdentifierId corresponding to this NameId, or an invalid
+  // IdentifierId if this is a special name.
+  auto AsIdentifierId() -> IdentifierId {
+    return index >= 0 ? IdentifierId(index) : IdentifierId::Invalid;
+  }
+
+  using IndexBase::IndexBase;
```
jonmeow: Declaration order: the constructor should probably be before the conversion.

jonmeow (approved):
Looks good! Just a couple small notes, and mind the test failure.

toolchain/sem_ir/ids.h
zygoloid: Fixed.

toolchain/sem_ir/ids.h
zygoloid: Done. Looks like the style guide wants this between the conversion from `IdentifierId` and the conversion to `IdentifierId` (factory functions, then constructors, then other methods), which seems a bit weird in this case.

PR #3342, opened 2023-10-26
Disable -Wmissing-field-initializers for clang-18 due to false positives.

jonmeow:
It warns on anonymous unions even when they have a field initialized. Reported at https://github.com/llvm/llvm-project/issues/70384

Depending on how it's fixed, we may be able to remove this. If it's fixed at head but still released in clang-18, we'd probably just change the conditional.

Reported by guille2718; this is a version-dependent approach from #3339 

Co-authored-by: Guillermo Rey <10690205+guille2718@users.noreply.github.com>


zygoloid (approved):

jonmeow:
FYI @guille2718, taking this as a variant from #3339 (I'll comment there too, for visibility)

PR #3343, opened 2023-10-27
Support for type-checking and lowering method calls.

zygoloid:
Adds a `BoundMethod` SemIR node to represent an `x.F` bound method, with a new builtin type `BoundMethodType`. Reorganized conversion of call expression arguments to also check and convert a `self` parameter in the implicit parameters list.

In passing, improved diagnostics and error recovery for bad call expressions. We now build a `call` node with the appropriate type and value category, but with invalid arguments, if the argument conversion failed, and diagnose calls to non-callable expressions.

`addr self` methods don't work properly yet; the `addr` is ignored for now.


toolchain/sem_ir/typed_nodes.h
```diff
@@ -129,6 +129,15 @@ struct BoolLiteral {
   BoolValue value;
 };
 
+struct BoundMethod {
+  static constexpr auto Kind = NodeKind::BoundMethod.Define("bound_method");
+
+  Parse::Node parse_node;
+  TypeId type_id;
+  NodeId object_id;
```
jonmeow, line 142: Minor question, I assume we'll use this for interfaces too; are interfaces objects, or would it be more precise to call this something like "bound_type_id"?

toolchain/check/handle_call_expression.cpp
```diff
@@ -2,33 +2,58 @@
...
-  if (!function_name) {
-    // TODO: Work on error.
-    context.TODO(parse_node, "Not a callable name");
+
+  auto diagnose_not_callable = [&, call_expr_parse_node = call_expr_parse_node,
+                                callee_id = callee_id] {
+    auto callee_type_id = context.nodes().Get(callee_id).type_id();
+    if (callee_type_id != SemIR::TypeId::Error) {
+      CARBON_DIAGNOSTIC(CallToNonCallable, Error,
+                        "Value of type {0} is not callable.", std::string);
```
jonmeow:
```suggestion
                        "Value of type `{0}` is not callable.", std::string);
```

jonmeow (approved):
Please make sure there's a test like:

```
class C {
  fn F[self: Self]() {}
}
fn Call() {
  // Can't call bound f
  C.F();
  // Might be curious if this happens to work, but really should fail for the same reason as above.
  C.F({});
}
```

I was looking around but not seeing this. Given you're getting into bound methods now, I want to make sure we have something.

toolchain/sem_ir/typed_nodes.h
zygoloid: This node represents `x.F` where the type of `x` is a class (or probably, eventually, a facet), not `Class.F` or `Interface.F`. So the value we're storing here is always the object argument -- the thing that will become `self` in the call, or `*self` for an `addr self` parameter.

Added some comments to explain.

zygoloid:
> Please make sure there's a test like:
> 
> ```
> class C {
>   fn F[self: Self]() {}
> }
> fn Call() {
>   // Can't call bound f
>   C.F();
>   // Might be curious if this happens to work, but really should fail for the same reason as above.
>   C.F({});
> }
> ```
> 
> I was looking around but not seeing this. Given you're getting into bound methods now, I want to make sure we have something.

The first part of that is already tested here: https://github.com/carbon-language/carbon-lang/blob/a9c845556e00143be05cda39a483d1a3820401bb/toolchain/check/testdata/class/fail_method.carbon#L17-L23

Added a test for the second part, but we happen to reject due to the explicit argument count not matching before we get to checking `self` has an argument.

jonmeow:
> > Please make sure there's a test like:
> > ```
> > class C {
> >   fn F[self: Self]() {}
> > }
> > fn Call() {
> >   // Can't call bound f
> >   C.F();
> >   // Might be curious if this happens to work, but really should fail for the same reason as above.
> >   C.F({});
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I was looking around but not seeing this. Given you're getting into bound methods now, I want to make sure we have something.
> 
> The first part of that is already tested here:
> 
> https://github.com/carbon-language/carbon-lang/blob/a9c845556e00143be05cda39a483d1a3820401bb/toolchain/check/testdata/class/fail_method.carbon#L17-L23
> 
> Added a test for the second part, but we happen to reject due to the explicit argument count not matching before we get to checking `self` has an argument.

Thanks, I was just not finding this test.

toolchain/sem_ir/typed_nodes.h
jonmeow: FWIW, I think this is still what I meant, but taking this as yes: an instance given an interface (class, facet, etc) would still be called an object.

PR #3345, opened 2023-10-27
Diagnose attempts to copy a non-copyable type.

zygoloid:
For now, we treat class types and `String` as non-copyable, because we don't know how to emit SemIR to copy them yet. This will change as we add support for copying those types when appropriate.


toolchain/check/convert.cpp
```diff
@@ -604,6 +604,35 @@ static auto PerformBuiltinConversion(Context& context, Parse::Node parse_node,
   return value_id;
 }
 
+// Given a value expression, form a corresponding initializer that copies from
+// that value, if it is possible to do so.
+auto PerformCopy(Context& context, SemIR::NodeId expr_id) -> SemIR::NodeId {
```
```suggestion
static auto PerformCopy(Context& context, SemIR::NodeId expr_id) -> SemIR::NodeId {
```

jonmeow (approved):
👍

PR #3346, opened 2023-10-27
Allow fields to be reordered in struct initialization.

zygoloid:


toolchain/check/convert.cpp
```diff
@@ -408,8 +408,8 @@ static auto ConvertStructToStruct(Context& context, SemIR::StructType src_type,
   }
 
   // Check that the structs are the same size.
-  // TODO: Check the field names are the same up to permutation, compute the
-  // permutation, and use it below.
+  // TODO: If not, include the name of the first source field that doesn't
+  // exist in the destination in the diagnostic.
```
jonmeow: Minor, but it could also be a dest field not in the source.

toolchain/check/convert.cpp
```diff
@@ -420,6 +420,15 @@ static auto ConvertStructToStruct(Context& context, SemIR::StructType src_type,
     return SemIR::NodeId::BuiltinError;
   }
 
+  // Prepare to look up fields in the source by index.
+  llvm::SmallDenseMap<StringId, int32_t> src_field_indexes;
```
jonmeow, line 424: The plural is indices. 😛

toolchain/check/convert.cpp
```diff
@@ -436,30 +445,48 @@ static auto ConvertStructToStruct(Context& context, SemIR::StructType src_type,
...
+      auto src_field_it = src_field_indexes.find(dest_field.name_id);
+      if (src_field_it == src_field_indexes.end()) {
+        CARBON_DIAGNOSTIC(
+            StructInitMissingFieldInLiteral, Error,
+            "Missing value for field `{0}` in struct initialization.",
+            llvm::StringRef);
+        CARBON_DIAGNOSTIC(StructInitMissingFieldInConversion, Error,
+                          "Cannot convert from struct type `{0}` to `{1}`: "
+                          "missing field `{2}` in source type.",
+                          std::string, std::string, llvm::StringRef);
```
jonmeow: These definitions seem like they should be in the respective scope that uses them, rather than outside. It'd leave less to read over to make sure emit arguments match.

toolchain/check/convert.cpp
```diff
@@ -420,6 +420,15 @@ static auto ConvertStructToStruct(Context& context, SemIR::StructType src_type,
...
   }
 
+  // Prepare to look up fields in the source by index.
+  llvm::SmallDenseMap<StringId, int32_t> src_field_indexes;
+  if (src_type.fields_id != dest_type.fields_id) {
+    for (auto [i, field_id] : llvm::enumerate(src_elem_fields)) {
+      src_field_indexes
+          [context.nodes().GetAs<SemIR::StructTypeField>(field_id).name_id] = i;
+    }
+  }
```
jonmeow: Note on structure... If you were to define the map as:

```suggestion
  llvm::SmallDenseMap<StringId, SemIR::StructTypeField> src_field_indexes;
  if (src_type.fields_id != dest_type.fields_id) {
    for (auto field_id : src_elem_fields) {
      auto field = context.nodes().GetAs<SemIR::StructTypeField>(field_id);
      CARBON_CHECK(src_field_indexes.insert({field.name_id, field}).second);
    }
  }
```

Then you could change the below to look like:

```
SemIR::StructTypeField src_field;
if (src_type.fields_id != dest_type.fields_id) {
  src_field_it = ...
  src_field = src_field_it.second;
} else {
  src_field = sem_ir.nodes().GetAs<SemIR::StructTypeField>(src_eleme_fields[i]);
}
```

But you could also really lean into that divergence of fields_id, move the common init_id code into a lambda, and refactor the code to fully separate out the two paths. Not sure how that'd affect readability here.

Anyways, to be clear -- just offering this up in case you think it's an improvement. I might at least consider `.insert` and CHECK success instead of `[]=`, even without the other changes.

jonmeow (approved):
Looks good

toolchain/check/convert.cpp
zygoloid: Our style guide says to use "indexes" (https://developers.google.com/style/word-list#index).

toolchain/check/convert.cpp
zygoloid: Fixed comment.

toolchain/check/convert.cpp

toolchain/check/convert.cpp
zygoloid: Converted to `insert` and a `CHECK`, though I note that we don't actually enforce uniqueness of field names in a struct yet! I need the index in addition to the field, in order to index into the source struct value, so I'm not making the other change for now.

PR #3347, opened 2023-10-30
Add parse support for 'import', brush up 'package' a little.

jonmeow:
This detects ordering issues with the `package` and `import` statements. `library` is changed from package-specific to instead be generic between the two, since structurally it's non-specific.

The next step would be to start exposing the results for the driver to make ordering decisions for checking. That'll involve further modifications to this code, but this felt like a reasonable change point because it's the extent of the parser enforcement, and still causes significant refactoring.


toolchain/parse/handle_import_and_package.cpp
```diff
@@ -0,0 +1,155 @@
...
+        context.AddLeafNode(NodeKind::PackageApi, context.Consume());
+        break;
+      }
+      case Lex::TokenKind::Impl: {
+        context.AddLeafNode(NodeKind::PackageImpl, context.Consume());
+        break;
+      }
+      default: {
+        CARBON_DIAGNOSTIC(ExpectedApiOrImpl, Error,
+                          "Expected a `api` or `impl`.");
```
zygoloid:
```suggestion
                          "Expected an `api` or `impl`.");
```
... or maybe drop the article entirely?
```suggestion
                          "Expected `api` or `impl`.");
```

toolchain/parse/context.h
```diff
@@ -44,6 +44,14 @@ class Context {
     NamedConstraint,
   };
 
+  // Used for restricting ordering of `package` and `import` directives.
+  enum class PackagingState : int8_t {
+    PackageOrImportsAllowed,
+    PackageEncountered,
+    ImportsAllowed,
+    PackageOrImportsInvalid,
+  };
```
zygoloid: I find the names of these enumerators confusing, because they're describing different kinds of information (the first and third enumerators say that something is allowed, the fourth says that something is disallowed, and the second describes our position in parsing). I'd find this easier to follow if these were all phrased as the same kind of information. For example:
```suggestion
  enum class PackagingState : int8_t {
    StartOfFile,
    AfterPackageDeclaration,
    AfterImportDeclaration,
    AfterNonPackagingDeclaration,
  };
```

It also seems to me that the second and third states aren't really different -- in either state, we can have imports but not a package declaration. I note that you currently use this distinction to produce a custom diagnostic for:

```carbon
package A;
package B;
```

that says we can't have two package declarations, whereas you get a different diagnostic (package declaration must be at the start of the file) for:

```carbon
import A;
package B;
```
and for
```carbon
package A;
import B;
package C;
```

... but I think it would be reasonable, and more consistent, to produce the "package declaration must be at the start of the file" diagnostic for all of these cases rather than special-casing two package declarations in a row but not two package declarations in general.

Alternatively, we could stay in the "after package declaration" state when we hit an import after a package, so that `package...; import...; package...;` also produces the "multiple package declarations" error instead of the "package declaration must be at the start of the file" error, but we'd still get the "package declaration must be at the start of the file" error for `package ...; fn ...; package ...;` in that case, so it'd still be a bit inconsistent. Maybe we could track three independent bits -- whether we've seen any package declarations, whether we've seen any import declarations, and whether we've seen any other kinds of declarations.

toolchain/parse/handle_import_and_package.cpp
zygoloid: Looks like this PR is just moving this diagnostic around, so I think it's fine to defer this, but if you want to fix it in passing, please do :)

toolchain/parse/testdata/packages/import/fail_after_decl.carbon
```diff
@@ -0,0 +1,25 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+
+fn A();
+
+// CHECK:STDERR: fail_after_decl.carbon:[[@LINE+3]]:8: ERROR: `import` directives must come after the `package` directive (if present) and before any other entities in the file.
+// CHECK:STDERR: import B api;
+// CHECK:STDERR:        ^
+import B api;
```
zygoloid: I think it'd be best to drop the `api` here so this is a valid `import` declaration other than its location. (I assume this is a copy-paste and not intended?)

toolchain/parse/context.h
jonmeow: My concern was that people may be confused by:

```
package Foo;
package Foo;
```

resulting in a "must be at the start of the file". And I could make it longer ("there must be only one, and it must be at the start of the file") but that felt a little too much. So that's how I ended up with the current setup, even though the `package import package` is a similar issue (but, clearer that the second `package` isn't at the top)

Anyways, doing the name changes.

toolchain/parse/testdata/packages/import/fail_after_decl.carbon
jonmeow: Yes, copy-paste, fixing

toolchain/parse/context.h
jonmeow: Or rather, maybe I should clarify.

First, we refer to `package` and `import` as "directives" instead of "declarations", so I'm trying to be careful about phrasing there.

Then we have:

```
// Foo
package Foo;
```

This isn't technically at the start of the file, but it is before declarations. So I feel I shouldn't just say "must be at the start of the file".

I could maybe say first non-comment line; but can we commit to never adding anything that would be valid above? And even then, is it as clear on a duplicate entry that's still adjacent to the first non-comment line?

jonmeow: I'll note, kind of tempting is a "first processed item is here" kind of diagnostic, though I wasn't sure it's worth the additional state.

Although, maybe it's worth it...

jonmeow:
Added notes to the errors, got rid of the state overlap. I think this is better now.

toolchain/parse/handle_import_and_package.cpp
```diff
@@ -0,0 +1,174 @@
...
+  auto state = context.PopState();
+
+  auto intro_token = context.Consume();
+  context.AddLeafNode(NodeKind::ImportIntroducer, intro_token);
+
+  switch (context.packaging_state()) {
+    case Context::PackagingState::StartOfFile:
+      // `package` is no longer allowed, but `import` may repeat.
+      context.set_packaging_state(Context::PackagingState::InImports,
+                                  intro_token);
```
zygoloid: I wonder if we should do this in the `AfterNonPackagingDeclaration` state too, so that we don't get repeated `ImportTooLate` errors if there are multiple imports in a row.

zygoloid: Hm, no, never mind, that'll make later "first non-comment line" notes point to the wrong place.

zygoloid (approved):
Very nice, thanks!

toolchain/parse/context.h
zygoloid: I think pointing at the previous line helps to clarify what we're talking about; I like the new approach of tracking where we switched state.

toolchain/parse/handle_import_and_package.cpp
jonmeow: How about this `AfterNonPackagingDeclarationImportsWarned` setup?

toolchain/parse/context.h
```diff
@@ -44,6 +44,17 @@ class Context {
...
 
+  // Used for restricting ordering of `package` and `import` directives.
+  enum class PackagingState : int8_t {
+    StartOfFile,
+    InImports,
+    AfterNonPackagingDeclaration,
+    // A warning about `import` placement has been issued so we don't keep
+    // issuing more (when `import` is repeated) until more non-`import`
+    // declarations come up.
+    AfterNonPackagingDeclarationImportsWarned,
```
zygoloid: Perhaps something like `AfterImportAfterNonPackagingDeclaration` so these are still all describing a position?

toolchain/parse/context.h
```diff
@@ -319,6 +340,11 @@ class Context {
   Lex::TokenIterator end_;
 
   llvm::SmallVector<StateStackEntry> state_stack_;
+
+  // The current packaging state, whether `import`/`package` are allowed.
+  PackagingState packaging_state_ = PackagingState::StartOfFile;
+  // The token that led to the current packaging state, starting as invalid.
+  Lex::Token packaging_state_token_ = Lex::Token::Invalid;
```
zygoloid: The comment here seems unclear now (is this the token that took us to the "imports warned" state or the token that took us to the "after non-packacing declaration" state?), and would become incorrect if you make the rename I suggested above.

I wonder if this'd be simpler to reason about if we tracked the location of the first non-packaging token instead of the token that led to the current state? The other thing we care about is the location of the first token overall / first non-comment line, but that's always `Lex::Token(0)`, so we don't need to track it explicitly. (We could start tracking it later if we start allowing anything before the `package` directive.)

zygoloid (approved):
Looks good. Some suggestions, but I'd be happy for you to land this with the comment tweaked slightly to clarify that entering the "imports warned" state doesn't change the token location.

toolchain/parse/context.h
jonmeow: I think `AfterXAfterY` is confusing, renaming to `AfterLateImport`.

jonmeow: Naming `InImportsAfterNonPackagingDeclaration`

toolchain/parse/context.h
jonmeow: Done

jonmeow: Actually, I believe that's incorrect, Lex::Token(0) is StartOfFile. Maybe Lex::Token(1)

PR #3349, opened 2023-10-30
Take the address of the object when calling an `addr self` method.

zygoloid:


toolchain/check/convert.cpp
```diff
@@ -887,13 +887,38 @@ auto ConvertCallArgs(Context& context, Parse::Node call_parse_node,
           &context.emitter(), [&](auto& builder) {
             CARBON_DIAGNOSTIC(
                 InCallToFunctionSelf, Note,
-                "Initializing self parameter of method declared here.");
-            builder.Note(self_param->parse_node, InCallToFunctionSelf);
+                "Initializing `{0}` parameter of method declared here.",
+                llvm::StringRef);
```
```suggestion
                llvm::StringLiteral);
```

Is StringLiteral valid here? StringRef is just something that makes me cautious on lifetime concerns, although this use looks valid.

jonmeow (approved):
Looks good, although I want to note that ConvertCallArgs is now about 125 lines. That's a bit long, but still borderline; not necessary to split here, but maybe likely to keep growing?

toolchain/check/convert.cpp
zygoloid: Yes, it's a bit awkward because we can't convert the conditional expression below to `StringLiteral`, but that's not a big problem. Done.

zygoloid:
> Looks good, although I want to note that ConvertCallArgs is now about 125 lines. That's a bit long, but still borderline; not necessary to split here, but maybe likely to keep growing?

I've factored out the `self` checking, which brings it down by 50 lines.

toolchain/check/convert.cpp
```diff
@@ -837,6 +837,64 @@ auto ConvertToBoolValue(Context& context, Parse::Node parse_node,
       context.GetBuiltinType(SemIR::BuiltinKind::BoolType));
 }
 
+namespace {
+CARBON_DIAGNOSTIC(InCallToFunction, Note, "Calling function declared here.");
+}
```
```suggestion
CARBON_DIAGNOSTIC(InCallToFunction, Note, "Calling function declared here.");
```

These are already static, shouldn't need a namespace.

jonmeow (approved):
Thanks!

PR #3352, opened 2023-11-01
Split IdentifierId and StringLiteralId from StringId

jonmeow:
Following up on discussion yesterday regarding this split.

Note, I'm expecting #3341 to do IdentifierId -> NameId in SemIR. It might be worth adding NameId creation directly to StringStore if you're content with this setup though.

### DIFF
--- a/toolchain/base/value_store.h
+++ b/toolchain/base/value_store.h
@@ -203,42 +203,51 @@ class ValueStore<StringId> : public Yaml::Printable<ValueStore<StringId>> {
   llvm::SmallVector<llvm::StringRef> values_;
 };
 
-class StringStore : public Yaml::Printable<StringStore> {
+// A thin wrapper around a `ValueStore<StringId>` that provides a different IdT,
+// while using a unified storage for values. This avoids potentially
+// duplicative string hash maps, which are expensive.
+template <typename IdT>
+class StringStoreWrapper : public Printable<StringStoreWrapper<IdT>> {
  public:
-  template <typename IdT>
-  static constexpr bool IsCompatibleId =
-      std::is_same_v<IdT, IdentifierId> || std::is_same_v<IdT, StringLiteralId>;
+  explicit StringStoreWrapper(ValueStore<StringId>* values) : values_(values) {}
 
-  template <typename IdT,
-            typename std::enable_if_t<IsCompatibleId<IdT>>* = nullptr>
   auto Add(llvm::StringRef value) -> IdT {
-    return IdT(values_.Add(value).index);
+    return IdT(values_->Add(value).index);
   }
 
-  template <typename IdT,
-            typename std::enable_if_t<IsCompatibleId<IdT>>* = nullptr>
   auto Get(IdT id) const -> llvm::StringRef {
-    return values_.Get(StringId(id.index));
+    return values_->Get(StringId(id.index));
   }
 
-  auto OutputYaml() const -> Yaml::OutputMapping {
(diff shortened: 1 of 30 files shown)


toolchain/sem_ir/file.h
```diff
@@ -268,8 +268,8 @@ class File : public Printable<File> {
   auto reals() const -> const ValueStore<RealId>& {
     return value_stores_->reals();
   }
-  auto strings() -> ValueStore<StringId>& { return value_stores_->strings(); }
-  auto strings() const -> const ValueStore<StringId>& {
+  auto strings() -> StringStore& { return value_stores_->strings(); }
+  auto strings() const -> const StringStore& {
     return value_stores_->strings();
   }
```
zygoloid: While I think it is reasonable to share the storage and the ID space here, I wonder if this is the interface we want -- in particular, do we want to expose the fact that we store IDs and string literals in the same storage to API clients?

We could instead provide two wrappers around the internal value store, one for string literals and one for identifiers, so that client code can use `context.identifiers().Get(ident_id)` or `context.string_literals().Add("blah")` without needing to know that they share storage.

What do you think? I don't think it's essential to hide this implementation detail, but it would seem a bit more consistent to me to do so, so that each kind of ID has its own function to access the value store.

zygoloid (approved):
Looks good. One substantive comment, but I think it's suggesting an incremental change so I'm fine with landing this to avoid backsliding and then discussing whether we want to make that change.

toolchain/sem_ir/file.h
jonmeow: Done

toolchain/base/value_store.h
```diff
@@ -186,16 +203,51 @@ class ValueStore<StringId> : public Yaml::Printable<ValueStore<StringId>> {
...
+
+  auto Print(llvm::raw_ostream& out) const -> void { out << *values_; }
+
+ private:
+  ValueStore<StringId>* values_;
+};
+
 // Stores that will be used across compiler phases for a given compilation unit.
 // This is provided mainly so that they don't need to be passed separately.
 class SharedValueStores : public Yaml::Printable<SharedValueStores> {
```
zygoloid, line 230: We should make this class non-copyable, because the implicitly copy constructor is no longer correct (the copies of the wrappers would end up pointing to the string store in the original).

Alternatively, we could create the wrappers on the fly like I did in #3341, rather than storing an instance. That's also probably slightly more efficient, because it avoids loading the pointer to the string value store from the stored wrapper, although I doubt that would make any real difference in practice.

jonmeow: > We should make this class non-copyable, because the implicitly copy constructor is no longer correct (the copies of the wrappers would end up pointing to the string store in the original).

Making non-copyable. I think incidental copies would only lead to problems anyways (e.g., if you assume you can just copy this [SharedValueStore] and a TokenizedBuffer using it, _that_ fails too).

> Alternatively, we could create the wrappers on the fly like I did in #3341, rather than storing an instance. That's also probably slightly more efficient, because it avoids loading the pointer to the string value store from the stored wrapper, although I doubt that would make any real difference in practice.

I'll leave it to you if you want to swap this, but I feel slightly faster to reason about this behavior unless we're going to do the on-the-fly creation as a more common semantic. [I'd also missed this in review, had focused more on just naive understanding of accessors -- now the `decltype(auto)` is both more and less helpful, in that it helped me miss this]

PR #3354, opened 2023-11-01
Rework name lookup to handle non-lexical scoping.

zygoloid:
When declaring a name such as `fn Ns.Class.F() { ... }`, enter the scopes of `Ns` and `Ns.Class` as we form the name, and remain in those non-lexical scopes until the end of the declaration.

When performing an unqualified lookup, look in any enclosing non-lexical scopes in addition to looking into the lexical name table.

We now track a scope index with each lookup result in the lexical name lookup table. This is used to determine whether a lexical or non-lexcial result is the innermost result and whether a declared name is in the same scope as some previous introduction of that name or in a nested scope. For now, this could just be the index into the scope_stack, but the intent is to also use this to detect names being declared after they are first looked up, which requires the indexes to outlive their scopes, so we use a persistent numbering of all scopes instead. The persistent numbering also permits more invariant checking.


toolchain/check/declaration_name_stack.h
```diff
@@ -11,6 +11,19 @@
 
 namespace Carbon::Check {
 
+// An index for a pushed scope, which may be a permanent scope with a
+// corresponding `NameScope`, or a temporary scope, such as is created for a
+// block.
```
jonmeow:
```suggestion
// An index for a pushed scope. The scope may be a permanent scope with a
// corresponding `NameScope` or a temporary scope, such as is created for a
// block.
```

Suggesting fewer commas in this paragraph.

The first "which" could be referring to "An index" rather than "pushed scope"; saying "The scope" drops a comma and clarifies.

Oxford comma only applies to lists of 3 or more, and makes it harder to read with the comma before "such as" (makes `, or a temporary scope,` look potentially like a parenthetical)

toolchain/check/declaration_name_stack.h
```diff
@@ -11,6 +11,19 @@
...
+//
+// These are comparable, and lower values correspond to scopes entered earlier
+// in the file.
```
jonmeow: Does "These" mean "indices" or "scopes"? Does "lower values" refer to values in lowering, the indexed values, or the comparison of indices?

Maybe:

```suggestion
// Indices are comparable. Lower indices correspond to scopes entered earlier in the file.
```

toolchain/check/declaration_name_stack.h
```diff
@@ -98,10 +115,16 @@ class DeclarationNameStack {
...
-  // node stack, which will be applied to the declaration name if appropriate.
-  auto Pop() -> NameContext;
+  // Finishes the current declaration name processing, returning the final
+  // context for adding the name to lookup. This also pops the final name node
+  // from the node stack, which will be applied to the declaration name if
+  // appropriate.
+  auto Finish() -> NameContext;
+
+  // Pops the declaration name from the declaration name stack, and leaves all
+  // scopes that were entered as part of handling the declaration name. This
```
jonmeow: Are "scopes that were entered as part of handling" the qualifying scopes, i.e. `A.B.C` in `fn A.B.C.D()`? Or are they something else?

toolchain/check/declaration_name_stack.h
```diff
@@ -98,10 +115,16 @@ class DeclarationNameStack {
...
 
-  // Pops the current declaration name processing, returning the final context
-  // for adding the name to lookup. This also pops the final name node from the
-  // node stack, which will be applied to the declaration name if appropriate.
-  auto Pop() -> NameContext;
+  // Finishes the current declaration name processing, returning the final
+  // context for adding the name to lookup. This also pops the final name node
+  // from the node stack, which will be applied to the declaration name if
+  // appropriate.
+  auto Finish() -> NameContext;
```
jonmeow: Both in the name and comments, the idea that this "finishes" processing seems misleading, given that `Pop` must still be called (which is not documented, and the "This also pops" comment can now be misread). Maybe qualifying it would help (e.g., `FinishName`) to hint that it's not doing everything, but you could also consider something like `GetName` or `MarkNameDone`.

As a side thought, you could do `PushAndStartName` (or `PushScopeAndStartName`), reflecting the two different actions needed to close what's started and getting more mirroring of a `Pop`/`PopScope`+`FinishName`, if you prefer the `Finish` as a form.

toolchain/check/declaration_name_stack.cpp
```diff
@@ -39,7 +40,12 @@ auto DeclarationNameStack::Pop() -> NameContext {
     ApplyNameQualifier(parse_node, name_id);
   }
 
-  return declaration_name_stack_.pop_back_val();
+  return declaration_name_stack_.back();
```
jonmeow: WDYT of having this set a state that prevents its reuse for the NameContext? Or is there some issue with doing so?

toolchain/check/declaration_name_stack.cpp
```diff
@@ -39,7 +40,12 @@ auto DeclarationNameStack::Pop() -> NameContext {
...
+}
+
+auto DeclarationNameStack::Pop() -> void {
+  context_->PopToScope(declaration_name_stack_.back().enclosing_scope);
+  declaration_name_stack_.pop_back();
```
jonmeow, line 56: It feels odd to me that the DeclarationNameStack mutates the context. Is that necessary, versus having the context mutate the DeclarationNameStack?

Should the function comment mention that it now mutates the context's scope?

toolchain/check/declaration_name_stack.cpp
```diff
@@ -129,16 +130,19 @@ auto DeclarationNameStack::UpdateScopeIfNeeded(NameContext& name_context)
       if (class_info.is_defined()) {
         name_context.state = NameContext::State::Resolved;
         name_context.target_scope_id = class_info.scope_id;
+        context_->PushScope(name_context.resolved_node_id, class_info.scope_id);
```
jonmeow: Similar to the above comment about `context_->PopScope`, I think the `PushScope` behavior needs to be reviewed -- maybe not moving to the `Context` object, but at least being documented behavior in the header, since it's an external mutation that may not be expected.

toolchain/check/context.cpp
```diff
@@ -108,59 +108,166 @@ auto Context::NoteIncompleteClass(SemIR::ClassId class_id,
...
-      return SemIR::NodeId::BuiltinError;
+  // Walk the non-lexical scopes and perform lookups into each of them.
+  for (std::size_t non_lexical_scope_index :
+       llvm::reverse(non_lexical_scope_indexes_)) {
+    auto& non_lexical_scope = scope_stack_[non_lexical_scope_index];
+
+    // If the innermost lexical result is within this non-lexical scope, then
+    // it shadows all further non-lexical results and we're done.
+    if (!lexical_results.empty() &&
+        lexical_results.back().scope_index > non_lexical_scope.index) {
```
jonmeow: I'm a little confused by this. `non_lexical_scope_index` is a `size_t` instead of a `ScopeIndex`, even though it ends with `scope_index`. But you use this to get `non_lexical_scope.index`, which _is_ a `ScopeIndex`.

I was going to suggest `indexes` -> `indices`, but that's a nit. Maybe the names should be more distinct?

Maybe `non_lexical_scope_indexes_` could just maintain the set of `ScopeIndex` + `NameScopeId`, i.e. a stack of `non_lexical_scopes_`? (which would be the same size as `size_t`, right?) That would remove the need to do the `scope_stack_` lookup on 184, as a small benefit.

toolchain/check/declaration_name_stack.h
zygoloid: Fixed. (I moved the remaining comma earlier to be clearer what the "such as" binds to.)

toolchain/check/declaration_name_stack.h
zygoloid: Fixed, and avoided taking a position on "indexes" versus "indices" ;)

toolchain/check/declaration_name_stack.h
zygoloid: -> `PushScopeAndStartName`, `FinishName`, `PopScope`.

toolchain/check/declaration_name_stack.h
zygoloid: Comment hopefully clarified. I also updated the class comment.

toolchain/check/declaration_name_stack.cpp
zygoloid: Done. Turned out to be slightly awkward because we want the original state in the name that we return here, for later use by the caller, but I think this adds some good error checking, thanks!

toolchain/check/declaration_name_stack.cpp
zygoloid: I think the `DeclarationNameStack` mutating the context when building up the declaration name is the simplest approach; otherwise I think we end up moving a lot of the logic currently in `DeclarationNameStack::UpdateScopeIfNeeded` and its calling logic out to somewhere else. And if `DeclarationNameStack` is pushing the scopes, I think logically it should be popping the scopes too.

Extended the comment to say more about popping the scopes, and updated the comment for `ApplyNameQualifier` to note that it can enter scopes.

toolchain/check/declaration_name_stack.cpp
zygoloid: Documented in the header, in both the class and public function comment.

toolchain/check/context.cpp
zygoloid: Switched to a list of `ScopeIndex` + `NameScopeId` as suggested. I think we'll probably eventually also want the `scope_inst_id`, but we can add that when needed.

toolchain/check/declaration_name_stack.h
jonmeow: I still think the comma before "or" is incorrect, grammatically. It suggests they're two independent clauses, which is when you'd typically add a comma when there's not a third item (which, given the length of both halves, is part of my difficulty in reading -- I want to use the comma to separate clauses).

This might read better with "may be" -> "may either be", emphasizing the binary options.

toolchain/check/declaration_name_stack.h
jonmeow: Lies! "`ScopeIndex` is comparable" would avoid taking a stance. :-P (or "`ScopeIndex` values are")

toolchain/check/declaration_name_stack.h
jonmeow: Thanks!

jonmeow (approved):
Thanks, looks great! Grammar bikesheds aside. 😉

toolchain/check/declaration_name_stack.h
zygoloid: Found some more things I wanted to say here and split this to two sentences. I think that should resolve the issue.

toolchain/check/declaration_name_stack.h
zygoloid: Sure, changed to "`ScopeIndex` values are" to further sidestep the question.

(For what it's worth, I do think that "`ScopeIndex`es" is not taking a position on "indexes" vs "indices" here. I would write -- and in other codebases, have written -- "`Child`s are ..." too, and I would use "`ScopeIndex`es" here even if we decide we want to use "indices" in general. I don't think it makes sense to apply irregular plurals to quoted technical text -- "`ScopeIndices`" or some attempt at a middle ground like "`ScopeIndic`es" would obviously be wrong.)

jonmeow (approved):

PR #3358, opened 2023-11-02
Support initializing a class from a struct.

zygoloid:


toolchain/check/convert.cpp
```diff
@@ -415,7 +419,13 @@ static auto ConvertStructToStruct(Context& context, SemIR::StructType src_type,
...
                       "with {1} element(s).",
                       size_t, size_t);
-    context.emitter().Emit(value.parse_node(), StructInitElementCountMismatch,
+    CARBON_DIAGNOSTIC(ClassInitElementCountMismatch, Error,
+                      "Cannot initialize class with {0} field(s) from struct "
+                      "with {1} element(s).",
+                      size_t, size_t);
+    context.emitter().Emit(value.parse_node(),
+                           is_class ? ClassInitElementCountMismatch
+                                    : StructInitElementCountMismatch,
```
jonmeow: Maybe make a `TokenKind` argument to replace `struct` and `class`, rather than a separate diagnostic?

toolchain/check/convert.cpp
```diff
@@ -494,12 +504,60 @@ static auto ConvertStructToStruct(Context& context, SemIR::StructType src_type,
...
                                                      new_block.id()})
-                 : context.AddNode(SemIR::StructValue{value.parse_node(),
-                                                      target.type_id, value_id,
-                                                      new_block.id()});
+         : is_init ? context.AddNode(SemIR::StructInit{value.parse_node(),
+                                                       target.type_id, value_id,
+                                                       new_block.id()})
+                   : context.AddNode(
+                         SemIR::StructValue{value.parse_node(), target.type_id,
+                                            value_id, new_block.id()});
```
jonmeow: I think this would be easier to read as if/else instead of a nested ternary, given the length.

toolchain/check/convert.cpp
```diff
@@ -622,6 +680,17 @@ static auto PerformBuiltinConversion(Context& context, Parse::Node parse_node,
     }
   }
 
+  // A struct {.f_1: T_1, .f_2: T_2, ..., .f_n: T_n} converts to a class type
+  // if it converts to the class's representation type.
```
jonmeow: I'm confused by the second line here; what's it trying to say? I don't understand the difference between a class and a class's representation, or how the code below gets a representation separate from `ClassType`.

toolchain/check/convert.cpp
```diff
@@ -383,12 +384,15 @@ static auto ConvertTupleToTuple(Context& context, SemIR::TupleType src_type,
...
 }
 
-// Performs a conversion from a struct to a struct type. Does not perform a
-// final conversion to the requested expression category.
-static auto ConvertStructToStruct(Context& context, SemIR::StructType src_type,
-                                  SemIR::StructType dest_type,
-                                  SemIR::NodeId value_id,
-                                  ConversionTarget target) -> SemIR::NodeId {
+// Performs a conversion from a struct to a struct or class type. Does not
+// perform a final conversion to the requested expression category.
```
jonmeow: Instead of "Does not perform", maybe "The caller should perform ... if necessary"? Or maybe a TODO if that's the intent? (the latter is where I'm a little uncertain from phrasing; similar elsewhere you have this comment)

toolchain/check/convert.cpp
```diff
@@ -494,12 +504,60 @@ static auto ConvertStructToStruct(Context& context, SemIR::StructType src_type,
...
+static auto ConvertStructToStruct(Context& context, SemIR::StructType src_type,
+                                  SemIR::StructType dest_type,
+                                  SemIR::NodeId value_id,
+                                  ConversionTarget target) -> SemIR::NodeId {
+  return ConvertStructToStructOrClass<SemIR::StructAccess>(
+      context, src_type, dest_type, value_id, target, /*is_class=*/false);
+}
+
+// Performs a conversion from a struct to a class type. Does not perform a
+// final conversion to the requested expression category.
```
jonmeow: This seems to be an identical comment to `ConvertStructToStructOrClass`. The comments should probably call out a difference.

toolchain/check/convert.cpp
```diff
@@ -383,12 +384,15 @@ static auto ConvertTupleToTuple(Context& context, SemIR::TupleType src_type,
...
-// Performs a conversion from a struct to a struct type. Does not perform a
-// final conversion to the requested expression category.
-static auto ConvertStructToStruct(Context& context, SemIR::StructType src_type,
-                                  SemIR::StructType dest_type,
-                                  SemIR::NodeId value_id,
-                                  ConversionTarget target) -> SemIR::NodeId {
+// Performs a conversion from a struct to a struct or class type. Does not
+// perform a final conversion to the requested expression category.
+template <typename TargetAccessNodeT>
+static auto ConvertStructToStructOrClass(Context& context,
```
jonmeow, line 391: Maybe add a hint to make it clearer that this shouldn't be called directly (I assume), or adjust names to more precisely reflect the difference between `ConvertStructToStructOrClass` and `ConvertStructToClass`.

jonmeow (approved):
The implementation looks good, although I was stumbling a little on the comments.

toolchain/check/convert.cpp
zygoloid: Made the diagnostics more similar and folded them together. We don't have a `TokenKind` for `struct`, so I used a `StringLiteral` instead.

toolchain/check/convert.cpp
zygoloid: I've expanded this comment a bit to help the reader understand what the representation type is. In each of these cases, the actual checking is done by the factored out `ConvertXToY` function; in this case, we get the representation type in `ConvertStructToClass`.

toolchain/check/convert.cpp
zygoloid: I'd prefer to state a fact here rather than give advice. Would changing "Does not" to "Intentionally does not" help explain that this is not a TODO?

toolchain/check/convert.cpp
zygoloid: Not identical -- that one says "to a struct or class type" and this one just says "to a class type". But I've changed the comment on the helper to say that it's a helper, which I think should resolve this.

toolchain/check/convert.cpp
zygoloid: Changed comment to say that this is just common implementation for the other two functions.

toolchain/check/convert.cpp
zygoloid: Added a few more words to make this sound less TODO-y.

jonmeow: Yes, thanks!

toolchain/check/convert.cpp

toolchain/check/convert.cpp
```diff
@@ -622,6 +680,19 @@ static auto PerformBuiltinConversion(Context& context, Parse::Node parse_node,
     }
   }
 
+  // A struct {.f_1: T_1, .f_2: T_2, ..., .f_n: T_n} converts to a class type
+  // if it converts to the struct type that is the class's representation type,
+  // that is, a struct with the same fields as the class, plus a base field
+  // where relevant.
```
```suggestion
  // if it converts to the struct type that is the class's representation type
  // (a struct with the same fields as the class, plus a base field
  // where relevant).
```

Try not to have comma-separated parentheticals inside of comma-separated parentheticals.

jonmeow (approved):
Thanks for the changes, one minor suggestion. :-P

PR #3360, opened 2023-11-02
Clarify some comments

josh11b:


toolchain/parse/node_kind.def
```diff
@@ -89,6 +89,9 @@ CARBON_PARSE_NODE_KIND_CHILD_COUNT(NameExpression, 0, CARBON_TOKEN(Identifier))
 //    is documented separately.
 // -  `_optional_` if this node may be present or omitted in valid parses,
 //    depending on which tokens are in the source code.
+//
+// Generally all of the non-external nodes in one block will be created in the
+// same `handle_*.cpp` file.
```
jonmeow: The `handle_*` files are named based on tokens rather than parse node kinds. While there is some correspondence, this seems to imply that the files are based on the parse tree nodes. You might instead say something like "There is generally a close correspondence between handling of tokens and the creation of non-external nodes in a given block."

jonmeow (approved):

toolchain/parse/node_kind.def
jonmeow: (for each of parse, check, and lower, the handle_* files are handling output from the prior step)

josh11b: I used your text.

PR #3362, opened 2023-11-03
Add a separate store for computed constant values.

zygoloid:
This moves the instructions generated for type values out of the block in which they happen to first be referenced, and into shared storage.

### DIFF
--- a/toolchain/check/testdata/array/base.carbon
+++ b/toolchain/check/testdata/array/base.carbon
@@ -8,6 +8,16 @@ var a: [i32; 1] = (1,);
 var b: [f64; 2] = (11.1, 2.2,);
 var c: [(); 5] = ((), (), (), (), (),);
 
+// CHECK:STDOUT: constants {
+// CHECK:STDOUT:   %.loc7_15: type = ptr_type [i32; 1]
+// CHECK:STDOUT:   %.loc7_22: type = tuple_type (i32)
+// CHECK:STDOUT:   %.loc8_15: type = ptr_type [f64; 2]
+// CHECK:STDOUT:   %.loc8_30: type = tuple_type (f64, f64)
+// CHECK:STDOUT:   %.loc9_10: type = tuple_type ()
+// CHECK:STDOUT:   %.loc9_14: type = ptr_type [(); 5]
+// CHECK:STDOUT:   %.loc9_38: type = tuple_type ((), (), (), (), ())
+// CHECK:STDOUT: }
+// CHECK:STDOUT:
 // CHECK:STDOUT: file "base.carbon" {
 // CHECK:STDOUT:   %.loc7_14: i32 = int_literal 1
 // CHECK:STDOUT:   %.loc7_15: type = array_type %.loc7_14, i32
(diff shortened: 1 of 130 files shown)


chandlerc (commented):
I like moving these instructions out of functions, but why a separate block as opposed to putting them into the file block? Having both the file and the constants block seems a bit odd, and it's not clear what the distinction is?

zygoloid:
> I like moving these instructions out of functions, but why a separate block as opposed to putting them into the file block? Having both the file and the constants block seems a bit odd, and it's not clear what the distinction is?

The `file` is sequentially executed, and can contain control flow (eg, if a global variable's type has an `if` expression within it). The `constants` block is atemporal and can be referenced from anywhere within the file, regardless of where we are in the overall evaluation. (I'm expecting that we would eventually lower the `constants` block up-front, before lowering anything else.) So, a couple of things motivate this approach:

1) Conditional execution.

    ```carbon
    var x: (if true then (i32, i32) as type else (i32, i32, i32) as type);
    ```

    ... will create global constants in this model, but would presumably put the canonical representation of the `(i32, i32)` and `(i32, i32, i32)` types in conditionally-executed blocks reachable from the `file` block if that's where we put constants.

2) Arbitrary placement.

    If we put constants in the `file` block, then they'd show up in an arbitrary location -- wherever we happened to reference them first. That means that the semir that we generate for a particular source construct would depend on which constants we happened to have executed beforehand, and which constants we happen to generate while checking a particular declaration (eg, which subexpressions we constant evaluate).

We could work around those things, but I think at least point (1) is likely to be hard to deal with: if a constant appears in an unreachable code block, lowering would usually not lower that code block, but the constant could be referenced from reachable code anywhere in the file.

chandlerc:
Thanks, that helps me understand a lot better.

One follow-up then -- could we keep the constants block *first* then? To make it clear that this is actually a dependency of even the file block?

A second question that doesn't really need an answer in this PR but seems worth thinking about -- the order *within* the constant block is still a little fragile. It's better in some senses than the CFG issues you mention for the file -- because the *visit* order is unambiguous. But the *construction* order is dependent on a fairly trivial aspect of the source file organization. Would it make sense to try to compute a total ordering for constants and sort them? My guess is this might be hard, maybe even impossible, and also difficult. But seemed worth pondering to see if we could get even the relative position within the constant block to be more stable long-term.

zygoloid:
> One follow-up then -- could we keep the constants block _first_ then? To make it clear that this is actually a dependency of even the file block?

Done.

> A second question that doesn't really need an answer in this PR but seems worth thinking about -- the order _within_ the constant block is still a little fragile. It's better in some senses than the CFG issues you mention for the file -- because the _visit_ order is unambiguous. But the _construction_ order is dependent on a fairly trivial aspect of the source file organization. Would it make sense to try to compute a total ordering for constants and sort them? My guess is this might be hard, maybe even impossible, and also difficult. But seemed worth pondering to see if we could get even the relative position within the constant block to be more stable long-term.

I think we could come up with an arbitrary total order. Currently the constants are generated in a topological order (by construction), which I suppose is better than nothing stability-wise but very lacking. This'd be easier we're OK with the total order having forward references. I think we should also think about a better way to name the constants -- right now we're using the source location of the first reference to them.

toolchain/check/testdata/array/base.carbon
```diff
@@ -8,56 +8,59 @@ var a: [i32; 1] = (1,);
...
+// CHECK:STDOUT:   %.loc9_10: type = tuple_type ()
+// CHECK:STDOUT:   %.loc9_14: type = ptr_type [(); 5]
+// CHECK:STDOUT:   %.loc9_38: type = tuple_type ((), (), (), (), ())
+// CHECK:STDOUT: }
+// CHECK:STDOUT:
 // CHECK:STDOUT: file "base.carbon" {
 // CHECK:STDOUT:   %.loc7_14: i32 = int_literal 1
-// CHECK:STDOUT:   %.loc7_15.1: type = array_type %.loc7_14, i32
-// CHECK:STDOUT:   %.loc7_15.2: type = ptr_type [i32; 1]
+// CHECK:STDOUT:   %.loc7_15: type = array_type %.loc7_14, i32
```
chandlerc, line 23: (Not blocking)

It's a bit odd that `ptr_type [i32; 1]` is a constant above, but the formulation of the `[i32; 1]` array type is here and not a constant.

Is this just a matter of finishing the hoisting into constants? Or is there something fundamental here?

chandlerc (approved):
Looks good!

I have another follow-up question inline but not for this PR.

PR #3363, opened 2023-11-03
Add SemIR instruction to track that a conversion was performed. 

zygoloid:
Instead of ad-hoc conversion tracking on some kinds of nodes that conversion creates, consolidate tracking into a single node kind. This frees up an operand on `Init` instructions that can be used to store the destination.


toolchain/sem_ir/typed_insts.h
```diff
@@ -244,6 +243,15 @@ struct ConstType {
   TypeId inner_id;
 };
 
+struct Conversion {
+  static constexpr auto Kind = InstKind::Conversion.Define("conversion");
```
jonmeow: Maybe "convert" or "convert_to" instead of "conversion"? Partly shorter, partly a better match for "initialize_from" versus "initialization_from" (word form, not as focused on "from" versus "to", although maybe "to" would help you on name collisions)

jonmeow (approved):
Looks good!

toolchain/sem_ir/typed_insts.h
zygoloid: Hm, I've tried this out, and to me it gives the suggestion that the `convert` node is actually performing a conversion itself, rather than just tracking that one was performed. By contrast, `initialize_from` actually performs initialization itself.

Maybe something like `converted_from` would work? But that is definitely not shorter any more :-)

jonmeow: I'd be fine with `converted`, `converted_from`, `convert_from`, etc.

Noting the example IR:

```
// CHECK:STDOUT:   %.loc13_26.1: {.a: i32, .b: i32} = struct_literal (%.loc13_17, %.loc13_25)
// CHECK:STDOUT:   %Class.ref: type = name_reference "Class", file.%Class
// CHECK:STDOUT:   %.loc13_26.2: ref Class = temporary_storage
// CHECK:STDOUT:   %.loc13_26.3: ref i32 = class_field_access %.loc13_26.2, member0
// CHECK:STDOUT:   %.loc13_26.4: init i32 = initialize_from %.loc13_17 to %.loc13_26.3
// CHECK:STDOUT:   %.loc13_26.5: ref i32 = class_field_access %.loc13_26.2, member1
// CHECK:STDOUT:   %.loc13_26.6: init i32 = initialize_from %.loc13_25 to %.loc13_26.5
// CHECK:STDOUT:   %.loc13_26.7: init Class = class_init (%.loc13_26.4, %.loc13_26.6), %.loc13_26.2
// CHECK:STDOUT:   %.loc13_26.8: ref Class = temporary %.loc13_26.2, %.loc13_26.7
// CHECK:STDOUT:   %.loc13_26.9: ref Class = conversion %.loc13_26.1, %.loc13_26.8
```

`initialize_from` has `to`, whereas `conversion` doesn't. But the second argument is still kind of a "to", isn't it?

Maybe:

```
// CHECK:STDOUT:   %.loc13_26.9: ref Class = convert_from %.loc13_26.1 to %.loc13_26.8
```

I _think_ "to" or something like that might make sense here? But feel free to do something you think makes sense here, which could include just merging as-is.

zygoloid: `to` is used to indicate the thing being initialized by an initializing expression, which doesn't fit here. In the interest of brevity and avoiding the link with `to`, I'll go with `converted`.

PR #3364, opened 2023-11-03
Split build keys so that a read-write key is only used for merged PRs, and read-only is used otherwise.

jonmeow:
The final test commit has an example of the push run:
https://github.com/carbon-language/carbon-lang/actions/runs/6748256043/job/18346225812

But pull request runs require merging to trunk, AFAIK. Unfortunately that means the remote upload disabling isn't really tested in full.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -132,24 +132,23 @@ jobs:
           which clang++
           clang++ --version
 
-      # Extract our access key for our build cache.
-      - name: Extract access key (read-only)
+      # Disable uploads when the remote cache is read-only.
+      - name: Set up remote cache access (read-only)
         if:
           steps.filter.outputs.ignore == 'false' && github.event_name ==
           'pull_request'
         run: |
-          # Disable uploads when the remote cache is read-only.
           echo "remote_cache_upload=--remote_upload_local_results=false" \
               >> $GITHUB_ENV
 
-      - name: Extract access key (read-write)
+      # Provide a cache key when the remote cache is read-write.
+      - name: Set up remote cache access (read-write)
         if:
           steps.filter.outputs.ignore == 'false' && github.event_name !=
           'pull_request'
         env:
-          REMOTE_CACHE_KEY: ${{ secrets.CARBON_GITHUB_BUILDS }}
+          REMOTE_CACHE_KEY: ${{ secrets.CARBON_BUILDS_GITHUB }}
         run: |
-          # Provide a cache key when the remote cache is read-write.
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yaml
```diff
@@ -133,13 +133,26 @@ jobs:
           clang++ --version
 
       # Extract our access key for our build cache.
-      - name: Extract access key
-        if: steps.filter.outputs.ignore == 'false'
+      - name: Extract access key (read-only)
+        if:
+          steps.filter.outputs.ignore == 'false' && github.event_name ==
+          'pull_request_target'
```
chandlerc: Just to confirm -- the goal is to first split the keys, and then pivot to just `pull_request` actions (and make they key visible there)?

chandlerc (approved):
LG (if I've understood the plan correctly, see inline)

.github/workflows/tests.yaml
jonmeow: No, this is the final state.

If you're referring to my comment, pull requests use the merged, trunk state of the action, which I can't modify with a pending PR. So I can test the results of a push event (because I can push to action-test) but not a pull request event (because I cannot push to trunk).

Can you please clarify whether you're okay with this, and if not, why?

chandlerc: My understanding is that we'll need to move from `pull_request_target` to `pull_request` to fully secure this: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

I still think this is a good first step though? Happy to chat more about this if useful.

jonmeow: Switched. Note, still only moderately tested.

.github/workflows/tests.yaml
```diff
@@ -133,13 +133,26 @@ jobs:
           clang++ --version
 
       # Extract our access key for our build cache.
-      - name: Extract access key
-        if: steps.filter.outputs.ignore == 'false'
+      - name: Extract access key (read-only)
```
chandlerc: Maybe rename?

```suggestion
      - name: Set up read-only cache access
```

chandlerc (approved):
Awesome, thanks!!!

.github/workflows/tests.yaml
jonmeow: Going with "Extract access key" -> "Set up remote cache access", keeping parenthetical for read-write vs read-only

chandlerc (approved):
Just confirming this LG to land (and begin triaging any failures) whenever you're ready and around for it.

As discussed, leaving it for you to merge when ready given that it can only be somewhat tested ahead of time and may need follow-ups to clean up things.

PR #3369, opened 2023-11-03
handle missing addr in self pointer pattern

### LINKED ISSUE #3367: Missing `addr` in self pointer pattern causes crash when called
### Description of the bug:

If you create a mutable member function (with a deduced `Self*`) and forget to include `addr` you get a crash when you call the function. 

https://godbolt.org/z/ns7Mq8TW7

Some sort of compiler warning, preferably one that suggests adding `addr`.

### What actually happened?

Compiler Crashed.

I have attempted to start fixing this and will create a PR shortly. There seems to exist many different spots where this could be caught including:

* While parsing, but I am not sure if there is enough type information at this point.
* `CheckAddrMeAccess` - where we type check `self` (nee `me`).
* `BindSelfIfPresent` - where the actual binding of self happens.
* `Interpreter::StepLocation` - where the actual crash occurs.

My PR is going to attempt to modify `CheckAddrMeAccess` however I have a question. In that function we check for an exact match between receiver and self type (only when using addr however). This results in [this](https://godbolt.org/z/P78PvEve7) not compiling, but [this](https://godbolt.org/z/hebcP8oMa) does. Is this difference on purpose? If not, which one has the correct behavior? I naively would think they should both work.

domisterwoozy:
Add validation to `CheckAddrSelfAccess` to additionally check for situations where `addr` is potentially missing. I also updated the name of the function since `me` was renamed to `self`.

Closes #3367

### DIFF
--- a/explorer/interpreter/type_checker.cpp
+++ b/explorer/interpreter/type_checker.cpp
@@ -2733,19 +2733,29 @@ static auto IsInstanceMember(Nonnull<const Element*> element) {
   }
 }
 
-auto TypeChecker::CheckAddrMeAccess(
+auto TypeChecker::CheckAddrSelfAccess(
     Nonnull<MemberAccessExpression*> access,
     Nonnull<const FunctionDeclaration*> func_decl, const Bindings& bindings,
     const ImplScope& impl_scope) -> ErrorOr<Success> {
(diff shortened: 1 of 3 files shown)


jonmeow (approved):
Thank you! The changes here look great.

PR #3372, opened 2023-11-07
Fix crash when an expression cannot convert to a type.

jonmeow:
This turns out to somewhat block `package` support because there's a fuzzer test-case that does similar. The parse is valid so we should probably handle it reasonably.

I suspect the `let` test case might work with a little effort, given it shouldn't really require much evaluation. On the other hand, `var` definitely shouldn't, and `fn` will probably require something like `constexpr` plus more substantial compile-time evaluation support.


toolchain/check/convert.cpp
```diff
@@ -1088,8 +1088,18 @@ auto ConvertCallArgs(Context& context, Parse::Node call_parse_node,
...
-      context, parse_node, value_id, SemIR::TypeId::TypeType));
+  auto type_inst_id = ConvertToValueOfType(context, parse_node, value_id,
+                                           SemIR::TypeId::TypeType);
+  if (type_inst_id == SemIR::InstId::BuiltinError) {
+    return SemIR::TypeId::Error;
+  }
+  auto type_id = context.CanonicalizeType(type_inst_id);
+  if (type_id == SemIR::TypeId::Error) {
+    CARBON_DIAGNOSTIC(TypeConversionFailure, Error,
+                      "Cannot convert expression to type `type`.");
```
zygoloid: Would something like "Cannot evaluate type expression" work better here? We only get here if the actual conversion to type `type` succeeded.

zygoloid (approved):
Thanks!

toolchain/check/convert.cpp
jonmeow: Done

PR #3374, opened 2023-11-08
Support for `returned var` and `return var`.

zygoloid:
Implement toolchain support for `returned var` and `return var`.

- Modeled `returned` in the parse tree as a `ReturnedSpecifier` appearing after the `VariableIntroducer`.
- Modeled `return var` in the parse tree as a `ReturnVarSpecifier` appearing after the `ReturnStatementStart`.
- Factored out the implementation of `return` statement and `returned var` handling in check into a new `return.{h,cpp}`. The parse nodes themselves are still handled in `handle_*.cpp`. This allows easy code reuse between `return` and `returned var`.


toolchain/check/return.cpp
```diff
@@ -0,0 +1,175 @@
...
+}
+
+static auto NoteReturnType(Context& context, Context::DiagnosticBuilder& diag,
+                           const SemIR::Function& function) {
+  // TODO: This is the location of the `fn` keyword. Find the location of the
+  // return type.
+  auto type_parse_node =
+      context.insts().Get(function.declaration_id).parse_node();
+  CARBON_DIAGNOSTIC(ReturnTypeHereNote, Note,
+                    "Return type of function is `{0}`.", std::string);
```
jonmeow, line 49: Maybe "The function's return type is"? The wording just sounds strange to me.

toolchain/parse/handle_statement.cpp
```diff
@@ -179,8 +180,15 @@ auto HandleStatementReturn(Context& context) -> void {
   context.PushState(state);
 
   context.AddLeafNode(NodeKind::ReturnStatementStart, context.Consume());
-  if (!context.PositionIs(Lex::TokenKind::Semi)) {
+
+  if (auto var_token = context.ConsumeIf(Lex::TokenKind::Var)) {
```
jonmeow, line 187: In state.def, please update StatementReturn comments according to this change.

toolchain/parse/node_kind.def
```diff
@@ -261,10 +266,12 @@ CARBON_PARSE_NODE_KIND_CHILD_COUNT(ContinueStatement, 1,
 
 // `return`:
 //   ReturnStatementStart
+//   _optional_ ReturnVarSpecifier
```
jonmeow: I think this should probably be something like "ReturnVarSpecifier or expression", not implying both are allowed.

toolchain/parse/state.def
```diff
@@ -730,8 +730,8 @@ CARBON_PARSE_STATE_VARIANTS3(TypeAfterParams, Class, Interface,
 // Always:
 //   1. PatternAsVariable
 //   2. VarAfterPattern
-//   3. VarFinishAs(Semicolon|For)
-CARBON_PARSE_STATE_VARIANTS2(Var, Semicolon, For)
+//   3. VarFinishAs(Declaration|For)
+CARBON_PARSE_STATE_VARIANTS2(Var, Declaration, For)
```
jonmeow: You've changed both what this state handles and its possible outcomes; please update the documentation.

toolchain/parse/state.def
```diff
@@ -229,7 +229,7 @@ CARBON_PARSE_STATE(DeclarationNameAndParamsAfterImplicit)
 // If `Semi`:
 //   1. DeclarationScopeLoop
 // If `Var`:
-//   1. VarAsSemicolon
+//   1. VarAsDeclaration
```
jonmeow: Please make sure `Returned` handling is documented.

toolchain/parse/handle_var.cpp
```diff
@@ -6,22 +6,41 @@
...
   context.PushState(State::VarAfterPattern);
 
+  // Consume the `var` or `returned var`.
+  auto returned_token = context.ConsumeIf(Lex::TokenKind::Returned);
+  if (returned_token && !context.PositionIs(Lex::TokenKind::Var)) {
+    CARBON_DIAGNOSTIC(ExpectedVarAfterReturned, Error,
+                      "Expected `var` after `returned`.");
+    context.emitter().Emit(*context.position(), ExpectedVarAfterReturned);
+    context.PopAndDiscardState();
+    context.PopAndDiscardState();
```
jonmeow: It looks a bit like you're pushing state that may immediately be discarded. Can the order be switched?

toolchain/parse/handle_var.cpp
```diff
@@ -6,22 +6,41 @@
...
 static auto HandleVar(Context& context, State finish_state) -> void {
   context.PopAndDiscardState();
 
-  // These will start at the `var`.
+  // These will start at the `var` or `returned`.
   context.PushState(finish_state);
   context.PushState(State::VarAfterPattern);
 
+  // Consume the `var` or `returned var`.
+  auto returned_token = context.ConsumeIf(Lex::TokenKind::Returned);
```
jonmeow: Given the `Returned` is already seen from the caller, had you considered a separate state for it so that this wouldn't need to check the token kind again?

toolchain/check/context.h
```diff
@@ -329,9 +343,13 @@ class Context {
     SemIR::NameScopeId scope_id;
 
     // Names which are registered with name_lookup_, and will need to be
-    // deregistered when the scope ends.
+    // unregistered when the scope ends.
     llvm::DenseSet<IdentifierId> names;
 
+    // Whether a `returned var` was introduced in this scope, and needs to be
+    // unregistered when the scope ends.
+    bool has_returned_var = false;
```
jonmeow, line 356: Why is this tracked separately from `return_scope_stack_.back().returned_var`? That is, it looks like checking there would give the same result; what prevented its use?

toolchain/check/context.h
```diff
@@ -115,6 +125,10 @@ class Context {
     return insts().Get(current_scope_inst_id).TryAs<InstT>();
   }
 
+  // Sets the given instruction to be the current `returned var`. If there is
+  // already a `returned var`, returns it instead.
+  auto SetReturnedVarOrGetExisting(SemIR::InstId bind_id) -> SemIR::InstId;
```
jonmeow: What's the return value when it's set?

toolchain/check/context.h
```diff
@@ -30,6 +30,16 @@ class Context {
     SemIR::InstBlockId continue_target;
   };
 
+  // A scope in which `return` can be used.
+  struct ReturnScope {
+    // The declaration from which we can return. Inside a function, this will
+    // be a `FunctionDeclaration`.
+    SemIR::InstId decl_id;
+
+    // The value corresponding to the current `returned var`, if any.
```
jonmeow: Would it be worth documenting that this can be set and unset as `returned var`s go in and out of scope?

toolchain/check/handle_statement.cpp
```diff
@@ -27,51 +28,46 @@ auto HandleExpressionStatement(Context& context, Parse::Node /*parse_node*/)
...
-  CARBON_CHECK(!context.return_scope_stack().empty());
-  auto fn_inst = context.insts().GetAs<SemIR::FunctionDeclaration>(
-      context.return_scope_stack().back());
-  const auto& callable = context.functions().Get(fn_inst.function_id);
-
-  if (context.parse_tree().node_kind(context.node_stack().PeekParseNode()) ==
-      Parse::NodeKind::ReturnStatementStart) {
-    context.node_stack()
-        .PopAndDiscardSoloParseNode<Parse::NodeKind::ReturnStatementStart>();
+auto HandleReturnStatementStart(Context& context, Parse::Node parse_node)
```
jonmeow: Would it be worth breaking this out to a separate `handle_return`, maybe leaving behind a small and boring `handle_expression_statement`?

Partly, I assume the BuildReturn functions can only be used here: it feels odd to me that I need to go looking in a separate file for their implementation. Maybe you won't want to re-merge, but even then, it feels odd to me to have `ReturnVarSpecifier` handled here. You'd already said you had trouble finding `return`, so maybe a split would help that?

toolchain/check/return.h
```diff
@@ -0,0 +1,37 @@
...
+auto CheckReturnedVar(Context& context, Parse::Node returned_node,
+                      Parse::Node name_node, IdentifierId name_id,
+                      Parse::Node type_node, SemIR::TypeId type_id)
+    -> SemIR::InstId;
+
+// Registers the given binding as the current `returned var` in this scope.
+auto RegisterReturnedVar(Context& context, SemIR::InstId bind_id) -> void;
+
+// Checks and builds SemIR for a `return;` statement.
+auto BuildReturnWithNoExpression(Context& context, Parse::Node parse_node)
```
jonmeow: There had been an argument for uniform abbreviations due to instance -> inst, which I believe included expression -> expr. Do you want to do that here?

toolchain/check/return.cpp
```diff
@@ -0,0 +1,175 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "toolchain/check/context.h"
+#include "toolchain/check/convert.h"
+#include "toolchain/parse/tree.h"
+#include "toolchain/sem_ir/inst.h"
+
+namespace Carbon::Check {
+
+static auto GetCurrentFunction(Context& context) -> SemIR::Function& {
```
jonmeow, line 13: Would it be worth quick comments on the `static` functions in this file?

toolchain/check/return.h
```diff
@@ -0,0 +1,37 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#ifndef CARBON_TOOLCHAIN_CHECK_RETURN_H_
```
jonmeow, line 5: I understand the desire for helper files like this, but maybe it should be named something like "return_and_returned_helpers" to make clearer the contents? "return" is a very overloaded term, so having a file with just that as the name might carry over that ambiguity, especially since it isn't specific to `return`.

jonmeow (approved):

toolchain/check/return.cpp
zygoloid: I think there's a bigger question here of how we want our diagnostics to be phrased and presented. "The function's return type is `X`." sounds really strange to me as a note in a compiler diagnostic, and it's very different from what C++ compilers produce. We're already departing from convention in various other ways, though. Let's discuss this on discord and see if we can establish a consensus position.

toolchain/parse/handle_statement.cpp
zygoloid: Done.

toolchain/parse/node_kind.def
zygoloid: Done.

toolchain/parse/state.def
zygoloid: Done.

toolchain/parse/state.def
zygoloid: Done.

toolchain/parse/handle_var.cpp
zygoloid: Done. It's slightly awkward because we need to enter these states before we create the introducer which ends up being a child of the nodes created for them, but want to parse the introducer before we decide whether to enter the states.

toolchain/parse/handle_var.cpp
zygoloid: I did try that, but previously had been trying to keep this `ConsumeIf` after the `PushState`s, which made it awkward. Seems to work a bit better now. Done.

toolchain/check/context.h
zygoloid: This is checking whether there's a `returned var` in the current scope, whereas the `return_scope_stack` value is checking whether there's a `returned var` for the current function. The difference is important because a function can have multiple `returned var`s in different scopes, and we regain the ability to declare a `returned var` when we leave the scope that the current `returned var` is declared within.

See [this section of the design](https://github.com/carbon-language/carbon-lang/blob/trunk/proposals/p0257.md#:~:text=We%20propose%20a%20set%20of%20restrictions%20to%20give%20simple%20and%20understandable%20behavior%20which%20remains%20reasonably%20expressive.) for the rules this is implementing, and `check/testdata/returned_var_scope.carbon` and `check/testdata/fail_returned_var_shadow.carbon`.

toolchain/check/context.h
zygoloid: Extended comment to specify.

toolchain/check/context.h
zygoloid: For me, that's clear from the "current" and "if any", but happy to add more words here if you would like?

toolchain/check/handle_statement.cpp
zygoloid: Sure, split `handle_statement.cpp` into `handle_expression_statement.cpp` and `handle_return_statement.cpp`.

I think it's useful to keep a separate, non-`handle_` file for the `return ...` and `returned var` processing, because they share logic. It wouldn't make sense to put `HandlePatternBinding` in the same file as `HandleReturnStatement`, and I don't think we want to have the `handle_*` files expose any functions other than their `Handle*` functions, so I wouldn't want to re-merge.

toolchain/check/return.h
zygoloid: I figure we're going to do a wholesale search-and-replace, so there's no point doing this one as a one-off.

toolchain/check/return.cpp
zygoloid: Done.

toolchain/check/return.h
zygoloid: I prefer it the way it is. "_helpers" tells me nothing, so I'd not want to add that part. I would be OK with `return_and_returned`, and will switch to that if you have a strong preference here, but I think they're closely-enough connected that simply `return` is sufficient.

toolchain/check/context.h
jonmeow:
```suggestion
    // The value corresponding to the current `returned var`, if any. Changes during scope processing.
```

I think this is my mindset that scope information doesn't typically mutate. e.g., `decl_id` won't, right?

toolchain/check/return.h
jonmeow: My concern is that there are all the `handle_*` files, and by being named after a statement, I'm slightly disoriented that this isn't another `handle_*` file that handles that statement; that is, `handle_return` would be sensible to see. The same would be true of `return_and_returned`, so it wouldn't offer clarity on this point.

On a related note, `convert` doesn't go the same way for me since there's no `convert` statement, though it's still a little weird of a name to me.

Not a strong opinion that some additional disambiguation should be blocking, it's more that I was hoping to reduce the source of this confusion somehow.

toolchain/parse/handle_var.cpp
```diff
@@ -6,22 +6,44 @@
...
+static auto HandleVar(Context& context, State finish_state,
+                      Lex::Token returned_token = Lex::Token::Invalid) -> void {
+  auto start_token = context.PopState().token;
 
-  // These will start at the `var`.
-  context.PushState(finish_state);
-  context.PushState(State::VarAfterPattern);
+  // These will start at the `var` or `returned`.
+  context.PushState(finish_state, start_token);
+  context.PushState(State::VarAfterPattern, start_token);
```
jonmeow: We typically take the full state, not just the token (even if we don't do more). Also, this variant of PushState says "Used when forming a new subtree with a token that isn't the start of the subtree.", which isn't quite what's happening here. Would you be okay with:

```suggestion
  auto state = context.PopState();

  // These will start at the `var` or `returned`.
  state.state = finish_state;
  context.PushState(state);
  state.state = State::VarAfterPattern;
  context.PushState(state);
```

Note, I think this may also be a little more resilient with code changes, because it propagates subtree_start and has_error. The current code works because there are no token being consumed and no nodes produced between, and implicit requirement.

Though, more broadly, maybe you'd be happier with something closer to what you had previously:

```
static auto HandleVar(Context& context, State finish_state,
                      bool is_returned) -> void {

  if (is_returned) {
    auto returned_token = context.ConsumeChecked(Lex::TokenKind::Returned);
    if (auto var_token = context.ConsumeIf(Lex::TokenKind::Var)) {
      context.AddLeafNode(NodeKind::VariableIntroducer, *var_token);
      context.AddLeafNode(NodeKind::ReturnedSpecifier, returned_token);
    } else {
      CARBON_DIAGNOSTIC(ExpectedVarAfterReturned, Error,
                        "Expected `var` after `returned`.");
      context.emitter().Emit(*context.position(), ExpectedVarAfterReturned);
      auto semi = context.SkipPastLikelyEnd(returned_token);
      context.AddLeafNode(NodeKind::EmptyDeclaration,
                          semi ? *semi : returned_token,
                          /*has_error=*/true);
      return;
    }
  } else {
    context.AddLeafNode(NodeKind::VariableIntroducer, context.ConsumeChecked(Lex::TokenKind::Var));
  }

  // These will start at the `var` or `returned`.
  state.state = finish_state;
  context.PushState(state);
  state.state = State::VarAfterPattern;
  context.PushState(state);

  // This will start at the pattern.
  context.PushState(State::PatternAsVariable);
}
```

I think the main thing that does is it removes the `context.PopAndDiscardState` in error handling, which I don't think we do elsewhere (typical is that we have one spot near the start that pops, with the only exception being loop states, which pop on the end of the loop). But, maybe you'd argue these kinds of patterns like ensuring we consistently pop at the start are wasteful and not doing enough to prevent issues?

toolchain/check/return.cpp
jonmeow: This should be non-blocking for the PR.

jonmeow (approved):
Looks good, two last suggestions. (I think you're good to submit whichever way you take `HandleVar`)

NB, the `HandleVar` comment does make me kind of want a test of the `for` interaction, because I started thinking the prior returned detection could have missed the error in `for (returned var x: i32 in y)`

toolchain/parse/handle_var.cpp
zygoloid: I know we do this in other places already, but I find this idiom of "mutate an existing state and then push it" to be confusing. It's unclear to me what that means -- which parts of the existing state are being intentionally preserved, and why -- and it's unclear that it would remain correct as more information is added to the state. (For example, this propagates the `has_error` bit from the parent state to two of the three child states, but is that really intentional or just happenstance?)

I'm making this change anyway, but I think we should think a bit about whether there's a way to make this kind of state manipulation clearer. Perhaps there's a different way that we can think about states that would help, such as, as a pending parse node.

(time passes)

I ended up using this modify-and-push approach for the first state only. That's the one that represents a pending parse node, and really should be inheriting information from the parent state. The second state doesn't represent a pending parse node, and its token and subtree start don't matter, so I think it's simpler to push a fresh state in that case. This also, I think, makes it clearer that the `has_error` flag is inherited onto only a single node.

> The current code works because there are no token being consumed and no nodes produced between, and implicit requirement.

I think it only requires that no nodes are produced between -- tokens *were* being consumed here, which is why we needed to pass the original start token to `PushState`. But yeah, it's definitely brittle -- and in fact I think maybe it's even more brittle than you were suggesting, because some changes work and some don't.

toolchain/check/context.h
zygoloid: Done.

PR #3375, opened 2023-11-09
Use abbreviation "expr" instead of "expression"

josh11b:


chandlerc (approved):
LG

Not sure about `Subexpr`, but let's try it as you have it here. We can revisit this easily if needed.

PR #3378, opened 2023-11-09
Extend description of parser states to show which tokens they consume.

zygoloid:
Also some minor improvements and typo fixes to the parser code for issues found while writing these descriptions.


toolchain/parse/state.def
```diff
@@ -387,347 +534,479 @@ CARBON_PARSE_STATE(FunctionIntroducer)
...
 CARBON_PARSE_STATE_VARIANTS3(TypeIntroducer, Class, Interface, NamedConstraint)
 
 // Handles processing of a type after its optional parameters.
 //
-// If `Semi`:
+// class/interface/constraint name ( ... ) ;
+//                                         ^
+// class/interface/constraint name ( ... ) ???
+//                                        ^
 //   (state done)
```
josh11b, line 1011: Move this after the `{` case?

josh11b (commented):
Thank you so much for doing this! The additional context really makes it much easier for me to understand.

toolchain/parse/state.def
zygoloid: Done.

josh11b (approved):

PR #3379, opened 2023-11-09
Refactor checking flow to allow for ordering based on import/package.

jonmeow:
As I was working on this, I noticed `import` and `library` syntax needs to be fixed for how it imports the current package, and for `Main` libraries. This mostly reflects the current state in its testing.

Otherwise, this should handle most of the errors I could think of: dependency cycles, redundant imports, etc.

It does not actually deal with the nuances of cross-IR references.

### DIFF
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -125,9 +125,8 @@ static auto TrackImport(
   if (package_key.first == ExplicitMainName) {
     // Implicit imports will have already warned.
     if (explicit_import_map) {
-      CARBON_DIAGNOSTIC(
-          ImportMainPackage, Error,
-          "Cannot import the `Main` package from another package.");
+      CARBON_DIAGNOSTIC(ImportMainPackage, Error,
+                        "Cannot import `Main` from other packages.");
       unit_info.emitter.Emit(import.node, ImportMainPackage);
     }
     return;
@@ -139,8 +138,8 @@ static auto TrackImport(
             explicit_import_map->insert({package_key, import.node});
         !success) {
       CARBON_DIAGNOSTIC(RepeatedImport, Error,
-                        "Library is imported more than once.");
-      CARBON_DIAGNOSTIC(FirstImported, Note, "First imported here.");
+                        "Library imported more than once.");
+      CARBON_DIAGNOSTIC(FirstImported, Note, "First import here.");
       unit_info.emitter.Build(import.node, RepeatedImport)
           .Note(insert_it->second, FirstImported)
           .Emit();
@@ -153,9 +152,8 @@ static auto TrackImport(
         import.package_id == package_directive->package.package_id &&
         import.library_id == package_directive->package.library_id) {
       CARBON_DIAGNOSTIC(ExplicitImportApi, Error,
-                        "The `impl` for a library implicitly imports the `api` "
-                        "and must not do so explicitly.");
-      CARBON_DIAGNOSTIC(ImportSelf, Error, "A file must not import itself.");
+                        "Omit import of library `api`.");
+      CARBON_DIAGNOSTIC(ImportSelf, Error, "Library cannot import itself.");
       unit_info.emitter.Emit(import.node, package_directive->api_or_impl ==
                                                   Parse::Tree::ApiOrImpl::Impl
                                               ? ExplicitImportApi
@@ -200,14 +198,13 @@ auto CheckParseTrees(const SemIR::File& builtin_ir,
       // APIs.
       if (package_key.first == ExplicitMainName) {
         if (package_key.second.empty()) {
-          CARBON_DIAGNOSTIC(
-              ExplicitMainPackage, Error,
-              "Omit the `package` directive to define the `Main` package.");
+          CARBON_DIAGNOSTIC(ExplicitMainPackage, Error,
+                            "Omit `package` directive for `Main` package.");
           unit_info.emitter.Emit(package->package.node, ExplicitMainPackage);
         } else {
-          CARBON_DIAGNOSTIC(ExplicitMainLibrary, Error,
-                            "Use the `library` directive to define libraries "
-                            "in the `Main` package.");
+          CARBON_DIAGNOSTIC(
+              ExplicitMainLibrary, Error,
+              "Use `library` directive in `Main` package libraries.");
           unit_info.emitter.Emit(package->package.node, ExplicitMainLibrary);
         }
         continue;
@@ -221,7 +218,7 @@ auto CheckParseTrees(const SemIR::File& builtin_ir,
       if (!success) {
         // TODO: Cross-reference the source, deal with library, etc.
         CARBON_DIAGNOSTIC(DuplicateLibraryApi, Error,
-                          "Multiple files declare the same library's API.");
+                          "Library's API declared in more than one file.");
         unit_info.emitter.Emit(package->package.node, DuplicateLibraryApi);
       }
     }
@@ -278,9 +275,9 @@ auto CheckParseTrees(const SemIR::File& builtin_ir,
           if (*import_unit->sem_ir) {
             ++import_it;
           } else {
-            CARBON_DIAGNOSTIC(ImportCycleDetected, Error,
-                              "Import is part of a cycle. The cycle must be "
-                              "fixed before the import can be used.");
+            CARBON_DIAGNOSTIC(
+                ImportCycleDetected, Error,
+                "Import is in cycle. Cycle must be fixed to import.");
             unit_info.emitter.Emit(import_it->first, ImportCycleDetected);
             import_it = unit_info.imports.erase(import_it);
           }
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -154,8 +154,8 @@ static auto TrackImport(
     if (packaging && import.package_id == packaging->names.package_id &&
         import.library_id == packaging->names.library_id) {
       CARBON_DIAGNOSTIC(ExplicitImportApi, Error,
-                        "Explicit import of `api` from `impl` file is redundant "
-                        "with implicit import.");
+                        "Explicit import of `api` from `impl` file is "
+                        "redundant with implicit import.");
       CARBON_DIAGNOSTIC(ImportSelf, Error, "File cannot import itself.");
       unit_info.emitter.Emit(
           import.node, packaging->api_or_impl == Parse::Tree::ApiOrImpl::Impl
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -97,7 +97,7 @@ static auto CheckParseTree(const SemIR::File& builtin_ir, UnitInfo& unit_info,
 #endif
 }
 
-// The package and library names, used as map keys
+// The package and library names, used as map keys.
 using ImportKey = std::pair<llvm::StringRef, llvm::StringRef>;
 
 // Returns a key form of the package object.
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -150,18 +150,17 @@ static auto TrackImport(
 
     // Check for explicit imports of the same library. The ID comparison is okay
     // in this case because both come from the same file.
-    auto package_directive = unit_info.unit->parse_tree->package();
-    if (package_directive &&
-        import.package_id == package_directive->package.package_id &&
-        import.library_id == package_directive->package.library_id) {
+    auto packaging = unit_info.unit->parse_tree->packaging_directive();
+    if (packaging && import.package_id == packaging->names.package_id &&
+        import.library_id == packaging->names.library_id) {
       CARBON_DIAGNOSTIC(ExplicitImportApi, Error,
                         "Explicit import `api` of `impl` file is redundant "
                         "with implicit import.");
       CARBON_DIAGNOSTIC(ImportSelf, Error, "File cannot import itself.");
-      unit_info.emitter.Emit(import.node, package_directive->api_or_impl ==
-                                                  Parse::Tree::ApiOrImpl::Impl
-                                              ? ExplicitImportApi
-                                              : ImportSelf);
+      unit_info.emitter.Emit(
+          import.node, packaging->api_or_impl == Parse::Tree::ApiOrImpl::Impl
+                           ? ExplicitImportApi
+                           : ImportSelf);
       return;
     }
   }
@@ -195,9 +194,9 @@ auto CheckParseTrees(const SemIR::File& builtin_ir,
   // Create a map of APIs which might be imported.
   llvm::DenseMap<ImportKey, UnitInfo*> api_map;
   for (auto& unit_info : unit_infos) {
-    const auto& package = unit_info.unit->parse_tree->package();
-    if (package) {
-      auto import_key = GetImportKey(unit_info, package->package);
+    const auto& packaging = unit_info.unit->parse_tree->packaging_directive();
+    if (packaging) {
+      auto import_key = GetImportKey(unit_info, packaging->names);
       // Catch explicit `Main` errors before they become marked as possible
       // APIs.
       if (import_key.first == ExplicitMainName) {
@@ -207,13 +206,13 @@ auto CheckParseTrees(const SemIR::File& builtin_ir,
         CARBON_DIAGNOSTIC(
             ExplicitMainLibrary, Error,
             "Use `library` directive in `Main` package libraries.");
-        unit_info.emitter.Emit(package->package.node,
+        unit_info.emitter.Emit(packaging->names.node,
                                import_key.second.empty() ? ExplicitMainPackage
                                                          : ExplicitMainLibrary);
         continue;
       }
 
-      if (package->api_or_impl == Parse::Tree::ApiOrImpl::Impl) {
+      if (packaging->api_or_impl == Parse::Tree::ApiOrImpl::Impl) {
         continue;
       }
 
@@ -222,7 +221,7 @@ auto CheckParseTrees(const SemIR::File& builtin_ir,
         // TODO: Cross-reference the source, deal with library, etc.
         CARBON_DIAGNOSTIC(DuplicateLibraryApi, Error,
                           "Library's API declared in more than one file.");
-        unit_info.emitter.Emit(package->package.node, DuplicateLibraryApi);
+        unit_info.emitter.Emit(packaging->names.node, DuplicateLibraryApi);
       }
     }
   }
@@ -231,10 +230,10 @@ auto CheckParseTrees(const SemIR::File& builtin_ir,
   llvm::SmallVector<UnitInfo*> ready_to_check;
   ready_to_check.reserve(units.size());
   for (auto& unit_info : unit_infos) {
-    const auto& package = unit_info.unit->parse_tree->package();
-    if (package && package->api_or_impl == Parse::Tree::ApiOrImpl::Impl) {
+    const auto& packaging = unit_info.unit->parse_tree->packaging_directive();
+    if (packaging && packaging->api_or_impl == Parse::Tree::ApiOrImpl::Impl) {
       // An `impl` has an implicit import of its `api`.
-      TrackImport(api_map, nullptr, unit_info, package->package);
+      TrackImport(api_map, nullptr, unit_info, packaging->names);
     }
 
     llvm::DenseMap<ImportKey, Parse::Node> explicit_import_map;
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -40,6 +40,8 @@ struct UnitInfo {
 };
 
 // Produces and checks the IR for the provided Parse::Tree.
+// TODO: Both valid and invalid imports should be recorded on the SemIR. Invalid
+// imports should suppress errors where it makes sense.
 static auto CheckParseTree(const SemIR::File& builtin_ir, UnitInfo& unit_info,
                            llvm::raw_ostream* vlog_stream) -> void {
   unit_info.unit->sem_ir->emplace(*unit_info.unit->value_stores,
@@ -99,8 +101,8 @@ static auto CheckParseTree(const SemIR::File& builtin_ir, UnitInfo& unit_info,
 using ImportKey = std::pair<llvm::StringRef, llvm::StringRef>;
 
 // Returns a key form of the package object.
-static auto GetImportKey(UnitInfo& unit_info, Parse::Tree::Package package)
-    -> ImportKey {
+static auto GetImportKey(UnitInfo& unit_info,
+                         Parse::Tree::PackagingNames package) -> ImportKey {
   auto* stores = unit_info.unit->value_stores;
   return {package.package_id.is_valid()
               ? stores->identifiers().Get(package.package_id)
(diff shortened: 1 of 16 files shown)


toolchain/check/check.cpp
```diff
@@ -56,16 +85,218 @@ auto CheckParseTree(SharedValueStores& value_stores,
...
+              ? stores->string_literals().Get(package.library_id)
+              : ""};
+}
+
+static constexpr llvm::StringLiteral ExplicitMainName = "Main";
 
-  return sem_ir;
+// Marks an import as required on both the source and target file. seen_imports
+// is optional, and will be omitted for implicit imports (which are disallowed
+// when explicit).
```
zygoloid: Is this comment out of date? `seen_imports` isn't declared here.

toolchain/check/check.cpp
```diff
@@ -56,16 +85,218 @@ auto CheckParseTree(SharedValueStores& value_stores,
...
+static auto TrackImport(
+    llvm::DenseMap<std::pair<llvm::StringRef, llvm::StringRef>, UnitInfo*>&
```
zygoloid: I think it'd help readability a lot to add a typedef for the key type here (maybe `LibraryName` or `LibraryKey`).

toolchain/check/check.cpp
```diff
@@ -56,16 +85,218 @@ auto CheckParseTree(SharedValueStores& value_stores,
...
+    UnitInfo& unit_info, Parse::Tree::Package import) -> void {
+  auto package_key = PackageKey(unit_info, import);
+
+  // Specialize the error for imports from `Main`.
+  if (package_key.first == ExplicitMainName) {
+    // Implicit imports will have already warned.
+    if (explicit_import_map) {
+      CARBON_DIAGNOSTIC(ImportMainPackage, Error,
+                        "The `Main` package cannot be mentioned explicitly on "
+                        "imports; it can only be imported implicitly");
```
zygoloid: I think it'd be useful to split this diagnostic into two:

- If the current library and the imported library are in the same package: "Use `import library` instead of `import package PackageName library` to import from the same package." or something like that.
- If the current package is not `Main`: "Cannot import `Main` package from another package."

It looks like the first of those two rules isn't being enforced yet.

toolchain/check/check.cpp
```diff
@@ -56,16 +85,218 @@ auto CheckParseTree(SharedValueStores& value_stores,
...
+    return;
+  }
+
+  if (explicit_import_map) {
+    // Check for redundant imports.
+    if (auto [insert_it, success] =
+            explicit_import_map->insert({package_key, import.node});
+        !success) {
+      CARBON_DIAGNOSTIC(RepeatedImport, Error,
+                        "Import matches an earlier import");
```
zygoloid: We use "matches" in the design to mean something else, so it might be better to avoid that word; I wonder if "Library imported more than once" or something like that would work here?

toolchain/check/check.cpp
```diff
@@ -56,16 +85,218 @@ auto CheckParseTree(SharedValueStores& value_stores,
...
+        if (package_key.second.empty()) {
+          CARBON_DIAGNOSTIC(ExplicitMainPackage, Error,
+                            "The `Main` package is implicit and not allowed in "
+                            "the `package` directive");
+          unit_info.emitter.Emit(package->package.node, ExplicitMainPackage);
+        } else {
+          CARBON_DIAGNOSTIC(ExplicitMainLibrary, Error,
+                            "The `Main` package is implicit; libraries should "
+                            "omit `package Main`",
+                            std::string);
```
zygoloid: Given that we can be confident we know what the developer meant here, I wonder if it'd be better to phrase these as instructions:
```suggestion
          CARBON_DIAGNOSTIC(ExplicitMainPackage, Error,
                            "Omit the `package` declaration to define the `Main` package.");
          unit_info.emitter.Emit(package->package.node, ExplicitMainPackage);
        } else {
          CARBON_DIAGNOSTIC(ExplicitMainLibrary, Error,
                            "Omit `package Main` when defining libraries in the `Main` package.");
```
What do you think? I think the first of these diagnostics in particular would benefit by giving more of an instruction.

toolchain/check/check.cpp
```diff
@@ -56,16 +85,218 @@ auto CheckParseTree(SharedValueStores& value_stores,
...
+          unit_info.emitter.Emit(package->package.node, ExplicitMainLibrary,
+                                 package_key.second.str());
```
zygoloid: You pass the library name here but don't use it in the diagnostic.

toolchain/check/check.cpp
```diff
@@ -56,16 +85,218 @@ auto CheckParseTree(SharedValueStores& value_stores,
...
+
+    llvm::DenseMap<std::pair<llvm::StringRef, llvm::StringRef>, Parse::Node>
+        explicit_import_map;
+    for (const auto& import : unit_info.unit->parse_tree->imports()) {
+      TrackImport(api_map, &explicit_import_map, unit_info, import);
+    }
+
+    if (unit_info.imports_remaining == 0) {
+      ready_to_check.push_back(&unit_info);
+    }
```
zygoloid, line 247: I think this needs to be done as a separate step after we've tracked all imports. Eg, if we have unit A with no imports, then unit B imports A, then when we finish processing A it will have 0 imports remaining, but isn't ready to be checked yet.

toolchain/check/check.cpp
```diff
@@ -56,16 +85,218 @@ auto CheckParseTree(SharedValueStores& value_stores,
...
+      }
+    }
+
+    // Check the remaining file contents, which are probably broken due to
+    // incomplete imports.
+    for (auto& unit_info : unit_infos) {
+      if (unit_info.imports_remaining > 0) {
+        CheckParseTree(builtin_ir, unit_info, vlog_stream);
+      }
+    }
```
zygoloid, line 295: Is this worth doing? I think we expect spurious errors here, as a result of the cycle error we already diagnosed.

toolchain/check/testdata/packages/explicit_imports.carbon
```diff
@@ -0,0 +1,35 @@
...
+
+package Api library "lib" api;
+
+// --- import_api.carbon
+
+import Api;
+import Api library "lib";
+
+// CHECK:STDOUT: file "api.carbon" {
+// CHECK:STDOUT: }
```
zygoloid, line 21: (Non-blocking.) It'd be nice to extend the print here to include details about the package, now we're tracking them in a convenient place.

toolchain/check/testdata/packages/explicit_imports.carbon
```diff
@@ -0,0 +1,35 @@
...
+// CHECK:STDOUT: constants {
```
zygoloid, line 22: (Non-blocking.) Some kind of separator between files in the output would be nice.

toolchain/check/check.cpp
```diff
@@ -56,16 +85,218 @@ auto CheckParseTree(SharedValueStores& value_stores,
...
+      if (unit_info.imports_remaining > 0) {
+        for (auto* import_it = unit_info.imports.begin();
+             import_it != unit_info.imports.end();) {
+          auto* import_unit = import_it->second->unit;
+          if (*import_unit->sem_ir) {
+            ++import_it;
+          } else {
+            CARBON_DIAGNOSTIC(
+                ImportCycleDetected, Error,
+                "Import is part of a cycle which could not be resolved");
```
zygoloid: The phrasing of this suggests to me that there might be some cycles that can be resolved.

zygoloid (commented):
Structure and approach looks great, some comments on the details.

The new diagnostics seem to omit a final period, that we include in (nearly, see #3380) our current diagnostics. Would be good to add those in for consistency.

toolchain/check/check.cpp
jonmeow: Removed.

toolchain/check/check.cpp
jonmeow: Done.

I'm inferring from your `LibraryKey` suggestion that you dislike the current `PackageKey`. I feel kind of odd with both `Package` and `Library` since it's a mix of package and library name, but the typical use-case is the package directive -- thus where I'd ended up with `PackageKey`.

Switching to `ImportKey` though because maybe that's similarly accurate and doesn't focus on either component.

toolchain/check/check.cpp
jonmeow: > I think it'd be useful to split this diagnostic into two:
> 
> * If the current library and the imported library are in the same package: "Use `import library` instead of `import package PackageName library` to import from the same package." or something like that.

I've added a TODO for this, as I'm already working on the implicit package name support. I'd like to test it together with that.

> * If the current package is not `Main`: "Cannot import `Main` package from another package."

I've rephrased the warning.

> It looks like the first of those two rules isn't being enforced yet.

I'm not sure it's splitting when one isn't handled. ;)

toolchain/check/check.cpp
jonmeow: Done.

toolchain/check/check.cpp
jonmeow: Done, emphasizing the `library` directive instead for the second (also, directive vs declaration).

toolchain/check/check.cpp
jonmeow: Cruft, removed.

toolchain/check/check.cpp
jonmeow: That would only be if checking immediately started. No checking is done in this loop.

toolchain/check/check.cpp
jonmeow: Rephrased, just "Cycle must be fixed to import"

toolchain/check/check.cpp
jonmeow: Isn't this consistent with error checking in general? We don't stop at the first error; instead, we keep producing errors even though the earlier error may render the later code invalid. This is true between phases, and inside phases. The only exception is we don't lower if any earlier phase failed.

Also, checking always produces a SemIR this way.

But the cyclic `import` might've also been unnecessary; remaining code could be correct.

toolchain/check/testdata/packages/explicit_imports.carbon
jonmeow: Had YAML been considered for SemIR output? It'd have pros and cons; I think this output is more LLVM IR-like, but we use YAML elsewhere, and YAML would provide a separator.

toolchain/check/testdata/packages/explicit_imports.carbon
jonmeow: There's already a TODO for this in the formatter. Since it would touch every semir test, and may be subject to preferences about format, I'd rather split it out.

jonmeow (commented):
Also did a pass of diagnostics to phrase tersely, stop trying for good grammar.

toolchain/check/check.cpp
```diff
@@ -56,16 +85,214 @@ auto CheckParseTree(SharedValueStores& value_stores,
...
+static constexpr llvm::StringLiteral ExplicitMainName = "Main";
+
+// Marks an import as required on both the source and target file.
+// TODO: When importing without a package name is supported, check that it's
+// used correctly.
+static auto TrackImport(
+    llvm::DenseMap<ImportKey, UnitInfo*>& api_map,
+    llvm::DenseMap<ImportKey, Parse::Node>* explicit_import_map,
+    UnitInfo& unit_info, Parse::Tree::Package import) -> void {
+  auto package_key = GetImportKey(unit_info, import);
```
zygoloid: `package_key` -> `import_key`

toolchain/check/check.cpp
```diff
@@ -56,16 +85,214 @@ auto CheckParseTree(SharedValueStores& value_stores,
...
 
 #ifndef NDEBUG
   if (auto verify = sem_ir.Verify(); !verify.ok()) {
     CARBON_FATAL() << sem_ir << "Built invalid semantics IR: " << verify.error()
                    << "\n";
   }
 #endif
+}
+
+// The package and library names, used as map keys
```
```suggestion
// The package and library names, used as map keys.
```

toolchain/check/check.cpp
```diff
@@ -56,16 +85,214 @@ auto CheckParseTree(SharedValueStores& value_stores,
...
+      return;
+    }
+
+    // Check for explicit imports of the same library.
+    auto package_directive = unit_info.unit->parse_tree->package();
+    if (package_directive &&
+        import.package_id == package_directive->package.package_id &&
+        import.library_id == package_directive->package.library_id) {
+      CARBON_DIAGNOSTIC(ExplicitImportApi, Error,
+                        "Omit import of library `api`.");
```
zygoloid: This error doesn't indicate to me what the problem was (that the import was of the same library that we're defining). Maybe we can just use the same "Library cannot import itself." diagnostic regardless of whether we're in an `impl`?

toolchain/check/check.cpp
```diff
@@ -56,16 +85,214 @@ auto CheckParseTree(SharedValueStores& value_stores,
...
+  for (auto& unit : units) {
+    unit_infos.emplace_back(unit);
+  }
+
+  // Create a map of APIs which might be imported.
+  llvm::DenseMap<ImportKey, UnitInfo*> api_map;
+  for (auto& unit_info : unit_infos) {
+    const auto& package = unit_info.unit->parse_tree->package();
+    if (package) {
+      auto package_key = GetImportKey(unit_info, package->package);
```
zygoloid: `package_key` -> `import_key`

toolchain/parse/tree.h
```diff
@@ -62,6 +62,26 @@ class Tree : public Printable<Tree> {
...
+    Impl,
+  };
+
+  // Used for both `package` and `import`. Links back to the node for
+  // diagnostics.
+  struct Package {
+    Node node;
+    IdentifierId package_id = IdentifierId::Invalid;
+    StringLiteralId library_id = StringLiteralId::Invalid;
+  };
```
zygoloid: I think this is more of a `LibraryName` than a `Package`, given that it contains both the package name and the library name within the package.

toolchain/parse/tree.h
```diff
@@ -62,6 +62,26 @@ class Tree : public Printable<Tree> {
...
+
+  // `package` information.
+  struct PackageDirective {
```
zygoloid: I expect we'll also use this for the `library` directives in the main package. Perhaps `PackagingDirective` would be a good name to cover both cases?

I'm fine with keeping it as-is for now, and only renaming once we add support for `library` directives.

toolchain/check/testdata/packages/explicit_imports.carbon
zygoloid: We explicitly wanted something that's represented more as an IR-style programming language, rather than a machine- (and hopefully human-)readable data structure serialization format like YAML.

But we could use a YAML-style `---` separator here anyway :)

In any case, please don't block on this.

toolchain/check/check.cpp
```diff
@@ -56,16 +85,214 @@ auto CheckParseTree(SharedValueStores& value_stores,
...
+      CARBON_DIAGNOSTIC(RepeatedImport, Error,
+                        "Library imported more than once.");
+      CARBON_DIAGNOSTIC(FirstImported, Note, "First import here.");
+      unit_info.emitter.Build(import.node, RepeatedImport)
+          .Note(insert_it->second, FirstImported)
+          .Emit();
+      return;
+    }
+
+    // Check for explicit imports of the same library.
```
zygoloid: I think it'd be useful for this comment to say that the ID comparison here is OK because the package and import directives are from the same file.

toolchain/check/check.cpp
```diff
@@ -56,16 +85,214 @@ auto CheckParseTree(SharedValueStores& value_stores,
...
+      if (!success) {
+        // TODO: Cross-reference the source, deal with library, etc.
+        CARBON_DIAGNOSTIC(DuplicateLibraryApi, Error,
+                          "Library's API declared in more than one file.");
+        unit_info.emitter.Emit(package->package.node, DuplicateLibraryApi);
+      }
+    }
+  }
+
+  // TODO: Detect self import.
```
zygoloid: This appears to be done.

toolchain/check/check.cpp
zygoloid: I just had the dependency edges reversed in my head. This is fine because the unit can't acquire any more imports from the processing of later units.

toolchain/check/check.cpp
```diff
@@ -56,16 +85,214 @@ auto CheckParseTree(SharedValueStores& value_stores,
...
+      if (unit_info.imports_remaining > 0) {
+        for (auto* import_it = unit_info.imports.begin();
+             import_it != unit_info.imports.end();) {
+          auto* import_unit = import_it->second->unit;
+          if (*import_unit->sem_ir) {
+            ++import_it;
+          } else {
+            CARBON_DIAGNOSTIC(
+                ImportCycleDetected, Error,
+                "Import is in cycle. Cycle must be fixed to import.");
```
zygoloid: The imports we identify here aren't necessarily in a cycle. Eg, if we have A imports B imports C imports D imports C, then only C and D are in a cycle, but this will identify A's import of B as well, which is not in a cycle, but can't be imported because of a cycle.

Maybe something like "Library cannot be imported due to a cycle." or "Import cannot be performed due to a cycle." would be more accurate than "Import is in a cycle."?

toolchain/check/check.cpp
zygoloid: In general we try to not produce follow-on errors that are caused by our recovery from a previously-reported error, but we do want to produce errors that are not a consequence of another error we've already reported.

Perhaps we could follow a more nuanced approach along those lines here: set a flag in check to say we had an import error, and check the parse tree in a mode that doesn't produce errors on failed unqualified lookups / lookups into other packages.

If that sounds good to you, maybe just add a TODO here to suppress follow-on errors caused by the missing imports?

toolchain/check/check.cpp
zygoloid: My thinking was that library here is more specific than package (a library is part of a package), so a `LibraryKey` would necessarily also identify the package, but perhaps that's ambiguous / unclear. I'm fine with `ImportKey`.

toolchain/check/check.cpp
zygoloid: > I'm not sure it's splitting when one isn't handled. ;)

You are currently enforcing the first rule for the `Main` package, but not for any other package, which is what I was suggesting you could split. The diagnostic we currently get from importing a `Main` library from another `Main` library  is incorrect: "Cannot import `Main` from other packages."

But no action required here: I'm fine with this as-is given the TODO you added.

toolchain/check/testdata/packages/explicit_imports.carbon
zygoloid: Yes, please don't block this PR on this.

toolchain/parse/tree.h
zygoloid: Given the other discussion on `LibraryKey`, maybe `LibraryName` is similarly ambiguous, and `LibraryPath` or `PackageAndLibraryName` or something like that would better avoid the ambiguity as to whether this is just a string library name or a full package-and-library name?

toolchain/check/check.cpp
jonmeow: Fixed

toolchain/check/check.cpp

toolchain/check/check.cpp
jonmeow: "Library cannot import itself." would confuse me because the `impl` is importing the `api` implicitly. Saying "Library cannot import itself explicitly." seems like it's missing information.

Switching to:

```
      CARBON_DIAGNOSTIC(ExplicitImportApi, Error,
                        "Explicit import `api` of `impl` file is redundant "
                        "with implicit import.");
      CARBON_DIAGNOSTIC(ImportSelf, Error, "File cannot import itself.");
```

Note, I'd rewritten this to mirror the suggested message:

```
          CARBON_DIAGNOSTIC(ExplicitMainPackage, Error,
                            "Omit `package` directive for `Main` package.");
```

I've rewritten this too, to:

```
        CARBON_DIAGNOSTIC(ExplicitMainPackage, Error,
                          "Default `Main` library must omit `package` directive.");
```

toolchain/check/check.cpp
jonmeow: It'll need more work, but removed.

toolchain/check/check.cpp
jonmeow: Done, with the latter.

toolchain/check/check.cpp
jonmeow: Added a TODO to CheckParseTree.

We should probably discuss the handling. For the simplest case of `import Foo library "broken"; Foo.Bar`, I think it's easy to say that shouldn't error. I think that's less obvious when we're talking about `import Foo library "working"; import Foo library "broken"; Foo.Bar`.

Part of this is that there are a lot of constructs that could lead to later name lookup errors, even basic things like `var 1y: ...; y` or `clss C; C`. The failure to correctly add the name should bubble up to the top. To me, that's akin to the case where you're suggesting not producing errors for failed unqualified lookup -- but I'm hesitant to invalidate a name lookup scope based on an error inside of it, preferring instead to indicate what does and does not succeed when the root cause is unclear.

toolchain/check/check.cpp

toolchain/parse/tree.h
jonmeow: How about `PackagingNames` to match `PackagingDirective`

Note, I'm finding the syntax frustrating: commented on #syntax.

toolchain/parse/tree.h
jonmeow: Done

toolchain/check/check.cpp
zygoloid: Looks like this edit got undone by a force-push.

toolchain/check/check.cpp
```diff
@@ -56,16 +87,213 @@ auto CheckParseTree(SharedValueStores& value_stores,
...
+      return;
+    }
+
+    // Check for explicit imports of the same library. The ID comparison is okay
+    // in this case because both come from the same file.
+    auto packaging = unit_info.unit->parse_tree->packaging_directive();
+    if (packaging && import.package_id == packaging->names.package_id &&
+        import.library_id == packaging->names.library_id) {
+      CARBON_DIAGNOSTIC(ExplicitImportApi, Error,
+                        "Explicit import `api` of `impl` file is redundant "
```
zygoloid: I had trouble parsing this diagnostic text. How about:
```suggestion
                        "Explicit import of `api` from `impl` file is redundant "
```
or
```suggestion
                        "Explicit import of `api` from corresponding `impl` file is redundant "
```

toolchain/check/check.cpp
jonmeow: Re-done

toolchain/check/check.cpp
zygoloid: Works for me, thanks.

toolchain/parse/tree.h
zygoloid: That name works for me.

zygoloid (approved):

PR #3382, opened 2023-11-10
Use abbreviation "decl" instead of "declaration"

josh11b:


chandlerc (approved):
LG

Issue #3384, opened 2023-11-10
are `abstract` / `base` specifiers permitted or required on class forward declarations?


zygoloid:
### Summary of issue:

Given this class:
```carbon
base class C {}
```
how does one write a forward declaration? Is
```carbon
base class C;
```
valid? Is
```carbon
class C;
```
valid?

### Details:

There doesn't seem to be a reason to require the `abstract` or `base` specifier to be present on a forward declaration. But usually our forward declaration syntax for classes is a prefix of the definition, and there also doesn't seem to be any particular hardship to developers in requiring the forward declaration to repeat the specifier.

In addition to the three answers suggested by the question (specifier is disallowed, specifier is allowed but not required, specifier is required), we could change the class definition syntax to put the specifier in the class body, as we did for inheritance:

```carbon
class C {
  base class;
  extend base: B;
}
```

justzh:
change it

chandlerc:
@justzh - Please keep comments on PRs and issues constructive and on-topic. I'm hiding yours, but you've had warnings now and so I will block if this continues.

josh11b:
This question was discussed at the end of [open discussion on 2024-05-09](https://docs.google.com/document/d/1s3mMCupmuSpWOFJGnvjoElcBIe2aoaysTIdyczvKX84/edit?pli=1&resourcekey=0-G095Wc3sR6pW1hLJbGgE0g&tab=t.0#heading=h.bu7djkos4xo). I believe the conclusions were:
* modifiers on forward declarations should match those on `extern` declarations (except for the `extern` modifier)
* otherwise prefer not repeating modifiers
* prefer putting modifiers on definitions over forward declarations

zygoloid:
This question was [addressed by #3762](https://docs.carbon-lang.dev/proposals/p3762.html#modifier-keywords) after [consideration of alternatives](https://docs.carbon-lang.dev/proposals/p3762.html#other-modifier-keyword-merging-approaches):

> - Other class, impl, and interface modifiers (abstract, base, final) exist only on the definition, not on the forward declaration.

PR #3385, opened 2023-11-10
Parsing and basic checking for `abstract class` and `base class`.

zygoloid:
For now, we require the same introducer to be used each time a class is declared, but see #3384.


toolchain/check/convert.cpp
```diff
@@ -539,9 +539,20 @@ static auto ConvertStructToClass(Context& context, SemIR::StructType src_type,
                                  SemIR::InstId value_id,
                                  ConversionTarget target) -> SemIR::InstId {
   PendingBlock target_block(context);
+  auto& class_info = context.classes().Get(dest_type.class_id);
+  if (class_info.inheritance_kind == SemIR::Class::Abstract) {
+    CARBON_DIAGNOSTIC(ConstructionOfAbstractClass, Error,
+                      "Cannot construct instance of abstract class `{0}`. "
+                      "Consider using `partial` class type instead.",
```
jonmeow: class type -> facet?

https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/classes.md#partial-facet

```suggestion
                      "Consider using `partial` facet type instead.",
```

toolchain/parse/testdata/class/fail_introducer.carbon
```diff
@@ -0,0 +1,29 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+
+// CHECK:STDERR: fail_introducer.carbon:[[@LINE+3]]:1: ERROR: Unrecognized declaration introducer.
```
jonmeow, line 9: The declaration introducer here seems like it should be recognized. I'd expect something more like "`abstract` is only allowed for `class` declarations", "`abstract` may not be combined with `base`", etc. Or at least, "Incorrect use of `abstract` keyword for declaration."

toolchain/parse/handle_declaration_scope_loop.cpp
```diff
@@ -22,75 +22,106 @@ static auto HandleUnrecognizedDecl(Context& context) -> void {
...
+      context.set_first_non_packaging_token(*context.position());
+    }
+    context.set_packaging_state(Context::PackagingState::AfterNonPackagingDecl);
+  }
+
+  // Remaining keywords are only valid after imports are complete, and so all
+  // result in a `set_packaging_state` call. Note, this may not always be
+  // necessary but is probably cheaper than validating.
+  switch (position_kind) {
+    case Lex::TokenKind::Abstract: {
```
jonmeow: What would you think of pushing `abstract` and `base` handling out of this file, and into their own state? That way, the general declaration scope handling doesn't start to become specialized towards handling specific syntaxes, which would make the file scale up in size. Perhaps that would also allow for aa setup that didn't need to change the handling of `interface` or `class` in this file, allowing `Consume` to continue to be done locally within handlers.

toolchain/parse/handle_declaration_scope_loop.cpp
```diff
@@ -22,75 +22,106 @@ static auto HandleUnrecognizedDecl(Context& context) -> void {
...
-          break;
-        }
-        default: {
-          HandleUnrecognizedDecl(context);
-          break;
-        }
+      break;
+    }
+    case Lex::TokenKind::Base: {
+      if (context.PositionIs(Lex::TokenKind::Class, 1)) {
```
jonmeow: Can `abstract` and `base` share their implementation? They look very similar, only differing on the specifier node kind.

toolchain/check/handle_class.cpp
```diff
@@ -39,6 +60,25 @@ static auto BuildClassDecl(Context& context)
...
       // This is a redeclaration of an existing class.
       class_decl.class_id = existing_class_decl->class_id;
+      auto& class_info = context.classes().Get(class_decl.class_id);
+
+      // The introducer kind must match the previous declaration.
+      // TODO: The rule here is not yet decided. See #3384.
+      if (class_info.inheritance_kind != inheritance_kind) {
+        CARBON_DIAGNOSTIC(
+            ClassRedeclarationDifferentIntroducer, Error,
+            "Class redeclared with a different inheritance kind.");
```
jonmeow:
```suggestion
            "Class redeclared with different inheritance kind.");
```

Should "a" be dropped here, for message terseness?

toolchain/check/convert.cpp
zygoloid: @josh11b I *think* that's out of date; we use "facet" to mean a value of facet type these days, and it seems confusing to also use it here. (It wouldn't be wrong, because every type is a facet, because `type` is a facet type, but it seems at least misleading.)

josh11b: I agree that we've changed how we use the word "facet" since https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/classes.md#partial-facet was written. I don't have a good replacement term at the moment. "partial class type" would be fine with me until we have that discussion, though should it be "partial-class type"?

zygoloid: I'm trying to suggest specifically using the `partial` keyword here. I suppose we could use
```
                      "Cannot construct instance of abstract class. "
                      "Consider using `partial {0}` instead.",
```
Thoughts, preferences?

zygoloid: I've switched to the `partial {0}` approach for now.

toolchain/parse/handle_declaration_scope_loop.cpp
zygoloid: I tried that first, but ended up with lots of extra states that seemed to be making the parsing more complicated. Maybe we can take another look as a follow-up?

One thing I have in the back of my mind here is what this will look like once we handle more modifiers: I've been imagining that we'd (effectively) have a loop consuming them until it hits an introducer, and then producing the introducer node followed by the modifiers (in their original order). And I think that doesn't easily fall out of a stack-based approach: the stack would reverse the order of modifiers, and we'll want to push an introducer-specific state to parse the rest of whatever we found which should be run *after* we add the modifier nodes, requiring pushing to somewhere that's not the top of the stack. It seems to me like using the state stack to model that would require that we do some mutation and/or reordering of the stack once we hit the introducer. Maybe that's OK though? We already do something like that for tuples.

Another possibility would be to do lookahead from the start of a declaration to find the introducer token, and enter the right state for it and immediately build its parse node. Then, it can handle its own modifiers in whatever way makes sense, and just skip the introducer token once it's reached. (We'd effectively be parsing as if we first reordered `public abstract class X` to `class public abstract X`.)

In any case, I'd prefer to dig into this in a separate PR, once we have more modifiers that we need to support, if that works for you?

toolchain/parse/handle_declaration_scope_loop.cpp
zygoloid: Nice, done.

toolchain/parse/testdata/class/fail_introducer.carbon
zygoloid: I've added a TODO. I think how we address this will depend on how we address the other question of how we parse modifier sequences.

toolchain/check/convert.cpp
jonmeow: I could go either way on this; I think the change is fine to keep.

toolchain/parse/handle_declaration_scope_loop.cpp
jonmeow: I'm a little wary here, particularly due to the addition of peeking in PositionIs and PositionKind to support something that I think should be removed. We can proceed, but can you add a TODO to those functions to suggest avoiding/removing the depth argument if possible?

jonmeow (approved):
Approving, hopefully the [related #toolchain discussion](https://discord.com/channels/655572317891461132/655578254970716160/1172710138520551504) yields a good improvement.

PR #3387, opened 2023-11-10
fix crash caused by unicode chars

domisterwoozy:
Prevent a crash during lexing for unicode chars.


toolchain/lex/tokenized_buffer.cpp
```diff
@@ -909,7 +909,8 @@ class [[clang::internal_linkage]] TokenizedBuffer::Lexer {
     // TODO: Need to add support for Unicode lexing.
     if (LLVM_LIKELY(position + 2 >= static_cast<ssize_t>(source_text.size()) ||
                     source_text[position + 1] != '#' ||
-                    !IsIdStartByteTable[source_text[position + 2]])) {
+                    !IsIdStartByteTable[static_cast<unsigned char>(
+                        source_text[position + 2])])) {
```
zygoloid: It looks like we have the same error in `LexKeywordOrIdentifier` (L869), and in the macro-generated `Dispatch##LexMethod##SymbolToken` functions (L1075, `OneCharTokenKindTable`), though neither of those can ever actually see a negative `char` value. Would you mind changing those to cast to `unsigned char` too, for consistency?

zygoloid (commented):
Thank you!

Can you also provide a test? (Eg, add a file to `toolchain/lex/testdata` with `r#\x80` or something like that in it.)

zygoloid (approved):
Perfect, thank you!

zygoloid:
Argh, the relevant parts of `tokenized_buffer.cpp` just got moved to `lex.cpp`.

chandlerc:
Thanks for updating, merging!

PR #3390, opened 2023-11-11
Reduce the compile time of the typed_insts_test by a further 40% or so.

zygoloid:
Move the non-dependent portions of templates out into separate functions so that we don't need to repeatedly generate code for them.

### DIFF
--- a/toolchain/sem_ir/typed_insts_test.cpp
+++ b/toolchain/sem_ir/typed_insts_test.cpp
@@ -31,14 +31,17 @@ namespace {
   static_assert(Name::Kind == InstKind::Name);
 #include "toolchain/sem_ir/inst_kind.def"
 
-template <typename TypedInst>
-auto CommonFieldOrder() -> void {
-  Inst inst = InstTestHelper::MakeInst(TypedInst::Kind, Parse::Node(1),
-                                       TypeId(2), 3, 4);
-  EXPECT_EQ(inst.kind(), TypedInst::Kind);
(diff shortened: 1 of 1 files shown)


toolchain/sem_ir/typed_insts_test.cpp
```diff
@@ -31,14 +31,17 @@ namespace {
...
+  Inst inst = InstTestHelper::MakeInst(kind, Parse::Node(1), TypeId(2), 3, 4);
+  EXPECT_EQ(inst.kind(), kind);
   EXPECT_EQ(inst.parse_node(), Parse::Node(1));
   EXPECT_EQ(inst.type_id(), TypeId(2));
+  return inst;
+}
 
+template <typename TypedInst>
+auto CommonFieldOrder() -> void {
+  Inst inst = MakeInstWithNumberedFields(TypedInst::Kind);
```
jonmeow, line 44: Shouldn't ExpectEqInsts be doing essentially the same comparisons that are happening below? That is, could this also be merged?

jonmeow (approved):

toolchain/sem_ir/typed_insts_test.cpp
zygoloid: This one is doing a comparison of the typed instruction's fields, whereas `ExpectEqInsts` is comparing the fields of an untyped instruction. I don't think we can avoid generating different code for each kind of typed instruction here.

PR #3391, opened 2023-11-11
Finish "declaration" -> "decl"

josh11b:
Some new uses of "declaration" introduced in #3374 were missed in #3382 since #3374 was still in flight.


jonmeow (approved):
Thanks!

PR #3392, opened 2023-11-13
Abbreviate "parameter" -> "param"

josh11b:


toolchain/sem_ir/typed_insts.h
```diff
@@ -335,8 +335,8 @@ struct NoOp {
   // This instruction doesn't produce a value, so has no type.
 };
 
-struct Parameter {
-  static constexpr auto Kind = InstKind::Parameter.Define("parameter");
+struct Param {
+  static constexpr auto Kind = InstKind::Param.Define("parameter");
```
jonmeow:
```suggestion
  static constexpr auto Kind = InstKind::Param.Define("param");
```

jonmeow: This should probably change. It'll cause a lot of test edits, but I don't think those need to be double-checked.

jonmeow (approved):
Looks good, thanks!

toolchain/sem_ir/typed_insts.h
josh11b: It doesn't cause any test edits.

PR #3393, opened 2023-11-13
Typed wrappers around parse tree nodes.

zygoloid:
These are intended to allow the structure of a parse tree node to be described more precisely in code, to support these use cases:

- Automated checking that the parse tree conforms to the expected structure.
- Easier reading and understanding of the structure of the parse tree by toolchain developers.
- Easier navigation of the parse tree, for example for tooling uses and for use when forming diagnostics.


chandlerc:
This is still marked as draft, but had a discussion in person that I wanted to mention and relay somewhat here...

Mostly, my comment was that I generally like having more a more typed layer on top of the flat untyped layer. Having both seems really nice in general (for example falling back to the untyped layer for errors if needed).

I'd like to keep the typed layer usable as broadly as possible. If possible, it seems good to have the type information even in layers like `check`. That does mean making the typeful layer needs to stay really minimal and efficient so it can be used even there. And the expansion to a struct of concrete children needs to have an explicit API edge that makes the complexity (walking the siblings) unsurprising. It still seems likely to be useful, for example in diagnostics or even in `CARBON_CHECK` error streams, logging, or debugging.

That does somewhat lean towards a different nomenclature from the current PR -- I had suggested "traverse" as a verb that maybe is effective at telegraphing the cost and thus appropriate contexts for creating the children (IE, don't traverse within an existing traversal unless you have some good reason to be a leaf or off the critical path). I'm happy with other names though. `NodeHandle` doesn't seem quite right though, it seems like this should be a bit more integral than just a handle.

Looking at the PR, there is also the question of whether the tree is bundled into the handle or passed into the traversal operation -- I lean somewhat toward the second here, but mabye that has issues?

Anyways, I think that's a rough summary of the high level discussion that I was present for. 

There was some longer discussion around whether the subtree size is actually needed or not with this. I was only around for the first part of this discussion, but my view is mostly that we could compute it lazily or store it on the side in the rare cases where it is needed if we want to get rid of it. But I feel like it may yet be an interesting tool and so I lean somewhat towards leaving in place for now, and removing it when the ecosystem is more complete and we have a better idea of how well the alternatives would realistically meet the needs of that ecosystem.

PR #3395, opened 2023-11-14
Detect duplicate member names in struct and struct type literals

josh11b:
A struct with the same member name twice can cause a `CARBON_CHECK` failure later when it is used (problem found by fuzzing).


jonmeow:
I'm nudging this to @zygoloid, both due to the comment about his PR (that could maybe use some elaboration -- I don't understand it), and because I think the error messages might not align with the direction we've discussed on #toolchain, but I'd like him to confirm.

toolchain/check/handle_struct.cpp
```diff
@@ -52,6 +52,64 @@ auto HandleStructFieldValue(Context& context, Parse::Node parse_node) -> bool {
...
+               Parse::NodeKind::StructFieldDesignator);
+  children = context.parse_tree().children(*struct_field_designator_iter);
+  return *children.begin();
+}
+
+static auto HasDuplicateNamesError(Context& context,
+                                   SemIR::InstBlockId type_block_id) -> bool {
+  auto& sem_ir = context.sem_ir();
+  auto fields = sem_ir.inst_blocks().Get(type_block_id);
+  llvm::DenseMap<SemIR::NameId, Parse::Node> names;
```
zygoloid: Maybe use a `SmallDenseMap` here to avoid allocating memory for small structs.

toolchain/check/handle_struct.cpp
```diff
@@ -52,6 +52,64 @@ auto HandleStructFieldValue(Context& context, Parse::Node parse_node) -> bool {
...
+      llvm::StringRef container;
+      Parse::Node prev_node = it->second;
+      if (context.parse_tree().node_kind(field_inst.parse_node) ==
+          Parse::NodeKind::StructFieldValue) {
+        container = "struct literal";
+        // This avoids using the wrong parse node in the struct literal case.
+        prev_node = StructFieldValueToName(context, prev_node);
+        field_inst.parse_node =
+            StructFieldValueToName(context, field_inst.parse_node);
+      } else {
```
zygoloid: I think it would be better if the parse node associated with a `StructTypeField` were consistent. Right now, within a `StructTypeLiteral`, it points at the name, whereas in a `StructLiteral` it points at the `=`. So perhaps instead of walking from the `StructFieldValue` to the `Name` here, we could collect and store the name node directly in `HandleStructFieldValue`:

```diff
-  SemIR::NameId name_id = context.node_stack().Pop<Parse::NodeKind::Name>();
+  auto [name_node, name_id] =
+      context.node_stack().PopWithParseNode<Parse::NodeKind::Name>();
   // Store the name for the type.
   context.args_type_info_stack().AddInst(SemIR::StructTypeField{
-      parse_node, name_id, context.insts().Get(value_inst_id).type_id()});
+      name_node, name_id, context.insts().Get(value_inst_id).type_id()});
```

toolchain/check/handle_struct.cpp
```diff
@@ -61,6 +119,10 @@ auto HandleStructLiteral(Context& context, Parse::Node parse_node) -> bool {
       .PopAndDiscardSoloParseNode<
           Parse::NodeKind::StructLiteralOrStructTypeLiteralStart>();
   auto type_block_id = context.args_type_info_stack().Pop();
+  if (HasDuplicateNamesError(context, type_block_id)) {
```
zygoloid: To me, this name isn't clear that it diagnoses the error if it finds it. Something like `CheckForDuplicateNames` or `DiagnoseDuplicateNames` would be clearer to me.

zygoloid (approved):
LG with comments addressed.

toolchain/check/handle_struct.cpp
```diff
@@ -52,6 +52,64 @@ auto HandleStructFieldValue(Context& context, Parse::Node parse_node) -> bool {
...
+  auto& insts = sem_ir.insts();
+  for (SemIR::InstId field_inst_id : fields) {
+    auto field_inst = insts.GetAs<SemIR::StructTypeField>(field_inst_id);
+    auto [it, added] =
+        names.insert({field_inst.name_id, field_inst.parse_node});
+    if (!added) {
+      // FIXME: Should this be renamed `MemberNameDuplicate`, and possibly
+      // made more widely available?
+      CARBON_DIAGNOSTIC(StructNameDuplicate, Error,
+                        "Member of {0} with a duplicated name `{1}`.",
```
zygoloid: I think the style we've been using for diagnostics recently would drop the "a" here.

```suggestion
                        "Member of {0} with duplicated name `{1}`.",
```
I think we could even shorten this to
```suggestion
                        "Duplicated field name `{1}` in {0}.",
```
("Member" -> "field" also seems like it'd make the diagnostic a bit more precise.)

toolchain/check/handle_struct.cpp
josh11b: What is going on here is that when there is a struct literal, it is deducing its type, and so the elements of that type are only loosely associated with the source code as written. It is fine to do what you are saying for now, but it is something of a hack relying on nothing else at the moment trying to use this data.

UPDATE: I see though that this is at least consistent with the other code path constructing a `StructTypeField`, so fine.

PR #3396, opened 2023-11-14
Add SemIR formatting file separators

jonmeow:
Moves the filename to suffix a `---` separator.

### DIFF
--- a/toolchain/check/testdata/array/array_in_place.carbon
+++ b/toolchain/check/testdata/array/array_in_place.carbon
@@ -10,7 +10,7 @@ fn G() {
   var v: [(i32, i32, i32); 2] = (F(), F());
 }
 
-// CHECK:STDOUT: ---
+// CHECK:STDOUT: --- array_in_place.carbon
 // CHECK:STDOUT: constants {
 // CHECK:STDOUT:   %.loc7_25.1: type = tuple_type (type, type, type)
 // CHECK:STDOUT:   %.loc7_25.2: type = tuple_type (i32, i32, i32)
@@ -19,7 +19,7 @@ fn G() {
 // CHECK:STDOUT:   %.loc10_42: type = tuple_type ((i32, i32, i32), (i32, i32, i32))
 // CHECK:STDOUT: }
 // CHECK:STDOUT:
-// CHECK:STDOUT: file "array_in_place.carbon" {
+// CHECK:STDOUT: file {
 // CHECK:STDOUT:   %F: <function> = fn_decl @F
 // CHECK:STDOUT:   %G: <function> = fn_decl @G
 // CHECK:STDOUT: }
@@ -52,4 +52,3 @@ fn G() {
 // CHECK:STDOUT:   assign %v.var, %.loc10_42.9
 // CHECK:STDOUT:   return
 // CHECK:STDOUT: }
-// CHECK:STDOUT: ...
--- a/toolchain/check/testdata/array/array_in_place.carbon
+++ b/toolchain/check/testdata/array/array_in_place.carbon
@@ -10,6 +10,7 @@ fn G() {
   var v: [(i32, i32, i32); 2] = (F(), F());
 }
 
+// CHECK:STDOUT: ---
 // CHECK:STDOUT: constants {
 // CHECK:STDOUT:   %.loc7_25.1: type = tuple_type (type, type, type)
 // CHECK:STDOUT:   %.loc7_25.2: type = tuple_type (i32, i32, i32)
@@ -51,3 +52,4 @@ fn G() {
 // CHECK:STDOUT:   assign %v.var, %.loc10_42.9
 // CHECK:STDOUT:   return
 // CHECK:STDOUT: }
+// CHECK:STDOUT: ...
--- a/toolchain/check/testdata/array/assign_return_value.carbon
+++ b/toolchain/check/testdata/array/assign_return_value.carbon
@@ -10,6 +10,7 @@ fn Run() {
   var t: [i32; 1] = F();
 }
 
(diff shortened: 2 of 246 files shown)


zygoloid (commented):
A lot of the extra output here seems unnecessary. Can we just put `---` between files, rather than adding a marker before and after each file? (I think that would even still be correct YAML.)

jonmeow:
> A lot of the extra output here seems unnecessary. Can we just put `---` between files, rather than adding a marker before and after each file? (I think that would even still be correct YAML.)

While it's true it would be valid, it wouldn't match the YAML formatting that we're using today. How strongly is that line bothering you? Is there a different separator that you would prefer?

[maybe 79 `=`, or similar, to make clear that we're not trying to be consistent?]

zygoloid:
> > A lot of the extra output here seems unnecessary. Can we just put `---` between files, rather than adding a marker before and after each file? (I think that would even still be correct YAML.)
> 
> While it's true it would be valid, it wouldn't match the YAML formatting that we're using today. How strongly is that line bothering you? Is there a different separator that you would prefer?

My main concern is to keep the SemIR readable and uncluttered. I expect the common case to be looking at the SemIR from a single file, where we don't need any additional markup at all. I don't mind having two (or more) separator lines *between* files, but I'd prefer that we don't add any extra decoration to the single-file case.

SemIR isn't YAML and doesn't pretend to be, so I don't think we should be anchored on YAML syntax, and in fact maybe using something that's "obviously YAML" like `---\n`...`...\n` may be counter-productive because it might give the wrong impression. I think a simple `---\n` between documents would be fine, but a sequence of some other length or using some other punctuation would work for me too. (Preferably not something that has special meaning in SemIR, such as `%`, `!`, or `@`, though.)

chandlerc:
> > > A lot of the extra output here seems unnecessary. Can we just put `---` between files, rather than adding a marker before and after each file? (I think that would even still be correct YAML.)
> > 
> > 
> > While it's true it would be valid, it wouldn't match the YAML formatting that we're using today. How strongly is that line bothering you? Is there a different separator that you would prefer?
> 
> My main concern is to keep the SemIR readable and uncluttered. I expect the common case to be looking at the SemIR from a single file, where we don't need any additional markup at all.

FWIW, this is I think the point that resonates the most with me -- having the single file case not change seems particularly valuable.


As a concrete suggestion, what about using the same structure as the split input files for now?

```
--- filename1.carbon

constants {
  ...
}

file "filename1.carbon" {
  ...
}

--- filename2.carbon

file "filename2.carobn" {
  ...
}
```

I'm not a huge fan of the repeated filename, but I like reusing the existing syntax... So for me it's a bit of a coin toss between this and just a `---` separator between the files.

But also maybe not worth over-stressing about getting this exactly right or perfect on the first shot. If it's not disrupting the single file cases, it seems easy to iterate if there is a different structure that works better.

jonmeow:
How about the `--- filename` syntax and removing the filename from `file {`? I'm not sure whether there would ever be multiple filenames in this scope, and this avoids the duplication.

chandlerc:
> How about the `--- filename` syntax and removing the filename from `file {`? I'm not sure whether there would ever be multiple filenames in this scope, and this avoids the duplication.

Sure.

I had started wondered about moving this more in the other direction, but as I said, it seems easy to iterate here and not the right time to ponder deeper / bigger changes.

jonmeow:
> > How about the `--- filename` syntax and removing the filename from `file {`? I'm not sure whether there would ever be multiple filenames in this scope, and this avoids the duplication.
> 
> Sure.
> 
> I had started wondered about moving this more in the other direction, but as I said, it seems easy to iterate here and not the right time to ponder deeper / bigger changes.

I think the advantage this way is that it more clearly indicates the constants are coming from the file, whereas the current syntax implies they come from elsewhere.

chandlerc:
> > How about the `--- filename` syntax and removing the filename from `file {`? I'm not sure whether there would ever be multiple filenames in this scope, and this avoids the duplication.
> 
> Sure.

Ah, I think the challenge is that I think this is a less clear output for single-file cases... =/ and so we still end up churning all the single file cases when I think it's hard to know what we actually want.

If the duplication really bothers you, I guess we can.

jonmeow:
> Ah, I think the challenge is that I think this is a less clear output for single-file cases... =/ and so we still end up churning all the single file cases when I think it's hard to know what we actually want.

Can you elaborate as to the loss of clarity? [perhaps I'm on the other end: I don't understand why `constants` is outside of `file` when they are part of the file]

chandlerc:
> > Ah, I think the challenge is that I think this is a less clear output for single-file cases... =/ and so we still end up churning all the single file cases when I think it's hard to know what we actually want.
> 
> Can you elaborate as to the loss of clarity? [perhaps I'm on the other end: I don't understand why `constants` is outside of `file` when they are part of the file]

I asked the same question when `constants` was added actually....

Richard's explanation was that the `file` section I think will include potentially *runtime* things that need to be executed (global initializers).

Part of my comment that I wonder if we will want to revisit the design more holistically -- I wonder if we will want to rename that section `globals`, and potentially nest it within a `file` group of some kind. Or whether we'll end up wanting to go a very different direction such as having the formatted SemIR be intrinsically multi-file, possibly because the each-file-owns-its-stuff won't pan out well, and we'll end up with a single merged `constants`.

If it makes sense to change the `file` section now, I'd do that in a separate change that focuses on that, and then have the change that introduces some split for multiple files?

jonmeow:
I'm closing this PR: we can discuss it further, but it's costing me significant time to keep merging conflicts. It'll be easier to start a new PR once there's agreement.

PR #3399, opened 2023-11-14
Retry infrastructure failures when building in continuous integration

josh11b:


chandlerc:
I've almost finished a version that has a bit more logic to handle things, sending it just a moment.

chandlerc:
See https://github.com/carbon-language/carbon-lang/pull/3401

PR #3401, opened 2023-11-15
Add up to 10 retries to our test action.

chandlerc:
This captures the exit code of Bazel and checks for success or permanent errors on each attempt. It also sleeps a small amount between attempts. We should be able to increase the retries and sleeps as needed to minimize flakiness here, and Bazel should even persist incremental progress efficiently. Hopefully this helps reduce the failure rate of our CI.

It also changes how we build on a `push` to use a single Bazel clause to hold this logic.

Managed to get one of the download failures when testing this, and the retry logic worked but there was a bug in the success logic.

Otherwise seems to work:
- Synthetic failure: https://github.com/carbon-language/carbon-lang/actions/runs/6872431707/job/18690894069
- Success: https://github.com/carbon-language/carbon-lang/actions/runs/6872461061

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -258,7 +258,14 @@ jobs:
           # build, so simply inject an explicit no-op test target.
           echo "//scripts:no_op_test" >> $TARGETS_FILE
 
-          for i in {1..10}; do
+          for i in {1..5}; do
+            if (( $i == 4 )); then
+              # Decrease the jobs sharply if we see repeated failures to try to
+              # work around transient network errors even if it makes things
+              # slower.
+              echo "build --jobs=4" >>user.bazelrc
+            fi
+
             bazel_exit=0
             bazelisk test -c ${{ matrix.build_mode }} \
               --target_pattern_file=$TARGETS_FILE || bazel_exit=$?
@@ -275,8 +282,10 @@ jobs:
             #         on execution timeout.
             # `4`  -- No tests found, which should be impossible here.
             # `8`  -- Explicitly interrupted build.
-            # `36` -- A suspected permanent error in the local execution
-            #         environment.
+            #
+            # Note that `36` is documented as "likely permanent", but we retry
+            # it as most of our transient failures actually produce that error
+            # code.
             if (( $bazel_exit == 1 || $bazel_exit == 2 || $bazel_exit == 3 || \
                   $bazel_exit == 4 || $bazel_exit == 8 || $bazel_exit == 8 ))
             then
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -215,7 +215,7 @@ jobs:
       # Build and run all targets on branch pushes to ensure we always have a
       # clean tree. We don't expect this to be an interactive path and so don't
       # optimize the latency of this step.
-      - name: Compute impacted pull request targets
+      - name: Compute impacted pull request targets (for push)
         if:
           steps.filter.outputs.ignore == 'false' && github.event_name == 'push'
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -261,8 +258,38 @@ jobs:
...
+
+            # Several error codes are reliably permanent, break immediately.
+            # `1`  -- The build failed.
+            # `2`  -- Command line or environment problem.
+            # `3`  -- Tests failed or timed out, we don't retry at this layer
+            #         on execution timeout.
+            # `4`  -- No tests found, which should be impossible here.
+            # `8`  -- Explicitly interrupted build.
+            # `36` -- A suspected permanent error in the local execution
+            #         environment.
```
josh11b: Comment doesn't match the code.
36 is the exit code I have observed on transient failures -- I think that is worth stating that explicitly since it is not what the documented exit codes say.

.github/workflows/tests.yaml
```diff
@@ -261,8 +258,38 @@ jobs:
           # build, so simply inject an explicit no-op test target.
           echo "//scripts:no_op_test" >> $TARGETS_FILE
 
-          bazelisk test -c ${{ matrix.build_mode }} \
-            --target_pattern_file=$TARGETS_FILE
+          for i in {1..10}; do
+            bazel_exit=0
+            bazelisk test -c ${{ matrix.build_mode }} \
+              --target_pattern_file=$TARGETS_FILE || bazel_exit=$?
```
josh11b, line 271: It seems like we should decrease the `--jobs` after a failure, to adapt to an overloaded environment.

.github/workflows/tests.yaml
```diff
@@ -215,15 +215,13 @@ jobs:
       # Build and run all targets on branch pushes to ensure we always have a
       # clean tree. We don't expect this to be an interactive path and so don't
       # optimize the latency of this step.
-      - name: Test (${{ matrix.build_mode }})
+      - name: Compute impacted pull request targets
```
josh11b:
```suggestion
      - name: Compute impacted pull request targets (for push)
```

.github/workflows/tests.yaml
```diff
@@ -261,8 +258,38 @@ jobs:
           # build, so simply inject an explicit no-op test target.
           echo "//scripts:no_op_test" >> $TARGETS_FILE
 
-          bazelisk test -c ${{ matrix.build_mode }} \
-            --target_pattern_file=$TARGETS_FILE
+          for i in {1..10}; do
```
josh11b: 10 seems like a lot? It seems like we should be made aware if this many retries are needed, and make adjustments like reducing `--jobs`. Maybe 4 would make more sense?

.github/workflows/tests.yaml
chandlerc: I'm fine with this, but we haven't been doing it consistently so I was a bit hesitant. Anyways, will add it for now, I do like it.

.github/workflows/tests.yaml
chandlerc: I'd much rather the actions finally succeed, especially with the merge queue... And we should still be able to find these because if it's a problem the action should show up as significantly slow?

Still, dropped this to 5, with 3 at the high jobs count (to try to get as much done with high throughput as we can), and then 2 more with a low jobs count as a last-ditch approach.

.github/workflows/tests.yaml
chandlerc: I was really hesitant to add even more logic here, and I'm not sure how important this will be compared to just retries... All the downloads that finish before the timeout or IO failure will be cached and so each retry should actually have significantly less work to do (and so end up with less load).

Anyways, I see a simpler way to add this now, so let's try it.

.github/workflows/tests.yaml
chandlerc: Doh, good call on the comment.

But extra doh on the fact that this is actually the exit code we keep seeing. Good suggestion to document it, comment updated and code left alone.

chandlerc (commented):
Thanks for review and suggestions!

josh11b (approved):

PR #3402, opened 2023-11-15
Add more package  and import support.

jonmeow:
This should cover:

```
library "lib" api;
import Foo library default;
import library default;
import library "lib";
```

This splits out `PackageName` and `LibraryName` to their own parse nodes so that checking can ignore them and still get a balanced parse tree (otherwise, we essentially need to implement handling of the parse nodes only to remove the identifiers/string literals -- the optional names mean we can't blindly do that as before). For reference, these nodes don't need to be handled because CheckParseTree will need to directly funnel import information along with checked IRs.


toolchain/check/check.cpp
```diff
@@ -148,21 +155,72 @@ static auto TrackImport(
...
-    // Check for explicit imports of the same library. The ID comparison is okay
-    // in this case because both come from the same file.
-    auto packaging = unit_info.unit->parse_tree->packaging_directive();
-    if (packaging && import.package_id == packaging->names.package_id &&
-        import.library_id == packaging->names.library_id) {
+    // True if the file's package is implicitly `Main` (by omission).
+    bool is_file_implicit_main =
+        !packaging || !packaging->names.package_id.is_valid();
+    // True if the import is using "current package" syntax (by omission).
+    bool is_import_current_package = !import.package_id.is_valid();
```
chandlerc: What do you think about calling this `implicit_package` to make it more clear that it's testing for the implicit syntax, not just package equality?

Something like:
```suggestion
    // True if the import is using the implicit "current package" syntax
    // (by omitting an explicit package).
    bool is_import_implicit_package = !import.package_id.is_valid();
```

(not blocking either way)

toolchain/check/check.cpp
```diff
@@ -148,21 +155,72 @@ static auto TrackImport(
...
+    }
+
+    if (!is_import_current_package) {
+      // Diagnose explicit imports of the same package that use the package
+      // name.
+      if (is_same_package || (is_file_implicit_main && is_explicit_main)) {
+        CARBON_DIAGNOSTIC(
+            ImportCurrentPackageByName, Error,
+            "Imports from the current package must omit the package name.");
+        unit_info.emitter.Emit(import.node, ImportCurrentPackageByName);
```
chandlerc, line 209: Seems like it would be a bit better to emit this on the actual package name parse node?

toolchain/parse/handle_import_and_package.cpp
```diff
@@ -5,113 +5,168 @@
...
-        context.emitter().Emit(*context.position(), ExpectedApiOrImpl);
-        ExitOnParseError(context, state, directive);
+      names.library_id = *library_id;
+      context.AddNode(NodeKind::LibrarySpecifier, library_token,
+                      library_subtree_start,
+                      /*has_error=*/false);
+    } else {
+      auto next_kind = context.PositionKind();
+      if (next_kind == Lex::TokenKind::StringLiteral ||
+          (accept_default && next_kind == Lex::TokenKind::Default)) {
```
chandlerc: I think this can just be:

```suggestion
    } else if (next_kind == Lex::TokenKind::StringLiteral ||
               (accept_default && next_kind == Lex::TokenKind::Default)) {
```

But I'm finding the conditions a bit hard to track here... Not sure I see anything substantially better you can do though.

chandlerc (approved):
The parser change was a bit hard for me, but I don't see any good improvements and it doesn't seem worth blocking on. Can always simplify the code if we come up with a better structure.

Left a few inline suggestions, but nothing blocking, merge when you're ready. =D

toolchain/check/check.cpp
jonmeow: At present, I'm just trying to issue diagnostics. There's a separate discussion about what level of detail to put into diagnostics, and how (what you're suggesting would currently require more manual traversal of the tree).

toolchain/parse/handle_import_and_package.cpp
jonmeow: Oops, fixed the redundant next_kind.

Yeah though -- the best I was coming up with was trying to simplify the syntax. ;)

toolchain/check/check.cpp
jonmeow: I've adjusted the comments, with a corresponding edit for is_package_implicit_main.

Renamed to `is_import_implicit_current_package`

The problem I have with `is_import_implicit_package` is that it's ambiguous in this context: `is_package_implicit_main` is right above, but `is_import_implicit_package` doesn't necessarily mean the implicit package of `Main`. I'd thought `current_package` was a balance of brevity and clarity, but since it doesn't work for  you, I've gone for the more verbose version.

PR #3403, opened 2023-11-15
Change Main//default to an api file

jonmeow:
When there is no `package` directive, default to `Main//default api` instead of
`Main//default impl`. This means:

- The extension will be `.carbon`, not `.impl.carbon`.
- There can only be one such file when compiling.

### DIFF
--- a/proposals/p3403.md
+++ b/proposals/p3403.md
@@ -1,0 +1,99 @@
+# Change Main//default to an api file
+
+<!--
+Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+Exceptions. See /LICENSE for license information.
+SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+-->
+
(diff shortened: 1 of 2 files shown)


proposals/p3403.md
```diff
@@ -0,0 +1,99 @@
...
+
+[Code and name organization](/docs/design/code_and_name_organization/#libraries)
+says:
+
+-   API filenames must have the `.carbon` extension. They must not have a
+    `.impl.carbon` extension.
+-   Implementation filenames must have the `.impl.carbon` extension.
+
+The `Main//default` library is currently specified as an `api` file. This means
+that it should be something like `main.impl.carbon`.
```
chandlerc: I think a typo here:

```suggestion
The `Main//default` library is currently specified as an `impl` file. This means
that it should be something like `main.impl.carbon`.
```

chandlerc (commented):
Nice! (Leaving to Kate to approve.)

KateGregory (approved):

PR #3404, opened 2023-11-16
Fix a crash on invalid found by fuzzing.

chandlerc:
When a `namespace` keyword has no `;` following it, we recover by building a parse tree `Namespace` node from the `Namespace` token (as there isn't a `;` token). Allow this correspondence on errors.

Also teach the diagnostics in this case to avoid the end-of-file token as that's almost always going to be a less meaningful location. Instead, we can point at the introducer which should at least be in the code that led to the error.

### DIFF
--- a/toolchain/parse/handle_decl_name_and_params.cpp
+++ b/toolchain/parse/handle_decl_name_and_params.cpp
@@ -27,7 +27,13 @@ static auto HandleDeclNameAndParams(Context& context, State after_name)
     CARBON_DIAGNOSTIC(ExpectedDeclName, Error,
                       "`{0}` introducer should be followed by a name.",
                       Lex::TokenKind);
-    context.emitter().Emit(*context.position(), ExpectedDeclName,
+    Lex::Token location = *context.position();
+    if (context.tokens().GetKind(location) == Lex::TokenKind::EndOfFile) {
+      // The end of file is often an especially unhelpful location. If that's
+      // the best we can do here, back up the location to the introducer itself.
+      location = state.token;
+    }
+    context.emitter().Emit(location, ExpectedDeclName,
                            context.tokens().GetKind(state.token));
     context.ReturnErrorOnState();
     context.AddLeafNode(NodeKind::InvalidParse, *context.position(),
--- a/toolchain/parse/node_kind.def
+++ b/toolchain/parse/node_kind.def
@@ -128,7 +128,9 @@ CARBON_PARSE_NODE_KIND_BRACKET(ImportDirective, ImportIntroducer,
 //   _external_: Name or QualifiedDecl
 // Namespace
 CARBON_PARSE_NODE_KIND_CHILD_COUNT(NamespaceStart, 0, CARBON_TOKEN(Namespace))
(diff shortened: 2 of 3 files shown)


toolchain/parse/node_kind.def
```diff
@@ -128,7 +128,9 @@ CARBON_PARSE_NODE_KIND_BRACKET(ImportDirective, ImportIntroducer,
 //   _external_: Name or QualifiedDecl
 // Namespace
 CARBON_PARSE_NODE_KIND_CHILD_COUNT(NamespaceStart, 0, CARBON_TOKEN(Namespace))
-CARBON_PARSE_NODE_KIND_CHILD_COUNT(Namespace, 2, CARBON_TOKEN(Semi))
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(Namespace, 2,
+                                   CARBON_TOKEN(Semi)
+                                       CARBON_IF_ERROR(CARBON_TOKEN(Namespace)))
```
jonmeow, line 133: Can you please add a test of:

```
namespace "x"
```

In this case, won't the location be the string literal token, and still crash? Maybe you should use `CARBON_ANY_TOKEN` here?

toolchain/parse/handle_decl_name_and_params.cpp
```diff
@@ -27,7 +27,13 @@ static auto HandleDeclNameAndParams(Context& context, State after_name)
     CARBON_DIAGNOSTIC(ExpectedDeclName, Error,
                       "`{0}` introducer should be followed by a name.",
                       Lex::TokenKind);
-    context.emitter().Emit(*context.position(), ExpectedDeclName,
+    Lex::Token location = *context.position();
+    if (context.tokens().GetKind(location) == Lex::TokenKind::EndOfFile) {
```
jonmeow, line 31: I should note too, there are probably a lot of places that we could get into logic like this. If you want to get started on this more generally, a context helper function would probably be suitable (e.g., PositionForDiagnostic with a default).

chandlerc: Definitely agree that we'll likely want something like this eventually.

That said, I'm not super confident on what it will look like yet... I worry that the code I've used here is somewhat specific to this diagnostic and we might need more interesting logic for others. My preference would be to wait until we have a few cases to look at before trying to generalize if that's OK?

toolchain/parse/node_kind.def
chandlerc: Good idea, added a bunch of tests.

Interestingly, none of them crash. The parser ends up reusing the namespace token reliably here it seems? Happy to add more tests or make another change though if useful.

toolchain/parse/handle_decl_name_and_params.cpp
jonmeow: Sure, it's not important.

jonmeow (approved):

toolchain/parse/node_kind.def
jonmeow: Thanks, I'm not clear on what the crash was caused by now, maybe I'll look separately.

PR #3406, opened 2023-11-16
Restore interface declaration parse test

josh11b:
The declaration test was made the same as the empty body test in #2649. This change restores it to test parsing a forward declaration of an interface.


jonmeow (approved):
Thanks!

PR #107, opened 2020-07-08
Code and name organization

jonmeow:
- [RFC](https://forums.carbon-lang.dev/t/rfc-code-and-name-organization/121)
- [Decision request](https://forums.carbon-lang.dev/t/request-for-decision-code-and-name-organization/140)
- [Examples doc](https://docs.google.com/document/d/1J8GX9uw5AxBz5Q22MLHJOfzLq4WJqKL-q1VwnKGHG-k/edit)
- [Approval announcement](https://forums.carbon-lang.dev/t/accepted-code-and-name-organization-107/150)

Files, packages, libraries, and imports for Carbon.

Contributors: chandlerc, fowles, tkoeppe, zygoloid

PR #3407, opened 2023-11-16
Clarify name bindings in namespaces.

jonmeow:
-   Require namespace members be declared in the same name scope as the
    namespace is declared.
-   Allow binding patterns to directly declare names in namespaces.
-   Disallow using different namespaces in the same binding pattern.

### DIFF
--- a/docs/design/code_and_name_organization/README.md
+++ b/docs/design/code_and_name_organization/README.md
@@ -33,6 +33,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
         -   [Imports from the current package](#imports-from-the-current-package)
     -   [Namespaces](#namespaces)
         -   [Re-declaring imported namespaces](#re-declaring-imported-namespaces)
+        -   [Declaring namespace members](#declaring-namespace-members)
         -   [Aliasing](#aliasing)
 -   [Caveats](#caveats)
     -   [Package and library name conflicts](#package-and-library-name-conflicts)
@@ -671,6 +672,66 @@ namespace Shapes;
 struct Shapes.Square { ... };
 ```
 
+#### Declaring namespace members
+
+Namespace members may only be declared in the same name scope which was used to
+declare the namespace. For example:
+
+```carbon
+namespace NS;
+
+// ✅ Allowed: declaration is in file scope, which also declared `NS`.
+class NS.ClassT {
+  // ❌ Error: A class body has its own name scope.
+  var NS.a: i32 = 0;
+}
+
+fn Function() {
+  // ❌ Error: A function body has its own name scope.
+  var NS.b: i32 = 1;
+
+  namespace FunctionNS;
+  // ✅ Allowed: Both `FunctionNS` and `FunctionNS.c` are defined in the same
+  // name scope.
+  var FunctionNS.c: i32 = 2;
+
+  if (true) {
+    // ❌ Error: This is a separate name scope from `FunctionNS`.
+    var FunctionNS.d: i32 = 3;
+  }
+}
+
+// ✅ Allowed: declaration is in file scope, which also declared `NS`.
+namespace NS.MemberNS;
+
+// ✅ Allowed: declaration is in file scope, which also declared `NS.MemberNS`.
+class NS.MemberNS.MemberClassT {}
+```
+
+When there are multiple declared names in binding patterns, particularly `var`
+and `let`, all names must be in the same namespace. For example:
+
+```carbon
+namespace NS;
+
+// ✅ Allowed: `a` and `b` use the default namespace.
+var (a: i32, b: i32) = (1, 2);
+
+// ✅ Allowed: `c` and `d` are in the same namespace.
+var (NS.c: i32, NS.d: i32) = (3, 4);
+
+// ❌ Error: `e` and `f` are not in the same namespace.
+var (e: i32, NS.f: i32) = (5, 6);
+```
+
+This restriction only applies to binding patterns, not more general uses of
+patterns.
+
+```
+// ✅ Allowed: not a binding pattern.
+(NS.namespace_var, local_var) = (0, 1);
+```
+
 #### Aliasing
 
 Carbon's [alias keyword](/docs/design/aliases.md) will support aliasing
@@ -969,6 +1030,8 @@ should be part of a larger testing plan.
 -   Namespaces
     -   [File-level namespaces](/proposals/p0107.md#file-level-namespaces)
     -   [Scoped namespaces](/proposals/p0107.md#scoped-namespaces)
+    -   [Allow prefixing a tuple binding pattern with a namespace](/proposals/p3407.md#allow-prefixing-a-tuple-binding-pattern-with-a-namespace)
+    -   [Allow mixing namespaces within a binding pattern](/proposals/p3407.md#allow-mixing-namespaces-within-a-binding-pattern)
 
 ## References
 
@@ -978,3 +1041,5 @@ should be part of a larger testing plan.
     [#2550: Simplified package declaration for the main package](https://github.com/carbon-language/carbon-lang/pull/2550)
 -   Proposal
     [#3403: Change Main//default to an api file](https://github.com/carbon-language/carbon-lang/pull/3403)
+-   Proposal
+    [#3453: Clarify name bindings in namespaces.](https://github.com/carbon-language/carbon-lang/pull/3407)
--- a/docs/design/code_and_name_organization/README.md
+++ b/docs/design/code_and_name_organization/README.md
@@ -699,7 +699,7 @@ namespace NS.MemberNS;
 class NS.MemberNS.MemberClassT {}
 ```
 
-When there are multiple declared names in binding patterns, particularly `var`
+When there are multiple declared names in binding patterns in the same pattern, particularly `var`
 and `let`, all names must be in the same namespace. For example:
 
 ```carbon
--- a/docs/design/code_and_name_organization/README.md
+++ b/docs/design/code_and_name_organization/README.md
@@ -1032,6 +1032,7 @@ should be part of a larger testing plan.
     -   [Scoped namespaces](/proposals/p0107.md#scoped-namespaces)
     -   [Allow prefixing a tuple binding pattern with a namespace](/proposals/p3407.md#allow-prefixing-a-tuple-binding-pattern-with-a-namespace)
     -   [Allow mixing namespaces within a binding pattern](/proposals/p3407.md#allow-mixing-namespaces-within-a-binding-pattern)
+    -   [Disallow declaring namespaces other than the file scope](/proposals/p3407.md#disallow-declaring-namespaces-other-than-the-file-scope)
 
 ## References
 
--- a/docs/design/code_and_name_organization/README.md
+++ b/docs/design/code_and_name_organization/README.md
@@ -699,8 +699,11 @@ namespace NS.MemberNS;
 class NS.MemberNS.MemberClassT {}
 ```
 
-When there are multiple declared names in binding patterns in the same pattern, particularly `var`
-and `let`, all names must be in the same namespace. For example:
+When there are multiple declared names in binding patterns in the same pattern,
+all names must be in the same namespace. Because namespace members can only be
+declared in the same scope as the namespace, a namespace-qualified pattern
+binding can only be used in the pattern of a `var` or `let` declaration. For
+example:
 
 ```carbon
 namespace NS;
--- a/docs/design/code_and_name_organization/README.md
+++ b/docs/design/code_and_name_organization/README.md
@@ -729,7 +729,7 @@ patterns.
 
 ```
 // ✅ Allowed: not a binding pattern.
-(NS.namespace_var, local_var) = (0, 1);
+case (NS.namespace_var, local_var) => ...;
 ```
 
 #### Aliasing
--- a/docs/design/code_and_name_organization/README.md
+++ b/docs/design/code_and_name_organization/README.md
@@ -1031,8 +1031,9 @@ should be part of a larger testing plan.
     -   [File-level namespaces](/proposals/p0107.md#file-level-namespaces)
     -   [Scoped namespaces](/proposals/p0107.md#scoped-namespaces)
     -   [Allow prefixing a tuple binding pattern with a namespace](/proposals/p3407.md#allow-prefixing-a-tuple-binding-pattern-with-a-namespace)
-    -   [Allow mixing namespaces within a binding pattern](/proposals/p3407.md#allow-using-namespaces-in-unrelated-scopes-or-mixing-them-within-a-binding-pattern)
-    -   [Disallow declaring namespaces outside the file scope](/proposals/p3407.md#disallow-declaring-namespaces-other-than-the-file-scope)
+    -   [Allow binding patterns to declare names in multiple namespaces](/proposals/p3407.md#allow-binding-patterns-to-declare-names-in-multiple-namespaces)
+    -   [Allow declaring names in namespaces not owned by the current scope](/proposals/p3407.md#allow-declaring-names-in-namespaces-not-owned-by-the-current-scope)
+    -   [Disallow declaring namespaces other than the file scope](/proposals/p3407.md#disallow-declaring-namespaces-other-than-the-file-scope)
 
 ## References
 
--- a/docs/design/code_and_name_organization/README.md
+++ b/docs/design/code_and_name_organization/README.md
@@ -699,8 +699,8 @@ namespace NS.MemberNS;
 class NS.MemberNS.MemberClassT {}
 ```
 
-When there are multiple declared names in binding patterns in the same pattern,
-all names must be in the same namespace. Because namespace members can only be
+When multiple names are declared by binding patterns in the same pattern, all
+names must be in the same namespace. Because namespace members can only be
 declared in the same scope as the namespace, a namespace-qualified pattern
 binding can only be used in the pattern of a `var` or `let` declaration. For
 example:
--- a/docs/design/code_and_name_organization/README.md
+++ b/docs/design/code_and_name_organization/README.md
@@ -1031,8 +1031,8 @@ should be part of a larger testing plan.
     -   [File-level namespaces](/proposals/p0107.md#file-level-namespaces)
     -   [Scoped namespaces](/proposals/p0107.md#scoped-namespaces)
     -   [Allow prefixing a tuple binding pattern with a namespace](/proposals/p3407.md#allow-prefixing-a-tuple-binding-pattern-with-a-namespace)
-    -   [Allow mixing namespaces within a binding pattern](/proposals/p3407.md#allow-mixing-namespaces-within-a-binding-pattern)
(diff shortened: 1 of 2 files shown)


docs/design/code_and_name_organization/README.md
```diff
@@ -671,6 +672,66 @@ namespace Shapes;
...
+// ❌ Error: `e` and `f` are not in the same namespace.
+var (e: i32, NS.f: i32) = (5, 6);
+```
+
+This restriction only applies to binding patterns, not more general uses of
+patterns.
+
+```
+// ✅ Allowed: not a binding pattern.
+(NS.namespace_var, local_var) = (0, 1);
```
chandlerc: Did we end up with this being a pattern on the LHS? I'm not sure...

Maybe just use a `case` example instead? The text above makes sense and sounds right to me.

proposals/p3407.md
```diff
@@ -0,0 +1,199 @@
...
+class ClassT {
+  var static class_member: i32 = init_data[0];
+}
+
+var NS.namespace_member: i32 = init_data[1];
+var package_member: i32 = init_data[2];
+```
+
+Rather than forcing a choice, this alternative is discarded in order to require
+syntax -- and resulting performance -- be clearer.
```
chandlerc: Could we can have a much simpler/shorter rationale here without fully exploring the static class variable space? Maybe something like:

````suggestion
The mixture of two different prefixes in a single declaration alone seems likely to be confusing, and doesn't have much motivation.

If we allow namespaced names within unrelated scopes, it would also imply the possibility of mixing use inside *class* scopes:

```carbon
namespace NS;

class ClassT {
  var (class_member: i32,
       NS.namespace_member: i32,
       package.package_member: i32);
}
```

Either these are all instance members, which would be surprising in terms of the naming conventions for instance members (for example requiring looking up the namespace name after the object in `x.NS.namespace_member`), or it would require the qualified names to be *static* members (in C++ terminology).

Mixing static and non-static members in a single declaration seems fundamentally problematic due to lifetimes.

We don't have a design for static class members like this yet, and it doesn't seem like we should introduce or motivate them with namespacing within scopes.

For now, we just disallow the mixing namespaces in single declarations for simplicity and to avoid having to resolve these issues. And we disallow namespaced names within unrelated scopes to avoid the name lookup complexity and any potential confusion over C++-style-`static` or other semantics.
````

Maybe could be further simplified or reduced...

File: proposals/p3407.md [hunk 3]
```diff
@@ -0,0 +1,199 @@
...
+```carbon
+var NS.(a: i32, b: i32) = (3, 4);
+```
+
+It's rare that we would have a single statement declare multiple names. As a
+consequence, the separation of the namespace qualifier from the declared
+identifier might end up unique to this syntax. In that context, we prefer `NS.a`
+for consistency with other cases, such as `class NS.class`.
+
+### Allow mixing namespaces within a binding pattern
```
chandlerc: This also seems to cover unrelated scopes?

```suggestion
### Allow using namespaces in unrelated scopes or mixing them within a binding pattern
```

chandlerc (commented):
Generally like this direction. Some suggestions inline.

proposals/p3407.md
jonmeow: I think these are two separate alternatives... [but maybe I don't understand what you're getting at? do you want me to add an alternative "Don't allow using namespaces in unrelated scopes?", i.e. that namespaces can only be used locally?]

chandlerc: Sure, but then I'd separate out the rationale too. I think the static member thing is more attached to the unrelated scope nesting than mixing within a binding, although certainly combining the two exacerbates the rationale.

proposals/p3407.md
jonmeow: I think this has actually gone in a different direction than the original text. The example does not have unrelated scopes: `ClassT` is within `package`.

I'm going to need to spend time cleaning this up, but not right now. Now I think there's probably a misunderstanding of the point the original text was trying to make.

docs/design/code_and_name_organization/README.md
jonmeow: Done

proposals/p3407.md
jonmeow: I've split and rephrased the sections. Note I've also dropped the `static` mentions: I think my original intent with that wasn't working for you, and it didn't feel like the rephrasing needed it.

chandlerc: I somewhat understood the `static` thing, but I think your new version is really nice and simple, so let's go with it. =D Thanks!!

chandlerc (commented):
Thanks for the updates, this looks great to me, but we should get another lead to look at it.

docs/design/code_and_name_organization/README.md
```diff
@@ -671,6 +672,66 @@ namespace Shapes;
 struct Shapes.Square { ... };
 ```
 
+#### Declaring namespace members
+
+Namespace members may only be declared in the same name scope which was used to
+declare the namespace. For example:
```
zygoloid, line 679: (No action requested.)

As an aside: I think we'll want a generalized form of this rule for other name contexts, such as classes, interfaces, and so on. Something like:

Members of a name context (namespace, class, interface, etc.) that are declared with a qualified name can only be declared in the same name scope that was used to declare the name context. Declarations with an unqualified name are members of the innermost enclosing scope, if that scope corresponds to a name context. Declarations with an unqualified name in a scope that does not correspond to a name context, such as the scope of a code block, a function signature, or the parameters of a generic class, are not members of a name context.

Then this rule is a special case of that rule. But I think we should leave this to another proposal and go ahead with this rule as-is. Mostly just noting that this seems consistent with what I think the bigger picture will be.

proposals/p3407.md
```diff
@@ -0,0 +1,191 @@
...
+-   Require namespace members be declared in the same name scope as the
+    namespace is declared.
+-   Allow binding patterns to directly declare names in namespaces.
+-   Disallow using different namespaces in the same binding pattern.
+
+## Problem
+
+While the trivial case of `var NS.a` seems to be supported as a consequence of
+proposal
+[#107: Code and name organization](https://github.com/carbon-language/carbon-lang/pull/107),
```
zygoloid: I went looking there and didn't see a mention of this case for patterns. Is there a specific part I should be looking at?

proposals/p3407.md
```diff
@@ -0,0 +1,191 @@
...
+there are a couple syntactic options when binding multiple names.
+
+Also, there's no clear decision around code such as:
+
+```carbon
+namespace NS;
+class ClassT {
+  // Is this a class member accessed through `NS`, or a file scope member inside
+  // `NS`? What is its lifetime?
+  var NS.a = 0;
```
zygoloid: Should this be something like:
```suggestion
  var NS.a: i32 = 0;
```

proposals/p3407.md
```diff
@@ -0,0 +1,191 @@
...
+for reference.
+
+## Rationale
+
+-   [Performance-critical software](/docs/project/goals.md#performance-critical-software)
+
+    -   Only allowing binding patterns to declare variables in the same
+        namespace allows for a guarantee that the memory can be sequentially
+        allocated. This allows for copy elision on initializers with function
+        calls.
```
zygoloid: I don't understand this. How does the choice of namespace affect the memory layout?

proposals/p3407.md
```diff
@@ -0,0 +1,191 @@
...
+
+// `D` must be declared within `A`, even if separately defined.
+class A.D {}
+```
+
+Both the `namespace` declaration and names declared within the `namespace` must
+be written in the same name scope. This avoids name lookup ambiguities, and
+builds consistency in name scope boundaries across declarations.
+
+### Disallow declaring namespaces other than the file scope
```
zygoloid: I believe disallowing this is the status quo ante, at least by implication. Looking for mentions of "namespace" in docs/design, I didn't find anywhere that directly and clearly states as a rule that namespaces are only allowed at file scope, but I found several places that assume that to be the case ([example](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/code_and_name_organization/README.md#namespaces:~:text=A%20namespace%20declaration%20adds%20the%20first%20identifier%20in%20the%20name%20path%20as%20a%20name%20in%20the%20file%27s%20namespace)).

If the intent of this proposal is to explicitly permit class- and function-scope namespaces, then I think that would be a design change, and I think we need more investigation into consequences and we need rationale. If not, I'd suggest we just drop this section for now (and the `FunctionNS` parts of the example in the other file) and leave it to another proposal to dig into this.

proposals/p3407.md
```diff
@@ -0,0 +1,191 @@
...
+```carbon
+class Complex {
+   namespace OptionSet1;
+   class OptionSet1.MemberClassA;
+   class OptionSet1.MemberClassB;
+
+   namespace OptionSet2;
+   class OptionSet2.MemberClassC;
+   class OptionSet2.MemberClassD;
+}
```
zygoloid: Our qualified name lookup rules definitely don't do the right thing for class-scope namespaces as they stand. (Eg, we won't do instance binding when naming a field that's a member of a namespace that's a member of a class.) I'm not sure how complicated that is to fix, but it would fit quite badly into the model in explorer for instance, where we want to know at the point where we form a member name whether it's an instance member or not, which we don't know when the member name is a namespace name.

zygoloid (commented):
Broadly, this looks good to me.

I think the restriction for where namespace members may be declared is good, and the restriction that a pattern can only introduce names in a single namespace seems fine and sufficiently motivated. I didn't understand one part of the rationale for the latter, but the argument that mixing namespaces would be confusing seems strong enough for me by itself, and this seems like something we can reconsider later if we have motivation.

This proposal doesn't have enough exploration for me to understand the consequences of allowing namespaces in function scope and especially in class scope, and I think we need more investigation before we make a rule change to (unambiguously) allow that. The rest of this proposal doesn't seem to depend on that decision, though.

docs/design/code_and_name_organization/README.md
```diff
@@ -671,6 +672,66 @@ namespace Shapes;
...
+var (e: i32, NS.f: i32) = (5, 6);
+```
+
+This restriction only applies to binding patterns, not more general uses of
+patterns.
+
+```
+// ✅ Allowed: not a binding pattern.
+case (NS.namespace_var, local_var) => ...;
+```
```
zygoloid: I didn't find this example helped my understanding, because it doesn't declare any names. Is there a case where it wouldn't be equivalent to say "all names declared by a pattern must be in the same namespace"?

proposals/p3407.md
jonmeow: Broadly speaking, #107 doesn't get into pattern syntax details. But a big reason namespaces were added was C++ compatibility. C++ allows variables in namespaces, and I'm not aware of any decisions to the contrary, so isn't this supported as a consequence?

Do you want more comments regarding #107 being ambiguous beyond just "seems to be supported"? An alternative evaluating not supporting `var NS.a`? Do you want to disallow `NS.a`?

proposals/p3407.md
jonmeow: I think there are significant design problems with the explorer, but we're archiving it. How much do you expect it to factor into language design decisions?

With toolchain, my expectation is that it wouldn't be complicated. Though certainly one choice would be to say that `namespace` is equivalent to a type declaration, not a variable declaration, and so cannot be used for instance members. But I think this change moots it.

proposals/p3407.md
jonmeow: This fell out of sync with the removal of related commentary in the doc. Removed for simplicity.

proposals/p3407.md
jonmeow: Removing -- or rather, rephrasing for the inverse.

For reference, I'd talked about this with @chandlerc and he'd supported doing it -- leads might want to discuss this. But I'm hoping this sets up the proposal for faster completion.

docs/design/code_and_name_organization/README.md
jonmeow: Removing the example, but I do think the clarification on binding patterns is important.

docs/design/code_and_name_organization/README.md
jonmeow: Agreed, it would be nice to have a general rule. The "Allow binding patterns to declare names in multiple namespaces" alternative alludes to the them but lacks a reference.

proposals/p3407.md
```diff
@@ -0,0 +1,200 @@
...
+<!-- tocstop -->
+
+## Abstract
+
+-   Require namespace members be declared in the same name scope as the
+    namespace is declared.
+    -   Declaring namespaces outside file scope is disallowed, so this means
+        only at file scope.
+-   Allow binding patterns to directly declare names in namespaces.
+-   Disallow using different namespaces in the same binding pattern.
```
zygoloid:
```suggestion
-   Disallow declaring names in multiple namespaces in the same pattern.
```
The previous phrasing read to me as if you can't use `N.a: M.T` if `N` and `M` are different namespaces, but you can use `(N.a: N.T, M.b: M.U)` because that's two different binding patterns.

proposals/p3407.md
zygoloid: The `var` syntax is `var PATTERN [...]`, and I couldn't find anywhere in #107 that says that you can use `NS.a: Type` as a pattern. The pattern matching proposal specifically says you can only use an identifier on the left-hand side of a `:`. So my conclusion was that `var NS.a: Type` was not supported by the design prior to this proposal -- to me, this part is an (important!) new syntactic feature, not something that was already supported.

Reading this again, I think this sentence wasn't supposed to be saying something about `var` and patterns, so maybe we could just use a different example:

```suggestion
While the trivial case of `class NS.C` seems to be supported as a consequence of
proposal
[#107: Code and name organization](https://github.com/carbon-language/carbon-lang/pull/107),
```

proposals/p3407.md
```diff
@@ -0,0 +1,200 @@
...
+Both the `namespace` declaration and names declared within the `namespace` must
+be written in the same name scope. This avoids name lookup ambiguities, and
+builds consistency in name scope boundaries across declarations.
+
+### Allow declaring namespaces in scopes other than the file scope
+
+We could allow declaring namespaces in scopes other than the file scope. It's
+ambiguous what path should have resulted from proposal
+[#107: Code and name organization](https://github.com/carbon-language/carbon-lang/pull/107),
+although some p
```
zygoloid: Incomplete sentence.

proposals/p3407.md
zygoloid: I think there are language semantics issues here that go beyond the implementation -- if we have `x: Complex`, what is the type of `x.OptionSet1`? If it can have instance members, then is it a class? Or is it some new kind of class-namespace-fragment thing? Or something else? Is it a first-class value that can be passed around, or like a namespace can it only be used in limited ways?

To be clear: I think these questions can be answered. But I think we shouldn't add the functionality without answering them, and I don't think we should block this proposal on figuring out the answers :) I'm very happy for you to be setting this to the side for now.

proposals/p3407.md
jonmeow: FYI, unresolving just in case chandlerc is wondering why this change occurred.

proposals/p3407.md
jonmeow: Fixed, finished the thought.

docs/design/code_and_name_organization/README.md
zygoloid: If I'm understanding correctly, what you're distinguishing between here isn't binding patterns versus other kinds of patterns, it's names being declared in patterns versus names being used in patterns.

Eg, `NS1.x: NS2.y` should be OK, despite being a binding pattern and mentioning two different namespaces. And if we had a hypothetical `name@pattern` syntax to give a name to a thing matched by another pattern (eg, [as-patterns in Haskell](https://www.haskell.org/tutorial/patterns.html), corresponding [future work](https://github.com/carbon-language/carbon-lang/blob/trunk/proposals/p2188.md#as-patterns) for Carbon), then `(NS1.x@(_: auto, _: auto), NS2.y: i32)` would not be OK because it declares two names in different namespaces, even though one of them isn't using a binding pattern.

Would it be correct to say:

> This restriction only applies to names declared in patterns, not names that are used by patterns.

... with the same example you used to have here?

jonmeow: > If I'm understanding correctly, what you're distinguishing between here isn't binding patterns versus other kinds of patterns, it's names being declared in patterns versus names being used in patterns.

Personally I would say when declaring names in binding patterns, but basically yes.

> `(NS1.x@(_: auto, _: auto), NS2.y: i32)`

Do I need to address `as` patterns? I'm not clear where the names introduced on the right would exist [I wouldn't have thought binding patterns work on the RHS of the syntax]; the lack of a design makes it hard to respond to.

docs/design/code_and_name_organization/README.md
```diff
@@ -671,6 +673,51 @@ namespace Shapes;
...
+
+// ✅ Allowed: declaration is in file scope, which also declared `NS`.
+namespace NS.MemberNS;
+
+// ✅ Allowed: declaration is in file scope, which also declared `NS.MemberNS`.
+class NS.MemberNS.MemberClassT {}
+```
+
+When there are multiple declared names in binding patterns, particularly `var`
+and `let`, all names must be in the same namespace. For example:
```
zygoloid: I think we should specify the scope over which the rule applies. For example:

```suggestion
When there are multiple declared names in binding patterns in the same pattern, particularly `var`
and `let`, all names must be in the same namespace. For example:
```
or
```suggestion
When multiple names are declared by binding patterns in the same pattern, particularly `var`
and `let`, all names must be in the same namespace. For example:
```
or perhaps more simply (given that you mention binding patterns below)
```suggestion
When multiple names are declared in the same pattern, particularly `var`
and `let`, all names must be in the same namespace. For example:
```

(As written, this could be read as disallowing `var NS.a: i32; var NS2.b: i32;` -- there are multiple declared names in binding patterns there.)

I also think it'd be useful to say the "particularly" part more directly, and less as an aside. Maybe we could say something along these lines:

> Because namespace members can only be declared in the same scope as the namespace, a namespace-qualified pattern binding can only be used in the pattern of a `var` or `let` declaration.

(I think that using "particularly" like this in a language rule could be unclear: does it mean that the rule only applies in that case, or applies in that case more than in other cases? Here, I think what's happening is that the rule applies uniformly to all patterns, but there happen to only be two cases where it *can* apply currently, due to other rules.)

docs/design/code_and_name_organization/README.md
zygoloid: What you have now looks fine to me. Saying that this is specific to binding patterns seems reasonable, given that they're the only way to declare names in a pattern currently. We can generalize this rule if and when we add as-patterns.

zygoloid (commented):
Proposal looks great to me.

Some suggestions for removing potential ambiguity from the design updates, but generally I'm very happy. Thank you!

docs/design/code_and_name_organization/README.md
jonmeow: Done, I think

zygoloid (approved):

PR #3408, opened 2023-11-17
Diagnose repeat Main//default files as a redundant api file.

jonmeow:


chandlerc (approved):
LG

jonmeow:
I'd missed an error with linking the error to the start of the file (specifically, it's incompatible with autoupdate), so I'm going with a different (probably better) fix. You might want to look at the delta again.

chandlerc (approved):
Nice, LG!

PR #3409, opened 2023-11-17
Diagnose incorrect file extensions.

jonmeow:
Note this is building on #3408

### DIFF
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -9,6 +9,7 @@
 #include "toolchain/base/value_store.h"
 #include "toolchain/check/context.h"
 #include "toolchain/diagnostics/diagnostic_emitter.h"
+#include "toolchain/lex/token_kind.h"
 #include "toolchain/parse/tree.h"
 #include "toolchain/parse/tree_node_location_translator.h"
 #include "toolchain/sem_ir/file.h"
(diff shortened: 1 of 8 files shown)


toolchain/check/check.cpp
```diff
@@ -279,29 +269,77 @@ auto CheckParseTrees(const SemIR::File& builtin_ir,
...
+    // the extension.
+    auto filename = unit_info.unit->tokens->filename();
+    static constexpr llvm::StringLiteral ApiExt = ".carbon";
+    static constexpr llvm::StringLiteral ImplExt = ".impl.carbon";
+    bool is_api_with_impl_ext = !is_impl && filename.ends_with(ImplExt);
+    auto want_ext = is_impl ? ImplExt : ApiExt;
+    if (is_api_with_impl_ext || !filename.ends_with(want_ext)) {
+      CARBON_DIAGNOSTIC(IncorrectExtension, Error,
+                        "File extension of `{0}` required for `{1}`.",
+                        llvm::StringLiteral, Lex::TokenKind);
```
chandlerc, line 310: What do you think about splitting the diagonstic approach up a bit?

It would seem a bit cleaner to me to have a diagnostic for any file that doesn't end in `.carbon`, without mentioning anything about `api` or `impl` at all.

And then a separate diagnostic, located at the actual `impl` keyword, that complains about the extension not being `.impl.carbon` specifically.

And then similar to what you have below, but located at the `api` keyword (or the package directive eventually if we remove that), and explains that this *isn't* an `impl` and so can't have the extension `.impl.carbon`.

WDYT?

Happy for this to be a follow-up though, this PR is still an improvement as now we're diagnosing things. Just thinking about how to make the errors be best understood.

chandlerc (approved):
LG!

A non-blocking thought inline about what might be an improved diagnostic structure.

toolchain/check/check.cpp
jonmeow: To be clear, I'm interpreting your suggestion as:

```
foo.x: ERROR: Incorrect file extension.
foo.x: File extension of `.carbon` required for `api`.
package Foo api;
^
```

```
bar.impl.carbon: ERROR: Incorrect file extension.
bar.impl.carbon: File extension of `.carbon` required for `api`.
package Bar api;
^
bar.impl.carbon: File extension of `.impl.carbon` only allowed for `impl`.
```

I think the difference here is an error declaring the classification of issue, versus an error trying to be informative [instead, I believe it could be skipped with minimal impact on comprehension]. It seemed like rust may have gone in that direction (e.g., `error[E0308]: mismatched types`, versus a message including the types). But I think if you look at errors elsewhere in the toolchain, that's a different direction than we've headed: I would advocate at least move uniformly in errors.

zygoloid (commented):
This change has broken one of my use cases for the carbon driver. I have been executing test cases from the command line as follows:

```console
$ ./bazel-bin/toolchain/driver/carbon compile --phase=check =(echo 'fn F() {}')
```

... which generates a file named (for example) `/tmp/zshGTmb4v`.

I think it's reasonable to require that `.impl.carbon` is not used as a suffix for API files. But do we really need the toolchain to require that all carbon source files are stored in files with a `.carbon` extension?

jonmeow:
> This change has broken one of my use cases for the carbon driver. I have been executing test cases from the command line as follows:
> 
> ```
> $ ./bazel-bin/toolchain/driver/carbon compile --phase=check =(echo 'fn F() {}')
> ```
> 
> ... which generates a file named (for example) `/tmp/zshGTmb4v`.
> 
> I think it's reasonable to require that `.impl.carbon` is not used as a suffix for API files. But do we really need the toolchain to require that all carbon source files are stored in files with a `.carbon` extension?

You brought this up in #toolchain, let's discuss there.

zygoloid:
For posterity: #3416 has added explicit support for reading source from stdin (`-`) and other non-regular files, and the filename restrictions are not applied in those cases. The `=(echo ...)` case still doesn't work, because it creates a regular file with a funny name, but `<(echo ...)` and `echo ... | carbon compile -` now work instead.

PR #3410, opened 2023-11-17
Binding pattern naming cleanup

geoffromer:
- Rename `PatternBinding` to `BindingPattern`.
- Use `BindingPattern` rather than `Pattern` in the names of binding-pattern-specific parse states.


toolchain/check/node_stack.h
```diff
@@ -302,7 +302,7 @@ class NodeStack {
       case Parse::NodeKind::MemberAccessExpr:
       case Parse::NodeKind::NameExpr:
       case Parse::NodeKind::ParenExpr:
-      case Parse::NodeKind::PatternBinding:
+      case Parse::NodeKind::BindingPattern:
```
jonmeow: Sorry, can you just run a quick sort here to keep the list in lexicographic order??

toolchain/check/handle_pattern_binding.cpp
```diff
@@ -24,12 +24,12 @@ auto HandleAddress(Context& context, Parse::Node parse_node) -> bool {
...
 
-auto HandleGenericPatternBinding(Context& context, Parse::Node parse_node)
+auto HandleGenericBindingPattern(Context& context, Parse::Node parse_node)
     -> bool {
-  return context.TODO(parse_node, "GenericPatternBinding");
+  return context.TODO(parse_node, "GenericBindingPattern");
 }
 
-auto HandlePatternBinding(Context& context, Parse::Node parse_node) -> bool {
+auto HandleBindingPattern(Context& context, Parse::Node parse_node) -> bool {
```
jonmeow: Should this file also be renamed (pattern_binding -> binding_pattern)?

I'll also note, I don't know whether a more clustered handle_pattern.cpp that includes binding patterns would be best. I note you rename this in check, and I'm not objecting, just mentioning it as something to think about (but maybe complexity of individual patterns will end up sufficiently large and distinct that we would prefer distinct files).

toolchain/parse/state.def
```diff
@@ -707,7 +707,7 @@ CARBON_PARSE_STATE_VARIANTS2(ParenConditionFinish, If, While)
 // ^
 //   1. Expr
 //   2. ParenExprParamFinishAsUnknown
-//   3. ParenExprFinishAsNormal    (SPECIAL: may be replaced)
+//   3. ParenExprFinishAsGrouping    (SPECIAL: may be replaced)
```
jonmeow: Where does the term "grouping" for this come from? I'm not familiar with this usage of it, and checking [expressions](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/expressions/README.md) I'm not seeing mentions of it. Similarly in the parse tree, we would refer to this case as "ParenExpr" (as in "ParenExprOrTupleLiteralStart"; "ParenExprAsParenExpr" probably felt redundant, but "ParensAs*" would've been an alternative [though it loses useful Expr context]). "Grouping" this would make me expect "GroupingOrTupleLiteralStart".

toolchain/check/handle_pattern_binding.cpp
geoffromer: Good catch, done.

And yeah, my hunch is that one source file for all pattern kinds will be too much code.

toolchain/parse/state.def
geoffromer: > Where does the term "grouping" for this come from? I'm not familiar with this usage of it, and checking [expressions](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/expressions/README.md) I'm not seeing mentions of it.

See e.g. [here](https://en.wikipedia.org/wiki/Symbols_of_grouping), especially "For example, in the expression 3(x+y) the parentheses are symbols of grouping, but in the expression (3, 5) the parentheses may indicate an open interval."

> Similarly in the parse tree, we would refer to this case as "ParenExpr" (as in "ParenExprOrTupleLiteralStart"; "ParenExprAsParenExpr" probably felt redundant, but "ParensAs*" would've been an alternative [though it loses useful Expr context]). "Grouping" this would make me expect "GroupingOrTupleLiteralStart".

That's a good point. I suggest using "paren expression" to refer to any expression delimited by parentheses, and "grouping expression" to refer to a paren expression that isn't a tuple literal. I've tried to apply that change consistently throughout.

jonmeow: How about separating this out, or getting a lead to review? I don't want the implementation getting out of sync [terminology-wise] with Carbon's own documentation, and I'm concerned that "grouping expression" isn't improving clarity. The wikipedia article seems more abstract.

jonmeow: And to be sure, we do have a few doc examples of "parenthesized expression": https://github.com/search?q=repo%3Acarbon-language%2Fcarbon-lang+%22parenthesized+expression%22&type=code

The similar search for "grouping expression" shows no results, so I think that's a naming change, not a cleanup. https://github.com/search?q=repo%3Acarbon-language%2Fcarbon-lang+%22grouping+expression%22&type=code

geoffromer: OK, I've gone in the other direction, consistently using `ParenExpr` only when it doesn't refer to a tuple literal, and using `ParenExprOrTupleLiteral` as the inclusive term. I'll pursue changing that terminology separately.

jonmeow: Just to be sure: while I'm fine with the PatternBinding rename, I'd like to avoid churn related to ParenExpr. Right now I'm going to wait for a resolution to #3417, but I'd be happy too if you wanted to change this to only be the PatternBinding rename. That'd mean all the misc test file changes wouldn't be combined with unrelated changes, also making it a little easier to review. [also, don't forget to update the PR description]

geoffromer: Sure, I'm happy to wait for #3417

geoffromer:
FYI I've made a couple very small new changes here. One is reverting an incomplete change from this PR, and the other is fixing a pre-existing typo.

toolchain/parse/state.def
geoffromer: I've now updated this based on the resolution of #3417.

toolchain/parse/state.def
```diff
@@ -697,7 +697,7 @@ CARBON_PARSE_STATE_VARIANTS2(ParenCondition, If, While)
 //   (state done)
 CARBON_PARSE_STATE_VARIANTS2(ParenConditionFinish, If, While)
 
-// Handles the `(` of a parenthesized expression.
+// Handles the `(` of a parenthesis expression or tuple literal
```
```suggestion
// Handles the `(` of a parenthesized single expression or tuple literal.
```

I believe this is what you intended here. Note this could also be reverted to just "parenthesized expression"; "parenthesized single expression" is just to accompany "tuple literal". You'd brought up "parenthesis expression" in discussion, but I don't think it was adopted (note that dictionary-wise, "parenthesized" means "wrapped in parentheses" -- "parenthesis" is singular).

toolchain/parse/state.def
```diff
@@ -707,12 +707,12 @@ CARBON_PARSE_STATE_VARIANTS2(ParenConditionFinish, If, While)
...
 //   1. Expr
 //   2. ParenExprParamFinishAsUnknown
-//   3. ParenExprFinishAsNormal    (SPECIAL: may be replaced)
+//   3. ParenExprFinishAsSingle    (SPECIAL: may be replaced)
 CARBON_PARSE_STATE(ParenExpr)
 
-// Handles the end of a parenthesized expression's parameter. This will start as
-// AsUnknown on the first parameter; if there are more, it switches to AsTuple
-// processing.
+// Handles the end of a parenthesis expression or tuple literal's parameter.
```
```suggestion
// Handles the end of a parenthesized single expression or tuple literal's parameter.
```

As above.

toolchain/parse/state.def
```diff
@@ -722,33 +722,33 @@ CARBON_PARSE_STATE(ParenExpr)
...
+//   2. ParenExprParamFinishAsTuple
 //   SPECIAL (variant is Unknown): parent becomes ParenExprFinishAsTuple
 //
 //  ...
 // ^
 //   (state done)
 CARBON_PARSE_STATE_VARIANTS2(ParenExprParamFinish, Unknown, Tuple)
 
-// Handles the `)` of a parenthesized expression.
+// Handles the `)` of a parenthesis expression.
```
```suggestion
// Handles the `)` of a parenthesized single expression or tuple literal.
```

As above.

jonmeow (approved):

jonmeow:
Look good!

toolchain/parse/state.def
geoffromer: You're right that this isn't what I intended, but what I intended was to drop "or tuple literal", so that "parenthesis expression" (abbreviated as `ParenExpr` in code) is the term for any expression headed by `(`. And yes, I intentionally used the noun form rather than the verb form, by analogy with e.g. "brace expression". That avoids the problem @zygoloid pointed out, that it's weird to use the term "parenthesized expression" for a tuple literal, where the thing inside parentheses is not an expression.

I've rephrased this to specify that a parenthesis is a parenthesized single expression or a tuple literal.

toolchain/parse/state.def
geoffromer: As above, what I had was wrong, but my intent was to drop "or tuple literal". I've done that now.

toolchain/parse/state.def
geoffromer: This one matches my intent: a parenthesis expression is a parenthesized single expression or a tuple literal.

geoffromer:
Revised based on Discord discussion

geoffromer:
Based on open discussion, I've reverted the changes related to `ParenExpr`. I'll send them as a separate PR.

jonmeow (approved):

PR #2761, opened 2023-04-12
[explorer] support abstract methods

maan2003:
updated #2528 

Remaining work:

- [x] rework
- [ ] more tests

Closes #2512

PR #3411, opened 2023-11-18
[explorer] Add support  for `abstract` virtual methods

### LINKED ISSUE #2512: Explorer: Support `abstract` virtual methods
### Background

The Carbon explorer currently supports `virtual` and `impl` prefixed virtual method, but not `abstract` methods (i.e. pure virtual methods in C++).

### Goal

Support `abstract` method, as described in the [class design doc](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/classes.md#virtual-override-keywords). This includes:
* Checking that the method has no implementation (`CallableDeclaration::body()`)
* Checking that child classes do implement the method

### Starting point

A recommended approach would be to modify the switch/case logic for virtual methods declaration in `explorer/interpreter/type_checker.cpp` (in the `TypeChecker::DeclareClassDeclaration` function, `VirtualOverride::Abstract` case) to remove the current error logged when `abstract` is used, and check if the method has no body first.

This can be tested by creating a running a new lit test under `explorer/testdata/class`, and run using `bazel run //explorer/testdata:class/<my_abstract_method_test.carbon>.run`.

Example of tests:

Error case 1
```
base class A {
  // Error, abstract method cannot have a body (note the curly braces)
  abstract fn Foo[self: Self]() {}
}
```

Valid case
```
abstract class A {
  // OK, no implementation
  abstract fn Foo[self: Self]();
}

class B extends A {
  // OK, implemented in child class
  fn Foo[self: Self]() {}
}
```

### Resources

* https://github.com/carbon-language/carbon-lang/blob/trunk/CONTRIBUTING.md
* https://github.com/carbon-language/carbon-lang/blob/trunk/docs/project/contribution_tools.md
* `#contributing-help` channel on the Discord

aneeshdurg:
Completing the work from #2761 by adding additional test cases. 
Closes #2512


geoffromer (approved):

geoffromer (changes requested):
It looks like the tests aren't passing. I can't tell exactly what the issue is, but running `explorer/autoupdate_testdata.sh` will probably fix it.

zygoloid (approved):

geoffromer (approved):

PR #3412, opened 2023-11-18
Support declaration modifier keywords

josh11b:


toolchain/parse/testdata/namespace/fail_modifiers.carbon
```diff
@@ -0,0 +1,23 @@
...
+// CHECK:STDERR: fail_modifiers.carbon:[[@LINE+3]]:37: ERROR: Unrecognized declaration introducer.
+// CHECK:STDERR: private abstract base default final namespace Foo;
+// CHECK:STDERR:                                     ^
+private abstract base default final namespace Foo;
+
+// CHECK:STDOUT: - filename: fail_modifiers.carbon
+// CHECK:STDOUT:   parse_tree: [
+// CHECK:STDOUT:     {kind: 'FileStart', text: ''},
+// CHECK:STDOUT:       {kind: 'InvalidParseStart', text: 'namespace', has_error: yes},
+// CHECK:STDOUT:       {kind: 'DeclModifierKeyword', text: 'private'},
```
jonmeow: Had you considered giving each modifier have its own kind? That's going to be more necessary for more complex modifiers, but even now it'd eliminate the need to re-examine the token to discern which keyword is in use.

toolchain/check/handle_modifier.h
```diff
@@ -0,0 +1,37 @@
...
+#ifndef CARBON_TOOLCHAIN_CHECK_HANDLE_MODIFIER_H_
+#define CARBON_TOOLCHAIN_CHECK_HANDLE_MODIFIER_H_
+
+#include "toolchain/check/context.h"
+
+namespace Carbon::Check {
+
+struct DeclModifierKeywords {
+  // At most one of these, and if present it must be first:
+  bool private_ = false;
```
jonmeow: Due to `bool` versus a bitmask, this is an 8-byte structure, right? https://cpp.compiler-explorer.com/z/G43qj57jd

(also, the suffix `_` looks like a style mistake)

toolchain/parse/handle_decl_scope_loop.cpp
```diff
@@ -58,24 +69,25 @@ auto HandleDeclScopeLoop(Context& context) -> void {
     context.set_packaging_state(Context::PackagingState::AfterNonPackagingDecl);
   }
 
+  // FIXME: Should we change from:
+  //   `AddLeafNode` here and `ReplacePlaceholderNode` in `HandleDeclModifier`
+  //   below
+  // to:
+  //   always adding a placeholder here, and `ReplacePlaceholderNode` in
+  //   the handlers for the different kinds of declarations
+  // ?
```
jonmeow: I think i already said this, but I lean towards the latter. You'd also want to store the subtree_start. But I think the result might be that you could write a helper function that takes in position_kind and a subtree_start, then switches on the various tokens that can actually introduce a declaration and pushes the corresponding state. Then I think a helper function could also share the switch between HandleDeclScopeLoop and HandleDeclModifier, rather than having overlapping switches.

toolchain/parse/node_kind.def
```diff
@@ -57,6 +57,15 @@ CARBON_PARSE_NODE_KIND_CHILD_COUNT(FileEnd, 0, CARBON_TOKEN(EndOfFile))
...
                                    CARBON_IF_ERROR(CARBON_ANY_TOKEN))
 
+// An invalid subtree. Always has an error.
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(InvalidParseStart, 0,
+                                   CARBON_IF_ERROR(CARBON_ANY_TOKEN))
+CARBON_PARSE_NODE_KIND_BRACKET(InvalidParseSubtree, InvalidParseStart,
+                               CARBON_IF_ERROR(CARBON_ANY_TOKEN))
+
+// A placeholder node to be replaced.
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(Placeholder, 0, CARBON_ANY_TOKEN)
```
jonmeow, line 80: Maybe a TODO, but we should probably update Tree::Verify to ensure placeholders are not present on completion (even for invalid trees).

toolchain/check/handle_modifier.cpp
```diff
@@ -0,0 +1,142 @@
...
+    CARBON_DIAGNOSTIC(ModifierNotAllowedWith, Error,
+                      "Modifier not allowed on the same declaration.");
+    CARBON_DIAGNOSTIC(ModifierNotAllowedWithPrevious, Note,
+                      "With this modifier.");
+    CARBON_DIAGNOSTIC(ModifierInWrongOrderSecond, Error,
+                      "Modifier must appear earlier.");
+    CARBON_DIAGNOSTIC(ModifierInWrongOrderFirst, Note, "Before this modifier.");
+
+    switch (context.tokens().GetKind(modifier_token)) {
+#define ACCESS(name)                                             \
```
jonmeow: Could this macro be a helper function?

toolchain/check/handle_modifier.h
```diff
@@ -0,0 +1,37 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#ifndef CARBON_TOOLCHAIN_CHECK_HANDLE_MODIFIER_H_
```
jonmeow: Structurally, I think it'd be better to separate ValidateModifiers (called from other files) from the handler. convert.* is an example of this. I'm not sure what the threshold has been for Context versus elsewhere.

toolchain/check/handle_modifier.h
```diff
@@ -0,0 +1,37 @@
...
+  bool virtual_ = false;
+};
+
+// Pops any DeclModifierKeyword parse nodes from `context` and then the
+// introducer node (using `pop_introducer`). Reports a diagnostic if they
+// contain repeated modifiers, modifiers in the incorrect order, or modifiers
+// not in `allowed`. Returns modifiers that were both found and allowed, and the
+// parse node corresponding to the first token of the declaration.
+auto ValidateModifiers(Context& context, DeclModifierKeywords allowed,
+                       std::function<Parse::Node()> pop_introducer)
```
jonmeow: Had you considered taking the introducer as NodeKind to pop, instead of a lambda? I think that'd require consider a PopForSoloParseNode version that doesn't template on the enum, which might've been the barrier here.

toolchain/check/handle_modifier.cpp
```diff
@@ -0,0 +1,142 @@
...
+                      "Modifier not allowed.");
+    CARBON_DIAGNOSTIC(ModifierDeclarationNote, Note, "On this declaration.");
+    CARBON_DIAGNOSTIC(ModifierDuplicated, Error,
+                      "Modifier repeated on the same declaration.");
+    CARBON_DIAGNOSTIC(ModifierDuplicatedPrevious, Note,
+                      "Previously appeared here.");
+    CARBON_DIAGNOSTIC(ModifierNotAllowedWith, Error,
+                      "Modifier not allowed on the same declaration.");
+    CARBON_DIAGNOSTIC(ModifierNotAllowedWithPrevious, Note,
+                      "With this modifier.");
```
jonmeow: I wonder if it'd be easier to read these as something like:

`base` repeated on declaration.
`abstract` not allowed on declaration with `base`.
`private` must appear before `base`.

With the same note:
`base` previously appeared here.

Note too, I think @zygoloid's comment on #toolchain had been to avoid "the" in diagnostics, thus a slightly different phrasing I'm suggesting.

jonmeow (commented):
I know this is draft, but trying to give a few comments that'll hopefully be helpful since I was looking at the structure for #toolchain.

toolchain/parse/testdata/namespace/fail_modifiers.carbon
josh11b: Looking at the token is annoying, but there are lots of places where I can handle all modifiers together because they all share a NodeKind

toolchain/check/handle_modifier.h
josh11b: This was only used to pass data in and out of the ValidateModifiers function, and so I expect it to be in a single register, but I can see that we will likely want to store it later (possibly with more modifiers). I've changed it to a bitmask.

Re: `_` suffix: this was to avoid collisions with C++ keywords. In the new version of this class they are constants that can be named with a capital letter instead.

toolchain/check/handle_modifier.h
josh11b: I didn't because the function took a template parameter, but if we could solve that issue then it would definitely be preferred. Is there a reason `RequireParseKind` is a template? Looks like it was changed in https://github.com/carbon-language/carbon-lang/pull/2987

toolchain/check/handle_modifier.cpp
josh11b: I've changed `DeclModifierKeywords` so that these can be made into lambdas.

toolchain/check/handle_modifier.cpp
josh11b: Updated.

toolchain/parse/node_kind.def
josh11b: Done.

toolchain/parse/handle_decl_scope_loop.cpp
josh11b: This is still in charge of `ReplacePlaceholderNode`, but I did use your suggestion to merge both functions into one. Thanks!

toolchain/parse/testdata/namespace/fail_modifiers.carbon
jonmeow: https://github.com/carbon-language/carbon-lang/pull/3419 with Literal is something zygoloid and I had discussed before. You might consider a similar approach to simplify handling of these, while still separating the node kind.

[I know there's still the operators, I need to check if he's on the same page about splitting those up too]

toolchain/check/handle_modifier.cpp
jonmeow: This function is about 130 lines right now, and I'd expect more modifiers in the future. Helper functions instead of lambda might break it up more, reducing some of the function content as more modifiers get added.

toolchain/check/handle_modifier.h
jonmeow: This might be clearer from the other parts of #2987... It was introducing `if constexpr` on the parse kind in order to allow compile-time evaluation of comparisons, versus runtime. In theory `RequireParseKind` alone could take it as a parameter, but then the cases that don't need to put a variable in the stack would need to for the call.

One option would be to hope that the variable and parameter would be optimized away, although I think that could only happen with inlining. The current state is the alternative, just don't pass it as a parameter. Then too, if we switch to CARBON_DCHECK, there's nothing to clean up.

Richard might have a better instinct about how useful an optimization this is, or if we should get rid of this altogether for `RequireParseKind` and just pass a value unconditionally. (I was sensitive here out of concern that we call it a lot)

toolchain/parse/handle_decl_scope_loop.cpp
```diff
@@ -61,64 +71,113 @@ auto HandleDeclScopeLoop(Context& context) -> void {
...
+        // We don't allow namespace or empty declarations after a modifier since
+        // they can't have modifiers and don't use bracketing parse nodes that
+        // would allow a variable number of modifier nodes.
+
+      case Lex::TokenKind::Namespace: {
+        if (saw_modifier) {
+          HandleUnrecognizedDecl(context, state.subtree_start);
+        } else {
+          introducer(NodeKind::NamespaceStart, State::Namespace);
+        }
```
jonmeow: FWIW I expect we'll want modifiers for namespaces, e.g.:
https://clang.llvm.org/docs/AttributeReference.html#abi-tag

jonmeow (commented):
FYI, I'm not looking too closely, just a few small comments. Please let me know on Discord or something if you want me to look closer again (just trying to limit how much I comment on draft PRs). But I'm also expecting that after last night's discussion you'll probably want to just work on the change a little more, maybe polishing? (I'm hoping you got a clear decision after I went home)

toolchain/parse/handle_decl_scope_loop.cpp
josh11b: We can deal with that when the time comes, no reason to make this PR bigger.

toolchain/check/handle_modifier.h
josh11b: Done.

toolchain/check/handle_named_constraint.cpp
```diff
@@ -8,11 +8,13 @@ namespace Carbon::Check {
 
 auto HandleNamedConstraintDecl(Context& context, Parse::Node parse_node)
     -> bool {
+  context.PopDeclState(DeclState::NamedConstraint);
```
jonmeow: The result of returning context.TODO should be that work ceases. What's the intent in these non-todo changes?

toolchain/check/handle_noop.cpp
```diff
@@ -15,4 +15,19 @@ auto HandleInvalidParse(Context& context, Parse::Node parse_node) -> bool {
...
+
+auto HandleInvalidParseSubtree(Context& context, Parse::Node parse_node)
+    -> bool {
+  return context.TODO(parse_node, "HandleInvalidParseSubtree");
+}
+
+auto HandlePlaceholder(Context& context, Parse::Node parse_node) -> bool {
+  CARBON_FATAL()
+      << "Placeholder node should always be replaced before parse completes";
+  return context.TODO(parse_node, "HandlePlaceholder");
```
jonmeow:
```suggestion
```

toolchain/check/decl_state.h
```diff
@@ -0,0 +1,80 @@
...
+  auto HasAbstract() const -> bool { return Has(Abstract); }
+  auto HasBase() const -> bool { return Has(Base); }
+  auto HasDefault() const -> bool { return Has(Default); }
+  auto HasFinal() const -> bool { return Has(Final); }
+  auto HasImpl() const -> bool { return Has(Impl); }
+  auto HasVirtual() const -> bool { return Has(Virtual); }
+  auto GetRaw() const -> RawEnum { return keywords; }
+
+ private:
+  RawEnum keywords;
```
jonmeow:
```suggestion
  RawEnum keywords_;
```

Style

toolchain/check/decl_state.h
```diff
@@ -0,0 +1,80 @@
...
+  auto SetPrivate() const -> KeywordModifierSet { return Set(Private); }
+  auto SetProtected() const -> KeywordModifierSet { return Set(Protected); }
+  auto SetAbstract() const -> KeywordModifierSet { return Set(Abstract); }
+  auto SetBase() const -> KeywordModifierSet { return Set(Base); }
+  auto SetDefault() const -> KeywordModifierSet { return Set(Default); }
+  auto SetFinal() const -> KeywordModifierSet { return Set(Final); }
+  auto SetImpl() const -> KeywordModifierSet { return Set(Impl); }
+  auto SetVirtual() const -> KeywordModifierSet { return Set(Virtual); }
+
+  auto Has(unsigned to_Has) const -> bool { return keywords & to_Has; }
```
jonmeow:
```suggestion
  auto Has(unsigned has_mask) const -> bool { return keywords & has_mask; }
```

"Has" -> "has" is for style.

"to_has" -> "has_mask" is a naming suggestion: I'm not sure "to" means for "has"; "has_mask" or similar might emphasize that this is a mask check.

toolchain/check/decl_state.h
```diff
@@ -0,0 +1,80 @@
...
+
+    // Sets of modifiers:
+    Access = Private | Protected,
+  };
+
+  KeywordModifierSet() : keywords(static_cast<RawEnum>(0)) {}
+  KeywordModifierSet(RawEnum k) : keywords(k) {}
+  auto Set(RawEnum to_set) const -> KeywordModifierSet {
+    KeywordModifierSet ret;
+    ret.keywords = static_cast<RawEnum>(keywords | to_set);
```
jonmeow: From "Set", I would expect direct assignment; i.e., `keywords = to_set;`. Maybe "MakeMerged"?

Also, should this be a static factory function to combine two? Why did you choose to not just set keywords? Maybe some comments would help here.

toolchain/check/decl_state.h
```diff
@@ -0,0 +1,80 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#ifndef CARBON_TOOLCHAIN_CHECK_DECL_STATE_H_
+#define CARBON_TOOLCHAIN_CHECK_DECL_STATE_H_
+
+namespace Carbon::Check {
+
+class KeywordModifierSet {
```
jonmeow: Please add some comments to explain the top-level types in this file.

toolchain/check/decl_state.h
```diff
@@ -0,0 +1,80 @@
...
+  auto HasFinal() const -> bool { return Has(Final); }
+  auto HasImpl() const -> bool { return Has(Impl); }
+  auto HasVirtual() const -> bool { return Has(Virtual); }
+  auto GetRaw() const -> RawEnum { return keywords; }
+
+ private:
+  RawEnum keywords;
+};
+
+struct DeclState {
```
jonmeow, line 46: What is this for? Can you please add comments?

toolchain/check/decl_state.h
jonmeow: Why `unsigned` here instead of `RawEnum`?

toolchain/check/modifiers_allowed.cpp
```diff
@@ -0,0 +1,47 @@
...
+        .Note(context_node, ModifierNotInContext)
+        .Emit();
+  }
+}
+
+auto ModifiersAllowedOnDecl(Context& context, KeywordModifierSet allowed,
+                            llvm::StringRef decl_name, Parse::Node context_node)
+    -> KeywordModifierSet {
+  auto& s = context.innermost_decl();
+  auto not_allowed = s.found.GetRaw() & ~allowed.GetRaw();
```
jonmeow: Versus exposing the RawEnum for this, had you considered adding `Diff` and `Intersection` helper functions?

toolchain/check/decl_state.h
```diff
@@ -0,0 +1,80 @@
...
+  // FIXME: `Fn` or `Function`?
+  enum DeclKind { FileScope, Class, NamedConstraint, Fn, Interface, Let, Var };
```
jonmeow: I'd typically expect `enum class`. I would expect it to be the default, with the enum outside the struct if you're concerned about type length. `KeywordModifierSet::RawEnum` might be a special case because it's a bitmask, where `enum class` might not work so well.

toolchain/check/decl_state.h
```diff
@@ -0,0 +1,80 @@
...
+
+  DeclState(DeclKind k, Parse::Node f) : first_node(f), kind(k) {}
+  DeclState() : DeclState(FileScope, Parse::Node::Invalid) {}
+
+  Parse::Node saw_access_mod = Parse::Node::Invalid;
+  Parse::Node saw_decl_mod = Parse::Node::Invalid;
```
jonmeow: Given these are APIs, I'd suggest `modifier` here. I'm not sure we're going to be able to abbreviate to `mod` due to ambiguity versus modulo syntax.

toolchain/check/decl_state.h
```diff
@@ -0,0 +1,80 @@
...
+  Parse::Node first_node;
+  // Invariant: members of the set are `saw_access_mod` and `saw_other_mod`
+  // if valid.
+  KeywordModifierSet found;
```
jonmeow: In callers, this would look like `decl_state.found`, right? That sounds a little like it's querying whether a declaration was found. Should this be named something like `modifier_set` instead?

toolchain/check/decl_state.h
```diff
@@ -0,0 +1,80 @@
...
+};
+
+struct DeclState {
+  // FIXME: `Fn` or `Function`?
+  enum DeclKind { FileScope, Class, NamedConstraint, Fn, Interface, Let, Var };
+
+  DeclState(DeclKind k, Parse::Node f) : first_node(f), kind(k) {}
+  DeclState() : DeclState(FileScope, Parse::Node::Invalid) {}
+
+  Parse::Node saw_access_mod = Parse::Node::Invalid;
```
jonmeow: Please add comments to explain what these are and the semantics for them.

toolchain/check/decl_state.h
```diff
@@ -0,0 +1,80 @@
...
+  Parse::Node saw_decl_mod = Parse::Node::Invalid;
+  Parse::Node first_node;
+  // Invariant: members of the set are `saw_access_mod` and `saw_other_mod`
+  // if valid.
```
jonmeow: Does this mean if the parse nodes are valid, or if the declaration is valid? What's the behavior when invalid?

toolchain/check/decl_state.h
```diff
@@ -0,0 +1,80 @@
...
+
+ private:
+  RawEnum keywords;
+};
+
+struct DeclState {
+  // FIXME: `Fn` or `Function`?
+  enum DeclKind { FileScope, Class, NamedConstraint, Fn, Interface, Let, Var };
+
+  DeclState(DeclKind k, Parse::Node f) : first_node(f), kind(k) {}
```
jonmeow: Why do you swap the order here? Can you please avoid single-character parameter names, too? (in IDEs, bear in mind the parameter names will show as hints)

toolchain/check/decl_state.h
```diff
@@ -0,0 +1,80 @@
...
+    Final = 1 << 5,
+    Impl = 1 << 6,
+    Virtual = 1 << 7,
+
+    // Sets of modifiers:
+    Access = Private | Protected,
+  };
+
+  KeywordModifierSet() : keywords(static_cast<RawEnum>(0)) {}
+  KeywordModifierSet(RawEnum k) : keywords(k) {}
```
jonmeow: Is this argument is intended to be a single keyword or a mask?

toolchain/check/modifiers_allowed.cpp
```diff
@@ -0,0 +1,47 @@
...
+                           context.TextForNode(modifier_node), decl_name);
+  } else {
+    CARBON_DIAGNOSTIC(ModifierNotInContext, Note,
+                      "Containing definition here.");
+    context.emitter()
+        .Build(modifier_node, ModifierNotAllowedOn,
+               context.TextForNode(modifier_node), decl_name)
+        .Note(context_node, ModifierNotInContext)
+        .Emit();
+  }
```
jonmeow:
```suggestion
  auto diag = context.emitter()
        .Build(modifier_node, ModifierNotAllowedOn,
               context.TextForNode(modifier_node), decl_name);
  if (context_node.is_valid()) {
    CARBON_DIAGNOSTIC(ModifierNotInContext, Note,
                      "Containing definition here.");
    diag.Note(context_node, ModifierNotInContext)
  }
  diag.Emit();
```

This kind of approach allows less duplication of the diagnostic construction.

Also note the `is_valid()` check instead of `== Parse::Node::Invalid`.

toolchain/check/modifiers_allowed.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+#include "toolchain/check/modifiers_allowed.h"
+
+namespace Carbon::Check {
+
+static auto ReportNotAllowed(Context& context, Parse::Node modifier_node,
+                             llvm::StringRef decl_name,
+                             Parse::Node context_node) {
+  CARBON_DIAGNOSTIC(ModifierNotAllowedOn, Error, "`{0}` not allowed on {1}.",
+                    llvm::StringRef, llvm::StringRef);
```
jonmeow:
```suggestion
                    std::string, std::string);
```

Please ensure diagnostic lifetimes can last the lifetime of the program, versus object lifetime. This typically means that `llvm::StringRef` won't work.

toolchain/check/context.h
```diff
@@ -252,9 +253,34 @@ class Context {
     params_or_args_stack_.AddInstId(inst_id);
   }
 
+  // Enter a declaration of kind `k`, with `f` parse node for the introducer
```
jonmeow: What is `f` here?

toolchain/check/context.h
```diff
@@ -252,9 +253,34 @@ class Context {
...
+  // token.
+  auto PushDeclState(DeclState::DeclKind k, Parse::Node parse_node) {
```
jonmeow: Please specify return types, even when `void`. It looks like you at least have `PopDeclState` doing the same, and `ReportNotAllowed`.

toolchain/check/context.h
```diff
@@ -252,9 +253,34 @@ class Context {
...
+    decl_state_stack_.push_back(DeclState(k, parse_node));
+  }
+  // Access the most recently entered declaration.
```
jonmeow: This is missing the verb phrasing from style:

> Function comments should be written with an implied subject of This function and should start with the verb phrase; for example, "Opens the file", rather than "Open the file".

https://google.github.io/styleguide/cppguide.html#Function_Comments

Note, this is also an issue below, such as "Get the state" (versus "Gets"). You might also consider the choice of words; e.g., "Leave" could also mean "allows to remain"; maybe "Pops" to echo that function name?

toolchain/check/decl_state.h
```diff
@@ -0,0 +1,80 @@
...
+ private:
+  RawEnum keywords;
+};
+
+struct DeclState {
+  // FIXME: `Fn` or `Function`?
+  enum DeclKind { FileScope, Class, NamedConstraint, Fn, Interface, Let, Var };
+
+  DeclState(DeclKind k, Parse::Node f) : first_node(f), kind(k) {}
+  DeclState() : DeclState(FileScope, Parse::Node::Invalid) {}
```
jonmeow:
```suggestion
  explicit DeclState(DeclKind k, Parse::Node f) : first_node(f), kind(k) {}
  explicit DeclState() : DeclState(FileScope, Parse::Node::Invalid) {}
```

Style: we're generally marking all constructors explicit unless otherwise needed (unfortunately, doesn't look like this made it into https://github.com/carbon-language/carbon-lang/blob/trunk/docs/project/cpp_style_guide.md, I'll try to remember to do that as a followup)

toolchain/check/decl_state.h
jonmeow:
```suggestion
  explicit KeywordModifierSet() : keywords(static_cast<RawEnum>(0)) {}
  explicit KeywordModifierSet(RawEnum k) : keywords(k) {}
```

Note this is particularly relevant for the second constructor, which is otherwise an implicit construction (which if you intend it, should be commented and explicitly marked to suppress the relevant clang-tidy warning).

toolchain/check/handle_class.cpp
```diff
@@ -13,43 +14,55 @@ auto HandleClassIntroducer(Context& context, Parse::Node parse_node) -> bool {
   context.inst_block_stack().Push();
   // Push the bracketing node.
   context.node_stack().Push(parse_node);
-  // A name should always follow.
+  // Optional modifiers and the name follow.
+  context.PushDeclState(DeclState::Class, parse_node);
```
jonmeow: This sounds a little odd to me: doesn't the introducer come after modifiers?

toolchain/check/modifiers_allowed.h
```diff
@@ -0,0 +1,24 @@
...
+#include "toolchain/check/context.h"
+
+namespace Carbon::Check {
+
+// Reports a diagnostic (using `decl_name`) if modifiers on this declaration are
+// not in `allowed`. Returns modifiers that were both found and allowed, and
+// updates the declaration state in `context`. Specify optional Parse::Node
+// `context_node` to add a note to the diagnostic about the containing
+// definition.
+auto ModifiersAllowedOnDecl(Context& context, KeywordModifierSet allowed,
```
jonmeow: Please change the name to add a verb explaining what this is doing. Does it get the modifiers or set them?

toolchain/check/decl_state.h
```diff
@@ -0,0 +1,80 @@
...
+class KeywordModifierSet {
+ public:
+  enum RawEnum {
+    // At most one of these access modifiers allowed for a given declaration,
+    // and if present it must be first:
+    Private = 1 << 0,
+    Protected = 1 << 1,
+
+    // At most one of these declaration modifiers allowed for a given
+    // declaration:
```
zygoloid: Can we have a `default base class C { ... }` in an `interface`?

toolchain/diagnostics/testdata/fail_multiline_token.carbon
```diff
@@ -6,7 +6,10 @@
 //
 // AUTOUPDATE
 
-// CHECK:STDERR: fail_multiline_token.carbon:[[@LINE+3]]:1: ERROR: Unrecognized declaration introducer.
+// CHECK:STDERR: fail_multiline_token.carbon:[[@LINE+6]]:1: ERROR: Unrecognized declaration introducer.
+// CHECK:STDERR: '''
+// CHECK:STDERR: ^~~
+// CHECK:STDERR: fail_multiline_token.carbon:[[@LINE+3]]:1: ERROR: Semantics TODO: `HandleInvalidParse`.
```
jonmeow: If you add `--phase=parse` in args, does it remove this diagnostic? I think we should be good with only the above diagnostic.

toolchain/diagnostics/diagnostic_kind.def
```diff
@@ -209,6 +209,14 @@ CARBON_DIAGNOSTIC_KIND(QualifiedExprUnsupported)
 CARBON_DIAGNOSTIC_KIND(QualifiedExprNameNotFound)
 CARBON_DIAGNOSTIC_KIND(UseOfNonExprAsValue)
 
+// Modifier checking
```
```suggestion
// Modifier checking.
```

Just for consistent punctuation; above similar comments have periods.

toolchain/parse/context.h
```diff
@@ -121,6 +121,24 @@ class Context {
...
+  //   ```
+  // It may be replaced with the intended leaf parse node with code like:
+  //   ```
+  //   auto HandleWillFillInPlaceholder(Context& context) -> void {
+  //     auto state = context.PopState();
+  //     context.ReplacePlaceholderNode(state.subtree_start, /* replacement */);
+  //   }
+  //   ```
+  auto ReplacePlaceholderNode(int32_t position, NodeKind kind, Lex::Token token,
+                              bool has_error = false) -> void;
```
jonmeow: How useful do you find the default value? It looks like it's used in one place. AddNode doesn't have the default in order to provide a reminder to consider errors.

toolchain/check/decl_state.h
zygoloid: Unusual parameter name, was this intentional? (Or was this maybe produced by a search and replace from something like `check` to `Has`?)

toolchain/parse/handle_decl_scope_loop.cpp
jonmeow: The concern I have is that `namespace` is a supported declaration introducer. Right now, the diagnostic is "Unrecognized declaration introducer." 

Maybe you could change the diagnostic to be appropriate for `namespace`, or call with no allowed modifiers? Or, just return using `context.TODO` instead of `HandleUnrecognizedDecl`

toolchain/check/handle_function.cpp
```diff
@@ -55,13 +56,110 @@ static auto BuildFunctionDecl(Context& context, bool is_definition)
...
+                                          "`fn` declaration");
+  switch (context.containing_decl().kind) {
+    case DeclState::FileScope:
+      modifiers =
+          ModifiersAllowedOnDecl(context, KeywordModifierSet().SetPrivate(),
+                                 "`fn` declaration at file scope");
+      break;
+
+    case DeclState::Class: {
+      auto access_modifiers = KeywordModifierSet().SetPrivate().SetProtected();
```
zygoloid: This logic that file-scope names can be private, and class-scope names can be private or protected, seems like it's getting duplicated across each kind of declaration. Can we handle that more centrally somehow? I wonder if some part of this can be combined with the `LookupOrAddName` call, which I guess is where we'd want to act on these modifiers.

Another reason to combine this with the `LookupOrAddName` call: I think we also want to look at the name specified for the declaration here. I think either we shouldn't allow `private fn Class.PrivateFn() {}` at file scope or we should allow `protected fn Class.ProtectedFn() {}` at file scope.

toolchain/check/handle_function.cpp
```diff
@@ -55,13 +56,110 @@ static auto BuildFunctionDecl(Context& context, bool is_definition)
...
+          context.containing_decl().first_node);
+      if (!context.containing_decl().found.HasAbstract()) {
+        modifiers = ModifiersAllowedOnDecl(
+            context, access_modifiers.SetImpl().SetVirtual(),
+            "`fn` declaration in a non-abstract `class` definition",
+            context.containing_decl().first_node);
+        if (!context.containing_decl().found.HasBase()) {
+          modifiers = ModifiersAllowedOnDecl(
+              context, access_modifiers.SetImpl(),
+              "`fn` declaration in a non-abstract non-base `class` definition",
```
zygoloid:
```suggestion
              "`fn` declaration in a final `class` definition",
```
Maybe?

toolchain/parse/handle_decl_scope_loop.cpp
```diff
@@ -7,23 +7,33 @@
...
+  if (*iter == context.tree().node_token(Node(subtree_start))) {
+    context.ReplacePlaceholderNode(subtree_start, NodeKind::InvalidParse,
+                                   cursor, /*has_error=*/true);
+  } else {
+    context.ReplacePlaceholderNode(subtree_start, NodeKind::InvalidParseStart,
+                                   cursor, /*has_error=*/true);
+    // Mark everything up to the last token consumed as invalid.
+    context.AddNode(NodeKind::InvalidParseSubtree, *iter, subtree_start,
+                    /*has_error=*/true);
+  }
```
jonmeow: Had you considered uniformly outputting a tree here? It'd reduce conditional behaviors in the output parse tree, also removing the math on `context.position()` which I wouldn't expect in general.

Also, why are you discarding the result of SkipPastLikelyEnd? I think that the prior selection of semi or cursor is consistent with the roots on declarations (valid and invalid), leading to this structure.

toolchain/check/handle_function.cpp
```diff
@@ -55,13 +56,110 @@ static auto BuildFunctionDecl(Context& context, bool is_definition)
...
+              context.containing_decl().first_node);
+        }
+      }
+      break;
+    }
+
+    case DeclState::NamedConstraint:
+      modifiers =
+          ModifiersAllowedOnDecl(context, KeywordModifierSet(),
+                                 "`fn` declaration in a `namespace` definition",
```
zygoloid: I think the diagnostics would be clearer if you handled all non-class cases together. That way if I write `impl fn F()` I get told I'm supposed to only write that in a class, not the less-helpful diagnostic that I can't write it in a namespace. That'd also avoid the need to invent diagnostics for impossible situations like an `impl fn` inside a `var` or `let`.

I think restructuring this would also make it easier for me to understand what the code's doing. Eg, with something more like

```
if (kind != DeclScope::Class) {
  modifiers = ModifiersAllowedOnDecl(context, modifiers.UnsetAbstract().UnsetVirtual().UnsetImpl(),
    "`fn` declaration outside of a class");
} else if (!containing_decl_modifiers.HasAbstract() && !containing_decl_modifiers.HasBase()) {
  modifiers = ModifiersAllowedOnDecl(context, modifiers.UnsetVirtual(),
    "`fn` declaration in final class");
} else if (!containing_decl_modifiers.HasAbstract()) {
  modifiers = ModifiersAllowedOnDecl(context, modifiers.UnsetAbstract(),
    "`fn` declaration in non-abstract class");
}
```

... I'd find it more clear which cases permit and disallow abstract / virtual / impl. (Maybe a `ModifiersDisallowedOnDecl` function would make more sense here?)

This restructuring would also allow factoring out handling for subsets of the modifiers into separate functions.

toolchain/check/testdata/let/fail_modifiers.carbon
```diff
@@ -0,0 +1,95 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+
+// CHECK:STDERR: fail_modifiers.carbon:[[@LINE+3]]:1: ERROR: Semantics TODO: `private`.
+// CHECK:STDERR: private let a: i32 = 1;
+// CHECK:STDERR: ^~~~~~~
+private let a: i32 = 1;
```
jonmeow: Should this implementation TODO be separated from the structural errors?

(also applies to other similar tests)

toolchain/check/testdata/class/fail_method_modifiers.carbon
```diff
@@ -0,0 +1,103 @@
...
+  // CHECK:STDERR:   private abstract fn Abstract[self: Self]();
+  // CHECK:STDERR:   ^~~~~~~
+  // CHECK:STDERR: fail_method_modifiers.carbon:[[@LINE+6]]:11: ERROR: `abstract` not allowed on `fn` declaration in a non-abstract `class` definition.
+  // CHECK:STDERR:   private abstract fn Abstract[self: Self]();
+  // CHECK:STDERR:           ^~~~~~~~
+  // CHECK:STDERR: fail_method_modifiers.carbon:[[@LINE-7]]:1: Containing definition here.
+  // CHECK:STDERR: class FinalClass {
+  // CHECK:STDERR: ^~~~~
+  private abstract fn Abstract[self: Self]();
+  // CHECK:STDERR: fail_method_modifiers.carbon:[[@LINE+9]]:3: ERROR: Semantics TODO: `protected`.
```
jonmeow:
```suggestion
  private abstract fn Abstract[self: Self]();

  // CHECK:STDERR: fail_method_modifiers.carbon:[[@LINE+9]]:3: ERROR: Semantics TODO: `protected`.
```

In these class tests, you seem to have put no blank lines between members. This is making it hard for me to identify where there's a line of code in the middle of blocks of output, since output also includes code.

What would you think of adding blank lines between lines of code in the various tests?

toolchain/check/handle_function.cpp
```diff
@@ -55,13 +56,110 @@ static auto BuildFunctionDecl(Context& context, bool is_definition)
...
+                                              .SetDefault()
+                                              .SetFinal()
+                                              .SetImpl()
+                                              .SetVirtual(),
+                                          "`fn` declaration");
+  switch (context.containing_decl().kind) {
+    case DeclState::FileScope:
+      modifiers =
+          ModifiersAllowedOnDecl(context, KeywordModifierSet().SetPrivate(),
+                                 "`fn` declaration at file scope");
```
jonmeow: In general, it looks like all these diagnostics start with "`fn` declaration". Structurally, that's true in general. Perhaps the keyword lex token kind could be passed, and only the "at file scope" part would need to be provided? Maybe that latter part could be structural, too, if there's duplication across declarations?

Part of this is making it easier to process diagnostics programmatically, without string parsing.

toolchain/check/handle_function.cpp
```diff
@@ -55,13 +56,110 @@ static auto BuildFunctionDecl(Context& context, bool is_definition)
       context.node_stack().PopIf<Parse::NodeKind::ImplicitParamList>().value_or(
           SemIR::InstBlockId::Empty);
   auto name_context = context.decl_name_stack().FinishName();
-  auto fn_node =
-      context.node_stack()
-          .PopForSoloParseNode<Parse::NodeKind::FunctionIntroducer>();
+  context.node_stack()
```
jonmeow, line 90: This function had been about 100 lines, but with your changes it's now about 200. Please refactor the function to reduce size: typically around 100 lines should be an upper limit, and the style guide says to aim much lower.

https://google.github.io/styleguide/cppguide.html#Write_Short_Functions

toolchain/check/handle_function.cpp
```diff
@@ -55,13 +56,110 @@ static auto BuildFunctionDecl(Context& context, bool is_definition)
...
+  if (modifiers.HasAbstract()) {
+    context.TODO(context.innermost_decl().saw_decl_mod, "abstract");
+  }
+  if (modifiers.HasDefault()) {
+    context.TODO(context.innermost_decl().saw_decl_mod, "default");
+  }
+  if (modifiers.HasFinal()) {
+    context.TODO(context.innermost_decl().saw_decl_mod, "final");
+  }
+  // Only for members of derived classes
```
jonmeow:
```suggestion
  // Only for members of derived classes.
```

Punctuation style.

toolchain/parse/node_kind.def
```diff
@@ -533,6 +555,7 @@ CARBON_PARSE_NODE_KIND_CHILD_COUNT(IfExprElse, 3,
 // StructLiteral
```
zygoloid, line 555: It's not clear to me what's being repeated here. I think two things would help:

- Indent the repeated parts.
- Put the `_repeated_` label *after* rather than before, like we do for other parent nodes.

So something like:

```c++
// Struct literals, such as `{.a = 0}`:
//   StructLiteralOrStructTypeLiteralStart
//         _external_: Name
//       StructFieldDesignator
//       _external_: expression
//     StructFieldValue
//     StructComma
//   _repeated_
// StructLiteral
```

(Maybe something more explicit like `_above nodes repeated_` would also make this clearer?)

That said, I'm planning to remove these comments as part of #3393 so maybe this isn't worth worrying about.

toolchain/check/handle_modifier.cpp
```diff
@@ -0,0 +1,116 @@
...
+        .Emit();
+  } else if (s.saw_decl_mod != Parse::Node::Invalid) {
+    context.emitter()
+        .Build(parse_node, ModifierMustAppearBefore,
+               context.TextForNode(parse_node),
+               context.TextForNode(s.saw_decl_mod))
+        .Note(s.saw_decl_mod, ModifierPrevious,
+              context.TextForNode(s.saw_decl_mod))
+        .Emit();
+  } else {
```
jonmeow: Can this share some of the diagnostic construction?

toolchain/check/handle_modifier.cpp
```diff
@@ -0,0 +1,116 @@
...
+                  llvm::StringRef, llvm::StringRef);
+CARBON_DIAGNOSTIC(ModifierRepeated, Error, "`{0}` repeated on declaration.",
+                  llvm::StringRef);
+CARBON_DIAGNOSTIC(ModifierMustAppearBefore, Error,
+                  "`{0}` must appear before `{1}`.", llvm::StringRef,
+                  llvm::StringRef);
+CARBON_DIAGNOSTIC(ModifierPrevious, Note, "`{0}` previously appeared here.",
+                  llvm::StringRef);
+
+static auto AccessModifierEnum(Lex::TokenKind token_kind) -> auto {
```
jonmeow: Why are you using an `auto` return here, versus specifying the type? I'm not clear this makes the code safer, [per style](https://google.github.io/styleguide/cppguide.html#Type_deduction).

jonmeow: What would you think of prefixing this and `DeclModifierEnum` with something like "Get" to clarify the returning behavior?

toolchain/check/handle_modifier.cpp
```diff
@@ -0,0 +1,116 @@
...
+CARBON_DIAGNOSTIC(ModifierNotAllowedWith, Error,
+                  "`{0}` not allowed on declaration with `{1}`.",
+                  llvm::StringRef, llvm::StringRef);
+CARBON_DIAGNOSTIC(ModifierRepeated, Error, "`{0}` repeated on declaration.",
+                  llvm::StringRef);
+CARBON_DIAGNOSTIC(ModifierMustAppearBefore, Error,
+                  "`{0}` must appear before `{1}`.", llvm::StringRef,
+                  llvm::StringRef);
+CARBON_DIAGNOSTIC(ModifierPrevious, Note, "`{0}` previously appeared here.",
+                  llvm::StringRef);
```
jonmeow: These should all be using strings, not StringRefs, due to lifetime concerns.

toolchain/check/context.h
```diff
@@ -252,9 +253,34 @@ class Context {
...
+    CARBON_CHECK(decl_state_stack_.back().kind == k);
+    decl_state_stack_.pop_back();
+    CARBON_CHECK(!decl_state_stack_.empty());
+  }
+
   // Prints information for a stack dump.
   auto PrintForStackDump(llvm::raw_ostream& output) const -> void;
 
+  // Get the text for the token of a node for diagnostics.
+  auto TextForNode(Parse::Node parse_node) -> llvm::StringRef {
```
```suggestion
  auto GetTextForNode(Parse::Node parse_node) -> llvm::StringRef {
```

What would you think of adding "Get" to the function name, to clarify the action?

toolchain/check/handle_class.cpp
jonmeow: When there is PushScopeAndStartName, is that always going to be preceded by PushDeclState? Should PushScopeAndStartName call PushDeclState?

zygoloid (commented):
Please consider the refactorings / restructurings suggested here as non-blocking. I'm happy for them to be dealt with in follow-ups as we start adding more modifier support.

toolchain/check/context.h
```diff
@@ -449,6 +475,10 @@ class Context {
   // The stack used for qualified declaration name construction.
   DeclNameStack decl_name_stack_;
 
+  // The stack of declarations that could have modifiers. Bottom of the stack
+  // always has a "DeclState::FileScope" entry.
+  llvm::SmallVector<DeclState> decl_state_stack_;
```
jonmeow: What would it look like to move this and related functions to a side-class, just to limit Context complexity? It looks like State is already in a separate file, so I'm essentially suggesting that the stack class be there too (similar to the way DeclNameStack went).

toolchain/parse/testdata/namespace/fail_modifiers.carbon
jonmeow: This would be good to hash out with @zygoloid. I understand the complexity concern, but this is a trade-off that I'm wary.

jonmeow (commented):
The structure is generally good, although I've gone over in a lot more detail. I think the main high-level topic I still would want to address is whether keywords should have separate nodes.

toolchain/check/decl_state.h
josh11b: I don't think we can have a `class` in an `interface` at all.

toolchain/check/decl_state.h
josh11b: Oops, yes that was a mistake, though the code has now been entirely replaced.

toolchain/check/decl_state.h
josh11b: For space reasons, I want the smaller fields last, which doesn't match the order that is more logical to me for the parameters.

I expanded the parameter names.

toolchain/check/decl_state.h
josh11b: Done.

toolchain/check/handle_function.cpp
josh11b: `final` modifier is not allowed on a `class` declaration, though it is the name of the default. This error seemed more actionable: a possible fix is to add `abstract` or `base`

toolchain/check/handle_function.cpp
josh11b: I don't think we want to repeat the access modifiers when defining a class function out-of-line. But we do need to support `private` for free functions. I agree the logic to decide which access modifiers allowed is duplicated, I've added a helper to do it centrally.

toolchain/check/decl_state.h
zygoloid: Hm, I was thinking that maybe we could define a default value for an associated type that way, but... yeah, we should require that to be written as `default let C:! type = C;` where `C` is defined somewhere else.

toolchain/check/decl_state.h
josh11b: Code has been replaced.

toolchain/check/handle_function.cpp
josh11b: Done.

toolchain/parse/node_kind.def
josh11b: My concern was we are currently using the indentation for parent-child tree structure, and I was worried about introducing confusion, but I see that by making both of these changes that it is possible to disambiguate. I still have some concern about `StructLiteralOrStructTypeLiteralStart` having a different indentation level than its siblings `StructFieldValue` and `StructComma`.  I made the change, but we can revert https://github.com/carbon-language/carbon-lang/pull/3412/commits/55527fb82835cd4814cecfdd96b605bdc110ed0e if leads to more confusion.

toolchain/check/handle_function.cpp
josh11b: The change is now adding 60 lines, and about half of that is the logic to call `context.TODO`. Do you have a suggestion of what to split out?

toolchain/check/decl_state.h
josh11b: Added a comment.

toolchain/check/handle_modifier.cpp
josh11b: Done.

toolchain/check/modifiers_allowed.h
josh11b: Renamed.

toolchain/diagnostics/testdata/fail_multiline_token.carbon
josh11b: Done.

toolchain/parse/context.h
josh11b: I copied this from `AddLeafNode`, which is closer to this function than `AddNode`.

toolchain/parse/handle_decl_scope_loop.cpp
josh11b: There are a lot of cases here, and this was the simplest logic that worked in all of them. `SkipPastLikelyEnd` advances the position, and that position is consistently what I want to use. I don't want extra cases to deal with the optional being empty.

toolchain/parse/handle_decl_scope_loop.cpp
josh11b: `context.TODO` seems to only be a member of `check`'s context, not `parse`'s. I've generated a custom error for this case.

toolchain/check/context.h
josh11b: Fixed.

toolchain/check/context.h
josh11b: Done.

toolchain/check/context.h
josh11b: Done.

toolchain/check/context.h
josh11b: Done, though it means any access to this stack takes really a lot of characters.

toolchain/check/decl_state.h
josh11b: Done.

toolchain/check/decl_state.h
josh11b: I much prefer the names being in the `DeclState` scope directly.

toolchain/check/decl_state.h
josh11b: Done.

toolchain/check/decl_state.h
josh11b: Done.

toolchain/check/decl_state.h
josh11b: I tried to rephrase. I mean, `found` contains one keyword per valid node.

toolchain/check/decl_state.h
josh11b: Done.

toolchain/check/handle_class.cpp
josh11b: Parse reorders the modifier nodes to come after the introducer.

josh11b: I don't know, but I do know that `impl` declarations are going to use this and don't have names, and `PushScopeAndStartName` is operating on a different stack.

toolchain/check/handle_modifier.cpp
josh11b: Fixed.

toolchain/check/handle_function.cpp
josh11b: Done.

toolchain/check/context.h
josh11b: I'm concerned it is verbose. I added this function specifically to make things more concise. Would `text_for_node` work?

toolchain/check/decl_state.h
josh11b: Code has been replaced

josh11b: Code has been replaced.

toolchain/check/decl_state.h
josh11b: Code has been replaced.

toolchain/check/decl_state.h
josh11b: Code has been replaced.

toolchain/check/handle_modifier.cpp
josh11b: Done.

toolchain/check/handle_modifier.cpp
josh11b: I moved common code into functions.

toolchain/check/handle_named_constraint.cpp
josh11b: I will revert, but we will want this code eventually.

toolchain/check/modifiers_allowed.cpp
josh11b: Code has been replaced.

toolchain/check/testdata/let/fail_modifiers.carbon
josh11b: Done.

josh11b (commented):
Sending my comments so I can see what I still have left to respond to.

toolchain/parse/testdata/namespace/fail_modifiers.carbon
josh11b: I split up the access from the other decl modifiers. Beyond that, I don't see any benefit to further splitting. @zygoloid did you want to weigh in?

josh11b (commented):
Okay, I think that was the last one. This should be ready for another review.

toolchain/check/handle_function.cpp
josh11b: I ended up with one function for "these are the modifiers that can ever be on this declaration" that was not contextual, and one function for "these are modifiers that should not be on this declaration in this particular context" (plus a couple of functions for specific groups of modifiers).

toolchain/parse/handle_decl_scope_loop.cpp
josh11b: I've added a comment about ignoring the return value of `SkipPastLikelyEnd`.

I could always output a subtree error, but it seems like that would be surprising to consumers if it represents a single token.

toolchain/check/decl_state.h
jonmeow: We often static_assert when there's a specific packing constraint we want to be sure of -- that may be good to add here.

e.g.: `static_assert(sizeof(DeclState) == 16, "Unexpected size of DeclState!");

https://github.com/carbon-language/carbon-lang/blob/18c3622becbd018208a0c4b53ca49528ce144e5d/toolchain/parse/tree.h#L233 as an example elsewhere

toolchain/check/handle_function.cpp
jonmeow: My expectation would be that implementing TODOs would typically increase LOC, so I don't think that solves the issue.

One possible cut would be moving all the modifier logic into a helper. Another would be to segment out function declaration/redeclaration handling (which looks like 128-167). Entry point logic might also be a cut.

toolchain/check/handle_function.cpp
jonmeow: In each place you use decl_name, it's "`xyz` declaration" where `xyz` is a token spelling. What's making you choose to pass as a string, versus passing a Lex::TokenKind to the diagnostic?

toolchain/parse/handle_decl_scope_loop.cpp
jonmeow: Can you please add a TODO to adjust the SkipPastLikelyEnd logic (possibly by adding something specific to decls) so that it returns a suitable value? I really think the `position` math should be replaced with using a return value, but maybe a different function would be more suitable.

For the subtree, we already have cases in the parse tree where two nodes can represent a single token. This is going to be especially common in error cases, because the error is often the _lack_ of a token, but we still associate a token with the parse node. As a consequence, I don't think it would be surprising here to have tokens be the same, since they're still parse errors.

toolchain/parse/handle_decl_scope_loop.cpp
```diff
@@ -61,64 +73,166 @@ auto HandleDeclScopeLoop(Context& context) -> void {
...
+          context.SkipPastLikelyEnd(cursor);
+          auto iter = context.position();
+          --iter;
+          // output an invalid parse subtree.
+          context.ReplacePlaceholderNode(state.subtree_start,
+                                         NodeKind::InvalidParseStart, cursor,
+                                         /*has_error=*/true);
+          context.AddNode(NodeKind::InvalidParseSubtree, *iter,
+                          state.subtree_start,
+                          /*has_error=*/true);
```
jonmeow: Please move this out to a helper function, or otherwise refactor the containing function to be smaller.

toolchain/parse/handle_decl_scope_loop.cpp
```diff
@@ -61,64 +73,166 @@ auto HandleDeclScopeLoop(Context& context) -> void {
...
+      case Lex::TokenKind::Let: {
+        introducer(NodeKind::LetIntroducer, State::Let);
+        return;
+      }
+
+        // We don't allow namespace or empty declarations after a modifier since
+        // they can't have modifiers and don't use bracketing parse nodes that
+        // would allow a variable number of modifier nodes.
+
+      case Lex::TokenKind::Namespace: {
```
jonmeow: The association of this comment looks a little odd.

toolchain/parse/handle_decl_scope_loop.cpp
```diff
@@ -61,64 +73,166 @@ auto HandleDeclScopeLoop(Context& context) -> void {
...
+          case Lex::TokenKind::Default:
+          case Lex::TokenKind::Final:
+          case Lex::TokenKind::Fn:
+          case Lex::TokenKind::Impl:
+          case Lex::TokenKind::Interface:
+          case Lex::TokenKind::Let:
+          case Lex::TokenKind::Private:
+          case Lex::TokenKind::Protected:
+          case Lex::TokenKind::Var:
+          case Lex::TokenKind::Virtual: {
```
jonmeow: It might be helpful for the contain function's size to make this an IsModifierOrIntroducer helper or similar.

toolchain/check/context.h
jonmeow: Maybe GetNodeText, or `node_text`?

Stepping back slightly for diagnostics, though. These are fetching the text from the source buffer from diagnostics, but I'm just thinking maybe I should be cautious about printing the raw source in diagnostics. Could they instead be passing the Lex::TokenKind and using the fixed spelling for the token?

There might also be good reasons to do that: if we get into metaprogramming, the source text might not be as expected. So far I think we've avoided having diagnostics doing that reach, instead we get things like StringifyType.

toolchain/check/context.h
jonmeow: Yeah, this is a trade-off of Context getting really big.

toolchain/check/decl_state.h
```diff
@@ -0,0 +1,109 @@
...
+
+  // Exits a declaration of kind `k`.
+  auto Pop(DeclState::DeclKind k) -> void {
+    CARBON_CHECK(s_.back().kind == k);
+    s_.pop_back();
+    CARBON_CHECK(!s_.empty());
+  }
+
+ private:
+  llvm::SmallVector<DeclState> s_;
```
jonmeow: Can this be something like "stack_" to explain the semantic? Again, I'm hesitant to adopt single-character member names, and style does advocate to use single-character names sparingly, particularly in class scopes.

toolchain/check/decl_state.h
```diff
@@ -0,0 +1,109 @@
...
+  // These fields are last because they are smaller.
+
+  // Invariant: contains just the modifiers represented by `saw_access_modifier`
+  // and `saw_other_modifier`.
+  KeywordModifierSet modifier_set = KeywordModifierSet::None;
+
+  DeclKind kind;
+};
+
+// Stack of `DeclStat` values, representing all the declarations we are
```
jonmeow:
```suggestion
// Stack of `DeclState` values, representing all the declarations we are
```

toolchain/parse/testdata/namespace/fail_modifiers.carbon
jonmeow: Note, I'm fine if you want to add a TODO to come to resolution on this.

jonmeow (commented):
Thanks! This is looking pretty close to done. I have a few more comments still, but please feel free to leave TODOs if they're difficult to address. I think it'd be worth getting this in and iterating onwards.

toolchain/check/decl_state.h
josh11b: Done.

toolchain/check/handle_function.cpp
josh11b: The issue with TODOs is they probably should not be put in the same function as diagnosing modifier issues. When they do get implemented, they are going to want to access a bunch of stuff that the helper function won't be able to see.

I've compressed the TODO code and moved the modifier diagnostic logic to a separate function.  I tried moving the redeclaration logic but it depends on too many locals.

toolchain/parse/handle_decl_scope_loop.cpp
josh11b: I've added a TODO and made it always generate a subtree error. Would it be reasonable for `SkipPastLikelyEnd` to always return an end token, even if it is not a `;`?

toolchain/parse/handle_decl_scope_loop.cpp
josh11b: Done.

toolchain/parse/handle_decl_scope_loop.cpp
josh11b: Sorry, I think VSCode changed it. I've moved it left and down.

toolchain/parse/handle_decl_scope_loop.cpp
josh11b: Done.

toolchain/check/context.h
josh11b: It is now `token_kind`.

toolchain/check/handle_function.cpp
josh11b: It now uses `Lex::TokenKind`

toolchain/parse/testdata/namespace/fail_modifiers.carbon
josh11b: I added a TODO, though the benefits of splitting have not been expressed to me.

toolchain/check/decl_state.h
josh11b: Done.

josh11b: I just changed the order to be consistent. If we need to do some performance tuning, we can do that based on benchmarks.

toolchain/check/context.h
```diff
@@ -264,9 +264,9 @@ class Context {
   // Prints information for a stack dump.
   auto PrintForStackDump(llvm::raw_ostream& output) const -> void;
 
-  // Get the text for the token of a node for diagnostics.
-  auto TextForNode(Parse::Node parse_node) -> std::string {
-    return tokens().GetTokenText(parse_tree().node_token(parse_node)).str();
+  // Get the lex::TokenKind of a node for diagnostics.
```
```suggestion
  // Get the Lex::TokenKind of a node for diagnostics.
```

toolchain/check/modifiers_allowed.h
```diff
@@ -0,0 +1,40 @@
...
+#ifndef CARBON_TOOLCHAIN_CHECK_MODIFIERS_ALLOWED_H_
+#define CARBON_TOOLCHAIN_CHECK_MODIFIERS_ALLOWED_H_
+
+#include "toolchain/check/context.h"
+
+namespace Carbon::Check {
+
+// Reports a diagnostic (using `decl_name`) if access control modifiers on this
+// are not allowed, and updates the declaration state in `context`.
+auto CheckAccessModifiersOnDecl(Context& context, Lex::TokenKind decl_kind)
```
jonmeow: You might consider renaming this file, e.g. to just "modifiers"

jonmeow (approved):
Per discussion, I'll look at the token vs parse node question once this is in. Just a couple small notes on final edits.

toolchain/check/modifiers_allowed.h
josh11b: Done.

PR #3413, opened 2023-11-20
Underline the entire token when producing diagnostic messages.

domisterwoozy:
This is an incremental improvement on our diagnostic messages that simply underlines an entire token if the token is larger than 1 char (else it points to the single char with a caret like it used to).


zygoloid (commented):
The discontinuity in output between 1-character and multi-character tokens here seems like something we could improve. (Eg, we get different formatting for an error mentioning local variable named `x` versus `x2`.) I don't have a strong preference for the behavior but I think it ought to be a bit more consistent. Maybe one of these options:

- Use `^` for a length of 0 (which you could set as the default in `DiagnosticLocation`) and `~~~` for a non-zero length (including a length of 1). In practice this means that lexing problems would use `^` most of the time, and parsing / checking problems would always use `~~~`.
- Use `^~~~` in all cases. (Always one `^` followed by max(1, length-1) `~`s.)
- Use `^^^^` in all cases. (Use max(1, length) `^`s.)

I think it's probably reasonable to use `^` for a point location and `~~~` for underlining a range, which means either of the first two options would work (the difference being that the first option treats a length of 0 as a point location and a non-zero length as a range, whereas the second option treats the location + length as both a point location (the start) and a range, with the `^` rendered "on top of" the first `~`). Eventually we'll want to be able to attach multiple underlined ranges to a diagnostic, which might push us more in one direction or another, but we can handle that later, once we extend the diagnostics machinery to support it.

toolchain/diagnostics/diagnostic_emitter.h
```diff
@@ -367,7 +369,14 @@ class StreamDiagnosticConsumer : public DiagnosticConsumer {
     if (message.location.column_number > 0) {
       *stream_ << message.location.line << "\n";
       stream_->indent(message.location.column_number - 1);
-      *stream_ << "^\n";
+      if (message.location.length <= 1) {
+        *stream_ << "^";
+      } else {
+        for (int i = 0; i < message.location.length; ++i) {
+          *stream_ << "~";
+        }
```
zygoloid: `message.location.length` could be longer than `message.location.line` in the case of a multi-line token. Would it make sense to end the `~`s early in that case?

Please can you also add a test for this, for example just:

```
'''
hello
'''
```

which currently produces
```
test.carbon:1:1: ERROR: Unrecognized declaration introducer.
'''
^
```
but I think after this change will produce something like
```
test.carbon:1:1: ERROR: Unrecognized declaration introducer.
'''
~~~~~~~~~~~~~
```

domisterwoozy: I simply limited the underline to end at the end of the line but will take note that we will have to truly handle this once we add inline messages and multiple ranges.

domisterwoozy:
I went with your second bullet point option. I also added a new testdata directory under diagnostics since I think it will be helpful once we start working more on the diagnostic machinery.

jonmeow:
> The discontinuity in output between 1-character and multi-character tokens here seems like something we could improve. (Eg, we get different formatting for an error mentioning local variable named `x` versus `x2`.) I don't have a strong preference for the behavior but I think it ought to be a bit more consistent. Maybe one of these options:
> 
> * Use `^` for a length of 0 (which you could set as the default in `DiagnosticLocation`) and `~~~` for a non-zero length (including a length of 1). In practice this means that lexing problems would use `^` most of the time, and parsing / checking problems would always use `~~~`.
> * Use `^~~~` in all cases. (Always one `^` followed by max(1, length-1) `~`s.)
> * Use `^^^^` in all cases. (Use max(1, length) `^`s.)
> 
> I think it's probably reasonable to use `^` for a point location and `~~~` for underlining a range, which means either of the first two options would work (the difference being that the first option treats a length of 0 as a point location and a non-zero length as a range, whereas the second option treats the location + length as both a point location (the start) and a range, with the `^` rendered "on top of" the first `~`). Eventually we'll want to be able to attach multiple underlined ranges to a diagnostic, which might push us more in one direction or another, but we can handle that later, once we extend the diagnostics machinery to support it.

Note, I think `^~~~~` could end up less helpful than `^^^` if we want to underline more, e.g. with the field designator diagnostic (from Clang):

```
<source>:7:13: error: field designator 'foo' does not refer to any field in type 'B'
    7 |     B c = {.foo = 0, .bar = 1};
      |            ~^~~~~~~
```

Versus if it used `^^^` to highlight the full name:

```
<source>:7:13: error: field designator 'foo' does not refer to any field in type 'B'
    7 |     B c = {.foo = 0, .bar = 1};
      |            ~^^^~~~~
```

toolchain/diagnostics/diagnostic_emitter.h
```diff
@@ -367,7 +371,17 @@ class StreamDiagnosticConsumer : public DiagnosticConsumer {
     if (message.location.column_number > 0) {
       *stream_ << message.location.line << "\n";
       stream_->indent(message.location.column_number - 1);
-      *stream_ << "^\n";
+      *stream_ << "^";
+      // We want to confirm the underlined portion is atleast 1 '~' long but it
+      // does not go past the end of the line in the case of a multiline token.
+      int underline_length = std::max(1, message.location.length - 1);
+      underline_length = std::min(
```
jonmeow: Note the diagnostics such as:

```
// CHECK:STDERR: var a: [1; 39999999999999999993];
// CHECK:STDERR:                                ^~
```

Here, only the close `]` should be highlighted. The `~` indicates the `;` is also part of the error, which seems incorrect; @zygoloid was this your intent?

toolchain/diagnostics/diagnostic_emitter.h
```diff
@@ -367,7 +371,17 @@ class StreamDiagnosticConsumer : public DiagnosticConsumer {
     if (message.location.column_number > 0) {
       *stream_ << message.location.line << "\n";
       stream_->indent(message.location.column_number - 1);
-      *stream_ << "^\n";
+      *stream_ << "^";
+      // We want to confirm the underlined portion is atleast 1 '~' long but it
```
jonmeow:
```suggestion
      // We want to confirm the underlined portion is at least 1 '~' long but it
```

toolchain/diagnostics/diagnostic_emitter.h
zygoloid: No, sorry, I meant max(0, ...) in the second option!

domisterwoozy: Ah, ok, I thought that was strange. Just updated it.

domisterwoozy:
> > The discontinuity in output between 1-character and multi-character tokens here seems like something we could improve. (Eg, we get different formatting for an error mentioning local variable named `x` versus `x2`.) I don't have a strong preference for the behavior but I think it ought to be a bit more consistent. Maybe one of these options:
> > 
> > * Use `^` for a length of 0 (which you could set as the default in `DiagnosticLocation`) and `~~~` for a non-zero length (including a length of 1). In practice this means that lexing problems would use `^` most of the time, and parsing / checking problems would always use `~~~`.
> > * Use `^~~~` in all cases. (Always one `^` followed by max(1, length-1) `~`s.)
> > * Use `^^^^` in all cases. (Use max(1, length) `^`s.)
> > 
> > I think it's probably reasonable to use `^` for a point location and `~~~` for underlining a range, which means either of the first two options would work (the difference being that the first option treats a length of 0 as a point location and a non-zero length as a range, whereas the second option treats the location + length as both a point location (the start) and a range, with the `^` rendered "on top of" the first `~`). Eventually we'll want to be able to attach multiple underlined ranges to a diagnostic, which might push us more in one direction or another, but we can handle that later, once we extend the diagnostics machinery to support it.
> 
> Note, I think `^~~~~` could end up less helpful than `^^^` if we want to underline more, e.g. with the field designator diagnostic (from Clang):
> 
> ```
> <source>:7:13: error: field designator 'foo' does not refer to any field in type 'B'
>     7 |     B c = {.foo = 0, .bar = 1};
>       |            ~^~~~~~~
> ```
> 
> Versus if it used `^^^` to highlight the full name:
> 
> ```
> <source>:7:13: error: field designator 'foo' does not refer to any field in type 'B'
>     7 |     B c = {.foo = 0, .bar = 1};
>       |            ~^^^~~~~
> ```

I am currently working on a much larger change that will allow us to explicitly specify which ranges we want to reference and how (which will hopefully make it easy to do things like the above). So I was thinking that this change is more of a stopgap/incremental improvement on what we have now.

zygoloid (approved):
Thank you! (Not merging in case @jonmeow has more comments.)

jonmeow:
Looks good, thanks!

jonmeow:
Looks like merging fails due to test changes. This may be an issue getting the PR in, just as other commits go by: @domisterwoozy do you want to take another try, or would you like someone with merge access to update for conflicts then merge?

domisterwoozy:
> Looks like merging fails due to test changes. This may be an issue getting the PR in, just as other commits go by: @domisterwoozy do you want to take another try, or would you like someone with merge access to update for conflicts then merge?

I think it should be good now.

PR #3414, opened 2023-11-21
Switch cross references to a ValueStore.

jonmeow:
As I get ready to add imports, I need an API to add values and this offers a consistent way of doing so.

### DIFF
--- a/toolchain/sem_ir/file.cpp
+++ b/toolchain/sem_ir/file.cpp
@@ -44,10 +44,11 @@ auto TypeInfo::Print(llvm::raw_ostream& out) const -> void {
 File::File(SharedValueStores& value_stores)
     : value_stores_(&value_stores),
       filename_("<builtins>"),
-      // Builtins are always the first IR, even when self-referential.
-      cross_reference_irs_({this}),
       type_blocks_(allocator_),
       inst_blocks_(allocator_) {
+  CARBON_CHECK(cross_reference_irs_.Add(this) == CrossReferenceIRId::Builtins)
+      << "Builtins must be the first IR, even if self-referential";
+
   // Default entry for InstBlockId::Empty.
   inst_blocks_.AddDefaultValue();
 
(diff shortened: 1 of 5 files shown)


toolchain/sem_ir/file.cpp
```diff
@@ -44,10 +44,11 @@ auto TypeInfo::Print(llvm::raw_ostream& out) const -> void {
 File::File(SharedValueStores& value_stores)
     : value_stores_(&value_stores),
       filename_("<builtins>"),
-      // Builtins are always the first IR, even when self-referential.
-      cross_reference_irs_({this}),
       type_blocks_(allocator_),
       inst_blocks_(allocator_) {
+  CARBON_CHECK(cross_reference_irs_.Add(this) == CrossReferenceIRId::Builtins)
+      << "Builtins must be the first IR, even if self-referential";
```
zygoloid: I would prefer that we don't put code with side-effects in a `CARBON_CHECK`: as a reader I expect that the `CHECK` expression is only computing the value that we're validating and will skip over it when looking for the behavior of the function.

toolchain/sem_ir/file.cpp
```diff
@@ -72,13 +73,12 @@ File::File(SharedValueStores& value_stores, std::string filename,
...
       filename_(std::move(filename)),
-      // Builtins are always the first IR.
-      cross_reference_irs_({builtins}),
       type_blocks_(allocator_),
       inst_blocks_(allocator_) {
   CARBON_CHECK(builtins != nullptr);
-  CARBON_CHECK(builtins->cross_reference_irs_[0] == builtins)
-      << "Not called with builtins!";
+  CARBON_CHECK(cross_reference_irs_.Add(builtins) ==
+               CrossReferenceIRId::Builtins)
```
zygoloid: Same comment here.

zygoloid (approved):
Tiny change then LG.

toolchain/sem_ir/file.cpp
jonmeow: Done

toolchain/sem_ir/file.cpp
jonmeow: Done

PR #3415, opened 2023-11-21
Pass imports to SemIR.

jonmeow:
This adds instructions so that we get printing. I may adjust the instruction format a little further to add a type, but I think the basic setup will remain.

Note this builds on #3414


toolchain/check/check.cpp
```diff
@@ -17,6 +17,20 @@
...
 
 struct UnitInfo {
+  // A given import within the file, with its destination.
+  struct Import {
+    Parse::Tree::PackagingNames names;
+    UnitInfo* unit_info;
+  };
+  // All imports corresponding to a single package, for the map.
+  struct PackageImports {
+    Parse::Node node;
```
zygoloid, line 34: A comment here would be useful: which node is this, and in which parse tree?

toolchain/check/check.cpp
```diff
@@ -227,11 +266,23 @@ static auto TrackImport(
...
 
+  // Get the package imports. Avoid creating unused values for insert conflicts
+  // because there's a SmallVector.
+  auto package_imports_it =
+      unit_info.package_imports_map.find(import.package_id);
+  if (package_imports_it == unit_info.package_imports_map.end()) {
+    package_imports_it = unit_info.package_imports_map
+                             .insert({import.package_id, {.node = import.node}})
+                             .first;
+  }
```
zygoloid: Hm, the designated initialization here makes this tricky. Maybe you could give `PackageImports` a constructor taking a `Parse::Node` and then use `try_emplace`:

```suggestion
  // Get the package imports.
  auto package_imports_it =
      unit_info.package_imports_map.try_emplace(import.package_id,
                                                import.node).first;
```

Non-blocking, though: I think this is OK as is.

toolchain/check/context.cpp
```diff
@@ -112,6 +114,41 @@ auto Context::NoteIncompleteClass(SemIR::ClassId class_id,
...
+    cross_reference_irs().Add(sem_ir);
+  }
+  if (has_failure) {
+    cross_reference_irs().Add(nullptr);
+  }
+  SemIR::CrossReferenceIRId last_id(cross_reference_irs().size() - 1);
+
+  auto inst_id = AddInst(SemIR::Import{.parse_node = import_node,
+                                       .first_cross_reference_ir_id = first_id,
+                                       .last_cross_reference_ir_id = last_id});
```
zygoloid: We switched to giving things that can be found by name lookup (namespaces, functions) a type in order to reduce the need to special-case them elsewhere and avoid crashes when downstream checking code tries to look at the type of a thing it found by name lookup. I think we should do the same here. Maybe we could use `BuiltinNamespaceType` for package names too, given that the design says that a package name is a kind of namespace. This'd also fix the SemIR printing, where you have references to a `%import` that is never declared (we don't print a name for untyped nodes because we don't expect them to be referenced).

toolchain/check/context.cpp
```diff
@@ -112,6 +114,41 @@ auto Context::NoteIncompleteClass(SemIR::ClassId class_id,
...
+  if (name_id.is_valid()) {
+    // Add the import to lookup. Should always succeed because imports will be
+    // uniquely named.
+    AddNameToLookup(import_node, name_id, inst_id);
+    // This binding is added primarily to show the name in formatted output.
+    AddInst(SemIR::BindName{.parse_node = import_node,
+                            .type_id = SemIR::TypeId::Error,
+                            .name_id = name_id,
+                            .value_id = inst_id});
```
zygoloid: Is this the representation you want in the longer term, or should we add a TODO here to revisit this?

zygoloid (commented):
Looks like good progress.

toolchain/check/check.cpp
jonmeow: Adding a comment.

toolchain/check/check.cpp
jonmeow: Done, I think this is a bit better.

toolchain/check/context.cpp
jonmeow: Yeah, I'd been considering whether a type would be needed, just hadn't gotten that far yet. Using BuiltinNamespaceType for now, since it exists. (I have some apprehension about divergent behaviors)

toolchain/check/context.cpp
jonmeow: Short of better ideas, I think this is long term: I assume it's desirable to show the name in formatted output, and we can't store it on the Import directly due to space constraints.

Added a couple explanatory comments of the tradeoff.

zygoloid (approved):
Thanks!

PR #3416, opened 2023-11-21
Support for reading source code from stdin and other unusual places.

zygoloid:
- Treat an input of `-` as meaning stdin.

- Fix building of an llvm::MemoryBuffer from a non-regular file.

- Do not enforce filename restrictions on non-regular files.

- Do not invent an output file name based on the name of a non-regular file.

### DIFF
--- a/toolchain/autoupdate_testdata.py
+++ b/toolchain/autoupdate_testdata.py
@@ -41,6 +41,8 @@ def main() -> None:
                 f"Args do not seem to be test files; for example, {sys.argv[1]}"
             )
         argv.append("--file_tests=" + ",".join(file_tests))
+    # Provide an empty stdin so that the driver tests that read from stdin
+    # don't block waiting for input. This matches the behavior of `bazel test`.
     subprocess.run(argv, check=True, stdin=subprocess.DEVNULL)
 
 
--- a/toolchain/autoupdate_testdata.py
+++ b/toolchain/autoupdate_testdata.py
@@ -41,7 +41,7 @@ def main() -> None:
                 f"Args do not seem to be test files; for example, {sys.argv[1]}"
             )
         argv.append("--file_tests=" + ",".join(file_tests))
-    subprocess.run(argv, check=True)
+    subprocess.run(argv, check=True, stdin=subprocess.DEVNULL)
 
 
 if __name__ == "__main__":
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -19,7 +19,8 @@ namespace Carbon::Check {
 struct UnitInfo {
   explicit UnitInfo(Unit& unit)
       : unit(&unit),
-        translator(unit.tokens, unit.tokens->filename(), unit.parse_tree),
+        translator(unit.tokens, unit.tokens->source().filename(),
+                   unit.parse_tree),
         err_tracker(*unit.consumer),
         emitter(translator, err_tracker) {}
 
@@ -47,9 +48,9 @@ struct UnitInfo {
 // imports should suppress errors where it makes sense.
 static auto CheckParseTree(const SemIR::File& builtin_ir, UnitInfo& unit_info,
                            llvm::raw_ostream* vlog_stream) -> void {
-  unit_info.unit->sem_ir->emplace(*unit_info.unit->value_stores,
-                                  unit_info.unit->tokens->filename().str(),
-                                  &builtin_ir);
+  unit_info.unit->sem_ir->emplace(
+      *unit_info.unit->value_stores,
+      unit_info.unit->tokens->source().filename().str(), &builtin_ir);
 
   // For ease-of-access.
   SemIR::File& sem_ir = **unit_info.unit->sem_ir;
@@ -279,7 +280,8 @@ static auto BuildApiMapAndDiagnosePackaging(
     if (!is_impl) {
       auto [entry, success] = api_map.insert({import_key, &unit_info});
       if (!success) {
-        llvm::StringRef prev_filename = entry->second->unit->tokens->filename();
+        llvm::StringRef prev_filename =
+            entry->second->unit->tokens->source().filename();
         if (packaging) {
           CARBON_DIAGNOSTIC(DuplicateLibraryApi, Error,
                             "Library's API previously provided by `{0}`.",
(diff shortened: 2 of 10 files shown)


toolchain/driver/driver.cpp
```diff
@@ -535,9 +539,22 @@ class Driver::CompilationUnit {
     } else {
       llvm::SmallString<256> output_file_name = options_.output_file_name;
       if (output_file_name.empty()) {
+        if (!source_->is_regular_file()) {
+          // Don't invent file names like `-.o` or `/dev/stdin.o`.
+          driver_->error_stream_
+              << "ERROR: Output file name must be specified for input "
+              << input_file_name_ << " that is not a regular file.\n";
```
chandlerc: Seems especially important to quote a bit here given that it may be something weird...

```suggestion
              << "ERROR: Output file name must be specified for input '"
              << input_file_name_ << "' that is not a regular file.\n";
```

toolchain/source/source_buffer.cpp
```diff
@@ -41,25 +83,21 @@ auto SourceBuffer::CreateFromFile(llvm::vfs::FileSystem& fs,
...
+  int64_t size = is_regular_file ? status->getSize() : -1;
 
   llvm::ErrorOr<std::unique_ptr<llvm::MemoryBuffer>> buffer =
       (*file)->getBuffer(filename, size, /*RequiresNullTerminator=*/false);
-  if (buffer.getError()) {
-    CARBON_DIAGNOSTIC(ErrorReadingFile, Error, "Error reading file: {0}",
-                      std::string);
-    emitter.Emit(filename, ErrorReadingFile, file.getError().message());
+
+  if (!CheckBuffer(buffer, "<stdin>", emitter)) {
```
chandlerc: Pretty sure this should be:
```suggestion
  if (!CheckBuffer(buffer, filename, emitter)) {
```

Is it worth trying to craft a test of this? I don't see any good way sadly -- would need an "actual" file that was un-readable or over 2G which seems ... not friendly for our testdata. =]

Maybe we could test this in a unit-test? With somewhat minor platform-specific if-defs, it's not too hard to write a unittest that ends up failing either of these diagnostics... With the vfs layer it might be quite easy to get a virtual file with a huge buffer that is never materialized...

chandlerc (commented):
FWIW, I'm happy with this approach and the code generally looks good to me, but probably good to wait for Jon to take a look before landing.

toolchain/driver/driver.cpp
jonmeow: I think we've used backticks elsewhere, including for filenames. :)

```suggestion
              << "ERROR: Output file name must be specified for input `"
              << input_file_name_ << "` that is not a regular file.\n";
```

toolchain/autoupdate_testdata.py
```diff
@@ -41,7 +41,7 @@ def main() -> None:
                 f"Args do not seem to be test files; for example, {sys.argv[1]}"
             )
         argv.append("--file_tests=" + ",".join(file_tests))
-    subprocess.run(argv, check=True)
+    subprocess.run(argv, check=True, stdin=subprocess.DEVNULL)
```
jonmeow, line 46: Why is stdin specified here? Should there be a comment, or could this be removed? I'm a little concerned if carbon won't play nicely with python defaults.

toolchain/check/check.cpp
```diff
@@ -298,13 +298,16 @@ static auto BuildApiMapAndDiagnosePackaging(
...
+    // is stdin, no filename checking is performed.
     auto filename = unit_info.unit->tokens->filename();
     static constexpr llvm::StringLiteral ApiExt = ".carbon";
     static constexpr llvm::StringLiteral ImplExt = ".impl.carbon";
     bool is_api_with_impl_ext = !is_impl && filename.ends_with(ImplExt);
     auto want_ext = is_impl ? ImplExt : ApiExt;
-    if (is_api_with_impl_ext || !filename.ends_with(want_ext)) {
+    if ((is_api_with_impl_ext || !filename.ends_with(want_ext)) &&
+        unit_info.unit->tokens->is_regular_file() &&
+        !unit_info.unit->tokens->filename().empty()) {
```
```suggestion
    if ((is_api_with_impl_ext || !filename.ends_with(want_ext)) &&
        unit_info.unit->tokens->is_regular_file() &&
        !filename.empty()) {
```

toolchain/lex/tokenized_buffer.h
```diff
@@ -234,6 +234,7 @@ class TokenizedBuffer : public Printable<TokenizedBuffer> {
   }
 
   auto filename() const -> llvm::StringRef { return source_->filename(); }
+  auto is_regular_file() const -> bool { return source_->is_regular_file(); }
```
jonmeow: Maybe we're starting to use enough of the source buffer elsewhere that it should be passed directly instead of having APIs exposed by the TokenizedBuffer? That'd be more consistent with how we treat TokenizedBuffer, Tree, and SemIR

toolchain/check/check.cpp
jonmeow: Can the is_regular_file check wrap more code? e.g., do we need to calculate is_api_with_impl_ext and want_ext for that case?

Also, given the regular file check, what's the use-case for a regular file that has an empty filename?

toolchain/source/source_buffer.h
```diff
@@ -50,13 +55,24 @@ class SourceBuffer {
     return text_->getBuffer();
   }
 
+  [[nodiscard]] auto is_regular_file() const -> bool {
```
jonmeow, line 58: I think we aren't typically doing nodiscard, we should really decide what to do about it.

```suggestion
  auto is_regular_file() const -> bool {
```

toolchain/source/source_buffer.cpp
```diff
@@ -18,6 +18,48 @@ struct FilenameTranslator : DiagnosticLocationTranslator<llvm::StringRef> {
...
+
+  llvm::ErrorOr<std::unique_ptr<llvm::MemoryBuffer>> buffer =
+      llvm::MemoryBuffer::getSTDIN();
+
+  if (!CheckBuffer(buffer, StdinName, emitter)) {
+    return std::nullopt;
+  }
+
+  return SourceBuffer(StdinName.str(), std::move(buffer.get()),
+                      /*is_regular_file=*/false);
```
jonmeow: FWIW it looks a bit like these two calls could be combined into a single CreateFromBuffer.

toolchain/source/source_buffer.cpp
```diff
@@ -18,6 +18,48 @@ struct FilenameTranslator : DiagnosticLocationTranslator<llvm::StringRef> {
...
+  static const llvm::StringLiteral StdinName = "<stdin>";
+
+  llvm::ErrorOr<std::unique_ptr<llvm::MemoryBuffer>> buffer =
+      llvm::MemoryBuffer::getSTDIN();
+
+  if (!CheckBuffer(buffer, StdinName, emitter)) {
+    return std::nullopt;
+  }
+
+  return SourceBuffer(StdinName.str(), std::move(buffer.get()),
```
jonmeow: Rather than passing a filename, can this switch to using `MemoryBuffer::getBufferIdentifier()`? Including for CreateFromFile, I get the impression that the end result might be identical, and reduce how much is passed around.

toolchain/source/source_buffer.cpp
```diff
@@ -41,25 +83,21 @@ auto SourceBuffer::CreateFromFile(llvm::vfs::FileSystem& fs,
...
-                      int64_t);
-    emitter.Emit(filename, FileTooLarge, size);
-    return std::nullopt;
-  }
+
+  // `stat` on a file without a known size gives a size of 0, which causes
+  // `llvm::vfs::File::getBuffer` to produce an empty buffer. Use a size of -1
+  // in this case so we get the complete file contents.
+  bool is_regular_file = status->isRegularFile();
+  int64_t size = is_regular_file ? status->getSize() : -1;
```
jonmeow, line 55: Since this doesn't use the size anymore, maybe the code should switch to `llvm::vfs::getBufferForFile`? That'd simplify this whole function [which was built around cautiously checking size before loading data].

toolchain/driver/driver.cpp
zygoloid: We use single-quotes a few lines below; going with that for now.

toolchain/source/source_buffer.cpp
zygoloid: I considered that. It would result in our performing an extra `fstat` call in the common case: we need to `fstat` the file to compute `is_regular_file`, and `getBuffer` will `fstat` the file itself to get the size if we don't pass one in. I don't know whether that's worth worrying about, but given that I think we want to stat the file here anyway it seems like we may as well pass in the size to skip the second stat.

toolchain/autoupdate_testdata.py
zygoloid: Added a comment to explain. Let me know if you'd prefer we handle this a different way.

I considered extending the testing infrastructure so we can specify the contents of stdin, but it seemed like excessive complexity. Happy to do something like that if you prefer.

toolchain/check/check.cpp

I added the check for an empty filename because it was requested on discord... but it can't happen because it's not possible to build a `SourceBuffer` with an empty filename with `is_regular_file()` being `true`. Dropped.

toolchain/lex/tokenized_buffer.h
zygoloid: Done.

toolchain/source/source_buffer.h
zygoloid: This file uses `[[nodiscard]]` on all its other accessors, so I'd prefer to be locally consistent here.

toolchain/source/source_buffer.cpp
zygoloid: Done.

toolchain/check/check.cpp
jonmeow: I think this was mixing signals, it was more that an empty filename would be one option to signal instead of is_regular_file. Anyways, SG

toolchain/source/source_buffer.h
jonmeow: Okay, re-raising the question on #toolchain (I'd brought this issue up a couple weeks ago -- I think the main result here is going to be I remove it later, and blame won't work as well as a consequence)

toolchain/source/source_buffer.cpp
zygoloid: We need to pass in the filename at least for the case where the buffer creation failed. I'm also feeling a little hesitant to use the undocumented result of `getSTDIN().getBufferIdentifier()` in our user-facing diagnostic output, even though it does produce exactly the string that I wanted here :)

toolchain/autoupdate_testdata.py
jonmeow: Probably okay for now.

toolchain/source/source_buffer.cpp
jonmeow: I think you missed this comment [edit: temporal issues]

jonmeow (approved):
LGTM, but please address https://github.com/carbon-language/carbon-lang/pull/3416#discussion_r1400899107 (if the suggestion doesn't work, go ahead and merge)

jonmeow:
You should also respond to https://github.com/carbon-language/carbon-lang/pull/3416#discussion_r1399961047

toolchain/source/source_buffer.cpp
zygoloid: I've added a test for the "can't open file" error. The easiest way I could find was opening a directory :)

I think we could test the other diagnostic via a unit test, if you think that's worth doing as part of this PR. We'd need to actually allocate 4GB of address space, which might hit a ulimit I suppose, but otherwise seems doable in a 64-bit-only test.

chandlerc: I think this is pretty solid, mostly wanted to do whatever testing was easy enough given the virtual FS stuff we have. Thanks!

chandlerc (approved):
LG

PR #3418, opened 2023-11-21
Remove nodiscard uses.

jonmeow:
Per [#toolchain discussion](https://discord.com/channels/655572317891461132/655578254970716160/1176632520834560211)

We'd at one point been trying to put `[[nodiscard]]` everywhere, but then we stopped because it had felt verbose without finding many issues (plus, people plain forgot to add it). Some history in #888.

Since newer code gets added without it, we now have code like:

```
  auto GetLineInfo(Line line) -> LineInfo&;
  [[nodiscard]] auto GetLineInfo(Line line) const -> const LineInfo&;
  auto AddLine(LineInfo info) -> Line;
  auto GetTokenInfo(Token token) -> TokenInfo&;
  [[nodiscard]] auto GetTokenInfo(Token token) const -> const TokenInfo&;
  auto AddToken(TokenInfo info) -> Token;
  [[nodiscard]] auto GetTokenPrintWidths(Token token) const -> PrintWidths;
```

Here, the lack of `[[nodiscard]]` doesn't mean anything: for example, `GetLineInfo` should not have its result discarded if it's called. But the mix could be confusing for readers.

As a resolution, remove the attribute. `[[nodiscard]]` should be treated like other attributes going forward, which essentially means "avoid in general, add a comment to explain why the attribute is needed" rather than use-as-default.

### DIFF
--- a/common/error.h
+++ b/common/error.h
@@ -20,7 +20,7 @@ namespace Carbon {
 struct Success {};
 
 // Tracks an error message.
-class [[nodiscard]] Error : public Printable<Error> {
+class Error : public Printable<Error> {
  public:
   // Represents an error state.
   explicit Error(llvm::Twine location, llvm::Twine message)
@@ -67,7 +67,7 @@ class [[nodiscard]] Error : public Printable<Error> {
 // Holds a value of type `T`, or an Error explaining why the value is
 // unavailable.
 template <typename T>
-class [[nodiscard]] ErrorOr {
+class ErrorOr {
  public:
   // Constructs with an error; the error must not be Error::Success().
   // Implicit for easy construction on returns.
@@ -137,7 +137,7 @@ class ErrorBuilder {
   // Accumulates string message to a temporary `ErrorBuilder`. After streaming,
   // the builder must be converted to an `Error` or `ErrorOr`.
   template <typename T>
-  [[nodiscard]] auto operator<<(const T& message) && -> ErrorBuilder&& {
+  auto operator<<(const T& message) && -> ErrorBuilder&& {
     *out_ << message;
     return std::move(*this);
   }
--- a/common/error.h
+++ b/common/error.h
@@ -20,7 +20,9 @@ namespace Carbon {
 struct Success {};
 
 // Tracks an error message.
-class Error : public Printable<Error> {
+//
+// This is nodiscard to enforce error handling prior to destruction.
+class [[nodiscard]] Error : public Printable<Error> {
  public:
   // Represents an error state.
   explicit Error(llvm::Twine location, llvm::Twine message)
@@ -66,8 +68,10 @@ class Error : public Printable<Error> {
 
 // Holds a value of type `T`, or an Error explaining why the value is
 // unavailable.
+//
+// This is nodiscard to enforce error handling prior to destruction.
 template <typename T>
-class ErrorOr {
+class [[nodiscard]] ErrorOr {
  public:
   // Constructs with an error; the error must not be Error::Success().
   // Implicit for easy construction on returns.
--- a/explorer/ast/bindings.h
+++ b/explorer/ast/bindings.h
@@ -78,7 +78,9 @@ class Bindings : public Printable<Bindings> {
   auto witnesses() const -> const ImplWitnessMap& { return witnesses_; }
 
   // Determine whether this is an empty set of bindings.
-  auto empty() const -> bool { return args_.empty() && witnesses_.empty(); }
+  [[nodiscard]] auto empty() const -> bool {
+    return args_.empty() && witnesses_.empty();
+  }
 
  private:
   BindingMap args_;
--- a/explorer/ast/bindings.h
+++ b/explorer/ast/bindings.h
@@ -78,9 +78,7 @@ class Bindings : public Printable<Bindings> {
   auto witnesses() const -> const ImplWitnessMap& { return witnesses_; }
 
   // Determine whether this is an empty set of bindings.
-  [[nodiscard]] auto empty() const -> bool {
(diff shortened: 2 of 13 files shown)


common/error.h
```diff
@@ -20,7 +20,7 @@ namespace Carbon {
 struct Success {};
 
 // Tracks an error message.
-class [[nodiscard]] Error : public Printable<Error> {
```
geoffromer, line 23: I definitely think we should keep the attribute here: when a function returns `Error`, there's a high risk of its return value being ignored, because it's not salient information on the happy path. Adding this attribute to `absl::Status` was a huge win for the Google codebase

common/error.h
```diff
@@ -67,7 +67,7 @@ class [[nodiscard]] Error : public Printable<Error> {
 // Holds a value of type `T`, or an Error explaining why the value is
 // unavailable.
 template <typename T>
-class [[nodiscard]] ErrorOr {
```
geoffromer, line 70: Optional, no change needed: I lean toward keeping `[[nodiscard]]` here, although it's less clear-cut than with `Error`. The risk of the error being discarded is much lower, but it's not zero, and the noise cost of a class-level attribute is much lower than a function-level attribute.

common/error.h
```diff
@@ -137,7 +137,7 @@ class ErrorBuilder {
   // Accumulates string message to a temporary `ErrorBuilder`. After streaming,
   // the builder must be converted to an `Error` or `ErrorOr`.
   template <typename T>
-  [[nodiscard]] auto operator<<(const T& message) && -> ErrorBuilder&& {
```
geoffromer, line 140: I think we should probably keep this one (and add it to the other `<<` overloads for this type: the usual convention is that you can ignore the return value of `<<` if you don't want to stream anymore, so the risk of mistakes is relatively high. Arguably the comment on this function already explains why the attribute is needed.

explorer/interpreter/pattern_match.h
```diff
@@ -26,12 +26,11 @@ class Arena;
 // PatternMatch call and then merged into an existing scope on success.
 // The matches for generic variables in the pattern are output in
 // `generic_args`.
-[[nodiscard]] auto PatternMatch(Nonnull<const Value*> p, ExpressionResult v,
```
geoffromer: This one seems worth keeping: the return value is easily overlooked because the function isn't named as though it returns a value, and disregarding the return value is almost certainly a bug.

explorer/ast/bindings.h
```diff
@@ -78,9 +78,7 @@ class Bindings : public Printable<Bindings> {
   auto witnesses() const -> const ImplWitnessMap& { return witnesses_; }
 
   // Determine whether this is an empty set of bindings.
-  [[nodiscard]] auto empty() const -> bool {
```
geoffromer: Optional, no change needed: I lean toward keeping this one, although I don't feel too strongly about it: "empty" is both an adjective and a verb, and people who naively read it as a verb are likely to disregard the return value, which is almost certainly a bug. On the other hand, experienced C++ programmers are unlikely to make that mistake, because of experience with `empty` methods on containers.

jonmeow: Reverted on the basis that explorer is archived, so I'm not going to feel bad having it have readability issues.

explorer/interpreter/pattern_match.h
jonmeow: Reverted on the basis that explorer is archived, so I'm not going to feel bad having it have readability issues.

common/error.h
jonmeow: My expectation is that the idea that it could have a mistake is little different from other API calls. That there's a non-nodiscard version is, to me, more indicative of the problem that we're adding it where no bugs result, and so it's more churn than value. Can you give an example of how this would be forgotten in code, or a bug that resulted from the overload? What makes this different from, for example, a factory function which we would not nodiscard?

common/error.h
jonmeow: llvm::Error uses nodiscard so keeping it here for consistency, with a comment.

common/error.h
jonmeow: As above, keeping.

common/error.h
geoffromer: I agree this is a closer call, and I don't feel too strongly about it. And yes, the other overload is much more debatable, because it's `&`-qualified, so the return value is presumably available by other means. 

> Can you give an example of how this would be forgotten in code, or a bug that resulted from the overload?

I'm thinking of things like

```c++
ProgramError(source_location) << "Error message";
```

> What makes this different from, for example, a factory function which we would not nodiscard?

Discarded-return-value bugs primarily happen when the programmer doesn't realize that the function has a return value in the first place, or is mistaken about the nature of the return value. Factory functions (and other functions that primarily return a value) are normally named after what they return, which makes that kind of mistake highly unlikely. This operator doesn't have that protection, because the name `<<` (in a streaming rather than bit-shifting context) is normally used for functions whose return value is ignorable.

jonmeow: `ProgramError` may not be too suggestive of its use, but naming is full of trade-offs. Also, streams must go somewhere. Then, context is important: functions using this would be returning an `Error` or `ErrorOr`. To forget to use the returned error, a return statement would need to be missing, which should be expected to cause other issues. As a consequence, I'm not clear there's utility for an exception here, or that the stream is distinct from other returned builders that may go unused (for example, with diagnostics).

geoffromer: I'm not sure I agree with this reasoning, but as I said, I don't feel too strongly.

geoffromer (approved):

PR #3421, opened 2023-11-27
Upgrade to latest `rules_python`.

chandlerc:
Also requires switching to using `pip_parse` and providing a fully resolved requirements lock file. This moves the input requirements to the `requirements.in` file, and processes it with:

```console
$ bazel run //github_tools:requirements.update
```

This will regenerate the `requirements.txt` file that is checked into the repository. The advice in the documentation is specifically to keep this file checked into the repository for hermetic builds with stable Python dependency versions.

This should fix builds on systems where the Python version is 3.12 and newer and the older version of `rules_python` stops working with errors due to removal of long-deprecated interfaces. I specifically hit this on Fedora 39.


chandlerc:
(I messed up creating this PR, deleting and trying again, sorry for noise.)

PR #3422, opened 2023-11-27
Add Linux AArch64 Bazel support and simplify.

chandlerc:
There was a lot of repetition and unnecessary cruft in our Bazel toolchain support. Switch to generating all of it with a single macro that handles everything. This should make no real difference but dramatically simplifies adding a new CPU.

Use this simplified system and add `aarch64` which is how Arm 64-bit CPU support shows up on a Linux host.

Also teach the basic scripts to map `aarch64` to `arm64` which is used in the released artifact strings.

### DIFF
--- a/bazel/cc_toolchains/clang_cc_toolchain_config.bzl
+++ b/bazel/cc_toolchains/clang_cc_toolchain_config.bzl
@@ -17,6 +17,7 @@ load(
     "variable_with_value",
     "with_feature_set",
 )
+load("@rules_cc//cc:defs.bzl", "cc_toolchain", "cc_toolchain_suite")
 load(
     ":clang_detected_variables.bzl",
     "clang_bindir",
@@ -886,7 +887,7 @@ def _impl(ctx):
 
     # Next, add the features based on the target platform. Here too the
     # features are order sensitive. We also setup the sysroot here.
-    if ctx.attr.target_cpu == "k8":
+    if ctx.attr.target_cpu in ["aarch64", "k8"]:
         features.append(sanitizer_static_lib_flags)
         features.append(linux_flags_feature)
         sysroot = None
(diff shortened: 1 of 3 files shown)


chandlerc:
(posted comment on wrong PR)

bazel/cc_toolchains/clang_cc_toolchain_config.bzl
```diff
@@ -906,8 +907,7 @@ def _impl(ctx):
     else:
         fail("Unsupported target platform!")
 
-    # TODO: Need to support non-macOS ARM platforms here.
-    if ctx.attr.target_cpu == "darwin_arm64":
+    if ctx.attr.target_cpu in ["aarch64", "darwin_arm64"]:
         features.append(aarch64_cpu_flags)
```
jonmeow, line 911: FWIW it feels slightly odd that this is unified as aarch64 here, but arm64  in script_utils.py. But maybe I'm missing something.

jonmeow (approved):
Looks good, nice cleanup.

bazel/cc_toolchains/clang_cc_toolchain_config.bzl
chandlerc: AFAIK, `aarch64` is more accurate term for the CPU. The `arm64` comes from macOS AFAIK, and isn't the nomenclature that Arm officially uses. It seems reasonable as a suffix to `darwin`, but anywhere else I would hope far `aarch64` to be the canonical name.

Sadly, Go picked `arm64` as *their* canonical name: https://github.com/golang/go/blob/master/src/go/build/syslist.go#L62

And that is why buildifier, buildozer, and the target determinator use `arm64` suffix even on Linux -- they're Go builds. So for anything where we have to match Go or macOS, we should canonicalize to `arm64`. For anything else or anything generic across, we should try to use `aarch64` wherever we can. At least, that's the strategy I've been trying to follow here.

PR #3423, opened 2023-11-27
Upgrade to latest rules_python.

chandlerc:
Also requires switching to using `pip_parse` and providing a fully
resolved requirements lock file. This moves the input requirements to
the `requirements.in` file, and processes it with:

```console
$ bazel run //github_tools:requirements.update
```

This will regenerate the `requirements.txt` file that is checked into
the repository. The advice in the documentation is specifically to keep
this file checked into the repository for hermetic builds with stable
Python dependency versions.

This should fix builds on systems where the Python version is 3.12 and
newer and the older version of `rules_python` stops working with errors
due to removal of long-deprecated interfaces.


chandlerc:
Had to fix Ubuntu by pinning to a specific interpreter version -- otherwise we'd end failing the test for the lock file because the Python interpreter version would be different.

I picked 3.11 because 3.9 gave me errors on Fedora 39 AArch64, and it seems fine to update to that for our testing as long as CI is happy with it.

jonmeow (approved):

jonmeow:
Approved, but I want to verify that github_tools still belongs in the repo due to the maintenance costs that crop up for it: https://discord.com/channels/655572317891461132/707150492370862090/1178735261228740618

chandlerc:
Happy to rip it all out if that's the decision, but would like to get a baseline that works on Linux AArch64 so merging for now. Thanks!

PR #3427, opened 2023-11-28
Auto-update pre-commit repositories and fix.

chandlerc:
The version of `flake8` was too old to support with Python 3.12 -- there is new F-string support that caused false positives sadly. The updated version has fixes for all of these.

This in turn updates codespell which has picked up several new fixes that actually fire in our code, so also fix everything. While we don't do more in-depth updates to old proposals, similar to simply fixing broken links, fixing automatically detected typos seems scalable and fine.

All edits were automatically generated here.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,7 +12,7 @@ default_language_version:
 
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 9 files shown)


chandlerc:
(Initially used the non-frozen update by mistake. Fixed to be frozen, sorry for the noise!)

jonmeow (approved):

PR #3429, opened 2023-11-28
Tidy up lint issues.

jonmeow:
These are manual fixes; mostly from clang-tidy, some from clangd (which notes unused includes).

In typed_insts, adding inlline due to misc-definitions-in-headers. Per discussion, clang-tidy is wrong, but inline silences it.

For parameter name skew in definition versus declaration, I'm just using the name from the definition.

### DIFF
--- a/toolchain/parse/node_kind.cpp
+++ b/toolchain/parse/node_kind.cpp
@@ -47,6 +47,7 @@ auto NodeKind::child_count() const -> int32_t {
   return child_count;
 }
 
+// NOLINTNEXTLINE(readability-function-size): It's hard to extract macros.
 void CheckNodeMatchesLexerToken(NodeKind node_kind, Lex::TokenKind token_kind,
                                 bool has_error) {
   switch (node_kind) {
@@ -69,9 +70,9 @@ void CheckNodeMatchesLexerToken(NodeKind node_kind, Lex::TokenKind token_kind,
     MatchActions                      \
   }
 
-#define CARBON_CASE(Name, MatchActions) \
-  case NodeKind::Name:                  \
-    MatchActions;                       \
+#define CARBON_CASE(Name, MatchActions)                    \
+  case NodeKind::Name:                                     \
+    MatchActions; /* NOLINT(bugprone-macro-parentheses) */ \
     break;
 
 #define CARBON_PARSE_NODE_KIND_BRACKET(Name, BracketName, MatchActions) \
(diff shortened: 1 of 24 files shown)


toolchain/parse/node_kind.cpp
```diff
@@ -47,6 +47,7 @@ auto NodeKind::child_count() const -> int32_t {
   return child_count;
 }
 
+// NOLINTNEXTLINE(readability-function-size): It's hard to extract macros.
 void CheckNodeMatchesLexerToken(NodeKind node_kind, Lex::TokenKind token_kind,
                                 bool has_error) {
```
zygoloid:
```suggestion
auto CheckNodeMatchesLexerToken(NodeKind node_kind, Lex::TokenKind token_kind,
                                bool has_error) -> void {
```
(Since we're here. Not sure why lint didn't catch this.)

toolchain/sem_ir/typed_insts.h
```diff
@@ -587,15 +587,15 @@ struct VarStorage {
...
-constexpr bool HasParseNode<T, decltype(&T::parse_node)> = true;
+static constexpr bool HasParseNode<T, decltype(&T::parse_node)> = true;
 
 // HasTypeId<T> is true if T has a `TypeId type_id` field.
 template <typename T, typename TypeIdType = TypeId T::*>
-constexpr bool HasTypeId = false;
+static constexpr bool HasTypeId = false;
 template <typename T>
-constexpr bool HasTypeId<T, decltype(&T::type_id)> = true;
+static constexpr bool HasTypeId<T, decltype(&T::type_id)> = true;
```
zygoloid: The linter is wrong here, and `static` is an ODR trap. Does adding `inline` instead silence the lint? If not, let's just suppress it here.

```suggestion
// HasParseNode<T> is true if T has a `Parse::Node parse_node` field.
template <typename T, typename ParseNodeType = Parse::Node T::*>
inline constexpr bool HasParseNode = false;
template <typename T>
inline constexpr bool HasParseNode<T, decltype(&T::parse_node)> = true;

// HasTypeId<T> is true if T has a `TypeId type_id` field.
template <typename T, typename TypeIdType = TypeId T::*>
inline constexpr bool HasTypeId = false;
template <typename T>
inline constexpr bool HasTypeId<T, decltype(&T::type_id)> = true;
```

toolchain/parse/node_kind.cpp
jonmeow: clang-tidy does `auto` when there's a return type, but explicitly doesn't hit `void`. It's sad there's not even an option.

toolchain/sem_ir/typed_insts.h
jonmeow: Done, works.

zygoloid (approved):

PR #3431, opened 2023-11-29
Error when passing StringRef to CARBON_DIAGNOSTIC.

jonmeow:
This gets to a lifetime subtlety, particularly with things like the sorting diagnostic consumer that delay output. In order to reduce the chance of accidental references, disallow StringRef in the diagnostics.

For example:

```
./toolchain/diagnostics/diagnostic_emitter.h:162:5: error: static_assert failed due to requirement '!std::is_same_v<llvm::StringRef, llvm::StringRef>' "Use std::string or llvm::StringLiteral for diagnostic lifetimes."
    static_assert(
    ^
toolchain/check/convert.cpp:477:11: note: in instantiation of member function 'Carbon::Internal::DiagnosticBase<std::string, std::string, llvm::StringRef>::DiagnosticBase' requested here
          CARBON_DIAGNOSTIC(StructInitMissingFieldInConversion, Error,
          ^
./toolchain/diagnostics/diagnostic_emitter.h:47:7: note: expanded from macro 'CARBON_DIAGNOSTIC'
      ::Carbon::Internal::DiagnosticBase<__VA_ARGS__>(        \
      ^
```

### DIFF
--- a/toolchain/diagnostics/diagnostic_emitter.h
+++ b/toolchain/diagnostics/diagnostic_emitter.h
@@ -158,7 +158,11 @@ template <typename... Args>
 struct DiagnosticBase {
   explicit constexpr DiagnosticBase(DiagnosticKind kind, DiagnosticLevel level,
                                     llvm::StringLiteral format)
-      : Kind(kind), Level(level), Format(format) {}
+      : Kind(kind), Level(level), Format(format) {
+    static_assert(
+        (... && !std::is_same_v<Args, llvm::StringRef>),
+        "Use std::string or llvm::StringLiteral for diagnostic lifetimes.");
+  }
 
   // Calls formatv with the diagnostic's arguments.
   auto FormatFn(const DiagnosticMessage& message) const -> std::string {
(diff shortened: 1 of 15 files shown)


toolchain/diagnostics/diagnostic_emitter.h
```diff
@@ -158,7 +158,11 @@ template <typename... Args>
 struct DiagnosticBase {
   explicit constexpr DiagnosticBase(DiagnosticKind kind, DiagnosticLevel level,
                                     llvm::StringLiteral format)
-      : Kind(kind), Level(level), Format(format) {}
+      : Kind(kind), Level(level), Format(format) {
+    static_assert(
+        (... && !std::is_same_v<Args, llvm::StringRef>),
+        "Use std::string or llvm::StringLiteral for diagnostic lifetimes.");
```
chandlerc: Maybe to make this a bit more clear:

```suggestion
        "Use std::string or llvm::StringLiteral for diagnostics to avoid lifetime issues.");
```

chandlerc (approved):
Good catch!

Would be nice to have a more general way to detect this, but it seems good to just get at least this safety net in place.

An optional suggestion on the static assert message below.

PR #3437, opened 2023-11-29
Refactor ValueStore to reduce template type repetition.

jonmeow:
There's a trade-off here of explicitness in the use versus repetition, but I'm hoping the Id offers sufficient info (also, some Ids already relied on this, so this builds consistency). The forward declarations I'm mixed on, but they are difficult to avoid if heading down this route due to interdependencies between ids and types which contain ids.

### DIFF
--- a/toolchain/base/value_store.h
+++ b/toolchain/base/value_store.h
@@ -49,7 +49,7 @@ class Real : public Printable<Real> {
 
 // Corresponds to an integer value represented by an APInt.
 struct IntId : public IdBase, public Printable<IntId> {
-  using IndexedType = const llvm::APInt;
+  using ValueType = const llvm::APInt;
   static const IntId Invalid;
   using IdBase::IdBase;
   auto Print(llvm::raw_ostream& out) const -> void {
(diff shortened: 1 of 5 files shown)


toolchain/base/value_store.h
```diff
@@ -117,14 +117,17 @@ class ValueStoreNotPrintable {};
...
-                              Yaml::Printable<ValueStore<IdT, ValueT>>,
-                              Internal::ValueStoreNotPrintable> {
+    : public std::conditional<
+          std::is_base_of_v<Printable<typename IdT::IndexedType>,
+                            typename IdT::IndexedType>,
+          Yaml::Printable<ValueStore<IdT>>, Internal::ValueStoreNotPrintable> {
  public:
   // Stores the value and returns an ID to reference it.
-  auto Add(ValueT value) -> IdT {
+  auto Add(typename IdT::IndexedType value) -> IdT {
```
zygoloid, line 130: It'd be nice to add a private `using ValueT = typename IdT::IndexedType;` to the class to reduce the repetition here (though that doesn't help for the base specifier).

toolchain/sem_ir/ids.h
```diff
@@ -13,8 +13,17 @@
...
+// Forward declare indexed types, for integration with ValueStore.
+class File;
+class Inst;
+struct Class;
+struct Function;
+struct TypeInfo;
+
 // The ID of an instruction.
 struct InstId : public IndexBase, public Printable<InstId> {
+  using IndexedType = Inst;
```
zygoloid: Given the shift from `Index` to `Id`, I think `IndexedType` isn't a great name. How about using `ValueType` here and something like `ElementType` for the elements of `BlockId`s?

jonmeow:
Accidentally deleted the comment about typedefs, but that should be done too.

zygoloid (approved):

PR #3438, opened 2023-11-30
Add a distinct semir file separator.

jonmeow:
This removes the filename from the file-scoped block, and places it above to make it clear where the full SemIR begins (with multifile, providing a barrier between).


jonmeow:
FYI @chandlerc this is the outcome of #3396 and discussion with zygoloid.

zygoloid (commented):
LGTM, but I'd like to hear @chandlerc's thoughts too.

toolchain/check/testdata/basics/multifile.carbon
```diff
@@ -14,15 +14,17 @@ package B api;
...
+// CHECK:STDOUT: file {
 // CHECK:STDOUT:   %A: <function> = fn_decl @A
 // CHECK:STDOUT: }
 // CHECK:STDOUT:
 // CHECK:STDOUT: fn @A() {
 // CHECK:STDOUT: !entry:
 // CHECK:STDOUT:   return
 // CHECK:STDOUT: }
-// CHECK:STDOUT: file "b.carbon" {
+// CHECK:STDOUT: --- b.carbon
```
chandlerc, line 28: Could we add a blank line below the file marker, and at the end of the file (or above the file marker) when there are multiple files? Just want it to stand apart from the other top-level things.

chandlerc (approved):
Nice!

toolchain/check/testdata/basics/multifile.carbon
jonmeow: Added newlines.

I'm not making it conditional because that would require awareness on the part of the driver of how output is constructed, which I'd like to avoid if possible. (I'm hoping an end-of-file newline is okay)

PR #3440, opened 2023-11-30
Have autoupdate discard conflict markers when possible.

jonmeow:
This now puts file content into a string, allowing conflict markers to be elided from file content. When code executes, this means it executes without seeing conflict markers, without a temporary update to the file that would only remove conflict markers.

Also refactors the main process flow, because it was getting a little too lengthy. This means passing a bunch of parameters passed around (partly because TestContext is private on the test class, and I don't want to change that). I'm hoping that overall it's easier to read the core loop now.

Mostly tested with some manually added conflict markers, and that current tests don't change.

### DIFF
--- a/testing/file_test/file_test_base.cpp
+++ b/testing/file_test/file_test_base.cpp
@@ -13,6 +13,7 @@
 #include "absl/flags/flag.h"
 #include "absl/flags/parse.h"
 #include "common/check.h"
+#include "common/error.h"
 #include "llvm/ADT/StringExtras.h"
 #include "llvm/ADT/Twine.h"
 #include "llvm/Support/FormatVariadic.h"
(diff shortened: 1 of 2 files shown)


testing/file_test/file_test_base.cpp
```diff
@@ -265,151 +266,119 @@ auto FileTestBase::DoArgReplacements(
...
+      test_files->push_back(FileTestBase::TestFile(
+          split->file_name.str(),
+          llvm::StringRef(
+              split->file_start,
+              line_trimmed.begin() - split->file_start - SplitPrefix.size())));
+    } else if (split->found_content_pre_split) {
+      // For the first split, we make sure there was no content prior.
+      return ErrorBuilder() << "When using split files, there must be no "
+                               "content before the "
+                               "first split file.";
```
zygoloid:
```suggestion
                               "content before the first split file.";
```

testing/file_test/file_test_base.cpp
```diff
@@ -265,151 +266,119 @@ auto FileTestBase::DoArgReplacements(
...
+    if (is_end) {
+      *inside_conflict_marker = false;
+      return true;
+    }
+
+    // Look for CHECK lines, which can be discarded.
+    if (line_trimmed.starts_with("// CHECK:STDOUT:") ||
+        line_trimmed.starts_with("// CHECK:STDERR:")) {
+      return true;
+    }
```
zygoloid, line 303: This somewhat duplicates the handling in `TryConsumeCheck`. I think that can be avoided though: if you reorder the call to `TryConsumeCheck` to before `TryConsumeConflictMarker`, then you could just error here if you see *any* line that's not a conflict marker.

zygoloid (commented):
Is the overall process here doing what we want? It looks like we will go through and strip out conflict markers *after* we run the test, which means we'll first run the toolchain on the source file with conflict markers still present in it, which will typically generate some bogus diagnostics, requiring an additional round of autoupdate. I think we'll need three rounds of autoupdate in total for check tests with conflicts:

- First round: generate errors about the conflict markers, and update the file to remove the conflict markers and add the error CHECKs.
- Second round: generate output with the right set of CHECK lines but wrong line numbers due to the CHECK lines for the conflict marker errors.
- Third round: generate output with correct line numbers.

I think either we should remove the conflict markers *before* we run the test when in autoupdate mode, or we should add back in the iterate-until-converged logic that the old python autoupdate script had. (Maybe limit it to at most, say, three attempts?)

(That said: I'm fine with you saying that adding a loop around the autoupdate logic should be done in a different PR.)

testing/file_test/file_test_base.cpp
jonmeow: If conflict marker checks weren't first, then content inside conflict markers would be incorrectly treated as valid.

jonmeow:
As discussed: after thinking further, I still lean against an update loop. But the latest commit should address the conflict marker problem by eliding them -- by manufacturing test file content.

testing/file_test/file_test_base.cpp
```diff
@@ -495,6 +491,196 @@ auto FileTestBase::TransformExpectation(int line_index, llvm::StringRef in)
...
+
+    bool is_consumed = false;
+    CARBON_ASSIGN_OR_RETURN(
+        is_consumed,
+        TryConsumeConflictMarker(line, line_trimmed, &inside_conflict_marker));
+    if (is_consumed) {
+      continue;
+    }
+
+    CARBON_ASSIGN_OR_RETURN(
```
zygoloid, line 625: Maybe something like:
```suggestion
    // At this point, remaining lines are part of the test input.
    CARBON_ASSIGN_OR_RETURN(
```
... to make it clear that we're removing conflict markers but nothing else before here.

testing/file_test/file_test_base.cpp
zygoloid: Including the CHECK lines from both sides of the conflict marker doesn't seem better or worse than including the CHECK lines from neither side to me -- either way, line numbering for the rest of the file will be wrong. And I don't think it matters: we only remove conflict markers in autoupdate mode, in which case the list of old check lines shouldn't be used for anything.

Or am I missing something?

(I don't mean to push hard on this; while I think it'd be a bit cleaner to not repeat these CHECK line checks here, I think it's not a big deal.)

zygoloid: OK, looking at the rest of the updated PR, I think this suggestion doesn't work any more, because you want to remove the conflict lines but not the check lines before you handle file splits and build the input file contents.

zygoloid (approved):

testing/file_test/file_test_base.cpp
jonmeow: Note, before I think I was also focused on the line_index increment and wanting to have one place to do that. It'd need to be before CHECK is processed but shouldn't be done until after conflict markers decide whether to drop the line. That could be addressed by conditionality around incrementing the line_index, or doing it in more locations. But I think any approach here ends up with some duplication of logic due to overlap in what's going on.

And my mistake in failing to drop the conflict marker fully didn't help.

PR #3443, opened 2023-12-01
Distribute autoupdate across threads.

zygoloid:
Also switch how we ensure that stdin is closed for tests, so that `bazel run` doesn't hang if invoked manually.

### DIFF
--- a/testing/file_test/file_test_base.cpp
+++ b/testing/file_test/file_test_base.cpp
@@ -728,7 +728,7 @@ static auto Main(int argc, char** argv) -> int {
   // Tests might try to read from stdin. Ensure those reads fail by closing
   // stdin and reopening it as /dev/null. Note that STDIN_FILENO doesn't exist
   // on Windows, but POSIX requires it to be 0.
-  llvm::sys::Process::SafelyCloseFileDescriptor(/*STDIN_FILENO*/ 0);
+  llvm::sys::Process::SafelyCloseFileDescriptor(0);
   llvm::sys::Process::FixupStandardFileDescriptors();
 
   llvm::SmallVector<std::string> tests = GetTests();
--- a/testing/file_test/file_test_base.cpp
+++ b/testing/file_test/file_test_base.cpp
@@ -6,8 +6,6 @@
 
 #include <filesystem>
 #include <fstream>
-#include <future>
-#include <mutex>
 #include <optional>
 #include <string>
 #include <utility>
@@ -23,6 +21,7 @@
 #include "llvm/Support/MemoryBuffer.h"
 #include "llvm/Support/PrettyStackTrace.h"
 #include "llvm/Support/Process.h"
+#include "llvm/Support/ThreadPool.h"
 #include "testing/file_test/autoupdate.h"
 
 ABSL_FLAG(std::vector<std::string>, file_tests, {},
@@ -33,8 +32,9 @@ ABSL_FLAG(std::string, test_targets_file, "",
 ABSL_FLAG(bool, autoupdate, false,
           "Instead of verifying files match test output, autoupdate files "
           "based on test output.");
-ABSL_FLAG(int, threads, 16,
-          "Number of threads to use when autoupdating tests.");
+ABSL_FLAG(unsigned int, threads, 0,
+          "Number of threads to use when autoupdating tests, or 0 to "
+          "automatically determine a thread count.");
 
 namespace Carbon::Testing {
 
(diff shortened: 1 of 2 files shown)


testing/file_test/file_test_base.cpp
```diff
@@ -533,14 +539,50 @@ static auto Main(int argc, char** argv) -> int {
...
+  // TODO: This won't work on Windows.
+  int dev_null = open("/dev/null", O_RDONLY);
+  if (dev_null == -1) {
+    llvm::errs() << "Failed to open /dev/null.\n";
+    return EXIT_FAILURE;
+  }
+  if (dup2(dev_null, 0) != 0) {
+    llvm::errs() << "Failed to override stdin.\n";
+    return EXIT_FAILURE;
+  }
```
jonmeow: Had you considered instead using `llvm::nulls` to replace the driver's stdin? Right now, it seems like there's a tradeoff where `bazel run` on a test which reads from stdin would wait on stdin: but such tests are rare, and doesn't `bazel run` hook up stdin? The trade-off of removing a cross-platform approach for incremental convenience with the two stdin tests seems difficult to me, but maybe we don't need to make that compromise.

zygoloid: Using `llvm::nulls` doesn't work -- the driver is using `llvm::MemoryBuffer::getSTDIN`, which directly reads from file descriptor 0. (And we can't even mock that out or similar, because part of what we're trying to test is that we actually read from file descriptor 0 when given `-` as the input file.)

I've changed this to support Windows. I don't have any way to test it, but according to the documentation this should work. *shrug*

We could move the stdin tests off to a separate mechanism -- maybe an `sh_test` -- if you prefer. Or I can just undo this part of the change and leave `bazel run` on these testing targets broken. (`bazel run` seems to run the test in the background, so ^D at the terminal doesn't help, and bazel eats ^Cs so they don't help either; you need to find and `kill` the test wrapper script to unwedge a `bazel run` of the driver tests.)

zygoloid: Found a way to do this in the LLVM API. (Still not tested on Windows, but it looks plausible.)

testing/file_test/file_test_base.cpp
```diff
@@ -720,18 +725,32 @@ static auto Main(int argc, char** argv) -> int {
     return EXIT_FAILURE;
   }
 
+  // Tests might try to read from stdin. Ensure those reads fail by closing
+  // stdin and reopening it as /dev/null.
+  llvm::sys::Process::SafelyCloseFileDescriptor(0);
```
jonmeow, line 731:
```suggestion
  llvm::sys::Process::SafelyCloseFileDescriptor(STDIN_FILENO);
```

jonmeow (approved):
Thanks! That approach looks good to me.

testing/file_test/file_test_base.cpp
zygoloid: That's a POSIX extension that doesn't exist on Windows.

jonmeow: Hmmm, maybe I need to object more.

These are noops on Windows:
https://github.com/llvm/llvm-project/blob/main/llvm/lib/Support/Windows/Process.inc#L273-L281

testing/file_test/file_test_base.cpp
```diff
@@ -720,18 +725,33 @@ static auto Main(int argc, char** argv) -> int {
     return EXIT_FAILURE;
   }
 
+  // Tests might try to read from stdin. Ensure those reads fail by closing
+  // stdin and reopening it as /dev/null. Note that STDIN_FILENO doesn't exist
+  // on Windows, but POSIX requires it to be 0.
+  llvm::sys::Process::SafelyCloseFileDescriptor(/*STDIN_FILENO*/ 0);
```
jonmeow:
```suggestion
  llvm::sys::Process::SafelyCloseFileDescriptor(0);
```

Comment syntax isn't correct for an argument comment.

testing/file_test/file_test_base.cpp
jonmeow: Well, fixup is anyway. Maybe it's okay.

josh11b:
FYI, there seems to be a race condition. The autoupdate script hung on exit for me:
```
~/carbon/c/extend/toolchain$ ./autoupdate_testdata.py                      4ms
Starting local Bazel server and connecting to it...
INFO: Build completed successfully, 80 total actions
Executing tests from //toolchain/testing:file_test
...........................................................!.........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................


^CTraceback (most recent call last):
  File "/usr/local/google/home/joshl/carbon/c/extend/toolchain/./autoupdate_testdata.py", line 50, in <module>
    main()
  File "/usr/local/google/home/joshl/carbon/c/extend/toolchain/./autoupdate_testdata.py", line 46, in main
    subprocess.run(argv, check=True)
  File "/usr/local/google/home/joshl/Cellar/python@3.11/3.11.3/lib/python3.11/subprocess.py", line 550, in run
    stdout, stderr = process.communicate(input, timeout=timeout)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/google/home/joshl/Cellar/python@3.11/3.11.3/lib/python3.11/subprocess.py", line 1199, in communicate
    self.wait()
  File "/usr/local/google/home/joshl/Cellar/python@3.11/3.11.3/lib/python3.11/subprocess.py", line 1262, in wait
    return self._wait(timeout=timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/google/home/joshl/Cellar/python@3.11/3.11.3/lib/python3.11/subprocess.py", line 2013, in _wait
    (pid, sts) = self._try_wait(0)
                 ^^^^^^^^^^^^^^^^^
  File "/usr/local/google/home/joshl/Cellar/python@3.11/3.11.3/lib/python3.11/subprocess.py", line 1971, in _try_wait
    (pid, sts) = os.waitpid(self.pid, wait_flags)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyboardInterrupt

```

zygoloid:
> FYI, there seems to be a race condition.

See #3449 (I can't be certain this is the same, but that's the only one I've seen).

PR #3444, opened 2023-12-01
Clarify naming around paren expressions

geoffromer:
(Split out of #3410)

Consistently use `ParenExpr` solely for parenthesized single expressions, and use more syntax-oriented terminology for states and nodes that might represent either a `ParenExpr` or a tuple literal.


toolchain/parse/state.def
```diff
@@ -697,45 +697,67 @@ CARBON_PARSE_STATE_VARIANTS2(ParenCondition, If, While)
...
+//       ^
+//   (state done)
+//
+// ( ... , ... )
+//       ^
+//   1. Expression
+//   2. TupleLiteralElementFinish
+//
+// ( ... )
+//      ^
```
jonmeow, line 779: Can this occur on TupleLiteralElementFinish? Should it be removed or replaced with something like `( ... , ... )` pointing to the close `)`?

jonmeow (approved):
Thanks for splitting this out!

toolchain/parse/state.def
```diff
@@ -402,7 +402,7 @@ CARBON_PARSE_STATE(Expr)
 //
 //  (
 // ^
-//   1. ParenExpr
+//   1. OpenParen
```
zygoloid: I don't see a definition of the `OpenParen` state here. Should this be `ParenExpr`?

toolchain/parse/state.def
```diff
@@ -697,45 +697,67 @@ CARBON_PARSE_STATE_VARIANTS2(ParenCondition, If, While)
...
-//   1. ParenExprFinishAsTuple
+//   1. TupleLiteralFinish
 //
 // ( ... )
 // ^
 //   1. Expr
-//   2. ParenExprParamFinishAsUnknown
-//   3. ParenExprFinishAsNormal    (SPECIAL: may be replaced)
+//   2. ExprAfterOpenParenFinish
+//   3. ParenExprFinish    (SPECIAL: may be replaced)
```
zygoloid, line 738: It looks like we don't need these SPECIAL actions any more: instead, we could:

- Not push the ParenExprFinish state here
- Push either a TupleLiteralFinish state or a ParenExprFinish state in ExprAfterOpenParenFinish, depending on whether we're in the tuple case or not.

geoffromer: As discussed in person yesterday, this doesn't work: we need two separate states to keep track of two separate locations (the open paren and the start of the expression inside it), unless we want to let the parser look back at already-parsed nodes for that information.

I tried revising the code to treat the two physical states as a single "logical" state that gets pushed and popped together, but that's not a straightforward change because `ConsumeListToken` expects to find the comma's parent at the top of the stack.

toolchain/parse/state.def
geoffromer: You may be right that at the moment the "..." always contains a comma, but that's not an invariant that `HandleTupleLiteralElementFinish` either enforces or relies on; it's just a contingent fact about how this state is currently used. In particular, if we chose to reimplement `CallExpr` in terms of `TupleLiteral`, that would no longer be true, and as far as I know, nothing about `TupleLiteralElementFinish` itself would need to change to accommodate that. As a result, I don't see much benefit to narrowing its contract in that way.

toolchain/parse/state.def
zygoloid: OK, let's drop this comment then.

toolchain/parse/handle_paren_expr.cpp
```diff
@@ -38,15 +36,13 @@ static auto HandleParenExprParamFinish(Context& context, bool as_tuple)
...
+  state.state = State::TupleLiteralElementFinish;
+
+  auto finish_state = context.PopState();
+  CARBON_CHECK(finish_state.state == State::ParenExprFinish)
+      << "Unexpected parent state, found: " << finish_state.state;
+  finish_state.state = State::TupleLiteralFinish;
+  context.PushState(finish_state);
 
   // On a comma, push another expression handler.
   if (list_token_kind == Context::ListTokenKind::Comma) {
```
zygoloid: I found this confusing, given that we always have a comma in this case.
```suggestion
  // If the close paren is not next, push another expression handler.
  if (list_token_kind != Context::ListTokenKind::CommaClose) {
```

toolchain/parse/handle_paren_expr.cpp
```diff
@@ -38,15 +36,13 @@ static auto HandleParenExprParamFinish(Context& context, bool as_tuple)
...
-  if (!as_tuple) {
-    state.state = State::ParenExprParamFinishAsTuple;
-
-    auto finish_state = context.PopState();
-    CARBON_CHECK(finish_state.state == State::ParenExprFinishAsNormal)
-        << "Unexpected parent state, found: " << finish_state.state;
-    finish_state.state = State::ParenExprFinishAsTuple;
-    context.PushState(finish_state);
-  }
+  state.state = State::TupleLiteralElementFinish;
```
zygoloid: Sink this down into the `if` below.

zygoloid (approved):
A couple of tiny comments, but then LGTM.

toolchain/parse/handle_paren_expr.cpp
geoffromer: Done.

toolchain/parse/handle_paren_expr.cpp
geoffromer: Done, but with some other comment changes in light of your other comment.

zygoloid (approved):
Looks good, not merging due to conflicts.

PR #3445, opened 2023-12-01
Add support for 'package' in expressions.

jonmeow:
This creates a namespace for `package` scope.

It looks like names of class_decls in namespaces lead to an unexpected instref. This is already true, as best as I can tell. I'm not sure if there's a preferred approach to address that, so I've left a TODO for now.

### DIFF
--- a/toolchain/sem_ir/formatter.cpp
+++ b/toolchain/sem_ir/formatter.cpp
@@ -146,9 +146,6 @@ class InstNamer {
     auto& [inst_scope, inst_name] = insts[inst_id.index];
     if (!inst_name) {
       // This should not happen in valid IR.
-      // TODO: This is encountered for names in namespaces: the namespace is
-      // declared before the name, and then the formatter prints the name scope
-      // for the name before it's been added.
       std::string str;
       llvm::raw_string_ostream(str) << "<unexpected instref " << inst_id << ">";
       return str;
--- a/toolchain/sem_ir/formatter.cpp
+++ b/toolchain/sem_ir/formatter.cpp
@@ -139,9 +139,16 @@ class InstNamer {
       return BuiltinKind::FromInt(inst_id.index).label().str();
     }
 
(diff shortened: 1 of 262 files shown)


toolchain/sem_ir/formatter.cpp
```diff
@@ -139,9 +139,16 @@ class InstNamer {
...
+    if (inst_id == InstId::PackageNamespace) {
+      return "package";
+    }
+
     auto& [inst_scope, inst_name] = insts[inst_id.index];
     if (!inst_name) {
       // This should not happen in valid IR.
+      // TODO: This is encountered for names in namespaces: the namespace is
+      // declared before the name, and then the formatter prints the name scope
+      // for the name before it's been added.
```
zygoloid: (Don't forget to remove this TODO now #3447 has landed.)

zygoloid (approved):
Yeah, this makes sense to me and the change looks good (needs to be merged with #3447).

Aside: I am starting to wonder if it would be better to rename SemIR's Namespaces to something like NamespaceFragment, given that the actual namespace is the union of a bunch of Namespaces in different Files. But definitely not in this PR :)

jonmeow:
> Aside: I am starting to wonder if it would be better to rename SemIR's Namespaces to something like NamespaceFragment, given that the actual namespace is the union of a bunch of Namespaces in different Files. But definitely not in this PR :)

I'd lean towards keeping `Namespace` with the understanding that they're composeable. `NamespaceFragment` is long and I also initially read it as "the namespace fragment in a qualified name" (before reading your comment more carefully). But I think too that, from the IR's perspective, the `Namespace` will try to present the composed view rather than a single fragment. Maybe other options will present themselves while implementing cross-IR references.

toolchain/sem_ir/formatter.cpp
jonmeow: Done, thanks for the reminder.

PR #3447, opened 2023-12-02
Associate a name with ClassDecl

jonmeow:
Note generics will probably need different behavior.

### DIFF
--- a/toolchain/sem_ir/formatter.cpp
+++ b/toolchain/sem_ir/formatter.cpp
@@ -439,7 +439,8 @@ class InstNamer {
         }
         case ClassDecl::Kind: {
           add_inst_name_id(
-              sem_ir_.classes().Get(inst.As<ClassDecl>().class_id).name_id);
+              sem_ir_.classes().Get(inst.As<ClassDecl>().class_id).name_id,
+              ".decl");
           continue;
         }
         case ClassType::Kind: {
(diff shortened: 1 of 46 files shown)


zygoloid (commented):
Please also add a `FormatInstructionLHS` overload  for `ClassDecl` to print out the name next to the `class_decl` instruction.

jonmeow:
> Please also add a `FormatInstructionLHS` overload for `ClassDecl` to print out the name next to the `class_decl` instruction.

Done, though please confirm you're okay with the format I've chosen. It felt type-ish so I headed in that direction.

toolchain/sem_ir/formatter.cpp
```diff
@@ -710,6 +716,12 @@ class Formatter {
     }
   }
 
+  // Print ClassDecl with type-like semantics even though it lacks a type_id.
+  auto FormatInstructionLHS(InstId inst_id, ClassDecl /*inst*/) -> void {
+    FormatInstName(inst_id);
+    out_ << ": type = ";
```
zygoloid:
```suggestion
    out_ << " = ";
```

Let's leave off the type because this instruction doesn't have one. (This is what we used to do for instructions that had values but not types, back before that functionality was removed.)

zygoloid (approved):

PR #3448, opened 2023-12-02
Parse support for tuple patterns in `var` and `let`

geoffromer:


toolchain/check/handle_binding_pattern.cpp
```diff
@@ -10,6 +10,7 @@
 namespace Carbon::Check {
 
 auto HandleAddress(Context& context, Parse::NodeId parse_node) -> bool {
+  // TODO: diagnose if this occurs outside an implicit parameter list
```
zygoloid: I think we already do, implicitly: we only allow `self` in an implicit parameter list (line 43-49), and we only allow `addr` to be applied to `self` (ilne 20-24).
```suggestion
```
See `check/testdata/class/fail_{,not_}self.carbon`, which test `addr` applied to non-`self` and `self` not in the implicit parameter list (though not both at once).

toolchain/check/handle_let.cpp
```diff
@@ -10,6 +10,10 @@ namespace Carbon::Check {
 
 auto HandleLetDecl(Context& context, Parse::NodeId parse_node) -> bool {
   auto value_id = context.node_stack().PopExpr();
+  if (context.parse_tree().node_kind(context.node_stack().PeekParseNode()) ==
+      Parse::NodeKind::ParamList) {
```
zygoloid:
```suggestion
  if (context.node_stack().PeekIs<Parse::NodeKind::ParamList>()) {
```
... although it would probably be more future-proof to use `PopIf<BindingPattern>` below instead, and reject all non-`BindingPattern` cases.

toolchain/check/handle_variable.cpp
```diff
@@ -32,15 +32,23 @@ auto HandleVariableInitializer(Context& context, Parse::NodeId parse_node)
...
   auto init_id = SemIR::InstId::Invalid;
-  bool has_init =
-      context.parse_tree().node_kind(context.node_stack().PeekParseNode()) !=
-      Parse::NodeKind::BindingPattern;
+  Parse::NodeKind next_kind =
+      context.parse_tree().node_kind(context.node_stack().PeekParseNode());
+  if (next_kind == Parse::NodeKind::ParamList) {
+    return context.TODO(parse_node, "tuple pattern in var");
+  }
+  bool has_init = next_kind != Parse::NodeKind::BindingPattern;
```
zygoloid: Please add a TODO to find a more robust way to determine if there was an initializer.

toolchain/parse/handle_binding_pattern.cpp
```diff
@@ -6,46 +6,22 @@
...
-    context.ConsumeIfBindingPatternKeyword(Lex::TokenKind::Template,
-                                           State::BindingPatternTemplate,
-                                           state.subtree_start);
-    context.ConsumeIfBindingPatternKeyword(Lex::TokenKind::Addr,
-                                           State::BindingPatternAddress,
-                                           state.subtree_start);
-  }
+  context.ConsumeIfBindingPatternKeyword(Lex::TokenKind::Template,
+                                         State::BindingPatternTemplate,
+                                         state.subtree_start);
```
zygoloid: Is there a reason that `ConsumeIfBindingPatternKeyword` needs to be a member of `Context`? It looks like it could be defined locally to this file, which is what I think we would generally prefer to do, to keep the interface of `Context` small and keep related functionality together.

toolchain/parse/handle_var.cpp
```diff
@@ -23,7 +23,11 @@ static auto HandleVar(Context& context, State finish_state,
     context.AddLeafNode(NodeKind::ReturnedModifier, returned_token);
   }
 
-  context.PushState(State::BindingPatternAsVariable);
+  if (context.PositionKind() == Lex::TokenKind::OpenParen) {
+    context.PushState(State::ParamListAsRegular);
+  } else {
+    context.PushState(State::BindingPattern);
+  }
```
zygoloid: I wonder if it's worth factoring out either a function or a state to do this. `State::Pattern`?

toolchain/parse/handle_binding_pattern.cpp
geoffromer: I agree, but actually I don't think this needs to be a function at all; I'm having a hard time coming up with a documentation comment (or even a name) that seems accurate but doesn't basically spell out the code in words, so I've inlined the function body into the two callsites.

toolchain/parse/handle_var.cpp
geoffromer: Good idea; done.

toolchain/check/handle_variable.cpp
```diff
@@ -32,15 +32,24 @@ auto HandleVariableInitializer(Context& context, Parse::NodeId parse_node)
...
+  // TODO: find a more robust way to determine if there was an initializer.
+  bool has_init = next_kind != Parse::NodeKind::BindingPattern;
   if (has_init) {
     init_id = context.node_stack().PopExpr();
     context.node_stack()
         .PopAndDiscardSoloParseNode<Parse::NodeKind::VariableInitializer>();
   }
 
+  if (context.parse_tree().node_kind(context.node_stack().PeekParseNode()) ==
+      Parse::NodeKind::ParamList) {
```
zygoloid:
```suggestion
  if (context.node_stack().PeekIs<Parse::NodeKind::ParamList>()) {
```

zygoloid (approved):

PR #3449, opened 2023-12-02
Move `llvm::Initialize*` calls to `main`.

zygoloid:
Per their documentation, the `llvm::Initialize*` functions are only supposed to be called by the main program, not by a library like toolchain/codegen. Fixes a hang due to a data race in multithreaded autoupdate.

Add a utility class `Carbon::InitLLVM` to do the common LLVM initialization shared by all Carbon tools, optionally including initializing the LLVM targets. Because the LLVM targets add a lot of binary size, only initialize them for binaries that opt in by depending on a new target `//common:all_llvm_targets`.

Also fix `//explorer:file_test` and `//explorer:file_test.trace` to share a binary rather than linking an identical binary twice.


chandlerc (approved):
LG

chandlerc (approved):
(still good, hoping it passes tests)

testing/base/gtest_main.cpp
```diff
@@ -13,5 +14,10 @@ auto main(int argc, char** argv) -> int {
       "https://github.com/carbon-language/carbon-lang/issues and include the "
       "crash backtrace.\n");
   llvm::InitLLVM init_llvm(argc, argv);
+  llvm::InitializeAllTargetInfos();
+  llvm::InitializeAllTargets();
+  llvm::InitializeAllTargetMCs();
+  llvm::InitializeAllAsmParsers();
+  llvm::InitializeAllAsmPrinters();
```
jonmeow: You might consider a Lower::Initialize call, to reduce the duplication.

zygoloid: I have a pending follow-up cleanup to do that and more, just trying to get this landed first.

zygoloid: Merged that in here, since I was making other substantial changes.

common/init_llvm.h
```diff
@@ -0,0 +1,28 @@
...
+#define CARBON_COMMON_INIT_LLVM_H_
+
+#include "llvm/Support/InitLLVM.h"
+
+namespace Carbon {
+
+// A RAII class to handle initializing LLVM and shutting it down. An instance of
+// this class should be created in the `main` function of each Carbon binary
+// that interacts with LLVM, before `argc` and `argv` are first inspected.
+class InitLLVM : public llvm::InitLLVM {
```
jonmeow: Why inheriting from llvm::InitLLVM instead of as a member? Granted it doesn't really have an API that's exposed, but it still seems odd to me since the class doesn't feel like a base class.

explorer/BUILD
```diff
@@ -86,9 +153,22 @@ file_test(
...
+        name = "%s_%s" % (test_dir, test_target),
+        size = "small",
+        srcs = ["explorer_tests"],
+        args = [
+            "--no_file_tests",
+            "--gtest_filter=" + test_suite + ".*",
+            "--fail_if_no_tests_run",
+        ],
+    )
+    for (test_dir, test_target, test_suite) in EXPLORER_GTEST_TESTS
```
chandlerc: Instead of a list comprehension and such, maybe it makes sense to write a small starlark macro for this?

I also wonder about just having a single test target and using test sharding?

explorer/ast/ast_test_matchers_test.cpp
```diff
@@ -23,7 +23,7 @@ using ::testing::Not;
 
 static constexpr SourceLocation DummyLoc("dummy", 0, FileKind::Main);
 
-TEST(BlockContentsAreTest, BasicUsage) {
+TEST(ASTTestMatchers, BlockContentsAreTest) {
```
chandlerc: Might make sense to split this off as a trivial improvement to the test you can easily land first?

explorer/main.cpp
```diff
@@ -32,11 +32,7 @@ namespace path = llvm::sys::path;
 
 auto ExplorerMain(int argc, char** argv, void* static_for_main_addr,
                   llvm::StringRef relative_prelude_path) -> int {
-  llvm::setBugReportMsg(
-      "Please report issues to "
-      "https://github.com/carbon-language/carbon-lang/issues and include the "
-      "crash backtrace.\n");
-  llvm::InitLLVM init_llvm(argc, argv);
+  Carbon::InitLLVM init_llvm(argc, argv);
```
chandlerc: I think I missed this the first time through -- we shouldn't need to initialize all the targets and such of LLVM in the explorer, should we? Maybe I'm confused, but I'd be very surprised if that were needed. And maybe that makes the explorer side of this much smaller / easier?

explorer/BUILD
```diff
@@ -86,9 +153,22 @@ file_test(
...
 
+[
+    sh_test(
+        name = "%s_%s" % (test_dir, test_target),
+        size = "small",
+        srcs = ["explorer_tests"],
+        args = [
+            "--no_file_tests",
+            "--gtest_filter=" + test_suite + ".*",
+            "--fail_if_no_tests_run",
```
jonmeow: Why do have two flags for file tests, versus just --gtest_filter?

That is, I don't understand why --no_file_tests is needed. Also, --fail_if_no_tests_run is only used in combination with both --no_file_tests and --gtest_filter: perhaps instead, if --gtest_filter results in no tests to run then it's an error? (perhaps a generally useful safety check that busyboxed tests still run something?)

chandlerc (commented):
(just a few quick comments -- sorry I've been a bit too jetlagged to help much with reviews recently)

explorer/BUILD
jonmeow: Regarding a single test target: file tests already use maximum sharding (50 shards total). So non-file-test tests would easily get mixed in and harder to directly execute. Trace versus non-trace issues would also be more difficult to see -- that's why there's already separation there. You could also end up with some flakiness because trace tests are slower to execute and may end up imbalanced, forcing test size from small to medium.

explorer/BUILD
```diff
@@ -42,8 +43,72 @@ cc_binary(
...
+    (
+        "syntax",
+        "parse_test",
+        "ParseTest",
+    ),
+    (
+        "syntax",
+        "unimplemented_example_test",
+        "UnimplementedExampleTest",
+    ),
```
jonmeow: Versus the --gtest_filter approach, had you considered replacing the built-in googletest main with something that'd dispatch based on the target name? That'd remove the need to add test names to the list, it could just do direct dependencies.

e.g., with a registration hook like `if (getenv(TEST_TARGET) == my_target && absl::GetFlag(gtest_filter).empty()) { absl::SetFlag(gtest_filter, my_tests); }` to set within the binary, rather than in the BUILD. My thought there is keeping the test unit names next to the code.

If you combined that with a .bzl rule that combined the per-cc_library with a sh_test that depended back on the main busy box binary, you might even be able to place such a list straight into `deps`.

explorer/parse_and_execute/BUILD
```diff
@@ -25,13 +25,14 @@ cc_library(
     ],
 )
 
-cc_test(
+# This is included in //explorer:explorer_tests.
```
jonmeow: Why the comment here but not elsewhere? I think consistency in either direction would be fine, but right now it seems the odd one out.

common/init_llvm.h
```diff
@@ -0,0 +1,28 @@
...
+ public:
+  // Initializes LLVM for use by a Carbon binary. On Windows, `argc` and `argv`
+  // are updated to refer to properly-encoded UTF-8 versions of the command line
+  // arguments.
+  InitLLVM(int& argc, char**& argv);
```
jonmeow: Reminding myself as I jot style notes, should this be explicit?

```suggestion
  explicit InitLLVM(int& argc, char**& argv);
```

common/init_llvm.h
zygoloid: For me it felt a bit more natural to implement one RAII type in terms of another this way, but I'm happy either way around. Changed to a data member.

explorer/main.cpp
zygoloid: Done, though I couldn't find a non-ugly and well-factored way to initialize LLVM only where we use it -- the place where we make the decision is often not the place where `main` is defined! Hopefully this `alwayslink` approach isn't too awful.

explorer/BUILD
zygoloid: (Historical, since I've removed this again, but: this is a limitation of the absl Flags library. You can't nicely have an optional flag that distinguishes between the flag being absent and being present with an empty value, because the library determines whether the flag is present by checking whether its string value differs from the default. There are workarounds like setting a default of "none" rather than an actually empty default, but that seemed too ugly.)

explorer/BUILD
zygoloid: (Comment no longer applies to latest version of this PR.)

zygoloid (commented):
Revised substantially to avoid linking the LLVM targets into anywhere that doesn't actually need them. This results in this change being a net reduction in the size of our build area (down 15-20%), rather than a net increase.

explorer/BUILD
zygoloid: (Resolving as this no longer applies to this PR.)

common/init_llvm.h
```diff
@@ -24,6 +25,7 @@ class InitLLVM {
 
  private:
   llvm::InitLLVM init_llvm;
+  llvm::SmallVector<char*> args;
```
jonmeow:
```suggestion
  llvm::InitLLVM init_llvm_;
  llvm::SmallVector<char*> args_;
```

common/init_llvm.cpp
```diff
@@ -8,16 +8,27 @@
...
+[[gnu::weak]] extern auto InitLLVMTargets() -> void;
+
+InitLLVM::InitLLVM(int& argc, char**& argv)
+    : init_llvm(argc, argv), args(argv, argv + argc) {
+  // LLVM assumes that argc and argv won't change, and registers them with an
+  // `llvm::PrettyStackTraceProgram` that will crash if an argv element gets
+  // nulled out, which for example `testing::InitGoogleTest` does. So make a
+  // copy of argv for use by the program to satisfy LLVM's assumptions.
+  argc = args.size();
+  argv = args.data();
```
jonmeow: Should this actually be handing `init_llvm` its own values, rather than passing a reference?

I'm asking because these changes to `argc` and `argv` looked odd to me: it's changing the values after they're passed to `init_llvm`. That seems inconsistent with the comments above.

Won't this undo the Windows-specific changes that `init_llvm` does? If not, does absl do the utf8 translation, potentially leading to both trying to do a conflicting transform?

Does it result in googletest still being able to modify the argc/argv seen by stack tracing?

common/init_llvm.h
```diff
@@ -0,0 +1,36 @@
...
+
+  // Shuts down LLVM.
+  ~InitLLVM() = default;
+
+ private:
+  llvm::InitLLVM init_llvm;
+  llvm::SmallVector<char*> args;
+
+  static auto (*InitializeTargets)() -> void;
+  static char RegisterTargets;
```
jonmeow: Style: Functions before variables.

jonmeow: Can you comment the semantics of these? I'm just not familiar with the pattern, so it'd help me.

common/all_llvm_targets.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+static auto InitLLVMTargets() -> void {
+  llvm::InitializeAllTargetInfos();
+  llvm::InitializeAllTargets();
+  llvm::InitializeAllTargetMCs();
+  llvm::InitializeAllAsmParsers();
+  llvm::InitializeAllAsmPrinters();
+}
+
+char InitLLVM::RegisterTargets = (InitializeTargets = &InitLLVMTargets, 0);
```
jonmeow, line 20: This is probably just my naivette, but why:

- Store `RegisterTargets` as a member of `InitLLVM` versus something local to this file?
- Initialize using a comma expression instead of something like `auto Register = InitializeTargets = &InitLLVMTargets;`?

Mild thought, had you considered something like two different targets, one for init-with-all-targets, one for init-without-all-targets, resulting in something like a cpp file that only contains `auto (*InitLLVM::InitializeTargets)() -> void = nullptr;` versus `auto (*InitLLVM::InitializeTargets)() -> void = &InitLLVMTargets;` here? i.e., creating `:init_llvm_with_targets` and `:init_llvm_without_targets`, never directly depending on `:init_llvm`.

jonmeow (commented):
Oh, so file_test was the main cost on explorer? Makes sense as I think about it.

common/init_llvm.cpp
zygoloid: > Should this actually be handing `init_llvm` its own values, rather than passing a reference?

We could do that, but it's awkward because we'd need to create some extra state to hold them before calling the constructor of the field. I don't think it's worth it.

> Won't this undo the Windows-specific changes that `init_llvm` does?

`args` is copied out of `argv, argv+argc` after the `init_llvm` constructor "fixes" the `argc` and `argv` values, so this is preserving the things that `llvm::InitLLVM` did, not undoing them.

> If not, does absl do the utf8 translation, potentially leading to both trying to do a conflicting transform?

`argc` and `argv` are already broken before you reach `main` on Windows because they get converted from Unicode into the ANSI code page, not into UTF-8. There's no way to fix them by transcoding to UTF-8, because a lossy translation was already performed. `llvm::InitLLVM` throws away `argc` and `argv` on Windows and recomputes them based on the original Unicode command line.

The absl API is required to process the `argc` and `argv` you give it, and not go off and work out its own ones (because the things you pass in might not be the original command line), so if we want to give it UTF-8 arguments, we need to form them before calling into absl's flag parser, which LLVM is now doing for us.

> Does it result in googletest still being able to modify the argc/argv seen by stack tracing?

It does not, but I think that is a bug fix: we want the pretty stack trace entry to show the original command line arguments given to the program, not the result of them being modified as gtest eats the arguments that it knows about.

jonmeow: Thanks for the explanation.

Do you want suggestions for changes to the comments, or will you take another look at those?

> No, which is a bug fix: we want the pretty stack trace entry to show the original command line arguments given to the program, not the result of them being modified as gtest eats the arguments that it knows about.

The argv now points at the same address that's later passed on to googletest, doesn't it? Oh, I see now -- it's making a copy and passing that back to googletest. Can you please clarify what is meant by "the program", for example? Also, that copy is already made, and it's copying llvm's processed arguments instead of the actual argc/argv?

Maybe the "make a copy" comments belong on `args`, rather than these two lines?

common/all_llvm_targets.cpp
zygoloid: > This is probably just my naivette, but why:
> 
> * Store `RegisterTargets` as a member of `InitLLVM` versus something local to this file?

So that it has access to the private member `InitLLVM::InitializeTargets`. That seems like it at least deserves a comment :)

> * Initialize using a comma expression instead of something like `auto Register = InitializeTargets = &InitLLVMTargets;`?

Just a trivial thing: that would make the type of `Register` be a function pointer, which would be a bit uglier, take up a bit more storage, and create a temptation to use `InitLLVM::RegisterTargets` instead of `InitLLVM::InitializeTargets` (given that they would store the same value).

> Mild thought, had you considered something like two different targets, one for init-with-all-targets, one for init-without-all-targets, resulting in something like a cpp file that only contains `auto (*InitLLVM::InitializeTargets)() -> void = nullptr;` versus `auto (*InitLLVM::InitializeTargets)() -> void = &InitLLVMTargets;` here? i.e., creating `:init_llvm_with_targets` and `:init_llvm_without_targets`, never directly depending on `:init_llvm`.

I did think about that. The consequence would be that every binary and test (transitively) using `//common:init_llvm` would need to link against one or the other, or get a link-time error. I'm fine with that, but it's a tradeoff -- do we want to burden (essentially) every `cc_binary` and `cc_test` rule with a dependency that says "don't link in the LLVM targets"? Maybe we should; it'll be clearer and more explicit that way, if more verbose.

I'll give that a go and see just how many extra deps I need to add. I think that would also address your other comments here (we can get rid of this registration machinery and function pointer shenanigans entirely, and just have two definitions of an `InitLLVM::InitializeTargets()` function).

jonmeow: I'm realizing the latter part may depends on where init_llvm is transitively used, versus directly. That is, a transitive use could actually result in a dependency conflict under the conflicting target model.

common/init_llvm.h
jonmeow: Oh, these are both variables.

jonmeow (approved):
Approving -- I suspect the current init_llvm dep model is correct. Comment changes to InitLLVM would still be appreciated though.

common/init_llvm.cpp
zygoloid: Comments updated.

common/init_llvm.h
zygoloid: Added comments.

common/all_llvm_targets.cpp
zygoloid: Added comments.

PR #3450, opened 2023-12-02
Support for `base: T;`, `.base`, `x.base`.

zygoloid:
No support for `extend base` yet, in an effort to minimize collisions with #3412.


toolchain/sem_ir/typed_insts.h
```diff
@@ -86,6 +86,18 @@ struct Assign {
   InstId rhs_id;
 };
 
+// A base in a class, of the form `base: base_type;`. A base class is an
+// element of the derived class, and the type of the `Base` instruction is an
+// `UnboundElementType`.
+struct Base {
```
jonmeow, line 92: `base class` also being a thing, and "base" just being a pretty common thing for type names anyway (e.g., `TestBase`), maybe something with another word to make usage clearer, such as `BaseElement` or `BaseValue`?

toolchain/parse/handle_class.cpp
```diff
@@ -0,0 +1,32 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "toolchain/parse/context.h"
+
+namespace Carbon::Parse {
+
+// Processing of the `class` declaration and definition themselves is shared
+// with interfaces and constraints, and is located in handle_type.cpp.
+
+// Handles processing of a complete `base: B` declaration.
```
jonmeow: Is this the file split that we want? I wonder if this could consolidate in one direction or the other:

- If there isn't much more class-specific handling, move this to handle_type or handle_base.
  - I'm not sure how much more you plan to put here, and lack an instinct myself.
- Shift handle_type to be more of a header, put more class logic in this file.

toolchain/parse/handle_decl_scope_loop.cpp
```diff
@@ -62,8 +62,20 @@ auto HandleDeclScopeLoop(Context& context) -> void {
...
+      if (context.PositionIs(Lex::TokenKind::Colon, Lookahead::NextToken)) {
+        context.PushState(State::BaseDecl);
+        context.PushState(State::Expr);
+        context.AddLeafNode(NodeKind::BaseIntroducer, context.Consume());
+        context.AddLeafNode(NodeKind::BaseColon, context.Consume());
+        return;
+      }
+      // TODO: If the next token isn't a colon or `class`, try to recover based
+      // on whether we're in a class, whether we have an `extend` modifier, and
+      // the following tokens.
```
jonmeow, line 75: While I see you have a check test for `base: A` in the file scope, I think parse should also test it.

But also, this TODO implies significant more code here: at what point should modifiers be their own helper function or state?

toolchain/check/handle_class.cpp
```diff
@@ -166,6 +167,98 @@ auto HandleClassDefinitionStart(Context& context, Parse::NodeId parse_node)
...
+    return true;
+  }
+
+  auto& class_info = context.classes().Get(*class_id);
+  if (class_info.base_id.is_valid()) {
+    CARBON_DIAGNOSTIC(
+        BaseRepeated, Error,
+        "Repeated `base` declaration. Multiple inheritance is not permitted.");
+    CARBON_DIAGNOSTIC(BasePrevious, Note,
+                      "Previous `base` declaration is here.");
```
jonmeow: Just to observe, we use several different phrasings for this type of note. Maybe we should start building a diagnostic style guide:

"Previously declared here."
https://github.com/carbon-language/carbon-lang/blob/1c4a35c76bbe436b7910e5ec059589595b0041c7/toolchain/check/handle_class.cpp#L68-L71

"Name is previously declared here."
https://github.com/carbon-language/carbon-lang/blob/1c4a35c76bbe436b7910e5ec059589595b0041c7/toolchain/check/context.cpp#L87C20-L87C56

"Previous definition was here." (maybe closest in phrasing, albeit "is" vs "was")
https://github.com/carbon-language/carbon-lang/blob/1c4a35c76bbe436b7910e5ec059589595b0041c7/toolchain/check/handle_class.cpp#L129-L132

"Field with the same name here."
https://github.com/carbon-language/carbon-lang/blob/1c4a35c76bbe436b7910e5ec059589595b0041c7/toolchain/check/handle_struct.cpp#L72C25-L72C57

(I've wondered similar for valid/allowed/permitted/legal)

toolchain/check/handle_class.cpp
```diff
@@ -166,6 +167,98 @@ auto HandleClassDefinitionStart(Context& context, Parse::NodeId parse_node)
...
+                      "`base` declaration can only be used in a class.");
+    context.emitter().Emit(parse_node, BaseOutsideClass);
+    return true;
+  }
+
+  auto& class_info = context.classes().Get(*class_id);
+  if (class_info.base_id.is_valid()) {
+    CARBON_DIAGNOSTIC(
+        BaseRepeated, Error,
+        "Repeated `base` declaration. Multiple inheritance is not permitted.");
```
jonmeow: "Repeated" to me implies duplicative, including the base class name. Maybe "Multiple `base` declarations found"?

Though you could also say something like "Cannot have multiple `base` declarations." or "Only one `base` declaration allowed."

```suggestion
        "Multiple `base` declarations found. Multiple inheritance is not permitted.");
```

toolchain/check/handle_class.cpp
```diff
@@ -166,6 +167,98 @@ auto HandleClassDefinitionStart(Context& context, Parse::NodeId parse_node)
...
+    context.emitter()
+        .Build(parse_node, BaseRepeated)
+        .Note(context.insts().Get(class_info.base_id).parse_node(),
+              BasePrevious)
+        .Emit();
+    return true;
+  }
+
+  auto base_type_id = ExprAsType(context, parse_node, base_type_expr_id);
+  if (!context.TryToCompleteType(base_type_id, [&] {
```
jonmeow, line 207: Just a thought, but maybe there should be some variant of this that sets base_type_id to error as needed (or return it as result instead of `bool`); it seems to be common for callers to want to do that, and might reduce the flow control at call sites. I see though that it can't _always_ do that, e.g. due to https://github.com/carbon-language/carbon-lang/blob/1c4a35c76bbe436b7910e5ec059589595b0041c7/toolchain/check/handle_function.cpp#L198

toolchain/check/handle_class.cpp
```diff
@@ -166,6 +167,98 @@ auto HandleClassDefinitionStart(Context& context, Parse::NodeId parse_node)
...
+        CARBON_DIAGNOSTIC(IncompleteTypeInBaseDecl, Error,
+                          "Deriving from incomplete type `{0}`.", std::string);
+        return context.emitter().Build(
+            parse_node, IncompleteTypeInBaseDecl,
+            context.sem_ir().StringifyType(base_type_id, true));
```
jonmeow, line 212: Please label the `true` here and below, I'd find it helpful trying to read the code.

toolchain/check/handle_class.cpp
```diff
@@ -166,6 +167,98 @@ auto HandleClassDefinitionStart(Context& context, Parse::NodeId parse_node)
...
+    auto base_class =
+        context.insts()
+            .Get(context.sem_ir().GetTypeAllowBuiltinTypes(base_type_id))
+            .TryAs<SemIR::ClassType>();
+    if (!base_class ||
+        context.classes().Get(base_class->class_id).inheritance_kind ==
+            SemIR::Class::Final) {
+      CARBON_DIAGNOSTIC(BaseIsFinal, Error,
+                        "Deriving from final type `{0}`. Only abstract and "
+                        "base classes can be derived from.",
```
jonmeow: Minor phrasing suggestion, reordering the "can" first.

```suggestion
                        "Deriving from final class `{0}`. Can only derive from `abstract` and `base` classes.",
```

toolchain/check/handle_class.cpp
```diff
@@ -166,6 +167,98 @@ auto HandleClassDefinitionStart(Context& context, Parse::NodeId parse_node)
...
+        .Note(context.insts().Get(class_info.base_id).parse_node(),
+              BasePrevious)
+        .Emit();
+    return true;
+  }
+
+  auto base_type_id = ExprAsType(context, parse_node, base_type_expr_id);
+  if (!context.TryToCompleteType(base_type_id, [&] {
+        CARBON_DIAGNOSTIC(IncompleteTypeInBaseDecl, Error,
+                          "Deriving from incomplete type `{0}`.", std::string);
```
jonmeow: Maybe "Cannot derive from" instead of just "Deriving from", to emphasize the error?

toolchain/check/handle_class.cpp
```diff
@@ -166,6 +167,98 @@ auto HandleClassDefinitionStart(Context& context, Parse::NodeId parse_node)
...
+    // declaration as being final classes.
+    auto base_class =
+        context.insts()
+            .Get(context.sem_ir().GetTypeAllowBuiltinTypes(base_type_id))
+            .TryAs<SemIR::ClassType>();
+    if (!base_class ||
+        context.classes().Get(base_class->class_id).inheritance_kind ==
+            SemIR::Class::Final) {
+      CARBON_DIAGNOSTIC(BaseIsFinal, Error,
+                        "Deriving from final type `{0}`. Only abstract and "
```
jonmeow: Would it be worth separating out inheritance from non-class types from the final class case? It's just odd to me to think this'll probably show up for things like `interface`, with `interface` labeled as a "final type".

toolchain/check/testdata/class/fail_base_bad_type.carbon
```diff
@@ -0,0 +1,145 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+
+base class Base {}
+class Final {}
+
+class DeriveFromi32 {
```
jonmeow, line 17: Can you test a base value, such as `base: 3`?

toolchain/check/handle_name.cpp
```diff
@@ -222,17 +232,30 @@ auto HandlePointerMemberAccessExpr(Context& context, Parse::NodeId parse_node)
   return context.TODO(parse_node, "HandlePointerMemberAccessExpr");
 }
 
+static auto GetParseNodeAsName(Context& context, Parse::NodeId parse_node)
```
jonmeow: With `self` and `Self`, we've used separate parse nodes. I was taking `package` in that same direction.

Here, you're making `base` use the same parse node as `name`, resulting in the need to look at token kinds. I think we should probably be using a consistent approach: which way are you leaning?

Note, I think this also gets towards the "Literal" parse node split, and whether operators should also be split. Should we require looking at tokens to determine how to read the parse node?

toolchain/check/handle_class.cpp
```diff
@@ -166,6 +167,98 @@ auto HandleClassDefinitionStart(Context& context, Parse::NodeId parse_node)
...
+auto HandleBaseColon(Context& /*context*/, Parse::NodeId /*parse_node*/)
+    -> bool {
+  return true;
+}
+
+auto HandleBaseDecl(Context& context, Parse::NodeId parse_node) -> bool {
+  auto base_type_expr_id = context.node_stack().PopExpr();
+
+  auto class_id =
+      context.node_stack().PeekIf<Parse::NodeKind::ClassDefinitionStart>();
```
jonmeow: Are we guaranteed that this is going to be the class definition start? For example:

```
class X {
  class Y {}
  fn F();
  var v: i32;

  base: Y;
}
```

Either way, please add a test of `base` following with other decls on a class.

Perhaps this could use the decl stack that's being added for modifiers, rather than the node stack?

toolchain/sem_ir/typed_insts.h
zygoloid: I chose this to parallel `Field`, so I'd like to pick a new name that works for both. Would `BaseDecl` and `FieldDecl` work for you?

toolchain/parse/handle_class.cpp
zygoloid: I don't think there's really anything class-specific to factor out of handle_type, at least not as it stands. There will be more class-specific stuff to come later -- at least adapters, maybe support for `extend`, maybe destructors and any other class-specific declarations.

But... given that we're parsing `base: B` in any declaration context and letting check figure it out, moving this to a separate handle_base seems reasonable to me. Done. We can add a handle_adapt, etc. for other kinds of (class-specific but perhaps allowed anywhere by parse) declarations.

toolchain/parse/handle_decl_scope_loop.cpp
zygoloid: #3412 is introducing the infrastructure here; we can look at this again after that.

Added testing.

toolchain/sem_ir/typed_insts.h
jonmeow: SGTM, though if you prefer to leave it as-is, my preference isn't strong.

toolchain/check/handle_class.cpp
zygoloid: Good idea. I've added https://docs.google.com/document/d/1RRYMm42osyqhI2LyjrjockYCutQ5dOf8Abu50kTrkX0/edit?resourcekey=0-kHyqOESbOHmzZphUbtLrTw#heading=h.vyg6tg3frugv but not added much content there for now.

toolchain/check/handle_class.cpp
zygoloid: In context, this looks like:
```console
ERROR: Deriving from incomplete type `X`.
```
... where I think the `ERROR:` part has the emphasis you're looking for here.

toolchain/check/handle_class.cpp
zygoloid: I'd like to look at this as a separate change after this PR, if that works for you: there are 13 calls to this as a diagnostic argument, most of them pre-existing, that pass a bare `true` in, and I'd like to make a consistent change across all of them. (Maybe switching to an enum.)

toolchain/check/handle_class.cpp
zygoloid: Looking at this again, I think "must" works better than "can only" here.

toolchain/check/handle_class.cpp
zygoloid: I thought about that. The issue is that we've vaguely said that we want almost(?) all types to be class types, but it's not clear exactly where we draw the line. `interface`s probably aren't classes, but tuples and structs and `i32` probably are, etc.

It seemed better to go with something that doesn't take an opinion on what is / is not considered to be a class for now.

Added a TODO to reconsider this once the language rules are clearer.

toolchain/check/handle_class.cpp
zygoloid: Yeah, this isn't in line with what we said about looking at the node stack only when handling a parent node.

Switched to using the scope stack instead. That matches what we do for fields.

toolchain/check/testdata/class/fail_base_bad_type.carbon
zygoloid: Done.

toolchain/check/handle_class.cpp
jonmeow: Thanks, and I've added a few more notes to the entry.

toolchain/check/handle_class.cpp
jonmeow: Maybe a style thing: when should errors be statements [of the detail that led to the error], and when is advice to be included? Maybe what I should really be suggesting is more like "`base` must be a complete type."

Chaining that thought, maybe wording should also focus on the keyword `base` rather than the semantic "deriving"?

Not really a strong suggestion here on either half, if you prefer the current message.

toolchain/check/handle_name.cpp
zygoloid: Switched to a different node kind. This necessitated adding node stack functionality to treat all parse nodes that produce names uniformly, like we do for expressions, so there's a fair bit more churn in this patch now. There's probably some refactoring we ought to be doing in the node stack, but I think it can wait until we have a third class of node kinds we want to treat uniformly.

toolchain/check/handle_class.cpp
zygoloid: Done.

toolchain/check/handle_class.cpp
jonmeow: SGTM

toolchain/check/handle_class.cpp
zygoloid: I have a plan, but I'll send a separate PR due to the cross-cutting nature of it.

toolchain/sem_ir/typed_insts.h
zygoloid: I'll send a separate PR with the renames.

toolchain/check/handle_class.cpp
zygoloid: Rephrased a bit.

toolchain/check/node_stack.h
```diff
@@ -212,6 +226,16 @@ class NodeStack {
                    << "; see value in ParseNodeKindToIdKind";
   }
 
+  // Peeks at the ID associated with the top of the name stack, if it's of the
+  // specified kind.
+  template <Parse::NodeKind::RawEnumType RequiredParseKind>
+  auto PeekIf() const -> std::optional<decltype(Peek<RequiredParseKind>())> {
```
jonmeow: Is this still used in the change?

jonmeow (approved):
Basically LGTM, I think remaining comments are small.

toolchain/check/node_stack.h
zygoloid: No! Thanks, removed.

toolchain/sem_ir/typed_insts.h
zygoloid: #3452.

toolchain/check/handle_class.cpp
zygoloid: #3454

PR #3453, opened 2023-12-04
Rename `Name` -> `IdentifierName` given that we have several other kinds of parse nodes that represent names.

zygoloid:
Factor out common checking handling for the different kinds of unqualified names


language_server/language_server.cpp
```diff
@@ -76,12 +76,12 @@ auto LanguageServer::onReply(llvm::json::Value /*id*/,
   return true;
 }
 
-// Returns the text of first child of kind Parse::NodeKind::Name.
+// Returns the text of first child of kind Parse::NodeKind::IdentifierName.
 static auto GetName(const SharedValueStores& value_stores,
```
jonmeow: Do you want to rename this to GetIdentifierName, or add a TODO that it should support names in general?

jonmeow (approved):
LGTM, one minor comment.

language_server/language_server.cpp
zygoloid: Renamed. I *think* this shouldn't be looking for keyword names, given that it's used for symbol indexing.

PR #3454, opened 2023-12-04
Factor out common pattern of trying to complete a type then falling back to an error type if that fails.

zygoloid:
As requested in [review of #3450](https://github.com/carbon-language/carbon-lang/pull/3450#discussion_r1414220143).


toolchain/check/context.h
```diff
@@ -222,6 +222,17 @@ class Context {
       std::optional<llvm::function_ref<auto()->DiagnosticBuilder>> diagnoser =
           std::nullopt) -> bool;
 
+  // Returns the type `type_id` as a complete type, or produces an incomplete
+  // type error and returns an error type. This is a convenience wrapper around
+  // TryToCompleteType.
+  auto AsCompleteType(
+      SemIR::TypeId type_id,
+      std::optional<llvm::function_ref<auto()->DiagnosticBuilder>> diagnoser =
+          std::nullopt) -> SemIR::TypeId {
```
```suggestion
      llvm::function_ref<auto()->DiagnosticBuilder> diagnoser) -> SemIR::TypeId {
```

All the callers pass a diagnoser, maybe it should be required?

toolchain/check/context.h
```diff
@@ -222,6 +222,17 @@ class Context {
       std::optional<llvm::function_ref<auto()->DiagnosticBuilder>> diagnoser =
           std::nullopt) -> bool;
 
+  // Returns the type `type_id` as a complete type, or produces an incomplete
+  // type error and returns an error type. This is a convenience wrapper around
+  // TryToCompleteType.
+  auto AsCompleteType(
```
jonmeow: Should we have TryAsCompleteType for consistency, or do you still want "To" because it's returning `bool` instead of an optional? [Hmm, maybe CanCompleteType to avoid the Try -> optional pattern entirely?] [or just TryCompleteType, dropping "To", so that they align differently?] [very optional]

jonmeow (approved):
Thanks!

toolchain/check/context.h
zygoloid: Yes, I would like to keep the `To` because I'd expect `TryAsCompleteType` to return an `optional<TypeId>`.

Also, because the diagnoser is now mandatory for `AsCompleteType` (which seems right because we shouldn't form a `TypeId::Error` without diagnosing) they're slightly less similar, so giving them slightly less similar names seems a bit better.

PR #3455, opened 2023-12-05
Simplify interface for getting an instruction from a type.

zygoloid:


toolchain/sem_ir/file.h
```diff
@@ -238,6 +239,30 @@ class File : public Printable<File> {
     }
   }
 
+  // Returns the instruction used to define the specified type.
+  auto GetType(TypeId type_id) const -> Inst {
```
jonmeow: This feels like it should be `GetTypeInst`, since `GetType` would really be `types().Get`. Similar thoughts on below.

Maybe these could be moved onto insts(), where you could do `insts().GetType` more naturally? Could that have a reference to types() without creating problems?

toolchain/sem_ir/file.h
```diff
@@ -226,7 +226,8 @@ class File : public Printable<File> {
     complete_types_.push_back(object_type_id);
   }
 
-  auto GetTypeAllowBuiltinTypes(TypeId type_id) const -> InstId {
+  // Returns the ID of the instruction used to define the specified type.
+  auto GetTypeInstId(TypeId type_id) const -> InstId {
```
jonmeow: As long as you're touching this, maybe `types().GetInstId`?

jonmeow (approved):

jonmeow:
Note, I'm okay with this as-is if you want to merge, but I do lean towards moving things off File.

toolchain/sem_ir/file.h
zygoloid: Last time I looked at this, referencing one `ValueStore` from another was problematic due to `File` being copyable (and actually copied in practice). But... after the recent changes for imports, we might not be copying `File`s any more, so maybe this is easier now.

jonmeow: Yeah, I think the ValueStores already aren't copyable.

jonmeow: There's also the BumpPtrAllocator

zygoloid: Moved onto `types()` , rather than `insts()`. I think this makes the call sites read a bit better, but I could go either way.

toolchain/sem_ir/file.h

toolchain/sem_ir/file.h
jonmeow: Works for me too.

toolchain/sem_ir/value_stores.h
```diff
@@ -58,6 +59,68 @@ class ConstantStore {
   llvm::SmallVector<InstId> values_;
 };
 
+// Provides a ValueStore wrapper with an API specific to types.
+class TypeStore : public ValueStore<TypeId> {
+ public:
+  TypeStore(InstStore* insts) : insts_(insts) {}
```
jonmeow:
```suggestion
  explicit TypeStore(InstStore* insts) : insts_(insts) {}
```

jonmeow (approved):

PR #3456, opened 2023-12-05
Remove support for disambiguating a stringified type as being a type.

zygoloid:
Most of the calls to `StringifyType` already passed `true` for `in_type_context`. Checking the rest, I found that every one of them was already sufficiently clear that they were printing a type, or could be made so with a very small change to the diagnostic text.


jonmeow (approved):
Yay!

PR #3459, opened 2023-12-06
Require `extend` before `base: BaseType;`

josh11b:


zygoloid (approved):
Thanks, looks good.

Would it be worth adding a test for `extend extend base: B`? Or maybe there's already test coverage for repeated modifiers elsewhere?

josh11b:
> Thanks, looks good.
> 
> Would it be worth adding a test for `extend extend base: B`? Or maybe there's already test coverage for repeated modifiers elsewhere?

There is some decent test coverage for repeated modifiers, but no reason not to add one here too.

PR #3461, opened 2023-12-06
Switch modifiers to use distinct parse nodes.

jonmeow:
This is supporting a direction that all parse nodes should correspond to a single token, allowing for reduced tokenized buffer access during checking (it's still necessary for diagnostics, and some literals).

One of the justifications for a unified parse node was implementation LOC: note this is slightly smaller, using macros to reduce some duplication. While this does add more switching in HandleDeclScopeLoop, that's offset by less explicit switching in the check handlers. Also, I think the duplication in HandleDeclScopeLoop can be reduced by shifting the flow there, which I'll do in a separate PR.


zygoloid (commented):
This seems to work out pretty nicely. LGTM, not approving because I'd also like @josh11b to review.

toolchain/check/handle_modifier.cpp
```diff
@@ -33,28 +34,16 @@ static auto EmitNotAllowedWithDiagnostic(Context& context,
...
-    default:
-      CARBON_FATAL() << "Unhandled access modifier keyword";
-  }
-}
-
-auto HandleAccessModifierKeyword(Context& context, Parse::NodeId parse_node)
-    -> bool {
-  auto keyword = GetAccessModifierEnum(
-      context.tokens().GetKind(context.parse_tree().node_token(parse_node)));
+auto HandleModifier(Context& context, Parse::NodeId parse_node,
```
josh11b: Shouldn't this have a name that doesn't start with `Handle` since it is not called by the same mechanism as the other `Handle` functions?

jonmeow: It should definitely be static, that's a mistake. But there are a number of functions that are `static auto Handle` which are helpers for the handle function, e.g. `HandleStatementKeywordFinish` -- I think the `static` and different arguments help make it apparent that they're not part of the boilerplate handlers.

josh11b (approved):

PR #3463, opened 2023-12-06
Refactor the decl scope loop to consolidate modifier handling.

jonmeow:
Building on #3462, trying to make the flow easier to see, also making a little more use of macros for boilerplate modifier handling.


toolchain/parse/handle_decl_scope_loop.cpp
```diff
@@ -2,12 +2,54 @@
...
+    }
+    case Lex::TokenKind::Library: {
+      context.PushState(State::Library);
+      return true;
+    }
+    case Lex::TokenKind::Package: {
+      context.PushState(State::Package);
+      return true;
+    }
+    default: {
```
zygoloid, line 35: Optional: to reduce indentation, consider `break;`ing here and putting the `default` case after the loop.

toolchain/parse/handle_decl_scope_loop.cpp
```diff
@@ -25,81 +67,192 @@ static auto OutputInvalidParseSubtree(Context& context, int32_t subtree_start)
                   /*has_error=*/true);
 }
 
-// Handles an unrecognized declaration.
-static auto HandleUnrecognizedDecl(Context& context, int32_t subtree_start)
-    -> void {
+// Reports an error and outputs an invalid parse tree.
```
zygoloid: I had trouble understanding what "outputs an invalid parse tree" means.

It seems like the purpose of `OutputInvalidParseSubtree` is to recover from an error by skipping to a good stopping place and fixing up the parse tree to match. I don't get that from the name. Maybe something more like `FinishAndSkipInvalidDecl` would be clearer?

toolchain/parse/handle_decl_scope_loop.cpp
```diff
@@ -35,71 +77,182 @@ static auto HandleUnrecognizedDecl(Context& context, int32_t subtree_start)
...
+              context.PositionKind(Lookahead::NextToken))) {
+        // `impl` is considered a declaration modifier if it is followed by
+        // another modifier or an introducer.
+        return false;
+      } else {
+        // TODO: Treat this `impl` token as a declaration introducer
+        HandleUnrecognizedDecl(context, state.subtree_start);
+        return true;
+      }
     }
```
zygoloid: It seems like there's still duplication here; can we check for the ambiguous cases first somehow? In all three cases, if a token is both an introducer and a modifier, the logic is the same: it's a modifier if the next token is a modifier or an introducer.

toolchain/parse/handle_decl_scope_loop.cpp
jonmeow: My leaning was this direction to keep everything in the switch, partly because `default` can't simply be removed. [I think it's visually more consistent]

toolchain/parse/handle_decl_scope_loop.cpp
jonmeow: Done

toolchain/parse/handle_decl_scope_loop.cpp
jonmeow: Refactored a little for this.

toolchain/parse/handle_decl_scope_loop.cpp
```diff
@@ -25,81 +68,185 @@ static auto OutputInvalidParseSubtree(Context& context, int32_t subtree_start)
...
+// Handles `base` as a declaration.
+static auto HandleBaseAsDecl(Context& context, Context::StateStackEntry state)
+    -> void {
+  // `base` may be followed by:
+  // - a colon
+  //   => assume it is an introducer, as in `extend base: BaseType;`.
+  // - a modifier or an introducer
+  //   => assume it is a modifier, as in `base class`; which is handled
+  //      by falling through to the next case.
+  // Anything else is an error.
```
zygoloid: I think this comment is out of date (at least the modifier part and "falling through")

toolchain/parse/handle_decl_scope_loop.cpp
```diff
@@ -110,164 +257,17 @@ auto HandleDeclScopeLoop(Context& context) -> void {
...
-  auto introducer = [&](NodeKind node_kind, State next_state) {
-    context.ReplacePlaceholderNode(state.subtree_start, node_kind,
-                                   context.Consume());
-    // Reuse state here to retain its `subtree_start`
-    state.state = next_state;
-    context.PushState(state);
-  };
-
   bool saw_modifier = false;
   while (true) {
```
zygoloid:
```suggestion
```

toolchain/parse/handle_decl_scope_loop.cpp
```diff
@@ -110,164 +257,17 @@ auto HandleDeclScopeLoop(Context& context) -> void {
...
+    if (TryHandleAsModifier(context, position_kind)) {
+      // Modifiers will continue through the loop, but we track we saw them.
+      saw_modifier = true;
+    } else if (TryHandleAsDecl(context, state, saw_modifier, position_kind)) {
+      return;
+    } else {
+      HandleUnrecognizedDecl(context, state.subtree_start);
+      return;
     }
   }
```
zygoloid:
```suggestion
```

toolchain/parse/handle_decl_scope_loop.cpp
```diff
@@ -110,164 +257,17 @@ auto HandleDeclScopeLoop(Context& context) -> void {
...
-      }
+    auto position_kind = context.PositionKind();
+    if (TryHandleAsModifier(context, position_kind)) {
+      // Modifiers will continue through the loop, but we track we saw them.
+      saw_modifier = true;
+    } else if (TryHandleAsDecl(context, state, saw_modifier, position_kind)) {
+      return;
+    } else {
+      HandleUnrecognizedDecl(context, state.subtree_start);
+      return;
```
zygoloid: I  think this would be a bit simpler if you moved the handling of the declaration out of the loop, eg:
```suggestion
  while (TryHandleAsModifier(context, context.PositionKind())) {
    saw_modifier = true;
  }
  if (!TryHandleAsDecl(context, state, saw_modifier, context.PositionKind())) {
    HandleUnrecognizedDecl(context, state.subtree_start);
  }
```
(Suggestion split in three parts to work around github limitations.)

zygoloid (approved):
Thanks!

toolchain/parse/handle_decl_scope_loop.cpp
jonmeow: Cleaning up the comment.

jonmeow:
Thanks, I'd totally overlooked the ability to shift the `while`.

PR #3466, opened 2023-12-06
Implement `operator+` for `i32`

aneeshdurg:
Implements addition via the `+` operator for i32.

### DIFF
--- a/toolchain/lower/handle.cpp
+++ b/toolchain/lower/handle.cpp
@@ -51,10 +51,11 @@ auto HandleAssign(FunctionContext& context, SemIR::InstId /*inst_id*/,
   context.FinishInit(storage_type_id, inst.lhs_id, inst.rhs_id);
 }
 
-auto HandleBinaryOperatorAdd(FunctionContext& /*context*/,
-                             SemIR::InstId /*inst_id*/,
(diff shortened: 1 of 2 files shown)


aneeshdurg:
I couldn't really tell what kind of contributions towards the toolchain the project is looking for right now, but this seemed like an easy gap to fill since we only support `i32` numeric types at the moment. If this kind of contribution isn't wanted at the moment, please let me know what other areas could use help instead!

jonmeow:
On Discord, #toolchain is used for discussion, and it may be helpful to query there if you're curious about where to start. Here's the doc we're using for toolchain tasks: https://docs.google.com/document/d/1O2rcDO70BamTERrKhJRLODarNCQOHQWUweNdSfLw-So/edit?resourcekey=0-gWJkSd-tlpCmk2lok1UcPg (also in #toolchain pinned messages if you forget)

As noted there, for operators we're really looking more at first implementing interfaces, then implementing operators as interfaces per the design, e.g.: https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/expressions/arithmetic.md#extensibility

However, interfaces are substantial work: something like `->` or the `alias` support noted at the end might be easier to pick up.

aneeshdurg:
Thanks for the pointers! I'll try to pick one of those up.

PR #3471, opened 2023-12-07
Upgrade clang-format version

jonmeow:
This apparently includes a fix for https://github.com/llvm/llvm-project/issues/47664 (previously https://bugs.llvm.org/show_bug.cgi?id=48320)

Note this continues with a clang-format version that's different from the compiler versions we're often using, but the differences in formatting seem worthwhile.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -72,14 +72,14 @@ repos:
         types_or: [c++, proto]
         language: python
         args: ['-i']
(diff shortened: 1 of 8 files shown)


zygoloid:
> This apparently includes a fix for https://bugs.llvm.org/show_bug.cgi?id=48320

This is https://github.com/llvm/llvm-project/issues/47664 in the new bug tracker.

zygoloid (approved):
LG, would be nice to use the new bug number in the commit message.

PR #3473, opened 2023-12-07
Parsing `impl`...`as`

josh11b:


toolchain/parse/tree.cpp
```diff
@@ -46,9 +46,7 @@ auto Tree::Parse(Lex::TokenizedBuffer& tokens, DiagnosticConsumer& consumer,
   context.AddLeafNode(NodeKind::FileEnd, *context.position());
 
   if (auto verify = tree.Verify(); !verify.ok()) {
-    if (vlog_stream) {
-      tree.Print(*vlog_stream);
-    }
+    tree.Print(llvm::errs());
     CARBON_FATAL() << "Invalid tree returned by Parse(): " << verify.error();
```
zygoloid, line 50:
```suggestion
    CARBON_FATAL() << "Invalid tree returned by Parse(): " << verify.error() << "\n" << tree;
```
Does this work?

toolchain/parse/handle_impl.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+    case Lex::TokenKind::As:
+      context.AddLeafNode(NodeKind::ImplAs, context.Consume());
+      context.PushState(State::Expr);
+      break;
+
+    default:
+      context.PushState(State::ImplAs);
+      context.PushStateForExpr(PrecedenceGroup::ForImplAs());
+      break;
```
jonmeow: Can you add some comments explaining the difference in logic, given both are treated as `ImplAs`?

toolchain/parse/handle_impl.cpp
```diff
@@ -0,0 +1,61 @@
...
+  }
+}
+
+auto HandleImplAfterForall(Carbon::Parse::Context& context) -> void {
+  auto state = context.PopState();
+  if (state.has_error) {
+    context.ReturnErrorOnState();
+    return;
```
jonmeow: In this state, you've encountered a `forall` token. Why do you return here instead of letting the `AddNode` call occur, adding a parse node for the error token?

(note too, with the `return`, using `state.has_error` below is equivalent to `/*has_error=*/false`, which might be clearer)

toolchain/parse/node_kind.def
```diff
@@ -681,6 +681,35 @@ CARBON_PARSE_NODE_KIND_BRACKET(InterfaceDecl, InterfaceIntroducer,
...
 
+// `impl ... as`:
+//     ImplIntroducer
+//     _repeated_ _external_: AccessModifierKeyword or DeclModifierKeyword
+//         _external_: ImplicitParamList
+//       ImplForall
+//     _optional_
+//     _optional_ _external_: expression
+//     ImplAs
+//     _external_: expression
```
jonmeow, line 702: Isn't the expression kind of an argument to `as`? Should it be a child instead of a sibling?

toolchain/parse/node_kind.def
```diff
@@ -681,6 +681,35 @@ CARBON_PARSE_NODE_KIND_BRACKET(InterfaceDecl, InterfaceIntroducer,
                                CARBON_TOKEN(Semi)
                                    CARBON_IF_ERROR(CARBON_TOKEN(Interface)))
 
+// `impl ... as`:
+//     ImplIntroducer
+//     _repeated_ _external_: AccessModifierKeyword or DeclModifierKeyword
+//         _external_: ImplicitParamList
+//       ImplForall
+//     _optional_
```
jonmeow: I was thinking about this, and I wonder if it'd be clearer to do something like:

(below):

```
// `forall ...`:
//   _external_: ImplicitParamList
// ImplForall
```

and here:

```
//     _optional_: ImplForAll (see below)
```

toolchain/parse/node_kind.def
```diff
@@ -681,6 +681,35 @@ CARBON_PARSE_NODE_KIND_BRACKET(InterfaceDecl, InterfaceIntroducer,
...
+// ImplDefinition
+//
+// The above is the structure for a definition; for a declaration,
+// ImplDefinitionStart and later nodes are removed and replaced by
+// ImplDecl.
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(ImplIntroducer, 0, CARBON_TOKEN(Impl))
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(ImplForall, 1, CARBON_TOKEN(Forall))
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(ImplAs, 0,
+                                   CARBON_TOKEN(As)
+                                       CARBON_IF_ERROR(CARBON_ANY_TOKEN))
```
jonmeow: Is this true of the current `ImplAs`? It looks like you're only emitting it after finding the `As` token.

toolchain/parse/state.def
```diff
@@ -1015,33 +1020,33 @@ CARBON_PARSE_STATE(StatementWhileBlockFinish)
 
 // Handles parsing after the declaration scope of a type.
 //
-// class/interface/constraint { ... }
-//                                  ^
+// class/impl/interface/constraint { ... }
```
jonmeow: This is starting to feel like it should just be `<introducer>`

toolchain/parse/state.def
```diff
@@ -1052,6 +1057,44 @@ CARBON_PARSE_STATE_VARIANTS3(TypeAfterParams, Class, Interface, NamedConstraint)
 //   (state done)
 CARBON_PARSE_STATE(BaseDecl)
 
+// Handles processing of an `impl...as` declaration after the introducer.
+//
+// impl forall [ ...
```
jonmeow, line 1119: I think more of these need to be tested. I'm seeing `impl X as Y` but no other syntax variations.

For example:

- impl as Y;
- impl forall;
- impl forall X;
- impl forall X as Y;
- impl forall X Y as Z;
- impl X as Y Z;

toolchain/parse/tree.cpp
```diff
@@ -46,9 +46,9 @@ auto Tree::Parse(Lex::TokenizedBuffer& tokens, DiagnosticConsumer& consumer,
   context.AddLeafNode(NodeKind::FileEnd, *context.position());
 
   if (auto verify = tree.Verify(); !verify.ok()) {
-    if (vlog_stream) {
-      tree.Print(*vlog_stream);
-    }
+    // TODO: Restrict this to the subtree with the error, or only output
+    // when vlogging is enabled.
```
jonmeow:
```suggestion
    // TODO: This is temporarily printing to stderr directly during development. If we can, restrict this to a subtree with the error and add it to the stack trace (such as with PrettyStackTraceFunction). Otherwise, switch back to vlog_stream prior to broader distribution so that end users are hopefully comfortable copy-pasting stderr when there are bugs in tree construction.
```

Trying to offer some more information for the TODO so that someone doesn't come along and just change it to vlog.

toolchain/parse/node_kind.def
josh11b: It seems much more useful to have it as a sibling to delimit the optional expression before it. Having `ImplAs` right before `ImplDefinitionStart` / `ImplDecl` doesn't seem helpful to the check stage. I don't think of this last expression as an argument to `as`, but all of these siblings are arguments of the impl declaration.

toolchain/parse/node_kind.def
josh11b: Fixed.

toolchain/parse/state.def
josh11b: Done, and fixed the code to make them work.

toolchain/parse/handle_impl.cpp
josh11b: I'm pretty uncertain about how to handle errors in the parser, could you perhaps suggest some code? I'm particularly uncertain about what can be assumed if `state.has_error` is true. When I try moving the `context.AddNode` call before the `if (state.has_error)`, I get a parse validation error since the `ImplForall` node is missing its expected child. I assume filling in the missing child is what you mean by "adding a parse node for the error token"? I took a stab at it, but I'm not sure what is expected here. I also deleted the `return;` on error, but that is just a guess.

It would be **_really_** helpful to add a section to https://docs.google.com/document/d/1RRYMm42osyqhI2LyjrjockYCutQ5dOf8Abu50kTrkX0/edit?resourcekey=0-kHyqOESbOHmzZphUbtLrTw#heading=h.d6o1d67xs8xu about what to do about errors in the parse stage:

* what does `state.has_error` mean? when should it be set? what can you assume when it has been set by another function? how should you handle it?
* what invariants need to be maintained when there are errors?
* what are common patterns?

toolchain/parse/handle_impl.cpp
zygoloid: Note that the first case is creating an `ImplAs` *node*, whereas the second case is pushing an `ImplAs` *state*. Maybe renaming the state would help (`ImplBeforeAs`?).

toolchain/parse/handle_impl.cpp
```diff
@@ -0,0 +1,76 @@
...
+      break;
+
+    case Lex::TokenKind::As:
+      context.AddLeafNode(NodeKind::ImplAs, context.Consume());
+      context.PushState(State::Expr);
+      break;
+
+    default:
+      context.PushState(State::ImplAs);
+      context.PushStateForExpr(PrecedenceGroup::ForImplAs());
```
zygoloid: I was expecting to see `ForType` here. Do we intend to have different rules for the type operands of `impl T1 as T2` than for `let x: T1`?

toolchain/parse/handle_impl.cpp
```diff
@@ -0,0 +1,76 @@
...
+  context.AddNode(NodeKind::ImplForall, state.token, state.subtree_start,
+                  state.has_error);
+
+  if (context.PositionKind() == Lex::TokenKind::As) {
+    context.AddLeafNode(NodeKind::ImplAs, context.Consume());
+    context.PushState(State::Expr);
+  } else {
+    context.PushState(State::ImplAs);
+    context.PushStateForExpr(PrecedenceGroup::ForImplAs());
+  }
```
zygoloid: Can this share code with `HandleImplAfterIntroducer`?

toolchain/parse/handle_impl.cpp
josh11b: I tried `ForType`, but that is currently defined as equivalent to `ForTopLevelExpr()`, which means it will include the `as` afterwards.

toolchain/parse/handle_impl.cpp
josh11b: Done.

toolchain/parse/handle_impl.cpp
josh11b: Done.

toolchain/parse/state.def
josh11b: Maybe, but this makes it easier to search for `class` or `interface` and find the relevant states.

toolchain/parse/node_kind.def
josh11b: Done.

toolchain/parse/handle_impl.cpp
```diff
@@ -0,0 +1,79 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "toolchain/parse/context.h"
+
+namespace Carbon::Parse {
+
+static auto ExpectAsOrTypeExpression(Carbon::Parse::Context& context) -> void {
+  if (context.PositionKind() == Lex::TokenKind::As) {
```
jonmeow:
```suggestion
  if (context.PositionIs(Lex::TokenKind::As)) {
```

toolchain/parse/handle_impl.cpp
```diff
@@ -0,0 +1,79 @@
...
+    context.PushStateForExpr(PrecedenceGroup::ForImplAs());
+  }
+}
+
+auto HandleImplAfterIntroducer(Carbon::Parse::Context& context) -> void {
+  auto state = context.PopState();
+  state.state = State::DeclOrDefinitionAsImpl;
+  context.PushState(state);
+
+  if (context.PositionKind() == Lex::TokenKind::Forall) {
```
jonmeow:
```suggestion
  if (context.PositionIs(Lex::TokenKind::Forall)) {
```

toolchain/parse/node_kind.def
jonmeow: Okay, just wanted to make sure you're considered it.

toolchain/parse/handle_impl.cpp
jonmeow: I'd meant just dropping the return.

I'd expected though that ParamListAsImplicit would always add a node, even in error cases. That may be a consistency issue. You adding an InvalidParse node is an unintended workaround.

Anyways, current state is fine, though it's probably worth someone (maybe me) changing ParamList behavior there.

Regarding the rest, jotted down some notes: https://docs.google.com/document/d/1RRYMm42osyqhI2LyjrjockYCutQ5dOf8Abu50kTrkX0/edit?resourcekey=0-kHyqOESbOHmzZphUbtLrTw#heading=h.f0xq7he13pm

jonmeow (approved):
Looks good!

toolchain/parse/handle_impl.cpp
josh11b: Based on the observed behavior in the test, it seems that ParamListAsImplicit sets error == true on the parent state if it did not output a node. If ParamListAsImplicit does output a node, that node has the error state set to true, but the parent state is left alone.

josh11b:
FYI, I had to make some changes after merging. The trickiest one is now that `impl` is parsed, `check_fuzzer` started executing some new unimplemented `interface` code paths.

jonmeow:
LGTM, todo's seem reasonable for that.

PR #3474, opened 2023-12-07
Add interface support to check

josh11b:
Largely copied from the `class` code


toolchain/check/handle_interface.cpp
```diff
@@ -3,26 +3,154 @@
...
+
+  // Enter the interface scope.
+  context.PushScope(interface_decl_id, interface_info.scope_id);
+
+  // TODO: Introduce `Self`.
+
+  context.inst_block_stack().Push();
+  context.node_stack().Push(parse_node, interface_id);
+  // TODO: do we need this for an interface?
+  context.args_type_info_stack().Push();
```
zygoloid: Do we want to create a witness table for interfaces? (I think we might -- even if we don't use them at runtime, I think we'll probably want a compile-time representation for them.) If we do, I think this stack would be the most consistent place to build one.

If not -- or not for now -- then we can remove this and the pop below.

toolchain/check/modifiers.cpp
```diff
@@ -88,7 +88,12 @@ auto CheckAccessModifiersOnDecl(Context& context, Lex::TokenKind decl_kind)
 
 auto RequireDefaultFinalOnlyInInterfaces(Context& context,
                                          Lex::TokenKind decl_kind) -> void {
-  // TODO: Skip this if *context.current_scope_kind() == SemIR::InterfaceDecl
+  if (auto kind = context.current_scope_kind()) {
+    if (*kind == SemIR::InterfaceDecl::Kind) {
+      // Both `default` and `final` allowed in an interface definition.
+      return;
+    }
+  }
```
zygoloid: Maybe you could simplify this to:
```suggestion
  if (context.GetCurrentScopeAs<SemIR::InterfaceDecl>()) {
    // Both `default` and `final` allowed in an interface definition.
    return;
  }
```
(I think this ends up doing slightly more work, but I'd expect it to optimize to the same thing.)

Or if not that, then maybe we could add a `CurrentScopeIs<Kind>`? It looks like the other caller of `current_scope_kind` would also benefit from this change.

zygoloid (approved):

toolchain/check/handle_interface.cpp
josh11b: I am not sure, but if so I think it would be in a later change.

toolchain/check/modifiers.cpp
josh11b: Created and switched to `CurrentScopeIs<Kind>`.

PR #3476, opened 2023-12-08
Fix LexicalLookupResult's node_id -> inst_id

jonmeow:
Noticed while working on #3475, but felt like it would be easy enough to separate out.


chandlerc (approved):
LG (not merging since in a stack).

jonmeow:
Thanks, rebased so that I can merge while the prior PR is still pending.

PR #3478, opened 2023-12-08
Avoid unnecessary const_cast<>

urielgarciarivas:
I noticed there is an unnecessary `const_cast` in the Explorer. `ExplorerMain(...)` can just receive `argv` as const in my opinion, I apologize if that casting was intended.

### DIFF
--- a/explorer/main_bin.cpp
+++ b/explorer/main_bin.cpp
@@ -5,7 +5,7 @@
 #include "common/bazel_working_dir.h"
 #include "explorer/main.h"
 
-auto main(int argc, char** argv) -> int {
+auto main(int argc, const char** argv) -> int {
   Carbon::SetWorkingDirForBazel();
 
   static int static_for_main_addr;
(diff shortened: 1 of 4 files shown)


explorer/main_bin.cpp
```diff
@@ -5,7 +5,7 @@
 #include "common/bazel_working_dir.h"
 #include "explorer/main.h"
 
-auto main(int argc, char** argv) -> int {
+auto main(int argc, const char** argv) -> int {
```
chandlerc, line 8: This isn't correct, the signature of `main` doesn't include `const` and we shouldn't deviate from that here. For reference, the spec is here:
https://eel.is/c++draft/basic.start.main#2.2

And given that spec, I think the rest of this code is reasonable as-is.

chandlerc (changes requested):
Thanks for the contribution, but this change isn't correct. See below for details.

explorer/main_bin.cpp
urielgarciarivas: Didn't know this guideline, thank you!

PR #3479, opened 2023-12-08
Rename `ParamList` to `TuplePattern`

geoffromer:


toolchain/check/handle_binding_pattern.cpp
```diff
@@ -115,7 +115,7 @@ auto HandleBindingPattern(Context& context, Parse::NodeId parse_node) -> bool {
     }
 
     case Parse::NodeKind::ImplicitParamListStart:
-    case Parse::NodeKind::ParamListStart:
+    case Parse::NodeKind::TuplePatternStart:
       // Parameters can have incomplete types in a function declaration, but not
       // in a function definition. We don't know which kind we have here.
```
zygoloid, line 120: This comment doesn't make a lot of sense any more. But I think this is deeper than the comment: the overall strategy of looking at the immediate context to determine whether you need a complete type -- or more broadly, what kind of instruction you're even creating -- doesn't make a lot of sense once we have nested patterns.

Maybe add a TODO to re-examine this?
```suggestion
      // Parameters can have incomplete types in a function declaration, but not
      // in a function definition. We don't know which kind we have here.
      // TODO: A tuple pattern can appear in other places than function parameters.
```

toolchain/parse/state.def
```diff
@@ -649,56 +649,61 @@ CARBON_PARSE_STATE(NamespaceFinish)
...
-//   2. ParamFinishAs(Implicit|Regular)
-CARBON_PARSE_STATE_VARIANTS2(Param, Implicit, Regular)
+//   2. PatternListElementFinishAs(Implicit|Tuple)
+CARBON_PARSE_STATE_VARIANTS2(PatternListElement, Implicit, Tuple)
 
-// Finishes parsing a parameter, including the optional trailing `,`. If there
-// are more parameters, enqueues another parameter parsing state.
+// Finishes parsing a pattern in a comma-separated list, including the optional
+// trailing `,`. If there are more patterns, enqueues another pattern
+// parsing state.
```
zygoloid:
```suggestion
// Finishes parsing a pattern in a comma-separated list, including the
// optional trailing `,`. If there are more patterns, enqueues another
// pattern parsing state.
```
80 columns

zygoloid (approved):

PR #3480, opened 2023-12-08
Split out and/or operator handling from infix.

jonmeow:
This is also doing the parse node split, allowing lower reliance in formatter on the tokenized buffer (something that I may be touching more due to import handling).


toolchain/parse/state.def
```diff
@@ -488,7 +493,7 @@ CARBON_PARSE_STATE(ExprLoop)
 // Completes an ExprLoop pass by adding an infix operator, then goes back
 // to ExprLoop.
 //
-// expr <binary operator> expr ...
+// expr <infix operator> expr ...
 //                            ^
```
zygoloid:
```suggestion
//                           ^
```

toolchain/parse/state.def
```diff
@@ -471,7 +471,7 @@ CARBON_PARSE_STATE(ExprInPostfixLoop)
 
 // Handles processing of an expression.
 //
-// expr <binary operator> ...
+// expr <infix operator> ...
 //      ^~~~~~~~~~~~~~~~~
```
zygoloid:
```suggestion
//      ^~~~~~~~~~~~~~~~
```

toolchain/parse/state.def
```diff
@@ -480,6 +480,11 @@ CARBON_PARSE_STATE(ExprInPostfixLoop)
 //      ^~~~~~~~~~~~~~~~~~
 //   1. ExprLoop
 //
+// expr <short circuit operator> ...
+//      ^~~~~~~~~~~~~~~~~
```
zygoloid:
```suggestion
// expr <short circuit operator> ...
//      ^~~~~~~~~~~~~~~~~~~~~~~~
```

toolchain/parse/state.def
```diff
@@ -501,6 +506,14 @@ CARBON_PARSE_STATE(ExprLoopForBinary)
 //   1. ExprLoop
 CARBON_PARSE_STATE(ExprLoopForPrefix)
 
+// Completes an ExprLoop pass by adding a short circuit operator, then goes back
+// to ExprLoop.
+//
+// expr <short circuit operator> expr ...
+//                              ^
```
zygoloid: Isn't this here:
```suggestion
// expr <short circuit operator> expr ...
//                                   ^
```

zygoloid (approved):
Sure, seems reasonable, and I like that the formatter doesn't need to look at tokens any more.

toolchain/parse/state.def
jonmeow: Done

toolchain/parse/state.def
jonmeow: done

toolchain/parse/state.def
jonmeow: done

toolchain/parse/state.def
jonmeow: Done

PR #3481, opened 2023-12-08
Split out infix and prefix operators to separate node kinds.

jonmeow:
This leaves a single state for each in the expr loop. I was trying to think through ways to have per-token states, but they felt sort of bulky. 

Note this is more verbose: but I think the long-term is going to be that when we start wanting to add handlers, we're going to need to switch to different names based on the token found. As a consequence, the parse state logic will end up diverging a little, and we'll just want to align towards boilerplate handlers.

Short-term, this opens up a path for saying that each parse node corresponds to precisely one token in success states, and separates out what were becoming big handler functions in check.


jonmeow:
(note, dependencies in stack)

toolchain/parse/handle_expr.cpp
```diff
@@ -279,44 +280,72 @@ auto HandleExprLoop(Context& context) -> void {
...
+        state.state = State::ExprLoopForInfixOperator;
         break;
     }
 
     context.PushState(state);
     context.PushStateForExpr(operator_precedence);
   } else {
-    context.AddNode(NodeKind::PostfixOperator, state.token, state.subtree_start,
-                    state.has_error);
+    context.AddNode(NodeKind::PostfixOperatorStar, state.token,
```
zygoloid, line 290: Even if we only have one postfix operator for now, I think we should add a `switch` here and a `..._POSTFIX` macro.

josh11b (approved):

toolchain/parse/handle_expr.cpp
jonmeow: Okay, I'll send a separate PR for that.

PR #3482, opened 2023-12-08
Replace the NodeKind StructFieldUnknown with InvalidParse.

jonmeow:
StructFieldUnknown was used previously for invalid parses. But we have added other, more common ways of talking about those; so this is removing the special-case.


zygoloid (approved):
Looks good. Not merging due to other (appear to be unrelated / not dependency) changes in the commit stack.

jonmeow:
Yes, dependencies are in the stack.

PR #3424, opened 2023-11-27
Fix crash when array contents aren't an expression.

jonmeow:
FloatingPointLiteral wrap is a clang-format thing. I didn't touch it manually.

PR #3484, opened 2023-12-08
On invalid parse nodes, if the token may vary, allow any token.

jonmeow:
Note #3486 rewrites the macro behavior, and is already approved: so this PR is only for the changed enforcement during error. Also, #3493 already changed several things to allow any token while this PR was awaiting review, but this still changes enforcement for `For` and `If`. 

This was brought up on [#toolchain](https://discord.com/channels/655572317891461132/655578254970716160/1182066616456970251), and I think this any-on-error approach gets at least some support. We could try setting it to the introducer, but it's quite possible we want it to be something like the token which led to the parse error, rather than a static token. That leads to a conclusion that, most typically, we'll expect arbitrary tokens when error conditions may lead to tokens which aren't the expected token.

A couple related, recent `CARBON_IF_ERROR` crash fixes can be found in #3404 and #3424. Something like #3404 would've been needed regardless because `namespace` didn't have `CARBON_IF_ERROR` before, although I might've missed the underlying issue with declarations because only `namespace` had a relevant test (that is, if #3404 had added `CARBON_ANY_TOKEN_ON_ERROR`, I wouldn't have had a crash in #3462). #3424 would've been avoided with this change because there was a `CARBON_IF_ERROR`, and it was just too restrictive.

### DIFF
--- a/toolchain/parse/node_kind.cpp
+++ b/toolchain/parse/node_kind.cpp
@@ -50,6 +50,11 @@ auto NodeKind::child_count() const -> int32_t {
 // NOLINTNEXTLINE(readability-function-size): It's hard to extract macros.
 auto CheckNodeMatchesLexerToken(NodeKind node_kind, Lex::TokenKind token_kind,
                                 bool has_error) -> void {
+  // As a special-case, a placeholder node may correspond to any lexer token.
+  if (node_kind == NodeKind::Placeholder) {
+    return;
+  }
+
   switch (node_kind) {
     // Use `CARBON_LOG CARBON_ANY_TOKEN` to discover which combinations happen
     // in practice.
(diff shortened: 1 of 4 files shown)


jonmeow:
(note, dependencies in stack)

jonmeow:
Note, some changes here overlapped with #3493, which would've incidentally fixed the `namespace` crash. However, since #3493 still didn't change *everything* do CARBON_IF_ERROR(CARBON_ANY_TOKEN), this is still a question.

zygoloid (approved):
I think this makes sense. The prior behaviour doesn't seem especially useful to me, given that if you wanted the `for` or `if` token you can easily find them.

PR #3485, opened 2023-12-08
Add `PushState` overload to combine `Parse::State` change with `StateStackEntry`

josh11b:
Shortens a common `PushState` pattern

### DIFF
--- a/toolchain/parse/context.h
+++ b/toolchain/parse/context.h
@@ -273,6 +273,12 @@ class Context {
         << "Excessive stack size: likely infinite loop";
   }
 
+  // Pushes a constructed state onto the stack, with a different parse state.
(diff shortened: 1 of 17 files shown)


jonmeow (approved):
Thanks! I was thinking about similar, I think this is a good simplification.

Should the PR description be more specific, though? e.g. something such as "Add PushState overload to combine Parse::State change with StateStackEntry" might make it easy to understand what's changing without opening the PR.

josh11b:
> Thanks! I was thinking about similar, I think this is a good simplification.
> 
> Should the PR description be more specific, though? e.g. something such as "Add PushState overload to combine Parse::State change with StateStackEntry" might make it easy to understand what's changing without opening the PR.

Done.

PR #3486, opened 2023-12-08
Refactor NodeKind to take advantage of a parse node only having one token kind.

jonmeow:
This builds on the series of changes to NodeKinds, aiming to simplify the NodeKind implementation a little, also making it clearer that there's a single associated token for each parse node (or, for placeholders/invalid parses, not validated).

Note that prior to the relevant changes, there were nodes with multiple tokens. This change is also locking in the approach of one token per parse node, by refactoring macros to stop supporting multiple.

### DIFF
--- a/toolchain/parse/node_kind.def
+++ b/toolchain/parse/node_kind.def
@@ -29,7 +29,7 @@
 //
 // LexTokenKind indicates the token kind required on a valid node.
 // It will be either `kind_name` to indicate it's required even if the node is
-// an error, or `CARBON_ONLY_IF_VALID(kind_name)` to indicate that the kind is
+// an error, or `CARBON_IF_VALID(kind_name)` to indicate that the kind is
 // not enforced on error. A token kind of `Error` may be used if the kind should
 // never be enforced; this should only be used when the node is either not in
 // trees (`Placeholder`) or is always invalid (such as `InvalidParse`).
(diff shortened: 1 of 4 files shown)


jonmeow:
(also a change with dependencies)

toolchain/parse/node_kind.def
```diff
@@ -94,57 +93,51 @@
...
-CARBON_PARSE_NODE_KIND_CHILD_COUNT(FileStart, 0, CARBON_TOKEN(FileStart))
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(FileStart, 0, FileStart)
 
 // The end of the file.
-CARBON_PARSE_NODE_KIND_CHILD_COUNT(FileEnd, 0, CARBON_TOKEN(FileEnd))
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(FileEnd, 0, FileEnd)
 
 // An invalid parse. Used to balance the parse tree. Always has an error.
-CARBON_PARSE_NODE_KIND_CHILD_COUNT(InvalidParse, 0, CARBON_ANY_TOKEN_ON_ERROR)
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(InvalidParse, 0, CARBON_ONLY_IF_VALID(Error))
```
zygoloid: This is confusing :)

Can we drop the `CARBON_ONLY_IF_VALID` here?

toolchain/parse/node_kind.def
```diff
@@ -94,57 +93,51 @@
...
-// valid, as a special-case in node_kind.cpp.
-CARBON_PARSE_NODE_KIND_CHILD_COUNT(Placeholder, 0, CARBON_ANY_TOKEN_ON_ERROR)
+// A placeholder node to be replaced; it will never exist in a valid parse tree.
+// Its token kind is not enforced even when valid.
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(Placeholder, 0, CARBON_ONLY_IF_VALID(Error))
 
 // An empty declaration, such as `;`.
-CARBON_PARSE_NODE_KIND_CHILD_COUNT(EmptyDecl, 0,
-                                   CARBON_TOKEN(Semi) CARBON_ANY_TOKEN_ON_ERROR)
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(EmptyDecl, 0, CARBON_ONLY_IF_VALID(Semi))
```
zygoloid: I'd find this more readable without the "ONLY" in the macro name. To me, this reads as "the parse node will only correspond to a semicolon if it's valid (and otherwise will not be a semicolon)", not "the parse node will correspond to a semicolon, but that's only guaranteed if it's valid (and otherwise nothing is guaranteed)".

zygoloid (approved):

toolchain/parse/node_kind.def
jonmeow: Done.

toolchain/parse/node_kind.def
jonmeow: Done.

PR #3487, opened 2023-12-09
Support derived to base conversions.

zygoloid:
This is enough to support calling methods that take a `Base` or `Base*` as their `self`. But name lookup doesn't look in the base class yet, so base class methods aren't actually found.


toolchain/check/testdata/class/derived_to_base.carbon
```diff
@@ -0,0 +1,193 @@
...
+}
+
+fn ConvertRef(c: C*) -> A* {
+  return &(*c as A);
+}
+
+fn ConvertInit() {
+  let a: A = {.base = {.base = {.a = 1}, .b = 2}, .c = 3} as C;
+}
+
```
jonmeow, line 36: I think the code works for it, but should there be some tests of non-class-type bases? e.g.:

```suggestion

class StructBase {
  extend base: {.x: i32};
}

fn ConvertStructBaseValue(v: StructBase) {
  let s: {.x: i32} = v;
}

fn ConvertStructBase(p: StructBase*) -> {.x: i32}* { return p; }
```

jonmeow (approved):
Looks good.

toolchain/check/testdata/class/derived_to_base.carbon
zygoloid: Done. Added to `fail_base_bad_type.carbon` because we don't allow such inheritance at all at the moment, but it seems good to make sure that it at least doesn't crash.

jonmeow: LGTM, for some reason I was thinking it was allowed.

PR #3489, opened 2023-12-09
Make tree's operators constexpr noexcept

urielgarciarivas:
Make tree's operators constexpr noexcept.

### DIFF
--- a/toolchain/parse/tree.h
+++ b/toolchain/parse/tree.h
@@ -85,7 +85,7 @@ class Tree : public Printable<Tree> {
                     llvm::raw_ostream* vlog_stream) -> Tree;
 
   // Tests whether there are any errors in the parse tree.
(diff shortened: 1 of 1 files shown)


jonmeow:
Regarding `noexcept`: We should be using `-fno-exceptions`, so `noexcept` won't have any effect in most cases. There are still going to be some performance-related uses, for example, [move constructors](https://clang.llvm.org/extra/clang-tidy/checks/performance/noexcept-move-constructor.html).

Regarding `constexpr`: We should only use `constexpr` where needed for compile-time evaluation. Here you're adding it to operators which aren't used at compile-time, so it could actually be misleading about how the classes are used.

There was some discussion about this on [#toolchain](https://discord.com/channels/655572317891461132/655578254970716160/1182936369752965260), which may be worth flipping through. If you have questions, Discord is really the best place to discuss potential toolchain work right now.

PR #3491, opened 2023-12-09
Move add invalid node. Follow up to #3473.

josh11b:
I think this matches expectations about how the parse code should be structured better.


jonmeow (approved):
Oh, yeah. This explains for me what's going on, thanks!

PR #3493, opened 2023-12-11
Change return of `SkipPastLikelyEnd` to be last consumed token

josh11b:
This approach means the parse subtree includes the full token range consumed.


toolchain/parse/handle_expr.cpp
```diff
@@ -401,14 +401,9 @@ auto HandleExprStatementFinish(Context& context) -> void {
...
-  if (auto semi_token = context.SkipPastLikelyEnd(state.token)) {
-    context.AddNode(NodeKind::ExprStatement, *semi_token, state.subtree_start,
-                    /*has_error=*/true);
-    return;
-  }
-
-  // Found junk not even followed by a `;`, no node to add.
-  context.ReturnErrorOnState();
+  auto last_consumed_token = context.SkipPastLikelyEnd(state.token);
+  context.AddNode(NodeKind::ExprStatement, last_consumed_token,
```
jonmeow:
```suggestion
  context.AddNode(NodeKind::ExprStatement, context.SkipPastLikelyEnd(state.token),
```

Maybe inline the variable in cases like this (also in other files)

jonmeow (approved):
LGTM, but leaving it to merge based on your #toolchain question.

zygoloid (approved):
I think this is an improvement. The code is cleaner and I think the new parse tree is no worse than it was before -- and may be better because the token ranges of the parse nodes now more accurately reflect the ranges of tokens we consumed when forming them.

PR #3494, opened 2023-12-11
Teach the parser to not crash on the spaceship `<=>` operator.

chandlerc:
This remains structurally valid although we don't use this operator or have a design for it, this at least fixes a fuzzer-found crash.


toolchain/parse/testdata/operators/spaceship.carbon
```diff
@@ -0,0 +1,38 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+
+fn F() {
+  var a: i32;
+  var b: i32;
+
```
josh11b: Since we are only parsing, this can be cut down
```suggestion
fn F() {
```

josh11b (approved):

PR #3499, opened 2023-12-12
Change contribution_tools.md to refer to toolchain instead of explorer

jonmeow:


chandlerc (commented):
Doh, for some reason I thought this had already been merged when I was sending out #3532 -- sorry about that.

What would work best? Landing this first, and then looking at #3532? Folding this into #3532? Largely happy to do whatever is eaisest here, no strong opinion.

chandlerc:
@zygoloid for thoughts on how best to land these.

zygoloid:
Sorry about the merge conflicts with #3532. Given that that PR includes the proposal for the change in direction I've gone ahead and merged it, but I think we should take the changes here too (at least the ones that aren't completely overlapping).

jonmeow:
The only non-overlapping change seems to be the one to contribution_tools.md, so I've reverted other work.

zygoloid (approved):

PR #3502, opened 2023-12-13
If a name is not found in a class, perform lookup into base classes.

zygoloid:
This builds out a little infrastructure for one name scope to `extend` another. We'll need more refinement here to cover other cases, but this should provide some foundation for that future work.


toolchain/sem_ir/value_stores.h
```diff
@@ -168,9 +168,22 @@ struct NameScope {
...
+  //
+  // TODO: A `NameScopeId` is not sufficient to describe an extended scope in
+  // general. For example:
+  //
+  //   class A(T:! type) {
+  //     extend base: B(T*);
+  //   }
+  //
+  // needs to describe the `T*` argument.
+  llvm::SmallVector<NameScopeId, 1> extended_scopes;
```
jonmeow, line 187: Please add a comment explaining why `1`. Also, maybe a TODO that we should consider doing something TinyPtrVector-like for IDs.

toolchain/sem_ir/value_stores.h
```diff
@@ -168,9 +168,22 @@ struct NameScope {
   // Names in the scope.
   llvm::DenseMap<NameId, InstId> names;
 
-  // Whether the scope corresponds to an import that failed. There may still be
-  // names from successful imports, or the current file.
-  bool has_load_error = false;
+  // Scopes extended by this scope.
+  //
+  // TODO: A `NameScopeId` is not sufficient to describe an extended scope in
+  // general. For example:
```
jonmeow: One possibility here is that NameScopeId is extended to address generic arguments, either by providing a name scope for them or a variant of name scope that includes the parameter and points at the base name scope.

(i.e.,`NameScopeId` might still be the right choice since we do need to support generic name scopes in general)

toolchain/sem_ir/value_stores.h
```diff
@@ -168,9 +168,22 @@ struct NameScope {
...
+  //
+  //   class A(T:! type) {
+  //     extend base: B(T*);
+  //   }
+  //
+  // needs to describe the `T*` argument.
+  llvm::SmallVector<NameScopeId, 1> extended_scopes;
+
+  // Whether the scope is incomplete, for example due to an import that failed
```
jonmeow: FWIW I'd been trying to avoid "incomplete" due to the mismatched meaning for "incomplete types"

toolchain/diagnostics/diagnostic_kind.def
```diff
@@ -174,6 +174,7 @@ CARBON_DIAGNOSTIC_KIND(DerefOfNonPointer)
 CARBON_DIAGNOSTIC_KIND(DerefOfType)
 CARBON_DIAGNOSTIC_KIND(FunctionPreviousDefinition)
 CARBON_DIAGNOSTIC_KIND(FunctionRedefinition)
+CARBON_DIAGNOSTIC_KIND(NameAmbiguousViaExtend)
```
```suggestion
CARBON_DIAGNOSTIC_KIND(NameAmbiguousDueToExtend)
```

Suggestion, mostly due to the advice to avoid "via" in documentation (https://developers.google.com/style/word-list#via). I don't generally comment on `e.g.` or `i.e.` in comments because I think they're common, but "via" seems like it's rarer and thus more likely to cause a stumble in understandability. The diagnostic codes are also somewhat of an API too, so clarity a little more valuable.

toolchain/check/check.cpp
```diff
@@ -96,7 +96,7 @@ static auto InitPackageScopeAndImports(Context& context, UnitInfo& unit_info)
 
       // If an import of the current package caused an error for the imported
       // file, it transitively affects the current file too.
-      package_scope.has_load_error |= import_scope.has_load_error;
+      package_scope.has_error |= import_scope.has_error;
```
jonmeow, line 99: To just note, it might've been helpful to split this rename out to a separate PR for easier review -- making it clearer which parts of this change affected what tests.

toolchain/check/handle_class.cpp
```diff
@@ -221,21 +221,27 @@ auto HandleBaseDecl(Context& context, Parse::NodeId parse_node) -> bool {
...
     // TODO: Once we have a better idea of which types are considered to be
     // classes, produce a better diagnostic for deriving from a non-class type.
     auto base_class = context.types().TryGetAs<SemIR::ClassType>(base_type_id);
-    if (!base_class ||
-        context.classes().Get(base_class->class_id).inheritance_kind ==
-            SemIR::Class::Final) {
+    auto* base_class_info =
+        base_class ? &context.classes().Get(base_class->class_id) : nullptr;
+    if (!base_class_info ||
+        base_class_info->inheritance_kind == SemIR::Class::Final) {
```
jonmeow: This took me a few reads to read correctly. What would you think of something like a `GetBaseScopeId` helper function in order to reduce the flow control paths? Alternately, what about swapping the if/else so that you use base_scope_info next to where you verify it's non-null? I'd prefer the former though, because I think it'd be easier to read -- e.g., I think you could get something like:

```
if (base_type_id == SemIR::TypeId::Error) {
  return SemIR::NameScopeId::Invalid;
}
auto base_class = context.types().TryGetAs<SemIR::ClassType>(base_type_id);
if (base_class) {
  auto base_class_info = context.classes().Get(base_class->class_id);
  if (base_class_info.inheritance_kind == SemIR::Class::Final) {
    CARBON_CHECK(base_class_info->scope_id.is_valid())
        << "Complete class should have a scope";
    return base_class_info.scope_id;
  }
}
// Emit diagnostic
return SemIR::NameScopeId::Invalid;
```

The nullness of base_class_info goes away in that approach.

toolchain/sem_ir/value_stores.h
jonmeow, line 187: Is it worth noting this vector is examined in reverse order? (I assume more entries will be added down the road, although it's just one right now?)

toolchain/check/context.cpp
```diff
@@ -296,26 +303,59 @@ auto Context::LookupUnqualifiedName(Parse::NodeId parse_node,
...
+      continue;
+    }
 
-  // TODO: Also perform lookups into `extend`ed scopes.
+    // If this is our second lookup result, diagnose an ambiguity.
+    if (result_id.is_valid()) {
+      CARBON_DIAGNOSTIC(
+          NameAmbiguousViaExtend, Error,
+          "Ambiguous use of name `{0}` found in multiple extended scopes.",
+          std::string);
```
jonmeow, line 345: Please add a test for this, or if nothing can reach it today, either a TODO to remind that it needs to be tested, or maybe a CHECK?

jonmeow (approved):
Approving -- I think tests are important to add, but I imagine the code's mostly working as intended right now so it's mostly a documentation question.

toolchain/check/context.cpp
```diff
@@ -296,26 +303,59 @@ auto Context::LookupUnqualifiedName(Parse::NodeId parse_node,
...
 
-  if (!required) {
-    return SemIR::InstId::Invalid;
+    result_id = scope_result_id;
   }
-  if (!scope.has_load_error) {
-    DiagnoseNameNotFound(parse_node, name_id);
+
+  if (required && !result_id.is_valid()) {
+    if (!has_error) {
```
jonmeow, line 356: Is *this* tested? I wasn't seeing it in the added tests. e.g.:

```
class C {
  extend base: A;
}

var x: i32 = C.v;
```

toolchain/sem_ir/value_stores.h
zygoloid: Yeah. I think we might want two layers here: a (nominal?) scope layer that has a name lookup table and a list of extended scopes, plus a (semantic?) scope layer that holds something like a reference to a nominal scope plus a list of generic arguments.

Did you want a change to the TODO? My thinking is that the *current* meaning of `NameScopeId` is insufficient, so we need to make *some* change here.

toolchain/check/handle_class.cpp
zygoloid: Refactored a bit, with an early exit for the non-class case; hopefully the logic is clearer now. I also changed error recovery here a bit to retain the information about the base class even if the base class is `final`.

toolchain/check/check.cpp
zygoloid: Noted for future.

toolchain/check/context.cpp
zygoloid: Added a TODO to add test coverage once this is reachable.

toolchain/check/context.cpp
zygoloid: Test added.

toolchain/diagnostics/diagnostic_kind.def
zygoloid: Done.

toolchain/sem_ir/value_stores.h
zygoloid: Done.

zygoloid: Thanks, nice catch. The order shouldn't affect whether name lookup succeeds, but it would affect which declarations we point to in ambiguity diagnostics (once we start producing ambiguity diagnostics...). Changed `LookupQualifiedName` to reverse these when they're pushed onto the queue. I think that'll mean we walk extended scopes in an unsurprising depth-first top-to-bottom preorder.

toolchain/sem_ir/value_stores.h
zygoloid: Seems like a good idea. Comment rewritten.

toolchain/check/handle_class.cpp
```diff
@@ -174,6 +174,77 @@ auto HandleBaseColon(Context& /*context*/, Parse::NodeId /*parse_node*/)
...
+// Information gathered about a base type specified in a `base` declaration.
+struct BaseInfo {
+  SemIR::TypeId type_id;
+  SemIR::NameScopeId scope_id;
+
+  // A `BaseInfo` representing an erroneous base.
+  static const BaseInfo Error;
+};
+constexpr BaseInfo BaseInfo::Error = {.type_id = SemIR::TypeId::Error,
+                                      .scope_id = SemIR::NameScopeId::Invalid};
```
jonmeow: Declaration order: static constants come early. FWIW you could also just make this a non-member.

```suggestion
// Information gathered about a base type specified in a `base` declaration.
struct BaseInfo {
  // A `BaseInfo` representing an erroneous base.
  static const BaseInfo Error;

  SemIR::TypeId type_id;
  SemIR::NameScopeId scope_id;
};
constexpr BaseInfo BaseInfo::Error = {.type_id = SemIR::TypeId::Error,
                                      .scope_id = SemIR::NameScopeId::Invalid};
```

```suggestion
// Information gathered about a base type specified in a `base` declaration.
struct BaseInfo {
  SemIR::TypeId type_id;
  SemIR::NameScopeId scope_id;
};
// A `BaseInfo` representing an erroneous base.
static constexpr BaseInfo BaseInfoError = {
    .type_id = SemIR::TypeId::Error, .scope_id = SemIR::NameScopeId::Invalid};
```

jonmeow (approved):
Looks good, thanks!

toolchain/sem_ir/value_stores.h
jonmeow: I mean, if you change anything, I'd just say "currently". But I get the intent, and maybe just happened to read it in an overly prescriptive way.

```suggestion
  // TODO: A `NameScopeId` is currently insufficient to describe an extended scope in
  // general. For example:
```

PR #3505, opened 2023-12-13
Start adapting to bzlmod configurations.

jonmeow:
Some background information is at https://bazel.build/external/migration

Trying to handle the simple cases first. This adds a requirement for bazel 7 due to differences in bzlmod handling between 6 and 7 (also discussed on [#infra](https://discord.com/channels/655572317891461132/707150492370862090/1184942191412510720)). Bazel seems to be okay with a partial migration such as ths.

The python import behavior has subtly shifted, so `carbon.` is no longer part of import paths. There's a version-incompatible change for `@@`. bzlmod makes repos sometimes show as `name~version`.

`target-determinator` seems to be okay with `@@` after a version update.

Things not moved here are things that basically need more dep work:

- clang_register_toolchains because I need to dive into its format.
- llvm-project because we need something slightly atypical, I need to make sure patching and the repo work carries over.
- com_google_libprotobuf_mutator is sufficiently atypical that it doesn't have a module already, but should be one of the easier things to fix.
- brotli/woff2: I think we should actually consider removing these. But again, they're not trivial moves.
- treesitter due to toolchain registration, which has shifted a bit.
  - rules_nodejs because treesitter depends on it in an awkward way to migrate.

### DIFF
--- a/bazel/check_deps/check_non_test_cc_deps.py
+++ b/bazel/check_deps/check_non_test_cc_deps.py
@@ -22,9 +22,7 @@ import sys
 from pathlib import Path
 
 runfiles = Path(os.environ["TEST_SRCDIR"])
-deps_path = (
-    runfiles / "carbon" / "bazel" / "check_deps" / "non_test_cc_deps.txt"
-)
+deps_path = runfiles / "_main" / "bazel" / "check_deps" / "non_test_cc_deps.txt"
 try:
     with deps_path.open() as deps_file:
         deps = deps_file.read().splitlines()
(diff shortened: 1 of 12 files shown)


bazel/check_deps/check_non_test_cc_deps.py
```diff
@@ -22,9 +22,7 @@
 from pathlib import Path
 
 runfiles = Path(os.environ["TEST_SRCDIR"])
-deps_path = (
-    runfiles / "carbon" / "bazel" / "check_deps" / "non_test_cc_deps.txt"
-)
+deps_path = runfiles / "_main" / "bazel" / "check_deps" / "non_test_cc_deps.txt"
```
zygoloid, line 25: Should we also remove the `bazel-execroot` symlink, or update it to point to `_main`?

zygoloid (approved):
LGTM, thank you.

bazel/check_deps/check_non_test_cc_deps.py
jonmeow: Updated for _main

PR #3506, opened 2023-12-14
Remove `SelfParam`, add an `AddrPattern` instead.

zygoloid:
This is intended to make the representation of a `self` pattern be more similar to other patterns.


toolchain/check/handle_binding_pattern.cpp
```diff
@@ -35,28 +39,21 @@ auto HandleBindingPattern(Context& context, Parse::NodeId parse_node) -> bool {
...
-
   // TODO: Handle `_` bindings.
 
   // Every other kind of pattern binding has a name.
   auto [name_node, name_id] = context.node_stack().PopNameWithParseNode();
 
+  // A `self` binding can only appear in an implicit parameter list.
+  if (name_id == SemIR::NameId::SelfValue &&
+      context.parse_tree().node_kind(context.node_stack().PeekParseNode()) !=
+          Parse::NodeKind::ImplicitParamListStart) {
```
jonmeow:
```suggestion
      context.PeekIs<Parse::NodeKind::ImplicitParamListStart>()) {
```

Looks pre-existing, but feels worth cleaning up while you're changing the code. Should probably do a pass trying to clean up things like this.

toolchain/check/handle_name.cpp
```diff
@@ -72,6 +72,26 @@ static auto GetClassElementIndex(Context& context, SemIR::InstId element_id)
...
+                             SemIR::FunctionId function_id) -> bool {
+  auto& function = sem_ir.functions().Get(function_id);
+  for (auto param_id :
+       sem_ir.inst_blocks().Get(function.implicit_param_refs_id)) {
+    auto inst = sem_ir.insts().Get(param_id);
+    if (auto addr = inst.TryAs<SemIR::AddrPattern>()) {
+      inst = sem_ir.insts().Get(addr->inner_id);
+    }
+    if (auto param = inst.TryAs<SemIR::Param>();
+        param && param->name_id == SemIR::NameId::SelfValue) {
```
jonmeow, line 87: This code matches what's in convert.cpp, differing only on what's done with `param`. Not sure it's worth a like `TryGetAsSelfValueParam` or similar, but might be worth considering.

toolchain/lower/handle.cpp
```diff
@@ -30,6 +30,11 @@ auto HandleAddressOf(FunctionContext& context, SemIR::InstId inst_id,
   context.SetLocal(inst_id, context.GetValue(inst.lvalue_id));
 }
 
+auto HandleAddrPattern(FunctionContext& /*context*/, SemIR::InstId /*inst_id*/,
+                       SemIR::AddrPattern /*inst*/) -> void {
+  CARBON_FATAL() << "`addr` should be lowered by `BuildFunctionDefinition`";
+}
```
jonmeow, line 36: Is it worth moving all the fatal handlers to their own file, like `handle_unused`?

toolchain/sem_ir/formatter.cpp
```diff
@@ -447,6 +447,12 @@ class InstNamer {
       };
 
       switch (inst.kind()) {
+        case AddrPattern::Kind: {
+          // TODO: Visit the entire parameter block, not just the parameter
+          // refs, so that we visit the parameter inside an `addr` pattern.
```
jonmeow: Is this an issue where patterns generally need to be recursed through? I'll offer some concern that walking the entire parameter block might not yield the right result either, due to separate use of blocks.

jonmeow (approved):
Looks good!

toolchain/lower/handle.cpp
zygoloid: Sure, will send a PR.

toolchain/sem_ir/formatter.cpp
zygoloid: Switched TODO to just describe the problem rather than suggesting a solution.

toolchain/check/handle_name.cpp
zygoloid: Maybe we can take another look once `ConvertCallArgs` is also handling deduced parameters. They might look less similar at that point.

PR #3508, opened 2023-12-14
Underline the complete declaration in diagnostics

josh11b:
Builds upon @domisterwoozy 's excellent #3442 . Removes the need to store the first node of a declaration in the declaration state stack.


zygoloid (approved):
Nice!

PR #3496, opened 2023-12-11
Update cc_toolchain configuration to be compatible with bazel 7.

zygoloid:

PR #3514, opened 2023-12-15
Move toolchain and llvm WORKSPACE rules to bzlmod

jonmeow:
Building on #3505, the toolchain and llvm rules require a little more special-casing to get them to work well. This also moves libprotobuffer_mutator, but that one's more minor. The migration encounters more quirks in repo naming as seen by various queries.

This changes some of the toolchain work that was recently done for bazel 7 in #3496, dropping a bzl file I'd suggested to add, instead using `:all` for toolchain registration. (somewhat as an improvement, somewhat just to avoid a `load`)

Remaining in the WORKSPACE are example code repos and tree sitter rules. Neither of these are part of the main toolchain builds, and so will probably be lower impact if there's a good solution for them.


MODULE.bazel
```diff
@@ -38,6 +43,79 @@ bazel_dep(
...
+    urls = ["https://github.com/google/libprotobuf-mutator/archive/v{0}.tar.gz".format(libprotobuf_mutator_version)],
+)
+
+bazel_cc_toolchain = use_extension(
+    "//bazel/cc_toolchains:clang_configuration.bzl",
+    "clang_toolchain_extension",
+)
+use_repo(bazel_cc_toolchain, "bazel_cc_toolchain")
+
+register_toolchains("@bazel_cc_toolchain//:all")
```
zygoloid, line 62: Pre-existing, please feel free to ignore: The `bazel_` prefix here seems redundant to me. `local_` or `clang_` would be more useful as a prefix I think.

```suggestion
clang_cc_toolchain = use_extension(
    "//bazel/cc_toolchains:clang_configuration.bzl",
    "clang_toolchain_extension",
)
use_repo(clang_cc_toolchain, "clang_cc_toolchain")

register_toolchains("@clang_cc_toolchain//:all")
```

zygoloid, line 62: Nice! I didn't know that `register_toolchains` supported this.

MODULE.bazel
```diff
@@ -38,6 +43,79 @@ bazel_dep(
...
+    build_file = "@llvm-raw//utils/bazel/third_party_build:zlib-ng.BUILD",
+    sha256 = "e36bb346c00472a1f9ff2a0a4643e590a254be6379da7cddd9daeb9a7f296731",
+    strip_prefix = "zlib-ng-2.0.7",
+    urls = [
+        "https://github.com/zlib-ng/zlib-ng/archive/refs/tags/2.0.7.zip",
+    ],
+)
+
+http_archive(
+    name = "llvm_zstd",
```
zygoloid: Do we not need the `maybe` any more? (I don't know what it was for.)

zygoloid (approved):

MODULE.bazel
jonmeow: It's supposed to only add the repo if not already present. AFAICT it's incompatible with bzlmod and not needed.

MODULE.bazel
jonmeow: I'd taken `bazel_` here to correspond to the directory that this is within, with a small mismatch in pluralization. My leaning would be to leave the path correspondence, possibly removing the prefix and making it just `cc_toolchains`, versus adding a difference prefix. But maybe that's best for a separate PR to maintain consistency here since this affects a few different files?

zygoloid: Sure.

jonmeow:
FYI I'm trying to work around the target-determinator conflict:

```
2023/12/16 00:29:12 failed to process change: failed to check out original commit during cleanup: <nil>
```

(unfortunately, there's very little info about why it's breaking)

PR #3515, opened 2023-12-15
Simplify handling of `VariableInitializer`

geoffromer:
Also stop supporting `var` with initializer inside `for`.

Resolves TODO in `handle_variable.cpp`


toolchain/check/handle_variable.cpp
```diff
@@ -26,24 +26,18 @@ auto HandleReturnedModifier(Context& context, Parse::NodeId parse_node)
...
-    return context.TODO(parse_node, "tuple pattern in var");
-  }
-  // TODO: find a more robust way to determine if there was an initializer.
-  bool has_init = next_kind != Parse::NodeKind::BindingPattern;
+  bool has_init = next_kind == Parse::NodeKind::VariableInitializer;
   if (has_init) {
     init_id = context.node_stack().PopExpr();
-    context.node_stack()
-        .PopAndDiscardSoloParseNode<Parse::NodeKind::VariableInitializer>();
   }
```
zygoloid:
```suggestion
  auto init_id = context.node_stack().PopIf<Parse::NodeKind::VariableInitializer>();
```

toolchain/parse/state.def
```diff
@@ -1154,17 +1154,27 @@ CARBON_PARSE_STATE(ImplBeforeAs)
...
+//
+// var ...             (variant is For)
+//    ^
+//   1. BindingPattern
+//   2. VarFinishAsFor
+//
+// var (...)           (variant is For)
+//    ^
+//   1. ParamListAsRegular
+//   2. VarFinishAsFor
```
zygoloid: These comments don't quite match the code: this state uses the `Pattern` state rather than dispatching on the `(` versus no-`(` itself.

toolchain/check/handle_variable.cpp
geoffromer: I'd assumed this code was intentionally avoiding `std::optional` for performance reasons, but that works for me!

toolchain/parse/state.def
geoffromer: Whoops, you're right. Looks like an incomplete edit from #3448.

toolchain/parse/state.def
```diff
@@ -1215,14 +1215,14 @@ CARBON_PARSE_STATE_VARIANTS2(VarFinish, Decl, For)
 // Handles the start of a `let`.
 //
 // let (...)
-// ^~~
-//   1. ParamListAsRegular
+//    ^
+//   1. Pattern
 //   2. LetAfterPattern
 //   3. LetFinish
 //
```
zygoloid: Looks like the first half of this comment can be removed.
```suggestion
```

zygoloid (approved):
Thanks, looks great!

toolchain/check/handle_variable.cpp
zygoloid: We could alternatively use `.value_or(SemIR::InstId::Invalid)` here but I doubt it'll make any real difference.

zygoloid (approved):

geoffromer:
I had to do a non-trivial merge, so this could probably use a fresh look.

toolchain/parse/handle_var.cpp
```diff
@@ -47,7 +47,13 @@ auto HandleVarAsReturned(Context& context) -> void {
 }
 
 auto HandleVarAsFor(Context& context) -> void {
-  HandleVar(context, State::VarFinishAsFor);
+  auto state = context.PopState();
+
+  // The finished variable declaration will start at the `var`.
+  state.state = State::VarFinishAsFor;
+  context.PushState(state);
```
zygoloid: Looks like we have a simpler way to write this now:
```suggestion
  context.PushState(state, State::VarFinishAsFor);
```

zygoloid (approved):

PR #3516, opened 2023-12-15
Correct accepted-proposal link

geoffromer:
It looks like we haven't used the "proposal accepted" label in about a year.


zygoloid (approved):
LG. I wonder if we should either auto-apply the label or remove it.

PR #3517, opened 2023-12-15
Verify bazel mod deps

jonmeow:
Query `bazel mod deps` with `--lockfile_mode=error` to detect out-of-sync lockfiles; print the output to support cross-platform updates.

Add a workaround for target-determinator problems to prevent them from blocking changes (e.g., https://github.com/carbon-language/carbon-lang/actions/runs/7225229412/job/19694000744?pr=3514)

Example verify failure to merge into end result:
https://github.com/carbon-language/carbon-lang/actions/runs/7227335045/job/19694765560?pr=3517

Success after updates:
https://github.com/carbon-language/carbon-lang/actions/runs/7227387163/job/19694933047

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -203,6 +203,22 @@ jobs:
         if: steps.filter.outputs.has_code == 'true'
         run: df -h
 
+      - name: Verify MODULE.bazel.lock
+        if: steps.filter.outputs.has_code == 'true'
+        run: |
+          exit_code=0
+          bazelisk mod deps --lockfile_mode=error || exit_code=$?
+          if (( $exit_code != 0 )); then
+            bazelisk mod deps --lockfile_mode=update
+            echo "MODULE.bazel.lock is out of date! Use below file for update."
+            echo "Platforms may require merging output, for example by applying"
+            echo "an update, re-running triggers, and applying the next update."
+            echo "============================================================"
+            cat MODULE.bazel.lock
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yaml
```diff
@@ -203,6 +203,22 @@ jobs:
...
+          exit_code=0
+          bazelisk mod deps --lockfile_mode=error || exit_code=$?
+          if (( $exit_code != 0 )); then
+            bazelisk mod deps --lockfile_mode=update
+            echo "MODULE.bazel.lock is out of date! Use below file for update."
+            echo "Platforms may require merging output, for example by applying"
+            echo "an update, re-running triggers, and applying the next update."
+            echo "============================================================"
+            cat MODULE.bazel.lock
+            echo "============================================================"
```
zygoloid, line 218:
```suggestion
            echo "MODULE.bazel.lock is out of date! Use below patch for update."
            echo "You may also need to apply patches for other platforms."
            echo "============================================================"
            git diff MODULE.bazel.lock
            echo "============================================================"
```
... maybe?

jonmeow: I'd started with something similar to that, but I was trying to clarify what it meant to apply patches for other platforms. It's really just copy-pasting the whole file, overwriting the old one.

[to be clear: I consider a patch to be a delta, wheras this is a full file]

zygoloid (approved):

PR #3518, opened 2023-12-15
Add a fallback for target-determinator failures

jonmeow:
Add a workaround for target-determinator problems to prevent them from blocking changes (e.g., https://github.com/carbon-language/carbon-lang/actions/runs/7225229412/job/19694000744?pr=3514)

I was trying to do this in #3517, but that fails due to a zstd issue. #3514 changes how zstd is set up, but fails due to this issue. So now I'm trying to separate out this workaround to see if things come together.


jonmeow:
Realizing this may not do what I want. :(

PR #3519, opened 2023-12-18
`startswith` -> `starts_with` to match upstream

josh11b:
This matches the change to upstream `llvm::StringRef`, see https://github.com/llvm/llvm-project/pull/75491 .


jonmeow (approved):
Thanks!

PR #3522, opened 2023-12-18
Rename `Carbon::StringLiteralId` -> `Carbon::StringLiteralValueId`.

zygoloid:
We have `StringLiteral`s in multiple other `Carbon` sub-namespaces. Rename to a more specific name to avoid collisions.

We should likely also rename `Carbon::IntId` -> `Carbon::IntValueId` and `Carbon::RealId` -> `Carbon::RealValueId`, but this collision is prioritized because it was blocking work on typed parse nodes which introduces a `Carbon::Parse::StringLiteralId`.

### DIFF
--- a/toolchain/base/value_store.h
+++ b/toolchain/base/value_store.h
@@ -97,17 +97,18 @@ struct IdentifierId : public IdBase, public Printable<IdentifierId> {
 };
 constexpr IdentifierId IdentifierId::Invalid(IdentifierId::InvalidIndex);
 
-// Adapts StringId for string literals.
-struct StringLiteralId : public IdBase, public Printable<StringLiteralId> {
-  static const StringLiteralId Invalid;
+// Adapts StringId for values of string literals.
+struct StringLiteralValueId : public IdBase,
+                              public Printable<StringLiteralValueId> {
+  static const StringLiteralValueId Invalid;
   using IdBase::IdBase;
   auto Print(llvm::raw_ostream& out) const -> void {
     out << "strLit";
     IdBase::Print(out);
   }
 };
-constexpr StringLiteralId StringLiteralId::Invalid(
-    StringLiteralId::InvalidIndex);
+constexpr StringLiteralValueId StringLiteralValueId::Invalid(
+    StringLiteralValueId::InvalidIndex);
 
 namespace Internal {
 // Used as a parent class for non-printable types. This is just for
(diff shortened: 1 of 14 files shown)


josh11b (approved):

toolchain/sem_ir/file.h
```diff
@@ -225,10 +225,11 @@ class File : public Printable<File> {
   auto reals() const -> const ValueStore<RealId>& {
     return value_stores_->reals();
   }
-  auto string_literals() -> StringStoreWrapper<StringLiteralId>& {
+  auto string_literals() -> StringStoreWrapper<StringLiteralValueId>& {
```
jonmeow: While I'm hesitant regarding renaming this entity versus others (wondering if this should get precedent for a shorter name since it's used cross-phase), I do think this should now be `string_literal_values` because you're reserving "string literal" to mean other things.

toolchain/base/value_store.h
```diff
@@ -257,10 +258,11 @@ class SharedValueStores : public Yaml::Printable<SharedValueStores> {
   auto ints() const -> const ValueStore<IntId>& { return ints_; }
   auto reals() -> ValueStore<RealId>& { return reals_; }
   auto reals() const -> const ValueStore<RealId>& { return reals_; }
-  auto string_literals() -> StringStoreWrapper<StringLiteralId>& {
+  auto string_literals() -> StringStoreWrapper<StringLiteralValueId>& {
     return string_literals_;
   }
-  auto string_literals() const -> const StringStoreWrapper<StringLiteralId>& {
+  auto string_literals() const
```
jonmeow: I'd make a similar comment here, for string_literal_values. Should IdentifierId also be IdentifierValueId, etc?

toolchain/sem_ir/file.h
zygoloid: Done. (@josh11b and I had discussed this and we didn't feel strongly either way, but happy to make the change since it sounds like you do.)

jonmeow (approved):

toolchain/sem_ir/file.h
jonmeow: Thanks -- it makes me sigh a little if we're headed in this direction, but I think if "Value" needs to be appended to all the types then it should also be appended to the stores.

toolchain/base/value_store.h
zygoloid: My thinking is that a "string literal" is a kind of token and a kind of expression, and so we have `Lex::TokenKind::StringLiteral`, `Parse::NodeKind::StringLiteral` and `SemIR::StringLiteral`, and we're about to gain a `Parse::StringLiteral` and a `Parse::StringLiteralId`, all representing that token / expression. What we're representing here is the *value* of the string literal expression, not the string literal expression / token itself.

On the other hand, we use plain "identifier" to refer only to the lexical element; the parse node is an `IdentifierName` or similar, and the semantic node is a `NameRef` or similar, so using `IdentifierId` to refer to the identity of the identifier seems reasonable. And I think the "value" of an identifier could be interpreted as meaning the value of the entity named by that identifier, rather than the name itself.

jonmeow: To the extent that "Identifier" is the token too, I'm a bit confused why that conflict isn't an issue.

To a point though, we also do IntegerId for the IntegerLiteral token. Maybe this should just be StringId for a StringLiteral token?

PR #3524, opened 2023-12-18
Add comments explaining module lock updates.

jonmeow:


MODULE.bazel
```diff
@@ -2,7 +2,16 @@
 # Exceptions. See /LICENSE for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-"""Bazel modules."""
+"""Bazel modules.
+
+If `MODULE.bazel.lock` changes locally, it means the host platform is not part
+of the GitHub supported set. Create a new PR with the result of
+`bazel mod deps`.
```
zygoloid: This isn't enough information for me to be able to follow these instructions. What would I do with the result of `bazel mod deps`?

MODULE.bazel
```diff
@@ -2,7 +2,16 @@
...
+
+Normally, test actions are used to require updates on platforms that GitHub will
+run tests on. Right now that's linux amd64 and osx x86_64. Because other
+platforms cannot be validated, they may fall out of sync on changes.
```
zygoloid:
```suggestion
platforms are not routinely validated, they may fall out of sync on changes.
```
Presumably, they *could* be, but currently aren't.

MODULE.bazel
jonmeow: Rephrased

MODULE.bazel
jonmeow: Rephrased

MODULE.bazel
```diff
@@ -2,7 +2,18 @@
 # Exceptions. See /LICENSE for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-"""Bazel modules."""
+"""Bazel modules.
+
+If `MODULE.bazel.lock` changes locally, it means the host platform is not part
+of the GitHub supported set. Running `bazel mod deps` provides a canonical
+update to `MODULE.bazel.lock`; create a PR with those changes in order to
+include the host platform.
```
chandlerc: I don't really follow this -- this seems to imply more ties to GitHub than I would expect. For example, we don't have any GitHub support for any Arm platforms. And it would seem odd for users to send PRs expanding it.

Maybe something more like:
```suggestion
If `MODULE.bazel.lock` changes locally, it means the host platform hasn't yet been added to the lock file. Running `bazel mod deps` provides a canonical
update to `MODULE.bazel.lock`; create a PR with those changes in order to
include the host platform.
```

MODULE.bazel
```diff
@@ -2,7 +2,18 @@
...
+
+Normally, test actions are used to require updates on platforms that GitHub will
+run tests on. Right now that's linux amd64 and osx x86_64. Because other
+platforms are not covered by CI, they may fall out of sync on dependency
+changes.
```
chandlerc: I had some trouble understanding this, what do you think about an alternate phrasing?

```suggestion
Platforms tested with GitHub actions are kept up-to-date. Other platforms may fall out of sync on dependency
changes, and should be updated with a PR the same way a platform is added.
```

chandlerc (approved):
LG, as long as @zygoloid is also happy.

PR #3525, opened 2023-12-18
Update pre-commit for bazel mod deps

jonmeow:
I feel kind of weird adding a pre-commit for this, but I think it may make validating #3524 easier.


chandlerc (approved):
LGTM!

PR #3526, opened 2023-12-19
Switch to bazel_dep and normal zlib.

chandlerc:
The use of zlib-ng seemed to work previously, but may have only done so because of some combination of an implementation detail of `strip_include_prefix` in older versions of Bazel and some happenstance of a system installed `zlib.h` being possible to find. With Bazel 7 it started breaking on my machine which doesn't have a system `zlib.h`.

Using the "normal" zlib and the bzlmod rigging for it is simple and seems to work.

Also adds the Linux AArch64 component of the Bazel lock file which was necessary to test this on my Arm Linux machine.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -73,17 +73,7 @@ llvm_project_version = "3d51010a3350660160981c6b8e624dcc87c208a3"
 bazel_dep(name = "platforms", version = "0.0.8")
 
 # zlib and zstd dependencies are copied from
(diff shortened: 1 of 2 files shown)


jonmeow (approved):

jonmeow:
With this fix, we might also want to try removing the zlib package from other notes (e.g. contribution_tools.md) but that can be handled later.

PR #3533, opened 2023-12-19
Try out hedronvision for compile_commands.json

jonmeow:
Directories in `external/` now include versions of dependencies, so continuing to make that work would require version-locking the compile_flags.txt. We've previously seen issues with directories being forgotten; this switches to letting bazel generate everything.

Setup follows instructions at https://github.com/hedronvision/bazel-compile-commands-extractor?tab=readme-ov-file#first-add-this-tool-to-your-bazel-setup

Maintains `./scripts/create_compdb.py`. Per discussion, new files seem to work okay -- I'm getting #include completion that clearly detects files.

Note this creates an `external` directory:

```
╚╡ls -ld external
lrwxrwxrwx 1 jperkins primarygroup 27 Dec 19 14:02 external -> bazel-out/../../../external
```

I felt sort of weird checking that in though, at least alongside `bazel-execroot` it feels like it should have a `bazel-` prefix. However, I can't rename it. So omitting it, and leaving it to the script to generate, felt like a reasonable compromise.

Running takes a few minutes (about 3 for me), and should be faster on reruns. It does print a couple warnings, just tree_sitter missing parser.c and  _GNUC_PREREQ errors from m4.


.gitignore
```diff
@@ -22,7 +22,8 @@
 /user.bazelrc
 
 # Compilation database used by clangd
-compile_commands.json
+/compile_commands.json
+/external
```
chandlerc: Filed an issue to be able to customize the name: https://github.com/hedronvision/bazel-compile-commands-extractor/issues/150

scripts/create_compdb.py
```diff
@@ -23,145 +23,16 @@
...
-    stderr=subprocess.DEVNULL,
-    universal_newlines=True,
-).splitlines()
-print("Found %d generated files..." % (len(generated_file_labels),))
-
-# Directly build these labels so that indexing can find them. Allow this to
-# fail in case there are build errors in the client, and just warn the user
-# that they may be missing generated files.
-print("Building the generated files so that tools can find them...")
-subprocess.check_call([bazel, "build", "--keep_going"] + generated_file_labels)
```
chandlerc, line 121: I think we need to keep building the generated files before running this? At least, the docs seem to indicate you still need to do a successful build to get generated files?

scripts/create_compdb.py
```diff
@@ -23,145 +23,16 @@
...
-scripts_utils.chdir_repo_root()
-directory = Path.cwd()
 
-# We use the `BAZEL` environment variable if present. If not, then we try to
-# use `bazelisk` and then `bazel`.
-bazel = scripts_utils.locate_bazel()
+def main() -> None:
+    scripts_utils.chdir_repo_root()
+    bazel = scripts_utils.locate_bazel()
+    subprocess.run([bazel, "run", "@hedron_compile_commands//:refresh_all"])
```
chandlerc, line 66: Should we hide the output from this without a flag to avoid users being confused by it printing out error messages that aren't actually errors?

.gitignore
jonmeow: Okay

scripts/create_compdb.py
jonmeow: Done, I don't think I was running into issues here because the toolchain doesn't benefit from this.

scripts/create_compdb.py
jonmeow: The output does include messages that build errors might be spurious, and I'm wary of suppressing output that *could* be errors (or even just actionable warnings). Are you sure you want to suppress all error messages?

.gitignore
chandlerc: I think https://github.com/hedronvision/bazel-compile-commands-extractor/issues/150#issuecomment-1863703726 is indicating we can just remove this section, WDYT?

scripts/create_compdb.py
```diff
@@ -23,145 +23,48 @@
...
-]
-llvm_files = [
-    Path("bazel-execroot/external").joinpath(
-        *f.parts[f.parts.index("llvm-project") :]
+
+def _build_generated_files(bazel: str) -> None:
+    print("Building the generated files so that tools can find them...")
+
+    # Now collect the generated file labels. Include some rules which generate
+    # files but aren't classified as "generated file".
```
chandlerc: Tiny comment tweak given new structure.

```suggestion
    # Collect the generated file labels. Include some rules which generate
    # files but aren't classified as "generated file".
```

scripts/create_compdb.py
chandlerc: No, wasn't sure. Let's keep it as is for now, your concerns make sense.

chandlerc (approved):
LGTM, regardless of how you want to handle `.gitignore`.

.gitignore
jonmeow: Removing /external because it's tool-dependent, but I feel weird about removing compile_commands.json since it's pre-existing (and may come from other tools, without ever running the hedronvision approach)

PR #3534, opened 2023-12-20
Typed wrappers around parse tree nodes

josh11b:
These are intended to allow the structure of a parse tree node to be described more precisely in code, to support these use cases:

- Automated checking that the parse tree conforms to the expected structure. (Added to `Tree::Verify`.)
- Easier reading and understanding of the structure of the parse tree by toolchain developers. (See `parse/typed_nodes.h`.)
- Easier navigation of the parse tree, for example for tooling uses and for use when forming diagnostics.

On this last point, an object representing the file may be inspecting using `Tree::ExtractFile`, as in:
```
auto file = tree->ExtractFile();
for (AnyDeclId decl_id : file.decls) {
  // `decl_id` is convertible to a `NodeId`.
  if (std::optional<FunctionDecl> fn_decl =
      tree->ExtractAs<FunctionDecl>(decl_id)) {
    // fn_decl->params is a `TuplePatternId` (which extends `NodeId`)
    // that is guaranteed to reference a `TuplePattern`.
    std::optional<TuplePattern> params = tree->Extract(fn_decl->params);
    // `params` has a value unless there was an error in that node.
  } else if (auto class_def = tree->ExtractAs<ClassDefinition>(decl_id)) {
    // ...
  }
}
```

The `Extract...` functions collect the child nodes into the typed parse node's fields (internally using a `Tree::SiblingIterator`) for easy access. However, this is not as fast as directly observing the tree structure using the postorder strategy being used by the check stage.

These functions rely on using struct reflection on the typed parse node definitions from `parse/typed_nodes.h` to get the expected structure of child nodes and then populate them.

Note that validating these in `Tree::Verify` adds significant cost to it, and is currently included in the parsing stage. Without this change, a 10 mloc test case of lex & parse takes 4.129 s ±  0.041 s. With this change, it takes 5.768 s ±  0.036 s.

This builds upon and completes #3393.

Co-authored-by: Richard Smith <richard@metafoo.co.uk>


toolchain/check/node_stack.h
```diff
@@ -92,6 +92,7 @@ class NodeStack {
   }
 
   // Pops the top of the stack and returns the parse_node.
+  // TODO: return a parse::NodeIdForKind<RequiredParseKind> instead.
```
zygoloid, line 95: (No action required.)

As future work, I think we should also look at `PopName` and `PopExpr` and think about whether we can add a `Pop<...>` taking a parse node category. Generally I think it'd be great if we base the kind of ID we get back from the node stack on the category of the parse node rather than the specific parse node to the extent that we can (eg, all expressions should result in an `InstId`).

toolchain/parse/extract.cpp
```diff
@@ -0,0 +1,334 @@
...
+    static_cast<void>(((std::get<Index>(fields) =
+                            Extractable<T>::Extract(tree, it, end, trace),
+                        unused) = ... = 0));
+
+    if (!(std::get<Index>(fields).has_value() && ...)) {
+      if (trace) {
+        *trace << "Tuple: error\n";
+      }
+      return std::nullopt;
+    }
```
zygoloid: Along the failure path here, I think we need to reset `it` back to where it was before we started, otherwise (eg) an `optional<tuple>` or a `SmallVector<struct>` won't work properly.

Hm. Also, the pack fold tries to extract all elements even if an earlier extraction fails. That seems OK (though perhaps wasteful) when we're not tracing, but when we are tracing it will lead to additional trace output after the point of failure, and potentially a lot of it. Maybe we should stop extracting at the first failure:

```suggestion
    // Use a fold over the `=` operator to parse fields from right to left.
    [[maybe_unused]] int unused;
    bool ok = true;
    static_cast<void>(((ok &&
                        ok = (std::get<Index>(fields) =
                              Extractable<T>::Extract(tree, it, end, trace)).has_value(),
                        unused) = ... = 0));

    if (!ok) {
      if (trace) {
        *trace << "Tuple: error\n";
      }
      return std::nullopt;
    }
```

toolchain/parse/extract.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+// Extract the fields of a simple aggregate type.
+template <typename T>
+struct Extractable {
+  static_assert(std::is_aggregate_v<T>, "Unsupported child type");
+  static auto ExtractImpl(const Tree* tree, Tree::SiblingIterator& it,
+                          Tree::SiblingIterator end, ErrorBuilder* trace)
+      -> std::optional<T> {
+    if (trace) {
+      *trace << "Aggregate: begin\n";
```
zygoloid: Maybe include `typeid(T).name()` here?

toolchain/parse/extract.cpp
```diff
@@ -0,0 +1,334 @@
...
+    return std::nullopt;
+  }
+  return result;
+}
+
+// Manually instantiate Tree::TryExtractNodeFromChildren
+#define CARBON_PARSE_NODE_KIND(KindName)                    \
+  template auto Tree::TryExtractNodeFromChildren<KindName>( \
+      llvm::iterator_range<Tree::SiblingIterator> children, \
+      ErrorBuilder * trace) const -> std::optional<KindName>;
```
zygoloid, line 332:
```suggestion
      ErrorBuilder* trace) const -> std::optional<KindName>;
```
I guess `clang-format` undoes this?

toolchain/parse/node_ids.h
```diff
@@ -0,0 +1,96 @@
...
+constexpr NodeIdInCategory<Category> NodeIdInCategory<Category>::Invalid =
+    NodeIdInCategory<Category>(NodeId::InvalidIndex);
+
+// Aliases for `NodeIdInCategory` to describe particular categories of nodes.
+using AnyDecl = NodeIdInCategory<NodeCategory::Decl>;
+using AnyExpr = NodeIdInCategory<NodeCategory::Expr>;
+using AnyModifier = NodeIdInCategory<NodeCategory::Modifier>;
+using AnyNameComponent = NodeIdInCategory<NodeCategory::NameComponent>;
+using AnyPattern = NodeIdInCategory<NodeCategory::Pattern>;
+using AnyStatement = NodeIdInCategory<NodeCategory::Statement>;
```
zygoloid: I think these types should have an `Id` suffix, eg `AnyDeclId`.

toolchain/parse/node_ids.h
```diff
@@ -0,0 +1,96 @@
...
+struct NodeIdNot : public NodeId {
+  // An explicitly invalid instance.
+  static const NodeIdNot<T> Invalid;
+
+  explicit NodeIdNot(NodeId node_id) : NodeId(node_id) {}
+};
+
+template <typename T>
+constexpr NodeIdNot<T> NodeIdNot<T>::Invalid =
+    NodeIdNot<T>(NodeId::InvalidIndex);
```
zygoloid, line 92: Would it make sense to take `const NodeKind&` here instead of `typename`? All extract wants is the kind, and passing the kind in will mean that these can be used in `typed_nodes.h` before the corresponding typed node is defined. I think this is the only remaining reason why `typed_nodes.h` needs to be in a somewhat-topological order.

toolchain/parse/node_kind.h
```diff
@@ -50,6 +87,40 @@ class NodeKind : public CARBON_ENUM_BASE(NodeKind) {
 // We expect the parse node kind to fit compactly into 8 bits.
 static_assert(sizeof(NodeKind) == 1, "Kind objects include padding!");
 
+// A definition of an parse node kind. This is an NodeKind value, plus
+// ancillary data such as the name to use for the node kind in LLVM IR. These
+// are not copyable, and only one instance of this type is expected to exist per
+// parse node kind, specifically `TypedNode::Kind`. Use `NodeKind` instead as a
+// thin wrapper around an parse node kind index.
```
zygoloid:
```suggestion
// A definition of a parse node kind. This is a NodeKind value, plus
// ancillary data such as the name to use for the node kind in LLVM IR. These
// are not copyable, and only one instance of this type is expected to exist per
// parse node kind, specifically `TypedNode::Kind`. Use `NodeKind` instead as a
// thin wrapper around a parse node kind index.
```

toolchain/parse/tree.cpp
```diff
@@ -235,6 +246,15 @@ auto Tree::Verify() const -> ErrorOr<Success> {
       return Error(llvm::formatv(
           "Node #{0} is a placeholder node that wasn't replaced.", n.index));
     }
+    // Should extract successfully if node not marked as having an error.
+    if (!n_impl.has_error && !TestExtract(this, n, n_impl.kind, nullptr)) {
```
zygoloid, line 253: Can you measure any performance impact from this additional verify step? (I don't know that we have a good benchmark here, but maybe you could generate a large input and run it through `carbon compile`.)

toolchain/parse/typed_nodes.h
```diff
@@ -0,0 +1,781 @@
...
+// The name of a package or library for `package`, `import`, and `library`.
+using PackageName = LeafNode<NodeKind::PackageName>;
+using LibraryName = LeafNode<NodeKind::LibraryName>;
+using DefaultLibrary = LeafNode<NodeKind::DefaultLibrary>;
+
+using PackageIntroducer = LeafNode<NodeKind::PackageIntroducer>;
+using PackageApi = LeafNode<NodeKind::PackageApi>;
+using PackageImpl = LeafNode<NodeKind::PackageImpl>;
+
+// `library` in `package` or `import`:
```
zygoloid:
```suggestion
// `library` in `package` or `import`.
```

toolchain/parse/typed_nodes.h
```diff
@@ -0,0 +1,781 @@
...
+// `import TheirPackage library "TheirLibrary";`
+using ImportIntroducer = LeafNode<NodeKind::ImportIntroducer>;
+struct ImportDirective {
+  static constexpr auto Kind = NodeKind::ImportDirective.Define();
+  ImportIntroducerId introducer;
+  std::optional<PackageNameId> name;
+  std::optional<LibrarySpecifierId> library;
+};
+
+// `library` as directive:
```
zygoloid:
```suggestion
// `library` as directive.
```

toolchain/parse/typed_nodes.h
```diff
@@ -0,0 +1,781 @@
...
+  using InfixOperator##Name = InfixOperator<NodeKind::InfixOperator##Name>;
+#define CARBON_PARSE_NODE_KIND_POSTFIX_OPERATOR(Name, ...) \
+  using PostfixOperator##Name =                            \
+      PostfixOperator<NodeKind::PostfixOperator##Name>;
+#include "toolchain/parse/node_kind.def"
+
+// The first operand of a short-circuiting infix operator: `a and` or `a or`.
+// The complete operator expression will be an InfixOperator with this as the
+// `lhs`.
+// FIXME: should this be a template?
```
zygoloid: Do you want this answered before the PR lands? If not:
```suggestion
// TODO: Should this be a template?
```
I don't really have a strong opinion here. It's not much duplication compared to the overhead of writing a template and two aliases. I'd be happy to say that we don't make this into a template until / unless we want to write generic code to cover both cases at once (eg, in check).

toolchain/parse/typed_nodes.h
```diff
@@ -0,0 +1,781 @@
...
+};
+
+// `.a: i32`
+struct StructFieldType {
+  static constexpr auto Kind = NodeKind::StructFieldType.Define();
+  StructFieldDesignatorId designator;
+  AnyExpr type_expr;
+};
+
+// Struct literals, such as `{.a = 0}`:
```
zygoloid:
```suggestion
// Struct literals, such as `{.a = 0}`.
```

toolchain/parse/typed_nodes.h
```diff
@@ -0,0 +1,781 @@
...
+struct StructLiteral {
+  static constexpr auto Kind =
+      NodeKind::StructLiteral.Define(NodeCategory::Expr);
+  StructLiteralOrStructTypeLiteralStartId introducer;
+  CommaSeparatedList<StructFieldValueId, StructCommaId> fields;
+};
+
+// Struct type literals, such as `{.a: i32}`:
```
zygoloid:
```suggestion
// Struct type literals, such as `{.a: i32}`.
```

toolchain/parse/typed_nodes.h
```diff
@@ -0,0 +1,781 @@
...
+
+// `constraint NC { ... }`
+struct NamedConstraintDefinition {
+  static constexpr auto Kind =
+      NodeKind::NamedConstraintDefinition.Define(NodeCategory::Decl);
+  NamedConstraintDefinitionStartId signature;
+  llvm::SmallVector<AnyDecl> members;
+};
+
+// Define `FooId` as the id for type `Foo`
```
zygoloid: Probably this was originally my phrasing, but I think this is clearer the other way around:
```suggestion
// Define `Foo` as the node type for the ID type `FooId`.
```

toolchain/parse/typed_nodes.h
```diff
@@ -0,0 +1,781 @@
...
+// This class provides a shorthand for defining parse node kinds for leaf nodes.
+template <const NodeKind& KindT, NodeCategory Category = NodeCategory::None>
+struct LeafNode {
+  static constexpr auto Kind = KindT.Define(Category);
+};
+
+// Each type defined below corresponds to a parse node kind, and describes the
+// expected child structure of that parse node.
+
+// Error nodes
```
zygoloid, line 70: I think it would be helpful to do something to make these subheadings stand out a bit more in the file. Given that we use markdown-inspired syntax in our comments already, how about one of these:
```suggestion
// ### Error nodes
```
or
```suggestion
// Error nodes
// -----------
```

toolchain/parse/typed_nodes.h
```diff
@@ -0,0 +1,781 @@
...
+
+#ifndef CARBON_TOOLCHAIN_PARSE_TYPED_NODES_H_
+#define CARBON_TOOLCHAIN_PARSE_TYPED_NODES_H_
+
+#include "toolchain/parse/node_ids.h"
+#include "toolchain/parse/node_kind.h"
+
+namespace Carbon::Parse {
+
+// A pair of a list item and its optional following comma.
```
zygoloid, line 16: A subheading comment before these three non-parse-node-kind helpers would be helpful.

toolchain/parse/typed_nodes_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+  llvm::vfs::InMemoryFileSystem fs_;
+  std::forward_list<SourceBuffer> source_storage_;
+  std::forward_list<Lex::TokenizedBuffer> token_storage_;
+  std::forward_list<Tree> tree_storage_;
+  DiagnosticConsumer& consumer_ = ConsoleDiagnosticConsumer();
+};
+
+TEST_F(TypedNodeTest, Empty) {
+  auto* tree = &GetTree("");
+  auto file = File::Make(tree);
```
zygoloid: I think we should rename `File::Make` to `File::Extract` for symmetry with other extraction operations. I also wonder if we can switch this around to being `tree->ExtractFile()`, but perhaps the dependencies there are awkward?

toolchain/parse/typed_nodes_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+  EXPECT_TRUE(tree->IsValid<FileStart>(file.start));
```
zygoloid: I wonder if we should make the type optional when the argument is a `NodeIdForKind` already, so we can write this as just `tree->IsValid(file.start)`. (Even if so, I'm happy for that to be left to a future PR!)

chandlerc:
Could you add a PR description that details a bit more of what the end result here is, what the API looks like, etc. This is a large and complex PR, and so it is quite hard to look at it and really understand the effect it has as opposed to the implementation strategy it leverages.

toolchain/parse/typed_nodes_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+  ASSERT_TRUE(f_fn.has_value());
+  auto f_sig = tree->Extract(f_fn->signature);
+  ASSERT_TRUE(f_sig.has_value());
+  EXPECT_FALSE(f_sig->return_type.has_value());
+  EXPECT_TRUE(f_sig->modifiers.empty());
+
+  auto g_fn = tree->ExtractAs<FunctionDecl>(file.decls[1]);
+  ASSERT_TRUE(g_fn.has_value());
+  EXPECT_TRUE(g_fn->return_type.has_value());
+  EXPECT_FALSE(g_fn->modifiers.empty());
```
zygoloid, line 91: It'd be nice to have an example with multiple modifiers so we can check that we extract them in the right order (reversed from sibling order).

toolchain/check/node_stack.h
josh11b: Added a TODO. Do note though that some parse node kinds are in multiple categories: `var` and `let` are both declarations and statements.

toolchain/parse/extract.cpp
josh11b: Good idea.

toolchain/parse/node_ids.h
josh11b: Done.

toolchain/parse/node_ids.h
josh11b: It would make the use of this more noisy: either you write `Foo::Kind` (which only works after `Foo` is defined anyway) or you write `NodeKind::Foo` which is pretty long. I can still do it, but I really haven't had any trouble coming up with a logical order to write typed_nodes.h in that respects the dependency edges. `NodeIdNot` specifically is only used in one place to say "not the introducer", which of course was already defined as is natural.

toolchain/parse/typed_nodes_test.cpp
josh11b: I took a stab at it and it seemed to have worked.

toolchain/parse/typed_nodes_test.cpp
josh11b: Done.

toolchain/parse/extract.cpp
josh11b: Indeed our autoformating undoes this.

toolchain/parse/typed_nodes_test.cpp
josh11b: I split it into pieces, which worked with a forward declaration.

toolchain/parse/tree.cpp
josh11b: Added a comment and to the PR description.

toolchain/parse/typed_nodes.h
josh11b: Done.

toolchain/parse/typed_nodes.h
josh11b: Updated comment.

toolchain/parse/typed_nodes.h
josh11b: Used the latter

josh11b:
> Could you add a PR description that details a bit more of what the end result here is, what the API looks like, etc. This is a large and complex PR, and so it is quite hard to look at it and really understand the effect it has as opposed to the implementation strategy it leverages.

Comments and PR description updated.

toolchain/check/node_stack.h
```diff
@@ -202,6 +202,9 @@ class NodeStack {
   // Pops a name from the top of the stack and returns the ID.
   auto PopName() -> SemIR::NameId { return PopNameWithParseNode().second; }
 
+  // TODO: Can we ad a `Pop<...>` that takes a parse node category? See
```
zygoloid:
```suggestion
  // TODO: Can we add a `Pop<...>` that takes a parse node category? See
```

toolchain/parse/tree.h
```diff
@@ -178,14 +175,53 @@ class Tree : public Printable<Tree> {
...
+  //   } else if (auto class_def = tree->ExtractAs<ClassDefinition>(decl_id)) {
+  //     // ...
+  //   }
+  // }
+  // ```
+
+  // Extract a `File` object representing the parse tree for the whole file.
+  // #include "toolchain/parse/typed_nodes.h" to get the definition of `File`
+  // and the types representing its children nodes.
+  auto ExtractFile() -> File const;
```
zygoloid:
```suggestion
  auto ExtractFile() const -> File;
```

toolchain/parse/tree.h
```diff
@@ -171,11 +175,51 @@ class Tree : public Printable<Tree> {
...
+
+  // Converts this node_id to a typed node of a specified type, if it is a valid
+  // node of that kind.
+  template <typename T>
+  auto ExtractAs(NodeId node_id) const -> std::optional<T>;
+
+  // Like ExtractAs(), but malformed tree errors are not fatal.
```
zygoloid: Would something like this make sense?
```suggestion
  // Like ExtractAs(), but malformed tree errors are not fatal. Should only be used
  // by `Verify()`.
```

toolchain/parse/node_ids.h
zygoloid: OK, let's leave as-is for now.

zygoloid (approved):
LGTM, please also check with @chandlerc.

toolchain/parse/typed_nodes.h
```diff
@@ -0,0 +1,829 @@
...
+// This class provides a shorthand for defining parse node kinds for leaf nodes.
+template <const NodeKind& KindT, NodeCategory Category = NodeCategory::None>
+struct LeafNode {
+  static constexpr auto Kind = KindT.Define(Category);
+};
+
+// ----------------------------------------------------------------------------
+// Each type defined below corresponds to a parse node kind, and describes the
+// expected child structure of that parse node. Each should have a
+// `static constexpr Kind` member initialized to calling `Define` on the
```
chandlerc:
```suggestion
// `static constexpr Kind` member initialized by calling `Define` on the
```

toolchain/parse/typed_nodes.h
```diff
@@ -0,0 +1,829 @@
...
+// corresponding `NodeKind`, in order to set its `NodeCategory`.
```
chandlerc:
```suggestion
// corresponding `NodeKind`, and passing in the `NodeCategory` of that kind.
// This will both associate the category with the node kind and create the
// necessary kind object for the typed node.
```

toolchain/parse/typed_nodes.h
```diff
@@ -0,0 +1,829 @@
...
+
+// This class provides a shorthand for defining parse node kinds for leaf nodes.
+template <const NodeKind& KindT, NodeCategory Category = NodeCategory::None>
+struct LeafNode {
+  static constexpr auto Kind = KindT.Define(Category);
+};
+
+// ----------------------------------------------------------------------------
+// Each type defined below corresponds to a parse node kind, and describes the
+// expected child structure of that parse node. Each should have a
```
chandlerc:
```suggestion
// Each node kind (in node_kind.def) should have a corresponding type defined here
// which describes the expected child structure of that parse node.
//
// Each of these types should start with a
```

toolchain/parse/typed_nodes.h
```diff
@@ -0,0 +1,829 @@
...
+// - `FooId` to require that child to have kind `Foo`,
+// - `AnyFooId` to require that child to have a kind in category `Foo`,
+// - `NodeIdOneOf<T, U>` to require the child to have kind `T::Kind` or
+//   `U::Kind`,
+// - `NodeIdNot<T>` to match any single child that doesn't have kind `T::Kind`,
+// - `llvm::SmallVector<T>` to match any number of children matching `T`
+// - `std::optional<T>` to match 0 or 1 children matching `T`
+// - `std::tuple<T...>` to match children matching `T...`
+// - `Aggregate`, for simple aggregate type `struct Aggregate { T x; U y; }`,
+//   to match children with types `T` and `U`.
```
chandlerc:
```suggestion
// This should be followed by field declarations that describe the child nodes, in order, that occur in the parse tree. The `Extract...` functions on the parse tree use struct reflection on these fields to guide the extraction of the child nodes from the tree into an object of this type with these fields for convenient access.
//
// The types of these fields are special and describe the specific child node structure of the parse node. Many of these types are defined in `node_ids.h`.
//
// Valid primitive types here are:
// - `NodeId` to match any single child node
// - `FooId` to require that child to have kind `NodeKind::Foo`
// - `AnyCatId` to require that child to have a kind in category `Cat`
// - `NodeIdOneOf<A, B>` to require the child to have kind `NodeKind::A` or `NodeKind::B`
// - `NodeIdNot<A>` to match any single child whose kind is not `NodeKind::A`
//
// There a few, restricted composite field types allowed that compose types in various ways, where all of the `T`s and `U`s below are themselves valid field types:
// - `llvm::SmallVector<T>` to match any number of children matching `T`
// - `std::optional<T>` to match 0 or 1 children matching `T`
// - `std::tuple<T...>` to match children matching `T...`
// - Any provided `Aggregate` type that is a simple aggregate type such as `struct Aggregate { T x; U y; }`,
//   to match children with types `T` and `U`.
```

toolchain/parse/typed_nodes.h
```diff
@@ -0,0 +1,829 @@
...
+using InvalidParse =
+    LeafNode<NodeKind::InvalidParse, NodeCategory::Decl | NodeCategory::Expr>;
+
+// An invalid subtree. Always has an error so can never be extracted.
+using InvalidParseStart = LeafNode<NodeKind::InvalidParseStart>;
+struct InvalidParseSubtree {
+  static constexpr auto Kind =
+      NodeKind::InvalidParseSubtree.Define(NodeCategory::Decl);
+  InvalidParseStartId start;
+  llvm::SmallVector<NodeIdNot<InvalidParseStart>> extra;
```
chandlerc: Could you add a blank line after the kind and the start of the fields? (Not suggesting the edit to all of them, but I can if needed.)

```suggestion
  static constexpr auto Kind =
      NodeKind::InvalidParseSubtree.Define(NodeCategory::Decl);
      
  InvalidParseStartId start;
  llvm::SmallVector<NodeIdNot<InvalidParseStart>> extra;
```

toolchain/parse/extract.cpp
```diff
@@ -0,0 +1,341 @@
...
+// interface Extractable {
+//   // Extract a value of this type from the sequence of nodes starting at
+//   // `it`, and increment `it` past this type. Returns `std::nullopt` if
+//   // the tree is malformed. If `trace != nullptr`, writes what actions
+//   // were taken to `*trace`.
+//   static auto Extract(Tree* tree, Tree::SiblingIterator& it,
+//                       Tree::SiblingIterator end,
+//                       ErrorBuilder* trace) -> std::optional<Self>;
+// }
+// ```
```
chandlerc: While I love the Carbon code in here, I think maybe it's better to use a more direct C++ example.

```suggestion
// ```cpp
// template<>
// struct Extractable<T> {
//   // Extract a value of this type from the sequence of nodes starting at
//   // `it`, and increment `it` past this type. Returns `std::nullopt` if
//   // the tree is malformed. If `trace != nullptr`, writes what actions
//   // were taken to `*trace`.
//   static auto Extract(Tree* tree, Tree::SiblingIterator& it,
//                       Tree::SiblingIterator end,
//                       ErrorBuilder* trace) -> std::optional<T>;
// };
// ```
```

chandlerc: Honestly, not sure its an improvement... the Cabron feature here is really nice for illustrating what is going on...

So consider this optional.

toolchain/parse/extract.cpp
```diff
@@ -0,0 +1,341 @@
...
+//
+// Note that `Tree::SiblingIterator`s iterate in reverse order through the
+// children of a node.
+//
+// This class is only in this file.
+template <typename T>
+struct Extractable;
+
+// Extract an `NodeId` as a single child.
```
chandlerc:
```suggestion
// Extract a `NodeId` as a single child.
```

toolchain/parse/extract.cpp
```diff
@@ -0,0 +1,341 @@
...
+      return std::nullopt;
+    }
+    if (trace) {
+      *trace << "NodeIdForKind: " << Kind << " consumed\n";
+    }
+    return NodeIdForKind<Kind>(*it++);
+  }
+};
+
+// Extract an `NodeIdInCategory<Category>` as a single child.
```
chandlerc:
```suggestion
// Extract a `NodeIdInCategory<Category>` as a single child.
```

toolchain/parse/extract.cpp
```diff
@@ -0,0 +1,341 @@
...
+    }
+    if (trace) {
+      *trace << "NodeIdNot " << T::Kind << ": " << tree->node_kind(*it)
+             << " consumed\n";
+    }
+    return NodeIdNot<T>(*it++);
+  }
+};
+
+// Extract a `llvm::SmallVector<T>` by extracting `T`s until we can't.
```
chandlerc: Note: I have zero opinion on how best to match articles, and for `std` in particular, there's no universal pronunciation. I asked @zygoloid to give me his intended article given the different letters following and have suggested those here. Consider all of this optional. =]

```suggestion
// Extract an `llvm::SmallVector<T>` by extracting `T`s until we can't.
```

toolchain/check/node_stack.h
```diff
@@ -102,6 +103,7 @@ class NodeStack {
 
   // Pops the top of the stack if it is the given kind, and returns the
   // parse_node. Otherwise, returns std::nullopt.
+  // TODO: return a parse::NodeIdForKind<RequiredParseKind> instead.
```
chandlerc:
```suggestion
  // TODO: Return a `Parse::NodeIdForKind<RequiredParseKind>` instead.
```

toolchain/check/check.cpp
```diff
@@ -154,6 +154,7 @@ static auto ProcessParseNodes(Context& context,
     // clang warns on unhandled enum values; clang-tidy is incorrect here.
     // NOLINTNEXTLINE(bugprone-switch-missing-default-case)
     switch (auto parse_kind = context.parse_tree().node_kind(parse_node)) {
+      // TODO: switch to parse::Name##Id(parse_node) here
```
chandlerc:
```suggestion
      // TODO: Switch to `Parse::Name##Id(parse_node)` here.
```

toolchain/parse/tree.h
```diff
@@ -171,11 +175,52 @@ class Tree : public Printable<Tree> {
...
+  // Converts this node_id to a typed node of a specified type, if it is a valid
+  // node of that kind.
+  template <typename T>
+  auto ExtractAs(NodeId node_id) const -> std::optional<T>;
+
+  // Like ExtractAs(), but malformed tree errors are not fatal. Should only be
+  // used by `Verify()`.
+  template <typename T>
+  auto TryExtractAs(NodeId node_id, ErrorBuilder* trace) const
+      -> std::optional<T>;
```
chandlerc: What do you think about calling this `VerifyExtractAs` (or something similar)? I worry about the name `TryExtractAs` seeming like it should be used to try one of two different node types to see which one works, and it sounds like that's not really the use case here.

```suggestion
  // Like ExtractAs(), but malformed tree errors are not fatal. Should only be
  // used by `Verify()`.
  template <typename T>
  auto VerifyExtractAs(NodeId node_id, ErrorBuilder* trace) const
      -> std::optional<T>;
```

Another idea, but this one entirely optional: would returning `ErrorOr<T>` make sense here, to parallel `Verify` itself?

toolchain/parse/tree.h
```diff
@@ -171,11 +175,52 @@ class Tree : public Printable<Tree> {
   // Verifies the parse tree structure. Checks invariants of the parse tree
   // structure and returns verification errors.
   //
-  // This is primarily intended to be used as a
-  // debugging aid. This routine doesn't directly CHECK so that it can be used
-  // within a debugger.
+  // This is fairly slow, and is primarily intended to be used as a debugging
+  // aid. This routine doesn't directly CHECK so that it can be used within a
+  // debugger.
   auto Verify() const -> ErrorOr<Success>;
```
chandlerc: Could we move this below the non-verify extract APIs, but keep it with the verifying extract API?

chandlerc (commented):
Just wanted to say, this is really cool. I love that it gives us a structural verification layer that we can both enforce and leverage in places where it makes sense.

A bunch of comments & suggestions inline just trying to help make sure the code comments make sense to a reader with less context on everything. (Many were paired with Josh and/or Richard.)

Some are super nit-picky, sorry for that, just saw them as I was reading and figured I would suggest edits. None of those are important if they get in the way of merges or anything. And I've tried to mark a few other places where things are just optional suggestions.

toolchain/parse/tree.h
josh11b: Issue with `ErrorOr<T>` is the one user includes more context in the error.

toolchain/parse/tree.h
josh11b: Done.

chandlerc:
Awesome, mashing the merge button!

chandlerc (approved):

PR #3535, opened 2023-12-20
Create a BindName for function parameters.

zygoloid:
The goal here is to make the representation more uniform so that we can start adding different kinds of binding -- checked generic bindings and template bindings -- across both function parameters and local `let` declarations.

With this change, the entry in the parameter list for the function is the name binding, not the Param itself, which has some ripple effects on consumers of that list that want to access the parameter rather than the binding. This is expected to change again when we start adding more of the pattern matching SemIR, but this seems good enough for now.


toolchain/sem_ir/file.h
```diff
@@ -59,6 +59,11 @@ struct Function : public Printable<Function> {
   // function, in lexical order. The first block is the entry block. This will
   // be empty for declarations that don't have a visible definition.
   llvm::SmallVector<InstBlockId> body_block_ids = {};
+
+  // Given a parameter reference instruction from `param_refs_id` or
+  // `implicit_param_refs_id`, returns the corresponding parameter instruction.
+  static auto GetParamIdFromParamRefId(const File& sem_ir, InstId param_ref_id)
+      -> InstId;
```
jonmeow: Static functions should be above data members.

toolchain/sem_ir/file.h
```diff
@@ -59,6 +59,11 @@ struct Function : public Printable<Function> {
   // function, in lexical order. The first block is the entry block. This will
   // be empty for declarations that don't have a visible definition.
   llvm::SmallVector<InstBlockId> body_block_ids = {};
+
+  // Given a parameter reference instruction from `param_refs_id` or
+  // `implicit_param_refs_id`, returns the corresponding parameter instruction.
```
jonmeow: I'm not sure what the meaning of "parameter instruction" is. Is it specifically a SemIR::Param?

```suggestion
  // Given a parameter reference instruction from `param_refs_id` or
  // `implicit_param_refs_id`, returns the corresponding SemIR::Param's ID,
  // resolving wrapper instructions.
```

Note, if you change to return a pair with `SemIR::Param` (see comments in cpp) I think this is moot.

toolchain/sem_ir/file.cpp
```diff
@@ -13,9 +13,26 @@
...
+  if (auto addr_pattern = ref.TryAs<SemIR::AddrPattern>()) {
+    param_ref_id = addr_pattern->inner_id;
+    ref = sem_ir.insts().Get(param_ref_id);
+  }
+
+  if (auto bind_name = ref.TryAs<SemIR::BindName>()) {
+    param_ref_id = bind_name->value_id;
+  }
+
+  return param_ref_id;
```
jonmeow:
```suggestion
  CARBON_CHECK(ref.Is<SemIR::Param>());
  return param_ref_id;
```

Note, I'm suggesting this mainly because I was trying to figure out whether a specific type of instruction should be returned, and a CHECK would've helped me be confident. The comment change in the header may be sufficient; this might be undesirable since it kind of wastes an insts() query.

toolchain/sem_ir/file.cpp
```diff
@@ -13,9 +13,26 @@
 #include "toolchain/sem_ir/ids.h"
 #include "toolchain/sem_ir/inst.h"
 #include "toolchain/sem_ir/inst_kind.h"
+#include "toolchain/sem_ir/typed_insts.h"
 
 namespace Carbon::SemIR {
 
+auto Function::GetParamIdFromParamRefId(const File& sem_ir, InstId param_ref_id)
+    -> InstId {
+  auto ref = sem_ir.insts().Get(param_ref_id);
```
jonmeow: In the places you call this, you invariably follow with a `GetAs<SemIR::Param>`. Would it be worth having this return `std::pair<InstId, SemIR::Param>`, since you already do the fetch here?

toolchain/check/handle_binding_pattern.cpp
```diff
@@ -44,6 +44,12 @@ auto HandleBindingPattern(Context& context, Parse::NodeId parse_node) -> bool {
   // Every other kind of pattern binding has a name.
   auto [name_node, name_id] = context.node_stack().PopNameWithParseNode();
 
+  auto make_bind_name = [name_node = name_node, name_id = name_id](
+                            SemIR::TypeId type_id,
+                            SemIR::InstId value_id) -> SemIR::Inst {
+    return SemIR::BindName{name_node, type_id, name_id, value_id};
+  };
```
jonmeow, line 54: Personal preference, but I think this lamba is increasing code/complexity rather than reducing it. If there's any difference beyond `make_bind_name(type_id, value_id)` versus `SemIR::BindName{name_node, type_id, name_id, value_id}`, I'm missing it. After some reading, I think the only reason `name_node` and `name_id` are taken as copies is the C++20 issue. The lambda requires 5 LOC in order to reduce ~2 LOC (a net increase), also adding more control flow.

If you keep this, please add a comment at least explaining why the lambda versus direct construction.

jonmeow (approved):

toolchain/sem_ir/file.h

toolchain/sem_ir/file.h
zygoloid: Updated to match new return type.

toolchain/sem_ir/file.cpp
zygoloid: We now call `As<SemIR::Param>` which implicitly performs this check.

toolchain/sem_ir/file.cpp
zygoloid: Done. (In the pattern matching doc, there's an idea that we'll have two different kinds of instruction for parameters, but we can deal with that later.)

toolchain/check/handle_binding_pattern.cpp
zygoloid: Added a comment; my next PR is going to add more logic to this lambda, so I'd rather not remove it then re-add it :)

PR #3536, opened 2023-12-21
Handle out-of-line declarations.

jonmeow:
By adding an InstId to the NameScope, we can determine whether the declaration is being added to a scoped entity (versus a namespace).

The choice of InstId on NameScope is chosen versus other solutions because, for imports, we want to just have a list of InstIds to import and, from those, get the containing namespaces for addition. Similar may also be desirable for printing fully qualified names given a singular InstId. That means an InstId must have a path to find enclosing name scopes.

What we're looking at here is:

- NameScopeId knows its InstId. (done here)
- Inst knows the enclosing NameScopeId. (future work)
- To walk up enclosing scopes for an Inst:
  1. Fetch the Inst.
  2. Find its enclosing NameScopeId (which will be per-declaration due to Function etc complexity).
  3. Fetch the NameScope if not Package scope. (if Package scope, done)
  4. Use the InstId on the NameScope to go back to step 1.


toolchain/check/handle_namespace.cpp
```diff
@@ -21,9 +22,12 @@ auto HandleNamespace(Context& context, Parse::NodeId parse_node) -> bool {
...
   LimitModifiersOnDecl(context, KeywordModifierSet::None,
                        Lex::TokenKind::Namespace);
+  auto future_namespace_id = SemIR::InstId(context.insts().size());
+  auto name_scope_id = context.name_scopes().Add(future_namespace_id);
   auto namespace_id = context.AddInst(SemIR::Namespace{
       parse_node, context.GetBuiltinType(SemIR::BuiltinKind::NamespaceType),
-      context.name_scopes().Add()});
+      name_scope_id});
+  CARBON_CHECK(future_namespace_id == namespace_id);
   context.decl_name_stack().AddNameToLookup(name_context, namespace_id);
```
zygoloid: Maybe:
```c++
  auto namespace_inst = SemIR::Namespace{
      parse_node, context.GetBuiltinType(SemIR::BuiltinKind::NamespaceType),
      SemIR::NameScopeId::Invalid};
  auto namespace_id = context.AddInst(namespace_inst);
  namespace_inst.name_scope_id = context.name_scopes().Add(namespace_id);
  context.insts().Set(namespace_id, namespace_inst);
```
This is what we do in other contexts where we have cyclic references between IR entities (eg, `Class` vs `ClassDecl`).

zygoloid (approved):

toolchain/check/handle_namespace.cpp
jonmeow: Done

PR #3537, opened 2023-12-21
Fix minor grammatical error in docs

nicdgonzalez:
Use "an" instead of "a" when the following word starts with a vowel sound.


nicdgonzalez:
I'm not sure how to rerun the checks, but it should be good to go now.

zygoloid (approved):
Thank you!

PR #3543, opened 2023-12-26
Do TODO to rename `QualifiedDecl` -> `QualifiedName`

josh11b:
Renaming since the parse node does not represent a declaration.


toolchain/check/decl_name_stack.cpp
```diff
@@ -76,11 +76,11 @@ auto DeclNameStack::LookupOrAddName(NameContext name_context,
           if (!inst.Is<SemIR::Namespace>()) {
             // TODO: Point at the declaration for the scoped entity.
             CARBON_DIAGNOSTIC(
-                QualifiedDeclOutsideScopeEntity, Error,
+                QualifiedNameOutsideScopeEntity, Error,
                 "Out-of-line declaration requires a declaration in "
                 "scoped entity.");
             context_->emitter().Emit(name_context.parse_node,
-                                     QualifiedDeclOutsideScopeEntity);
+                                     QualifiedNameOutsideScopeEntity);
```
chandlerc: These renames lose a little bit of information. These *are* declarations, but of a qualified name. Maybe `DeclaringQualifiedName...` would be a more clear name?

But given the text "qualified declaration outside scope entity" reads about right to me, so maybe this one doesn't need the rename?

josh11b: Done.

toolchain/check/decl_name_stack.cpp
```diff
@@ -188,25 +188,24 @@ auto DeclNameStack::TryResolveQualifier(NameContext& name_context,
       if (auto class_decl = context_->insts()
                                 .Get(name_context.resolved_inst_id)
                                 .TryAs<SemIR::ClassDecl>()) {
-        CARBON_DIAGNOSTIC(QualifiedDeclInIncompleteClassScope, Error,
+        CARBON_DIAGNOSTIC(QualifiedNameInIncompleteClassScope, Error,
```
chandlerc: I think this is also talking about a erroneous declaration due to qualification.

chandlerc (approved):
One more diagnostic that I'd leave with the other name. LG with that.

toolchain/check/decl_name_stack.cpp
josh11b: Done.

PR #3545, opened 2023-12-28
Test trace output of `Tree::VerifyExtractAs`, fix found bugs

josh11b:
Tests previously uncovered code. Fix uncovered problems:
* formatting of trace output
* package & import directives need to be classified as declarations
* the problem that meant the previous problem wasn't caught by existing tests (since `Tree::Verify` didn't check that top-level declarations match `AnyDeclId`, as required by `Tree::ExtractFile()`).

### DIFF
--- a/toolchain/parse/typed_nodes_test.cpp
+++ b/toolchain/parse/typed_nodes_test.cpp
@@ -137,6 +137,149 @@ TEST_F(TypedNodeTest, For) {
   ASSERT_TRUE(for_var_name.has_value());
 }
 
+TEST_F(TypedNodeTest, VerifyExtractTraceLibrary) {
+  auto* tree = &GetTree(R"carbon(
+    library default impl;
+  )carbon");
+  auto file = tree->ExtractFile();
+
+  ASSERT_EQ(file.decls.size(), 1);
+  ErrorBuilder trace;
+  auto library = tree->VerifyExtractAs<LibraryDirective>(file.decls[0], &trace);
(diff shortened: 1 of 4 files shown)


toolchain/parse/typed_nodes_test.cpp
```diff
@@ -137,6 +137,149 @@ TEST_F(TypedNodeTest, For) {
...
+  auto file = tree->ExtractFile();
+
+  ASSERT_EQ(file.decls.size(), 1);
+  ErrorBuilder trace;
+  auto library = tree->VerifyExtractAs<LibraryDirective>(file.decls[0], &trace);
+  EXPECT_TRUE(library.has_value());
+  Error err = trace;
+  // Use Regex matching to avoid hard-coding the result of `typeinfo(T).name()`.
+  EXPECT_THAT(err.message(), testing::MatchesRegex(
+                                 R"Trace(Aggregate [^:]*: begin
```
chandlerc, line 153: I think this would also work and might be a bit more readable?
```suggestion
                                 R"Trace(\
Aggregate [^:]*: begin
```

(same pattern below if it does work)

chandlerc (approved):
Really nice, thanks!

Suggestion to maybe make the output raw string a touch more readable, but feel free to just merge if it doesn't work.

toolchain/parse/typed_nodes_test.cpp
josh11b: Doesn't work.

PR #3546, opened 2023-12-28
Use `LLVM_UNLIKELY` reduce cost of `Tree::Verify`

josh11b:


chandlerc:
No significant change to the performance here. I don't think the branches on trace or even the generated code in the trace-specific parts are really what's making things slow from looking at the profile a bit.

PR #3547, opened 2023-12-28
Start using typed parse node ids in the check stage

josh11b:
Goal is to increase type safety, though more work needs to be done (see added TODOs).

Note that, after this change, check handlers corresponding to deleted parse node kinds will no longer compile.

### DIFF
--- a/toolchain/check/handle_operator.cpp
+++ b/toolchain/check/handle_operator.cpp
@@ -7,17 +7,19 @@
 
 namespace Carbon::Check {
 
-auto HandleInfixOperatorAmp(Context& context, Parse::NodeId parse_node)
-    -> bool {
+auto HandleInfixOperatorAmp(Context& context,
+                            Parse::InfixOperatorAmpId parse_node) -> bool {
   return context.TODO(parse_node, "HandleInfixOperatorAmp");
 }
 
-auto HandleInfixOperatorAmpEqual(Context& context, Parse::NodeId parse_node)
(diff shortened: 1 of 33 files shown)


toolchain/parse/node_ids.h
```diff
@@ -50,6 +50,9 @@ struct NodeIdInCategory : public NodeId {
   // An explicitly invalid instance.
   static const NodeIdInCategory<Category> Invalid;
 
+  // TODO: Support conversion from `NodeIdForKind<Kind>` if `Kind::category()`
+  // overlaps with `Category.
```
chandlerc:
```suggestion
  // overlaps with `Category`.
```

toolchain/check/handle_operator.cpp
```diff
@@ -262,18 +281,14 @@ auto HandlePrefixOperatorNot(Context& context, Parse::NodeId parse_node)
   return true;
 }
 
-auto HandlePrefixOperatorPlus(Context& context, Parse::NodeId parse_node)
-    -> bool {
-  return context.TODO(parse_node, "HandlePrefixOperatorPlus");
-}
```
chandlerc, line 268: Maybe worth mentioning in the PR description that this already prevents us from leaving stale handlers around when removing parse nodes, and citing this one as an example?

chandlerc (approved):
Nice!

toolchain/check/handle_operator.cpp
josh11b: Done.

PR #3548, opened 2023-12-28
Make `Parse::NodeCategory` printable

josh11b:
Completes TODO in `parse/extract.cpp`, moving the printing code for `NodeCategory` out of the implementation of `Extractable<NodeIdInCategory<T>>`.


chandlerc (approved):
LG

PR #3549, opened 2023-12-28
Add a `MemberName` parse node category for member access expressions

josh11b:
Use this new category to replace the unconstrained `NodeId` child of `MemberAccessExpr ` and `PointerMemberAccessExpr`. For now this new category matches `IdentifierName` and `BaseName`, but later this will be expanded to support `a.(b.c)` and `p->(b.c)` syntactic forms.

QUESTION: Is it time to make a `node_category.def` x-macro file?
ANSWER: Not yet.

### DIFF
--- a/toolchain/parse/extract.cpp
+++ b/toolchain/parse/extract.cpp
@@ -99,6 +99,7 @@ struct Extractable<NodeIdInCategory<Category>> {
   }
       CARBON_NODE_CATEGORY(Decl);
       CARBON_NODE_CATEGORY(Expr);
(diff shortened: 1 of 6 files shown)


chandlerc (approved):
Very nice.

My answer to the question is "not yet" -> I'd want us to need to associate more metadata with each category or be growing towards mid 10s of categories to factor it out personally.

But I'm also happy for you or others working here that feel like a `.def` would be an improvement to switch to it at that point. It's really whenever it starts to feel like an improvement IMO.

PR #3552, opened 2024-01-02
Implement a list of exported names for imports.

jonmeow:
This adds a block for exported InstIds, rather than scanning the package scope. This working down a path discussed last month, which we'll need to add enclosing namespaces to the Inst in order to complete import of something like `namespace NS; var NS.a;`

Exports could've been a separate `vector<InstId>` on `SemIR::File`, but using an entry in `inst_blocks` felt more consistent.

### DIFF
--- a/toolchain/sem_ir/ids.h
+++ b/toolchain/sem_ir/ids.h
@@ -241,7 +241,8 @@ struct InstBlockId : public IdBase, public Printable<InstBlockId> {
   // 0-index block.
   static const InstBlockId Empty;
 
-  // Exported instructions. Always the 1-index block.
+  // Exported instructions. Always the 1-index block. Empty until the File is
+  // fully checked; intermediate state is in the Check::Context.
   static const InstBlockId Exports;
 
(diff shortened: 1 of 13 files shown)


toolchain/sem_ir/ids.h
```diff
@@ -237,9 +237,13 @@ struct InstBlockId : public IdBase, public Printable<InstBlockId> {
   using ElementType = InstId;
   using ValueType = llvm::MutableArrayRef<ElementType>;
 
-  // All File instances must provide the 0th instruction block as empty.
+  // An empty block, reused to avoid allocating empty vectors. Always the
+  // 0-index block.
   static const InstBlockId Empty;
 
+  // Exported instructions. Always the 1-index block.
```
zygoloid: I think it'd be useful to mention here that this block is empty until the file has been fully checked.

zygoloid (approved):

toolchain/sem_ir/ids.h

PR #3555, opened 2024-01-03
Basic support for generic bindings.

zygoloid:
This change adds a `BindSymbolicName` instruction for generic bindings, paralleling the existing `BindName`. A mechanism is also added to allow both kinds of binding to be accessed uniformly, for convenience in the case where the two different kinds of binding are treated the same.

Generic bindings of type `type` are allowed to be used as types, although no operations are provided for such types. For now lowering treats these types as empty structs, which seems like a reasonable lowering for non-monomorphized unconstrained types.


toolchain/check/handle_binding_pattern.cpp
```diff
@@ -10,10 +10,10 @@
 namespace Carbon::Check {
 
 auto HandleAddress(Context& context, Parse::AddressId parse_node) -> bool {
-  auto self_param_id =
-      context.node_stack().Pop<Parse::NodeKind::BindingPattern>();
-  auto self_param = context.insts().TryGetAs<SemIR::BindName>(self_param_id);
-  if (self_param && self_param->name_id == SemIR::NameId::SelfValue) {
+  auto self_param_id = context.node_stack().PopPattern();
```
chandlerc, line 13: Maybe leave a TODO to restore some amount of type safety here when the infrastructure needed is available?

chandlerc (commented):
Should these be "symbolic bindings" rather than "generic bindings" from the terminology? https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/generics/terminology.md#bindings

Also, if its easy to separate out the typed instruction infrastructure changes from the addition of the new binding patterns that motivates them, it would make review a bit easier IMO. If not, maybe add a bit more detail on the more general nature of changes to the infrastructure? As is, the short sentence in the description doesn't give me much context -- I thought it referred to just the common code factoring in `check`, and then was a bit surprised at the new machinery in `inst.h`, `typed_insts.h` and such.

zygoloid:
> Should these be "symbolic bindings" rather than "generic bindings" from the terminology? https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/generics/terminology.md#bindings

Done.

> Also, if its easy to separate out the typed instruction infrastructure changes from the addition of the new binding patterns that motivates them, it would make review a bit easier IMO.

Done, and landed separately.

toolchain/check/handle_binding_pattern.cpp
zygoloid: I think this general `PopPattern` might actually be what we want here: according to the typed node, an `AddressPattern` can have `AnyPatternId` within it, and this function now diagnoses the case where it's applied to a pattern other than a binding pattern.

toolchain/check/handle_binding_pattern.cpp
```diff
@@ -28,14 +28,8 @@ auto HandleAddress(Context& context, Parse::AddressId parse_node) -> bool {
...
-auto HandleGenericBindingPattern(Context& context,
-                                 Parse::GenericBindingPatternId parse_node)
-    -> bool {
-  return context.TODO(parse_node, "GenericBindingPattern");
-}
-
-auto HandleBindingPattern(Context& context, Parse::BindingPatternId parse_node)
-    -> bool {
+auto HandleAnyBindingPattern(Context& context, Parse::NodeId parse_node,
+                             bool is_generic) -> bool {
```
chandlerc, line 34: I think this should be:
```suggestion
                             bool is_comptime) -> bool {
```

Specifically, my understanding of the terminology is that the entity with the pattern becomes a generic thing, and these bindings result in generic parameters, but the bindings are "compile-time bindings".

(Somewhat commenting to see if I've internalized the terminology correctly, this isn't super important so feel free to defer some of this if needed. Also happy if we need to adjust the terminology.)

toolchain/check/handle_binding_pattern.cpp
```diff
@@ -72,6 +69,17 @@ auto HandleBindingPattern(Context& context, Parse::BindingPatternId parse_node)
               context.node_stack().PeekParseNodeKind()) {
     case Parse::NodeKind::ReturnedModifier:
     case Parse::NodeKind::VariableIntroducer: {
+      if (is_generic) {
+        CARBON_DIAGNOSTIC(
+            GenericBindingInVarDecl, Error,
+            "`var` declaration cannot declare a generic binding.");
+        context.emitter().Emit(type_node, GenericBindingInVarDecl);
```
chandlerc: Similar to the bool renaming:

```suggestion
            CompTimeBindingInVarDecl, Error,
            "`var` declaration cannot declare a compile-time binding.");
        context.emitter().Emit(type_node, CompTimeBindingInVarDecl);
```

Also, should this be `CompTime`, `CompileTime`, or `Comptime`? I have no opinion really, happy for you to pick whichever, but may be reflected in the name of the bool as well.

toolchain/lower/file_context.cpp
```diff
@@ -219,7 +219,9 @@ auto FileContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
             sem_ir().insts().TryGetAs<SemIR::AddrPattern>(param_ref_id)) {
       bind_name_id = addr->inner_id;
     }
-    CARBON_CHECK(sem_ir().insts().TryGetAs<SemIR::BindName>(bind_name_id));
+    auto bind_name = sem_ir().insts().Get(bind_name_id);
+    // TODO: Should we stop passing generic bindings at runtime?
```
chandlerc:
```suggestion
    // TODO: Should we stop passing compile-time bindings at runtime?
```

chandlerc (approved):
Generally LG, just some terminology tweaks. Feel free to submit whenever.

toolchain/check/handle_binding_pattern.cpp
zygoloid: I think you're right, but this is reflecting whether we had a `GenericBindingPattern` or a `BindingPattern` as the parse node, which is also using the old terminology. Probably best to update these names together.

I've updated the other places you pointed out and added a TODO to rename the `GenericBindingPattern` parse node.

toolchain/check/handle_binding_pattern.cpp
zygoloid: Done.

PR #3556, opened 2024-01-03
Split Parse out to its own target.

jonmeow:
This is mirroring the structure of codegen/codegen.h, lower/lower.h, and check/check.h. I recently did lex/lex.h, so parse/parse.h is the last. Now, the directory's main API file is eponymous with the directory.

I could've used a friend function to avoid making the Tree constructor public, but in other places we make less use of `friend`, just leaving things public. This felt more consistent, and simple because it only affects the constructor.

### DIFF
--- a/toolchain/parse/BUILD
+++ b/toolchain/parse/BUILD
@@ -37,6 +37,7 @@ cc_test(
     srcs = ["typed_nodes_test.cpp"],
     deps = [
         ":node_kind",
+        ":parse",
         ":tree",
         "//testing/base:gtest_main",
         "//toolchain/diagnostics:diagnostic_emitter",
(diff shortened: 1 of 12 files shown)


toolchain/parse/BUILD
```diff
@@ -47,6 +48,35 @@ cc_test(
...
+    name = "parse",
+    srcs = [
+        "context.cpp",
+        "context.h",
+        "parse.cpp",
+    ] +
+    # Glob handler files to avoid missing any.
+    glob([
+        "handle_*.cpp",
+    ]),
```
zygoloid, line 61: Do we still need these to be combined together? It looks like we have some nice layering we could extract now:

```mermaid
graph BT
  parse --> handlers & context & tree
  handlers --> context & tree
  context --> tree
```

(We couldn't do this before only because `Parse` was inside tree, creating a cycle.)

zygoloid (approved):
Thanks, looks good. I think this allows us to get a bit more layering in the parser too, though I'm happy to defer that to a later PR.

toolchain/parse/BUILD
jonmeow: I can look at that in a separate PR -- I wasn't trying to split that out here because the handlers are still declared in context.h, so I hadn't been sure of the value for splitting.

PR #3557, opened 2024-01-03
Shift handler loops to use a dispatch table.

jonmeow:
The intent here is that while the optimizer may optimize a large switch to a similar construct, it may get a little confused by the code structure. The switch to an array offers a stronger guarantee of how the code structure will execute.

This is intended to mirror the path of lex, minus `[[clang::musttail]]`: mainly, I wasn't sure if we should be restructuring code for `[[clang::musttail]]`.

### DIFF
--- a/toolchain/lower/function_context.cpp
+++ b/toolchain/lower/function_context.cpp
@@ -43,10 +43,10 @@ using DispatchFunctionT = auto(FunctionContext& context, SemIR::InstId inst_id,
 
 // Transforms a parse node ID to a typed parse node, which is then passed to the
 // handler.
-#define CARBON_SEM_IR_INST_KIND(Name)                                  \
-  auto Dispatch##Name(FunctionContext& context, SemIR::InstId inst_id, \
-                      SemIR::Inst inst) -> void {                      \
-    return Handle##Name(context, inst_id, inst.As<SemIR::Name>());     \
+#define CARBON_SEM_IR_INST_KIND(Name)                                         \
+  static auto Dispatch##Name(FunctionContext& context, SemIR::InstId inst_id, \
+                             SemIR::Inst inst) -> void {                      \
+    return Handle##Name(context, inst_id, inst.As<SemIR::Name>());            \
   }
 #include "toolchain/sem_ir/inst_kind.def"
 
--- a/toolchain/lower/function_context.cpp
+++ b/toolchain/lower/function_context.cpp
@@ -37,19 +37,33 @@ auto FunctionContext::TryToReuseBlock(SemIR::InstBlockId block_id,
   return true;
 }
 
+// Structure for the core handler dispatch.
+using DispatchFunctionT = auto(FunctionContext& context, SemIR::InstId inst_id,
+                               SemIR::Inst inst) -> void;
+using DispatchTableT =
+    std::array<DispatchFunctionT*, SemIR::InstKind::EnumCount>;
+
+// Transforms a parse node ID to a typed parse node, which is then passed to the
+// handler.
+#define CARBON_SEM_IR_INST_KIND(Name)                                  \
+  auto Dispatch##Name(FunctionContext& context, SemIR::InstId inst_id, \
+                      SemIR::Inst inst) -> void {                      \
+    return Handle##Name(context, inst_id, inst.As<SemIR::Name>());     \
+  }
+#include "toolchain/sem_ir/inst_kind.def"
+
+// The main dispatch table. This is used instead of a switch in order to
+// optimize for common functionality.
+static constexpr DispatchTableT DispatchTable = {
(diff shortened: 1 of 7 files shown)


jonmeow:
I was building this on top of #3556 because I figured this one was more likely to change, or might not be desirable. Happy to get feedback on the approach and iterate here.

toolchain/lower/function_context.cpp
```diff
@@ -37,19 +37,33 @@ auto FunctionContext::TryToReuseBlock(SemIR::InstBlockId block_id,
...
+    return Handle##Name(context, inst_id, inst.As<SemIR::Name>());     \
+  }
+#include "toolchain/sem_ir/inst_kind.def"
+
+// The main dispatch table. This is used instead of a switch in order to
+// optimize for common functionality.
+static constexpr DispatchTableT DispatchTable = {
+#define CARBON_SEM_IR_INST_KIND(Name) &Dispatch##Name,
+#include "toolchain/sem_ir/inst_kind.def"
+};
```
zygoloid: This enum count / array sizing technique seems to be adding a bit of complexity to the enum definitions that we otherwise haven't needed so far. I'm not opposed, but I wonder if you considered an inferred array bound instead:

```suggestion
// Structure for the core handler dispatch.
using DispatchFunctionT = auto(FunctionContext& context, SemIR::InstId inst_id,
                               SemIR::Inst inst) -> void;

// Transforms a parse node ID to a typed parse node, which is then passed to the
// handler.
#define CARBON_SEM_IR_INST_KIND(Name)                                  \
  auto Dispatch##Name(FunctionContext& context, SemIR::InstId inst_id, \
                      SemIR::Inst inst) -> void {                      \
    return Handle##Name(context, inst_id, inst.As<SemIR::Name>());     \
  }
#include "toolchain/sem_ir/inst_kind.def"

// The main dispatch table. This is used instead of a switch in order to
// optimize for common functionality.
static constexpr DispatchFunctionT* DispatchTable[] = {
#define CARBON_SEM_IR_INST_KIND(Name) &Dispatch##Name,
#include "toolchain/sem_ir/inst_kind.def"
};
```
We could use `std::array` CTAD and also avoid having to write out a function type:
```suggestion
// Transforms a parse node ID to a typed parse node, which is then passed to the
// handler.
#define CARBON_SEM_IR_INST_KIND(Name)                                  \
  auto Dispatch##Name(FunctionContext& context, SemIR::InstId inst_id, \
                      SemIR::Inst inst) -> void {                      \
    return Handle##Name(context, inst_id, inst.As<SemIR::Name>());     \
  }
#include "toolchain/sem_ir/inst_kind.def"

// The main dispatch table. This is used instead of a switch in order to
// optimize for common functionality.
static constexpr std::array DispatchTable = {
#define CARBON_SEM_IR_INST_KIND(Name) &Dispatch##Name,
#include "toolchain/sem_ir/inst_kind.def"
};
```
We could also shorten this a little by generating lambdas inline:
```suggestion
// The main dispatch table. This is used instead of a switch in order to
// optimize for common functionality.
static constexpr std::array DispatchTable = {
#define CARBON_SEM_IR_INST_KIND(Name)                                \
   +[](FunctionContext& context, SemIR::InstId inst_id,              \
       SemIR::Inst inst) -> void {                                   \
      return Handle##Name(context, inst_id, inst.As<SemIR::Name>()); \
   },
#include "toolchain/sem_ir/inst_kind.def"
};
```

I'm not going to push for a change here if you prefer it as-is.

toolchain/lower/function_context.cpp
```diff
@@ -37,19 +37,33 @@ auto FunctionContext::TryToReuseBlock(SemIR::InstBlockId block_id,
...
+// Structure for the core handler dispatch.
+using DispatchFunctionT = auto(FunctionContext& context, SemIR::InstId inst_id,
+                               SemIR::Inst inst) -> void;
+using DispatchTableT =
+    std::array<DispatchFunctionT*, SemIR::InstKind::EnumCount>;
+
+// Transforms a parse node ID to a typed parse node, which is then passed to the
+// handler.
+#define CARBON_SEM_IR_INST_KIND(Name)                                  \
+  auto Dispatch##Name(FunctionContext& context, SemIR::InstId inst_id, \
```
zygoloid: If you don't switch to using a lambda as I suggest in the other comment, this should be declared `static`.

zygoloid (commented):
Thanks!

toolchain/lower/function_context.cpp
jonmeow: Done

toolchain/lower/function_context.cpp
jonmeow: I'm keeping the separate `Dispatch` for now, I think it's a little easier to read that way. But switched to `std::array`

jonmeow:
Per discussion, sticking with a switch until there's benchmark to motivate.

PR #3559, opened 2024-01-03
Add support for '--config=clang-tidy'

jonmeow:
This sets things up to use `bazel` to run `clang-tidy` using https://github.com/erenon/bazel_clang_tidy.

I'm fixing issues outside of explorer, and disabling clang-tidy for targets in explorer that have legacy issues. I was going to disable clang-tidy for targets in explorer such as interpreter anyways, because they're slow to parse, and just extended that to the currently failing targets.


chandlerc (approved):
Nice!!!

PR #3560, opened 2024-01-03
Use typed parse node ids in SemIR instruction types

josh11b:
This involves a number of supporting changes:
* The `parse_node;` member of instruction types may now have any type derived from `Parse::NodeId` and is no longer required to have that exact type.
* `Parse::Node::Invalid` is now a singleton object of a separate type that is convertible to `Parse::NodeId` and its descendants. This replaces the `Invalid` member of its descendants, and avoids having to write long `NodeIdOneOf<...>` types when initializing variables to invalid.
* `IndexBase` now allows `==` and `!=` comparisons between its derived classes and types that are convertible to those types.
* A number of functions in the check stage have been changed to preserve more type information instead of using `Parse::NodeId`.
* `NodeIdForKind<K>` (also known as `KId`) now has a `Kind` member so it may be used to declare `NodeIdOneOf<T, U>` types without #including `parse/typed_nodes.h`.
* `NodeIdForKind<K>` (also known as `KId`) may be implicitly converted to `NodeIdOneOf<T, U>` if `T::Kind == K` or `U::Kind == K` (executing a TODO).

Many of the `parse_node` members were not converted since they would have required more extensive changes. They have been marked with "TODO" comments.

### DIFF
--- a/toolchain/sem_ir/typed_insts.h
+++ b/toolchain/sem_ir/typed_insts.h
@@ -326,6 +326,7 @@ struct FunctionDecl {
 struct Import {
   static constexpr auto Kind = InstKind::Import.Define("import");
 
+  // TODO: Should always be an ImportDirectiveId?
   Parse::NodeId parse_node;
   TypeId type_id;
   CrossRefIRId first_cross_ref_ir_id;
@@ -349,7 +350,8 @@ struct InitializeFrom {
 struct InterfaceDecl {
   static constexpr auto Kind = InstKind::InterfaceDecl.Define("interface_decl");
 
-  Parse::NodeId parse_node;
+  Parse::NodeIdOneOf<Parse::InterfaceDecl, Parse::InterfaceDefinitionStart>
+      parse_node;
   // No type: an interface declaration is not itself a value. The name of an
   // interface declaration becomes a facet type value.
   // TODO: For a generic interface declaration, the name of the interface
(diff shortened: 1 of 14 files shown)


toolchain/sem_ir/typed_insts.h
```diff
@@ -612,16 +655,19 @@ struct ValueOfInitializer {
...
-// HasParseNodeMember<T> is true if T has a `Parse::NodeId parse_node` field.
-template <typename T, typename ParseNodeType = Parse::NodeId T::*>
+// HasParseNodeMember<T> is true if T has a `U parse_node` field,
+// where `U` extends `Parse::NodeId`.
+template <typename T, typename NodeIdType = Parse::NodeId>
 inline constexpr bool HasParseNodeMember = false;
 template <typename T>
-inline constexpr bool HasParseNodeMember<T, decltype(&T::parse_node)> = true;
+inline constexpr bool HasParseNodeMember<
+    T, decltype(Parse::NodeId(std::declval<T>().parse_node))> = true;
```
zygoloid: Maybe it's better to check for inheritance rather than explicit convertibility? Not a strong preference though.
```suggestion
    T, std::is_base_of_v<Parse::NodeId, decltype(T::parse_node)>> = true;
```

zygoloid (approved):

toolchain/sem_ir/typed_insts.h
josh11b: Accepted and fixed.

PR #3561, opened 2024-01-03
Factor code out of templates to reduce compile time of `parse/extract.cpp`

josh11b:
Unfortunately, it is a relatively small improvement (~5%, though noisy) in compile time for that file in my testing.


jonmeow (approved):
Per mention, where you had comments on the struct, you might keep those on the struct (versus where they're now above the helper)

PR #3564, opened 2024-01-04
Roadmap for 2024 and a retrospective for 2023

chandlerc:
We propose a roadmap for 2024 focused on a working Carbon toolchain that
supports Carbon ↔ C++ interop.


docs/project/roadmap.md
```diff
@@ -10,120 +10,108 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
...
+## Beyond 2024
 
 Longer term goals are hard to pin down and always subject to change, but we want
 to give an idea of what kinds of things are expected at a high level further out
-in order to illustrate how the goals and priorities we have in 2023 feed into
+in order to illustrate how the goals and priorities we have in 2024 feed into
 subsequent years.
 
-### Potential 2024 goals: ship a working 0.1 language for evaluation
+### Potential 2025 goals: ship a working 0.1 language for evaluation
```
jonmeow: Given the mention of 0.1 is being removed from the roadmap, should a link to the milestones doc be added?

docs/project/roadmap.md
```diff
@@ -10,120 +10,108 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
...
+Carbon. We still expect some work on language design, but for its priority to be
+driven largely as a function of being in the critical path of some aspect of our
+implementation work.
 
-While we have talked about our 0.1 language, or our Minimum Viable Product (MVP)
-for evaluation purposes, we need to pin down exactly what this includes. We need
-concrete milestones that need to be reached for us to have confidence in
-potential users and communities being able to evaluate Carbon as a successor to
-C++.
+## Key results in 2024
```
jonmeow, line 47: Diagnostic readability has been a significant focus of development, and leads have encouraged working on it, but it's not mentioned either here or in milestones.md. It's significant effort and I don't think it's necessary for other features. Should diagnostic readability be in at least one of these two docs?

docs/project/roadmap.md
```diff
@@ -10,120 +10,108 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
...
+toolchain. This doesn't include supporting everything or even moderately complex
+builds, only the simplest of builds using these build systems need to work at
+first.
 
-We expect to complete the level of Carbon Explorer validation of the design
-needed for 0.1 in 2023 so that we have high confidence in the design's cohesion
-and behavior.
-
-### A toolchain that can build a minimal mixed C++ and Carbon program
+### Carbon's has a design and toolchain implementation of basic C++ interop
```
jonmeow:
```suggestion
### Carbon has a design and toolchain implementation of basic C++ interop
```

docs/project/roadmap.md
```diff
@@ -10,120 +10,108 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
...
 
-While we have talked about our 0.1 language, or our Minimum Viable Product (MVP)
-for evaluation purposes, we need to pin down exactly what this includes. We need
-concrete milestones that need to be reached for us to have confidence in
-potential users and communities being able to evaluate Carbon as a successor to
-C++.
+## Key results in 2024
 
-We expect this to include a reasonably precise set of requirements across:
+### Carbon's toolchain implements the language enough to support interop with C++
```
jonmeow: This goal seems to overlap with "### Carbon's has a design and toolchain implementation of basic C++ interop". FWIW, the latter list seems more focused on features to support interop, whereas this list seems more broad language features not directly associated with interop; for example, it's not clear to me how operator overloading would fit in with interop; are you setting a goal for cross-language operator overloading?

docs/project/roadmap.md
```diff
@@ -10,120 +10,108 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
...
--   Necessary documentation, strategy, or other supporting material
+Some example language features that we think are key to success here but far
+from an exhaustive list:
 
-We expect this to also include at least enough of Carbon's features and C++
-interop to support [our basic example](/docs/images/snippets.md#mixed).
+-   Imports
+-   Operator overloading and dispatch for expressions
+-   Generic types and functions
+-   Templates (at least enough for interop)
```
jonmeow: To be sure, if we don't implement any interop that would need template support this year, does that mean no template support?

Note though, template support might have substantial cost, even minimal template support.

docs/project/roadmap.md
chandlerc: I'm not sure exactly what things you're thinking of here, so sorry if this is too vague of an answer...

But generally, I think the milestones document is the better place to try to cover the set of things needed for 0.1. I suspect some level of diagnostic quality is going to be needed for 0.1, but there are also probably some levels of investment into diagnostics not needed. If we have good ways of calling out the things that would be needed, we should, but I've not thought deeply about how to identify that.

Separately, there are times where we want to build our infrastructure so that it *can* support the level of diagnostic quality we expect to want long-term. That's a tricky thing to defer even though we may not be trying to completely leverage it within the 0.1 timeframe. Sometimes, we can. But other times, we can't and I suspect we should build "the right" infrastructure from the start. Again, if anywhere, I would expect to cover this in the milestones document, but not sure exactly how. Happy for suggestions on that front though.

docs/project/roadmap.md
chandlerc: No, this is about implementing enough of the Carbon language design that we can write useful code on the Carbon side of the interop. The goal below is the only one talking about actual interop features. Happy for suggestions to clarify the text?

As a concrete example, all I mean by "operator" stuff below is that we likely need things like `+` to work, and work for more than one type in Carbon code. Otherwise we won't really be able to write interesting code on the Carbon side of interop.

docs/project/roadmap.md
chandlerc: Sure, I would prioritize template support here pretty specifically at the same level as template support in interop. If we aren't getting towards wanting interop to support that, then it we also probably aren't going to need it on the Carbon side.

docs/project/roadmap.md
```diff
@@ -10,120 +10,108 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
...
 
-Note that we don't expect to finish implementing the 0.1 design in 2023. Our
-goal is to make sufficient progress that we can complete the implementation in
-2024, but there are still many things that can go wrong and cause significant
-delays.
+### Carbon's toolchain can build C++ code
 
-### Complete design coverage of the 0.1 language's necessary features
+We need the toolchain to be able to build C++ code as-if it were Clang in order
```
zygoloid:
```suggestion
We need the toolchain to be able to build C++ code as if it were Clang in order
```

docs/project/roadmap.md
```diff
@@ -10,120 +10,108 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
...
 
-This year we plan to finish the design of the necessary feature-set that we
-define above for the 0.1 language.
+### Carbon's toolchain works with existing, simple C++ build systems
 
-### Complete 0.1 language implementation coverage in the Carbon Explorer
+We should be able to drop Carbon's toolchain into at least simple `Makefile` or
+CMake build systems as a replacement for the C++ toolchain and provide a Carbon
+toolchain. This doesn't include supporting everything or even moderately complex
+builds, only the simplest of builds using these build systems need to work at
```
zygoloid:
```suggestion
builds; only the simplest of builds using these build systems need to work at
```
(comma splice)

docs/project/roadmap.md
```diff
@@ -10,120 +10,108 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
...
-the Carbon 0.1 language becomes increasingly concrete. This should help set the
-stage for evaluations of Carbon when 0.1 is finished and available. To broaden
-our engagement, we want to give talks at 2-3 conferences spanning 2-3 geographic
-regions.
-
-We also want these talks to provide the C++ community a deeper understanding of
-the Carbon language and project, spanning 3-4 different topics. For example, we
-might share the details of the language design, governance, and implementation.
+We want to continue to engage with the external C++ community as the Carbon
+toolchain becomes a more real and working toolchain. We specifically want to
```
zygoloid:
```suggestion
toolchain becomes a more real and complete toolchain. We specifically want to
```

docs/project/roadmap.md
```diff
@@ -10,120 +10,108 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
...
+share as interop becomes something people can experiment with and explore.
```
zygoloid:
```suggestion
share when interop becomes something people can experiment with and explore.
```
maybe?

zygoloid (commented):
I'm happy with the content here, just some copy-edits for the roadmap.

docs/project/roadmap.md
jonmeow: Per discussion, I think we concluded that we probably won't be investing much into more advanced diagnostics next year (and possibly for 0.1). So not really needing changes.

docs/project/roadmap.md
jonmeow: Maybe something like "Carbon's toolchain supports interesting native code" or "supports generic code" (since that's the focus)?

zygoloid:
```suggestion
### Carbon's toolchain implements the pure-Carbon prerequisites for C++ interop
```

chandlerc: I think this gets a bit *too* precise...

I've tried a hybrid of Jon's suggestion and the wording from the objective with a clarifying paragraph. What do folks think?

docs/project/roadmap.md
chandlerc: Yeah, good idea. Done.

chandlerc (commented):
Thanks for all the updates, hopefully converging on the feedback so far!

docs/project/roadmap.md
```diff
@@ -45,19 +45,24 @@ implementation work.
...
+
+This goal is not necessarily about complete support for the entire language
+design, but rather enough of it to support building the realistic and
+interesting Carbon code that interoperates with C++. This does include
+prerequisites like the prelude.
 
 Some example language features that we think are key to success here but far
 from an exhaustive list:
 
 -   Imports
```
jonmeow:
```suggestion
interesting Carbon code that interoperates with C++.

Some example language features that we think are key to success here but far
from an exhaustive list:

-   Imports and prelude
```

docs/project/roadmap.md
jonmeow: Thanks, looks good. (although... bouncing in my head what the "realistic" standard is)

chandlerc: Yeah, the realistic is a bit of a judgement call. But I think it's OK to not have that level of precision. Mostly, I don't think we're going to be that confused about the direction or priorities because of different ideas of what's realistic here.

docs/project/roadmap.md
chandlerc: Sure -- had to manually apply this as Github didn't give me a button and added a couple of words.

zygoloid (approved):
LGTM

chandlerc:
After chatting with @KateGregory, going to add that we should be at least starting to explore building some more documentation and educational material by the end of the year given where other things will be. Stay tuned for concrete wording update....

proposals/p3564.md
```diff
@@ -0,0 +1,104 @@
...
+
+Our [roadmap for 2023] focused on getting both the language and tools ready for
+evaluation, and building of context in the C++ community to support that
+evaluation. We executed on this really well, but with mixed results.
+
+[Roadmap for 2023]:
+    https://github.com/carbon-language/carbon-lang/blob/840cb1bed7cf9bd57e000cb4a61e986c383d3038/docs/project/roadmap.md
+
+On getting ready for evaluation, we made fantastic progress on getting the
+language (design) ready. We have milestone definitions [link], and closed the
```
josh11b: Need to fill in this link

chandlerc: Good catch, fixed!

chandlerc (commented):
Added a key result around actually starting to explore what our tutorial or introductory material needs to look like as the interop starts to come online in the toolchain.

Goal is just to indicate that we should be *starting* this year, this will in large part feed into what we need to do going forward.

Can see the specific commit adding this here:
[`cb12fb1` (#3564)](https://github.com/carbon-language/carbon-lang/pull/3564/commits/cb12fb1d5bd526d399d2a8d1ff611389a3a337af)

Also another fix from code review!

docs/project/roadmap.md
```diff
@@ -95,6 +98,18 @@ We want to continue to engage with the external C++ community as the Carbon
...
+
+Because of the nature of Carbon's experiment, the tutorial and introductory
+material won't be focused on typical teaching of the language to general
+developers. Instead, it will be focused on enabling C++ developers to start
+evaluating specific aspects of Carbon for interoperating with existing C++
+codebases.
+
+We only expect to _start_ building this material in 2024. We want to learn about
+any critical gaps are for folks to start evaluating C++ interop and how best to
+close them going into 2025.
```
zygoloid, line 111:
```suggestion
what any critical gaps are for folks to start evaluating C++ interop and how best
to close them going into 2025.
```

zygoloid (approved):

chandlerc:
Checked with all the leads and we're happy with this roadmap, so landing! Thanks all!!!

PR #3565, opened 2024-01-04
Add `Tree::TryAs` to cast to a typed parse node id

josh11b:


toolchain/parse/tree.h
```diff
@@ -125,6 +125,18 @@ class Tree : public Printable<Tree> {
     return !node_has_error(id);
   }
 
+  // Converts to `n` to a constrained node id `T` if the `node_kind(n)` matches
```
zygoloid:
```suggestion
  // Converts `n` to a constrained node id `T` if the `node_kind(n)` matches
```

toolchain/parse/tree.h
```diff
@@ -125,6 +125,18 @@ class Tree : public Printable<Tree> {
...
+  template <typename T>
+  auto TryAs(NodeId n) const -> std::optional<T> {
+    CARBON_DCHECK(n.is_valid());
+    if (ConvertTo<T>::AllowedFor(node_kind(n))) {
+      return T(n);
+    } else {
+      return std::nullopt;
+    }
+  }
+
```
zygoloid, line 139: I think a non-`Try` version would also be useful: I'd use it [here](https://github.com/carbon-language/carbon-lang/pull/3555/files#diff-7787b79ed2ced26e22d4cde2cc2fc0f77f1474430b2cb551dee017e1992c3d8aR78) if it existed.

```suggestion

  // Converts to `n` to a constrained node id `T`. Checks that the `node_kind(n)`
  // matches the constraint on `T`.
  template <typename T>
  auto As(NodeId n) const -> T {
    CARBON_DCHECK(n.is_valid());
    CARBON_CHECK(ConvertTo<T>::AllowedFor(node_kind(n)));
    return T(n);
  }

```

zygoloid (approved):
LG, happy for this to go head without `As`.

toolchain/parse/tree.h
josh11b: Added, with a test.

PR #3566, opened 2024-01-04
Refactor BindName to support tracking the enclosing scope.

jonmeow:
This is a step towards adding enclosing scopes for imports. It creates an indirection for all bind names.

We discussed specializing for bindings that are in function scope (i.e., not a useful enclosing scope for imports or diagnostics). However, the thought is to go ahead with this singular approach for now, and only change structure if it's a performance issues so that we have incrementally fewer instructions to handle.


chandlerc (approved):
LG

PR #3567, opened 2024-01-04
Check clang-tidy in tests.yaml

jonmeow:
Example success: https://github.com/carbon-language/carbon-lang/actions/runs/7414437667/job/20175507192
Example failure: https://github.com/carbon-language/carbon-lang/actions/runs/7414486486/job/20175656728

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -6,7 +6,7 @@ name: test
 
 on:
   push:
-    branches: [trunk, action-test]
+    branches: [trunk]
   pull_request:
   merge_group:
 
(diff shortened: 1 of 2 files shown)


josh11b (approved):

.github/workflows/tests.yaml
```diff
@@ -307,6 +313,20 @@ jobs:
...
+      # The `-k` flag is used to print all clang-tidy errors.
+      - name: clang-tidy
+        if:
+          steps.filter.outputs.has_code == 'true' && matrix.os == 'ubuntu-22.04'
+          && matrix.build_mode == 'clang-tidy'
+        env:
+          TARGETS_FILE: ${{ runner.temp }}/targets
+        run: |
+          bazelisk build --config=clang-tidy -k \
+            --target_pattern_file=$TARGETS_FILE
```
chandlerc, line 328: Hmm, I'm a bit worried we'll want similar retry logic here as we have for tests. But maybe not?

I'm happy to try it out and see how it goes.

chandlerc (approved):

.github/workflows/tests.yaml
jonmeow: I *think* we don't need retry here due to the difference in what's executing, although I may be misunderstanding the source of failures.

PR #3569, opened 2024-01-04
Support defining instruction categories with a common representation

zygoloid:
Add a mechanism to define instruction categories, to support inspecting the common representation of similar kinds of instruction. Use that mechanism to make formatting of branch instructions slightly more type-safe.

The idea here is to use the existing `Inst` mechanism for converting to and from structs, extended to operate on a struct representing multiple different kinds of instruction. In this case, the concrete kind of instruction is stored in the struct in a `kind` field, rather than being implied by the type.

Factored out of #3555 where this mechanism is used to provide a common interface for runtime and symbolic name bindings.

### DIFF
--- a/toolchain/sem_ir/inst.h
+++ b/toolchain/sem_ir/inst.h
@@ -64,6 +64,8 @@ struct InstLikeTypeInfo<
                 "Instruction type should not have a kind field");
   static auto GetKind(TypedInst) -> InstKind { return TypedInst::Kind; }
   static auto IsKind(InstKind kind) -> bool { return kind == TypedInst::Kind; }
+  // A name that can be streamed to an llvm::raw_ostream.
+  static auto DebugName() -> InstKind { return TypedInst::Kind; }
 };
 
 // An instruction category is instruction-like.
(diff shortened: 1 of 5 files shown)


toolchain/sem_ir/inst.h
```diff
@@ -88,31 +136,40 @@ class Inst : public Printable<Inst> {
...
   // instruction's kind, and returns the typed instruction.
-  template <typename TypedInst, typename Info = TypedInstArgsInfo<TypedInst>>
+  template <typename TypedInst, typename Info = InstLikeTypeInfo<TypedInst>>
   auto As() const -> TypedInst {
-    CARBON_CHECK(Is<TypedInst>()) << "Casting inst of kind " << kind()
-                                  << " to wrong kind " << TypedInst::Kind;
-    auto build_with_type_id_and_args = [&](auto... type_id_and_args) {
+    CARBON_CHECK(Is<TypedInst>())
+        << "Casting inst of kind " << kind() << " to wrong kind "
+        << typeid(TypedInst).name();
```
josh11b: My experience was that typeid names were not pretty. Can we get the kind name in those cases where it is a single kind by having a name member of the info type?

josh11b (approved):

toolchain/sem_ir/inst.h
zygoloid: Done.

josh11b (approved):

PR #3571, opened 2024-01-04
Update setup instructions for recent debian / ubuntu

zygoloid:
`pip3 install` refuses to perform a system-wide installation, but we can install `pre-commit` with `apt` instead.

### DIFF
--- a/docs/project/contribution_tools.md
+++ b/docs/project/contribution_tools.md
@@ -50,7 +50,10 @@ sudo apt install \
   libc++-dev \
   lld \
   python3 \
-  pre-commit
+  pipx
+
+# Install pre-commit.
+pipx install pre-commit
 
 # Set up git.
 # If you don't already have a fork:
--- a/docs/project/contribution_tools.md
+++ b/docs/project/contribution_tools.md
@@ -49,10 +49,8 @@ sudo apt install \
   gh \
   libc++-dev \
   lld \
-  python3
-
-# Install pre-commit.
-pip3 install pre-commit
+  python3 \
+  pre-commit
 
 # Set up git.
 # If you don't already have a fork:


jonmeow:
Knowing your environment, I might suggest `pipx` instead of `pip3` (this is probably mentioned in the error message you're seeing). I'm not sure how generally applicable that advice is, though... Equally, note that the version of pre-commit you have available via apt is out-of-date.

Have you checked how this does on general distros though? https://packages.ubuntu.com/search?keywords=pre-commit&searchon=names&suite=all&section=all suggests versions aren't up-to-date for anything but noble (which I think is bleeding edge). We've had some issues with version skew in pre-commit so I'm hesitant here.

zygoloid:
> Knowing your environment, I might suggest `pipx` instead of `pip3` (this is probably mentioned in the error message you're seeing).

Yeah, `pipx` is mentioned as a fallback at least for non-Debian-packaged python binaries. That seems like a better thing for us to suggest here than `pip3`.

> Equally, note that the version of pre-commit you have available via apt is out-of-date.

In my environment at least, `apt` has version 3.5.0 available, which I've been using successfully on another machine; that seems to be new enough. We don't seem to list a minimum version; maybe we should.

> Have you checked how this does on general distros though? https://packages.ubuntu.com/search?keywords=pre-commit&searchon=names&suite=all&section=all suggests versions aren't up-to-date for anything but noble (which I think is bleeding edge). We've had some issues with version skew in pre-commit so I'm hesitant here.

Yeah, that doesn't look all that recent. Perhaps we should be suggesting installing and using `pipx` instead.

jonmeow (approved):
The sad thing for me is I'd kind of prefer recommending apt because it autoupdates (versus pip, which doesn't). But I think this is probably better.

PR #3572, opened 2024-01-05
Run clang-tidy on headers

jonmeow:
This patches bazel_clang_tidy handling of headers. I found an equivalent change at https://github.com/erenon/bazel_clang_tidy/pull/13, but that was [already rejected](https://github.com/erenon/bazel_clang_tidy/pull/13#issuecomment-1047007424). Per the criticism, this will result in redundant processing of headers.

The project instead uses `HeaderFilterRegex: ".*"`, but that results in two problems:

1. When running with `-k`, errors are repeated when a header is included more than once, which is common.
2. clang-tidy including errors from headers that are included from other modules (e.g., abseil-cpp); filtering correctly is difficult.

Given the trade-offs and options (including forking), I thought patching was preferable so long as it remains narrow.


.pre-commit-config.yaml
```diff
@@ -232,6 +232,7 @@ repos:
 exclude: |
   (?x)^(
       MODULE.bazel.lock|
+      bazel/clang_tidy/.*\.patch|
```
chandlerc: Maybe `bazel_clang_tidy` to avoid any ambiguity with a patch to `clang-tidy` itself?

chandlerc (approved):
LG, and makes sense about the tradeoff.

I'm gonna follow-up on the PR to try and advocate for it to get merged, but we'll see what happens.

.pre-commit-config.yaml
jonmeow: Done

PR #3573, opened 2024-01-05
Put check stage handle functions in execution order

josh11b:
No changes other than moving code and adding section comments.


chandlerc:
While I often like orders based on the order in which code is executed, it does seem a little hard to keep up-to-date, etc. Curious what others think about this (or maybe there was already a discussion I missed?) -- @jonmeow @zygoloid

zygoloid:
@josh11b and I had discussed this late last year, and we both thought it would make these files more readable for them to be in execution order.

chandlerc (approved):
LG given the discussion.

PR #3574, opened 2024-01-06
Choice parsing in toolchain

kshokhin:
Implement Choice parsing in toolchain according to [design](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/sum_types.md).


toolchain/parse/typed_nodes_test.cpp
```diff
@@ -284,5 +284,84 @@ TEST_F(TypedNodeTest, CategoryMatches) {
 #include "toolchain/parse/node_kind.def"
 }
 
+TEST_F(TypedNodeTest, Choice) {
```
zygoloid: We don't generally expect unit tests for each kind of typed node. We check that the file tests (eg, testdata/choice/basic.carbon) can be successfully extracted as typed nodes as part of parse tree verification, so there is test coverage for extraction of these new typed nodes even without this added test.

toolchain/parse/typed_nodes.h
```diff
@@ -677,6 +677,37 @@ struct IfExprElse {
...
+// ------------
+
+using ChoiceIntroducer = LeafNode<NodeKind::ChoiceIntroducer>;
+
+struct ChoiceSignature {
+  static constexpr auto Kind =
+      NodeKind::ChoiceDefinitionStart.Define(NodeCategory::None);
+
+  ChoiceIntroducerId introducer;
+  AnyNameComponentId name;
```
zygoloid, line 691: We probably need a `llvm::SmallVector<AnyModifierId> modifiers;` here for things like `public choice X { ... }`.

zygoloid (commented):
The changes here look good to me and seem to be largely complete, and the tests are passing, but this PR has been in draft status for a week or so, so I wanted to check: are you looking for review / feedback, or are you still planning to do further work on this prior to asking for a review? I've left a couple of comments in case you want feedback (additionally: some negative tests for parse errors would be great), but generally I'm excited to see this getting worked on :) Thank you!

kshokhin:
> The changes here look good to me and seem to be largely complete, and the tests are passing, but this PR has been in draft status for a week or so, so I wanted to check: are you looking for review / feedback, or are you still planning to do further work on this prior to asking for a review? I've left a couple of comments in case you want feedback (additionally: some negative tests for parse errors would be great), but generally I'm excited to see this getting worked on :) Thank you!

Thanks for the feedback!

The PR is a draft mostly because there's almost no parse errors handling, it's still work to be done.

I have also a couple of questions and wanted to ask them in the discord, and waited for the access to the toolchain channel. However there's still no access for some reason.

The main purpose of making this PR as a draft was to gain the initial feedback on the overall approach and make sure that I'm not completely messing the things up :)

toolchain/parse/typed_nodes.h
```diff
@@ -677,6 +677,39 @@ struct IfExprElse {
...
+using ChoiceAlternativeListComma =
+    LeafNode<NodeKind::ChoiceAlternativeListComma>;
+
+struct ChoiceDefinition {
+  static constexpr auto Kind =
+      NodeKind::ChoiceDefinition.Define(NodeCategory::Decl);
+
+  ChoiceDefinitionStartId signature;
+  struct Alternative {
+    AnyNameComponentId name;
```
zygoloid: I think we only want to allow simple identifier names here. I don't have a strong preference as to whether that's validated in the parser or in check, but either way, we should have a test covering this case, for example:
```
choice X { A.B }
```

toolchain/parse/handle_choice.cpp
```diff
@@ -0,0 +1,97 @@
...
+                             ExpectedChoiceAlternativeName);
+    }
+
+    context.SkipPastLikelyEnd(*context.position());
+
+    state.state = State::ChoiceAlternativeFinish;
+    state.subtree_start = {};
+    state.token = *context.position();
+    state.has_error = true;
+    context.PushState(state);
```
zygoloid: I think we're not getting any benefit from reusing the previous state here, given that we overwrite all of its fields. I think it'd be a little more obvious what we're doing if we push a fresh state:
```suggestion
    context.PushState(State::ChoiceAlternativeFinish);
    context.ReturnErrorOnState();
```
(It might also be reasonable to move the `PushState(ChoiceAlternativeFinish)` calls from here and line 68 to before the `if (!context.PositionIs...`, given that we do that unconditionally.)

zygoloid (commented):
Thank you, this looks great! Just a couple of trivial things, then we can merge this.

zygoloid (approved):
Awesome, thank you!

PR #3575, opened 2024-01-08
Add imports of enclosing scopes.

jonmeow:
Namespaces are copied, which means also adding their name to the underlying instruction. It happened not to be done previously; the name was only in name lookup.

Since the only import supported right now is the default import, functionality is limited; in the future I'll need to deal with namespace vs package conflicts.

Tests of namespace imports are under "namespace" -- I figured this would be best for scaling as more instructions get support.

This also improves some debugging-related output that I was trying to use while trying to build the support.


toolchain/sem_ir/ids.h
```diff
@@ -187,8 +187,8 @@ struct NameId : public IdBase, public Printable<NameId> {
   static auto ForIdentifier(IdentifierId id) -> NameId {
     // NOLINTNEXTLINE(misc-redundant-expression): Asserting to be sure.
     static_assert(NameId::InvalidIndex == IdentifierId::InvalidIndex);
-    CARBON_CHECK(id.index >= 0 || id.index == InvalidIndex)
-        << "Unexpected identifier ID";
+    CARBON_CHECK(id.index >= 0 || !id.is_valid())
```
zygoloid: For me this change makes it less clear that the code is correct. Previously the combination of the `static_assert` and the `CHECK` demonstrated that it's correct to map `id.index` directly into `NameId` without changing the value, and now as a reader I'm wondering if there are other values of `id.index` for which `is_valid()` might return `false`, and whether they're handled properly.

toolchain/sem_ir/value_stores.cpp
```diff
@@ -47,4 +47,32 @@ auto NameStoreWrapper::GetIRBaseName(NameId name_id) const -> llvm::StringRef {
...
+  for (auto [key, _] : names) {
+    keys.push_back(key);
+  }
+  llvm::sort(keys,
+             [](NameId lhs, NameId rhs) { return lhs.index < rhs.index; });
+  llvm::ListSeparator key_sep;
+  for (auto key : keys) {
+    out << key_sep << key << ": " << names.find(key)->second;
+  }
+  out << "}";
```
zygoloid, line 73: Is it feasible to print the names out as a `-`-bulleted list rather than an inline dictionary? I think this formatting will make the raw IR hard to read for large name scopes, but I don't think it's urgent to address this if it's not trivial to do so.

toolchain/check/check.cpp
```diff
@@ -74,46 +77,169 @@ struct UnitInfo {
...
+    if (auto it = copied_namespaces.find(import_enclosing_scope_id);
+        it != copied_namespaces.end()) {
+      // We inject names at the provided scope, and don't need to keep
+      // traversing parents.
+      scope_cursor = it->second;
+      break;
+    }
+
+    // The namespace hasn't been copied yet, so add it to our list.
+    auto scope = import_sem_ir.name_scopes().Get(import_enclosing_scope_id);
```
zygoloid: Accidental `NameScope` copy.
```suggestion
    auto& scope = import_sem_ir.name_scopes().Get(import_enclosing_scope_id);
```

toolchain/check/check.cpp
```diff
@@ -74,46 +77,169 @@ struct UnitInfo {
...
-      return import_sem_ir.bind_names().Get(bind_name.bind_name_id).name_id;
+      auto bind_name = import_sem_ir.bind_names().Get(
+          import_inst.As<SemIR::BindName>().bind_name_id);
+      return {bind_name.name_id, bind_name.enclosing_scope_id};
     }
 
     case SemIR::InstKind::FunctionDecl: {
-      auto bind_name = import_inst.As<SemIR::FunctionDecl>();
-      return import_sem_ir.functions().Get(bind_name.function_id).name_id;
+      auto function = import_sem_ir.functions().Get(
```
```suggestion
      auto& function = import_sem_ir.functions().Get(
```
Accidental `Function` copy.

toolchain/check/check.cpp
```diff
@@ -74,46 +77,169 @@ struct UnitInfo {
...
+          import_inst.As<SemIR::FunctionDecl>().function_id);
+      return {function.name_id, function.enclosing_scope_id};
+    }
+
+    case SemIR::InstKind::Namespace: {
+      auto namespace_inst = import_inst.As<SemIR::Namespace>();
+      auto scope =
```
```suggestion
      auto& scope =
```
Accidental `NameScope` copy.

toolchain/check/check.cpp
```diff
@@ -74,46 +77,169 @@ struct UnitInfo {
...
+    new_namespaces.push_back({scope_inst.name_scope_id, scope_inst.name_id});
+    import_enclosing_scope_id = scope.enclosing_scope_id;
+  }
+
+  // Add enclosing namespace names, starting with the outermost.
+  for (auto import_namespace : llvm::reverse(new_namespaces)) {
+    auto name_id =
+        CopyNameFromImportIR(context, import_sem_ir, import_namespace.second);
+    auto scope = context.name_scopes().Get(scope_cursor);
+    auto [it, success] = scope.names.insert({name_id, SemIR::InstId::Invalid});
```
zygoloid, line 197: Note that `scope` here is a *copy* of the `NameScope`, so this will be inserting into a copy of the scope's `names` list, not into the list actually stored on the `NameScope`. It took me a while to figure out how this is giving the right results anyway. It turns out that we insert this namespace name twice -- once here, into the copy of the scope list, and again in `AddNamespace` below, into the actual scope.

Aside from not making the `scope` copy, I think we should refactor these insertions so we only do the lookup of `name_id` in `scope_cursor` once.

toolchain/sem_ir/value_stores.cpp
jonmeow: It requires correct indentation, so it would be non-trivial. Note the entire `Print` function would need to use `-` formatting and be indent aware.

The easiest way to do this would probably be to change ValueStore to require YamlPrintable on every data type, so that yaml-structured data could be returned here. Note that'd also require changing various ids and structured data used with ValueStore to be YamlPrintable, too.

Although I find the raw IR useful for debugging, I'm not sure how much others use it: Class and Interface lack detail, newer features haven't been getting added to Function printing. Given the whole push towards textual IR, to the extent it _could_ be done, is it worthwhile?

This does make me note the equivalent format in textual IR, though. Whereas the raw IR is "raw" and not really expected to be easy to read, the textual IR is AFAIU intended to be simpler. Also, I think the necessary changes would be more constrained since indentation is a known factor. Maybe there's more utility in changing the textual IR format of name scopes?

toolchain/check/check.cpp
jonmeow: Fixed

toolchain/sem_ir/ids.h
jonmeow: Does reversing the order help your reading by shifting emphasis to more "if it's valid, then we also need to check the index"?

For contrast, I think we use `is_valid()` in a lot of places, and preferentially to `Invalid` (or `InvalidIndex`) comparisons. Whereas you express uncertainty about `is_valid`, I have uncertainty about why it wasn't being used -- if I'm misunderstanding what's to be achieved. My hope is that consistent API use improves understandability. I think the ordering might've lent to my confusion here, so I'm theorizing the swap causes a similar problem for you.

zygoloid: I don't think the reordering really gets to the heart of my concern. We're converting an `IdentifierId` index into a `NameId` index, and the purpose of the prior asserts and `CHECK`s was to make sure that each index we map across is mapped properly -- either it's a non-negative index, which we map directly, or it's `IdentifierId::InvalidIndex`, which we map to `NameId::InvalidIndex` (which we check is the same value, so we don't need to add a special case). But the new code doesn't check the same property -- for example, if we add a new `IdentifierId` special value for which `is_valid` returned `false`, say `IdentifierId::AlsoInvalid = IdentifierId(-2)`, the `CHECK` would not fire, but it should.

Looking for something that addresses both of our desires here, I'd be fine with something like this:

```c++
if (id.index >= 0) {
  return NameId(id.index);
} else if (!id.is_valid()) {
  return NameId::Invalid;
} else {
  CARBON_FATAL("Unexpected identifier ID");
}
```

I think this is just all-round better than how I wrote this function originally: we don't need any `CARBON_CHECK`s or `static_assert`s or `NOLINTNEXTLINE`s this way. And it's clearer that each input produces a meaningful `NameId`.

toolchain/sem_ir/value_stores.cpp
zygoloid: OK, let's leave this as-is then :)

For textual IR, I think we should display namespaces the same way we display classes: as a separate braced top-level entity with a symbolic name. We don't have a `NamespaceId` and a separate `Namespace` object for them, just a `NameScope`, but I guess we can display them as if we did.

zygoloid (approved):

toolchain/sem_ir/ids.h
jonmeow: > for example, if we add a new `IdentifierId` special value for which `is_valid` returned `false`, say `IdentifierId::AlsoInvalid = IdentifierId(-2)`, the `CHECK` would not fire, but it should.

Is there some reason behind your suggestion that we may add more invalid values?

If we added new invalid values, I believe that could make the suggested code erroneous too [and similar for the prior code] -- it depends on why another invalid value is being added. I think the status quo is that _because_ there is a single invalid value, code should be expected to have built-in assumptions that it is singular.

To me, having a singular invalid value is a standard coding pattern. As with most coding patterns, following it allows certain simplifications, and violating can cause cascading issues.

> Looking for something that addresses both of our desires here, I'd be fine with something like this:
> 
> ```c++
> if (id.index >= 0) {
>   return NameId(id.index);
> } else if (!id.is_valid()) {
>   return NameId::Invalid;
> } else {
>   CARBON_FATAL("Unexpected identifier ID");
> }
> ```
> 
> I think this is just all-round better than how I wrote this function originally: we don't need any `CARBON_CHECK`s or `static_assert`s or `NOLINTNEXTLINE`s this way. And it's clearer that each input produces a meaningful `NameId`.

Done.

PR #3576, opened 2024-01-08
Split import logic to its own file.

jonmeow:
This is currently built on top of #3575. I'm just getting to the point where imports have enough logic that I'm thinking they may be able to stand on their own.


chandlerc:
Going to let @zygoloid handle the review here as I'm guessing he'll have a lot more context based on reviewing the underlying import PR. But happy to help out if needed.

zygoloid (approved):

PR #3577, opened 2024-01-09
Add support to decl/init globals of primitive types

gysddn:
This is an experimental strategy to implement globals.

It's only handled in the lowering phase.
I added a new function named `__global_init` and lowered the top_inst_blck with that function. That way I was able to  override the behaviors of some handle functions, only 3 of them at the moment:

- VarStorage  → Creates a global
- BindName   → Sets a global value
- Assign          → Adds an initialization instruction to the init_function

The initialization function is very much like what `clang` generates but only simpler, instead of seperate functions for each global, there is only one for all of them. So llvm.global_ctors "only" points to that.

### DIFF
--- a/toolchain/lower/file_context.cpp
+++ b/toolchain/lower/file_context.cpp
@@ -37,20 +37,18 @@ auto FileContext::Run() -> std::unique_ptr<llvm::Module> {
     types_[type_id.index] = BuildType(sem_ir_->types().Get(type_id).inst_id);
   }
 
+  LowerTopInstBlock();
+
   // Lower function declarations.
   functions_.resize_for_overwrite(sem_ir_->functions().size());
   for (auto i : llvm::seq(sem_ir_->functions().size())) {
     functions_[i] = BuildFunctionDecl(SemIR::FunctionId(i));
   }
 
-  // TODO: Lower global variable declarations.
-
   // Lower function definitions.
   for (auto i : llvm::seq(sem_ir_->functions().size())) {
     BuildFunctionDefinition(SemIR::FunctionId(i));
   }
-
-  // TODO: Lower global variable initializers.
 
   return std::move(llvm_module_);
 }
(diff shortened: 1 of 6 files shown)


jonmeow:
Sorry, I think there may have been a misunderstanding of [my comments here](https://discord.com/channels/655572317891461132/655578254970716160/1191826810179162262).

What I was trying to describe is a way to create `__global_init` during the _check_ phase, rather than _lower_. There are a few reasons we want to go this way, but let me try to outline key reason...

Compile-time constants, including things such as function declarations, have so far gone into the top inst block. So far, most of these have been distinct enough to tell them apart (`FatalErrorIfEncountered`). However, as we get into generics, we're going to have more constants used in the creation of types (e.g., `FixedArray(17)` creates a constant `17`). The instruction for `17` would end up in the top inst block.

Under an approach where lowering uses the top inst block for instructions to put in `__global_init`, it means that instructions which are part of compile-time constant evaluation are more likely to leak into runtime evaluation by accident. We may need to take efforts to prevent that.

While there are probably ways to make that work, we do generally want `SemIR` textual output to reflect the way that programs are expected to evaluate. So for example, [global_decl_with_init](https://github.com/carbon-language/carbon-lang/blob/trunk/toolchain/check/testdata/var/global_decl_with_init.carbon) has:

```
--- global_decl_with_init.carbon

file {
  package: <namespace> = namespace {.x = %x}
  %x.var: ref i32 = var x
  %x: ref i32 = bind_name x, %x.var
  %.loc7: i32 = int_literal 0
  assign %x.var, %.loc7
}
```

What we would really want is something more like:

```
--- global_decl_with_init.carbon

file {
  package: <namespace> = namespace {.x = %x}
  %__global_init: <function> = fn_decl @__global_init
  %x.var: ref i32 = var x
  %x: ref i32 = bind_name x, %x.var
}

fn @__global_init() {
!entry:
  %.loc7: i32 = int_literal 0
  assign %x.var, %.loc7
  return
}
```

In SemIR, we may want to augment the storage of global variables to help lowering emit it; for example, storing them in a way that they're easy to traverse, similar to how functions are easy to find without looking at the top block.

Once checking emits a `__global_init` function, the amount of work needed in lowering to hook that up to `global_ctors` should be more minimal. It's really shifting the implementation burden more to checking, because we want to limit the semantic logic done within lowering. Does that make sense?

gysddn:
Oh, it makes so much more sense now.
I had lots of doubts going this way, but this cleared things up.

Think I'll go ahead and try that approach.

PR #3579, opened 2024-01-09
Add `reify_constant` SemIR node wrapping each constant-phase expression.

zygoloid:
Draft: exploration of adding a SemIR node after each constant-phase expression to reify the constant and track the constant value.


zygoloid:
Closing, based on toolchain discussion we prefer #3581, at least for now.

PR #3581, opened 2024-01-09
Very basic support for constant evaluation of expressions.

zygoloid:
Form a side table with constant values for each instruction. Evaluation is only supported for a few very simple kinds of instruction for now. This is not observable outside of the SemIR output, because nothing depends on expressions having a constant value phase yet.


toolchain/check/context.cpp
```diff
@@ -437,29 +449,7 @@ auto Context::FollowNameRefs(SemIR::InstId inst_id) -> SemIR::InstId {
...
-      case SemIR::FieldDecl::Kind:
-      case SemIR::FunctionDecl::Kind:
-        return inst_id;
-
-      default:
-        // TODO: Handle the remaining cases.
-        return SemIR::InstId::Invalid;
-    }
-  }
+  return constant_values().Get(inst_id);
```
jonmeow: Should this function just be removed in favor of the direct call?

toolchain/sem_ir/value_stores.h
```diff
@@ -53,6 +53,49 @@ class InstStore {
...
+  // if it is not constant.
+  auto Get(InstId inst_id) const -> InstId {
+    CARBON_CHECK(inst_id.index >= 0);
+    return static_cast<size_t>(inst_id.index) >= values_.size()
+               ? InstId::Invalid
+               : values_[inst_id.index];
+  }
+
+  // Sets the constant value of the given instruction.
+  auto Set(InstId inst_id, InstId const_id) -> void {
```
jonmeow, line 70: Should const_id always be valid for calls, or is invalid expected sometimes?

toolchain/sem_ir/value_stores.h
```diff
@@ -53,6 +53,49 @@ class InstStore {
   ValueStore<InstId> values_;
 };
 
+// Provides a ValueStore wrapper for tracking the constant values of
+// instructions.
```
jonmeow, line 57: Would it be worth explicitly documenting the semantic that values_ will be sparse with InstId::Invalid in entries that lack a constant? That's kind of on Get, but affects semantics of other functions too, and things like OutputYaml don't document what's going on.

toolchain/sem_ir/value_stores.h
```diff
@@ -53,6 +53,49 @@ class InstStore {
...
+      for (auto [id, value] : llvm::enumerate(values_)) {
+        if (value.is_valid()) {
+          map.Add(PrintToString(InstId(id)), Yaml::OutputScalar(value));
+        }
+      }
+    });
+  }
+
+  auto array_ref() const -> llvm::ArrayRef<InstId> { return values_; }
+  auto size() const -> int { return values_.size(); }
```
jonmeow: Are these (and Reserve) used? I can't help but think uses might cause awkward outcomes. I don't really think there's a need for uniformity in API on these details, if that's the only reason they're there. In other cases, I believe they were added due to direct use.

toolchain/sem_ir/value_stores.h
```diff
@@ -53,6 +53,49 @@ class InstStore {
...
+    CARBON_CHECK(inst_id.index >= 0);
+    return static_cast<size_t>(inst_id.index) >= values_.size()
+               ? InstId::Invalid
+               : values_[inst_id.index];
+  }
+
+  // Sets the constant value of the given instruction.
+  auto Set(InstId inst_id, InstId const_id) -> void {
+    CARBON_CHECK(inst_id.index >= 0);
+    if (values_.size() <= static_cast<size_t>(inst_id.index)) {
```
jonmeow: Maybe this is just me, but comparing with line 64, I'd typically write the comparison with index on the left and size on the right. Swapping the order feels like it makes it harder to see differences in how the comparison should function, even if the comparison weren't identical.

```suggestion
    if (static_cast<size_t>(inst_id.index) >= values_.size()) {
```

toolchain/sem_ir/value_stores.h
```diff
@@ -53,6 +53,49 @@ class InstStore {
...
+  // Returns the constant value of the requested instruction, or InstId::Invalid
+  // if it is not constant.
+  auto Get(InstId inst_id) const -> InstId {
+    CARBON_CHECK(inst_id.index >= 0);
+    return static_cast<size_t>(inst_id.index) >= values_.size()
+               ? InstId::Invalid
+               : values_[inst_id.index];
+  }
+
+  // Sets the constant value of the given instruction.
```
jonmeow, line 69: Should this mention the resize semantic?

toolchain/check/eval.cpp
```diff
@@ -0,0 +1,193 @@
...
+  // original ID.
+  return context.inst_blocks().Add(const_insts);
+}
+
+// If the specified operands of the given instruction have constant values,
+// replaces the operands with their constant values and builds a corresponding
+// constant value. Otherwise returns `SemIR::InstId::Invalid`.
+template <typename InstT, typename... SpecificNodeId>
+static auto TryRebuildIfConstOperands(Context& context, SemIR::Inst inst,
+                                      SpecificNodeId InstT::*... each_op_id)
```
jonmeow: Are these really NodeIds? It looks a lot like they're either going to be an InstBlockId or an InstId. Maybe just `IdT`?

toolchain/check/eval.cpp
```diff
@@ -0,0 +1,193 @@
...
+      auto value = context.insts().TryGetAs<SemIR::BoolLiteral>(const_id);
+      if (!value) {
+        // TODO: Can we CHECK this instead?
+        break;
+      }
+      value->value =
+          (value->value == SemIR::BoolValue::False ? SemIR::BoolValue::True
+                                                   : SemIR::BoolValue::False);
+      return context.AddConstantInst(*value);
+    }
```
jonmeow: I see a lot of test edits, but where should I look to see specific tests that are ensuring this rebuild logic is working? For example, with `not`, where in check testdata is `not` constant propagation tested? Are there any examples of either `not true` or `not false`?

toolchain/check/eval.cpp
```diff
@@ -0,0 +1,193 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "toolchain/check/eval.h"
+
+#include "toolchain/sem_ir/typed_insts.h"
+
+namespace Carbon::Check {
+
+// If the given instruction is constant, returns its constant value.
+static auto UnwrapIfConstant(Context& context, SemIR::InstId inst_id)
```
jonmeow: Please document the overload semantic that you're going for somewhere. IIUC you're relying on this for implementing TryRebuild, but it took me some time to puzzle that out.

toolchain/check/eval.cpp
```diff
@@ -0,0 +1,193 @@
...
+    // constant, so it's likely that the entire block is constant. Resize to the
+    // target size given that we're going to allocate memory now anyway.
+    if (const_insts.size() == const_insts.capacity()) {
+      const_insts.reserve(insts.size());
+    }
+
+    const_insts.push_back(const_inst_id);
+  }
+  // TODO: If the new block is identical to the original block, return the
+  // original ID.
```
jonmeow, line 43: Maybe we should just work on more general deduplication for inst_blocks... Even here, I think two independent inst blocks could result in the same constants.

toolchain/check/eval.cpp
```diff
@@ -0,0 +1,193 @@
...
+  return context.inst_blocks().Add(const_insts);
+}
+
+// If the specified operands of the given instruction have constant values,
+// replaces the operands with their constant values and builds a corresponding
+// constant value. Otherwise returns `SemIR::InstId::Invalid`.
+template <typename InstT, typename... SpecificNodeId>
+static auto TryRebuildIfConstOperands(Context& context, SemIR::Inst inst,
```
jonmeow: Should const/constant be consistently abbreviated on function names? e.g. UnwrapIfConstant vs TryRebuildIfConstOperands

toolchain/check/eval.cpp
```diff
@@ -0,0 +1,193 @@
...
+  auto typed_inst = inst.As<InstT>();
+  if (([&] {
+        auto unwrapped = UnwrapIfConstant(context, typed_inst.*each_op_id);
+        if (!unwrapped.is_valid()) {
+          return false;
+        }
+        typed_inst.*each_op_id = unwrapped;
+        return true;
+      }() &&
+       ...)) {
```
jonmeow: What would you think of writing this without a lambda? e.g.:

```
template <typename InstT, typename IdT>
static auto UnwrapIfConstOperand(InstT& typed_inst, IdT InstT::*op_id) {
  auto unwrapped = UnwrapIfConstant(context, typed_inst.*op_id);
  if (!unwrapped.is_valid()) {
    return false;
  }
  typed_inst.*op_id = unwrapped;
  return true;
}

  if (UnwrapIfConstOperand(typed_inst, each_op_id) && ...) {
```

Note, I comment mainly because the lengthy lambda inside the `if`, also with variadics, is a bit difficult to follow.

toolchain/sem_ir/value_stores.h
zygoloid: Sure, done.

(FWIW, for me the difference between these two is that the line 64 comparison is saying "if the index is out of range, then use this other value", and this case is saying "if the size is insufficient, then change it" -- so that the "object" of the comparison is on the left and the "subject" is on the right -- but it seems equally clear to me here to say "if the index doesn't fit in the container then resize the container so it fits".)

toolchain/sem_ir/value_stores.h
zygoloid: Yeah, these are unused. Removed.

I think we might eventually want a `Reserve` if we start reserving an estimated final size for the `InstStore`, but no need to add it speculatively. Removed that too.

toolchain/sem_ir/value_stores.h
zygoloid: Done.

toolchain/sem_ir/value_stores.h
zygoloid: I don't think so; that's an implementation detail, not part of the API contract.

toolchain/check/eval.cpp
zygoloid: Fixed. Also added an `Each` prefix to match the parameter and the nascent carbon variadics design.

toolchain/check/eval.cpp
zygoloid: Done.

toolchain/check/eval.cpp
zygoloid: We already have `CopyOnWriteBlock` in convert.cpp, which could be used directly here. But I'd prefer to address this in a separate PR, if that's OK with you.

toolchain/sem_ir/value_stores.h
zygoloid: Added a CHECK that is is valid.

toolchain/check/eval.cpp
jonmeow: Absolutely -- I was expecting the TODO to remain, just commenting on scope.

toolchain/check/eval.cpp
zygoloid: Switched to `Constant` for now, to match existing uses (eg, `GetConstantValue`). Perhaps we should reserve `const` for `const`-qualified types using the `const` keyword, rather than constant evaluation, but in any case, let's discuss that separately.

toolchain/check/eval.cpp
zygoloid: Sure, done.

toolchain/check/context.cpp

toolchain/check/eval.cpp
jonmeow: FWIW, I think we already use `Addr` as an abbreviation of `Address` which does not specifically refer to the `addr` keyword. I'd be fine with `Const` in general on the same terms -- but fine for that to be separately handled due to existing cases.

jonmeow (approved):

toolchain/check/eval.cpp
zygoloid: Added tests for `not` specifically to `check/testdata/operators/unary_op.carbon`. I also did a pass over the other cases and made sure there was testing for the ones that are reachable, but some of these aren't currently reachable because (for example) we don't have any constant-phase reference or initializing expressions yet.

Added a TODO.

PR #3582, opened 2024-01-09
Change lexical lookup to use an array instead of hash map.

jonmeow:
This is primarily being done for performance reasons, removing hash lookups. The increased memory consumption is accepted.

Refactors LexicalLookup out to its own structure.


toolchain/check/lexical_lookup.h
```diff
@@ -0,0 +1,70 @@
...
+
+ private:
+  // The offset used when accessing lookup_ with NameId::index.
+  static constexpr int Offset = SemIR::NameId::SpecialValueCount + 1;
+
+  StringStoreWrapper<IdentifierId>* identifiers_;
+
+  // Maps identifiers to name lookup results.
+  // TODO: Consider TinyPtrVector<Result> or similar.
+  llvm::SmallVector<llvm::SmallVector<Result>> lookup_;
```
zygoloid: Maybe we could specify a small size of 1 or 2 for the inner vector for the time being, given that we generally don't expect any collisions at all and the default small size (`sizeof(SmallVector<T>) == 64`) is ... not very small when multiplied by the number of identifiers. Though even with a small size of 1, the inner vector will be 24 bytes.

One possibility to get a more dense representation would be to have the vector store a single `Result`, with a special `Result` value meaning "more than one result", and store a separate `llvm::DenseMap<IdentifierId, llvm::SmallVector<Result, N>>` for the uncommon case of more than one result for an identifier.

For what it's worth, Clang stores [an `llvm::SmallVector<NamedDecl*, 2>`](https://github.com/llvm/llvm-project/blob/c9124adfd8291a5f5b1d23295308d8940648c596/clang/include/clang/Sema/IdentifierResolver.h#L45) on each identifier for lookup, so we'd at least be in good company if we just set a small size of 2 here.

```suggestion
  // TODO: Consider TinyPtrVector<Result> or similar. For now, use a small
  // size of 2 to cover the common case.
  llvm::SmallVector<llvm::SmallVector<Result, 2>> lookup_;
```

toolchain/check/scope_index.h
```diff
@@ -0,0 +1,29 @@
...
+// An index for a pushed scope. This may correspond to a permanent scope with a
+// corresponding `NameScope`, in which case a different index will be assigned
+// each time the scope is entered. Alternatively, it may be a temporary scope
+// such as is created for a block, and will only be entered once.
+//
+// `ScopeIndex` values are comparable. Lower `ScopeIndex` values correspond to
+// scopes entered earlier in the file.
+//
+// TODO: Move this struct and the name lookup code in context.h to a separate
+// file.
```
zygoloid: TODO seems out of date. Thanks for addressing this part of it!

toolchain/sem_ir/ids.h
```diff
@@ -225,6 +229,10 @@ constexpr NameId NameId::SelfType = NameId(NameId::InvalidIndex - 2);
 constexpr NameId NameId::ReturnSlot = NameId(NameId::InvalidIndex - 3);
 constexpr NameId NameId::PackageNamespace = NameId(NameId::InvalidIndex - 4);
 constexpr NameId NameId::Base = NameId(NameId::InvalidIndex - 5);
+constexpr int NameId::SpecialValueCount = 5;
+// Enforce the link between SpecialValueCount and the last special value.
+static_assert(NameId::SpecialValueCount ==
+              NameId::InvalidIndex - NameId::Base.index);
```
zygoloid: Given that the uses of this all add one, maybe it'd be simpler to include `InvalidIndex` in the count of special values?

jonmeow: Echoing the discussion around #3575 and invalid, I would feel more comfortable distinguishing between a valid special value and invalid.

Maybe I'll go with NonIndexValueCount.

toolchain/check/lexical_lookup.h
jonmeow: Done

toolchain/check/scope_index.h
jonmeow: Removed (it felt weird to put this in lexical_lookup.h given the other uses -- I chose the separate header to avoid a specific association)

zygoloid (approved):

toolchain/sem_ir/ids.h
zygoloid: Works for me. I think something like `LowestValidValue` could also work; using `id.index - LowestValidValue` to convert an ID to an array index seems like it'd be reasonably readable and clear.

PR #3587, opened 2024-01-10
Change LexicalLookup to not resize, due to lack of current need.

jonmeow:
Per discussion, unqualified lookup should only occur on identifiers that existed at the time Context was initialized. As a consequence, the resize logic in LexicalLookup may not be necessary. Even considering metaprogramming, if metaprogramming is restricted to qualified name lookup, it may not be necessary in the future. Support should be easy to add if we need it. Trying to clearly document in the CHECK message what the rationale is.

### DIFF
--- a/toolchain/check/lexical_lookup.h
+++ b/toolchain/check/lexical_lookup.h
@@ -29,35 +29,26 @@ class LexicalLookup {
     ScopeIndex scope_index;
   };
 
-  explicit LexicalLookup(StringStoreWrapper<IdentifierId>& identifiers)
-      : identifiers_(&identifiers),
-        lookup_(identifiers_->size() + SemIR::NameId::NonIndexValueCount) {}
-
-  ~LexicalLookup() {
-    CARBON_CHECK(lookup_.size() ==
-                 identifiers_->size() + SemIR::NameId::NonIndexValueCount)
-        << lookup_.size() << " must match " << identifiers_->size() << " + "
-        << SemIR::NameId::NonIndexValueCount
-        << "; something may have been added incorrectly";
-  }
-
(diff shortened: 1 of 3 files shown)


toolchain/check/lexical_lookup.h
```diff
@@ -29,35 +29,26 @@ class LexicalLookup {
...
   auto Get(SemIR::NameId name_id) -> llvm::SmallVector<Result, 2>& {
+    CARBON_CHECK(name_id.index < original_identifiers_size_)
+        << "An identifier was added after the Context was initialized. "
+           "Currently, we expect that new identifiers will never be used with "
+           "lexical lookup (they're added for things like detecting name "
+           "collisions in imports). That might change with metaprogramming: if "
+           "it does, we may need to start resizing `lookup_`, either on each "
+           "identifier addition or in Get` where this CHECK currently fires.";
     return lookup_[name_id.index + SemIR::NameId::NonIndexValueCount];
   }
```
zygoloid: Optionally:
```suggestion
  auto Get(SemIR::NameId name_id) -> llvm::SmallVector<Result, 2>& {
    CARBON_CHECK(name_id.index + SemIR::NameID::NonIndexValueCount < lookup_.size())
        << "An identifier was added after the Context was initialized. "
           "Currently, we expect that new identifiers will never be used with "
           "lexical lookup (they're added for things like detecting name "
           "collisions in imports). That might change with metaprogramming: if "
           "it does, we may need to start resizing `lookup_`, either on each "
           "identifier addition or in Get` where this CHECK currently fires.";
    return lookup_[name_id.index + SemIR::NameId::NonIndexValueCount];
  }
```
... and remove `original_identifiers_size_`.

zygoloid (approved):

PR #3590, opened 2024-01-11
Split parse nodes out from instructions because they're rarely used.

jonmeow:
The parse nodes are still tracked as part of the same value store interface in order to ensure parity, but they're split out from Inst itself in order to reduce the size of Inst -- the expectation is that they don't need to be passed around quite as much.

This change doesn't actually reduce the passing very much, although there are hints of it: AddInstAndPush doesn't typically need a separate parse node from the one on the Inst itself, for example. In a couple spots I changed code to rely a little more on the InstId until the ParseNode is needed, but it's very low hanging fruit where done. I think convert could do more to not eagerly fetch the parse node before its use, but more cleanup felt it would be easier to handle separately. I'm currently viewing this as making such cleanup _possible_ rather than executing on it up-front.

But also, I want to make sure there's a consensus to head in this direction before pulling the trigger. We speculated that this would result in the parse node being passed around less, and I do think that's the case, although it's a bit fuzzy in the change.

### DIFF
--- a/toolchain/check/context.cpp
+++ b/toolchain/check/context.cpp
@@ -61,8 +61,9 @@ auto Context::VerifyOnFinish() -> void {
   CARBON_CHECK(params_or_args_stack_.empty()) << params_or_args_stack_.size();
 }
 
-auto Context::AddInst(SemIR::Inst inst) -> SemIR::InstId {
-  auto inst_id = inst_block_stack_.AddInst(inst);
+auto Context::AddInst(Parse::NodeId parse_node, SemIR::Inst inst)
+    -> SemIR::InstId {
+  auto inst_id = inst_block_stack_.AddInst(parse_node, inst);
   CARBON_VLOG() << "AddInst: " << inst << "\n";
 
   auto const_id = TryEvalInst(*this, inst_id, inst);
@@ -75,10 +76,10 @@ auto Context::AddInst(SemIR::Inst inst) -> SemIR::InstId {
   return inst_id;
 }
 
-auto Context::AddConstant(SemIR::Inst inst, bool is_symbolic)
-    -> SemIR::ConstantId {
+auto Context::AddConstant(Parse::NodeId parse_node, SemIR::Inst inst,
+                          bool is_symbolic) -> SemIR::ConstantId {
   // TODO: Deduplicate constants.
-  auto inst_id = insts().AddInNoBlock(inst);
+  auto inst_id = insts().AddInNoBlock(parse_node, inst);
   constants().Add(inst_id);
 
   auto const_id = is_symbolic ? SemIR::ConstantId::ForSymbolicConstant(inst_id)
@@ -91,7 +92,7 @@ auto Context::AddConstant(SemIR::Inst inst, bool is_symbolic)
 
 auto Context::AddInstAndPush(Parse::NodeId parse_node, SemIR::Inst inst)
     -> void {
-  auto inst_id = AddInst(inst);
+  auto inst_id = AddInst(parse_node, inst);
   node_stack_.Push(parse_node, inst_id);
 }
 
@@ -101,9 +102,8 @@ auto Context::DiagnoseDuplicateName(Parse::NodeId parse_node,
                     "Duplicate name being declared in the same scope.");
   CARBON_DIAGNOSTIC(NameDeclPrevious, Note,
                     "Name is previously declared here.");
-  auto prev_def = insts().Get(prev_def_id);
   emitter_->Build(parse_node, NameDeclDuplicate)
-      .Note(prev_def.parse_node(), NameDeclPrevious)
+      .Note(insts().GetParseNode(prev_def_id), NameDeclPrevious)
       .Emit();
 }
 
@@ -122,10 +122,10 @@ auto Context::NoteIncompleteClass(SemIR::ClassId class_id,
   const auto& class_info = classes().Get(class_id);
   CARBON_CHECK(!class_info.is_defined()) << "Class is not incomplete";
   if (class_info.definition_id.is_valid()) {
-    builder.Note(insts().Get(class_info.definition_id).parse_node(),
+    builder.Note(insts().GetParseNode(class_info.definition_id),
                  ClassIncompleteWithinDefinition);
   } else {
-    builder.Note(insts().Get(class_info.decl_id).parse_node(),
+    builder.Note(insts().GetParseNode(class_info.decl_id),
                  ClassForwardDeclaredHere);
   }
 }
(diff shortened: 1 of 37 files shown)


toolchain/check/context.cpp
```diff
@@ -85,9 +87,8 @@ auto Context::DiagnoseDuplicateName(Parse::NodeId parse_node,
                     "Duplicate name being declared in the same scope.");
   CARBON_DIAGNOSTIC(NameDeclPrevious, Note,
                     "Name is previously declared here.");
-  auto prev_def = insts().Get(prev_def_id);
   emitter_->Build(parse_node, NameDeclDuplicate)
-      .Note(prev_def.parse_node(), NameDeclPrevious)
+      .Note(insts().GetParseNode(prev_def_id), NameDeclPrevious)
```
zygoloid, line 107: Would it make sense to extend the location translator to also accept an `InstId`, so that we don't need to separately grab the parse nodes here? Seems fine for that to be a follow-up.

toolchain/check/eval.cpp
```diff
@@ -174,7 +177,8 @@ auto TryEvalInst(Context& context, SemIR::InstId inst_id, SemIR::Inst inst)
       value->value =
           (value->value == SemIR::BoolValue::False ? SemIR::BoolValue::True
                                                    : SemIR::BoolValue::False);
-      return context.AddConstantInst(*value);
+      return context.AddConstantInst(context.insts().GetParseNode(const_id),
```
jonmeow: Note, this matches, but should it actually have used the inst_id's parse node?

zygoloid: Yes, it probably should. However, I would like to remove the `NodeId` parameter from `AddConstantInst` entirely, as a follow-up after this PR lands. Because constant instructions get reused across different parse nodes (eg, currently, types get deduplicated, and eventually I'd like to deduplicate all constants), I think storing a parse node for them will invite bugs.

toolchain/sem_ir/typed_insts.h
```diff
@@ -207,57 +199,54 @@ struct AnyBranch {
...
   InstBlockId target_id;
   InstId arg_id;
 };
 
 struct Builtin {
-  static constexpr auto Kind = InstKind::Builtin.Define("builtin");
+  static constexpr auto Kind =
+      InstKind::Builtin.Define<Parse::InvalidNodeId>("builtin");
 
   // Builtins don't have a parse node associated with them.
```
zygoloid:
```suggestion
  // Builtins don't have a parse node associated with them.
  static constexpr auto Kind =
      InstKind::Builtin.Define<Parse::InvalidNodeId>("builtin");

```

toolchain/sem_ir/typed_insts.h
```diff
@@ -424,7 +410,7 @@ struct IntLiteral {
 // if it's referenced through name resolution.
 struct LazyImportRef {
   static constexpr auto Kind =
-      InstKind::LazyImportRef.Define("lazy_import_ref");
+      InstKind::LazyImportRef.Define<Parse::InvalidNodeId>("lazy_import_ref");
 
   // No parse node: an instruction's parse tree node must refer to a node in the
```
zygoloid, line 412: This comment should probably be moved before the definition of Kind.

toolchain/sem_ir/inst_kind.h
```diff
@@ -92,8 +91,11 @@ static_assert(sizeof(InstKind) == 1, "Kind objects include padding!");
 // are not copyable, and only one instance of this type is expected to exist per
 // instruction kind, specifically `TypedInst::Kind`. Use `InstKind` instead as a
 // thin wrapper around an instruction kind index.
+template <typename TypedNodeArg>
 class InstKind::Definition : public InstKind {
  public:
+  using TypedNode = TypedNodeArg;
```
zygoloid: Maybe an `Id` suffix would help readers know what this represents:
```suggestion
  using TypedNodeId = TypedNodeArg;
```

toolchain/sem_ir/value_stores.h
```diff
@@ -21,7 +22,11 @@ class InstStore {
   // instruction block. Check::Context::AddInst or InstBlockStack::AddInst
   // should usually be used instead, to add the instruction to the current
   // block.
-  auto AddInNoBlock(Inst inst) -> InstId { return values_.Add(inst); }
+  auto AddInNoBlock(Parse::NodeId parse_node, Inst inst) -> InstId {
+    // TODO: CARBON_CHECK the parse node kind against the instruction kind.
+    parse_nodes_.push_back(parse_node);
+    return values_.Add(inst);
+  }
```
zygoloid: I think the loss of type safety here is something we should address in the short term. I think this can be improved by making `AddInNoBlock` a template:

```suggestion
  template <typename InstT>
  auto AddInNoBlock(typename decltype(InstT::Kind)::TypedNode parse_node, InstT inst) -> InstId {
    parse_nodes_.push_back(parse_node);
    return values_.Add(inst);
  }
```

We might also need an overload for the case where `InstT` is exactly `Inst` that checks the parse node kind at runtime.

```c++
  template <typename InstT>
  auto AddInNoBlock(Parse::NodeId parse_node, InstT inst)
      -> std::enable_if_t<std::is_same_v<InstT, Inst>, InstId> {
    // TODO: CARBON_CHECK the parse node kind against the instruction kind.
    parse_nodes_.push_back(parse_node);
    return values_.Add(inst);
  }
```

We'll presumably want to do the same thing everywhere we're taking a `NodeId` and an `Inst` in order to not regress type safety, such as in `Context::AddInst`, and in `PendingBlock`. Might be worth adding a node-and-inst pair type that carries an inst and a suitable kind of node for it, to factor out the checks?

zygoloid (approved):
Looks good. Ideally I'd like the type safety issue with parse node kinds to be addressed before this lands because I'm worried about backsliding there, but  given the risk of regressions and the size of the change I won't insist :) So don't treat any of these comments as blocking.

toolchain/check/context.cpp
jonmeow: I agree it'd make sense, but it's something I think would be better done separately. This change is already pretty sweeping, and such a change is significant (e.g., requiring a new translator class).

Added an explicit todo in check.cpp next to `Parse::NodeLocationTranslator`.

toolchain/sem_ir/inst_kind.h
jonmeow: Done

toolchain/sem_ir/typed_insts.h
jonmeow: Done.

toolchain/check/eval.cpp
jonmeow: Okay, I'm not going to bother noting a TODO then.

toolchain/sem_ir/value_stores.h
jonmeow: Most of the calls are indirect. Switched to ParseNodeAndInst to handle validation.

jonmeow: Noting the type-checking changes are significant, I figured it's better to go through another review round than just merge. Changing the call semantics in order to get compile-time type checking changes all the call sites, so I didn't think it's worth just merging -- I don't think it materially changes the churn, but at least history will show most of the churn from one PR.

zygoloid (approved):
Awesome, thank you!

PR #3595, opened 2024-01-11
Distinguish between template constants and symbolic constants.

zygoloid:
This is accomplished by tracking an extra bit on the ID we store in the constant values table, and propagating that from subexpressions to the enclosing expression. This extra bit is not yet computed correctly for types; that will be addressed in later PRs.

### DIFF
--- a/toolchain/sem_ir/formatter.cpp
+++ b/toolchain/sem_ir/formatter.cpp
@@ -769,11 +769,11 @@ class Formatter {
     out_ << InstT::Kind.ir_name();
     FormatInstructionRHS(inst);
     if (auto const_id = sem_ir_.constant_values().Get(inst_id);
-        const_id.is_valid()) {
-      out_ << ", const";
-      if (const_id != inst_id) {
+        const_id.is_constant()) {
+      out_ << (const_id.is_symbolic() ? ", sym" : ", const");
+      if (const_id.inst_id() != inst_id) {
         out_ << " = ";
-        FormatInstName(const_id);
+        FormatInstName(const_id.inst_id());
       }
     }
     out_ << "\n";
--- a/toolchain/sem_ir/formatter.cpp
+++ b/toolchain/sem_ir/formatter.cpp
@@ -770,11 +770,12 @@ class Formatter {
     FormatInstructionRHS(inst);
     if (auto const_id = sem_ir_.constant_values().Get(inst_id);
         const_id.is_constant()) {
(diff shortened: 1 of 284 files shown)


toolchain/sem_ir/ids.h
```diff
@@ -68,6 +68,65 @@ constexpr InstId InstId::Invalid = InstId(InstId::InvalidIndex);
...
+  }
+
+  auto Print(llvm::raw_ostream& out) const -> void {
+    if (is_template()) {
+      out << "const " << inst_id();
+    } else if (is_symbolic()) {
+      out << "sym " << inst_id();
+    } else {
+      out << "nonconst";
+    }
```
chandlerc: Curious, what about:

```suggestion
    if (is_template()) {
      out << "template " << inst_id();
    } else if (is_symbolic()) {
      out << "symbolic " << inst_id();
    } else {
      out << "runtime";
    }
```

toolchain/sem_ir/formatter.cpp
```diff
@@ -769,11 +769,11 @@ class Formatter {
     out_ << InstT::Kind.ir_name();
     FormatInstructionRHS(inst);
     if (auto const_id = sem_ir_.constant_values().Get(inst_id);
-        const_id.is_valid()) {
-      out_ << ", const";
-      if (const_id != inst_id) {
+        const_id.is_constant()) {
+      out_ << (const_id.is_symbolic() ? ", sym" : ", const");
```
chandlerc: Similar to the below, what do you think about two formatting changes:

1) Render these differently from arguments, as some kind of annotation? My best guess is `[...]`, but happy for other "annotation"-y formats.

2) Spell out the kinds, and use the actual names?

```suggestion
      out_ << (const_id.is_symbolic() ? " [symbolic" : " [template");
```

toolchain/sem_ir/formatter.cpp
```diff
@@ -769,11 +769,11 @@ class Formatter {
...
+      if (const_id.inst_id() != inst_id) {
         out_ << " = ";
-        FormatInstName(const_id);
+        FormatInstName(const_id.inst_id());
       }
```
chandlerc, line 777: ... and:

```suggestion
        FormatInstName(const_id.inst_id());
      }
      out_ << "]";
```

chandlerc (commented):
Nice, WDYT about a tweak to the names?

chandlerc (approved):
LG

PR #3598, opened 2024-01-12
Ensure we evaluate instructions created in uncommon ways.

zygoloid:
Instructions created by splices during conversion are now evaluated, as are instructions created in cases where we first create a placeholder instruction and later replace it by a different instruction.

This also removes the ability to set a parse node and instruction independently after creating an `InstId`, which could lead to them accidentally not matching.


chandlerc (approved):
Very nice!

PR #3599, opened 2024-01-13
Remove parse nodes from constants.

zygoloid:
These instructions are intended to be shared across all uses and so don't have a meaningful location. So far, only type constants are shared.


chandlerc (approved):
Niiiice!

I really like this. At first glance I didn't fully understand it but seeing the type conversions with the constant attached, it makes tons of sense now.

PR #3600, opened 2024-01-14
Fixed `var` keyword in Carbon code example. Incorrect keyword color.

x7cF:
This PR is changing the color of the var keyword in the `PrintTotalArea` source example which is referenced in the README. The original color was white and changed to `#FF7B72`.


x7cF:
I updated my CLA username

jonmeow:
Thanks for pointing this out!

Looking at the image, I think there are some subtle encoding differences from the tool you used to change the color (e.g., it sets a height/width where the prior tool didn't, and changes values such that the resulting image grows 271 KB -> 364 KB). This manifests in GitHub in the diff view -- note the images don't line up.

The canonical copy of this is [in Docs](https://docs.google.com/drawings/d/1qQe9eYjkRXt1RY5FZXk6ex3mw84S0UItQ6Pn-6WKyJI/edit?resourcekey=0-HXNhCiARVJXGe9q1X57A0A). I've updated the color there. Would you please update your PR with this?

Here's a copy of the updated file:
carbon_snippet (from File -> Download -> Scalable Vector Graphic)

jonmeow:
@chandlerc had asked where this was, so I've pushed a commit with the different image so that we can get the fix in.

x7cF:
So should I still update my commit with the correct image

x7cF:
I must also apologies for my late response. I forgot I even was a programmer for a few days

chandlerc:
> I must also apologies for my late response. I forgot I even was a programmer for a few days

No problem at all, and thanks for bringing this to our attention and helping get it fixed!

chandlerc (approved):
LG

PR #3601, opened 2024-01-15
Add a native Bazel build for libpfm and use it in benchmarks.

chandlerc:
The `libpfm` in the Bazel central repository uses `make` to build it, which is difficult to integrate with our toolchain. Rather than try to fix all the issues there, it's easy to just add a native Bazel build for the library. I don't know that any of the relevant upstream folks are interested in this kind of build, but it seems easy for us to maintain as a Carbon project build configuration. I've also not tried to port all of the different configurations as a consequence, and only 64-bit x86 and Arm as that seems the only likely architectures we'll care about in the near term.

I've kept this using the `bzlmod` stuff as best I can, and I *think* I'm holding all of those pieces correctly, but if not, happy for suggestions on adjustments.

The `google_benchmark` package also has an awkward way of enabling `libpfm` support using a top-level `bazel` command line flag. I think this is because of how brittle the Bazel build of `libpfm` is, but I'm not sure. With the new build, it seems easy to patch `google_benchmark` to detect the same conditions as we build `libpfm` under, and enable it there. So I've done this to avoid folks having to pass a command line flag on platforms where it is supported.

The result is that we now get really nice CPU counter support in our benchmarks out-of-the-box on Linux x86-64 and AArch64. For example on my Fedora Asahi install on a Mac Mini I get:

```console
$ bazel run -c opt --copt=-gmlt //common:hashing_benchmark --run_under="taskset -c 4" -- --benchmark_counters_tabular=true --benchmark_perf_counters=CYCLES,INSTRUCTIONS
INFO: Invocation ID: 4aaeb9e9-7df5-4f1f-b56b-c03411790268
INFO: Analyzed target //common:hashing_benchmark (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //common:hashing_benchmark up-to-date:
  bazel-bin/common/hashing_benchmark
INFO: Elapsed time: 0.360s, Critical Path: 0.02s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Running command line: /bin/bash -c 'taskset -c 4 bazel-bin/common/hashing_benchmark '\''--benchmark_counters_tabular=true'\'' '\''--benchmark_perf_counters=CYCLES,INSTRUCTIONS'\'''
2024-01-15T00:10:50-08:00
Running /home/chandlerc/.cache/bazel/_bazel_chandlerc/b686aa8910e0845b88c21d715819b076/execroot/_main/bazel-out/aarch64-opt/bin/common/hashing_benchmark
Run on (8 X 2064 MHz CPU s)
CPU Caches:
  L1 Data 64 KiB (x8)
  L1 Instruction 128 KiB (x8)
  L2 Unified 4096 KiB (x2)
Load Average: 0.01, 0.08, 0.08
--------------------------------------------------------------------------------------------------------------------------------------------------------------
Benchmark                                                                           Time             CPU   Iterations     CYCLES INSTRUCTIONS bytes_per_second
--------------------------------------------------------------------------------------------------------------------------------------------------------------
BM_LatencyHash<RandValues<uint8_t>, CarbonHashBench>                             4.11 ns         4.11 ns    170200064    13.1321      9.00587      232.116Mi/s
BM_LatencyHash<RandValues<uint8_t>, AbseilHashBench>                             4.82 ns         4.82 ns    145643520    15.3657      12.0059      197.946Mi/s
BM_LatencyHash<RandValues<uint8_t>, LLVMHashBench>                               7.96 ns         7.95 ns     87956480    25.3737      17.0068      119.991Mi/s
BM_LatencyHash<RandValues<uint16_t>, CarbonHashBench>                            4.11 ns         4.11 ns    170365952    13.1247      9.00587      464.573Mi/s
BM_LatencyHash<RandValues<uint16_t>, AbseilHashBench>                            5.51 ns         5.51 ns    127568896    17.5578      14.0059      346.225Mi/s
BM_LatencyHash<RandValues<uint16_t>, LLVMHashBench>                              8.00 ns         7.99 ns     87085056     25.377      17.0068      238.834Mi/s
BM_LatencyHash<RandValues<std::pair<uint8_t, uint8_t>>, CarbonHashBench>         4.91 ns         4.90 ns    136013824    15.6456      14.0059      389.006Mi/s
BM_LatencyHash<RandValues<std::pair<uint8_t, uint8_t>>, AbseilHashBench>         6.85 ns         6.85 ns    102630400    21.8041      18.0059      278.637Mi/s
BM_LatencyHash<RandValues<std::pair<uint8_t, uint8_t>>, LLVMHashBench>           7.57 ns         7.56 ns     92798976    24.1437      20.0068      252.151Mi/s
BM_LatencyHash<RandValues<uint32_t>, CarbonHashBench>                            4.12 ns         4.12 ns    170229760    13.1272      9.00587      926.444Mi/s
BM_LatencyHash<RandValues<uint32_t>, AbseilHashBench>                            4.93 ns         4.92 ns    145304576    15.3738      12.0059      775.224Mi/s
BM_LatencyHash<RandValues<uint32_t>, LLVMHashBench>                              8.11 ns         8.10 ns     87127040     25.373      17.0068       470.98Mi/s
```

### DIFF
--- a/bazel/google_benchmark/0001-Use-libpfm-by-default-on-supported-platforms.patch
+++ b/bazel/google_benchmark/0001-Use-libpfm-by-default-on-supported-platforms.patch
@@ -1,0 +1,62 @@
+From 4110acfa9a131a5a93ae4caef26bd34dd5509155 Mon Sep 17 00:00:00 2001
+From: Chandler Carruth <chandlerc@gmail.com>
+Date: Sat, 13 Jan 2024 02:58:47 -0800
+Subject: [PATCH] Use libpfm by default on supported platforms.
+
+---
+ BUILD.bazel | 26 ++++++++++++++++++--------
+ 1 file changed, 18 insertions(+), 8 deletions(-)
+
+diff --git a/BUILD.bazel b/BUILD.bazel
+index 60d31d2..c313c8a 100644
+--- a/BUILD.bazel
++++ b/BUILD.bazel
+@@ -1,3 +1,5 @@
++load("@bazel_skylib//lib:selects.bzl", "selects")
++
+ licenses(["notice"])
+
+ config_setting(
+@@ -24,12 +26,20 @@ config_setting(
+     visibility = ["//visibility:public"],
+ )
+
+-config_setting(
+-    name = "perfcounters",
+-    define_values = {
+-        "pfm": "1",
+-    },
+-    visibility = [":__subpackages__"],
++selects.config_setting_group(
++    name = "aarch64_or_x86_64",
++    match_any = [
++        "@platforms//cpu:aarch64",
++        "@platforms//cpu:x86_64",
++    ],
++)
++
++selects.config_setting_group(
++    name = "linux_aarch64_or_x86_64",
++    match_all = [
++        "@platforms//os:linux",
++        ":aarch64_or_x86_64",
++    ],
+ )
+
+ cc_library(
+@@ -62,11 +72,11 @@ cc_library(
+     defines = [
+         "BENCHMARK_STATIC_DEFINE",
+     ] + select({
+-        ":perfcounters": ["HAVE_LIBPFM"],
++        ":linux_aarch64_or_x86_64": ["HAVE_LIBPFM"],
+         "//conditions:default": [],
+     }),
+     deps = select({
+-        ":perfcounters": ["@libpfm//:libpfm"],
++        ":linux_aarch64_or_x86_64": ["@libpfm//:libpfm"],
+         "//conditions:default": [],
+     }),
+ )
+--
+2.43.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -233,6 +233,8 @@ exclude: |
   (?x)^(
       MODULE.bazel.lock|
       bazel/bazel_clang_tidy/.*\.patch|
+      bazel/google_benchmark/.*\.patch|
+      bazel/libpfm/.*\.patch|
       bazel/llvm_project/.*\.patch|
       third_party/examples/.*/carbon/.*|
   )$
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -37,11 +37,41 @@ bazel_dep(
     version = "1.14.0.bcr.1",
     repo_name = "com_google_googletest",
 )
+
+google_benchmark_version = "1.8.3"
+
 bazel_dep(
     name = "google_benchmark",
-    version = "1.8.3",
+    version = google_benchmark_version,
     repo_name = "com_github_google_benchmark",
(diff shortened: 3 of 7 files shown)


bazel/google_benchmark/0001-Use-libpfm-by-default-on-supported-platforms.patch
```diff
@@ -0,0 +1,62 @@
...
++selects.config_setting_group(
++    name = "aarch64_or_x86_64",
++    match_any = [
++        "@platforms//cpu:aarch64",
++        "@platforms//cpu:x86_64",
++    ],
++)
++
++selects.config_setting_group(
++    name = "linux_aarch64_or_x86_64",
```
zygoloid, line 39: Would it make sense to keep this named `perfcounters` to make the patch smaller?

zygoloid (approved):
I do wonder whether BCR would take some or all of this as patches. The previous libpfm bazelification was done here: https://github.com/bazelbuild/bazel-central-registry/pull/645 and it doesn't look like the option of moving away from using the `Makefile` was considered there. I couldn't find any BCR policy guidance as to whether reusing an external build system or reimplementing as native Bazel build rules is preferred.

In any case I don't see a reason to block on upstreaming this, even if it would be possible to do so.

chandlerc:
> I do wonder whether BCR would take some or all of this as patches. The previous libpfm bazelification was done here: [bazelbuild/bazel-central-registry#645](https://github.com/bazelbuild/bazel-central-registry/pull/645) and it doesn't look like the option of moving away from using the `Makefile` was considered there. I couldn't find any BCR policy guidance as to whether reusing an external build system or reimplementing as native Bazel build rules is preferred.
> 
> In any case I don't see a reason to block on upstreaming this, even if it would be possible to do so.

I'm a bit hesitant because one aspect that made this (much) simpler is only supporting the CPU names that our Bazel build happens to use on Linux. I'm a little nervous about scope creep there, but TBH, it probably isn't going to be worse than trying to use the `make` based approach.

Worth thinking about to see if we can tidy up more.

bazel/google_benchmark/0001-Use-libpfm-by-default-on-supported-platforms.patch
chandlerc: 🤷

My mild preference was to make the name reflect the reality of how I re-configured them. But if you'd rather, happy to switch it around. I don't feel strongly either way.

zygoloid: I don't feel strongly either, just thinking about keeping the patch minimal. Let's merge as is.

PR #3611, opened 2024-01-17
Deduplicate and canonicalize all constants.

zygoloid:
Rather than producing multiple constants with the same value, fold all instances of a given constant to the same constant instruction.

A future PR will use this to replace the current type canonicalization system.


toolchain/sem_ir/value_stores.h
```diff
@@ -146,17 +146,35 @@ class ConstantValueStore : public Yaml::Printable<ConstantValueStore> {
...
+      : allocator_(&allocator), constants_(sem_ir) {}
 
-  auto array_ref() const -> llvm::ArrayRef<InstId> { return values_; }
-  auto size() const -> int { return values_.size(); }
+  // Adds a new constant instruction, or gets the existing constant with this
+  // value. Returns the instruction ID and a bool indicating whether a new
+  // instruction was added.
+  auto GetOrAdd(Inst inst) -> std::pair<SemIR::InstId, bool>;
+
+  auto vector() const -> std::vector<InstId>;
```
jonmeow: For naming, had you considered something like `GetAsVector` or `CopyToVector`? I think the accessor-like name obscures how expensive the copy occurring here is.

toolchain/sem_ir/formatter.cpp
```diff
@@ -592,7 +592,7 @@ class Formatter {
     llvm::SaveAndRestore constants_scope(scope_,
                                          InstNamer::ScopeIndex::Constants);
     out_ << "constants {\n";
-    FormatCodeBlock(sem_ir_.constants().array_ref());
+    FormatCodeBlock(sem_ir_.constants().vector());
```
jonmeow: Is this going to be creating two copies of the constants, once on line 53 and once here? Can the copy be shared?

toolchain/sem_ir/value_stores.h
jonmeow: Why `std::vector` instead of `llvm::SmallVector`?

toolchain/sem_ir/value_stores.h
```diff
@@ -146,17 +146,35 @@ class ConstantValueStore : public Yaml::Printable<ConstantValueStore> {
...
 };
 
-// Provides storage for instructions representing global constants.
+// Provides storage for instructions representing deduplicated global constants.
 class ConstantStore {
  public:
-  // Add a constant instruction.
-  auto Add(InstId inst_id) -> void { values_.push_back(inst_id); }
+  explicit ConstantStore(SemIR::File* sem_ir, llvm::BumpPtrAllocator& allocator)
+      : allocator_(&allocator), constants_(sem_ir) {}
```
jonmeow:
```suggestion
  explicit ConstantStore(SemIR::File& sem_ir, llvm::BumpPtrAllocator& allocator)
      : allocator_(&allocator), constants_(&sem_ir) {}
```

It seems odd to me that you take in one as pointer, and the other as reference. Maybe both as reference?

toolchain/check/context.cpp
```diff
@@ -100,14 +100,13 @@ auto Context::AddPlaceholderInst(SemIR::ParseNodeAndInst parse_node_and_inst)
...
-  auto inst_id = sem_ir().insts().AddInNoBlock(
-      SemIR::ParseNodeAndInst::Untyped(Parse::NodeId::Invalid, inst));
-  constants().Add(inst_id);
-
+  auto [inst_id, added] = constants().GetOrAdd(inst);
   auto const_id = is_symbolic ? SemIR::ConstantId::ForSymbolicConstant(inst_id)
                               : SemIR::ConstantId::ForTemplateConstant(inst_id);
-  constant_values().Set(inst_id, const_id);
+  if (added) {
+    // TODO: Should `ConstantStore` do this for us?
```
jonmeow, line 107: +1 for this. Had you avoided it because of the extra constant_values propagation, or is there a different concern?

toolchain/sem_ir/inst_profile.h
```diff
@@ -0,0 +1,21 @@
...
+
+#include "llvm/ADT/FoldingSet.h"
+#include "toolchain/sem_ir/file.h"
+#include "toolchain/sem_ir/inst.h"
+
+namespace Carbon::SemIR {
+
+// Computes a profile of the given constant instruction, for canonicalization /
+// deduplication.
+auto ProfileConstant(llvm::FoldingSetNodeID& id, File* sem_ir, Inst inst)
```
jonmeow: Can the `sem_ir` be `const` here? If not, should the comments explain the `sem_ir` modifications?

jonmeow: In a lot of functions, you have `&id` and `*sem_ir`: why a reference for one but a pointer for the other? Is it because the `sem_ir` is optional?

Note, [style](https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs) says:

> Non-optional input parameters should usually be values or const references, while non-optional output and input/output parameters should usually be references (which cannot be null).

toolchain/sem_ir/inst_profile.cpp
```diff
@@ -0,0 +1,123 @@
...
+  real.exponent.Profile(id);
+  id.AddBoolean(real.is_decimal);
+}
+
+// Selects the function to use to profile argument N of instruction InstT. This
+// default implementation is used when there is no such argument, and adds
+// nothing to the profile.
+template <typename InstT, int N,
+          bool InRange = (N < InstLikeTypeInfo<InstT>::NumArgs)>
+static constexpr ProfileArgFunction* SelectProfileArgFunction =
```
jonmeow: Had you considered interleaving these with the XYZFunction definitions that correspond? In my mind, that may emphasize the way the function definitions are used by these (that they're not called directly).

Also, are these functions separated out in order to reduce generated code on overlapping template uses when the table use produced? It seems like the `constexpr` function values could've been more direct. Should that be noted in a comment?

toolchain/sem_ir/inst_profile.cpp
```diff
@@ -0,0 +1,123 @@
...
+                        int32_t arg1) -> void {
+  ProfileArg0(id, context, arg0);
+  ProfileArg1(id, context, arg1);
+}
+
+static constexpr auto KindIndex(InstKind::RawEnumType kind) -> int {
+  return static_cast<InstKind::UnderlyingType>(kind);
+}
+
+auto ProfileConstant(llvm::FoldingSetNodeID& id, File* sem_ir, Inst inst)
```
jonmeow: Should the `context` names be made consistent with the `ProfileConstant` `sem_ir` name? We typically use `context` for `Context`, although I could also see changing `sem_ir` to `context`.

toolchain/sem_ir/inst_profile.cpp
```diff
@@ -0,0 +1,123 @@
...
+template <ProfileArgFunction* ProfileArg0, ProfileArgFunction* ProfileArg1>
+static auto ProfileArgs(llvm::FoldingSetNodeID& id, File* context, int32_t arg0,
+                        int32_t arg1) -> void {
+  ProfileArg0(id, context, arg0);
+  ProfileArg1(id, context, arg1);
+}
+
+static constexpr auto KindIndex(InstKind::RawEnumType kind) -> int {
+  return static_cast<InstKind::UnderlyingType>(kind);
+}
```
jonmeow: Had you considered exposing `EnumBase::AsInt`? i.e., `KindIndex(inst.kind())` -> `inst.kind().AsInt()`

toolchain/sem_ir/value_stores.h
```diff
@@ -146,17 +146,35 @@ class ConstantValueStore : public Yaml::Printable<ConstantValueStore> {
...
+  // TODO: We store two copies of each constant instruction: one in insts() and
+  // one here. We could avoid one of those copies and store just an InstId here,
+  // at the cost of some more indirection when recomputing profiles during
+  // lookup. Once we have a representative data set, we should measure the
+  // impact on compile time from that change.
+  struct ConstantNode : llvm::FoldingSetNode {
+    Inst inst;
+    InstId inst_id;
+
+    auto Profile(llvm::FoldingSetNodeID& id, SemIR::File* sem_ir) -> void;
```
jonmeow: Style: functions before members.
```suggestion
    auto Profile(llvm::FoldingSetNodeID& id, SemIR::File* sem_ir) -> void;

        Inst inst;
    InstId inst_id;
```

toolchain/sem_ir/value_stores.h
jonmeow: Should this have a comment mentioning the sort semantic?

toolchain/sem_ir/value_stores.h
```diff
@@ -146,17 +146,35 @@ class ConstantValueStore : public Yaml::Printable<ConstantValueStore> {
...
+  };
+
+  llvm::BumpPtrAllocator* allocator_;
```
jonmeow, line 179: In typed_nodes_, I'd done `llvm::SmallVector<std::unique_ptr<TypeNode>> type_node_storage_;` -- should we move that to this same allocator approach?

(not in this PR, but more about if I'm understanding the intent)

jonmeow (approved):
Essentially LGTM, albeit nitpicks.

toolchain/sem_ir/formatter.cpp
zygoloid: Yes, this makes two copies. I don't think it's worth putting effort into sharing them, given that the performance of the textual SemIR formatter isn't really a priority.

toolchain/sem_ir/inst_profile.cpp
zygoloid: Yes, these are separated out to reduce the code size. I've added a comment saying so. I also switched to a `constexpr` function with `if constexpr` to do the selection; I think that's a lot easier to understand than these template specializations :)

toolchain/sem_ir/inst_profile.cpp
zygoloid: Done. I'd been meaning to rework this but forgot, thanks!

toolchain/sem_ir/inst_profile.cpp
zygoloid: Switched to `sem_ir`. (Just an oversight / incomplete refactoring: this used to be in `value_stores.cpp` where it was trying to be consistent with the `context` parameter from `llvm::ContextualFoldingSet`, but I don't think it makes sense to use that naming here. I hadn't intended to change from the `sem_ir` naming convention.)

toolchain/sem_ir/inst_profile.h
zygoloid: Sorry, too much thinking in Carbon :) (I was passing `FoldingSetNodeID` by reference because that's idiomatic and forced upon us in various places by the LLVM API, but otherwise defaulting to using pointers because that's how we'd do it in Carbon.)

Switched to references for parameters and pointers for class members.

toolchain/sem_ir/value_stores.h
zygoloid: Fixed.

toolchain/sem_ir/value_stores.h

zygoloid: Largely because this isn't supposed to be used anywhere performance- or size-sensitive, but I suppose we're generally using `SmallVector<T, 0>` for that these days. Switched.

toolchain/sem_ir/value_stores.h
zygoloid: I'd prefer to keep this as-is. Style [allows either order](https://google.github.io/styleguide/cppguide.html#Declaration_Order) for structs, and data-first is more consistent with classes -- at least for aggregate structs (those without constructors) -- because it puts construction information at the start.

toolchain/sem_ir/value_stores.h
zygoloid: In principle yes, but I'll be removing that storage entirely in an upcoming PR -- once we canonicalize all constants, we won't need separate type canonicalization any more.

toolchain/sem_ir/inst_profile.h
zygoloid: It can. Done.

toolchain/check/context.cpp
zygoloid: Mostly I think I was thinking of consistency: adding to `insts()` doesn't change `constant_values()`, so adding to `constants()` seems like maybe it shouldn't do so either. And symmetrically, `Check::Context::AddInst` is responsible for updating `constant_values` in most cases, so having `SemIR` do it in some cases seems inconsistent.

I'm not sure what the right balance here is. The `ConstantStore` is a higher-level layer that is already modifying `insts()` (and reading from `inst_blocks()`), so it doesn't seem all that unreasonable for it to modify `constant_values()` too, but at the same time that seems like something that `Check` ought to be doing in the current setup.

I'm happy for you to tell me what you'd prefer here. If `ConstantStore` does start updating `constant_values()` too, should we even have a `Context::AddConstant` function at all, given that it would be a simple one-liner?

toolchain/sem_ir/value_stores.h
jonmeow: I would even say `SmallVector<T>` is appropriate if you're saying performance doesn't matter, since it's easier to write and explain.

toolchain/sem_ir/inst_profile.cpp
jonmeow: LGTM. Yeah, that was why I was thinking `context` might be okay.

toolchain/sem_ir/value_stores.h
jonmeow: It would really be nice to be internally consistent within Carbon...

toolchain/sem_ir/value_stores.h
jonmeow: SGTM, thanks.

toolchain/check/context.cpp
jonmeow: > If `ConstantStore` does start updating `constant_values()` too, should we even have a `Context::AddConstant` function at all, given that it would be a simple one-liner?

The TODO had seemed reasonable to me in terms of not needing to return the "success" info. And actually, removing `AddConstant` would be the best argument for moving it to `constants`: it reduces the API surface of `Context`.

I suppose it wouldn't bother me much if `insts()` starting gaining more connectivity if that helped callers, too.

toolchain/sem_ir/inst_profile.cpp
```diff
@@ -11,113 +11,108 @@
...
+#define CARBON_SEM_IR_INST_KIND(KindName)              \
+  ProfileArgs<SelectProfileArgFunction<KindName, 0>(), \
+              SelectProfileArgFunction<KindName, 1>()>,
 #include "toolchain/sem_ir/inst_kind.def"
   };
 
   inst.kind().Profile(id);
-  ProfileFunctions[KindIndex(inst.kind())](id, sem_ir, inst.arg0(),
-                                           inst.arg1());
+  ProfileFunctions[inst.kind().AsInt()](id, sem_ir, inst.arg0(), inst.arg1());
```
jonmeow, line 115: The refactoring made me think to suggest a different approach:

```
static constexpr std::pair<ProfileArgFunction*, ProfileArgFunction*>
    ProfileArgFunctions[] = {
#define CARBON_SEM_IR_INST_KIND(KindName)   \
  {SelectProfileArgFunction<KindName, 0>(), \
   SelectProfileArgFunction<KindName, 1>()},
};

inst.kind().Profile(id);
auto functions = ProfileArgFunctions[inst.kind().AsInt()];
functions.first(id, sem_ir, inst.arg0());
functions.second(id, sem_ir, inst.arg1());
```

I think this eliminates the `ProfileArgs` function, resulting in slightly more compact code to execute.

Totally optional -- not sure what you prefer for readability, or if I'm really right about codegen.

jonmeow (approved):
LGTM -- feel free to merge. Even if there are still bits you choose to change, such as the AddConstant TODO, I think it's up to you whether it's easiest to address here or in a separate PR.

toolchain/sem_ir/value_stores.h
zygoloid: I did some checking, and we're currently not especially consistent. For aggregates with both fields and methods, we have an approximately 2:1 ratio of methods first versus fields first (though it's closer to being 1:1 if you count a cluster of types using the same convention as one):

- Fields first: `Check::ConversionTarget`, `StringLiteral::Introducer`, `CharSet` (string_literal.cpp), `RandomSourceOptions` (tokenized_buffer_benchmark.cpp), `Step` (in `File::StringifyTypeExpr`), `SemIR::InitRepr`, `InstNamer::Namespace`, `ConstantStore::ConstantNode`

- Methods first: `Check::DeclNameStack::NameContext`, `Driver::{CompileOptions,Options}`, `NibbleLUT` (lex.cpp), `ExpectedTokens` (tokenized_buffer_test_helpers.h), `TokenizedBuffer::PrintWidths`, `Parse::Context::StateStackEntry`, `SemIR::{BindNameInfo,Function,Class,Interface}`, `SemIR::ValueRepr`, `SemIR::TypeInfo`, `SemIR::NameScope`, `Printer` (yaml_test_helpers.cpp)

(Granted, I added most of the types in the first category and you added most of the types in the second!) Personally I think the more important consistency is putting the most important information about a type first, rather than putting either fields or methods first, so I would prefer that we allow the author of the code to choose the option that best communicates their intent rather than enforcing one option or the other.

toolchain/sem_ir/inst_profile.cpp
zygoloid: FWIW, I started with that, but switched to this approach so that we make only one indirect branch instead of two. Honestly I'm not sure which is better; I can see arguments in both directions, but I *think* the single indirect branch approach will likely be more efficient.

PR #3613, opened 2024-01-18
Fix reference to package_scope.

jonmeow:
The reference may be invalidated by an `Import` call reallocating the `name_scopes`.


chandlerc (approved):
LG

PR #3617, opened 2024-01-18
Use constant evaluation to determine the identity of types.

zygoloid:
Remove the type canonicalization mechanism and instead rely on constant canonicalization to deduplicate types.

Rename the `Canonicalize*Type` functions to reflect that they're no longer performing canonicalization. Switch code that creates types due to semantic checking, rather than due to source syntax, to directly create type constants through evaluation rather than creating an instruction and evaluating it to produce a separate constant representation.

The mapping from `const (const T)` that was previously performed by type canonicalization is now implemented in expression evaluation instead.

The value `<error>` is now treated as a constant value, with a special property that an instruction involving `<error>` that could possibly be constant evaluates to `<error>`. This helps avoid producing follow-on errors when an error occurs as a subexpression of an expression, such as a type, that is intended to be constant.


toolchain/sem_ir/file.cpp
```diff
@@ -105,9 +110,11 @@ File::File(SharedValueStores& value_stores, std::string filename,
   static constexpr auto BuiltinIR = CrossRefIRId(0);
   for (auto [i, inst] : llvm::enumerate(builtins->insts_.array_ref())) {
     // We can reuse builtin type IDs because they're special-cased values.
```
jonmeow: This line feels like it should be more closely associated with the `inst.type_id()`; it sort of was before, but you've added `builtin_id` on its own line now. Maybe the comment is a little unclear about what it means though?

toolchain/check/testdata/basics/multifile_raw_and_textual_ir.carbon
```diff
@@ -37,6 +37,15 @@ fn B() {}
 // CHECK:STDOUT:     inst+1:          {kind: FunctionDecl, arg0: function0, type: type1}
 // CHECK:STDOUT:     inst+2:          {kind: Return}
 // CHECK:STDOUT:   constant_values:
+// CHECK:STDOUT:     instTypeType:    template instTypeType
```
jonmeow: Should the special-casing that `insts` has to exclude builtins be added here?

toolchain/check/context.h
```diff
@@ -578,16 +553,15 @@ class Context {
...
-  llvm::SmallVector<std::unique_ptr<TypeNode>> type_node_storage_;
+  // Cache of reverse mapping from type constants to types.
+  //
+  // TODO: Instead of mapping to a dense `TypeId` space, we could make `TypeId`
+  // be a thin wrapper around `ConstantId` and only perform the lookup only when
+  // we want to access the completeness and value representation of a type. It's
+  // not clear whether that would result in more or fewer lookups.
+  //
+  // TODO: Should this be part of the `TypeStore`?
+  llvm::DenseMap<SemIR::ConstantId, SemIR::TypeId> type_ids_for_type_constants_;
```
jonmeow, line 564: Should this be a vector sized to constants in order to avoid hash lookups?

toolchain/check/eval.cpp
```diff
@@ -4,31 +4,82 @@
...
+static auto MakeConstantResult(Context& context, SemIR::Inst inst, Phase phase)
+    -> SemIR::ConstantId {
+  switch (phase) {
+    case Phase::Template:
+    case Phase::Symbolic:
+      return context.AddConstant(inst, phase == Phase::Symbolic);
+    case Phase::Error:
+      return SemIR::ConstantId::Error;
+    case Phase::Runtime:
+      return SemIR::ConstantId::NotConstant;
```
jonmeow, line 64: Maybe rename Runtime to NotConstant to match, and to disambiguate slightly from runtime constants?

toolchain/check/eval.cpp
```diff
@@ -4,31 +4,82 @@
...
+// ordered so that the phase of an expression is the numerically highest phase
+// of its constituent evaluations.
+enum class Phase : uint8_t {
+  // Value could be entirely and concretely computed.
+  Template,
+  // Evaluation encountered a reference to a symbolic binding.
+  Symbolic,
+  // Evaluation encountered an error, which is treated as being potentially
+  // constant, but with an unknown phase.
+  Error,
```
jonmeow: Why is Error an earlier phase than Runtime, or after Symbolic even though it carries from Template? Is this worth some comments to explain the intended interaction with LatestPhase (which I assume is why)?

toolchain/check/eval.cpp
```diff
@@ -184,19 +235,36 @@ auto TryEvalInst(Context& context, SemIR::InstId inst_id, SemIR::Inst inst)
...
+        return MakeConstantResult(context, value, Phase::Template);
+      }
+      if (phase == Phase::Error) {
+        return SemIR::ConstantId::Error;
+      }
+      break;
+    }
+
+    // `const (const T)` evaluates to `const T`. Otherwise, `const T` evaluates
+    // to itself.
```
jonmeow, line 264: This use of `const` feels like it may be confusing if we use `const` for runtime constants.

toolchain/check/eval.cpp
```diff
@@ -4,31 +4,82 @@
...
+  return Phase(std::max(static_cast<uint8_t>(a), static_cast<uint8_t>(b)));
+}
+
+// Forms a `constant_id` describing a given evaluation result.
+static auto MakeConstantResult(Context& context, SemIR::Inst inst, Phase phase)
+    -> SemIR::ConstantId {
+  switch (phase) {
+    case Phase::Template:
+    case Phase::Symbolic:
+      return context.AddConstant(inst, phase == Phase::Symbolic);
```
jonmeow: This comparison feels a little odd with the switch, but maybe that's just me.

jonmeow (approved):

toolchain/check/testdata/basics/multifile_raw_and_textual_ir.carbon
zygoloid: Done.

toolchain/check/context.h
zygoloid: We don't have a dense array of constants; a `ConstantId` is just an `InstId` plus a flag saying whether it's symbolic or template. We could make this a vector sized to `insts()`, but it'd be quite sparse.

toolchain/sem_ir/file.cpp
```diff
@@ -159,28 +159,38 @@ auto File::OutputYaml(bool include_builtins) const -> Yaml::OutputMapping {
...
+                    for (int i : llvm::seq(start, insts_.size())) {
+                      auto id = InstId(i);
+                      map.Add(PrintToString(id),
+                              Yaml::OutputScalar(insts_.Get(id)));
+                    }
+                  }));
+          map.Add("constant_values",
+                  Yaml::OutputMapping([&](Yaml::OutputMapping::Map map) {
+                    int start = include_builtins ? 0 : BuiltinKind::ValidCount;
+                    for (int i : llvm::seq(start, insts_.size())) {
```
jonmeow, line 186:
```suggestion
                    for (int i : llvm::seq(start, constant_values_.size())) {
```

toolchain/check/eval.cpp
zygoloid: The `Phase` constants are intended to correspond to the [expression phases in the design](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/README.md#expression-phases).

toolchain/check/eval.cpp
zygoloid: Resolving based on in-person discussion.

toolchain/check/eval.cpp
zygoloid: Expanded the comments to try to make this clearer.

toolchain/check/eval.cpp
zygoloid: I'm indifferent. Switched :-)

toolchain/sem_ir/file.cpp
zygoloid: Done.

toolchain/sem_ir/file.cpp
zygoloid: This was what I intended as-is -- `constant_values_` doesn't have a `size()`, and is able to return a constant value for any `InstId`. (It's notionally a map from `InstId` to `ConstantId`.) Would a comment help?

PR #3619, opened 2024-01-19
Improve diagnosing of name conflicts in imports.

jonmeow:
Name conflicts weren't previously tested, and the diagnostics were just a TODO, so this is also adding testing for that. But handling too.

Removing AddEntry because I think it's hard to make helpful for this use-case when we want to do a diagnostic followup (because really, callers want the full `.insert` result of pointer + success), and unused otherwise.

### DIFF
--- a/toolchain/check/import.cpp
+++ b/toolchain/check/import.cpp
@@ -105,7 +105,7 @@ static auto CopySingleNameScopeFromImportIR(
   context.ReplaceInstBeforeConstantUse(
       namespace_id, {Parse::NodeId::Invalid, namespace_inst});
 
-  // If there's a name conflict, produce the diagnose, but still produce the
+  // Diagnose if there's a name conflict, but still produce the
   // namespace to supersede the name conflict in order to avoid repeat
   // diagnostics.
   if (!success) {
--- a/toolchain/check/import.cpp
+++ b/toolchain/check/import.cpp
@@ -62,21 +62,6 @@ static auto CopyNameFromImportIR(Context& context,
   return import_name_id;
 }
 
-// Creates a namespace. The type ID is builtin, and reused to avoid duplicative
-// canonicalization.
-static auto AddNamespace(Context& context,
-                         SemIR::NameScopeId enclosing_scope_id,
-                         SemIR::NameId name_id, SemIR::TypeId namespace_type_id)
-    -> std::pair<SemIR::InstId, SemIR::NameScopeId> {
-  // Use the invalid node because there's no node to associate with.
-  auto inst =
-      SemIR::Namespace{namespace_type_id, name_id, SemIR::NameScopeId::Invalid};
-  auto id = context.AddPlaceholderInst({Parse::NodeId::Invalid, inst});
-  inst.name_scope_id = context.name_scopes().Add(id, enclosing_scope_id);
-  context.ReplaceInstBeforeConstantUse(id, {Parse::NodeId::Invalid, inst});
-  return {id, inst.name_scope_id};
-}
-
 static auto CacheCopiedNamespace(
     llvm::DenseMap<SemIR::NameScopeId, SemIR::NameScopeId>& copied_namespaces,
     SemIR::NameScopeId import_scope_id, SemIR::NameScopeId to_scope_id)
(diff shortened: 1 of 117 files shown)


toolchain/check/import.cpp
```diff
@@ -87,10 +72,56 @@ static auto CacheCopiedNamespace(
...
+      // Namespaces are open, so we can append to the existing one even if it
+      // comes from a different file.
+      CacheCopiedNamespace(copied_namespaces, import_scope_id,
+                           namespace_inst->name_scope_id);
+      return namespace_inst->name_scope_id;
+    }
+  }
+
+  // Produce the namespace for the entry. Use the invalid node because there's
+  // no node to associate with.
```
chandlerc, line 98: Fine to use the invalid node in this iteration, but you could also use the `import` parse node maybe?

toolchain/check/import.cpp
```diff
@@ -87,10 +72,56 @@ static auto CacheCopiedNamespace(
...
+  auto namespace_id =
+      context.AddPlaceholderInst({Parse::NodeId::Invalid, namespace_inst});
+  namespace_inst.name_scope_id =
+      context.name_scopes().Add(namespace_id, enclosing_scope_id);
+  context.ReplaceInstBeforeConstantUse(
+      namespace_id, {Parse::NodeId::Invalid, namespace_inst});
+
+  // If there's a name conflict, produce the diagnose, but still produce the
+  // namespace to supersede the name conflict in order to avoid repeat
+  // diagnostics.
```
chandlerc: Slight wording tweak maybe:

```suggestion
  // Diagnose if there's a name conflict, but still produce the
  // namespace to supersede the name conflict in order to avoid repeat
  // diagnostics.
```

toolchain/check/testdata/packages/fail_conflict_no_namespaces.carbon
```diff
@@ -0,0 +1,52 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+// CHECK:STDERR: conflict.carbon: ERROR: Duplicate name being declared in the same scope.
+// CHECK:STDERR: conflict.carbon: Name is previously declared here.
+
+// --- namespace.carbon
```
chandlerc: Maybe:
```suggestion
// --- fn.carbon
```

toolchain/check/testdata/packages/fail_conflict_no_namespaces.carbon
```diff
@@ -0,0 +1,52 @@
...
+var Foo: i32;
+
+// --- conflict.carbon
+
+package Example library "conflict" api;
+
+import library "fn";
+import library "var";
+
+// CHECK:STDOUT: --- namespace.carbon
```
chandlerc:
```suggestion
// CHECK:STDOUT: --- fn.carbon
```

toolchain/check/testdata/packages/fail_conflict_no_namespaces.carbon
chandlerc, line 1: Rather than "no namespaces" maybe "fail_conflict_function_variable" or something? Maybe we'll have other cases outside of namespaces as well in the future...

chandlerc (approved):
Generally LG. Some minor suggestions on a test case, LG to merge with those.

Also a forward-looking idea that's optional and/or fine for a future PR if it makes sense.

toolchain/check/import.cpp
jonmeow: I think the import would be something belonging in a note; it's not the name conflict and would be misleading.

toolchain/check/testdata/packages/fail_conflict_no_namespaces.carbon
jonmeow: I'm concerned that getting more specific creates an incorrect implication that there are more code paths that are untested.

That is, consider this as a grid of what's tested:

|  | namespace | not namespace
| --- | --- | ---
| namespace | ✅  | ✅ 
| not namespace | ✅  | TODO: merge  

Versus the implication when you start getting more specific:

| | namespace | function | variable | class | interface
| --- | --- | --- | --- | --- | ---
| namespace | ✅  | ✅  | ❌  | ❌  | ❌ 
| function | ✅  | TODO: merge  | ✅  | ❌  | ❌ 
| variable | ❌  | ❌  | TODO: merge  | ❌  | ❌ 
| class | ❌  | ❌  | ❌  | TODO: merge  | ❌ 
| interface | ❌  | ❌  | ❌  | ❌  | TODO: merge

toolchain/check/import.cpp
jonmeow: And to be clear, adding it I think would create technical debt towards a correct implementation. [It's easier to find and address invalid than to find and address a valid-but-incorrect argument]

PR #3624, opened 2024-01-19
Constant evaluation support for initializing expressions.

zygoloid:
The constant value we associate with an initializing representation is the object representation that the initializing expression will store to its destination.

Also include the type in the profile of an instruction. This is now necessary for array values, which are represented as tuple_value instructions with array type, to avoid instructions with different types being merged by constant canonicalization.


chandlerc (approved):
Nice!

PR #3625, opened 2024-01-20
Constant evaluation for aggregate access.

zygoloid:
Also, form `addr_of error` instead of `addr_of operand` when `operand` is not a reference expression, so that we don't try to constant-evaluate a meaningless expression.

Also, expect a constant for an array bound rather than specifically an integer literal. This allows constant evaluation results to be more easily tested by inspecting array bounds.

### DIFF
--- a/toolchain/check/handle_array.cpp
+++ b/toolchain/check/handle_array.cpp
@@ -32,21 +32,25 @@ auto HandleArrayExpr(Context& context, Parse::ArrayExprId parse_node) -> bool {
   context.node_stack()
       .PopAndDiscardSoloParseNode<Parse::NodeKind::ArrayExprSemi>();
   auto element_type_inst_id = context.node_stack().PopExpr();
-  auto bound_inst = context.insts().Get(bound_inst_id);
-  if (auto literal = bound_inst.TryAs<SemIR::IntLiteral>()) {
-    const auto& bound_value = context.ints().Get(literal->int_id);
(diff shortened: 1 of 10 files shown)


toolchain/check/handle_array.cpp
```diff
@@ -32,21 +32,25 @@ auto HandleArrayExpr(Context& context, Parse::ArrayExprId parse_node) -> bool {
...
-          {parse_node, SemIR::ArrayType{SemIR::TypeId::TypeType, bound_inst_id,
-                                        ExprAsType(context, parse_node,
-                                                   element_type_inst_id)}});
-      return true;
-    }
+
+  // The array bound must be a constant.
+  //
+  // TODO: Should we support runtime-phase bounds in cases such as:
+  //   comptime fn F(n: i32) -> type { return [i32; n]; }
```
chandlerc, line 39: Not blocking here, but FWIW: I think no? Because it's only valid when the function guaranteed-compile-time, it should be trivial to write it as `template n:! i32`?

If we want something like `if consteval` that might motivate this, but I'm somewhat dubious...

chandlerc (approved):
> Also, expect a constant for an array bound rather than specifically an integer literal. This allows constant evaluation results to be more easily tested by inspecting array bounds.

Expecting a constant of course makes sense here, but I'd love to have a more direct way of testing...

Does `let template i:! i32 = expr;` not work? (Anyways, not an issue for merging this PR.)

PR #3631, opened 2024-01-20
Fix syntax highlighting in VSCode (#3630).

### LINKED ISSUE #3630: Invalid syntax highlighting in VSCode
There is an issue in VSCode syntax highlighting.
Highlighting works as expected in Compiler Explorer demo:
image

The same code in VSCode:
image

Seems that everything gets colored as string literal after the first one.

Tested with Kubuntu 22.04.

czapiga:
Add missing escape backslash to `string_escapes` matching string.

Closes #3630

VSCode highlighting working after change:
image


chandlerc (approved):
LG, thanks for the fix, merging!

PR #3632, opened 2024-01-22
Split github_tools into its own bazel repo.

jonmeow:
The pip dependencies in github_tools are the reason the MODULE.bazel.lock is platform-dependent. Following complaints about the platform-dependence, split apart github_tools from the rest of the bazel repo and make it not track the lockfile: while there's an incremental safety risk due to not tracking checksums, it's unlikely the tools there would ever be part of a Carbon release process. If we eventually add Python tools that need pip to the release, it might be desirable to go back to re-unify the bazezl repos.

This does make running pr_comments incrementally more inconvenient because a "bazel run" needs to be run from the github_tools subdir.

As a consequence of separating the dependency, this means tests will not be continuously run in github_tools. They're now a separate repo, and we cannot add a dependency without restoring the platform-dependent issue. I think pr_comments is sufficiently low value and unchanging that it is not worth building separate CI for it.

Cleans up some legacy references to third_party/llvm-project, since now github_tools needs to be added to the main bazelignore.


chandlerc (approved):
Yeah this feels like a reasonable way to split the difference. And I agree that CI doesn't seem terribly important here.

LG

PR #3635, opened 2024-01-22
Fix block string literals highlighting

### LINKED ISSUE #3634: Missing block string literals highlighting
According to docs, current implementation and tests:
https://github.com/carbon-language/carbon-lang/blob/3822c1b526d3568a60127595bba632471f92faac/docs/design/README.md?plain=1#L644-L646

Highlighting in VSCode is not working as expected:
image

Block string literals are getting colored when invalid quotes (`"`) are used, but still not good enough (escaped triple quote):
image

czapiga:
Replace block string literals quotes.

Highlighting after change:
image

Closes #3634

### DIFF
--- a/utils/textmate/Syntaxes/Carbon.plist
+++ b/utils/textmate/Syntaxes/Carbon.plist
@@ -82,9 +82,9 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
                 <key>name</key>
                 <string>string.quoted.triple.carbon</string>
                 <key>begin</key>
-                <string>"""</string>
+                <string>'''</string>
                 <key>end</key>
-                <string>"""</string>
+                <string>'''</string>
 
                 <key>patterns</key>
                 <array>
@@ -143,7 +143,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
                 <key>name</key>
                 <string>constant.character.escape.carbon</string>
                 <key>match</key>
-                <string>\\([tnr'"0\0]|x[0-9A-F]{2}|u\{[0-9A-F]{4,}\})</string>
+                <string>\\('''|[tnr'"0\0]|x[0-9A-F]{2}|u\{[0-9A-F]{4,}\})</string>
             </dict>
         </dict>
     </dict>
--- a/utils/textmate/Syntaxes/Carbon.plist
+++ b/utils/textmate/Syntaxes/Carbon.plist
@@ -143,7 +143,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
                 <key>name</key>
                 <string>constant.character.escape.carbon</string>
                 <key>match</key>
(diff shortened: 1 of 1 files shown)


utils/textmate/Syntaxes/Carbon.plist
```diff
@@ -143,7 +143,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
                 <key>name</key>
                 <string>constant.character.escape.carbon</string>
                 <key>match</key>
-                <string>\\([tnr'"0\0]|x[0-9A-F]{2}|u\{[0-9A-F]{4,}\})</string>
+                <string>\\('''|[tnr'"0\0]|x[0-9A-F]{2}|u\{[0-9A-F]{4,}\})</string>
```
chandlerc: I would have expected this is fine to just have with the single escaped `'`? The escape sequence is just `\'`, and it can then be followed by two `'`s without forming the close `'''`.... At least, that's my reading of this section:
https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/lexical_conventions/string_literals.md#:~:text=whose%20first%20%27%20is%20not%20part%20of%20a%20%5C%27%20escape%20sequence

Or does that result in VS Code still thinking the block literal is closed?

chandlerc (commented):
Awesome to see all the improvements here! A question on this one inline...

utils/textmate/Syntaxes/Carbon.plist
czapiga: Single escaped `'` results in only the first `'` of `'''` getting colored (look at #3634, it's visible in last image). Also single `'` is already covered by the second capturing group of this regex.

chandlerc: > Single escaped `'` results in only the first `'` of `'''` getting colored (look at #3634, it's visible in last image).

I think that's correct though? There is one escaped character (the first `'`), and two normal characters in the body of the block string literal.

Here is an example to illustrate what I'm talking about maybe:

```carbon
let example: auto = '''
    One and two quotes are fine w/o escaping: ' ''
    Can also be escaped: \' \'\'
    First quote escaped makes the rest normal, and not the end:
      \'''
    Could also escape all three:
      \'\'\'
    But without any escapes, we end the string with three:
    ''';
```

> Also single `'` is already covered by the second capturing group of this regex.

Yeah, and I think that's probably all we need here?

chandlerc:
Also maybe worth coordinating with the author of #3642 who seems to be working on an overlapping area?

czapiga:
I believe it won't be needed now - I want to make a change in different area next time (probably parsing or adding diagnostic). It was just quick followup to my previous PR.

utils/textmate/Syntaxes/Carbon.plist
czapiga: > First quote escaped makes the rest normal, and not the end:
      \'''

Ok, I see.. I'll revert this part.

chandlerc (approved):
Awesome, thanks, and LG!

Merging this as it'll let the file type highlighting be a more focused PR.

PR #3636, opened 2024-01-22
Provide diagnostic locations for imports.

jonmeow:
Note namespaces aren't handled well, because they can be generated from merging imports. That's on my internal TODO list.

### DIFF
--- a/toolchain/check/testdata/packages/fail_conflict_namespace_first.carbon
+++ b/toolchain/check/testdata/packages/fail_conflict_namespace_first.carbon
@@ -3,8 +3,6 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
 // AUTOUPDATE
-// CHECK:STDERR: namespace_then_fn.carbon: ERROR: Duplicate name being declared in the same scope.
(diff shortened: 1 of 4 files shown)


toolchain/check/testdata/packages/fail_conflict_namespace_first.carbon
```diff
@@ -17,6 +15,10 @@ fn NS.Foo() {}
 
 package Example library "fn" api;
 
+// CHECK:STDERR: fn.carbon:[[@LINE+4]]:1: ERROR: Duplicate name being declared in the same scope.
+// CHECK:STDERR: fn NS() {}
+// CHECK:STDERR: ^~~~~~~~~
+// CHECK:STDERR: namespace_then_fn.carbon: Name is previously declared here.
```
zygoloid, line 21: Do you know why we don't get a line and column number here?

zygoloid (approved):

toolchain/check/testdata/packages/fail_conflict_namespace_first.carbon
jonmeow: Yes, see #3640 (different code, support is already pending)

PR #3640, opened 2024-01-23
Provide diagnostic locations for imported namespaces.

jonmeow:
Building on #3636 which handles the general import case, add special casing for namespaces. Namespaces can be combined cross-IR, so it's a little more complex.

The implementation adds import_id to the Namespace instruction as a reference to find the original using the normal structure. This is achieved by moving the name_id to NameScope to free up space.

---

I considered a few alternatives...

I considered adding import_id to NameScope, but:

1. It's more consistent with things such as Function or Class that provide name_id on the info object rather than the instruction.
2. I thought it more likely that there would be more NameScope cases that might want a name_id rather than the import_id, since ImportRef will typically be used.

I considered putting the import source (cross-ref IR id + inst id) on the NameScope versus a separate ImportRef, which seems like the strongest argument towards the NameScope approach because it removes an instruction. That just felt inconsistent though, and the overhead of instruction-per-imported-namespace should be low (theoretically few namespaces should be used). Plus I feel a bit odd adding two generally-unused ids to NameScope.

A specialized Namespace structure could also have been created to store the import_id, but that would add an indirection to the NameScope.


toolchain/check/check.cpp
```diff
@@ -30,20 +30,61 @@ class SemIRLocationTranslator
...
-  Parse::NodeLocationTranslator* node_translator_;
+  auto GetLocationInFile(const SemIR::File* sem_ir,
+                         Parse::NodeLocation node_location) const
+      -> DiagnosticLocation {
+    auto it = node_translators_->find(sem_ir);
+    CARBON_CHECK(it != node_translators_->end());
+    return it->second->GetLocation(node_location);
+  }
+
+  const llvm::DenseMap<const SemIR::File*, Parse::NodeLocationTranslator*>*
```
chandlerc: Minor, and fine to do in a follow-up: might be worth extracting a typedef for this map. It's used in more than one function declaration it seems in addition to here.

toolchain/check/testdata/array/array_in_place.carbon
```diff
@@ -25,7 +25,7 @@ fn G() {
 // CHECK:STDOUT: }
 // CHECK:STDOUT:
 // CHECK:STDOUT: file {
-// CHECK:STDOUT:   package: <namespace> = namespace package, {.F = %F, .G = %G} [template]
+// CHECK:STDOUT:   package: <namespace> = namespace {.F = %F, .G = %G}, invalid [template]
```
chandlerc: Hmm, should this be special cased in the formatter? Specifically, that `invalid` is used to represent a normal state here that just doesn't have any extra information to display?

chandlerc (commented):
Generally really nice!

One super minor thing inline, and a question about the formatted IR representation after this change inline.

toolchain/check/check.cpp
jonmeow: I tend to prefer not doing typedefs in cases like this. https://google.github.io/styleguide/cppguide.html#Aliases I've always read as being reticent about typedefs [and I like to see the map].

toolchain/check/testdata/array/array_in_place.carbon
jonmeow: I guess I have a different interpretation here. `invalid` is the representation of `InstId::Invalid` being present (or probably any `Invalid` value); it's not "no extra information", it _is_ the information.

jonmeow: Changed the code to elide invalid values.

toolchain/check/check.cpp
chandlerc: FWIW, I wasn't suggesting this really be public.

But regardless, this is optional.

toolchain/check/testdata/array/array_in_place.carbon
chandlerc: Thanks, I do somewhat prefer viewing this as a "not used" part of the instruction and not rendering it.

That said, a) you shouldn't block on this as its super easy to iterate here if we change our minds.

And b) good to double check with @zygoloid to get another view maybe?

But definitely go ahead and merge!

chandlerc (approved):

PR #3642, opened 2024-01-23
Add filetype highlighting to block string literals

### LINKED ISSUE #3641: Missing filetype highlighting in block string literals
From the docs on [string literals](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/lexical_conventions/string_literals.md):

> A block string literal starts with '''. Characters on the same line following the ''' are an optional file type indicator. The...

> A file type indicator is any sequence of non-whitespace characters other than ' or #. The file type indicator has no semantic meaning to the Carbon compiler, but some file type indicators are understood by the language tooling (for example, syntax highlighter, code formatter) as indicating the structure of the string literal's content.

However, these filetypes aren't being highlighted:

image

skoriop:
This PR adds filetype highlighting support to block string literals in the TextMate grammar.

After adding the regex in the `begin` of the block string literal, `beginCaptures` is used to access the first capturing group of the regex (which captures the filetype) and apply the `constant.character.escape` rule.

image

Closes #3641

### DIFF
--- a/utils/textmate/Syntaxes/Carbon.plist
+++ b/utils/textmate/Syntaxes/Carbon.plist
@@ -82,7 +82,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
                 <key>name</key>
                 <string>string.quoted.triple.carbon</string>
                 <key>begin</key>
-                <string>'''([^']*\n)?</string>
+                <string>'''([^\s'#]*\n)?</string>
                 <key>end</key>
                 <string>'''</string>
                 <key>beginCaptures</key>
--- a/utils/textmate/Syntaxes/Carbon.plist
+++ b/utils/textmate/Syntaxes/Carbon.plist
@@ -82,10 +82,17 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
                 <key>name</key>
                 <string>string.quoted.triple.carbon</string>
                 <key>begin</key>
-                <string>'''</string>
+                <string>'''([^']*\n)?</string>
                 <key>end</key>
                 <string>'''</string>
-
+                <key>beginCaptures</key>
+                <dict>
+                    <key>1</key>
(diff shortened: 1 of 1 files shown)


chandlerc:
Exciting -- but may be worth coordinating a bit with #3635 as there seems to be some overlap?

chandlerc:
> Exciting -- but may be worth coordinating a bit with #3635 as there seems to be some overlap?

After looking at both, I've queued #3635 to merge as it'll give an isolated fix to the quote style. If you can rebase on top of that, happy to review the highlighting for the file type marker.

Also, thanks again for contributing, really appreciated!

skoriop:
> Exciting -- but may be worth coordinating a bit with https://github.com/carbon-language/carbon-lang/pull/3635 as there seems to be some overlap?

My bad, don't know how I missed that when making my PR :no_mouth: 

I've rebased my commits onto `trunk` now.

utils/textmate/Syntaxes/Carbon.plist
```diff
@@ -82,10 +82,17 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
...
                 <string>'''</string>
-
+                <key>beginCaptures</key>
+                <dict>
+                    <key>1</key>
+                    <dict>
+                        <key>name</key>
+                        <string>constant.character.escape.carbon</string>
+                    </dict>
+                </dict>
```
chandlerc, line 95: This may just be that I don't know as much about the vscode syntax (sorry for that), but could you mention this part of the change and what it does in the description?

skoriop: Yup, I've updated the description

chandlerc: Ah, I see! Thanks!

chandlerc (approved):
LG

PR #3643, opened 2024-01-23
Refactor instruction profile functions.

jonmeow:
I was thinking about different ways to implement our discussion. Note I have two commits with slightly different approaches here. I lean towards the "Different approach" commit.

The main change here is probably SelectProfileArgFunction versus SelectProfileArgs. I was trying to write a `TryArgType<InstT, N>` which would allow removing SelectProfileArgs but couldn't get that to work. Approaches I thought of either ran into tuple element bounds checking, or unsupported C++ features.

I backed away from casting (as in the earlier commit) because it felt like it created more wrappers, due to BuiltinKind and int32_t direct use (on AnyBranch, "kind-specific data"). These are things that, if we added generic `arg0_typed<ArgType>() -> ArgType` support on Inst, could probably be factored a little better. But I was trying to operate within a narrow refactoring scope to try to see what I could do in a narrow scope, without modifying Inst.


zygoloid:
I think this will result in more code size than we had before, because we won't use the same implementation for the different ID kinds that hash as just an integer any more. I think we can change this a little and avoid that, though.

I've sent you a PR against your PR: https://github.com/jonmeow/carbon-lang/pull/1 :)

zygoloid (dismissed):
LG, commit message should probably be updated before this is merged.

jonmeow:
Sorry, to try to explain... I merged in your suggestion so that I could look at it more closely, but I was still looking for ways to provide a simpler approach. I made some changes, could you review now?

toolchain/sem_ir/inst_profile.cpp
```diff
@@ -4,110 +4,115 @@
...
-    } else if constexpr (std::is_same_v<ArgT, IntId>) {
-      return IntProfileArgFunction;
-    } else if constexpr (std::is_same_v<ArgT, RealId>) {
-      return RealProfileArgFunction;
+    if constexpr (IsInvocable<ArgT>(
+                      [](auto&& arg)
+                          -> decltype(ProfileArg(
+                              std::declval<llvm::FoldingSetNodeID&>(),
+                              std::declval<const File&>(), arg)) {})) {
+      return std::declval<ArgT>();
```
zygoloid, line 91: I think it's better to have an explicit list of the ID types that use the `int32_t` profiling rather than using that profiling automatically for newly-added ID types, so that explicit action is required when adding a new ID type to choose the profiling behavior. (In particular, while making the list I noticed that we were using `int32_t` profiling for `BindNameId` and for `StringLiteralValueId`, both of which seem questionable.)

jonmeow:
Per discussion... We may look instead at deduplicating blocks and ints/reals before they get added, so that the IDs would be profiled directly. That'd get overloading working directly. Neither of us are super happy about the options here, but generally agreeing that that kind of approach would be better. The current code is reasonable, so agreement was towards closing this and focusing more on a long-term where we wouldn't need to special-case IDs for profiling here.

PR #3645, opened 2024-01-24
Propagate the phase of a type from its constituent types.

zygoloid:
To avoid bouncing through `constant_values()` to determine whether a type is symbolic or template, store the `ConstantId` on the `TypeInfo` not just the `InstId`.

In addition to propagating the symbolic / template phase, this also propagates whether a type contains an error, resulting in our no longer producing types such as `<error>*` -- these now evaluate to simply `<error>`. While this makes our types less precise after an error, it also removes some follow-on diagnostics, so it seems to be an improvement on the whole.

### DIFF
--- a/toolchain/check/eval.cpp
+++ b/toolchain/check/eval.cpp
@@ -83,6 +83,14 @@ static auto GetConstantValue(Context& context, SemIR::InstId inst_id,
   return const_id.inst_id();
 }
 
+// A type is always constant, but we still need to extract its phase.
+static auto GetConstantValue(Context& context, SemIR::TypeId type_id,
+                             Phase* phase) -> SemIR::TypeId {
+  auto const_id = context.types().GetConstantId(type_id);
+  *phase = LatestPhase(*phase, GetPhase(const_id));
+  return type_id;
+}
+
 // If the given instruction block contains only constants, returns a
 // corresponding block of those values.
 static auto GetConstantValue(Context& context, SemIR::InstBlockId inst_block_id,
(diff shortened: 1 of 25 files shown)


toolchain/sem_ir/value_stores.h
```diff
@@ -189,18 +189,24 @@ class TypeStore : public ValueStore<TypeId> {
   explicit TypeStore(InstStore* insts) : insts_(insts) {}
 
   // Returns the ID of the instruction used to define the specified type.
```
jonmeow: Should this say "constant" instead of "instruction"?
```suggestion
  // Returns the ID of the constant used to define the specified type.
```

toolchain/check/eval.cpp
```diff
@@ -109,6 +117,17 @@ static auto GetConstantValue(Context& context, SemIR::InstBlockId inst_block_id,
   return context.inst_blocks().Add(const_insts);
 }
 
+// The constant value of a type block is that type block, but we still need to
+// extract its phase.
+static auto GetConstantValue(Context& context, SemIR::TypeBlockId type_block_id,
+                             Phase* phase) -> SemIR::TypeBlockId {
+  auto types = context.type_blocks().Get(type_block_id);
+  for (auto type_id : types) {
+    GetConstantValue(context, type_id, phase);
```
jonmeow, line 126: Should this short-circuit if the phase is runtime?

toolchain/check/testdata/struct/fail_nested_incomplete.carbon
```diff
@@ -14,9 +14,6 @@ class Incomplete;
 // CHECK:STDERR: ^~~~~~~~~~~~~~~~~
 var s: {.a: Incomplete};
 
-// CHECK:STDERR: fail_nested_incomplete.carbon:[[@LINE+3]]:1: ERROR: Cannot implicitly convert from `<error>*` to `Incomplete*`.
-// CHECK:STDERR: var p: Incomplete* = &s.a;
-// CHECK:STDERR: ^~~~~~~~~~~~~~~~~~~~~~~~~~
 var p: Incomplete* = &s.a;
```
jonmeow, line 17: Should this be moved out of this fail test, since it no longer fails?

jonmeow (approved):
LGTM with a few small questions.

toolchain/check/testdata/struct/fail_nested_incomplete.carbon
zygoloid: This is testing error recovery, and so has to be in a `fail_` file.

toolchain/check/eval.cpp
zygoloid: That case is not possible: we only form a`TypeId`s for type constants, so we can't find a runtime subexpression here. That does mean that we could bail out if we hit an error, but given that it's a bit subtle why that would be correct, and it would only improve performance along an error path (and marginally slow down the other paths), maybe it's not worth it.

toolchain/sem_ir/value_stores.h

Issue #2191, opened 2022-09-16
one-tuples and one-tuple syntax


zygoloid:
We have a longstanding [open question](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/tuples.md#single-value-tuples) on tuple syntax: can you write a one-tuple, and if so, how? (See also #2038 for some recent discussion).

Possibilities:

0) Carbon has no 1-tuples (like Haskell, Swift, Scala). `(a)` is the same as `a` and in simple cases can be used as if it were a 1-tuple.
1) Allow specifically `(` _expression_ `,` `)` for this. Do not allow a trailing comma in any other cases (previously discussed but seems unprecedented in other languages).
2) Allow a trailing comma in tuple syntax for non-empty tuples, that is mandatory when constructing a 1-tuple but otherwise optional; don't allow trailing commas in other parenthesized lists like function calls (like Chapel).
3) Allow a trailing comma in all non-empty parenthesized comma-separated lists; it is mandatory when the parentheses could otherwise be interpreted as grouping parens (like Python, Rust, Julia).
4) Some blurring of the line between values and 1-tuples, so we have both but they're usually interchangeable.
5) Change tuple syntax entirely to avoid the problem, such as using `(@, first_elem, second_elem)` (suggested by @tkoeppe on #2038).

zygoloid:
I think that we need 1-tuples for generic code, if nothing else. The case of exactly one result should be expected to be common, and treating it as some kind of special case seems especially problematic. Note that `GetFirstElement(MakeTuple(a, b...))` would misbehave under option 0, particularly if `b` is empty and `a` is itself a tuple. I think option 0 is a bad fit given our goals for generic programming.

Of the remaining options, 4 and 5 seem too inventive, and 4 seems likely to lead to confusion. Option 3 is the most consistent choice, has plenty of precedent, and doesn't seem to solicit complaints -- in fact, while researching this I found a lot of user sentiment for Swift to switch from option 0 to option 3.

So I propose we select option 3, at least for now. This seems like a relatively easy syntactic decision to revisit later, unless the decision is to switch to option 0, but as noted I think that's a poor fit for Carbon and will remain so.

josh11b:
I'd just like to express that allowing a trailing comma has been used to make it easier to edit lists and to make diffs smaller in source control when making changes. See, for example, https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#trailing-commas

[...]

PR #3646, opened 2024-01-24
Tuples and tuple indexing

zygoloid:
Add support for extracting elements of a tuple by their numerical index.

Also formally add the well-established basic syntactic and semantic rules for
tuples, for which we have had leads issues but no proposal, into the design.

### DIFF
--- a/docs/design/expressions/member_access.md
+++ b/docs/design/expressions/member_access.md
@@ -863,6 +863,11 @@ let a: i32 = (41, 42, 43).1;
 let b: i32 = (1, 2, 3).0x1;
 // ❌ Error: no tuple element named `2`.
 let c: i32 = (1, 2).2;
+
+var t: (i32, i32, i32) = (1, 2, 3);
+let p: (i32, i32, i32)* = &t;
+// ✅ `m == 3`.
+let m: i32 = p->2;
 ```
 
 In a compound member access of the form:
@@ -874,6 +879,18 @@ in which the first _expression_ is a tuple and the second _expression_ is of
 integer or integer literal type, the second _expression_ is required to be a
 non-negative template constant that is less than the number of tuple elements,
 and the result is the corresponding positional element of the tuple.
+
+```
+// ✅ `d == 43`.
+let d: i32 = (41, 42, 43).(1 + 1);
+// ✅ `e == 2`.
+let template e:! i32 = (1, 2, 3).(0x1);
+// ❌ Error: no tuple element with index 4.
+let f: i32 = (1, 2).(2 * 2);
+
+// ✅ `n == 3`.
+let n: i32 = p->(e);
+```
 
 ## Precedence and associativity
 
--- a/docs/design/expressions/member_access.md
+++ b/docs/design/expressions/member_access.md
@@ -849,13 +849,20 @@ A tuple indexing expression is of the form:
 -   _expression_ `.` _integer-literal_
 -   _expression_ `->` _integer-literal_
 
-The _expression_ is required to be of tuple type, and _integer-literal_ is
-required to be less than the number of tuple elements. The result is the
-corresponding positional element of the tuple.
+The _expression_ is required to be of tuple type.
+
+Each positional element of the tuple is considered to have a name that is the
+corresponding decimal integer: `0`, `1`, and so on. The spelling of the
+_integer-literal_ is required to exactly match one of those names, and the
+result is the corresponding element of the tuple.
 
 ```
-// `n == 42`.
-let n: i32 = (41, 42, 43).1;
+// ✅ `a == 42`.
+let a: i32 = (41, 42, 43).1;
+// ❌ Error: no tuple element named `0x1`.
+let b: i32 = (1, 2, 3).0x1;
+// ❌ Error: no tuple element named `2`.
+let c: i32 = (1, 2).2;
 ```
 
 In a compound member access of the form:
--- a/docs/design/expressions/member_access.md
+++ b/docs/design/expressions/member_access.md
@@ -24,6 +24,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 -   [Instance binding](#instance-binding)
 -   [Non-instance members](#non-instance-members)
 -   [Non-vacuous member access restriction](#non-vacuous-member-access-restriction)
+-   [Tuple indexing](#tuple-indexing)
 -   [Precedence and associativity](#precedence-and-associativity)
 -   [Alternatives considered](#alternatives-considered)
 -   [References](#references)
@@ -840,6 +841,32 @@ alias X3 = (i32 as Factory).Make;
 // ❌ Error, compound access without impl lookup or instance binding.
 alias X4 = i32.((i32 as Factory).Make);
 ```
+
+## Tuple indexing
+
+A tuple indexing expression is of the form:
+
+-   _expression_ `.` _integer-literal_
+-   _expression_ `->` _integer-literal_
+
+The _expression_ is required to be of tuple type, and _integer-literal_ is
+required to be less than the number of tuple elements. The result is the
+corresponding positional element of the tuple.
+
+```
+// `n == 42`.
+let n: i32 = (41, 42, 43).1;
+```
+
+In a compound member access of the form:
+
+-   _expression_ `.` `(` _expression_ `)`
+-   _expression_ `->` `(` _expression_ `)`
+
+in which the first _expression_ is a tuple and the second _expression_ is of
+integer or integer literal type, the second _expression_ is required to be a
+non-negative template constant that is less than the number of tuple elements,
+and the result is the corresponding positional element of the tuple.
 
 ## Precedence and associativity
 
--- a/docs/design/lexical_conventions/README.md
+++ b/docs/design/lexical_conventions/README.md
@@ -36,7 +36,7 @@ The sequence of lexical elements is formed by repeatedly removing the longest
 initial sequence of characters that forms a valid lexical element, with the
 following exception:
 
--   When a numeric literal immediately follows a `.` or `->` symbol, with no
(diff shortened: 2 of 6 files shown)


proposals/p3646.md
```diff
@@ -0,0 +1,252 @@
...
+If a future proposal introduces tuple slicing, it should revisit this question,
+because this kind of indexing from the end is often desirable when forming a
+slice. The possibility of using a different notation for this operation should
+be considered, such as `tuple.(.size - 1)`.
+
+### Trailing commas
+
+Carbon permits optional trailing commas in tuples, with mandatory trailing
+commas for one-tuples. Alternatives to this choice were considered in
+[leads issue #111](https://github.com/carbon-language/carbon-lang/issues/2191).
```
josh11b: Is this #111 or #2191?

docs/design/expressions/member_access.md
```diff
@@ -841,6 +842,32 @@ alias X3 = (i32 as Factory).Make;
...
+## Tuple indexing
+
+A tuple indexing expression is of the form:
+
+-   _expression_ `.` _integer-literal_
+-   _expression_ `->` _integer-literal_
+
+The _expression_ is required to be of tuple type, and _integer-literal_ is
+required to be less than the number of tuple elements. The result is the
+corresponding positional element of the tuple.
```
chandlerc: Can the index be negative?

proposals/p3646.md
```diff
@@ -0,0 +1,252 @@
...
+
+// Error, no field named `0x0`.
+(1, 2, 3).0x0;
+```
+
+We could apply the same restriction in Carbon. However, this restriction would
+introduce inconsistency between `.0x0` and `.(0x0)`, the latter of which does
+not exist in Rust. While we don't imagine a lot of use for these indexes, it's
+possible that a binary index might be useful in some special cases, and there
+doesn't seem to be a reason to disallow this beyond coding style.
```
chandlerc: FWIW, I lean towards the Rust restriction.

There are already lexical inconsistencies here, for example the fact that it doesn't form a real literal when followed by a `.`.

Personally, I feel like it will be easier to teach that these are "positional names" with their own, much narrower set of rules than integer literals.

But interested in what others think about this specific issue -- and I wonder if it would be worth extract this into a specific issue for leads to record this aspect of the decision?

chandlerc (commented):
Super excited to see this in a concrete proposal!

Some quick initial thoughts inline.

proposals/p3646.md
```diff
@@ -0,0 +1,252 @@
...
+
+If the tuple index is not between 0 and one less than the number of elements in
+the tuple, inclusive, the indexing is invalid.
+
+### Tuple slicing
+
+The current skeleton design suggests using `tuple[a .. b]` to slice tuples. For
+example, `tuple[0 .. 2]` could be used to extract the first two elements of a
+tuple. Tuple slicing support is not covered by this proposal, but could be added
+in the future with syntax such as `tuple.(0 .. 2)`.
```
josh11b: I like this approach. I think it is worth documenting the concern that it might lead to an incorrect theory about how Carbon works, where `tuple.__` gives you an element value but `tuple.(__)` gives you a tuple, though I don't think that should stop us from adopting this approach.

jonmeow:
To be sure:

```
let A: i32 = 1;
let Foo: auto = (41, 42);
// Invalid:
var b: i32 = Foo.A;
// Valid:
var c: i32 = Foo.(A);
```

Note, I'm not sure how I feel about this; the `(1,).0` syntax will probably take some getting used to, it looks a lot like other syntaxes [and I'm super-familiar with Python syntax]. Just want to be sure my interpretation is correct.

proposals/p3646.md
zygoloid: Hah, both, but I meant #2191.
```suggestion
[leads issue #2191](https://github.com/carbon-language/carbon-lang/issues/2191).
```

docs/design/expressions/member_access.md
zygoloid: No, because an *integer-literal* token is never negative.

proposals/p3646.md
zygoloid: I don't feel strongly about this at all. If we want to say that we treat `.0`, `.1`, ... as the *names* of the elements of a tuple, I think that's a reasonable reason to add a restriction. If we do add such a restriction, I think we should also forbid digit separators in these indexes. (I can't really imagine ever wanting digit separators here -- indexing a `1_000`-tuple seems ... implausible, and allowing them would not be compatible with the idea that these are numeric field names.)

zygoloid:
> To be sure:
> 
> ```
> let A: i32 = 1;
> let Foo: auto = (41, 42);
> // Invalid:
> var b: i32 = Foo.A;
> // Valid:
> var c: i32 = Foo.(A);
> ```

You would need `template A:! i32 = 1;`, because we need a template constant as the index, but otherwise yes.

proposals/p3646.md
chandlerc: FWIW, you'd still be able to use digit separators if useful: `huge_tuple.(1_000)`.

zygoloid: Done.

proposals/p3646.md
zygoloid: Added some text mentioning that concern.

docs/design/lexical_conventions/README.md
```diff
@@ -33,4 +33,11 @@ A _lexical element_ is one of the following:
...
 The sequence of lexical elements is formed by repeatedly removing the longest
-initial sequence of characters that forms a valid lexical element.
+initial sequence of characters that forms a valid lexical element, with the
+following exception:
+
+-   When a numeric literal immediately follows a `.` or `->` symbol, with no
+    intervening whitespace, a real literal is never formed. Instead, the token
+    will end no later than the next `.` character. For example, `tuple.1.2` is
+    five tokens, `tuple` `.` `1` `.` `2`, not three tokens, `tuple` `.` `1.2`.
+    However, `tuple . 1.2` is lexed as three tokens.
```
chandlerc: Hmm, I feel like there is a slight clarification that might be needed here:

```suggestion
-   When a numeric literal immediately follows a `.` (that isn't itself part of a numeric literal) or `->` symbol, with no
    intervening whitespace, a real literal is never formed. Instead, the token
    will end no later than the next `.` character. For example, `tuple.1.2` is
    five tokens, `tuple` `.` `1` `.` `2`, not three tokens, `tuple` `.` `1.2`.
    However, `tuple . 1.2` is lexed as three tokens.
```

Not sure this is quite right, and I don't think it's strictly needed -- we'd never consider such a `.` because it'd be consumed forming the numeric literal that contains it, and so it may not even technically be a symbol... But I feel like it's too easy to imagine an ambiguity here, and so might be worth making it a bit more explicit.

proposals/p3646.md
```diff
@@ -0,0 +1,284 @@
...
+```python
+tup = (1, 2, 3)
+# Prints 2.
+print(tup[1])
+```
+
+In C++, `std::pair` is indexed using `.first` and `.second`, and `std::tuple` is
+indexed using `std::get<I>`.
+
+In Rust, a tuple is indexed using `.N`, where `N` is a decimal integer literal.
```
chandlerc:
```suggestion
In both Rust and Swift, a tuple is indexed using `.N`, where `N` is a decimal zero-based integer literal.
```

I think this is true. Is it worth including links? This is a bit minimally documented in Swift, but I found it here:
https://docs.swift.org/swift-book/documentation/the-swift-programming-language/expressions#Explicit-Member-Expression:~:text=The%20members%20of%20a%20tuple%20are%20implicitly%20named%20using%20integers%20in%20the%20order%20they%20appear%2C%20starting%20from%20zero.%20For%20example%3A

josh11b (approved):

proposals/p3646.md
zygoloid: Interesting. This does seem to be true in Swift, but there's a quirk: Swift does not allow digit separators, nor non-decimal field indexes, but it *does* allow leading zeroes: `(10, 20, 30).01` is permitted and treated as `(10, 20, 30).1`.

I'm not sure if that's intentional or a bug :-)

docs/design/lexical_conventions/README.md
zygoloid: Replaced "symbol" with "token", to avoid a possible misunderstanding that the `.` might be part of some other token. Does that work for you?

proposals/p3646.md
zygoloid: Updated the text to cover this, and generally compare the proposed feature to Swift.

docs/design/expressions/member_access.md
```diff
@@ -841,6 +842,39 @@ alias X3 = (i32 as Factory).Make;
 alias X4 = i32.((i32 as Factory).Make);
 ```
 
+## Tuple indexing
+
+A tuple indexing expression is of the form:
+
+-   _expression_ `.` _integer-literal_
+-   _expression_ `->` _integer-literal_
```
chandlerc, line 850: Kate pointed out to me that none of the examples here or in the proposal use this syntax. Could you add some just so that it's clear to folks that anchor on the examples that this is included?

zygoloid: Done.

chandlerc (approved):
Approving for leads based on our discussion and merging!

docs/design/expressions/member_access.md
```diff
@@ -841,6 +842,56 @@ alias X3 = (i32 as Factory).Make;
...
+
+The _expression_ is required to be of tuple type.
```
geoffromer, line 852: Presumably this only applies to the `.` form, and the _expression_ in the `->` form is required to be of type pointer to tuple?

There's a similar problem on line 878.

PR #3647, opened 2024-01-24
Implement merging of namespace declarations.

jonmeow:
Per discussion with zygoloid, namespace declarations will merge both when repeated in a given file, and across files. This echoes how forward declarations of other entities are allowed to repeat. When merging with an imported namespace, fill in the parse node so that future diagnostics point at the declaration in the same file rather than a declaration in a different file, just for locality.

Although it might be desirable to issue a diagnostic when a namespace declaration is repeated within a given file, that's similarly true in other cases, but may be more desirable as a tidy-style issue rather than preventing code from compiling. Allowing the repetition also makes the import versus non-import cases more consistent: the namespace declarations merge regardless of the source.


chandlerc (approved):
LG

PR #3648, opened 2024-01-24
Support determining `IdKind` from `NodeCategory`, in addition to `NodeKind`

josh11b:
The categories `Expr`, `MemberName`, `Decl`, `Statement`, and `Modifier` are usable since they are associated with a consistent `IdKind`. The mapping to `IdKind` for NodeKinds that have those categories are no longer listed explicitly, ensuring that the `NodeCategory` mapping is the source of truth.

Also: fixes the category of the `FunctionDefinitionStart` and `ArrayExprStart` node kinds.

Note: I've added [a section on defining constexpr constants to the Toolchain architecture doc](https://docs.google.com/document/d/1RRYMm42osyqhI2LyjrjockYCutQ5dOf8Abu50kTrkX0/edit?resourcekey=0-kHyqOESbOHmzZphUbtLrTw&tab=t.0#heading=h.f7682a2tpvxr).

FUTURE:

* We should switch `TuplePattern` to put an `InstId` on the `NodeStack` instead of an `InstBlockId`, so we can handle the pattern category.
* We should make a category for names to replace uses of the `NameId` `IdKind`.
* We should make use of these new APIs more, and propagate more-precise types through the codebase.

QUESTION: Should I use a different approach for determining the number of members of the `NodeKind` enum?


toolchain/check/node_stack.h
```diff
@@ -278,6 +354,27 @@ class NodeStack {
...
+    RequireParseCategory<RequiredParseCategory>(back.parse_node);
+    constexpr std::optional<IdKind> RequiredIdKind =
+        ParseNodeCategoryToIdKind(RequiredParseCategory);
+    CARBON_CHECK(RequiredIdKind.has_value())
+        << "NodeCategory " << RequiredParseCategory
+        << " doesn't correspond to an IdKind";
+    if constexpr (*RequiredIdKind == IdKind::InstId) {
+      return back.id<SemIR::InstId>();
+    }
+    if constexpr (*RequiredIdKind == IdKind::NameId) {
```
zygoloid: `PopWithParseNode` also has an `InstBlockId` case. Should the two functions be consistent?

Can we avoid duplicating the list here somehow? Maybe we could implement `Pop` as `Peek` followed by a `pop_back`?

toolchain/check/node_stack.h
```diff
@@ -194,6 +208,38 @@ class NodeStack {
                    << "; see value in ParseNodeKindToIdKind";
   }
 
+  // Pops the top of the stack and returns the parse_node and the ID.
+  template <Parse::NodeCategory RequiredParseCategory>
+  auto PopWithParseNode() -> auto {
+    constexpr std::optional<IdKind> RequiredIdKind =
+        ParseNodeCategoryToIdKind(RequiredParseCategory);
+    CARBON_CHECK(RequiredIdKind.has_value())
```
zygoloid: Can we `static_assert` this?

toolchain/check/node_stack.h
```diff
@@ -278,6 +354,27 @@ class NodeStack {
...
   }
 
+  // Peeks at the ID associated with the top of the name stack.
+  template <Parse::NodeCategory RequiredParseCategory>
+  auto Peek() const -> auto {
+    Entry back = stack_.back();
+    RequireParseCategory<RequiredParseCategory>(back.parse_node);
+    constexpr std::optional<IdKind> RequiredIdKind =
+        ParseNodeCategoryToIdKind(RequiredParseCategory);
+    CARBON_CHECK(RequiredIdKind.has_value())
```
zygoloid: Same comment as above re `static_assert`.

toolchain/check/node_stack.h
```diff
@@ -194,6 +208,38 @@ class NodeStack {
...
+      return node_id_cast(back);
+    }
+    if constexpr (RequiredIdKind == IdKind::NameId) {
+      auto back = PopWithParseNode<SemIR::NameId>();
+      RequireParseCategory<RequiredParseCategory>(back.first);
+      return node_id_cast(back);
+    }
+    CARBON_FATAL() << "Unpoppable IdKind for parse category: "
+                   << RequiredParseCategory
+                   << "; see value in ParseNodeCategoryToIdKind";
```
zygoloid: `static_assert` here too would be good I think. (We'll need to change most of the `if constexpr`s to `else if constexpr`s and put this in a final `else`.)

toolchain/check/node_stack.h
```diff
@@ -363,91 +461,109 @@ class NodeStack {
   };
   static_assert(sizeof(Entry) == 8, "Unexpected Entry size");
 
+  // Translate a parse node category to the enum ID kind it should always
+  // provide, if it is consistent.
+  static constexpr auto ParseNodeCategoryToIdKind(Parse::NodeCategory category)
```
zygoloid, line 439: It's not obvious to me what `category` means here in the case where it's not just a single bit: does it mean that we want an `IdKind` that works for anything with an intersection with `category`, or an `IdKind` that works for anything whose category is a superset of `category`?

If the former, then I think we should also return `std::nullopt` for categories including bits outside the five we can handle, because (eg) if we're asked for an `Expr | Whatever`, and it's actually just a `Whatever`, it might not have an `instId`.

zygoloid (commented):
Generally LG

toolchain/check/node_stack.h
josh11b: Done.

Note that this was correct for the categories of parse nodes, but not for categories of queries. To see the difference, observe that the `IdentifierName` node is in `NodeCategory::NameComponent | NodeCategory::MemberName`, which the prior code correctly identified as `NameId`, but the new code does not. This new code is more conservative, though, and so will be less likely to be misused.

toolchain/check/node_stack.h
josh11b: Done.

toolchain/check/node_stack.h
josh11b: Done.

toolchain/check/node_stack.h
josh11b: How do you write that? I tried
```
if constexpr (...) {
} else if constexpr (...) {
} else {
  static_assert(false);
}
```
and it gave me a compile error.

toolchain/check/node_stack.h
josh11b: Done. `InstId` and `NameId` are the only ones currently associated with a category.

toolchain/check/node_stack.h
zygoloid: It looks like we don't require a new enough Clang for that: https://clang.llvm.org/cxx_dr_status.html#2518

In the mean time we could do this:
```c++
if constexpr (cond1) {
  // ...
} else if constexpr (cond2) {
  // ...
// instead of `} else if constexpr (cond3) {`:
} else {
  static_assert(cond3);
  // ...
}
```

toolchain/check/node_stack.h
```diff
@@ -278,6 +331,25 @@ class NodeStack {
...
+    static_assert(RequiredIdKind.has_value());
+    if constexpr (*RequiredIdKind == IdKind::InstId) {
+      return back.id<SemIR::InstId>();
+    }
+    if constexpr (*RequiredIdKind == IdKind::NameId) {
+      return back.id<SemIR::NameId>();
+    }
+    CARBON_FATAL() << "Unpeekable IdKind for parse category: "
+                   << RequiredParseCategory
+                   << "; see value in ParseNodeCategoryToIdKind";
```
zygoloid:
```suggestion
    if constexpr (*RequiredIdKind == IdKind::InstId) {
      return back.id<SemIR::InstId>();
    } else {
      static_assert(*RequiredIdKind == IdKind::NameId,
                    "Unpeekable IdKind for parse category");
      return back.id<SemIR::NameId>();
    }
```

zygoloid (approved):
LG either with suggested change or (if you prefer) as-is.

PR #3649, opened 2024-01-24
Update project to require Clang 16 or newer.

chandlerc:
This should also unblock our switch to C++20 and other improvements.

There are three core parts of the change --

1) Updating our infrastructure to fetch and find Clang-16.
2) Updating our documentation to reflect this and help folks with any
   system issues they encounter.

The infrastructure change is unfortunately tricky. We can't get Clang 16 easily on GitHub's runner images, and in the past we've had persistent problems with flakiness when our actions download this much during their runs. Due to the flakiness, we've previously removed all downloading of dependencies outside of Bazel itself, and added retry loops around Bazel specifically to overcome flaky downloads.

This change tries to address these problems by populating the Clang and LLVM toolchain in a place that we can then cache using the built-in GitHub action caching infrastructure. This seems like by far the least likely to flake way of downloading extra things into our runs. And since these are relatively slow moving dependencies, we should populate this cache very, very rarely.

For Linux, this downloads the binary release artifact from GitHub, prunes out large parts of it that we don't need, and then caches this as a local toolchain. This proves both small and fast.

For macOS, this uses a trick to cache the destination of Homebrew installs. It unfortunately caches the *entire* Homebrew installation though, and so it also goes to some lengths to prune and minimize how much is installed from Homebrew. The result is "only" a 2gb cache image. Because of the size and slower download and filesystem, the macOS runs see a 1 - 2 minute slowdown.

We might extend the Linux infrastructure here usefully if we want to test multiple LLVM versions. We might also extend the macOS version to get a cheaper way to prune parts of the system and free up disk space, or to cache other Homebrew installed tools if needed.

Last but not least, this brought to the forefront an issue with our C++ toolchain integration which relied on a specific CMake build option being set in the LLVM toolchain install. This option isn't used in the official release artifacts. Instead, switch to a more robust approach to linking libc++abi statically that shouldn't have these problems.

Beyond the infrastructure changes, this also updates the documentation to reflect requiring Clang 16 or newer, and adds some extra tips for folks that are missing this.

The documentation is also updated to address a problem with getting the right libc++abi files installed to support the more robust linking strategy. This may reduce the problems we've seen in the past around libc++abi and linking on other Linux distros as well.

### DIFF
--- a/docs/project/contribution_tools.md
+++ b/docs/project/contribution_tools.md
@@ -42,12 +42,17 @@ These commands should help set up a development environment on your machine.
 # Update apt.
 sudo apt update
 
+# Check that the `clang` version is at least 16, our minimum version. That needs
+# the number of the `:` in the output to be over 16. For example, `1:16.0-57`.
+apt-cache show clang | grep 'Version:'
+
 # Install tools.
 sudo apt install \
   bazel \
   clang \
   gh \
   libc++-dev \
+  libc++abi-dev \
   lld \
   python3 \
   pipx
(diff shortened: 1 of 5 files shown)


docs/project/contribution_tools.md
```diff
@@ -65,7 +70,24 @@ pre-commit install
 bazel test //...:all
 ```
 
-> NOTE: Most LLVM 14+ installs should build Carbon. If you're having issues, see
+If the version of `clang` is earlier than 16, you may still gave version 16
```
josh11b:
```suggestion
If the version of `clang` is earlier than 16, you may still have version 16
```

josh11b (commented):
I feel pretty out of my depth reviewing this change. Perhaps you could get some additional eyes on it?

zygoloid (approved):

zygoloid (approved):

PR #3651, opened 2024-01-25
Add import support for remaining decl types.

jonmeow:
I'd excluded these initially just because I was thinking towards copies, but under the current model I'm trying to catch all the decl types just for consistency. Note references will still be a TODO error (LazyImportRef is already tested for this, it just didn't feel necessary to add individual tests while I try to sort out behavior).

Fixes an oversight where declarations in an entity's scope were being added to the list of exports.

Note I'm trimming some Import API arguments as now-unused.


chandlerc (approved):
LG!

PR #3653, opened 2024-01-25
Remove some C++17 workarounds now we build in C++20 mode.

zygoloid:


chandlerc:
I suspect we should wait at least a day or two to make sure the new GitHub stuff sticks without becoming flaky before we start going in and leveraging it.

One thing that might make sense to update alongside this is our style guide to clarify that we're actively using C++20 now.

chandlerc:
The new infra seems to be surviving with quite a few runs without issue, so I think we should be good to move forward here.

But I think you'll need to rebase on top of #3654 at least and maybe fix some new issues with clang-tidy.

Beyond that, up to you if you'd like to do the style guide update here or I'm happy to do it myself as a follow-up.

toolchain/sem_ir/inst.h
```diff
@@ -103,6 +93,10 @@ struct InstLikeTypeInfo<
   }
 };
 
+// A type is InstLike if InstLikeTypeInfo is defined for it.
+template <typename T>
+concept InstLikeType = requires { sizeof(InstLikeTypeInfo<T>); };
```
chandlerc, line 100: The Google style guide ended up recommending keeping new concept definitions private rather than part of the visible API:
https://google.github.io/styleguide/cppguide.html#Concepts

Could this potentially be defined as a private member of `Inst` below?

toolchain/sem_ir/inst.h
```diff
@@ -124,8 +118,7 @@ struct InstLikeTypeInfo<
 //   data where the instruction's kind is not known.
 class Inst : public Printable<Inst> {
  public:
-  template <typename TypedInst,
-            typename Info = typename InstLikeTypeInfo<TypedInst>::Self>
+  template <InstLikeType TypedInst>
```
chandlerc: The style guide suggests to avoid this syntax, and prefers something like this from my reading:

```suggestion
  template <typename TypedInst>
    requires InstLikeType<TypedInst>
```

toolchain/sem_ir/typed_insts.h
```diff
@@ -737,18 +737,23 @@ struct VarStorage {
   NameId name_id;
 };
 
+// HasParseNode is true if T has an associated parse node.
+template <typename T>
+concept HasParseNode = !std::is_same_v<typename decltype(T::Kind)::TypedNodeId,
+                                       Parse::InvalidNodeId>;
```
chandlerc: Similar to above, is there a way to make these internal to the library and not something exposed as part of its API?

toolchain/sem_ir/inst.h
```diff
@@ -55,13 +50,10 @@ struct InstLikeTypeInfoBase {
...
 // A particular type of instruction is instruction-like.
 template <typename TypedInst>
-struct InstLikeTypeInfo<
-    TypedInst,
-    static_cast<bool>(
-        std::is_same_v<const InstKind::Definition<
-                           typename decltype(TypedInst::Kind)::TypedNodeId>,
-                       decltype(TypedInst::Kind)>)>
-    : InstLikeTypeInfoBase<TypedInst> {
+  requires std::is_same_v<const InstKind::Definition<
```
chandlerc: Style guidance is to use named concepts when available instead of predicates, here and elsewhere:

```suggestion
  requires std::same_as<const InstKind::Definition<
```

chandlerc (commented):
After spending a bit of time going through the style guidance Google published for concepts, have some suggestions inline.

That said, while still useful as a starting point, totally open if there are aspects of this that don't work out well in practice -- the guidance is inherently less mature than much other guidance.

toolchain/sem_ir/inst.h
zygoloid: We also use it in the `operator<<` at the bottom of this file, which can't be a member of `Inst`.

It seems like the style guide is motivated in large part by brittleness issues with concepts, which I don't think apply to the toolchain in the same way, given that we don't have any kind of API stability, so I'm not sure to what extent we should follow that guidance. If we want to stick with that, we could replace these concepts (indeed, all the concepts in this PR) with `constexpr bool` templates instead, like they were before, given that we don't intend to allow them to be used as *type-constraint*s. What do you think?

chandlerc: On the one hand, I agree that we don't quite have the same level of concern currently as the Google style guide is designed for -- that's true for quite a few aspects of it... But I'm hesitant to fully disregard this aspect.

Fundamentally, `sem_ir` really *should* be an API we expect to have users, and I don't think we'd want this to be part of that API directly, and instead be an implementation detail (of that API).

That said, I don't think we need to defend heavily against mis-use, maybe just a detail namespace or 'internal' namespace?

As for switching to `constexpr bool` functions -- I think the risks are basically the same either way. I'd pick between the two based on what makes the definition simpler / easier-to-understand.

zygoloid: I meant `constexpr bool` variable templates -- which is what these are today. I don't really see a reason to apply different rules for a `constexpr bool` variable template versus for a `concept`, if we don't care about subsumption (which we don't in this case, because this will be an atomic constraint).

chandlerc: I didn't mean to distinguish between variable templates and functions, sorry. And also understand if we have some of these already, mostly trying to figure out how we want to handle things going forward. And so also happy if it shouldn't block this PR but be something we fix forward (to the extent there is any fix).

Anyways, I think the point of the style rule is that when the constraint (whether a constexpr bool variable template or a concept or ...) is likely to be used as a constraint in user-code APIs, it creates overly easily propagated brittleness.

So my feeling is that if this is an issue worth defending against anywhere in the toolchain, it seems likely to be here in what would be a very public API. And I think that would speak to being very cautious about having these in the public API.

If these are important for client code to use as constraints, then maybe they're the rare exception and need to be public concepts. Probably just worth documenting that a bit?

But if they're not used by clients of `sem_ir`, is there harm to putting them in an internal namespace? If so, that seeems like you see an argument to diverge from this aspect of the style guide. I'm open to that I think, but would like to understand it a bit.

zygoloid: Actually, reflecting on that a bit -- I think the key difference between what's happening here and the restriction in the style guide is that these concepts are only intended to be applied to exactly the small set of types defined in `typed_insts.h`, not to an open, extensible set of types. Would it make a difference if we guaranteed that somehow?

toolchain/sem_ir/inst.h
zygoloid: Done, in the code affected by this patch.

Do we want to use `std::same_as` instead of `std::is_same_v` even in places unrelated to constraints, such as in `if constexpr` checks?

toolchain/sem_ir/inst.h
zygoloid: Done.

toolchain/sem_ir/inst.h
zygoloid: I've moved these into a `Detail` namespace, along with the related implementation details in `typed_insts.h`.

toolchain/sem_ir/typed_insts.h
zygoloid: Moved to a `Detail` namespace.

toolchain/sem_ir/inst.h
chandlerc: I don't feel strongly at all, but I would probably just switch to `std::same_as` where it applies for consistency. It seems useful in constraints, and not worth the confusion of having a mixture.

That said, totally open to other approaches here too.

chandlerc: Also, I'd switch to include `<concepts>` -- ClangD doesn't seem to always get this right as yet.

toolchain/sem_ir/inst.h
```diff
@@ -19,20 +19,18 @@
 
 namespace Carbon::SemIR {
 
+// InstLikeTypeInfo is an implementation detail, and not public API.
+namespace Detail {
```
chandlerc: We've called these `Internal` in other parts of the codebase, probably not worth diverging here.

There is a thread in Discord to see if that's the name we want to stick with or change to something else.

toolchain/sem_ir/typed_insts.h
```diff
@@ -737,18 +737,28 @@ struct VarStorage {
   NameId name_id;
 };
 
+// These concepts are an implementation detail of the library, not public API.
+namespace Detail {
```
chandlerc: Just flagging the other place to do s/Detail/Internal/

chandlerc (approved):
A couple of minor suggestions inline, happy for you to merge with those applied. We may still need to iterate a bit, but this looks like a nice step!

toolchain/sem_ir/inst.h
zygoloid: Done.

toolchain/sem_ir/typed_insts.h
zygoloid: Acknowledged.

toolchain/sem_ir/inst.h
zygoloid: Done.

chandlerc (approved):

PR #3654, opened 2024-01-25
Fix outstanding `clang-tidy` errors.

chandlerc:
Recent runs of `clang-tidy` for me started showing more errors, and this is a collection of changes to address them.

First, I've systematically applied the disabling tag to all C++ rules under //explorer/... with `buildozer` so we don't spend time analyzing this code or reporting errors from it. Not sure this was strictly necessary, but it seemed like a nice consistency improvement.

Next, I disabled a buggy check for missing `default` cases in `switch`es. It seems to get confused by the fancy conversions in our `enum_base.h`. We don't miss much with this as the Clang compiler warnings for `switch` catch most of our actual bugs. I also removed the local disabling of this now that it is turned off centrally.

Lastly, I added error checking to two file descriptor manipulating calls in the `file_test` infrastructure.


.clang-tidy
```diff
@@ -16,6 +16,8 @@ WarningsAsErrors: '*'
 #       inst_id = name_ref->value_id;
 #                 ^ unchecked access to optional value
 #     }
+# - bugprone-switch-missing-default-case has false positives for `enum_base.h`
+# and is redundant with Clang's enum warnings.
```
jonmeow:
```suggestion
# - bugprone-switch-missing-default-case has false positives for `enum_base.h`
#   and is redundant with Clang's enum warnings.
```

jonmeow: Can you clarify in your comment how this is redundant? I checked https://clang.llvm.org/extra/clang-tidy/checks/bugprone/switch-missing-default-case.html:

"Enum types are already covered by compiler warnings (comes under -Wswitch) when a switch statement does not handle all enum values. This check focuses on non-enum types where the compiler warnings may not be present."

From the documentation, it sounds like this is actually a false positive because it's incorrectly treating the `enum_base` as a non-enum type (missing the cast behavior).

Note though, I've definitely seen this false positive. And actually, we have `NOLINTNEXTLINE(bugprone-switch-missing-default-case)` in 11 places. Maybe this check doesn't offer enough value because it doesn't have enough _true_ positives on non-enum types, but if your preference is to use .clang-tidy to fully disable instead of NOLINTNEXTLINE, would you please remove all NOLINTNEXTLINEs that correspond?

```
╚╡git grep bugprone-switch-missing-default-case .
toolchain/check/check.cpp:    // NOLINTNEXTLINE(bugprone-switch-missing-default-case)
toolchain/check/context.cpp:    // NOLINTNEXTLINE(bugprone-switch-missing-default-case)
toolchain/check/context.cpp:    // NOLINTNEXTLINE(bugprone-switch-missing-default-case)
toolchain/check/eval.cpp:  // NOLINTNEXTLINE(bugprone-switch-missing-default-case)
toolchain/lower/function_context.cpp:    // NOLINTNEXTLINE(bugprone-switch-missing-default-case)
toolchain/parse/parse.cpp:    // NOLINTNEXTLINE(bugprone-switch-missing-default-case)
toolchain/sem_ir/file.cpp:  // NOLINTNEXTLINE(bugprone-switch-missing-default-case)
toolchain/sem_ir/file.cpp:    // NOLINTNEXTLINE(bugprone-switch-missing-default-case)
toolchain/sem_ir/file.cpp:    // NOLINTNEXTLINE(bugprone-switch-missing-default-case)
toolchain/sem_ir/formatter.cpp:    // NOLINTNEXTLINE(bugprone-switch-missing-default-case)
toolchain/sem_ir/inst.cpp:  // NOLINTNEXTLINE(bugprone-switch-missing-default-case)
```

jonmeow (approved):
Approving, but I think bugprone-switch-missing-default-case might deserve a closer look so I'm not merging.

.clang-tidy
jonmeow: I was thinking about this further... https://github.com/carbon-language/carbon-lang/commit/b9df8ca765f10f15231d3d6b0742a84f23fe68fb added 3 `default` cases [plus the yaml_test_helpers.cpp change], they're probably all due to this warning. I'll let you consider their relative value for true positives.

chandlerc: Doh, should have searched for the nolint lines, sorry about that.

Yeah, seems this is more about true-positive vs. false-positive rate. It seems we've had basically one true positive and all of these false positives. I lean towards just disabling it at the `.clang-tidy` level and ripping out the nolint lines, but curious what others think.

zygoloid: The clang warning seems to work properly for `EnumBase`, so I think this is only adding noise for that case. `EnumBase` `switch`es are *much* more common than `switch`es over actual integers, so even if this were catching a bug every time in the latter case (and I think the false positive rate there for actual bugs rather than stylistic issues is still going to be quite high), I think it'd still be worth turning off.

If we can change `EnumBase` in a way that means we don't trip the lint, that'd work for me too, but it looks to me like we're already doing the best we can there.

chandlerc: I've not found any way to get `EnumBase` to play nicely with the check. =/ I tried a few things but they didn't work.

I've removed all the local disabling though as Jon suggested. And I updated the documentation here to better explain that it's more that Clang's warnings give us enough coverage given we have to weigh against the extra false positives from `EnumBase`.

chandlerc (commented):
Thanks for the thoughts.

I've removed the now-redundant local disablings so I think this should be good, but it is a bit different from the first version so PTAL.

jonmeow:
LGTM

.clang-tidy
```diff
@@ -16,6 +16,8 @@ WarningsAsErrors: '*'
 #       inst_id = name_ref->value_id;
 #                 ^ unchecked access to optional value
 #     }
+# - bugprone-switch-missing-default-case has false positives for `enum_base.h`
```
jonmeow:
```suggestion
# - bugprone-switch-missing-default-case has false positives for `enum_base.h`.
```

PR #3655, opened 2024-01-25
Fix missing abi requirement

jonmeow:
#3649 added a requirement for libc++abi-dev but missed updating the README to mention it.


chandlerc (approved):
LG! I remembered and then promptly forgot again about updates too, thanks for catching it.

PR #3656, opened 2024-01-25
Refactor LazyImportRef into ImportRefUnused.

jonmeow:
This makes some changes to the formatter so that ImportRefUnused and ImportRefUsed will both be labeled as "import_ref" with an "unused -> used" argument change in textual IR, but is otherwise not changing logic.


chandlerc (approved):
LG

PR #3657, opened 2024-01-25
Start adding ImportRefUsed, removing older copy behavior.

jonmeow:
Builds on #3656.

Under the prior "lazy" model, we had been planning to copy instructions. Under the current "unused+used" model, we're restricting that to constants. I'm trimming back some of the ResolveIfImportRefUnused logic because it was more appropriate for the former model.

Right now, I'm adding ImportRefUsed direct creation in import.cpp for namespaces. I think I'll need to do something similar in RseolveIfImportRefUnused... I'm still trying to think about how to manage type information there (which needs to come in as an import reference itself, and probably have some amount of deduplication before forming a TypeId). So ImportRefUnused lacks a type because I'm hesitant to aggressively load it, whereas ImportRefUsed should *always* have a type but it's just an error while I think things through.

For reference, ImportRefUnused and ImportRefUsed are mainly split in order to track the boolean "used" without making fundamental modifications to Inst for bit packing (this effectively instead packs a bit into InstKind). AnyImportRef currently excludes the type because it's mainly for diagnostic printing at the moment. It could end up with a TypeId that would end up Invalid for ImportRefUnused, though it could also be that the TypeId is only accessed when using ImportRefUsed explicitly.


toolchain/check/testdata/packages/fail_todo_lazy_import_ref.carbon
chandlerc, line 1: May want to rename this test case here or in a subsequent PR given that these aren't lazy import refs any more.

chandlerc (approved):
LG

toolchain/check/testdata/packages/fail_todo_lazy_import_ref.carbon
jonmeow: Renamed

jonmeow: Actually, removing, since the tests added by #3656 should supersede this.

jonmeow: And then I realized I didn't have anything for `var`, so I'm just moved it there to correspond with other tests.

PR #3660, opened 2024-01-27
Enable C++20 and fix infrastructure to work with it.

chandlerc:
C++20 mode exposes bugs in `clang-tidy` 16 that are challenging to work around, but this should manage to do so. The patch to `bazel_clang_tidy` has been sent upstream but no need to wait for it.

We don't actually specify a specific version of C++ in our style guide docs, and the Google style guide has already been updated to be based on C++20 so nothing to be done there.


.clang-tidy
```diff
@@ -63,3 +63,9 @@ CheckOptions:
       key: readability-identifier-naming.MethodIgnoredRegexp,
       value: '^classof$',
     }
+  - {
+      # This erroneously fires in C++20 mode with LLVM 16 clang-tidy, likely due
+      # to: https://github.com/llvm/llvm-project/issues/61969
```
zygoloid:
```suggestion
      # This erroneously fires in C++20 mode with LLVM 16 clang-tidy, due to:
      # https://github.com/llvm/llvm-project/issues/46097
```

zygoloid (approved):

josh11b (approved):

.clang-tidy
josh11b: Are  you sure 46097 is right? That seems about `error: invalid case style for template parameter 'expr-type'`, not `clang-diagnostic-builtin-macro-redefined` or suppressed warnings being treated as errors.

zygoloid: The error that's being suppressed here is for a template parameter named `expr-type`, which is 46097. 61969 is the bug number corresponding to the patch instead.

PR #3662, opened 2024-01-29
CrossRefIRId -> ImportIRId

jonmeow:
One more (hopefully last) rename on the Import instruction renaming.

I was kind of tempted to rename to just "IRId", since the IRs aren't all imports. However, this felt easier to read, and a better choice than CrossRef because it's more consistent with the other ways imports exist in code. (even if IRs aren't all imports, most use-cases are derived from imports)

Note though that import_irs may include IRs not just from direct imports. Beyond the builtin IR, I'm thinking that for indirect imports, or the prelude, we may end up adding them. e.g., so that constants can be generated for indirect imports and still correspond to a directly known IR, and for a given IR that's indirectly imported multiple times to be deduplicated locally. I'm not there yet, I'm just mentioning this to help give background for naming thoughts.

### DIFF
--- a/toolchain/sem_ir/file.h
+++ b/toolchain/sem_ir/file.h
@@ -270,9 +270,9 @@ class File : public Printable<File> {
   auto interfaces() const -> const ValueStore<InterfaceId>& {
     return interfaces_;
   }
-  auto cross_ref_irs() -> ValueStore<CrossRefIRId>& { return cross_ref_irs_; }
-  auto cross_ref_irs() const -> const ValueStore<CrossRefIRId>& {
-    return cross_ref_irs_;
+  auto import_irs() -> ValueStore<ImportIRId>& { return import_irs_; }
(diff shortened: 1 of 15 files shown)


toolchain/sem_ir/file.h
```diff
@@ -339,10 +339,9 @@ class File : public Printable<File> {
   // Storage for interfaces.
   ValueStore<InterfaceId> interfaces_;
 
-  // Related IRs. There will always be at least 2 entries, the builtin IR (used
-  // for references of builtins) followed by the current IR (used for references
-  // crossing instruction blocks).
-  ValueStore<CrossRefIRId> cross_ref_irs_;
+  // Related IRs. There will always be at least one entry, the builtin IR (used
+  // for references of builtins).
```
jonmeow, line 343: Note, this is just a fix: the 2nd entry is an old comment.

zygoloid (approved):

PR #3663, opened 2024-01-30
Implement some basic ImportRef handling for builtins.

jonmeow:
I'm basically just nudging down the path I think is right here. Adding a small bit more support, but also more tests to capture cases that I think will need to be verified as working.

With diagnostics like "Value of type `<function>` is not callable.", that's because it expects a FunctionDecl but is instead finding a ImportRefUsed. I'll need to work out the necessary support for a callable function.

### DIFF
--- a/toolchain/check/context.cpp
+++ b/toolchain/check/context.cpp
@@ -15,6 +15,7 @@
 #include "toolchain/check/inst_block_stack.h"
 #include "toolchain/lex/tokenized_buffer.h"
 #include "toolchain/parse/node_kind.h"
+#include "toolchain/sem_ir/builtin_kind.h"
 #include "toolchain/sem_ir/file.h"
 #include "toolchain/sem_ir/ids.h"
 #include "toolchain/sem_ir/inst.h"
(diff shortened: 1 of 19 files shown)


toolchain/check/context.cpp
```diff
@@ -227,22 +228,35 @@ auto Context::ResolveIfImportRefUnused(SemIR::InstId inst_id) -> void {
...
+        {SemIR::ImportRefUsed{import_inst.type_id(), unused_inst->ir_id,
+                              unused_inst->inst_id}});
+    return;
   }
+
+  auto import_type_inst_id = import_ir.types().GetInstId(import_inst.type_id());
+  CARBON_CHECK(import_type_inst_id.is_valid());
+
+  // If the type of the instruction is a builtin, use it directly.
+  auto type_id = SemIR::TypeId::Invalid;
```
chandlerc, line 245: I'd avoid the dead initialization. Sanitizers and static analysis tools seem useful to catch bugs here.

```suggestion
  SemIR::TypeId type_id;
```

chandlerc (approved):
Makes sense as an incremental step.

toolchain/check/context.cpp
jonmeow: So far we haven't been supporting default construction of ID types, so the suggestion doesn't compile. We've been relying more on preventing accidental default construction, since use-cases seem relatively rare (we have about a dozen `Invalid` inits right now). Just enabling default construction doesn't seem to catch uninit bugs. Maybe #toolchain is a good place to discuss if you'd like to add default construction?

PR #3666, opened 2024-01-30
Use concepts and better comparisons from C++20 in `IndexBase`.

chandlerc:
This is one of the nicer improvements to boiler-plate as we benefit from concepts simplifying the constraints on the templates, defaulted `!=` finding `==`, and the `<=>` operator for a single definition of all relational operators.

I've checked and the `!=` usages do seem to be effective at finding the `==` definitions, but we still need both the RHS- and LHS-based overloads of `==` as far as I can tell.

### DIFF
--- a/toolchain/base/index_base.h
+++ b/toolchain/base/index_base.h
@@ -53,8 +53,9 @@ struct IndexBase : public IdBase {
   using IdBase::IdBase;
 };
 
-// Support equality comparison when one operand is a child of IdBase (including IndexBase) and the other operand is either the same type or convertible. Inequality
-// automatically uses these in C++20.
+// Support equality comparison when one operand is a child of `IdBase`
+// (including `IndexBase`) and the other operand is either the same type or
+// convertible to that type.
 template <typename IndexType>
   requires std::derived_from<IndexType, IdBase>
 auto operator==(IndexType lhs, IndexType rhs) -> bool {
@@ -66,14 +67,8 @@ template <typename IndexType, typename RHSType>
 auto operator==(IndexType lhs, RHSType rhs) -> bool {
   return lhs.index == IndexType(rhs).index;
 }
-template <typename LHSType, typename IndexType>
-  requires std::derived_from<IndexType, IdBase> &&
-           std::convertible_to<LHSType, IndexType>
-auto operator==(LHSType lhs, IndexType rhs) -> bool {
-  return IndexType(lhs).index == rhs.index;
-}
 
-// Relational comparisons for only IndexBase.
+// Relational comparisons are only supported for types derived from `IndexBase`.
 template <typename IndexType>
   requires std::derived_from<IndexType, IndexBase>
 auto operator<=>(IndexType lhs, IndexType rhs) -> std::strong_ordering {
--- a/toolchain/base/index_base.h
+++ b/toolchain/base/index_base.h
@@ -5,7 +5,8 @@
 #ifndef CARBON_TOOLCHAIN_BASE_INDEX_BASE_H_
 #define CARBON_TOOLCHAIN_BASE_INDEX_BASE_H_
 
-#include <type_traits>
+#include <compare>
+#include <concepts>
 
 #include "common/ostream.h"
 #include "llvm/ADT/DenseMapInfo.h"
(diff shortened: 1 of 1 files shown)


toolchain/base/index_base.h
```diff
@@ -52,78 +53,32 @@ struct IndexBase : public IdBase {
   using IdBase::IdBase;
 };
 
-// Equality comparison for both IdBase and IndexBase.
-template <
-    typename IndexType,
-    typename std::enable_if_t<std::is_base_of_v<IdBase, IndexType>>* = nullptr>
+// Equality comparison for both IdBase and IndexBase, including the index type
+// on either the LHS or RHS with something convertible to it. Inequality
+// automatically uses these in C++20.
```
jonmeow: Is this comment about it being a C++20 feature something you find useful, or could we assume developers will assume != and == will share implementations going forward?

toolchain/base/index_base.h
```diff
@@ -52,78 +53,32 @@ struct IndexBase : public IdBase {
   using IdBase::IdBase;
 };
 
-// Equality comparison for both IdBase and IndexBase.
-template <
-    typename IndexType,
-    typename std::enable_if_t<std::is_base_of_v<IdBase, IndexType>>* = nullptr>
+// Equality comparison for both IdBase and IndexBase, including the index type
+// on either the LHS or RHS with something convertible to it. Inequality
```
jonmeow: As you're adding "including the index type on either the LHS or RHS with something convertible to it", I'm having trouble understanding what it's trying to say. On my first reading, index's type is `int32_t`, and "it" means the same, so this is saying "including int32_t with something convertible to int32_t". Would something like the below work for you?

```suggestion
// Support equality comparison when one operand is a child of IdBase (including IndexBase) and the other operand is either the same type or convertible. Inequality
```

Note though, I'm also wondering if we really need the "convertible" half of this -- #3670

jonmeow (approved):
LGTM, just a couple comments

toolchain/base/index_base.h
zygoloid: I think this comment will not age well -- once we're used to the C++20 rules, this will not be a useful thing to say. I think we should remove it.

toolchain/base/index_base.h
```diff
@@ -52,78 +53,32 @@ struct IndexBase : public IdBase {
...
-}
-
-template <
-    typename IndexType, typename RHSType,
-    typename std::enable_if_t<std::is_base_of_v<IdBase, IndexType>>* = nullptr,
-    typename std::enable_if_t<std::is_convertible_v<RHSType, IndexType>>* =
-        nullptr>
+template <typename IndexType, typename RHSType>
+  requires std::derived_from<IndexType, IdBase> &&
+           std::convertible_to<RHSType, IndexType>
```
zygoloid, line 66: Not arguing for a change here, just a data point for the style rule: I think type-constraints in the template parameter list would be significantly easier to read than this requires-clause. `template <std::derived_from<IdBase> IndexType, std::convertible_to<IndexType> LHSType>` is clearly looking for one type that is derived from `IdBase` and another type that is convertible to that first type, whereas this formulation makes it pretty unclear which type we expect to be derived from which and which type we expect to be convertible to which.

The standard library concepts are named in such a way that use in a template parameter list is readable, which I think will tend to make use in a requires-clause confusing. This isn't a new problem -- `is_base_of` and `is_convertible` have the same lack of clarity, and there are some especially bad examples like `is_convertible` and `is_constructible` making opposite choices (they model "is convertible *to*" versus "is constructible *from*") -- but use of a type constraint would solve this problem.

toolchain/base/index_base.h
```diff
@@ -52,78 +53,32 @@ struct IndexBase : public IdBase {
...
-    typename LHSType, typename IndexType,
-    typename std::enable_if_t<std::is_base_of_v<IdBase, IndexType>>* = nullptr,
-    typename std::enable_if_t<std::is_convertible_v<LHSType, IndexType>>* =
-        nullptr>
+template <typename LHSType, typename IndexType>
+  requires std::derived_from<IndexType, IdBase> &&
+           std::convertible_to<LHSType, IndexType>
 auto operator==(LHSType lhs, IndexType rhs) -> bool {
   return IndexType(lhs).index == rhs.index;
 }
```
zygoloid:
```suggestion
```

We shouldn't need this; the compiler will automatically rewrite `a == b` into `b == a` if needed.

chandlerc: I thought so as well, tried removing it, and got failures. 🤷

jonmeow: I tried removing each separately, and it seems to still compile. Per chat with zygoloid, if you have the matching `!=` it won't (assuming you meant to use an older approach, so it won't rewrite). Maybe you checked before removing `!=`?

toolchain/base/index_base.h
chandlerc: FWIW, I found it suprising that this aspect of C++20 worked this way, which is why I wrote it. But I'm happy to remove.

toolchain/base/index_base.h
chandlerc: There was some discussion of this, but I think no consensus to move away from the style guidance here as it adds two different ways these need to be read with (to me at least) surprisingly similar syntax.

Jon also taught me a helpful way of reading these -- move the name inbetween the two arguments -> `std::derived_from<A, B>` means `A` is `derived_from` `B`. This is apparently a common idiom for reading and designing these kinds of APIs in Python.

toolchain/base/index_base.h
chandlerc: Removed.

chandlerc (commented):
Should all be addressed, PTAL?

zygoloid (approved):

PR #3668, opened 2024-01-30
Switch remaining relational operator overloads to spaceship.

chandlerc:
This doesn't matter too much as these were expanded by the LLVM iterator
facade, but eventually this should enable that facade to be a little
less fancy and should also simplify the dispatch to directly use the
three-way comparison.

One case is a bit subtle and didn't have a comment so I added one to
explain a bit what is going on there.


jonmeow (approved):
Approving, not merging given the stack.

PR #3670, opened 2024-01-30
Remove is_convertible comparisons for IndexBase

jonmeow:
These convertible checks are only used in one test, so that index types can be compared with NodeId. We aren't doing this otherwise: can we remove them?

### DIFF
--- a/toolchain/base/index_base.h
+++ b/toolchain/base/index_base.h
@@ -53,19 +53,12 @@ struct IndexBase : public IdBase {
   using IdBase::IdBase;
 };
 
-// Support equality comparison when one operand is a child of `IdBase`
-// (including `IndexBase`) and the other operand is either the same type or
-// convertible to that type.
+// Support equality comparison for types derived from `IdBase` (including
+// `IndexBase`).
 template <typename IndexType>
   requires std::derived_from<IndexType, IdBase>
 auto operator==(IndexType lhs, IndexType rhs) -> bool {
   return lhs.index == rhs.index;
-}
-template <typename IndexType, typename RHSType>
-  requires std::derived_from<IndexType, IdBase> &&
-           std::convertible_to<RHSType, IndexType>
-auto operator==(IndexType lhs, RHSType rhs) -> bool {
-  return lhs.index == IndexType(rhs).index;
 }
 
 // Relational comparisons are only supported for types derived from `IndexBase`.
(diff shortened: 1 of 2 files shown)


jonmeow:
Note, if you don't want to do this so that `AnyFunctionId == FunctionId` works, WDYT of at least ensuring both operands are IndexBase types to discourage int32_t use (as in the test)? Could it be restricted to the Any* NodeId types, or are there other planned uses of it?

zygoloid:
These are not permitting comparisons with `int32_t`; they're only permitting comparisons between a type that is derived from `IdBase` (eg, `FunctionId`) and a type that implicitly converts to that type (eg, `AnyFunctionId`). `int32_t` does not implicitly convert to `FunctionId` -- the conversion is explicit -- so comparisons of `Id`s / `Index`es against `int32_t` should already not be permitted here.

Note that in the test, `n` is a `NodeId`, not an `int32_t`.

chandlerc:
For future reference, it'd be easier for me to follow if we kept the discussion on the original PR thread and only forked off a new PR once we have a conclusion.

jonmeow:
> These are not permitting comparisons with `int32_t`; they're only permitting comparisons between a type that is derived from `IdBase` (eg, `FunctionId`) and a type that implicitly converts to that type (eg, `AnyFunctionId`). `int32_t` does not implicitly convert to `FunctionId` -- the conversion is explicit -- so comparisons of `Id`s / `Index`es against `int32_t` should already not be permitted here.
> 
> Note that in the test, `n` is a `NodeId`, not an `int32_t`.

Thanks for the correction. But to the extent that this is _only_ used in a unit test, is it necessary to continue to support?

jonmeow:
> For future reference, it'd be easier for me to follow if we kept the discussion on the original PR thread and only forked off a new PR once we have a conclusion.

Sorry, I hadn't wanted to block your PR on a question that seemed peripheral. I'll keep in mind your preference for the future.

chandlerc:
> > These are not permitting comparisons with `int32_t`; they're only permitting comparisons between a type that is derived from `IdBase` (eg, `FunctionId`) and a type that implicitly converts to that type (eg, `AnyFunctionId`). `int32_t` does not implicitly convert to `FunctionId` -- the conversion is explicit -- so comparisons of `Id`s / `Index`es against `int32_t` should already not be permitted here.
> > Note that in the test, `n` is a `NodeId`, not an `int32_t`.
> 
> Thanks for the correction. But to the extent that this is _only_ used in a unit test, is it necessary to continue to support?

(Mostly posting this to see if my understanding is correct -- good for @zygoloid to confirm here.)

I _think_ that this reflects the typed nodes not having all that many users yet, and the unittest reflects a specific intended API for them?

jonmeow:
Closing since this PR isn't making any progress.

PR #3671, opened 2024-01-31
Keep parameters in scope throughout the entity that they parameterize.

zygoloid:
Previously, we created scopes for implicit parameter lists and tuple patterns, but that meant that bindings went out of scope too soon. We now keep them in scope until the end of the enclosing declaration. This is accomplished by pushing a scope for parameters when we handle a name that might have them, and then popping the scope again if it turns out that there were no parameters.

For a case such as:

```carbon
fn A(T:! type).B(U:! type).F(x: T, y: U) {
  var z: T;
}
```

... we now have the following scopes in the stack:

-   A parameter scope containing `T`.
-   A class scope for `A(T:! type)`.
-   A parameter scope containing `U`.
-   A class scope for `A(T:! type).B(U:! type)`.
-   A parameter scope containing `x: T` and `y: U`.
-   A function body scope containing `z: T`.

The innermost scope when check processes a declaration of a function, class, or similar is now often a parameter scope rather than the enclosing scope in which the class or function is declared, so the target scope is now passed explicitly into the modifier checking code that wants to inspect that enclosing scope.


toolchain/check/testdata/class/fail_todo_generic_method.carbon
chandlerc, line 1: s/method/member_function/ ? (Or make it a method?)

toolchain/check/testdata/let/fail_duplicate_decl.carbon
```diff
@@ -4,13 +4,14 @@
...
   // CHECK:STDERR: fail_duplicate_decl.carbon:[[@LINE+6]]:7: ERROR: Duplicate name being declared in the same scope.
   // CHECK:STDERR:   let a: i32 = 1;
   // CHECK:STDERR:       ^
-  // CHECK:STDERR: fail_duplicate_decl.carbon:[[@LINE-4]]:6: Name is previously declared here.
-  // CHECK:STDERR: fn F(a: i32) {
-  // CHECK:STDERR:      ^
+  // CHECK:STDERR: fail_duplicate_decl.carbon:[[@LINE-4]]:7: Name is previously declared here.
+  // CHECK:STDERR:   let a: i32 = 1;
+  // CHECK:STDERR:       ^
   let a: i32 = 1;
```
chandlerc: Maybe make it a touch easier to tell apart?

```suggestion
  let a: i32 = 2;
```

chandlerc (approved):

PR #3672, opened 2024-01-31
Scaffolding for checking impls.

zygoloid:
Consume the components of the `impl` declaration, and set up scopes for the child elements. We don't yet build a representation for the impl itself.

Also, add an interface type value. This is necessary so that we have a value for the expression on the right-hand side of `as` in an `impl`.


toolchain/check/handle_impl.cpp
```diff
@@ -8,31 +8,82 @@ namespace Carbon::Check {
...
+
+  // Push the bracketing node.
+  context.node_stack().Push(parse_node);
+
+  // Optional modifiers follow.
+  context.decl_state_stack().Push(DeclState::Impl);
+
+  // Create a scope for implicit parameters. We may not use it, but it's simpler
+  // to create it unconditionally than to track whether it exists.
+  context.PushScope();
```
chandlerc, line 23: Not for this PR, but a thought: would it make sense to tag or type scopes with *which* scope it is? Specifically both in pushing adn popping.

That might both make the code a bit more self-documenting, and it might also provide some nice checking that the *right* scope is found, given that we have separate code doing the push and pop (rather than something more RAII to make it constructively correct).

chandlerc (approved):
LG

PR #3673, opened 2024-01-31
Remove `operator!=` now that C++20 lets us use `operator==`.

chandlerc:
Also switches to defaulting `operator==` where we can, and adds a CRTP helper to let us do this even when using CRTP classes. It turns out to be quite surprising what is needed to allow a CRTP base class to not interfere in the defaulting of comparison functions, and so I've extracted it into its own helper and tried to document it. We don't have a lot of users yet, but it should ensure things like `Printable` continue to work without interfering with other patterns.

### DIFF
--- a/common/crtp_helpers.h
+++ b/common/crtp_helpers.h
@@ -1,0 +1,71 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#ifndef CARBON_COMMON_CRTP_HELPERS_H_
+#define CARBON_COMMON_CRTP_HELPERS_H_
+
+#include <compare>
+#include <concepts>
+
+namespace Carbon {
+
+// Helper for building correct empty or stateless [CRTP] base classes.
+//
+// [CRTP]: https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern
+//
+// C++ conflates _compositional_ type authoring with inheritance. An effective
+// pattern for doing compositional type design in C++ are CRTP base classes.
+// However, because they are ultimately base classes, they have to be specially
+// crafted in order to not disrupt useful C++ features such as defaulted
+// comparisons.
+//
+// For an empty or stateless CRTP base class, we provide this helper that should
+// itself follow the CRTP pattern when building a CRTP base class:
+//
+// ```cpp
+// template <typename DerivedT>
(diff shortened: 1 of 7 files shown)


chandlerc:
As mentioned in Discord, not at all sure about this one. Especially the infrastructure to support defaulting `operator==`. =/

common/crtp_helpers.h
```diff
@@ -0,0 +1,71 @@
...
+  // defaulted operator in the derived type.
+  //
+  // The parameters are also accepted by value rather than by const reference
+  // because this base is required to be empty. We don't want to force an
+  // address to exist for it, and empty types have special low-cost
+  // implementations in the ABI.
+  template <typename LeftT, typename RightT>
+    requires std::same_as<LeftT, CRTPBaseT> && std::same_as<RightT, CRTPBaseT>
+  friend constexpr auto operator==(LeftT /*lhs*/, RightT /*rhs*/) -> bool {
+    static_assert(std::is_empty_v<CRTPBaseT>,
```
zygoloid, line 52: Would it make sense to move this assert out of this function and into the class?

zygoloid (commented):
I'm not sure whether this is worth it as-is, but I suppose we'll get more benefit from this as the codebase grows, so this is probably the right thing to do. Would appreciate other opinions.

common/crtp_helpers.h
chandlerc: Yeah, that was my original idea. But the class has to be completed while the derived class is still incomplete, so I didn't see a way to make it work. It needs to be in an instantiated method I think, but one that definitely is instantiated and doesn't interfere with special member functions. I couldn't think of a better way than duplicating it into the two comparison functions, but if you have ideas, lemme know!

zygoloid: Oh, of course. Yeah, this seems like the best we can do.

common/crtp_helpers.h
```diff
@@ -0,0 +1,71 @@
...
+// ```cpp
+// template <typename DerivedT>
+// class ComposingWidget : EmptyCRTPBase<ComposingWidget<DerivedT>> {
+//   ...
+// };
+// ```
+//
+// This will provide the necessary comparison functions to allow `DerivedT` to
+// default its comparison functions (or not) as desired. These will be stateless
+// comparison functions, and this helper is only valid when `ComposableWidget`
```
geoffromer, line 35:
```suggestion
// comparison functions, and this helper is only valid when `ComposingWidget`
```

chandlerc:
Based on discussions, dropping this.

PR #3674, opened 2024-01-31
Implement ConstantId::Invalid

jonmeow:
Right now, ConstantValueStore defaults to having unknown values use NotConstant. This generally works for the current IR, but with imports we're expecting sparse entries which are generally unknown -- and distinguishing between NotConstant and simply unset would be helpful. As a consequence, add Invalid.

We discussed whether to simply have ConstantValueStore default to Invalid going forward, or to make the default flexible. The upside to the former is consistency, the upside to the latter is that it should result in fewer Sets when operating on the current IR (which will more frequently have known non-constant values). This PR offers both approaches in separate commits, but I somewhat lean towards the latter for fewer array resizes.

Note, a totally different approach would be to use a different class (not ConstantValueStore) for imported IRs -- then the default of Invalid versus NotConstant would be type-dependent. However, I expect we're going to want to do at least somewhat consistent lookups, and using the same ConstantValueStore for both cases allows avoiding a virtual interface or templating. Also, I'm hoping to only maintain the ConstantValueStore for an imported IR as part of Context (not File), which would mean the SmallVector storage overhead is ephemeral, mitigating one of the potential advantages of using a different type for imported IRs.

### DIFF
--- a/toolchain/sem_ir/formatter.cpp
+++ b/toolchain/sem_ir/formatter.cpp
@@ -788,6 +788,15 @@ class Formatter {
     out_ << "\n";
   }
 
+  // Don't print a constant for ImportRefUnused.
+  auto FormatInstruction(InstId inst_id, ImportRefUnused inst) -> void {
+    Indent();
+    FormatInstructionLHS(inst_id, inst);
+    out_ << ImportRefUnused::Kind.ir_name();
+    FormatInstructionRHS(inst);
+    out_ << "\n";
+  }
+
   auto FormatInstructionLHS(InstId inst_id, Inst inst) -> void {
     switch (inst.kind().value_kind()) {
       case InstValueKind::Typed:
--- a/toolchain/sem_ir/formatter.cpp
+++ b/toolchain/sem_ir/formatter.cpp
@@ -772,11 +772,16 @@ class Formatter {
     out_ << InstT::Kind.ir_name();
     FormatInstructionRHS(inst);
     if (auto const_id = sem_ir_.constant_values().Get(inst_id);
-        const_id.is_valid() && const_id.is_constant()) {
-      out_ << (const_id.is_symbolic() ? " [symbolic" : " [template");
-      if (const_id.inst_id() != inst_id) {
-        out_ << " = ";
-        FormatInstName(const_id.inst_id());
+        !const_id.is_valid() || const_id.is_constant()) {
+      out_ << " [";
+      if (const_id.is_valid()) {
+        out_ << (const_id.is_symbolic() ? "symbolic" : "template");
+        if (const_id.inst_id() != inst_id) {
+          out_ << " = ";
+          FormatInstName(const_id.inst_id());
+        }
+      } else {
+        out_ << const_id;
       }
       out_ << "]";
     }
--- a/toolchain/sem_ir/formatter.cpp
+++ b/toolchain/sem_ir/formatter.cpp
@@ -772,7 +772,7 @@ class Formatter {
     out_ << InstT::Kind.ir_name();
     FormatInstructionRHS(inst);
     if (auto const_id = sem_ir_.constant_values().Get(inst_id);
-        const_id.is_constant()) {
+        const_id.is_valid() && const_id.is_constant()) {
       out_ << (const_id.is_symbolic() ? " [symbolic" : " [template");
       if (const_id.inst_id() != inst_id) {
         out_ << " = ";
--- a/toolchain/sem_ir/ids.h
+++ b/toolchain/sem_ir/ids.h
@@ -94,12 +94,14 @@ struct ConstantId : public IdBase, public Printable<ConstantId> {
   // An ID for an expression whose phase cannot be determined because it
   // contains an error. This is always modeled as a template constant.
   static const ConstantId Error;
+  // An explicitly invalid ID.
+  static const ConstantId Invalid;
 
   // Returns the constant ID corresponding to a template constant, which should
   // either be in the `constants` block in the file or should be known to be
   // unique.
   static constexpr auto ForTemplateConstant(InstId const_id) -> ConstantId {
-    return ConstantId(const_id.index + 1);
+    return ConstantId(const_id.index + IndexOffset);
   }
 
   // Returns the constant ID corresponding to a symbolic constant, which should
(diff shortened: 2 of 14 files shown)


toolchain/sem_ir/formatter.cpp
```diff
@@ -772,7 +772,7 @@ class Formatter {
     out_ << InstT::Kind.ir_name();
     FormatInstructionRHS(inst);
     if (auto const_id = sem_ir_.constant_values().Get(inst_id);
-        const_id.is_constant()) {
+        const_id.is_valid() && const_id.is_constant()) {
```
zygoloid: I wonder if it's worth adding an `[<invalid>]` tag or similar to the invalid case, especially if we don't expect that to happen.

toolchain/sem_ir/ids.h
```diff
@@ -137,13 +149,22 @@ struct ConstantId : public IdBase, public Printable<ConstantId> {
   }
 
  private:
-  // ConstantIds don't have an invalid state.
-  using IdBase::is_valid;
+  // The offset of InstId indices to ConstantId indices.
+  static constexpr int32_t IndexOffset = 2;
+  static constexpr int32_t NotConstantIndex = IndexOffset + InvalidIndex;
```
zygoloid: It's not obvious to me what the pattern is here (eg, how this would be extended if we added more special indexes). I can see that this is right because I know that `InvalidIndex == -1`, but if we had a second special value or a third, I'm not clear on what calculation we'd use.

Maybe it'd be more consistent to use all negative values for special indices, like other `Id` types:

```suggestion
  static constexpr int32_t NotConstantIndex = InvalidIndex - 1;
  // The offset of InstId indices to ConstantId indices.
  static constexpr int32_t IndexOffset = -NotConstantIndex + 1;
```

(I think keeping the offset for both the positive and negative cases is probably still a good thing, so that we quickly catch cases where an `InstId` is confused for a `ConstantId`, but I'd be OK with having the offset only for the negative cases if you'd prefer.)

toolchain/sem_ir/ids.h
```diff
@@ -94,40 +94,52 @@ struct ConstantId : public IdBase, public Printable<ConstantId> {
...
   static constexpr auto ForTemplateConstant(InstId const_id) -> ConstantId {
-    return ConstantId(const_id.index + 1);
+    return ConstantId(const_id.index + IndexOffset);
   }
 
   // Returns the constant ID corresponding to a symbolic constant, which should
   // either be in the `constants` block in the file or should be known to be
   // unique.
   static constexpr auto ForSymbolicConstant(InstId const_id) -> ConstantId {
     // Avoid allocating index -1.
```
zygoloid: I think this comment is redundant now. (And it looks like it was wrong before, oops!)

toolchain/sem_ir/value_stores.h
```diff
@@ -119,12 +119,15 @@ class InstStore {
 // instructions.
 class ConstantValueStore {
  public:
-  // Returns the constant value of the requested instruction, or
-  // `ConstantId::NotConstant` if it is not constant.
+  explicit ConstantValueStore(ConstantId default_value)
+      : default_(default_value) {}
+
+  // Returns the constant value of the requested instruction, or default_ if it
+  // is not constant.
```
zygoloid:
```suggestion
  // has not been set.
```

toolchain/sem_ir/ids.h
jonmeow: I was trying to have the `inst_id` calculation trivial, are you intending that there should be control flow? i.e., instead of:

```
return InstId(std::abs(index) - IndexOffset);
```

(where `std::abs` might be a `neg` instruction)

We have multiple branches, such as:

```
if (index >= 0) {
  return InstId(index);
} else if (!is_valid() || *this == NotConstant) {
  return InstId::Invalid;
} else {
  return InstId(-index - IndexOffset);
}
```

zygoloid: No, I was intending that the `inst_id` calculation would remain the same -- we would keep the offset for both the positive and negative cases.

jonmeow: Okay, but the `inst_id` calculation still needs to change to handle invalid _somehow_. Assuming you're wanting this, I've added a `CARBON_CHECK` that it's only used for valid constants.

toolchain/sem_ir/ids.h
jonmeow: Removing.

toolchain/sem_ir/value_stores.h
jonmeow: Done

toolchain/sem_ir/formatter.cpp
jonmeow: Done

toolchain/sem_ir/formatter.cpp
```diff
@@ -772,11 +772,16 @@ class Formatter {
...
+        !const_id.is_valid() || const_id.is_constant()) {
+      out_ << " [";
+      if (const_id.is_valid()) {
+        out_ << (const_id.is_symbolic() ? "symbolic" : "template");
+        if (const_id.inst_id() != inst_id) {
+          out_ << " = ";
+          FormatInstName(const_id.inst_id());
+        }
+      } else {
+        out_ << const_id;
```
zygoloid, line 784:
```suggestion
        out_ << "unknown";
```
Looking at how this appears for `import_ref`s, I think this might be clearer.

toolchain/sem_ir/ids.h
```diff
@@ -137,13 +149,20 @@ struct ConstantId : public IdBase, public Printable<ConstantId> {
...
-  // ConstantIds don't have an invalid state.
-  using IdBase::is_valid;
+  static constexpr int32_t NotConstantIndex = InvalidIndex - 1;
+  // The offset of InstId indices to ConstantId indices.
+  static constexpr int32_t IndexOffset = -NotConstantIndex + 1;
+
+  // Double-check inst_id calculations work. C++23 makes std::abs constexpr, but
+  // until then we mirror std::abs logic here.
+  static_assert(NotConstantIndex < 0 &&
+                -NotConstantIndex - IndexOffset == InstId::InvalidIndex);
```
zygoloid: Maybe we could change `inst_id` to compute `abs(index)` longhand so we can make it `constexpr`? LLVM [optimizes that to the same thing](https://godbolt.org/z/aT1bWPWj3), and it avoids a function call at `-O0`.

It'd definitely be nice to be able to rewrite this as
```
static_assert(!ConstantId::NotConstant.inst_id().is_valid());
```
or `<...>inst_id() == InstId::Invalid` or something like that.

toolchain/sem_ir/ids.h
```diff
@@ -94,40 +94,52 @@ struct ConstantId : public IdBase, public Printable<ConstantId> {
...
-  auto is_template() const -> bool { return index > 0; }
+  // Returns whether this represents a constant. Requires is_valid.
+  auto is_constant() const -> bool {
+    CARBON_CHECK(is_valid());
+    return *this != ConstantId::NotConstant;
+  }
+  // Returns whether this represents a symbolic constant. Requires is_valid.
+  auto is_symbolic() const -> bool {
+    CARBON_CHECK(is_valid());
+    return index <= IndexOffset;
```
zygoloid:
```suggestion
    return index <= -IndexOffset;
```

toolchain/sem_ir/ids.h

toolchain/sem_ir/formatter.cpp
jonmeow: I think the problem is that putting anything here for an ImportRefUnused is unclear: unused imports _never_ have a constant. How about this approach which special-cases ImportRefUnused to not print the constant?

(of course, now ImportRefUnused is very special-cased, but it feels like this should mirror ImportRefUsed where possible)

zygoloid (approved):

PR #2198, opened 2022-09-20
Fix LLVM compilation issues

jonmeow:
We've been having issues with asan builds on linux. This change should fix all of that. A build run can be found at:
https://github.com/carbon-language/carbon-lang/actions/runs/3093378863/jobs/5005683059

(currently in progress, but I'm expecting it to succeed at this point)

It may be that the issues with asan builds were actually related to caching. That is, maybe the brew build command didn't change enough between v14 and v15 that the cache hits were still an issue. We did notice this with 15.0.0 versus 15.0.1 include paths (that is, bazel wasn't happy using the cached results of a 15.0.0 build due to the skew in include paths). In order to address this, I've added CACHE_VERSION to the remote_cache setup. I've also set up corresponding buckets in Cloud.

However, I'm also switching Linux to llvm-15 and apt. I'd originally been looking at this because the issues were linux-specific, and we've previously had linux-specific issues with Homebrew. Although it may have been the cache all along, I would prefer to keep this setup (if nothing else, it made the caching issues more obvious, even though we were still confused by the include path manifestation).

PR #3677, opened 2024-02-01
Switch our main macOS actions to Arm.

chandlerc:
GitHub has publicly available Arm macOS runners now: https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

These runners are used when the OS label is `macos-14` instead of `macos-12`.

This PR switches the main PR and merge queue macOS runs to that OS. It adds a push-only `macos-12` run to check post-merge whether Intel macOS keeps working as we don't want that to break too badly. Given how much faster the M1 is, it didn't seem worth it to keep this in the critical path of development. And the Intel path here is more likely than others to diminish in importance over time.

All our OS predicates in the action encoded a specific OS version which no longer works given two macOS versions, so I've switch all the OS predicates to a generic pattern that doesn't encode a version.

Example run on a PR: https://github.com/carbon-language/carbon-lang/actions/runs/7766629200?pr=3686
Example run on push: https://github.com/carbon-language/carbon-lang/actions/runs/7766480822


.github/workflows/tests.yaml
```diff
@@ -23,19 +23,26 @@ jobs:
...
+        os: [ubuntu-22.04, macos-14]
         build_mode: [fastbuild, opt]
-        # The clang-tidy config doesn't work on macos (missing `truncate`).
         include:
+          # The clang-tidy config doesn't work on macos (missing `truncate`).
           - os: 'ubuntu-22.04'
             build_mode: 'clang-tidy'
+          # On push (post-submit), also run on `macos-12` to get Intel macOS
+          # coverage. This is less and less critical so we skip it on the PR and
+          # merge queue path by switching to a redundant OS.
```
jonmeow: Just to be sure, is this still important enough that it's worth the additional costs involved?

.github/workflows/tests.yaml
```diff
@@ -23,19 +23,26 @@ jobs:
...
+          - os: ${{ github.event_name == 'push' && 'macos-12' || 'macos-14' }}
```
jonmeow: I'm trusting this works, but had you considered on line 26 instead writing:

```
- os: ${{ github.event_name == 'push' && [ubuntu-22.04, macos-14, macos-12] || [ubuntu-22.04, macos-14] }}
```

(if this works, I think it'd reduce repetition)

.github/workflows/tests.yaml
```diff
@@ -74,33 +81,43 @@ jobs:
         if: steps.filter.outputs.has_code == 'true'
         with:
           # Match the min version listed in docs/project/contribution_tools.md
-          python-version: '3.9'
+          # or the oldest version available on the OS.
+          python-version: ${{ matrix.os == 'macos-14' && '3.11' || '3.9' }}
```
jonmeow: Had you considered updating the min version? I don't think we've changed for a couple years now, maybe updating it to 3.11 everywhere would be fine?

.github/workflows/tests.yaml
```diff
@@ -23,19 +23,26 @@ jobs:
...
+          - os: ${{ github.event_name == 'push' && 'macos-12' || 'macos-14' }}
+            build_mode: opt
     runs-on: ${{ matrix.os }}
 
     steps:
       # Ubuntu images start with 23GB available, and this adds 14GB more. For
       # comparison, MacOS images have >100GB free.
       - name: Free up disk space (Ubuntu)
-        if: matrix.os == 'ubuntu-22.04'
+        if: contains(matrix.os, 'ubuntu')
```
jonmeow: Why contains instead of startsWith?

.github/workflows/tests.yaml
```diff
@@ -74,33 +81,43 @@ jobs:
...
+          'macos')
         id: cache-homebrew-macos
         uses: actions/cache@v3
         env:
           cache-name: cache-homebrew
         with:
-          # Cover all the critical parts of Homebrew here.
+          # Cover all the critical parts of Homebrew here. Homebrew on Arm macOS
+          # uses its own prefix making this easy to cover, but we need a few
+          # different paths for Intel.
```
jonmeow, line 98: Just to be sure, this is arm rather than the macos-14 setup?

.github/workflows/tests.yaml
```diff
@@ -227,12 +248,15 @@ jobs:
 
       # We need to replace the `.` with a `_` for the build cache.
       - name: Setup LLVM and Clang (macOS)
-        if: steps.filter.outputs.has_code == 'true' && matrix.os == 'macos-12'
+        if:
+          steps.filter.outputs.has_code == 'true' && contains(matrix.os,
+          'macos')
```
jonmeow: Why are you using macos-12 for the macos-14 cache? Will the build outputs all be the same, even with the environment and CPU change?

jonmeow (commented):
Generally looks good, just a few questions about structure.

.github/workflows/tests.yaml
```diff
@@ -227,12 +248,15 @@ jobs:
...
         run: |
           echo "os_for_cache=macos-12" >> $GITHUB_ENV
       - name: Setup LLVM and Clang (Ubuntu)
         if:
-          steps.filter.outputs.has_code == 'true' && matrix.os == 'ubuntu-22.04'
+          steps.filter.outputs.has_code == 'true' && contains(matrix.os,
+          'ubuntu')
```
jonmeow: Similar to the macos-12 comment, are you sure all ubuntu builds are sufficiently hermetic that they will yield the same outputs regardless of version? Note, I think we've seen some corrupt builds from caching of non-hermetic build outputs before -- IIRC that's https://github.com/carbon-language/carbon-lang/pull/2198 and its note about a slight difference in include paths.

jonmeow: Note, maybe if the build hermeticism issue is solved, we could just remove the os suffix now?

.github/workflows/tests.yaml
chandlerc: Happy to try that, but seemed like maybe better to do separately to minimize how much changes with this PR?

.github/workflows/tests.yaml
chandlerc: I think it's nice to keep for now. If nothing else, the fact that so much GitHub action infra is still using Intel macOS seems like a good reason to avoid letting it break just yet.

.github/workflows/tests.yaml
chandlerc: It's a bit harder to do than that because you can't put JSON directly into an expression, but you can parse a string into JSON, so I've done that. It does seem simpler, even if it breaks VSCode's YAML highlighting...

.github/workflows/tests.yaml
chandlerc: It was shorter, but maybe that's not a great reason.

That said, this still was really repetitive. I've extracted a job-level environment context variable `os` to just contain `macos` or `ubuntu`, and renamed the matrix variable to `runner`. This lets us simply use `runner` when we want a *specific* runner image, and `os` when we just need to know whether its macOS or Ubuntu.

PTAL and let me know what you think. Happy to go back or try different approaches here.

.github/workflows/tests.yaml
chandlerc: Yep, all of Arm Homebrew uses this layout.

My guess is they made the change to a parallel root because there was a pretty long period of time where the *x86* Homebrew was in use on Arm machines via Rosetta. This "just worked" for a bunch of use cases, and there were a lot of installs like this. When they got ready to have a native Arm Homebrew, they wanted it to be able to be installed in parallel with the x86 one to avoid breakage, so they moved to their own prefix.

.github/workflows/tests.yaml
chandlerc: No, good catch. I had missed that the version was included here. See below for a full answer.

.github/workflows/tests.yaml
chandlerc: No, they're not. Adding ARM works "fine", but by accident -- the *CPU* part of this is part of the action hash used to query the cache. But the OS versions are not. Nor are the distinctions between Ubuntu and some other Linux distro if we ever care.

But there *is* a great way to solve the hermeticity issues here more simply than partitioning our buckets! We can use the `--action_env` flag to thread a key for the remote cache in from the workflow file, and set that in a way that allows us to use a single bucket.

I've done this to make it more obvious that the two macOS runs are correct, and left a TODO to completely remove this -- that seems like it's worth keeping as a separate change as we'll need to fix up the name of the buckets, the keys, and then do a round of fresh builds to re-populate them.

chandlerc (commented):
Thanks for the great feedback! Addressing it resulted in quite some changes, so PTAL.

One run using the new techniques is here: https://github.com/carbon-language/carbon-lang/actions/runs/7753457932

This had to populate the now-keyed remote caches. I can push some fresh versions to double check that the cache hits work.

.github/workflows/tests.yaml
```diff
@@ -358,6 +381,7 @@ jobs:
 
             bazel_exit=0
             bazelisk test -c ${{ matrix.build_mode }} \
+              --action_env=BAZEL_REMOTE_CACHE_KEY \
```
jonmeow: Can this be added to the bazelrc instead of being passed to individual executions, to ensure it also applies even if `bazel mod deps` or the commands executed as part of diff do cache reads/writes?

.github/workflows/tests.yaml
jonmeow: SGTM

.github/workflows/tests.yaml
jonmeow: Can separate out if you want, but I've created a carbon-builds-github-v1. (just dropping the os suffix)

> that seems like it's worth keeping as a separate change as we'll need to fix up the name of the buckets, the keys, and then do a round of fresh builds to re-populate them.

I think all you'd need to do is change the names of the bucket in this PR. Fresh builds are also necessary with either the macos-14 or the --action_env addition (both require cache population), you'd just be putting builds into a different bucket.

Keys are unaffected since it's still the same account.

jonmeow (approved):

jonmeow:
Feel free to merge when you've had a chance to respond on comments, but I do wonder if action_env should be in the bazelrc. I'm glad you found that though.

.github/workflows/tests.yaml
chandlerc: Doh, yes. Not sure why I didn't think of that, thanks.

.github/workflows/tests.yaml
chandlerc: Ah cool, nice that it's that easy. I've done this and tested that it all works, and we get caching afterward.

chandlerc (commented):
I think this is all ready to go, but I'll merge on Monday in case it creates some problem.

PR #3678, opened 2024-02-01
Parse tree for `impl` that is better for check stage

josh11b:
Use two different nodes for "<type> followed by `as`" and "<type> omitted before `as`, use `self`", so it is easier to determine which case. Later the second case will push the type id for `self` onto the node stack, making the two paths more similar.


toolchain/parse/node_kind.def
```diff
@@ -683,7 +683,14 @@ CARBON_PARSE_NODE_KIND_CHILD_COUNT(ImplIntroducer, 0, Impl)
...
-CARBON_PARSE_NODE_KIND_CHILD_COUNT(ImplAs, 0, As)
+
+// `... as`:
+//   _external_: expression
+// TypeImplAs
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(TypeImplAs, 1, As)
+// `as` without a type before it
+CARBON_PARSE_NODE_KIND_CHILD_COUNT(DefaultSelfImplAs, 0, As)
+
 CARBON_PARSE_NODE_KIND_BRACKET(ImplDefinitionStart, ImplIntroducer,
```
jonmeow: Should these parse nodes be more clearly associated with the comment that talks about them? Particularly as more is in the middle, I also wonder if the usual "external" would be better than "see below"

toolchain/parse/typed_nodes.h
```diff
@@ -859,6 +857,16 @@ struct ImplForall {
   ImplicitParamListId params;
 };
 
+// `as` with no type before it
```
jonmeow, line 860:
```suggestion
// `as` with no type before it.
```

Similarly, this feels more like a sentence with the additional words, so punctuation might make sense.

https://google.github.io/styleguide/cppguide.html#Punctuation,_Spelling_and_Grammar

toolchain/check/handle_impl.cpp
```diff
@@ -32,20 +32,24 @@ auto HandleImplForall(Context& context, Parse::ImplForallId parse_node)
...
     -> bool {
+  auto self_id = context.node_stack().PopExpr();
+  context.node_stack().Push(parse_node, self_id);
+  return true;
+}
+
+auto HandleDefaultSelfImplAs(Context& /*context*/,
+                             Parse::DefaultSelfImplAsId /*parse_node*/)
+    -> bool {
+  // TODO: Determine self_id and push it onto node stack
```
```suggestion
  // TODO: Determine self_id and push it onto node stack.
```

Punctuation: https://google.github.io/styleguide/cppguide.html#Punctuation,_Spelling_and_Grammar

jonmeow (approved):
LGTM. Comments are minor, feel free to merge regardless of whether you make the suggested changes.

toolchain/parse/handle_impl.cpp
```diff
@@ -64,17 +63,16 @@ auto HandleImplAfterForall(Carbon::Parse::Context& context) -> void {
...
 auto HandleImplBeforeAs(Carbon::Parse::Context& context) -> void {
   auto state = context.PopState();
-  if (state.has_error) {
-    context.ReturnErrorOnState();
-    return;
-  }
   if (auto as = context.ConsumeIf(Lex::TokenKind::As)) {
-    context.AddLeafNode(NodeKind::ImplAs, *as);
+    context.AddNode(NodeKind::TypeImplAs, *as, state.subtree_start,
+                    state.has_error);
```
zygoloid, line 68: It looks like we don't have any test coverage for the case where there's a parse error before `as`, where we'd be building a subtree with `has_error == true` here. I *think* this is handling that properly, but it'd be good to have a test case to cover that.

toolchain/parse/typed_nodes.h
josh11b: This is a noun phrase not a sentence, and adding a period would be inconsistent with the parallel comments

toolchain/parse/handle_impl.cpp
josh11b: Added.

toolchain/parse/node_kind.def
josh11b: Done.

PR #3679, opened 2024-02-01
Integrate description of tuple indexing into member access better.

zygoloid:
Improve the exposition of the design changes from #3646 to integrate better into the overall description of member access design.

This fixes the incorrect description of the rules for `->` by instead relying on the general rule that `->` is rewritten to use `*` and `.` before any other processing is done, and generally makes *integer-literal* names be less of a special case.


geoffromer (approved):

geoffromer:
(Not merging in case you wanted a review from both of us)

zygoloid:
@chandlerc I think it'd be useful for a lead other than myself to confirm this isn't inadvertently changing the direction in some important way.

chandlerc (approved):
This looks like a good update to the docs without a direction change. Merging.

PR #3683, opened 2024-02-02
Build Impl entity to represent an impl declaration.

zygoloid:
Collect the contents of an `impl` into a scope, and start doing very basic checking for `impl` declarations and definitions.

This change adds two new `Id` types to the set of type that `NodeStack` supports -- `ImplId` and `NameScopeId`.


toolchain/check/handle_impl.cpp
```diff
@@ -34,34 +38,143 @@ auto HandleImplForall(Context& context, Parse::ImplForallId parse_node)
...
+
+namespace {
+struct ImplDeclResult {
+  SemIR::ImplId impl_id;
+  SemIR::InstId impl_decl_id;
+  SemIR::NameScopeId enclosing_scope_id;
+};
+}  // namespace
+
+static auto BuildImplDecl(Context& context, Parse::AnyImplDeclId parse_node)
```
jonmeow, line 133: Should this have a comment?

toolchain/check/handle_impl.cpp
```diff
@@ -34,34 +38,143 @@ auto HandleImplForall(Context& context, Parse::ImplForallId parse_node)
...
+    context.TODO(parse_node, "extending interface without definition");
+    enclosing_scope.has_error = true;
+    return;
+  }
+
+  enclosing_scope.extended_scopes.push_back(interface.scope_id);
+}
+
+namespace {
+struct ImplDeclResult {
```
jonmeow: Should this have a comment?

toolchain/sem_ir/file.h
```diff
@@ -171,7 +171,44 @@ struct Interface : public Printable<Interface> {
...
+  TypeId constraint_id;
+
+  // The following members are set at the `{` of the impl definition.
+
+  // The definition of the impl. This is a ImplDecl.
+  InstId definition_id = InstId::Invalid;
+  // The enclosing scope of the `impl` definition. Because the same `impl` can
+  // be declared in multiple scopes, the enclosing scope is only known when we
+  // see the definition, unlike for other kinds of entity.
+  NameScopeId enclosing_scope_id = NameScopeId::Invalid;
```
jonmeow: Just to be sure -- while possibly helpful, I'd mainly been adding this for named entities [particularly forward declarations], which I'm not sure applies here since there's no NameId. If there's no name to associate with this, not sure how much utility it'll have. Perhaps scope_id's enclosing_scope_id would work if needed?

toolchain/check/node_stack.h
```diff
@@ -303,49 +264,18 @@ class NodeStack {
...
-    if constexpr (RequiredIdKind == IdKind::NameId) {
-      return back.id<SemIR::NameId>();
-    }
-    if constexpr (RequiredIdKind == IdKind::TypeId) {
-      return back.id<SemIR::TypeId>();
-    }
-    CARBON_FATAL() << "Unpeekable IdKind for parse kind: " << RequiredParseKind
-                   << "; see value in ParseNodeKindToIdKind";
+    RequireParseKind<RequiredParseKind>(stack_.back().parse_node);
+    return Peek<ParseNodeKindToIdKind(RequiredParseKind)>();
```
jonmeow: Minor, but this refactoring looks big enough that it might've been nice as a separate PR, given it looks tangential to the core Impl change.

Could you please describe the node stack changes as part of the PR description?

toolchain/sem_ir/ids.h
```diff
@@ -210,6 +211,22 @@ struct InterfaceId : public IdBase, public Printable<InterfaceId> {
...
+  static const ImplId Invalid;
+
+  using IdBase::IdBase;
+  auto Print(llvm::raw_ostream& out) const -> void {
+    out << "impl";
+    IdBase::Print(out);
+  }
+};
+
+constexpr ImplId ImplId::Invalid = ImplId(ImplId::InvalidIndex);
```
```suggestion
constexpr ImplId ImplId::Invalid = ImplId(InvalidIndex);
```

toolchain/sem_ir/formatter.cpp
```diff
@@ -138,6 +149,14 @@ class InstNamer {
         interface_id.index);
   }
 
+  // Returns the scope index corresponding to an impl.
+  auto GetScopeFor(ImplId impl_id) -> ScopeIndex {
+    return static_cast<ScopeIndex>(
+        static_cast<int32_t>(ScopeIndex::FirstFunction) +
+        sem_ir_.functions().size() + sem_ir_.classes().size() +
+        sem_ir_.interfaces().size() + impl_id.index);
+  }
```
jonmeow: I wonder, would rewriting with constexpr reduce duplication? e.g.:

```
template<typename IdT>
auto GetScopeFor(IdT id) -> ScopeIndex {
  int32_t index = static_cast<int32_t>(ScopeIndex::FirstFunction);
  if constexpr (std::is_same(IdT, FunctionId)) {
    index += fn_id.index;
  } else {
    index += sem_ir_.functions().size();
    if constexpr (std::is_same(IdT, ClassId)) {
       ...
    }
  }
  return static_cast<ScopeIndex>(index);
}
```

Ending nesting with `} else { static_assert(false, "Unsupported IdT"); }`

toolchain/sem_ir/formatter.cpp
```diff
@@ -138,6 +149,14 @@ class InstNamer {
         interface_id.index);
   }
 
+  // Returns the scope index corresponding to an impl.
+  auto GetScopeFor(ImplId impl_id) -> ScopeIndex {
+    return static_cast<ScopeIndex>(
```
jonmeow: Side-comment: would it be worth changing this type name to differ from check's ScopeIndex? It feels like the way it's used has significantly different semantics: in order of scope creation versus just putting all scopes into an array.

toolchain/sem_ir/formatter.cpp
```diff
@@ -575,6 +602,10 @@ class Formatter {
       FormatInterface(InterfaceId(i));
     }
 
+    for (int i : llvm::seq(sem_ir_.impls().size())) {
+      FormatImpl(ImplId(i));
+    }
```
jonmeow, line 579: Why the particular ordering here? Note, someone unsure if it should match the ordering in scopes (maybe there'd be a way to make that work as a loop over the scopes rather than as a loop over each type?)

toolchain/sem_ir/file.h
```diff
@@ -171,7 +171,44 @@ struct Interface : public Printable<Interface> {
...
+
+  // TODO: Track the generic parameters for `impl forall`.
+  // The type for which the impl is implementing a constraint.
+  TypeId self_id;
+  // The constraint that the impl implements.
+  TypeId constraint_id;
+
+  // The following members are set at the `{` of the impl definition.
+
+  // The definition of the impl. This is a ImplDecl.
```
jonmeow:
```suggestion
  // The definition of the impl. This is an ImplDecl.
```

toolchain/check/node_stack.h
```diff
@@ -440,20 +383,31 @@ class NodeStack {
       -> std::optional<IdKind> {
     // TODO: Patterns should also produce an `InstId`, but currently
     // `TuplePattern` produces an `InstBlockId`.
-    if (!!(category & Parse::NodeCategory::Expr)) {
+    constexpr Parse::NodeCategory InstCategories = Parse::NodeCategory::Expr;
```
jonmeow: FWIW, these constexprs seem verbose to me versus what they're saving, but if you want them I'd at least suggest using `auto` to avoid repeating the type on the line.

```
constexpr auto InstCategories = Parse::NodeCategory::Expr;
```

toolchain/check/node_stack.h
```diff
@@ -566,14 +522,46 @@ class NodeStack {
...
     }
   }
 
+  // Peeks at the ID associated with the top of the name stack.
+  template <IdKind RequiredIdKind>
+  auto Peek() const -> auto {
+    Entry back = stack_.back();
+    if constexpr (RequiredIdKind == IdKind::InstId) {
+      return back.id<SemIR::InstId>();
+    } else if constexpr (RequiredIdKind == IdKind::InstBlockId) {
```
jonmeow: Why `else if` here but no `else` in other functions using `if constexpr`?

toolchain/check/handle_impl.cpp
```diff
@@ -34,34 +38,143 @@ auto HandleImplForall(Context& context, Parse::ImplForallId parse_node)
...
 }
 
-auto HandleDefaultSelfImplAs(Context& /*context*/,
-                             Parse::DefaultSelfImplAsId /*parse_node*/)
-    -> bool {
-  // TODO: Determine self_id and push it onto node stack.
+// If the specified name scope corresponds to a class, returns the corresponding
+// class declaration.
+// TODO: Should this be somewhere more central?
+static auto AsClassScope(Context& context, SemIR::NameScopeId scope_id)
```
```suggestion
static auto TryAsClassScope(Context& context, SemIR::NameScopeId scope_id)
```

Maybe TryAs to reflect the optional result?

toolchain/check/handle_impl.cpp
```diff
@@ -34,34 +38,143 @@ auto HandleImplForall(Context& context, Parse::ImplForallId parse_node)
...
+  if (!scope.inst_id.is_valid()) {
+    return std::nullopt;
+  }
+  return context.insts().TryGetAs<SemIR::ClassDecl>(scope.inst_id);
+}
+
+auto HandleDefaultSelfImplAs(Context& context,
+                             Parse::DefaultSelfImplAsId parse_node) -> bool {
+  // Find the enclosing scope for the `impl`.
+  // TODO: Do this without modifying the node stack.
```
jonmeow, line 68: Is this something you're actively working on? Rather than pushing the enclosing_scope_id onto the stack (which seems to be resulting in NameScopeId support being added to NodeStack IIUC) would a `PeekScope(1)` work? Could there be an intervening scope which would interfere with that? The `current_scope` API is maybe a little restrictive; perhaps that should be more accessible in general.

[note, mainly commenting here because the comment didn't feel like it meshed well with the `implicit_param_list` code it's on]

toolchain/check/handle_impl.cpp
```diff
@@ -34,34 +38,143 @@ auto HandleImplForall(Context& context, Parse::ImplForallId parse_node)
...
+  if (implicit_param_list) {
+    context.node_stack().Push(implicit_param_list->first,
+                              implicit_param_list->second);
+  }
+
+  // TODO: This is also valid in a mixin.
+  auto class_decl = AsClassScope(context, enclosing_scope_id);
+  if (!class_decl) {
+    context.TODO(parse_node, "recover from `impl as` in non-class scope");
+    return false;
```
```suggestion
    return context.TODO(parse_node, "recover from `impl as` in non-class scope");
```

toolchain/check/handle_impl.cpp
```diff
@@ -34,34 +38,143 @@ auto HandleImplForall(Context& context, Parse::ImplForallId parse_node)
...
+    return;
+  }
+  auto& enclosing_scope = context.name_scopes().Get(enclosing_scope_id);
+
+  auto interface_type =
+      context.types().TryGetAs<SemIR::InterfaceType>(constraint_id);
+  if (!interface_type) {
+    context.TODO(parse_node, "extending non-interface constraint");
+    enclosing_scope.has_error = true;
+    return;
```
jonmeow, line 109:
```suggestion
    enclosing_scope.has_error = true;
    return context.TODO(parse_node, "extending non-interface constraint");
```

toolchain/check/handle_impl.cpp
```diff
@@ -34,34 +38,143 @@ auto HandleImplForall(Context& context, Parse::ImplForallId parse_node)
...
+  }
+
+  auto& interface = context.interfaces().Get(interface_type->interface_id);
+  if (!interface.scope_id.is_valid()) {
+    // TODO: Should this be valid?
+    context.TODO(parse_node, "extending interface without definition");
+    enclosing_scope.has_error = true;
+    return;
```
jonmeow, line 117:
```suggestion
    enclosing_scope.has_error = true;
    return context.TODO(parse_node, "extending interface without definition");
```

toolchain/sem_ir/formatter.cpp
zygoloid: Yeah, it's not great that we've ended up with `SemIR::InstNamer::ScopeIndex` shadowing `SemIR::ScopeIndex`. I think we could rename this one to `ScopeId` in line with our normal conventions for `Id` / `Index` suffixes, given that the numerical value here isn't supposed to be meaningful. (I'm assuming you don't want that done in *this* PR :).)

toolchain/sem_ir/formatter.cpp
zygoloid: I'll have a look in a separate PR.

toolchain/sem_ir/formatter.cpp
zygoloid: Maybe this is a place where renaming `InstNamer::ScopeIndex` to `ScopeId` would help to avoid the suggestion that the order is meaningful. :-)

The general idea is to order declarations in a way that makes the IR more readable with fewer forward references, so (for example) we put interfaces before impls, and classes before functions. I debated a bit whether impls should go before or after classes, and while it seemed a bit nicer to put interfaces and impls together, I don't really feel strongly, and putting impls after classes would be logical because impls will often refer to classes in their signature.

Let me know if you'd prefer a different order.

toolchain/sem_ir/file.h
zygoloid: Yeah, I added this for consistency, but given that it's always the same as the enclosing scope on the scope_id, it's not needed. Removed.

toolchain/check/node_stack.h
zygoloid: This is necessary in order for the `static_assert` at the end to check for an unhandled case. (What I really wanted to do was end this whole chain with `else { static_assert(false); }`, but we need a newer compiler for that to work.)

I thought about changing the other ones but wanted to keep the changes to node_stack in this PR small.

toolchain/check/node_stack.h
zygoloid: Done.

toolchain/check/node_stack.h
zygoloid: Done.

toolchain/check/handle_impl.cpp
zygoloid: Added.

toolchain/check/handle_impl.cpp
zygoloid: Renamed to better indicate that this is just "the stuff that `BuildImplDecl` returns". Hopefully that obviates a need for a comment.

toolchain/check/handle_impl.cpp
zygoloid: That won't work here, because this function returns `void` not `bool`.

toolchain/check/handle_impl.cpp
zygoloid: (Likewise)

toolchain/check/handle_impl.cpp
zygoloid: Yes, I'll be working on this and the other TODOs in this PR as I continue on impl support.

I didn't like making an assumption about which scopes are on the scope stack; that seemed more brittle than making an assumption about the node stack. In particular, we've been thinking about not pushing a parameter scope for the case of `impl` without `forall`, in which case we'd want to look at the top scope, not the second-to-top scope. I suppose in part I'm feeling this way because #3671 had to undo some assumptions about the scope stack that became incorrect, and maybe it'd actually be fine here given that we're setting up all the relevant scopes locally to this file.

Do you want this changed in this PR?

toolchain/check/node_stack.h
zygoloid: Also sent out #3685 with further node stack cleanups. If you like, I can land that one first and rebase this on top to remove the node stack refactorings from here.

zygoloid: Landed #3685 and merged.

toolchain/sem_ir/formatter.cpp
chandlerc: (Just LG-ed the refactoring PR, so resolving)

toolchain/check/handle_impl.cpp
zygoloid: #3691 removes this, making this `Build` function like all the others.

toolchain/check/handle_impl.cpp
zygoloid: This is cleaned up in #3691.

jonmeow (approved):
LGTM. I'm going to leave merging to you because I'm not sure what the PR stacking is like.

PR #3685, opened 2024-02-03
Tidy up and refactor the node stack.

zygoloid:
Add a type representing a non-discriminated union of IDs. Refactor the node stack to use it. Plus a few other refactorings aiming to clean up and simplify the code. The overall goal here is that adding a new kind of ID, node category, or instruction should only require changing one place in the node stack rather than a bunch of different changes.

One minor functionality change: crash backtraces now use the correct type for IDs when dumping the node stack rather than using `InstID` printing for all but one case.


toolchain/check/node_stack.cpp
```diff
@@ -4,22 +4,39 @@
...
+      constexpr auto IdKind = static_cast<Id::Kind>(Is);
+      if constexpr (IdKind == Id::Kind::None) {
+        output << " -> no value";
+      } else if constexpr (IdKind == Id::Kind::Invalid) {
+        CARBON_FATAL() << "Should not be in node stack";
+      } else {
+        output << id.As<IdKind>();
+      }
+    }...};
+  }(std::make_index_sequence<Id::NumValidKinds + 2>());
```
josh11b: I found this hard to understand

zygoloid: Replaced with an X macro

josh11b (approved):

PR #3687, opened 2024-02-05
Add GlobalInit to blocks to emit initialization instructions for globals

gysddn:
`GlobalInit` block is now static block within a `SemIR` which will be used to emit initialization instructions for variables in the `Package` scope.

inst_block_stack now has additional methods to handle `GlobalInit` block separately, this block can be popped without being finalized allowing to accumulate between all instances of variables.

At the end of the `check` phase, if this block is not empty , the function `__global_init` will be added with this block being inserted into it.

This block is pushed to `inst_block_scope` at the end `BindName`, allowing instruction to be emitted into it, then popped at the semicolon (VariableDecl).

This significantly changes the `SemIR` output, that's why this commit updates a lot of the test cases.


gysddn:
Only implemented for 'var', not 'let' at the moment.

toolchain/check/handle_binding_pattern.cpp
```diff
@@ -115,6 +115,10 @@ auto HandleAnyBindingPattern(Context& context, Parse::NodeId parse_node,
       if (context_parse_node_kind == Parse::NodeKind::ReturnedModifier) {
         RegisterReturnedVar(context, bind_id);
       }
+      if (auto current_scope_id = context.current_scope_id();
+          current_scope_id == SemIR::NameScopeId::Package) {
```
geoffromer: The named variable doesn't seem necessary here. Would this work instead?

```suggestion
      if (context.current_scope_id() == SemIR::NameScopeId::Package) {
```

toolchain/check/handle_variable.cpp
```diff
@@ -95,6 +95,10 @@ auto HandleVariableDecl(Context& context, Parse::VariableDeclId parse_node)
   }
 
   context.decl_state_stack().Pop(DeclState::Var);
+  if (auto current_scope_id = context.current_scope_id();
+      current_scope_id == SemIR::NameScopeId::Package) {
```
geoffromer:
```suggestion
  if (context.current_scope_id() == SemIR::NameScopeId::Package) {
```

toolchain/check/inst_block_stack.h
```diff
@@ -29,6 +29,8 @@ class InstBlockStack {
   // called in order to support lazy allocation.
   auto Push() -> void { Push(SemIR::InstBlockId::Invalid); }
 
+  auto PushInit() -> void;
```
geoffromer: Could you add comments to this and `PopInit`? It'd be great to have some of the information from the PR description in the code itself.

toolchain/check/context.cpp
```diff
@@ -618,6 +618,28 @@ auto Context::ParamOrArgEnd(Parse::NodeKind start_kind) -> SemIR::InstBlockId {
   return ParamOrArgPop();
 }
 
+auto Context::FinalizeInits() -> void {
+  inst_block_stack().PushInit();
+  if (!inst_block_stack().PeekCurrentBlockContents().empty()) {
+    AddInst({Parse::NodeId::Invalid, SemIR::Return{}});
+    // Pop the Inits block here to finalize it
```
```suggestion
    // Pop the Inits block here to finalize it.
```

Style nit: we use [standard punctuation](https://google.github.io/styleguide/cppguide.html#Punctuation,_Spelling_and_Grammar) in comments.

toolchain/check/context.cpp
```diff
@@ -618,6 +618,28 @@ auto Context::ParamOrArgEnd(Parse::NodeKind start_kind) -> SemIR::InstBlockId {
...
+    inst_block_stack().Pop();
+
+    // __global_init is only added if there are initialization instructions
```
```suggestion
    // __global_init is only added if there are initialization instructions.
```

Style: punctuation

toolchain/check/context.cpp
```diff
@@ -618,6 +618,28 @@ auto Context::ParamOrArgEnd(Parse::NodeKind start_kind) -> SemIR::InstBlockId {
...
+    auto nameid = sem_ir().identifiers().Add("__global_init");
+    sem_ir().functions().Add(
+        {.name_id = SemIR::NameId::ForIdentifier(nameid),
```
```suggestion
    auto name_id = sem_ir().identifiers().Add("__global_init");
    sem_ir().functions().Add(
        {.name_id = SemIR::NameId::ForIdentifier(name_id),
```

Style: underscores are used to separate words in [variable names](https://google.github.io/styleguide/cppguide.html#Variable_Names).

toolchain/check/inst_block_stack.cpp
```diff
@@ -21,6 +21,11 @@ auto InstBlockStack::Push(SemIR::InstBlockId id) -> void {
   ++size_;
 }
 
+auto InstBlockStack::PushInit() -> void {
+  Push(SemIR::InstBlockId::Inits);
+  stack_[size_ - 1].content = init_block_;
```
geoffromer: Is it possible for`size_` to not be 0 here? I'd naively expect initializers to always be at the bottom of the stack. If so, maybe rather than pushing and popping them, we could just always have them at the bottom of the stack?

toolchain/check/context.cpp
```diff
@@ -618,6 +618,28 @@ auto Context::ParamOrArgEnd(Parse::NodeKind start_kind) -> SemIR::InstBlockId {
   return ParamOrArgPop();
 }
 
+auto Context::FinalizeInits() -> void {
+  inst_block_stack().PushInit();
+  if (!inst_block_stack().PeekCurrentBlockContents().empty()) {
```
jonmeow: Had you considered manipulating `inits_` directly here, rather than adding it to the block stack?

toolchain/sem_ir/ids.h
```diff
@@ -362,6 +362,11 @@ struct InstBlockId : public IdBase, public Printable<InstBlockId> {
   // fully checked; intermediate state is in the Check::Context.
   static const InstBlockId Exports;
 
+  // Global declaration initialization instructions. Empty if none are present.
+  // Otherwise, __global_init function will be generated and this block will
+  // be inserted into it.
+  static const InstBlockId Inits;
```
jonmeow: Why do you choose `Inits` versus something such as `GlobalInit` to match the produced function? It may be helpful to be more specific in order to disambiguate from local initializers.

toolchain/check/context.h
```diff
@@ -363,6 +363,9 @@ class Context {
     inst_blocks().Set(SemIR::InstBlockId::Exports, exports_);
   }
 
+  // Finalizes the initialization function (__global_init)
```
```suggestion
  // Finalizes the initialization function (__global_init).
```

Style: Punctuation.

toolchain/check/inst_block_stack.cpp
```diff
@@ -52,6 +57,14 @@ auto InstBlockStack::Pop() -> SemIR::InstBlockId {
   return back.id;
 }
 
+auto InstBlockStack::PopInit() -> void {
+  CARBON_CHECK(stack_[size_ - 1].id == SemIR::InstBlockId::Inits)
+      << "Trying to pop Inits block from " << name_
+      << " but a different block is present!";
+  init_block_ = stack_[size_ - 1].content;
```
jonmeow:
```suggestion
  init_block_ = std::move(stack_[size_ - 1].content);
```

toolchain/check/inst_block_stack.cpp
jonmeow:
```suggestion
  stack_[size_ - 1].content = std::move(init_block_);
```

At present, this would perform a copy on PushInit and PopInit. That would create scaling issues when there are many or complex initializers. Using `move` should alleviate the issue.

jonmeow: I believe the package's inst block is always at the bottom of the stack. If you consider something like a global `let x: MyType(1 + 7) = Foo;`, the `let x: MyType(1 + 7)` goes into the package's inst block, while the `= Foo;` is going into the global init block. That's going to inhibit leaving global inits on the bottom of the stack.

jonmeow (commented):
Thanks, the output looks good! My comments are mostly style-related, this approach looks right to me.

toolchain/check/handle_binding_pattern.cpp
gysddn: Definitely! This totally has slipped my attention.
I updated them.

toolchain/check/inst_block_stack.h
gysddn: Done!

toolchain/check/inst_block_stack.cpp
gysddn: Done!

toolchain/check/context.h
gysddn: I updated them all.

toolchain/sem_ir/ids.h
gysddn: I agree.
I renamed the block to `GlobalInit`.

toolchain/check/context.cpp
gysddn: I did, but I just couldn't come up with a minimal way to do it.

The `GlobalInit` is being constructed inside the stack object, so there is no direct way to either "view" or "manipulate" it. I had to check if it was empty or not, and if it wasn't, I had to push it onto the stack again to add the `Return` instruction. I just thought I would push it regardless, that way I would be able to both "check its emptiness" and "add the return".

I realize it is not the most straight-forward, but I didn't want to make any more API changes to `InstBlockStack` than I had to.

toolchain/check/context.h
```diff
@@ -363,6 +363,9 @@ class Context {
     inst_blocks().Set(SemIR::InstBlockId::Exports, exports_);
   }
 
+  // Finalizes the initialization function (__global_init).
+  auto FinalizeInits() -> void;
```
jonmeow: For FinalizeInits/PushInit/PopInit, could you please rename them using Globalnit similar to the block rename?

toolchain/check/context.cpp
jonmeow: Makes sense, with the `return` function. Makes me wonder if we should switch around how we organize `Finalize` logic, now that we're adding more (if we had a single entry point for both exports and __global_init then it might look a little more reasonable to put this full function on InstBlockStack), but this seems good for now.

toolchain/check/context.h
gysddn: Done.

toolchain/check/context.cpp
gysddn: That makes sense. I might try that and see how it would work.

jonmeow (approved):
Looks good, thanks!

jonmeow:
Looking at https://github.com/carbon-language/carbon-lang/actions/runs/7833531743/job/21374710204, I think the PR needs to pull in changes made at head. Specifically #3688 factored out current_scope_id and related functionality to its own class.

gysddn:
I added to merge commit, but if that seems too excessive to review. I can rebase this branch and make it more minimal.

PR #3688, opened 2024-02-05
Factor the scope stack and lexical lookups out of Check::Context.

zygoloid:


toolchain/check/scope_stack.h
```diff
@@ -0,0 +1,195 @@
...
+  // Runs verification that the processing cleanly finished.
+  auto VerifyOnFinish() -> void;
+
+  auto return_scope_stack() -> llvm::SmallVector<ReturnScope>& {
+    return return_scope_stack_;
+  }
+
+  auto break_continue_stack() -> llvm::SmallVector<BreakContinueScope>& {
+    return break_continue_stack_;
+  }
```
jonmeow, line 124: Had you considered leaving these on Context? Versus the main scope stack, the SmallVector return means this doesn't really control either, and they're directly exposed on Context itself. You might need a pointer to update returned_var, but I don't even see logic for break_continue_stack.

toolchain/check/scope_stack.h
```diff
@@ -0,0 +1,195 @@
...
+  // Returns the instruction associated with the current scope, or Invalid if
+  // there is no such instruction, such as for a block scope.
+  auto current_scope_inst_id() const -> SemIR::InstId {
+    return current_scope().scope_inst_id;
+  }
+
+  // Returns the current scope, if it is of the specified kind. Otherwise,
+  // returns nullopt.
+  template <typename InstT>
+  auto GetCurrentScopeAs(const SemIR::File& sem_ir) -> std::optional<InstT> {
```
jonmeow, line 85: Similar to `PushScope` -> `Push`, I think you could probably rename `current_scope` -> `current` or `peek`

toolchain/check/scope_stack.h
```diff
@@ -0,0 +1,195 @@
...
+
+  // Pops the top scope from scope_stack_ if it contains no names.
+  auto PopIfEmpty() -> void {
+    if (scope_stack_.back().names.empty()) {
+      Pop();
+    }
+  }
+
+  // Pops scopes until we return to the specified scope index.
+  auto PopToScope(ScopeIndex index) -> void;
```
jonmeow: Had you considered just `PopTo`?

jonmeow (approved):
LGTM, just a few minor comments so feel free to merge.

toolchain/check/scope_stack.h
zygoloid: Yeah, I started that way, but the `returned var` handling seemed to fit better on `ScopeStack`, and the return scope stack is mixed up with that too. I was also thinking that we might actually want to switch to maintaining these stacks more like the non-lexical scope stacks, so that when you push a scope you say whether it's a return / break / continue scope, and don't have to remember to update these stacks when you enter a different scope. It probably doesn't matter right now that we're still in an enclosing return/break/continue scope when we enter, say, a nested class, but it could matter if we started allowing statements in subexpressions (eg, block expressions of some kind).

toolchain/check/scope_stack.h
zygoloid: Done.

toolchain/check/scope_stack.h
zygoloid: Done.

PR #3690, opened 2024-02-06
Refactor `SemIR::InstNamer`

zygoloid:
- Rename `ScopeIndex` to `ScopeId`, because the order is not meaningful. This avoids collisions with `SemIR::ScopeIndex`.
- Change `GetScopeFor` to use an `if constexpr` chain rather than duplicating the numbering logic across multiple functions.
- Simplify `GetNameFor`, avoiding multiple identical overloads.

As requested in review of #3683.


chandlerc (approved):
LG

PR #3691, opened 2024-02-06
Use the DeclNameStack for impl declarations.

zygoloid:
They don't have names, but using the DeclNameStack anyway keeps our
behavior more consistent, and keeps track of the enclosing name scope
and the prior state of the scope stack for us.

Depends on #3683.


jonmeow (approved):
LGTM, thanks!

PR #3693, opened 2024-02-07
Implement constant handling for types of imports.

jonmeow:
This provides support for Const, Pointer, Struct, and Tuple types. It does not cover Class, Function, or Interface which have their own Id and are tracked slightly differently.


toolchain/check/import_ref.cpp
```diff
@@ -0,0 +1,266 @@
...
+      case SemIR::InstKind::StructType:
+        return TryResolveTypedInst(inst.As<SemIR::StructType>());
+
+      case SemIR::InstKind::TupleType:
+        return TryResolveTypedInst(inst.As<SemIR::TupleType>());
+
+      default:
+        context_.TODO(
+            Parse::NodeId::Invalid,
+            llvm::formatv("TryResolveNextInst on {0}", inst.kind()).str());
```
chandlerc: Stale function name here?

toolchain/check/import_ref.cpp
```diff
@@ -0,0 +1,266 @@
...
+    return context_.types().GetConstantId(
+        context_.GetStructType(context_.inst_blocks().Add(fields)));
+  }
+
+  auto TryResolveTypedInst(SemIR::TupleType inst) -> SemIR::ConstantId {
+    CARBON_CHECK(inst.type_id == SemIR::TypeId::TypeType);
+
+    // First verify that all used types are ready.
+    bool added_todos = false;
+    auto orig_el_type_ids = import_ir_.type_blocks().Get(inst.elements_id);
```
chandlerc: Minor, but `el` as an abbreviation tripped me up for a bit reading this.

toolchain/check/import_ref.cpp
```diff
@@ -0,0 +1,266 @@
...
+      el_type_ids.push_back(context_.GetTypeIdForTypeConstant(el_const_id));
+    }
+
+    return context_.types().GetConstantId(context_.GetTupleType(el_type_ids));
+  }
+
+  Context& context_;
+  const SemIR::File& import_ir_;
+  SemIR::ConstantValueStore& import_ir_constant_values_;
+  llvm::SmallVector<SemIR::InstId> todo_;
```
chandlerc: Particularly given the use of a to-do API elsewhere, id prefer a different name. My canonical name for this pattern is "worklist".

jonmeow: WDYT of `stack_` to better label the access pattern?

toolchain/check/import_ref.cpp
jonmeow: `elem_`?

toolchain/check/import_ref.cpp
jonmeow: Yeah, fixed.

toolchain/check/import_ref.cpp
chandlerc: (as discussed, let's go with `work_stack_`)

chandlerc (approved):
LG!

PR #3695, opened 2024-02-08
Add diagnostic support so that we can just pass in TypeId.

jonmeow:
Note we may also want to do this with NameId, maybe some other things, but the TypeId use is pretty broad and repetitive -- I thought I'd start with it first.

### DIFF
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -73,6 +73,19 @@ class SemIRLocationTranslator
     }
   }
 
+  auto TranslateArg(DiagnosticTypeTranslation translation, llvm::Any arg) const
+      -> llvm::Any override {
+    switch (translation) {
+      case DiagnosticTypeTranslation::TypeId: {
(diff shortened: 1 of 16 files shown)


chandlerc (approved):
Niiiice!

I have some ideas for potential tweaks / improvements to the mechanics behind this, but not sure any even work, and it seems much better to get this in with the cleanups to all the call sites. That way I can see if anything even works (decent chance you even tried some and in fact, they don't work). =]

jonmeow:
> Niiiice!
> 
> I have some ideas for potential tweaks / improvements to the mechanics behind this, but not sure any even work, and it seems much better to get this in with the cleanups to all the call sites. That way I can see if anything even works (decent chance you even tried some and in fact, they don't work). =]

SGTM, I'm sure there are other ways the diagnostic support could be set up. It strikes me that the struct could probably be a constexpr function and might work better that way, now. But happy to let that be handled separately.

PR #3698, opened 2024-02-09
Add --dump_output to file_test to help debugging.

jonmeow:
When we always had a single file in the test file, we could run the driver over it to see output. This adds the ability to do something similar for multi-file tests. So for example in test output, there's now:

```
[ RUN      ] ToolchainFileTest.toolchain/check/testdata/class/fail_todo_import.carbon

To test this file alone, run:
  bazel test //toolchain/testing:file_test --test_arg=--file_tests=toolchain/check/testdata/class/fail_todo_import.carbon

To view output, run:
  bazel run //toolchain/testing:file_test --test_arg=--file_tests=toolchain/check/testdata/class/fail_todo_import.carbon --test_arg=--dump_output

[       OK ] ToolchainFileTest.toolchain/check/testdata/class/fail_todo_import.carbon (64 ms)
```

And here's what a run looks like:

```
$ bazel run //toolchain/testing:file_test --test_arg=--file_tests=toolchain/check/testdata/class/fail_todo_import.carbon --test_arg=--dump_output
(elided bazel output)
Executing tests from //toolchain/testing:file_test
===============================================================================
= toolchain/check/testdata/class/fail_todo_import.carbon
===============================================================================
= stderr
===============================================================================
b.carbon: ERROR: Semantics TODO: `TryResolveImportRefUnused on ClassDecl`.
b.carbon: ERROR: Semantics TODO: `TryResolveImportRefUnused on ClassDecl`.
b.carbon: ERROR: Semantics TODO: `TryResolveInst on ClassType`.
b.carbon: ERROR: Semantics TODO: `TryResolveInst on ClassType`.
b.carbon:18:29: ERROR: Name `d_ref` not found.
var d: (ForwardDeclared,) = d_ref;
                            ^~~~~
===============================================================================
= stdout
===============================================================================
--- a.carbon
(eliding semir output)
}

===============================================================================
= Exit with success: false
===============================================================================

Done!
```

This also switches from PrettyStackTraceFormat to PrettyStackTraceString. This is partly so that we can share the produced command (avoiding two different format strings corresponding to the same value), but also it's in my mind to step away from https://github.com/llvm/llvm-project/pull/77351.

### DIFF
--- a/testing/file_test/file_test_base.cpp
+++ b/testing/file_test/file_test_base.cpp
@@ -35,6 +35,9 @@ ABSL_FLAG(bool, autoupdate, false,
 ABSL_FLAG(unsigned int, threads, 0,
           "Number of threads to use when autoupdating tests, or 0 to "
           "automatically determine a thread count.");
+ABSL_FLAG(bool, dump_output, false,
+          "Instead of verifying files match test output, directly dump output "
+          "to stderr.");
 
 namespace Carbon::Testing {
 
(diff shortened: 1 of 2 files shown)


testing/file_test/file_test_base.cpp
```diff
@@ -66,20 +69,25 @@ static auto SplitOutput(llvm::StringRef output)
...
 auto FileTestBase::TestBody() -> void {
-  std::optional<llvm::PrettyStackTraceFormat> stack_trace_entry;
+  std::string test_command;
+  std::optional<llvm::PrettyStackTraceString> stack_trace_entry;
 
   // If we're being run from bazel, provide some assistance for understanding
   // and reproducing failures.
   const char* target = getenv("TEST_TARGET");
   if (target) {
+    constexpr const char* CommandFormat =
```
chandlerc, line 79: Maybe `llvm::StringLiteral`?

chandlerc (approved):
Minor suggestion below, LG either way.

testing/file_test/file_test_base.cpp
jonmeow: This seemed odd to me, but `llvm::formatv` takes a `const char*`, so if I were to use `llvm::StringLiteral` I'd need to `.data()`. I think `llvm::StringLiteral` would only add indirection in this situation.

https://llvm.org/doxygen/FormatVariadic_8h_source.html

PR #3699, opened 2024-02-11
Add help information printing for a specific subcommand

### LINKED ISSUE #3694: Subcommand help doesn't work
### Description of the bug:

`carbon help` says "you can pass a specific subcommand to the `help` subcommand to get details about its usage", but that doesn't seem to work.

`bazel run toolchain/driver:carbon help compile`

I expected it to print an explanation of the `compile` subcommand.

### What actually happened?

It printed "ERROR: Found unexpected positional argument or subcommand: 'compile'".

_No response_

clavin:
The `command_library` provides functionality for end users to query documentation and help information about commands through the command line interface itself. At present, this is implemented as:

* A flag, `--help`, on every command
* A meta subcommand, `help`, on every command that has other subcommands

To match with common functionality in other command line interfaces, this PR amends the `help` meta subcommand to accept an optional positional string argument. This argument specifies which subcommand to print help information for. When not specified, the present behavior is maintained, printing the help information for its parent command.

A potentially unexpected consequence of this implementation is that you can query for help on meta subcommands as well: `help help` is a valid input which prints the help information about the `help` meta subcommand (e.g., same for `version`). I don't see any harm in this behavior so I chose not to explicitly prevent it, but I'm open to preventing it if deemed unwanted. (I'm also happy to workshop the strings in this PR.)

Closes #3694.

#### Before

```
$ carbon help compile
```
```
ERROR: Found unexpected positional argument or subcommand: 'compile'
```

#### After

```
$ carbon help compile
```
```
Compile Carbon source code.

This subcommand runs the Carbon compiler over input source code, checking it for errors and producing the requested output.

Error messages are written to the standard error stream.

Different phases of the compiler can be selected to run, and intermediate state can be written to standard output as these phases progress.

Subcommand 'compile' usage:
  carbon [-v] compile [OPTIONS] <FILE>...

...
```

### DIFF
--- a/common/command_line.cpp
+++ b/common/command_line.cpp
@@ -92,6 +92,8 @@ class MetaPrinter {
   void RegisterWithCommand(const Command& command, CommandBuilder& builder);
 
   void PrintHelp(const Command& command) const;
+  void PrintHelpForSubcommandName(const Command& command,
+                                  llvm::StringRef subcommand_name) const;
   void PrintVersion(const Command& command) const;
   void PrintSubcommands(const Command& command) const;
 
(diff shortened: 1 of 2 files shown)


chandlerc (approved):
Awesome chango! Thanks for the contribution!

PR #3700, opened 2024-02-12
Add 'let' handling for ImportRefUsed.

jonmeow:
- File::StringifyTypeExpr now has a case that hits the ImportRefUsed TODO, so implementing that. I think the `static` approach will be helpful in ensuring there aren't access bugs, particularly when future support is added.

- `let` wasn't adding to exports because it doesn't use `decl_name_stack` the way `var` does. This now adds to exports, but we might want to unify logic for issues such as this.

- BuildImportRefUsedValueRepr is now called for another inst kind, and it seemed like calling back to BuildValueRepr was the best way to resolve this. I don't *think* that's going to cause recursion.

### DIFF
--- a/toolchain/check/context.cpp
+++ b/toolchain/check/context.cpp
@@ -712,17 +712,9 @@ class TypeCompleter {
     return value_rep;
   };
 
-  auto BuildImportRefUsedValueRepr(SemIR::TypeId type_id,
-                                   SemIR::ImportRefUsed import_ref) const
-      -> SemIR::ValueRepr {
-    CARBON_CHECK(import_ref.inst_id.is_builtin())
-        << "TODO: Handle non-builtin ImportRefUsed cases, such as functions, "
-           "classes, and interfaces";
-
-    const auto& import_ir = context_.import_irs().Get(import_ref.ir_id);
-    auto import_inst = import_ir->insts().Get(import_ref.inst_id);
-
-    switch (import_inst.As<SemIR::Builtin>().builtin_kind) {
+  auto BuildBuiltinValueRepr(SemIR::TypeId type_id,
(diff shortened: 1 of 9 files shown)


toolchain/check/context.cpp
```diff
@@ -743,6 +735,16 @@ class TypeCompleter {
     llvm_unreachable("All builtin kinds were handled above");
   }
 
+  auto BuildImportRefUsedValueRepr(SemIR::TypeId type_id,
+                                   SemIR::ImportRefUsed import_ref) const
+      -> SemIR::ValueRepr {
+    const auto& import_ir = context_.import_irs().Get(import_ref.ir_id);
+    auto import_inst = import_ir->insts().Get(import_ref.inst_id);
+    CARBON_CHECK(import_inst.kind() != SemIR::InstKind::ImportRefUsed)
+        << "If ImportRefUsed can point at another, this would be recursive.";
```
chandlerc, line 744: My guess is this will come up when re-exporting an import maybe? And then it might be valid to hit that here, but with a different `import_ir` to hop back through the chain? Or maybe that will have a different representation entirely?

(Clearly this isn't about this PR or blocking, just curious about the future here.)

chandlerc (approved):
Nice, and merging. There is a question below, but that's just for discussion of future stuff.

toolchain/check/context.cpp
jonmeow: I'm thinking to have this happen the ImportRef may need to be in the export list directly. As a consequence, really it depends on the representation of the re-export. For example, if the re-export is via `alias`, it's likely the instruction we'd see exported is actually the `alias` rather than for the Imports. If you can re-export an entire library, we'd probably just do that by dumping more direct ImportRefs. So this logic might actually not be triggered.

PR #3701, opened 2024-02-13
Implement 'alias' with transparent semantics.

jonmeow:
Adds `BindAlias` with a hybrid of `BindName` and `NameRef` semantics. I think it's slightly closer to `BindName` because it introduces a name, so I'm going more in that direction. This also matches the need for `bind_name_id` with imports on enclosing scopes.

Note, only things that look like a name reference are being allowed on the RHS of `alias`. This includes builtins that look like name references, such as `bool`, but not ones that turn into values underneath, such as `false`.


jonmeow:
Note, another option I could go with for now is that anything on the right must be either a builtin or a name ref, dropping generic expression support. Happy to change, although being able to alias `{.a: i32}` would probably be helpful things that don't work if it's restricted to a name ref. That's about how I reached the template-constant requirement.

chandlerc:
> Note, another option I could go with for now is that anything on the right must be either a builtin or a name ref, dropping generic expression support. Happy to change, although being able to alias `{.a: i32}` would probably be helpful things that don't work if it's restricted to a name ref. That's about how I reached the template-constant requirement.

FWIW, I think making the RHS a name makes sense. Our current provisional [design](https://github.com/carbon-language/carbon-lang/tree/trunk/docs/design#aliases):

> Note that the right-hand side of the equal sign (=) is a name not a value, so alias four = 4; is not allowed. This allows alias to work with entities like namespaces, which aren't values in Carbon.

It's pretty provisional, but seems reasonable to stick to a minimal model for now?

FWIW, I think you could do the template constant thing if you want, w/o really needing anything more than the modifiers we already have I think?

```carbon
let template Flag:! bool = true;
let template OtherFlag:! bool = Flag;
```

That seem reasonable to implement even w/o other forms of `template` support?

jonmeow:
> > Note that the right-hand side of the equal sign (=) is a name not a value, so alias four = 4; is not allowed. This allows alias to work with entities like namespaces, which aren't values in Carbon.
> 
> It's pretty provisional, but seems reasonable to stick to a minimal model for now?

Switching to a nameref then.

Note that does prevent things like `alias MyAlias = MyGenericType(param);`, not sure if that's deliberate.

jonmeow:
> `let template Flag:! bool = true;`

Note, I think `true` and `false` should still work with `alias`. The discussion has been to configure them as choice values, so I'd expect we'll have something along the lines of:

`let template true:! bool = BoolValue.True;`

Then, `alias f = true;` should still be valid, since it's a name reference to a defined constant, rather than a name reference to a literal value. Note, I think this is a little different from integer literal values.

[currently I might leave this failing, but I might treat that more as a bug]

jonmeow:
> That seem reasonable to implement even w/o other forms of `template` support?

I should add, specifically -- I'm trying not to implement parsing for `template` in `var` or `let`. These are divergent implementations and while I think the intent is to work towards pattern binding support, that's not something I think is feasible in a similar time span as to `alias`, particularly with zygoloid out.

toolchain/parse/state.def
```diff
@@ -672,6 +677,35 @@ CARBON_PARSE_STATE(NamespaceFinish)
 //   (state done)
 CARBON_PARSE_STATE(Package)
 
+// Handles `alias`.
```
josh11b:
```suggestion
// Handles `alias` after the introducer.
```

toolchain/parse/state.def
```diff
@@ -672,6 +677,35 @@ CARBON_PARSE_STATE(NamespaceFinish)
...
+//
+// alias ...
+//      ^
+//   1. DeclNameAndParamsAsNone
+//   2. AliasInitializer
+CARBON_PARSE_STATE(Alias)
+
+// Handles `alias`.
```
josh11b:
```suggestion
// Handles the part after `=` of `alias`.
```
or
```suggestion
// Handles `alias` after the name.
```

toolchain/parse/state.def
```diff
@@ -672,6 +677,35 @@ CARBON_PARSE_STATE(NamespaceFinish)
...
+//
+// alias ... = ... ;
+//           ^
+//   1. Expression
+//   2. AliasFinish
+//
+// alias ???
+//       ^
+//   (state done)
+CARBON_PARSE_STATE(AliasInitializer)
```
josh11b: I'm wondering if there is a better name than "initializer" here, since this isn't just the initial value. Maybe `AliasAssignment` or `AliasSource`?

toolchain/parse/state.def
```diff
@@ -672,6 +677,35 @@ CARBON_PARSE_STATE(NamespaceFinish)
...
+
+// Handles `alias` after the name.
```
josh11b:
```suggestion
// Concludes `alias` declarations.
```

toolchain/parse/state.def
```diff
@@ -672,6 +677,35 @@ CARBON_PARSE_STATE(NamespaceFinish)
...
+//      ^
+//   1. DeclNameAndParamsAsNone
+//   2. AliasInitializer
+CARBON_PARSE_STATE(Alias)
+
+// Handles `alias`.
+//
+// alias ... = ... ;
+//           ^
+//   1. Expression
```
josh11b: I see both `Expression` and `Expr` in this file, but `Expr` looks to be correct. This is probably a result me incompletely renaming "expression" in comments.

```suggestion
//   1. Expr
```

toolchain/parse/testdata/alias/fail_syntax.carbon
```diff
@@ -0,0 +1,99 @@
...
+// CHECK:STDERR: fail_syntax.carbon:[[@LINE+3]]:8: ERROR: `alias` requires a `=` for the source.
+// CHECK:STDERR: alias a;
+// CHECK:STDERR:        ^
+alias a;
+
+// CHECK:STDERR: fail_syntax.carbon:[[@LINE+3]]:7: ERROR: `alias` introducer should be followed by a name.
+// CHECK:STDERR: alias =;
+// CHECK:STDERR:       ^
+alias =;
+
```
josh11b, line 31:
```suggestion

alias a = b c d;
```

toolchain/check/handle_alias.cpp
```diff
@@ -0,0 +1,70 @@
...
+  return true;
+}
+
+auto HandleAlias(Context& context, Parse::AliasId /*parse_node*/) -> bool {
+  auto [expr_node, expr_id] = context.node_stack().PopExprWithParseNode();
+
+  auto name_context = context.decl_name_stack().FinishName();
+
+  LimitModifiersOnDecl(context, KeywordModifierSet::None,
+                       Lex::TokenKind::Namespace);
```
josh11b:
```suggestion
                       Lex::TokenKind::Alias);
```

Please add a test that uses modifiers.

toolchain/check/handle_alias.cpp
```diff
@@ -0,0 +1,70 @@
...
+  context.decl_state_stack().Pop(DeclState::Alias);
```
josh11b: I suspect we'll want to at least support access modifiers.
```suggestion
  LimitModifiersOnDecl(context, KeywordModifierSet::Access,
                       Lex::TokenKind::Namespace);
   if (!!(modifiers & KeywordModifierSet::Access)) {
    context.TODO(context.decl_state_stack().innermost().saw_access_modifier,
                 "access modifier");
  }
  context.decl_state_stack().Pop(DeclState::Alias);
```

toolchain/check/handle_alias.cpp
```diff
@@ -0,0 +1,70 @@
...
+    // Type (`bool`) and value (`false`) literals provided by the builtin
+    // structure should be turned into name references.
+    // TODO: Look into handling `false`, this doesn't do it right now because it
+    // sees a value instruction instead of a builtin.
+    alias_id = context.AddInst(
+        {name_context.parse_node,
+         SemIR::BindAlias{context.insts().Get(expr_id).type_id(), bind_name_id,
+                          expr_id}});
+  } else if (auto inst = context.insts().TryGetAs<SemIR::NameRef>(expr_id)) {
+    // Pass through name references, albeit changing the name in use.
```
josh11b, line 55: Can we have a test that shows the error message for:
```
let a: i32 = 1;
alias b = a;
let c: b = 2;
```
?

toolchain/check/handle_alias.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+auto HandleAliasInitializer(Context& /*context*/,
+                            Parse::AliasInitializerId /*parse_node*/) -> bool {
+  return true;
+}
+
+auto HandleAlias(Context& context, Parse::AliasId /*parse_node*/) -> bool {
+  auto [expr_node, expr_id] = context.node_stack().PopExprWithParseNode();
+
+  auto name_context = context.decl_name_stack().FinishName();
```
josh11b, line 29: Can we have some tests showing name conflicts with aliases? Like:
```
alias a = i32;
var a: i32 = 1;

var b: i32 = 2;
alias b = i32;
```

toolchain/check/testdata/namespace/fail_decl_in_alias.carbon
```diff
@@ -0,0 +1,38 @@
...
+
+alias ns = NS;
+
+// CHECK:STDERR: fail_decl_in_alias.carbon:[[@LINE+6]]:7: ERROR: Name qualifiers are only allowed for entities that provide a scope.
+// CHECK:STDERR: fn ns.A() -> i32 { return 0; }
+// CHECK:STDERR:       ^
+// CHECK:STDERR: fail_decl_in_alias.carbon:[[@LINE+3]]:4: Non-scope entity referenced here.
+// CHECK:STDERR: fn ns.A() -> i32 { return 0; }
+// CHECK:STDERR:    ^~
+fn ns.A() -> i32 { return 0; }
```
josh11b, line 18: Why is this an error?

toolchain/parse/state.def
jonmeow: Done

toolchain/parse/state.def
jonmeow: Done

toolchain/parse/state.def
jonmeow: Done

toolchain/parse/state.def
jonmeow: I think we should probably be using "initializer" for consistency, in the same way that we'd probably call the thing for `let` an initializer even though it isn't can't be changed after initialization.

However, this at least I'll switch to AfterName since it's really including the `=`.

toolchain/parse/state.def
jonmeow: Going with:

```
// Handles `alias` parsing at the end.
```

I'm trying to be consistent with other similar comments:

> // Handles `var` parsing at the end.
> // Handles `let` parsing at the end.

toolchain/parse/testdata/alias/fail_syntax.carbon
jonmeow: Done

toolchain/check/handle_alias.cpp
jonmeow: Done

toolchain/check/handle_alias.cpp
jonmeow: Done

toolchain/check/handle_alias.cpp
jonmeow: Done

toolchain/check/testdata/namespace/fail_decl_in_alias.carbon
jonmeow: Stemming from some recollections in explorer discussion, aliases are not supported when declaring names. This is WAI even if the error isn't easy to understand.

toolchain/check/handle_alias.cpp
jonmeow: Added fail_alias_misuse (suggestions welcome for a better name, I'm not sure what feature you're trying to test with this)

josh11b: I was expecting to see an error that either said `a` or `b` (and I wasn't sure which would happen), but it turns out that is a bad example.

toolchain/check/testdata/namespace/fail_decl_in_alias.carbon
josh11b: Could you add a comment explaining then, please, since this wasn't what I expected? (No need to change the semantics until we have a spec for what `alias` does.)

josh11b (approved):

toolchain/check/handle_alias.cpp
jonmeow: Okay, renaming to "fail_aliased_name_in_error`

toolchain/check/testdata/namespace/fail_decl_in_alias.carbon
jonmeow: Done

PR #3703, opened 2024-02-14
Add import handling for FunctionDecl

jonmeow:
This works by creating a faux FunctionDecl in the context of the current IR, which seems to be working for function calls. Deduced params are there, but won't really be tested until classes are up and running. Also I may need to look further at return_slot_id to ensure it's working. But the basics, I think, are here.

Reorganizes some other ImportRef work from `has_unresolved` that'd relied on manual calls to a more detection-based `HasUnresolved` approach that doesn't require as much checking.


toolchain/check/import_ref.cpp
```diff
@@ -73,8 +73,25 @@ class ImportRefResolver {
...
+      SemIR::NameScopeId::Invalid;
+
+  // Returns true if new unresolved constants were found.
+  //
+  // At the start of a function, do:
+  //   auto initial_work = work_stack_.size();
+  // Then when determining:
+  //   if (HasUnresolved(initial_work)) { ... }
+  auto HasUnresolved(size_t initial_work) -> bool {
+    CARBON_CHECK(initial_work <= work_stack_.size());
```
geoffromer: Could this have an error message? It wasn't quite clear to me what a failure here would mean.

toolchain/check/import_ref.cpp
```diff
@@ -73,8 +73,25 @@ class ImportRefResolver {
...
+    return initial_work < work_stack_.size();
+  }
+
+  // Returns the ConstantId for an InstId. Adds it to the stack and sets
+  // has_unresolved to true if it's not ready.
```
geoffromer: What is has_unresolved?

toolchain/check/import_ref.cpp
```diff
@@ -83,12 +100,56 @@ class ImportRefResolver {
...
+    const_ids.reserve(param_refs.size());
+    for (auto inst_id : param_refs) {
+      const_ids.push_back(
+          GetConstantId(import_ir_.insts().Get(inst_id).type_id()));
+    }
+    return const_ids;
+  }
+
+  // Given a param_refs_id and const_ids from GetParamConstantIds, returns a
+  // version localized to the current IR.
```
geoffromer:
```suggestion
  // version of param_refs_id localized to the current IR.
```

toolchain/check/import_ref.cpp
```diff
@@ -83,12 +100,56 @@ class ImportRefResolver {
...
+    llvm::SmallVector<SemIR::ConstantId> const_ids;
+    const_ids.reserve(param_refs.size());
+    for (auto inst_id : param_refs) {
+      const_ids.push_back(
+          GetConstantId(import_ir_.insts().Get(inst_id).type_id()));
+    }
+    return const_ids;
+  }
+
+  // Given a param_refs_id and const_ids from GetParamConstantIds, returns a
```
geoffromer: Would it work to have this function call `GetParamConstantIds`, rather than having the caller do it? Might it even make sense to inline `GetParamConstantIds` into the body of this function?

toolchain/check/import_ref.cpp
```diff
@@ -83,12 +100,56 @@ class ImportRefResolver {
...
-  // Returns the ConstantId for a TypeId, or adds it to the stack and
-  // returns Invalid if the ConstantId is not ready.
+  // Returns the ConstantId for a TypeId. Adds it to the stack and sets
+  // has_unresolved to true if it's not ready.
   auto GetConstantId(SemIR::TypeId type_id) -> SemIR::ConstantId {
     return GetConstantId(import_ir_.types().GetInstId(type_id));
   }
 
+  // Given a param_refs_id, returns the necessary constants to convert it. Sets
+  // has_unresolved to true if any aren't ready.
```
geoffromer:
```suggestion
  // Returns the ConstantId for the type of each parameter in the parameter list param_refs_id, in order. Sets
  // has_unresolved to true if any aren't ready.
```

That's just a guess; it may be inaccurate. The main point of my suggestion is to clarify what the return value means (rather than what it's used for), and in particular to make it explicit that the elements of the return value correspond to the parameters.

toolchain/check/import_ref.cpp
```diff
@@ -83,12 +100,56 @@ class ImportRefResolver {
...
+    for (auto [ref_id, const_id] : llvm::zip(param_refs, const_ids)) {
+      new_param_refs.push_back(context_.AddInstInNoBlock(
+          {SemIR::ImportRefUsed{context_.GetTypeIdForTypeConstant(const_id),
+                                import_ir_id_, ref_id}}));
+    }
+    return context_.inst_blocks().Add(new_param_refs);
+  }
+
+  // Translates a NameId from the import IR to a local NameId.
+  auto GetNameId(SemIR::NameId name_id) -> SemIR::NameId {
```
geoffromer: Maybe call this `GetLocalNameId` or something like that?

As a less important suggestion along the same lines, maybe call the parameter `imported_name_id`?

toolchain/check/import_ref.cpp
```diff
@@ -83,12 +100,56 @@ class ImportRefResolver {
...
   }
 
+  // Given a param_refs_id, returns the necessary constants to convert it. Sets
+  // has_unresolved to true if any aren't ready.
+  auto GetParamConstantIds(SemIR::InstBlockId param_refs_id)
+      -> llvm::SmallVector<SemIR::ConstantId> {
+    if (param_refs_id == SemIR::InstBlockId::Empty) {
+      return {};
+    }
+    const auto& param_refs = import_ir_.inst_blocks().Get(param_refs_id);
```
geoffromer, line 117: The style guide [says](https://google.github.io/styleguide/cppguide.html#Type_deduction) "Use type deduction only if it makes the code clearer to readers who aren't familiar with the project, or if it makes the code safer". This code would have been clearer to me if this type was explicit.

toolchain/check/import_ref.cpp
```diff
@@ -83,12 +100,56 @@ class ImportRefResolver {
...
+
+  // Given a param_refs_id and const_ids from GetParamConstantIds, returns a
+  // version localized to the current IR.
+  auto GetParamRefsId(SemIR::InstBlockId param_refs_id,
+                      const llvm::SmallVector<SemIR::ConstantId>& const_ids)
+      -> SemIR::InstBlockId {
+    if (param_refs_id == SemIR::InstBlockId::Empty) {
+      return SemIR::InstBlockId::Empty;
+    }
+    const auto& param_refs = import_ir_.inst_blocks().Get(param_refs_id);
```
geoffromer, line 136: Here again, I think an explicit type would be clearer.

toolchain/check/import_ref.cpp
```diff
@@ -173,35 +235,86 @@ class ImportRefResolver {
         SemIR::ConstType{SemIR::TypeId::TypeType, inner_type_id});
   }
 
+  auto TryResolveTypedInst(SemIR::FunctionDecl inst) -> SemIR::ConstantId {
+    auto initial_work = work_stack_.size();
+    auto type_const_id = GetConstantId(inst.type_id);
+
+    const auto& function = import_ir_.functions().Get(inst.function_id);
```
geoffromer, line 245: Another `auto` that isn't clear to me.

toolchain/check/import_ref.cpp
```diff
@@ -73,8 +73,25 @@ class ImportRefResolver {
...
+  static constexpr SemIR::NameScopeId NoEnclosingScopeForImports =
+      SemIR::NameScopeId::Invalid;
+
+  // Returns true if new unresolved constants were found.
+  //
+  // At the start of a function, do:
+  //   auto initial_work = work_stack_.size();
+  // Then when determining:
+  //   if (HasUnresolved(initial_work)) { ... }
+  auto HasUnresolved(size_t initial_work) -> bool {
```
geoffromer: At the callsite, `HasUnresolved(initial_work)` reads like it means "is any of the initial work still unresolved?". How about something like `HasNewUnresolvedSince`?

toolchain/check/import_ref.cpp
jonmeow: This flow is explained on TryResolveInst

toolchain/check/import_ref.cpp
jonmeow: See https://github.com/carbon-language/carbon-lang/pull/3703/files#r1489964250

toolchain/check/import_ref.cpp
jonmeow: Done.

toolchain/check/import_ref.cpp
jonmeow: has_unresolved is mentioned in the PR description, but I left this behind by mistake. Cleaned up.

toolchain/check/import_ref.cpp
jonmeow: See https://github.com/carbon-language/carbon-lang/pull/3703/files#r1489964250

toolchain/check/import_ref.cpp
jonmeow: I believe this has been discussed several times, e.g. here on #toolchain:
https://discord.com/channels/655572317891461132/655578254970716160/1169674453932851210

This is the coding style we've been using on the toolchain, and if you look, you'll find a lot of equivalent code. If you're in disagreement with this style, would you be okay bringing this up on either #style or #toolchain and proposing changes separate from this PR?

toolchain/check/import_ref.cpp

toolchain/check/import_ref.cpp
jonmeow: Sure, added local before all the functions, renamed to import_name_id.

toolchain/check/import_ref.cpp
jonmeow: Rephrased, but focusing more on the API (that it's used by the other function) than the implementation (that it's iterating through the refs and only grabs the type, which seems more like an implementation detail).

toolchain/check/import_ref.cpp
jonmeow: Renamed to just "HasNewWork"

toolchain/check/import_ref.cpp
geoffromer: I'm not seeing this change.

toolchain/check/import_ref.cpp
geoffromer: Let me put it a different way: the problem I'm trying to address is that I couldn't make sense of the body of `GetParamRefsId` except by also reading the body of `GetParamConstantIds`. That sort of coupling between the internals of separate functions makes the code substantially harder for me to read, so one of the main things I look for in API comments is that they enable me to reason locally about a single function body at a time.

I see what you're getting at, in the sense that this information is a sort of joint implementation detail of `GetParamConstantIds` and `GetParamRefsId` that can be hidden from the mutual caller that acts as an intermediary, but the contract between them still needs to be documented somewhere. There are ways of doing that while still hiding it from other functions (such as factoring out a separate class that it can be an internal invariant of), but I'm not sure that degree of complexity is necessary here.

toolchain/check/import_ref.cpp
geoffromer: I'm not trying to take a policy position on the toolchain codebase as a whole. All I'm saying is that this code was meaningfully easier for me to understand once I knew the type of this variable, but figuring out that type involved a digression of a couple of minutes or so while I dug through the layers of indirection. Does the toolchain style have a different way of solving that kind of problem?

jonmeow: Yes, inst_blocks.get returns an inst block. insts.get returns an inst. functions.get returns a function. classes.get returns a class. etc etc etc

Note, this is what I'm describing on the discord thread that I linked to. I understand you might not be trying to argue a policy position, but this is a codebase style we've been using and I think disagreements with the approach shouldn't be handled in isolation.

toolchain/check/import_ref.cpp
jonmeow: Sorry, my mistake -- fixed.

toolchain/check/import_ref.cpp
jonmeow: I've tried rephrasing this again, can you take another look?

toolchain/check/import_ref.cpp
geoffromer: OK, I've started a discussion in #toolchain.

chandlerc: FWIW, we had a long discussion about this there and in the toolchain sync.

I don't think we have a definitive conclusion, but two takeaways seem particularly relevant here:

1) I think the toolchain code should work to use `auto` or not in a reasonably consistent way. To that end, I think it makes sense to try to address this holistically rather than in cases that happen to show up in PRs. That's not because the cases that show up aren't real and don't cause real cost to the reviewers and readers, but just so we have a plan to be consistent going forward.

2) We're going to try to document the uses of `auto`. Jon has already added the first one to the toolchain doc, and we should add others where needed. As that happens, we can then evaluate that underlying pattern more holistically and decide whether we're comfortable keeping it or we should change the toolchain to use explicit types fairly consistently there.

I think (2) will ensure we don't lose track of this, and I don't think we need to block this review on sorting that out, so let's resolve this thread (and the two others below) and if we need to make a change do so with an update to the doc and working to systematically apply it across the code.

chandlerc (approved):
(Taking over this review as Geoff requested.)

LG overall.

I have some lingering questions about the exact representation we want for imported references, and how they in turn materialize in the SemIR... but I think it's too early to really worry about them. I'd rather get this and a bit more of the import logic in place, and then maybe it makes sense to chat about the overall structure and whether there are any tweaks that make sense?

The core architecture to have an iterative resolver all makes sense and seems important regardless of what tweaks we make to how we represent the end result.

PR #3705, opened 2024-02-14
Switch example code from Create to Make

jonmeow:
There's not explicit guidance about this, but there was an [old discussion](https://discord.com/channels/655572317891461132/963846118964350976/998010464237785189) that loosely settled on Make for now.


chandlerc (approved):
Sure, LG!

PR #3707, opened 2024-02-15
Fix for MakeUnqualifiedName conflicting with a namespace

jonmeow:
Unqualified names don't handle scopes the way that typical names do, so a name conflict with a namespace needs to be handled specially. I'm still favoring keeping code close as much as possible, particularly since long-term this syntax will probably shift to be more consistent. For now I'm just flagging when we shouldn't push scopes, so that MakeUnqualifiedName doesn't need to clean up.

Note, a different approach would basically be:

```
PushScopeAndStartName
ApplyNameQualifierTo
result = decl_name_stack_.back();
decl_name_stack_.back().state = NameContext::State::Finished;
PopScope
return result;
```

But that approach feels worse to me, due to the additional stack manipulations and the need to duplicate some of the FinishName logic just to be able to pop the scope that didn't really need to be added.

### DIFF
--- a/toolchain/check/testdata/var/fail_namespace_conflict.carbon
+++ b/toolchain/check/testdata/var/fail_namespace_conflict.carbon
@@ -6,27 +6,21 @@
 
 namespace A;
 
-// CHECK:STDERR: fail_namespace_conflict.carbon:[[@LINE+9]]:5: ERROR: Duplicate name being declared in the same scope.
-// CHECK:STDERR: var A: b;
+// CHECK:STDERR: fail_namespace_conflict.carbon:[[@LINE+6]]:5: ERROR: Duplicate name being declared in the same scope.
+// CHECK:STDERR: var A: i32;
 // CHECK:STDERR:     ^
 // CHECK:STDERR: fail_namespace_conflict.carbon:[[@LINE-5]]:1: Name is previously declared here.
 // CHECK:STDERR: namespace A;
 // CHECK:STDERR: ^~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


toolchain/check/testdata/var/fail_namespace_conflict.carbon
```diff
@@ -0,0 +1,54 @@
...
+// CHECK:STDERR: fail_namespace_conflict.carbon:[[@LINE+9]]:5: ERROR: Duplicate name being declared in the same scope.
+// CHECK:STDERR: var A: b;
+// CHECK:STDERR:     ^
+// CHECK:STDERR: fail_namespace_conflict.carbon:[[@LINE-5]]:1: Name is previously declared here.
+// CHECK:STDERR: namespace A;
+// CHECK:STDERR: ^~~~~~~~~~~~
+// CHECK:STDERR: fail_namespace_conflict.carbon:[[@LINE+3]]:8: ERROR: Name `b` not found.
+// CHECK:STDERR: var A: b;
+// CHECK:STDERR:        ^
+var A: b;
```
chandlerc: I'd prefer something that avoids an error on the type so its clear what this is trying to test. The type name lookup errors feel like noise in this context.

Happy for you to fix this here and submit w/o another round, or do this in a follow-up.

(Here and below as well of course.)

chandlerc (approved):
LG -- a comment inline to fix before submitting or to fix in a follow-up, whatever your preference is.

toolchain/check/testdata/var/fail_namespace_conflict.carbon
jonmeow: Done, verified it's not necessary for the crash.

PR #3709, opened 2024-02-17
Add ClassDecl/Type import functionality.

jonmeow:
I believe this PR is sufficient to pull in all current class features, including the current bits of inheritance which have been implemented. Because a class declaration can reference its own type, this creates an incomplete type prior to constant loading.

Right now, the object representation is imported proactively, but individual fields are left as ImportRefUnused. This means that member functions and similar will only be imported if called.

This also adjusts how function parameters are being handled, to match the expectations of Self param structure.

When formatting, I'm starting to look into constants. Otherwise we get "unexpected instref".

Overall, there are a few things that may be worth further discussion:

- The lack of a constant corresponding to the ClassType on ClassDecl is inconvenient -- I'd like to see how zygoloid feels about trying to restructure this. i.e., I'm setting a constant in order to be able to track things down later, it'd be nice if the normal IR did this simply for consistency, or if we were able to combine these rather than having separate instructions.
- Should we shift the parse node tracking further, and go with a setup wherein imports can embed import references into that? e.g., negative values go to another array which includes a ImportIRId for printing diagnostics, replacing the invalid NodeId.
- Can the formatter switch to a more general scan of instructions for naming, to eliminate the ImportRef constant approach added here?
- GetExprValueForLookupResult special-casing instructions felt surprising, I might see if there's a way to restructure to avoid that.

But I think these issues are things that can be separated out.


toolchain/check/handle_name.cpp
```diff
@@ -49,21 +49,41 @@ static auto GetExprValueForLookupResult(Context& context,
...
+                             ->insts()
+                             .Get(import_ref.inst_id)
+                             .kind();
+      if (import_kind == SemIR::ClassDecl::Kind ||
+          import_kind == SemIR::InterfaceDecl::Kind) {
+        return GetExprValueForLookupResult(
+            context, context.constant_values().Get(lookup_result_id).inst_id());
+      } else {
+        return lookup_result_id;
+      }
```
chandlerc, line 77: Any reason not to recurse immediately and regardless of `import_kind`, and rely on the recursing `GetExprValueForLookupResult` to diagnose any unsupported kind? For me, that would make the logic simpler of first recurse to the leaf IR, and then have a single place where we handle the supported leaves or diagnose others.

chandlerc: But maybe I've misunderstood how this is working? Let me know if so.

toolchain/check/import_ref.cpp
```diff
@@ -136,9 +142,51 @@ class ImportRefResolver {
     const auto& param_refs = import_ir_.inst_blocks().Get(param_refs_id);
     llvm::SmallVector<SemIR::InstId> new_param_refs;
     for (auto [ref_id, const_id] : llvm::zip(param_refs, const_ids)) {
-      new_param_refs.push_back(context_.AddInstInNoBlock(
-          {SemIR::ImportRefUsed{context_.GetTypeIdForTypeConstant(const_id),
-                                import_ir_id_, ref_id}}));
+      // Figure out the param structure. This echoes
+      // Function::GetParamFromParamRefId.
```
chandlerc, line 182: Not blocking at all, but curious if you have anything in mind to avoid the parallel code structure?

toolchain/check/import_ref.cpp
```diff
@@ -73,6 +74,11 @@ class ImportRefResolver {
   }
 
  private:
+  struct Work {
+    SemIR::InstId inst_id;
+    bool incomplete_type = false;
```
chandlerc: From reading the rest of the PR, I think I have some idea of how all this is used, but I think it's a somewhat hard to piece together. Could you add some comments, maybe on the `Work` type rather than just this field that try to pull together how this is used throughout the resolution?

toolchain/check/import_ref.cpp
```diff
@@ -224,6 +312,132 @@ class ImportRefResolver {
...
+    auto initial_work = work_stack_.size();
+
+    auto object_repr_const_id = GetLocalConstantId(import_class.object_repr_id);
+    auto base_const_id = import_class.base_id.is_valid()
+                             ? GetLocalConstantId(import_class.base_id)
+                             : SemIR::ConstantId::Invalid;
+
+    if (HasNewWork(initial_work)) {
+      return SemIR::ConstantId::Invalid;
+    }
```
chandlerc: A bit more comments to help explain the initial work logic here would help me understand the flow of the function I think.

toolchain/check/import_ref.cpp
```diff
@@ -224,6 +312,132 @@ class ImportRefResolver {
...
+
+    // Update the class info with the definition.
+    if (incomplete_type) {
+      CARBON_CHECK(!const_id.is_valid());
+      const_id = import_ir_constant_values_.Get(inst_id);
+    }
+    auto& new_class = context_.classes().Get(
```
chandlerc, line 384: Would it make sense to extract some of this logic into a helper function similar to `CreateIncompleteClass` and leave this to just wire things together?

toolchain/check/testdata/class/fail_todo_import_forward_decl.carbon
```diff
@@ -0,0 +1,104 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+// ARGS: compile --dump-raw-sem-ir --phase=check %s
```
chandlerc: Any particular reason to switch to the raw dump here?

toolchain/sem_ir/formatter.cpp
```diff
@@ -32,8 +32,9 @@ class InstNamer {
   enum class ScopeId : int32_t {
     None = -1,
     File = 0,
-    Constants = 1,
-    FirstFunction = 2,
+    ImportRef = 1,
```
chandlerc, line 35: I'm not seeing this used in the tests, am I misunderstanding how this works and it doesn't show up like `file` and such?

chandlerc (commented):
Some comments inline!

toolchain/check/import_ref.cpp
jonmeow: Most of my thoughts are along the lines of either flattening out `addr` (it's a bit of an odd approach right now, maybe bind names need to have more of a modifier concept) + making the presence of a bind name reliable (removing a conditional), or delving into the _other_ end and rewriting param handling to be more agnostic (but that's pretty substantial). It's also possible some combination of both would work best.

[added a TODO; note, I'm thinking about this, but "Figure out" is an implementation comment, not a TODO]

toolchain/check/testdata/class/fail_todo_import_forward_decl.carbon
jonmeow: Oops, this was debug. I'd just missed cleaning this up.

toolchain/sem_ir/formatter.cpp
jonmeow: It's visible in template constants. In something such as:

```
// CHECK:STDOUT: %B.ref.loc6: <function> = name_ref B, file.%import_ref [template = imports.%B]
```

The reason `imports.` is present there is due to this.

Note, there may be some better approach to this, as noted in the PR description ("Can the formatter switch to a more general scan of instructions for naming, to eliminate the ImportRef constant approach added here?").

Added an explicit TODO.

toolchain/check/handle_name.cpp
jonmeow: Added a comment. Note that in the "else" case it's returning the ImportRefUsed as the value, so a recursion is not equivalent logic.

toolchain/check/import_ref.cpp
jonmeow: Added some comments here, moved the comments on TryResolveInst up to the class level and elaborated on them.

toolchain/check/import_ref.cpp
jonmeow: I'm not adding anything specific here, but do the other comment changes address this?

toolchain/check/import_ref.cpp
jonmeow: Done

toolchain/check/import_ref.cpp
```diff
@@ -19,6 +19,33 @@ namespace Carbon::Check {
...
+//
+// Calling Resolve on an instruction operates in an iterative manner, tracking
+// Work items on work_stack_. At a high level, the loop is:
+//
+// 1. If Work has received a constant, it's considered resolved.
+//    - If incomplete_type is marked, resolve unconditionally.
+//    - The constant check avoids performance costs of deduplication on add.
+// 2. Resolve the instruction: (TryResolveInst/TryResolveTypedInst)
+//    A. For cases with incomplete types, when incomplete_type isn't marked:
+//      i. Make the incomplete type with associated constant.
```
chandlerc: I'm not sure I'm understanding this sentence...

Do you mean `Import the type with its associated constant`?

Something else?

toolchain/check/import_ref.cpp
```diff
@@ -19,6 +19,33 @@ namespace Carbon::Check {
...
+//        - If the imported type is incomplete, return the constant.
+//      ii. Mark incomplete_type.
+//    B. Gather all input constants.
+//      - Gathering constants directly adds unresolved values to work_stack_.
+//    C. If any need to be resolved (HasNewWork), return Invalid; this
+//       instruction needs two calls to complete.
+//    D. Build any necessary IR structures, and return the output constant.
+//    - For trivial cases with zero or one input constants, this may return
+//      a constant (if one, potentially Invalid) directly.
```
chandlerc, line 44: I think this is missing an indent?
```suggestion
//      - For trivial cases with zero or one input constants, this may return
//        a constant (if one, potentially Invalid) directly.
```

toolchain/check/import_ref.cpp
```diff
@@ -224,6 +333,145 @@ class ImportRefResolver {
...
+    auto base_const_id = import_class.base_id.is_valid()
+                             ? GetLocalConstantId(import_class.base_id)
+                             : SemIR::ConstantId::Invalid;
+
+    if (HasNewWork(initial_work)) {
+      return SemIR::ConstantId::Invalid;
+    }
+
+    // On the first pass, we build the incomplete type's constant above. On the
+    // second pass, we need to fetch it.
```
chandlerc: Maybe a bit more reminder of the connection between the specific bool variable and the passes?

```suggestion
    // On the first pass, we build the incomplete type's constant above. If we
    // get here on a subsequent pass we need to fetch the one we built in the
    // first pass.
```

toolchain/check/testdata/class/import_member_cycle.carbon
```diff
@@ -0,0 +1,154 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// ARGS: compile --dump-raw-sem-ir --phase=check %s
```
chandlerc: Another raw dump maybe for debugging?

toolchain/check/import_ref.cpp
```diff
@@ -73,6 +102,16 @@ class ImportRefResolver {
...
 
  private:
+  // A step in work_stack_.
+  struct Work {
+    // The instruction to work on.
+    SemIR::InstId inst_id;
+
+    // True if an incomplete type has been generated; only used by instructions
+    // with incomplete types.
+    bool incomplete_type = false;
```
chandlerc: What do you think about calling this `resolved_incomplete_type` to make it a bit more clear how it is used? See also the naming suggestion below.

Also happy with other names if a different approach would work.

toolchain/check/import_ref.cpp
```diff
@@ -156,34 +238,42 @@ class ImportRefResolver {
...
-  // when all input constants are ready, a common scenario is that some inputs
-  // will still be unresolved, and it'll return Invalid between phases. On the
-  // second call, all previously unready constants will have been resolved, so
-  // it should run to completion. As a consequence, it's important to reserve
-  // all expensive logic for the second phase; this in particular includes
-  // GetTypeIdForTypeConstant calls which do a hash table lookup.
-  //
   // TODO: Error is returned when support is missing, but that should go away.
-  auto TryResolveInst(SemIR::InstId inst_id) -> SemIR::ConstantId {
+  auto TryResolveInst(SemIR::InstId inst_id, bool incomplete_type)
```
chandlerc: If the `Work` member is renamed, I'd rename this to match (`resolved_incomplete_type` or whichever).

toolchain/check/import_ref.cpp
```diff
@@ -224,6 +333,145 @@ class ImportRefResolver {
...
+      const auto& base_class = context_.classes().Get(
+          context_.insts().GetAs<SemIR::ClassType>(base_inst_id).class_id);
+      new_scope.extended_scopes.push_back(base_class.scope_id);
+    }
+    CARBON_CHECK(new_scope.extended_scopes.size() ==
+                 old_scope.extended_scopes.size());
+  }
+
+  auto TryResolveTypedInst(SemIR::ClassDecl inst, SemIR::InstId inst_id,
+                           bool incomplete_type) -> SemIR::ConstantId {
```
chandlerc: And here, since we're also manipulating the member of `Work`, I wonder if there is a different name for the function variable to indicate the separation?

`already_resolved_incomplete_type` is awfully long.

Alternatively, could consider a different naming structure: `pre_incomplete_type` to indicate the first phase?

Or to keep `false`-by-default, maybe `type_second_pass`? And then these comments can remind the reader that the second pass doesn't need to create the incomplete type but does need to load. WDYT? Happy to chat about naming options if none of these work.

chandlerc (approved):
The multi-phase processing I think was what was tripping me up. Thanks for all the comments and improvements here, I think this makes sense to me now.

The change LG structurally. Some trivial stuff inline and a test fix, LGTM with those in place.

I've also left some comments discussing some different options for naming the bool used to track the two passes here. Happy for you to merge with whatever names seem best to you (even the current ones) and continue that discussion in Discord or with more folks to see if there is a better name. That's a somewhat minor tweak to maybe make things read a bit better in the future, but I don't want to block on it so approving. =]

toolchain/check/import_ref.cpp
jonmeow: Restructuring with "For most cases:" to make the intended overlap clear.

toolchain/check/import_ref.cpp
jonmeow: Rephrased as discussed.

toolchain/check/import_ref.cpp
jonmeow: Renamed to `made_incomplete_type`, hoping it helps (`resolved` just implies to me a more complete state)

toolchain/check/import_ref.cpp

toolchain/check/import_ref.cpp
jonmeow: As above, `made_incomplete_type`

`type_second_pass` isn't quite the semantic I was aiming for because we still might need to check the constant to see if the instruction was previously filled in for some other edge cases.

toolchain/check/import_ref.cpp

jonmeow (commented):
Per discussion, also added a struct cycle test.

toolchain/check/testdata/class/import_member_cycle.carbon
jonmeow: Done

toolchain/check/import_ref.cpp
```diff
@@ -19,6 +19,37 @@ namespace Carbon::Check {
...
+//
+// 1. If Work has received a constant, it's considered resolved.
+//    - If made_incomplete_type, resolve unconditionally.
+//    - The constant check avoids performance costs of deduplication on add.
+// 2. Resolve the instruction: (TryResolveInst/TryResolveTypedInst)
+//    - For most cases:
+//      A. For types which _can_ be incomplete, when not made_incomplete_type:
+//        i. Start by making an incomplete type to address circular references.
+//        ii. If the imported type is incomplete, return the constant.
+//        iii. Set made_incomplete_type.
```
chandlerc: WDYT:

```suggestion
//        iii. Otherwise, continue resolving but set made_incomplete_type.
```

chandlerc (approved):
Niiiiice!

PR #3717, opened 2024-02-22
Redeclaration support for `impl` declarations.

zygoloid:
Maintain a mapping from (self type, constraint) to `ImplId` on the `Impl` value store so that we can perform redeclaration lookup.

### DIFF
--- a/toolchain/sem_ir/impl.h
+++ b/toolchain/sem_ir/impl.h
@@ -45,29 +45,39 @@ struct Impl : public Printable<Impl> {
 
 // A collection of `Impl`s, which can be accessed by the self type and
 // constraint implemented.
-class ImplStore : private ValueStore<ImplId> {
+class ImplStore {
+  using Store = ValueStore<ImplId>;
+
  public:
   // Looks up the impl with this self type and constraint, or creates a new
   // `Impl` if none exists.
   // TODO: Handle parameters.
   auto LookupOrAdd(TypeId self_id, TypeId constraint_id) -> ImplId {
     auto [it, added] =
-        values_.insert({{self_id, constraint_id}, ImplId::Invalid});
+        lookup_.insert({{self_id, constraint_id}, ImplId::Invalid});
     if (added) {
       it->second =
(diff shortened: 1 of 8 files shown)


toolchain/sem_ir/impl.h
```diff
@@ -0,0 +1,75 @@
...
+ public:
+  // Looks up the impl with this self type and constraint, or creates a new
+  // `Impl` if none exists.
+  // TODO: Handle parameters.
+  auto LookupOrAdd(TypeId self_id, TypeId constraint_id) -> ImplId {
+    auto [it, added] =
+        values_.insert({{self_id, constraint_id}, ImplId::Invalid});
+    if (added) {
+      it->second =
+          ValueStore::Add({.self_id = self_id, .constraint_id = constraint_id});
```
jonmeow: Is the `ValueStore::` required? (will be moot with the inheritance change, I suppose)

toolchain/sem_ir/impl.h
```diff
@@ -0,0 +1,75 @@
...
+  // TODO: Handle control flow in the impl body, such as if-expressions.
+  InstBlockId body_block_id = InstBlockId::Invalid;
+
+  // The following members are set at the `}` of the impl definition.
+  bool defined = false;
+};
+
+// A collection of `Impl`s, which can be accessed by the self type and
+// constraint implemented.
+class ImplStore : private ValueStore<ImplId> {
```
jonmeow: Style: "All inheritance should be public. If you want to do private inheritance, you should be including an instance of the base class as a member instead."

https://google.github.io/styleguide/cppguide.html#Inheritance

You might also want a comment about doing this approach in order to restrict mutation (`Add` and `Set`)

jonmeow (approved):
LGTM. I've been thinking about whether we should do more things like `impl.h` for the value store variants, I support this direction.

I think the private inheritance should be switched to a member, but please just fix it before merging.

toolchain/sem_ir/impl.h
zygoloid: Moot.

toolchain/sem_ir/impl.h

PR #3718, opened 2024-02-22
Refactor DiagnosticAnnotationScope to remove private inheritance

jonmeow:
I noticed this when commenting on #3717; I think we can achieve similar results without the base-child class separation. Also using llvm::function_ref, but maybe I'm misunderstanding the reason for the template parameter.

### DIFF
--- a/toolchain/diagnostics/diagnostic_emitter.h
+++ b/toolchain/diagnostics/diagnostic_emitter.h
@@ -237,7 +237,7 @@ using NoTypeDeduction = std::type_identity_t<Arg>;
 
 }  // namespace Internal
 
-template <typename LocationT, typename AnnotateFn>
+template <typename LocationT>
 class DiagnosticAnnotationScope;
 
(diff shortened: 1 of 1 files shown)


zygoloid (commented):
The switch to `llvm::function_ref` here will lead to use-after-scope bugs for the intended use cases. The problem is that we want to support

```c++
DiagnosticAnnotationScope scope(emitter, [&] (...) { ... });
```

... where the lambda is a temporary. Using `function_ref` means we'll call the lambda after it's destroyed. So instead we're building a type-erased callable wrapper here so that we can store the lambda as a member of `scope`.

If we want to avoid the private inheritance here, we could add a wrapper class instead. I'll put together a PR for that.

zygoloid:
> If we want to avoid the private inheritance here, we could add a wrapper class instead. I'll put together a PR for that.

#3719 removes the private inheritance while keeping the type erasure.

PR #3719, opened 2024-02-22
Avoid private inheritance in DiagnosticAnnotationScope.

zygoloid:
The style guide doesn't permit private inheritance, so switch to a member instead. Keep storing the callable to ensure it lives long enough, but type-erase it using `function_ref` instead of a hand-rolled mechanism.

### DIFF
--- a/toolchain/diagnostics/diagnostic_emitter.h
+++ b/toolchain/diagnostics/diagnostic_emitter.h
@@ -282,8 +282,8 @@ class DiagnosticEmitter {
     // For the expected usage see the builder API: `DiagnosticEmitter::Build`.
     template <typename... Args>
     auto Emit() -> void {
-      for (auto annotator : emitter_->annotators_) {
-        annotator(*this);
+      for (auto annotate_fn : emitter_->annotate_fns_) {
+        annotate_fn(*this);
       }
       emitter_->consumer_->HandleDiagnostic(std::move(diagnostic_));
     }
@@ -356,8 +356,6 @@ class DiagnosticEmitter {
   }
 
  private:
-  using AnnotateFn = llvm::function_ref<auto(DiagnosticBuilder& builder)->void>;
-
   // Converts an argument to llvm::Any for storage, handling input to storage
   // type translation when needed.
   template <typename Arg>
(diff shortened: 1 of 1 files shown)


toolchain/diagnostics/diagnostic_emitter.h
```diff
@@ -356,30 +356,7 @@ class DiagnosticEmitter {
...
-    }
-    ~DiagnosticAnnotationScopeBase() {
-      CARBON_CHECK(emitter_->annotators_.back() == this);
-      emitter_->annotators_.pop_back();
-    }
-
-   private:
-    DiagnosticEmitter* emitter_;
-  };
+  using AnnotateFn = llvm::function_ref<auto(DiagnosticBuilder& builder)->void>;
```
jonmeow: Had you considered inlining this since it's only used once? It might also avoid confusion over different meaning of AnnotateFn versus the template.

toolchain/diagnostics/diagnostic_emitter.h
```diff
@@ -486,23 +463,20 @@ class ErrorTrackingDiagnosticConsumer : public DiagnosticConsumer {
 // given emitter. That function can annotate the diagnostic by calling
 // `builder.Note` to add notes.
 template <typename LocationT, typename AnnotateFn>
-class DiagnosticAnnotationScope
-    : private DiagnosticEmitter<LocationT>::DiagnosticAnnotationScopeBase {
-  using Base =
-      typename DiagnosticEmitter<LocationT>::DiagnosticAnnotationScopeBase;
+class DiagnosticAnnotationScope {
+ private:
+  using Emitter = DiagnosticEmitter<LocationT>;
```
jonmeow: You might consider something like making the private `using` public because it's a bit inconsistent with style to have two private sections (https://google.github.io/styleguide/cppguide.html#Class_Format, "The public section should be first" etc)

jonmeow (approved):
Looks good!

toolchain/diagnostics/diagnostic_emitter.h
zygoloid: Done.

toolchain/diagnostics/diagnostic_emitter.h
zygoloid: I don't think it makes sense to add undesirable things to the class's public API to work around a style rule. I've just duplicated this type into the two places it was used, but I think this was better before. *grumble grumble*.

PR #3721, opened 2024-02-22
Add diagnostics for `extend impl` misuse.

zygoloid:
To make the implementation simpler, make `PopWithParseNodeIf` return `pair<NodeId, optional<value>>` rather than `optional<pair<NodeId, value>>`. While wrapping the whole result in `optional` seems more principled, it's significantly harder to work with.


toolchain/check/node_stack.h
```diff
@@ -320,6 +299,30 @@ class NodeStack {
     return std::nullopt;
   }
 
+  // Pops the top of the stack and returns the parse_node and the ID if it is
+  // of the specified kind.
+  template <const Parse::NodeKind& RequiredParseKind>
+  auto PopWithParseNodeIf()
+      -> std::pair<Parse::NodeIdForKind<RequiredParseKind>,
+                   decltype(PopIf<RequiredParseKind>())> {
```
jonmeow, line 307: This switch makes me wonder, would it be more consistent to drop optional for both, using Invalid instead of nullopt? Does that work consistently or do we ever push Invalid in these cases?

Put differently, if we're tossing principle, maybe we should toss optional since it's just overhead. ;)

toolchain/check/handle_impl.cpp
```diff
@@ -99,7 +107,36 @@ static auto ExtendImpl(Context& context, Parse::AnyImplDeclId parse_node,
...
+    enclosing_scope.has_error = true;
+    return;
+  }
+
+  if (context.parse_tree().node_kind(self_type_node) ==
+      Parse::NodeKind::TypeImplAs) {
+    CARBON_DIAGNOSTIC(ExtendImplSelfAs, Error,
+                      "Cannot `extend` an `impl` with an explicit self type.");
+    CARBON_DIAGNOSTIC(ExtendImplSelfAsDefault, Note,
+                      "Remove the explicit `Self` type here.");
```
jonmeow: FWIW I kind of lean towards the "declare variables close to where they're used" in cases like this, due to the two-scope separation. (it means I wouldn't be looking to see if there are other spots this diagnostic is produced)

toolchain/check/handle_impl.cpp
```diff
@@ -99,7 +107,36 @@ static auto ExtendImpl(Context& context, Parse::AnyImplDeclId parse_node,
...
+      // recover.
+      if (auto self_as = context.parse_tree().ExtractAs<Parse::TypeImplAs>(
+              self_type_node)) {
+        diag.Note(self_as->type_expr, ExtendImplSelfAsDefault);
+      }
+      diag.Emit();
+    } else {
+      diag.Emit();
+      enclosing_scope.has_error = true;
+      return;
```
jonmeow, line 129: This return seems a little odd, why didn't you structure this as:

```
if (self_type_id == ...) {
  ... (dropping emit)
} else {
  enclosing_scope.has_error = true;
}
diag.Emit()
```

jonmeow (approved):

toolchain/check/handle_impl.cpp
zygoloid: The first arm of this `if` recovers from the error and does not return. I've flipped the arms around and changed the success / recovery case to be below the `if` rather than an `else` so this control flow should be clearer now.

toolchain/check/node_stack.h
zygoloid: I think the node stack asserts if you try to push `Invalid`, so we could do that. I can investigate in a follow-up if that's OK?

toolchain/check/handle_impl.cpp
zygoloid: Done.

toolchain/check/node_stack.h
jonmeow: Oh, sorry, yeah -- I agree a TODO is fine.

jonmeow:
Note this PR hasn't been passing tests.

PR #3722, opened 2024-02-23
Assign a constant to ClassDecl/InterfaceDecl for name references.

jonmeow:
By adding a constant to ClassDecl/InterfaceDecl, we're able to remove name reference special-casing. Use TryEvalInst on the Decl to generate the Type. For ClassDecl, then use the generated constant for self_type_id.

### DIFF
--- a/toolchain/check/handle_class.cpp
+++ b/toolchain/check/handle_class.cpp
@@ -56,18 +56,21 @@ static auto BuildClassDecl(Context& context, Parse::AnyClassDeclId parse_node)
   auto decl_block_id = context.inst_block_stack().Pop();
 
   // Add the class declaration.
-  auto class_decl = SemIR::ClassDecl{SemIR::ClassId::Invalid, decl_block_id};
+  auto class_decl = SemIR::ClassDecl{SemIR::TypeId::TypeType,
+                                     SemIR::ClassId::Invalid, decl_block_id};
   auto class_decl_id = context.AddPlaceholderInst({parse_node, class_decl});
 
   // Check whether this is a redeclaration.
   auto existing_id =
       context.decl_name_stack().LookupOrAddName(name_context, class_decl_id);
+  auto const_id = SemIR::ConstantId::Invalid;
   if (existing_id.is_valid()) {
     if (auto existing_class_decl =
             context.insts().Get(existing_id).TryAs<SemIR::ClassDecl>()) {
       // This is a redeclaration of an existing class.
       class_decl.class_id = existing_class_decl->class_id;
       auto& class_info = context.classes().Get(class_decl.class_id);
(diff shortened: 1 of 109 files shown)


toolchain/check/handle_class.cpp
```diff
@@ -105,11 +108,19 @@ static auto BuildClassDecl(Context& context, Parse::AnyClassDeclId parse_node)
...
+    // Provide the ClassType for name references.
+    // TODO: This may need to differ for generic types, but should still take
+    // name lookup into account.
+    const_id = self_const_id;
   }
 
   // Write the class ID into the ClassDecl.
-  context.ReplaceInstBeforeConstantUse(class_decl_id, {parse_node, class_decl});
+  context.ReplaceInstBeforeConstantUse(class_decl_id, {parse_node, class_decl},
+                                       const_id);
```
zygoloid: I would prefer that we create the class type in `TryEvalInst` rather than bypassing constant evaluation here. `constant_values` is currently a cache for the result of the constant evaluator and I think it's valuable that we keep it that way. We can still have the evaluator pull the constant out of the class's `self_type_id` rather than redoing `ClassType` canonicalization, and the small extra cost per class declaration seems fine to me.

toolchain/check/handle_interface.cpp
```diff
@@ -63,6 +63,8 @@ static auto BuildInterfaceDecl(Context& context,
     if (auto existing_interface_decl =
             context.insts().Get(existing_id).TryAs<SemIR::InterfaceDecl>()) {
       // This is a redeclaration of an existing interface.
+      // TODO: class_decl = *existing_class_decl;
+      interface_decl.type_id = existing_interface_decl->type_id;
```
zygoloid: I don't understand this change: don't we always want the type to be `TypeType`? I would think the only thing we need to update here is the `interface_id`.

zygoloid (commented):
This looks nice.

toolchain/check/handle_interface.cpp
jonmeow: Sorry, missed cleaning this up (leftover from while editing)

toolchain/check/handle_class.cpp
jonmeow: I was thinking about this, and how about this approach instead, also dropping GetClassType?

zygoloid (approved):
Looks great, thanks!

(Not in this PR, but I'm wondering if we can change the existing uses of `self_type_id` to use the constant value of the class declaration, and remove that field from the `Class` entirely.)

toolchain/check/handle_class.cpp
zygoloid: Thanks, this is great.

PR #3723, opened 2024-02-26
Remove obsolete `__init__.py` files

clavin:
I noticed that there are some empty `__init__.py` files in the repo, with no immediately clear reason to have them. I asked [in Discord](https://discord.com/channels/655572317891461132/655578254970716160/1211088334894534686) and it looks like these are likely artifacts from when `lit` was used for testing, now left over and obsolete from the migration. To keep the repo tidy, this PR deletes the files.


jonmeow (approved):
Thanks!

PR #3724, opened 2024-02-26
Remove self_type_id from Class

jonmeow:
This relies more on the constant built for ClassDecl's new constant to provide class information, but as a consequence, does more type lookups for the constant.


jonmeow:
Note, I'm going to respond a little in #toolchain since I think you'd expected only Self to need to resolve the type; I'm not sure this PR makes sense to merge but I thought it'd be helpful to send it.

PR #3725, opened 2024-02-26
Split classes in file.h and value_stores.* to separate files

jonmeow:
I'm proposing a different split, along the line of "what does this relate to". I view impl.h as having started down this route. Moving the inst store stuff to inst.h feels odd to me given how much else is there right now, but maybe it's still the best approach. Some files only contain a store, no structured class, but I felt the consistency in file naming (without _store suffixes) might help.


zygoloid (approved):
Awesome, thank you!

PR #3726, opened 2024-02-26
Add support for importing interfaces.

jonmeow:
Interface support is pretty skeletal so this may need additions later, but I think it's still worthwhile to fill in the necessary bits now. With this change, the expectation is then that everything we have right now which _can_ be imported, is supported for import (at least for the "current package, no overlap" case).


chandlerc (approved):
LG, especially as its largely a starting point skeleton.

PR #3729, opened 2024-02-27
Support for name lookup into interfaces.

zygoloid:
Follow the existing support for classes.

None of this is especially useful until other features land: we don't yet have any use for defining methods of an interface out of line, because we don't support `default` or `final` interface methods, and we don't have impl lookup, so referring to an interface member is also not especially useful. But this is a nice piece to factor out that's a prerequisite for effectively testing other interface features.


toolchain/check/decl_name_stack.cpp
```diff
@@ -245,6 +260,18 @@ auto DeclNameStack::TryResolveQualifier(NameContext& name_context,
             context_->classes().Get(class_decl->class_id).self_type_id);
         context_->NoteIncompleteClass(class_decl->class_id, builder);
         builder.Emit();
+      } else if (auto interface_decl = context_->insts()
+                                           .Get(name_context.resolved_inst_id)
+                                           .TryAs<SemIR::InterfaceDecl>()) {
```
```suggestion
                                           .TryGetAs<SemIR::InterfaceDecl>(name_context.resolved_inst_id)) {
```

toolchain/check/decl_name_stack.cpp
```diff
@@ -245,6 +260,18 @@ auto DeclNameStack::TryResolveQualifier(NameContext& name_context,
...
+        CARBON_DIAGNOSTIC(
+            QualifiedDeclInUndefinedInterfaceScope, Error,
+            "Cannot declare a member of undefined interface `{0}`.",
```
jonmeow, line 267: Why undefined here versus incomplete for the equivalent class diagnostic?

toolchain/check/decl_name_stack.cpp
```diff
@@ -245,6 +260,18 @@ auto DeclNameStack::TryResolveQualifier(NameContext& name_context,
...
+            SemIR::NameId);
+        auto builder = context_->emitter().Build(
+            name_context.parse_node, QualifiedDeclInUndefinedInterfaceScope,
+            context_->interfaces().Get(interface_decl->interface_id).name_id);
```
```suggestion
            std::string);
        auto builder = context_->emitter().Build(
            name_context.parse_node, QualifiedDeclInUndefinedInterfaceScope,
            context.sem_ir().StringifyTypeExpr(name_context.resolved_inst_id));
```

Why NameId here but TypeId for the equivalent class diagnostic? Could this be something more like thea bove?

jonmeow (approved):
Looks good, just some small optional suggestions.

toolchain/check/decl_name_stack.cpp
zygoloid: For a class type, after `class C;` you can't write `let c: C`, but for an interface, after `interface I;` you can write `let T:! I;` -- an interface becomes complete after a declaration whereas a class only becomes complete after a definition. So "incomplete" wouldn't be correct terminology for an interface. We could use "undefined" in both cases, although I think "incomplete" is a more familiar term for classes coming from C++.

jonmeow: It may be nice to find a consistent term... "undefined" might be better for consistency, but it has an awkward [plain[er]-English] meaning to say a name is undefined.

jonmeow: (I'm maybe less swayed by C++ here because I think it might help to give people the same term for when members can be defined for both)

jonmeow: Maybe: "Interface `{0}` must be defined before defining out-of-line members." Or something like that.

toolchain/check/decl_name_stack.cpp
zygoloid: Sure. We need to look up the constant value of the interface decl, though, since `StringifyTypeExpr` can't stringify an `InterfaceDecl`.

toolchain/check/decl_name_stack.cpp
zygoloid: Done, and likewise for the class case.

toolchain/check/decl_name_stack.cpp
zygoloid: The diagnostic has a note attached to it which I think helps clarify the meaning:

```console
foo.carbon:5:1: ERROR: Cannot declare a member of undefined interface `Undefined`.
fn Undefined.F();
   ^~~~~~~~~
foo.carbon:3:1: Interface was forward declared here.
interface Undefined;
^~~~~~~~~~~~~~~~~~~~
```

PR #3730, opened 2024-02-27
Add an associated entity instruction and corresponding type for interface elements.

zygoloid:
When declaring an associated entity in an interface -- just associated functions for now -- create an associated entity value and corresponding type to represent a "slot in a witness table". Also track the list of associated entities on the interface so that we will eventually be able to check impls against them.

Associated entities are represented as the integer index of their slot in a witness table.


toolchain/sem_ir/interface.h
```diff
@@ -20,6 +20,12 @@ struct Interface : public Printable<Interface> {
   // until we reach the `}` of the interface definition.
   auto is_defined() const -> bool { return defined; }
 
+  // Determines whether we're currently defining the interface. This is true
+  // between the braces of the interface.
+  auto is_being_defined() const -> bool {
```
jonmeow, line 25: FWIW, I was considering "partially defined" for this state. But I was also mulling doing equivalent for other `definition_id.is_valid()` checks (class and fn), so maybe the `!defined` check is extra we don't generally want? *shrug*

toolchain/sem_ir/interface.h
```diff
@@ -20,6 +20,12 @@ struct Interface : public Printable<Interface> {
...
+    return definition_id.is_valid() && !defined;
```
jonmeow:
```suggestion
    return definition_id.is_valid() && !is_defined();
```

Just to enforce the relationship.

toolchain/check/decl_name_stack.cpp
```diff
@@ -5,7 +5,9 @@
 #include "toolchain/check/decl_name_stack.h"
 
 #include "toolchain/check/context.h"
+#include "toolchain/check/scope_index.h"
```
jonmeow: Mirroring includes from the header?

toolchain/lower/handle.cpp
```diff
@@ -60,6 +60,12 @@ auto HandleAssign(FunctionContext& context, SemIR::InstId /*inst_id*/,
   context.FinishInit(storage_type_id, inst.lhs_id, inst.rhs_id);
 }
 
+auto HandleAssociatedEntity(FunctionContext& context, SemIR::InstId inst_id,
+                            SemIR::AssociatedEntity /*inst*/) -> void {
+  context.SetLocal(inst_id,
+                   llvm::ConstantStruct::getAnon(context.llvm_context(), {}));
```
jonmeow: Should there be lowering tests?

toolchain/sem_ir/inst.h
```diff
@@ -343,6 +343,16 @@ class InstStore {
...
   }
 
+  // Returns the requested instruction as the specified type, if it is valid and
+  // of that type. Otherwise returns nullopt.
+  template <typename InstT>
+  auto TryGetAsIfValid(InstId inst_id) const -> std::optional<InstT> {
+    if (!inst_id.is_valid()) {
+      return std::nullopt;
+    }
+    return Get(inst_id).TryAs<InstT>();
```
jonmeow:
```suggestion
    return TryGetAs<InstT>(inst_id);
```

?

toolchain/sem_ir/interface.h
```diff
@@ -42,6 +48,8 @@ struct Interface : public Printable<Interface> {
 
   // The following members are set at the `}` of the interface definition.
   bool defined = false;
+  // The declarations of the associated entities.
+  InstBlockId associated_entities_id = InstBlockId::Invalid;
```
jonmeow: Is this always set? If so, can `defined` be removed? If not, maybe the comment should note it?

toolchain/check/testdata/impl/impl_forall.carbon
```diff
@@ -16,19 +16,23 @@ impl forall [T:! type] T as Simple {
...
 // CHECK:STDOUT:   %.1: type = interface_type @Simple [template]
+// CHECK:STDOUT:   %.2: type = assoc_entity_type @Simple, <function> [template]
+// CHECK:STDOUT:   %.3: <associated <function> in Simple> = assoc_entity element0, @Simple.%F [template]
 // CHECK:STDOUT: }
 // CHECK:STDOUT:
 // CHECK:STDOUT: file {
 // CHECK:STDOUT:   package: <namespace> = namespace {.Simple = %Simple.decl} [template]
 // CHECK:STDOUT:   %Simple.decl = interface_decl @Simple, () [template = constants.%.1]
-// CHECK:STDOUT:   impl_decl @impl, (<unexpected instref inst+4>, <unexpected instref inst+5>, <unexpected instref inst+6>, <unexpected instref inst+7>)
+// CHECK:STDOUT:   impl_decl @impl, (<unexpected instref inst+7>, <unexpected instref inst+8>, <unexpected instref inst+9>, <unexpected instref inst+10>)
```
jonmeow, line 26: I think these "unexpected instref" should be fixed because this is a non-fail test.

jonmeow (approved):
Please at least fix the "unexpected instref", that looks like it should be fixed here. Other than that, generally looks good, let me know if you want another round of review.

toolchain/check/decl_name_stack.cpp
zygoloid: VSCode did it, I dunno why. Undone.

toolchain/check/testdata/impl/impl_forall.carbon
zygoloid: I agree. Will take a look after this lands (they're not new here or related to this change at all).

toolchain/lower/handle.cpp
zygoloid: Added, and in fact this didn't work :) Fixed.

toolchain/sem_ir/interface.h
zygoloid: Sure, done.

PR #3731, opened 2024-02-28
Format the decl block for a declaration as a trailing block.

zygoloid:
Previously we used the default inst block formatting, which writes out a parenthesized list of references, which would always be 'unexpected instref's because nothing else prints the instructions in the decl block.

In addition, track the decl block for function declarations like we do for other kinds of declaration. This means that the parameter declarations for a function are now properly rendered into the formatted IR. Note that this adds a lot of verbosity to `function_decl`, but it does accurately reflect the IR, and we'll probably want this information to be printed once we start supporting more complex generic function declarations.

This removes almost all the 'unexpected instref's in our formatted output. There are remaining cases when a declarative scope contains multiple blocks, where we only track one of those blocks. That happens when there is control flow within declarative scopes, and for error recovery when a class or interface or similar is defined more than once.

As requested in #3730.


jonmeow (approved):
Looks good. I agree more verbose, but seems like the right path.

PR #3732, opened 2024-02-28
Make namespace formatting more consistent with other declaration block instructions.

zygoloid:
List names vertically rather than horizontally.


josh11b (approved):

toolchain/sem_ir/formatter.cpp
```diff
@@ -1073,8 +1075,11 @@ class Formatter {
   auto FormatArg(ElementIndex index) -> void { out_ << index; }
 
   auto FormatArg(NameScopeId id) -> void {
-    out_ << '{';
-    FormatNameScope(id, ", ", "");
+    out_ << "{\n";
```
jonmeow, line 1111: Should this be inserting the `\n` even for an empty name scope?

zygoloid: I have a follow-up PR to address this.

PR #3734, opened 2024-02-29
If a SemIR instruction has a braced block and a constant value, print the constant value first.

zygoloid:


toolchain/sem_ir/formatter.cpp
```diff
@@ -924,6 +918,31 @@ class Formatter {
     out_ << "\n";
   }
 
+  auto FormatPendingConstantValue(bool space_before) -> void {
```
jonmeow: I was trying to figure out at call sites, is this saying there _is_ a space before the call, or that a space should be inserted _before_ the call?

I think as written, it's being told whether to add a space:

```suggestion
  auto FormatPendingConstantValue(bool needs_space_before) -> void {
```

But consider flipping the bool so that it'd be "has_space_before", since "has" is a typical form.

jonmeow: Should this have comments explaining its semantics? e.g., to me it's subtle that it should take care of only printing the constant once if called multiple times, or that it does the space balancing.

toolchain/sem_ir/formatter.cpp
```diff
@@ -1262,6 +1281,8 @@ class Formatter {
   bool in_terminator_sequence_ = false;
   int indent_ = 0;
   bool after_open_brace_ = false;
+  ConstantId pending_constant_value_ = ConstantId::NotConstant;
```
jonmeow, line 1306: Would it be worth starting to add comments to explain what these members are?

toolchain/sem_ir/formatter.cpp
```diff
@@ -924,6 +918,31 @@ class Formatter {
     out_ << "\n";
   }
 
+  auto FormatPendingConstantValue(bool space_before) -> void {
+    if (!pending_constant_value_.is_valid() ||
+        pending_constant_value_.is_constant()) {
```
jonmeow: I'll note the irony here of changing an invalid value to not constant, because the assignment is set up to run unconditionally. That makes the `is_constant` consistently run on the second pass, even for initially invalid values. Had you considered an early return before assigning, dropping the indent? :)

toolchain/sem_ir/formatter.cpp
jonmeow, line 1306: Why NotConstant instead of Invalid?

jonmeow (approved):

jonmeow:
Looks good, sorry for the nits.

toolchain/sem_ir/formatter.cpp
zygoloid: Changed to an enum, hopefully this clarifies.

zygoloid: Added comment.

toolchain/sem_ir/formatter.cpp
zygoloid: Changed to early return, added comment to explain the use of `NotConstant` as a sentinel, and simplified the condition to a direct comparison against `NotConstant` to make it more obvious that's the value we're excluding here.

toolchain/sem_ir/formatter.cpp
zygoloid: Done.

zygoloid: Hopefully the new comment clarifies.

PR #3735, opened 2024-02-29
Fix up the C++ snippet to reflect the carbon snippet

PJBoy:
Offers a more precise apples-to-apples language comparison. Replaces the non-portable `M_PI` and removes the extraneous arg parameters. It does use C++23, so I can understand if that's unacceptable for the time being.

### DIFF
--- a/docs/images/snippets.md
+++ b/docs/images/snippets.md
@@ -55,7 +55,7 @@ struct Circle {
 
 void PrintTotalArea(std::span<Circle> circles) {
   std::float32_t area = 0;
-  for (Circle c : circles) {
+  for (const Circle& c : circles) {
     area += std::numbers::pi * c.r * c.r;
   }
   std::print("Total area: {}\n", area);
@@ -101,8 +101,9 @@ fn Main() -> i32 {
 
 ```cpp
 // C++ code used in both Carbon and C++:
+import std;
 struct Circle {
-  float r;
+  std::float32_t r;
 };
 
 // Carbon exposing a function for C++:
@@ -119,9 +120,9 @@ fn PrintTotalArea(circles: Slice(Cpp.Circle)) {
 }
 
 // C++ calling Carbon:
-#include <vector>
-#include "circle.h"
-#include "geometry.carbon.h"
+import std;
+import "circle.h";
+import "geometry.carbon.h";
 
 auto main() -> int {
   std::vector<Circle> circles = {{1.0}, {2.0}};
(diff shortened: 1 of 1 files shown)


docs/images/snippets.md
```diff
@@ -47,26 +47,23 @@ fn QuickSort[T:! Comparable & Movable](s: Slice(T)) {
...
-  float r;
+  std::float32_t r;
 };
 
 void PrintTotalArea(std::span<Circle> circles) {
-  float area = 0;
-  for (const Circle& c : circles) {
-    area += M_PI * c.r * c.r;
+  std::float32_t area = 0;
+  for (Circle c : circles) {
```
zygoloid: I think the `const&` here is a slightly better match for the Carbon code, though neither is exactly right -- the Carbon code doesn't force a copy to be made, but it has value semantics rather than reference semantics.

jonmeow:
Note, the CLA is required. But glancing at the change, I don't feel it's an unambiguous choice; I'll let one of the leads decide whether they want to adopt these changes (note, the images would also need to be updated).

docs/images/snippets.md
```diff
@@ -126,7 +123,7 @@ fn PrintTotalArea(circles: Slice(Cpp.Circle)) {
 #include "circle.h"
 #include "geometry.carbon.h"
```
zygoloid, line 130: It'd be good to consistently use `import` here too, at least in place of `#include <vector>`.

If you wanted to update the other includes, we don't have rules yet for how to map Carbon libraries to C++ module names, but `import Carbon:Geometry;` seems like a good guess to use as a placeholder. Or you could use `import "geometry.carbon.h";` for now.

docs/images/snippets.md
PJBoy: Thanks for the feedback. Agree that `const&` is a more faithful approximation

docs/images/snippets.md
PJBoy: I have opted for the latter choice. I felt the former choice would be more appropriate if we wrote circle.h to export a module (say `export module circle;`), but then I didn't want to guess what the consuming Carbon code would look like (`import Cpp module circle;`?)

zygoloid (commented):
Sorry for the latency here. @chandlerc and I have talked about this a bit, and we'd like to hold off on using `import` in the C++ examples for now. While we eventually will want to write this example this way, we think the majority of C++ developers aren't yet able to use `import std;`, so using that in the examples is probably going to be a distraction more than it's helpful.

What do you think about changing the examples to something more like:

```c++
#include <numbers>
#include <print>
#include <span>
#include <stdfloat>
#include <vector>
// or: import std;
```

for the time being?

zygoloid (approved):
Thank you!

PR #3737, opened 2024-03-01
Basic support for associated constant declarations

zygoloid:


toolchain/parse/state.def
```diff
@@ -1265,6 +1265,10 @@ CARBON_PARSE_STATE(Let)
 //         ^~~~~
 //   1. Expr
 //
+// let ... ;
+//        ^
+//   (state done)
+//
```
```suggestion
```

Small suggestion, to merge the cases. Does make me wonder if it'd help to be a bit more specific about what "..." is though.

toolchain/parse/typed_nodes.h
```diff
@@ -345,8 +345,12 @@ struct LetDecl {
...
   llvm::SmallVector<AnyModifierId> modifiers;
   AnyPatternId pattern;
-  LetInitializerId equals;
-  AnyExprId initializer;
+
+  struct Initializer {
+    LetInitializerId equals;
+    AnyExprId initializer;
+  };
+  std::optional<Initializer> initializer;
```
jonmeow, line 353: It might be worth updating the comments at https://github.com/carbon-language/carbon-lang/blob/trunk/toolchain/parse/node_kind.def#L321-L323 to indicate the initializer is now optional, similar to how VariableDecl does.

toolchain/sem_ir/formatter.cpp
```diff
@@ -1203,6 +1203,11 @@ class Formatter {
   auto FormatArg(InstId id) -> void { FormatInstName(id); }
 
   auto FormatArg(InstBlockId id) -> void {
+    if (!id.is_valid()) {
+      out_ << "invalid";
```
jonmeow, line 1211: Minor, but I think printing `id` here would result in `<invalid>`, which maybe is worth the `<>` bracketing / standardized print?

```suggestion
      out_ << id;
```

toolchain/sem_ir/inst_kind.def
```diff
@@ -23,6 +23,7 @@ CARBON_SEM_IR_INST_KIND(ArrayIndex)
 CARBON_SEM_IR_INST_KIND(ArrayInit)
 CARBON_SEM_IR_INST_KIND(ArrayType)
 CARBON_SEM_IR_INST_KIND(Assign)
+CARBON_SEM_IR_INST_KIND(AssociatedConstantDecl)
```
jonmeow, line 26: As you're adding this, would it be worth adding a todo test for import? i.e., in file A, declare an interface with an associated constant; then in file B, reference the interface (e.g., with an alias). The result should be that the interface members, including the associated constant, try to resolve; I'd expect that to currently generate a TODO (from import_ref, without code changes).

toolchain/check/testdata/interface/fail_assoc_const_not_constant.carbon
```diff
@@ -0,0 +1,32 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+
+interface I {
+  // CHECK:STDERR: fail_assoc_const_not_constant.carbon:[[@LINE+3]]:7: ERROR: Pattern in associated constant declaration must be a single `:!` binding.
+  // CHECK:STDERR:   let a: i32;
+  // CHECK:STDERR:       ^
+  let a: i32;
+}
```
jonmeow, line 12: As part of this, you poison the name scope. Would it be worth testing the result?

```suggestion
  // CHECK:STDERR: fail_assoc_const_not_constant.carbon:[[@LINE+3]]:7: ERROR: Pattern in associated constant declaration must be a single `:!` binding.
  // CHECK:STDERR:   let a: i32;
  // CHECK:STDERR:       ^
  let a: i32;
}

// Should fail silently.
alias Invalid = I.Nonexistent;
```

jonmeow, line 12: As part of this, you poison the name scope. Would it be worth adding a failed name lookup?

```suggestion
  // CHECK:STDERR: fail_assoc_const_not_constant.carbon:[[@LINE+3]]:7: ERROR: Pattern in associated constant declaration must be a single `:!` binding.
  // CHECK:STDERR:   let a: i32;
  // CHECK:STDERR:       ^
  let a: i32;
}

// All name lookups fail silently due to the invalid name.
alias Invalid = I.Nonexistent;
```

jonmeow (approved):
Looks good! Happy to see this moving forward.

toolchain/check/testdata/interface/fail_assoc_const_not_constant.carbon
zygoloid: Done. I had a couple of thoughts here:

- I'm testing both the name `a` and an unrelated name, because I think it would be fine (and probably desirable) if we started rejecting `I.Nonexistent`, but we shouldn't start rejecting uses of `I.a`.
- It'll be easy to overlook a future change that makes this start producing errors, and we can't fix that by making this a non-`fail_` test. It seems like we don't yet have a good way to test error recovery. No particular suggestions for how to improve that, but mentioning it in case you have an idea.

toolchain/parse/state.def
zygoloid: Done. (I had kept them apart to avoid merging the valid case and the error case, but we don't do that anywhere else and looking at it again I think it's not helpful to keep these separate.)

The wider context (not visible in the diff) gives more clarity about what the `...` is -- the state is called `LetAfterPattern` and the comment on it says "Handles `let` after the pattern", but I've updated the comment to mention the initializer being optional.

toolchain/parse/typed_nodes.h
zygoloid: Done.

toolchain/sem_ir/formatter.cpp
zygoloid: Generally the formatter tries not to print out raw IDs, because they're not meaningful to a reader of the formatted IR, so I think using the ID type's printing machinery in this case would make us marginally less consistent. I also think the formatter should be owning the details of the IR format, even for invalid IR, and not delegating.

toolchain/sem_ir/inst_kind.def
zygoloid: Done.

Issue #2477, opened 2022-12-17
Matching forward declaration of functions and classes


josh11b:
### Description of the bug:

- Question-for-leads issue #472 and proposal #875 established that Carbon will use forward declarations.
- Question-for-leads issue #1132 described the general rules for how to match forward declarations with their definitions.
- Proposal #1084 implemented the decision in #1132 to allow forward declaration of interfaces and implementations.
- Proposal #722 and PR #1026 added support for forward declaring classes
- PR #1017 explained how to forward declare functions

What is missing is an update to the description of function and class declarations to include the matching rules from #1132. In particular, that they are matched by name and if a declaration and definition have matching names but are different in other details, then an error should be reported at compile time.

It isn't super clear to me if making these changes should be a proposal or just a PR.

### Link to documentation:

- [forward declaration of classes](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/classes.md#forward-declaration)
- [forward declaration of functions](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/functions.md#function-declarations)
- [Declarations, Definitions, and Scopes in the overview](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/README.md#declarations-definitions-and-scopes)

Issue #1132, opened 2022-03-15
How do we match forward declarations with their definitions?


josh11b:
With the decision on #472 , Carbon has separate forward declarations from definitions and we need some way to match them up. This comes up for functions, classes, interfaces, named constraints, interface implementations, and probably others.

There are a three main components to this question:
1. What parts of a declaration are used to connect a forward declaration to its definition?
2. What constitutes a match for any given part?
3. Once we have identified that a declaration should be associated with a particular definition, what additional checking must be done to make sure they are consistent?

zygoloid:
What agreement rule do we want for the case where the forward declaration and definition are in different scopes -- specifically, when checking whether a declaration of a member function in a class (or a defaulted function in an interface or a member of an impl) agrees with a definition written outside that class? For example:

```carbon
class A {
  fn F();
}

let T:! type = A;
// OK?
fn T.F() {}

class B(T:! type) {
  fn F();
}
constraint Type {}

// OK?
fn B(T:! Type).F() {}
```

I suggest we require the part before the `.` to agree with the class declaration, in the same way that we require the part after the `.` to agree with the function declaration, so both of the above definitions are rejected -- and in general, the first part of `fn <tokens that identify a class>.F()` is allowed if and only if `class <tokens that identify a class>;` would be allowed. That seems to be in line with what's proposed here, but @josh11b and I weren't certain the prior decision covered this.

chandlerc:
I agree.

PR #3739, opened 2024-03-04
Expand handling of function redeclarations.

jonmeow:
This detects more cases of incorrect matches between declaration and definition. It also factors the logic out to a separate file for easier sharing, particularly when it comes to merging imports (I'm not sure if this exact API will be reshared, but the core logic should apply).


toolchain/check/function.cpp
```diff
@@ -0,0 +1,214 @@
...
+          .Emit();
+      return false;
+    }
+  }
+  return true;
+}
+
+// Returns true if the provided function signatures are equal, in the sense that
+// declarations can be merged.
+static auto IsFunctionSignatureEqual(Context& context,
```
zygoloid: I think it'd be useful to use the "match" / "agree" terminology from the design here (https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/generics/details.md#matching-and-agreeing). Generally, when we say two declarations "match", we mean they're similar enough that they must either declare the same thing or be an error, and when we say they "agree" we mean that they're similar enough that they validly declare the same thing. So I think what we're checking here is whether two function signatures agree. (For now we're saying function declarations match if they have the same name; maybe overloading will change that, or maybe not.)

jonmeow: Done, although noting that documentation, it might've been helpful to provide an explanation of what it means to agree but not match. It *sounds* like the only difference is that agreeing is simply "names are optional but if present must match"  versus matching being "names must match" [and other than that, requirements are identical], but maybe I'm missing something?

jonmeow: I think we're requiring matching, since param names must match. [also, the doc says: "Carbon needs to determine if two declarations match in order to say which definition a forward declaration corresponds to and to verify that nothing is defined twice."]

jonmeow: Also, "MatchParam" because I can't think of a good name format that would allow for "Match" and "Agree" to be interchangeable. "MatchParam" is maybe "ParamAgrees"; "IsParamMatch" would be "DoesParamAgree".

zygoloid: See also #1132 and #2477. We don't have a design update for functions yet, but the current direction is:

- Two function declarations match if they have the same name.
- Two function declarations agree if they have the same declaration syntax with names resolving to the same thing, except that parameter names can be omitted, parentheses can be added / removed if they don't change anything, and aliases can be expanded in one but not both declarations.

And the general rule is that if two declarations match, then it's an error if they don't agree.

(I don't think we need to get all the corner case details right in this PR; it's probably better to iterate there once this is in place.)

But the thing we're checking here is agreement, not matching.

toolchain/check/function.cpp
```diff
@@ -0,0 +1,216 @@
...
+  if (new_param_ref.Is<SemIR::AnyBindName>()) {
+    new_param_ref =
+        context.insts().Get(new_param_ref.As<SemIR::AnyBindName>().value_id);
+    prev_param_ref =
+        context.insts().Get(prev_param_ref.As<SemIR::AnyBindName>().value_id);
+  }
+
+  CARBON_CHECK(new_param_ref.Is<SemIR::Param>()) << new_param_ref;
+  CARBON_CHECK(prev_param_ref.Is<SemIR::Param>()) << prev_param_ref;
+
```
zygoloid: Do these produce better diagnostics than the ones we get from `As`? If so, would it make sense to improve the CHECK there instead of adding one here?

toolchain/check/function.cpp
```diff
@@ -0,0 +1,216 @@
...
+      new_param_ref.type_id() != prev_param_ref.type_id()) {
+    return false;
+  }
+
+  if (new_param_ref.Is<SemIR::AddrPattern>()) {
+    new_param_ref =
+        context.insts().Get(new_param_ref.As<SemIR::AddrPattern>().inner_id);
+    prev_param_ref =
+        context.insts().Get(prev_param_ref.As<SemIR::AddrPattern>().inner_id);
+  }
```
zygoloid: Hm, I'm torn. I want to suggest using `TryAs` here, but that'd make the two statements in the `if` body less similar. I wonder if it makes sense to get rid of the early `kind` check and instead use

```suggestion
  if (auto new_addr = new_param_ref.TryAs<SemIR::AddrPattern>(),
         prev_addr = prev_param_ref.TryAs<SemIR::AddrPattern>();
       new_addr && prev_addr) {
    new_param_ref = context.insts().Get(new_addr.inner_id);
    prev_param_ref = context.insts().Get(prev_addr.inner_id);
  }
```

What do you think? In part I'm thinking that we'll eventually allow the `unused` wrapper to differ, so the early kind check might not be desirable in general.

(I'm also fine with this as-is, just for your consideration.)

toolchain/check/function.cpp
```diff
@@ -0,0 +1,216 @@
...
+  CARBON_CHECK(new_param_ref.Is<SemIR::Param>()) << new_param_ref;
+  CARBON_CHECK(prev_param_ref.Is<SemIR::Param>()) << prev_param_ref;
+
+  auto new_param = new_param_ref.As<SemIR::Param>();
+  auto prev_param = prev_param_ref.As<SemIR::Param>();
+  return new_param.name_id == prev_param.name_id;
+}
+
+// Returns true if two param refs agree.
+static auto DoParamAgrees(Context& context, SemIR::InstId new_decl_id,
```
zygoloid:
```suggestion
static auto DoParamsAgree(Context& context, SemIR::InstId new_decl_id,
```

I think? (I'd read "DoParamAgrees" as "do the 'param-agrees' process", which might be accurate but seems a bit awkward.)

But given that this is producing diagnostics rather than just computing a `bool`, I think naming this so it doesn't sound like a predicate might be better. Maybe:

```suggestion
static auto CheckParamsAgree(Context& context, SemIR::InstId new_decl_id,
```

toolchain/check/function.cpp
```diff
@@ -0,0 +1,216 @@
...
+  // This will often occur for empty params.
+  if (new_param_refs_id == prev_param_refs_id) {
+    return true;
+  }
+  const auto new_param_ref_ids = context.inst_blocks().Get(new_param_refs_id);
+  const auto prev_param_ref_ids = context.inst_blocks().Get(prev_param_refs_id);
+  if (new_param_ref_ids.size() != prev_param_ref_ids.size()) {
+    CARBON_DIAGNOSTIC(
+        FunctionSignatureParamCountDisagree, Error,
+        "Function declaration with {0}{1} parameter(s) disagrees.", int32_t,
```
zygoloid: I think our choice of terminology isn't helping us present a clear diagnostic here. Maybe we can massage it a bit:

"Function redeclaration with {0}{1} parameter(s) does not agree with previous declaration."
"Previous matching declaration has {0}{1} parameter(s)."

... but really the match / agree terminology doesn't seem to be working well here. :(

If we keep the current terminology, I think we should always be saying "[dis]agrees *with something*" not just abstractly "[dis]agrees" -- it should be a relation between two things.

[I also really don't like the `(s)`, but that's not new in this PR.]

toolchain/check/function.cpp
```diff
@@ -0,0 +1,216 @@
...
+  auto& prev_function = context.functions().Get(prev_function_id);
+
+  // TODO: Disallow redeclarations within classes?
+  if (!DoesFunctionSignatureAgree(context, new_function, prev_function)) {
+    return false;
+  }
+
+  if (!is_definition) {
+    CARBON_DIAGNOSTIC(FunctionRedeclaration, Error,
+                      "Redeclaration of function {0}.", SemIR::NameId);
```
zygoloid:
```suggestion
                      "Redundant redeclaration of function {0}.", SemIR::NameId);
```
I think this'll make it a little clearer why this is an error.

zygoloid (approved):

toolchain/check/function.cpp
jonmeow: Removed as redundant. Not investigating diagnostics.

jonmeow: Removed as redundant, I don't think I looked at diagnostics, this was just not cleaned up.

toolchain/check/function.cpp
jonmeow: I feel weird about this. As discussed, my sense is we need to flatten params more, but I'm not sure what the path we'll take is. Maybe we should just move addr to Param now? But I'm hesitant to make this particular change due to the inconsistency, and partly because I see it as doing two TryAs even if the first fails.

toolchain/check/function.cpp
jonmeow: Done

toolchain/check/function.cpp
jonmeow: Rephrasing, avoiding `(s)`

toolchain/check/function.cpp
jonmeow: This was some sloppy naming, but how about just `CheckRedecl` + `Param` + `s` as naming. :)

zygoloid: That name works for me, but I think it'd be better to have more difference between the names of `CheckRedeclParams` and `CheckRedeclParam` to reflect that one of them is producing diagnostics and the other is just performing a comparison. (Or maybe make them both produce diagnostics per the FIXME.)

zygoloid (approved):

toolchain/check/function.cpp
jonmeow: Moved (not really addressing the FIXME, I don't think the structure is worth addressing)

zygoloid (approved):

PR #3740, opened 2024-03-05
Add `Self` into scope in interface definitions and using facets as types.

zygoloid:
`Self` is modeled as a `bind_symbolic_name` with no corresponding value, for now at least. In the future it might make sense to model it as a new kind of instruction, or as a `bind_symbolic_name` whose value is a `param`, but for now we just want it to introduce a symbolic constant.

In order to convert a value like the `Self` of an interface to a type, a new instruction `facet_type_access` is introduced. This notionally accesses the "type" field within a facet, converting it from a pair of (type, witness) into just the type.


toolchain/check/convert.cpp
```diff
@@ -833,6 +833,19 @@ static auto PerformBuiltinConversion(Context& context, Parse::NodeId parse_node,
...
+
+    // Facet type conversions: a value T of facet type F1 can be implicitly
+    // converted to facet type F2 if T satisfies the requirements of F2.
+    //
+    // TODO: Support this conversion in general. For now we only support it in
+    // the case where F1 is an interface type and F2 is `type`.
+    // TODO: Support converting tuple and struct values to facet types,
+    // combining the above conversions and this one in a single conversion.
+    if (auto src_interface_type =
+            sem_ir.types().TryGetAs<SemIR::InterfaceType>(value_type_id)) {
```
jonmeow: Why does this do TryGetAs when it doesn't use src_interface_type, vs Is? Is this about the TODOs? Still looks kind of odd to give it a name.

```suggestion
    if (sem_ir.types().TryGetAs<SemIR::InterfaceType>(value_type_id)) {
```

```suggestion
    if (sem_ir.types().GetAsInst(value_type_id).Is<SemIR::InterfaceType>()) {
```

toolchain/check/handle_interface.cpp
```diff
@@ -126,14 +128,34 @@ auto HandleInterfaceDefinitionStart(
...
+        context.constant_values().Get(interface_decl_id));
+
+    // We model `Self` as a symbolic binding whose type is the interface.
+    interface_info.self_param_id = context.AddInst(
+        {Parse::NodeId::Invalid,
+         SemIR::BindSymbolicName{
+             self_type_id,
+             context.bind_names().Add(
+                 {.name_id = SemIR::NameId::SelfType,
+                  .enclosing_scope_id = interface_info.scope_id}),
```
jonmeow: Soft suggestion, but breaking this out to a local might make the AddInst easier to read.

toolchain/check/handle_interface.cpp
```diff
@@ -126,14 +128,34 @@ auto HandleInterfaceDefinitionStart(
...
+             SemIR::InstId::Invalid}});
```
jonmeow: This feels odd that it doesn't have an id, could we use context.constant_values().Get(interface_decl_id).inst_id()?

toolchain/check/handle_interface.cpp
```diff
@@ -126,14 +128,34 @@ auto HandleInterfaceDefinitionStart(
...
+  // Declare and introduce `Self`.
+  if (!interface_info.is_defined()) {
+    // TODO: Once we support parameterized interfaces, this won't be the right
+    // type. For `interface X(T:! type)`, the type of `Self` is `X(T)`, whereas
+    // this will be simply `X`.
+    auto self_type_id = context.GetTypeIdForTypeConstant(
+        context.constant_values().Get(interface_decl_id));
+
+    // We model `Self` as a symbolic binding whose type is the interface.
+    interface_info.self_param_id = context.AddInst(
```
jonmeow: Should import_ref.cpp be population this now too? [maybe a TODO?]

jonmeow (approved):
Looks good!

toolchain/check/convert.cpp
zygoloid: Added `TypeStore::Is`.

toolchain/check/handle_interface.cpp
zygoloid: Done.

toolchain/check/handle_interface.cpp
zygoloid: The `inst_id` on a `BindSymbolicName` is a value that the symbolic binding will eventually be equal to. For the `Self` of an interface, we don't know what that is. It's not the interface type, though.

As mentioned in the overall PR comment, I'm not certain whether we will want to use a `BindSymbolicName` like this in the long term, but this seems like a reasonable path forwards for now.

toolchain/check/handle_interface.cpp
zygoloid: Added a TODO. While investigating, I found the existing import_ref support for importing `BindName` and `BindSymbolicName` doesn't work, so I added a couple more TODOs there.

PR #3742, opened 2024-03-06
Offer GetTypeIdForTypeInst

jonmeow:
#3740 added another similar call. This may be temporary, maybe we'll stop using and remove later, but it's a small simplification right now and we might also retain a similar inst -> constant -> type flow.


zygoloid (approved):
I'm slightly hesitant here because both of the existing callers are wrong / temporary (though we only have the TODO on one of them): once we add generics, we'll want to form a different type for `Self` than the type of the declaration. But we can easily remove this again if we find it has no users after fixing that.

zygoloid:
(Removed from queue because we probably want a different commit message.)

PR #3743, opened 2024-03-06
Initial scaffolding for building a witness table for an `impl`.

zygoloid:
Add an instruction to hold the witness table, along with a corresponding type to keep things simpler. Add `check/impl.{h,cpp}` to house the new logic. No checking of impls against interfaces is performed yet.


toolchain/check/impl.h
```diff
@@ -0,0 +1,18 @@
...
+
+#ifndef CARBON_TOOLCHAIN_CHECK_IMPL_H_
+#define CARBON_TOOLCHAIN_CHECK_IMPL_H_
+
+#include "toolchain/check/context.h"
+#include "toolchain/sem_ir/ids.h"
+
+namespace Carbon::Check {
+
+// Builds and returns a witness table for the impl `impl_id`.
```
jonmeow: FWIW, right now you return an InterfaceWitness, which I assume would contain the table, but I'd tend to say that it's returning a witness (or very precisely "InterfaceWitness") rather than a witness table. When it says "returns a witness table", I'd expect just that. Not sure if this is something that'll change, but might be worth thinking about how much precision to use in the comments.

toolchain/sem_ir/formatter.cpp
```diff
@@ -747,7 +747,18 @@ class Formatter {
       out_ << ' ';
       OpenBrace();
       FormatCodeBlock(impl_info.body_block_id);
-      FormatNameScope(impl_info.scope_id, "!members:\n");
+
+      // Always include the !members label because we always list the witness in
+      // this section.
```
```suggestion
      // Print the !members label even if the name scope is empty because we always list the witness in
      // this section.
```

Suggestion to clarify what "Always include" is talking about.

jonmeow (approved):
LGTM

toolchain/check/impl.h
zygoloid: Eventually this will be a different kind of witness, once we support the more general  `impl Type as Constraint`, not merely `impl Type as Interface`, but it'll always be a witness, so I'll just drop the "table".

zygoloid:
```suggestion
// Builds and returns a witness for the impl `impl_id`.
```

toolchain/sem_ir/formatter.cpp
```diff
@@ -747,7 +747,18 @@ class Formatter {
       out_ << ' ';
       OpenBrace();
       FormatCodeBlock(impl_info.body_block_id);
-      FormatNameScope(impl_info.scope_id, "!members:\n");
+
+      // Print the !members label even if the name scope is empty because we always list the witness in
+      // this section.
```
zygoloid:
```suggestion
      // Print the !members label even if the name scope is empty because we
      // always list the witness in this section.
```

PR #3746, opened 2024-03-07
Use suggestion-bot to convert pre-commit errors in PRs with fixes into code review suggestions.

zygoloid:
This can't be done directly from a `pull_request` action, because that's run without write privileges. This is done for security reasons, because it runs in the context of the pull request branch. So instead, we perform this in two steps:

- The `pull_request` action runs `pre-commit` and uploads an artifact containing the diffs and the event information (which is only used to extract the pull request number).
- A separate `workflow_run` action is triggered when the `pre-commit` action finishes. This action is privileged, and should be able to download the artifact and create corresponding suggestions.

Unfortunately, due to the permissions model in play here, the second half of this appears to only be testable live in production.

For now, we're splitting the pre-commit action into two actions -- one to run on PRs and one to run when actually merging commits -- so that the suggestions are only triggered in the former case. It might be possible to recombine these using data in the `workflow_run` invocation to tell them apart, but the documentation here isn't very good so I've made this PR dump out that event information so that we can look at it and see if it contains the relevant information.

### DIFF
--- a/.github/workflows/pre_commit.yaml
+++ b/.github/workflows/pre_commit.yaml
@@ -6,9 +6,6 @@ name: pre-commit
 
 on:
   pull_request:
-  merge_group:
-  push:
-    branches: [trunk]
 
 jobs:
   pre-commit:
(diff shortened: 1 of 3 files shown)


.github/workflows/pre_commit.yaml
```diff
@@ -10,10 +10,22 @@ on:
...
       - uses: pre-commit/action@v3.0.0
+      - uses: actions/setup-node@v1
+        if: ${{ failure() && github.event_name == 'pull_request' }}
+        with:
+          node-version: 18
+      - name: Suggest edits from pre-commit
+        if: ${{ failure() && github.event_name == 'pull_request' }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: git diff | yarn suggestion-bot
```
zygoloid, line 31: https://github.com/tido64/suggestion-bot/issues/279

jonmeow (approved):

PR #3747, opened 2024-03-07
Augment the file_test framework to allow per-file fail checks.

jonmeow:
This handles toolchain failures per-file. The intent is to allow placing both "success" and "fail" tests in the same file, using splits. However, this PR only adds support and updates existing tests to continue passing.

### DIFF
--- a/toolchain/driver/driver.cpp
+++ b/toolchain/driver/driver.cpp
@@ -333,13 +333,13 @@ auto Driver::ParseArgs(llvm::ArrayRef<llvm::StringRef> args, Options& options)
       [&](CommandLine::CommandBuilder& b) { options.Build(b); });
 }
 
-auto Driver::RunCommand(llvm::ArrayRef<llvm::StringRef> args) -> bool {
+auto Driver::RunCommand(llvm::ArrayRef<llvm::StringRef> args) -> RunResult {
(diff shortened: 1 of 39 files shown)


toolchain/driver/driver.cpp
```diff
@@ -644,31 +646,36 @@ auto Driver::Compile(const CompileOptions& options) -> bool {
       }
     }
   });
+  RunResult result = {.success = true};
   for (const auto& input_file_name : options.input_file_names) {
     units.push_back(
         std::make_unique<CompilationUnit>(this, options, input_file_name));
+    result.per_file_success[input_file_name] = true;
```
zygoloid: I wonder if it'd be better to represent this as an array of {filename, success} pairs instead of a string map -- a duplicate filename might be passed to the driver, so using filename as the primary key seems suspect. An array would also be lower overhead than a stringmap for non-test uses. (Using a string map for the file_test version still makes sense, though, I think.)

zygoloid (approved):

toolchain/driver/driver.cpp
jonmeow: Done.

zygoloid (approved):

PR #3750, opened 2024-03-07
Unify pre-commit actions and attempt to fix suggestion-bot

zygoloid:


.github/workflows/pre_commit_suggestions.yaml
```diff
@@ -16,42 +24,22 @@ permissions:
...
+          name: pre-commit output
+          run-id: ${{ github.event.workflow_run.id }}
 
+      # Use https://github.com/tido64/suggestion-bot to create PR suggestions
+      # matching the diff that pre-commit created.
       - name: Create suggestions
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: GITHUB_EVENT_PATH=~/output/event npx suggestion-bot ~/output/diff
+        run: GITHUB_EVENT_PATH=./event cat ./diff | npx suggestion-bot -f -m "pre-commit checks made changes to the following files:"
```
jonmeow: Can you get a wrap here

.github/workflows/pre_commit_suggestions.yaml
```diff
@@ -16,42 +24,22 @@ permissions:
 
 jobs:
   pull-request-suggestions:
-    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
+    # Only generate suggestions if pre-commit for a PR failed.
+    # prettier-ignore
```
jonmeow: FWIW I would've just let it reflow it (I think we do that elsewhere)

.github/workflows/pre_commit_suggestions.yaml
```diff
@@ -16,42 +24,22 @@ permissions:
...
+    if: >
```
jonmeow: Why is this > versus |?

jonmeow (approved):

PR #3755, opened 2024-03-08
Add framework for the extern keyword.

jonmeow:
This doesn't add full support. I'm separating it out to make the effects of the modifier changes clearer for review. I'm restructuring a little with the expectation that we'll have some more categories of modifier keywords in the future (similar to `extern`, these may not be in a "set" such as access), and thus easily scaling up to a few more would be useful.


zygoloid (approved):
LG with pre-commit fix applied.

PR #3756, opened 2024-03-08
Form the contents of the witness table.

zygoloid:
Check that functions in an `impl` are declared properly. For now, simply reuse the redeclaration checking logic. Longer-term, we need to check that the signatures match in a more approximate way, after substituting the `Self` type into the signature in the `interface`.

### DIFF
--- a/toolchain/check/impl.cpp
+++ b/toolchain/check/impl.cpp
@@ -5,13 +5,88 @@
 #include "toolchain/check/impl.h"
 
 #include "toolchain/check/context.h"
+#include "toolchain/check/function.h"
 #include "toolchain/sem_ir/ids.h"
 #include "toolchain/sem_ir/impl.h"
+#include "toolchain/sem_ir/typed_insts.h"
 
 namespace Carbon::Check {
 
+// Checks that `impl_function_id` is a valid implementation of the function
+// described in the interface as `interface_function_id`. Returns the value to
+// put into the corresponding slot in the witness table, which can be
+// `BuiltinError` if the function is not usable.
+static auto CheckAssociatedFunctionImplementation(
+    Context& context, SemIR::FunctionId interface_function_id,
+    SemIR::InstId impl_decl_id) -> SemIR::InstId {
+  auto decl_node = context.insts().GetParseNode(impl_decl_id);
+  auto impl_function_decl =
+      context.insts().TryGetAs<SemIR::FunctionDecl>(impl_decl_id);
+  if (!impl_function_decl) {
+    context.TODO(decl_node, "diagnose non-function implementing function");
+    return SemIR::InstId::BuiltinError;
+  }
+
+  // TODO: Substitute the `Self` from the `impl` into the type in the interface
+  // before checking. Also, this should be a semantic check rather than a
+  // syntactic one. The functions should be allowed to have different signatures
+  // as long as we can synthesize a suitable thunk.
+  if (!CheckFunctionRedecl(context, impl_function_decl->function_id,
+                           interface_function_id)) {
+    return SemIR::InstId::BuiltinError;
+  }
+  return impl_decl_id;
+}
+
+// Builds a witness that the specified impl implements the given interface.
+static auto BuildInterfaceWitness(
+    Context& context, const SemIR::Impl& impl, SemIR::InterfaceId interface_id,
+    llvm::SmallVectorImpl<SemIR::InstId>& used_decl_ids) -> SemIR::InstId {
+  const auto& interface = context.interfaces().Get(interface_id);
+  if (!interface.is_defined()) {
+    context.TODO(context.insts().GetParseNode(impl.definition_id),
+                 "impl of non-defined interface");
+    return SemIR::InstId::BuiltinError;
+  }
+
+  auto& impl_scope = context.name_scopes().Get(impl.scope_id);
+
+  llvm::SmallVector<SemIR::InstId, 32> table;
+  auto assoc_entities = context.inst_blocks().Get(interface.associated_entities_id);
+  table.reserve(assoc_entities.size());
+
+  for (auto decl_id : assoc_entities) {
+    auto decl = context.insts().Get(decl_id);
+    if (auto fn_decl = decl.TryAs<SemIR::FunctionDecl>()) {
+      auto& fn = context.functions().Get(fn_decl->function_id);
+      auto impl_decl_id = context.LookupNameInExactScope(fn.name_id, impl_scope);
+      if (impl_decl_id.is_valid()) {
+        used_decl_ids.push_back(impl_decl_id);
+        table.push_back(CheckAssociatedFunctionImplementation(
+            context, fn_decl->function_id, impl_decl_id));
+      } else {
+        context.TODO(context.insts().GetParseNode(impl.definition_id),
+                     "diagnose missing decl in impl");
+        table.push_back(SemIR::InstId::BuiltinError);
+      }
+    } else if (auto const_decl = decl.TryAs<SemIR::AssociatedConstantDecl>()) {
+      // TODO: Check we have a value for this constant in the constraint.
+      context.TODO(context.insts().GetParseNode(impl.definition_id),
+                   "impl of interface with associated constant");
+      return SemIR::InstId::BuiltinError;
+    } else {
+      CARBON_FATAL() << "Unexpected kind of associated entity " << decl;
+    }
+  }
+
+  auto table_id = context.inst_blocks().Add(table);
+  return context.AddInst(SemIR::InterfaceWitness{
+      context.GetBuiltinType(SemIR::BuiltinKind::WitnessType), interface_id,
+      table_id});
+}
+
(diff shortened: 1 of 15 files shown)


toolchain/check/impl.cpp
```diff
@@ -5,11 +5,120 @@
...
 #include "toolchain/check/context.h"
+#include "toolchain/check/function.h"
+#include "toolchain/diagnostics/diagnostic_emitter.h"
 #include "toolchain/sem_ir/ids.h"
 #include "toolchain/sem_ir/impl.h"
+#include "toolchain/sem_ir/typed_insts.h"
 
 namespace Carbon::Check {
 
+static auto NoteAssociatedFunction(Context& context,
```
jonmeow, line 17:
```suggestion
// Adds the location of the associated function to a diagnostic.
static auto NoteAssociatedFunction(Context& context,
```

toolchain/check/impl.cpp
```diff
@@ -24,14 +133,14 @@ auto BuildImplWitness(Context& context, SemIR::ImplId impl_id)
     return SemIR::InstId::BuiltinError;
   }
 
-  auto interface_id = interface_type->interface_id;
+  llvm::SmallVector<SemIR::InstId, 32> used_decl_ids;
```
jonmeow: What do you think of moving the "32" out to a constant to document why this value (also versus using SmallVector's default)? I see it's also used in `table`, so maybe at the file level.

toolchain/check/impl.cpp
```diff
@@ -5,11 +5,120 @@
...
+    builder.Emit();
+
+    return SemIR::InstId::BuiltinError;
+  }
+
+  // TODO: Substitute the `Self` from the `impl` into the type in the interface
+  // before checking. Also, this should be a semantic check rather than a
+  // syntactic one. The functions should be allowed to have different signatures
+  // as long as we can synthesize a suitable thunk.
+  if (!CheckFunctionRedecl(context, impl_function_decl->function_id,
```
jonmeow, line 52: To the extent that we're planning to add syntactic checking to redecl checking, is this an API constraint that should be noted on the CheckFunctionRedecl comments? It might also help us avoid forgetting if this is something like `CheckFunctionRedeclSemanticOnly` or such.

toolchain/check/impl.cpp
```diff
@@ -5,11 +5,120 @@
...
+        used_decl_ids.push_back(impl_decl_id);
+        table.push_back(CheckAssociatedFunctionImplementation(
+            context, fn_decl->function_id, impl_decl_id));
+      } else {
+        CARBON_DIAGNOSTIC(
+            ImplMissingFunction, Error,
+            "Missing implementation of {0} in impl of interface {1}.",
+            SemIR::NameId, SemIR::NameId);
+        auto builder = context.emitter().Build(
+            context.insts().GetParseNode(impl.definition_id),
```
jonmeow:
```suggestion
            impl.definition_id,
```

toolchain/check/impl.cpp
```diff
@@ -5,11 +5,120 @@
...
+static auto BuildInterfaceWitness(
+    Context& context, const SemIR::Impl& impl, SemIR::InterfaceId interface_id,
+    llvm::SmallVectorImpl<SemIR::InstId>& used_decl_ids) -> SemIR::InstId {
+  const auto& interface = context.interfaces().Get(interface_id);
+  if (!interface.is_defined()) {
+    CARBON_DIAGNOSTIC(ImplOfUndefinedInterface, Error,
+                      "Implementation of undefined interface {0}.",
+                      SemIR::NameId);
+    auto builder = context.emitter().Build(
+        context.insts().GetParseNode(impl.definition_id),
```
jonmeow:
```suggestion
        impl.definition_id,
```

toolchain/check/impl.cpp
```diff
@@ -5,11 +5,120 @@
...
+    Context& context, SemIR::FunctionId interface_function_id,
+    SemIR::InstId impl_decl_id) -> SemIR::InstId {
+  auto impl_function_decl =
+      context.insts().TryGetAs<SemIR::FunctionDecl>(impl_decl_id);
+  if (!impl_function_decl) {
+    CARBON_DIAGNOSTIC(ImplFunctionWithNonFunction, Error,
+                      "Associated function {0} implemented by non-function.",
+                      SemIR::NameId);
+    auto builder = context.emitter().Build(
+        context.insts().GetParseNode(impl_decl_id), ImplFunctionWithNonFunction,
```
jonmeow:
```suggestion
        impl_decl_id, ImplFunctionWithNonFunction,
```

toolchain/check/impl.cpp
```diff
@@ -5,11 +5,120 @@
...
 
 namespace Carbon::Check {
 
+static auto NoteAssociatedFunction(Context& context,
+                                   Context::DiagnosticBuilder& builder,
+                                   SemIR::FunctionId function_id) -> void {
+  CARBON_DIAGNOSTIC(ImplAssociatedFunctionHere, Note,
+                    "Associated function {0} declared here.", SemIR::NameId);
+  const auto& function = context.functions().Get(function_id);
+  builder.Note(context.insts().GetParseNode(function.decl_id),
```
jonmeow:
```suggestion
  builder.Note(function.decl_id,
```

I think these cast?

jonmeow (approved):
LGTM with fixes.

toolchain/check/impl.cpp
```diff
@@ -5,11 +5,121 @@
...
 
+// Adds the location of the associated function to a diagnostic.
+static auto NoteAssociatedFunction(Context& context,
+                                   Context::DiagnosticBuilder& builder,
+                                   SemIR::FunctionId function_id) -> void {
+  CARBON_DIAGNOSTIC(ImplAssociatedFunctionHere, Note,
+                    "Associated function {0} declared here.", SemIR::NameId);
+  const auto& function = context.functions().Get(function_id);
+  builder.Note(function.decl_id,
+               ImplAssociatedFunctionHere, function.name_id);
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  builder.Note(function.decl_id, ImplAssociatedFunctionHere, function.name_id);
```

toolchain/check/impl.cpp
```diff
@@ -5,11 +5,121 @@
...
+    Context& context, const SemIR::Impl& impl, SemIR::InterfaceId interface_id,
+    llvm::SmallVectorImpl<SemIR::InstId>& used_decl_ids) -> SemIR::InstId {
+  const auto& interface = context.interfaces().Get(interface_id);
+  if (!interface.is_defined()) {
+    CARBON_DIAGNOSTIC(ImplOfUndefinedInterface, Error,
+                      "Implementation of undefined interface {0}.",
+                      SemIR::NameId);
+    auto builder = context.emitter().Build(
+        impl.definition_id,
+        ImplOfUndefinedInterface, interface.name_id);
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
        impl.definition_id, ImplOfUndefinedInterface, interface.name_id);
```

toolchain/check/impl.cpp
```diff
@@ -5,11 +5,121 @@
...
+        table.push_back(CheckAssociatedFunctionImplementation(
+            context, fn_decl->function_id, impl_decl_id));
+      } else {
+        CARBON_DIAGNOSTIC(
+            ImplMissingFunction, Error,
+            "Missing implementation of {0} in impl of interface {1}.",
+            SemIR::NameId, SemIR::NameId);
+        auto builder = context.emitter().Build(
+            impl.definition_id,
+            ImplMissingFunction, fn.name_id, interface.name_id);
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
        auto builder =
            context.emitter().Build(impl.definition_id, ImplMissingFunction,
                                    fn.name_id, interface.name_id);
```

toolchain/check/impl.cpp
zygoloid: I think we'll want to move fully away from reusing the function redeclaration checking here -- I see this as being a short-term placeholder only. Longer-term, I think what we want to do is to generate a thunk that has the substituted signature, and in its body call the function in the interface (with an optimization for the case where they happen to exactly match). So for now, using the exact redeclaration checking logic, as it evolves, seems fine to me, and I'd rather not add complexity to it that we're not going to need down the road.

toolchain/check/impl.cpp
zygoloid: The default (12) is probably fine, and I'd imagine it'll cover >90% of all cases. Removed the explicit argument.

ger1964jou:
1. Uefi

PR #3759, opened 2024-03-08
Update to checkout v4

jonmeow:
Noticed due to https://github.com/carbon-language/carbon-lang/actions/runs/8207272518

```
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
```

I'm expecting we can just update the version number on these.

### DIFF
--- a/.github/workflows/pre_commit.yaml
+++ b/.github/workflows/pre_commit.yaml
@@ -14,7 +14,7 @@ jobs:
   pre-commit:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


zygoloid (approved):

zygoloid:
Looking at the actions/checkout changelog, the v3->v4 incompatibility is just the dependency on a new version of node, so... LG :-)

PR #3760, opened 2024-03-08
Rename parse_node -> node_id

jonmeow:
This was previously discussed at https://discord.com/channels/655572317891461132/655578254970716160/1209975051588210729. I'm initiating this mainly because we typically use "id" suffixes to indicate an `IdBase` being passed around and the non-id suffix of `parse_node` suggests at it carrying more data than it actually does. There used to be more reason for avoiding `node_id` because `SemIR::InstId` used to be named `NodeId`, but that's no longer necessary. As a consequence, I'd like to rename `parse_node` to more precisely reflect its type.

In full, this is doing:

```
parse_node_kind -> node_kind
parse_node -> node_id
ParseNodeCategory -> NodeCategory
ParseNodeKind -> NodeKind
ParseNode -> NodeId
```

This is primarily in check and sem_ir, but with some `parse_node_kind` references in parse too.

Pluralization is consistent with name forms on both sides, so that wasn't part of my replacements.


toolchain/check/node_stack.h
```diff
@@ -385,8 +378,8 @@ class NodeStack {
 
   // Translate a parse node category to the enum ID kind it should always
   // provide, if it is consistent.
-  static constexpr auto ParseNodeCategoryToIdKind(Parse::NodeCategory category,
-                                                  bool for_node_kind)
+  static constexpr auto NodeIdCategoryToIdKind(Parse::NodeCategory category,
```
geoffromer: Should this be `NodeCategoryToIdKind`?

jonmeow: Yes, fixed. Also verified there are no other "NodeIdCategory" occurrences.

geoffromer (approved):

PR #3762, opened 2024-03-08
Merging forward declarations

jonmeow:
-   Add the `extern` keyword for forward declarations in libraries that don't
    provide the definition.
-   Treat repeated forward declarations as redundant.
    -   Allow them when they prevent a dependence on an imported name.
-   Clarify rules for when modifier keywords should be on forward declarations
    and definitions.

### DIFF
--- a/proposals/p3762.md
+++ b/proposals/p3762.md
@@ -28,8 +28,8 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
         -   [Libraries cannot both define an entity and declare it `extern`](#libraries-cannot-both-define-an-entity-and-declare-it-extern)
         -   [`impl` files with a forward declaration must contain the definition](#impl-files-with-a-forward-declaration-must-contain-the-definition)
         -   [Type scopes may contain both a forward declaration and definition](#type-scopes-may-contain-both-a-forward-declaration-and-definition)
-    -   [Merging `extern` declarations](#merging-extern-declarations)
     -   [Using `extern` declarations in `extern impl`](#using-extern-declarations-in-extern-impl)
+    -   [`impl` lookup involving `extern` types](#impl-lookup-involving-extern-types)
     -   [Modifier keywords](#modifier-keywords)
     -   [Caveats](#caveats)
         -   [Migration of C++ forward declarations](#migration-of-c-forward-declarations)
@@ -134,25 +134,28 @@ For example:
 
 ## Proposal
 
-1.  `extern` is used to mark forward declarations of entities defined in a
-    different library.
-    -   `extern` must only be used for declarations in libraries that don't
-        contain the definition. It is invalid for the defining library to
-        declare as `extern`.
-    -   In modifiers, `extern` comes immediately after access control keywords,
+1.  Add `extern` declarations.
+    1.  `extern` is used to mark forward declarations of entities defined in a
+        different library.
+        -   The defining library must declare the entity as public in an `api`
+            file.
+        -   It is invalid for the defining library to declare as `extern`.
+        -   `extern` must only be used for declarations in libraries that don't
+            contain the definition.
+    2.  In modifiers, `extern` comes immediately after access control keywords,
         and before other keywords.
+    3.  Declaring an entity as `extern` anywhere in a file means it must be used
+        as `extern` _throughout_ that file.
+        -   The entity cannot be used _at all_ prior to the `extern`
+            declaration, even if there is an imported declaration.
+        -   This allows refactoring to add and remove declarations without
+            impacting files that have an `extern`.
+    4.  If a library declares an entity as non-`extern` in either the `api` or
+        `impl`, it is invalid to declare the same entity as `extern` elsewhere
+        within the library.
 2.  An entity may only be forward declared once in a given file. A forward
     declaration is only allowed before a definition.
-3.  Declaring an entity as `extern` anywhere in a file means it must be used as
-    `extern` _throughout_ that file.
-    -   The entity cannot be used _at all_ prior to the `extern` declaration,
-        even if there is an imported declaration.
-    -   This allows refactoring to add and remove declarations without impacting
-        files that have an `extern`.
-4.  If a library declares an entity as non-`extern` in either the `api` or
-    `impl`, it is invalid to declare the same entity as `extern` elsewhere
-    within the library.
-5.  Modifier keywords will be handled on a case-by-case basis for merging
+3.  Modifier keywords will be handled on a case-by-case basis for merging
     declarations.
 
 ## Details
@@ -209,7 +212,7 @@ class A;
 
 class B;
 // Invalid: Disallowed due to repetition.
-class B:
+class B;
 
 class C;
 // Valid: Allowed because the definition is added.
@@ -259,22 +262,6 @@ fn Foo(c: C);
 extern class C;
 ```
 
-This should have limited effect on `api` and `impl` combinations, but will still
-have some effect. For example:
-
-```carbon
-package Bar library "a" api;
-
-class C { ... }
-```
-
-```carbon
-package Bar library "a" impl;
-
-// Invalid: Imported the definition from the `api` file.
-class C;
-```
-
 #### Libraries cannot both define an entity and declare it `extern`
 
 In a library, if the `impl` defines an entity, the `api` must not use `extern`
@@ -313,6 +300,19 @@ the `impl` is not defined locally. Note that an entity declared in the `api` may
 still not get a compile-time diagnostic unless the compiler is told it's seeing
 _all_ available `impl` files.
 
+```carbon
+package Bar library "a" api;
+
+class C { ... }
+```
+
+```carbon
+package Bar library "a" impl;
+
+// Invalid:
+class C;
+```
+
 #### Type scopes may contain both a forward declaration and definition
 
 The combination of a forward declaration and a definition is allowed in type
@@ -339,38 +339,42 @@ class C {
 This is necessary because type bodies are not automatically moved out-of-line,
 unlike function bodies.
 
-### Merging `extern` declarations
-
-When merging imported declarations from different libraries, multiple
-declarations for the same entity may be found. Only one declaration is retained
-for semantic checking and diagnostics. If available, a non-`extern` declaration
-(potentially with a definition) is used; otherwise, the `extern` declaration
-from the earliest import is used.
-
-For a new declaration in the importing file, an imported `extern` declaration is
-always superseded. It is invalid to use the imported declaration before the new
-declaration. It is considered the same entity for type checking, but diagnostics
-should use the local definition.
-
-For a new declaration in the importing file, an imported non-`extern`
-declaration will:
-
--   If the new declaration is `extern`, the imported non-`extern` is superseded
-    as above. It is invalid to use the imported declaration before the new
-    declaration.
--   If the new declaration is non-`extern`, the imported declaration must be
-    from the same library.
-    -   The `api` may forward declare when the `impl` defines, or the `impl` may
-        repeat a forward declaration and also define.
-
 ### Using `extern` declarations in `extern impl`
 
-An `extern` type may only be used in the type structure of an `impl` if it's an
-`extern impl`. They are incompatible with non-`extern` `impl`s.
+It is invalid for a non-`extern` `impl` declaration to use an `extern` type in
+its type structure.
 
 Consider two libraries, one defining `A` and declaring `B` as `extern`, and the
 other defining `B` and declaring `A` as `extern`. Neither should be able to
 define an `impl` involving both `A` and `B`, otherwise both could.
+
+### `impl` lookup involving `extern` types
+
+If `impl` lookup involving `extern` types finds a non-`final` parameterized
+`impl`, the result is that the lookup succeeds, but none of the values of the
+associated entities of interface are known. This is because there may be another
+more specialized `impl` that applies that is not visible (as can also happen
+with constrained generics).
+
+For example:
+
+```
+library "A" api;
+class C { }
+class D(T:! type) { }
+impl forall [T:! type] D(T) as I where .Result = i32;
+impl D(C) as I where .Result = bool;
+```
+
+```
+library "B" api;
+extern class C;
+extern class D(T:! type);
+extern impl forall [T:! type] D(T) as I where .Result = i32;
+```
+
+In the above, `D(C)` impls `I`, but with unknown `.Result`, since it might not
+be `i32`.
 
 ### Modifier keywords
 
@@ -382,8 +386,8 @@ When considering various modifiers on a forward declaration versus definition:
     [Forward `impl` declaration](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/generics/details.md#forward-impl-declaration).
 -   Other class, impl, and interface modifiers (`abstract`, `base`, `final`)
     exist only on the definition, not on the forward declaration.
--   Function modifiers (`impl`, `virtual`, `default`, `abstract`, `base`,
-    `final`) must match between forward declaration and definition.
+-   Function modifiers (`impl`, `virtual`, `default`, `abstract`, `final`) must
+    match between forward declaration and definition.
     -   This only affects type-scoped names because they are invalid on
         namespace names.
     -   `abstract` won't have a definition so is moot here.
@@ -521,8 +525,19 @@ The trade-offs we consider are:
         -   Requiring them would be somewhat inconsistent with C++.
     -   `extern` is a special-case where its presence is intrinsic to the
         keyword's semantics.
--   For a function, we are choosing to duplicate modifiers between the
+-   For type-scoped members, we are choosing to duplicate modifiers between the
     declaration and definition.
+
+    -   For example:
+
+        ```
+        class A {
+          private class B;
+        }
+        // `private` is required here.
+        private class A.B { ... }
+        ```
+
     -   There is more emphasis on being able to copy-paste a function
         declaration. This in particular may help developers more than classes
         because all the parameters must also be copied. Things such as `static`
@@ -530,6 +545,9 @@ The trade-offs we consider are:
         friction rather than a benefit.
     -   Modifiers such as `virtual` affect the calling convention of functions,
         and as a consequence _must_ be on the first declaration.
+    -   A downside of this approach is that it means the class name is inserted
+        in the middle of the out-of-scope definition, rather than near the
+        front.
 
 The most likely alternative would be to disallow most modifiers on out-of-line
 definitions after a forward declaration, for type member functions in specific.
@@ -698,7 +716,7 @@ imported library.
 3.  Risks of a public extern
 
     When making an `extern fn`, it is callable. This creates a risk of a
-    function being declared as `extern` for intern use, but accidentally
+    function being declared as `extern` for internal use, but accidentally
     allowing dependencies on the function. This risk could be mitigated by
     requiring access control of an `extern` to be either equal to or more
     restrictive than the original symbol (which might be hard to validate, but
--- a/proposals/p3762.md
+++ b/proposals/p3762.md
@@ -39,6 +39,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
     -   [`extern` naming](#extern-naming)
     -   [Default `extern` to private](#default-extern-to-private)
     -   [Opaque types](#opaque-types)
+    -   [Require a library provide its own `extern` declarations](#require-a-library-provide-its-own-extern-declarations)
 
 <!-- tocstop -->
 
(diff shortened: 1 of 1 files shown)


jonmeow:
To note commits, the "Convert from Google doc" should pretty much be what's [in Docs](https://docs.google.com/document/d/1odDdc0bgoR-U5PiacrJvzJcloAKgrHhDmtaD_E_iIAQ/edit?tab=t.0), which has edits to address comments that were made.  "TODO cleanup" is cleaning up a couple TODOs that I'd left in the doc, including filling out some discord background that @josh11b highlighted.

proposals/p3762.md
```diff
@@ -0,0 +1,667 @@
...
+    -   [`extern` naming](#extern-naming)
+    -   [Default `extern` to private](#default-extern-to-private)
+    -   [Opaque types](#opaque-types)
+
+<!-- tocstop -->
+
+## Abstract
+
+-   Add the `extern` keyword for forward declarations in libraries that don't
+    have the definition.
```
chandlerc: Super minor wording tweak?

```suggestion
    provide the definition.
```

zygoloid (commented):
What is our story for coherence in the presence of `extern` declarations, in particular `extern` type and interface declarations?

The best answer I can think of off the top of my head is that an `impl` lookup that involves an `extern` declaration in any way is invalid -- but that will have some serious implications if we define function calls in terms of `impl` lookup!

@josh11b Is this a problem that you've considered previously?

proposals/p3762.md
```diff
@@ -0,0 +1,667 @@
...
+```
+
+```carbon
+package Foo library "d" api;
+import library "a";
+
+fn Foo(c: C);
+
+// Invalid: `F` used the imported `C`, so an `extern` declaration of `C` is now
+// invalid.
```
chandlerc:
```suggestion
// Invalid: `Foo` used the imported `C`, so an `extern` declaration of `C` is now
// invalid.
```

proposals/p3762.md
```diff
@@ -0,0 +1,667 @@
...
+  fn F();
+  class D {
+    fn G() { return C.F(); }
+  }
+  fn F() { D.G(); }
+}
+```
+
+This is necessary because type bodies are not automatically moved out-of-line,
+unlike function types.
```
chandlerc:
```suggestion
This is necessary because type bodies are not automatically moved out-of-line,
unlike function bodies.
```

proposals/p3762.md
```diff
@@ -0,0 +1,667 @@
...
+
+```carbon
+class C {
+  class D;
+
+  fn F();
+  class D {
+    fn G() { return C.F(); }
+  }
+  fn F() { D.G(); }
```
chandlerc: This example seems a bit odd, as `F` doesn't return anything, and `G` doesn't declare a return? And I think it's useful to actually have declarations here because that uses `D` first. Can also emphasize that there is only one `D` type with `Self`...

So maybe:
```suggestion
  fn F() -> D;
  
  class D {
    fn G() -> Self { return C.F(); }
    
    var x: i32;
  }
  
  fn F() -> D { return {.x = 42}; }
```

proposals/p3762.md
```diff
@@ -0,0 +1,667 @@
...
+
+For diagnostics, when merging imported declarations that don't differ (from
+different imported libraries):
+
+-   Two `extern` declarations will prefer the first.
+-   An `extern` and non-`extern` declaration will always prefer the non-`extern`
+    declaration.
+-   Two non-`extern` declarations is invalid.
+    -   While a given file may both forward declare and have a separate
+        declaration with definition, they only export one declaration.
```
chandlerc: Maybe a simpler way to say this:

```suggestion
When merging imported declarations that don't differ (from
different imported libraries), diagnostics should use the non-`extern` declaration if it is imported, and otherwise the first `extern` declaration imported.
```

proposals/p3762.md
```diff
@@ -0,0 +1,667 @@
...
+
+For a new declaration in the importing file, an imported `extern` declaration is
+always superseded. It is invalid to use the imported declaration before the new
+declaration. It is considered the same entity for type checking, but diagnostics
+will lean towards the local definition.
```
chandlerc: Just to keep wording consistent, maybe:

```suggestion
should use the local definition.
```

proposals/p3762.md
```diff
@@ -0,0 +1,667 @@
...
+starting point that we can implement towards. This will likely evolve in future
+proposals, particularly as more keywords are added. However, this still offers a
+baseline.
+
+The trade-offs we consider are:
+
+-   Consistency in when a modifier keyword is expected (if applicable) is
+    valuable. While we may change specifics about when or where a keyword is
+    required, making it either allowed or disallowed is preferred over making it
+    optional.
```
chandlerc: Slight wording tweak, as "allowed" could be read as the as "optional"...

```suggestion
    required, when it is in effect it should be either required or disallowed
    rather than it being allowed but optional without changing semantics.
```

proposals/p3762.md
```diff
@@ -0,0 +1,667 @@
...
+    -   We could require similar consistency on `extern`, and should have more
+        nuanced rules if the access control rules go beyond public and
+        `private`. For example, a package-private type shouldn't be allowed to
+        be made public through an `extern` declaration; but package-private
+        isn't actually part of the language right now.
+-   For a type, we are choosing to have minimal modifiers on the declaration.
+    -   The modifiers we disallow (`abstract`, `base`, `final`, and `default`)
+        have no effect on uses because the forward declared type is incomplete.
+    -   Requiring them would be somewhat inconsistent with C++, and may
+        primarily create toil.
```
chandlerc:
```suggestion
    -   The modifiers we disallow (`abstract`, `base`, `final`, and `default`)
        have no effect on uses because the forward declared type is incomplete.
        -   Requiring them would end up leaking an implementation detail and
            create toil.
    -   Requiring them would be somewhat inconsistent with C++.
```

chandlerc: (mostly trying to capture that it leaks implementation detail, which seems important in the rationale)

proposals/p3762.md
```diff
@@ -0,0 +1,667 @@
...
+    -   There is more emphasis on being able to copy-paste a function
+        declaration. This in particular may help developers more than classes
+        because all the parameters must also be copied. Things such as `static`
+        in C++ being declaration-only is also something we view as a source of
+        friction rather than a benefit.
+    -   Modifiers such as `virtual` affect the calling convention of functions,
+        and as a consequence _must_ be on the first declaration.
+
+The most likely alternative would be to, for type member functions in specific,
+disallow most modifiers. This would be because, unlike other situations, a
```
chandlerc:
```suggestion
The most likely alternative would be to, for type member functions in specific,
disallow most modifiers on definitions after a declaration. This would be because, unlike other situations, a
```

proposals/p3762.md
```diff
@@ -0,0 +1,667 @@
...
+member must have a forward declaration if there is an out-of-line definition. We
+would probably want to drop these collectively in order to maximize copy-paste
+ability (ideally, everything before `fn` is dropped). However, it would shift
+understandability of the definition in a way that may be harmful. For now this
+proposal suggests adopting the more verbose approach and seeing how it goes.
+
+Similarly, we could allow flexibility to choose which modifier keywords are
```
chandlerc: I think this is more "alternatively"?

```suggestion
Alternatively, we could allow flexibility to choose which modifier keywords are
```

proposals/p3762.md
```diff
@@ -0,0 +1,667 @@
...
+declared similarly, but have different keywords on the definition, implying a
+difference in behavior that would not exist. With consideration for the
+[Principle: Prefer providing only one way to do a given thing](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/project/principles/one_way.md),
+this proposal takes the more prescriptive approach, rather than offering
+flexibility.
+
+Note a common aspect between types and functions in this model is that modifiers
+on the definition are typically a superset of modifiers on the declaration.
+While looking at the declaration always gives an incomplete view, looking at the
+definition can give a complete view.
```
chandlerc: Above it says keywords on a function definition must be some *subset* of the keywords on the declaration. So I don't think there is a commonality (at the moment) between types and functions here....

I think the real issue here is that the modifiers we have today are of two different kinds for functions and types outside of access control:

- Type modifiers today: *not* semantically needed to use a declaration w/o definition. Only needed to specify semantics of the definition.
- Function modifiers today: *mostly* semantically needed to use a declaration. (The only one that is less clear is `impl` here I think.)

The distinction also isn't (to me) all that surprising -- there is extremely little you can do with just a declaration of a type, so there isn't much utility for modifiers there. And conversely, most of a functions semantics are available with just a declaration, and so most of the modifiers for functions will be needed before a definition.

Eventually, we may end up with some definition-only modifiers for functions (`inline` is my first guess). My feeling is that this is more about the nature of the modifier, and that happens to inherently reflect the nature of how much semantic information a declaration provides (as opposed to a definition). And so it *happens* that this seems like only a distinction between types and functions now, but I think long term its just going to (as the main proposal says) case-by-case with the nature of each modifier indicating what rules we want.

Anyways, not sure what if anything to capture from this beyond correct the superset statement here... I think the proposal already points us in the right direction.

proposals/p3762.md
```diff
@@ -0,0 +1,667 @@
...
+    really useful if there are functions which take a pointer as a parameter,
+    but an instance of the pointer could only be created by either unsafe casts
+    or if there's a function that returns the pointer type.
+
+    Unsafe casts carry an inherent risk. In the case of a returned pointer, that
+    type could be captured by `auto`. Having `extern class` default to private
+    does not prevent the type's use.
+
+Considering the trade-offs involved combination of these three points, this
+proposal suggests using the regular access control semantic.
```
chandlerc: FWIW, I'm not quite that sold on the trade-offs here. But what I think *is* compelling is that we should use the simple regular access control semantics for now, and revisit this later, ideally when we can consider how exactly we want to address your point (2) above. I mostly don't want to imply much weight behind the current direction.

At least personally, I'm really nervous about this aspect of the proposal, but I'm not at all nervous about waiting until we figure out more of the access control generally and then looking carefully at the options in this space. Not sure how others feel.

chandlerc (commented):
First pass of comments.

josh11b:
> What is our story for coherence in the presence of `extern` declarations, in particular `extern` type and interface declarations?
> 
> The best answer I can think of off the top of my head is that an `impl` lookup that involves an `extern` declaration in any way is invalid -- but that will have some serious implications if we define function calls in terms of `impl` lookup!

@zygoloid We can't use an `extern` type to satisfy the orphan rule in an `impl` definition, and `extern` types are incomplete, which limits them further.

There is an issue with `impl` lookup if you don't import the library with the definition (which is likely the point), but for one narrow case we can allow it: when we know all the implementations of a specific interface for the `extern` type. For this we need two ingredients:
* an `extern` declaration saying what is implemented, and
* a reason to understand that is complete.

We have these two ingredients in the case of an `extern` function declaration that is not [an open extension point](https://docs.google.com/document/d/1rTsDJ_ZNENH0mhi5E64MGjKJ29OYiI_dzBRz2GzYySU/edit?resourcekey=0-z6_xQ9xuJ5CZ6qJqiuSxuw&tab=t.0#heading=h.rzpb0tjipu1h):
* The `extern` declaration specifies what impl of the call operator exists for the type of the function in the other library.
* [Function overloading is closed](https://docs.google.com/document/d/1rTsDJ_ZNENH0mhi5E64MGjKJ29OYiI_dzBRz2GzYySU/edit?resourcekey=0-z6_xQ9xuJ5CZ6qJqiuSxuw&tab=t.0#heading=h.4qjd5al6gppj), which says that the specified impls are a complete list.

This requires agreement of the whole overload set between the library that defines it and the `extern` declaration.

proposals/p3762.md
jonmeow: Sure, rephrasing a little. Also remembering why I hadn't originally said "diagnostics" here, so augmenting that.

proposals/p3762.md
jonmeow: Elaborating for clarity, trying to break this down more.

proposals/p3762.md
jonmeow: Done.

proposals/p3762.md
jonmeow: Done, rephrased the surrounding slightly.

proposals/p3762.md
jonmeow: Rephrasing "As another alternative"

proposals/p3762.md
jonmeow: I think you're reading "this model" as "the above alternative model", but I meant "the proposed model". Rephrased for clarity.

It sounds like you're agreeing, once I've fixed that?

proposals/p3762.md
jonmeow: Does it help to add:

"That choice may be reconsidered later, based on how the semantics work out and any issues that arise, particularly around access control."

(note, somewhat copying phrasing from "opaque types")

jonmeow (commented):
Thanks for the comments!

jonmeow:
@josh11b @zygoloid Are you suggesting some change here, or is this just something to discuss?

josh11b:
> @josh11b @zygoloid Are you suggesting some change here, or is this just something to discuss?

@jonmeow I talked with @zygoloid yesterday, and I think we ended up in a position in between our original statements:

* @zygoloid said `impl` lookup can't involve an `extern` type, but I think it is allowed to find `extern impl` declarations. It will be rare for the compiler to be able to succeed at resolving an impl lookup to a single result unless the found result is `final`, or it is a function overload. However, impl lookup can use `extern impl` declarations to prove that some impl exists, for example to be able to show that a call to a generic is allowed.
* I said you might be able to use an `extern` type in an `impl` definition as long as it is not used to satisfy the orphan rule, and that is wrong. You may not use an `extern` type in the type structure of a non-`extern` `impl`. To see why, consider two libraries, one defining `A` and declaring `B` `extern`, and the other defining `B` and declaring `A` `extern`. Neither should be able to define an `impl` involving both `A` and `B`, otherwise both could.

Bottom line, I think we should include some statement about `impl`s in this proposal, perhaps:

> You may only use an `extern` type in the type structure of an `extern` `impl`.

It might be good to include the `A` and `B` example to justify this restriction.

jonmeow:
@josh11b Added "Using `extern` declarations in `extern impl`" for this.

@zygoloid Added "`impl` files with a forward declaration must contain the definition" per discussion on #3763

proposals/p3762.md
chandlerc: I probably would have said "should" be reconsidered later, but it probably doesn't matter.

(Sorry for the individual reply, I can't reply to comments in a batch because of the `git-stack` force pushing...)

chandlerc (commented):
I think I'm largely happy here, but would like @zygoloid to take a look as well before it goes in. Happy for it to merge when it LG to him as well.

proposals/p3762.md
```diff
@@ -0,0 +1,828 @@
...
+For example:
+
+```carbon
+class A { ... }
+// Invalid: Disallowed after the definition.
+class A;
+
+class B;
+// Invalid: Disallowed due to repetition.
+class B:
```
zygoloid:
```suggestion
class B;
```

proposals/p3762.md
```diff
@@ -0,0 +1,828 @@
...
+restrictions.
+
+For example:
+
+-   `fn F(); fn F() {}` could merge because the function signatures match,
+    forward declaring a function is valid, and only one definition is provided.
+-   `fn F(); fn F(x: i32);` won't merge because they differ in parameters,
+    resulting in a diagnostic.
+
+## Proposal
```
zygoloid, line 135: There's another rule I'd like to see here:

> - If a declaration is declared `extern`, it must match a public declaration in an `api` file in another library in the same package.

Specifically: there must be an owning library, and the declaration must not be private in that library, and must not be declared only in an `impl` file.

proposals/p3762.md
```diff
@@ -0,0 +1,828 @@
...
+```carbon
+package Bar library "a" api;
+
+class C { ... }
+```
+
+```carbon
+package Bar library "a" impl;
+
+// Invalid: Imported the definition from the `api` file.
```
zygoloid: I would not have expected this result from the description of the rule above:

> In a file, if a declaration or definition of an entity is imported, the file must choose between either using that version or declaring its own. It cannot do both.

In this file, we only use an incomplete, forward-declared version of `C`, and don't use the definition, so I'd expect this to behave the same as `library "b"` above, where we import a definition and then add an incomplete declaration.

If the intent is that an `extern` declaration can follow a definition but a non-`extern` declaration cannot, then I think it'd be good to make that a bit more explicit in the rules.

Do we allow:

```carbon
package Baz library "a" api;
class C;
```

```carbon
package Baz library "a" impl;
class C;
```

? I think if we disallow the imported-definition-followed-by-definition case in a single library, we should do the same for imported-declaration-followed-by-declaration. That way, moving the definition from an `impl` file to the `api` file doesn't break other `impl` files, and redundant declarations in `impl` files are disallowed regardless of whether the declaration in the `api` file is a definition.

Perhaps we could generalize the rule in the previous section:

> In a <del>file</del> <ins>library</ins>, a forward declaration must never follow a forward declaration or definition for the same entity<ins>, including the case where one declaration is in the `api` file and the following declaration is in an `impl` file</ins>.

proposals/p3762.md
```diff
@@ -0,0 +1,828 @@
...
+  }
+
+  fn F() -> D { return {.x = 42}; }
+}
+```
+
+This is necessary because type bodies are not automatically moved out-of-line,
+unlike function bodies.
+
+### Merging `extern` declarations
```
zygoloid: It looks like this section is describing an implementation strategy for the toolchain, rather than part of the language design. If so, mentioning that would be useful.

```suggestion
### Merging `extern` declarations in the toolchain
```

proposals/p3762.md
```diff
@@ -0,0 +1,828 @@
...
+    declaration.
+-   If the new declaration is non-`extern`, the imported declaration must be
+    from the same library.
+    -   The `api` may forward declare when the `impl` defines, or the `impl` may
+        repeat a forward declaration and also define.
+
+### Using `extern` declarations in `extern impl`
+
+An `extern` type may only be used in the type structure of an `impl` if it's an
+`extern impl`. They are incompatible with non-`extern` `impl`s.
```
zygoloid: I found this a little hard to read. What do you think of something like:
```suggestion
It is invalid for a non-`extern` `impl` declaration to use an `extern` type in its type structure.
```

proposals/p3762.md
```diff
@@ -0,0 +1,828 @@
...
+-   Access modifiers (`private` and `protected`) must match.
+    -   As an exception, an `extern` name may be `private` when the actual name
+        is public.
+        -   This allows a library to forward declare another library's type
+            without allowing clients to depend on its forward declaration.
+        -   On merging, the more public declaration will take precedence, hiding
+            `private extern` declarations.
+    -   This affects both type-scoped names and namespace names.
+
+### Caveats
```
zygoloid:
```suggestion
### Future work
```
IIRC, the style guide suggests preferring English over Latin loan words.

proposals/p3762.md
```diff
@@ -0,0 +1,828 @@
...
+
+When considering various modifiers on a forward declaration versus definition:
+
+-   `extern` is only valid on a forward declaration. Rules are detailed above.
+-   `extend` in `extend impl` is only on the declaration in the class body
+    (whether that is a forward declaration or definition), as described at
+    [Forward `impl` declaration](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/generics/details.md#forward-impl-declaration).
+-   Other class, impl, and interface modifiers (`abstract`, `base`, `final`)
+    exist only on the definition, not on the forward declaration.
+-   Function modifiers (`impl`, `virtual`, `default`, `abstract`, `base`,
```
zygoloid: I think the `base` modifier is only permitted on class declarations (and as an introducer for `base: B`).

```suggestion
-   Function modifiers (`impl`, `virtual`, `default`, `abstract`,
```

proposals/p3762.md
```diff
@@ -0,0 +1,828 @@
...
+            as meaning "`D` is not a base class" when it actually means "`D` may
+            or may not be a base class".
+-   Access control has a certain necessity for consistency, so that a consumer
+    of a forward declaration would still be allowed to use the definition if
+    refactored.
+    -   We could require similar consistency on `extern`, and should have more
+        nuanced rules if the access control rules go beyond public and
+        `private`. For example, a package-private type shouldn't be allowed to
+        be made public through an `extern` declaration; but package-private
+        isn't actually part of the language right now.
```
zygoloid, line 511: This argument does not apply to type-scope access control, where we are deviating from C++. I think it's worth saying something about that -- presumably the copy-paste argument given below for functions is the relevant concern here, but I believe this also applies to nested classes (and any other kinds of nested entities, such as interfaces):
```
class A {
  private class B;
}
// `private` is required here
private class A.B {
  // ...
}
```

proposals/p3762.md
```diff
@@ -0,0 +1,828 @@
...
+    | Default visibility             | Public (proposed)         | Private (alternative)            |
+    | ------------------------------ | ------------------------- | -------------------------------- |
+    | Public                         | `extern class C;`         | `export extern class C;`         |
+    | Library-private                | `private extern class C;` | `extern class C;`                |
+    | Package-private (hypothetical) | `package extern class C;` | `export package extern class C;` |
+
+3.  Risks of a public extern
+
+    When making an `extern fn`, it is callable. This creates a risk of a
+    function being declared as `extern` for intern use, but accidentally
```
zygoloid:
```suggestion
    function being declared as `extern` for internal use, but accidentally
```

I think restrictions on which functions interns can call are best enforced by code review rather than language rules ;-)

zygoloid (commented):
Looks good, thanks. A few nuances, but nothing major.

proposals/p3762.md
jonmeow: I'm familiar with [avoiding latin abbreviations](https://developers.google.com/style/abbreviations#dont-use) (exception for "etc") but not latin loan words altogether, which would also disallow things such as "[quota](https://developers.google.com/style/word-list#quota)" or "[appendix](https://developers.google.com/style/word-list#appendix)". Is this getting mixed up?

"Future work" is not the appropriate label here, regardless -- I'm using "Caveat" in its intended form, a warning about implications for migration. I'd been considering "Appendix" (technically also a latin loan word), but "Additional notes", "Supplementary information", or simply "Warnings" would also work if you have a strong preference.

proposals/p3762.md
jonmeow: Done

proposals/p3762.md
jonmeow: Done. FYI @josh11b since I was copying suggested text.

proposals/p3762.md
jonmeow: There are a number of validity comments here, and they shouldn't be considered to be toolchain-specific. I've rewritten a little to omit diagnostics comments.

proposals/p3762.md
jonmeow: Done

proposals/p3762.md
jonmeow: Done

proposals/p3762.md
jonmeow: Done, with some rephrasing and restructuring.

proposals/p3762.md
jonmeow: Moved this example under "`impl` files with a forward declaration must contain the definition" -- I think it fits well there. It is an impl file with a forward declaration, which is clearly invalid under that rule.

I think we don't want to generally disallow forward declarations in `impl` files after the `api` file, due to the separate rule: "Files must either use an imported declaration or declare their own". It would make it more difficult to read an `impl` file which half uses a forward declaration of something forward declared in the `api` file if we carve out a special case for it.

proposals/p3762.md
jonmeow: Added an explicit example and rephrased the function note below to just be type-scoped members.

proposals/p3762.md
zygoloid: Thanks for addressing this, I see you've moved the example next to where we say that case is invalid.

Just to confirm, is the case I mentioned in my previous comment valid? I don't feel strongly, but I think it makes sense to reject given that we reject in the case where the `api` file has a definition. The `api` file is "close enough" to the `impl` file that the declaration in the `impl` file seems like it should be rejected for being redundant.

zygoloid (approved):
I'm happy for this to go forward as-is. Please feel free to merge.

I'd also be happy to add a restriction on forward-declaring an entity in an `impl` file if it's also forward-declared in the `api` file. I don't see a need for that to be in this proposal, but I'm not hitting the merge button in case you would like to add that rule here.

jonmeow (commented):
Oops, forgot to publish responses

proposals/p3762.md
jonmeow: Published the response that I forgot to. Regarding the `api` case, let me give the more complex example of why I think it should be allowed:

```
package Baz library "a" api;
class C;
```

```
package Baz library "a" impl;
class C;
class D {
  fn E(c: C*) { ... }
}
class C {
  fn F(d: D) { ... }
}
```

If `class C;` is invalid in the `impl`, then `E` violates "Files must either use an imported declaration or declare their own". That can't just replace "Files" with "Libraries" because we already have an `api`/`impl` split in how that's handled, so there would be a finely tailored exception for this case. Rather, I think it's better to handle this way, allowing a marginally simpler rule set.

proposals/p3762.md
zygoloid: I'd be much more confident that an English reader would understand "quota" and "appendix" than "caveat", and I don't think there is an alternative English word for those.

The reason I was pushing back a little on this is that I would interpret this heading as meaning "Warnings" or "Beware" or maybe "Limitations" or something like that, but that doesn't seem to fit the content here, making me unsure what you mean by this Latin word. To my reading, this isn't a list of warnings, because it's not discussing negative consequences or dangers of this proposal. Rather, this reads to me as a description of some anticipated future work that is not part of this proposal, for which you've considered interactions with this proposal. That's the kind of information I'd expect to see in a "Future work" section.

But in any case, I don't want to block the proposal on this. We've established that proposals are written in the voice of the author, so while I find this heading to be confusing and unhelpful, I won't insist that it's changed, and I approved the proposal on that basis.

proposals/p3762.md
zygoloid: I agree that we should allow your example, so my suggested rule generalization doesn't work.

But I think that doesn't answer the question about my example. To my reading, my example *is* disallowed by this proposal, specifically because it violates: "In an `impl` file, if the `impl` file forward declares an entity, it must also provide the definition." But yours is not disallowed, because the `impl` file provides the definition.

Does that match your expectation and intent?

proposals/p3762.md
jonmeow: To me, "future work" is anticipated design work, whereas this is just something that people will need to do in the future. "appendix" is a fairly bland name so I've gone with it.

proposals/p3762.md
jonmeow: Per discussion, yes. Tried to make the example clearer.

PR #3763, opened 2024-03-08
Matching redeclarations

zygoloid:
Require exact syntactic matching in redeclarations. Provide new terminology for
redeclaration matching and agreement. Specify non-redeclaration rules for the
other contexts where we require multiple declarations to match, such as `impl`s
of `interfaces`, `impl`s of `virtual fn`s.

### DIFF
--- a/proposals/p3763.md
+++ b/proposals/p3763.md
@@ -564,9 +564,26 @@ redeclarations. However, this turns out to be a poor choice:
     associated constants for the interface, which it may be reasonable to use
     directly in the declaration of the function, as in the preceding example.
 
-Therefore, a different rule is used. In the general case, when matching an
-associated function in an `impl` against a function declaration in an
-`interface`, a synthetic function called a _thunk_ is generated:
+Therefore, different rules are used. In the specific case where an associated
+function declaration in the `impl` can be used directly to satisfy a requirement
+introduced by a function declaration in the `interface`, it is used directly.
+For example, this is necessary to avoid infinite recursion when implementing the
+`Call` interface.
+
+The function in the `impl` is used directly when:
+
+-   Each parameter in the `impl` function has the same type as the parameter in
+    the `interface`. This includes the `self` parameter, which must be present
+    in both functions if it is present in either.
+-   Each parameter in the `impl` has the same category -- either `var` or not --
+    as the parameter in the `interface`.
+-   The return type in the `interface` and `impl` are the same type.
+
+> **Note:** More constraints are expected to appear here over time. The key
+> property we aim to identify is whether the two functions have the same calling
+> convention.
+
+Otherwise, a synthetic function called a _thunk_ is generated:
 
 -   The declaration of the thunk is formed by substituting the `Self` type of
     the `impl` into the declaration in the `interface`. This implicitly also
@@ -586,15 +603,13 @@ Implicit conversions are performed as necessary to initialize the parameters of
 the function in the `impl` from the parameters of the thunk, and to initialize
 the return value of the function from the result of the call.
 
-As a special case, if a `var` parameter in the callee is initialized from a
-`var` parameter of the thunk of the same type, no copy is performed, and the
-parameters refer to the same variable. In this case, the `var` parameter is
-destroyed by either the caller of the thunk or the callee, depending on the
-calling convention, and not by the thunk itself.
+It is an error if a thunk is needed to wrap a function declaration with a `var`
+parameter, because otherwise a copy would always be performed when initializing
+the parameter.
 
-As an optimization, if the parameter types and return type are the same between
-the thunk and its callee, and the calling convention is otherwise the same, the
-implementation should use the callee function directly and not generate a thunk.
+> **Note:** These rules do not cover the case where the function declaration in
+> the `interface` or `impl` has implicit generic bindings. That case is
+> considered to be out of scope for this proposal.
 
 ### `virtual` functions
 
--- a/proposals/p3763.md
+++ b/proposals/p3763.md
@@ -1,4 +1,4 @@
-# Redeclarations
+# Matching redeclarations
 
 <!--
 Part of the Carbon Language project, under the Apache License v2.0 with LLVM
@@ -17,10 +17,10 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 -   [Background](#background)
     -   [#1132 rule](#1132-rule)
     -   [C++ rule](#c-rule)
+    -   [One-definition rule](#one-definition-rule)
+    -   [Where declarations can appear](#where-declarations-can-appear)
 -   [Proposal](#proposal)
 -   [Details](#details)
-    -   [Where redeclarations can appear -- recap of #3762](#where-redeclarations-can-appear----recap-of-3762)
-    -   [One-definition rule](#one-definition-rule)
     -   [Syntactic vs semantic matching](#syntactic-vs-semantic-matching)
         -   [Unqualified name lookup](#unqualified-name-lookup)
         -   [`impl`s that have not yet been declared](#impls-that-have-not-yet-been-declared)
@@ -28,7 +28,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
     -   [`_` parameter names and `unused` modifier](#_-parameter-names-and-unused-modifier)
     -   [Scope differences](#scope-differences)
     -   [`impl` declarations](#impl-declarations)
-        -   [Redeclarations](#redeclarations-1)
+        -   [Redeclarations](#redeclarations)
         -   [Differences between `impl` declarations](#differences-between-impl-declarations)
         -   [Out-of-line definitions of associated functions](#out-of-line-definitions-of-associated-functions)
         -   [`impl` members vs `interface` members](#impl-members-vs-interface-members)
@@ -159,6 +159,31 @@ Indeed, it's desirable to support a simple syntactic rule for transforming an
 in-class declaration into an out-of-class definition, both for tooling and for
 programmers.
 
+### One-definition rule
+
+Proposal #3762 establishes the following rules:
+
+If an entity is declared in the `api` file of a library, it can be redeclared in
+`impl` files subject to other rules on redundant redeclarations, and a
+definition is required in exactly one `impl` file.
+
+If the entity is _not_ declared in the `api` file, and is instead introduced as
+non-`extern` in an `impl` file, then it must be defined in that `impl` file.
+Note that this means that such an entity can only be declared in a single `impl`
+file; if an entity is intended to be shared by multiple `impl` files, it must be
+declared in the `api` file, typically as a `private` member of the library.
+
+### Where declarations can appear
+
+Because declarations always declare their entity within the lexically-enclosing
+package, and the package forms part of the scope, declarations for an entity can
+only appear in a single package.
+
+As specified in #3762, entities can be declared in more than one library: Each
+entity has an owning library, and all declarations of that entity outside the
+owning library are declared `extern`. All declarations of an entity without the
+`extern` keyword are required to appear in the same library.
+
 ## Proposal
 
 Replace the terminology ("match" and "agree") and corresponding rules here as
@@ -191,42 +216,6 @@ fn A.G(n: (i32)) {}
 
 ## Details
 
-### Where redeclarations can appear -- recap of #3762
-
-Because declarations always declare their entity within the lexically-enclosing
-package, and the package forms part of the scope, declarations for an entity can
-only appear in a single package.
-
-As specified in #3762, entities can be declared in more than one library: Each
-entity has an owning library, and all declarations of that entity outside the
-owning library are declared `extern`. Some
-[special considerations](#extern-declarations) apply to `extern` declarations.
-All declarations of an entity without the `extern` keyword are required to
-appear in the same library.
-
-As a consequence of those rules, excluding `extern` declarations, all
-declarations of an entity are in the same library. This is important: our chosen
-redeclaration rule closely couples the syntactic choices made in all
-declarations of the same entity, and this would be a significant burden to
-refactoring if the scope across which redeclarations could appear were large.
-Requiring an `api` file and its `impl` files to be changed at the same time is,
-in comparison, a minor burden to refactoring.
-
-### One-definition rule
-
-If the entity is declared in the `api` file of a library, it can be redeclared
-in any `impl` file. A definition is then required in exactly one `impl` file.
-
-If the entity is _not_ declared in the `api` file, and is instead introduced as
-non-`extern` in an `impl` file, then it must be defined in that `impl` file.
-Note that this means that such an entity can only be declared in a single `impl`
-file. An entity shared by multiple `impl` files must be declared in the `api`
-file, typically as a `private` member of the library. If it is not defined in
-the `api` file, it must be defined in exactly one `impl` file in that library.
-
-Note that we cannot in general enforce these one-definition constraints until
-link time, but we can enforce them.
-
 ### Syntactic vs semantic matching
 
 Redeclarations are primarily checked syntactically. The intention is that
@@ -246,6 +235,13 @@ we expect implementations of Carbon to be able to call out semantic differences
 between redeclarations -- for example, "type of parameter `x` in redeclaration
 is different from type in previous declaration" -- not only point to where the
 syntax diverged.
+
+All non-`extern` declarations of an entity are in the same library, and all
+non-`extern` declarations see the first non-`extern` declaration of the entity,
+which is either in the same file or, for a declaration in an `impl` file, in the
+`api` file, so we can always perform a precise syntactic redeclaration check if
+we persist syntactic information from the `api` file to `impl` files. Some
+[special considerations](#extern-declarations) apply to `extern` declarations.
 
 #### Unqualified name lookup
 
--- a/proposals/p3763.md
+++ b/proposals/p3763.md
@@ -615,7 +615,31 @@ the parameter.
 
 For a `virtual` function, the same approach is taken as for `impl` functions: a
 thunk is generated that differs from the declaration in the base class by
-replacing the type of `self` with the derived class.
+replacing the type of `self` with the derived class, unless the function chosen
+for a class can be used directly. When a virtual function is used directly in a
+base class and not overridden in the derived class, it is also used directly in
+the derived class, even though its declared `self` parameter does not have a
+matching type.
+
+```carbon
+base class B {
+  // No thunk used.
+  virtual fn F[addr self: B*]();
+  virtual fn G[addr self: B*]();
+}
+base class C {
+  extend B;
+  // No thunk used: `self` has expected type `C*`.
+  impl fn F[addr self: C*]();
+  // Uses a thunk due to unexpected `self` type.
+  impl fn G[addr self: B*]();
+}
+class D {
+  // No thunk for `F`, because no thunk was used in `C`.
+  // Uses thunk for `F`, because thunk was used in `C`.
+  extend C;
+}
+```
 
 Note that this supports covariant return types automatically, as well as any
 other case where the return value from the derived class function can be
@@ -642,6 +666,9 @@ not useful.
 > so would have return type `B*`. The downside is that the vtable for `B` would
 > have two `Clone` slots, for `A.Clone` and `B.Clone`, whereas a covariant
 > return in C++ would only need a single vtable slot to express the same thing.
+
+It is an error for a class with a custom value representation to declare or
+implement a virtual function that passes `self` by value.
 
 ### `extern` declarations
 
--- a/proposals/p3763.md
+++ b/proposals/p3763.md
@@ -38,7 +38,8 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 -   [Rationale](#rationale)
 -   [Alternatives considered](#alternatives-considered)
     -   [Use a partially or fully semantic rule](#use-a-partially-or-fully-semantic-rule)
-    -   [Disallow shadowing in `impl` after use in `api`](#disallow-shadowing-in-impl-after-use-in-api)
+    -   [Use package-wide name poisoning](#use-package-wide-name-poisoning)
+    -   [Allow shadowing in implementation file after use in API file](#allow-shadowing-in-implementation-file-after-use-in-api-file)
 
 <!-- tocstop -->
 
--- a/proposals/p3763.md
+++ b/proposals/p3763.md
@@ -28,13 +28,17 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
     -   [`_` parameter names and `unused` modifier](#_-parameter-names-and-unused-modifier)
     -   [Scope differences](#scope-differences)
     -   [`impl` declarations](#impl-declarations)
-    -   [`impl` members vs `interface` members](#impl-members-vs-interface-members)
+        -   [Redeclarations](#redeclarations-1)
+        -   [Differences between `impl` declarations](#differences-between-impl-declarations)
+        -   [Out-of-line definitions of associated functions](#out-of-line-definitions-of-associated-functions)
+        -   [`impl` members vs `interface` members](#impl-members-vs-interface-members)
     -   [`virtual` functions](#virtual-functions)
     -   [`extern` declarations](#extern-declarations)
     -   [`let` and `var` declarations](#let-and-var-declarations)
 -   [Rationale](#rationale)
 -   [Alternatives considered](#alternatives-considered)
     -   [Use a partially or fully semantic rule](#use-a-partially-or-fully-semantic-rule)
+    -   [Disallow shadowing in `impl` after use in `api`](#disallow-shadowing-in-impl-after-use-in-api)
 
 <!-- tocstop -->
 
@@ -100,6 +104,11 @@ that the terminology in use here is problematic:
     example, functions with different signatures are said to match.
 -   The term "match" is easy to confuse with pattern-matching terminology.
 -   The term "agree" sounds like it is expressing an opinion.
+-   These terms don't fit well into diagnostics. For example, we considered
+    diagnostic text similar to "Declaration X does not agree with matching
+    declaration Y", which doesn't clearly describe the problem. While it's not
+    necessary for us to use the formal, technical terminology in diagnostics, it
+    would be useful to have terminology that also works in that context.
 
 The chosen rule has also been found to be problematic to implement in practice,
 because it combines the purely-syntactic concern of whether the same syntax is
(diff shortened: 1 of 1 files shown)


proposals/p3763.md
```diff
@@ -0,0 +1,637 @@
...
+2.  Grouping parentheses can be present in one declaration and absent in another
+    if it doesn't change the parse result.
+3.  An alias can be used in one declaration, where the alias target or a
+    different alias to the same target is used in the other declaration.
+
+[Concerns have been raised](https://discord.com/channels/655572317891461132/963846118964350976/1214372759224983562)
+that the terminology in use here is problematic:
+
+-   Declarations are said to "match" even when they are quite dissimilar -- for
+    example, functions with different signatures are said to match.
```
jonmeow, line 105: Is it worth noting the particular challenges that match + agree raises for diagnostic phrasing, as in https://github.com/carbon-language/carbon-lang/pull/3739#discussion_r1513399921? I generally take this as an issue where "match" and "agree" are synonyms in English, but we could have a diagnostic such as: `Declaration X does not agree with matching declaration Y.` -- which might confuse developers.

proposals/p3763.md
```diff
@@ -0,0 +1,637 @@
...
+### One-definition rule
+
+If the entity is declared in the `api` file of a library, it can be redeclared
+in any `impl` file. A definition is then required in exactly one `impl` file.
+
+If the entity is _not_ declared in the `api` file, and is instead introduced as
+non-`extern` in an `impl` file, then it must be defined in that `impl` file.
+Note that this means that such an entity can only be declared in a single `impl`
+file. Entities shared by multiple `impl` files must be defined in the `api`
+file, typically as `private` members of the library.
```
jonmeow: Why the restriction here? I didn't include it in #3762 (maybe subtly so, I don't think I addressed multi-`impl` directly) because it didn't seem necessary to allow a guarantee when building a library. One `impl` should be able to declare as non-`extern` when the other defines it. Multi-`impl` libraries are going to be awkward because each would independently see the declaration in the `api`, and one would see a definition and the other wouldn't -- a diagnostic could only be issued if compiled together, at which point having the `api` define wouldn't matter.

This is also why I was going for requiring non-`extern` in the `api` even though it wouldn't affect compilation: I take `extern` as more about "library is responsible for defining".

proposals/p3763.md
```diff
@@ -0,0 +1,637 @@
...
+
+In this function definition:
+
+-   `F(a: T, b: U)` does not differ from the declaration of `F`.
+-   `class N.C(T:! type).D(U:! type);` would be a valid redeclaration of `D`,
+    because:
+-   `D(U:! type)` does not differ from the declaration of `D`.
+-   `class N.C(T:! type);` would be a valid redeclaration of `C`, because:
+-   `C(T:! type)` does not differ from the declaration of `C`.
+-   `namespace N;` would be a valid redeclaration of `N`.
```
jonmeow: Should this be indented:
```suggestion
-   `class N.C(T:! type).D(U:! type);` would be a valid redeclaration of `D`,
    because:
    -   `D(U:! type)` does not differ from the declaration of `D`.
    -   `class N.C(T:! type);` would be a valid redeclaration of `C`, because:
        -   `C(T:! type)` does not differ from the declaration of `C`.
        -   `namespace N;` would be a valid redeclaration of `N`.
```

proposals/p3763.md
```diff
@@ -0,0 +1,637 @@
...
+    make refactoring harder.
+
+We address these points as follows:
+
+-   Checking whether `extern` declarations properly redeclare their targets
+    during compilation is best-effort. Some level of semantic checks may be
+    important for an implementation, because constant evaluation may cross
+    between file boundaries, but checks that the declarations are syntactically
+    identical can be skipped.
+    -   In place of compile-time checking, `extern` declarations are expected to
```
jonmeow:
```suggestion
    -   As a fallback for compile-time checking, `extern` declarations are expected to
```

Note, you also say things like "This rule cannot in general be checked and "Some level of semantic checks may be important". The argument being made seems to be that compile-time checks should be omitted; I'm wary of that. e.g., I think the common case for `extern` on a class might be that the `api` doesn't import the definition but the `impl` does, in order to break a dependency loop. That could still get compile-time checking from the `impl`. But, is it your intent to say that we should not have compile-time checks?

proposals/p3763.md
```diff
@@ -0,0 +1,637 @@
...
+    burden, because it introduces syntactic coupling across libraries that may
+    make refactoring harder.
+
+We address these points as follows:
+
+-   Checking whether `extern` declarations properly redeclare their targets
+    during compilation is best-effort. Some level of semantic checks may be
+    important for an implementation, because constant evaluation may cross
+    between file boundaries, but checks that the declarations are syntactically
+    identical can be skipped.
```
jonmeow: What's intended here? I'm stuck on the phrasing and it seems like it's going to have an impact on what code is valid (syntactic differences should not be expected to be caught by linker errors). I think we could probably do really well with syntactic checking, but do you mean to just disallow it? Is it worth considering the alternative?

```suggestion
    between file boundaries, but checks that the declarations are syntactically
    identical will be best-effort.
```
```suggestion
    between file boundaries, but checks that the declarations are syntactically
    identical will be skipped.
```

jonmeow (commented):
I had a couple questions about `extern` handling as mentioned here. I skimmed through everything else and 👍

proposals/p3763.md
zygoloid: Added some text on that.

proposals/p3763.md
zygoloid: Sorry, what I wrote here wasn't what I meant -- the restriction I was going for is: if you declare something in multiple `impl` files, then it needs to be *declared* in the `api` file, not *defined* in the `api` file. Multiple `impl`s can't share something just by declaring it the same way as each other. Instead, if something is first declared in an `impl` file, then it's local to that `impl` file, and if `impl` files want to communicate with each other, they do that via things declared `private` in the `api` file.

proposals/p3763.md
zygoloid: I've reworked this section a bit to try to make my intent clearer.

proposals/p3763.md
zygoloid: The intent is that syntactic differences *are* caught by the link-time check, so compile-time syntactic checks aren't required and we don't need to include a syntactic description of every declaration in the precompiled form of the `api` file. (We do need to include it in the object file so the linker can check, but I don't see a way to avoid that.) Hopefully clarified by the new updates.

proposals/p3763.md
jonmeow: That's how I understood it, I think. I'm trying to figure out what the rationale is (and this may be something to put in #3762 instead, given I'm trying to hammer out `extern` behaviors there).

My understanding from how this affects diagnostics is:
- _with_ this restriction, at compile time:
  - If an `api` file has a declaration, we can only issue an "undefined" diagnostic if we have some way to say "all `impl`s are being compiled at once" (note, I'm sad about this because I think this will be the common case)
  - If an `impl` file has a declaration, we can issue an "undefined" diagnostic.
- _without_ this restriction, at compile time, the behavior is the same either way: no diagnostic unless "all `impl`s are being compiled at once"

In other words, the only diagnostic this enables is for the case it disallows. For that, it takes some incremental flexibility: an API could be fully internal to a library without needing to modify the header.

proposals/p3763.md
```diff
@@ -0,0 +1,755 @@
...
+```carbon
+class A {
+  virtual fn F[self: Self]();
+}
+
+// OK, `virtual` not required to match.
+fn A.F[self: Self]() {}
+```
+
+### `_` parameter names and `unused` modifier
```
jonmeow, line 336: I should ask here, do we want to continue to allow `_` parameters in extern declarations?

With `_`, a library could rename a function parameter with per-file changes, even if one of those steps is to change the name in `extern` declarations to `_`. Without `_`, the entire package must change in sync.

If the perspective is that this package-scope refactoring requirement is okay for function parameters, maybe we should re-evaluate how other merging works, such as whether redundant redeclarations are allowed? Most of the reason for those had been about avoiding package-scope changes, as I understood it. Is this just a split trade-off, weighed differently here?

proposals/p3763.md
jonmeow: Is this just disallowing the case because you don't think it's helpful? Maybe because you want to not provide reference-able symbols for anything not in the `api`?

zygoloid: I want redeclarations to be checked during compilation, not at link time, as much as possible; I don't want to have two different `impl` files both contain a declaration that we match up without ever having seen both of them existing at once. And I want to maximize the number of cases where we can diagnose a declaration without a definition during compilation rather than at link time.

C++ lets you play "declare the thing multiple times in independent files that never see each other" games, and it's much more harmful to software evolution than it is helpful. For `extern` declarations, I think we essentially need to put up with it, but I don't think we should be allowing it within the scope of a single library.

Happy to add an "Alternative considered" here if you like.

proposals/p3763.md
zygoloid: `_` doesn't allow renaming function parameters in general -- specifically, function parameters that are used later in the signature (eg, type parameters) aren't helped by that. So it's only a half-measure, and a general parameter renaming tool still needs to make package-wide changes. Allowing `_` also makes a purely-syntactic check more complex -- we'd have to disallow later uses of the parameter name somehow, because those might resolve to something else in the context of the declaration using `_`.

I think I wouldn't want to encourage a proliferation of `extern` declarations in arbitrary other parts of a package -- I'd want to encourage a style where there is a single `extern` declaration that's kept alongside the library that owns the real declaration. So to the extent that a change has package-wide impact, that seems like an avoidable problem in the existing code. (Though I do wonder whether we should have some enforcement that people don't create this style issue.)

proposals/p3763.md
jonmeow: I'm fine with this, but can I just ask to pull this into #3762 so that it's considered alongside the other rules for where declarations may be placed?

proposals/p3763.md
jonmeow: Marking this resolved, I think we have other things to discuss regarding `extern` anyways.

proposals/p3763.md
```diff
@@ -0,0 +1,783 @@
...
+  impl Self as I where _ {}
+}
+```
+
+> **Note:** These rules assume that `impl` declarations have an associated
+> scope, and can only be redeclared in that scope, like all other declarations.
+> This was
+> [agreed in discussion on 2024-03-11](https://docs.google.com/document/d/1s3mMCupmuSpWOFJGnvjoElcBIe2aoaysTIdyczvKX84/edit?resourcekey=0-G095Wc3sR6pW1hLJbGgE0g&tab=t.0#heading=h.p69b78lovqb7)
+> but has not yet been incorporated into a design proposal. This will be the
+> subject of a separate proposal.
```
chandlerc: I'd prefer to avoid using "agreed" or "decided" when referring to open discussions, just to avoid ambiguity with issues that document more formal decisions.

Maybe "This was the consensus in our discussion" or something similar?

proposals/p3763.md
```diff
@@ -0,0 +1,783 @@
...
+
+```carbon
+impl forall [T:! type] T as Interface(T) {
+  fn F();
+}
+fn (forall [T:! type] T as Interface(T)).F() {}
+```
+
+> **Note:** As part of associating `impl` declarations with a scope, we
+> [agreed](https://docs.google.com/document/d/1s3mMCupmuSpWOFJGnvjoElcBIe2aoaysTIdyczvKX84/edit?resourcekey=0-G095Wc3sR6pW1hLJbGgE0g&tab=t.0#heading=h.p69b78lovqb7)
```
chandlerc: Another place where I feel like "agreed" should be avoided. In this case, I think its even more important because I think this is still very fresh...

proposals/p3763.md
```diff
@@ -0,0 +1,783 @@
...
+declare the same entity. Instead, two `impl` declarations declare the same
+entity if the portion of the declaration from the introducer keyword until the
+`;` or `{` does not differ, except:
+
+-   If the constraint type in the `impl` is of the syntactic form
+    `expression where constraints`, then the `constraints` portion is not
+    considered.
+-   `impl as` is rewritten to `impl Self as` before looking for a previous
+    declaration and checking for differences from any previous declaration that
+    is found.
```
chandlerc, line 447: I think this removes the ability to provide an out-of-line definition after a declaration of the form `impl as I`?

Or do the scope rules mean that you "just" have to form a valid redeclaration of `Self`?

Either way, would be good to document the consequence clearly here I think.

proposals/p3763.md
```diff
@@ -0,0 +1,783 @@
...
+> ```
+> fn [T:! type](T as Interface(T)).F() {}
+> ```
+>
+> However, that change is out of scope for this proposal.
+
+#### `impl` members vs `interface` members
+
+Associated functions in an `impl` are required to match the corresponding
+declarations in the `interface`:
```
chandlerc: This is using the "match" terminology again...

proposals/p3763.md
```diff
@@ -0,0 +1,783 @@
...
+
+As a special case, if a `var` parameter in the callee is initialized from a
+`var` parameter of the thunk of the same type, no copy is performed, and the
+parameters refer to the same variable. In this case, the `var` parameter is
+destroyed by either the caller of the thunk or the callee, depending on the
+calling convention, and not by the thunk itself.
+
+As an optimization, if the parameter types and return type are the same between
+the thunk and its callee, and the calling convention is otherwise the same, the
+implementation should use the callee function directly and not generate a thunk.
```
chandlerc: These special cases worry me a bit about the approach.

Is it possible to frame the semantics here to constructively avoid the need for this? The implicit conversions thing is something we already get when calling a function -- is there a way to avoid the indirection through a thunk when framing this?

chandlerc (commented):
Largely looks good. Mostly cosmetic comments on my end except for the thunk part for `impl` functions. But even that is really just trying to find a somewhat cleaner model, not suggesting shifting direction.

proposals/p3763.md
zygoloid: Maybe there's some nice way to do that, but I've not found it yet.

I think there end up somewhat fundamentally being two separate implicit conversion steps here. For example, given:

```carbon
interface A {
  fn F(a: i16);
}
class C {}
impl C as A {
  fn F(a: i32) { Print("F"); }
}
fn G(T:! A, a: i8) { T.F(a); }
fn H(a: i8) { G(C, a); }
```

... I think we first want to convert from `i8` to `i16` in the call to `T.F`, and then from `i16` to `i32` in the `impl`. I think that's going to be necessary because in general we'll be type-checking the two halves of this conversion separately. And in the general case there might not be an implicit conversion directly from whichever type is in the place of `i8` to whichever type is in the place of `i32`, or the combined conversion might do something different from the two individual conversions (though the latter case concerns me less than the former).

In the above example, I *think* I also want to reject `(C as A).F(n)` given `n: i32` -- even though the function in the `impl` supports an `i32` argument, I think `(C as A).F` should implement `Call(like i16)`, like `T` does. But that seems less clear-cut to me: we know more about `C as A` than we do about `T`, so perhaps it's OK for us to also know more about `(C as A).F` than about `T.F` -- but I think it'd be problematic if it *only* implements `Call(like i32)` and not `Call(like i16)`, because that'd mean that a checked generic call that works before monomorphization doesn't work afterwards / if the call is manually monomorphized (if the argument is a type that implicitly converts to `i16` but not to `i32`). I think this works out better if `(C as A).F` is something like "thunk for `F` with the signature of `A.F`" rather than simply being `F`.

However, one potentially nice observation is that these special cases aren't actually special just for thunks. We get the exact same special cases in the converting `impl` of `Call` for an `fn` as we get in the converting `impl` of `Call` for a thunk. In both cases, we get an implicit-conversion `impl` of `Call` that doesn't copy `var`s, and an exact-match `impl` of `Call` that doesn't recursively use `Call` again. Maybe once we understand better how `var` interacts with the `Call` interface, the special case for `var` might even disappear here.

proposals/p3763.md
zygoloid: Done.

proposals/p3763.md
zygoloid: Done.

proposals/p3763.md
zygoloid: Added some extra text on this, mostly in the note.

proposals/p3763.md
zygoloid: Fixed.

proposals/p3763.md
```diff
@@ -0,0 +1,816 @@
...
+-   The associated function in an `impl` is expected to have different syntax
+    than that in the interface in some cases. The two declarations are in
+    different scopes, so will refer to the same types in different ways. And the
+    declaration in the `impl` is declared with knowledge of the `Self` type and
+    associated constants for the interface, which it may be reasonable to use
+    directly in the declaration of the function, as in the preceding example.
+
+Therefore, a different rule is used. In the general case, when matching an
+associated function in an `impl` against a function declaration in an
+`interface`, a synthetic function called a _thunk_ is generated:
```
chandlerc: FYI, this somewhat re-introduces the `matching` term... But see below.

proposals/p3763.md
chandlerc: I agree that we'll want a thunk in the general case, and I like that.

My suggestion was that it would be nice to have a preferred mode where there is no thunk at all that is tried first rather than framing eliding the thunk as an optimization. Specifically, I'm imagining we say that if the types match exactly between impl and interface, we do the non-thunk thing, but if there is any difference, we then build a thunk. And I wonder if we want to preclude `var` parameters when the thunk is required?

If it's important to build from the thunk up, we could talk about notionally generating a thunk, and then if the call is an *exact* match, we call the `impl`  function directly rather than the thunk. And again, maybe only then is `var` allowed?

Either way, I think it would be good to make the text here a bit more clear that the `impl` declaration *doesn't* redeclare the `interface` declaration in any way, that it is a new thing, but that these declarations \<TERM\> (correspond?) through the notional thunk building?

FWIW, also happy to just leave the `var` behavior as a future work thing, it doesn't seem important to get right here.

chandlerc (commented):
Sorry this somewhat fell off my radar, I think there's mostly just a wording/nuance tweak that I'd like to try to get working before landing still...

proposals/p3763.md
zygoloid: Done.

proposals/p3763.md
```diff
@@ -0,0 +1,858 @@
...
+> `impl`s to add a return type in anticipation of the `interface` later adding
+> one. However, such a feature is considered out of scope for this proposal.
+
+Implicit conversions are performed as necessary to initialize the parameters of
+the function in the `impl` from the parameters of the thunk, and to initialize
+the return value of the function from the result of the call.
+
+It is an error if a thunk is needed to wrap a function declaration with a `var`
+parameter, because otherwise a copy would always be performed when initializing
+the parameter.
```
chandlerc, line 626: Should we add a note about potentially revisiting this when we have moves?

Something like:

```suggestion
It is an error if a thunk is needed to wrap a function declaration with a `var`
parameter, because otherwise a copy would always be performed when initializing
the parameter.

> **Note:** Here as well another rule may work better and we should revisit in
> the future. Specifically, it might be better to allow these cases and simply
> move from the outer `var` to the inner `var` even though this still leaves
> two allocations in principle.
```

proposals/p3763.md
```diff
@@ -0,0 +1,858 @@
...
+>
+> Here, a call to `b->Clone()` would find `B.Clone` rather than `A.Clone`, and
+> so would have return type `B*`. The downside is that the vtable for `B` would
+> have two `Clone` slots, for `A.Clone` and `B.Clone`, whereas a covariant
+> return in C++ would only need a single vtable slot to express the same thing.
+
+It is an error for a class with a custom value representation to declare or
+implement a virtual function that passes `self` by value.
+
+### `extern` declarations
```
chandlerc, line 696: Just wanted to make sure that @jonmeow had a chance to look over the latest version of this and it makes sense to him as well.

jonmeow: Still looks good, it actually looks like the edits focused elsewhere since my last review. We've discussed `extern` a lot since then and I think we'll want to change it, but I'm not even sure the planned changes to `extern` really change the approach here.

Note, just for example:

> -   This rule cannot in general be checked during compilation. For example,
>     there may be no file that imports both an `extern` declaration and the
>     library that owns the entity.

This will become a requirement, that the library that owns the entity imports the `extern` declaration. As a consequence, we might be able to check whether `extern` declarations properly redeclare at compile time, although invalid code (e.g., an extern declaration that _wasn't_ included correctly) would still only be caught at link time.

But cross-library syntactic coupling issues probably still apply.

proposals/p3763.md
```diff
@@ -0,0 +1,858 @@
...
+        during compilation in order to improve the quality of diagnostics and
+        produce diagnostics earlier, but can defer harder cases until link time.
+        For example, an implementation could choose to not track the syntactic
+        form of a declaration, and perform only semantic checks during
+        compilation, deferring syntactic checking until link time.
+-   To support `extern` declarations, an additional constraint is imposed: the
+    lookup results for names used in each declaration of an entity are required
+    to be the same. This rule actually applies to all declarations of all
+    entities, but it only has an effect -- and needs to be checked -- for
+    `extern` declarations.
```
jonmeow, line 734: This feels minor, but you have the considered alternative "Disallow shadowing in `impl` after use in `api`". Isn't that expressing that this has effects and needs to be checked for `api`/`impl` splits too, or put differently, any cross-file redeclaration?

proposals/p3763.md
```diff
@@ -0,0 +1,863 @@
...
+the return value of the function from the result of the call.
+
+It is an error if a thunk is needed to wrap a function declaration with a `var`
+parameter, because otherwise a copy would always be performed when initializing
+the parameter.
+
+> **Note:** Here as well another rule may work better and we should revisit in
+> the future. Specifically, it might be better to allow these cases and simply
+> move from the outer `var` to the inner `var` even though this still leaves
+> two allocations in principle.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
> move from the outer `var` to the inner `var` even though this still leaves two
> allocations in principle.
```

proposals/p3763.md
zygoloid: Hm. For the `api` / `impl` case, we use this rule from above:

> If an entity is declared as non-`extern` in an `api` file and redeclared in an `impl` file of the same library, it is an error for the declaration to use any unqualified name that was introduced in the `impl` file.

My intent was that this would be a lot easier to check than the rule for `extern` declarations, because we don't need to do a comparison of two different sets of lookup results, we only need to track whether name lookup found a local-only result while processing a declaration -- and the idea was that if you satisfy that rule, then you don't need to check this one (and thus we don't need to track what the lookup results were for all declarations, only for `extern` declarations).

But I think maybe I was wrong. For example:

```
package Foo library "other";
namespace N;
class N.C {}
```

```
package Foo library "problem";
namespace N;
class C {}
fn N.F(c: C);
```

```
impl package Foo library "problem";
import library "other";
fn N.F(c: C) {}
```

Here, unqualified lookup for `C` in the implementation file will find `N.C`, which wasn't declared in the same file, but the declaration of `N.F` is syntactically identical to that in "other".

So I think we have two choices:

1) Change the rule above to reject a declaration in an implementation file that redeclares something in an API file unless all the lookups found something that the API file could also see (which seems like a much less clean rule), or
2) Apply this "lookup results are the same" check everywhere.

I think we should pick whichever choice is best for the implementation. How difficult would it be to track whether a lookup finds results that would have been visible in the API file? (Eg, can we put a tag on each lookup result to say whether it could only be found due to declarations *and imports* in the current file?) Conversely, how expensive would it be to perform the "lookup results are the same" check as part of all redeclaration checks?

jonmeow: That's similar to the example I had in mind. Maybe more significant than I was thinking?

Regarding (2), I think it's easy:

All parameters should get types assigned while parsing the function. The incremental expense of verifying they're the same seems small: we've already implemented the code that does it, and it's just comparing instruction types in an instruction block.

Regarding (1), I guess I'm not too sure how you intended to do the checking at present. By the proposal, you're planning to allow something like:

```
class C;
namespace NS;
fn NS.F(x: C);
fn NS.F(x: C) {}
class NC.C;
```

And disallow something like:

```
class C;
namespace NS;
fn NS.F(x: C);
fn NS.G(x: C);
fn NS.F(x: C) {}
class NC.C;
fn NS.G(x: C) {}
```

Also disallowing:

```
class C;
namespace NS;
fn NS.F(x: C);
alias NS.C = C;
fn NS.F(x: C);
```

Is that correct, that you are allowed to shadow a name so long as you don't reuse it in a redeclaration?

If shadowing that case is disallowed, you can poison name lookup. Each poison entry can be exported and imported just by the implementation. Note, I think the implementation file design doesn't say we implicitly import everything the API does: we would need to add that to make this work (to ensure that not just poison values matched, but everything inbetween matches too). Then, the error could be on the `import library "other";` that introduces a conflict with a poisoned value.

Alternately, if shadowing is allowed so long as it's not later used in a redeclaration, it isn't straight poisoning of name lookup -- that would render the first example invalid. It's not just tracking which type a name resolved to, because that wouldn't catch (3). Were you planning on, for each poisoned name, keep a vector of all declarations that used it? That same vector could be exported, and validated against the redeclaration site, although it too feels a little expensive.

zygoloid: Regarding (2): For a templated function, we don't have a type assigned to a parameter until we instantiate, so comparing parameter types is not enough to ensure that lookups produce the same result in that case. Also, it would be possible for the lookup results to differ and for us to get the same type anyway (though that's a less important problem I suppose, because we can at least be confident the declarations are consistent). I think we could do a lock-step comparison of the instruction blocks and make sure that all `name_ref`s have the same constant value, though that's also not quite right, because the constant values being equal doesn't imply that the lookups found the same thing, but I think we could make that work. Another option would be to track a list of all name lookups and the corresponding lookup results on the entity for an `extern` declaration, and compare that as part of the redeclaration logic (eg, by redoing all the lookups and checking they result in the same thing).

Regarding (1): I think your first example is invalid in this proposal as it stands because it violates

> In a declarative scope, it is an error if a name is first looked up and not found, and later introduced.

for the name `N.C`, but would be valid if the final declaration were moved to an implementation file. But as you say:

> If shadowing that case is disallowed, you can poison name lookup. Each poison entry can be exported and imported just by the implementation. Note, I think the implementation file design doesn't say we implicitly import everything the API does: we would need to add that to make this work (to ensure that not just poison values matched, but everything inbetween matches too). Then, the error could be on the import library "other"; that introduces a conflict with a poisoned value.

I think I like that idea. So then, to summarize, the plan would be:

- The implementation files import not just the API file, but also everything the API file imports, and all the poisoned names. So in general, they start in the same name lookup state that the API file ended in.
- Same-package imports check against the poisoned state as they add names. (For cross-package names, there's not a problem because unqualified lookup never looks in other packages.)

Plus additional checks for `extern` declarations.

jonmeow: > Regarding (1): I think your first example is invalid in this proposal as it stands because it violates
> 
> > In a declarative scope, it is an error if a name is first looked up and not found, and later introduced.

Good, I was missing this interaction.

> I think I like that idea. So then, to summarize, the plan would be:
> 
> * The implementation files import not just the API file, but also everything the API file imports, and all the poisoned names. So in general, they start in the same name lookup state that the API file ended in.
> * Same-package imports check against the poisoned state as they add names. (For cross-package names, there's not a problem because unqualified lookup never looks in other packages.)
> 
> Plus additional checks for `extern` declarations.

SGTM -- I think this will be a usability improvement for implementation files in general, beyond this.

chandlerc: SG to me too -- lemme know when updated with this? Or is this really for a future proposal?

jonmeow: The changes look good to me. Feel free to mark this as resolved, leaving it open in case chandlerc has more comments.

chandlerc (approved):
LG, shipping it!

PR #3764, opened 2024-03-09
Homebrew recommends installing pre-commit directly

nathany:
On macOS, Homebrew recommends `pipx` to install Python binaries, or installing `pre-commit` directly, since it is available.

Here is the message in full. 

```console
❯ pip3 install pre-commit

error: externally-managed-environment
 × This environment is externally managed
╰─> To install Python packages system-wide, try brew install
    xyz, where xyz is the package you are trying to
    install.

    If you wish to install a non-brew-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip.
    
    If you wish to install a non-brew packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
```

~~Also, unless there is a compelling reason to install a particular 3.x version of Python, it may be better to let Homebrew use whatever version people may already have installed (or let it automatically install the version they've tested).~~

Here are the details of the pre-commit Homebrew package:

```
❯ brew info pre-commit
==> pre-commit: stable 3.6.2 (bottled), HEAD
Framework for managing multi-language pre-commit hooks
https://pre-commit.com/
/opt/homebrew/Cellar/pre-commit/3.6.2 (679 files, 9.9MB) *
  Poured from bottle using the formulae.brew.sh API on 2024-03-08 at 22:22:11
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/p/pre-commit.rb
License: MIT
==> Dependencies
Required: libyaml ✔, python@3.12 ✔
```

Installing directly with Homebrew is a supported option in the pre-commit documentation. https://pre-commit.com/#install

### DIFF
--- a/docs/project/contribution_tools.md
+++ b/docs/project/contribution_tools.md
@@ -107,12 +107,9 @@ brew install \
   bazelisk \
   gh \
   llvm \
-  python@3.10
+  pre-commit
 
 # IMPORTANT: Make sure `llvm` is added to the PATH! It's separate from `brew`.
-
-# Install pre-commit.
-pip3 install pre-commit
 
 # Set up git.
 gh repo fork --clone carbon-language/carbon-lang
--- a/docs/project/contribution_tools.md
+++ b/docs/project/contribution_tools.md
@@ -107,6 +107,7 @@ brew install \
   bazelisk \
   gh \
   llvm \
+  python@3.10 \
   pre-commit
 
 # IMPORTANT: Make sure `llvm` is added to the PATH! It's separate from `brew`.


docs/project/contribution_tools.md
```diff
@@ -107,13 +107,10 @@ brew install \
   bazelisk \
   gh \
   llvm \
-  python@3.10
```
jonmeow, line 110: Carbon has a number of Python scripts, and part of the intent here was to ensure there's a good version to use. I think it's fine to switch pre-commit to a brew install, but this should probably stay.

nathany: Okay. That makes sense. Updated.

I had excluded Python because it's a transitive dependency of pre-commit, but it's probably better to be explicit.

Should we leave it at Python 3.10, or up it to 3.12? Python 3.12 is what pre-commit currently depends on in the Homebrew recipe.

jonmeow: Generally we've been saying we require 3.9 or newer (lower on the page), which is where this came from. I think part of the desire for an older version here stems from wanting to stop people from depending on a feature specific to a newer version of python, breaking compatibility. It's a delicate balance. :-/

jonmeow (approved):
Thank you!

PR #3766, opened 2024-03-10
rephrase a few sentences

nathany:
Minor tweaks while reading through the docs to hopefully improve readability.

### DIFF
--- a/docs/project/principles/safety_strategy.md
+++ b/docs/project/principles/safety_strategy.md
@@ -694,8 +694,8 @@ Advantages:
 -   Probabilistic safety should come at lower performance cost (including CPU,
     memory, and disk space).
     -   This will sometimes be significant, and as a result of multiple checks,
-        could be the difference between the hardened build mode being 50% slower
-        than the performance build mode and being 200% slower.
+        could result in the hardened build mode being 50% slower than the
+        performance build mode instead of being 200% slower.
 
 Disadvantages:
 
@@ -767,5 +767,5 @@ addition. If build modes are added for non-safety-related reasons, this may lead
 to moving build modes out of the safety strategy.
 
 **Experiment**: This can be considered an experiment. Carbon may eventually add
-more than the initial three build modes, although the reticence to add more is
+more than the initial three build modes, although the reluctance to add more is
 likely to remain.
--- a/docs/project/roadmap.md
+++ b/docs/project/roadmap.md
@@ -52,8 +52,8 @@ This goal is not necessarily about complete support for the entire language
 design, but rather enough of it to support building the realistic and
 interesting Carbon code that interoperates with C++.
 
-Some example language features that we think are key to success here but far
-from an exhaustive list:
+Here are some example language features that we think are key to success, but
+this is far from an exhaustive list:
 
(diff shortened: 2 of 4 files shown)


nathany:
FYI, I noticed that [p0257](https://github.com/carbon-language/carbon-lang/blob/trunk/proposals/p0257.md) uses the `struct` keyword a few times rather than `class`, but I'm not sure if past proposals should be updated.

docs/project/principles/safety_strategy.md
```diff
@@ -695,7 +695,7 @@ Advantages:
     memory, and disk space).
     -   This will sometimes be significant, and as a result of multiple checks,
         could be the difference between the hardened build mode being 50% slower
-        than the performance build mode and being 200% slower.
+        than the performance build mode instead of being 200% slower.
```
zygoloid: I think we should rephrase this, but I think this rephrasing doesn't quite work: we've gone from "difference between X and Y" to "difference between X instead of Y", which doesn't sound grammatical to me.

How about something like "could *result in* the hardened build mode being 50% slower than the performance build mode instead of being 200% slower"?

docs/project/principles/safety_strategy.md
```diff
@@ -767,5 +767,5 @@ addition. If build modes are added for non-safety-related reasons, this may lead
 to moving build modes out of the safety strategy.
 
 **Experiment**: This can be considered an experiment. Carbon may eventually add
-more than the initial three build modes, although the reticence to add more is
+more than the initial three build modes, although the resistance to add more is
```
zygoloid: Maybe "reluctance"? "resistance" sounds a bit too active to me.

docs/project/roadmap.md
```diff
@@ -52,8 +52,8 @@ This goal is not necessarily about complete support for the entire language
 design, but rather enough of it to support building the realistic and
 interesting Carbon code that interoperates with C++.
 
-Some example language features that we think are key to success here but far
-from an exhaustive list:
+Some example language features that we think are key to success (this is far
+from an exhaustive list):
```
zygoloid: Maybe avoid the parenthetical:

```suggestion
Here are some example language features that we think are key to success, but
this is far from an exhaustive list:
```

zygoloid (commented):
Thank you!

nathany:
@zygoloid Thanks for the feedback. I've made the updates.

zygoloid (approved):
Thank you!

(Minor point, but for future reference: we prefer updates to PRs to be pushed as separate commits rather than rebasing and amending -- the github review tools do not support the rebase / amend / force push workflow very well, unfortunately.)

nathany:
> (Minor point, but for future reference: we prefer updates to PRs to be pushed as separate commits rather than rebasing and amending -- the github review tools do not support the rebase / amend / force push workflow very well, unfortunately.)

Thanks for the tip. Does the merge queue do squash commits?

zygoloid:
> Does the merge queue do squash commits?

Yes, it does. We have some documentation of the process and our guidance [here](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/project/code_review.md#responding-to-review-comments).

PR #3768, opened 2024-03-11
Handle arrow operator

CJ-Johnson:
This change implements the check behavior for the arrow operator.

`ptr->Foo()` is rewritten as `(*ptr).Foo()` and `ptr->(X.y)` is rewritten as `(*ptr).(X.y)`


toolchain/check/handle_name.cpp
```diff
@@ -238,10 +237,56 @@ auto HandleMemberAccessExpr(Context& context, Parse::MemberAccessExprId node_id)
...
-  return context.TODO(node_id, "HandlePointerMemberAccessExpr");
+  auto [name_id, name] = context.node_stack().PopNameWithNodeId();
+
+  if (!FollowReference(context, node_id)) {
+    return false;
+  }
+
+  context.node_stack().Push(name_id, name);
+
+  return AccessMember(context, node_id);
```
zygoloid: I think this would be cleaner and better factored if `FollowReference` and `AccessMember` didn't interact with the node stack at all. That is: make each `Handle` function start by popping all of its state off the stack, then call those functions, and finish by having each `Handle` function push whatever state is relevant. That'll make it easier to check that the `Handle` functions are interacting with the node stack properly, and make the helper functions more reusable.

toolchain/check/handle_name.cpp
```diff
@@ -238,10 +237,56 @@ auto HandleMemberAccessExpr(Context& context, Parse::MemberAccessExprId node_id)
...
+          "To form a pointer type, write the `*` after the pointee type.");
+      builder.Note(TokenOnly(node_id), DerefOfType);
+    }
+    builder.Emit();
+  }
+  context.AddInstAndPush({node_id, SemIR::Deref{result_type_id, value_id}});
+  return true;
+}
+
+auto HandlePrefixOperatorStar(Context& context,
```
zygoloid: Let's move `FollowReference` and `AccessMember` to new files. Maybe `name.cpp` or `member.cpp` or something for `AccessMember`, and `pointer.cpp` for `FollowReference`. That'll let us keep this `Handle` function in `handle_operator.cpp`, which I think is where it belongs.

toolchain/check/handle_name.cpp
```diff
@@ -238,10 +237,56 @@ auto HandleMemberAccessExpr(Context& context, Parse::MemberAccessExprId node_id)
...
+                      SemIR::TypeId);
+    auto builder =
+        context.emitter().Build(TokenOnly(node_id), DerefOfNonPointer, type_id);
+    // TODO: Check for any facet here, rather than only a type.
+    if (type_id == SemIR::TypeId::TypeType) {
+      CARBON_DIAGNOSTIC(
+          DerefOfType, Note,
+          "To form a pointer type, write the `*` after the pointee type.");
+      builder.Note(TokenOnly(node_id), DerefOfType);
+    }
```
zygoloid: This is a very syntactic check -- we're looking for the case where someone writes `*T` where they meant `T*` here. So I think this belongs in the `Handle` function that's handling the unary prefix `*` syntax, rather than here in the shared code.

The preceding diagnostic ("Cannot dereference operand ...") also seems pretty closely connected to the `*` syntax, so perhaps all of this type checking should be in the caller rather than here, and we should have a different diagnostic path for dealing with `->`. Maybe we could pass in a `function_ref` for diagnosing the case where the operand is not a pointer?

toolchain/check/handle_name.cpp
```diff
@@ -185,7 +184,7 @@ auto HandleMemberAccessExpr(Context& context, Parse::MemberAccessExprId node_id)
             << " of function type";
         if (IsInstanceMethod(context.sem_ir(), function_decl->function_id)) {
           context.AddInstAndPush(
-              {node_id,
+              {Parse::MemberAccessExprId{node_id},
```
zygoloid: Do we know that the `node_id` is definitely a `MemberAccessExprId` here? I think it could be a `PointerMemberAccessExprId` instead. You'll probably need to change the definition of `BoundMethod` to allow either kind of `NodeId` -- use `Parse::NodeIdOneOf` -- and change the type of `node_id` in this function to match. (A typedef for `AnyMemberAccessExprId` might be useful.)

toolchain/check/handle_name.cpp
```diff
@@ -238,10 +237,56 @@ auto HandleMemberAccessExpr(Context& context, Parse::MemberAccessExprId node_id)
   return true;
 }
 
+auto HandleMemberAccessExpr(Context& context, Parse::MemberAccessExprId node_id)
+    -> bool {
+  return AccessMember(context, node_id);
+}
+
+static auto FollowReference(Context& context, Parse::NodeId node_id) -> bool {
```
zygoloid: Per #2006, our term for this operation is "dereference", so I think we should call this something like `DereferencePointer`.

zygoloid (commented):
Thanks, this all makes sense; comments here are largely about how to organize / factor the code.

jonmeow:
Don't forget to add/update tests. :)

Can probably also fix this TODO: https://github.com/carbon-language/carbon-lang/blob/86a7c9ff451f4cf49dd70e97d34cb6e9065f74fd/toolchain/check/testdata/class/fail_incomplete.carbon#L43-L46

CJ-Johnson:
Thanks for the review, yall! I'm having a bit of trouble with debugging now. I went ahead and pushed the changes so you could see the current status. I'm unable to run `toolchain/autoupdate_testdata.py` because some tests are still failing. I'll keep at the debug and hopefully will have a more complete update soon.

CJ-Johnson:
If you're curious, the errors I'm hitting are:
```
Running: toolchain/check/fuzzer_corpus/743bcc2582c9d359eb545c9c2234b31d20686ab1
CHECK failure at ./toolchain/check/node_stack.h:526: NodeKindToIdKind(parse_kind) == id_kind: Unexpected Id::Kind mapping for PointerMemberAccessExpr
```
and
```
Stack dump:
0.	Program arguments: /private/var/tmp/_bazel_cjjohnson/2964d713fedb8a64e79b1315050e3e56/sandbox/darwin-sandbox/7730/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/toolchain/testing/file_test.runfiles/_main/toolchain/testing/file_test --test_targets_file=toolchain/testing/file_test.tests.txt
1.	bazel test //toolchain/testing:file_test --test_arg=--file_tests=toolchain/check/testdata/class/fail_incomplete.carbon
2.	Program arguments: compile --phase=check --dump-sem-ir fail_incomplete.carbon
3.	NodeStack:
	0.	FunctionDefinitionStart -> function3
	1.	ReturnStatementStart -> no value
inst_block_stack_:
	0.	block<invalid>	{inst+0, inst+1, inst+3, inst+5, inst+9, inst+10, inst+11, inst+14, inst+24}
	1.	block10	{inst+25, inst+26}
param_and_arg_refs_stack:
	0.	block8	{inst+22}
args_type_info_stack_:
	0.	block<invalid>	{}
CHECK failure at ./toolchain/check/node_stack.h:534: RequiredParseKind == actual_kind: Expected ReturnStatementStart, found PointerMemberAccessExpr
```

toolchain/check/handle_name.cpp
```diff
@@ -19,10 +21,35 @@ auto HandleMemberAccessExpr(Context& context, Parse::MemberAccessExprId node_id)
...
 auto HandlePointerMemberAccessExpr(Context& context,
                                    Parse::PointerMemberAccessExprId node_id)
     -> bool {
-  return context.TODO(node_id, "HandlePointerMemberAccessExpr");
+  auto [name_node_id, name_id] = context.node_stack().PopNameWithNodeId();
+  auto base_id = context.node_stack().PopExpr();
+  auto dereference_id = PerformPointerDereference(context, node_id, base_id);
+  context.AddInstAndPush({node_id, dereference_id});
+  auto member_id = PerformMemberAccess(context, node_id, base_id, name_id);
+  context.node_stack().Push(node_id, member_id);
```
zygoloid: You're pushing `node_id` onto the node stack twice here, once in `AddInstAndPush` and again in `node_stack().Push`. That's responsible for at least one of the two test issues you're seeing. We don't actually want the dereferenced base to be on the stack, because we don't need to pass it to the next check action, so:

```suggestion
  auto deref_base_id = context.AddInst({node_id, dereference_id});
  auto member_id = PerformMemberAccess(context, node_id, deref_base_id, name_id);
  context.node_stack().Push(node_id, member_id);
```

I think the `AddInst` call for the dereference should be in `PerformPointerDereference` rather than here; we typically have check operations create and add instructions themselves. (That way, if they want to create multiple instructions, they don't need to treat the last one differently, and if they end up creating no instructions and just returning an operand, they can do so.)

CJ-Johnson:
Awesome! Thanks for the tips. The PR is now ready for review, IIUC :D

toolchain/check/handle_name.cpp
```diff
@@ -19,10 +21,34 @@ auto HandleMemberAccessExpr(Context& context, Parse::MemberAccessExprId node_id)
...
+  auto type_id =
+      context.GetUnqualifiedType(context.insts().Get(base_id).type_id());
+  // TODO: Check for any facet here, rather than only a type.
+  if (type_id == SemIR::TypeId::TypeType) {
+    CARBON_DIAGNOSTIC(
+        DerefOfType, Error,
+        "To form a pointer type, write the `*` after the pointee type.");
+    auto builder = context.emitter().Build(TokenOnly(node_id), DerefOfType);
+    builder.Emit();
+  }
```
zygoloid: Previously this was a note attached to another error, now it's a separate error (see changes in `check/testdata/pointer/fail_deref_type.carbon`). I think that the earlier diagnosis was better.

toolchain/check/handle_name.cpp
```diff
@@ -19,10 +21,34 @@ auto HandleMemberAccessExpr(Context& context, Parse::MemberAccessExprId node_id)
   return true;
 }
 
+auto HandlePrefixOperatorStar(Context& context,
+                              Parse::PrefixOperatorStarId node_id) -> bool {
```
zygoloid: I think this function should remain in `handle_operator.cpp` rather than moving into `handle_name.cpp`.

zygoloid (commented):
Thanks, I think we're very close here.

toolchain/check/pointer_dereference.cpp
```diff
@@ -20,13 +19,6 @@ auto PerformPointerDereference(Context& context,
...
           context.types().TryGetAs<SemIR::PointerType>(type_id)) {
     result_type_id = pointer_type->pointee_id;
-  } else if (type_id != SemIR::TypeId::Error) {
-    CARBON_DIAGNOSTIC(DerefOfNonPointer, Error,
-                      "Cannot dereference operand of non-pointer type `{0}`.",
-                      SemIR::TypeId);
-    auto builder =
-        context.emitter().Build(TokenOnly(node_id), DerefOfNonPointer, type_id);
-    builder.Emit();
   }
```
zygoloid, line 27: For the case of `not_pointer->member`, we should produce a diagnostic of some kind.

Perhaps you could make `PerformPointerDereference` take an `llvm::function_ref<auto (SemIR::TypeId not_pointer_type_id) -> void> diagnose_not_pointer` parameter and call it here if the type isn't a pointer type, and pass a lambda that emits a diagnostic from each of the two callers. That way we can avoid the duplicated check in `HandlePrefixOperatorStar`, and emit a custom diagnostic from `HandlePointerMemberAccessExpr` -- perhaps something like "Cannot apply `->` operator to non-pointer type `{0}`."

We should also make sure that case is tested.

(It'd be cool for the `->` case to recover by treating the `->` operator as a `.` instead, especially if we find that the left-hand operand's type has a member with the specified name, but I think that's too much complexity to add in this patch. So let's not do that in this PR.)

zygoloid (approved):
Awesome, thanks!

PR #3770, opened 2024-03-12
Disable modernize-use-designated-initializers check for now.

zygoloid:
This check is adding a lot of noise to the clang-tidy output for construction of `SemIR::Inst` types, and we don't want to switch to using designated initialization for them at this time.


jonmeow (approved):
I've been trying to add comments along the way about why we're disabling checks, would it make sense to add one for this too?

zygoloid:
> I've been trying to add comments along the way about why we're disabling checks, would it make sense to add one for this too?

Done.

PR #3772, opened 2024-03-13
Remove legacy repo_name settings

jonmeow:
I'd kept these in to separate the bazel module update from the BUILD file changes, then forgot about it. I think all of these can be cleanly removed now. I think it's something we should clean up for consistency with the bazel central repository names; I think it's best to reduce that divergence.

llvm_zlib and llvm_zstd remain because of how llvm depends on the particular names.

### DIFF
--- a/bazel/check_deps/check_non_test_cc_deps.py
+++ b/bazel/check_deps/check_non_test_cc_deps.py
@@ -82,7 +82,7 @@ for dep in deps:
     # message.
     if repo_base in (
         "@google_benchmark",
-        "@com_github_protocolbuffers_protobuf",
+        "@protobuf",
         "@abseil-cpp",
         "@googletest",
     ):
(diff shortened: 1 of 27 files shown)


bazel/check_deps/check_non_test_cc_deps.py
```diff
@@ -81,10 +81,10 @@
     # exist. Specially diagnose them to try to provide a more helpful
     # message.
     if repo_base in (
-        "@com_github_google_benchmark",
+        "@google_benchmark",
         "@com_github_protocolbuffers_protobuf",
```
zygoloid: Where does this name come from? This is the only mention of it in our codebase.

zygoloid (approved):

bazel/check_deps/check_non_test_cc_deps.py
jonmeow: Looks like this was simply incorrect before, replacing.

PR #3773, opened 2024-03-13
Add support to fix_cc_deps for canonicalizing #include style.

jonmeow:
Note this builds on #3772 (I was partly cleaning up because I was looking at this again).

This stops printing "Ignore missing" for ignored includes because it was feeling noisy. This has been bugging me for a bit, and now I'm here. To get an idea of what I mean, here it is masking a fix:

```
Querying bazel for Carbon targets...
Querying bazel for external targets...
Building header map...
Building generated file list...
Parsing headers from source files...
Ignored missing '#include "testing/fuzzing/carbon.pb.h"' in 'explorer/fuzzing/ast_to_proto_main.cpp'
Ignored missing '#include "testing/fuzzing/carbon.pb.h"' in 'explorer/fuzzing/ast_to_proto.h'
Ignored missing '#include "testing/fuzzing/carbon.pb.h"' in 'explorer/fuzzing/fuzzer_util.h'
Fixing include format in 'testing/file_test/file_test_base.h': '#include "gtest/gtest.h"' to '#include <gtest/gtest.h>'
Ignored missing '#include "testing/fuzzing/carbon.pb.h"' in 'testing/fuzzing/proto_to_carbon.cpp'
Ignored missing '#include "testing/fuzzing/carbon.pb.h"' in 'testing/fuzzing/proto_to_carbon.h'
Ignored missing '#include "testing/fuzzing/carbon.pb.h"' in 'testing/fuzzing/proto_to_carbon_test.cpp'
Done!
```

To help make this work, I'm also adjusting the gtest handling to use the same EXTERNAL_REPO logic as the rest. I don't recall if there'd been some other reason for the special casing, but AFAICT it works fine (auto-adds a missing gtest dependency) this way.

### DIFF
--- a/scripts/fix_cc_deps.py
+++ b/scripts/fix_cc_deps.py
@@ -27,6 +27,15 @@ class ExternalRepo(NamedTuple):
     remap: Callable[[str], str]
     # The target expression to gather rules for within the repo.
     target: str
+    # Whether to use "" or <> for the include.
+    use_system_include: bool = False
+
+
+class RuleChoice(NamedTuple):
+    # Whether to use "" or <> for the include.
+    use_system_include: bool
+    # Possible rules that may be used.
+    rules: Set[str]
 
 
 # Maps external repository names to a method translating bazel labels to file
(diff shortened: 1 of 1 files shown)


scripts/fix_cc_deps.py
```diff
@@ -194,42 +223,56 @@ def get_missing_deps(
...
+                        f"{', '.join(rule_choice.rules)}"
+                    )
+                    ambiguous = True
+                # Use the single dep without removing it.
+                missing_deps.add(next(iter(rule_choice.rules)))
+
+            # If the include style should change, update file content.
+            if is_system_include != rule_choice.use_system_include:
+                if rule_choice.use_system_include:
+                    new_include = f"#include <{header}>"
```
geoffromer, line 264: Could this use single-quotes, for consistency with the `else` branch?

geoffromer (approved):

scripts/fix_cc_deps.py
jonmeow: The black formatter is making this choice, preferring `"` for strings by default. Note black is really great, and will also translate strings like `"#include \"{header}\""` to `'#include "{header}"'`

PR #3774, opened 2024-03-13
Factor member name lookup out of `handle_name.cpp`.

zygoloid:
The member access logic is fairly large, and will be growing with the addition of impl lookup. Factor it out to separate the logic for dealing with handling the parse node and updating the node stack from the logic that checks and builds the member access expression.


jonmeow (approved):
Note, I'm not looking too closely since this is just moving code.

PR #3775, opened 2024-03-13
Refactor member lookup in preparation for adding impl lookup.

zygoloid:
A couple of minor functional changes here:

- We now always create a `name_ref` for the name referred to by the right-hand operand of member access. Previously we skipped creating this instruction if the referenced name was a field, and just created the field access instruction. This makes our processing of member accesses and our SemIR representation a bit more uniform.
- We now perform lookup into the type of the left-hand operand if it's any type with a scope, not just for classes. This means we do lookup into interface types. However, doing so isn't really useful yet because it always finds an associated entity that isn't usable by itself. This changes the diagnostic in `toolchain/check/testdata/interface/fail_todo_facet_lookup.carbon`.

### DIFF
--- a/toolchain/check/handle_name.cpp
+++ b/toolchain/check/handle_name.cpp
@@ -5,7 +5,6 @@
 #include "toolchain/check/context.h"
 #include "toolchain/check/member_access.h"
 #include "toolchain/lex/token_kind.h"
-#include "toolchain/parse/node_ids.h"
 #include "toolchain/sem_ir/inst.h"
 #include "toolchain/sem_ir/typed_insts.h"
 
--- a/toolchain/check/handle_name.cpp
+++ b/toolchain/check/handle_name.cpp
@@ -5,6 +5,7 @@
 #include "toolchain/check/context.h"
 #include "toolchain/check/member_access.h"
 #include "toolchain/lex/token_kind.h"
+#include "toolchain/parse/node_ids.h"
 #include "toolchain/sem_ir/inst.h"
 #include "toolchain/sem_ir/typed_insts.h"
 
--- a/toolchain/check/member_access.cpp
+++ b/toolchain/check/member_access.cpp
@@ -61,6 +61,8 @@ static auto GetAsNameScope(Context& context, SemIR::InstId base_id)
   return std::nullopt;
 }
 
+// Returns the index of the specified class element within the class's
+// representation.
 static auto GetClassElementIndex(Context& context, SemIR::InstId element_id)
     -> SemIR::ElementIndex {
   auto element_inst = context.insts().Get(element_id);
@@ -91,6 +93,8 @@ static auto IsInstanceMethod(const SemIR::File& sem_ir,
   return false;
 }
 
+// Performs a member name lookup into the specified scope. If the scope is
+// invalid, assume an error has already been diagnosed, and return BuiltinError.
 static auto LookupMemberNameInScope(Context& context,
                                     Parse::MemberAccessExprId node_id,
                                     SemIR::InstId /*base_id*/,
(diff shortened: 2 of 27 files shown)


toolchain/check/handle_name.cpp
```diff
@@ -5,6 +5,7 @@
 #include "toolchain/check/context.h"
 #include "toolchain/check/member_access.h"
 #include "toolchain/lex/token_kind.h"
+#include "toolchain/parse/node_ids.h"
```
jonmeow: This feels a bit like a stray edit...

toolchain/check/member_access.cpp
```diff
@@ -87,20 +91,90 @@ static auto IsInstanceMethod(const SemIR::File& sem_ir,
...
+                                    Parse::MemberAccessExprId node_id,
+                                    SemIR::InstId /*base_id*/,
+                                    SemIR::NameId name_id,
+                                    SemIR::NameScopeId name_scope_id)
+    -> SemIR::InstId {
+  auto inst_id = name_scope_id.is_valid() ? context.LookupQualifiedName(
+                                                node_id, name_id, name_scope_id)
+                                          : SemIR::InstId::BuiltinError;
+  auto inst = context.insts().Get(inst_id);
+  // TODO: Track that this instruction was named within `base_id`.
```
jonmeow: I may be missing something here... Isn't this just a matter of adding a tombstone to name_scope_id if LookupQualifiedName failed? Why is base_id necessary?

Is this even something that should be done here, versus as part of LookupQualifiedName?

Note, part of my thinking is that if `base_id` is an incomplete type and `name_scope_id` is invalid, the lookup will always be an error, so not sure how important it is to track that the field was named. (it will be diagnosed, and probably should not create a conflict with a later definition)

toolchain/check/member_access.cpp
```diff
@@ -87,20 +91,90 @@ static auto IsInstanceMethod(const SemIR::File& sem_ir,
   return false;
 }
 
+static auto LookupMemberNameInScope(Context& context,
```
jonmeow, line 98: Please add comments to the new functions which explain their behavior.

toolchain/check/member_access.cpp
```diff
@@ -87,20 +91,90 @@ static auto IsInstanceMethod(const SemIR::File& sem_ir,
...
+        << " of unbound element type";
+    auto index = GetClassElementIndex(context, element_id.inst_id());
+    auto access_id = context.AddInst(
+        {node_id, SemIR::ClassElementAccess{
+                      unbound_element_type->element_type_id, base_id, index}});
+    if (SemIR::GetExprCategory(context.sem_ir(), base_id) ==
+            SemIR::ExprCategory::Value &&
+        SemIR::GetExprCategory(context.sem_ir(), access_id) !=
+            SemIR::ExprCategory::Value) {
+      // Class element access on a value expression produces an
```
jonmeow: You're lowering the indent of this, maybe it could re-wrap?

toolchain/check/member_access.cpp
```diff
@@ -207,24 +217,26 @@ auto PerformMemberAccess(Context& context, Parse::MemberAccessExprId node_id,
...
-  // Should only be reached on error.
-  return SemIR::InstId::BuiltinError;
+  // Perform lookup into the base type.
+  auto member_id = LookupMemberNameInScope(context, node_id, base_id, name_id,
+                                           *name_scope_id);
+
+  // Perform instance binding if we found an instance member.
+  member_id = PerformInstanceBinding(context, node_id, base_id, member_id);
+
+  return member_id;
```
jonmeow, line 247: FWIW this might be nested less if you switch the `if(!` above; you could also put `name_scope_id` as a condition-scoped variable.

jonmeow (approved):
Looks good.

toolchain/check/handle_name.cpp
zygoloid: Hm. VS Code added that. It's right -- we're using types from this file -- but we don't include it for other handle_* files, so I'm removing it for consistency.

toolchain/check/member_access.cpp
zygoloid: This TODO isn't about that at all :)

Currently, for `A.B` we create a `name_ref` that refers to `B`, but doesn't mention that we looked for it in `A` at all.  The point of the TODO is that the `name_ref` instruction created on the next line should track the `base_id`.

I've changed the TODO to try to make this clearer (but this is pre-existing, just moved).

toolchain/check/member_access.cpp
zygoloid: I think it's easier to follow if the error case is the early exit, rather than the success case.

toolchain/check/member_access.cpp
zygoloid: Done.

toolchain/check/member_access.cpp
zygoloid: Done.

PR #3776, opened 2024-03-14
Basic support for looking up impl members when naming an associated entity.

zygoloid:
When a member access names an interface member, perform impl lookup to find the impl and its corresponding member.


toolchain/check/member_access.cpp
```diff
@@ -93,12 +90,111 @@ static auto IsInstanceMethod(const SemIR::File& sem_ir,
...
+                              SemIR::InstId member_id) -> SemIR::InstId {
+  auto witness_id =
+      LookupInterfaceWitness(context, type_const_id, assoc_type.interface_id);
+  if (!witness_id.is_valid()) {
+    CARBON_DIAGNOSTIC(MissingImplInMemberAccess, Error,
+                      "Cannot access member of interface {0} in type {1} "
+                      "that does not implement that interface.",
+                      SemIR::NameId, std::string);
+    context.emitter().Emit(
+        context.insts().GetNodeId(member_id), MissingImplInMemberAccess,
```
jonmeow:
```suggestion
        member_id, MissingImplInMemberAccess,
```

Just a reminder, but GetNodeId should rarely be required. Probably I should also change TODO to use a SemIRLocation.

jonmeow (approved):

PR #3777, opened 2024-03-14
Add upcoming conferences section to README

nathany:
Add upcoming conferences section to README so people are aware and can consider attending.


chandlerc (approved):
Thanks!

PR #3787, opened 2024-03-15
Factor `IdKind` enum out of node stack.

zygoloid:
Provide a general mechanism for determining the kind of the args of an instruction. Use this to simplify instruction profiling a little. The intent is to also use this mechanism as the basis of a substitution mechanism, which will be part of a future patch.

Note that this causes us to do three table lookups and two indirect calls in inst_profile per instruction, instead of one table lookup and one indirect call. We can revisit this if it shows up in profiles.

### DIFF
--- a/toolchain/sem_ir/id_kind.h
+++ b/toolchain/sem_ir/id_kind.h
@@ -36,7 +36,7 @@ class TypeEnum {
   // Accesses the type given an enum value.
   template <RawEnumType K>
     requires(K != RawEnumType::Invalid)
-  using AsType = __type_pack_element<static_cast<size_t>(K), Types...>;
+  using TypeFor = __type_pack_element<static_cast<size_t>(K), Types...>;
 
   // Workarond for Clang bug https://github.com/llvm/llvm-project/issues/85461
   template <RawEnumType Value>
--- a/toolchain/sem_ir/id_kind.h
+++ b/toolchain/sem_ir/id_kind.h
@@ -1,0 +1,120 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#ifndef CARBON_TOOLCHAIN_SEM_IR_ID_KIND_H_
+#define CARBON_TOOLCHAIN_SEM_IR_ID_KIND_H_
+
+#include <algorithm>
+
+#include "toolchain/sem_ir/ids.h"
+
+namespace Carbon::SemIR {
+
+// An enum whose values are the specified types.
+template <typename... Types>
+class TypeEnum {
+ public:
+  static constexpr std::size_t NumTypes = sizeof...(Types);
+  static constexpr std::size_t NumValues = NumTypes + 2;
+
+  static_assert(NumValues <= 256, "Too many types for raw enum.");
+
+  // The underlying raw enumeration type.
+  enum class RawEnumType : uint8_t {
+    // The first sizeof...(Types) values correspond to the types.
+
+    // An explicitly invalid value.
+    Invalid = NumTypes,
+
+    // Indicates that no type should be used.
+    // TODO: This doesn't really fit the model of this type, but it's convenient
+    // for all of its users.
+    None,
+  };
+
+  // Accesses the type given an enum value.
+  template <RawEnumType K>
+    requires(K != RawEnumType::Invalid)
+  using AsType = __type_pack_element<static_cast<size_t>(K), Types...>;
+
+  // Workarond for Clang bug https://github.com/llvm/llvm-project/issues/85461
+  template <RawEnumType Value>
+  static constexpr auto FromRaw = TypeEnum(Value);
+
+  // Names for the `Invalid` and `None` enumeration values.
+  static constexpr const TypeEnum& Invalid = FromRaw<RawEnumType::Invalid>;
+  static constexpr const TypeEnum& None = FromRaw<RawEnumType::None>;
+
+  // Accesses the enumeration value for the type `IdT`. If `AllowInvalid` is
+  // set, any unexpected type is mapped to `Invalid`, otherwise an invalid type
+  // results in a compile error.
+  //
+  // The `Self` parameter is an implementation detail to allow `ForImpl` to be
+  // defined after this template, and should not be specified.
+  template <typename IdT, bool AllowInvalid = false, typename Self = TypeEnum>
+  static constexpr auto For = Self::template ForImpl<IdT, AllowInvalid>();
+
+  // This bool indicates whether the specified type corresponds to a value in
+  // this enum.
+  template <typename IdT>
+  static constexpr bool Contains = For<IdT, true>.is_valid();
+
+  // Explicitly convert from the raw enum type.
+  explicit constexpr TypeEnum(RawEnumType value) : value_(value) {}
+
+  // Implicitly convert to the raw enum type, for use in `switch`.
+  //
+  // NOLINTNEXTLINE(google-explicit-constructor)
+  constexpr operator RawEnumType() const { return value_; }
+
+  // Conversion to bool is deleted to prevent direct use in an `if` condition
+  // instead of comparing with another value.
+  explicit operator bool() const = delete;
+
+  // Returns the raw enum value.
+  constexpr auto ToRaw() const -> RawEnumType { return value_; }
+
+  // Returns a value that can be used as an array index. Returned value will be
+  // < NumValues.
+  constexpr auto ToIndex() const -> std::size_t {
+    return static_cast<std::size_t>(value_);
+  }
+
+  // Returns whether this is a valid value, not `Invalid`.
+  constexpr auto is_valid() const -> bool {
+    return value_ != RawEnumType::Invalid;
+  }
+
+ private:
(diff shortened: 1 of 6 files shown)


toolchain/sem_ir/inst_profile.cpp
```diff
@@ -65,54 +65,32 @@ static auto RealProfileArgFunction(llvm::FoldingSetNodeID& id,
...
-  ProfileArg1(id, sem_ir, arg1);
+// Profiles the given instruction argument, which is of the specified kind.
+static auto ProfileArg(llvm::FoldingSetNodeID& id, const File& sem_ir,
+                       IdKind arg_kind, int32_t arg) -> void {
+  static constexpr std::array<ProfileArgFunction*, IdKind::NumValues>
+      ProfileFunctions = [] {
+        std::array<ProfileArgFunction*, IdKind::NumValues> array = {};
+        for (auto& elem : array) {
+          elem = DefaultProfileArgFunction;
+        }
```
jonmeow:
```suggestion
        std::array<ProfileArgFunction*, IdKind::NumValues> array;
        array.fill(DefaultProfileArgFunction);
```

Is this what you want? https://en.cppreference.com/w/cpp/container/array/fill

Is the default init (`= {}`) necessary when you `fill`?

toolchain/sem_ir/id_kind.h
```diff
@@ -0,0 +1,120 @@
...
+  template <typename IdT>
+  static constexpr bool Contains = For<IdT, true>.is_valid();
+
+  // Explicitly convert from the raw enum type.
+  explicit constexpr TypeEnum(RawEnumType value) : value_(value) {}
+
+  // Implicitly convert to the raw enum type, for use in `switch`.
+  //
+  // NOLINTNEXTLINE(google-explicit-constructor)
+  constexpr operator RawEnumType() const { return value_; }
```
jonmeow, line 69: Is this in use? I'm trying to figure out how it'd work: RawEnumType only contains Invalid and None, wouldn't this be quirky for a switch?

If this is removed, can the `operator bool() = delete` be deleted?

toolchain/sem_ir/id_kind.h
```diff
@@ -0,0 +1,120 @@
...
+
+  // Conversion to bool is deleted to prevent direct use in an `if` condition
+  // instead of comparing with another value.
+  explicit operator bool() const = delete;
+
+  // Returns the raw enum value.
+  constexpr auto ToRaw() const -> RawEnumType { return value_; }
```
jonmeow, line 76: Why AsType but ToRaw/ToIndex?

toolchain/sem_ir/id_kind.h
```diff
@@ -0,0 +1,120 @@
...
+
+  // Returns a value that can be used as an array index. Returned value will be
+  // < NumValues.
+  constexpr auto ToIndex() const -> std::size_t {
```
jonmeow, line 80: Comping this with EnumBase::AsInt, is this a spot where matching APIs would make sense? Did this need to be `size_t` instead of the underlying `uint8_t` to avoid conversions?

toolchain/sem_ir/id_kind.h
```diff
@@ -0,0 +1,120 @@
...
+    requires(K != RawEnumType::Invalid)
+  using AsType = __type_pack_element<static_cast<size_t>(K), Types...>;
+
+  // Workarond for Clang bug https://github.com/llvm/llvm-project/issues/85461
+  template <RawEnumType Value>
+  static constexpr auto FromRaw = TypeEnum(Value);
+
+  // Names for the `Invalid` and `None` enumeration values.
+  static constexpr const TypeEnum& Invalid = FromRaw<RawEnumType::Invalid>;
+  static constexpr const TypeEnum& None = FromRaw<RawEnumType::None>;
```
jonmeow, line 47: Why this, particularly with the `&`, versus the out-of-line constexpr we do elsewhere? Is this something we should actually be switching the other cases to?

toolchain/sem_ir/inst.h
```diff
@@ -209,6 +210,18 @@ class Inst : public Printable<Inst> {
...
   auto type_id() const -> TypeId { return type_id_; }
 
+  // Gets the kinds of IDs used for arg0 and arg1 of the specified kind of
+  // instruction.
+  //
+  // TODO: This would ideally live on InstKind, but can't be there for layering
+  // reasons.
+  static auto ArgKinds(InstKind kind) -> std::pair<IdKind, IdKind> {
+    return ArgKindTable[kind.AsInt()];
+  }
```
jonmeow, line 220: Does this need to be public?

toolchain/sem_ir/inst.cpp
```diff
@@ -32,4 +32,20 @@ auto Inst::Print(llvm::raw_ostream& out) const -> void {
   out << "}";
 }
 
+template <typename InstKind, int Arg>
```
jonmeow: ArgIndex?

toolchain/sem_ir/inst.cpp
```diff
@@ -32,4 +32,20 @@ auto Inst::Print(llvm::raw_ostream& out) const -> void {
...
+static constexpr auto IdKindFor() -> IdKind {
```
jonmeow:
```suggestion
// Returns the IdKind of an instruction's argument, or None if the index exceeds the argument count.
template <typename InstKind, int Arg>
static constexpr auto IdKindFor() -> IdKind {
```

toolchain/sem_ir/id_kind.h
```diff
@@ -0,0 +1,120 @@
...
+    static_assert(Index != NumTypes || AllowInvalid,
+                  "Unexpected type passed to TypeEnum::For<...>");
+    return TypeEnum(static_cast<RawEnumType>(Index));
+  }
+
+  RawEnumType value_;
+};
+
+// An enum of all the ID types used as instruction operands.
+using IdKind = TypeEnum<
```
jonmeow, line 108: FWIW, the contrast here between IdKind and TypeEnum -- from a naming perspective -- feels odd to me.

jonmeow (approved):
Looks good! I have some comments for cleanup, but nothing that seems worth slowing down for.

toolchain/sem_ir/inst_profile.cpp
zygoloid: Thanks! I didn't know about `fill`. And the `= {}` is a relic of a C++17-and-before rule; we don't need that thanks to a language change by @CJ-Johnson =)

toolchain/sem_ir/id_kind.h
zygoloid: This is used in #3788.

toolchain/sem_ir/id_kind.h
zygoloid: They operate on different things: `x.ToY` converts `x` to `Y`, whereas `IdKind::AsType<K>` converts `K` to a type.

Hm, on reflection, I think `IdKind::TypeFor<K>` would be a better name for `AsType`, and more consistent with `IdKind::For<T>`. Will make that change.

toolchain/sem_ir/id_kind.h
zygoloid: No, unfortunately this only works in a class template. Also, this is needed in a class template to work around the clang bug referenced in the comment -- Clang gets confused by the technique we use elsewhere if you use it in a class template. I've updated the toolchain doc to describe this technique.

toolchain/sem_ir/inst.h
zygoloid: I intended for this to be part of the public API of `Inst` -- I think it's reasonable to want to query this when you don't have an `Inst` instance, just a kind.

toolchain/sem_ir/inst.cpp
zygoloid: Done.

toolchain/sem_ir/id_kind.h
zygoloid: Renamed to `TypeFor`.

toolchain/sem_ir/id_kind.h
zygoloid: I think this is different because this is specifically intended for use as an array index, whereas `AsInt` just returns the underlying integer value, which could be negative. That's also why I'm using `std::size_t` -- it seems like the most obvious type to use for an array index.

The `To` vs `As` seems like something we should be consistent on, but `AsInt` is the only `As`-prefixed function in the codebase, and we also only have one existing `To` name, so we don't have a strong precedent here. Let's discuss this in #naming then consolidate on one or the other.

toolchain/sem_ir/id_kind.h
zygoloid: Noted. Though I don't have a better idea right now; suggestions welcome :)

jonmeow: IdKind[Enum]Base?

toolchain/sem_ir/id_kind.h
jonmeow: FWIW, my sense is we're pretty typically using `AsInt` for array indices.

toolchain/sem_ir/inst.h
jonmeow: If you're not using it, I might still suggest it be private. It'd kind of hammer on the TODO

PR #3788, opened 2024-03-15
Substitute `Self` in associated function signatures before checking them against `impl`s.

zygoloid:
Add a general substitution mechanism to support substituting symbolic bindings with their values throughout symbolic constants and, more specifically, types. This is done by decomposing the constant instruction into its operands, substituting into the operands, and then rebuilding the constant value by invoking the constant evaluator.

### DIFF
--- a/toolchain/check/subst.cpp
+++ b/toolchain/check/subst.cpp
@@ -133,6 +133,10 @@ static auto Rebuild(Context& context, Worklist& worklist, SemIR::InstId inst_id)
     return inst_id;
   }
 
+  // TODO: Updating the arguments might result in the instruction having a
+  // different type. We should consider either recomputing the type or
+  // substituting into it. In the latter case, consider caching, as we may
+  // substitute into related types repeatedly.
   inst.SetArgs(arg0, arg1);
   auto result_id = TryEvalInst(context, SemIR::InstId::Invalid, inst);
   CARBON_CHECK(result_id.is_constant())
--- a/toolchain/check/subst.cpp
+++ b/toolchain/check/subst.cpp
@@ -29,15 +29,22 @@ struct WorklistItem {
 // into. For details of the algorithm used here, see `SubstConstant`.
 class Worklist {
  public:
-  auto Get(int index) -> WorklistItem& { return worklist[index]; }
-  auto Size() -> int { return worklist.size(); }
-  auto Push(SemIR::InstId inst_id) -> void {
-    worklist.push_back({.inst_id = inst_id,
-                        .is_expanded = false,
-                        .next_index = static_cast<int>(worklist.size() + 1)});
-    CARBON_CHECK(worklist.back().next_index > 0) << "Constant too large.";
+  explicit Worklist(SemIR::InstId root_id) {
+    worklist_.push_back(
+        {.inst_id = root_id, .is_expanded = false, .next_index = -1});
   }
-  auto Pop() -> SemIR::InstId { return worklist.pop_back_val().inst_id; }
+
+  auto operator[](int index) -> WorklistItem& { return worklist_[index]; }
+  auto size() -> int { return worklist_.size(); }
+  auto back() -> WorklistItem& { return worklist_.back(); }
+
+  auto Push(SemIR::InstId inst_id) -> void {
+    worklist_.push_back({.inst_id = inst_id,
+                         .is_expanded = false,
+                         .next_index = static_cast<int>(worklist_.size() + 1)});
+    CARBON_CHECK(worklist_.back().next_index > 0) << "Constant too large.";
+  }
+  auto Pop() -> SemIR::InstId { return worklist_.pop_back_val().inst_id; }
 
  private:
   // Constants can get pretty large, so use a large worklist. This should be
@@ -158,9 +165,7 @@ auto SubstConstant(Context& context, SemIR::ConstantId const_id,
     return const_id;
   }
 
-  Worklist worklist;
-  worklist.Push(const_id.inst_id());
-  worklist.Get(0).next_index = -1;
+  Worklist worklist(const_id.inst_id());
 
   // For each instruction that forms part of the constant, we will visit it
   // twice:
@@ -176,7 +181,7 @@ auto SubstConstant(Context& context, SemIR::ConstantId const_id,
   // instruction will not need to be rebuilt.
   int index = 0;
   while (index != -1) {
-    auto& item = worklist.Get(index);
+    auto& item = worklist[index];
 
     if (item.is_expanded) {
       // Rebuild this item if necessary. Note that this might pop items from the
@@ -216,14 +221,14 @@ auto SubstConstant(Context& context, SemIR::ConstantId const_id,
     // modifies the worklist, so it's not safe to use `item` after
     // `ExpandOperands` returns.
     item.is_expanded = true;
-    int first_operand = worklist.Size();
+    int first_operand = worklist.size();
     int next_index = item.next_index;
     ExpandOperands(context, worklist, item.inst_id);
 
     // If there are any operands, go and update them before rebuilding this
     // item.
-    if (worklist.Size() > first_operand) {
-      worklist.Get(worklist.Size() - 1).next_index = index;
+    if (worklist.size() > first_operand) {
+      worklist.back().next_index = index;
       index = first_operand;
     } else {
       // No need to rebuild this instruction.
@@ -231,9 +236,9 @@ auto SubstConstant(Context& context, SemIR::ConstantId const_id,
     }
   }
 
-  CARBON_CHECK(worklist.Size() == 1)
+  CARBON_CHECK(worklist.size() == 1)
       << "Unexpected data left behind in work list";
-  return context.constant_values().Get(worklist.Get(0).inst_id);
+  return context.constant_values().Get(worklist.back().inst_id);
 }
 
 auto SubstType(Context& context, SemIR::TypeId type_id,
--- a/toolchain/check/subst.cpp
+++ b/toolchain/check/subst.cpp
@@ -12,6 +12,7 @@ namespace Carbon::Check {
 
 namespace {
 
+// Information about an instruction that we are substituting into.
 struct WorklistItem {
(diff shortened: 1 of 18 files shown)


toolchain/sem_ir/id_kind.h
```diff
@@ -0,0 +1,123 @@
...
+  static constexpr std::size_t NumTypes = sizeof...(Types);
+  static constexpr std::size_t NumValues = NumTypes + 2;
+
+  static_assert(NumValues <= 256, "Too many types for raw enum.");
+
+  // The underlying raw enumeration type.
+  //
+  // The enum_extensibility attribute indicates that this enum is intended to
+  // take values that do not correspond to its declared enumerators.
+  enum class [[clang::enum_extensibility(open)]] RawEnumType : uint8_t{
```
zygoloid: Bad formatting here filed as a [clang-format bug](https://github.com/llvm/llvm-project/issues/85476).

toolchain/check/subst.h
```diff
@@ -0,0 +1,35 @@
...
+  SemIR::InstId bind_id;
+  // The replacement constant value to substitute.
+  SemIR::ConstantId replacement_id;
+};
+
+using Substitutions = llvm::ArrayRef<Substitution>;
+
+// Replaces the `BindSymbolicName` instruction `bind_id` with `replacement_id`
+// throughout the constant `const_id`, and returns the substituted value.
+auto SubstConstant(Context& context, SemIR::ConstantId const_id,
```
jonmeow, line 25: It feels odd to me that you do `Substitution` but `Subst`. `Subst` seems like an unusual abbreviation (I'm used to things like `re.sub` from Python); maybe just write out `Substitute`?

On locals I don't think about it as much as on function names, FWIW. Maybe I'll just get used to it. But then it's probably an abbreviation to add to the arch doc.

toolchain/check/subst.cpp
```diff
@@ -0,0 +1,241 @@
...
+                        .next_index = static_cast<int>(worklist.size() + 1)});
+    CARBON_CHECK(worklist.back().next_index > 0) << "Constant too large.";
+  }
+  auto Pop() -> SemIR::InstId { return worklist.pop_back_val().inst_id; }
+
+ private:
+  // Constants can get pretty large, so use a large worklist. This should be
+  // about 4KiB, which should be small enough to comfortably fit on the stack,
+  // but large enough that it's unlikely that we'll need a heap allocation.
+  llvm::SmallVector<WorklistItem, 512> worklist;
```
jonmeow:
```suggestion
  llvm::SmallVector<WorklistItem, 512> worklist_;
```

toolchain/check/subst.cpp
```diff
@@ -0,0 +1,241 @@
...
+  // root, this is -1.
+  int next_index : 31;
+};
+
+// A list of instructions that we're currently in the process of substituting
+// into. For details of the algorithm used here, see `SubstConstant`.
+class Worklist {
+ public:
+  auto Get(int index) -> WorklistItem& { return worklist[index]; }
+  auto Size() -> int { return worklist.size(); }
```
jonmeow: FWIW I'd be fine with this as size()

toolchain/sem_ir/id_kind.h
jonmeow: Given the odd formatting here (`uint8_t{` and indent), would it be worth something like:

```suggestion
// TODO: Works around a clang-format bug:
// https://github.com/llvm/llvm-project/issues/85476
#define CARBON_OPEN_ENUM [[clang::enum_extensibility(open)]]
  enum class CARBON_OPEN_ENUM RawEnumType : uint8_t {
```

I think this makes clang-format happier. (note, this may not be the best place for the #define)

toolchain/check/subst.cpp
```diff
@@ -0,0 +1,241 @@
...
+  // Constants can get pretty large, so use a large worklist. This should be
+  // about 4KiB, which should be small enough to comfortably fit on the stack,
+  // but large enough that it's unlikely that we'll need a heap allocation.
+  llvm::SmallVector<WorklistItem, 512> worklist;
+};
+
+}  // namespace
+
+// Pushes the specified operand onto the worklist.
+static auto PushOperand(Context& context, Worklist& worklist,
```
jonmeow, line 59: Had you considered making these as members of a class rather than passing `Context`+`Worklist` as an argument?

toolchain/check/subst.cpp
```diff
@@ -0,0 +1,241 @@
...
+    case SemIR::IdKind::For<SemIR::TypeId>:
+      return context.GetTypeIdForTypeInst(worklist.Pop()).index;
+    case SemIR::IdKind::For<SemIR::InstBlockId>: {
+      auto old_inst_block_id = static_cast<SemIR::InstBlockId>(arg);
+      auto size = context.inst_blocks().Get(old_inst_block_id).size();
+      SemIR::CopyOnWriteInstBlock new_inst_block(context.sem_ir(),
+                                                 old_inst_block_id);
+      for (auto i : llvm::index_range(0, size)) {
+        new_inst_block.Set(size - i - 1, worklist.Pop());
+      }
```
jonmeow: Can you:

```suggestion
      for (auto i : llvm::reverse(llvm::seq(size))) {
        new_inst_block.Set(i, worklist.Pop());
      }
```

If that works, it'd apply below too.

FWIW, there's probably an option here where you add two APIs to worklist:
- `Back(N)` to return an ArrayRef of the last N items
- `Pop(N)` to pop the last N items

Then do:

```
for (auto [i, inst_id] : llvm::enumerate(llvm::reverse(worklist.Back(N))) {
  new_inst_block.Set(i, N);
}
worklist.Pop(N);
```

But I'm not sure this is really worth it...

toolchain/check/subst.cpp
```diff
@@ -0,0 +1,241 @@
...
+  }
+
+  if (!const_id.is_symbolic()) {
+    // A template constant can't contain a reference to a symbolic binding.
+    return const_id;
+  }
+
+  Worklist worklist;
+  worklist.Push(const_id.inst_id());
+  worklist.Get(0).next_index = -1;
```
jonmeow: Maybe move this to a constructor which can do it without a separate `Get`?

toolchain/check/subst.cpp
```diff
@@ -0,0 +1,241 @@
...
+      for (auto [bind_id, replacement_id] : substitutions) {
+        if (item.inst_id == bind_id) {
+          // This is the binding we're replacing. Perform substitution.
+          item.inst_id = replacement_id.inst_id();
+          break;
+        }
+      }
+
+      // Even if it's not being substituted, don't look through it. Its constant
+      // value doesn't spend on its operand.
```
jonmeow:
```suggestion
      // If it's not being substituted, don't look through it. Its constant
      // value doesn't depend on its operand.
```

Minor phrasing suggestion, feel free to reject if I'm misunderstanding, although I think "spend" is a typo.

jonmeow (approved):
I'm glad to see the progress here!

toolchain/check/subst.h
zygoloid: I already added it to the arch doc. Maybe it seems unsurprising to me because it's used heavily in Clang. How about we check back and see if you've got used to it; if it remains weird then we can unabbreviate it. Based on experience with explorer, I think this is going to be something we do a lot, though maybe that also argues for making the name `Substitutions` shorter.

toolchain/sem_ir/id_kind.h
```diff
@@ -21,16 +21,22 @@ class TypeEnum {
...
-    // for all of its users.
-    None,
+  //
+  // The enum_extensibility attribute indicates that this enum is intended to
+  // take values that do not correspond to its declared enumerators.
+// TODO: Works around a clang-format bug:
+// https://github.com/llvm/llvm-project/issues/85476
+#define CARBON_OPEN_ENUM [[clang::enum_extensibility(open)]]
+  enum class CARBON_OPEN_ENUM RawEnumType : uint8_t {
+      // The first sizeof...(Types) values correspond to the types.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
    // The first sizeof...(Types) values correspond to the types.
```

toolchain/sem_ir/id_kind.h
```diff
@@ -21,16 +21,22 @@
...
+
+      // An explicitly invalid value.
+      Invalid = NumTypes,
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
    // An explicitly invalid value.
    Invalid = NumTypes,
```

toolchain/sem_ir/id_kind.h
```diff
@@ -21,16 +21,22 @@
...
+
+      // Indicates that no type should be used.
+      // TODO: This doesn't really fit the model of this type, but it's
+      // convenient for all of its users.
+      None,
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
    // Indicates that no type should be used.
    // TODO: This doesn't really fit the model of this type, but it's
    // convenient for all of its users.
    None,
```

toolchain/check/subst.cpp
zygoloid: Done.

toolchain/check/subst.cpp
zygoloid: Done.

toolchain/check/subst.cpp
zygoloid: I thought about putting them on `Worklist`, but I like the separation between the worklist and these operations. I've not really thought about adding a class to hold the context and the worklist, and having now considered it I think it's probably not worthwhile -- passing them both around seems like less work than maintaining a separate declaration and definition of all the functions -- but I find this pretty borderline and would be happy to change it if you feel more strongly.

PR #3789, opened 2024-03-15
Start pulling in names from cross-package imports.

jonmeow:
Note, I'm annotating the lookup partly so that the reason the conflict comes up is clear, partly so that there's actually a diagnostic line associated with the root cause as more tests get packed into a single file.

### DIFF
--- a/toolchain/base/value_store.h
+++ b/toolchain/base/value_store.h
@@ -200,6 +200,14 @@ class ValueStore<StringId> : public Yaml::Printable<ValueStore<StringId>> {
     return values_[id.index];
   }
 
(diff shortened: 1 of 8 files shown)


zygoloid (approved):
I think it'd be a good idea to cache failed lookups, because we might end up performing them over and over again, and it'd be unfortunate to look for the name in all imports each time. But that definitely doesn't need to be done in this PR, so I'm merging.

jonmeow:
Yeah, I'd meant to leave to TODO for that. My thought was towards a unified story with name lookup failures and the impact on later declarations; I thought it best to implement those together. I also wasn't yet sure that something like storing Invalid would yield the right result here.

PR #3790, opened 2024-03-16
Parse and check support for compound member access.

zygoloid:
On the parsing side, we treat `a.(b)` as a member access whose second operand is a `ParenExpr` rather than a `MemberName`. A new node category is added for the union of `MemberName` and `ParenExpr` to support this.

Checking is mostly reusing the same pieces we already have for simple member access. Compound member access is in most ways a simplified form of simple member access because it doesn't need to do any lookup.


toolchain/check/member_access.cpp
```diff
@@ -366,4 +370,42 @@ auto PerformMemberAccess(Context& context, Parse::MemberAccessExprId node_id,
...
+          member.type_id())) {
+    member_id =
+        PerformImplLookup(context, base_type_const_id, *assoc_type, member_id);
+  }
+
+  // Perform instance binding if we found an instance member.
+  member_id = PerformInstanceBinding(context, node_id, base_id, member_id);
+
+  // If we didn't perform impl lookup or instance binding, that's an error
+  // because the base expression is not used for anything.
```
chandlerc, line 398: Just flagging that there is a comment from me in the bindings proposal that suggests this might not always be an error.

Probably not worth changing anything at this stage, but flagging in review just in case you want to add a TODO or anything.

chandlerc (approved):
Niiiice! LG and will merge. Left a comment inline just as an FYI.

PR #3791, opened 2024-03-18
file_name -> filename

jonmeow:
I wanted to choose one or the other. I think some code has been using each from early on. We're predominately using `filename`, so consolidating on that. This conveniently matches the [Google dev doc style guide](https://developers.google.com/style/word-list#filename) (which we use for docs) which says "filename: Not file name".

In toolchain:

```
╚╡git grep file_name . | wc -l
35
╚╡git grep filename . | wc -l
489
```


chandlerc (approved):
LG

PR #3792, opened 2024-03-18
Associate cross-file errors with the file being compiled.

jonmeow:
Import-related compilation failures can be associated with lines in other files. As we move to have multiple tests in a given file, this can be particularly confusing because errors can end up associated with the wrong file.

This adds a message in front of the compilation error to indicate what was being compiled, before getting into the error. Putting it first allows the test infrastructure to file-associate the full error message. This is not the same as a Note diagnostic because it's at the start of the diagnostic, rather than at the end.

I'm associating the filename with each emitter because it felt a little more appropriate than having it as an api on the location translator. I think having Emit/Build take an additional Location for the source would be a fair alternative, since we do need this information on the start of a message, although that would require changing each message. Setting up an annotator-style approach for this doesn't feel quite right to me (i.e., setting the location on each parse node, before calling out to the handler).

Format-wise I considered moving the ERROR prefix but thought that might not be considered the "right" thing to flag for a human trying to read these diagnostics. I could hide this note with a flag, keeping it primarily for tests, but thought we might want something like this for developers. That said, if we want a note at the end of the diagnostic for developers (versus in addition to), that would get me back to the flag approach because a note wouldn't quite offer the right context to achieve what's done here.

### DIFF
--- a/testing/file_test/autoupdate.cpp
+++ b/testing/file_test/autoupdate.cpp
@@ -15,8 +15,12 @@
 
 namespace Carbon::Testing {
 
-// Converts a matched line number to an int, trimming whitespace.
+// Converts a matched line number to an int, trimming whitespace. Returns -1 if
+// no line number was matched.
 static auto ParseLineNumber(absl::string_view matched_line_number) -> int {
+  if (matched_line_number.empty()) {
+    return -1;
+  }
   llvm::StringRef trimmed = matched_line_number;
   trimmed = trimmed.trim();
   // NOLINTNEXTLINE(google-runtime-int): API requirement.
@@ -41,7 +45,6 @@ auto FileTestAutoupdater::CheckLine::RemapLineNumbers(
     return;
   }
 
-  bool found_one = false;
   // Use a cursor for the line so that we can't keep matching the same
   // content, which may occur when we keep a literal line number.
   int line_offset = 0;
@@ -60,10 +63,8 @@ auto FileTestAutoupdater::CheckLine::RemapLineNumbers(
       RE2::PartialMatch(line_cursor, *replacement_->re, &matched_line_number);
     }
     if (matched_line_number.empty()) {
-      CARBON_CHECK(found_one) << line_;
       return;
     }
-    found_one = true;
 
     // Update the cursor offset from the match.
     line_offset = matched_line_number.begin() - line_.c_str();
--- a/testing/file_test/file_test_base.cpp
+++ b/testing/file_test/file_test_base.cpp
@@ -240,7 +240,7 @@ auto FileTestBase::GetLineNumberReplacements(
     -> llvm::SmallVector<LineNumberReplacement> {
   return {{.has_file = true,
            .re = std::make_shared<RE2>(
-               llvm::formatv(R"(({0}):(\d+))", llvm::join(filenames, "|"))),
+               llvm::formatv(R"(({0}):(\d+)?)", llvm::join(filenames, "|"))),
            .line_formatv = R"({0})"}};
 }
 
--- a/testing/file_test/file_test_base_test.cpp
+++ b/testing/file_test/file_test_base_test.cpp
@@ -131,6 +131,15 @@ class FileTestBaseTest : public FileTestBase {
       result.per_file_success.push_back({"a.carbon", true});
       result.per_file_success.push_back({"fail_b.carbon", false});
       return result;
+    } else if (filename == "fail_no_line.carbon") {
+      stderr << "In fail_no_line.carbon:\nan error\n";
+      return {{.success = false}};
+    } else if (filename == "fail_multi_no_line.carbon") {
+      stderr << "In b.carbon:\n"
+             << "an error\n"
+             << "In c.carbon:\n"
+             << "c.carbon:5: an error\n";
+      return {{.success = false}};
     } else {
       return ErrorBuilder() << "Unexpected file: " << filename;
     }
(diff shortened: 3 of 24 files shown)


chandlerc:
I wonder what you or others think about a subtly different approach here -- specifically, insisting that the initial error location is in the compiling file?

Specifically, I'm wondering what cases we have where the error wouldn't be fully detected by compiling the API file in question, and I think we want those to be (very) rare. Currently, I think we have:

1) cross-library name conflicts
2) template instantiation failures

Are there others? The example I saw in this PR is focused on (1) I think, but I just did a quick skim and so might have missed some.

If so, I could imagine (1) being phrased as an error about *importing* an API file with conflicting declarations, and then noting both conflicts from the two importees. I think that structure to the error message would actually be a bit more clear for me getting the message.

I think (2) is a trickier case, but I think even there it might make sense to diagnose these in a different way from how Clang does by first giving an error of "Unable to instantiate template `Foo` here as it doesn't type check with [parameterization]", and give the location that triggers the error in the currently compiling file. And then attach diagnostics from the instantiation itself. But very unsure of this -- wonder what @zygoloid thinks here?

jonmeow:
> I wonder what you or others think about a subtly different approach here -- specifically, insisting that the initial error location is in the compiling file?

Can you please be specific about how what you're asking differs from the approach? That is, can you please provide a specific example of what you're proposing, e.g. versus a changed error message:

```
// CHECK:STDERR: While compiling fail_main_use_other_extern.carbon:
// CHECK:STDERR: other_fn_extern.carbon:5:1: ERROR: Duplicate name being declared in the same scope.
// CHECK:STDERR: fn F();
// CHECK:STDERR: ^~~~~~~
// CHECK:STDERR: other_fn.carbon:4:1: Name is previously declared here.
// CHECK:STDERR: fn F() {}
// CHECK:STDERR: ^~~~~~~~
```

Are you thinking about something like:

```
// CHECK:STDERR: ERROR: fail_main_use_other_extern.carbon:4: Error in import.
// CHECK:STDERR: import Other library "fn_extern";
// CHECK:STDERR: ^~~~~~
// CHECK:STDERR: other_fn_extern.carbon:5:1:  Duplicate name being declared in the same scope.
// CHECK:STDERR: fn F();
// CHECK:STDERR: ^~~~~~~
// CHECK:STDERR: other_fn.carbon:4:1: Name is previously declared here.
// CHECK:STDERR: fn F() {}
// CHECK:STDERR: ^~~~~~~~
```

I'd kind of expected we might want "imported here" after each of those, but *after* versus *before* is extremely important for the related test infrastructure, which very much wants it *before*. i.e., I'd thought a human might prefer:

```
// CHECK:STDERR: While compiling fail_main_use_other_extern.carbon:
// CHECK:STDERR: ERROR: other_fn_extern.carbon:5:1:  Duplicate name being declared in the same scope.
// CHECK:STDERR: fn F();
// CHECK:STDERR: ^~~~~~~
// CHECK:STDERR: fail_main_use_other_extern.carbon:4: Imported here.
// CHECK:STDERR: import Other library "fn_extern";
// CHECK:STDERR: ^~~~~~
// CHECK:STDERR: other_fn.carbon:4:1: Name is previously declared here.
// CHECK:STDERR: fn F() {}
// CHECK:STDERR: ^~~~~~~~
// CHECK:STDERR: fail_main_use_other_extern.carbon:3: Imported here.
// CHECK:STDERR: import Other library "fn";
// CHECK:STDERR: ^~~~~~
```

And *maybe* once we have that we stop printing "While compiling fail_main_use_other_extern.carbon:", maybe not.

It's also worth considering how this also affects an example message that currently doesn't note the "While compiling" bit, if it does.

> Specifically, I'm wondering what cases we have where the error wouldn't be fully detected by compiling the API file in question, and I think we want those to be (very) rare. Currently, I think we have:
> 
> 1. cross-library name conflicts
> 2. template instantiation failures
> 
> Are there others? The example I saw in this PR is focused on (1) I think, but I just did a quick skim and so might have missed some.

Intra-library, there can be declarations missing definitions, and duplicate definitions. i.e., ODR-related issues.

Note that cross-library name conflicts might be related to a broader category of any differences, including modifier keyword differences.

I could also see some generic-related messages cropping up due to incompatible interface things in subtle ways.

> If so, I could imagine (1) being phrased as an error about _importing_ an API file with conflicting declarations, and then noting both conflicts from the two importees. I think that structure to the error message would actually be a bit more clear for me getting the message.

Note, another choice here would be to just change the message; instead of "While compiling", I could twiddle it to "ERROR: foo.carbon: [Warning|Error] in import."

> I think (2) is a trickier case, but I think even there it might make sense to diagnose these in a different way from how Clang does by first giving an error of "Unable to instantiate template `Foo` here as it doesn't type check with [parameterization]", and give the location that triggers the error in the currently compiling file. And then attach diagnostics from the instantiation itself. But very unsure of this -- wonder what @zygoloid thinks here?

jonmeow:
> And _maybe_ once we have that we stop printing "While compiling fail_main_use_other_extern.carbon:", maybe not.

To clarify -- maybe we stop printing for end users, and move it to a test-only flag.

jonmeow:
Per discussion, will aim for:

```
// CHECK:STDERR: fail_main_use_other_extern.carbon:4: In import.
// CHECK:STDERR: import Other library "fn_extern";
// CHECK:STDERR: ^~~~~~
// CHECK:STDERR: other_fn_extern.carbon:5:1: ERROR: Duplicate name being declared in the same scope.
// CHECK:STDERR: fn F();
// CHECK:STDERR: ^~~~~~~
// CHECK:STDERR: fail_main_use_other_extern.carbon:3: In import.
// CHECK:STDERR: import Other library "fn";
// CHECK:STDERR: ^~~~~~
// CHECK:STDERR: other_fn.carbon:4:1: Name is previously declared here.
// CHECK:STDERR: fn F() {}
// CHECK:STDERR: ^~~~~~~~
```

with a blank line between (after?) all diagnostics.

jonmeow:
Closing this PR as I think the changes requested are going to be mostly unrelated.

PR #3793, opened 2024-03-18
Update the file_test readme with API and fail_ notes.

jonmeow:
Also a minor example update.

### DIFF
--- a/testing/file_test/README.md
+++ b/testing/file_test/README.md
@@ -33,7 +33,6 @@ A typical implementation will look like:
 ```
 #include "my_library.h"
 
-#include "llvm/ADT/StringExtras.h"
 #include "testing/file_test/file_test_base.h"
 
 namespace Carbon::Testing {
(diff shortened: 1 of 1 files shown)


testing/file_test/README.md
```diff
@@ -65,6 +64,17 @@ CARBON_FILE_TEST_FACTORY(MyFileTest);
...
+## Filename `fail_` prefixes
+
+When a run fails, information about what pieces failed are returned on
+`RunResult`. This affects whether a `fail_` prefix on the file is required,
+including in combination with split-file tests (using the `// --- <filename>`
+comment marker).
+
+The main test file and any split-files must have a `fail_` prefix if and only if
+they have an associated error. An exception is that the main test file may omit
+`fail_` when it contains split-files that have a `fail_` prefix.
```
josh11b, line 76: Are you saying:
* If you don't use split files, then `fail_` prefix on main test file if and only if it has an error.
* If you do use split files, then `fail_` prefix on each split that has a failure. If there is any failure then the main test file may optionally also start with `fail_`.

?

jonmeow: > Are you saying:
> 
> * If you don't use split files, then `fail_` prefix on main test file if and only if it has an error.
> * If you do use split files, then `fail_` prefix on each split that has a failure. If there is any failure then the main test file may optionally also start with `fail_`.
>   ?

Yes, that's a different way to phrase it. I prefer this phrasing because it makes behavior sound less divergent.

jonmeow: Actually, there's a nuance your phrasing is missing. There can be no per-file failures, and the main test file may still fail. Then `fail_` is required.

josh11b (approved):

PR #3795, opened 2024-03-18
Start filling in `extern` support on functions.

jonmeow:
Still needs more merge/redeclaration logic for import semantics, but this felt like a reasonable point to send a PR.


toolchain/check/handle_function.cpp
```diff
@@ -39,13 +40,25 @@ auto HandleReturnType(Context& context, Parse::ReturnTypeId node_id) -> bool {
...
   CheckAccessModifiersOnDecl(context, decl_kind, target_scope_id);
+  if (is_definition) {
+    ForbidExternModifierOnDefinition(context, decl_kind);
+  }
+  if (target_scope_id.is_valid()) {
+    auto target_id = context.name_scopes().Get(target_scope_id).inst_id;
+    if (target_id.is_valid() &&
+        !context.insts().Is<SemIR::Namespace>(target_id)) {
+      ForbidModifiersOnDecl(context, KeywordModifierSet::Extern, decl_kind,
+                            " on member functions");
```
chandlerc: I don't really have a better idea for how to phrase this, but I think this includes rejecting `extern` nested functions, correct?

Anyways, not really worth changing now, mostly wanted to make sure I understood it correctly.

toolchain/check/handle_function.cpp
```diff
@@ -96,16 +109,16 @@ static auto BuildFunctionDecl(Context& context,
...
                      ModifierOrder::Access),
                  "access modifier");
   }
-  if (!!(modifiers & KeywordModifierSet::Extern)) {
-    context.TODO(context.decl_state_stack().innermost().modifier_node_id(
-                     ModifierOrder::Extern),
-                 "extern modifier");
+  bool is_extern = !!(modifiers & KeywordModifierSet::Extern);
+  if (is_extern && is_definition) {
+    is_extern = false;
```
chandlerc: It would seem nice if `DiagnoseModifiers` pruned modifiers that it diagnosed for you. Then you could simplify this.

Maybe a TODO? Doesn't seem worth blocking on, just maybe a nice refactoring to queue up.

toolchain/check/testdata/function/declaration/extern.carbon
```diff
@@ -0,0 +1,118 @@
...
+// CHECK:STDERR: ^~~~~~~
+// CHECK:STDERR: fail_redecl_extern.carbon:[[@LINE-4]]:1: Previously declared here.
+// CHECK:STDERR: extern fn F();
+// CHECK:STDERR: ^~~~~~~~~~~~~~
+fn F();
+
+// --- fail_member_extern.carbon
+
+class C {
+  // CHECK:STDERR: fail_member_extern.carbon:[[@LINE+3]]:3: ERROR: `extern` not allowed on `fn` declaration on member functions.
```
chandlerc: The wording of the diagnostic with the context string seems a bit awkward here.

I feel like we really want a fairly customized description of the entity with the invalid modifier:

- `fn` declaration
- member `fn` declaration
- `fn` definition (see below)

Not sure whether its best to have dedicated APIs for each of these entities or if there is a good way to let the caller customize the string here given that I think `fn` is coming from the token itself.

That said, as it will expand changes in the modifiers, happy if you'd rather leave some TODOs and fix this in a future iteration.

toolchain/check/testdata/function/declaration/fail_modifiers.carbon
```diff
@@ -70,7 +70,7 @@ base fn InvalidModifier();
 // CHECK:STDERR: ^~~~~~~
 default final virtual fn ModifiersConflict2() {}
 
-// CHECK:STDERR: fail_modifiers.carbon:[[@LINE+9]]:1: ERROR: `extern` not allowed on `fn` declaration.
+// CHECK:STDERR: fail_modifiers.carbon:[[@LINE+9]]:1: ERROR: `extern` not allowed on `fn` declaration on a definition.
```
chandlerc: Similar but a bit worse I think here is the context string not working well. Keeping discussion focused on the above comment.

toolchain/check/function.cpp
```diff
@@ -224,6 +225,17 @@ auto MergeFunctionRedecl(Context& context, Parse::NodeId node_id,
         .Emit();
     // The second definition will be unused as a consequence of the error.
     return true;
+  } else if (prev_function.is_extern) {
+    CARBON_DIAGNOSTIC(FunctionDefiningExtern, Error,
+                      "Cannot define extern function `{0}`.", SemIR::NameId);
```
chandlerc:
```suggestion
                      "Cannot define `extern` function `{0}`.", SemIR::NameId);
```

chandlerc (approved):
Mostly a question on the diagnostics we end up with here.

Happy for this to go in with some TODOs (and the suggested edit), so approving.

If implementing more configurability in the diagnostics in this PR, happy to take another look.

toolchain/check/handle_function.cpp
jonmeow: Do you mean function declarations in functions? Those aren't supported right now, they'd be rejected at function body processing.

toolchain/check/testdata/function/declaration/fail_modifiers.carbon
jonmeow: Changes to "that provides a definition".

toolchain/check/testdata/function/declaration/extern.carbon
jonmeow: Changed phrasing to "declaration that is a member"

It's not clear to me what you mean. Are you suggesting that the API accept an arbitrary string instead of a token to describe the entity? Note, one thought towards taking tokens and similar in the APIs was trying to keep diagnostic messages programmatically accessible; I've been wary of strings because it feels like it pushes tool developers towards string parsing semantics on diagnostic messages. But I'm too concerned here, we effectively do this for types already, and we should just delve into more strings?

I've added a TODO, regardless -- hopefully you're okay with that, although I'd just say all the diagnostics could use polish, but most don't have specific TODOs.

toolchain/check/handle_function.cpp
jonmeow: Removed the && check. It prunes, I'd just shifted the validation without removing removing this bit.

Issue #2113, opened 2022-08-25
Structure, scope, and naming of the prelude and syntax aliases


chandlerc:
Short version: we should pick a scheme for naming things in the prelude of Carbon, and to a limited extent in the standard library more generally as they overlap. I'm suggesting the rough following scheme:

- The prelude works *exactly* as an import of one (or more) packages from the standard library.
  - The prelude itself doesn't introduce aliases or other functionality.
  - Everything imported by the prelude works just like a normally imported Carbon package, and the package name must be used to reference it.
- The prelude should at *least* cover the standard library types that are directly referenced, used, or connected to language constructs.
  - Developers shouldn't need to import something to use language syntax, and language syntax shouldn't require the power to reference un-imported names.
- We provide keywords and literal syntaxes in the _language_ that give brief syntax for the sufficiently common entities (regardless of what package-qualified name they otherwise need).
  - This builds on #543 which suggests `bool`, `i32`, and so on.
  - For example, a keyword `foo` might refer to `SomePackageName.Foo`.

This raises two more questions: what package names should the Carbon standard library use? (Or at least, the parts of the standard library we need to consider to establish what is in or not-in the prelude?) And how much of the standard library should be imported by the prelude?

For the first, after a bunch of thought, I'm being persuaded that Carbon's standard library shouldn't try to live in a single package, and the package names should -- to the extent possible -- be tied to the purpose or utility provided by that part of the standard library.

For example (but not suggested in this issue):
- `Math` package for things like the definition of π.
- `Containers` package maybe for ... well ... containers
- ...

I'd like to suggest that to start with, the only part of the standard library we put in the prelude are the parts of the standard library that are directly connected to the language itself: things like `bool` as that's the return type of `and` operators, all other types with dedicated syntax (`i32`, etc), and interfaces used by language syntax like `CommonType` and `Add`.

Note that these wouldn't get any preferential syntax by being in the prelude with this suggestion, *just* be available without any explicit import.

My suggestion is that we have a single package that is only the constructs that directly interact with the syntax in some way as part of the prelude to start. We can expand this in the future if desired, but I think its a safe place to start. I think its reasonable for all of those constructs to be in one package, and the fact that they're part of the language itself helps suggest how to name that package. Some options seem especially appealing to me:

- a) `Core`, as in interacting with the core language.
- b) `Base`, as in the base of the library dependency graph.
- c) `Language`, due to being part of the language.
- d) `Prelude`, as it is the contents of the prelude.

[...]

PR #3796, opened 2024-03-19
Overloaded operator support.

zygoloid:
Support is added for all overloaded operator interfaces in the current design apart from `Assign`, which is going to require some more work to properly handle, given that primitive assignment currently has a special implementation for quite a few builtin types.

As we don't have support for generics yet -- in particular, generic interfaces -- there is no support for `*With` interfaces, but homogenous interfaces such as `Add` are supported instead.

Factor out building of call expressions so that overloaded operators can generate calls.

Switch a few places from using specific kinds of NodeId to a general NodeId. Because overloaded operators and other things like implicit conversions can result in member access and function calls, those operations can't require a specific kind of NodeId.

Add import support for associated entities, and fix import support for interfaces and symbolic bindings. We now import interfaces in two steps, first importing a forward declaration then a definition, just like we do for classes. For symbolic bindings, we ensure that each BindSymbolicName is imported only once, because its ID is used as its symbolic identity. This is necessary because we (only) support operator interfaces that are defined in an imported Carbon package for now.

The entire contents of `check/operator.cpp` should probably be rethought. In particular, doing a lot of name lookups on each operator is likely to be bad for performance. But this gets us to the point where overloaded operators are basically working, which seems like a good place to iterate from.

For now, the tests that the individual operators map to the right interfaces are mostly generated by a script, but that's just because I'm expecting a fair bit of churn in how we define the prelude and the `impl`s -- in particular, when we add support for `AddWith`, we'll need to update all the tests. The plan is to remove the script once things settle down.


.pre-commit-config.yaml
```diff
@@ -180,7 +180,7 @@ repos:
             Exceptions. See /LICENSE for license information.
             SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
           - --custom_format
-          - '\.(carbon|proto|ypp)$'
+          - '\.(carbon|proto|ypp)(\.in)?$'
```
jonmeow: What's a ".in" file? Should this be covered by the PR description?

It looks like you're using it for a script, but it's still Carbon code. We use .def a bit for that in C++; would .def be appropriate here too? Maybe it should be .in.carbon instead of .carbon.in? Perhaps .template or .tmpl instead of .in, to be precise about the use-case?

toolchain/check/testdata/operators/overloaded/negate.carbon
```diff
@@ -0,0 +1,242 @@
...
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+
+// This file was generated from unary_op.carbon.in. Run make_tests.sh to regenerate.
+
+// --- prelude.carbon
+
+package Carbon api;
```
jonmeow: Why "Carbon"? If you're trying to create a package name conflict, should this be "Core"? If not, maybe something like "FakePrelude"?

toolchain/sem_ir/typed_insts.h
```diff
@@ -273,8 +273,7 @@ struct BoolLiteral {
 // `self` parameter, such as `object.MethodName`.
 struct BoundMethod {
   static constexpr auto Kind =
-      InstKind::BoundMethod.Define<Parse::AnyMemberAccessExprId>(
```
jonmeow, line 276: Are you removing all uses of AnyMemberAccessExprId? Should the definition be deleted?

toolchain/check/testdata/operators/overloaded/make_tests.sh
```diff
@@ -0,0 +1,38 @@
+#! /bin/bash
+# Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+# Exceptions. See /LICENSE for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+make_test() {
+  HEADER="// This file was generated from $4. Run make_tests.sh to regenerate."
+  sed "s,INTERFACE,$1,g; s,OP,$2,g; s,HEADER,$HEADER," $4 > $3.carbon
+}
```
jonmeow: I have concerns about this script; putting a script in testdata feels like it hides it from developers (I would only expect data files here, not code). Additionally, it adds another manual execution step separate from autoupdate. Shell scripts are additionally a barrier if we try to support Windows -- I would suggest a preference for Python, as it has better cross-platform compatibility.

What alternatives had you considered? A few I might suggest if you haven't considered them already:

- Use a genrule to generate boilerplate files.
  - file_test should be okay taking any BUILD output as input, so I think this would mostly work.
  - You'd probably and to specify ARGS to exclude the IR (i.e., to avoid update issues, making this just a test that it compiles)
  - genrules can still be written to be platform-specific, so please be careful about this -- I think we currently only have genrules in explorer code.
- Write a traditional unit test that generates content and runs the driver on it directly.
  - This feels a little related to the genrule, although shifting away from file_test a little more. End results are probably the same.
- Add a utility in file_test for templating.
  - A way to still have autoupdate support.

toolchain/check/operator.h
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Carbon::Check {
+
+struct Operator {
+  llvm::StringLiteral interface_name;
+  llvm::StringLiteral op_name = "Op";
+};
+
+// Checks and builds SemIR for a unary operator expression, `$operand` or
+// `operand$`.
```
jonmeow:
```suggestion
// Checks and builds SemIR for a unary operator expression. For example, `$operand` or
// `operand$`.
```

I think clearly indicating this is an example would help avoid confusion. You might also consider replacing `$` with an operator Carbon uses, but maybe my mind is too much on the shell script.

toolchain/check/operator.h
```diff
@@ -0,0 +1,31 @@
...
+auto BuildUnaryOperator(Context& context, Parse::AnyExprId node_id, Operator op,
+                        SemIR::InstId operand_id) -> SemIR::InstId;
+
+// Checks and builds SemIR for a binary operator expression, `lhs_id $ rhs_id`.
```
jonmeow:
```suggestion
// Checks and builds SemIR for a binary operator expression. For example, `lhs_id $ rhs_id`.
```

toolchain/check/handle_operator.cpp
```diff
@@ -4,20 +4,46 @@
 
 #include "toolchain/check/context.h"
 #include "toolchain/check/convert.h"
+#include "toolchain/check/operator.h"
 #include "toolchain/check/pointer_dereference.h"
 #include "toolchain/diagnostics/diagnostic_emitter.h"
 
 namespace Carbon::Check {
 
+auto HandleUnaryOperator(Context& context, Parse::NodeId node_id, Operator op)
```
jonmeow:
```suggestion
// Common logic for unary operator handlers.
static auto HandleUnaryOperator(Context& context, Parse::NodeId node_id, Operator op)
```

The `static` is the more important part of my suggestion.

toolchain/check/handle_operator.cpp
```diff
@@ -4,20 +4,46 @@
...
+  // TODO: Support implicit conversion from specific node IDs to node ID
+  // categories and change this function to take an `AnyExprId` directly.
+  auto expr_node_id = static_cast<Parse::AnyExprId>(node_id);
+  auto operand_id = context.node_stack().PopExpr();
+  auto result_id = BuildUnaryOperator(context, expr_node_id, op, operand_id);
+  context.node_stack().Push(expr_node_id, result_id);
+  return true;
+}
+
+auto HandleBinaryOperator(Context& context, Parse::NodeId node_id, Operator op)
```
jonmeow:
```suggestion
// Common logic for binary operator handlers.
static auto HandleBinaryOperator(Context& context, Parse::NodeId node_id, Operator op)
```

toolchain/check/operator.cpp
```diff
@@ -0,0 +1,145 @@
...
+  auto name_id = SemIR::NameId::ForIdentifier(ident_id);
+
+  // Look up `package.Carbon`.
+  auto package_id = context.LookupQualifiedName(
+      node_id, name_id, SemIR::NameScopeId::Package, /*required=*/false);
+  if (!package_id.is_valid()) {
+    return SemIR::NameScopeId::Invalid;
+  }
+
+  package_id = context.constant_values().Get(package_id).inst_id();
```
jonmeow, line 38: Why does this re-resolve through the constant? (a comment might help)

toolchain/check/operator.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+#include "toolchain/check/call.h"
+#include "toolchain/check/context.h"
+#include "toolchain/check/member_access.h"
+#include "toolchain/sem_ir/ids.h"
+#include "toolchain/sem_ir/typed_insts.h"
+
+namespace Carbon::Check {
+
+// Returns the scope of the Carbon package, or Invalid if it's not found.
```
jonmeow: Why "Carbon" versus "Core"? Wasn't "Core" the outcome of https://github.com/carbon-language/carbon-lang/issues/2113#issuecomment-1301589247?

toolchain/check/operator.cpp
```diff
@@ -0,0 +1,145 @@
...
+  // Lookup `Interface.Op`.
+  auto op_ident_id = context.identifiers().Add(op.op_name);
+  auto op_id = context.LookupQualifiedName(
+      node_id, SemIR::NameId::ForIdentifier(op_ident_id), interface_scope_id,
+      /*required=*/false);
+  if (!op_id.is_valid()) {
+    return SemIR::InstId::Invalid;
+  }
+
+  op_id = context.constant_values().Get(op_id).inst_id();
```
jonmeow, line 100: Similar to package_id comment, this feels odd.

jonmeow (commented):
Awesome!

Note regarding name lookup, I've had similar concerns -- I think we need some way of caching operator lookup results. But I'm not sure what the shape of that is. (do we store this on types?)

toolchain/check/operator.cpp
zygoloid: Ah right. Switched to Core throughout.

We should have a proposal for that, and update the documentation to match :-)

toolchain/check/testdata/operators/overloaded/negate.carbon
zygoloid: Updated to `Core` to match general switch from `Carbon` to `Core`.

toolchain/check/handle_operator.cpp
```diff
@@ -4,20 +4,48 @@
...
 #include "toolchain/check/convert.h"
+#include "toolchain/check/operator.h"
 #include "toolchain/check/pointer_dereference.h"
 #include "toolchain/diagnostics/diagnostic_emitter.h"
 
 namespace Carbon::Check {
 
+// Common logic for unary operator handlers.
+static auto HandleUnaryOperator(Context& context, Parse::NodeId node_id, Operator op)
+    -> bool {
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
static auto HandleUnaryOperator(Context& context, Parse::NodeId node_id,
                                Operator op) -> bool {
```

toolchain/check/handle_operator.cpp
```diff
@@ -4,20 +4,48 @@
...
+  auto expr_node_id = static_cast<Parse::AnyExprId>(node_id);
+  auto operand_id = context.node_stack().PopExpr();
+  auto result_id = BuildUnaryOperator(context, expr_node_id, op, operand_id);
+  context.node_stack().Push(expr_node_id, result_id);
+  return true;
+}
+
+// Common logic for binary operator handlers.
+static auto HandleBinaryOperator(Context& context, Parse::NodeId node_id, Operator op)
+    -> bool {
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
static auto HandleBinaryOperator(Context& context, Parse::NodeId node_id,
                                 Operator op) -> bool {
```

toolchain/check/operator.h
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Carbon::Check {
+
+struct Operator {
+  llvm::StringLiteral interface_name;
+  llvm::StringLiteral op_name = "Op";
+};
+
+// Checks and builds SemIR for a unary operator expression. For example, `$operand` or
+// `operand$`.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
// Checks and builds SemIR for a unary operator expression. For example,
// `$operand` or `operand$`.
```

toolchain/check/operator.h
```diff
@@ -0,0 +1,31 @@
...
+auto BuildUnaryOperator(Context& context, Parse::AnyExprId node_id, Operator op,
+                        SemIR::InstId operand_id) -> SemIR::InstId;
+
+// Checks and builds SemIR for a binary operator expression. For example, `lhs_id $ rhs_id`.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
// Checks and builds SemIR for a binary operator expression. For example,
// `lhs_id $ rhs_id`.
```

toolchain/check/operator.cpp
zygoloid: Comment added.

toolchain/check/operator.cpp
zygoloid: Comment added.

toolchain/sem_ir/typed_insts.h
zygoloid: Done.

jonmeow (approved):

toolchain/check/testdata/operators/overloaded/make_tests.sh
jonmeow: Just to note, discussed this offline: agreed to a TODO to remove the script, with the expectation that the [actual] prelude would lead to these being pretty small and not worth autoupdate anymore.

[avoiding various template approaches to avoid abstraction of test files and golden output]

zygoloid: Per discussion, added a TODO to say the script is short-term and should be removed once things stabilize here.

.pre-commit-config.yaml
zygoloid: Switched to `.tmpl`. (`.def` suggests C preprocessor input. `.in.carbon` would get picked up by `file_test`.)

PR #3798, opened 2024-03-20
Lowering support for overloaded operators.

zygoloid:
Fix a collection of issues that were preventing lowering for overloaded operators from working.

Instead of creating `import_ref` instructions during name lookup in the current block, whatever that might be, we now create them in the `file` block always. This avoids inserting them into blocks that might not be intended to contain them, such as functions, and avoids the IR generated for a function depending on which names we happen to have looked up first.

When importing a class, function, or interface, import its enclosing scope ID. This is necessary to allow us to distinguish between functions at interface scope, which shouldn't be lowered, and other functions, and will also be used in future to provide qualified names for declarations when printing types. In order to support this:

- Track the constant values of namespaces created during importing so that we can find them when resolving an import ref. Use those constant values to convert an enclosing scope ID from the imported IR into a corresponding ID in the current IR.
- Change how we do two-pass import of classes and namespaces so that we can do two-pass import even for non-defining declarations, so that we can import the enclosing scope.

While working on the final point above, I reworked `TryResolveInst` to return a flag indicating whether another pass is necessary instead of implicitly encoding this in the `ConstantId`. This permits the handling of classes to be simplified; now `import_ir_constant_values` is only accessed in a single place.

### DIFF
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -162,7 +162,7 @@ static auto InitPackageScopeAndImports(Context& context, UnitInfo& unit_info)
   // First create the constant values map for all imported IRs. We'll populate
   // these with mappings for namespaces as we go.
   size_t num_irs = context.import_irs().size();
-  for (auto& [package_id, package_imports] : unit_info.package_imports_map) {
+  for (auto& [_, package_imports] : unit_info.package_imports_map) {
     num_irs += package_imports.imports.size();
   }
   context.import_ir_constant_values().resize(
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -159,6 +159,15 @@ struct UnitInfo {
 // Add imports to the root block.
 static auto InitPackageScopeAndImports(Context& context, UnitInfo& unit_info)
     -> void {
+  // First create the constant values map for all imported IRs. We'll populate
+  // these with mappings for namespaces as we go.
+  size_t num_irs = context.import_irs().size();
+  for (auto& [package_id, package_imports] : unit_info.package_imports_map) {
+    num_irs += package_imports.imports.size();
+  }
+  context.import_ir_constant_values().resize(
+      num_irs, SemIR::ConstantValueStore(SemIR::ConstantId::Invalid));
+
   // Importing makes many namespaces, so only canonicalize the type once.
   auto namespace_type_id =
       context.GetBuiltinType(SemIR::BuiltinKind::NamespaceType);
@@ -217,9 +226,8 @@ static auto InitPackageScopeAndImports(Context& context, UnitInfo& unit_info)
                                     sem_irs, package_imports.has_load_error);
   }
 
-  context.import_ir_constant_values().resize(
-      context.import_irs().size(),
-      SemIR::ConstantValueStore(SemIR::ConstantId::Invalid));
+  CARBON_CHECK(context.import_irs().size() == num_irs)
+      << "Created an unexpected number of IRs";
 }
 
 // Loops over all nodes in the tree. On some errors, this may return early,
--- a/toolchain/check/import_ref.cpp
+++ b/toolchain/check/import_ref.cpp
@@ -68,7 +68,6 @@ class ImportRefResolver {
   auto Resolve(SemIR::InstId inst_id) -> SemIR::ConstantId {
     work_stack_.push_back({inst_id});
     while (!work_stack_.empty()) {
-      auto initial_work = work_stack_.size();
       auto work = work_stack_.back();
       CARBON_CHECK(work.inst_id.is_valid());
 
@@ -84,6 +83,7 @@ class ImportRefResolver {
       if (existing_const_id.is_valid() && !work.retry) {
         work_stack_.pop_back();
       } else {
+        auto initial_work = work_stack_.size();
         auto [new_const_id, finished] =
             TryResolveInst(work.inst_id, existing_const_id);
         CARBON_CHECK(finished == !HasNewWork(initial_work));
--- a/toolchain/check/import_ref.cpp
+++ b/toolchain/check/import_ref.cpp
@@ -288,10 +288,7 @@ class ImportRefResolver {
   // unresolved constants to the work stack.
   auto GetLocalNameScopeId(SemIR::NameScopeId name_scope_id)
       -> SemIR::NameScopeId {
-    if (!name_scope_id.is_valid()) {
-      return SemIR::NameScopeId::Invalid;
-    }
-    auto inst_id = import_ir_.name_scopes().Get(name_scope_id).inst_id;
+    auto inst_id = import_ir_.name_scopes().GetInstIdIfValid(name_scope_id);
     if (!inst_id.is_valid()) {
       // Map scopes that aren't associated with an instruction to invalid
       // scopes. For now, such scopes aren't used, and we don't have a good way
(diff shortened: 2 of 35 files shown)


toolchain/check/import_ref.cpp
```diff
@@ -22,29 +22,30 @@ namespace Carbon::Check {
...
+//        declaration, pass it to Retry(). It will be passed to future attempts
+//        to resolve this instruction so the earlier work can be found, and will
+//        be made available for other instructions to use.
+//      - The second attempt to resolve this instruction must produce the same
+//        constant, because the value may have already been used by resolved
+//        instructions.
+//    - Build any necessary IR structures, and return the output constant.
+// 3. If resolve didn't return Retry(), pop the work. Otherwise, it needs to
+//    remain, and may no longer be at the top of the stack; set `retry` on it so
+//    we'll make sure to run it again later.
```
zygoloid, line 48: @jonmeow I had thought these changes were necessary for correctness, but now that I've finished debugging the issues I was seeing, I don't think they are. However, I do prefer where I ended up: returning a retry / done flag seems nicer to me than carrying that information partly in whether `Invalid` is returned and partly in whether `made_forward_decl` is set.

I'm happy to roll back the refactoring here if you don't like it so much, or split it out to a separate PR. Let me know what  you'd prefer.

toolchain/check/check.cpp
```diff
@@ -159,6 +159,15 @@ struct UnitInfo {
 // Add imports to the root block.
 static auto InitPackageScopeAndImports(Context& context, UnitInfo& unit_info)
     -> void {
+  // First create the constant values map for all imported IRs. We'll populate
+  // these with mappings for namespaces as we go.
+  size_t num_irs = context.import_irs().size();
+  for (auto& [package_id, package_imports] : unit_info.package_imports_map) {
```
```suggestion
  for (auto& [_, package_imports] : unit_info.package_imports_map) {
```

toolchain/check/import_ref.cpp
jonmeow: I still see code like `work_stack_.back().retry = true;`. This makes it look like information is still only partly in the return value, but there's more code now to understand. Can that be improved, so that the only interaction with work_stack_ is here?

toolchain/check/import_ref.cpp
```diff
@@ -258,13 +284,47 @@ class ImportRefResolver {
...
+    auto name_scope_inst = context_.insts().Get(const_id.inst_id());
+    if (auto namespace_inst = name_scope_inst.TryAs<SemIR::Namespace>()) {
+      return namespace_inst->name_scope_id;
+    }
+    if (auto class_inst = name_scope_inst.TryAs<SemIR::ClassType>()) {
+      return context_.classes().Get(class_inst->class_id).scope_id;
+    }
+    if (auto interface_inst = name_scope_inst.TryAs<SemIR::InterfaceType>()) {
+      return context_.interfaces().Get(interface_inst->interface_id).scope_id;
+    }
```
jonmeow: Should this be a switch on the kind rather than repeated cast attempts? I think it might be easier to read, but couldn't the repeated TryAs also be a performance issue, particularly if we add more scoped entities?

toolchain/check/import_ref.cpp
```diff
@@ -258,13 +284,47 @@ class ImportRefResolver {
...
+    if (const_id == SemIR::ConstantId::Error) {
+      return SemIR::NameScopeId::Invalid;
+    }
```
jonmeow: Why is this validation after const_id.inst_id() is used? Wouldn't using the result of insts().Get be an issue for an Error?

toolchain/check/import_ref.cpp
zygoloid: Thanks, I forgot to delete that line; it's not needed any more. Fixed.

toolchain/check/import_ref.cpp

While I actually find the result a little harder to read and more error-prone due to writing each type twice rather than once (so I marginally prefer the old form in isolation), I think I'd prefer a `switch` once we add a fourth case here, which we probably will :)

FWIW, I don't think performance is a factor here. After inlining, this'll just be repeated comparisons against the kind, which the optimizer should canonicalize into the same form as the `switch` anyway. (And at `-O0` a `switch` over a few cases gets lowered into a series of `if`s. Though we would avoid building the `optional`s at `-O0`.)

toolchain/check/import_ref.cpp
zygoloid: `ConstantId::Error` is just the constant form of `InstId::BuiltinError`, so it's a `Builtin` instruction, and `insts().Get` on it is fine. Given that this is a rare error case, I put it here so we don't need to check it on the common case.

toolchain/check/import_ref.cpp
```diff
@@ -67,21 +68,36 @@ class ImportRefResolver {
...
+            << "Constant value changed in second pass.";
         import_ir_constant_values_.Set(work.inst_id, new_const_id);
-        work_stack_.pop_back();
+        if (finished) {
+          CARBON_CHECK(!HasNewWork(initial_work))
+              << "TryResolveInst succeeded but added work for "
+              << import_ir_.insts().Get(work.inst_id);
+          work_stack_.pop_back();
+        } else {
+          work_stack_[initial_work - 1].retry = true;
```
jonmeow: Is this assuming that at most one constant was added to the stack?

jonmeow: Nevermind.

toolchain/check/import_ref.cpp
```diff
@@ -112,15 +128,25 @@ class ImportRefResolver {
...
+    bool retry = false;
   };
 
-  // For imported entities, we use an invalid enclosing scope. This will be okay
-  // if the scope isn't used later, but we may need to change logic for this if
-  // the behavior changes.
-  static constexpr SemIR::NameScopeId NoEnclosingScopeForImports =
-      SemIR::NameScopeId::Invalid;
+  // The result of attempting to resolve an imported instruction to a constant.
+  struct ResolveResult {
```
jonmeow, line 131: I wonder, have you considered:

- Use std::optional<ConstantId> instead of ResolveResult
- Keep the Retry function, but as  resolver member instead of here

Mainly wondering if that'd be a useful reduction of code.

jonmeow: Maybe that doesn't work due to the combinatorics aspect.

toolchain/check/import_ref.cpp
```diff
@@ -67,21 +68,36 @@ class ImportRefResolver {
   auto Resolve(SemIR::InstId inst_id) -> SemIR::ConstantId {
     work_stack_.push_back({inst_id});
     while (!work_stack_.empty()) {
+      auto initial_work = work_stack_.size();
```
jonmeow: If you're now tracking initial work here, and we're also tracking initial work in the helpers to determine whether retry is needed, is the finished indicator on ResolveResult needed?

Sorry, piecing this together as I think through approaches. Maybe std::optional<ConstantId> actually could work as I think through we ResolveResult couldn't be replaced without this.

toolchain/check/import_ref.cpp
jonmeow: FWIW, I kind of agree on the redundancy, but I find the switch a lot easier to see consistencies in -- the `if (...)` block I spend time trying to figure out the behavior. For example versus the `TryAs` code, I can more trivially see the consistent `return` behavior, and the cross-case flow is enforced by `switch` (and we do have things like the `Param` `TryAs` uses which do more of a fallthrough behavior, emphasizing the need to check). The `default` handling is similarly clearer for me.

If we did more macros, it'd kind of be tempting to write something that encapsulated:

```
#define CARBON_INST_CASE(InstName) \
case InstName::Kind: \
   auto typed_inst = inst.As<InstName>();
```

jonmeow: Hmmm, I think we could probably get:

```
TypedInstSwitch(context_.insts().Get(const_id.inst_id()))
    .Case<SemIR::Namespace>([](auto inst) { return inst.name_scope_id; })
    .Case<SemIR::ClassType>([](auto inst) { return context_.classes().Get(inst.class_id).scope_id; })
etc
```

[my mind has been on StringSwitch today]

Is this worthwhile, do you think?

jonmeow: Per discussion, I'm going to try this out.

toolchain/check/import_ref.cpp
zygoloid: I think that's possible, and a resolution step should be asking for a retry if and only if it didn't add any more work. However, I think it'd be fragile to rely on that, and would make the resolve code less clear. I've added a `CHECK` that each resolve step says it's finished if and only if it added no new work, and updated the comment on the `finished` flag to point out that it's technically not necessary.

jonmeow: Why is it fragile? I think this should be a strong requirement for the logic: if there's no new work added, we shouldn't enter a retry state.

jonmeow: Note, regarding clarity, initial_work should probably be declared closer to where it's used. [The flow here is a bit hard to read now, maybe it could have reduced nesting with a `continue`]

toolchain/check/import_ref.cpp
```diff
@@ -67,21 +68,36 @@ class ImportRefResolver {
...
       // Double-check that the constant still doesn't have a calculated value.
       // This should typically be checked before adding it, but a given
       // instruction may be added multiple times before its constant is
       // evaluated.
-      if (!work.made_forward_decl &&
-          import_ir_constant_values_.Get(work.inst_id).is_valid()) {
+      //
+      // If a retry was requested, then rerun this work item even if it already
+      // has a constant value from the previous run. It may not have been fully
+      // imported, even though it was imported enough to produce a constant.
```
jonmeow: Suggestion, you might collapse this text together with something like:

Step 1: If a constant value is already known for the work item, and we're
processing it for the first time, it's considered resolved.

Maybe those comments should be moved into the function body so that we only describe behavior once? [Or maybe just briefer callbacks?]

toolchain/check/import_ref.cpp
zygoloid: > Why is it fragile?

It's a subtle rule for a resolver function to follow, for which we'd have no checking. Eg, if a resolver function adds a call to `GetLocalConstantId` in the wrong place, we have an infinite loop, or if it intends to retry but somehow doesn't add any work (as it might be reasonable to do after creating a forward declaration in some cases) then it wouldn't actually get a retry. It seems much less error-prone to me for a resolver function to need to explicitly say whether it wants to be called again or not, rather than inferring that from other state. I think it also makes the resolve functions more readable.

zygoloid: > Note, regarding clarity, initial_work should probably be declared closer to where it's used. [The flow here is a bit hard to read now, maybe it could have reduced nesting with a `continue`]

Done.

toolchain/check/import_ref.cpp
zygoloid: Made the comments in this function more directly call back to the class comment.

I think it's useful to have the big comment on the class because it's useful context for the `TryResolve*` functions too.

jonmeow (approved):
Thanks! I think this is good progress.

PR #3799, opened 2024-03-20
Support implicit conversion from specific node IDs to node ID categories

josh11b:


zygoloid (approved):
Thanks!

I looked briefly into doing a compile-time check here but there are some layering problems. I guess we can see how well doing the check at runtime goes in practice.

PR #3801, opened 2024-03-20
Provide a TypedInstSwitch for simpler cases

jonmeow:
We're frequently writing code like:

```
      case ArrayType::Kind: {
        auto array = inst.As<ArrayType>();
```

TypedInstSwitch aims to provide a simpler API for this, loosely inspired by StringSwitch. Although simple compiles will probably be worse, the rough expectation is that the optimizer will close the performance gap for us in cases where it really matters.

### DIFF
--- a/toolchain/sem_ir/file.cpp
+++ b/toolchain/sem_ir/file.cpp
@@ -310,7 +310,7 @@ static auto StringifyTypeExprImpl(const SemIR::File& outer_sem_ir,
     };
 
     TypedInstSwitch(sem_ir.insts().Get(step.inst_id))
-        .Case<ArrayType>([&](auto inst) {
+        .Case([&](ArrayType inst) {
           if (step.index == 0) {
             out << "[";
             steps.push_back(step.Next());
@@ -319,7 +319,7 @@ static auto StringifyTypeExprImpl(const SemIR::File& outer_sem_ir,
             out << "; " << sem_ir.GetArrayBoundValue(inst.bound_id) << "]";
           }
         })
-        .Case<AssociatedEntityType>([&](auto inst) {
+        .Case([&](AssociatedEntityType inst) {
           if (step.index == 0) {
             out << "<associated ";
             steps.push_back(step.Next());
@@ -331,16 +331,16 @@ static auto StringifyTypeExprImpl(const SemIR::File& outer_sem_ir,
                 << ">";
           }
         })
-        .Case<AnyBindName>([&](auto inst) {
+        .Case([&](AnyBindName inst) {
           auto name_id = inst.bind_name_id;
           out << sem_ir.names().GetFormatted(
               sem_ir.bind_names().Get(name_id).name_id);
         })
-        .Case<ClassType>([&](auto inst) {
+        .Case([&](ClassType inst) {
           auto class_name_id = sem_ir.classes().Get(inst.class_id).name_id;
           out << sem_ir.names().GetFormatted(class_name_id);
         })
-        .Case<ConstType>([&](auto inst) {
+        .Case([&](ConstType inst) {
           if (step.index == 0) {
             out << "const ";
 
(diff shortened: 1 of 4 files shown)


jonmeow:
I'd discussed this with @zygoloid on related to #3798, so mainly looking to get your thoughts @chandlerc (though I do want to make sure @zygoloid agrees with the end result)

toolchain/sem_ir/file.cpp
```diff
@@ -308,214 +309,135 @@ static auto StringifyTypeExprImpl(const SemIR::File& outer_sem_ir,
...
-        if (step.index == 0) {
-          out << "<associated ";
-          steps.push_back(step.Next());
-          push_inst_id(sem_ir.types().GetInstId(assoc.entity_type_id));
-        } else {
+    TypedInstSwitch(sem_ir.insts().Get(step.inst_id))
+        .Case([&](ArrayType inst) {
+          if (step.index == 0) {
+            out << "[";
+            steps.push_back(step.Next());
```
zygoloid, line 316: The lambda bodies end up indented three steps from the switch call here. I wonder if we can do better.

I've looked at some [other syntactic options](https://godbolt.org/z/a1YPnTn15) to try to reduce the indentation here:

- (This PR, LLVM `StringSwitch`-style): `Switch(arg).Case(a).Case(b)`: 3 step indentation to lambda body.
- Variadic call: `Switch(arg, a, b)`: 3 step indentation to lambda body. However, may make table dispatch easier to write.
- `Switch(arg) | a | b`: 1 step indentation to lambda body. `clang-format` gives this fairly dense formatting:

    ```c++
    Switch(arg) | [&](TypeA inst) {
      // ...
    } | [&](TypeB inst) {
      // ...
    } | [&](TypeC inst) {
      // ...
    };
    ```
    Most other non-keyword binary operators (`*`, `||`, `^`, `+`, `=`, `<<`, `>>`) give similar results.
- `Switch(arg) or a or b`: 1 step indentation to lambda body. Less densely formatted:

    ```c++
    Switch(arg) or [&](TypeA inst) {
      // ...
    }
    or [&](TypeB inst) {
      // ...
    }
    or [&](TypeC inst) {
      // ...
    };
    ```

    but formatting goes wrong for lambdas with empty bodies.
- `Switch(arg), a, b`: 3 step indent to lambda body.

I think the non-keyword binary operator option gives the best formatting here -- it's reasonably dense, and there are no `}` lines that might suggest we're at the end of the statement. But are we prepared to use an operator for this?

zygoloid: For the shorter lambda bodies in `GetExprCategory`, we get [a formatting pretty similar to with `.Case`](https://godbolt.org/z/o3hqWqoWz) if we use a binary operator.

zygoloid (commented):
Is there any noticeable difference in compile time for `file.cpp`, particularly with `-c opt`? We'll be giving the inliner a real workout here :)

jonmeow:
> Is there any noticeable difference in compile time for `file.cpp`, particularly with `-c opt`? We'll be giving the inliner a real workout here :)

Maybe about a 10% increase? On trunk, I'm seeing about 6s. On branch, I'm seeing about 6.7s.

toolchain/sem_ir/file.cpp
jonmeow: The variadic might be appealing due to the `()` wrapping the full body of the switch, similar to an actual switch statement. In general, I'm a bit cautious about the understandability of passing lambdas to operators, and `or` in particular I think might not carry the right semantic for readers. I also think at least _some_ indentation makes it easier to identify the start and end.

That said regarding indentation, comparing your two example code links, you seem to get significantly different indentation for the `|` operator. I wonder if this may not be working as well as hoped? The inconsistency may also make it less desirable.

One thing that makes me hesitant about these alternatives is how they handle the `Cases` equivalent, where it might cast but not use the result. Maybe the variadic approach would make writing a dispatch table easier though, which I suspect we might like to end up at even though I'm not writing it here.

Either way, I'm currently pausing on digging into the syntax until we can get agreement at least in _principle_. If Chandler has a strong preference towards a `switch` statement, I think that would nix work.

jonmeow:
Going to start from scratch

PR #3720, opened 2024-02-22
Member binding operators

josh11b:
Define the binding operation used to compute the result of `x.y`, `p->y`, `x.(C.y)`, and `p->(C.y)` as calling a method from user-implementable interfaces.

PR #3802, opened 2024-03-20
`extend api`

josh11b:
Allow types to `extend api` other types, adding the names from the other type to its namespace, for forwarding and delegation use cases.


proposals/p3802.md
```diff
@@ -0,0 +1,1183 @@
...
+
+-   `extend impl as I` adds lookup into the implementation of an interface `I`.
+-   `extend base: B` adds lookup into the base class `B`.
+-   `extend adapt C` adds lookup into the adapted class `C`.
+-   [mixins are also planning](/proposals/p2760.md#future-work-mixins) on using
+    an `extend` syntax
+
+Interfaces and named constraints also
+[use the `extend` keyword](/docs/design/generics/details.md#interface-extension)
+to lookup into another, which is also required.
```
zygoloid: The antecedent of "which" here was unclear to me. I first read this as the lookup into another constraint being required, and then as the `extend` keyword being required (unlike in `extend impl`, where it can be omitted to turn off the `extend api` behavior). But I think this is "required" in the sense of a requirement -- implementing the other constraint is a requirement of the enclosing constraint.

Maybe:
```suggestion
to indicate that lookups should be performed in another constraint,
and to add a requirement that `Self` implements the other constraint.
```

proposals/p3802.md
```diff
@@ -0,0 +1,1183 @@
...
+
+For some use cases, we want to be able to add name to lookup into that type
```
zygoloid:
```suggestion
For some use cases, we want to be able to add to name lookup into a type
```

proposals/p3802.md
```diff
@@ -0,0 +1,1183 @@
...
+> [comment on #3720](https://github.com/carbon-language/carbon-lang/pull/3720/files/97004ac6748a4f39e664d7409401df1811adc25c#r1527084183).
+
+> **Note:** This is an alternative to defining an `alias` for each member of the
+> extended type. This avoids having to repeat them, which is both lengthy and
+> could get out of sync as the class evolves. The `extend api` approach works
+> even if, as in this example, we don't know the names of the members of the
+> type being extended.
+
+Like other uses of `extend`, an `extend api` declaration do not have access
+control modifiers and only operate on public names.
```
zygoloid:
```suggestion
Like other uses of `extend`, an `extend api` declaration does not have access
control modifiers and only operates on public names.
```

proposals/p3802.md
```diff
@@ -0,0 +1,1183 @@
...
+with consistent lookup rules as the other uses of `extend`, with conflicts
+handled as determined by leads issues
+[#2355](https://github.com/carbon-language/carbon-lang/issues/2355) and
+[#2745](https://github.com/carbon-language/carbon-lang/issues/2745).
+
+The extended type is required to be complete at the point of the `extend api`
+declaration, so these lookups into `T` can succeed.
+
+The general rule for resolving ambiguity for `extend`, which we apply here as
+well, is that if lookup into `Box(T)` succeeds, then that result is used and no
```
zygoloid: This would be easier to follow if it appeared after the `Box` example below or had a link to that example.

proposals/p3802.md
```diff
@@ -0,0 +1,1183 @@
...
+}
+var b: Box(String) = ...;
+var position: i32 = b.Search(32);
+```
+
+There are two ingredients to make this work:
+
+-   We need some way to make `b.Search` be equivalent to `b.(String.Search)` not
+    `b.(Box(String).Search)`.
+-   We need the act of binding a method of `String` to a `Box(String)` value
```
zygoloid:
```suggestion
-   We need the act of binding a method of `String` to a `Box(String)` value to
```

proposals/p3802.md
```diff
@@ -0,0 +1,1183 @@
...
+
+```carbon
+class Ref(T:! type) {
+  var ptr: T*;
+  extend api T;
+}
+impl forall [T:! type, U:! RefBind(T)]
+    U as ValueBind(Ref(T)) where .Result = Ref(U.Result) {
+  fn Op[self: Self](p: Ref(T)) -> Ref(Result) {
+    return {.ptr = self.Op(p->ptr)};
```
zygoloid:
```suggestion
    return {.ptr = self.Op(p.ptr)};
```

proposals/p3802.md
```diff
@@ -0,0 +1,1183 @@
...
+also be done using `extend api` and a different set of binding implementations:
+
+```carbon
+class Ref(T:! type) {
+  var ptr: T*;
+  extend api T;
+}
+impl forall [T:! type, U:! RefBind(T)]
+    U as ValueBind(Ref(T)) where .Result = Ref(U.Result) {
+  fn Op[self: Self](p: Ref(T)) -> Ref(Result) {
```
zygoloid:
```suggestion
  fn Op[self: Self](p: Ref(T)) -> Ref(U.Result) {
```
or
```suggestion
  fn Op[self: Self](p: Ref(T)) -> Result {
```
I think `Result` here names `(U as ValueBind(Ref(T))).Result`, which is `Ref(U.Result)`.

proposals/p3802.md
```diff
@@ -0,0 +1,1183 @@
...
+    return {.ptr = self.Op(p->ptr)};
+  }
+}
+// (Plus impls for the other three combinations of ValueBind and RefBind.)
```
zygoloid: Exploring these myself:

```carbon
// <Ref(T) value>.<ref bindable> produces a Ref(R) value expression.
impl forall [T:! type, U:! RefBind(T)]
    U as ValueBind(Ref(T)) where .Result = Ref(U.Result) {
  fn Op[self: Self](p: Ref(T)) -> Result {
    return {.ptr = self.Op(p->ptr)};
  }
}

// <Ref(T) ref expr>.<ref bindable> produces a reference expression.
// Reference collapse: no added Ref(...) here.
impl forall [T:! type, U:! RefBind(T)]
    U as RefBind(Ref(T)) where .Result = U.Result {
  fn Op[self: Self](p: Ref(T)*) -> Result* {
    return self.Op(p->ptr);
  }
}

// <Ref(T) value>.<value bindable> produces a value expression.
// A value binding is performed to turn the Ref(T) into a T.
impl forall [T:! type, U:! ValueBind(T)]
    U as ValueBind(Ref(T)) where .Result = U.Result {
  fn Op[self: Self](p: Ref(T)) -> Result {
    return self.Op(*p.ptr);
  }
}
```

I think implementing `U as RefBind(Ref(T))` in terms of `U as ValueBind(T)` is not possible. We can't produce a pointer as a result of the binding.

There's also overlap between the first and third `impl`s. And the first and second violate the new coherence rule in #3720:  given `ref: Ref({.n: i32})`, `ref.n` will either be a value expression of type `Ref(i32)` or a reference expression of type `i32` depending on the expression category of `ref`.

I think maybe the right thing to do is to drop the second `impl` entirely, and have only two `impl`s, both for `ValueBind`, with a `match_first`:

```carbon
match_first {

// <Ref(T) value>.<ref bindable> produces a value expression of type `Ref(R)`.
impl forall [T:! type, U:! RefBind(T)]
    U as ValueBind(Ref(T)) where .Result = Ref(U.Result) {
  fn Op[self: Self](p: Ref(T)) -> Result {
    return {.ptr = self.Op(p->ptr)};
  }
}

// <Ref(T) value>.<value bindable> produces a value expression of type `R`.
// A value binding is performed to turn the Ref(T) into a T.
impl forall [T:! type, U:! ValueBind(T)]
    U as ValueBind(Ref(T)) where .Result = U.Result {
  fn Op[self: Self](p: Ref(T)) -> Result {
    return self.Op(*p.ptr);
  }
}

}
```

For this to really work, I think we'd need binding to fall back from `RefBind` to `ValueBind` if reference binding isn't possible, as discussed in #3720. The above pair of `impl`s would then be implementing the same fallback semantics for `Ref(T)`.

proposals/p3802.md
```diff
@@ -0,0 +1,1183 @@
...
+}
+```
+
+Note, however, `Ref(T)` doesn't preserve the interface implementations of `T`,
+see [future work](#future-work-customizing-impl-binding).
+
+This application of `extend api` and binding was first described in
+[this comment on #3720](https://github.com/carbon-language/carbon-lang/pull/3720/files/37e967ff53e89e345eecb49ec79c6dfbe18a3c54#r1513712572).
+
+### Use case: with implicit conversion
```
zygoloid, line 279: Note to self: I'm parking this review here while looking at other things.

proposals/p3802.md
```diff
@@ -0,0 +1,1184 @@
...
+-   `extend impl as I` adds lookup into the implementation of an interface `I`.
+-   `extend base: B` adds lookup into the base class `B`.
+-   `extend adapt C` adds lookup into the adapted class `C`.
+-   [mixins are also planning](/proposals/p2760.md#future-work-mixins) on using
+    an `extend` syntax
+
+Interfaces and named constraints also
+[use the `extend` keyword](/docs/design/generics/details.md#interface-extension)
+to indicate that lookups should be performed in another constraint,
+and to add a requirement that `Self` implements the other constraint.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
to indicate that lookups should be performed in another constraint, and to add a
requirement that `Self` implements the other constraint.
```

proposals/p3802.md
```diff
@@ -0,0 +1,1184 @@
...
+
+For some use cases, we want to be able to add to name lookup into a type
+without making other changes to the type, for which we introduce `extend api`,
+with consistent lookup rules as the other uses of `extend`, with conflicts
+handled as determined by leads issues
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
For some use cases, we want to be able to add to name lookup into a type without
making other changes to the type, for which we introduce `extend api`, with
consistent lookup rules as the other uses of `extend`, with conflicts handled as
determined by leads issues
```

proposals/p3802.md
josh11b: I don't think this works. #3720 only ever produces a reference expression (by automatically dereferencing the returned pointer) when the expression on the left side of the `.` is a reference expression. I think this makes `Ref(T)` impossible, and I should probably delete this section.

josh11b: Sorry, I had not understood what you were proposing (that you can't get a reference expression, but can get a `Ref(T)` value that acts like one). I've made these changes, added the additional support for calls on `Ref(T)` values, and moved this section to future work, [as discussed in open discussion on 05-16](https://docs.google.com/document/d/1s3mMCupmuSpWOFJGnvjoElcBIe2aoaysTIdyczvKX84/edit?resourcekey=0-G095Wc3sR6pW1hLJbGgE0g&tab=t.0#heading=h.3qlpp5e56u46).

proposals/p3802.md
josh11b: I rewrote it a bit, but let me know if it would be better to move this until after the `Box` example.

josh11b (commented):
Should be ready for another look now.

zygoloid:
@josh11b and I have both lost track of our state on this. We intend to return to this at some point but this PR is not a current priority.

josh11b:
It recently came up in discussion that this feature might be accessed using `extend alias` instead of `extend api`. Rationale:
* `api` is no longer a keyword, not worth creating it for this case
* The effects of this feature can be simulated by using `alias`.
Similarly, if #5028 replaces `alias` with something else (like `using`), we could use the same keyword here (`extend using`).

What do you think?

PR #3803, opened 2024-03-21
Initial support for builtin functions.

zygoloid:
For now, a builtin function is defined by specifying a string literal initializer in a function declaration:

```carbon
fn MyBuiltin(a: i32) -> i32 = "builtin.name";
```

End-to-end support is included for a sample `"int.add"` builtin performing integer addition, covering constant evaluation and code generation.

The implementation here needs substantial refactoring before we'll be ready to start adding more builtins. That refactoring work will be coming next. This change is aiming to checkpoint some incremental progress.


jonmeow:
> ```
> fn MyBuiltin(a: i32) -> i32 = "builtin.name";
> ```

I feel really weird about this syntax, I'm still cheering on something attribute based, e.g.:

`@__builtin.name fn MyBuiltin(a: i32) -> i32;`

Do you expect assignment is the direction we want to go? Because a chunk of this doesn't matter if we instead end up with something like:

`fn MyBuiltin(a: i32) -> i32 = __builtin_name`

Because all that really is is "how do we spell the token".

[though I will note I think string parsing is inefficient relative to identifier or keyword parsing]

toolchain/sem_ir/formatter.cpp
```diff
@@ -795,6 +796,15 @@ class Formatter {
       FormatType(fn.return_type_id);
     }
 
+    // TODO: Move this conversion of kind to string elsewhere.
```
jonmeow, line 799: What about using EnumBase for BuiltinFunctionKind? Note, that'd also give you a name automatically if you were fine with `IntAdd` instead of `int.add`

Note, fine if it's not part of this PR

toolchain/parse/typed_nodes.h
```diff
@@ -312,6 +312,20 @@ struct FunctionDefinition {
   llvm::SmallVector<AnyStatementId> body;
 };
 
+using BuiltinFunctionDefinitionStart =
+    FunctionSignature<NodeKind::BuiltinFunctionDefinitionStart,
+                      NodeCategory::None>;
+using BuiltinName = LeafNode<NodeKind::BuiltinName>;
+
+// A builtin function definition: `fn F() -> i32 = "builtin-name";`
```
jonmeow:
```suggestion
// A builtin function definition: `fn F() -> i32 = "builtin.name";`
```

This seems to have fallen out of sync?

toolchain/parse/handle_function.cpp
```diff
@@ -52,6 +52,30 @@ auto HandleFunctionSignatureFinish(Context& context) -> void {
       context.PushState(State::StatementScopeLoop);
       break;
     }
+    case Lex::TokenKind::Equal: {
```
jonmeow, line 55: Is it fair to discuss whether assignment is the right syntax versus a modifier keyword due to the effects on parse complexity?

toolchain/lower/testdata/builtins/method_vs_nonmethod.carbon
```diff
@@ -0,0 +1,24 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+
+fn AddNonmethod(a: i32, b: i32) -> i32 = "int.add";
```
jonmeow, line 7: Just to note though, but not having a `__builtin` or anything like that here makes this feel like it's really something intended for developers at the large, even if poorly documented.

toolchain/lower/handle.cpp
```diff
@@ -166,14 +167,75 @@ auto HandleBuiltin(FunctionContext& /*context*/, SemIR::InstId /*inst_id*/,
   CARBON_FATAL() << "TODO: Add support: " << inst;
 }
 
+static auto GetCalleeBuiltinFunctionKind(const SemIR::File& sem_ir,
```
jonmeow, line 172: Is this and HandleBuiltinCall worth a function comment, since it's not a boilerplate handler?

toolchain/lower/handle.cpp
```diff
@@ -166,14 +167,75 @@ auto HandleBuiltin(FunctionContext& /*context*/, SemIR::InstId /*inst_id*/,
...
+      if (lhs_type != rhs_type || lhs_type != result_type ||
+          context.sem_ir().types().GetInstId(lhs_type) !=
+              SemIR::InstId::BuiltinIntType) {
+        break;
+      }
+      constexpr bool SignedOverflowIsUB = false;
+      context.SetLocal(
+          inst_id, context.builder().CreateAdd(context.GetValue(arg_ids[0]),
+                                               context.GetValue(arg_ids[1]), "",
+                                               false, SignedOverflowIsUB));
```
jonmeow:
```suggestion
                                               context.GetValue(arg_ids[1]), /*Name=*/"",
                                               /*HasNUW=*/false, SignedOverflowIsUB));
```

Trying to figure out what the not-really-provided parameters are... not that I intuitively understand HasNUW/HasNSW as "has no unsigned/signed wrap", but at least now I know the parameter names are a little opaque. Makes me kind of wonder if the constant should just be "OverflowIsUB" and pass it into both, e.g.:

```suggestion
                                               context.GetValue(arg_ids[1]), /*Name=*/"",
                                               /*HasNUW=*/OverflowIsUB, /*HasNSW=*/OverflowIsUB));
```

But I assume here your thought was this only would work for signed integers, so the unsigned value doesn't matter?

toolchain/lower/handle.cpp
```diff
@@ -166,14 +167,75 @@ auto HandleBuiltin(FunctionContext& /*context*/, SemIR::InstId /*inst_id*/,
...
+
+static auto HandleBuiltinCall(FunctionContext& context, SemIR::InstId inst_id,
+                              SemIR::BuiltinFunctionKind builtin_kind,
+                              llvm::ArrayRef<SemIR::InstId> arg_ids) -> void {
+  switch (builtin_kind) {
+    case SemIR::BuiltinFunctionKind::None:
+      CARBON_FATAL() << "No callee in function call.";
+
+    case SemIR::BuiltinFunctionKind::IntAdd: {
+      // TODO: Move type checking to the point where we make the call.
```
jonmeow, line 199: FWIW I think I comment about this elsewhere, but I think type checking should be done on the function declaration rather than callers.

toolchain/check/testdata/function/builtin/method.carbon
```diff
@@ -0,0 +1,97 @@
...
+//
+// AUTOUPDATE
+
+interface I {
+  fn F[self: Self](other: Self) -> Self;
+}
+
+impl i32 as I {
+  fn F[self: i32](other: i32) -> i32 = "int.add";
+}
```
jonmeow, line 13: Should there be a test of:

```
interface I {
  fn F[self: Self](other: Self) -> Self = "int.add";
}
```

toolchain/check/handle_function.cpp
```diff
@@ -284,4 +284,53 @@ auto HandleFunctionDefinition(Context& context,
...
+auto HandleBuiltinFunctionDefinition(
+    Context& context, Parse::BuiltinFunctionDefinitionId /*node_id*/) -> bool {
+  auto name_id =
+      context.node_stack().PopForSoloNodeId<Parse::NodeKind::BuiltinName>();
+  auto function_id =
+      context.node_stack()
+          .Pop<Parse::NodeKind::BuiltinFunctionDefinitionStart>();
+
+  auto& function = context.functions().Get(function_id);
+  function.builtin_kind = LookupBuiltinFunctionKind(context, name_id);
```
jonmeow, line 330: Why are parameters only validated on call instead of here? i.e., why doesn't a builtin function constrain the parameters of the declaration?

toolchain/check/testdata/builtins/int_add.carbon
```diff
@@ -0,0 +1,231 @@
...
+var bad_return_type: [i32; BadReturnType(1, 2)];
+
+// CHECK:STDERR: fail_bad_decl.carbon:[[@LINE+6]]:21: ERROR: 3 argument(s) passed to function expecting 2 argument(s).
+// CHECK:STDERR: var bad_call: [i32; JustRight(1, 2, 3)];
+// CHECK:STDERR:                     ^~~~~~~~~~
+// CHECK:STDERR: fail_bad_decl.carbon:[[@LINE-18]]:1: Calling function declared here.
+// CHECK:STDERR: fn JustRight(a: i32, b: i32) -> i32 = "int.add";
+// CHECK:STDERR: ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+var bad_call: [i32; JustRight(1, 2, 3)];
+
```
jonmeow, line 46:
```suggestion
var bad_call: [i32; JustRight(1, 2, 3)];

fn WrapTooFew(a: i32) -> i32 {
  return TooFew(a);
}

fn WrapTooMany(a: i32) -> i32 {
  return TooMany(a, a, a);
}

fn WrapBadReturnType(a: i32) -> i32 {
  return BadReturnType(a, a);
}

```

You're testing cases with constant parameters, should it also be tested with non-constant parameters?

toolchain/sem_ir/function.h
```diff
@@ -59,6 +68,11 @@ struct Function : public Printable<Function> {
   // Whether the declaration is extern.
   bool is_extern;
 
+  // The following members are set at the end of a builtin function definition.
+
+  // If this is a builtin function, the corresponding builtin kind.
```
jonmeow, line 73:
```suggestion
  // If this is a builtin function, the corresponding builtin kind.
  // TODO: Propagate across imports.
```

toolchain/check/handle_function.cpp
zygoloid: This PR is trying to just do the simplest thing possible for now.

We should definitely think about what checks should be applied to the declaration. I am not sure that we'll be able to fully check the declaration in general -- the function declaration could be generic or in a generic context, and we may not know enough to see whether the declaration is valid until we have the concrete arguments. But we could do at least some of the checking when we see the declaration. Also, I am expecting that most of the builtins declared in the prelude won't be used in most compilations, so deferring the checks until use may give a performance advantage. I think this is something we should revisit once we have a more complete picture of what kinds of builtins we'll have.

(In Clang, we ended up with a hybrid situation where some builtins have specified parameter types and others are checked on call, resulting in a somewhat inconsistent state. I think if we do checks on declarations we'll end up in a similar place. It's not terrible, but it would be nice to find a more consistent answer.)

toolchain/check/testdata/builtins/int_add.carbon
zygoloid: Sure, added with a TODO to diagnose these before we get to lower.

toolchain/check/testdata/function/builtin/method.carbon
zygoloid: Hm, we don't have any implementation or testing of defined-but-not-defaulted associated entities in interfaces at all. Added some.

toolchain/lower/handle.cpp
zygoloid: Sure.

toolchain/lower/handle.cpp
zygoloid: Added parameter comments for NSW / NUW, since they could easily be transposed. I've added a name of "add" for the add instruction, which should obviate the need for a parameter comment there.

Regardless of whether the addition is signed or unsigned, we never want to set `NUW` -- that indicates that we're performing unsigned arithmetic for which overflow is UB, which for us it never is.

toolchain/lower/testdata/builtins/method_vs_nonmethod.carbon
zygoloid: Perhaps we can restrict this syntax to the `Core` package. (I think I'd also be fine with starting all the builtin names with `builtin.` or `core.` or `lang.` or something.)

We could consider having a mode for the toolchain that says "I'm compiling the core package itself" that turns off the automatic dependency on `Core`, enables defining builtins, and causes lookup for `Core` names to look in the current package scope instead? (But not in this PR.)

toolchain/parse/handle_function.cpp
zygoloid: I thought this approach would be simpler in the short term than adding support for modifiers with arguments, given where our current modifier support is. This approach also doesn't require adding any new token kinds. (From a design perspective rather than an implementation perspective, I also favor putting something after the function signature rather than before, because this is adding a definition for the function, and that's where we put definitions in general.)

But the goal here was just to add something that works for now, and explicitly not to block on working out what's the right long-term approach.

toolchain/parse/typed_nodes.h
zygoloid: This is just a placeholder to demonstrate that a string literal goes there. `"builtin.name"` might suggest that builtin names start with `builtin.`, which is not the current intent. Switched to a space instead; does that work for you?

toolchain/sem_ir/formatter.cpp
zygoloid: Noted.

Part of the refactoring that I'm planning is to use a `.def` file to define the builtins, or something like that. I'm not yet sure whether we want `EnumBase` or something more like `IdBase` here -- Clang has a mechanism for dynamically adding builtins, which would suggest something more like `IdBase`, but we might not want to support that.

toolchain/sem_ir/function.h
zygoloid: Implemented import support instead :)

toolchain/lower/handle.cpp
zygoloid: Noted, let's keep the discussion with the other comment.

toolchain/lower/testdata/builtins/method_vs_nonmethod.carbon
jonmeow: Yeah, that might be worthwhile as enforcement. Although, from your suggestions -- maybe "Internal."? [or "private."?]

toolchain/check/handle_function.cpp
jonmeow: I wonder if we could instead constrain instantiation. I'd still expect a signature on the generic that would correspond to something the builtin could support -- e.g., the builtin could require constraints, maybe only applying to a particular specialization of the generic.

But I think that's not an issue for this PR.

jonmeow (approved):
Looks good, feel free to merge once tests are passing.

PR #3804, opened 2024-03-21
Rename DiagnosticLocationTranslator -> DiagnosticConverter

jonmeow:
Since the addition of TranslateArg, I don't think this type is going to go away (cutting a TODO). Refactoring names slightly to fit the current role, and adding const to ConvertLocation.


zygoloid (commented):
I think the name `DiagnosticTranslator` really strongly suggests that it's going to do localization of diagnostic text. Maybe something like `Converter` is a better word to use here than `Translator`?

jonmeow:
> I think the name `DiagnosticTranslator` really strongly suggests that it's going to do localization of diagnostic text. Maybe something like `Converter` is a better word to use here than `Translator`?

Done

zygoloid (approved):

PR #3805, opened 2024-03-21
Collapse diagnostic errors and notes into a single vector.

jonmeow:
On #3792 it was requested to put context-related diagnostic messages prior to the diagnostic error. In order to support that, remove the distinction that an error needs to come first on DiagnosticMessage.

### DIFF
--- a/toolchain/diagnostics/sorting_diagnostic_consumer.h
+++ b/toolchain/diagnostics/sorting_diagnostic_consumer.h
@@ -35,10 +35,12 @@ class SortingDiagnosticConsumer : public DiagnosticConsumer {
   void Flush() override {
     llvm::stable_sort(diagnostics_,
                       [](const Diagnostic& lhs, const Diagnostic& rhs) {
-                        return std::tie(lhs.message.location.line_number,
-                                        lhs.message.location.column_number) <
-                               std::tie(rhs.message.location.line_number,
-                                        rhs.message.location.column_number);
+                        const auto& lhs_loc = lhs.messages[0].location;
+                        const auto& rhs_loc = rhs.messages[0].location;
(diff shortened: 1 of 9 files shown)


jonmeow:
Note, stacking changes here.

toolchain/diagnostics/sorting_diagnostic_consumer.h
```diff
@@ -35,10 +35,12 @@ class SortingDiagnosticConsumer : public DiagnosticConsumer {
   void Flush() override {
     llvm::stable_sort(diagnostics_,
                       [](const Diagnostic& lhs, const Diagnostic& rhs) {
-                        return std::tie(lhs.message.location.line_number,
-                                        lhs.message.location.column_number) <
-                               std::tie(rhs.message.location.line_number,
-                                        rhs.message.location.column_number);
+                        const auto& lhs_loc = lhs.messages[0].location;
+                        const auto& rhs_loc = rhs.messages[0].location;
```
zygoloid, line 39: Should we still track the primary diagnostic index in the vector, so we can use it for sorting?

jonmeow: I think this *is* the value we want for sorting, isn't it?

jonmeow: Or put differently, we could iterate through both messages for comparing: I'm being lazy because there's currently nothing being affected. But I believe the primary location is the first location.

zygoloid: Hm, if we're just using prefixes of the primary diagnostic for "imported here" information, then yes, I suppose it is what we want for sorting.

zygoloid (approved):

PR #3806, opened 2024-03-21
Print a blank line after a diagnostic.

jonmeow:
The purpose of the newline is to make it clearer where a given diagnostic begins and ends, particularly as the first message of a diagnostic may not be the error.

This is a trivial code change, but ripples edits through test files.

### DIFF
--- a/toolchain/check/testdata/alias/fail_control_flow.carbon
+++ b/toolchain/check/testdata/alias/fail_control_flow.carbon
@@ -4,12 +4,14 @@
 //
 // AUTOUPDATE
 
-// CHECK:STDERR: fail_control_flow.carbon:[[@LINE+9]]:11: ERROR: Semantics TODO: `Control flow expressions are currently only supported inside functions.`.
+// CHECK:STDERR: fail_control_flow.carbon:[[@LINE+11]]:11: ERROR: Semantics TODO: `Control flow expressions are currently only supported inside functions.`.
 // CHECK:STDERR: alias a = true or false;
 // CHECK:STDERR:           ^~~~~~~
-// CHECK:STDERR: fail_control_flow.carbon:[[@LINE+6]]:11: ERROR: Semantics TODO: `Control flow expressions are currently only supported inside functions.`.
+// CHECK:STDERR:
+// CHECK:STDERR: fail_control_flow.carbon:[[@LINE+7]]:11: ERROR: Semantics TODO: `Control flow expressions are currently only supported inside functions.`.
 // CHECK:STDERR: alias a = true or false;
 // CHECK:STDERR:           ^~~~~~~~~~~~~
+// CHECK:STDERR:
 // CHECK:STDERR: fail_control_flow.carbon:[[@LINE+3]]:11: ERROR: Alias initializer must be a name reference.
 // CHECK:STDERR: alias a = true or false;
 // CHECK:STDERR:           ^~~~~~~~~~~~~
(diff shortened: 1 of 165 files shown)


jonmeow:
Note, stacking changes here.

toolchain/diagnostics/diagnostic_consumer.cpp
```diff
@@ -41,6 +41,7 @@ auto StreamDiagnosticConsumer::HandleDiagnostic(Diagnostic diagnostic) -> void {
       *stream_ << "\n";
     }
   }
+  *stream_ << "\n";
```
zygoloid: Would it make sense to suppress the blank line after the final diagnostic? I think I wouldn't expect an additional blank line there.

zygoloid (commented):
Have you considered making `file_test` ignore blank lines, at least by default? I think the check lines for the blank lines between diagnostics aren't really making our tests more readable.

jonmeow:
> Have you considered making `file_test` ignore blank lines, at least by default? I think the check lines for the blank lines between diagnostics aren't really making our tests more readable.

I'd be really hesitant to conditionally ignore output. How do we establish that we're only ignoring the blank lines we mean to?

zygoloid:
> > Have you considered making `file_test` ignore blank lines, at least by default? I think the check lines for the blank lines between diagnostics aren't really making our tests more readable.
> 
> I'd be really hesitant to conditionally ignore output. How do we establish that we're only ignoring the blank lines we mean to?

Perhaps we could add a `// KEEP-BLANK-LINES` marker in the file preamble for the tests where we want to check the exact output.

jonmeow:
> > > Have you considered making `file_test` ignore blank lines, at least by default? I think the check lines for the blank lines between diagnostics aren't really making our tests more readable.
> > 
> > 
> > I'd be really hesitant to conditionally ignore output. How do we establish that we're only ignoring the blank lines we mean to?
> 
> Perhaps we could add a `// KEEP-BLANK-LINES` marker in the file preamble for the tests where we want to check the exact output.

I'd really rather not: I think the common case is that we should just have the blank line and skim past it. The red flag would be having _more_ blank lines that shouldn't be there.

If we really want to do a better job here, I'd be happier with something like restructuring the diagnostics *only* for tests, maybe a --dont-print-blank-lines-between-diagnostics flag. If you really want this, that's the route I'd go, to just not test these blank lines I'm adding for clarity.

jonmeow:
Note, the blank lines also serve an important utility in these tests: they still make it clearer where one diagnostic ends and another begins when the autoupdate association results in something unclear, which is the issue that led to this whole PR chain.

toolchain/diagnostics/diagnostic_consumer.cpp
jonmeow: I've updated this PR with a force-push to do that. The force-push is so that I don't have multiple commits editing every test to need to rebase on branch updates.

toolchain/driver/driver.cpp
```diff
@@ -394,17 +394,16 @@ auto Driver::ValidateCompileOptions(const CompileOptions& options) const
 // Ties together information for a file being compiled.
 class Driver::CompilationUnit {
  public:
-  explicit CompilationUnit(Driver* driver, const CompileOptions& options,
+  explicit CompilationUnit(Driver* driver, const CompileOptions& options, DiagnosticConsumer* consumer,
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  explicit CompilationUnit(Driver* driver, const CompileOptions& options,
                           DiagnosticConsumer* consumer,
```

toolchain/check/testdata/alias/fail_control_flow.carbon
```diff
@@ -4,12 +4,14 @@
 //
 // AUTOUPDATE
 
-// CHECK:STDERR: fail_control_flow.carbon:[[@LINE+9]]:11: ERROR: Semantics TODO: `Control flow expressions are currently only supported inside functions.`.
+// CHECK:STDERR: fail_control_flow.carbon:[[@LINE+11]]:11: ERROR: Semantics TODO: `Control flow expressions are currently only supported inside functions.`.
```
jonmeow, line 7: To note on blank lines, we already have non-import-related tests which print multiple diagnostics for the same line of code. I do think that the blank line makes the separation clearer in tests.

zygoloid (approved):

PR #3807, opened 2024-03-21
Add import context for locations.

jonmeow:
As discussed around #3792, identify the import a diagnostic message came from prior to the diagnostic message itself. This occurs during location translation so that the logic can be central.

I'd considered associating the parse node with ImportRef instructions, but I realized about halfway through that because I need to store the ImportDirectiveId on the ImportIR for cross-package imports, it's there for use in location translation without extra work. That saves a fair amount of stringing it through declarations, as well as an oddity where ImportRef instructions would have a node that didn't really represent them.

### DIFF
--- a/toolchain/diagnostics/diagnostic_converter.h
+++ b/toolchain/diagnostics/diagnostic_converter.h
@@ -15,14 +15,16 @@ namespace Carbon {
 template <typename LocationT>
 class DiagnosticConverter {
  public:
+  // Callback type used to report context messages from ConvertLocation.
+  // Note that the first parameter type is DiagnosticLocation rather than
+  // LocationT, because ConvertLocation must not recurse.
   using ContextFnT = llvm::function_ref<void(
       DiagnosticLocation, const Internal::DiagnosticBase<>&)>;
 
   virtual ~DiagnosticConverter() = default;
 
   // Converts a LocationT to a DiagnosticLocation. ConvertLocation may invoke
-  // context_fn to provide context messages, but they must already have
-  // converted locations.
+  // context_fn to provide context messages.
   virtual auto ConvertLocation(LocationT loc, ContextFnT context_fn) const
       -> DiagnosticLocation = 0;
 
--- a/toolchain/diagnostics/diagnostic_converter.h
+++ b/toolchain/diagnostics/diagnostic_converter.h
@@ -20,9 +20,9 @@ class DiagnosticConverter {
 
   virtual ~DiagnosticConverter() = default;
 
-  // Converts a LocationT to a DiagnosticLocation. The function may optionally
-  // provide context messages using context_fn, but these must be simple and
-  // already have converted locations.
+  // Converts a LocationT to a DiagnosticLocation. ConvertLocation may invoke
+  // context_fn to provide context messages, but they must already have
+  // converted locations.
   virtual auto ConvertLocation(LocationT loc, ContextFnT context_fn) const
       -> DiagnosticLocation = 0;
 
--- a/toolchain/diagnostics/diagnostic_converter.h
+++ b/toolchain/diagnostics/diagnostic_converter.h
@@ -15,9 +15,16 @@ namespace Carbon {
 template <typename LocationT>
 class DiagnosticConverter {
  public:
+  using ContextFnT = llvm::function_ref<void(
+      DiagnosticLocation, const Internal::DiagnosticBase<>&)>;
+
   virtual ~DiagnosticConverter() = default;
 
-  virtual auto ConvertLocation(LocationT loc) const -> DiagnosticLocation = 0;
+  // Converts a LocationT to a DiagnosticLocation. The function may optionally
+  // provide context messages using context_fn, but these must be simple and
+  // already have converted locations.
+  virtual auto ConvertLocation(LocationT loc, ContextFnT context_fn) const
+      -> DiagnosticLocation = 0;
 
   // Converts arg types as needed. Not all uses require conversion, so the
   // default returns the argument unchanged.
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -36,10 +36,11 @@ class SemIRDiagnosticConverter : public DiagnosticConverter<SemIRLocation> {
       const SemIR::File* sem_ir)
       : node_converters_(node_converters), sem_ir_(sem_ir) {}
 
-  auto ConvertLocation(SemIRLocation loc) const -> DiagnosticLocation override {
+  auto ConvertLocation(SemIRLocation loc, ContextFnT context_fn) const
+      -> DiagnosticLocation override {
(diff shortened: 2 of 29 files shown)


jonmeow:
Note, stacking changes here. But I think this is the end of this stack.

toolchain/diagnostics/diagnostic_converter.h
```diff
@@ -27,9 +27,16 @@ enum class DiagnosticTypeConversion : int8_t {
...
  public:
+  using ContextFnT = llvm::function_ref<void(
+      DiagnosticLocation, const Internal::DiagnosticBase<>&)>;
+
   virtual ~DiagnosticConverter() = default;
 
-  virtual auto ConvertLocation(LocationT loc) const -> DiagnosticLocation = 0;
+  // Converts a LocationT to a DiagnosticLocation. The function may optionally
+  // provide context messages using context_fn, but these must be simple and
+  // already have converted locations.
```
geoffromer: I'm having trouble following this comment, but I can't tell if that's because I'm not familiar with this code or if it needs clarification. What are these context messages, and what do they have to do with the conversion? What counts as "simple"? Isn't the fact that they already have converted locations guaranteed by the fact that `context_fn` takes a `DiagnosticLocation` as its argument?

Also, this comment seems to be expressed in terms of requirements on the implementation ("...these _must_ be simple...") and permission given to the implementation ("the function may _optionally_..."). I think it would be more conventional to express it in terms of requirements on/permissions to the caller, and guarantees made by the implementation, e.g. "ConvertLocation may invoke context_fn to provide context messages, but they will be simple, and already have converted locations."

geoffromer (commented):
(Porting a comment from my review of #3811)

toolchain/diagnostics/diagnostic_converter.h
jonmeow: Done, dropping "simple".

geoffromer: Is it possible for a `DiagnosticLocation` to not be a "converted location"?

jonmeow: If I understand what you're trying to ask, while a `DiagnosticLocation` could in theory be tracked directly, it's unlikely [would impose overhead] and not how the code works. If I'm misunderstanding, maybe you could rephrase your question?

geoffromer: I'm trying to understand what "they must already have converted locations" means. In this context, I would ordinarily assume it means that their locations must already have been converted to `DiagnosticLocation`, but the relevant parameter of `context_fn` has type `DiagnosticLocation`, so that statement seems equivalent to "calls to `context_fn` must type-check". So I feel like I must have misunderstood something; my question was a guess at where I went wrong, but I don't understand your answer.

(By the way, in line with my original comment, I suggest changing "must" to "will").

jonmeow: > I'm trying to understand what "they must already have converted locations" means. In this context, I would ordinarily assume it means that their locations must already have been converted to `DiagnosticLocation`, but the relevant parameter of `context_fn` has type `DiagnosticLocation`, so that statement seems equivalent to "calls to `context_fn` must type-check". So I feel like I must have misunderstood something; my question was a guess at where I went wrong, but I don't understand your answer.

That's essentially what it's saying. I think the difference is that I'm expecting this to be more subtle than you; I expect callers might be surprised given that every other API deals with LocationT, and I'm emphasizing that this is a deliberate choice through the comment.

> (By the way, in line with my original comment, I suggest changing "must" to "will").

Maybe you could clarify why? I think "must" as a good grammatical match for "may": when something _may_ happen, then it _must_ occur in a given way; it's an obligation. For contrast, "will" implies inevitability that doesn't exist due to "may". [when...will]

Note, some of this may come down to personal writing styles.

toolchain/diagnostics/diagnostic_converter.h
```diff
@@ -15,9 +15,16 @@ namespace Carbon {
 template <typename LocationT>
 class DiagnosticConverter {
  public:
+  using ContextFnT = llvm::function_ref<void(
```
geoffromer, line 21:
```suggestion
  // Callback type used to report context messages from the conversion.
  // Note that the first parameter type is DiagnosticLocation rather than
  // LocationT, because it's an output of the conversion.
  using ContextFnT = llvm::function_ref<void(
```

toolchain/diagnostics/diagnostic_converter.h
geoffromer: > I expect callers might be surprised given that every other API deals with LocationT, and I'm emphasizing that this is a deliberate choice through the comment.

In that case, it might be clearer as a comment on `ContextFnT`, where it can talk about the parameter type more directly. I've left a suggestion for doing that.

> > (By the way, in line with my original comment, I suggest changing "must" to "will").
>
> Maybe you could clarify why? I think "must" as a good grammatical match for "may": when something _may_ happen, then it _must_ occur in a given way; it's an obligation. For contrast, "will" implies inevitability that doesn't exist due to "may". [when...will]
> 
> Note, some of this may come down to personal writing styles.

This is a moot point if you accept my suggestion above, but in case it's helpful: in my experience, people typically use language of factual assertion to give guarantees and take on obligations (e.g. "The project _will_ be done by Friday"), and use language of obligation and permission to impose requirements on others (e.g. "The project _must_ be done by Friday"). I automatically read function comments as speaking to callers on behalf of the implementers, not the other way around, so when a comment says "must", I assume that it's imposing a requirement on the callers, not making a guarantee to them (i.e. imposing a requirement on implementers). Deviating from that convention is particularly rough for me in this case, because the inversion of control makes it harder than usual to work out the intent from context.

My suggestion of using "may" caused confusion here because it can be read as granting permission, or as stating a factual possibility (like "I may work from home tomorrow"). The latter is what I intended, but it might have been better to use "can" or "might" instead, to reduce the ambiguity.

zygoloid (approved):
LGTM. There's quite a long stack of work on top of this, so if we can land this to unblock that work, and finish any remaining discussion on the DiagnosticConverter comments in a separate PR review, I think that'd be helpful.

geoffromer:
> There's quite a long stack of work on top of this, so if we can land this to unblock that work, and finish any remaining discussion on the DiagnosticConverter comments in a separate PR review, I think that'd be helpful.

That's fine with me.

toolchain/diagnostics/diagnostic_converter.h
jonmeow: Adopted the other suggested comment on ContextFn, removed this clause.

toolchain/diagnostics/diagnostic_converter.h
jonmeow: Done, with some edits to specify ConvertLocation

PR #3809, opened 2024-03-25
Start adding extern logic for functions.

jonmeow:
This starts propagating is_extern on import, and warns when merging an imported non-extern declaration with a local non-extern declaration. Note this doesn't address import conflicts yet (i.e., two libraries define an equivalent name) because they don't call merge logic.

On merge, I'm only setting values when new_is_definition because I think it better matches the comment and resulting behavior. Note I now set them even for bad redefinitions; I think this matches the comment, and there's not a perfect choice here. I could change the flow back if preferred.

Note this puts a spotlight on invalid nodes on imported decls, which I think I'm going to need to address now. This isn't addressed by ImportRef logic directly because the Function's decl_id is a FunctionDecl, rather than the ImportRef that led to it. While I could add the ImportRef link to each decl, I think adjusting the associated NodeId is a better approach.


toolchain/check/function.cpp
```diff
@@ -191,60 +191,88 @@ auto CheckFunctionTypeMatches(Context& context,
...
+          FunctionNonExternRedecl, Error,
+          "Only one library can declare function {0} without `extern`.",
+          SemIR::NameId);
+      context.emitter()
+          .Build(node_id, FunctionNonExternRedecl, prev_function.name_id)
+          .Note(prev_function.decl_id, FunctionPreviousDecl)
+          .Emit();
+      return;
+    }
+  } else {
```
zygoloid, line 215: (Optional, no action needed.)

```suggestion
    }
    return;
  }
```

Could maybe reduce indentation here.

toolchain/check/function.cpp
```diff
@@ -191,60 +191,88 @@ auto CheckFunctionTypeMatches(Context& context,
                      context.functions().Get(prev_function_id), substitutions);
 }
 
+// Checks to see if a structurally valid redeclaration is allowed in context.
+// These all still merge.
+static auto CheckIsAllowedRedecl(Context& context, Parse::NodeId node_id,
+                                 SemIR::Function& new_function,
```
zygoloid, line 197: I expect we'll eventually want to factor this out and use the same implementation for classes, impls, and interfaces. No need to do that now, but I suppose we should start thinking about how to provide a way to access the common fields of `SemIR::Function`, `SemIR::Class`, etc.

toolchain/check/function.cpp
```diff
@@ -191,60 +191,88 @@ auto CheckFunctionTypeMatches(Context& context,
...
+  if (new_is_definition) {
+    // Track the signature from the definition, so that IDs in the body
+    // match IDs in the signature.
+    prev_function.definition_id = new_function.definition_id;
+    prev_function.implicit_param_refs_id = new_function.implicit_param_refs_id;
+    prev_function.param_refs_id = new_function.param_refs_id;
+    prev_function.return_type_id = new_function.return_type_id;
+    prev_function.return_slot_id = new_function.return_slot_id;
+  }
   return true;
```
zygoloid, line 281: Should we also set `prev_function.is_extern` to `new_function.is_extern` here? Once we've seen a local declaration, I think it should no longer matter whether a previous (eg, imported) declaration was `extern`.

toolchain/check/function.cpp
```diff
@@ -191,60 +191,88 @@ auto CheckFunctionTypeMatches(Context& context,
...
+                        "Redefinition of function {0}.", SemIR::NameId);
+      CARBON_DIAGNOSTIC(FunctionPreviousDefinition, Note,
+                        "Previously defined here.");
+      context.emitter()
+          .Build(node_id, FunctionRedefinition, prev_function.name_id)
+          .Note(prev_function.definition_id, FunctionPreviousDefinition)
+          .Emit();
+      return;
+    }
+    if (new_is_definition && prev_function.is_extern) {
```
zygoloid: I think we're missing a diagnostic for:
```
fn F();
extern fn F() {}
```
I wonder if perhaps we could split this up as:

1) Diagnose if `prev_function.is_extern != new_function.is_extern`.
2) Diagnose if `new_is_definition && new_function.is_extern`.

(I'm thinking of approaching it this way because, for `impl`s, we can have more than one non-defining declaration in a file (due to `match_first` blocks and `extend impl` declarations), so we'd be able to reach these checks for `impl`s when neither declaration is a definition.)

toolchain/check/function.cpp
jonmeow: I feel kind of odd about reducing the indent this way. It has an odd implication to me that imported versus non-imported is more significant; the `else` was intended to treat the cases as equal.

toolchain/check/function.cpp
jonmeow: Relevant test:
https://github.com/carbon-language/carbon-lang/blob/trunk/toolchain/check/testdata/function/definition/extern.carbon#L11-15

That's invalid due to the use of `extern` on a definition, unrelated to the merge. Merging assumes valid declarations as input. I don't think there's a code change here: do you specifically want this case tested, or is the point that the second line is already diagnosed sufficient?

Note, I'd expect similar elsewhere: given `extern` is only valid on forward declarations, and a definition might _not_ merge with a declaration, the `extern` definition needs to be diagnosed on its own.

toolchain/check/function.cpp
jonmeow: Yeah, agreed. In a draft commit I'm adding a general merge.h that I think will probably be where I start addressing things like this, but I'm not totally sure how generic we want to get here (to minimize fetches of data). This may be more a shared way of printing a diagnostic rather than accessing fields.

toolchain/check/function.cpp
jonmeow: Done, with a small test that the diagnostic doesn't print in extern.carbon.

jonmeow:
Note, depends on #3807

toolchain/check/function.cpp
zygoloid: Thanks, looks like we already perform my second check elsewhere. I would find this easier to reason about if we checked for `prev_function.is_extern != new_function.is_extern` here: the current check relies on conditions established elsewhere, which for me makes it harder to see why it's sufficient to correctly handle all cases. But I think this is correct as-is.

Perhaps a better question to ask is, for:
```carbon
extern fn F();
fn F() {}
```
... do we want a diagnostic that says "Cannot define `extern` function `F`.", or do we want a diagnostic that says "Redeclaring `extern` function `F` as non-`extern`."?

To me, the latter is better (though it's marginal) because it's describing a simpler problem -- declaration A is extern, declaration B is not -- which we know is invalid without considering whether declaration B is a definition or not.

But I think this is all a fairly weak argument. Feel free to keep the logic as-is; if you do, for consistency with the previous `if` it'd make sense to drop the tautological check of `new_is_definition`:
```suggestion
    if (prev_function.is_extern) {
```

toolchain/check/function.cpp
```diff
@@ -191,60 +191,91 @@ auto CheckFunctionTypeMatches(Context& context,
...
+    // match IDs in the signature.
+    prev_function.definition_id = new_function.definition_id;
+    prev_function.implicit_param_refs_id = new_function.implicit_param_refs_id;
+    prev_function.param_refs_id = new_function.param_refs_id;
+    prev_function.return_type_id = new_function.return_type_id;
+    prev_function.return_slot_id = new_function.return_slot_id;
+  }
+  if (!new_function.is_extern) {
+    prev_function.is_extern = false;
+  }
```
zygoloid, line 280: What happens for:
```carbon
package A library "foo" api;
fn F();
```
```carbon
package A library "bar" api;
// 1
import library "foo";
// 2
extern fn F();
// 3
fn F() {}
```
To my reading of the code:
At (1), we import a non-`extern` declaration of `F`. `is_extern` is `false`.
At (2), we process a valid redeclaration, and don't change `is_extern` here.
At (3), we process a non-`extern` definition, the previous declaration is not imported, and the function is not extern, so we allow the definition.

Perhaps:
```suggestion
  // The first non-imported declaration determines whether the function is extern in this file. For error recovery, if we see both an extern declaration and a non-extern declaration in the same file, consider the function to be non-extern.
  if (prev_is_import || !new_function.is_extern) {
    prev_function.is_extern = new_function.is_extern;
  }
```

zygoloid (approved):

toolchain/check/function.cpp
jonmeow: The suggested prev_is_import won't be the logic that's desired long-term, because you can also get cases where the previous is an import and the current case is *also* an import. You would be fixing one case, but I believe it introduces other bugs.

To be sure, there is a TODO at head for this, and this change leaves that TODO in place.

```
// TODO: Detect conflicting cross-file declarations, as well as uses of imported
// declarations followed by a redeclaration.
```

https://github.com/carbon-language/carbon-lang/blob/trunk/toolchain/check/function.cpp#L194-L195

jonmeow: Also, adding fail_mix_extern_decl.carbon to demonstrate behavior.

toolchain/check/function.cpp
jonmeow: Sure, dropping the tautological check, I'd missed that.

Rephrased the diagnostic per request, I agree that's better.

I'm wary of a check because we don't necessarily update the function to mark is_extern when is_extern would be invalid. But adding a comment.

toolchain/check/function.cpp
zygoloid: Has that been pushed yet? I can't find the test file in this PR.

jonmeow: It's a split-file in the final commit: https://github.com/carbon-language/carbon-lang/pull/3809#discussion_r1542046945

toolchain/check/testdata/function/definition/import.carbon
```diff
@@ -34,12 +43,26 @@ import library "a";
...
+library "mix_extern_decl" api;
+
+import library "a";
+
+extern fn D();
+// CHECK:STDERR: fail_mix_extern_decl.carbon:[[@LINE+4]]:1: ERROR: Only one library can declare function D without `extern`.
+// CHECK:STDERR: fn D() {}
+// CHECK:STDERR: ^~~~~~~~
+// CHECK:STDERR: fail_mix_extern_decl.carbon: Previously declared here.
+fn D() {}
```
jonmeow, line 64: @zygoloid Here

zygoloid (approved):
Thanks!

PR #3810, opened 2024-03-25
Replace NodeId with a hybrid LocationId in SemIR diagnostics.

jonmeow:
The purpose of this change is to allow something such as a FunctionDecl instruction to note an imported instruction as the "loc_id". Note that doesn't occur here: this change is already very sweeping in edits. There is no testdata affected, intended to show equivalent behavior.

We might want to consolidate NodeId references towards LocationId, but if that's preferred, I'd still like to split it out. A lot of this just piping through LocationId where it's a build error otherwise, enough that imports should be able to start using it for diagnostics. ValueStores are added but still unused -- just flushing out structure for review.

Restructuring SemIRLocation is necessary to use LocationId this way. For TokenOnly, it's not getting used in Parse, so I migrated it to Check and it's now specific to SemIRLocation.

I also considered making LocationId reference an InstId (which would need to be an ImportRef) instead of an ImportIRInstId. However, that would've required import.cpp to add instructions for decls which are reached during resolution -- we typically don't have an inst ready for use. An extra inst is essentially 16 bytes in InstId's ValueStore + 4 bytes in LocationId's ValueStore, whereas this is 8 bytes per.

### DIFF
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -40,16 +40,21 @@ class SemIRDiagnosticConverter : public DiagnosticConverter<SemIRLocation> {
       -> DiagnosticLocation override {
     // Parse nodes always refer to the current IR.
     if (!loc.is_inst_id) {
-      return ConvertLocationInFile(sem_ir_, loc.node_location, context_fn);
+      CARBON_CHECK(loc.loc_id.is_node_id() || !loc.loc_id.is_valid())
+          << "TODO: Handle non-NodeId locs";
+      return ConvertLocationInFile(sem_ir_, loc.loc_id.node_id(),
+                                   loc.token_only, context_fn);
     }
 
     const auto* cursor_ir = sem_ir_;
     auto cursor_inst_id = loc.inst_id;
     while (true) {
       // If the parse node is valid, use it for the location.
-      if (auto node_id = cursor_ir->insts().GetNodeId(cursor_inst_id);
-          node_id.is_valid()) {
-        return ConvertLocationInFile(cursor_ir, node_id, context_fn);
+      if (auto loc_id = cursor_ir->insts().GetLocationId(cursor_inst_id);
+          loc_id.is_valid()) {
+        CARBON_CHECK(loc_id.is_node_id()) << "TODO: Handle non-NodeId locs";
+        return ConvertLocationInFile(cursor_ir, loc_id.node_id(),
+                                     loc.token_only, context_fn);
       }
 
       // If the parse node was invalid, recurse through import references when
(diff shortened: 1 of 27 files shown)


jonmeow:
As I'm looking again at TryResolveImportRef, I'm changing my mind about using an InstId versus separate store. Converting to draft.

toolchain/check/check.cpp
```diff
@@ -21,7 +21,7 @@ namespace Carbon::Check {
 
 // Parse node handlers. Returns false for unrecoverable errors.
 #define CARBON_PARSE_NODE_KIND(Name) \
-  auto Handle##Name(Context& context, Parse::Name##Id node_id) -> bool;
+  auto Handle##Name(Context& context, Parse::Name##Id node_id)->bool;
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  auto Handle##Name(Context& context, Parse::Name##Id node_id) -> bool;
```

toolchain/sem_ir/inst.h
```diff
@@ -290,36 +290,46 @@ inline auto operator<<(llvm::raw_ostream& out, TypedInst inst)
 
 // Associates a NodeId and Inst in order to provide type-checking that the
```
geoffromer:
```suggestion
// Associates a NodeId and LocationId in order to provide type-checking that the
```

toolchain/sem_ir/ids.h
```diff
@@ -472,6 +475,64 @@ struct ElementIndex : public IndexBase, public Printable<ElementIndex> {
...
+// The ID of an ImportIRInst.
+struct ImportIRInstId : public IdBase, public Printable<InstId> {
+  using ValueType = ImportIRInst;
+
+  using IdBase::IdBase;
+};
+
+// A SemIR location used exclusively for diagnostic locations. If positive, it
+// corresponds to a Parse::NodeId in the current IR. If valid and negative, it
+// corresponds to an ImportIRInstId.
```
geoffromer: What about zero?

Also, the code isn't implemented in terms of "positive" and "negative", it's implemented in terms of comparisons with `InvalidIndex`. Might it be clearer to phrase the comment in those terms as well? Alternatively, maybe the mapping from integers to sub-kinds of location should be treated as an implementation detail, and not specified here at all?

geoffromer (commented):
(Porting some comments from my review of #3811)

toolchain/sem_ir/inst.h
jonmeow: Fixed (NodeId -> LocationId, not Inst)

toolchain/sem_ir/ids.h
jonmeow: Adjusted comments. I think the implementation is nuanced and should be commented somewhere; here seems best.

jonmeow:
Note, depends on #3809

zygoloid (approved):
Looks good! A couple of thoughts (no request for changes here):

- Would it make sense to abbreviate `Location` to `Loc`? It seems like it's going to come up a lot.
- I think we'll eventually need at least a third kind of location, for code generated by metaprogramming. We can presumably share the negative `LocationId` space between this and imports. But I'm mostly just happy that we'll have an abstraction in place to let us do that :)

jonmeow:
> Looks good! A couple of thoughts (no request for changes here):
> 
> * Would it make sense to abbreviate `Location` to `Loc`? It seems like it's going to come up a lot.

I think it'd be reasonable. I was tempted to, and obviously I'm already doing that in local variables, but we already had DiagnosticLocation so I wanted to split discussion of that.

> * I think we'll eventually need at least a third kind of location, for code generated by metaprogramming. We can presumably share the negative `LocationId` space between this and imports. But I'm mostly just happy that we'll have an abstraction in place to let us do that :)

Yeah, we can probably find another way to steal a bit for this. Note, depending on the setup, it might also be worth discussing the other thing of replacing the whole loc_ids_ array with an 8 byte structure that contains one of the three setups, particularly if 8 bytes is right for both metaprogramming and import refs, even if it's burning 4 bytes for node ids.

PR #3811, opened 2024-03-25
Provide locations for indirectly imported instructions.

jonmeow:
This replaces all invalid node IDs in import_ref.cpp with references to the imported instruction.

This splits out ReplaceInstBeforeConstantUse into a separate function when the LocationId is replaced, as for splicing. That's the less common case, whereas others would need to provide the LocationId in order just to not change the value.

### DIFF
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -37,12 +37,17 @@ class SemIRDiagnosticConverter : public DiagnosticConverter<SemIRLocation> {
       const SemIR::File* sem_ir)
       : node_converters_(node_converters), sem_ir_(sem_ir) {}
 
+  // Converts an instruction's location to a diagnostic location, which will be
+  // the underlying line of code. Adds context for any imports used in the
+  // current SemIR to get to the underlying code.
   auto ConvertLocation(SemIRLocation loc, ContextFnT context_fn) const
       -> DiagnosticLocation override {
+    // Cursors for the current IR and instruction in that IR.
     const auto* cursor_ir = sem_ir_;
     auto cursor_inst_id = SemIR::InstId::Invalid;
 
-    // Notes an import on the diagnostic and updates cursors.
+    // Notes an import on the diagnostic and updates cursors to point at the
+    // imported IR.
     auto follow_import_ref = [&](SemIR::ImportIRId ir_id,
                                  SemIR::InstId inst_id) {
       const auto& import_ir = cursor_ir->import_irs().Get(ir_id);
@@ -64,15 +69,16 @@ class SemIRDiagnosticConverter : public DiagnosticConverter<SemIRLocation> {
         follow_import_ref(import_ir_inst.ir_id, import_ir_inst.inst_id);
         return std::nullopt;
       } else {
+        // Parse nodes always refer to the current IR.
         return ConvertLocationInFile(cursor_ir, loc_id.node_id(),
                                      loc.token_only, context_fn);
       }
     };
 
+    // Handle the base location.
     if (loc.is_inst_id) {
       cursor_inst_id = loc.inst_id;
     } else {
-      // Parse nodes always refer to the current IR.
       if (auto diag_loc = handle_loc(loc.loc_id)) {
         return *diag_loc;
       }
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -40,8 +40,9 @@ class SemIRDiagnosticConverter : public DiagnosticConverter<SemIRLocation> {
   auto ConvertLocation(SemIRLocation loc, ContextFnT context_fn) const
       -> DiagnosticLocation override {
     const auto* cursor_ir = sem_ir_;
-    auto cursor_inst_id = loc.inst_id;
+    auto cursor_inst_id = SemIR::InstId::Invalid;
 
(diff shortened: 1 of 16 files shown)


toolchain/sem_ir/inst.h
```diff
@@ -373,31 +383,37 @@ class InstStore {
...
+    CARBON_CHECK(inst_id.index < (int)loc_ids_.size())
+        << inst_id.index << " " << loc_ids_.size();
+    return loc_ids_[inst_id.index];
   }
 
-  // Overwrites a given instruction and parse node with a new value.
-  auto Set(InstId inst_id, NodeIdAndInst node_id_and_inst) -> void {
-    values_.Get(inst_id) = node_id_and_inst.inst;
-    node_ids_[inst_id.index] = node_id_and_inst.node_id;
+  // Overwrites a given instruction and location ID with a new value.
```
geoffromer: Is the "and location ID" part still accurate?

toolchain/check/check.cpp
```diff
@@ -35,36 +36,55 @@ class SemIRDiagnosticConverter : public DiagnosticConverter<SemIRLocation> {
 
   auto ConvertLocation(SemIRLocation loc, ContextFnT context_fn) const
       -> DiagnosticLocation override {
+    const auto* cursor_ir = sem_ir_;
+    auto cursor_inst_id = loc.inst_id;
+
+    auto follow_import_ref = [&](SemIR::ImportIRId ir_id,
```
geoffromer, line 51: I think it would help me follow this code if these lambdas had function-style API comments

toolchain/check/check.cpp
```diff
@@ -35,36 +36,55 @@ class SemIRDiagnosticConverter : public DiagnosticConverter<SemIRLocation> {
 
   auto ConvertLocation(SemIRLocation loc, ContextFnT context_fn) const
       -> DiagnosticLocation override {
+    const auto* cursor_ir = sem_ir_;
+    auto cursor_inst_id = loc.inst_id;
```
geoffromer: What if this isn't the active member of the union?

toolchain/check/context.h
```diff
@@ -101,8 +101,15 @@ class Context {
   // evaluation, either because it's newly created and entirely unused, or
   // because it's only used in a position that constant evaluation ignores, such
   // as a return slot.
-  auto ReplaceInstBeforeConstantUse(SemIR::InstId inst_id,
-                                    SemIR::LocationIdAndInst loc_id_and_inst)
+  auto ReplaceLocationIdAndInstBeforeConstantUse(
+      SemIR::InstId inst_id, SemIR::LocationIdAndInst loc_id_and_inst) -> void;
+
+  // Replaces the value of the instruction `inst_id` with `inst`.
```
geoffromer: Is the value of an instruction ID an `Inst` (as this comment suggests) or a `LocationIdAndInst` (as the previous comment suggests)?

toolchain/sem_ir/inst.h
jonmeow: Fixed

toolchain/check/check.cpp
jonmeow: In that case the value wasn't used; changed the flow to make this more obvious.

toolchain/check/check.cpp
jonmeow: Added.

toolchain/check/context.h
jonmeow: The "value" is could mean either or both depending on context. I've dropped "the value" since it doesn't seem to help.

toolchain/check/check.cpp
geoffromer: That helps some, but what does "updates cursors" mean? I gather it's the next step of some sort of traversal, but I don't quite follow what the traversal is actually doing.

toolchain/check/check.cpp
```diff
@@ -63,11 +66,14 @@ class SemIRDiagnosticConverter : public DiagnosticConverter<SemIRLocation> {
       }
     };
 
-    // Parse nodes always refer to the current IR.
-    if (!loc.is_inst_id) {
+    if (loc.is_inst_id) {
+      cursor_inst_id = loc.inst_id;
+    } else {
+      // Parse nodes always refer to the current IR.
```
geoffromer: If so, do you need to use `handle_loc` here, or could you call `ConvertLocationInFile` directly?

I ask because omitting the `handle_loc` call here would enable a couple other changes:
- It looks like `handle_loc` is only called in one other place, and I think it would be clearer to inline it there, avoiding the indirection of a function call.
- I also think it would be clearer if this block were above the lambdas, since it partly deals with the initialization of `cursor_inst_id`, and the lambdas seem more related to the body of the loop below.

toolchain/check/check.cpp
jonmeow: It's following imports to find the originating code. I've added some more comments to try to clarify.

toolchain/check/check.cpp
jonmeow: Sorry, moved this comment to a more appropriate spot.

jonmeow:
Note, depends on #3810

geoffromer (approved):

PR #3815, opened 2024-03-26
Factor out some parts of builtin handling and add declaration checking.

zygoloid:
In preparation for adding more builtins, factor out the handling of builtin function kinds into separate files.

Add checking for builtin function signatures. The mechanism used here is intended to provide a lot of flexibility for declaring generic builtin functions and pretty arbitrary constraints on the types of parameters of builtin functions. For now, these constraints are checked when the builtin function is declared. The hope is that this will suffice, but if not, it should be straightforward to switch to doing some of the checking on call and share logic between the checks.


toolchain/sem_ir/builtin_function_kind.h
```diff
@@ -0,0 +1,50 @@
...
+#define CARBON_TOOLCHAIN_SEM_IR_BUILTIN_FUNCTION_KIND_H_
+
+#include <cstdint>
+
+#include "common/enum_base.h"
+
+namespace Carbon::SemIR {
+
+class File;
+struct InstId;
```
jonmeow: Can some of these be included instead of forward declared?

toolchain/sem_ir/builtin_function_kind.h
```diff
@@ -0,0 +1,50 @@
...
+struct TypeId;
+
+CARBON_DEFINE_RAW_ENUM_CLASS(BuiltinFunctionKind, std::uint8_t) {
+#define CARBON_SEM_IR_BUILTIN_FUNCTION_KIND(K) CARBON_RAW_ENUM_ENUMERATOR(K)
```
jonmeow: In other similar invocations, I think we use "Name". Single-letter parameter names are discouraged by style, and I think it's clearer than "K"

toolchain/sem_ir/builtin_function_kind.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+#include <utility>
+
+#include "toolchain/sem_ir/file.h"
+#include "toolchain/sem_ir/ids.h"
+#include "toolchain/sem_ir/typed_insts.h"
+
+namespace Carbon::SemIR {
+
+namespace BuiltinFunctionInfo {
```
jonmeow: Why not use an anonymous namespace, particularly given you have one on line 21?

If this is so that you can do `constexpr BuiltinInfo None = {"", nullptr};`, I wonder if there would be easier to understand ways to do this. For example, had you considered something like creating a std::array of BuiltinInfo objects, and indexing into that? It could be assigned by index at construction in order to avoid ambiguity about ordering, which I assume was one thing you were trying to avoid with the current approach.

toolchain/check/handle_function.cpp
```diff
@@ -318,17 +317,51 @@ static auto LookupBuiltinFunctionKind(Context& context,
   return kind;
 }
 
+static auto IsValidBuiltinDeclaration(Context& context,
```
jonmeow, line 322: Please you add a small comment on behavior.

toolchain/check/handle_function.cpp
```diff
@@ -318,17 +317,51 @@ static auto LookupBuiltinFunctionKind(Context& context,
...
+                                      const SemIR::Function& function,
+                                      SemIR::BuiltinFunctionKind builtin_kind)
+    -> bool {
+  // Form the list of parameter types for the declaration.
+  llvm::SmallVector<SemIR::TypeId> param_type_ids;
+  auto param_refs = context.inst_blocks().Get(function.param_refs_id);
+  param_type_ids.reserve(param_refs.size());
+  for (auto param_id : param_refs) {
+    // TODO: We also need to track whether the parameter is declaraed with
```
jonmeow:
```suggestion
    // TODO: We also need to track whether the parameter is declared with
```

toolchain/sem_ir/builtin_function_kind.cpp
```diff
@@ -0,0 +1,177 @@
...
+template <typename ParamT, typename TypeConstraint>
+struct Param {
+  static auto Check(const File& sem_ir, ValidateState& state) -> bool {
+    return TypeConstraint::Check(sem_ir, state, ParamT::Get(state));
+  }
+};
+
+// Constraint that a type is the first generic parameter of the builtin. See
+// ValidateSignature for details.
+struct T {
```
jonmeow: I'm really used to "T" being a template argument. It's not obvious to me what it's doing here, or why "T" would be an appropriate name outside a template parameter. Can you elaborate?

toolchain/sem_ir/builtin_function_kind.cpp
```diff
@@ -0,0 +1,177 @@
...
+struct BuiltinInfo {
+  llvm::StringLiteral name;
+  ValidateFn* validate;
+};
+
+// State used when validating a builtin signature that persists between
+// individual checks.
+struct ValidateState {
+  // The type `T`, if any.
+  TypeId t = TypeId::Invalid;
```
jonmeow: What is "t" short for? This doesn't seem to be an instance of "T". Note, I think the single character names are making it harder to read this code.

toolchain/sem_ir/builtin_function_kind.cpp
jonmeow: Maybe "AnyType" if that's appropriate?

toolchain/sem_ir/builtin_function_kind.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+// Descriptions of builtin functions follow. For each builtin, a corresponding
+// `BuiltinInfo` constant is declared describing properties of that builtin.
+
+// Not a builtin function.
+constexpr BuiltinInfo None = {"", nullptr};
+
+// "int.add": integer addition.
+constexpr BuiltinInfo IntAdd = {
+    "int.add", ValidateSignature<auto(T, T)->T, Param<T, AnyInt>>};
```
jonmeow: I'm confused by this, maybe you could explain directly what this example is doing in ValidateSignature's comment?

jonmeow (commented):
Note, it looks a little like you're implementing a type system in BuiltinFunctionKind. That will be flexible for sure, but I worry about the amount of complexity involved... What does this look like when you're trying to parameterize multiple types? Which builtins will need this complexity?

toolchain/check/handle_function.cpp
zygoloid: Done.

toolchain/sem_ir/builtin_function_kind.cpp
zygoloid: The purpose here is partly to limit the scope of names like `T`, and partly to give a scope to put the builtin definitions into.

If you'd prefer that we define the builtin info as an array with a lambda, we can do that, but I think it's going to be more verbose and less readable. The pattern I'm using here is intended to be comparable to how we define `InstKind`s and the like, with one variable definition for each enumerator.

toolchain/sem_ir/builtin_function_kind.cpp
zygoloid: Renamed and expanded comment.

toolchain/sem_ir/builtin_function_kind.cpp
zygoloid: Here, `T` is a generic parameter of a builtin signature. I've reworked this to try to be a little clearer.

toolchain/sem_ir/builtin_function_kind.cpp
zygoloid: I've expanded the comment to walk through a similar but slightly more interesting example.

zygoloid:
> Note, it looks a little like you're implementing a type system in BuiltinFunctionKind. That will be flexible for sure, but I worry about the amount of complexity involved... What does this look like when you're trying to parameterize multiple types? Which builtins will need this complexity?

With suitable typedefs, something like left-shift for an integer: `fn LeftShift[T:! AnyInt, U:! AnyInt](a: T, b: U) -> T` would be represented as `auto (IntT, IntU) -> IntT`. (With the previous formulation, this'd be `auto (T, AnyInt) -> T, Param<T, AnyInt>`, but I think the new approach is a bit cleaner.)

toolchain/sem_ir/builtin_function_kind.h
zygoloid: Done.

toolchain/sem_ir/builtin_function_kind.h
zygoloid: Done.

toolchain/sem_ir/builtin_function_kind.cpp
```diff
@@ -0,0 +1,179 @@
...
+// specifies that the builtin takes values of two integer types and returns a
+// value of a third integer type. Types used within the signature should provide
+// a `Check` function that validates that the Carbon type is expected:
+//
+//   auto Check(const File&, ValidateState&, TypeId) -> bool;
+//
+// To constrain that the same type is used in multiple places in the signature,
+// `TypeParam<I, T>` can be used. For example:
+//
+// `auto (TypeParam<0, AnyInt>, AnyInt) -> TypeParam<0, AnyInt>`
```
jonmeow: Consistency suggestion:

```suggestion
// a `Check` function that validates that the Carbon type is expected:
//
// `auto Check(const File&, ValidateState&, TypeId) -> bool;`
//
// To constrain that the same type is used in multiple places in the signature,
// `TypeParam<I, T>` can be used. For example:
//
// `auto (TypeParam<0, AnyInt>, AnyInt) -> TypeParam<0, AnyInt>`
```

or

```suggestion
// a `Check` function that validates that the Carbon type is expected:
//
//   auto Check(const File&, ValidateState&, TypeId) -> bool;
//
// To constrain that the same type is used in multiple places in the signature,
// `TypeParam<I, T>` can be used. For example:
//
//   auto (TypeParam<0, AnyInt>, AnyInt) -> TypeParam<0, AnyInt>
```

toolchain/sem_ir/builtin_function_kind.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+// Returns the builtin function kind with the given name, or None if the name
+// is unknown.
+auto BuiltinFunctionKind::ForBuiltinName(llvm::StringRef name)
+    -> BuiltinFunctionKind {
+  return llvm::StringSwitch<BuiltinFunctionKind>(name)
+#define CARBON_SEM_IR_BUILTIN_FUNCTION_KIND(Name) \
+  .Case(BuiltinFunctionInfo::Name.name, BuiltinFunctionKind::Name)
+#include "toolchain/sem_ir/builtin_function_kind.def"
+      .Default(BuiltinFunctionKind::None);
```
jonmeow:
```suggestion
#define CARBON_SEM_IR_BUILTIN_FUNCTION_KIND(Name) \
  if (name == BuiltinFunctionInfo::Name.name) { return BuiltinFunctionKind::Name; }
#include "toolchain/sem_ir/builtin_function_kind.def"
  return BuiltinFunctionKind::None;
```

How much does StringSwitch buy here? I think this may be similar, but with less to optimize. But maybe I've misread what's going on there.

toolchain/sem_ir/builtin_function_kind.cpp
```diff
@@ -0,0 +1,179 @@
...
+          sem_ir.insts().TryGetAs<SemIR::BoundMethod>(callee_id)) {
+    callee_id = bound_method->function_id;
+  }
+  callee_id = sem_ir.constant_values().Get(callee_id).inst_id();
+  if (!callee_id.is_valid()) {
+    return SemIR::BuiltinFunctionKind::None;
+  }
+  if (auto callee = sem_ir.insts().TryGetAs<SemIR::FunctionDecl>(callee_id)) {
+    const auto& function = sem_ir.functions().Get(callee->function_id);
+    return function.builtin_kind;
```
jonmeow:
```suggestion
    return sem_ir.functions().Get(callee->function_id).builtin_kind;
```

jonmeow (approved):
Looks good. Thanks for the changes, I do think it's easier to understand.

toolchain/sem_ir/builtin_function_kind.cpp
jonmeow: My one last comment here would be, and up to you, but you could take the individual names and put them in an array for a macro, then use the single array in place of both the StringSwitch and validate function. Even name() could be replaced with that, I think.

toolchain/sem_ir/builtin_function_kind.cpp
zygoloid: I think this is basically what `StringSwitch` does under the hood, so it's probably not buying us anything.

toolchain/sem_ir/builtin_function_kind.cpp
```diff
@@ -0,0 +1,177 @@
...
+  BuiltinFunctionInfo::Name.name,
+#include "toolchain/sem_ir/builtin_function_kind.def"
+};
+
+// Returns the builtin function kind with the given name, or None if the name
+// is unknown.
+auto BuiltinFunctionKind::ForBuiltinName(llvm::StringRef name)
+    -> BuiltinFunctionKind {
+#define CARBON_SEM_IR_BUILTIN_FUNCTION_KIND(Name) \
+  if (name == BuiltinFunctionInfo::Name.name) { return BuiltinFunctionKind::Name; }
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  if (name == BuiltinFunctionInfo::Name.name) {   \
    return BuiltinFunctionKind::Name;             \
  }
```

toolchain/sem_ir/builtin_function_kind.cpp
zygoloid: I've looked into this. `EnumBase` really wants its own names array, so we'd end up storing the names twice in that direction: once in the `BuiltinFunctionInfo` table and again in the names table. This is probably worth looking into again if we add another field into `BuiltinFunctionInfo`, but I'm inclined to not do this for now.

PR #3816, opened 2024-03-26
Add builtins for some basic integer operations.

zygoloid:
Supports unary `-`, and binary `+`, `-`, `*`, `/`, `%`, `==`, `!=`.


toolchain/check/eval.cpp
```diff
@@ -286,37 +286,174 @@ static auto PerformAggregateIndex(Context& context, SemIR::Inst inst)
...
+                                     SemIR::BuiltinFunctionKind builtin_kind,
+                                     SemIR::InstId arg_id)
+    -> SemIR::ConstantId {
+  CARBON_CHECK(builtin_kind == SemIR::BuiltinFunctionKind::IntNegate)
+      << "Unexpected builtin kind";
+
+  auto op = context.insts().GetAs<SemIR::IntLiteral>(arg_id);
+  auto op_val = context.ints().Get(op.int_id);
+
+  if (op_val.isMinSignedValue()) {
```
josh11b, line 307: Are we distinguishing literals (like `-2,147,483,648`) from constant values of a type with a range (like `-2,147,483,648 as i32`)?

toolchain/sem_ir/ids.h
```diff
@@ -265,6 +265,11 @@ struct BoolValue : public IdBase, public Printable<BoolValue> {
   static const BoolValue False;
   static const BoolValue True;
 
+  // Returns the `BoolValue` corresponding to `b`.
+  static constexpr auto FromBool(bool b) -> BoolValue {
```
jonmeow, line 269: Maybe just "From"?

jonmeow, line 269: Is this used as constexpr?
```suggestion
  static auto FromBool(bool b) -> BoolValue {
```

toolchain/sem_ir/builtin_function_kind.cpp
```diff
@@ -123,10 +136,38 @@ using IntT = TypeParam<0, AnyInt>;
...
 constexpr BuiltinInfo IntAdd = {"int.add",
                                 ValidateSignature<auto(IntT, IntT)->IntT>};
 
+// "int.sub": integer subtraction.
+constexpr BuiltinInfo IntSub = {"int.sub",
+                                ValidateSignature<auto(IntT, IntT)->IntT>};
+
+// "int.mul": integer multiplication.
+constexpr BuiltinInfo IntMul = {"int.mul",
+                                ValidateSignature<auto(IntT, IntT)->IntT>};
```
jonmeow, line 145: You might consider something like "BinaryIntToInt", "BinaryIntToBool", but this seems fine too.

toolchain/check/eval.cpp
```diff
@@ -286,37 +286,174 @@ static auto PerformAggregateIndex(Context& context, SemIR::Inst inst)
...
+      CARBON_FATAL() << "Unexpected operation kind.";
+  }
+
+  if (overflow) {
+    CARBON_DIAGNOSTIC(CompileTimeIntegerOverflow, Error,
+                      "Integer overflow in calculation {0} {1} {2}.",
+                      llvm::APSInt, llvm::StringLiteral, llvm::APSInt);
+    context.emitter().Emit(loc, CompileTimeIntegerOverflow,
+                           llvm::APSInt(lhs_val, false), op_str,
+                           llvm::APSInt(rhs_val, false));
```
jonmeow, line 377: You might consider adding a type conversion for the literal value to APSInt to the diagnostics, rather than writing this in each spot (similar to TypeId -> str). Partly I don't know why "false" (maybe comment with parameter name?)

toolchain/check/eval.cpp
```diff
@@ -286,37 +286,174 @@ static auto PerformAggregateIndex(Context& context, SemIR::Inst inst)
...
+                                      SemIR::InstId rhs_id)
+    -> SemIR::ConstantId {
+  auto lhs = context.insts().GetAs<SemIR::IntLiteral>(lhs_id);
+  auto rhs = context.insts().GetAs<SemIR::IntLiteral>(rhs_id);
+  auto lhs_val = context.ints().Get(lhs.int_id);
+  auto rhs_val = context.ints().Get(rhs.int_id);
+
+  bool overflow = false;
+  llvm::APInt result_val;
+  llvm::StringLiteral op_str = "<error>";
```
jonmeow, line 333: This value should never be used, right? Should it be assigned at all?

toolchain/check/eval.cpp
```diff
@@ -286,37 +286,174 @@ static auto PerformAggregateIndex(Context& context, SemIR::Inst inst)
...
+
+  bool result;
+  switch (builtin_kind) {
+    case SemIR::BuiltinFunctionKind::IntEq:
+      result = (lhs_val == rhs_val);
+      break;
+    case SemIR::BuiltinFunctionKind::IntNeq:
+      result = (lhs_val != rhs_val);
+      break;
+
```
```suggestion
```
Sorry, this just looks inconsistent to me with the two cases.

jonmeow (approved):
Looks good.

toolchain/check/eval.cpp
zygoloid: It should never be used, but `StringLiteral` doesn't have a default constructor.

toolchain/check/eval.cpp
zygoloid: I have an idea here, will send a follow-up PR.

toolchain/sem_ir/ids.h
zygoloid: Done.

zygoloid: I'm inclined to keep the `constexpr` -- it would be reasonable to use this from constant expressions even though we happen to not use it that way right now.

toolchain/check/testdata/array/fail_bound_negative.carbon
```diff
@@ -4,27 +4,37 @@
...
+// CHECK:STDERR: fail_bound_negative.carbon:[[@LINE+3]]:14: ERROR: Array bound of -1 is negative.
+// CHECK:STDERR: var a: [i32; Negate(1)];
+// CHECK:STDERR:              ^~~~~~~
+var a: [i32; Negate(1)];
 
 // CHECK:STDOUT: --- fail_bound_negative.carbon
 // CHECK:STDOUT:
 // CHECK:STDOUT: constants {
 // CHECK:STDOUT:   %.1: i32 = int_literal 1 [template]
+// CHECK:STDOUT:   %.2: i32 = int_literal 4294967295 [template]
```
josh11b, line 18: Looks like we don't handle negative integers here.

toolchain/check/eval.cpp
zygoloid: We haven't implemented literal semantics in the toolchain yet. For now, we only have `i32`. The big missing piece here would be support for generic `impl`s, so we can define a suitable `impl` for `Core.IntLiteral(N)` for all `N`.

toolchain/check/testdata/array/fail_bound_negative.carbon
zygoloid: Sent #3822 to fix this.

PR #3817, opened 2024-03-27
Simplify diagnostic conversion.

zygoloid:
Remove the enum of converted types, and use `llvm::any_cast` to determine the type instead.


jonmeow (approved):
Thanks, this is a nice improvement.

PR #3819, opened 2024-03-27
Implement merging of conflicts found during import.

jonmeow:
Note we only identify conflicts between libraries in the current package during import.

This restructures the BUILD because of dependency cycles between cpp files... We're going to need context's name lookup to handle things such as merging, merging requires function logic, function logic requires context access. Per discussion, going with a single large cc_library for now rather than trying to split out small libraries.

I'm envisioning the new merge.* as a hub for cross-declaration merge logic. Note function.cpp is already pretty sizable, and I think it may lean a little function-specific even if there are some utilities that could be split out.

### DIFF
--- a/toolchain/check/BUILD
+++ b/toolchain/check/BUILD
@@ -54,10 +54,13 @@ cc_library(
         "convert.cpp",
         "decl_name_stack.cpp",
         "eval.cpp",
+        "function.cpp",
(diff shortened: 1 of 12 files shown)


jonmeow:
Depends on #3811

zygoloid (approved):

PR #3820, opened 2024-03-27
Add CARBON_KIND_SWITCH to better handle typed inst switches.

jonmeow:
Converts a switch in formatter.cpp as an example of resulting syntax.

### DIFF
--- a/toolchain/base/kind_switch.h
+++ b/toolchain/base/kind_switch.h
@@ -59,9 +59,9 @@ auto Cast(ValueT&& kind_switch_value) -> auto {
   return kind_switch_value.template As<CaseT>();
 }
 
-#define CARBON_KIND_MERGE_(Prefix, Line) Prefix##Line
-#define CARBON_KIND_LABEL_(Line) \
-  CARBON_KIND_MERGE_(carbon_internal_kind_case_, Line)
+#define CARBON_INTERNAL_KIND_MERGE(Prefix, Line) Prefix##Line
+#define CARBON_INTERNAL_KIND_LABEL(Line) \
+  CARBON_INTERNAL_KIND_MERGE(carbon_internal_kind_case_, Line)
 
 }  // namespace Carbon::Internal::Kind
 
@@ -88,7 +88,7 @@ auto Cast(ValueT&& kind_switch_value) -> auto {
                 decltype([]([[maybe_unused]] typed_variable_declaration) {})>( \
                 carbon_internal_kind_switch_value);                            \
             false) {}                                                          \
-  else [[maybe_unused]] CARBON_KIND_LABEL_(__LINE__)
+  else [[maybe_unused]] CARBON_INTERNAL_KIND_LABEL(__LINE__)
 // NOLINTEND(bugprone-macro-parentheses)
 
 #endif  // CARBON_TOOLCHAIN_BASE_KIND_SWITCH_H_
--- a/toolchain/base/kind_switch.h
+++ b/toolchain/base/kind_switch.h
@@ -1,0 +1,90 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#ifndef CARBON_TOOLCHAIN_BASE_KIND_SWITCH_H_
+#define CARBON_TOOLCHAIN_BASE_KIND_SWITCH_H_
+
+#include "llvm/ADT/STLExtras.h"
+
+// This library provides switch-like behaviors for Carbon's kind-based types.
+//
+// An expected use case is to mix regular switch `case` statements and
+// `CARBON_KIND`. However, the `switch` must be defined using
+// `CARBON_KIND_SWITCH`. For example:
+//
+//   CARBON_KIND_SWITCH(untyped_inst) {
+//     case CARBON_KIND(SomeInstType inst): {
+//       return inst.typed_field;
+//     }
+//     case OtherType1::Kind:
+//     case OtherType2::Kind:
+//       return value;
+//     default:
+//       return default_value;
+//   }
+//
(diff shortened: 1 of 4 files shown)


toolchain/base/kind_switch.h
```diff
@@ -0,0 +1,92 @@
...
+//   `CaseT::Kind`, which is passed to the `case` keyword.
+//
+// Note, this is currently used primarily for Inst in toolchain. When more
+// use-cases are added, it would be worth considering whether the API
+// requirements should change.
+namespace Carbon::Internal {
+
+// Given `CARBON_KIND_SWITCH(value)` this handles calling `value.kind()`.
+template <typename T>
+auto KindSwitch(T&& switch_value) -> auto {
```
chandlerc: Maybe, since this doesn't do the switch, `ValueKind`?

toolchain/base/kind_switch.h
```diff
@@ -0,0 +1,92 @@
...
+  return switch_value.kind();
+};
+
+// Given `CARBON_KIND(CaseT name)` this generates `CaseT::Kind`.
+template <typename FnT, typename KindT>
+constexpr auto KindCaseValue() -> KindT {
```
chandlerc: Does our Clang version let us write this?

```suggestion
consteval auto KindCaseValue() -> KindT {
```

toolchain/base/kind_switch.h
```diff
@@ -0,0 +1,92 @@
...
+//
+// - The type passed to `CARBON_KIND_SWITCH` has `.kind()` to switch on, and
+//   `.As<CaseT>` for `CARBON_KIND` to cast to.
+// - Each type passed to `CARBON_KIND` (`CaseT` above) provides
+//   `CaseT::Kind`, which is passed to the `case` keyword.
+//
+// Note, this is currently used primarily for Inst in toolchain. When more
+// use-cases are added, it would be worth considering whether the API
+// requirements should change.
+namespace Carbon::Internal {
```
chandlerc: How about adding a namespace here to avoid repeating it below? (To be clear, I don't really mind the repeatiton, but we don't have a good way to separate the part of the name that is really just a prefix vs. signifying the behavior, and there is ambiguity with the word `Kind` for me.)

```suggestion
namespace Carbon::Internal::Kind {
```

toolchain/base/kind_switch.h
```diff
@@ -0,0 +1,92 @@
...
+  return ArgT::Kind;
+}
+
+// Given `CARBON_KIND_SWITCH(value)` and `CARBON_KIND(CaseT name)` this
+// generates `value.As<CaseT>()`.
+template <typename FnT, typename ValueT>
+auto KindCaseAs(ValueT&& kind_switch_value) -> auto {
+  using CaseT = llvm::function_traits<FnT>::template arg_t<0>;
+  return kind_switch_value.template As<CaseT>();
+}
```
chandlerc: Unlike in the `switch` condition variable, I think here you can do something simpler, and its a bit cleaner:

```suggestion
auto KindCaseAs(const ValueT& kind_switch_value) -> auto {
  using CaseT = llvm::function_traits<FnT>::template arg_t<0>;
  return kind_switch_value.template As<CaseT>();
}
```

We shouldn't be modifying the incoming value, and we can't really move out of it either, so this should be correct to just be a `const &` input parameter (I think).

toolchain/base/kind_switch.h
```diff
@@ -0,0 +1,92 @@
...
+
+// Produces a case-compatible block of code that also instantiates a local typed
+// variable.
+//
+// This uses `if` to scope the variable, and provides a dangling `else` in order
+// to prevent accidental `else` use. The label allows `:` to follow the macro
+// name, making it look more like a typical `case`.
+//
+// NOLINTBEGIN(bugprone-macro-parentheses)
+#define CARBON_KIND(type_and_name)                                \
```
chandlerc: Maybe a slightly different name of:

```suggestion
#define CARBON_KIND(typed_variable_declaration)                                \
```

To make it clear that this should be structured like a variable declaration?

toolchain/base/kind_switch.h
chandlerc: Is `KindT` needed? Or rather, I assume it is -- mostly curious what goes wrong when you try to deduce the return type?

chandlerc (commented):
Mostly cosmetic stuff / nit picks. This is really fun, and the code quality improvements when using it are kind of amazing (in a good way).

toolchain/base/kind_switch.h
jonmeow: Done

toolchain/base/kind_switch.h
jonmeow: We already have `InstValueKind`, so this feels odd, particularly with your request above making this `Internal::Kind::ValueKind`.

How about `SwitchOn`? In fact, how about `Kind::SwitchOn`, `Kind::ForCase`, `Kind::Cast`?

toolchain/base/kind_switch.h
jonmeow: Done, compiles.

jonmeow: If the return type is not an enum (e.g., if it's an inst Definition type) it tries to copy. Casting to the case type seemed best.

toolchain/base/kind_switch.h
jonmeow: Done

toolchain/base/kind_switch.h
jonmeow: Done

toolchain/base/kind_switch.h
chandlerc: SG

toolchain/base/kind_switch.h
chandlerc: (as discussed live, we found a way to make this work)

chandlerc (approved):
Ship it!

PR #3821, opened 2024-03-27
Improve conversion to / from `BoolValue`.

zygoloid:
Rename `BoolValue::FromBool` to `BoolValue::From` as requested in #3816.

Add `BoolValue::ToBool`.

Convert existing code to use these where appropriate.


zygoloid:
@jonmeow Sorry, looks like I forgot to push the rename from `FromBool` to `From` before merging #3816. Here it is.

jonmeow (approved):
FromBool seems like a better name once you have ToBool, but do what you prefer here.

You could also just have `value()` as the thing that returns a `bool`, then `From` would seem fine.

Issue #3714, opened 2024-02-22
need name shadowing rules for `Self`


zygoloid:
Given

```carbon
class A {
  impl i32 as AddWith(Self) {
    // ... Self ...
  }
}
```

... does `Self` inside the `impl` refer to `A`, to `i32`, or is it invalid to use it due to ambiguity? Same question applies for `Self` on the right-hand side of the `as`. Explorer treats `Self` here as being `A` -- an `impl` only introduces a `Self` if there's not already a `Self` in scope.

The same question arises for nested classes:

```carbon
class A {
  class B {
    // ... Self ...
  }
}
```

I think I would expect that `Self` is `B` within `class B`, rather than naming `A` or being ambiguous. Especially given that a lookup  for `Self` within the body of `B` should presumably behave the same way if `B` is defined out of line:

```carbon
class A {
  class B;
}
class A.B {
  // ... Self ...
}
```

... and it would be very surprising if `Self` here didn't work.

Presumably `interface`s can be nested within `class`es, where the same question would arise:

```carbon
class A {
  interface B {
    // ... Self ...
  }
}
```

Is `Self` now the type implementing the interface, or is it `A`? I think in this case it *must* be the type implementing the interface, because there would otherwise be no way to name that type.

Some options to consider:

1) A `class`, `interface`, or `impl` declaration always introduces `Self`, as the class type or implementing type. `Self` refers to the innermost such declaration.
2) A `class` or `interface` always introduces `Self`. An `impl` only introduces `Self` if there's no `Self` in scope.
3) An `impl` doesn't introduce `Self` ever.

josh11b:
Another option is that an `impl` _declaration_ doesn't introduce `Self`, but `Self` is introduced in the _body_ of an `impl` definition.

I am not excited about option 3, since I think `Self` will shorten long type names in many circumstances, and no `Self` would make it hard to copy from the `interface` definition into the `impl` body.

zygoloid:
We have some [examples in explorer's prelude](https://github.com/carbon-language/carbon-lang/blob/eed33c3b8cfd81844cc5e1ebc746408fbb8dff64/explorer/data/prelude.carbon#L108) that would be broken by `Self` coming into scope at the `{` rather than the `as`, for example:

```carbon
impl String as EqWith(Self) {
```

[...]

PR #3824, opened 2024-03-27
Allow simple uses of `Self` in `impl`s.

zygoloid:
Per #3714, some of the details here are not yet settled. In particular, we might want `Self` to come into scope at the start of the definition, not at the `as` keyword. However, this change allows us to accept the uncontroversial examples.


jonmeow (approved):
Looks good!

PR #3825, opened 2024-03-27
Disable bugprone-macro-parentheses and let clang-format insert braces.

jonmeow:
- [bugprone-macro-parentheses](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/macro-parentheses.html) -- this is just a false positive issue, I don't think it's helping us catch bugs.
- [InsertBraces](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#insertbraces) -- although there's a warning about this creating issues due to incomplete semantic information, it seems to be happy with our code, and allows clang-format to fix something that clang-tidy would otherwise warn about.


chandlerc (approved):
LG

PR #3826, opened 2024-03-28
location -> loc abbreviation

jonmeow:


toolchain/diagnostics/diagnostic_kind.def
```diff
@@ -127,7 +127,7 @@ CARBON_DIAGNOSTIC_KIND(ExpectedAliasInitializer)
 
 CARBON_DIAGNOSTIC_KIND(SemanticsTodo)
 
-// Location context.
+// Loc context.
```
geoffromer: This seems more like a use of the English word "location" rather than a type name, so it seems like it shouldn't be abbreviated.

toolchain/parse/handle_decl_name_and_params.cpp
```diff
@@ -25,13 +25,13 @@ static auto HandleDeclNameAndParams(Context& context, State after_name)
     CARBON_DIAGNOSTIC(ExpectedDeclName, Error,
                       "`{0}` introducer should be followed by a name.",
                       Lex::TokenKind);
-    Lex::TokenIndex location = *context.position();
-    if (context.tokens().GetKind(location) == Lex::TokenKind::FileEnd) {
+    Lex::TokenIndex loc = *context.position();
```
geoffromer: I wonder if this should be renamed to `position` instead, since neither the accessor nor the type name mentions "location". I bring this up here because changing to the abbreviation seems to emphasize (inaccurately, I think) that this is a "location" in the toolchain sense of the word, not just a coincidental use of the English word "location".

jonmeow: Went with "token" since we often use that in this sort of context (note it can be reassigned from state.token)

toolchain/diagnostics/diagnostic_kind.def
jonmeow: Done

geoffromer (approved):
I'll leave merging to you since I don't want to interfere with your other in-flight changes.

PR #3827, opened 2024-03-28
Start attempting ambiguous import merges.

jonmeow:
Just connecting this together. Note the diagnostics still need to be cleaned up/improved, although this is correctly handling some cases.

This exposed an incorrect parameter passed in DiagnosticEmitter, which is fixed here -- context_fn shouldn't include args. Note this was caught by the CHECK, I'm just expanding on the message since I used the detail to help figure out the bug.


chandlerc (approved):
Makes sense!

PR #3829, opened 2024-03-28
Address fuzzer-caught error introduced in #3768

CJ-Johnson:
The fuzzer identified that tuple literals following member accesses and pointer member accesses caused the compiler to crash. This PR fixes that by calling PeekIsName() before assuming a name follows the access operator.


toolchain/check/testdata/pointer/fail_arrow.carbon
```diff
@@ -0,0 +1,12 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+
+fn Foo(x: i32) {
+  x.();
```
josh11b, line 8: Move to toolchain/parse/testdata instead of toolchain/check/testdata?

josh11b (commented):
I don't think these changes are affecting the problematic input. It was failing in parse validation before getting to the check stage.

josh11b (commented):
I think this change is now subsumed by https://github.com/carbon-language/carbon-lang/pull/3830

PR #3830, opened 2024-03-28
Prevent tuples after `.` and `->`

josh11b:


toolchain/parse/handle_paren_expr.cpp
```diff
@@ -6,6 +6,38 @@
...
+    if (!state.has_error) {
+      CARBON_DIAGNOSTIC(UnexpectedTokenInCompoundMemberAccess, Error,
+                        "Expected `)`.");
+      context.emitter().Emit(*context.position(),
+                             UnexpectedTokenInCompoundMemberAccess);
+      context.ReturnErrorOnState();
+    }
+
+    // Recover from the invalid token.
+    auto end_of_element = context.FindNextOf({Lex::TokenKind::CloseParen});
```
jonmeow: Please add a test for if there are multiple close parens and the first one is the wrong one, such as: `.((()))`

See Context::SkipMatchingGroup() for an example of how to handle this, I think you want something like `context.SkipTo(context.tokens().GetMatchedClosingToken(state.token));`

jonmeow: Hmm, maybe that's not the best example for a bug. Perhaps something like: `.(x x x ())`

toolchain/parse/handle_paren_expr.cpp
```diff
@@ -6,6 +6,38 @@
...
 namespace Carbon::Parse {
 
+auto HandleOnlyParenExpr(Context& context) -> void {
+  auto state = context.PopState();
+
+  // Advance past the open paren.
+  context.AddLeafNode(NodeKind::ExprOpenParen,
+                      context.ConsumeChecked(Lex::TokenKind::OpenParen));
+
+  context.PushState(state, State::ParenExprFinish);
```
jonmeow: WDYT of moving the shared logic out to a helper function instead of pushing a state? I was looking at the SkipTo in OnlyParenExprFinish and it took me a moment to figure out where the position was being advanced past the close paren, which might've been more apparent with a helper function. Since there isn't intermediate processing, I don't think the state stack is strictly required.

toolchain/parse/handle_paren_expr.cpp
josh11b: Done. The differences are not visible since the output doesn't indicate which `)` token is being referenced.

toolchain/parse/handle_paren_expr.cpp
josh11b: The helper function is only one line, which makes it a bit suspect. But making that change ended up with a better tree, I think, so I kept the change.

jonmeow (approved):
Thanks, looks good!

toolchain/parse/handle_paren_expr.cpp
jonmeow: Yeah, I was expecting it to only be visible in the old path because I'd expect processing to resume at the second `)`, yielding a parse error because `)` shouldn't be encountered like that. Skipping to it directly means that parse error is avoided.

PR #3831, opened 2024-03-28
Change ImportRef to a triple state: Unloaded, Loaded, Used

jonmeow:
This doesn't significantly change logic, although I'm trying to add the location to used state.

The issue I'm trying to address is how to identify a declaration as "allowed to be redeclared". Consider:

```
library "a" api;
extern fn F();
```
```
library "b" api;
extern fn F();
```
```
library "c" api;
import library "a";
import library "b";
var x: auto = F();
fn F();
```

What currently happens is:

1. On import of "a", `F` becomes ImportRefUnused
2. On import of "b", `F` becomes ImportRefUsed in order to merge.
3. In "c", the call `F()` doesn't change the state.
4. In "c", the declaration `fn F();` needs some breadcrumb to understand whether "F" has been referenced, as in step (3) here.

What I want to happen is:

1. On import of "a", `F` becomes ImportRefUnloaded
2. On import of "b", `F` becomes ImportRefLoaded in order to merge.
3. In "c", the call `F()` causes `F` to become ImportRefUsed
4. In "c", the declaration `fn F();` detects that `F` is already ImportRefUsed, and can use the associated `used_id` for a diagnostic about why redeclaring is invalid.

Note this PR isn't implementing (4). I'm focused on the refactoring to add a new ImportRef state here.


jonmeow:
Depends on #3827

chandlerc (approved):
LG (once the underlying PR is merged). The whole flow makes sense, and I like just talking about "loaded" when all we're doing is merging and not creating a constructive use, so the whole model makes lots of sense.

PR #3832, opened 2024-03-29
Defer parsing of method bodies until the end of a suitable enclosing scope.

zygoloid:
In parse, form a list of methods that are defined inline, tracking where they start, where they end, and which other inline methods are nested within them.

In check, when we reach an inline method body, skip it and add it to a worklist to be processed later. We also track when we reach the start and end of a context in which inline method bodies are deferred, so that we know when to replay the bodies.

When suspending a function definition to be processed later, the `DeclNameStack` entry is moved to separate storage, including popping the corresponding scopes from the scope stack and removing the corresponding lexical names from lexical lookup. Later, when we return to the function and parse its definition, the `DeclNameStack` entry is restored. The same is done when we reach the end of a nested context that can have inline methods, so that we can reenter the nested scope before processing its members.


common/variant_helpers.h
```diff
@@ -0,0 +1,42 @@
...
+//
+// ```
+// auto overloaded = Overload{[] (int) {}, [] (float) {}};
+// ```
+template <typename... Fs>
+struct Overload : Fs... {
+  using Fs::operator()...;
+};
+template <typename... Fs>
+Overload(Fs...) -> Overload<Fs...>;
```
jonmeow, line 27: https://en.cppreference.com/w/cpp/utility/variant/visit says this isn't needed as of c++20, does the compiler implement this feature?

And also, that this is a common enough pattern to be on cppreference but not stl...

```suggestion
```

toolchain/check/check.cpp
```diff
@@ -231,14 +250,360 @@ static auto InitPackageScopeAndImports(Context& context, UnitInfo& unit_info)
...
+  // Set the specified inline method index as being the next one to parse.
+  auto SkipTo(Parse::InlineMethodIndex next_index) -> void {
+    index = next_index;
+    if (static_cast<std::size_t>(index.index) ==
+        tree->inline_methods().size()) {
+      start_id = Parse::NodeId::Invalid;
+    } else {
+      start_id = tree->inline_methods().Get(index).start_id;
+    }
+  }
```
jonmeow: This feels like more class-y logic... for example, should `tree` be private rather than public?

toolchain/parse/context.cpp
```diff
@@ -427,6 +430,46 @@ auto Context::EmitExpectedDeclSemiOrDefinition(Lex::TokenKind expected_kind)
   emitter().Emit(*position(), ExpectedDeclSemiOrDefinition, expected_kind);
 }
 
+// Returns whether we are currently parsing in a context in which methods can be
+// declared, such as a class or interface.
+static auto ParsingInMethodContext(Context& context) -> bool {
```
jonmeow: To be sure I'm understanding correctly, in [the classes design](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/classes.md#methods), we define "method" to mean -- roughly -- a member function associated with an instance:

> Method declarations are distinguished from class function declarations by having a `self` parameter in square brackets `[`...`]` before the explicit parameter list in parens `(`...`)`.

Shouldn't class functions be processed out-of-line in addition to methods? I think that may even be how this code works right now, so maybe there's only a terminology mismatch?

WDYT of generally replacing "method" with "member" or "member function" (depending on use)?

toolchain/parse/context.h
```diff
@@ -357,6 +366,7 @@ class Context {
   Lex::TokenIterator end_;
 
   llvm::SmallVector<StateStackEntry> state_stack_;
+  llvm::SmallVector<InlineMethodIndex> enclosing_inline_method_stack_;
```
jonmeow: Should this have comments explaining its semantics?

toolchain/check/lexical_lookup.h
```diff
@@ -44,7 +52,26 @@ class LexicalLookup {
...
+
+  // Restore a previously-suspended lookup result.
+  auto Restore(SuspendedResult sus, ScopeIndex index) -> void {
+    lookup_[sus.index].push_back({sus.inst_id, index});
+  }
+
  private:
+  // Get the index at which the specified name is stored in `lookup_`.
+  auto GetLookupIndex(SemIR::NameId name_id) -> size_t {
+    return name_id.index + SemIR::NameId::NonIndexValueCount;
```
jonmeow:
```suggestion
  auto GetLookupIndex(SemIR::NameId name_id) -> int32_t {
    return name_id.index + SemIR::NameId::NonIndexValueCount;
```

Isn't name_id.index an int32_t? Why does this use size_t?

toolchain/check/lexical_lookup.h
```diff
@@ -28,12 +28,20 @@ class LexicalLookup {
     ScopeIndex scope_index;
   };
 
+  // A lookup result that has been temporarily removed from scope.
+  struct SuspendedResult {
+    // The lookup index.
+    size_t index;
```
jonmeow:
```suggestion
    int32_t index;
```

Per below common on GetLookupIndex

toolchain/check/check.cpp
```diff
@@ -25,6 +30,20 @@ namespace Carbon::Check {
...
+// HandleFunctionDefinitionStart, except that the various context stacks are
+// cleared out in between.
+auto HandleFunctionDefinitionSuspend(Context& context,
+                                     Parse::FunctionDefinitionStartId node_id)
+    -> SuspendedFunction;
+
+// Handle resuming the definition of a function, after a previous suspension.
+auto HandleFunctionDefinitionResume(Context& context,
+                                    Parse::FunctionDefinitionStartId node_id,
+                                    SuspendedFunction sus_fn) -> void;
```
jonmeow: Had you considered other places to declare these versus handle_function.cpp, perhaps `function.*` or (since I assume there'll be more suspend-related functions) a new `suspend.*`? Looking at dependencies, maybe helpers should move to `function.*`?

It's just the forward declaration that looks odd to me here because it's defined in a separate `.cpp` file. Maybe we should move the #define'd decls to a handle.h on the same basis (maybe it used to be in context.h?). I assume we'll have more of these Suspend/Resume functions as more types are added, and I'd be wary of accumulating more forward declarations here as a consequence.

toolchain/check/check.cpp
```diff
@@ -231,14 +250,360 @@ static auto InitPackageScopeAndImports(Context& context, UnitInfo& unit_info)
...
+      return true;
+    default:
+      return false;
+  }
+}
+
+namespace {
+// A worklist of pending tasks to perform to parse skipped inline method
+// bodies.
+struct InlineMethodWorklist {
```
jonmeow: Should this and NodeIdTraversal be classes? [Style says](https://google.github.io/styleguide/cppguide.html#Structs_vs._Classes) "Use a struct only for passive objects that carry data; everything else is a class." With the amount of methods on this, is this really passive?

Note too, with the length, I was kind of wondering if it'd be helpful to move longer function definitions out-of-line in order to more easily skim the API.

toolchain/check/check.cpp
```diff
@@ -231,14 +250,360 @@ static auto InitPackageScopeAndImports(Context& context, UnitInfo& unit_info)
...
+  // A worklist task that indicates we should parse a skipped method.
+  struct ParseSkippedMethod {
+    // The method that we skipped.
+    Parse::InlineMethodIndex method_index;
+    // The suspended function. Only present when `method_index` is valid.
+    std::optional<SuspendedFunction> sus_fn;
```
jonmeow: Had you considered options other than "sus" as an abbreviation, because of the word "sus"'s meaning? As an abbreviation, "susp" is at least a little better.

Going a different path, "Delayed"/"Deferred" instead of "Suspended"? `delayed_fn` might be short enough to type, although `del_fn` and `def_fn` might be misinterpreted abbreviations. Maybe something more like `relocate`/`relo`? (note, "relocate" might actually be a good fit semantically)

toolchain/check/check.cpp
```diff
@@ -231,14 +250,360 @@ static auto InitPackageScopeAndImports(Context& context, UnitInfo& unit_info)
...
+  struct LeaveMethodScope {
+    // Whether this scope is itself within a method scope.
+    bool in_method_scope;
+  };
+
+  // A pending parsing task.
+  using Task =
+      std::variant<ParseSkippedMethod, EnterMethodScope, LeaveMethodScope>;
+
+  InlineMethodWorklist() { worklist.reserve(64); }
```
jonmeow: It might be worth noting the reason for reserving is noted on the member; that's not where I would typically look.

toolchain/check/check.cpp
```diff
@@ -231,14 +250,360 @@ static auto InitPackageScopeAndImports(Context& context, UnitInfo& unit_info)
...
+
+    default:
+      return false;
+  }
+}
+
+// Determines whether this node kind is the start of a delayed method scope.
+static auto IsStartOfDelayedMethodScope(Parse::NodeKind kind) -> bool {
+  switch (kind) {
+    case Parse::NodeKind::ClassDefinitionStart:
```
jonmeow, line 352: Had you considered using `NodeIdOneOf` for this and the `End` function?

toolchain/check/check.cpp
```diff
@@ -231,14 +250,360 @@ static auto InitPackageScopeAndImports(Context& context, UnitInfo& unit_info)
...
+  // Push a task to re-enter a method scope, so that functions defined within it
+  // are parsed in the right context.
+  auto PushEnterMethodScope(Context& context) -> void {
+    method_scopes.push_back(worklist.size());
+    worklist.push_back(
+        EnterMethodScope{std::nullopt, IsInDelayedMethodScope(context)});
+  }
+
+  // Suspend the current method scope, which is finished but still on the
+  // decl_names_stack, and push a task to leave the scope when we're parsing
```
```suggestion
  // decl_name_stack, and push a task to leave the scope when we're parsing
```

toolchain/check/check.cpp
```diff
@@ -231,14 +250,360 @@ static auto InitPackageScopeAndImports(Context& context, UnitInfo& unit_info)
...
+  // Performs any processing necessary after handling a node.
+  auto AfterHandle(Parse::NodeKind parse_kind) -> void {
+    // When we reach the end of a delayed method scope, add a task to the
+    // worklist to leave the scope. If this is not a nested scope, start parsing
+    // the skipped methods now.
+    if (IsEndOfDelayedMethodScope(parse_kind)) {
+      if (worklist.SuspendFinishedScopeAndPush(context)) {
+        chunks.back().parsing_skipped_methods = true;
+      }
+    }
```
```suggestion
    if (IsEndOfDelayedMethodScope(parse_kind) &&
        worklist.SuspendFinishedScopeAndPush(context)) {
      chunks.back().parsing_skipped_methods = true;
    }
```

Mild preference for a single `if` where possible.

toolchain/check/check.cpp
```diff
@@ -231,14 +250,360 @@ static auto InitPackageScopeAndImports(Context& context, UnitInfo& unit_info)
...
+
+  // Indexes in `worklist` of method scopes that are currently still open.
+  llvm::SmallVector<size_t> method_scopes;
+};
+}  // namespace
+
+namespace {
+// A traversal of the node IDs in the parse tree, in the order in which we need
+// to check them.
+struct NodeIdTraversal {
```
jonmeow: Had you considered moving ProcessNodeIds into this too?

toolchain/parse/context.cpp
```diff
@@ -427,6 +430,46 @@ auto Context::EmitExpectedDeclSemiOrDefinition(Lex::TokenKind expected_kind)
...
+         state == State::DeclDefinitionFinishAsImpl ||
+         state == State::DeclDefinitionFinishAsInterface ||
+         state == State::DeclDefinitionFinishAsNamedConstraint;
+}
+
+auto Context::AddFunctionDefinitionStart(Lex::TokenIndex token,
+                                         int subtree_start, bool has_error)
+    -> void {
+  if (ParsingInMethodContext(*this)) {
+    enclosing_inline_method_stack_.push_back(tree_->inline_methods_.Add(
```
jonmeow: I wonder... beyond just having the parse tree track members to out-of-line, could the parse tree produce an ordering of parse nodes as they should be analyzed, e.g. as the chunks? Not blocking this PR; just generally wondering if you think that might be viable/worthwhile.

toolchain/check/scope_stack.cpp
```diff
@@ -142,4 +139,34 @@ auto ScopeStack::SetReturnedVarOrGetExisting(SemIR::InstId inst_id)
...
+  CARBON_CHECK(!result.entry.has_returned_var)
+      << "Should not suspend a scope with a returned var.";
+  return result;
+}
+
+auto ScopeStack::Restore(SuspendedScope scope) -> void {
+  for (auto entry : scope.suspended_lookups) {
+    // clang-tidy warns that the `std::move` below has no effect. While that's
+    // true, this `move` defends against the suspended lookup growing more state
+    // later.
```
jonmeow, line 161: I'm a little confused by this, the parameter is taken by value so I wouldn't expect it to be that defensive.

jonmeow (commented):
High level I think this looks good, though I have a couple terminology questions regarding `method` and `sus`

common/variant_helpers.h
zygoloid: The compiler implements the feature, but we enable a warning for using CTAD on a type that doesn't declare any deduction guides.

toolchain/check/lexical_lookup.h
zygoloid: It's converting a `NameId` to an array index, which (for `SmallVector`) is expected to be a `size_t`. (The internal representation of `NameId` doesn't seem important here.)

toolchain/check/lexical_lookup.h
zygoloid: I'm inclined to go with `uint32_t` here, and an assert in `Suspend` that it fits (which should never fail because a converted `NameId` can only barely exceed the range of `int32_t`).

toolchain/check/check.cpp
zygoloid: I think it's very important that the definitions of these functions live next to `HandleFunctionDefinitionStart` and `HandleFunctionDefinition`, because they need to be maintained together and need to behave consistently.

How would you feel about adding a `handle_function.h`?

(As far as I'm aware, this is it in terms of suspending and resuming node actions. But who knows how we'll change the design in future.)

toolchain/check/check.cpp
zygoloid: Hm. I structured it like this because the `SuspendFinished...` call has important side-effects that should only happen if the `IsEnd...` call returns `true`, and I was concerned that `&&` wouldn't give a sufficient hint to the reader that the short-circuiting is important. Perhaps there's another structuring that would help here; how about this:

```suggestion
    if (IsEndOfDelayedMethodScope(parse_kind)) {
      chunks.back().parsing_skipped_methods =
          worklist.SuspendFinishedScopeAndPush(context);
    }
```

toolchain/check/check.cpp
jonmeow: handle_function.h seems okay, or handle.h both the main #define and this part.

toolchain/parse/context.cpp
zygoloid: Hm. Yes, I like the idea, and I think that could work well. I suppose this would amount to the parser producing a list of actions:

- process chunk from NodeId A to NodeId B
- suspend function
- resume function
- suspend scope
- resume scope
- leave scope

and the check side of things going back to a simple in-order execution of the actions given to it by the parser.

But yeah, I'd prefer to land this PR first. Most of the pieces here should be reusable in that approach.

toolchain/check/check.cpp
jonmeow: SG

toolchain/check/scope_stack.cpp
zygoloid: It's taken by value, but we're moving into the function parameter. If the suspension object owned something (eg, a tiny vector), this'd mean we move that thing rather than copying it.

The clang-tidy check is just bogus in this situation: it's preventing us from expressing a semantic move because it's looked at the implementation details of the type and determined that it would happen to be the same as a copy, but that is not a reasonable reason to issue a complaint -- it's expecting us to rely on implementation details of a type we don't own, and not express what we mean.

However, I think the check is probably what we want in the majority of cases, and sadly the same clang-tidy check name also covers other situations where the check is detecting a bug (moving a const object), so I don't think we should turn it off.

toolchain/check/check.cpp
zygoloid: It's classified.

toolchain/check/check.cpp
zygoloid: Ooh, I really like the `handle.h` idea. Thanks, done!

toolchain/check/check.cpp
zygoloid: Hm, I hadn't considered that. I don't think we have a way to ask "does this node match that `NodeIdOneOf` type?" that doesn't CHECK-fail if not, currently, but we could add that if you think it would aid the clarity here.

(FWIW, I've been trying to arrange this so that LLVM will smartly combine the three `switch`es per `kind` into a single `switch`. I'm not sure whether that's as likely to happen if we use something other than `switch` here, but we can dig into that if we see performance concerns.)

I want to keep the list of start and end node kinds here, because it's important that these three lists all match, so I wouldn't want to use a typedef declared out in `node_ids.h`, in case that affects your opinion here. This is also why I'm reluctant to use a `NodeCategory` for this instead of `NodeIdOneOf` or an explicit list of kinds.

toolchain/check/check.cpp
zygoloid: Made the whole PR classier, and moved `NodeIdTraversal::Next` out of line. `InlineMethodWorklist::SuspendFinishedScopeAndPush` is borderline for me for a class defined in a `.cpp` file, would you like that one out of line too?

toolchain/check/check.cpp
zygoloid: Should be less sus now. I don't think we need an abbreviation here, really.

For other names: I think "delayed", "deferred", and "relocated" do a less good job of capturing "already started but not active right now" for me. I could go with "paused" (and `Pause` instead of `Suspend`) if that's better? Looking through a thesaurus, I also considered "interrupted", which I don't think works well because of other potential meanings, and "adjourned", which seems a bit too obscure.

toolchain/check/check.cpp

toolchain/check/lexical_lookup.h
zygoloid: Changed to `uint32_t`.

toolchain/parse/context.h
zygoloid: Added.

toolchain/parse/context.cpp
zygoloid: Yes, I agree. Let's find better names. Will start a discussion on discord in #naming.

zygoloid: Updated based on names favored on discord.

toolchain/check/check.cpp
zygoloid: We don't really have much layering here, but given the amount of state we're tracking I've been trying to split this up based on which state the different pieces need to interact with. `ProcessNodeIds` doesn't need to interact with the private members of `NodeIdTraversal`, so I've kept it separate for now. Happy to change that if you'd find that clearer or better organized.

toolchain/check/check.cpp
jonmeow: Eh, maybe it's not common enough? I was just trying to think of a way to write shorter cde.

toolchain/check/lexical_lookup.h
jonmeow: Most of my thought is that it seems a little inconsistent with surrounding code. But also, you do:

```
auto index = GetLookupIndex(name_id);
auto& results = lookup_[index];
...
return {static_cast<uint32_t>(index), results.pop_back_val().inst_id};
```

So yes it's used as an array index, but because it comes from a NameId required to fit inside a uint32_t. Additionally, it's natively a int32_t. So it's being upsized and downsized, but it's practically restricted to be a valid int32 (arguably, max values might overflow, but that feels really edge-casey).

Anyways, I'm not insisting here, but maybe it's worth thinking about that you have something that must be a valid int32_t, is converted to a size_t, then to a uint32_t. int32_t stores compactly and would be a consistent type.

toolchain/check/lexical_lookup.h
jonmeow: Per above, I'd suggest size_t or int32_t. My leaning would be away from uint32_t per [style](https://google.github.io/styleguide/cppguide.html#Integer_Types): "You should not use the unsigned integer types such as uint32_t, unless there is a valid reason such as representing a bit pattern rather than a number, or you need defined overflow modulo 2^N. In particular, do not use unsigned types to say a number will never be negative."

But maybe there's a specific intent here that I've missed.

toolchain/check/check.cpp
jonmeow: FWIW, I'm now wondering if the whole traversal logic should be its own .h/.cpp, but that seems out of scope for this PR. If the class were split to its own file, the function might make more sense to merge in since there's not much of an abstraction boundary.

jonmeow (approved):
Thanks for the changes! Tests are failing, but feel free to merge once they're passing. I don't think any of my comments require changes at this point.

toolchain/check/check.cpp
jonmeow: This seems okay right now.

toolchain/check/check.cpp
jonmeow: "Suspend" seems fine to me, was just trying to think if you wanted something with a clean abbreviation.

I don't know if "deferred" would work better for you now,j given the other changes. Maybe it's more complex; PartialDeferredFunction? :)

toolchain/check/lexical_lookup.h
zygoloid: I've added a comment to make sure the intent is documented.

toolchain/check/check.cpp
zygoloid: I think we want two different words here. We're using "deferred" in a timeless sense to refer to functions and scopes that need special processing, whether or not that special processing has already begun or finished, whereas we're using "suspended" to refer to things that are suspended *right now*. For example, we have comments like "Suspend the current deferred definition scope" that would be confusing if we used the same word for both senses.

PR #3833, opened 2024-03-29
SemIR fidelity when representing rewrite semantics

chandlerc:
The toolchain's [Semantic IR][semir] should start off modeling the full,
complex, and rich library-based and generic extension point semantics of Carbon
without eliding any layers or rewrites for compile time efficiency. We shouldn't
front-load elision or optimization when implementing the designs.

Once we have a full-fidelity implementation, we should work to build an
efficient elision, short-circuit, or common case simplification into the design
itself sufficient to make the SemIR model efficient. Only if we cannot find a
reasonable approach for that should we diverge the SemIR model to optimize its
efficiency, and we should preserve full fidelity in an optional mode.

[semir]:
    https://docs.google.com/document/d/1RRYMm42osyqhI2LyjrjockYCutQ5dOf8Abu50kTrkX0/edit?resourcekey=0-kHyqOESbOHmzZphUbtLrTw#heading=h.503m6lfcnmui

### DIFF
--- a/proposals/p3833.md
+++ b/proposals/p3833.md
@@ -115,6 +115,10 @@ result in sunk cost.
 
 ### Example: overloaded operators
 
+> Note: this example isn't trying to capture all the nuance of the current
+> design or suggest any change to the design. It is merely trying to provide an
+> illustration of how we might follow the proposal in practice.
+
 As an example to help illustrate this, how should the toolchain model overloaded
 operators? This proposal suggests that _initially_, the implementation should
 aim to fully model the rewrite-based dispatch through interfaces in the prelude.
@@ -123,9 +127,30 @@ be used to model the rewritten semantics of `x.(Core.AddWith(typeof(y)).Op)(y)`.
 This in turn would dispatch to an exact-type implementation which would provide
 any implicit conversions, and so-on.
 
-> Note: this example isn't trying to capture all the nuance of the current
-> design or suggest any change to the design. It is merely trying to provide an
-> illustration of how we might follow the proposal in practice.
+To be precise, the expectation is that the SemIR for `x + y` should as a
+consequence model all of:
+
+-   Looking up the `Core.AddWith` interface.
+-   Passing the type of `y` as a parameter to the interface (but likely using
+    that type directly rather than forming a literal `typeof(y)` kind of
+    expression, see below).
+-   Looking up the `Op` method.
+-   Calling that `Op` method on `x` with `y` as its parameter just as we would
+    model any call to `object.(Interface.Method)(y)`.
+
+However, specific things that would still be reasonable without deviating from
+the full semantic modeling here:
+
+-   Not representing a user-written `typeof` style expression to compute the
+    type, or duplicating the `y` expression.
+-   Caching and re-using SemIR for immutable things like successfully resolved
+    name lookup. Concretely, re-using a single cached lookup of `Core.AddWith`
+    would still provide the full model. And similarly, re-using an already
+    computed parameterization such as `Core.AddWith(i32)` repeatedly, and the
+    lookup of `Op` within that parameterization.
+
+The reason is that these don't change the semantics in any way -- they are
+constructively the same results.
 
 This should be the initial implementation _goal_, not a requirement for any
 incremental progress on implementation. Short-circuiting or other approximations
--- a/proposals/p3833.md
+++ b/proposals/p3833.md
@@ -126,7 +126,7 @@ That is, each use of `x + y` should turn into roughly equivalent SemIR as would
 be used to model the rewritten semantics of
 `x.(package#Core.AddWith(typeof(y)).Op)(y)`. This in turn would dispatch to an
 exact-type implementation which would provide any implicit conversions, and
-so-on. And importantly, this would occur even when we know that the types of `x`
+so on. And importantly, this would occur even when we know that the types of `x`
 and `y` are exactly `i32` and we could alternatively directly hard code the
 toolchain to use the desired builtin implementation.
 
--- a/proposals/p3833.md
+++ b/proposals/p3833.md
@@ -127,8 +127,8 @@ be used to model the rewritten semantics of
 `x.(package#Core.AddWith(typeof(y)).Op)(y)`. This in turn would dispatch to an
 exact-type implementation which would provide any implicit conversions, and
 so-on. And importantly, this would occur even when we know that the types of `x`
-and `y` are exactly `i32` and we might be able to directly hard coded to a
-builtin implementation.
+and `y` are exactly `i32` and we could alternatively directly hard code the
+toolchain to use the desired builtin implementation.
 
 To be precise, the expectation is that the SemIR for `x + y` should as a
 consequence model all of:
--- a/proposals/p3833.md
+++ b/proposals/p3833.md
@@ -112,6 +112,34 @@ understand how the design works in complex cases and to debug unexpected
 behavior in the toolchain. This means starting with that implementation doesn't
 result in sunk cost.
 
+### Example: overloaded operators
+
+As an example to help illustrate this, how should the toolchain model overloaded
+operators? This proposal suggests that _initially_, the implementation should
+aim to fully model the rewrite-based dispatch through interfaces in the prelude.
+That is, each use af `x + y` should turn into roughly equivalent SemIR as would
+be used to model the rewritten semantics of `x.(Core.AddWith(typeof(y)).Op)(y)`.
+This in turn would dispatch to an exact-type implementation which would provide
+any implicit conversions, and so-on.
+
+> Note: this example isn't trying to capture all the nuance of the current
+> design or suggest any change to the design. It is merely trying to provide an
+> illustration of how we might follow the proposal in practice.
+
+This should be the initial implementation _goal_, not a requirement for any
+incremental progress on implementation. Short-circuiting or other approximations
+used in incrementally building towards this goal should always be reasonable to
+consider based on the tradeoffs at hand. But they should be seen as temporary,
+incremental measures and where possible removed so we can evaluate the full
+semantic model when the implementation is complete.
+
+Based on the practical costs that this kind of model incurs once we have it
+implemented, we should first look for ways to adjust the design itself to avoid
+forming the complex semantic expansion in enough cases to make the model
+scalable. And only when we've exhausted those options begin adding long-term
+shortcuts in the SemIR model for common cases with a flag to disable them for
+testing and debugging.
+
 ## Rationale
 
 -   [Language tools and ecosystem](/docs/project/goals.md#language-tools-and-ecosystem)
--- a/proposals/p3833.md
+++ b/proposals/p3833.md
@@ -125,7 +125,9 @@ aim to fully model the rewrite-based dispatch through interfaces in the prelude.
 That is, each use of `x + y` should turn into roughly equivalent SemIR as would
 be used to model the rewritten semantics of `x.(Core.AddWith(typeof(y)).Op)(y)`.
 This in turn would dispatch to an exact-type implementation which would provide
-any implicit conversions, and so-on.
+any implicit conversions, and so-on. And importantly, this would occur even when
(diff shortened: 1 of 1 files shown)


proposals/p3833.md
```diff
@@ -0,0 +1,157 @@
...
+    -   [Directly implement the optimized model](#directly-implement-the-optimized-model)
+
+<!-- tocstop -->
+
+## Abstract
+
+The toolchain's [Semantic IR][semir] should start off modeling the full,
+complex, and rich library-based and generic extension point semantics of Carbon
+without eliding any layers or rewrites for compile time efficiency. We shouldn't
+front-load elision or optimization when implementing the designs.
```
jonmeow, line 34: Is the emphasis here "for compile time efficiency"? e.g., we've [temporarily] elided layers in order to reduce implementation cost -- is that okay? Or what does it mean to "start off" modeling everything -- does it really mean, it won't be considered complete until it models everything? [What about the 0.1 milestone, which doesn't have everything?]

Note too, we do also optimize a little as we go along. I think there's a balance here where we probably don't want to spend too much time, but we should probably front-load a little.

jonmeow: For a proposal, maybe the focus should be more on the goal of what the toolchain should be doing at milestones, rather than the implementation and how it evolves to get there?

proposals/p3833.md
```diff
@@ -0,0 +1,157 @@
...
+lesser degree. Some examples:
+
+-   [Proposal #820 - Implicit conversions](https://github.com/carbon-language/carbon-lang/pull/820)
+-   [Proposal #845 - `as` expressions](https://github.com/carbon-language/carbon-lang/pull/845)
+-   [Proposal #1083 - Arithmetic expressions](https://github.com/carbon-language/carbon-lang/pull/1083)
+
+The most significant step that motivates capturing an explicit strategy here is
+[proposal #3720 - Binding operators](https://github.com/carbon-language/carbon-lang/pull/3720).
+
+## Proposal
```
josh11b, line 73: I feel like an example would make this proposal a lot clearer.

proposals/p3833.md
chandlerc: Not sure I understand this enough to address it based on the review alone -- might be worth having a VC discussion around it.

First off -- I'm not sure a proposal is the right way to document this, was an idea from @zygoloid and I figured we could try it. If a different model works, happy with that.

Secondly, definitely don't want to preclude whatever is needed when making incremental progress towards an implementation. I've added an example as suggested by Josh below and clarified this point in the example but happy to add other clarifications around that.

Beyond that, I think the example may also answer some of this. If it does, happy to adjust the rest of the wording to be more clear?

proposals/p3833.md
chandlerc: Good idea, added one below.

proposals/p3833.md
```diff
@@ -112,6 +113,34 @@ understand how the design works in complex cases and to debug unexpected
 behavior in the toolchain. This means starting with that implementation doesn't
 result in sunk cost.
 
+### Example: overloaded operators
+
+As an example to help illustrate this, how should the toolchain model overloaded
+operators? This proposal suggests that _initially_, the implementation should
+aim to fully model the rewrite-based dispatch through interfaces in the prelude.
+That is, each use af `x + y` should turn into roughly equivalent SemIR as would
```
josh11b:
```suggestion
That is, each use of `x + y` should turn into roughly equivalent SemIR as would
```

proposals/p3833.md
```diff
@@ -112,6 +113,34 @@ understand how the design works in complex cases and to debug unexpected
...
+be used to model the rewritten semantics of `x.(Core.AddWith(typeof(y)).Op)(y)`.
```
josh11b: @zygoloid Is this syntax allowed?

proposals/p3833.md
```diff
@@ -112,6 +113,34 @@ understand how the design works in complex cases and to debug unexpected
...
+This in turn would dispatch to an exact-type implementation which would provide
+any implicit conversions, and so-on.
```
josh11b: Could you explain the alternative(s) that is/are being decided against in terms of this example as well?

testing/file_test/autoupdate.cpp
```diff
@@ -103,7 +103,7 @@ auto FileTestAutoupdater::CheckLine::RemapLineNumbers(
 }
 
 auto FileTestAutoupdater::GetFileAndLineNumber(
-    llvm::DenseMap<llvm::StringRef, int> file_to_number_map,
+    const llvm::DenseMap<llvm::StringRef, int>& file_to_number_map,
```
jonmeow: It looks like some commits have bled over, maybe a mis-merge.

proposals/p3833.md
jonmeow: A discussion might help. It sounds like this is deliberately making an implementation choice; I would like to understand the review process for future toolchain implementation choices, and it would probably be good to discuss checking in the toolchain architecture doc since this proposal would be the first thing deciding how the toolchain (and SemIR specifically) works.

testing/file_test/autoupdate.cpp
chandlerc: `git-stack` misbehaved in a way I cannot figure out.... trying to repair the history now

chandlerc: The diff should be fixed now.

chandlerc:
(the brief diff issues should be fixed now, sorry for that)

proposals/p3833.md
zygoloid: Roughly, yes, I think so. I see a couple of issues here:

- `typeof` doesn't exist (yet).
- I would not expect us to need `Core` to be in scope -- I think we mean "the `AddWith` that lives in the prelude" rather than "the result of name lookup into the name `Core`" here. In #1083, the rule was (supposed to be) that we look for an implementation of the interface from the standard library, not that we do a name lookup to find the name `Core` and look for these various names within whatever we find.

proposals/p3833.md
zygoloid: Maybe we could point out that this happens even when adding together (say) `i32`s.

proposals/p3833.md
```diff
@@ -0,0 +1,211 @@
...
+aim to fully model the rewrite-based dispatch through interfaces in the prelude.
+That is, each use of `x + y` should turn into roughly equivalent SemIR as would
+be used to model the rewritten semantics of `x.(Core.AddWith(typeof(y)).Op)(y)`.
+This in turn would dispatch to an exact-type implementation which would provide
+any implicit conversions, and so-on.
+
+To be precise, the expectation is that the SemIR for `x + y` should as a
+consequence model all of:
+
+-   Looking up the `Core.AddWith` interface.
```
zygoloid: Maybe:
```suggestion
-   Looking up the `AddWith` interface within the standard library package.
```
In particular, an unqualified lookup of the name `Core` isn't part of the design. (Performing a qualified lookup for the interface names wasn't my intent in #1083 either, but I don't think that's an important distinction, and I'd be OK with the rule being that you do perform that qualified lookup. But I'd be pretty strongly opposed to using an unqualified lookup for the name `Core`.)

proposals/p3833.md
```diff
@@ -0,0 +1,211 @@
...
+semantic model when the implementation is complete.
+
+Based on the practical costs that this kind of model incurs once we have it
+implemented, we should first look for ways to adjust the design itself to avoid
+forming the complex semantic expansion in enough cases to make the model
+scalable. And only when we've exhausted those options begin adding long-term
+shortcuts in the SemIR model for common cases with a option to disable them for
+testing and debugging.
+
+## Rationale
```
zygoloid, line 180: One thing I think would be worth discussing here is the impact of this proposal on cycles in the design: it'd be easy for us to accidentally define (for example) binding in terms of function calls and function calls in terms of bindings in such a way that we'd never bottom out. By requiring the toolchain to follow the design and not take shortcuts, at least under a flag, we can empirically test that our foundations actually work.

This means, for example, however we cut the loop of a function call desugaring into another function call needs to be determined in the design, not merely done in the toolchain at a convenient spot. So we shouldn't have the toolchain say "this is a direct call to a function; I'm not using the full function call machinery to dispatch it, because that would end up resulting in another function call", and instead we should have the *design* say that, if that's the intended approach. And if the intended approach is instead that we do use the full function call machinery, including `impl` lookup, but only *once*, then the design should say that instead.

proposals/p3833.md
chandlerc: Ah, that makes sense, I think done.

proposals/p3833.md
chandlerc: Ah, now I understand the confusion when we've discussed this in the past.

Definitely not suggesting we should be doing arbitrary unqualified lookup. I agree that would be bad.

Really, this seems to suggest that there is a fundamental gap -- we should have some unambiguous way of looking up a name in the `Core` package without doing that name lookup. We have `package.Name`, we kind of need `(package Core).Name` but with some much better syntax. Is there a way I could frame this in pseudo syntax here?

proposals/p3833.md
chandlerc: I tried adding a section about this. Not sure I really explained it well though, so ssuggestions welcome. =]

chandlerc (commented):
Thanks for comments, most addressed, and one follow-up question inline!

proposals/p3833.md
zygoloid: I'd be fine with the syntax you suggested, eg:
```suggestion
-   Looking up the `(package Core).AddWith` interface, where `package Core` is pseudo-syntax that directly names the `Core` package.
```

chandlerc:
(Fixed weird mis-merge, sorry about that)

proposals/p3833.md
chandlerc: I've added a slightly different imagined syntax and clarified the behavior and that these are all imagined syntaxes.

proposals/p3833.md
chandlerc: I think the syntax is now clarified as imagined syntax and not doing weird things with name lookup.

chandlerc (commented):
PTAL, I think the comments are addressed now?

chandlerc:
Ping? It's been a while since update...

proposals/p3833.md
```diff
@@ -0,0 +1,231 @@
...
+> illustration of how we might follow the proposal in practice.
+
+As an example to help illustrate this, how should the toolchain model overloaded
+operators? This proposal suggests that _initially_, the implementation should
+aim to fully model the rewrite-based dispatch through interfaces in the prelude.
+That is, each use of `x + y` should turn into roughly equivalent SemIR as would
+be used to model the rewritten semantics of
+`x.(package#Core.AddWith(typeof(y)).Op)(y)`. This in turn would dispatch to an
+exact-type implementation which would provide any implicit conversions, and
+so-on. And importantly, this would occur even when we know that the types of `x`
```
zygoloid:
```suggestion
so on. And importantly, this would occur even when we know that the types of `x`
```

zygoloid (approved):
Looks good, let's move forward with this. Apologies for the delay.

proposals/p3833.md
```diff
@@ -0,0 +1,231 @@
...
+> illustration of how we might follow the proposal in practice.
+
+As an example to help illustrate this, how should the toolchain model overloaded
+operators? This proposal suggests that _initially_, the implementation should
+aim to fully model the rewrite-based dispatch through interfaces in the prelude.
+That is, each use of `x + y` should turn into roughly equivalent SemIR as would
+be used to model the rewritten semantics of
+`x.(package#Core.AddWith(typeof(y)).Op)(y)`. This in turn would dispatch to an
+exact-type implementation which would provide any implicit conversions, and
+so on. And importantly, this would occur even when we know that the types of `x`
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
exact-type implementation which would provide any implicit conversions, and so
on. And importantly, this would occur even when we know that the types of `x`
```

PR #3834, opened 2024-03-29
Make file_test dramatically faster.

chandlerc:
Our testing had started to take more noticable time so I looked at where the time went to see if it could be improved easily. Almost all of the time was spent building regex matchers. That code path allocates a lot of memory and does a lot of processing in order to make the regex fast to run over the input. While that's not really a great tradeoff for how we use these matchers, the bigger issue is that we don't need a regex is the vast majority of cases.

This change inspects the pattern more deeply to avoid most of the work.

First, it looks for the
cases where the pattern doesn't require any adjustment at all and directly builds a matcher from that if it can. This avoids an extra string allocation entirely. This is probably only a minor improvement, but it is also very easy.

The larger change is to process the string in two phases. First, we expand the keywords and check for a regex region. If we find no regex region, we can directly build a string equality matcher for the expanded string. This still allocates an extra copy of the string but is *dramatically* cheaper that building a regex.

Finally, if we *do* find a regex, we re-process the string to escape everything and transform the regex sequence into its valid form before building the matcher.

The result for me is an over 5x reduction in test time. Profiling afterward shows a lot more opportunities for optimization here if things get slow again. The actual toolchain code isn't really visible in the profile yet.

Note, I was profiling the normal build, which has asserts and ASan and such. I've not looked at the optimized build.

### DIFF
--- a/testing/file_test/file_test_base.cpp
+++ b/testing/file_test/file_test_base.cpp
@@ -472,10 +472,10 @@ static auto TryConsumeSplit(
   return true;
 }
 
-// Transforms a `FileCheck` pattern into a single complete regex by escaping all
-// regex characters outside of the designated `{{...}}` regex sequences, and
-// switching those to a normal regex sub-pattern syntax.
-static void TransformFileCheckToRegex(std::string& str) {
+// Converts a `FileCheck`-style expectation string into a single complete regex
+// string by escaping all regex characters outside of the designated `{{...}}`
+// regex sequences, and switching those to a normal regex sub-pattern syntax.
+static void ConvertExpectationStringToRegex(std::string& str) {
   for (int pos = 0; pos < static_cast<int>(str.size());) {
     switch (str[pos]) {
       case '(':
@@ -588,7 +588,7 @@ static auto TransformExpectation(int line_index, llvm::StringRef in)
   // Otherwise, we need to turn the entire string into a regex by escaping
   // things outside the regex region and transforming the regex region into a
   // normal syntax.
-  TransformFileCheckToRegex(str);
+  ConvertExpectationStringToRegex(str);
   return Matcher<std::string>{MatchesRegex(str)};
 }
 
--- a/testing/file_test/file_test_base.cpp
+++ b/testing/file_test/file_test_base.cpp
@@ -482,69 +482,55 @@ static auto TransformExpectation(int line_index, llvm::StringRef in)
   if (!in.consume_front(" ")) {
     return ErrorBuilder() << "Malformated CHECK line: " << in;
   }
-  // Try to avoid an extra string copy if the input is already in a viable form.
-  if (in.find("[[") == llvm::StringRef::npos &&
-      in.find("{{") == llvm::StringRef::npos) {
+
+  // Check early if we have a regex component as we can avoid building an
+  // expensive matcher when not using those.
+  bool has_regex = in.find("{{") != llvm::StringRef::npos;
+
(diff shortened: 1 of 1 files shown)


testing/file_test/file_test_base.cpp
```diff
@@ -479,14 +479,84 @@ static auto TransformExpectation(int line_index, llvm::StringRef in)
...
+      case '{': {
+        if (pos + 1 != static_cast<int>(str.size()) && str[pos + 1] != '{') {
+          // Save that we found a regex region, we'll expand it below.
+          has_regex = true;
+          // Do an extra increment to move past the second curly.
+          ++pos;
+        }
+        ++pos;
+        break;
+      }
```
jonmeow: Above, you do a `.find("{{"}`. Does it work if you use the result of that for `has_regex`, i.e.:

```
  // Try to avoid an extra string copy if the input is already in a viable form.
  bool has_regex = in.find("{{") != llvm::StringRef::npos;
  if (in.find("[[") == llvm::StringRef::npos && !has_regex) {
    return Matcher<std::string>{StrEq(in)};
  }
```

If that's equivalent, you could remove this case, at which point the loop would only be handling `[[`. That in turn you might be able to handle as:

```
for (int pos = str.find("[["); pos != std::string::npos; pos = str.find("[[", pos))
```

That also has a side-effect of skipping straight past single `[` characters (removing a minor condition), and you could reuse the `in.find("[[")` result above to initialize `pos` too.

In combination, that might also reduce this code enough to address something else I might comment on -- you're adding about 40 lines to this function, taking it to 120-ish lines, which seems a bit high of a line count [for style](https://google.github.io/styleguide/cppguide.html#Write_Short_Functions) right now.

jonmeow (approved):
Nice!

testing/file_test/file_test_base.cpp
chandlerc: Done and extracted a function to keep the size down. It seemed much more readable to extract the regex expansion as that doesn't have lots of error paths or contextual short circuiting, it just processes the string top to bottom. Seemed pretty clean, but lemme know if you'd like anything different. PTAL.

testing/file_test/file_test_base.cpp
```diff
@@ -596,6 +520,75 @@ static auto TransformExpectation(int line_index, llvm::StringRef in)
...
+
+  // If there was no regex, we can directly match the adjusted string.
+  if (!has_regex) {
+    return Matcher<std::string>{StrEq(str)};
+  }
+
+  // Otherwise, we need to turn the entire string into a regex by escaping
+  // things outside the regex region and transforming the regex region into a
+  // normal syntax.
+  TransformFileCheckToRegex(str);
```
jonmeow: Minor, but given the other function name "TransformExpectation", you might consider "TransformExpectationToRegex" or rename "TransformExpectation" to "TransformFileCheck".

jonmeow (approved):
Looks good, thanks!

testing/file_test/file_test_base.cpp
chandlerc: I was trying to differentiate it because this *doesn't* transform an entire expectation, just the pattern-to-be-matched. But maybe "FileCheck" isn't a good name of the non-regex-pattern-syntax. Happy for another name, but would prefer there is a somewhat clear distinction between this transform name and the one below.

chandlerc: Picked a name, merging.

PR #3835, opened 2024-03-29
Some minor efficiency fixes.

chandlerc:
This doesn't move the needle too much, but cleans some things up. It let me see an underlying issue that I'll try to fix next.


jonmeow (approved):
Looks good, merging though I wonder if it's worth switching to llvm::Twine instead of `.append` for simplicity (I'm not sure what relative performance is)

PR #3836, opened 2024-03-29
When merging functions, detect uses of imports prior to a redecl.

jonmeow:
LookupNameInDecl is only called from DeclNameStack, but I'm adding mark_imports_used there because it feels more consistent. Not sure if we want a better API boundary. I admit I'm also suspicious of its call to LookupInCurrentScope but maybe it's okay due to how imports work.

I was choosing to print multiple diagnostics when a declaration is previously used _and_ doesn't match because I think the "previously used" is more important, but the "doesn't match" may give an additional hint about why it didn't work.

The merge.h utility function is because I think we can follow a similar model for identifying errors with other declarations: classes, interfaces, etc.


jonmeow:
Depends on #3831

chandlerc (approved):
LG -- to be honest, pretty light review of the diagnostics, as they seem like a great starting point.

PR #3842, opened 2024-04-01
Address invalid redeclaration and definitions across imports.

jonmeow:
Note, there's still a TODO "Allow non-extern declarations in the same library."


jonmeow:
Depends on #3836

chandlerc (approved):
Nice!

PR #3843, opened 2024-04-02
Special-case api/impl implicit imports and verify relevant redeclarations.

jonmeow:
Adds ImportIRId::ApiForImpl to reserve a specific slot for the `api` import, so that the code can trivially determine whether an import is from the same library. This is then used for merging function declarations, because the rules for redeclarations in the same library slightly differ as compared to other imports (note they're also not identical to same-file rules).

The main thing this leaves from the recent #3762 is verifying that entities forward declared in the `impl` file are also defined, but that's not in-scope for merging; it's moreso post-checking validation.

Note, a lot of our `invalid <entity> ID` comments in ids.h were incorrectly copy-pasted, so I've cut `<entity>`.

### DIFF
--- a/toolchain/check/merge.h
+++ b/toolchain/check/merge.h
@@ -15,8 +15,7 @@ struct InstForMerge {
   // The resolved instruction.
   SemIR::Inst inst;
   // The imported instruction, or invalid if not an import. This should
-  // typically only be used for the ImportIRId, but we only resolve that if
-  // needed.
+  // typically only be used for the ImportIRId, but we only load it if needed.
   SemIR::ImportIRInstId import_ir_inst_id;
 };
 
--- a/toolchain/check/merge.h
+++ b/toolchain/check/merge.h
@@ -7,14 +7,17 @@
 
 #include "toolchain/check/context.h"
 #include "toolchain/sem_ir/ids.h"
+#include "toolchain/sem_ir/import_ir.h"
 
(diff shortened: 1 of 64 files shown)


josh11b (approved):
I'm not super familiar with this code, so you may want to get another review, but I didn't see any problems.

toolchain/check/merge.h
```diff
@@ -7,22 +7,25 @@
...
 
-struct PrevInstForMerge {
+struct InstForMerge {
   // The resolved instruction.
   SemIR::Inst inst;
-  // Whether an import was encountered during resolution.
-  bool is_import;
+  // The imported instruction, or invalid if not an import. This should
+  // typically only be used for the ImportIRId, but we only resolve that if
+  // needed.
```
zygoloid: I don't understand what the part of this second sentence after the comma is telling me. What do we only resolve if needed -- the ImportIRId? If so, what does it mean to resolve an ImportIRId?

zygoloid (approved):

toolchain/check/merge.h
jonmeow: rephrasing: resolve that -> load it

ImportIRInstId is a reference to an ImportIRId and an InstId. Only the ImportIRId should be used in this context, but it's not actually available unless the ImportIRInst that contains the actual ImportIRId is loaded/resolved.

PR #3844, opened 2024-04-02
Fix determination of whether we're in a nested deferred definition scope.

zygoloid:
Instead of attempting to reconstruct this based on the properties of the target scope of the declaration name, check whether the enclosing scope of the inner deferred definition scope is the enclosing deferred definition scope.

Also add verbose logging to the deferred definition handling code to make it easier to see when we switch between handling the regular parse nodes and the deferred function bodies.

### DIFF
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -8,6 +8,8 @@
 
 #include "common/check.h"
 #include "common/error.h"
+#include "common/variant_helpers.h"
+#include "common/vlog.h"
 #include "toolchain/base/pretty_stack_trace_function.h"
 #include "toolchain/check/context.h"
 #include "toolchain/check/diagnostic_helpers.h"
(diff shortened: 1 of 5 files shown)


toolchain/check/check.cpp
```diff
@@ -455,20 +470,21 @@ class DeferredDefinitionWorklist {
   llvm::SmallVector<Task, 0> worklist_;
 
   // Indexes in `worklist` of deferred definition scopes that are currently
-  // still open.
-  llvm::SmallVector<size_t> enclosing_scopes_;
+  // still open, and the corresponding scope indexes.
+  llvm::SmallVector<std::pair<size_t, ScopeIndex>> enclosing_scopes_;
```
geoffromer: Could this be a struct instead of a pair? Meaningful member names would probably be clearer than `.first` and `.second`.

geoffromer (approved):

PR #3848, opened 2024-04-03
Lambdas

CJ-Johnson:
This document proposes a path forward to add lambdas to Carbon. It further proposes augmenting function declarations to create a more continuous syntax between the two categories of functions. In short, both lambdas and function declarations will be introduced with the `fn` keyword. The presence of a name distinguishes a function declaration from a lambda expression, and the rest of the syntax applies to both kinds. By providing a valid lambda syntax in Carbon, migration from from C++ to Carbon will be made easier and more idiomatic. In C++, lambdas are defined at their point of use and are often anonymous, meaning replacing them solely with function declarations would create an ergonomic burden compounded by the need for the migration tool to select a name.

Associated discussion docs:

* [Lambdas Discussion 1](https://docs.google.com/document/d/1rZ9SXL4Voa3z20EQz4UgBMOZg8xc8xzKqA1ufPQdTao/)
* [Lambdas Discussion 2](https://docs.google.com/document/d/14K_YLjChWyyNv3wv5Mn7uLFHa0JZTc21v_WP8RzC8M4/)
* [Lambdas Discussion 3](https://docs.google.com/document/d/1VVOlRuPGt8GQpjsygMwH2B7Wd0mBsS3Qif8Ve2yhX_A/)
* [Lambdas Discussion 4](https://docs.google.com/document/d/1Sevhvjo06Bc6wTigNL1pK-mlF3IXvzmU1lI2X1W9OYA/)


proposals/p3848.md
```diff
@@ -0,0 +1,576 @@
...
+fn Foo() {
+  let lambda: auto = fn[T:! Printable](t: T) { Print(t); };
+
+  fn FunctionDeclaration[T:! Printable](t: T) { Print(t); }
+}
+```
+
+To understand how the syntax between lambdas and function declarations is
+reasonably "continuous", refer to this table of syntactic positions and the
+following code examples.
```
geoffromer, line 281: Consider presenting this information more like this, instead:

> Function definitions have one of the following syntactic forms (where items in square brackets are optional and independent):
>
> `fn` [_name_] [ _implicit-params_ ] [_tuple-pattern_] `=>` _expression_ `;`\
> `fn` [_name_] [ _implicit-params_ ] [_tuple-pattern_] [`->` _return-type_] `{` _statements_ `}`
>
> The first form is a shorthand for the second: "`=>` _expression_ `;`" is equivalent to "`-> auto { return ` _expression_ `; }`".
> 
> _implicit-params_ consists of square brackets enclosing an optional default capture mode and any number of explicit captures, function fields, and deduced parameters, all separated by commas. The default capture mode (if any) must come first; the other items can appear in any order. If _implicit-params_ is omitted, it is equivalent to `[]`.
>
> The presence of _name_ determines whether this is a function declaration or a lambda expression.
>
> The presence of _tuple-pattern_ determines whether the function body uses named or positional parameters.
>
> The presence of "`->` _return-type_" determines whether the function body can (and must) return a value.

That's more abstract, but at least for me, it would make it much easier to see how this design is (and isn't) continuous.

proposals/p3848.md
```diff
@@ -0,0 +1,576 @@
...
+-   [Function Fields](#function-fields)
+-   [Copy Semantics](#copy-semantics)
+-   [Self and Recursion](#self-and-recursion)
+-   [Rationale](#rationale)
+
+<!-- tocstop -->
+
+## Abstract
+
+To support migration from C++ to Carbon, there must be valid syntax to capture
```
geoffromer: Stating the rationale solely in terms of C++ migration seems too narrow -- for example, it doesn't explain why we have positional parameters, or the `=>` syntax, which are unlikely to be used for C++ migration purposes. I think we want lambdas because they can make code easier to read, understand, and write, as well as because we need them for migration.

proposals/p3848.md
```diff
@@ -0,0 +1,576 @@
...
+the behavior of C++ lambdas. They are defined at their point of use and are
+often anonymous, meaning replacing them solely with function declarations will
+create an ergonomic burden compounded by the need for the migration tool to
+select a name. This proposal proposes a path forward to add lambdas to Carbon
+and augment function declarations accordingly.
```
geoffromer: I think of an abstract as a kind of summary of the document, but most of the content here is unique to the abstract, as far as I can tell. I'd reccommend moving most of this to the Rationale section, and focus the Abstract more on the key points of what's being proposed.

proposals/p3848.md
```diff
@@ -0,0 +1,576 @@
...
+There are even functions that provide a body of statements but no return value.
+
+```
+let lambda: auto = fn { Print(T.Make()); };
+
+fn FunctionDeclaration { Print(T.Make()); }
+```
+
+Functions support [captures](#function-captures) and [fields](#function-fields)
+in the square brackets.
```
geoffromer:
```suggestion
in the square brackets (in addition to deduced parameters and the `Self` parameter).
```

proposals/p3848.md
```diff
@@ -0,0 +1,576 @@
...
+
+```
+fn Foo() {
+  let lambda: auto = fn { Print($0); };
+
+  fn FunctionDeclaration { Print($0); }
+}
+```
+
+This is of course in addition to supporting explicit parameters.
```
geoffromer: This makes it sound kind of like a single function can have both.

```suggestion
Of course, functions can also have named parameters, but a single function can't have both
named and positional parameters.
```

proposals/p3848.md
```diff
@@ -0,0 +1,576 @@
...
+fn G(D) { E3; }
+
+fn G(D) -> F { E4; }
+
+fn G[B, C](D) { E3; }
+
+fn G[B, C](D) -> F { E4; }
+```
+
+**Alternative Considered**: As opposed to a continuous syntax between lambdas
```
geoffromer: We often link to relevant alternatives from the relevant design docs. In order to make them linkable, I think they need to be marked with ### section headings.

proposals/p3848.md
```diff
@@ -0,0 +1,576 @@
...
+
+**Proposal**: Positional parameters, introduced in the body of a function by way
+of the dollar sign and a corresponding non-negative parameter position integer
+(ex: `$3`), are `auto` parameters to the function in which they are defined.
+They can be used in any lambda or function declaration that lacks an explicit
+parameter list (parentheses). They are variadic by design, meaning an unbounded
+number of arguments can be passed to any function that lacks an explicit
+parameter list. Only the parameters that are named in the body will be read
+from, meaning the highest named parameter denotes the minimum number of
+arguments required by the function. Users are free to omit lower-numbered
```
geoffromer:
```suggestion
arguments required by the function. The function body is free to omit lower-numbered
```

proposals/p3848.md
```diff
@@ -0,0 +1,576 @@
...
+```
+fn Foo() {
+  let handle: Handle = Handle.Get();
+  fn MyThread[copy]() { handle.Process(); }
+  var thread: Thread = Thread.Make(MyThread);
+  thread.Join();
+}
+```
+
+### Default Capture Mode
```
geoffromer: I suggest putting this below the Capture Modes section. It's kind of disorienting to read about how to default a particular setting before reading about what the setting means.

proposals/p3848.md
```diff
@@ -0,0 +1,576 @@
...
+that it's valuable to be able to intermix explicit captures with deduced
+parameters and fields in any order that makes the most sense for the context.
+Without a clear justification for a rule that says they can't intermix, the more
+flexible behavior was favored.
+
+### Capture Modes
+
+**Proposal**: `copy` and `ref` are the two primary forms of function captures.
+They each have companion `const` versions. They are syntactically distinct from
+`let` and `var`. `let` and `var` can appear in binding patterns, such as
```
geoffromer: It feels tautological to say that `copy` and `ref` are syntactically distinct from `let` and `var`: they're different tokens, so of course they're syntactically distinct. I'd suggest instead saying that function captures are syntactically distinct from function fields. And probably that should go in the Function Captures section, because it seems more of a fact about function captures as a whole, not so much about the mode specifier tokens in particular.

proposals/p3848.md
```diff
@@ -0,0 +1,576 @@
...
+fn Example {
+  var by_copy: i32 = 0;
+  var by_const_copy: i32 = 0;
+  var by_reference: i32 = 0;
+  var by_const_reference: i32 = 0;
+
+  let lambda: auto = fn [copy by_copy,
+                         const copy by_const_copy,
+                         ref by_reference,
+                         const ref by_const_reference] {
```
geoffromer: I'm having a hard time getting my brain to distinguish the keywords from the identifiers in this example. I can't tell if that's a problem with the syntax itself, or an artifact of the fact that these identifiers all have the keywords embedded in them. Could you rename the variables to avoid the latter problem?

proposals/p3848.md
CJ-Johnson: Awesome! I added this content right above the part that you highlighted. My reading of it is that these two blocks of text are not mutually exclusive. Do you disagree?

proposals/p3848.md
CJ-Johnson: Done! Should I move Rationale to the top or leave it at the bottom?

proposals/p3848.md
CJ-Johnson: Good call, thanks!

proposals/p3848.md
CJ-Johnson: Worded this in a bit more of a verbose way, only to make it clear that `self: Self` and `addr self: Self*` are specifically for function *declarations* and specifically for such that exist inside a class/interface.

proposals/p3848.md
CJ-Johnson: Perfect. Thanks!

proposals/p3848.md
CJ-Johnson: Had to employ embedded HTML tags to make this work. Hopefully that's ok.

CJ-Johnson: Unfortunately when I ran precommit on this, it removed these html tags from the table of contents, even though I manually added them. The sections are still linkable, they just don't show up in the TOC unfortunately.

proposals/p3848.md
```diff
@@ -0,0 +1,619 @@
...
+_implicit-parameters_ consists of square brackets enclosing a optional default
+capture mode and any number of explicit captures, function fields, and deduced
+parameters, all separated by commas. The default capture mode (if any) must come
+first; the other items can appear in any order. If _implicit-parameters_ is
+omitted, it is equivalent to `[]`.
+
+The presence of _name_ determines whether this is a function declaration or a
+lambda expression.
+
+The presence of tuple-pattern determines whether the function body uses named or
```
geoffromer:
```suggestion
The presence of _tuple-pattern_ determines whether the function body uses named or
```

proposals/p3848.md
geoffromer: I agree they're not mutually exclusive. Personally I don't find the table and examples helpful, but they may work better for other people.

proposals/p3848.md
geoffromer: I think the way you have it now is good.

proposals/p3848.md
geoffromer: Would it not work to do something like this?

```md
### Alternative Considered: Multiple kinds of lambdas

As opposed to a continuous syntax between lambdas...
```

proposals/p3848.md
```diff
@@ -0,0 +1,576 @@
...
+additional restriction was considered. That being, visibility of functions with
+positional parameters could be restricted to only non-public interfaces. **This
+alternative will be put forth as a leads question before a decision is made.**
+
+## Function Captures
+
+**Proposal**: Function captures in Carbon mirror the non-init captures of C++.
+Bindings defined in a scope visible to a function can be captured in the
+function instance, be it a function declaration or a lambda. The lifetime of a
+capture is the lifetime of the function in which it exists. For example...
```
geoffromer:
```suggestion
**Proposal**: Function captures in Carbon mirror the non-init captures of C++.
A function capture declaration consists of a capture mode followed by the name of a
binding from the enclosing scope, and makes that identifier available in the
inner function body. The lifetime of a capture is the lifetime of the function in
which it exists. For example...
```

(I'm mainly trying to explicitly spell out the high-level syntax and semantics, so that the statement that they mirror C++ doesn't have to do quite so much heavy lifting.)

proposals/p3848.md
```diff
@@ -0,0 +1,576 @@
...
+}
+```
+
+## Function Fields
+
+**Proposal**: To mirror the behavior of init captures in C++, function fields
+will support nothing-implies-`let` and `var` binding patterns. These will be
+annotated with a type and initialized with the right-hand-side of an equals
+sign. The lifetime of a function field is the same as the lifetime of the
+function declaration or lambda in which it exists.
```
geoffromer: I think we should specify this slightly differently:

```suggestion
**Proposal**: Function fields mirror the behavior of init captures in C++.
A function field definition consists of an irrefutable pattern, `=`, and an initializer.
It matches the pattern with the initializer when the function definition is evaluated.
The bindings in the pattern have the same lifetime as the function, and their scope
extends to the end of the function body.
```

This is more general than what we've discussed so far, because it allows things like `fn [(a: auto, b: auto) = Foo()] {...}`, but that generalization seems desirable.

proposals/p3848.md
```diff
@@ -0,0 +1,638 @@
...
+equivalent to "`-> auto { return` _expression_ `; }`".
+
+_implicit-parameters_ consists of square brackets enclosing a optional default
+capture mode and any number of explicit captures, function fields, and deduced
+parameters, all separated by commas. The default capture mode (if any) must come
+first; the other items can appear in any order. If _implicit-parameters_ is
+omitted, it is equivalent to `[]`.
+
+The presence of _name_ determines whether this is a function declaration or a
+lambda expression.
```
zygoloid: I think we're missing a description of the behavior of the trailing `;` here. How about something like:

```suggestion
Function definitions and lambda expressions have one of the following syntactic forms (where items in
square brackets are optional and independent):

`fn` \[_name_\] \[_implicit-parameters_\] \[_tuple-pattern_\] `=>` _expression_
\[`;`\]

`fn` \[_name_\] \[_implicit-parameters_\] \[_tuple-pattern_\] \[`->`
_return-type_\] `{` _statements_ `}`

The first form is a shorthand for the second: "`=>` _expression_ `;`" is
equivalent to "`-> auto { return` _expression_ `; }`".

_implicit-parameters_ consists of square brackets enclosing a optional default
capture mode and any number of explicit captures, function fields, and deduced
parameters, all separated by commas. The default capture mode (if any) must come
first; the other items can appear in any order. If _implicit-parameters_ is
omitted, it is equivalent to `[]`.

The presence of _name_ determines whether this is a function declaration or a
lambda expression. The trailing `;` in the first form is required for a function declaration, but is not part of the syntax of a lambda expression.
```

proposals/p3848.md
```diff
@@ -0,0 +1,638 @@
...
+
+**Proposal**: `copy`, `ref`, `const copy` and `const ref` can appear as function
+captures. They behave as specified in the following table:
+
+| Capture Mode Syntax |                     Corresponding Semantics                      |
+| :-----------------: | :--------------------------------------------------------------: |
+|       `copy`        |     Capture "by-copy" where the resulting capture is mutable     |
+|    `const copy`     | Capture "by-const-copy" where the resulting capture is immutable |
+|        `ref`        |        Capture "by-reference" behaving as a C++ reference        |
+|     `const ref`     |  Capture "by-const-reference" behaving as a C++ const reference  |
```
zygoloid: Suppose I want to write this:
```carbon
let a: String = "long string I'd rather not make a copy of";
DoThingWithLazyGetter(fn [???] => a);
```

Can a `ref` capture be used to capture a `let` binding? If so, what happens -- does that create a temporary and capture a reference to it, or does that capture a value as if by `[a: auto = a]`?

If not, I think the outcome is that there isn't a way to capture a `let` binding without renaming it. You can use a function field, but our name shadowing rules would suggest that you must use a new name for the function field. I wonder if it'd be worth adding syntax for capturing a value as a value, rather than as an object.

proposals/p3848.md
```diff
@@ -0,0 +1,638 @@
...
+applies to by-copy and by-const-copy captures.
+
+By-reference and by-const-reference captures are a bit different, because even
+if the underlying type is not copyable, the reference to that object is always
+copyable.
+
+The final case is by-value function fields. Since C++ const references, when
+made into fields of a class, prevent the class from being copied, so too should
+by-value function fields prevent the function in which it is contained from
+being copied.
```
zygoloid: Should we distinguish between copy construction and copy assignment here? Const reference fields in C++ only prevent copy assignment, not copy construction.

proposals/p3848.md
```diff
@@ -0,0 +1,638 @@
...
+purpose binding patterns. The lack of an initializing expression would create an
+ergonomic burden.
+
+## Copy Semantics
+
+**Proposal**: To mirror the behavior of C++, function declarations and lambdas
+will be as copyable as their contained function fields and function captures.
+This means that, if a function holds a by-object function field, if the type of
+the field is copyable, so too is the function that contains it. This also
+applies to by-copy and by-const-copy captures.
```
zygoloid: Just to check my understanding:
```carbon
// Returns the number of times it's been called.
fn Counter[var n: i32 = 0]() { ++n; return n; }

fn Run() {
  Print(Counter());
  Print(Counter());
  var my_counter: auto = Counter;
  Print(my_counter());
  Print(my_counter());
  Print(Counter());
}
```
... would print 1 2 3 4 3.

proposals/p3848.md
```diff
@@ -0,0 +1,638 @@
...
+```
+
+#### Alternative Considered: Additional Restriction
+
+In addition to the proposed restrictions, an additional restriction was
+considered. That being, visibility of functions with positional parameters could
+be restricted to only non-public interfaces. **This alternative will be put
+forth as a leads question before a decision is made.**
+
+## Function Captures
```
zygoloid, line 435: Something it might be good to mention here:

If a function object `F` has mutable state, either because it has a non-`const` `copy` capture or because it has a `var` function field, then a call to `F` should require the callee to be a reference expression rather than a value expression. We need a mutable handle to the function in order to be able to mutate its mutable state.

proposals/p3848.md
```diff
@@ -0,0 +1,638 @@
...
+By-reference and by-const-reference captures are a bit different, because even
+if the underlying type is not copyable, the reference to that object is always
+copyable.
+
+The final case is by-value function fields. Since C++ const references, when
+made into fields of a class, prevent the class from being copied, so too should
+by-value function fields prevent the function in which it is contained from
+being copied.
+
+## Self and Recursion
```
zygoloid, line 583: Given the direction in #3720, an expression of the form `x.(F)`, where `F` is a function with a `self` or `addr self` parameter, produces a callable that holds the value of `x`, and does not hold the value of `F`. As a consequence, I think we can't support combining captures and function fields with a `self` parameter under that model.

Would that be a reasonable restriction to impose here?

proposals/p3848.md
```diff
@@ -0,0 +1,638 @@
...
+  }
+}
+
+fn Foo5 {
+  fn Bar5() {}  // ✅ Valid: Bar5 has explicit parameters
+}
+
+fn Foo6() {
+  my_list.Sort(
+    fn => $0 < $1  // ✅ Valid: Foo6 has explicit parameters
```
jonmeow, line 430:
```suggestion
    fn => ($0 < $1)  // ✅ Valid: Foo6 has explicit parameters
```

Operator precedence thing: is this intended to evaluate as `(fn => $0) < $1` or `fn => ($0 < $1)`? I suspect we might not want to define precedence between `fn =>` and `<` and, as a consequence, parentheses would be required to disambiguate. It doesn't seem like the proposal is addressing precedence, which seems fine, but just suggesting the edit in the example for clarity.

proposals/p3848.md
CJ-Johnson: Yes that is correct

proposals/p3848.md
CJ-Johnson: Added a note with this

proposals/p3848.md
CJ-Johnson: Sure! I'm fine with that. Added a note about this.

proposals/p3848.md
```diff
@@ -0,0 +1,652 @@
...
+<!-- tocstop -->
+
+## Abstract
+
+This document proposes a path forward to add lambdas to Carbon. It further
+proposes augmenting function declarations to create a more continuous syntax
+between the two categories of functions. In short, both lambdas and function
+declarations will be introduced with the `fn` keyword. The presence of a name
+distinguishes a declaration from a lambda expression, and the rest of the syntax
+applies to both kinds. See [Syntax Overview](#syntax-overview) for more.
```
chandlerc: We have a convention of keeping the abstract in the document and the proposal summary th e same -- I feel like there is a bit of content in each, could you merge them?

proposals/p3848.md
```diff
@@ -0,0 +1,652 @@
...
+  );
+}
+```
+
+#### Alternative Considered: Additional Restriction
+
+In addition to the proposed restrictions, an additional restriction was
+considered. That being, visibility of functions with positional parameters could
+be restricted to only non-public interfaces. **This alternative will be put
+forth as a leads question before a decision is made.**
```
zygoloid: This leads question has now been answered: #3860

proposals/p3848.md
zygoloid: I think we still need an answer here -- I think wanting to capture a local `let` binding will be a common case, and requiring it to be renamed seems unsatisfying from an ergonomic perspective.

proposals/p3848.md
```diff
@@ -0,0 +1,587 @@
...
+should require the callee to be a reference expression rather than a value
+expression. We need a mutable handle to the function in order to be able to
+mutate its mutable state.
+
+#### Future Work: Reference Captures
+
+Much discussion has been had so far about the implications of capturing by
+reference. For now, such behavior is supported not through captures but instead
+through function fields formed from the address of an object in the outer scope.
+It is imperative that more work be done in this area to address the erganomic
```
zygoloid:
```suggestion
It is imperative that more work be done in this area to address the ergonomic
```

zygoloid (commented):
Looks good to me, thanks, and sorry for the slow review cycle!

proposals/p3848.md
CJ-Johnson: Thanks! Fixed :)

proposals/p3848.md
```diff
@@ -0,0 +1,587 @@
...
+
+## Abstract
+
+This document proposes a path forward to add lambdas to Carbon. It further
+proposes augmenting function declarations to create a more continuous syntax
+between the two categories of functions. In short, both lambdas and function
+declarations will be introduced with the `fn` keyword. The presence of a name
+distinguishes a function declaration from a lambda expression, and the rest of
+the syntax applies to both kinds. See [Syntax Overview](#syntax-overview) for
+more.
```
chandlerc: We should probably merge the content here and the PR description.

proposals/p3848.md
```diff
@@ -0,0 +1,587 @@
...
+
+**Proposal**: A largely continuous syntax between lambdas and function
+declarations.
+
+At a high level, lambdas and function declarations will look like the following.
+
+```
+let lambda1: auto = fn => T.Make();
+
+let lambda2: auto = fn []() -> T { return T.Make(); };
```
chandlerc: I understand this and the below are just illustrating where the `[]`s and `()`s go, but I suspect we'll want to not accept empty `[]`s at least... Would it make sense to use some pseudo syntax to indicate that there's something inside? Or to have a more examples?

proposals/p3848.md
```diff
@@ -0,0 +1,587 @@
...
+
+```
+fn Foo() {
+  let lambda: auto = fn[T:! Printable](t: T) { Print(t); };
+
+  fn FunctionDeclaration[T:! Printable](t: T) { Print(t); }
+}
+```
+
+### Succinctly
```
chandlerc: Not sure what this section heading means....

(I fear it may be ... too succinct? I'll see myself out.)

proposals/p3848.md
```diff
@@ -0,0 +1,587 @@
...
+
+_implicit-parameters_ consists of square brackets enclosing a optional default
+capture mode and any number of explicit captures, function fields, and deduced
+parameters, all separated by commas. The default capture mode (if any) must come
+first; the other items can appear in any order. If _implicit-parameters_ is
+omitted, it is equivalent to `[]`.
+
+The presence of _name_ determines whether this is a function declaration or a
+lambda expression. The trailing `;` in the first form is required for a function
+declaration, but is not part of the syntax of a lambda expression.
```
chandlerc: I think this is just trying to speak to definition syntaxes.

```suggestion
The presence of _name_ determines whether this is a function definition or a
lambda expression. The trailing `;` in the first form is required for a function
definition, but is not part of the syntax of a lambda expression.
```

proposals/p3848.md
```diff
@@ -0,0 +1,587 @@
...
+
+fn[B, C] -> F { E2; }
+
+fn(D) { E3; }
+
+fn(D) -> F { E4; }
+
+fn[B, C](D) { E3; }
+
+fn[B, C](D) -> F { E4; }
```
chandlerc: Adding some whitespace based on some bikeshed-y discussion in the syntax channel...

```suggestion
fn [B, C] => A1

fn (D) => A2

fn [B, C](D) => A2

fn { E1; }

fn -> F { E2; }

fn [B, C] { E1; }

fn [B, C] -> F { E2; }

fn (D) { E3; }

fn (D) -> F { E4; }

fn [B, C](D) { E3; }

fn [B, C](D) -> F { E4; }
```

proposals/p3848.md
```diff
@@ -0,0 +1,587 @@
...
+fn G(D) { E3; }
+
+fn G(D) -> F { E4; }
+
+fn G[B, C](D) { E3; }
+
+fn G[B, C](D) -> F { E4; }
+```
+
+### Alternative Considered: Terse vs Elaborated
```
chandlerc: The alternatives considered are spread throughout the proposal rather than grouped at the bottom which I think makes it a little bit harder to get a sense of what the actual proposal is.

But maybe more importantly, it would be good to try to summarize the pros, cons, and what tipped the scales to not pursue the alternative for each one. That's an important part of the structure we typically want in proposals.

proposals/p3848.md
```diff
@@ -0,0 +1,587 @@
...
+        -   [Alternative Considered: Middle vs Beginning](#alternative-considered-middle-vs-beginning)
+-   [Function Fields](#function-fields)
+-   [Copy Semantics](#copy-semantics)
+-   [Self and Recursion](#self-and-recursion)
+    -   [Alternative Considered: Recursive Self](#alternative-considered-recursive-self)
+-   [Rationale](#rationale)
+
+<!-- tocstop -->
+
+## Abstract
```
chandlerc, line 38: It would be good to add a background section that provides reference links for Lambdas in other languages -- at least C++, Swift, and Rust.

proposals/p3848.md
```diff
@@ -0,0 +1,587 @@
...
+syntax between lambdas and function declarations, this alternative was decided
+against.
+
+```
+let lambda1: auto = @ => T.Make();
+
+let lambda2: auto = @[]() -> T { return T.Make(); };
+```
+
+## Positional Parameters
```
chandlerc, line 372: As I think Swift is one of the few other languages that has a similar feature, it would be good to a) mention that and b) indicate what if any things really differ between them.

proposals/p3848.md
```diff
@@ -0,0 +1,587 @@
...
+
+### Capture Modes
+
+**Proposal**: `let` and `var` can appear as function captures. They behave as
+they would in regular bindings.
+
+To prevent ambiguities, captures can only exist on functions where the
+definition is attached to the declaration. This means they are supported on
+lambdas and they are supported on function declarations that are immediately
+defined, but they are not supported on forward-declared functions.
```
chandlerc: Should this also mention only being supported inside the body of another function?

Or, alternatively, in an expression (lambda) or statement (function) context?

chandlerc (commented):
Very happy with the direction here. A bunch of comments, but they're all focused on the proposal write up, not the specific proposed feature.

chandlerc:
Another suggestion for the proposal -- I think it'd be nice to add some comparison for how the Carbon lambdas would look in C++.

This doesn't need to get into features that C++ lambdas have and Carbon's don't, just showing how analogous C++ lambdas would look so that folks can compare.

proposals/p3848.md
CJ-Johnson: Made this change to both captures and function fields sections :)

KateGregory:
I'm really liking this. A few small comments. More calling examples, ideally right after the first syntax example. "This would be invoked like this:" kind of thing. I see just one (though I may have missed others) example of a lambda not being put into a variable but being used right away, the comparator in the sort example. This is terrific but may be missed by someone less interested in positional parameters, so some other examples of unnamed lambdas just being used would be great. I would hate for a reader to think that the #1 use of lambdas is to put them into a named variable or value.

CJ-Johnson:
> I'm really liking this. A few small comments. More calling examples, ideally right after the first syntax example. "This would be invoked like this:" kind of thing. I see just one (though I may have missed others) example of a lambda not being put into a variable but being used right away, the comparator in the sort example. This is terrific but may be missed by someone less interested in positional parameters, so some other examples of unnamed lambdas just being used would be great. I would hate for a reader to think that the #1 use of lambdas is to put them into a named variable or value.

Thanks for the feedback, Kate! I tried to address it by adding both `// In a variable:` and `// In a function call:` variants to the top level examples. Hopefully this will clarify things :)

proposals/p3848.md
chandlerc: Just to clarify -- this is now addressed.

proposals/p3848.md
chandlerc: Resolving as this is nicely listed as an alternative and cites the issue.

proposals/p3848.md
chandlerc: (Pretty sure this is addressed, resolving)

proposals/p3848.md
chandlerc: (I like having both, let's resolve this for now.)

chandlerc (commented):
I think this LGTM, checking with leads before approving.

proposals/p3848.md
```diff
@@ -93,10 +176,20 @@ deduced parameters in the square brackets. In addition, `self: Self` or
...
 
-  fn FunctionDeclaration[var x, var y: i32 = 0] { Print(x, y); }
+  // In a function call:
+  Foo(fn [var x, var y: i32 = 0] { Print(++x, ++y); });
+  // Equivalent in C++23:
+  // Foo([x, y = int32_t{0}] mutable -> void { Print(++x, ++y); });
+
+  fn FunctionDeclaration[var x, var y: i32 = 0] { Print(++x, ++y); }
+  // Equivalent in C++23:
+  // Such a function declaration is not possible in C++.
```
zygoloid:
```suggestion
  // auto FunctionDeclaration = [x, y = int32_t{0}] mutable -> void { Print(++x, ++y); };
```

proposals/p3848.md
```diff
@@ -108,8 +201,16 @@ of type `auto`.
 ```
 fn Foo() {
   let lambda: auto = fn { Print($0); };
+  // Equivalent in C++23:
+  // Such a lambda is not possible in C++.
```
zygoloid:
```suggestion
  // auto lambda = [](auto _0, auto...) -> void { Print(_0); };
```

proposals/p3848.md
```diff
@@ -108,8 +201,16 @@ of type `auto`.
...
+  // Equivalent in Swift:
+  // let lambda = { Print($0) };
 
   fn FunctionDeclaration { Print($0); }
+  // Equivalent in C++23:
+  // Such a function declaration is not possible in C++.
+  // Equivalent in Swift:
+  // Such a function declaration is not possible in Swift.
```
zygoloid:
```suggestion
  // Equivalent in C++23:
  // auto FunctionDeclaration = [](auto _0, auto...) -> void { Print(_0); };
  // Equivalent in Swift:
  // let FunctionDeclaration = { Print($0) };
```
given that
```carbon
  fn FunctionDeclaration { Print($0); }
```
is equivalent to
```carbon
  let FunctionDeclaration: auto = fn { Print($0); };
```

zygoloid (approved):
LGTM. Some extremely minor nit-picking with equivalent versions of some of the examples in C++ here, but I'm happy with this either way; I don't think these affect the understandability of the proposal and certainly don't affect the proposed approach.

chandlerc:
Just confirming that @KateGregory gave a thumbs up offline here -- this is good to go.

@CJ-Johnson - wanted to give you a chance to merge the suggestions from @zygoloid, but also happy to merge them for you and merge the PR whenever.

chandlerc:
Accepting the suggestions as [requested](https://discord.com/channels/655572317891461132/999638000126394370/1286505898277208155) and merging!

PR #3853, opened 2024-04-03
Add builtins for integer bitwise, bit-shift, and relational comparison operators.

zygoloid:


toolchain/sem_ir/builtin_function_kind.cpp
```diff
@@ -152,6 +156,30 @@ constexpr BuiltinInfo IntDiv = {"int.div",
...
+constexpr BuiltinInfo IntOr = {"int.or",
+                               ValidateSignature<auto(IntT, IntT)->IntT>};
+
+// "int.xor": integer bitwise xor.
+constexpr BuiltinInfo IntXor = {"int.xor",
+                                ValidateSignature<auto(IntT, IntT)->IntT>};
+
+// "int.leftshift": integer left shift.
+constexpr BuiltinInfo IntLeftShift = {
+    "int.leftshift", ValidateSignature<auto(IntT, IntU)->IntT>};
```
jonmeow: Had you considered an underscore as in left_shift, etc? Might be easier to read that way, and consistent with other instruction word separation.

toolchain/lower/handle.cpp
```diff
@@ -168,10 +167,38 @@ auto HandleBuiltin(FunctionContext& /*context*/, SemIR::InstId /*inst_id*/,
...
+}
+
 // Handles a call to a builtin function.
 static auto HandleBuiltinCall(FunctionContext& context, SemIR::InstId inst_id,
                               SemIR::BuiltinFunctionKind builtin_kind,
                               llvm::ArrayRef<SemIR::InstId> arg_ids) -> void {
+  auto is_signed = [&](SemIR::InstId int_id) {
+    return context.sem_ir().types().IsSignedInt(
+        context.sem_ir().insts().Get(int_id).type_id());
+  };
```
jonmeow: FWIW this feels like it could be a small helper function instead, `static auto IsSignedInt(FunctionContext& context, SemIR::InstId inst_id)`

jonmeow (approved):
LGTM

toolchain/sem_ir/builtin_function_kind.cpp
zygoloid: Done.

toolchain/lower/handle.cpp
zygoloid: Done.

PR #3855, opened 2024-04-03
Rework raw identifier lexing to avoid slowing down regular identifiers.

zygoloid:
Instead of special-casing tokens starting with `r`, lex them as normal identifiers, and add a special case to `#` handling to detect if the previous token was an `r` identifier.

This roughly doubles the time to lex a raw identifier, because we do two hash table insertions rather than one, and probably slightly slows down lexing string literals starting with `#`, but should remove the 2% overhead to identifier lexing from the previous approach.


toolchain/lex/lex.cpp
```diff
@@ -1104,40 +1103,43 @@ auto Lexer::LexKeywordOrIdentifier(llvm::StringRef source_text,
...
-    return LexKeywordOrIdentifier(source_text, position);
+auto Lexer::LexHash(llvm::StringRef source_text, ssize_t& position)
+    -> LexResult {
+  // For `r#`, we already lexed an `r` identifier token. Detect that case and
+  // replace that token with a raw identifier. We do this to keep identifier
+  // lexing as fast as possible.
+
+  // Look for the `r` token. Note that this is always in bounds because we
+  // create a start of file token.
+  TokenIndex maybe_r_index = buffer_.tokens().end()[-1];
```
jonmeow: Can you add a test where `#` is the first token?

jonmeow: As mentioned, this feels oddly expensive versus `token_infos_.back()` -- creates an iterator range, offsets by one, then uses the result. Not sure how much of that would be optimized away.

toolchain/lex/testdata/fail_bad_raw_identifier.carbon
```diff
@@ -5,34 +5,69 @@
...
 // CHECK:STDERR: r#á
 // CHECK:STDERR:  ^
+// CHECK:STDERR:
 r#á
-// CHECK:STDOUT:     { index: 6, kind: 'Identifier', line: {{ *}}[[@LINE-1]], column:  1, indent: 1, spelling: 'r', identifier: 0 },
-// CHECK:STDOUT:     { index: 7, kind:      'Error', line: {{ *}}[[@LINE-2]], column:  2, indent: 1, spelling: '#á', has_trailing_space: true },
+// CHECK:STDOUT:     { index:  6, kind:    'Identifier', line: {{ *}}[[@LINE-1]], column:  1, indent: 1, spelling: 'r', identifier: 0 },
+// CHECK:STDOUT:     { index:  7, kind:         'Error', line: {{ *}}[[@LINE-2]], column:  2, indent: 1, spelling: '#á', has_trailing_space: true },
+
+// Raw `r` identifier doesn't start a second raw string.
```
jonmeow: Raw string or identifier? (in general)

jonmeow (approved):
LGTM with side-comments

toolchain/lex/testdata/fail_bad_raw_identifier.carbon
zygoloid: Fixed, oops

PR #3856, opened 2024-04-04
Add core library with operator interface definitions and an example program.

zygoloid:
This is almost certainly not how we'll want these to be organized, and if the examples/ directory is kept it should have BUILD files. But this at least lets me park these files somewhere that's more global than my own checkout.


jonmeow:
What do you think of adding tests that these compile? You could have a file_test that at least ensures there aren't any errors compiling things in the "examples" directory, indirectly ensuring the Core imports work [not passing stderr/stdout to the framework, disabling autoupdate to avoid in-file checks]. Without that I'm concerned these would fall out of sync quickly [over the next two weeks].

Also, regarding "lib/core", do we generally want a "lib" directory? We also have things like "language_server" and "utils" that have library-like things, but haven't used a "lib" directory so far. I think a couple alternatives would be just "core" (because it's important enough to be top-level, and we aren't clearly subdividing libraries elsewhere) or "toolchain/core" (because it's bound to the toolchain's implementation).

zygoloid:
> What do you think of adding tests that these compile? 

As noted in the PR description, I think we should do something like that. How strongly do you want that in this PR? I'd like to approach that "properly", by adding carbon build rules to bazel, which will likely take a bit of effort.

jonmeow:
> > What do you think of adding tests that these compile?
> 
> As noted in the PR description, I think we should do something like that. How strongly do you want that in this PR? I'd like to approach that "properly", by adding carbon build rules to bazel, which will likely take a bit of effort.

I've published the draft #3861 as the approach I'm thinking of. I suppose since I'm building this anyways, it's best to avoid conflicting work. But I wasn't planning to cover the "example" directory.

Build rules may be okay while we have a very small number of examples. At a certain point though, if that's used to construct separate targets for each test which must each individually be subprocessed, I would expect similar issues as we had with `lit` performance. We need a fast test that code is correct for ourselves; I would view carbon build rules as something more aimed at end users.

If you really want to get this in, I can approve.

lib/core/i32.carbon
```diff
@@ -0,0 +1,130 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
```
jonmeow: Directory comment: maybe core/prelude/i32.carbon? You already have my `core/` comment on #naming, but I'll note here that I was also thinking maybe splitting off all the `prelude` files to a subdir (since we'll presumably have both libraries that users must import manually, and prelude libraries that are implicit)

jonmeow (approved):

PR #3857, opened 2024-04-04
Add functions for instantiating builtin types through lookup.

jonmeow:
Note, my instinct is that `Float(dyn_size)` should be invalid. However, I think the constant evaluation doesn't result in the call being evaluated in eval.cpp when the size is non-constant. I think I could get an error for symbolic phase calls, but that seems a little less interesting already. Long-term maybe we want a way to mark functions as _must_ be evaluated during constant phase?

Also, I think there may be a bug with literal value parse node locations, I should be able to point at the position of `arg_ids[0]` but it's missing a line number so I point at `loc` instead.


toolchain/check/eval.cpp
```diff
@@ -302,6 +302,20 @@ static auto PerformAggregateIndex(Context& context, SemIR::Inst inst)
...
+// Enforces that the byte width is 64 for a float.
+static auto ValidateFloatByteWidth(Context& context, SemIRLoc loc,
+                                   SemIR::InstId inst_id) -> bool {
+  auto inst = context.insts().GetAs<SemIR::IntLiteral>(inst_id);
+  if (context.ints().Get(inst.int_id) == 64) {
+    return true;
+  }
+
+  CARBON_DIAGNOSTIC(CompileTimeFloatByteWidth, Error,
+                    "Byte width must be 64.");
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  CARBON_DIAGNOSTIC(CompileTimeFloatByteWidth, Error, "Byte width must be 64.");
```

toolchain/check/eval.cpp
```diff
@@ -302,6 +302,19 @@ static auto PerformAggregateIndex(Context& context, SemIR::Inst inst)
   return context.constant_values().Get(elements[index_val.getZExtValue()]);
 }
 
+// Enforces that the byte width is 64 for a float.
```
zygoloid: I think you mean "bit" not "byte" throughout.

toolchain/check/eval.cpp
```diff
@@ -497,6 +510,30 @@ static auto PerformBuiltinCall(Context& context, SemIRLoc loc, SemIR::Call call,
     case SemIR::BuiltinFunctionKind::None:
       CARBON_FATAL() << "Not a builtin function.";
 
+    case SemIR::BuiltinFunctionKind::IntMake32: {
+      if (phase != Phase::Template) {
+        break;
+      }
```
zygoloid: I think this check is redundant -- there are no arguments, so there's no phase to check.
```suggestion
```

toolchain/check/eval.cpp
```diff
@@ -497,6 +510,30 @@ static auto PerformBuiltinCall(Context& context, SemIRLoc loc, SemIR::Call call,
...
+      if (!ValidateFloatByteWidth(context, loc, arg_ids[0])) {
+        return SemIR::ConstantId::Error;
+      }
+      return context.constant_values().Get(SemIR::InstId::BuiltinFloatType);
+    }
+
+    case SemIR::BuiltinFunctionKind::BoolMake: {
+      if (phase != Phase::Template) {
+        break;
+      }
```
zygoloid:
```suggestion
```

toolchain/check/eval.cpp
```diff
@@ -497,6 +510,30 @@ static auto PerformBuiltinCall(Context& context, SemIRLoc loc, SemIR::Call call,
...
 
+    case SemIR::BuiltinFunctionKind::IntMake32: {
+      if (phase != Phase::Template) {
+        break;
+      }
+      return context.constant_values().Get(SemIR::InstId::BuiltinIntType);
+    }
+
+    case SemIR::BuiltinFunctionKind::FloatMake: {
+      if (phase != Phase::Template) {
```
zygoloid, line 519:
```suggestion
      // TODO: Support a symbolic constant width.
      if (phase != Phase::Template) {
```
I assume we'd eventually want to allow
```
fn F[N: i32](f: Core.Float(N))
```

toolchain/check/testdata/function/builtin/types.carbon
```diff
@@ -0,0 +1,203 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
```
zygoloid: `check/testdata/function/builtin` is for tests checking our handling of function declarations with builtin definitions in general, not for the behavior of specific builtins. For specific builtins, I've been adding one file per builtin to `check/testdata/builtins/<name>` (with `.`s in the name replaced by `/`s).

toolchain/sem_ir/builtin_function_kind.cpp
```diff
@@ -68,6 +68,9 @@ using Bool = BuiltinType<InstId::BuiltinBoolType>;
 // TODO: This only matches i32 for now. Support iN for all N, and the
 // Core.BigInt type we use to implement for integer literals.
 using AnyInt = BuiltinType<InstId::BuiltinIntType>;
+
+// Constraint that requires the type to be the type type.
+using TypeT = BuiltinType<InstId::BuiltinTypeType>;
```
zygoloid: The `T` suffix suggests to me that this would be a constraint on a generic parameter, eg`fn F[T:! type](x: T) -> T` would be `auto(TypeT) -> TypeT`, matching how `IntT` behaves. Maybe just `Type`, to match `Bool`?

toolchain/check/testdata/function/builtin/types.carbon
```diff
@@ -0,0 +1,203 @@
...
+//
+// AUTOUPDATE
+
+// --- types.carbon
+
+library "types" api;
+
+fn Int() -> type = "int.make_32";
+fn Float(size: i32) -> type = "float.make";
+fn Bool() -> type = "bool.make";
```
zygoloid: I wonder if `"float.type"` or `"float.maketype"` would be a better builtin name than `"float.make"` -- the latter sounds like a function `Float.Make` which I'd expect to produce a floating-point value, not a floating-point type.

But I'm happy with anything here; these names are just an implementation detail after all.

toolchain/check/eval.cpp
jonmeow: Fixed

toolchain/check/eval.cpp
jonmeow: Done

toolchain/check/eval.cpp
jonmeow: Done, although I would hope these are actually types before that happens.

toolchain/check/eval.cpp
jonmeow: Done

toolchain/check/testdata/function/builtin/types.carbon
jonmeow: Done, make_type due to the `make` pattern

toolchain/sem_ir/builtin_function_kind.cpp
jonmeow: Done (I just feel weird writing `Type` alone)

toolchain/check/testdata/function/builtin/types.carbon
jonmeow: Done

zygoloid (approved):

PR #3858, opened 2024-04-04
Opt out of pre-commit suggestions

jonmeow:
Due to my workflow, I prefer to run `pre-commit` locally when I forget to after an amend. While reviewdog can be helpful in the UI, it's mostly making me mark comments as resolved as part of re-pushing. Rather than continuing with this, maybe opting out is best?

### DIFF
--- a/.github/workflows/pre_commit_suggestions.yaml
+++ b/.github/workflows/pre_commit_suggestions.yaml
@@ -27,7 +27,8 @@ jobs:
     # Only generate suggestions if pre-commit for a PR failed.
     if: |
       github.event.workflow_run.conclusion == 'failure' &&
-      github.event.workflow_run.event == 'pull_request'
+      github.event.workflow_run.event == 'pull_request' &&
+      github.actor != 'jonmeow'
     runs-on: ubuntu-latest
     steps:
       - uses: reviewdog/action-setup@v1


zygoloid (approved):
I *think* reviewdog is supposed to be able to automatically update or resolve its comments if they no longer apply on rerun. Maybe we need to do something to make that work, though. Would that address your needs, or would you prefer that the comments don't appear at all?

Approving anyway; this is just intended to be a convenience to the PR author.

jonmeow:
I'd suspect to make reviewdog resolve them, it'd need to be run with no delta. i.e., that it might be setting the specific set of comments to make on each execution. Since it's not run on success, it won't clear. But note e.g. that https://github.com/reviewdog/reviewdog/issues/568 is an open issue. There's a mention of https://github.com/Ardiannn08/resolve-outdated-comment but it's not clear to me that the permissions will be right.

I do worry about a situation where I really miss a bunch of pre-commit things and there are a couple dozen things to fix. Then even if I fix it up, it's sort of just noise on the review thread. So better, maybe enough that the noise is better than special-casing.

zygoloid:
Yeah, also see https://github.com/reviewdog/reviewdog/issues/1150 -- it looks like I was misremembering, and this is just a requested feature not an implemented one.

PR #3861, opened 2024-04-04
Start including the prelude for testing.

jonmeow:
- Adds an empty prelude.carbon file
- Imports that file in any non-Core package file
  - Adds --disable-prelude-import to avoid that
- Adds --exclude-dump-file-prefix to be able to hide files from dumping
   - Used to hide core files (we can't do this by package name due to lex dumps, for example)
- Restructures some tests to not rely on `i32`, particularly `alias` tests (which rely on a name ref) and tests with no prelude.

I'm adding the framework for switching i32 to calling Int32 in the prelude, but I'm running into a separate error actually switching over. So that *mostly* works, but isn't quite ready for prime time. However, maybe the current state of this PR is still useful to review since it does a lot of the infrastructure work and adds the %Core everywhere?


core/prelude/prelude.carbon
```diff
@@ -0,0 +1,13 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+
+package Core library "prelude" api;
```
zygoloid: I think it'd be a nice rule for the library name to be the file name without the `.carbon` suffix, so perhaps this file should be `core/prelude.carbon` instead of `core/prelude/prelude.carbon`?

toolchain/check/check.cpp
```diff
@@ -1047,17 +1048,31 @@ auto CheckParseTrees(const SemIR::File& builtin_ir,
...
+      auto implicit_names = packaging->names;
+      implicit_names.package_id = IdentifierId::Invalid;
+      TrackImport(api_map, nullptr, unit_info, implicit_names);
     }
 
     llvm::DenseMap<ImportKey, Parse::NodeId> explicit_import_map;
+
+    // Add the prelude import. It's added to explicit_import_map so that it can
+    // conflict with an explicit import of the prelude.
+    // TODO: Add --disable-prelude-import for `/no_prelude/` subdirs.
```
zygoloid: Presumably this is just for tests?
```suggestion
    // TODO: Add --disable-prelude-import for `/no_prelude/` subdirs in testdata.
```

toolchain/check/testdata/alias/fail_local_in_namespace.carbon
```diff
@@ -6,7 +6,11 @@
 
 namespace NS;
 
-fn F() -> bool {
+// CHECK:STDERR: fail_local_in_namespace.carbon:[[@LINE+4]]:11: ERROR: Name `C` not found.
+// CHECK:STDERR: fn F() -> C {
+// CHECK:STDERR:           ^
+// CHECK:STDERR:
+fn F() -> C {
```
zygoloid:
```suggestion
class C {}

fn F() -> C {
```
or maybe
```suggestion
fn F() -> {} {
```
(I assume you didn't intend to make the declaration of `F` invalid here, given that this is a test of alias declarations.)

toolchain/check/testdata/index/fail_negative_indexing.carbon
```diff
@@ -5,6 +5,10 @@
 // AUTOUPDATE
 
 var a: (i32, i32) = (12, 6);
+// CHECK:STDERR: fail_negative_indexing.carbon:[[@LINE+7]]:16: ERROR: Name `Negate` not found.
+// CHECK:STDERR: var b: i32 = a[-10];
+// CHECK:STDERR:                ^~~
+// CHECK:STDERR:
 // CHECK:STDERR: fail_negative_indexing.carbon:[[@LINE+3]]:16: ERROR: Semantics TODO: `missing or invalid operator interface`.
```
zygoloid: Maybe add a TODO for the diagnostic here?
```suggestion
// TODO: Only diagnose this once, and make the diagnostic clearer.
// CHECK:STDERR: fail_negative_indexing.carbon:[[@LINE+7]]:16: ERROR: Name `Negate` not found.
// CHECK:STDERR: var b: i32 = a[-10];
// CHECK:STDERR:                ^~~
// CHECK:STDERR:
// CHECK:STDERR: fail_negative_indexing.carbon:[[@LINE+3]]:16: ERROR: Semantics TODO: `missing or invalid operator interface`.
```

zygoloid (approved):
I think we should establish a convention of using `--no-blah` for negated boolean flags, so `--no-prelude-import`. But I'm happy for this to land without that change, and for us to discuss that separately.

core/prelude/prelude.carbon
jonmeow: Done for now, but I'll note the way that `operators`/`types` versus `operators.carbon` is presented is confusing me and feels inconsistent. I think I would advocate towards either having a directory or a main file within the directory: `prelude/operators/<something>.carbon`, not `prelude/operators.carbon` + `prelude/operators/*`. [i.e., closer to index.html, README.md, or //bazel/target]

toolchain/check/check.cpp

toolchain/check/testdata/alias/fail_local_in_namespace.carbon
jonmeow: Fixed, didn't mean to change this substantially since the `false` failure was irrelevant to these changes.

toolchain/check/testdata/index/fail_negative_indexing.carbon
jonmeow: Amended the context.TODO

PR #3865, opened 2024-04-05
Import all `impl`s in imported IRs.

zygoloid:
This allows impl lookup to find such impls.

Eventually we'll want to do this more lazily, and filter to the relevant subset of `impl`s needed for a query. But for simplicity, for now just import all `impl`s.

### DIFF
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -16,6 +16,7 @@
 #include "toolchain/check/function.h"
 #include "toolchain/check/handle.h"
 #include "toolchain/check/import.h"
+#include "toolchain/check/import_ref.h"
 #include "toolchain/diagnostics/diagnostic.h"
 #include "toolchain/diagnostics/diagnostic_emitter.h"
 #include "toolchain/lex/token_kind.h"
@@ -754,6 +755,10 @@ static auto CheckParseTree(
   context.inst_block_stack().Push();
 
   InitPackageScopeAndImports(context, unit_info);
+
+  // Import all impls declared in imports.
+  // TODO: Do this selectively when we see an impl query.
+  ImportImpls(context);
 
   if (!ProcessNodeIds(context, vlog_stream, unit_info.err_tracker)) {
     context.sem_ir().set_has_errors(true);
--- a/toolchain/check/import_ref.cpp
+++ b/toolchain/check/import_ref.cpp
@@ -1032,11 +1032,11 @@ static auto ImportImpl(Context& context, SemIR::ImportIRId import_ir_id,
   // Import the definition if the impl is defined.
   // TODO: Do we need to check for multiple definitions?
   auto impl_id = context.impls().LookupOrAdd(self_id, constraint_id);
-  auto& impl = context.impls().Get(impl_id);
   if (import_impl.is_defined()) {
     // TODO: Create a scope for the `impl` if necessary.
     // TODO: Consider importing the definition_id.
 
+    auto& impl = context.impls().Get(impl_id);
     impl.witness_id = context.AddImportRef(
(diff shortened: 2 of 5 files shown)


toolchain/check/import_ref.cpp
```diff
@@ -987,4 +1019,44 @@ auto LoadImportRef(Context& context, SemIR::InstId inst_id, SemIRLoc loc)
...
+
+  // Import the definition if the impl is defined.
+  // TODO: Do we need to check for multiple definitions?
+  auto impl_id = context.impls().LookupOrAdd(self_id, constraint_id);
+  auto& impl = context.impls().Get(impl_id);
+  if (import_impl.is_defined()) {
+    // TODO: Create a scope for the `impl` if necessary.
+    // TODO: Consider importing the definition_id.
+
+    impl.witness_id = context.AddImportRef(
```
```suggestion
  if (import_impl.is_defined()) {
    // TODO: Create a scope for the `impl` if necessary.
    // TODO: Consider importing the definition_id.

    auto& impl = context.impls().Get(impl_id);
    impl.witness_id = context.AddImportRef(
```

toolchain/check/import_ref.cpp
```diff
@@ -987,4 +1019,44 @@ auto LoadImportRef(Context& context, SemIR::InstId inst_id, SemIRLoc loc)
...
+        {.ir_id = import_ir_id, .inst_id = import_impl.witness_id});
+  }
+}
+
+// TODO: This doesn't belong in this file. Consider moving the import resolver
+// and this file elsewhere.
+auto ImportImpls(Context& context) -> void {
```
jonmeow, line 1047: Should there be a TODO for handling indirect imports?

i.e.:

```
library "a" api;

interface HasF {
  fn F();
}
```

```
library "b" api;

import library "a";

class C {}
impl C as HasF {
  fn F() {}
}
```

```
library "c" api;

import library "b";

alias D = C;
```

```
library "d" api;

import library "a";
import library "c";

fn G(c: D) {
  c.(HasF.F)();
}
```

toolchain/check/check.cpp
```diff
@@ -755,6 +756,10 @@ static auto CheckParseTree(
 
   InitPackageScopeAndImports(context, unit_info);
 
+  // Import all impls declared in imports.
+  // TODO: Do this selectively when we see an impl query.
+  ImportImpls(context);
```
jonmeow, line 761: Just to note, it might be worth considering a similar kind of ImportRefUnused approach long-term. i.e., maybe adding impls to exports, rather than scanning this way (can there be private impls?). I understand that you're maybe going to want to find impls close to the type in question, but maybe this late-import model isn't needed for that.

It just looks odd to me since I think we probably don't want to do other import logic quite this way, although maybe it's right for impls. High-level comment, not for this PR.

jonmeow (approved):

toolchain/check/import_ref.cpp
zygoloid: The `impl` handling here should already work: library "c" will contain the `impl C as HasF` too. But it doesn't work for unrelated reasons -- library "d" sees two different `HasF` interfaces, one from library "a" and one from library "c" (copied from "b", copied from "a") -- so I don't think I can write a test that checks this, unfortunately.

toolchain/check/check.cpp
zygoloid: Maybe this viewpoint will make it look a little less odd in the short term.: I'm thinking of pulling in an `Impl` as being analogous to pulling in a name lookup table entry in a name scope -- we import just the bits we need for lookup (the type and constraint here, versus the name in a name scope), and add an `ImportRefUnused` for the rest (the witness here, the lookup result in a name scope).

But in any case, I definitely want to consider other options long-term.

PR #3866, opened 2024-04-05
Fix missing return slot assignment

jonmeow:
Per your comment on #toolchain. Note, not sure what a test would look like.

### DIFF
--- a/toolchain/check/import_ref.cpp
+++ b/toolchain/check/import_ref.cpp
@@ -715,7 +715,7 @@ class ImportRefResolver {
     if (function.return_slot_id.is_valid()) {
       auto import_ir_inst_id = context_.import_ir_insts().Add(
           {.ir_id = import_ir_id_, .inst_id = function.return_slot_id});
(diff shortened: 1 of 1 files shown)


zygoloid (approved):
Maybe you'd be able to see a difference in a lower test, eg

```
// --- a.carbon
package A api;
class C {
  var x: [i32: 8];
}
fn F() -> C;

// --- b.carbon
import A;
fn G() {
  var c: A.C = A.F();
}
```

If not, maybe this doesn't affect anything right now, in which case let's just merge this.

jonmeow:
> Maybe you'd be able to see a difference in a lower test, eg
> 
> ```
> // --- a.carbon
> package A api;
> class C {
>   var x: [i32: 8];
> }
> fn F() -> C;
> 
> // --- b.carbon
> import A;
> fn G() {
>   var c: A.C = A.F();
> }
> ```
> 
> If not, maybe this doesn't affect anything right now, in which case let's just merge this.

Code fails due to lack of support (VarStorage in general). I think things [import inst support] just aren't far enough yet for me to get a clean repro.

PR #3868, opened 2024-04-05
Add support for `iN` and `uN` for all suitable `N`.

zygoloid:
`i32` is retained as a special case for now, for bootstrapping purposes, and maps to `BuiltinIntType`, which is distinct from `Core.Int(32)`. This will be removed later once we support `Core.BigInt`.

For now this provides both the `iN` types and also the builtins to support `Core.Int(N)`. The intent is that we'll change the `iN` support to rewrite to calls here when we do that for the other type literals and type keywords.

No conversions between integer types are supported yet, and all literals are of type `i32`, so we can't actually form values of any of these new types.


toolchain/sem_ir/ids.h
```diff
@@ -305,6 +305,32 @@ struct BoolValue : public IdBase, public Printable<BoolValue> {
...
+// This might eventually capture any other properties of an integer type that
+// affect its semantics, such as overflow behavior.
+struct IntKind : public IdBase, public Printable<IntKind> {
+  static const IntKind Unsigned;
+  static const IntKind Signed;
+
+  // Returns whether this type is signed.
+  constexpr auto is_signed() -> bool { return *this == Signed; }
+
+  using IdBase::IdBase;
```
```suggestion
  using IdBase::IdBase;

    // Returns whether this type is signed.
  constexpr auto is_signed() -> bool { return *this == Signed; }

```

Since IdBase is a constructor...

toolchain/sem_ir/typed_insts.h
```diff
@@ -567,6 +567,17 @@ struct IntLiteral {
...
 
+struct IntType {
+  static constexpr auto Kind =
+      InstKind::IntType.Define<Parse::NodeId>("int_type");
+
+  TypeId type_id;
+  IntKind int_kind;
+  // TODO: Consider adding a more compact way of representing either a small
+  // unsigned integer bit width or an inst_id.
+  InstId bit_width_id;
```
jonmeow, line 578: Why InstId instead of an IntId? (I think it's required to be a literal, in essence?)

toolchain/lower/handle.cpp
```diff
@@ -260,15 +262,23 @@ static auto HandleBuiltinCall(FunctionContext& context, SemIR::InstId inst_id,
       return;
     }
     case SemIR::BuiltinFunctionKind::IntDiv: {
-      context.SetLocal(inst_id, context.builder().CreateSDiv(
-                                    context.GetValue(arg_ids[0]),
-                                    context.GetValue(arg_ids[1]), "div"));
+      context.SetLocal(inst_id, IsSignedInt(context, inst_id)
```
jonmeow, line 265: FWIW, I'm a little surprised you don't either make Uint* for all the operations, or more generally do a BuiltinDiv that accepts whatever types (since it could morph based on arguments, as here)

toolchain/check/eval.cpp
```diff
@@ -302,6 +302,56 @@ static auto PerformAggregateIndex(Context& context, SemIR::Inst inst)
...
+                           TypedInt{bit_width->type_id, bit_width_val});
+    return false;
+  }
+  // TODO: Pick a maximum size and document it in the design. For now
+  // we use 2^^23, because that's the largest size that LLVM supports.
+  constexpr int MaxIntWidth = 1 << 23;
+  if (bit_width_val.ugt(MaxIntWidth)) {
+    CARBON_DIAGNOSTIC(IntWidthTooLarge, Error,
+                      "Integer type width of {0} is greater than the "
+                      "maximum supported size of {1}.",
```
```suggestion
                      "maximum supported width of {1}.",
```

Seems like "width" should be used in both places in the diagnostic?

toolchain/check/handle_literal.cpp
```diff
@@ -2,7 +2,10 @@
 // Exceptions. See /LICENSE for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+#include <cwchar>
+
```
jonmeow:
```suggestion
```

Am I missing a use, or was this added due to some intermediate change?

toolchain/check/handle_literal.cpp
```diff
@@ -73,21 +83,49 @@ auto HandleBoolTypeLiteral(Context& context, Parse::BoolTypeLiteralId node_id)
   return true;
 }
 
+static auto HandleIntOrUnsignedIntTypeLiteral(Context& context,
```
jonmeow, line 90: Is it worth remarking on the semantics of this?

jonmeow (approved):

jonmeow:
LGTM, just a few small comments.

toolchain/sem_ir/typed_insts.h
zygoloid: We support `Core.Int(N)` where `N` is a symbolic expression. That will eventually be important for things like `impl forall [N:! BigInt] Core.Int(N) as Add`, but isn't much use right now given the general lack of implementation of generics.

toolchain/lower/handle.cpp
zygoloid: There are definitely some other options we could pursue here. I'm going to land this as-is because I don't think any of them is obviously substantially better than the rest (and this patch is just fixing a bug here; most of the builtins already dispatched on signedness), but I think we should keep discussing this especially as we add the bigint type and floating-point support.

Some thoughts on specific approaches:

- Current approach: it's a bit weird that we have builtins with different semantics (eg, overflow behavior) with the same name.
- Provide an `Int*` set and a `UInt*` set of builtins: we'd end up with some redundancy for the operations that are identical for signed and unsigned types (eg, bitwise operations).
- Provide a more general `BuiltinDiv` that also supports floating-point types too: the implementation of that (especially in `eval.cpp`) would be substantially different in that case, because the types have a different representation, whereas it's pretty easy to handle all the integer types together, with just a small branch for signed vs unsigned operations in the cases where it matters.
- Provide different signed builtins for operations where it matters, eg add `int.sdiv` and `int.udiv` but keep a single `int.and`, and leave it up to the library to map the operations to the right type based on signedness: maybe this would be a nice option to pursue, but would add a bunch of builtin names. Would also support providing (eg) a wrapping add on signed integers as a library function, if desired. Matches LLVM better.

The last of those options seems pretty tempting to me.

toolchain/check/handle_literal.cpp
zygoloid: Yeah, more vs code automatic wrong `#include` insertion.

toolchain/check/handle_literal.cpp
zygoloid: Sure.
```suggestion
// Shared implementation for handling `iN` and `uN` literals.
static auto HandleIntOrUnsignedIntTypeLiteral(Context& context,
```

toolchain/sem_ir/ids.h
```diff
@@ -305,6 +305,33 @@ struct BoolValue : public IdBase, public Printable<BoolValue> {
...
+//
+// This might eventually capture any other properties of an integer type that
+// affect its semantics, such as overflow behavior.
+struct IntKind : public IdBase, public Printable<IntKind> {
+  static const IntKind Unsigned;
+  static const IntKind Signed;
+
+  using IdBase::IdBase;
+
+    // Returns whether this type is signed.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  // Returns whether this type is signed.
```

toolchain/lower/handle.cpp
chandlerc: +1 to the last of those options from me FWIW -- mirroring the LLVM model for the builtins makes a lot of sense to me.

PR #3873, opened 2024-04-09
Switch check code to use CARBON_KIND_SWITCH

jonmeow:
This is intended to have no user-visible effect, only aiming for a style change/simplification.


chandlerc (approved):
LG!

PR #3874, opened 2024-04-09
Basic check support for adapters.

zygoloid:
No support is provided yet for converting between a type and an adapter for that type.


toolchain/sem_ir/typed_insts.h
```diff
@@ -189,6 +189,15 @@ struct AssociatedEntityType {
   TypeId entity_type_id;
 };
 
+// An adapted type declaration in a class, of the form `adapt T;`.
+struct AdaptDecl {
```
jonmeow: Order relative to other declarations? (before Assoc*?)

toolchain/check/handle_class.cpp
```diff
@@ -172,15 +183,79 @@ auto HandleClassDefinitionStart(Context& context,
...
-auto HandleAdaptDecl(Context& context, Parse::AdaptDeclId /*node_id*/) -> bool {
-  auto adapted_expr_id = context.node_stack().PopExpr();
-  (void)adapted_expr_id;
-  // TODO: Process the `adapt` declaration.
+auto HandleAdaptDecl(Context& context, Parse::AdaptDeclId node_id) -> bool {
+  auto [adapted_type_node, adapted_type_expr_id] =
+      context.node_stack().PopExprWithNodeId();
+
+  // Process modifiers. `extend` is permitted, no others are allowed.
+  LimitModifiersOnDecl(context, KeywordModifierSet::Extend,
```
jonmeow, line 239: For other declarations, we have fail_modifiers.carbon testing various invalid modifiers. I'm not noticing that here.

toolchain/check/handle_class.cpp
```diff
@@ -172,15 +183,79 @@ auto HandleClassDefinitionStart(Context& context,
...
+    context.emitter().Emit(node_id, AdaptOutsideClass);
+    return true;
+  }
+
+  auto& class_info = context.classes().Get(enclosing_class_decl->class_id);
+  if (class_info.adapt_id.is_valid()) {
+    CARBON_DIAGNOSTIC(AdaptRepeated, Error,
+                      "Multiple `adapt` declarations in class.");
+    CARBON_DIAGNOSTIC(AdaptPrevious, Note,
+                      "Previous `adapt` declaration is here.");
```
jonmeow: FWIW, we probably have equivalent diagnostics for `base`, it might be worth parameterizing the keyword with the Lex::TokenKind (adapt or base) rather than having a separate diagnostic, given these are closely related.

Other bits of this (incomplete type checking as another example) might be similarly possible to share.

jonmeow (approved):
Looks good!

toolchain/check/testdata/class/fail_adapt_modifiers.carbon
```diff
@@ -0,0 +1,110 @@
...
+}
+
+class C4 {
+  // CHECK:STDERR: fail_adapt_modifiers.carbon:[[@LINE+6]]:10: ERROR: `extend` repeated on declaration.
+  // CHECK:STDERR:   extend extend adapt B;
+  // CHECK:STDERR:          ^~~~~~
+  // CHECK:STDERR: fail_adapt_modifiers.carbon:[[@LINE+3]]:3: `extend` previously appeared here.
+  // CHECK:STDERR:   extend extend adapt B;
+  // CHECK:STDERR:   ^~~~~~
+  extend extend adapt B;
```
jonmeow, line 41: `extend base adapt` should probably be tested due to the special-ness of `base`. :)

toolchain/sem_ir/typed_insts.h
zygoloid: Done.

toolchain/check/handle_class.cpp
zygoloid: Done.

toolchain/check/handle_class.cpp
zygoloid: Done in a couple of places. This one is a bit awkward because we have extra text in the base case, but I think it's still worth factoring out.

toolchain/check/handle_class.cpp
```diff
@@ -345,23 +368,15 @@ auto HandleBaseDecl(Context& context, Parse::BaseDeclId node_id) -> bool {
 
   auto enclosing_class_decl = context.GetCurrentScopeAs<SemIR::ClassDecl>();
   if (!enclosing_class_decl) {
-    CARBON_DIAGNOSTIC(BaseOutsideClass, Error,
-                      "`base` declaration can only be used in a class.");
-    context.emitter().Emit(node_id, BaseOutsideClass);
+    DiagnoseClassSpecificDeclOutsideClass(context, node_id,
+                                          Lex::TokenKind::Base);
     return true;
   }
```
jonmeow: Note the wrapping calls are also similar.

toolchain/check/handle_class.cpp
jonmeow: Yeah, we do similar elsewhere, although maybe we should provide a better way to achieve it. You might consider llvm::StringLiteral instead of std::string though, given it's constant.

toolchain/check/testdata/class/fail_adapt_modifiers.carbon
zygoloid: Done.

toolchain/check/handle_class.cpp
zygoloid: Factored that out too.

PR #3875, opened 2024-04-09
Allow incomplete return types in function declarations.

zygoloid:
Move completeness check to the point where the function is defined or first called. This means we also defer deciding whether the function has a return slot until that point. Instead of storing a return slot per function, store the location of the return storage, which may or may not be used, and compute and store a separate flag saying whether to use it at the point of first use or definition.

This is the final piece in supporting simple `Make` functions in classes as a replacement for constructors.


toolchain/sem_ir/function.h
```diff
@@ -52,14 +76,30 @@ struct Function : public Printable<Function> {
...
+  // function, depending on whether the return type needs a return slot.
+  InstId return_storage_id;
   // Whether the declaration is extern.
   bool is_extern;
 
+  // The following member is set on the first call to the function, or at the
+  // point where the function is defined.
+
+  // Whether the function uses a return slot.
+  enum class ReturnSlot : int8_t {
```
jonmeow: Declaration order. I know mixed feelings on this, but I think we should be following the style guide in spots like this.

toolchain/check/import_ref.cpp
```diff
@@ -733,8 +730,10 @@ class ImportRefResolver {
          .param_refs_id =
              GetLocalParamRefsId(function.param_refs_id, param_const_ids),
          .return_type_id = new_return_type_id,
-         .return_slot_id = new_return_slot,
+         .return_storage_id = new_return_storage,
          .is_extern = function.is_extern,
+         // There has been no call or definition in this file.
+         .return_slot = SemIR::Function::ReturnSlot::NotComputed,
```
jonmeow: Should this be forwarded when Absent or similar? If it's an Error in an import, should it become valid when imported? i.e., if known from the API, I'm not confident it should be ignored.

As an example test:

```
library "a" api;
class C;
// Diagnoses incomplete type, right?
fn F() -> C;
```

```
library "b" api;
import library "a";

fn G() {
  // Will this diagnose incomplete?
  F();
}
```

```
library "c" api;
import library "a";

class C {}

fn H() {
  // Does this get a different result for the return slot?
  F();
}
```

Note, that'd also have implications for merging, since it's not part of the prev_function merge right now.

toolchain/check/function.cpp
```diff
@@ -327,4 +327,42 @@ auto MergeFunctionRedecl(Context& context, SemIR::LocId loc_id,
...
+  // Check the return type is complete. Only diagnose incompleteness if we've
+  // not already done so.
+  auto diagnose_incomplete_return_type = [&] {
+    CARBON_DIAGNOSTIC(IncompleteTypeInFunctionReturnType, Error,
+                      "Function returns incomplete type `{0}`.", SemIR::TypeId);
+    return context.emitter().Build(function.return_storage_id,
+                                   IncompleteTypeInFunctionReturnType,
+                                   function.return_type_id);
+  };
+  std::optional<llvm::function_ref<auto()->Context::DiagnosticBuilder>>
```
jonmeow: Would it be worth a typedef on Context for this?

toolchain/check/function.cpp
```diff
@@ -315,7 +315,7 @@ auto MergeFunctionRedecl(Context& context, SemIR::LocId loc_id,
     prev_function.implicit_param_refs_id = new_function.implicit_param_refs_id;
     prev_function.param_refs_id = new_function.param_refs_id;
     prev_function.return_type_id = new_function.return_type_id;
-    prev_function.return_slot_id = new_function.return_slot_id;
+    prev_function.return_storage_id = new_function.return_storage_id;
```
jonmeow, line 340: Would it be worth asserting new_function.return_slot is NotComputed? (tieing in with my other comment that maybe it shouldn't always be NotComputed)

jonmeow (approved):
Approving, though not sure about the behavior on import.

toolchain/check/import_ref.cpp
zygoloid: When the return type is incomplete, the `Error` state gets re-checked on every call anyway; the difference between `NotChecked` and `Error` is that `Error` suppresses diagnostics. So I think it's just a question of whether we want a diagnostic per file, or one in total.

Per in-person discussion, more error suppression is probably better, so changed this to import and merge the return slot information.

toolchain/check/function.cpp
zygoloid: Changed to merge the slot info.

toolchain/check/function.cpp
zygoloid: Rearranged to avoid the need for this type.

toolchain/sem_ir/function.h
zygoloid: Sure, done.

jonmeow (approved):
Thanks, all the changes look great.

jonmeow:
Might be a merge conflict to investigate -- tests look like they need to be updated.

PR #3877, opened 2024-04-10
Move the injection of the prelude compile to the driver.

chandlerc:
The `file_test` layer still adds it to the VFS and filters it from the dump output, but the driver now injects it into the compilation unites depending on the `--import-prelude` file (enabled by default).

This doesn't address improvements to how we find the prelude code, just sinking the needed logic into the driver itself.

### DIFF
--- a/toolchain/driver/driver.cpp
+++ b/toolchain/driver/driver.cpp
@@ -671,7 +671,17 @@ auto Driver::Compile(const CompileOptions& options) -> RunResult {
   // Prepare CompilationUnits before building scope exit handlers.
   StreamDiagnosticConsumer stream_consumer(error_stream_);
   llvm::SmallVector<std::unique_ptr<CompilationUnit>> units;
-  units.reserve(options.input_filenames.size());
+
+  // Directly insert the core package into the compilation units.
+  // TODO: Should expand this into a more rich system to search for the core
+  // package source code.
+  if (options.prelude_import) {
+    units.push_back(std::make_unique<CompilationUnit>(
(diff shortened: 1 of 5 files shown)


toolchain/driver/driver.cpp
```diff
@@ -671,7 +671,17 @@ auto Driver::Compile(const CompileOptions& options) -> RunResult {
...
+  // Directly insert the core package into the compilation units.
+  // TODO: Should expand this into a more rich system to search for the core
+  // package source code.
+  if (options.prelude_import) {
+    units.push_back(std::make_unique<CompilationUnit>(
+        this, options, &stream_consumer, "core/prelude.carbon"));
+  }
+
+  // Add the input source files.
+  units.reserve(units.size() + options.input_filenames.size());
```
```suggestion
    units.reserve(options.input_filenames.size() + 1);
    units.push_back(std::make_unique<CompilationUnit>(
        this, options, &stream_consumer, "core/prelude.carbon"));
  } else {
    units.reserve(options.input_filenames.size());
  }

  // Add the input source files.
```

Thoughts on switching to reserve prior to pushing?

jonmeow (approved):
LGTM

toolchain/driver/driver.cpp
chandlerc: Sure, used a slightly different code pattern. PTAL.

jonmeow: We'll probably want to switch (we will swiftly have more than one prelude file) but this is fine for now.

PR #3879, opened 2024-04-11
Use https://app.stepsecurity.io/secureworkflow to improve workflow security.

jonmeow:
This is coming out of advice from https://securityscorecards.dev/viewer/?uri=github.com/carbon-language/carbon-lang

Updates action versions and pins by checksum. This is something we already do in other places, and doing it with actions just seems consistent.

Trying to add token permissions at the same time.

step-security/harden-runner is being set up to monitor network traffic for actions, with the idea that we'd set it to block unexpected traffic in the future.

Note, https://app.stepsecurity.io/secureworkflow generates the pinned checksums and adds the harden-runner. In some cases it added token permissions, but we have a couple it doesn't recognize (`gh` executions, other custom commands, jlumbroso/free-disk-space, reviewdog/action-setup as examples) so I'm trying to guess my best.

Because these are workflows, testing is limited.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -10,6 +10,10 @@ on:
   pull_request:
   merge_group:
 
+permissions:
+  contents: read # For actions/checkout.
+  pull-requests: read # For dorny/paths-filter to read pull requests.
+
 # Cancel previous workflows on the PR when there are multiple fast commits.
 # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency
 concurrency:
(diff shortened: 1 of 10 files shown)


chandlerc (approved):
LGTM, nice!

.github/workflows/tests.yaml
```diff
@@ -39,43 +43,46 @@ jobs:
     steps:
       # Ubuntu images start with 23GB available, and this adds 14GB more. For
       # comparison, MacOS images have >100GB free.
+      #
+      # Although we could delete more, if we run into a limit, not deleting
+      # everything provides a little flexibility to get space while trying
+      # to shrink the build.
```
zygoloid, line 54: Should we keep this comment next to the action it refers to?

zygoloid (approved):

.github/workflows/tests.yaml
jonmeow: Fixed

PR #3880, opened 2024-04-12
Update calls to std::fstream constructors

rupprecht:
Constructing types like `std::ifstream` with `std::string_view` is no longer allowed; one must pass a different type, such as `std::string`. A recent libc++ change requires this: https://github.com/llvm/llvm-project/commit/4761e74a276ee1f38596f4849daa9d633929f2ae


chandlerc (approved):
Thanks for the fix! LG and merging!

PR #3882, opened 2024-04-15
Detect the current build mode for toolchain autoupdate.

jonmeow:
Particularly with the changes to file_test, I believe it's faster to use the current build mode than to rebuild in opt. This was particularly noticeable while working on my laptop...

### DIFF
--- a/toolchain/autoupdate_testdata.py
+++ b/toolchain/autoupdate_testdata.py
@@ -19,7 +19,14 @@ def main() -> None:
     # `bazel-bin`.
     build_mode = "fastbuild"
     workspace = subprocess.check_output(
-        ["bazel", "info", "workspace"], encoding="utf-8"
+        [
+            "bazel",
+            "info",
+            "workspace",
+            # Unlike below, we need stdout.
+            "--ui_event_filters=-info,-stderr,-finish",
+        ],
+        encoding="utf-8",
     ).strip()
     bazel_bin_path = Path(workspace).joinpath("bazel-bin")
     if bazel_bin_path.exists():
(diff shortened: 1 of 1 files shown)


zygoloid (approved):

toolchain/autoupdate_testdata.py
```diff
@@ -8,17 +8,33 @@
...
 from pathlib import Path
 
 
 def main() -> None:
+    # Use the most recently used build mode, or `fastbuild` if missing
+    # `bazel-bin`.
+    build_mode = "fastbuild"
+    workspace = subprocess.check_output(
+        ["bazel", "info", "workspace"], encoding="utf-8"
+    ).strip()
```
zygoloid: For me this produces two lines of output (an `INFO: Invocation ID: ...` line and a path). Do we need:
```suggestion
    workspace = subprocess.check_output(
        ["bazel", "info", "workspace", "--ui_event_filters=-info"], encoding="utf-8"
    ).strip()
```
to turn off the `INFO` logging?

zygoloid: Looks like the INFO line goes to stderr, but it's still extra noise that we generally turn off in this script.

jonmeow: It works because this is reading stdout, and that's on stderr. Is it important to disable? (I'm hesitant to turn off output in case there's an issue)

zygoloid: So far in this script we've been turning off extra noise like this -- see the `ui_event_filters` flag below.

jonmeow: Oh, I hadn't even noticed that. Not really a fan of turning this stuff off, but adding for consistency.

zygoloid (approved):

PR #3884, opened 2024-04-15
Refactor pre-merge redeclaration checking for sharing.

jonmeow:
I'm poking at adding similar validation for `class` merging; this refactors support so that it can easily be shared.

### DIFF
--- a/toolchain/check/merge.cpp
+++ b/toolchain/check/merge.cpp
@@ -12,6 +12,135 @@
 
 namespace Carbon::Check {
 
+CARBON_DIAGNOSTIC(RedeclPrevDecl, Note, "Previously declared here.");
+
+// Diagnoses a redundant redeclaration diagnostic.
+static auto DiagnoseRedeclRedundant(Context& context, Lex::TokenKind decl_kind,
+                                    SemIR::NameId name_id, SemIRLoc new_loc,
+                                    SemIRLoc prev_loc) {
+  CARBON_DIAGNOSTIC(RedeclRedundant, Error,
+                    "Redundant redeclaration of `{0} {1}`.", Lex::TokenKind,
+                    SemIR::NameId);
+  context.emitter()
+      .Build(new_loc, RedeclRedundant, decl_kind, name_id)
+      .Note(prev_loc, RedeclPrevDecl)
+      .Emit();
+}
+
+// Diagnoses a redefinition diagnostic.
+static auto DiagnoseRedeclRedef(Context& context, Lex::TokenKind decl_kind,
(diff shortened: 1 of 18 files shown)


jonmeow:
Depends on #3883

toolchain/check/testdata/function/definition/extern.carbon
```diff
@@ -19,7 +19,7 @@ extern fn F() {}
 library "def_for_extern_decl" api;
 
 extern fn F();
-// CHECK:STDERR: fail_def_for_extern_decl.carbon:[[@LINE+7]]:1: ERROR: Redeclaring `extern` function `F` as non-`extern`.
+// CHECK:STDERR: fail_def_for_extern_decl.carbon:[[@LINE+7]]:1: ERROR: Defining `extern fn F` is disallowed.
```
geoffromer: Is this change in the phrasing intentional? I think I'd have a hard time understanding the new version if I encountered it in ordinary usage.

toolchain/check/merge.cpp
```diff
@@ -12,6 +12,135 @@
 
 namespace Carbon::Check {
 
+CARBON_DIAGNOSTIC(RedeclPrevDecl, Note, "Previously declared here.");
+
+// Diagnoses a redundant redeclaration diagnostic.
```
geoffromer:
```suggestion
// Diagnoses a redundant redeclaration.
```

toolchain/check/merge.cpp
```diff
@@ -12,6 +12,135 @@
...
+static auto DiagnoseRedeclRedundant(Context& context, Lex::TokenKind decl_kind,
```
geoffromer: Optional: the original name `DiagnoseRedundantRedecl` reads more naturally to me (notice how it corresponds more directly to the comment, for example).

jonmeow: Dropped "redecl", since all of these are redeclaration diagnostics. In this case, I'm putting more weight on keeping the names of diagnostics consistent cross-diagnostic (trying to highlight the cluster of redeclaration diagnostics), and the function names consistent with diagnostics.

toolchain/check/merge.cpp
jonmeow: Done.

toolchain/check/testdata/function/definition/extern.carbon
jonmeow: The previous phrasing was bothering me because it didn't read as well for me when working with `class`. Maybe collapsing this diagnostic into the other `extern` check will work better?

geoffromer (approved):

PR #3885, opened 2024-04-15
Add a minimal SECURITY.md

jonmeow:
Also mentioned on DIscord [here](https://discord.com/channels/655572317891461132/707150492370862090/1229512261832409238) for context.

The template may seen be people with commit access at https://github.com/carbon-language/carbon-lang/security/policy, by clicking "Start setup". We don't have any releases so I'm eliding support information there.

This stems from the security policy advice at [OSSF](https://github.com/ossf/scorecard/blob/b118c1950f34e9d89237c1c6e69d273a192c490c/docs/checks.md#security-policy).  It's common enough for these to be minimal ([step-security](https://github.com/step-security/harden-runner/security) is just an email address). I considered setting up a security list for Carbon, but not sure we need it at the moment; it seems slightly duplicative of the reporting which was previously enabled.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,16 @@
+<!--
+Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+Exceptions. See /LICENSE for license information.
+SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+-->
+
+# Security policy
+
+# Reporting a vulnerability
+
+Please use
+<https://github.com/carbon-language/carbon-lang/security/advisories/new> to
+report security vulnerabilities.
+
+We use GitHub's vulnerability reporting for intake. We will respond to reports
+within two weeks. For valid issues we will coordinate and disclose on GitHub.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -6,6 +6,10 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 # Security policy
 
+The Carbon Language is still an
+[experimental project](/README.md#project-status), so please be careful if using
+it in environments with security concerns.
+
 # Reporting a vulnerability
 
 Please use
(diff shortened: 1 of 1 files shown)


jonmeow:
@chandlerc Any concerns with this change?

SECURITY.md
```diff
@@ -0,0 +1,16 @@
+<!--
+Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+Exceptions. See /LICENSE for license information.
+SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+-->
+
+# Security policy
+
```
chandlerc, line 8: WDYT about a brief mention of the project status here?

Maybe something like:

```suggestion
# Security policy

The Carbon Language is still an experimental project, and as such we don't recommend using it in a production environment or somewhere with significant security concerns.

However, security and [safety](/docs/project/goals.md#practical-safety-and-testing-mechanisms) are important priorities for the project long-term. It is an area we take seriously and want to make sure we are set up for success. Even as an experimental project, please follow our process below to report any security vulnerabilities or issues.

```

chandlerc (commented):
(Sorry I didn't see this last week, just missed it.)

Generally like it. Suggested a pretty minor addition just to add some context for folks that may land here. Happy to have suggestions or tweaks of the wording here though, this is just off the top of my head.

SECURITY.md
jonmeow: I've added some of this, however, trying a different phrasing (trying to lower the emphasis on "experimental", "long-term" because I think it would discourage reporting). Also adding a couple other steps (this is me glancing at https://www.rust-lang.org/policies/security for phrasing ideas, too).

chandlerc: Nice!

chandlerc (approved):
LG!

PR #3888, opened 2024-04-15
More explicitly disable ui events.

jonmeow:
Rather than individually disabling single events, resulting in no events being printed, clear all events.

### DIFF
--- a/explorer/autoupdate_testdata.sh
+++ b/explorer/autoupdate_testdata.sh
@@ -5,6 +5,6 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 bazel run -c opt --experimental_convenience_symlinks=ignore \
-  --ui_event_filters=-info,-stdout,-stderr,-finish \
+  --ui_event_filters= \
   --test_sharding_strategy=disabled \
   //explorer:file_test -- --autoupdate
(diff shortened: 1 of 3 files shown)


zygoloid (commented):
I think this is turning off too much -- this loses fatal, error, warning, and progress in addition to the spam we were trying to filter out. In particular, this change means we won't get compile errors from autoupdate any more.

But it looks like the only filters that are still working are the `-info,-stderr`, so we could reduce the list to just those two. The other things we were trying to turn off here now seem to make it through all the filters :(

jonmeow:
The precise behavior doesn't seem to be well documented ([flag reference](https://bazel.build/reference/command-line-reference#flag--ui_event_filters) is vague, [related enum](https://github.com/bazelbuild/bazel/blob/76a214932da4a9dcf73d3cfe9313e0172d20e630/src/main/java/com/google/devtools/build/lib/events/EventKind.java#L21) doesn't seem to match behavior). All I really have is the error message from specifying an invalid arg (can't specify EventKind groups): "ERROR: While parsing option --ui_event_filters=errors: Not a valid event kind: 'errors' (should be fatal, error, warning, info, progress, debug, start, finish, subcommand, stdout, stderr, pass, fail, timeout, cancelled or depchecker)")

I'll leave this alone since maybe behavior will keep changing.

PR #3889, opened 2024-04-16
Support explicit conversion between adapters and their adapted types.

zygoloid:
Allow an explicit `as` conversion to convert between adapters and their adapted types. Also make the value representation of an adapter be the same as the value representation of the adapted type so that the conversion is always possible.


toolchain/check/testdata/as/adapter_conversion.carbon
```diff
@@ -0,0 +1,447 @@
...
+
+class B {
+  adapt A;
+}
+
+var a_ref: A = {.x = 1, .y = 2};
+let a_val: A = a_ref;
+
+let b_val: B = a_val as B;
+let b_ptr: B* = &(a_ref as B);
```
josh11b, line 29: Comment that `as` to an adapter type preserves category?

toolchain/check/testdata/class/extend_adapt.carbon
```diff
@@ -94,8 +94,7 @@ class StructAdapter {
 // CHECK:STDOUT:   %.1: type = unbound_element_type SomeClass, i32 [template]
 // CHECK:STDOUT:   %.2: type = struct_type {.a: i32, .b: i32} [template]
 // CHECK:STDOUT:   %.3: type = ptr_type {.a: i32, .b: i32} [template]
-// CHECK:STDOUT:   %.4: type = ptr_type SomeClass [template]
```
josh11b, line 97: Is this expected churn?

toolchain/sem_ir/class.h
```diff
@@ -68,7 +68,8 @@ struct Class : public Printable<Class> {
   // The following members are set at the `}` of the class definition.
 
   // The object representation type to use for this class. This is valid once
-  // the class is defined.
+  // the class is defined. For an adapter, this is the non-adapter type that
+  // this class directly or indirectly adapts.
```
josh11b: Maybe?
```suggestion
  // this class directly or transitively adapts.
```

toolchain/check/testdata/as/adapter_conversion.carbon
zygoloid: Done.

toolchain/check/testdata/class/extend_adapt.carbon
zygoloid: Yes -- the value representation for the adapter is now copied from the adapted type. It used to be `SomeClass*`, which was why this type showed up, but it's now `{.a: i32, .b: i32}*`.

josh11b (approved):

PR #3891, opened 2024-04-16
Initial, extern-ignoring support for extern class decls.

jonmeow:
This doesn't actually track whether a declaration is `extern`. It does, however:

- Factor out and expand merge support for classes, sharing handling with functions.
  - This makes the ClassRedefinition diagnostic redundant, as the redeclaration checking overlaps.
- Add partial `extern` handling to class handling; just some verifications of correct use.
- Factor out `extern` on member handling for sharing with `fn`.
- Fixes a bug in import_ref where a class's definition_id wasn't assigned when defining.

This changes how a redefinition is handled (replaced, rather than merged). I don't know whether that's ideal, but I think it results in easy-to-understand consequences, and it's more consistent with how `fn` works.

There's enough work here that this felt like a decent cut point, particularly as the amount of work to actually add `extern` tracking will be significant.


toolchain/check/testdata/class/cross_package_import.carbon
```diff
@@ -62,12 +61,18 @@ var c: Other.C = {};
...
 
 import Other library "define";
-// CHECK:STDERR: fail_merge_define_extern.carbon:[[@LINE+6]]:1: In import.
+// CHECK:STDERR: fail_merge_define_extern.carbon:[[@LINE+12]]:1: In import.
 // CHECK:STDERR: import Other library "extern";
 // CHECK:STDERR: ^~~~~~
-// CHECK:STDERR: other_extern.carbon:5:1: ERROR: Semantics TODO: `Merging ClassType not yet supported.`.
-// CHECK:STDERR: class C;
-// CHECK:STDERR: ^~~~~~~~
+// CHECK:STDERR: other_extern.carbon:4:1: ERROR: Only one library can declare `class C` without `extern`.
```
geoffromer, line 67: Maybe add a TODO comment flagging the fact that this error message is wrong, or at least give this file a `todo` name? That way there's less risk of confusion about the intent of the test.

toolchain/check/testdata/class/extern.carbon
```diff
@@ -0,0 +1,762 @@
...
+// CHECK:STDERR:
+import library "extern_decl_copy";
+
+// --- fail_todo_extern_decl_after_import_extern_decl.carbon
+
+library "extern_decl_after_import_extern_decl" api;
+
+import library "extern_decl";
+
+// CHECK:STDERR: fail_todo_extern_decl_after_import_extern_decl.carbon:[[@LINE+10]]:1: ERROR: Duplicate name being declared in the same scope.
```
geoffromer, line 244: It surprises me that this is an error. Is this the behavior we want, or should this case be a todo?

toolchain/check/testdata/class/fail_redefinition.carbon
```diff
@@ -10,40 +10,30 @@ class Class {
...
-  // CHECK:STDERR:   fn I() {}
-  // CHECK:STDERR:   ^~~~~~~~
-  // CHECK:STDERR: fail_redefinition.carbon:[[@LINE-23]]:3: Previously defined here.
-  // CHECK:STDERR:   fn I() {}
-  // CHECK:STDERR:   ^~~~~~~~
-  // CHECK:STDERR:
   fn I() {}
 }
 
+// CHECK:STDERR: fail_redefinition.carbon:[[@LINE+4]]:10: ERROR: Out-of-line declaration requires a declaration in scoped entity.
```
geoffromer, line 26: This seems like a regression to me; this error message is confusing if I'm looking at the first definition of `Class` instead of the second. I don't think this needs to be fixed now, but is there a path to fixing it later, if we decide it's worth doing?

toolchain/check/testdata/class/extern.carbon
```diff
@@ -0,0 +1,762 @@
...
+extern class C;
+
+// --- def.carbon
+
+library "def" api;
+
+class C {}
+
+// ============================================================================
+// Test files
```
geoffromer, line 36: Do I understand correctly that the test files are all independent of each other, and these cases could instead be written as a bunch of separate _actual_ files, with the setup duplicated across them? If so, would it make sense to do it that way instead? It seems like that would make the tests easier to work with, because the golden IR would be hundreds of lines closer to the corresponding source code, and it would be easy to tell at a glance that there's no interaction between the test files.

toolchain/check/testdata/class/extern.carbon
```diff
@@ -0,0 +1,762 @@
...
+// CHECK:STDOUT:   %C.decl.2: invalid = class_decl @C.2 [template = constants.%C.2] {}
+// CHECK:STDOUT:   %Core: <namespace> = namespace [template] {}
+// CHECK:STDOUT: }
+// CHECK:STDOUT:
+// CHECK:STDOUT: class @C.1 {
+// CHECK:STDOUT: !members:
+// CHECK:STDOUT:   .Self = file.%import_ref.3
+// CHECK:STDOUT: }
+// CHECK:STDOUT:
+// CHECK:STDOUT: class @C.2 {
```
geoffromer, line 597: It looks like we wind up with two class definitions in the IR, even though there's only one class definition in the source files. Am I reading that right?

toolchain/check/class.h
```diff
@@ -0,0 +1,28 @@
...
+#include "toolchain/sem_ir/class.h"
+#include "toolchain/sem_ir/ids.h"
+
+namespace Carbon::Check {
+
+// Tries to merge new_class into prev_class_id. Since new_class won't have a
+// definition even if one is upcoming, set is_definition to indicate the planned
+// result.
+//
+// If merging is successful, updates the ClassId on new_class and returns
```
geoffromer: I don't see the function body modifying new_class anywhere. Am I misunderstanding something?

toolchain/check/testdata/class/extern.carbon
jonmeow: "todo" is in the filename. Note the `fail_` prefix is required for the current failing result.

toolchain/check/testdata/class/fail_redefinition.carbon
jonmeow: I've noted this to @zygoloid as a concern for this approach. Right now I would say it's WAI.

There's probably a path, it's just time and effort.

toolchain/check/class.h
jonmeow: Fixed, also the mirrored comment in function.h

toolchain/check/testdata/class/extern.carbon
jonmeow: Similar to how when you write a unit test you might have test infrastructure [and] multiple TEST_F files in a single file, we are moving towards a single file with multiple split-file test cases. The number of physical files is just becoming unwieldy, and copying setup will result in substantial duplication too.

Although the golden IR is significant, this is the trade-off we've chosen right now, putting the IR at the bottom of the file so that the test cases are easier to read through.

toolchain/check/testdata/class/extern.carbon
jonmeow: At present, yes. Note though that this is a fail test, and also there is a TODO in merge.cpp lines 202-203 about spoiling new classes when merges occur.

toolchain/check/testdata/class/cross_package_import.carbon
jonmeow: I think this should succeed, so added `todo` to the filename.

geoffromer (approved):

PR #3892, opened 2024-04-17
Use separate builtins for signed / unsigned arithmetic.

zygoloid:
Per offline discussion with chandlerc and jonmeow, use different builtins for signed versus unsigned integer ops instead of looking at the type. In this commit, the arithmetic builtins (add, sub, negate, mul, div, mod) are split. I'll apply the same change to comparisons and to right shift in separate PRs.


jonmeow:
Note, removing `@`s in the main commit description. Otherwise, when committed (e.g. to trunk), it keeps creating notifications in GitHub.

toolchain/sem_ir/builtin_function_kind.cpp
```diff
@@ -164,29 +164,53 @@ constexpr BuiltinInfo FloatMakeType = {"float.make_type",
 constexpr BuiltinInfo BoolMakeType = {"bool.make_type",
                                       ValidateSignature<auto()->Type>};
 
-// "int.negate": integer negation.
-constexpr BuiltinInfo IntNegate = {"int.negate",
-                                   ValidateSignature<auto(IntT)->IntT>};
+// "int.snegate": integer negation.
+constexpr BuiltinInfo IntSNegate = {"int.snegate",
+                                    ValidateSignature<auto(IntT)->IntT>};
```
jonmeow, line 169: You now have "snegate" but "make_type_signed". Is this something worth being consistent on, either something like "make_stype" or "negate_signed"? Or perhaps "int." + "uint." instead of "int.s" + "int.u".

Maybe familiarity with C++ types is making me like "uint", although "sadd" kind of makes me want an underscore between 'words' (s_add), and it seems more consistent with "i32" + "u32" naming. I understand if you're trying to name after llvm ops, though it's kind of masked by the calls (and things like mod vs rem).

toolchain/lower/handle.cpp
```diff
@@ -242,48 +251,70 @@ static auto HandleBuiltinCall(FunctionContext& context, SemIR::InstId inst_id,
...
+    case SemIR::BuiltinFunctionKind::IntSMod: {
+      context.SetLocal(inst_id, context.builder().CreateSRem(
+                                    context.GetValue(arg_ids[0]),
+                                    context.GetValue(arg_ids[1]), "rem"));
+      return;
+    }
+    case SemIR::BuiltinFunctionKind::IntUAdd: {
+      context.SetLocal(inst_id, context.builder().CreateAdd(
+                                    context.GetValue(arg_ids[0]),
+                                    context.GetValue(arg_ids[1]), "add"));
```
jonmeow, line 298: Is this at all something we could make a helper function for, just passing in &IRBuilder::CreateAdd and "add" as arguments in order to reduce boilerplate? Note, I'm not totally convinced that's worthwhile, but in my head this could look like:

```
static auto CreateBinaryOp(..., bool signed_overflow_is_ub = false) -> void {
  context.SetLocal(inst_id, (context.builder().*op_fn)(context.GetValue(arg_ids[0]), context.GetValue(arg_ids[1]), "add", /*HasNUW=*/false, /*HasNSW=*/signed_overflow_is_ub);
}

  // IntUAdd
  CreateBinaryOp(context, inst_id, arg_ids, &IRBuilder::CreateAdd, "add");

  // IntSAdd
  CreateBinaryOp(context, inst_id, arg_ids, &IRBuilder::CreateAdd, "add", /*signed_overflow_is_ub=*/true);
```

toolchain/lower/handle.cpp
```diff
@@ -242,48 +251,70 @@ static auto HandleBuiltinCall(FunctionContext& context, SemIR::InstId inst_id,
...
-                                    ? context.builder().CreateSRem(
-                                          context.GetValue(arg_ids[0]),
-                                          context.GetValue(arg_ids[1]), "rem")
-                                    : context.builder().CreateURem(
-                                          context.GetValue(arg_ids[0]),
-                                          context.GetValue(arg_ids[1]), "rem"));
+    case SemIR::BuiltinFunctionKind::IntSMod: {
+      context.SetLocal(inst_id, context.builder().CreateSRem(
+                                    context.GetValue(arg_ids[0]),
+                                    context.GetValue(arg_ids[1]), "rem"));
```
jonmeow, line 292: Would it be helpful for names to reflect the builtin function rather than the call? e.g., "smod" (including signed versus unsigned) instead of "rem".

jonmeow (approved):

toolchain/check/eval.cpp
```diff
@@ -391,6 +391,9 @@ static auto PerformBuiltinUnaryIntOp(Context& context, SemIRLoc loc,
       }
       op_val.negate();
       break;
+    case SemIR::BuiltinFunctionKind::IntUNegate:
+      op_val.negate();
+      break;
```
josh11b, line 396: Do unsigned types implement negate?

toolchain/lower/handle.cpp
zygoloid: The default arguments are a pain here. Some of the binary operations we create have the `nsw` / `nuw` flags, but others of them don't, or have flags with other meanings. I'll take a look, but I'm not feeling optimistic. (This is pre-existing; I'm assuming you don't want this done in this PR?)

toolchain/sem_ir/builtin_function_kind.cpp
zygoloid: Yeah, I think we wanted to mirror the names that LLVM uses here, hence `sadd` / `uadd` matching `APInt`. But these aren't public API so I don't really care too much, and I'm fine with reconsidering that if these names are making you sadd.

I think we could just about get rid of `make_type_signed` / `make_type_unsigned` after all the builtins are changed to not care about the signedness of the type -- we could have a single `int.make_type(N)` builtin that produces a "sequence of N bits" type, like LLVM's `iN`. The remaining issue, I think, is that we use the signedness in diagnostics and SemIR to determine how to print values of the type to the user. That too could be fixed if we use an interface to determine how to print values for diagnostics; I don't know if that's something we would want to explore in the short term, but we could consider that as a long-term approach here. In the short term, I'm open to renaming the `make_type_*` functions to whatever's consistent with our other decisions.

I think I like keeping a consistent "int." prefix, given that I think it would be reasonable to expose the "unsigned" (wrapping) operations on `Core.Int(N)` with different names -- I think treating the choice as part of the operation rather than part of its type or namespace at the core level is likely to be a good thing.

One possible option here would be to focus on the meaning of the operations rather than on the "signedness" -- eg, `int.add` (signed overflow is an error) vs `int.add.wrap` (overflow wraps). That would work for `add`, `sub`, `mul`, and `negate`, but not for `div`, `mod`, and comparisons.

We also need to decide how to name the right-shift operations. I'd been planning on `int.ashr` (arithmetic right shift) vs `int.lshr` (logical right shift) to match LLVM, but if we deviate from LLVM naming then we could call those "signed" vs "unsigned" if that's more consistent.

toolchain/check/eval.cpp
zygoloid: Per [the design](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/expressions/arithmetic.md#integer-types:~:text=Signed%20and%20unsigned%20integer%20types%20support%20all%20the%20arithmetic%20operators), yes. I'm not sure we explicitly considered the alternative, but we probably would find use for a builtin for a wrapping negate even if we don't expose it as a unary `-` operator.

toolchain/lower/handle.cpp
zygoloid: Given that the builtin names are an API between the compiler and the core library, rather than user-facing, I don't know whether / how much a Carbon developer reading the LLVM IR will care about them matching; I think matching the name of the LLVM IR instruction might be more useful. But this all seems to connect to the broader question of naming; I think we should settle that before proceeding with this PR.

toolchain/lower/handle.cpp
jonmeow: Nothing here is important for this PR.

toolchain/lower/handle.cpp
jonmeow: FWIW, I thought these names would be used in the LLVM IR and mainly it might help match back operations to the thing that created them (at least, I thought that's what the purpose of these were). I was expecting the LLVM IR instruction to already be visible, as in:

```
// CHECK:STDOUT: define i32 @TestMod(i32 %a, i32 %b) {
// CHECK:STDOUT:   %rem = srem i32 %a, %b
// CHECK:STDOUT:   ret i32 %rem
// CHECK:STDOUT: }
```

->

```
// CHECK:STDOUT: define i32 @TestMod(i32 %a, i32 %b) {
// CHECK:STDOUT:   %smod = srem i32 %a, %b
// CHECK:STDOUT:   ret i32 %smod
// CHECK:STDOUT: }
```

Note, maybe this could even just pass the builtin name?

toolchain/sem_ir/builtin_function_kind.cpp
jonmeow: I was expecting the `make_type` calls to want to return something semi-typed. The return value needs to work with Carbon's type system (i.e., `i32` versus `u32` cannot be distinguished based on bit width).

But I also think it's temporary: once we have a native `Int(32)` typing capacity, wouldn't we just have that contain storage such as `bits.allocate(N)`?

FWIW I lean towards distinguishing builtins based on how Carbon plans to use them, rather than how LLVM IR defines them, if we're only exposing these for the core library. In part there, I expect that things that are more broadly exposed, as in for end users, would probably have a different API.

zygoloid: We can't just use something like `bits.allocate(N)` -- we need enough typing in the type primitives to know how to represent the type in LLVM IR (eg, `iN` versus any other N-bit type) and how to represent it during constant evaluation / how to type-check the inputs to the builtin functions. We still need somewhat high-level primitive types for both purposes (eg, "this is a primitive N-bit integer", not just "this is N bits wide"), but we should be able to get to the point where we have something like:

```carbon
package Core api;

fn MakeIntN(N:! BigInt) -> type = "int.make";

class Int(N:! BigInt) {
  adapt MakeIntN(N);

  impl as Add {
    fn Op[self: Self](other: Self) = "int.sadd";
  }
}

class UInt(N:! BigInt) {
  adapt MakeIntN(N);

  impl as Add {
    fn Op[self: Self](other: Self) = "int.uadd";
  }
}
```

toolchain/lower/handle.cpp
zygoloid: I think the builtin name is an implementation detail of the core lib <-> compiler interface, and shouldn't be showing up in the IR -- except if, by "coincidence", it's also a good name for the result, which I'd expect it usually is.

But I do somewhat question how useful the instruction names we're adding here really are. It doesn't seem very valuable to me to add names that could be inferred from the IR itself, and giving instructions names does add some cost. Names like `%a` and `%b` above are clearly very useful, and give information that would not otherwise 
be present in the IR, but names that just describe the IR don't seem so useful to me.

Maybe we can just remove these names?

jonmeow: Maybe they're not important -- I'll leave that to you. I'm not sure what the effect of just removing would be. You could also just call it "ret" in all cases. It seems like these are all short functions, so the value of naming the result is negligible.

FWIW %a %b could be %lhs %rhs in context.

toolchain/sem_ir/builtin_function_kind.cpp
jonmeow: Wouldn't the casting behavior of llvm types be sufficient? I had thoughts types were mainly aesthetic, that these were actually always force-cast in essence?

That is, the builtin function could partly be defining how to convert parameters for llvm's purposes. I'm not sure it needs to be defined by the result of "int.make". [But you're more familiar with llvm.]

zygoloid: The pointee types of pointers don't matter to LLVM, but types do otherwise matter quite a lot. Maybe the biggest thing is that the type used to represent function parameters and return values affects the calling convention used for the function. For local values, if we get the type "wrong" we'll end up at best inserting a lot of `bitcast` instructions in the IR we generate, but that's not always possible -- I think the most obvious representation for "N bytes of memory" would be `[i8 x N]`, which is not even `bitcast`able and would require going through memory.

We need our types to specify their representation enough that we get the LLVM IR types we want; I don't think there's any way around that. I think the idea of using a `bits.allocate(N)` for everything, while appealing, won't work. We can probably do something like `allocate(Size, Alignment, LLVM type)` -- but even then, I think having stronger typing will make the constant evaluation side of things much simpler for us to deal with, and give us the ability to evaluate more efficiently.

toolchain/lower/handle.cpp
zygoloid: I've sent out #3898 to handle this more generally.

toolchain/sem_ir/builtin_function_kind.cpp
zygoloid: I'm going to go ahead and land this split of the builtins; we can iterate on the naming separately, based on the results of the discussion on discord.

PR #3893, opened 2024-04-17
Add ExternDecl and ExternType for extern classes.

jonmeow:
This expands `extern` handling for most cases.


jonmeow:
Depends on #3891

toolchain/check/handle_class.cpp
```diff
@@ -35,6 +38,51 @@ auto HandleClassIntroducer(Context& context, Parse::ClassIntroducerId node_id)
...
+    auto prev_class_id = SemIR::ClassId::Invalid;
+    CARBON_KIND_SWITCH(prev_inst_for_merge.inst) {
+      case CARBON_KIND(SemIR::ClassDecl class_decl): {
+        prev_class_id = class_decl.class_id;
+        break;
+      }
+      case CARBON_KIND(SemIR::ClassType class_type): {
+        prev_class_id = class_type.class_id;
+        break;
+      }
```
zygoloid, line 64: When does this case happen? I'm cautious here because I don't think we want *any* previous name that names a class type to be merged. Eg, if the previous declaration was an `ImportRef` naming `alias C = OtherClass;`, then that shouldn't merge with `class C` even though it was a class type. But if I'm reading `ResolvePrevInstForMerge` correctly, you'll get a `ClassType` in both the case of an imported class declaration and in the case of any other imported declaration that constant-evaluates to a class type.

zygoloid (approved):
Looks good. One potential concern, but if it is a real problem, it can be addressed in a separate PR.

toolchain/check/handle_class.cpp
jonmeow: Theoretically I agree, but I'd say what you're mentioning (with alias) is not supported right now.

Bear in mind an ImportRef only has the constant value to point at the instruction in the current IR. This is all we have to work with, without a major refactoring. Which maybe should occur, and I have already have spots I'm thinking about this, but that's very much out-of-scope.

PR #3894, opened 2024-04-17
Fix prelude finding for direct driver execution.

jonmeow:
This is just to be able to directly run the driver (e.g., `bazel run //toolchain/driver:carbon compile foo.carbon`); it shouldn't affect anything else.

### DIFF
--- a/testing/file_test/file_test_base.h
+++ b/testing/file_test/file_test_base.h
@@ -60,7 +60,7 @@ class FileTestBase : public testing::Test {
     bool success;
 
     // Per-file success results. May be empty.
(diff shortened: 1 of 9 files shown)


chandlerc (approved):
LG (took a quick look as I've been looking at related stuff recently).

Some of this may make sense to factor (I'm seeing some overlap with the Clang driver code integration I've been working on), but I think that can happen later / when needed. This seems like a good initial approach and a strict improvement.

jonmeow:
Switches StringRefs to strings in a few spots because this creates lifetime issues (visible in opt builds).

PR #3895, opened 2024-04-17
Make driver find all prelude files. Add build rule for //examples:sieve.

zygoloid:
The driver now looks for all files under core/prelude/ and considers them all to be part of the prelude. The driver also now only processes the prelude in `--phase=check` and later, when it would actually be imported.

With that done, add a simple `carbon_binary` build rule and use it to build the example in `//examples`. This should cause the example to be built as part of our continuous integration.


examples/sieve.carbon
```diff
@@ -8,8 +8,9 @@ import Core library "prelude/operators/arithmetic";
 import Core library "prelude/operators/comparison";
 
 // TODO: Copied from core/prelude/types/i32.carbon.
-// Remove the following, once we do cross-file impl lookup.
-// import Core library "prelude/types/i32";
+// Because we don't deduplicate interfaces, the implementations in that file are
+// treated as implementing a different interface from the one we import above.
+// Remove the following, once we deduplicaate interfaces.
```
jonmeow:
```suggestion
// Remove the following, once we deduplicate interfaces.
```

bazel/carbon_rules/defs.bzl
```diff
@@ -0,0 +1,40 @@
...
+        # Build each source file. For now, we pass all sources to each compile
+        # because we don't have visibility into dependencies and have no way to
+        # specify multiple output files. Object code for each input is written
+        # into the output file in turn, so the final carbon source file
+        # specified ends up determining the contents of the object file.
+        #
+        # We also pass in the prelude files, but not prelude.carbon because the
+        # driver adds that itself. For now, we assume that the prelude doesn't
+        # produce any necessary object code, and don't include the .o files for
+        # //core/prelude... in the final linked binary.
```
jonmeow: I'm concerned what you're doing here is going to be brittle. Your changes to prelude.carbon affect file_test, which needs to also load any dependencies. Right now that's handled through the driver. By creating a divergence here, I expect file_test to break.

Had you considered instead making dependencies of the main prelude file load the same way that the main prelude file is loaded?

zygoloid: Done.

bazel/carbon_rules/defs.bzl
```diff
@@ -0,0 +1,47 @@
...
+        name = "%s.objs" % name,
+        objects = [src + ".compile" for src in srcs],
+    )
+
+    # For now, we assume that the prelude doesn't produce any necessary object
+    # code, and don't include the .o files for //core/prelude... in the final
+    # linked binary.
+    #
+    # TODO: This will need to be revisited eventually.
+    cc_binary(name = name, deps = ["%s.objs" % name])
```
jonmeow, line 47: Just to note, although this says "revisited eventually", I'm concerned this is actually going to be a barrier to prelude development somewhat soon since it means we can't put in any logic or function bodies. But maybe we can just comment this out if it gets in the way.

core/BUILD
```diff
@@ -2,8 +2,9 @@
 # Exceptions. See /LICENSE for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+# The prelude, and all of its dependencies.
 filegroup(
-    name = "core",
-    data = glob(["**/*.carbon"]),
+    name = "prelude",
+    data = ["prelude.carbon"] + glob(["prelude/**/*.carbon"]),
```
jonmeow, line 8: What model do you expect when we start having more files in core than just the prelude? Shouldn't the driver and tests have those available too?

toolchain/driver/BUILD
```diff
@@ -72,6 +72,7 @@ cc_fuzz_test(
 cc_binary(
     name = "carbon",
     srcs = ["driver_main.cpp"],
+    data = ["//core:prelude"],
```
jonmeow: The driver library already has this, and driver_main.cpp doesn't use it; why is it required here?

toolchain/driver/driver.h
```diff
@@ -48,6 +48,13 @@ class Driver {
   // error stream (stderr by default).
   auto RunCommand(llvm::ArrayRef<llvm::StringRef> args) -> RunResult;
 
+  // Finds the source files that define the prelude and returns a list of their
+  // filenames. On error, writes a message to `error_stream` and returns an
+  // empty list.
+  static auto FindPreludeFiles(llvm::StringRef data_dir,
```
jonmeow, line 54: This is taking in members of Driver, so why a static function? I'd expect everywhere that uses it is creating an instance of Driver, and it could be called before RunCommand.

toolchain/driver/driver.cpp
```diff
@@ -694,7 +739,9 @@ auto Driver::Compile(const CompileOptions& options) -> RunResult {
     // Shared values will always be printed after per-file printing.
     if (options.dump_shared_values) {
       for (const auto& unit : units) {
-        unit->PrintSharedValues();
+        if (unit->IncludeInDumps()) {
+          unit->PrintSharedValues();
+        }
```
jonmeow: Rather than making IncludeInDumps public, maybe PrintSharedValues should just query IncludeInDumps? I mean, it's really just for dumping, not generic past that.

toolchain/driver/driver.cpp
```diff
@@ -30,6 +30,45 @@
...
+                          "core/prelude");
+  std::error_code ec;
+  for (llvm::sys::fs::recursive_directory_iterator prelude_files_it(
+           prelude_dir, ec, /*follow_symlinks=*/false);
+       prelude_files_it != llvm::sys::fs::recursive_directory_iterator();
+       prelude_files_it.increment(ec)) {
+    if (ec) {
+      error_stream << "ERROR: Could not find prelude: " << ec.message() << "\n";
+      result.clear();
+      break;
```
```suggestion
      return {};
```

I think it'd be clearer here to return directly.

toolchain/driver/driver.cpp
```diff
@@ -30,6 +30,45 @@
...
+    }
+
+    // TODO: Should we ignore case?
```
jonmeow: Can you elaborate on this todo? What's the situation where this function would encounter file extensions with other casing choices? Note, we already have other file extension checking code: I would not expect case-insensitive logic here to be sufficient.

bazel/carbon_rules/defs.bzl
zygoloid: I think probably the right way to handle this would be to add build rules to //core to build a library and add it as a dependency here. I think that should be pretty straightforward to do when the time comes.

core/BUILD
zygoloid: My expectation is that toolchain tests should not depend on any part of core other than the prelude. I think tests for the rest of core should live in `//core` instead, treating the rest of core as a layer on top of the base toolchain.

For `carbon_binary` or `carbon_library` targets, I'm expecting we would take a similar stance: each target should be explicitly declaring a dependency on any libraries they use (including core libraries). If that turns out to be excessively verbose or non-ergonomic, we could inject an implicit dependency on `//core` into each target, but to the extent possible, I think we should strive to treat core as "just another package" to the extent we can. The prelude is an exception to this because we implicitly import it (but even then, I think we should be looking at ways we can make it less special in the longer term).

toolchain/driver/BUILD
zygoloid: You're right, it's redundant. Removed.

toolchain/driver/driver.h
zygoloid: It's used to construct the `llvm::vfs::FileSystem` that is passed to the driver, and doesn't respect that file system, so I felt it was clearer to make it static rather than to make it non-static and have `file_test` mutate the VFS after passing it into the driver.

toolchain/driver/driver.cpp
zygoloid: Done. Factored out all the per-CU finalization code to make this consistent with the other things we do with CUs.

toolchain/driver/driver.cpp
zygoloid: Sure. But let's not disable NRVO:
```suggestion
      result.clear();
      return result;
```

toolchain/driver/driver.cpp
zygoloid: OK, if we're already doing a case-sensitive extension comparison, let's just drop the TODO. I think it's fine to insist that the carbon installation preserves the lowercase extensions (we're already using a more-than-three-character extension, so it's not like we need to worry about FAT filenames anyway, and everyone else I think we might care about is case-preserving).

```suggestion
```

core/BUILD
jonmeow: To try to wrap back from discussion yesterday -- I think when we run the driver through bazel, or when we install, we want to be pulling in _all_ core files as dependencies, not just the prelude. I think a developer should be able to run the toolchain directly on a source file, and get compiled output -- behind the scenes, package management logic should make that just work.

Carbon's toolchain tests are going to be difficult to split up. We're running into linking size issues, and that's going to be a significant limiting factor for splitting things up.

With regards to the ability to set narrow dependencies, I don't think we're there right now. Also, for a small project, I would be concerned that having to individually set dependencies _in addition to_ individually `import`ing dependencies is going to be regarded as duplicative. (would this even come up in a local install or cmake setup?)

jonmeow (approved):
Approving, because even with my disagreement about narrowing of core support to just the prelude, it's something we can evolve in the future if you want to proceed with it.

PR #3897, opened 2024-04-18
Add builtins for basic float operations

ppaglilla:
Adds support for unary `-` and binary `+`, `-`, `*`, `/` for floating point types.

Real literals are now transformed to `llvm::APFloat`s during the check phase into the `FloatLiteral` instruction.

This PR likely collides a bit with #3892 and might need to be updated when that one is merged.


toolchain/sem_ir/file.cpp
```diff
@@ -382,6 +384,21 @@ static auto StringifyTypeExprImpl(const SemIR::File& outer_sem_ir,
...
+      case CARBON_KIND(FloatType inst): {
+        // TODO: Is this okay?
+        if (step.index == 1) {
+          out << ")";
+        } else if (auto width_value =
+                       sem_ir.insts().TryGetAs<IntLiteral>(inst.bit_width_id)) {
+          out << "f";
+          sem_ir.ints().Get(width_value->int_id).print(out, /*isSigned=*/false);
+        } else {
+          out << "Float(";
```
zygoloid:
```suggestion
          out << "Core.Float(";
```

toolchain/sem_ir/typed_insts.h
```diff
@@ -444,6 +444,25 @@ struct FacetTypeAccess {
...
+  TypeId type_id;
+  FloatId float_id;
+};
+
+struct FloatType {
+  static constexpr auto Kind =
+      InstKind::FloatType.Define<Parse::NodeId>("float_type");
+
+  TypeId type_id;
+  FloatKind float_kind;
```
zygoloid: Do you anticipate a use for this, or is this just here to match `IntType`? I think we can probably do without this `kind` value for now.

(The bit width might not be quite enough to distinguish, e.g. a 128-bit extended IEEE format from PPC double-double, or to distinguish between different 16-bit FP formats, but nor can our `fN` notation. I think we can defer worrying about this until we have a design for a wider set of floating point types.)

toolchain/sem_ir/formatter.cpp
```diff
@@ -1161,6 +1161,14 @@ class Formatter {
         .print(out_, sem_ir_.types().IsSignedInt(inst.type_id));
   }
 
+  auto FormatInstructionRHS(FloatLiteral inst) -> void {
+    // Arbitrarily uses 5 digits of precision when formatting
+    // float literals.
+    llvm::SmallVector<char, 16> buffer;
+    sem_ir_.floats().Get(inst.float_id).toString(buffer, 5);
```
zygoloid: Does LLVM provide a way to produce a shortest representation that correctly round-trips? In principle we'd like our formatted SemIR to be parseable back to an equivalent file, so we should use a lossless printing mechanism here.

toolchain/check/handle_literal.cpp
```diff
@@ -58,11 +58,42 @@ auto HandleIntLiteral(Context& context, Parse::IntLiteralId node_id) -> bool {
...
 
 auto HandleRealLiteral(Context& context, Parse::RealLiteralId node_id) -> bool {
+  // Convert the real literal to an llvm::APFloat and add it to the floats
+  // ValueStore. In the future this would use an arbitrary precision Rational
+  // type.
+  auto real_id =
+      context.tokens().GetRealLiteral(context.parse_tree().node_token(node_id));
+  auto real_value = context.sem_ir().reals().Get(real_id);
+
+  if (real_value.mantissa.getActiveBits() > 64) {
```
zygoloid, line 72: The constraints that we want to apply here are [described as follows in the design](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/expressions/implicit_conversions.md#data-types):

> A floating-point constant can be implicitly converted to any type fM in which that value is between the least representable finite value and the greatest representable finite value (inclusive), and converts to the nearest representable finite value, with ties broken by picking the value for which the mantissa is even.

I think it's fine to apply a different set of checks for now, but if so, please can you add a `TODO` comment to implement the above instead?

zygoloid (commented):
Thank you, this looks awesome! I still need to look through the tests but the code changes are great.

toolchain/check/eval.cpp
```diff
@@ -553,6 +573,62 @@ static auto PerformBuiltinIntComparison(Context& context,
...
+static auto PerformBuiltinUnaryFloatOp(Context& context,
+                                       SemIR::BuiltinFunctionKind builtin_kind,
+                                       SemIR::InstId arg_id)
+    -> SemIR::ConstantId {
+  auto op = context.insts().GetAs<SemIR::FloatLiteral>(arg_id);
+  auto op_val = context.floats().Get(op.float_id);
+
+  switch (builtin_kind) {
+    case SemIR::BuiltinFunctionKind::FloatNegate:
+      op_val = llvm::APFloat(0.0) - op_val;
```
zygoloid:
```suggestion
      op_val.changeSign();
```
`APFloat(0.0)` will assume IEEE754 64-bit semantics for the `0.0`, which won't be right once we start supporting more kinds of floating-point type.

toolchain/check/eval.cpp
```diff
@@ -553,6 +573,62 @@ static auto PerformBuiltinIntComparison(Context& context,
...
+                                        SemIR::BuiltinFunctionKind builtin_kind,
+                                        SemIR::InstId lhs_id,
+                                        SemIR::InstId rhs_id)
+    -> SemIR::ConstantId {
+  auto lhs = context.insts().GetAs<SemIR::FloatLiteral>(lhs_id);
+  auto rhs = context.insts().GetAs<SemIR::FloatLiteral>(rhs_id);
+  auto lhs_val = context.floats().Get(lhs.float_id);
+  auto rhs_val = context.floats().Get(rhs.float_id);
+
+  llvm::APFloat result_val(0.0);
```
zygoloid:
```suggestion
  llvm::APFloat result_val(lhs_val.getSemantics());
```
to avoid assuming IEEE-754 64-bit.

toolchain/check/eval.cpp
```diff
@@ -553,6 +573,62 @@ static auto PerformBuiltinIntComparison(Context& context,
...
+      result_val = lhs_val - rhs_val;
+      break;
+    case SemIR::BuiltinFunctionKind::FloatMul:
+      result_val = lhs_val * rhs_val;
+      break;
+    case SemIR::BuiltinFunctionKind::FloatDiv:
+      if (rhs_val.isZero()) {
+        DiagnoseDivisionByZero(context, loc);
+        return SemIR::ConstantId::Error;
+      }
```
zygoloid: We can drop this check; per [the design](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/expressions/arithmetic.md#:~:text=division%20by%20zero%20results%20in%20either%20%C2%B1%E2%88%9E%20or%2C%20for%200.0%20/%200.0%2C%20a%20quiet%20NaN.), floating-point division by zero is permitted.

toolchain/lower/handle.cpp
```diff
@@ -339,6 +339,39 @@ static auto HandleBuiltinCall(FunctionContext& context, SemIR::InstId inst_id,
...
       return;
     }
+    case SemIR::BuiltinFunctionKind::FloatNegate: {
+      // Lower `-x` as `0 - x`
+      auto* operand = context.GetValue(arg_ids[0]);
+      context.SetLocal(inst_id,
+                       context.builder().CreateFSub(
+                           llvm::ConstantInt::getNullValue(operand->getType()),
+                           operand, "fneg"));
+      return;
```
zygoloid: Use `CreateFNeg` here instead of `CreateFSub`. Unlike for integers, `0.0 - x` isn't equivalent to `- x`, specifically when `x` is `0.0` -- `-0.0` should give a negative zero, but `0.0 - 0.0` is a positive zero.

toolchain/sem_ir/file.cpp
ppaglilla: Resolved!

toolchain/sem_ir/typed_insts.h
ppaglilla: That was me following `IntType` and not realizing I didn't end up using the field 😅 . Just removed it.

Thanks for pointing out the bit width as well! Let's defer it for now like you said. I'd be happy to work on that once the design we need is in place.

zygoloid (commented):
Went through this again and found a few more small things.

toolchain/sem_ir/formatter.cpp
ppaglilla: It does. If we keep the precision argument at 0 (the default), LLVM will [use the full precision of the `APFLoat`](https://github.com/llvm/llvm-project/blob/08a787c91be51e6a561dc50ec34ce28abc04ac54/llvm/include/llvm/ADT/APFloat.h#L461) in it [will correctly roundtrip](https://github.com/llvm/llvm-project/blob/aac4d03423dd6b7bdef0f2eb03c570f3e2ca6630/llvm/lib/Support/APFloat.cpp#L4089).

Just updated the formatter to use that, meaning floats will now round-trip. The only caveat is "visual", since we now show the full precision (e.g. `0.10000000000000001` for `1/10`, such as [here](https://github.com/ppaglilla/carbon-lang/blob/9c4d94e0a12dd2d3fa80ae7e7265040e1c123476/toolchain/check/testdata/basics/builtin_types.carbon#L16)).

ppaglilla:
Thank you for the thorough review @zygoloid! I'll continue addressing the comments tomorrow!

toolchain/check/eval.cpp
ppaglilla: Resolved!

toolchain/check/eval.cpp
ppaglilla: Resolved!

toolchain/check/eval.cpp
ppaglilla: Done!

toolchain/lower/handle.cpp
ppaglilla: Done! Thanks for the context!

toolchain/check/handle_literal.cpp
ppaglilla: Added a TODO comment to implement the implicit conversions as documented.

I'm happy to work on it on a follow up PR. I'd just need to read up on floating point internals first, it's been a while since I've worked on this kind of lowish-level things 😅.

zygoloid (approved):
Thank you so much!

PR #3898, opened 2024-04-19
Unify instruction naming between SemIR and LLVM IR

zygoloid:
Factor out `SemIR::InstNamer` and also use it when lowering to LLVM IR. Automatically name all instructions created with our `IRBuilder` based on the name computed by the `InstNamer`, and likewise name basic blocks using the label generated by the `InstNamer`.

Move some of the existing naming logic out from lower into `InstNamer` so that it's also used in SemIR. In particular, we now name call instructions after their callee, or after the builtin name for calls to builtins.

Computing and adding these names isn't completely free. This instruction naming is designed to be optional, so that we can turn it off for builds where the LLVM IR will only be converted to assembly and won't be seen by a human, but so far it's enabled unconditionally. We can tune that later as needed.


toolchain/check/testdata/as/identity.carbon
```diff
@@ -97,10 +97,10 @@ fn Initializing() {
 // CHECK:STDOUT:   %x.var: ref X = var x
 // CHECK:STDOUT:   %x: ref X = bind_name x, %x.var
 // CHECK:STDOUT:   %Make.ref: <function> = name_ref Make, file.%Make [template = file.%Make]
-// CHECK:STDOUT:   %.loc24_7: ref X = splice_block %x.var {}
-// CHECK:STDOUT:   %.loc24_19: init X = call %Make.ref() to %.loc24_7
+// CHECK:STDOUT:   %.loc24: ref X = splice_block %x.var {}
+// CHECK:STDOUT:   %Make: init X = call %Make.ref() to %.loc24
 // CHECK:STDOUT:   %X.ref.loc24_25: type = name_ref X, file.%X.decl [template = constants.%X]
```
chandlerc: Given the `Make.ref` and other `Make` things, maybe `Make.call` would be a bit more clear for the actual call?

chandlerc (commented):
One quick comment, still going through though...

chandlerc (commented):
One quick comment, still going through though...

toolchain/check/testdata/as/identity.carbon
zygoloid: Done.

toolchain/lower/function_context.cpp
```diff
@@ -116,4 +126,17 @@ auto FunctionContext::CopyValue(SemIR::TypeId type_id, SemIR::InstId source_id,
...
+    llvm::Instruction* inst, const llvm::Twine& name, llvm::BasicBlock* block,
+    llvm::BasicBlock::iterator insert_pt) const -> void {
+  llvm::StringRef base_name = (inst_namer_ && !inst->getType()->isVoidTy())
+                                  ? inst_namer_->GetUnscopedNameFor(inst_id_)
+                                  : "";
+  IRBuilderDefaultInserter::InsertHelper(
+      inst,
+      (!base_name.empty() && !name.isTriviallyEmpty()) ? base_name + "." + name
+                                                       : base_name + name,
+      block, insert_pt);
```
chandlerc: I know `Twine` fights hard, but maybe something like this would work?

```suggestion
  auto prefix_name = [&] -> llvm::Twine {
    if (inst_namer_ && !inst->getType()->isVoidTy() &&
        !name.isTriviallyEmpty()) {
      return llvm::Twine(inst_namer_->GetUnscopedNameFor(inst_id_)) + "." + name;
    }
    
    return name;
  };
  IRBuilderDefaultInserter::InsertHelper(
      inst, prefix_name(), block, insert_pt);
```

chandlerc (approved):
Maybe one improvement to the inserter logic? But LG with whatever compiles there. Twine is ... special.

Really like the naming approach, and the result seems really nice for reading the IR.

toolchain/lower/function_context.cpp
zygoloid: I don't think that works, sadly -- the twine returned by the lambda has another twine as a child, and that inner twine is destroyed too early.

Switched to a different approach.

PR #3899, opened 2024-04-19
Add support for builtin float comparison operations

ppaglilla:
Adds support for builtin comparison operations for floats (`==`, `!=`, `<`, `<=`, `>`, `>=`).


toolchain/lower/handle.cpp
```diff
@@ -200,6 +200,28 @@ static auto GetBuiltinICmpPredicate(SemIR::BuiltinFunctionKind builtin_kind,
   }
 }
 
+// Get the predicate to use for an `fcmp` instruction generated for the
+// specified builtin.
+static auto GetBuiltinFCmpPredicate(SemIR::BuiltinFunctionKind builtin_kind)
```
ppaglilla, line 205: This is using the ordered variants of `fcmp` conditions, which I'm not sure is accurate with Carbon's intended behavior. Would love feedback on whether this is okay!

zygoloid: Yes, using ordered comparisons is the intent. I don't think we state this explicitly, but it's [implicitly mentioned here](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/expressions/comparison_operators.md#:~:text=the%20implementation%20for%20floating%2Dpoint%20types%20is%20none%20of%20these%20because%20NaN%20values%20do%20not%20compare%20less%20than%20or%20equivalent%20to%20themselves) and [here](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/expressions/comparison_operators.md#:~:text=floating%2Dpoint%20NaN%20values%2C%20where%20two%20values%20can%20compare%20neither%20equal%20nor%20not%2Dequal) in the design.

zygoloid (approved):
Looks great, thank you!

PR #3902, opened 2024-04-22
Fix incorrect application of invalid type on imports.

jonmeow:
In handle_class and handle_interface we assign TypeType, so is more consistent. I think this had been missed because we haven't really been using these declarations (historically, declarations didn't have a type). It seems not to significantly affect output at present, although I found this while trying to change merge behavior.


jonmeow:
Depends on #3901

zygoloid (approved):

PR #3903, opened 2024-04-22
Shift merge logic into import_ref

jonmeow:
Previously, cross-import merge logic was done as part of the loading locations. This has problems that in particular arise when indirectly importing (not explicitly name referencing) types that are equivalent, but come from separate libraries. There may be nothing linking them other than the name and enclosing scope, but they may still merge.

This PR works towards addressing that by adding imported names to the enclosing scope during import_ref. The intent is that indirectly resolved names will still become "visible". A separate change will be needed to make sure types which should be unavailable cannot be accessed through name lookup.


jonmeow:
Depends on #3904

PR #3904, opened 2024-04-22
Split apart some tests which don't need the prelude.

jonmeow:
This could probably go further, but I'm just trying to get the ability to run some significant tests without the prelude. This already requires some migration of `i32` and `bool` to non-prelude-dependent types (technically not at the moment, but I'm trying to stick with the philosophical model of the prelude).

Note this undoes part of #3895 which had made all lex/parse tests depend on the prelude -- I don't think that was a necessary change. Rather, it seems better to isolate individual lex/parse tests from the prelude. I'd been on the fence in that PR (it was fixing one of the issues that contributed to wanting to not include the prelude), but now it feels more consistent. The prelude could have a parse error and we shouldn't break every parse test on that.


jonmeow:
Depends on #3902

josh11b (approved):

PR #3906, opened 2024-04-23
Allocate de Bruijn levels to symbolic bindings.

zygoloid:
Use a level comparison during substitution to determine whether we're substituting a particular binding. Evaluate symbolic bindings with the same name and the same level to the same symbolic constant, for example across redeclarations of a generic function.

### DIFF
--- a/toolchain/check/scope_stack.h
+++ b/toolchain/check/scope_stack.h
@@ -115,6 +115,13 @@ class ScopeStack {
   auto LookupOrAddName(SemIR::NameId name_id, SemIR::InstId target_id)
       -> SemIR::InstId;
 
+  // Adds a compile-time binding in the current scope, and returns its index.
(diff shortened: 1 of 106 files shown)


toolchain/check/scope_stack.h
```diff
@@ -159,7 +169,7 @@ class ScopeStack {
 
     // Names which are registered with lexical_lookup_, and will need to be
     // unregistered when the scope ends.
-    llvm::DenseSet<SemIR::NameId> names;
+    llvm::DenseSet<SemIR::NameId> names = {};
```
jonmeow, line 172: Out of curiosity, why this addition? Some warning?

jonmeow (approved):
Thank you!

toolchain/check/scope_stack.h
zygoloid: Yes, this fixes a clang-tidy missing initializer warning.

PR #3908, opened 2024-04-23
Track on `InstKind` whether each instruction kind can define a constant.

zygoloid:
The intent is to use this information in lowering to determine which kinds of instruction need constant lowering code and which kinds need runtime lowering code. For now we just validate that we only create constants with instructions that should be able to form constants.

### DIFF
--- a/toolchain/sem_ir/constant.cpp
+++ b/toolchain/sem_ir/constant.cpp
@@ -10,14 +10,21 @@ namespace Carbon::SemIR {
 
 auto ConstantStore::GetOrAdd(Inst inst, bool is_symbolic) -> ConstantId {
   // Check that we're allowed to form this kind of constant.
-  CARBON_CHECK(inst.kind().constant_kind() != InstConstantKind::Never)
-      << "Should not form a constant from instruction " << inst;
-  CARBON_CHECK(inst.kind().constant_kind() != InstConstantKind::SymbolicOnly ||
-               is_symbolic)
-      << "Should only form a symbolic constant from instruction " << inst;
-  CARBON_CHECK(inst.kind().constant_kind() != InstConstantKind::Always ||
-               !is_symbolic)
-      << "Should only form a template constant from instruction " << inst;
+  switch (inst.kind().constant_kind()) {
+    case InstConstantKind::Never:
+      CARBON_FATAL() << "Should not form a constant from instruction " << inst;
+      break;
+    case InstConstantKind::SymbolicOnly:
+      CARBON_CHECK(is_symbolic)
+          << "Should only form a symbolic constant from instruction " << inst;
+      break;
+    case InstConstantKind::Conditional:
+      break;
+    case InstConstantKind::Always:
+      CARBON_CHECK(!is_symbolic)
+          << "Should only form a template constant from instruction " << inst;
+      break;
+  }
 
   // Compute the instruction's profile.
   ConstantNode node = {.inst = inst, .constant_id = ConstantId::NotConstant};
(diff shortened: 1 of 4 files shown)


toolchain/sem_ir/constant.cpp
```diff
@@ -9,6 +9,16 @@
...
 auto ConstantStore::GetOrAdd(Inst inst, bool is_symbolic) -> ConstantId {
+  // Check that we're allowed to form this kind of constant.
+  CARBON_CHECK(inst.kind().constant_kind() != InstConstantKind::Never)
+      << "Should not form a constant from instruction " << inst;
+  CARBON_CHECK(inst.kind().constant_kind() != InstConstantKind::SymbolicOnly ||
+               is_symbolic)
+      << "Should only form a symbolic constant from instruction " << inst;
+  CARBON_CHECK(inst.kind().constant_kind() != InstConstantKind::Always ||
+               !is_symbolic)
+      << "Should only form a template constant from instruction " << inst;
```
jonmeow: Can this be written differently so that either one CHECK runs or something? My mind just doesn't operate straight on this kind of "!=" code structure. A `switch` might make this easier to reason about.

toolchain/sem_ir/inst_kind.h
```diff
@@ -16,14 +16,37 @@ namespace Carbon::SemIR {
...
+// kind of instruction.
+enum class InstConstantKind : int8_t {
+  // This instruction never defines a constant value. For example,
+  // `UnaryOperatorNot` never defines a constant value; if its operand is a
+  // template constant, its constant value will instead be a `BoolLiteral`. This
+  // is also used for instructions that don't produce a value at all.
+  Never,
+  // This instruction may be a symbolic constant, depending on its operands, but
+  // is never a template constant. For example, a `Call` instruction can be a
+  // symbolic constant but never a template constant.
```
jonmeow, line 39: Is this going to remain true with metaprogramming? I'm kind of wondering if BindSymbolicName might be a better example, although it's a little more tautological.

toolchain/sem_ir/inst_kind.h
```diff
@@ -16,14 +16,37 @@ namespace Carbon::SemIR {
...
 
+// Whether an instruction can be used to define a constant value. This specifies
+// whether the instruction can be added to the `constants()` list. Note that
+// even instructions that cannot define a constant value can still have an
+// associated `constant_value()`, but the constant value will be a different
+// kind of instruction.
+enum class InstConstantKind : int8_t {
+  // This instruction never defines a constant value. For example,
+  // `UnaryOperatorNot` never defines a constant value; if its operand is a
+  // template constant, its constant value will instead be a `BoolLiteral`. This
```
jonmeow, line 34: Seems a little odd, structurally, that a BoolLiteral constant isn't a template constant; but maybe I'm missing the purpose...

jonmeow (approved):

toolchain/sem_ir/inst_kind.h
zygoloid: That's reversed from what the comment is trying to say: a `BoolLiteral` can be a constant value; a `UnaryOperatorNot` cannot, because evaluating a `UnaryOperatorNot` produces a `BoolLiteral` constant not a `UnaryOperatorNot` constant.

toolchain/sem_ir/inst_kind.h
zygoloid: I would expect this to remain true -- note that we're classifying the kind of instructions that define constants, not merely the ones that can evaluate to constants, so if we have a call that produces a constant value, the corresponding instruction kind will be the one corresponding to the result of the call, not the `Call` instruction itself. (For example, evaluating a `Call` instruction to a function that returns 42 will produce a constant of kind `IntLiteral`, not of kind `Call`.) I picked this example to draw attention to that subtlety.

toolchain/sem_ir/constant.cpp
zygoloid: Done.

jonmeow (approved):

PR #3909, opened 2024-04-23
Remove ExternDecl and ExternType

jonmeow:
These should no longer be helpful under the new `extern` strategy.


jonmeow:
Depends on #3904

zygoloid (approved):

PR #3910, opened 2024-04-23
Remove the builtin IR, and instead define builtin types locally.

zygoloid:
We don't need it any more, and removing it simplifies a few things:

- One fewer predefined `File` and reserved ID.
- We now have simply `Builtin` instructions for builtins, instead of having an `ImportRef` that indirectly references a `Builtin`.
- `ConstantId`s now always refer directly to a local constant, instead of sometimes referring to an `ImportRef` for a constant in the builtins IR.

### DIFF
--- a/toolchain/sem_ir/file.cpp
+++ b/toolchain/sem_ir/file.cpp
@@ -72,7 +72,7 @@ File::File(SharedValueStores& value_stores, std::string filename)
   auto api_placeholder_id =
       import_irs_.Add({.node_id = Parse::NodeId::Invalid, .sem_ir = nullptr});
   CARBON_CHECK(api_placeholder_id == ImportIRId::ApiForImpl)
-      << "ApiForImpl must be the second IR";
+      << "ApiForImpl must be the first IR";
 
   insts_.Reserve(BuiltinKind::ValidCount);
 // Error uses a self-referential type so that it's not accidentally treated as
--- a/toolchain/sem_ir/file.cpp
+++ b/toolchain/sem_ir/file.cpp
@@ -62,39 +62,19 @@ auto TypeInfo::Print(llvm::raw_ostream& out) const -> void {
   out << "{constant: " << constant_id << ", value_rep: " << value_repr << "}";
 }
 
-File::File(SharedValueStores& value_stores, std::string filename,
-           const File* builtins, llvm::function_ref<void()> init_builtins)
+File::File(SharedValueStores& value_stores, std::string filename)
     : value_stores_(&value_stores),
       filename_(std::move(filename)),
       type_blocks_(allocator_),
       constant_values_(ConstantId::NotConstant),
       inst_blocks_(allocator_),
       constants_(*this, allocator_) {
-  CARBON_CHECK(builtins != nullptr);
-  auto builtins_id =
-      import_irs_.Add({.node_id = Parse::NodeId::Invalid, .sem_ir = builtins});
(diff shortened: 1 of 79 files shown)


toolchain/sem_ir/file.cpp
```diff
@@ -62,39 +62,19 @@ auto TypeInfo::Print(llvm::raw_ostream& out) const -> void {
...
-  CARBON_CHECK(builtins != nullptr);
-  auto builtins_id =
-      import_irs_.Add({.node_id = Parse::NodeId::Invalid, .sem_ir = builtins});
-  CARBON_CHECK(builtins_id == ImportIRId::Builtins)
-      << "Builtins must be the first IR";
-
   auto api_placeholder_id =
       import_irs_.Add({.node_id = Parse::NodeId::Invalid, .sem_ir = nullptr});
   CARBON_CHECK(api_placeholder_id == ImportIRId::ApiForImpl)
       << "ApiForImpl must be the second IR";
```
jonmeow:
```suggestion
      << "ApiForImpl must be the first IR";
```

jonmeow (approved):
The model for imports that you've suggested is that everything should point to a canonical IR location. This seems to be going in a different direction: builtins are equal everywhere, but rather than being in a canonical IR location, they are fully special-cased instructions that have no singular IR address. This is a simplification since we typically special-case them, although this feels somewhat inconsistent with the design direction put forward in #3833 of prioritizing fidelity first.

That said, I think this change is okay; I think the simplification is probably more valuable than what I would call fidelity.

zygoloid:
> The model for imports that you've suggested is that everything should point to a canonical IR location. This seems to be going in a different direction: builtins are equal everywhere, but rather than being in a canonical IR location, they are fully special-cased instructions that have no singular IR address.

FWIW, I'm thinking of these builtins as more like constant values (like `3`) than entities with an owning file and canonical IR location. I'm also anticipating that we'll replace at least some of the builtin types with builtin functions that return types -- we're nearly there with `IntType` and `FloatType`.

jonmeow:
I think of them a little differently... With `type` in particular, it's more like an actual type. That is, if I write `class C;` in separate files, those are distinct types and lead to a collision. For contrast, if I write `type` in separate files, it is the same type type. It's not really performing a conversion. `bool` and `namespace` may be in a similar class, although it's hard to tell.

That said, builtin_kinds.def does have a TODO. That TODO was originally there for the specific `IntType` and `FloatType` examples that you mention. I think the anticipation of removal should be clear from the TODO, as well as pending changes.

By using those as a basis, are you saying that you would expect *every* entry to be removed?

jonmeow:
Maybe the better way for me to put this is that I think of the constant value as the output of an instruction. Typically constant values for types are the same output because they come from the same instruction, even through an import chain. Before, there was an import chain for builtins. Now, they're the same value because builtins are special-cased.

zygoloid:
> Maybe the better way for me to put this is that I think of the constant value as the output of an instruction. Typically constant values for types are the same output because they come from the same instruction, even through an import chain. Before, there was an import chain for builtins. Now, they're the same value because builtins are special-cased.

I think they don't *need* to be special-cased. We could import a `Builtin` instruction the same way we import anything else, and that would also work -- we deduplicate constants, and by extension types, and that would be enough to deduplicate builtin types. But special-casing them avoids creating extra `ImportRef`s and `Builtin` instructions, so it's probably worth doing anyway.

PR #3911, opened 2024-04-24
Use def file to avoid listing all non-type insts in type switches.

zygoloid:
Depends on #3910.


jonmeow (approved):
Looks good!

Is it worth using this to assert the presence of type_id on the relevant instruction?

jonmeow:
Just to note, not expecting such a check here, only didn't merge because of deps.

PR #3912, opened 2024-04-24
Don't canonicalize prefixes when locating the driver binary.

zygoloid:
The approach here directly follows the one used by Clang's driver.

### DIFF
--- a/toolchain/driver/driver_main.cpp
+++ b/toolchain/driver/driver_main.cpp
@@ -9,6 +9,7 @@
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Support/Path.h"
+#include "llvm/Support/Program.h"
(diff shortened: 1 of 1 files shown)


toolchain/driver/driver_main.cpp
```diff
@@ -18,6 +19,16 @@ auto main(int argc, char** argv) -> int {
     return EXIT_FAILURE;
   }
 
+  // Find the executable without resolving symlinks. Do a PATH lookup if argv[0]
+  // isn't a valid path.
+  llvm::SmallString<128> exe_path(argv[0]);
```
chandlerc, line 24: Guessing you want to use `exe_path` below and remove the call to `getMainExecutable`?

zygoloid: Great idea :-)

Done.

chandlerc (approved):
LG!

PR #3915, opened 2024-04-25
Start adding tracking of the complete list of IRs under check.

jonmeow:
As we talk about how imports should work, we want to start having indirectly imported IRs in `import_irs`, where it's currently only directly imported IRs. The tracking of `CheckIRId` is set up to be able to determine whether an IR being indirectly imported is actually already tracked (and may be a direct import). Changes here to how ApiForImpl is handled start using the logic.

Note that indirect imports may be added while processing even the first import from the current library. As a consequence, it's necessary to prepare this map before starting imports, in particular, setting ApiForImpl before any indirect imports may encounter the same import. Also, prior validation that `num_irs` matches the final size are no longer relevant.

check_ir_id is tracked on SemIR because I want to be able to figure it out given an ImportIR, but the mapping is ephemeral after checking and so I store that in Context.

I'm putting relevant logic into import_ref.cpp because the primary alternative is import.cpp, and as more logic is added, ImportRefResolver will be adding import IRs.

### DIFF
--- a/toolchain/sem_ir/ids.h
+++ b/toolchain/sem_ir/ids.h
@@ -220,6 +220,16 @@ struct FunctionId : public IdBase, public Printable<FunctionId> {
 
 constexpr FunctionId FunctionId::Invalid = FunctionId(InvalidIndex);
 
+// The ID of an IR within the set of all IRs being evaluated in the current
+// check execution.
+struct CheckIRId : public IdBase, public Printable<CheckIRId> {
+  using IdBase::IdBase;
+  auto Print(llvm::raw_ostream& out) const -> void {
+    out << "check_ir";
+    IdBase::Print(out);
+  }
+};
+
 // The ID of a class.
 struct ClassId : public IdBase, public Printable<ClassId> {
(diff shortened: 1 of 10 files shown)


jonmeow:
Depends on #3913

toolchain/sem_ir/ids.h
```diff
@@ -268,7 +278,7 @@ struct ImplId : public IdBase, public Printable<ImplId> {
 
 constexpr ImplId ImplId::Invalid = ImplId(InvalidIndex);
 
-// The ID of an imported IR.
+// The ID of an IR within the set of imported IRs, both direct and indirect.
```
zygoloid, line 281: Is this comment right? Looking at `InitPackageScopeAndImports`, it looks to me like we only assign `ImportIRId`s to directly imported IRs.

zygoloid (approved):

toolchain/sem_ir/ids.h
zygoloid: Oh, I see this changes in the next PR.

PR #3916, opened 2024-04-26
Separate constant emission from function emission.

zygoloid:
Make constant emission non-recursive, and stop building a bogus FunctionContext to emit constants.

To support this, move `InstConstantKind` from the typed instruction definition into the `.def` file, and add more macros to allow us to generate case labels based on whether an instruction is a constant.


toolchain/sem_ir/inst_kind.def
```diff
@@ -11,14 +11,31 @@
...
-// default expand to their argument:
+// The invocation of the above macro will be wrapped in one macro from each of
+// the following sets, which by default expand to their argument:
+//
+// Whether the instruction can define a type:
 // - CARBON_SEM_IR_INST_KIND_TYPE(...)
-//   Invoked for each instruction that might define a type constant.
+//   Invoked for each instruction that is always of type `type`, and might
+//   define a type constant.
+// - CARBON_SEM_IR_INST_KIND_MAYBE_TYPE(...)
```
jonmeow, line 21: The CONSTANT macros seem to be using a prefix, whereas TYPE is suffixed.

WDYT of MAYBE_TYPE -> TYPE_IN_CONSTANT or TYPE_CONDITIONAL, NOT_TYPE -> TYPE_NEVER (consistent with CONSTANT_NEVER)?

toolchain/sem_ir/inst_kind.def
```diff
@@ -11,14 +11,31 @@
...
+//   Invoked for each instruction that might define a type constant if it's of
+//   type `type`, but isn't necessarily of type `type`.
```
jonmeow: The current "isn't necessarily of type `type`" phrasing feels slightly odd to me. e.g., if the type constant is of type `type`, wouldn't the type be `type`? Is it needed, or can we say similar without it?

```suggestion
//   Invoked for each instruction that might define a type constant if it's of
//   type `type`.
```

or more closely echoing above phrasing:

```suggestion
//   Invoked for each instruction that is *sometimes* of type `type`, and might define a a type constant.
```

toolchain/lower/handle_type.cpp
```diff
@@ -7,64 +7,22 @@
...
-  context.SetLocal(inst_id, context.GetTypeAsValue());
-}
-
-auto HandleConstType(FunctionContext& context, SemIR::InstId inst_id,
-                     SemIR::ConstType /*inst*/) -> void {
-  context.SetLocal(inst_id, context.GetTypeAsValue());
-}
+// For potentially non-constant instructions that are always of type `type`
+// and might define a type constant, produce the trivial runtime representation
+// of type `type`.
```
jonmeow: FWIW, the "potentially" here throws me a little. Would the below suggestion carry similar meaning for you, trying to say similar with less?

```suggestion
// For instructions that are always of type `type`, produce the trivial runtime representation
// of type `type`.
```

toolchain/lower/function_context.cpp
```diff
@@ -60,11 +60,16 @@ auto FunctionContext::LowerInst(SemIR::InstId inst_id) -> void {
...
   switch (inst.kind()) {
+#define CARBON_SEM_IR_INST_KIND_CONSTANT_ALWAYS(Name)
 #define CARBON_SEM_IR_INST_KIND(Name)                     \
   case SemIR::Name::Kind:                                 \
     Handle##Name(*this, inst_id, inst.As<SemIR::Name>()); \
     break;
 #include "toolchain/sem_ir/inst_kind.def"
+
+    default:
+      CARBON_FATAL() << "Missing constant value for constant instruction "
```
jonmeow, line 72: Is this missing, or not expected to be called with an always-constant instruction?

toolchain/lower/file_context.h
```diff
@@ -91,16 +91,19 @@ class FileContext {
 
   // Maps callables to lowered functions. SemIR treats callables as the
   // canonical form of a function, so lowering needs to do the same.
-  llvm::SmallVector<llvm::Function*> functions_;
+  // We resize this directly to the correct size rather than using a small size.
```
jonmeow: Isn't this a little misleading about the behavior? If a file has something like 64 functions or types, wouldn't it avoid a heap allocation?

FWIW, maybe:

```suggestion
  // We resize this directly to the correct size, and don't try to fit into the same allocation.
```

or just:

```suggestion
  // We resize this directly to the (often large) correct size.
```

(and below)

toolchain/lower/testdata/function/call/struct_param.carbon
```diff
@@ -13,7 +13,7 @@ fn Main() {
 // CHECK:STDOUT: ; ModuleID = 'struct_param.carbon'
 // CHECK:STDOUT: source_filename = "struct_param.carbon"
 // CHECK:STDOUT:
-// CHECK:STDOUT: @struct = internal constant { i32, i32 } { i32 2, i32 3 }
```
jonmeow, line 16: It's curious to me that the `@struct`, `@tuple` names are lost. Is that something that can be avoided? It feels like it reduces clarity of the LLVM IR.

toolchain/lower/constant.cpp
```diff
@@ -0,0 +1,270 @@
...
+      ConstantContext& context, SemIR::Name /*inst*/) -> llvm::Constant* { \
+    return context.GetTypeAsValue();                                       \
+  }
+#include "toolchain/sem_ir/inst_kind.def"
+
+// Emits an aggregate constant of LLVM type `Type` whose elements are the
+// contents of `refs_id`.
+template <typename ConstantType, typename Type>
+static auto EmitAggregateConstant(ConstantContext& context,
+                                  SemIR::InstBlockId refs_id, Type* llvm_type)
```
jonmeow, line 100: Maybe this is just me, but I'd lean towards these as member functions on ConstantContext rather than static functions. I'm not sure how much you're getting from the separation: it feels like it's actually increasing the code.

Alternately, given there's not file separation, maybe a big `switch` statement for a lot of these (rather than the function separation) would actually be okay? It might actually result in less code overall, considering duplication.

toolchain/lower/constant.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+    auto inst = file_context.sem_ir().insts().Get(const_id.inst_id());
+    llvm::Constant* value = nullptr;
+    switch (inst.kind()) {
+#define CARBON_SEM_IR_INST_KIND_CONSTANT_NEVER(...)
+#define CARBON_SEM_IR_INST_KIND_CONSTANT_SYMBOLIC_ONLY(...)
+#define CARBON_SEM_IR_INST_KIND(Name)                                \
+  case SemIR::Name::Kind:                                            \
+    value = Emit##Name##AsConstant(context, inst.As<SemIR::Name>()); \
+    break;
```
jonmeow, line 259:
```suggestion
    CARBON_KIND_SWITCH(inst) {
#define CARBON_SEM_IR_INST_KIND_CONSTANT_NEVER(...)
#define CARBON_SEM_IR_INST_KIND_CONSTANT_SYMBOLIC_ONLY(...)
#define CARBON_SEM_IR_INST_KIND(Name)                                \
  case CARBON_KIND(SemIR::Name const_inst):                                            \
    value = Emit##Name##AsConstant(context, const_inst); \
    break;
```

toolchain/lower/constant.cpp
```diff
@@ -0,0 +1,270 @@
...
+                                        SemIR::StringLiteral inst)
+    -> llvm::Constant* {
+  CARBON_FATAL() << "TODO: Add support: " << inst;
+}
+
+static auto EmitStructTypeFieldAsConstant(ConstantContext& /*context*/,
+                                          SemIR::StructTypeField /*inst*/)
+    -> llvm::Constant* {
+  // A StructTypeField isn't a value, so this constant value won't ever be used.
+  return nullptr;
```
jonmeow: Why isn't this returning GetUnusedConstant?

jonmeow (approved):
Interesting, nice how this works.

toolchain/sem_ir/inst_kind.def
zygoloid: Sure:
```suggestion
//   Invoked for each instruction that is sometimes of type `type`, and might define a type constant.
```

toolchain/sem_ir/inst_kind.def
```diff
@@ -11,14 +11,30 @@
...
+// The invocation of the above macro will be wrapped in one macro from each of
+// the following sets, which by default expand to their argument:
+//
+// Whether the instruction can define a type:
 // - CARBON_SEM_IR_INST_KIND_TYPE(...)
-//   Invoked for each instruction that might define a type constant.
+//   Invoked for each instruction that is always of type `type`, and might
+//   define a type constant.
+// - CARBON_SEM_IR_INST_KIND_MAYBE_TYPE(...)
+//   Invoked for each instruction that is sometimes of type `type`, and might define a type constant.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
//   Invoked for each instruction that is sometimes of type `type`, and might
//   define a type constant.
```

toolchain/lower/constant.cpp
zygoloid: My intent was to model this directly after `FunctionContext` and `Handle*` for consistency. Given the patterns we use elsewhere in lower and more broadly in the toolchain -- of having either a `switch` or a bunch of free functions, and not having a collection of member functions that grow with the number of kinds of instruction / node / state / whatever -- I'd be reluctant to make these member functions.

I think one big `switch` would work, and would be consistent with patterns we use elsewhere, though it'd be a fairly large function. My main concern there is that I'd expect it to continue to grow over time, and it'd seem at least on the borderline of being too large for a single function as-is. I can give it a go and see how large a function we get, as a follow up.

toolchain/lower/constant.cpp
zygoloid: I'd considered making the same change to `FunctionContext::LowerInst`, but my feeling at the time was that the macros weren't really helping. But maybe for consistency we should be using them everywhere they apply?

toolchain/lower/file_context.h
zygoloid: I meant "small size" as in the "small size optimization" -- I don't think that's misleading, but I can see how you might instead read it as "a size that's too small for the content" or something like that. I'll use the latter suggestion.

toolchain/lower/function_context.cpp
zygoloid: `LowerBlock` would skip an always-constant instruction. But I think this'd be clearer if I moved that `if` into `LowerInst`; I've done that.

toolchain/lower/handle_type.cpp
zygoloid: I'm fine with this change. For me the new comment doesn't explain why we filter out the `CONSTANT_ALWAYS` case below, but perhaps that's obvious enough that it's not worth calling out.

toolchain/sem_ir/inst_kind.def
zygoloid: Sounds good, I'll send a separate PR.

toolchain/lower/testdata/function/call/struct_param.carbon
zygoloid: We can make `InstNamer` include them in the names of constants. I'll send a separate PR for that.

toolchain/lower/constant.cpp
zygoloid: Added a comment to explain.

toolchain/sem_ir/inst_kind.def
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
//   Invoked for each instruction that is sometimes of type `type`, and might
//   define a type constant.
```

toolchain/lower/handle_type.cpp
```diff
@@ -7,64 +7,21 @@
...
-                     SemIR::ClassType /*inst*/) -> void {
-  context.SetLocal(inst_id, context.GetTypeAsValue());
-}
-
-auto HandleConstType(FunctionContext& context, SemIR::InstId inst_id,
-                     SemIR::ConstType /*inst*/) -> void {
-  context.SetLocal(inst_id, context.GetTypeAsValue());
-}
+// For instructions that are always of type `type`, produce the trivial runtime representation
+// of type `type`.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
// For instructions that are always of type `type`, produce the trivial runtime
// representation of type `type`.
```

toolchain/sem_ir/inst_kind.def
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
//   Invoked for each instruction that is sometimes of type `type`, and might
//   define a type constant.
```

PR #3919, opened 2024-04-26
Use `CARBON_KIND_SWITCH` for inst dispatch in `FunctionContext` and `ConstantContext`.

zygoloid:
As requested in #3916.


josh11b (approved):

zygoloid:
FWIW: I'm pretty neutral on this change. I don't think the macros really help here and they'll add some compile-time cost, but it seems to be pretty negligible and it's more consistent to always use them wherever they'd apply.

jonmeow:
FWIW I agree it's small, but the consistency seems worth it.

PR #3921, opened 2024-04-27
Remove unused include statement in driver.cpp

prabhatexit0:
This removes unused include
```
```
from `./toolchain/driver/driver.cpp`


jonmeow (approved):
Thanks. For reference, this unused because some contents were split to other files; still, it's "unused" because diagnostic types are indirectly gotten from sorting_diagnostic_consumer instead, and that's directly used so it's probably fine.

PR #3922, opened 2024-04-29
Introduce basic linking with the Clang driver.

chandlerc:
This adds the most rudimentary support for linking in the Carbon driver by calling out to the Clang driver and letting it do the linking. This is a super minimal version to start with, just enough to be somewhat useful and to exercise calling into Clang for this.

Also adds a Clang runner to help run the Clang driver. This is a bit more complete, although it sill needs some significant work. For example, it will need some significant enhancements to be able to compile C++ code, etc.

One thing that will likely change here is to better manage streams and work better as a library. For now, this is a bit limited because Clang and LLVM don't provide the necessary support for this.

### DIFF
--- a/toolchain/driver/clang_runner.cpp
+++ b/toolchain/driver/clang_runner.cpp
@@ -1,0 +1,152 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "toolchain/driver/clang_runner.h"
+
+#include <algorithm>
+#include <memory>
+#include <numeric>
+#include <optional>
+
+#include "clang/Basic/Diagnostic.h"
+#include "clang/Basic/DiagnosticOptions.h"
+#include "clang/Driver/Compilation.h"
+#include "clang/Driver/Driver.h"
+#include "clang/Frontend/CompilerInvocation.h"
+#include "clang/Frontend/TextDiagnosticPrinter.h"
+#include "common/command_line.h"
+#include "common/vlog.h"
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/ScopeExit.h"
+#include "llvm/ADT/StringExtras.h"
+#include "llvm/ADT/StringRef.h"
+#include "llvm/IR/LLVMContext.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/Path.h"
+#include "llvm/Support/Program.h"
+#include "llvm/Support/VirtualFileSystem.h"
+#include "llvm/TargetParser/Host.h"
+
+namespace Carbon {
+
+static auto GetExecutablePath(llvm::StringRef exe_name) -> std::string {
+  // If the `exe_name` isn't already a valid path, look it up.
+  if (!llvm::sys::fs::exists(exe_name)) {
+    if (llvm::ErrorOr<std::string> path_result =
+            llvm::sys::findProgramByName(exe_name)) {
+      return *path_result;
+    }
+  }
+
+  return exe_name.str();
+}
+
+ClangRunner::ClangRunner(llvm::StringRef exe_name, llvm::StringRef target,
+                         llvm::raw_ostream* vlog_stream)
+    : exe_name_(exe_name),
+      exe_path_(GetExecutablePath(exe_name)),
+      target_(target),
+      vlog_stream_(vlog_stream),
+      diagnostic_ids_(new clang::DiagnosticIDs()) {}
+
+auto ClangRunner::Run(llvm::ArrayRef<llvm::StringRef> args) -> bool {
+  // TODO: Maybe handle response file expansion similar to the Clang CLI?
+
+  // If we have a verbose logging stream, and that stream is the same as
+  // `llvm::errs`, then add the `-v` flag so that the driver also prints verbose
+  // information.
+  bool inject_v_arg = vlog_stream_ == &llvm::errs();
+  std::array<llvm::StringRef, 1> v_arg_storage;
+  llvm::ArrayRef<llvm::StringRef> maybe_v_arg;
+  if (inject_v_arg) {
+    v_arg_storage[0] = "-v";
+    maybe_v_arg = v_arg_storage;
+  }
+
+  CARBON_CHECK(!args.empty());
+  CARBON_VLOG() << "Running Clang driver with arguments: \n";
+
+  // Render the arguments into null-terminated C-strings for use by the Clang
+  // driver. Command lines can get quite long in build systems so this tries to
+  // minimize the memory allocation overhead.
+  std::array<llvm::StringRef, 1> exe_arg = {exe_name_};
+  auto args_range =
+      llvm::concat<const llvm::StringRef>(exe_arg, maybe_v_arg, args);
+  int total_size = 0;
+  for (llvm::StringRef arg : args_range) {
+    // Accumulate both the string size and a null terminator byte.
+    total_size += arg.size() + 1;
+  }
+
+  // Allocate one chunk of storage for the actual C-strings and a vector of
+  // pointers into the storage.
+  llvm::OwningArrayRef<char> cstr_arg_storage(total_size);
+  llvm::SmallVector<const char*, 64> cstr_args;
+  cstr_args.reserve(args.size() + inject_v_arg + 1);
+  for (ssize_t i = 0; llvm::StringRef arg : args_range) {
+    cstr_args.push_back(&cstr_arg_storage[i]);
+    memcpy(&cstr_arg_storage[i], arg.data(), arg.size());
+    i += arg.size();
+    cstr_arg_storage[i] = '\0';
+    ++i;
+  }
+  for (const char* cstr_arg : llvm::ArrayRef(cstr_args).drop_front()) {
+    CARBON_VLOG() << "    '" << cstr_arg << "'\n";
+  }
+
+  CARBON_VLOG() << "Preparing Clang driver...\n";
+
+  // Create the diagnostic options and parse arguments controlling them out of
+  // our arguments.
+  llvm::IntrusiveRefCntPtr<clang::DiagnosticOptions> diagnostic_options =
+      clang::CreateAndPopulateDiagOpts(cstr_args);
+
+  // TODO: We don't yet support serializing diagnostics the way the actual
+  // `clang` command line does. Unclear if we need to or not, but it would need
+  // a bit more logic here to set up chained consumers.
+  clang::TextDiagnosticPrinter diagnostic_client(llvm::errs(),
+                                                 diagnostic_options.get());
+
+  clang::DiagnosticsEngine diagnostics(
+      diagnostic_ids_, diagnostic_options.get(), &diagnostic_client,
+      /*ShouldOwnClient=*/false);
+  clang::ProcessWarningOptions(diagnostics, *diagnostic_options);
+
+  clang::driver::Driver driver(exe_path_, target_, diagnostics);
+
+  // TODO: Directly run in-process rather than using a subprocess. This is both
+  // more efficient and makes debugging (much) easier. Needs code like:
+  // driver.CC1Main = [](llvm::SmallVectorImpl<const char*>& argv) {};
+  std::unique_ptr<clang::driver::Compilation> compilation(
+      driver.BuildCompilation(cstr_args));
+  CARBON_CHECK(compilation) << "Should always successfully allocate!";
+  if (compilation->containsError()) {
+    // These should have been diagnosed by the driver.
+    return false;
+  }
+
+  CARBON_VLOG() << "Running Clang driver...\n";
+
+  llvm::SmallVector<std::pair<int, const clang::driver::Command*>>
+      failing_commands;
+  int result = driver.ExecuteCompilation(*compilation, failing_commands);
+
+  // Finish diagnosing any failures before we verbosely log the source of those
+  // failures.
+  diagnostic_client.finish();
+
+  CARBON_VLOG() << "Execution result code: " << result << "\n";
+  for (const auto& [command_result, failing_command] : failing_commands) {
+    CARBON_VLOG() << "Failing command '" << failing_command->getExecutable()
+                  << "' with code '" << command_result << "' was:\n";
+    if (vlog_stream_) {
+      failing_command->Print(*vlog_stream_, "\n\n", /*Quote=*/true);
+    }
+  }
+
+  // Return whether the command was executed successfully.
+  return result == 0 && failing_commands.empty();
+}
+
+}  // namespace Carbon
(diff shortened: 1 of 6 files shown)


toolchain/driver/driver.h
```diff
@@ -70,7 +72,12 @@ class Driver {
   auto ValidateCompileOptions(const CompileOptions& options) const -> bool;
 
   // Implements the compile subcommand of the driver.
-  auto Compile(const CompileOptions& options) -> RunResult;
+  auto Compile(const CompileOptions& options, const CodegenOptions& cg_options)
```
jonmeow: Maybe codegen_options? "cg" isn't an abbreviation I'm familiar enough with for an API; I think "computer graphics", but maybe I'm not familiar enough with compilers.

toolchain/driver/clang_runner.cpp
```diff
@@ -0,0 +1,152 @@
...
+  // Factor in the executable name and all the null bytes.
+  total_size += exe_name_.size() + args.size() + 1;
+  if (inject_v_flag) {
+    total_size += strlen("-v") + 1;
+  }
+
+  // Allocate one chunk of storage for the actual C-strings and a vector of
+  // pointers into the storage.
+  llvm::OwningArrayRef<char> cstr_arg_storage(total_size);
+  llvm::SmallVector<const char*, 64> cstr_args;
```
jonmeow, line 85: Isn't 64 going to be the default value? Why do you specify it here?

toolchain/driver/clang_runner.cpp
```diff
@@ -0,0 +1,152 @@
...
+  // driver. Command lines can get quite long in build systems so this tries to
+  // minimize the memory allocation overhead.
+  int total_size = std::accumulate(
+      args.begin(), args.end(), 0,
+      [](int sum, llvm::StringRef arg) { return sum + arg.size(); });
+  // Factor in the executable name and all the null bytes.
+  total_size += exe_name_.size() + args.size() + 1;
+  if (inject_v_flag) {
+    total_size += strlen("-v") + 1;
+  }
```
jonmeow: There were a couple things that I was wondering about this:
- Why std::accumulate instead of a range-based for loop? 
- Is the `+ args.size()` for performance? I was having trouble reading this, how significant would the trade-off be to do `+ 1`? (alternately, making the comment more specific to `args.size()` might help)
- Had you considered something like `maybe_v_flag[0].size()` instead of `strlen("-v")`?

But I'm wondering, had you considered the below loop (reusing `args_range` there):

```
  auto args_range = llvm::concat<const llvm::StringRef>(
      llvm::ArrayRef({exe_name_}), maybe_v_flag, args);
  int total_size = 0;
  for (llvm::StringRef arg : args_range) {
    total_size += arg.size() + 1;
  }
```

toolchain/driver/clang_runner.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+  // Finish diagnosing any failures before we verbosely log the source of those
+  // failures.
+  diagnostic_client.finish();
+
+  CARBON_VLOG() << "Execution result code: " << result << "\n";
+  for (const auto& [command_result, failing_command] : failing_commands) {
+    CARBON_VLOG() << "Failing command '" << failing_command->getExecutable()
+                  << "' with code '" << command_result << "' was:\n";
+    if (verbose()) {
```
jonmeow: FWIW, this seems to only be done in this one spot, it might be worth inlining to reduce how much needs to be read.

toolchain/driver/clang_runner.h
```diff
@@ -0,0 +1,62 @@
...
+#include "common/ostream.h"
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/StringRef.h"
+
+namespace Carbon {
+
+// Runs Clang in a similar fashion to invoking it with the provided arguments on
+// the command line. We use a textual command line interface to allow easily
+// incorporating custom command line flags from user invocations that we don't
+// parse but pass transparently along to Clang itself.
```
jonmeow:
```suggestion
// parse, but will pass transparently along to Clang itself.
```

toolchain/driver/clang_runner.h
```diff
@@ -0,0 +1,62 @@
...
+//
+// This doesn't literally use a subprocess to invoke Clang, it instead tries to
```
jonmeow:
```suggestion
// This doesn't literally use a subprocess to invoke Clang; it instead tries to
```

toolchain/driver/clang_runner_test.cpp
```diff
@@ -0,0 +1,150 @@
...
+  const auto* unit_test = ::testing::UnitTest::GetInstance();
+  const auto* test_info = unit_test->current_test_info();
+  std::filesystem::path test_file =
+      test_tmpdir / llvm::formatv("{0}_{1}_{2}", test_info->test_suite_name(),
+                                  test_info->name(), name_suffix)
+                        .str();
+  // Make debugging a bit easier by cleaning up any files from previous runs.
+  // This is only necessary when not run in Bazel's test environment.
+  std::filesystem::remove(test_file);
+  CARBON_CHECK(!std::filesystem::exists(test_file));
```
jonmeow, line 100: Had you considered using gtest's teardown to remove created files, rather than deleting name collisions this way? You might then also be able to create a simpler filename, rather than the complex formula.

Note this may also be masking a duplicate "foo.o" mistake below.

toolchain/driver/clang_runner_test.cpp
```diff
@@ -0,0 +1,150 @@
...
+// Clang driver because they work hard to interact with the underlying
+// filesystem and operating system. For now, we just check that a link command
+// is echoed back with plausible contents.
+//
+// TODO: We should eventually strive to have a more complete setup that lets us
+// test more complete Clang functionality here.
+TEST(ClangRunnerTest, LinkCommandEcho) {
+  // Just create some empty files to use in a synthetic link command below.
+  std::filesystem::path foo_file = WriteTestFile("foo.o", "");
+  std::filesystem::path bar_file = WriteTestFile("foo.o", "");
```
jonmeow: Is this intentionally using the same filename? It looks like the code would create one, then delete it, then create it again.

toolchain/driver/driver.cpp
```diff
@@ -830,4 +901,59 @@ auto Driver::Compile(const CompileOptions& options) -> RunResult {
...
+      args.push_back(
+          "--sysroot=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk");
+      // We also need to insist on a modern linker, otherwise the driver tries
+      // to old and deprecated flags. The specific number here comes from an
+      // inspection of the Clang driver source code to understand where features
+      // were enabled, and this appears to be the latest version to control
+      // driver behavior.
+      //
+      // TODO: We should replace this with use of `lld` eventually.
+      args.push_back("-mlinker-version=705");
```
jonmeow, line 924: Is it worth validating that these aren't already in args?

jonmeow (commented):
This looks really good!

jonmeow:
Is it worth updating the sieve compile to use this? Or is that more a separate PR?

toolchain/driver/driver.h
chandlerc: Sure. It's familiar to me, but not important.

toolchain/driver/driver.cpp
chandlerc: My mental model for all the arg manipulation is basically around appending-to-override. And so the idea is that this will override any earlier versions, but we can allow further appending to override these back again.

I don't think the first will come up, so I can check that these don't exist yet, but it'll add some expense at minimal benefit.

I expect us to potentially want to allow overriding of these with later flags though. This is essentially setting up the defaults that users might override.

It's not in this PR, but thinking of exposing something like `--clang-option=-xyz` that lets us pass arbitrary extra flags to Clang long term.

toolchain/driver/clang_runner_test.cpp
chandlerc: Errr... no... good catch.

toolchain/driver/clang_runner.cpp
chandlerc: Sorry, some of this complexity got added later on in the evolution of teh PR, I can try to simplify. I like the idea of extracting the entire range now that we have so many things in it.

My inclination is always to use an algorithm rather than a range based for loop here, but the code for that gets quite bad here so switched to a loop.

toolchain/driver/clang_runner.h
josh11b: FYI the original text reads fine to me.

toolchain/driver/clang_runner_test.cpp
josh11b: Likely this is supposed to be:
```suggestion
  std::filesystem::path foo_file = WriteTestFile("foo.o", "");
  std::filesystem::path bar_file = WriteTestFile("bar.o", "");
```

toolchain/driver/driver.cpp
```diff
@@ -829,4 +900,59 @@ auto Driver::Compile(const CompileOptions& options) -> RunResult {
...
+      // On macOS we need to set the sysroot to a viable SDK. Currently, this
+      // hard codes the path to be the unversioned symlink. The prefix is also
+      // hard coded in Homebrew and so this seems likely to work reasonably
+      // well. Homebrew and I suspect the Xcode Clang both have this hard coded
+      // at build time, so this seems reasonably safe but we can revisit if/when
+      // needed.
+      args.push_back(
+          "--sysroot=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk");
+      // We also need to insist on a modern linker, otherwise the driver tries
+      // to old and deprecated flags. The specific number here comes from an
```
josh11b: Some mistake here, perhaps:
```suggestion
      // We also need to insist on a modern linker, otherwise the driver tries
      // too old and deprecated flags. The specific number here comes from an
```

toolchain/driver/clang_runner.cpp
chandlerc: These are the *pointers*, I wouldn't expect the small size to be 64 pointers? I thought it was roughly a cache line (8 pointers)?

toolchain/driver/clang_runner.cpp
chandlerc: Sure. This too was a holdover from an older factoring of everything with different stream setup.

toolchain/driver/clang_runner_test.cpp
chandlerc: I understand this masked the error, and happy to consider how to avoid this.

However, I don't quite see how using teardown would help with the complex temporary formula... I added this in another test and I thought this was needed in part because of test sharding and such.... But maybe I'm mis-remembering.

And when debugging, it was specifically useful to support either `TEST_TMPDIR` or the system temporary directory, and to have the file be left around after runs.

So I could switch this to the teardown, but I think it'd be somewhat of a wash in terms of code complexity and as I think the complexity above would still be needed, and would trade off debugging a specific failure for masking the typo.... I'm not really sure that's a benefit?

But also not hugely attached to the particular solution I have here, so maybe you see something that is more all-around better?

toolchain/driver/clang_runner_test.cpp
chandlerc: (yes, Jon already flagged and I fixed)

chandlerc (commented):
(think I got all the comments, PTAL)

toolchain/driver/clang_runner.cpp
jonmeow: Ah, I see, 64 bytes in CalculateSmallVectorDefaultInlinedElements.

Still though, why do you specify 64 here? Is there some method you're using to choose the number of arguments? Maybe a comment could be added to explain the source?

toolchain/driver/clang_runner.cpp
jonmeow: (not sure if changes aren't published, but not seeing edits here)

toolchain/driver/clang_runner_test.cpp
jonmeow: > However, I don't quite see how using teardown would help with the complex temporary formula... I added this in another test and I thought this was needed in part because of test sharding and such.... But maybe I'm mis-remembering.

I've never heard of this issue with sharding: each shard gets its own tmpdir.

That said, you might consider something like `llvm::sys::fs::createUniqueDirectory(std::filesystem::temp_directory_path() / test_info->test_suite_name(), result_path)` [either off the TEST_TMPDIR path, or using test_tmpdir as the prefix instead], then you shouldn't need to remove files and would still have files available after execution.

https://llvm.org/doxygen/namespacellvm_1_1sys_1_1fs.html#abf677459f5137cfb49f6d81b27ea54a1

jonmeow (approved):
Approving, just not the minor thing with `verbose()`.

Also, note in tests:

```
FAIL: //toolchain/driver:clang_runner_test (see /private/var/tmp/_bazel_runner/e591f63ed099023de1f206992dfce127/execroot/_main/bazel-out/darwin_arm64-opt/testlogs/toolchain/driver/clang_runner_test/test.log)
==================== Test output for //toolchain/driver:clang_runner_test:
INFO: From Testing //toolchain/driver:clang_runner_test:
[==========] Running 2 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 2 tests from ClangRunnerTest
[ RUN      ] ClangRunnerTest.Version
================================================================================
```

There seems to be something weird going on there, in opt mode. I'm assuming any fix would be small though.

toolchain/driver/clang_runner.cpp
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  CARBON_CHECK(!args.empty());
+  CARBON_VLOG() << "Running Clang driver with arguments: \n";
+
+  // Render the arguments into null-terminated C-strings for use by the Clang
+  // driver. Command lines can get quite long in build systems so this tries to
+  // minimize the memory allocation overhead.
+  auto args_range = llvm::concat<const llvm::StringRef>(
+      llvm::ArrayRef({exe_name_}), maybe_v_flag, args);
```
jonmeow: I'm wondering, is the lifetime of the ArrayRef bad for the concat? (noting the opt issue)

[I'd been considering suggesting this as a lambda, maybe that'd fix it?]

toolchain/driver/clang_runner.cpp
chandlerc: I tried to add a comment above about avoiding allocations more of the time, in large part to explain using a custom number here.

But not sure I can give a better or more specific number at this point.

Is that OK? I was trying to address previous comments about small vector sizes, but not sure it's actually working....

toolchain/driver/clang_runner.cpp
chandlerc: Nope, missed it, thanks!

toolchain/driver/clang_runner_test.cpp
chandlerc: I did think about that, but it seemed to be even more complexity to be managing directories in addition to files...

Anyways, happy to revisit this if there is a better pattern, or maybe both this and the other place where we have similar technique can be simplified if I misunderstood the issue or there is a better  way to avoid it.

toolchain/driver/clang_runner.cpp
chandlerc: Yeah, pretty sure its the temporary array ref. The lambda sounds like a good approach.

chandlerc:
(merging as discussed, this seems in a good enough state for the initial cut)

PR #3925, opened 2024-04-30
Fix CHECK failure when there is control flow in a global variable initializer

josh11b:
Problem found by fuzzing.


toolchain/check/inst_block_stack.h
```diff
@@ -130,6 +130,9 @@ class InstBlockStack {
   // The size of the stack. Entries after this in `stack_` are kept around so
   // that we can reuse the allocated buffer for their content.
   int size_ = 0;
+
+  // Current global init block to push.
+  SemIR::InstBlockId init_block_id_ = SemIR::InstBlockId::GlobalInit;
```
zygoloid: Maybe reorder this next to `init_block_`?

zygoloid (approved):

toolchain/check/inst_block_stack.h
josh11b: Done.

PR #3926, opened 2024-05-01
Include parse node being checked in crash backtrace.

zygoloid:
When we crash, include the source location of the parse node that we were handling, as well as the name of the check function that we were calling. Also include the source snippet, since it's easy to do so, and may avoid the need to look at the input file in some cases.

### DIFF
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -706,12 +706,28 @@ auto NodeIdTraversal::Next() -> std::optional<Parse::NodeId> {
 // for example if an unrecoverable state is encountered.
 // NOLINTNEXTLINE(readability-function-size)
 static auto ProcessNodeIds(Context& context, llvm::raw_ostream* vlog_stream,
-                           ErrorTrackingDiagnosticConsumer& err_tracker)
-    -> bool {
+                           ErrorTrackingDiagnosticConsumer& err_tracker,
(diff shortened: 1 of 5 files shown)


toolchain/check/check.cpp
```diff
@@ -706,12 +706,28 @@ auto NodeIdTraversal::Next() -> std::optional<Parse::NodeId> {
...
+    auto loc = converter->ConvertLoc(
+        node_id, [](DiagnosticLoc, const Internal::DiagnosticBase<>&) {});
+    int32_t length_this_line = std::min(
+        loc.length, static_cast<int32_t>(loc.line.size()) - loc.column_number);
+    output << loc.filename << ":" << loc.line_number << ":" << loc.column_number
+           << ": Check::Handle" << context.parse_tree().node_kind(node_id)
+           << "\n"
+           << loc.line << "\n"
+           << std::string(std::max(loc.column_number - 1, 0), ' ') << '^'
+           << std::string(std::max(length_this_line - 1, 0), '~');
```
jonmeow: Is this underline logic that could be shared with diagnostics?

jonmeow (approved):

toolchain/check/check.cpp
zygoloid: Good idea. Done.

jonmeow (approved):
Thanks!

PR #3927, opened 2024-05-01
More consistent package syntax

zygoloid:
Change the syntax for `package` declarations from:

```carbon
[package Foo] [library "bar"] api;
[package Foo] [library "bar"] impl;
```

to

```carbon
[package Foo] [library "bar"];
impl [package Foo] [library "bar"];
```

### DIFF
--- a/docs/design/README.md
+++ b/docs/design/README.md
@@ -194,7 +194,7 @@ The `import` declaration imports a
 [library from a package](#files-libraries-packages). It must appear at the top
 of a Carbon source file, the first thing after the
 [optional `package` declaration](#package-declaration). Libraries can optionally
-be split into [api and implementation files](#files-libraries-packages), like
+be split into [API and implementation files](#files-libraries-packages), like
 C++'s header and source files but without requiring a source file in any cases.
 This declaration from the example:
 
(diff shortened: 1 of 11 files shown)


proposals/p3927.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+## Problem
+
+The `package` declaration is currently inconsistent with other Carbon
+declarations:
+
+-   Modifier keywords for other declarations precede the introducer keyword.
+    However, for package declarations, the `api` / `impl` modifier appears at
+    the end of the declaration.
```
chandlerc, line 50: FWIW, this part I have mixed feelings about. It had seemed really nice for us to have a reliably introductory string at the top of the file, and we somewhat lose that.

Maybe we already lost that with the default package?

That said, it seems like this should really go before `library`, not `package` here?

zygoloid: > It had seemed really nice for us to have a reliably introductory string at the top of the file

Is that from an aesthetic perspective, or are there technical advantages to having a fixed initial token sequence that's shared across all files in a library that we should be considering here?

FWIW, having made the change across our documentation for this PR, I find that I do like how `impl package Foo library "bar";` reads -- the first thing that tells me is the kind of file that I'm in, which seems pretty useful, whereas putting it at the end feels more like an afterthought or ignorable detail.

> That said, it seems like this should really go before `library`, not `package` here?

Like so?
```carbon
package Foo impl library "bar";
package Foo impl;
impl library "bar";
```
I can see the logic, but it feels less satisfying to me than the status quo or the proposed approach.

jonmeow: It might also be worth considering how a prefix looks with imports:

```
impl package Foo library "wiz";

import package Pitter library "patter";
import library "bar";
// Discussed options:
export import package Other library "baz";
import reexport package Other library "baz";
```

The positioning is more consistent as a prefix, and not requiring `api` feels like an incremental writability improvement. I expect we would not insert a keyword before `library` inside an `import`, but we do put `import` before `package`.

(tbh, I would be happy to argue `library` in `package ... library` is redundant and should be removed for simplicity)

docs/design/README.md
```diff
@@ -2190,13 +2191,15 @@ selected yet.
...
+    -   `impl` files implement part of the interface of a library.
 -   **Libraries** are the granularity of code reuse through imports.
 -   **Packages** are the unit of distribution.
 
-Each library must have exactly one `api` file. This file includes declarations
-for all public names of the library. Definitions for those declarations must be
-in some file in the library, either the `api` file or an `impl` file.
+Each library must have exactly one API file. This file includes declarations for
+all public names of the library. Definitions for those declarations must be in
+some file in the library, either the API file or an `impl` file.
```
jonmeow: In documentation, perhaps we should say "API or implementation file" instead of "`impl` file" in general? i.e., to consistently use terms, since we can no longer consistently use keywords.

proposals/p3927.md
chandlerc: I was a bit worried about tools that want to look for a magic initial sequence, but I don't think that's too important. This was mostly an aesthetic question for me.

The comparison with `import` is a really nice point though, and seems to really clearly show my idea about `library` was wrong. Especially given that this *does* do an implicit import of exactly the API named with the string following `impl`.

Ok, I think I'm sold on the leading `impl`. =]

docs/design/README.md
zygoloid: Yeah, that seems like a good idea. Done.

proposals/p3927.md
zygoloid: Added an "Alternative considered" for putting `impl` immediately before where `library` is / would be.

Updated the "Alternatives considered" to mention @jonmeow's observation about consistency with `import`.

chandlerc (approved):
LGTM, ship it!

PR #3928, opened 2024-05-01
Remove dead code in HandleInterfaceWitnessAccess

jonmeow:
This is handled through constant logic, and seems to have been missed during constant refactoring, so replacing with FatalErrorIfEncountered.


chandlerc (approved):
LG

PR #3929, opened 2024-05-01
Add VSCode recommendation for carbonizer syntax highlighter by @RohanVashisht1234.

zygoloid:


zygoloid:
Looking at #3930 instead.

PR #3917, opened 2024-04-26
Switch import logic to rely on constants.

jonmeow:
Breaks extern handling, makes other things work. TODOs in import_ref for extern, the tests need rewritten too under the new model.

PR #3931, opened 2024-05-02
Promote FunctionType to a standard instruction.

jonmeow:
This removes the builtin FunctionType, replacing it with a FunctionType instruction. The constant for a FunctionDecl is now a StructValue with type of FunctionType.

Note this means a function declaration produces _both_ a type, and a value of the type. This has some consequences in terms of circularity, and makes the importing of function declarations a little more complex.

It'll get particularly peculiar for imports because of the behavior of the reference, but that's a known issue due to other things such as `alias`. The impact will hopefully be contained to ResolvePrevInstForMerge (and ImportRefs).

To note a small formatting change in diagnostics:

```
-  // CHECK:STDERR: fail_member_lookup.carbon:[[@LINE+4]]:3: ERROR: Value of type `<associated <function> in Interface>` is not callable.
+  // CHECK:STDERR: fail_member_lookup.carbon:[[@LINE+4]]:3: ERROR: Value of type `<associated F in Interface>` is not callable.

-  // CHECK:STDERR: fail_todo_facet_lookup.carbon:[[@LINE+4]]:3: ERROR: Value of type `<associated <function> in Interface>` is not callable.
+  // CHECK:STDERR: fail_todo_facet_lookup.carbon:[[@LINE+4]]:3: ERROR: Value of type `<associated F in Interface>` is not callable.
```


toolchain/check/context.cpp
```diff
@@ -1065,6 +1065,14 @@ auto Context::GetBuiltinType(SemIR::BuiltinKind kind) -> SemIR::TypeId {
   return type_id;
 }
 
+auto Context::GetFunctionType(SemIR::FunctionId fn_id) -> SemIR::TypeId {
+  auto type_id = GetTypeImpl<SemIR::FunctionType>(*this, fn_id);
+  // To keep client code simpler, complete function types before returning them.
+  bool complete = TryToCompleteType(type_id);
+  CARBON_CHECK(complete) << "Failed to complete builtin type";
```
zygoloid:
```suggestion
  CARBON_CHECK(complete) << "Failed to complete function type";
```

toolchain/check/handle_function.cpp
```diff
@@ -60,6 +61,55 @@ static auto DiagnoseModifiers(Context& context, bool is_definition,
...
+static auto GetRedeclFunctionId(Context& context, SemIR::Inst prev_inst)
+    -> SemIR::FunctionId {
+  CARBON_KIND_SWITCH(prev_inst) {
+    case CARBON_KIND(SemIR::StructValue struct_value): {
+      if (auto fn_type = context.types().TryGetAs<SemIR::FunctionType>(
+              struct_value.type_id)) {
+        return fn_type->function_id;
+      }
+      return SemIR::FunctionId::Invalid;
+    }
```
zygoloid, line 74: Do we still need this case at all? I have a suspicion that we stopped needing it as of #3917.

toolchain/check/impl.cpp
```diff
@@ -33,6 +34,7 @@ static auto NoteAssociatedFunction(Context& context,
 static auto CheckAssociatedFunctionImplementation(
     Context& context, SemIR::FunctionId interface_function_id,
     SemIR::InstId impl_decl_id, Substitutions substitutions) -> SemIR::InstId {
+  CARBON_CHECK(!context.insts().Is<SemIR::FunctionType>(impl_decl_id));
```
zygoloid: This seems like an unnecessary CHECK given that we have a diagnostic for the general case where `impl_decl_id` isn't a `FunctionDecl` a couple of lines below. Is it important to CHECK-fail on this, or is this a debugging aid left behind from development?
```suggestion
```

toolchain/check/import_ref.cpp
```diff
@@ -879,11 +881,26 @@ class ImportRefResolver {
...
+    CARBON_CHECK(inst.type_id == SemIR::TypeId::TypeType);
+    auto fn_const_id = GetLocalConstantId(
+        import_ir_.functions().Get(inst.function_id).decl_id);
+    if (HasNewWork(initial_work)) {
+      return ResolveResult::Retry();
+    }
+    auto fn_val =
+        context_.insts().GetAs<SemIR::StructValue>(fn_const_id.inst_id());
+    CARBON_CHECK(context_.types().Is<SemIR::FunctionType>(fn_val.type_id));
+    return {context_.types().GetConstantId(fn_val.type_id)};
```
zygoloid: I think it'd be simpler to import the function type constant here, rather than importing the function value constant and then grabbing its type:
```suggestion
    auto fn_const_id = GetLocalConstantId(
        import_ir_.insts()
                  .Get(import_ir_.functions().Get(inst.function_id).decl_id)
                  .type_id());
    if (HasNewWork(initial_work)) {
      return ResolveResult::Retry();
    }
    return {context_.GetTypeIdForTypeConstant(fn_val.type_id)};
```
This is fine as-is if that doesn't work or is undesirable for some reason.

zygoloid (approved):
Awesome, thanks!

toolchain/check/handle_function.cpp
jonmeow: Yes, because this is the constant value.

zygoloid: Sure, but shouldn't `prev_inst` here be the previous instruction found by name lookup? I would expect that to be a `FunctionDecl`, not a `StructValue`.

In any case, consider this non-blocking: this PR is preserving the existing behavior here.

jonmeow: That's not how it works (see `ResolvePrevInstForMerge`). As a reminder, this is pretty much the same issue that came up when I worked on `class`, discussed here: https://github.com/carbon-language/carbon-lang/pull/3893#discussion_r1574889278

toolchain/check/context.cpp
jonmeow: Done

toolchain/check/impl.cpp
jonmeow: Done (I think I added this while debugging)

toolchain/check/import_ref.cpp
jonmeow: Get(function.decl_id).type_id() will correspond to the StructValue whose type is the FunctionType, creating a circular reference.

By querying the constant of the decl, failures add the decl itself to the resolve stack, which results in the type generation.

jonmeow: (or maybe it's actually more direct than that -- type_id() is this FunctionType)

zygoloid: Right, of course. We create the `FunctionType` as a side effect of importing the `FunctionDecl`, so we need to import the `FunctionDecl` here. Makes sense.

zygoloid: The `GetAs` seems redundant here, given we don't need any of the `StructValue`-specific fields:
```suggestion
    auto fn_val = context_.insts().Get(fn_const_id.inst_id());
    CARBON_CHECK(context_.types().Is<SemIR::FunctionType>(fn_val.type_id()));
    return {context_.types().GetConstantId(fn_val.type_id())};
```

toolchain/check/handle_function.cpp
zygoloid: Yeah, thanks. So the issue is that name lookup will find an `ImportRef`, not a `FunctionDecl`, and we don't have a good way to determine if the `ImportRef` actually refers to a `FunctionDecl` because all we have is the constant value. That makes sense, and while it's not going to be correct in the long term, that's an issue for another day.

PR #3932, opened 2024-05-02
Update LLVM

### LINKED ISSUE #3136: Remove clangd BUILD patch
#3112 patched the LLVM build to expose files from clangd, [here](https://github.com/carbon-language/carbon-lang/blob/trunk/bazel/patches/llvm/0004_Add_library_for_clangd.patch). If possible, this should be upstreamed: we should be trying to stay as close to LLVM upstream as we can, avoiding patches which we would need to maintain.

If it can't be upstreamed because upstream doesn't want these files used as a library, the dependency should be re-evaluated. In general we don't want to add dependencies on internal libraries, because of a lack of long-term stability guarantees.

FYI @josh11b @Maan2003

jonmeow:
Fixes #3136, wherein the clangd patch is upstreamed so we don't need to apply it anymore.

Updates goldens relative to a yaml output change in LLVM.


zygoloid (approved):

chandlerc (approved):
LG

PR #3935, opened 2024-05-04
Add a GenericClassType as the type of the name of a generic class.

zygoloid:
This is modeled analogously to FunctionType. So far, GenericClassType has no operations, but eventually values of that type will be callable like values of FunctionType.


toolchain/sem_ir/typed_insts.h
```diff
@@ -492,6 +492,17 @@ struct FunctionType {
   FunctionId function_id;
 };
 
+// The type of the name of a generic class. The corresponding value is an empty
+// `StructValue`.
+struct GenericClassType {
+  static constexpr auto Kind =
+      InstKind::GenericClassType.Define<Parse::AnyFunctionDeclId>(
```
jonmeow: I understand modeling after functions, but this seems like it should be using a ClassDecl node, not FunctionDecl. Maybe this works right now because GetGenericClassType doesn't validate?

toolchain/check/context.cpp
```diff
@@ -1066,6 +1067,15 @@ auto Context::GetFunctionType(SemIR::FunctionId fn_id) -> SemIR::TypeId {
   return type_id;
 }
 
+auto Context::GetGenericClassType(SemIR::ClassId class_id) -> SemIR::TypeId {
+  auto type_id = GetTypeImpl<SemIR::GenericClassType>(*this, class_id);
+  // To keep client code simpler, complete generic class types before returning
+  // them.
+  bool complete = TryToCompleteType(type_id);
+  CARBON_CHECK(complete) << "Failed to complete generic class type";
```
jonmeow: At this point we're at 3 of these that always complete, perhaps GetTypeImpl should have a GetCompleteTypeImpl variant which handles it?

toolchain/check/import_ref.cpp
```diff
@@ -694,24 +697,30 @@ class ImportRefResolver {
...
         .enclosing_scope_id = SemIR::NameScopeId::Invalid,
-        .implicit_param_refs_id = SemIR::InstBlockId::Invalid,
-        .param_refs_id = SemIR::InstBlockId::Invalid,
-        // `.self_type_id` depends on the ClassType, so is set below.
+        .implicit_param_refs_id = import_class.implicit_param_refs_id.is_valid()
+                                      ? SemIR::InstBlockId::Empty
+                                      : SemIR::InstBlockId::Invalid,
+        .param_refs_id = import_class.param_refs_id.is_valid()
+                             ? SemIR::InstBlockId::Empty
+                             : SemIR::InstBlockId::Invalid,
```
jonmeow, line 711: Setting Empty as incorrect-but-fully-functional leaves me concerned that it exposes things to incorrect reads. Had you considered a different special value, such as a new Uninitialized (incorrect-and-broken), or perhaps expanding ClassDecl to have a GenericClassDecl (assuming the below ReplaceInstBeforeConstantUse is at the root of this [not necessarily in this PR])?

toolchain/check/import_ref.cpp
```diff
@@ -755,10 +764,23 @@ class ImportRefResolver {
...
+      // declaration.
+      if (auto class_type = context_.insts().TryGetAs<SemIR::ClassType>(
+              class_const_id.inst_id())) {
+        class_id = class_type->class_id;
+      } else {
+        auto generic_class_type =
+            context_.types().GetAs<SemIR::GenericClassType>(
+                context_.insts().Get(class_const_id.inst_id()).type_id());
+        class_id = generic_class_type.class_id;
+      }
```
jonmeow: It's probably better to only read the instruction once.

```suggestion
      auto class_const_inst = context_.insts().Get(class_const_id.inst_id());
      if (auto class_type = class_const_inst.TryAs<SemIR::ClassType>()) {
        class_id = class_type->class_id;
      } else {
        auto generic_class_type =
            context_.types().GetAs<SemIR::GenericClassType>(class_const_inst.type_id());
        class_id = generic_class_type.class_id;
      }
```

jonmeow (approved):

toolchain/check/context.cpp
zygoloid: Done, but I could only make it cover two of these functions -- `GetBuiltinType` uses a predefined `InstId` directly instead of looking up the `ConstantId` in the constants table.

toolchain/sem_ir/typed_insts.h
zygoloid: Oops, copy-paste error. We actually never assign a location to this inst at all, because it's only ever created as a constant. Changed to `InvalidNodeId`.

toolchain/check/import_ref.cpp
zygoloid: Yeah, this is subtly depending on `TryEvalInst` / `Class::is_generic` caring about whether there is a parameter list, not about the *contents* of the parameter list. I've added a TODO; will consider this further.

toolchain/check/import_ref.cpp
zygoloid: I think / hope that it shouldn't matter -- the compiler should be smart enough to see the array can't have changed between the two accesses -- but it definitely can't hurt to help it out :) Done.

PR #3936, opened 2024-05-06
Replace ResolvePrevInstForMerge with decl logic.

jonmeow:
This addresses the const reliance of ResolvePrevInstForMerge by checking the imported instruction, testing with `alias` name conflicts. Given the need to examine structure, the helper function feels like it now gets in the way, so I'm just removing it.

### DIFF
--- a/toolchain/check/testdata/class/no_prelude/implicit_import.carbon
+++ b/toolchain/check/testdata/class/no_prelude/implicit_import.carbon
@@ -71,10 +71,10 @@ alias B = C;
 
 library "def_alias" impl;
 
-// CHECK:STDERR: def_alias.impl.carbon:[[@LINE+9]]:1: ERROR: Duplicate name being declared in the same scope.
+// CHECK:STDERR: fail_def_alias.impl.carbon:[[@LINE+9]]:1: ERROR: Duplicate name being declared in the same scope.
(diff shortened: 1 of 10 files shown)


zygoloid (approved):
Thanks!

toolchain/check/testdata/class/no_prelude/implicit_import.carbon
```diff
@@ -0,0 +1,242 @@
...
+class C {}
+
+// --- def_alias.carbon
+
+library "def_alias" api;
+
+class C;
+alias B = C;
+
+// --- def_alias.impl.carbon
```
zygoloid:
```suggestion
// --- fail_def_alias.impl.carbon
```

toolchain/check/testdata/function/definition/no_prelude/implicit_import.carbon
```diff
@@ -95,8 +95,31 @@ library "redef_after_def" impl;
...
 fn A() {}
 
+// --- def_alias.carbon
+
+library "def_alias" api;
+
+fn A();
+alias B = A;
+
+// --- def_alias.impl.carbon
```
zygoloid:
```suggestion
// --- fail_def_alias.impl.carbon
```

PR #3937, opened 2024-05-06
Move decl-specific merge logic back to respective handlers.

jonmeow:
I'd split out the decl-specific handlers so that import_ref.cpp could also call them. However, with the current model for imports and constants, we shouldn't use them there. As a consequence, merge them back into individual files for greater consistency with other helper functions (and better visibility when making related code changes).


jonmeow:
Depends on #3936

zygoloid (approved):

PR #3938, opened 2024-05-06
Exporting imported names

jonmeow:
In order to support exporting imported names, add `export import library <library>` and `export <name reference>` syntax.

### DIFF
--- a/proposals/p3938.md
+++ b/proposals/p3938.md
@@ -24,6 +24,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 -   [Rationale](#rationale)
 -   [Alternatives considered](#alternatives-considered)
     -   [Other `export` syntax structures](#other-export-syntax-structures)
+    -   [Other `export name` placements](#other-export-name-placements)
     -   [Re-exporting cross-package](#re-exporting-cross-package)
 
 <!-- tocstop -->
@@ -107,7 +108,8 @@ would be exported, or it should just be moved.
 ## Proposal
 
 Support the `export` keyword as a modifier to `import library <library>`
-(excluding cross-package imports). For example:
+(excluding cross-package imports). This is `export import` for short. For
+example:
 
 ```carbon
 export import library "lib";
@@ -115,7 +117,7 @@ export import library "lib";
 
 Additionally, support the `export` keyword on individual, file-scoped or
 namespace-scoped entities (excluding entities in other packages, and namespaces
-themselves). For example:
+themselves). This is `export name` for short. For example:
 
 ```carbon
 // Export an entity:
@@ -150,29 +152,34 @@ invalid in files which cannot be imported: implementation files and
 
 In the
 [source file introduction](/docs/design/code_and_name_organization/README.md#source-file-introduction),
-the `export` directive is intermingled with `import` directives. It is invalid
-later in the file.
+`export import` directives are intermingled with other `import` directives.
+`export name` directives are intermingled with the the `export` directive is
+intermingled with `import` directives.
 
 This allows:
 
 ```
 import library "foo";
-export FooType;
-
-import library "bar";
-export BarType;
-
 export import library "wiz";
+import library "bar";
+
+export FooType;
+export BarType;
 ```
 
 It disallows `export` directives following other code:
 
 ```
 import library "foo";
+// Invalid: comes before an import.
+export FooType;
+
+import library "bar";
 
 class C { ... };
 
-export FooType;
+// Invalid: comes after a declaration.
+export import library "wiz";
 ```
 
 ## Future work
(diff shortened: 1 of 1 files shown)


jonmeow:
Note, not sure what the best way is to see if people agree on the syntax choice -- I was thinking if there's a preference for an alternative, I can adjust the proposal. But `export import` at least sounded like people though it'd be okay, in the discussion.

zygoloid (commented):
LGTM

jonmeow:
@zygoloid FYI I want to be sure you notice the namespace commit. I don't think it really changes the proposal, but it I was trying to close some design gaps I realized existed (naively, `export NS;` would only export the namespace itself, not contents). [also, `export OtherPackage.Name;` wasn't explicitly disallowed, and is now]

zygoloid:
Still LG with namespace changes. I agree that disallowing namespace exports for now is the right choice, consistent with the existing rule that namespaces are exported iff they contain anything exported.

proposals/p3938.md
```diff
@@ -144,6 +145,35 @@ alias Obj = Other.Obj;
...
+```
+
+It disallows `export` directives following other code:
+
+```
+import library "foo";
+
+class C { ... };
+
+export FooType;
```
zygoloid: I think we should explore the tradeoff here further. We require all `import`s (whether or not they are `export`ed) to be listed first, and I think that's reasonable to preserve, but then it seems we have three choices:

1) As proposed: `export name` can (only) appear in the preamble, with the imports, and cannot appear with the other declarations in the library. This seems potentially a bit unintuitive to me, because other names exported by the library (those the library owns or declares `extern`, or aliases to imported names) can appear later on -- and indeed must.
2) Exact opposite: `export name` can (only) appear after the preamble, so we'd have (a) package declaration, then (b) imports, then (c) declarations that can add new names but not new imports. That seems a bit unsatisfying if you want to group the name exports with the libraries they're exported from.
3) Allow either: `export name` can appear in the preamble, or afterwards.

I *think* the option that seems best to me is (2): treat `export name` like any other non-import declaration. The consideration I find most persuasive here is how we'll handle name lookup within the `name` in `export name` declarations. Either we'll need to carefully process import declarations in the order they appear, intermingled with processing of `export name` declarations, so that the name lookup behavior for `export name` only sees the names that were already imported, *or* name lookup in `export name` syntax will find things that haven't been introduced yet, which seems surprising and inconsistent with our normal behavior for name lookup.

jonmeow: I can add this alternative, switching to (2) if that's your preference. To make sure I'm getting this right, trying to note which advantages/disadvantages are particular to each option:

1. `export name` with `import`s

Semantically, names not introduced by the current package are addressed first.

Advantages:

- No need to teach developers they cannot `export` newly introduced names.

Disadvantages:

(mostly the lack of advantages from others)

2. `export name` with other declarations

Semantically, names introduced by by other packages come first, then operations on those names and/or introducing new names.

Advantages:

- `import` remains very special.
- Avoids name lookup questions regarding exporting names added by later `import` directives.

Disadvantages:

- Prevents placing `export name` next to the import that is expected to add the name.
- Means `export import` and `export name` will be in different sections: no single place to look for re-exports.

3. No ordering for `export name`

Semantically, let developers choose what the prefer.

Advantages:

- Maximum flexibility, HOA rule.

Disadvantages:

- Most inconsistent with the desire to treat `import` as special.

zygoloid: That sounds good to me. I've asked for feedback on discord, but if there are no opposing opinions by Monday I think we should go ahead with option 2. We can always revisit if we get new information.

jonmeow: FYI merged this into the proposal, along with a couple extra thoughts (not sure how you feel about adding a dedicated `export name` section, but I'm expecting a negative response)

proposals/p3938.md
```diff
@@ -0,0 +1,333 @@
...
+invalid in files which cannot be imported: implementation files and
+`Main//default`.
+
+### Source file introduction
+
+In the
+[source file introduction](/docs/design/code_and_name_organization/README.md#source-file-introduction),
+`export import` directives are intermingled with other `import` directives.
+`export name` directives are intermingled with the the `export` directive is
+intermingled with `import` directives.
```
chandlerc: This sentence doesn't parse for me...

Maybe you're trying to say something like:

```suggestion
`export name` directives are normal code and cannot be intermingled with any `import` directives, including `export import` directives.
```

proposals/p3938.md
```diff
@@ -0,0 +1,333 @@
...
+```
+import library "foo";
+export import library "wiz";
+import library "bar";
+
+export FooType;
+export BarType;
+```
+
+It disallows `export` directives following other code:
```
chandlerc: Maybe this is easier to phrase as -- "these are always `import` directives, whether modified by `export` or not, and those can never follow other code" or something similar?

chandlerc (commented):
LGTM, minor wording suggestions below, nothing substantive.

I agree with the option selected for where you can write `export` on a name. Checking with other leads to see if I can approve...

chandlerc (approved):
And approving with wording fixes.

chandlerc (approved):
And approving with wording fixes.

proposals/p3938.md
jonmeow: Shortened to just "This disallows" since the comments already mark. (making the comments a little more explicit too)

chandlerc:
And merging (confirmed with the author).

PR #3942, opened 2024-05-07
Unify more invalid declaration recovery.

jonmeow:
When we expect a semicolon, we should be able to provide more standard recovery.

Note this affects a test of `base`, but it looks like a partial improvement (the prior line was moving too far). Still, it looks like recovery isn't handling `{}` quite right. I'm not trying to address that here though -- leaving a TODO.

### DIFF
--- a/toolchain/parse/context.h
+++ b/toolchain/parse/context.h
@@ -281,6 +281,13 @@ class Context {
   // Propagates an error up the state stack, to the parent state.
   auto ReturnErrorOnState() -> void { state_stack_.back().has_error = true; }
 
+  // Adds a node expecting a decl semi. Includes error recovery when the token
+  // is not a semi. `is_def_allowed` is only selecting the diagnostic if not a
+  // semi; a definition must be handled separately.
+  auto AddNodeExpectingDeclSemi(StateStackEntry state, Lex::TokenKind decl_kind,
+                                NodeKind node_kind, bool is_def_allowed)
+      -> void;
+
   // Emits a diagnostic for a declaration missing a semi.
   auto DiagnoseExpectedDeclSemi(Lex::TokenKind expected_kind) -> void;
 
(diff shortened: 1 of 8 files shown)


jonmeow:
Depends on #3941

toolchain/parse/context.h
```diff
@@ -281,6 +281,13 @@ class Context {
   // Propagates an error up the state stack, to the parent state.
   auto ReturnErrorOnState() -> void { state_stack_.back().has_error = true; }
 
+  // Adds a node expecting a decl semi. Includes error recovery when the token
+  // is not a semi. `is_def_allowed` is only selecting the diagnostic if not a
+  // semi; a definition must be handled separately.
```
josh11b: Could you explain the arguments more explicitly?
- `decl_kind` I think is only used in error message
- `node_kind` is the kind of the node being added

jonmeow: > Could you explain the arguments more explicitly?
> 
> * `decl_kind` I think is only used in error message

Done.

> * `node_kind` is the kind of the node being added

Can you please elaborate what you're wanting here? There's only one mention of a node in the comment, so I'm missing possible other meanings: "`node_kind` is the kind of the node" sounds redundant to me, but lacking an understanding of the ambiguity I'm not sure what else to write. Note this is also similar to other `AddNode` and `AddLeafNode` functions.

toolchain/parse/context.h
```diff
@@ -281,6 +281,13 @@ class Context {
   // Propagates an error up the state stack, to the parent state.
   auto ReturnErrorOnState() -> void { state_stack_.back().has_error = true; }
 
+  // Adds a node for a declaration's semicolon. Includes error recovery when the
```
josh11b, line 284: Maybe:
```suggestion
  // Adds a node with kind `node_kind` for a declaration's semicolon. Includes error recovery when the
```
It also would have been a bit less surprising to me if `node_kind` came earlier instead of after something only used as text when there is an error. You mention `AddNode` and `AddLeafNode` don't specifically document the `node_kind` parameter, but in those methods the `node_kind` parameter is first, right after the `Add...Node`.

jonmeow: I've reordered parameters so that the diagnostic-specific things are last, does that work for you?

If not, I'll use your suggestion, but I'd like to be clear that saying "Adds a node with kind `node_kind`" seems no less redundant.

josh11b (approved):

PR #3944, opened 2024-05-08
Add export keyword handling.

jonmeow:
This provides `export import` logic in lex, parse, and check; `export name` logic is only in lex and parse, not check.

I think with `export name` I'm going to need to modify import_ref and some consolidation logic, whereas `export import` seems feasible to keep as primarily import logic. Given the implementations were looking like they'd diverge more substantially, I thought it'd be helpful to cut the PR here.


toolchain/check/check.cpp
```diff
@@ -200,6 +200,81 @@ struct UnitInfo {
...
+                                 SemIR::TypeId namespace_type_id) -> void {
+  // Add imports from the current package.
+  auto self_import = unit_info.package_imports_map.find(IdentifierId::Invalid);
+  if (self_import == unit_info.package_imports_map.end()) {
+    // Push the scope; there are no names to add.
+    context.scope_stack().Push(package_inst_id, SemIR::NameScopeId::Package);
+    return;
+  }
+
+  auto imported_irs = llvm::SmallVector<bool>(total_ir_count, false);
```
```suggestion
  llvm::SmallVector<bool> imported_irs(total_ir_count, false);
```
Also, would it make sense to use `llvm::SmallBitVector`?

... but see comment below; I think we can avoid this vector entirely.

toolchain/check/check.cpp
```diff
@@ -200,6 +200,81 @@ struct UnitInfo {
...
+  bool error_in_import = self_import->second.has_load_error;
```
zygoloid: Incrementally computing this flag is requiring a little plumbing here. I think it'd be simpler overall to make `ImportLibraryFromCurrentPackage` set the package name scope to invalid itself if the imported package name scope is invalid. That function is otherwise responsible for importing names from the other file's name scope into our name scope, so also importing the `has_error` flag there seems logical to me.

(We'd still need to update it here in the `self_import->second.has_load_error` case, so maybe this doesn't buy us anything in terms of code size, but I still think it'd be an improvement.)

toolchain/check/check.cpp
```diff
@@ -200,6 +200,81 @@ struct UnitInfo {
...
+      for (const auto& indirect_ir : import_sem_ir.import_irs().array_ref()) {
+        if (indirect_ir.is_export) {
+          context.import_irs()
+              .Get(context
+                       .check_ir_map()[indirect_ir.sem_ir->check_ir_id().index])
+              .is_export = true;
+        }
+      }
+    }
+  }
```
zygoloid: There's some duplication here handling the `is_export` flag in two different ways (first import / later import). I think we should be able to unify this by calling `ImportLibraryFromCurrentPackage` unconditionally, and having it bail out after calling `AddImportIR` if the library is already imported.

I think you could also remove the bit vector that way too -- we can determine whether the IR has already been imported by looking at whether a value for `InstId::PackageNamespace` has already been set in the IR's `import_ir_constant_values`, or by making `AddImportIR` also return a flag to say whether the IR is new.

toolchain/check/testdata/packages/no_prelude/export_import.carbon
```diff
@@ -0,0 +1,487 @@
...
+
+library "fail_use_non_export_import" api;
+
+import library "non_export_import";
+
+// CHECK:STDERR: fail_use_non_export_import.carbon:[[@LINE+3]]:15: ERROR: Name `C` not found.
+// CHECK:STDERR: alias Local = C;
+// CHECK:STDERR:               ^
+alias Local = C;
+
```
zygoloid, line 147: I think it'd be useful to add tests for the cases where an imported library transitions from non-exported to exported after being first imported:

```
import library "foo";
export import library "bar";
```
... where either:
- "bar" exports "foo" (and so this library should also export "foo" even though we don't say so), or
- "foo" contains an exported import of "bar" (and so the first time we import "bar" it's non-exported), or
- both "foo" and "bar" contain an exported import of "baz"

toolchain/parse/handle_import_and_package.cpp
```diff
@@ -157,49 +166,106 @@ static auto HandleDirectiveContent(Context& context,
...
+auto HandleExport(Context& context) -> void {
+  auto state = context.PopState();
+
+  context.ConsumeChecked(Lex::TokenKind::Export);
+
+  // Error for both Main//default and every implementation file.
+  auto packaging = context.tree().packaging_directive();
+  if (!packaging || packaging->api_or_impl == Tree::ApiOrImpl::Impl) {
+    CARBON_DIAGNOSTIC(ExportFromImpl, Error,
+                      "`export` is only allowed in API files.");
```
zygoloid, line 237: Given that we decided to consider the default file to be an API file rather than an implementation file, this error seems to be technically incorrect in that case. Is it worth adding a custom diagnostic for that?

toolchain/parse/typed_nodes.h
```diff
@@ -143,7 +143,7 @@ struct QualifiedName {
   IdentifierNameId rhs;
 };
 
-// Library, package, import
+// Library, package, import, export
 // ------------------------
```
zygoloid:
```suggestion
// --------------------------------
```

zygoloid (approved):
Some suggestions but nothing blocking.

toolchain/check/check.cpp
jonmeow: Done (FWIW, I was writing this more Carbon-ish, as `var imported_irs: auto = llvm.SmallVector(bool).Make(total_ir_count, false);` seems most likely)

I'm not sure SmallBitVector is needed here. One trade-off is that I'm keeping the reference to avoid the re-indexing, whereas SmallBitVector would require separate get/set calls. Since these are ephemeral, not sure the memory footprint matters as much.

toolchain/check/check.cpp
jonmeow: Done.

toolchain/check/check.cpp
jonmeow: Added a comment to explain what's going on:

```
  // Track whether an IR was imported in full, including `export import`. This
  // distinguishes from IRs that are indirectly added without all names being
  // exported to this IR.
```

If we were to use `AddImportIR` we'd need to track equivalent information, and I think this functions-scoped tracking is more efficient.

toolchain/parse/handle_import_and_package.cpp
jonmeow: We could, but I've been wondering: is this technicality helpful, or should we change the meaning of nothing to be just "the main file", not an API file? Semantically, things like `extern` should also only be allowed in API files except for the default file. At the same time, it can't have an impl.

Maybe I'll take this to #naming.

toolchain/check/testdata/packages/no_prelude/export_import.carbon
jonmeow: Done, also tried improving test names given the additions.

jonmeow:
To just note, planning to adjust `export` parse order separately, based on #3938 comments.

toolchain/check/check.cpp
zygoloid: Yeah, a single allocation once per invocation seems kinda irrelevant. I think that keeping the reference / re-indexing is also irrelevant (and I would expect there to actually be zero cost to re-indexing -- the optimizer should see right through that) for the same reason. So maybe `SmallVector` is better just to avoid using a less-well-known type.

toolchain/check/check.cpp
zygoloid: While I think it should still be possible to simplify this a bit, it's not worth blocking on.

toolchain/parse/handle_import_and_package.cpp
zygoloid: Let's address any decision from #naming in a separate PR.

PR #3947, opened 2024-05-09
Don't attempt to merge multiple class definitions.

### LINKED ISSUE #3945: Infinite loop in `LookupQualifiedName`
### Description of the bug:

Infinite loop found by fuzzer.

`bazel run toolchain/driver:carbon compile test.carbon`, where the contents of `test.carbon` are:

```
base class B {
}

base class R {
  extend base: B;
}

base class B {
  extend base: B;
  var c: c32;
}
```

The compiler terminates in well under a second, probably with some sort of error.

### What actually happened?

The compiler runs indefinitely (at least a minute).

Initial debugging indicates that this is caused by an infinite loop in `LookupQualifiedName`, where `scope.extended_scopes()` consists of the ID of `scope` itself.

geoffromer:
Mutating classes after they're complete leads to inconsistencies such as complete class types that have incomplete class definitions.

Closes #3945

### DIFF
--- a/toolchain/check/handle_class.cpp
+++ b/toolchain/check/handle_class.cpp
@@ -68,6 +68,11 @@ static auto MergeClassRedecl(Context& context, SemIRLoc new_loc,
                         .is_extern = prev_is_extern},
                        prev_import_ir_id);
 
+  if (new_is_definition && prev_class.is_defined()) {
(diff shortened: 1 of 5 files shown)


josh11b (approved):

jonmeow (approved):
Looks good, thanks!

PR #3948, opened 2024-05-15
On NameContext, rename enclosing_scope and target_scope_id.

jonmeow:
I think the new name is more consistent for how `enclosing_scope_id` is used relative to `name_id` (even removing the clarifying note on `enclosing_scope_id_for_new_inst`). Suggesting `initial_scope_index` as a replacing for the old `enclosing_scope`, hoping it's a little clearer.

I'm replacing `target_scope_id` uses in modifier logic because they seemed to be based on the NameContext use.

### DIFF
--- a/toolchain/check/handle_interface.cpp
+++ b/toolchain/check/handle_interface.cpp
@@ -38,7 +38,7 @@ static auto BuildInterfaceDecl(Context& context,
 
   // Process modifiers.
   CheckAccessModifiersOnDecl(context, Lex::TokenKind::Interface,
-                             name_context.target_scope_id);
+                             name_context.enclosing_scope_id);
   LimitModifiersOnDecl(context, KeywordModifierSet::Access,
                        Lex::TokenKind::Interface);
 
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -488,7 +488,7 @@ class DeferredDefinitionWorklist {
   auto PushEnterDeferredDefinitionScope(Context& context) -> void {
     bool nested = !enclosing_scopes_.empty() &&
                   enclosing_scopes_.back().scope_index ==
-                      context.decl_name_stack().PeekEnclosingScope();
+                      context.decl_name_stack().PeekInitialScopeIndex();
     enclosing_scopes_.push_back(
         {.worklist_start_index = worklist_.size(),
          .scope_index = context.scope_stack().PeekIndex()});
--- a/toolchain/check/decl_name_stack.cpp
+++ b/toolchain/check/decl_name_stack.cpp
@@ -35,8 +35,8 @@ auto DeclNameStack::NameContext::prev_inst_id() -> SemIR::InstId {
 
 auto DeclNameStack::MakeEmptyNameContext() -> NameContext {
   return NameContext{
-      .enclosing_scope = context_->scope_stack().PeekIndex(),
-      .target_scope_id = context_->scope_stack().PeekNameScopeId()};
+      .initial_scope_index = context_->scope_stack().PeekIndex(),
+      .enclosing_scope_id = context_->scope_stack().PeekNameScopeId()};
 }
 
 auto DeclNameStack::MakeUnqualifiedName(SemIR::LocId loc_id,
@@ -83,7 +83,7 @@ auto DeclNameStack::FinishImplName() -> NameContext {
 auto DeclNameStack::PopScope() -> void {
   CARBON_CHECK(decl_name_stack_.back().state == NameContext::State::Finished)
       << "Missing call to FinishName before PopScope";
-  context_->scope_stack().PopTo(decl_name_stack_.back().enclosing_scope);
+  context_->scope_stack().PopTo(decl_name_stack_.back().initial_scope_index);
   decl_name_stack_.pop_back();
 }
 
@@ -91,7 +91,7 @@ auto DeclNameStack::Suspend() -> SuspendedName {
   CARBON_CHECK(decl_name_stack_.back().state == NameContext::State::Finished)
       << "Missing call to FinishName before Suspend";
   SuspendedName result = {decl_name_stack_.pop_back_val(), {}};
-  auto enclosing_index = result.name_context.enclosing_scope;
+  auto enclosing_index = result.name_context.initial_scope_index;
   auto& scope_stack = context_->scope_stack();
   while (scope_stack.PeekIndex() > enclosing_index) {
     result.scopes.push_back(scope_stack.Suspend());
@@ -105,7 +105,7 @@ auto DeclNameStack::Suspend() -> SuspendedName {
 auto DeclNameStack::Restore(SuspendedName sus) -> void {
   // The enclosing state must be the same when a name is restored.
   CARBON_CHECK(context_->scope_stack().PeekIndex() ==
-               sus.name_context.enclosing_scope)
+               sus.name_context.initial_scope_index)
       << "Name restored at the wrong position in the name stack.";
 
   // clang-tidy warns that the `std::move` below has no effect. While that's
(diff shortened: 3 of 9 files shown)


toolchain/check/handle_interface.cpp
```diff
@@ -38,7 +38,7 @@ static auto BuildInterfaceDecl(Context& context,
 
   // Process modifiers.
   CheckAccessModifiersOnDecl(context, Lex::TokenKind::Interface,
-                             name_context.target_scope_id);
+                             name_context.enclosing_scope_id);
```
zygoloid, line 41: I think this is wrong (pre-existing bug), because `enclosing_scope_id` refers to the resolved prefix of the name, which isn't necessarily the enclosing scope. For example, for 
```carbon
class C { 
  protected interface I {}    
  protected interface I;  
}
```
... we currently reject the `protected` on the second declaration of `I`:
```console
/proc/self/fd/11:4:3: ERROR: `protected` not allowed on `interface` declaration, `protected` is only allowed on class members.
  protected interface I;
  ^~~~~~~~~
```
... because `name_context.enclosing_scope_id` is resolved to the interface declaration here, and so we're incorrectly checking this interface declaration as if it appeared within itself.

(The above example would be invalid for other reasons -- it's redeclaring `I` after a definition -- but I think the point remains that the current name and the new name don't really convey what this field means.)

zygoloid (commented):
I think making a change here is a good idea. We're currently using "enclosing scope" to mean both "the scope that lexically encloses an entire declaration" and also "the scope that owns the declared entity". Eg, given
```carbon
class A {
  class B {
    fn F();
  }
  fn B.F() {}
}
```
the `NameContext::enclosing_scope` for the name `B.F` refers to the class scope of `A` (the scope that lexically encloses the name), but the `enclosing_scope_id` for the `Function` `B.F` is `B`.

That said: for me, the change from `NameContext::target_scope_id` to `NameContext::enclosing_scope_id` doesn't help -- I'd expect the latter to be the scope that encloses the entire name context, not the scope that the name will be redeclared within. But I think a bigger issue is that the ID we're storing in this field in `NameContext` isn't always the parent / enclosing scope for the name at all: if the name context so far was resolved to the name of an existing scope, then it's the ID of that scope itself, not its parent scope. So I wonder if it would make sense to use a name like `resolved_prefix_scope_id` in `NameContext`? That might also reduce the need for a lengthy comment in `PeekTargetScope` explaining that this sometimes represents the entire name so far, and sometimes represents the portion of the name before the final `.`, and would parallel the naming of `unresolved_name_id` and `resolved_inst_id`.

I also wonder if it'd be clearer to use `parent_scope_id` instead of `enclosing_scope_id` in the various entity classes and in modifiers, where we mean the owning scope rather than the one that lexically encloses the declaration. But I think there's less scope for confusion there, because the two are usually the same for the first declaration of an entity. (The exception would be names in namespaces, given that namespaces never lexically enclose anything.)

(FWIW, C++ [uses the term "target scope"](https://eel.is/c++draft/basic#scope.scope-2.4) to mean the scope that owns the entity (approximately, the scope where its name uniquely refers to it, but C++ has some extra complications here), uses "parent scope" to mean the scope that immediately contains another scope, and uses "enclosing scope" to mean "any scope that lexically encloses another scope".)

jonmeow:
> I think making a change here is a good idea. We're currently using "enclosing scope" to mean both "the scope that lexically encloses an entire declaration" and also "the scope that owns the declared entity". Eg, given
> 
> ```
> class A {
>   class B {
>     fn F();
>   }
>   fn B.F() {}
> }
> ```
> 
> the `NameContext::enclosing_scope` for the name `B.F` refers to the class scope of `A` (the scope that lexically encloses the name), but the `enclosing_scope_id` for the `Function` `B.F` is `B`.
> 
> That said: for me, the change from `NameContext::target_scope_id` to `NameContext::enclosing_scope_id` doesn't help -- I'd expect the latter to be the scope that encloses the entire name context, not the scope that the name will be redeclared within. But I think a bigger issue is that the ID we're storing in this field in `NameContext` isn't always the parent / enclosing scope for the name at all: if the name context so far was resolved to the name of an existing scope, then it's the ID of that scope itself, not its parent scope. So I wonder if it would make sense to use a name like `resolved_prefix_scope_id` in `NameContext`? That might also reduce the need for a lengthy comment in `PeekTargetScope` explaining that this sometimes represents the entire name so far, and sometimes represents the portion of the name before the final `.`, and would parallel the naming of `unresolved_name_id` and `resolved_inst_id`.

I'm confused here, rename which to `resolved_prefix_scope_id`? Neither bit of scope information is set only for resolved names [also, if you mean the ScopeIndex, I don't view that as an ID]. Note that resolved_inst_id is only valid for resolved names. I'm also not understanding what the "prefix" is.

> I also wonder if it'd be clearer to use `parent_scope_id` instead of `enclosing_scope_id` in the various entity classes and in modifiers, where we mean the owning scope rather than the one that lexically encloses the declaration. But I think there's less scope for confusion there, because the two are usually the same for the first declaration of an entity. (The exception would be names in namespaces, given that namespaces never lexically enclose anything.)
> 
> (FWIW, C++ [uses the term "target scope"](https://eel.is/c++draft/basic#scope.scope-2.4) to mean the scope that owns the entity (approximately, the scope where its name uniquely refers to it, but C++ has some extra complications here), uses "parent scope" to mean the scope that immediately contains another scope, and uses "enclosing scope" to mean "any scope that lexically encloses another scope".)

In theory I'm fine with this, but that seems like I pretty big change since probably every `enclosing_scope` should be renamed to `parent_scope` [including digging through comments]. I lack the energy for that right now so, if it's your preference, I'll close this PR and leave it to be addressed later.

toolchain/check/handle_interface.cpp
jonmeow: Note, while I agree it seems suspicious, this PR is not changing behavior.

zygoloid (approved):
> I'm confused here, rename which to `resolved_prefix_scope_id`?

`NameContext::target_scope_id` (or in this PR, `NameContext::enclosing_scope_id`).

> Neither bit of scope information is set only for resolved names [also, if you mean the ScopeIndex, I don't view that as an ID]. Note that resolved_inst_id is only valid for resolved names. I'm also not understanding what the "prefix" is.

The idea that I think it'd be good to communicate with the field name is that the field describes the prefix of the name that we could resolve as a scope. Eg, for `A.B.C`, it'll either be:

- `A.B`, if `C` didn't resolve to a scope, or
- `A.B.C` if `C` did resolve to a scope.

For example, the uses of this field to check modifiers in `handle_*` are subtly incorrect because they're misinterpreting this field as the parent scope, which is usually true, but only holds if the final name component doesn't resolve to an existing scope.

Instead of the renaming I'm suggesting, perhaps we could change `NameContext` to delay resolving the name to a scope until we try to add another name after that scope. So then the `NameScopeId` we're storing *would* always be the enclosing / parent scope for the name.

If you think that direction is reasonable, then I think this PR is good as-is, as a step towards that approach, and we can actually make that change as future work.

=> Approving on that basis.

jonmeow:
> > I'm confused here, rename which to `resolved_prefix_scope_id`?
> 
> `NameContext::target_scope_id` (or in this PR, `NameContext::enclosing_scope_id`).
> 
> > Neither bit of scope information is set only for resolved names [also, if you mean the ScopeIndex, I don't view that as an ID]. Note that resolved_inst_id is only valid for resolved names. I'm also not understanding what the "prefix" is.
> 
> The idea that I think it'd be good to communicate with the field name is that the field describes the prefix of the name that we could resolve as a scope. Eg, for `A.B.C`, it'll either be:
> 
> * `A.B`, if `C` didn't resolve to a scope, or
> * `A.B.C` if `C` did resolve to a scope.

I think the nuance that makes me view this differently is `A` [versus `A.B.C`]. There are actually three possibilities:

- If a scope was never resolved [as in an unresolved `A`], the scope corresponding to the overall declaration (which could be `package`, or a scoped entity).
- If a scope was resolved, the scope.
  - `A.B`, if `C` didn't resolve to a scope, or
  - `A.B.C` if `C` did resolve to a scope.

Given your definition of it as based on "the prefix of the name", I would say that `A` has no prefix. However, it does have a parent scope.

> For example, the uses of this field to check modifiers in `handle_*` are subtly incorrect because they're misinterpreting this field as the parent scope, which is usually true, but only holds if the final name component doesn't resolve to an existing scope.
> 
> Instead of the renaming I'm suggesting, perhaps we could change `NameContext` to delay resolving the name to a scope until we try to add another name after that scope. So then the `NameScopeId` we're storing _would_ always be the enclosing / parent scope for the name.
> 
> If you think that direction is reasonable, then I think this PR is good as-is, as a step towards that approach, and we can actually make that change as future work.

Possibly fine, I'm not sure whether there's a reason for the current resolution pattern beyond it just being a first stroke. What you're describing may have an incremental advantage of dropping the ResolvedNonScope state, and reducing redeclaration scope lookups by one.

PR #3949, opened 2024-05-15
Initial 'export name' handling.

jonmeow:
This adds a `BindExport` instruction in order to better track the location of the `export` itself, but a `bind_name_id` is also added to `ImportRef` so that we know quickly where to put it in name lookup.

Merging identical names is a TODO. I haven't quite decided how best to achieve that, because I do think the BindExport should be what's actually added to name lookup.

Also, I will probably add a mode to DeclNameStack that blocks non-namespace scopes. This seems to already be an error, but the wrong one (maybe due to lack of support for cross-file decl/def support).


chandlerc (approved):
Generally looks good.

A bit odd that we call this a 'bind', but I don't have an immediate suggestion for a better name. And even if we want to rename it, seems best to do that as a follow-up, so approving.

PR #3950, opened 2024-05-15
Move 'export name' out of the import section.

jonmeow:
This implements recent updates to #3938

PR #3954, opened 2024-05-16
Add support for combining conflicting exported names.

jonmeow:
I think this is a key remaining bit of polish for intra-package exports. Next I'm going to be dealing with cross-package exports, though I think the fundamentals here will remain intact.

Extern declarations could cause issues for the current approach, but I think that can be addressed separately. I don't think it's a fundamental problem, more just incremental.

### DIFF
--- a/toolchain/check/import.cpp
+++ b/toolchain/check/import.cpp
@@ -12,6 +12,7 @@
 #include "toolchain/parse/node_ids.h"
 #include "toolchain/sem_ir/file.h"
 #include "toolchain/sem_ir/ids.h"
+#include "toolchain/sem_ir/import_ir.h"
 #include "toolchain/sem_ir/inst.h"
 #include "toolchain/sem_ir/typed_insts.h"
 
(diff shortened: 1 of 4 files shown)


jonmeow:
Depends on #3950 (dependency merged)

toolchain/check/import.cpp
```diff
@@ -216,6 +217,77 @@ static auto CopyEnclosingNameScopesFromImportIR(
   return scope_cursor;
 }
 
+// Returns the canonical IR inst for an entity. Returns an invalid ir_id for the
+// current IR.
+static auto GetCanonicalImportIRInst(Context& context,
+                                     const SemIR::File* cursor_ir,
+                                     SemIR::InstId cursor_inst_id)
+    -> SemIR::ImportIRInst {
+  for (;;) {
```
```suggestion
  while (true) {
```
for consistency with how we write this elsewhere.

toolchain/sem_ir/import_ir.h
```diff
@@ -30,6 +30,10 @@ struct ImportIRInst : public Printable<ImportIRInst> {
     out << ir_id << ":" << inst_id;
   }
 
+  auto operator==(const ImportIRInst& rhs) const -> bool {
+    return ir_id == rhs.ir_id && inst_id == rhs.inst_id;
+  }
```
zygoloid, line 35:
```suggestion
  auto operator==(const ImportIRInst& rhs) const -> bool = default;
```

zygoloid (approved):

toolchain/sem_ir/import_ir.h
jonmeow:
```
In file included from toolchain/sem_ir/name.cpp:8:
In file included from ./toolchain/sem_ir/file.h:20:
./toolchain/sem_ir/import_ir.h:33:8: error: explicitly defaulted equality comparison operator is implicitly deleted [-Werror,-Wdefaulted-function-deleted]
  auto operator==(const ImportIRInst& rhs) const -> bool = default;
       ^
./toolchain/sem_ir/import_ir.h:28:30: note: defaulted 'operator==' is implicitly deleted because there is no viable three-way comparison function for base class 'Printable<Carbon::SemIR::ImportIRInst>'
struct ImportIRInst : public Printable<ImportIRInst> {
```

jonmeow: Adding an explicit operator== to Printable fixes _this_, but makes all other children comparable too. I'm not sure that's what we want. Will follow up separately.

PR #3955, opened 2024-05-17
Fix accidental name exposure on cross-package import use.

jonmeow:
Looking at `export` handling, I noticed and figured to fix this case first.

Also makes the test file prelude-agnostic.


jonmeow:
Depends on #3954

chandlerc (approved):
LG (when other commits are landed)

PR #3958, opened 2024-05-18
Disable `google-readability-casting` as it is broken.

chandlerc:
I had been bothered by the confusing suggestions to replace calls to a constructor with ... calls to a constructor, claiming the original code was a C-style cast. In some code, we've worked around this by using `static_cast` but that really doesn't seem to be a readability win.

I did some research and from what I can tell this tidy check is deeply broken, with both Chromium and internal Google users disabling it due to widespread issues:
https://chromium.googlesource.com/chromium/src/+/f828a2248f0da4e7bfadfa5db324230043aaea02

We should disable it as well for now. I've removed one `NOLINT` comment for it, but not tried to remove all the `static_cast`s that are now unnecessary as that seems hard to do in any automatic way.

adding, and why.**

Closes #ISSUE

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -21,6 +21,10 @@ WarningsAsErrors: '*'
 #       inst_id = name_ref->value_id;
 #                 ^ unchecked access to optional value
 #     }
+# - google-readability-casting is deeply broken, and fires regularly on calls
+#   to constructors. Worse, it suggests switching to a call to a constructor
+#   which doesn't fix the warning. This has been disabled by many major users
+#   of clang-tidy (Chromium, etc).
 # - google-readability-function-size overlaps with readability-function-size.
 # - modernize-use-designated-initializers is disabled because it fires on
 #   creation of SemIR typed insts, for which we do not currently want to use
@@ -33,7 +37,7 @@ Checks:
   -*, bugprone-*, -bugprone-branch-clone, -bugprone-easily-swappable-parameters,
   -bugprone-exception-escape, -bugprone-macro-parentheses,
   -bugprone-narrowing-conversions, -bugprone-switch-missing-default-case,
-  -bugprone-unchecked-optional-access, google-*,
+  -bugprone-unchecked-optional-access, google-*, -google-readability-casting,
   -google-readability-function-size, -google-readability-todo,
   misc-definitions-in-headers, misc-misplaced-const, misc-redundant-expression,
   misc-static-assert, misc-unconventional-assign-operator,
--- a/toolchain/lex/lex.cpp
+++ b/toolchain/lex/lex.cpp
@@ -195,7 +195,6 @@ static constexpr SIMDMaskArrayT PrefixMasks = []() constexpr {
   for (int i = 1; i < static_cast<int>(masks.size()); ++i) {
     masks[i] =
         // The SIMD types and constexpr require a C-style cast.
-        // NOLINTNEXTLINE(google-readability-casting)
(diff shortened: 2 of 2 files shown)


.clang-tidy
```diff
@@ -21,6 +21,10 @@ WarningsAsErrors: '*'
 #       inst_id = name_ref->value_id;
 #                 ^ unchecked access to optional value
 #     }
+# - google-readability-casting is deeply broken, and fires regularly on calls
```
jonmeow, line 24: If we're moving forward with this, is there an issue you can link to?

.clang-tidy
```diff
@@ -21,6 +21,10 @@ WarningsAsErrors: '*'
...
+#   to constructors. Worse, it suggests switching to a call to a constructor
+#   which doesn't fix the warning. This has been disabled by many major users
+#   of clang-tidy (Chromium, etc).
```
jonmeow, line 27: Note, the way you're talking about this issue is pretty strong. Would it be worth softening up the wording for other readers? e.g., maybe we could focus on the issue as it affects us:

```suggestion
# - google-readability-casting fires regularly on calls to constructors. It suggests switching to a call to a constructor, which doesn't fix the warning.
```

jonmeow:
> We should disable it as well for now. I've removed one `NOLINT` comment for it, but not tried to remove all the `static_cast`s that are now unnecessary as that seems hard to do in any automatic way.

Isn't this tidy check only warning about C-style casts? i.e., from a style perspective, doesn't [Google's casting style](https://google.github.io/styleguide/cppguide.html#Casting) still apply? If there are false positives for constructors, that would change. But we would still frequently be casting due to things like int vs size_t comparisons (which are probably the most frequent case). As a consequence, `static_cast` would remain. I see the one false positive here, but aren't the other uses of `static_cast` [ed: and `reinterpret_cast`, `const_cast`] correct?

That is to say, we have the occasional false positive, but also we use C++ casts a lot. If someone unfamiliar with Google C++ style comes along and starts adding C-style casts, is it going to be a net improvement if we're correcting the style in review? Should style change to instead encourage C-style casts?

jonmeow:
I should also ask here, is this also an issue with `{}` constructor syntax? What's the code that you were trying to write that you ran into this?

[ed: to be sure -- it's not clear to me from <https://crbug.com/1282228> whether they considered brace-based init as a possible solution]

.clang-tidy
jonmeow: Note too, you might also consider rephrasing the PR title and description.

chandlerc:
Well, after a huge debugging session that should have been much shorter (Jon told me exactly what command I should run to confirm my assumption, and I took way to long to run it), I understand the issue better now.

There *was* some bug at some point or in some patched versions of `clang-tidy` and `clangd` here. But the reason there are few references to it any more is that these bugs got fixed. And in fact modern clang-tidy and clangd are completely correct here.

At some point, I was sure there weren't other versions of `clangd` available on macOS, but if that was true once, it isn't true now -- there is a version shipped in Xcode that VSCode knew how to find and has the specific bug in this check I was seeing. Anyways, I'm using the right version and we can delete this PR. I'll see abotu adding an FAQ entry about accidentally getting the wrong `clangd` in your IDE.

zygoloid:
I'm still [seeing problems](https://github.com/carbon-language/carbon-lang/actions/runs/9308830717/job/25623009126?pr=4001) with this check in our CI. Should we disable the check?

PR #3959, opened 2024-05-19
Stop skipping JSON copyright & license headers by default.

chandlerc:
A lot of the JSON we use is parsed by VSCode or other systems that are happy to have comments in them, and we should retain a license header there. We can exclude specific files where this is a problem. I think the `package.json` and generated lock file from it are two such cases where we can't have the license header, but we can directly exclude more as needed.

This should help make sure that when we *do* have a license header in a JSON file it stays up to date and is formatted canonically.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -179,8 +179,12 @@ repos:
             Part of the Carbon Language project, under the Apache License v2.0 with LLVM
             Exceptions. See /LICENSE for license information.
             SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+          - --skip_pattern
+          # The default skip pattern skips JSON which we use with VSCode and
+          # it, like most file-based parsers, can support comments easily.
+          - '(^LICENSE|\.(ico))$'
           - --custom_format
-          - '\.(carbon|proto|ypp)(\.tmpl)?$'
+          - '\.(carbon|c|json|proto|ypp)(\.tmpl)?$'
           - ''
           - '// '
           - ''
@@ -209,11 +213,6 @@ repos:
           - ''
           - '-- '
           - ''
-          - --custom_format
-          - '\.c$'
-          - ''
-          - '// '
-          - ''
         exclude: |
           (?x)^(
               .bazelversion|
@@ -221,6 +220,7 @@ repos:
               compile_flags.txt|
               github_tools/requirements.txt|
               third_party/.*|
+              utils/vscode/package(-lock)?.json|
               .*\.def|
               .*\.svg|
               .*/fuzzer_corpus/.*|
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -217,6 +217,7 @@ repos:
           (?x)^(
               .bazelversion|
               .github/pull_request_template.md|
+              .vscode/extensions.json|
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -179,8 +179,12 @@ repos:
             Part of the Carbon Language project, under the Apache License v2.0 with LLVM
             Exceptions. See /LICENSE for license information.
             SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+          - --skip_pattern
+          # The default skip pattern skips JSON which we use with VSCode and
+          # it, like most file-based parsers, can support comments easily.
+          - '(^LICENSE|\.(ico))$'
           - --custom_format
-          - '\.(carbon|proto|ypp)(\.tmpl)?$'
+          - '\.(carbon|c|json|proto|ypp)(\.tmpl)?$'
```
jonmeow, line 187: If you want comments in json, let's make sure we're using a json-compatible comment format, e.g.:

```
  "_comment": "Put your JSON comment here"
```
https://reqbin.com/json/5wzepqmt/json-comment-example

chandlerc: FWIW, many consumers of JSON, at least the JSON in our project, specifically use C++ line style comments from what I can see. We already had those in one place. I tried to explain this above, but happy to improve that comment if it wasn't clear?

My intent here was for us to exclude any JSON files that need to follow strict JSON syntax and cannot support comments. I added two such below as exceptions.

jonmeow: In GitHub, in extensions.json the comments are highlighted in red. That's actually what got me to dig, why it was marked as an error. I'm not sure what else would break.

I think you could put this in json files with:

```
  "_copyright": "Part of the Carbon Language project, under the Apache License v2.0 with LLVM Exceptions. See /LICENSE for license information. SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception",
```

But if you strongly prefer to stick with `//` for familiarity reasons, I can accept and we can see if anything else breaks.

jonmeow (approved):

.pre-commit-config.yaml
chandlerc: Wild, I hadn't seen that GH highlighted these as incorrect.

Amazingly, GitHub doesn't seem to complain about the comments here:
https://github.com/carbon-language/carbon-lang/blob/trunk/utils/vscode/language-configuration.json

I have no idea why it complains in one place and not the other.

I'm worried about using the `"_copyright"` because that too is just a convention, and I'm worried that there will be schemas that don't support it. =/

I feel like we should use `//` comments where they work, and just not worry about the file header where they don't work. I guess the question becomes whether we:

- Manually exclude all the `.json` files where they don't work. I'll add `extensions.json` to the list, and we can keep adding more as discovered.
- Manually include the few `.json` files where they *do* work. I can try to figure out how to do this with the precommit stuff.
- Just never use a header in a `.json` file. I'll change this PR to remove the one case we have then.

What do you think? I'm happy with any of these, although I lean towards the first or third as those are the ones I know how to do.

jonmeow: We can try (1)

chandlerc: Sounds good, should be implemented now.

PR #3962, opened 2024-05-20
Fix clang-tidy issues in `//common`.

chandlerc:
These likely predate the CI integration for `clang-tidy` runs.

Most of these seem good generally, even though I disabled some with nolint comments. The multilevel pointer one seems almost like a bug in the check to detect the specific case of `memcpy`, but otherwise seems like a solid lint.

### DIFF
--- a/common/hashing_test.cpp
+++ b/common/hashing_test.cpp
@@ -74,14 +74,14 @@ TEST(HashingTest, Integers) {
         EXPECT_THAT(hash, Ne(hash_zero));
       }
     };
-    test_int_hash(i);
     test_int_hash(static_cast<int8_t>(i));
     test_int_hash(static_cast<uint8_t>(i));
     test_int_hash(static_cast<int16_t>(i));
     test_int_hash(static_cast<uint16_t>(i));
     test_int_hash(static_cast<int32_t>(i));
     test_int_hash(static_cast<uint32_t>(i));
-    test_int_hash(static_cast<int64_t>(i));
+    // `i` is already an int64_t variable.
+    test_int_hash(i);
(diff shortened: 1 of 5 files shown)


common/hashing_test.cpp
```diff
@@ -332,11 +332,15 @@ template <typename T>
 auto PrintFullWidthHex(llvm::raw_ostream& os, T value) {
   static_assert(sizeof(T) == 1 || sizeof(T) == 2 || sizeof(T) == 4 ||
                 sizeof(T) == 8);
+  // Given the nature of a format string and the good formatting, a nested
+  // conditional seems like the most readable structure.
+  // NOLINTBEGIN(readability-avoid-nested-conditional-operator)
```
jonmeow, line 337: Maybe not good for this change, but had you considered:

```
os << llvm::formatv(llvm::formatv("{{0:x{0}}", 2 * sizeof(T)), static_cast<uint64_t>(value);
```

jonmeow (approved):

jonmeow:
Thanks for the cleanup, I must've missed this before. I'm surprised it didn't cause problems earlier.

common/hashing_test.cpp
chandlerc: Mostly it seems really unfortunate to have a dynamically computed format string rather than a string literal. But it's hard to get string literals without something like the nested conditions here.

Maybe it's just me, but I find the way `clang-format` renders the conditions to result in something very readable.

jonmeow: FWIW, I was mostly thinking that since this is essentially debug code, the approach might just offer less to understand.

PR #3963, opened 2024-05-20
Implement proposal #3927.

zygoloid:
Change syntax for package declaration to put the `impl` keyword at the start and remove the `api` keyword.

To support this, rearrange processing of package, library, and import declarations to use the general modifier handling support in declaration parsing rather than special-case logic.

There is an ambiguity in `impl package.Foo as Bar`, which we resolve by treating `package` as an introducer after a modifier only if it's not followed by `.`.


toolchain/check/handle_import_and_package.cpp
```diff
@@ -3,44 +3,53 @@
...
-auto HandleImportExport(Context& /*context*/, Parse::ImportExportId /*node_id*/)
-    -> bool {
-  return true;
-}
-
-auto HandleImportDirective(Context& /*context*/,
+auto HandleImportDirective(Context& context,
                            Parse::ImportDirectiveId /*node_id*/) -> bool {
+  LimitModifiersOnDecl(context, KeywordModifierSet::Export,
+                       Lex::TokenKind::Import);
```
jonmeow, line 23: Is this (and equivalent on `library` and `package`) tested to ensure it's working as intended?

toolchain/parse/handle_base.cpp
```diff
@@ -6,6 +6,28 @@
 
 namespace Carbon::Parse {
 
+// Handles a `base` declaration after the introducer.
+auto HandleBaseAfterIntroducer(Context& context) -> void {
```
jonmeow, line 10: Thanks for refactoring these.

toolchain/parse/handle_decl_scope_loop.cpp
```diff
@@ -88,118 +54,112 @@ static auto ApplyIntroducer(Context& context, Context::StateStackEntry state,
...
-    }
-    case Lex::TokenKind::Namespace: {
-      ApplyIntroducer(context, state, NodeKind::NamespaceStart,
-                      State::Namespace);
-      return true;
-    }
-    case Lex::TokenKind::Let: {
-      ApplyIntroducer(context, state, NodeKind::LetIntroducer, State::Let);
-      return true;
+  const auto& info = DeclIntroducers[context.PositionKind().AsInt()];
```
jonmeow, line 114: Since you're heading in the direction of more generalized introducer logic, can you please make sure that invalid modifiers on `package` and `import` are tested to parse/error correctly? Previously those probably didn't need it, but that wouldn't be the case anymore.

toolchain/parse/handle_decl_scope_loop.cpp
```diff
@@ -8,51 +8,17 @@
 
 namespace Carbon::Parse {
 
-// Handles positions which are end of scope and packaging declarations. Returns
-// true when either applies. When the position is neither, returns false, and
-// may still update packaging state.
-static auto TryHandleEndOrPackagingDecl(Context& context) -> bool {
+// Returns true if at the end of the scope.
+static auto IsEndOfScope(Context& context) -> bool {
```
jonmeow: Could this even be `context.PositionIs(Lex::TokenKind::CloseCurlyBrace) || context.PositionIs(Lex::TokenKind::FileEnd)`?

toolchain/parse/handle_decl_scope_loop.cpp
```diff
@@ -88,118 +54,112 @@ static auto ApplyIntroducer(Context& context, Context::StateStackEntry state,
...
+};
+
+// Information about a keyword that might be an introducer keyword.
+struct DeclIntroducerInfo {
+  DeclIntroducerKind introducer_kind;
+  NodeKind node_kind;
+  State state;
+};
+
+static constexpr auto DeclIntroducers = [] {
```
jonmeow, line 60: Technically I think you could end the namespace before this instead of after.

toolchain/parse/handle_decl_scope_loop.cpp
```diff
@@ -88,118 +54,112 @@ static auto ApplyIntroducer(Context& context, Context::StateStackEntry state,
...
+      }
+      return false;
     }
-    case Lex::TokenKind::Var: {
-      ApplyIntroducer(context, state, NodeKind::VariableIntroducer,
-                      State::VarAsDecl);
-      return true;
+
+    case DeclIntroducerKind::PackagingDecl: {
+      break;
```
jonmeow, line 131:
```suggestion
      // Packaging-related keywords manage their packaging state.
      break;
```

Small comment from line 22 that is being removed, do you think i's still helpful?

toolchain/parse/handle_decl_scope_loop.cpp
```diff
@@ -88,118 +54,112 @@ static auto ApplyIntroducer(Context& context, Context::StateStackEntry state,
...
-      return true;
-    }
-    case Lex::TokenKind::Let: {
-      ApplyIntroducer(context, state, NodeKind::LetIntroducer, State::Let);
-      return true;
+  const auto& info = DeclIntroducers[context.PositionKind().AsInt()];
+
+  switch (info.introducer_kind) {
+    case DeclIntroducerKind::Unrecognized: {
+      // A `;` with no modifiers is an empty declaration.
```
jonmeow, line 118: Maybe a minor consequence of this: do we test / want to support `;` on its own at the top of a file? I think it would've become a packaging state error before. However, now you handle this _after_ packaging keywords, and so I think it's becoming valid...?

Note, this probably isn't tested, but maybe it should be. e.g.:

```
;

package Foo;
```

jonmeow (approved):

toolchain/check/handle_import_and_package.cpp
zygoloid: It is, but I think it wasn't when you started reviewing. So, done :)

toolchain/parse/handle_decl_scope_loop.cpp
zygoloid: Done.

toolchain/parse/handle_decl_scope_loop.cpp
zygoloid: Done and inlined.

toolchain/parse/handle_decl_scope_loop.cpp
zygoloid: Done.

toolchain/parse/handle_decl_scope_loop.cpp
zygoloid: Readded a comment.

toolchain/parse/handle_decl_scope_loop.cpp
zygoloid: Great point! Added tests.

We still reject a `;` before `package` and `library` declarations, but we now allow a `;` before or between `import` declarations. I hadn't thought about this before your comment, but now I have, I think that's the behavior that I would want here.

PR #3966, opened 2024-05-21
Update some additional package rules where structure changed in #3963.

jonmeow:
I think the template files are simply an oversight. The fuzz files should probably be updated because failing on the `package` line isn't an interesting test of logic. Plus one minor comment edit.


geoffromer (approved):
(Not merging in case @zygoloid has comments)

zygoloid (approved):
Thanks!

PR #3967, opened 2024-05-21
Handling for cross-package imports of "export name".

jonmeow:
Starts warning on duplicate "export name" because I'm updating name lookup to make it work.

Need to think harder about how to handle cross-package "export import". But adding some failing tests to sketch out what *should* work.

### DIFF
--- a/toolchain/check/testdata/packages/no_prelude/cross_package_export.carbon
+++ b/toolchain/check/testdata/packages/no_prelude/cross_package_export.carbon
@@ -1,0 +1,429 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+
+// ============================================================================
+// Setup files
+// ============================================================================
+
+// --- base.carbon
+
+package Other library "base";
+
+class C {
+  var x: ();
+};
+
+// --- export_import.carbon
+
+package Other library "export_import";
+
+export import library "base";
+
+// --- export_import_copy.carbon
+
+package Other library "export_import_copy";
+
+export import library "base";
+
+// --- export_name.carbon
+
+package Other library "export_name";
+
+import library "base";
+
+export C;
+
+// --- export_name_copy.carbon
+
+package Other library "export_name_copy";
+
+import library "base";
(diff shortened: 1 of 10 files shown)


toolchain/diagnostics/diagnostic_consumer.cpp
```diff
@@ -19,8 +19,15 @@ auto StreamDiagnosticConsumer::HandleDiagnostic(Diagnostic diagnostic) -> void {
...
-      *stream_ << "ERROR: ";
+    switch (message.level) {
+      case DiagnosticLevel::Error:
+        *stream_ << "ERROR: ";
+        break;
+      case DiagnosticLevel::Warning:
+        *stream_ << "WARNING: ";
+        break;
+      case DiagnosticLevel::Note:
+        break;
```
chandlerc: Could you split this out to its own PR? Mostly because it causess some churn in diagnostics and it seems a bit cleaner to have those separate from the import part.

jonmeow: Done, #3972

jonmeow:
Now depends on #3972 [ed: now merged]

toolchain/check/testdata/packages/no_prelude/cross_package_export.carbon
```diff
@@ -0,0 +1,429 @@
...
+// CHECK:STDERR: ^~~~~~
+// CHECK:STDERR: export_name_copy.carbon:4:1: In import.
+// CHECK:STDERR: import library "base";
+// CHECK:STDERR: ^~~~~~
+// CHECK:STDERR: base.carbon:4:1: ERROR: Duplicate name being declared in the same scope.
+// CHECK:STDERR: class C {
+// CHECK:STDERR: ^~~~~~~~~
+// CHECK:STDERR: fail_todo_use_export_name_both.carbon:[[@LINE-10]]:1: In import.
+// CHECK:STDERR: import Other library "export_name";
+// CHECK:STDERR: ^~~~~~
```
chandlerc, line 100: Not about this PR at all, but maybe for future stuff -- it's a bit awkward taht we end up doing the import walk twice here. Is it once for the error and once for the note? If so, I at least have an idea of how this might be addressed.

chandlerc (approved):
Nice, LG!

(The comment below is totally unrelated to this PR)

PR #3968, opened 2024-05-21
Refactor check_ir_map to encapsulate it.

jonmeow:
This is mostly to reduce code complexity at call sites. I was considering pulling out a wrapper type, but it seems a bit small for that right now.


jonmeow:
Depends on #3967

chandlerc (approved):
LG (once underlying PR lands)

PR #3971, opened 2024-05-21
Unify handling of transitive imports between current and other packages.

jonmeow:
This makes cross-package `export import` work. Note collisions still occur with cross-package "export name".

### DIFF
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -212,6 +212,56 @@ struct UnitInfo {
   UnitInfo* api_for_impl = nullptr;
 };
 
+// Collects transitive imports, handling deduplication.
+static auto CollectTransitiveImports(const UnitInfo::PackageImports& imports,
+                                     int total_ir_count)
+    -> llvm::SmallVector<SemIR::ImportIR> {
+  llvm::SmallVector<SemIR::ImportIR> results;
+
+  // Track whether an IR was imported in full, including `export import`. This
+  // distinguishes from IRs that are indirectly added without all names being
+  // exported to this IR.
+  llvm::SmallVector<int> imported_irs(total_ir_count, -1);
+
+  // First add direct imports. This means that if an entity is imported both
+  // directly and indirectly, the import path will reflect the direct import.
+  for (const auto& import : imports.imports) {
+    const auto& direct_ir = **import.unit_info->unit->sem_ir;
+    imported_irs[direct_ir.check_ir_id().index] = results.size();
+    results.push_back({.node_id = import.names.node_id,
+                       .sem_ir = &direct_ir,
+                       .is_export = import.names.is_export});
+  }
+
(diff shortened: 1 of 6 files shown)


jonmeow:
Depends on #3968 [ed: merged]

toolchain/check/check.cpp
```diff
@@ -212,6 +212,56 @@ struct UnitInfo {
...
+// Collects transitive imports, handling deduplication.
+static auto CollectTransitiveImports(const UnitInfo::PackageImports& imports,
+                                     int total_ir_count)
+    -> llvm::SmallVector<SemIR::ImportIR> {
+  llvm::SmallVector<SemIR::ImportIR> results;
+
+  // Track whether an IR was imported in full, including `export import`. This
+  // distinguishes from IRs that are indirectly added without all names being
+  // exported to this IR.
+  llvm::SmallVector<int> imported_irs(total_ir_count, -1);
```
zygoloid: Perhaps `imported_ir_indexes`? It would have been useful to me for the comment to mention this is tracking indexes in `results`.

toolchain/check/check.cpp
```diff
@@ -212,6 +212,56 @@ struct UnitInfo {
...
+    const auto& direct_ir = **import.unit_info->unit->sem_ir;
+    imported_irs[direct_ir.check_ir_id().index] = results.size();
+    results.push_back({.node_id = import.names.node_id,
+                       .sem_ir = &direct_ir,
+                       .is_export = import.names.is_export});
+  }
+
+  // Loop through direct imports for any indirect exports. The underlying vector
+  // is appended during iteration, so take the size at the start.
+  for (int direct_index : llvm::seq(results.size())) {
```
zygoloid: How would you feel about:
```suggestion
  // Loop through direct imports for any indirect exports. The underlying vector
  // is appended during iteration, so take the size first.
  const int direct_imports = results.size();
  for (int direct_index : llvm::seq(direct_imports)) {
```
... so that the code also expresses that we're intentionally computing the size before the loop?

toolchain/check/check.cpp
```diff
@@ -212,6 +212,56 @@ struct UnitInfo {
...
+    for (const auto& indirect_ir :
+         results[direct_index].sem_ir->import_irs().array_ref()) {
+      if (!indirect_ir.is_export) {
+        continue;
+      }
+
+      auto& indirect_index =
+          imported_irs[indirect_ir.sem_ir->check_ir_id().index];
+      bool is_export = results[direct_index].is_export;
```
zygoloid: Maybe move this out of the inner `for` loop?

zygoloid (approved):

toolchain/check/check.cpp

toolchain/check/check.cpp

toolchain/check/check.cpp
jonmeow: How about `ir_to_result_index`

PR #3973, opened 2024-05-21
Handle merging of conflicting 'export name' cross-package imports.

jonmeow:


jonmeow:
Depends on #3971

zygoloid (approved):
Looks good. Given that we import namespace names more eagerly than other names, would it be useful to have a test for a conflict between a namespace name and a non-namespace name?

jonmeow:
I think you're thinking of same-library imports, where namespaces are intended to dominate. Cross-package imports are handled very differently, with namespaces just being a conflict, and cross_package_import.carbon should already be testing.

PR #3975, opened 2024-05-21
Attempt to improve the performance of function value profiling.

zygoloid:
Don't look at the contents of the empty instruction block.


jonmeow (approved):

zygoloid:
No evidence this is faster: closing.

PR #3976, opened 2024-05-21
Switch the prelude to using 'export import'.

jonmeow:
Note, I assume this doesn't affect all the TODOs (like for i32.carbon), but does cut some things down (and requires updating some tests that have prelude name conflicts, but I think they were intended to be updated this way).


jonmeow:
Depends on #3973

toolchain/check/testdata/function/builtin/call_from_operator.carbon
```diff
@@ -24,53 +14,6 @@ impl i32 as Core.Add {
 
 var arr: [i32; 1 + 2] = (3, 4, 3 + 4);
```
zygoloid: Given that this is a low-level test for calling a builtin via an operator interface, I think we should make it a no-prelude test.

toolchain/check/testdata/index/fail_negative_indexing.carbon
```diff
@@ -6,11 +6,7 @@
 
 var a: (i32, i32) = (12, 6);
 // TODO: Only diagnose this once, and make the diagnostic clearer.
```
zygoloid: This TODO no longer applies here. (The underlying issue hasn't been addressed, but we have a `context.TODO` call to track that.)
```suggestion
```

zygoloid (approved):

toolchain/check/testdata/function/builtin/call_from_operator.carbon
jonmeow: Done

toolchain/check/testdata/index/fail_negative_indexing.carbon
jonmeow: Done

PR #3977, opened 2024-05-22
Fix handling for indirectly imported entities.

jonmeow:
A member of an imported entity can be an ImportRefUnloaded in an imported SemIR. To address this, find a loaded version of it for use.

### DIFF
--- a/toolchain/check/import_ref.cpp
+++ b/toolchain/check/import_ref.cpp
@@ -1266,12 +1266,12 @@ class ImportRefResolver {
   llvm::SmallVector<Work> work_stack_;
 };
 
-// Returns a pair of ImportIRInsts equivalent to the ImportRef currently being
+// Returns a list of ImportIRInsts equivalent to the ImportRef currently being
 // loaded (including the one pointed at directly by the ImportRef), and the
 // final instruction's type ID.
 //
 // This addresses cases where an ImportRefUnloaded may point at another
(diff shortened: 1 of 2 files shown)


jonmeow:
Depends on #3976

toolchain/check/import_ref.cpp
```diff
@@ -1265,20 +1266,58 @@ class ImportRefResolver {
   llvm::SmallVector<Work> work_stack_;
 };
 
+// Returns a pair of ImportIRInsts equivalent to the ImportRef currently being
```
```suggestion
// Returns a list of ImportIRInsts equivalent to the ImportRef currently being
```
(To me the old comment suggested that this would return two (a pair of) `ImportIRInst`s.)

toolchain/check/import_ref.cpp
```diff
@@ -1265,20 +1266,58 @@ class ImportRefResolver {
...
+// loaded (including the one pointed at directly by the ImportRef), and the
+// final instruction's type ID.
+//
+// This addresses cases where an ImportRefUnloaded may point at another
+// ImportRefUnloaded. The ImportRefResolver requires a SemIR with
```
```suggestion
// ImportRefUnloaded. The ImportRefResolver requires a SemIR with a
```

zygoloid (approved):

toolchain/check/import_ref.cpp

toolchain/check/import_ref.cpp

PR #3978, opened 2024-05-22
Switch type literals to use builtin functions.

jonmeow:
Modifies the core package and literal handling to use factory functions for standard type literals.

Updates function/builtin/import.carbon to stop depending on the prelude, since it would list all the impls in the core file. Updates alias/builtins.carbon to be failing (the type values cannot be aliased).


jonmeow:
Depends on #3977

core/prelude.carbon
```diff
@@ -6,11 +6,6 @@
 
 package Core library "prelude";
 
+export import library "prelude/types/bool";
 export import library "prelude/operators";
 export import library "prelude/types";
```
zygoloid, line 10: Should "prelude/types" export "prelude/types/bool"?

zygoloid (approved):
Looks good. I wonder how much all these extra function calls will slow down @chandlerc's performance benchmark.

jonmeow:
> Looks good. I wonder how much all these extra function calls will slow down @chandlerc's performance benchmark.

Noted this on #toolchain, but I think it's WAI

core/prelude.carbon
jonmeow: Done

PR #3979, opened 2024-05-22
Replace `FoldingSet` with `DenseMap` for instruction canonicalization.

zygoloid:
Switch from recursing into non-canonical instruction fields to separately canonicalizing those fields. This means we now form canonical `InstBlockId`s, `TypeBlockId`s, `IntId`s, `FloatId`s, and `BindNameId`s at least in the cases when they're referenced by a constant instruction.

This reduces the overall runtime for @chandlerc's 10MLoC example by 27.5% on my machine.


toolchain/base/value_store.h
```diff
@@ -72,9 +77,30 @@ struct FloatId : public IdBase, public Printable<FloatId> {
 };
 constexpr FloatId FloatId::Invalid(FloatId::InvalidIndex);
 
+// DenseMapInfo for llvm::APFloat, for use in the canonical float value store.
+// TODO: Remove this once our new hash table is available.
+struct APFloatDenseMapInfo {
```
jonmeow: Why this approach of a value that needs to be passed in, instead of overloading `struct llvm::DenseMapInfo` as we do elsewhere? i.e., removing the need for a `FloatValueStore` alias.

Is the overload in LLVMContextImpl.h an issue? Would it be unreasonable to include it? Though, that raises a different question -- why does your implementation diverge from LLVM's, avoiding the Bogus value?

https://github.com/llvm/llvm-project/blob/main/llvm/lib/IR/LLVMContextImpl.h#L77-L90

toolchain/sem_ir/bind_name.h
```diff
@@ -0,0 +1,74 @@
...
+  static auto getEmptyKey() -> BindNameInfo {
+    return BindNameInfo{.name_id = NameId(-1),
+                        .enclosing_scope_id = NameScopeId::Invalid,
+                        .bind_index = CompileTimeBindIndex::Invalid};
+  }
+  static auto getTombstoneKey() -> BindNameInfo {
+    return BindNameInfo{.name_id = NameId(-1),
+                        .enclosing_scope_id = NameScopeId::Invalid,
+                        .bind_index = CompileTimeBindIndex::Invalid};
+  }
```
jonmeow: Should these values be different?

toolchain/sem_ir/bind_name.h
```diff
@@ -0,0 +1,74 @@
...
+    -> HashCode {
+  Hasher hasher(seed);
+  hasher.Hash(value);
+  return static_cast<HashCode>(hasher);
+}
+
+// DenseMapInfo for BindNameInfo.
+struct BindNameInfoDenseMapInfo {
+  static auto getEmptyKey() -> BindNameInfo {
+    return BindNameInfo{.name_id = NameId(-1),
```
jonmeow: NameId::Invalid?

toolchain/sem_ir/block_value_store.h
```diff
@@ -47,6 +48,25 @@ class BlockValueStore : public Yaml::Printable<BlockValueStore<IdT>> {
...
+      it->first.data = Get(id);
+      it->second = id;
+    }
+    return it->second;
+  }
+
+  // Promotes an existing block ID to a canonical block ID, or returns an
+  // existing canonical block ID if the block was already added. The specified
+  // block must not be modified after this point.
+  auto GetOrAddCanonical(IdT id) -> IdT {
```
jonmeow: I'd kind of prefer a different name versus the above, since it's deduplicating an existing block instead of potentially adding a new one. Maybe something like `Finalize`, `GetAs`, or `MakeCanonical`?

toolchain/sem_ir/bind_name.h
```diff
@@ -0,0 +1,74 @@
...
+
+  // The name.
+  NameId name_id;
+  // The enclosing scope.
+  NameScopeId enclosing_scope_id;
+  // The index for a compile-time binding. Invalid for a runtime binding.
+  CompileTimeBindIndex bind_index;
+};
+
+// Hashing for BindNameInfo.
```
jonmeow, line 28: Where you do these, would it be worth more specifically pointing at hashing.h? I'm not sure the "CarbonHashValue" name was helping me here, it just looks odd next to the way DenseMapInfo works.

toolchain/sem_ir/bind_name.h
```diff
@@ -0,0 +1,74 @@
...
+    return static_cast<uint64_t>(HashValue(val));
+  }
+  static auto isEqual(const BindNameInfo& lhs, const BindNameInfo& rhs)
+      -> bool {
+    return std::memcmp(&lhs, &rhs, sizeof(BindNameInfo)) == 0;
+  }
+};
+
+// Value store for BindNameInfo. In addition to the regular ValueStore
+// functionality, this can provide optional canonical IDs for BindNameInfos.
```
jonmeow, line 60: Why doesn't this use CanonicalValueStore?

toolchain/sem_ir/block_value_store.h
```diff
@@ -82,6 +102,37 @@ class BlockValueStore : public Yaml::Printable<BlockValueStore<IdT>> {
...
+
+    friend auto CarbonHashValue(CanonicalBlock block, uint64_t seed)
+        -> HashCode {
+      Hasher hasher(seed);
+      hasher.HashSizedBytes(block.data);
+      return static_cast<HashCode>(hasher);
+    }
+  };
+
+  struct CanonicalBlockDenseMapInfo {
```
jonmeow, line 120: You seem to be doing some nuanced work here with SpecialData, would it be worth some comments explaining how this is intended to work?

toolchain/sem_ir/block_value_store.h
```diff
@@ -82,6 +102,37 @@ class BlockValueStore : public Yaml::Printable<BlockValueStore<IdT>> {
...
+    static constexpr ElementType SpecialData[1] = {ElementType::Invalid};
+    static auto getEmptyKey() -> CanonicalBlock {
+      return CanonicalBlock{{SpecialData, SpecialData}};
+    }
+    static auto getTombstoneKey() -> CanonicalBlock {
+      return CanonicalBlock{{SpecialData, SpecialData + 1}};
```
jonmeow: I was trying to figure out what's happening here, would be below be okay/equivalent?

```suggestion
      return CanonicalBlock{llvm::ArrayRef(SpecialData, 0)};
    }
    static auto getTombstoneKey() -> CanonicalBlock {
      return CanonicalBlock{llvm::ArrayRef(SpecialData, 1)};
```

toolchain/sem_ir/block_value_store.h
```diff
@@ -82,6 +102,37 @@ class BlockValueStore : public Yaml::Printable<BlockValueStore<IdT>> {
...
+      Hasher hasher(seed);
+      hasher.HashSizedBytes(block.data);
+      return static_cast<HashCode>(hasher);
+    }
+  };
+
+  struct CanonicalBlockDenseMapInfo {
+    static constexpr ElementType SpecialData[1] = {ElementType::Invalid};
+    static auto getEmptyKey() -> CanonicalBlock {
+      return CanonicalBlock{{SpecialData, SpecialData}};
```
jonmeow: You seem to be relying on it tracking a data pointer here, but does it need to? Is it worth verifying that LLVM never optimizes away the "empty ArrayRef" case?

toolchain/sem_ir/inst.h
```diff
@@ -445,6 +453,29 @@ class InstBlockStore : public BlockValueStore<InstBlockId> {
   }
 };
 
+inline auto CarbonHashValue(const Inst& value, uint64_t seed) -> HashCode {
```
jonmeow, line 459: Similar to above, I'd appreciate some comments explaining what's going on here.

toolchain/sem_ir/inst.h
```diff
@@ -445,6 +453,29 @@ class InstBlockStore : public BlockValueStore<InstBlockId> {
...
 
+template <>
+struct llvm::DenseMapInfo<Carbon::SemIR::Inst> {
+  using Inst = Carbon::SemIR::Inst;
+  static auto getEmptyKey() -> Inst {
+    return Inst(-1, Carbon::SemIR::TypeId::Invalid, 0, 0);
+  }
+  static auto getTombstoneKey() -> Inst {
+    return Inst(-2, Carbon::SemIR::TypeId::Invalid, 0, 0);
+  }
```
jonmeow, line 475: Instead of putting custom int values into the InstKind, had you considered something like a new InstKind::Invalid and instead modifying arg0?

Alternately, should this be validating that we never allow these values as an InstKind?

jonmeow (commented):
The high level approach looks good, though I have some questions about how the various hash infos are implemented.

toolchain/base/value_store.h
zygoloid: We can't specialize `llvm::DenseMapInfo<llvm::APFloat>` ourselves because we don't own either of those types. I didn't find that implementation in `LLVMContextImpl.h` when searching -- I wonder why it's *there* of all places; that doesn't make any sense at all. But we can't use that anyway because it's not in a public header file.

I looked at `Bogus` but from a brief inspection it looked like `APFloat` would assert if you tried to copy it in that state. But maybe I was wrong -- if LLVM itself is doing this then I assume it's fine. Switching to using that.

toolchain/sem_ir/bind_name.h
zygoloid: Oops, fixed, thanks.

toolchain/sem_ir/bind_name.h
zygoloid: Changed to be different from `NameId::Invalid` to avoid colliding with a theoretically possible value.

jonmeow (approved):

jonmeow:
(approving to unblock)

toolchain/sem_ir/bind_name.h
jonmeow: Just to note -- I'm not seeing how you're doing this, but would it be worth adding NameId::HashAsEmpty, HashAsTombstone so that it's clearer on the NameId that these need to be reserved values?

toolchain/sem_ir/bind_name.h
zygoloid: We don't need to canonicalize most `BindNameInfo`s -- the intent here is to only canonicalize the ones that form part of constants, so we're not performing an extra hash table insertion here for every name we declare. `CanonicalValueStore` is intended for the case where we're only storing canonical values.

toolchain/sem_ir/bind_name.h
zygoloid: I ended up switching to storing the intentionally-bad value in the compile-time binding index, where I think the risk of having more reserved values is much lower. I understand the motivation behind adding names for the extra states, but that would seem to legitimize using those values as valid values, and the point here is to form a representation that's not a valid value, so I'm hesitant. I've added a comment to the declaration of the type to further reduce the risk.

toolchain/sem_ir/block_value_store.h
zygoloid: Done.

toolchain/sem_ir/block_value_store.h
zygoloid: Added.

toolchain/sem_ir/block_value_store.h
zygoloid: Done.

toolchain/sem_ir/block_value_store.h
zygoloid: It's documented to behave as a pointer and a size, even when the size is zero.

toolchain/sem_ir/inst.h
zygoloid: Added a cross-reference to the documentation in hashing.h.

toolchain/sem_ir/inst.h
zygoloid: Added a CHECK to the `Inst` constructor.

PR #3980, opened 2024-05-23
Singular `extern` declarations

jonmeow:
Each entity is restricted to one, optional `extern` declaration. If used, it must be imported by the defining library. The defining library annotates the existence of an `extern` with the `has_extern` modifier.


proposals/p3980.md
```diff
@@ -0,0 +1,397 @@
...
+### Type coherency
+
+In the context of the example that is the [problem](#problem), `C` will produce
+the same type regardless of whether `C` is the `extern` or non-`extern`
+declaration. This means that both function signatures have identical types.
+
+We do this by having the non-`extern` declaration import the `extern`
+declaration. Because one can see the other, the compiler can easily use the same
+type for both declarations. This makes it easy for compilation of other
+libraries, which may be importing either or both declarations, to easily
```
josh11b:
```suggestion
libraries, which may be importing either the `extern` or both declarations, to easily
```
or
```suggestion
libraries, which may be directly importing either or both declarations, to easily
```
or
```suggestion
libraries, which may be importing both declarations, to easily
```

proposals/p3980.md
```diff
@@ -0,0 +1,397 @@
...
+
+The information that an `extern` declaration exists somewhere else is expected
+to offer long-term refactoring benefits.
+
+### Alternate names for `has_extern`
+
+We're not entirely happy with the `has_extern` name, although it is ambiguous. A
+few other names that might be worth considering are `is_extern` and `externed`.
+However, it's not clear either of these names are better. The name may change in
+the future if a better model is brought up.
```
chandlerc: Just as a quick note, I think it'd be worth filing a leads question for the `has_extern` aspect of the proposal. Maybe it just ends up affirming "yes, as suggested here", but this aspect in particular is the one that I at least am most ambivalent about, and it seems good to have a issue to focus any discussion _which direction_ there, and let the review focus on the proposal itself and documenting the other, less controversial aspects.

proposals/p3980.md
```diff
@@ -0,0 +1,397 @@
...
+### Versus current state
+
+The key changes are in comparison to the design from
+[#3762](https://github.com/carbon-language/carbon-lang/pull/3762):
+
+-   There may only be one `extern` declaration.
+-   The owning library is required to import the `extern` declaration.
+-   When there is an `extern` declaration, the first owning declaration (either
+    forward declaration or definition) must be marked as `has_extern`.
+    -   If there are two, it is not on the second.
```
josh11b: Also that there must be an owning declaration in the api file?

proposals/p3980.md
```diff
@@ -0,0 +1,397 @@
...
+`extern` and `has_extern` declarations, and offers additional parity in
+modifiers. It also makes it easy for a tool to know if it's missing a
+declaration.
+
+The information that an `extern` declaration exists somewhere else is expected
+to offer long-term refactoring benefits.
+
+### Alternate names for `has_extern`
+
+We're not entirely happy with the `has_extern` name, although it is ambiguous. A
```
josh11b:
```suggestion
We're not entirely happy with the `has_extern` name, although it is not ambiguous. A
```

proposals/p3980.md
jonmeow: Filed #3986

proposals/p3980.md
jonmeow: Done

proposals/p3980.md
```diff
@@ -0,0 +1,398 @@
...
+
+The `has_extern` modifier must be present on the first owning declaration, when
+there is an `extern` declaration. Because the first owning declaration must be
+in an API file, this will only be present in API files.
+
+The modifier allows libraries to restrict whether they have an `extern`
+declaration. It will also be used to support tool-based validation that the
+`extern` declaration is imported as required.
+
+### Versus current state
```
zygoloid: One thing I'm unclear on here, mostly because I'm not entirely sure what the current state is:

Given
```
library "extern";
extern class C;
fn F() -> C;
```
```
library "owning";
import "extern";
has_extern class C {}
```
```
library "distant";
import "owning";
```
```
library "client";
import "extern";
import "distant";
fn CallF() {
  F();
}
```
... is the call to `F()` in `CallF()` valid? That is: is `C` a complete type in that compilation?

If so, then removing the `import "owning";` from library `"distant"` breaks the build of the `"client"` library. That seems problematic.

If not, then presumably that only applies for `has_extern` types. That is, I think we want this to work:
```
library "owning";
class C {}
```
```
library "distant";
import "owning";
fn F() -> C;
```
```
library "client";
import "distant";
fn CallF() {
  F();
}
```
... and I think the relevant difference here is whether `C` is known to have `extern` declarations.

A rule we've previously discussed, but I don't remember whether we've agreed to or not, is:

- Definitions of types are transitively visible (even if not explicitly re-exported) if they are not `has_extern`
- Definitions of `has_extern` types are not transitively visible, and those types are instead incomplete if their owning library (or a library that `export name;`s the type, or maybe `export name;`s something involving the type?) is not directly imported.

I'm fine saying this is off-topic for this PR, but if so, it'd be useful to at least mention what the current state is -- if we don't have such a rule, then capturing that there is a known issue with removing distant imports in this PR would be useful.

zygoloid (commented):
Looks good to me other than the open leads issue around `has_extern` syntax.

proposals/p3980.md
jonmeow: > ... is the call to `F()` in `CallF()` valid? That is: is `C` a complete type in that compilation?

`C` is never imported by `client` so no. There is no reason its type would be complete.

It might be different if something pulled the complete definition in for `client`.

> If so, then removing the `import "owning";` from library `"distant"` breaks the build of the `"client"` library. That seems problematic.

N/A, although you might get interactions such as this, just not from the example. e.g., if you also had something like your later example used in combination, which triggered the definition of `C` to be pulled in.

> If not, then presumably that only applies for `has_extern` types. That is, I think we want this to work:
> 
[elided]
> 
> ... and I think the relevant difference here is whether `C` is known to have `extern` declarations.

Yes, because this dealing with a complete type, and we have only one type.

> A rule we've previously discussed, but I don't remember whether we've agreed to or not, is:
> 
> * Definitions of types are transitively visible (even if not explicitly re-exported) if they are not `has_extern`
> * Definitions of `has_extern` types are not transitively visible, and those types are instead incomplete if their owning library (or a library that `export name;`s the type, or maybe `export name;`s something involving the type?) is not directly imported.
> 
> I'm fine saying this is off-topic for this PR, but if so, it'd be useful to at least mention what the current state is -- if we don't have such a rule, then capturing that there is a known issue with removing distant imports in this PR would be useful.

I think declarations are transitively visible regardless of `has_extern`. The only difference with `has_extern` is whether declarations being transitively visible might have significant effect.

Note, I think this kind of thing is an unavoidable consequence of type consolidation: we can't expect `extern` to be using the same type as `has_extern`, but have the same type behave differently in use. [as I recall prior discussion, the idea behind having separate extern and non-extern types was to address this problem differently]

A similar example might be:

```
class C;
fn F() -> C*;
class C {}
fn G() -> C { return *F; }
```

Here, even though `F()` returns an incomplete type at declaration, it's the same type as later defined, and so the code is valid.

jonmeow: Discussing the awkward consequences of `has_extern` in the combination case, in particular if you have code that imports functions which return `has_extern` and `extern` versions of the type (indirectly imported, and with extra consideration cross-package where names are lazy loaded), I think there are three main options:

1. The order of calling functions matters, under the information accumulation principle. If you call the `has_extern` version first, the type is complete when the `extern` version is called. In the other order, it's still incomplete.

2. Whenever a library has an exported entity which uses a `has_extern` type, it adds that `has_extern` type to a list of types which _must_ be evaluated on import. This is not transitive unless re-used. Thus call order doesn't matter, and a complete type is available.

    - There's a simpler option here where all imported `has_extern` can be treated this way, but the restriction to exported entities may simplify matters. Although in turn, it may become really complex for class members.

3. If a `has_extern` type is indirectly returned (and not directly imported) the type is always incomplete.

    - I think this is the rule you're describing. Note it makes `has_extern` really special.

To be sure, I think (1) is status quo.

jonmeow: Filed #4025

jonmeow:
I've updated the proposal based on discussion from #3986 (the new `extern` + `extern library`) and #4025 (new wording around indirect imports), plus some side discussion (future work "`extern` and template interactions", alternative "Have types with `extern` members re-export them").

@zygoloid @chandlerc Note in particular the alternative "Separate require-direct-import from non-owning declarations", which is what I was trying to ask for the "as a consequence" in https://github.com/carbon-language/carbon-lang/issues/3986#issuecomment-2204765916. I think there might have been a misunderstanding that I was suggesting "extern" should mean *both* effects; what I was thinking instead is that maybe it should be *modular*, probably with some other keyword meaning "require direct import".

PS: Sorry if I've missed an old-style `extern` to new-style `extern` update, please let me know if you see something.

proposals/p3980.md
```diff
@@ -0,0 +1,660 @@
...
+import library "a";
+extern fn F() -> C*;
+```
+
+Here, the return types of `F` differ.
+
+This proposal aims to address the differing return types by unifying the type of
+`C` regardless of whether it's `extern`. This could be done under multiple
+different approaches, and this proposal aims for one which is efficient to
+process.
```
chandlerc: nit: Maybe make it more explicit the processing in question is the implementation?

```suggestion
different approaches, and this proposal aims for one which enables efficient
implementation strategies.
```

proposals/p3980.md
```diff
@@ -0,0 +1,660 @@
...
+```
+package Other library "type";
+extern class MyType {
+  fn Print();
+};
+```
+
+```
+package Other library "use_type";
+fn Make() -> MyType*;
```
chandlerc: I think this is missing:

```suggestion
package Other library "use_type";
import library "type";
fn Make() -> MyType*;
```

proposals/p3980.md
```diff
@@ -0,0 +1,660 @@
...
+```
+
+```
+package Other library "extern" api;
+extern library "type" class MyType;
+```
+
+```
+package Other library "use_extern";
+fn Print(val: MyType*);
```
chandlerc: And here I think is missing:

```suggestion
package Other library "use_extern";
import library "extern";
fn Print(val: MyType*);
```

proposals/p3980.md
```diff
@@ -0,0 +1,660 @@
...
+};
+```
+
+```
+package Other library "use_type";
+fn Make() -> MyType*;
+```
+
+```
+package Other library "extern" api;
```
chandlerc:
```suggestion
package Other library "extern";
```

proposals/p3980.md
```diff
@@ -0,0 +1,660 @@
...
+
+[Issue #3986](https://github.com/carbon-language/carbon-lang/issues/3986)
+discussed other syntaxes for `extern` + `extern library`. These were mainly
+`has_extern`/`is_extern`/`externed` + `extern`. The current syntax is more
+verbose, but was chosen because:
+
+-   The behavior of imports being incomplete was something leads wanted to offer
+    separate from the non-owning declaration, and `extern` was decided to be a
+    good name for that.
+-   It has extra detail in `extern library "<owning_library>"`.
```
chandlerc: I feel like this isn't capturing all of the rationale from the issue here...

Working off of the text in that issue, but also putting the emphasis on different parts and trying to structure it into a bullet point list here I ended up with something more like:
```suggestion
-   Providing behavior of imports being incomplete orthogonally but using the same keyword allowed us have a single meaning for the `export` keyword and consistent usage of it on declarations.
    -   Enabling this behavior independently wasn't seen as a problem by the leads and potentially useful even if not the primary motivation. The primary motivation was orthogonality of features.
    -   Within this context `extern` seemed like an acceptable name, and alternative names seemed significantly less good. It also gives us only one new keyword.
-   Adding the owning library with `extern library "<owning_library"` will hopefully give us better diagnostics, better human readability and understandability of the code.
    -   It is _very_ verbose, but this verbosity goes on the forward declaration in the non-owning library, which we expect to be read less often.
    -   And if in practice we find the verbosity becomes a significant issue we can revisit syntaxes to address that specifically. For example, if we have significant repetiton, we might consider a grouping structure such as `extern library "..." { <many forward declarations> }`.
```

Not sure this wording works for you, and happy to have something different. But I think it would be good to incorporate some of these points:
- the specific diagnostic & understandability benefits expected
- the goal of a somewhat orthogonal decomposition of features
- the fact that `extern` wasn't endorsed as a *great* name, just that none of teh alternatives we found were better.
- the mitigating factors of the verbosity

chandlerc (commented):
Thanks, overall I think the proposal looks great. Added some super minor suggestions in line, but they don't seem to be very significant.

I did suggest another attempt at writing up rationale for the alternative, although I'm not confident in the words I ended up with either. Maybe you or others have better ways of summarizing this.

In general, with whatever improvements to the alternative syntax rationale we converge on, I'm happy with this landing.

proposals/p3980.md
jonmeow: Thanks, this was why I asked for rationale on #3986. I've adopted this, although I'm having trouble understanding it. In particular:
 
> Providing behavior of imports being incomplete orthogonally but using the same keyword allowed us have a single meaning for the `export` keyword and consistent usage of it on declarations.

Orthogonally from what? Did you mean `extern` instead of `export`, or is there a connection to `export` I'm missing?

(if you mean having `extern` on a declaration accesses two orthogonal features, creating a single meaning, wouldn't some developers use it for each feature separately -- i.e. there are three possible meanings [first feature/second/both]?)

chandlerc: > Thanks, this was why I asked for rationale on #3986. I've adopted this, although I'm having trouble understanding it. In particular:
> 
> > Providing behavior of imports being incomplete orthogonally but using the same keyword allowed us have a single meaning for the `export` keyword and consistent usage of it on declarations.
> 
> Orthogonally from what? Did you mean `extern` instead of `export`, or is there a connection to `export` I'm missing?

I meant `extern`, sorry about that.

> 
> (if you mean having `extern` on a declaration accesses two orthogonal features, creating a single meaning, wouldn't some developers use it for each feature separately -- i.e. there are three possible meanings [first feature/second/both]?)

All I meant was separating the "must import directly" and "having a forward declaration in an unowning library". We can now think about them independently, and have the latter depend on the former. ("indpendent" maybe a better word than "orthogonal" here)

jonmeow: Thanks, I've tried to rephrase this, can you make sure I've captured it correctly?

jonmeow: (did a couple small rephrasings to try to better mix the suggested phrasing with examples)

chandlerc: Reading the new text, I feel like it conveys a very different emphasis than what I intended.

The new text seems to frame this as having access to both aspects of this as fundamentally important, whereas I was trying to suggest that this wasn't *important*, just acceptable and it let us factor the syntax in a way that ended up giving a more declarative meaning to each component, and layered semantics as you compose them.

But maybe my suggested edit wasn't clear here, maybe @zygoloid can help with an a different phrasing that is more clear?

jonmeow: Trying another rephrasing, switching emphasis on what's important versus acceptable.

Note, I'm not sure if you and I just view the terms used (orthogonal, independent, declarative meaning) differently.

As implemented, I would say `extern class C;` has two possible meanings:

1. There is a `extern library "foo" class C;` declaration in a differently library.
2. This was intended to require direct imports, and shouldn't have another declaration in a different library.

If I understand correctly, you're saying that because (1) is optional, it's orthogonal and independent of (2). `extern` has multiple possible behaviors that apply, so has more declarative meaning. Am I interpreting that correctly?

For me, to try to clarify my interpretation of the words:

- (1) requires (2), so the features are related rather than orthogonal.
- The approach means using `extern` for (2) makes (1) optional, and shifts control to other libraries rather than the owning library, so I would say they're intertwined rather than independent.
- Determining whether (1) applies [when reading `extern class C;`] requires reading all imported libraries (or, clear comments). I would say the declaration has less meaning on its own because it requires more context to understand the behavior applied.

[this different interpretation is why I'm trying different phrasings, to try to find something that works for us both]

proposals/p3980.md
```diff
@@ -0,0 +1,696 @@
...
+
+-   The primary motivation is to provide feature (1).
+-   Leads consider it acceptable to support (2) separate from (1), and don't see
+    a problem with making (1) optional to achieve this (in other words, making
+    the `extern library "<owning_library>"` declaration optional).
+    -   It's okay that that `extern library "<owning_library>"` can be added and
+        removed from imported libraries without modifying the owning library.
+    -   If a developer considers it important to disambiguate the intended use
+        of a declaration `extern class C;` and whether there should be a
+        declaration in a separate library, they can add comments.
```
zygoloid: Some small tweaks and a bit more motivation:

```suggestion
-   Leads wanted a syntax on the owning declaration that states something
    positive about the owning declaration itself, rather than expressing that
    other declarations exist, which suggests that the syntax on the owning
    declaration should provide feature (2).
-   Leads consider it valuable, though secondary, to support (2) separate from (1), and
    find it acceptable to make (1) optional to achieve this (in other words, making
    the `extern library "<owning_library>"` declaration optional).
    -   It's okay that that `extern library "<owning_library>"` can be added and
        removed from imported libraries without modifying the owning library.
    -   If a developer considers it important to disambiguate the intended use
        of a declaration `extern class C;` and whether there should be a
        declaration in a separate library, they can add comments.
```

... then I think this reasonably captures our discussions.

proposals/p3980.md
jonmeow: Merging phrasing suggestion from @zygoloid

zygoloid: I suggested some tweaks that Jon has applied; @chandlerc, how does it look to you now?

jonmeow:
FYI, added extended notes on syntactic matching (specifically, why it won't apply for `extern library`).

proposals/p3980.md
chandlerc: Yeah, this works well for me. Thanks (and sorry it took a while to circle back here).

chandlerc (approved):
LGTM, and approved for leads!

PR #3981, opened 2024-05-23
Replace image file in /docs/images/ directory and update the svg edit to reflect proposal #3927.

### LINKED ISSUE #3969: Update sample code in README
### Description of the bug:

The examples in the README need to be updated to reflect the change to package syntax from proposal https://github.com/carbon-language/carbon-lang/pull/3927 .

### Link to documentation:

https://github.com/carbon-language/carbon-lang/blob/trunk/README.md

`package <something> api;` should just be `package <something>;`

_No response_

ryoari:
Replace image file in /docs/images/ directory and update the svg editing document to reflect the change to package syntax from proposal #3927

Closes #3969


jonmeow:
In quicksort, we should probably keep the same formatting (font and colors) that was used previously. Can you please restore the changes there to those?

jonmeow:
Note, the other two look good.

ryoari:
yeah looks a bit odd, I am trying to fix it while keeping the updated carbon syntax, would you mind another iteration or do you feel consistent would be better?

ryoari:
quicksort_snippet
how does this look?

jonmeow:
I think we really would like to keep a consistent look-and-feel. We're trying to keep a similar look as with GitHub (visible in [snippets.md](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/images/snippets.md), in particular the dark default theme) so that it "feels" like other code snippets that would be seen in GitHub.

Note too, there's already a skew with snippets.md (the C++ parts of snippets was changed without updating images); that's a separate issue though, and this does fix #3969.

jonmeow (approved):
Thanks for the edits here.

PR #3982, opened 2024-05-23
Delete CarbonInfraBot comments with no replies when they become outdated.

zygoloid:


.github/workflows/pre_commit_suggestions.yaml
```diff
@@ -60,3 +60,16 @@ jobs:
...
             reviewdog -f=diff -f.diff.strip=1 -reporter=github-pr-review
+
+  delete-stale-suggestions:
+    if: |
+      github.event.workflow_run.event == 'pull_request' &&
+      github.actor != 'jonmeow'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Delete stale suggestions
+      - uses: Ardiannn08/resolve-outdated-comment@v1.1
```
jonmeow: Why 1.1 when 1.3 is available?

https://github.com/Ardiannn08/resolve-outdated-comment

Also, please pin this dependency. You can use https://app.stepsecurity.io/secureworkflow for that.

.github/workflows/pre_commit_suggestions.yaml
```diff
@@ -60,3 +60,16 @@ jobs:
           cat ./diff | \
           GITHUB_EVENT_PATH=./event \
             reviewdog -f=diff -f.diff.strip=1 -reporter=github-pr-review
+
+  delete-stale-suggestions:
+    if: |
+      github.event.workflow_run.event == 'pull_request' &&
+      github.actor != 'jonmeow'
```
jonmeow:
```suggestion
      github.event.workflow_run.event == 'pull_request'
```

Don't think I need to be mentioned here

jonmeow (dismissed):
This looks helpful, though please be sure to resolve the security issue before merging.

.github/workflows/pre_commit_suggestions.yaml
jonmeow: (the secureworkflow thing will also add other features, similar to what's done above)

zygoloid: I was using 1.1 because that's what [the example uses](https://github.com/marketplace/actions/resolve-outdated-comments); changed to 1.3 and pinned.

.github/workflows/pre_commit_suggestions.yaml
zygoloid: Done.

chandlerc (commented):
If this works, will be a delight. Really the only friction I hit w/ the suggested edits, and this seems like it gives us really the best of both worlds.

(Provided it works... and thanks for the review and such Jon!)

jonmeow (approved):

PR #3983, opened 2024-05-24
Expand bazel retries in tests.

jonmeow:
Trying to better address ephemeral issues such as:
https://github.com/carbon-language/carbon-lang/actions/runs/9226025647/job/25384877352

`bazelisk test` currently has some retries, and this migrates those to a script and copies it to other places that we invoke bazelisk.

### DIFF
--- a/scripts/run_bazel.py
+++ b/scripts/run_bazel.py
@@ -12,15 +12,66 @@ Exceptions. See /LICENSE for license information.
 SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 """
 
-import os
-import sys
+import argparse
+import subprocess
+import time
 
 import scripts_utils
 
 
 def main() -> None:
+    parser = argparse.ArgumentParser(description="Runs bazel.")
+    parser.add_argument(
+        "--attempts",
+        metavar="COUNT",
+        type=int,
+        default=1,
+        help="The number of attempts to execute the command, automatically "
+        "retrying errors that may be transient.",
+    )
+    parser.add_argument(
+        "--jobs-on-last-attempt",
+        metavar="COUNT",
+        type=int,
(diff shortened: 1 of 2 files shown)


scripts/run_bazel.py
```diff
@@ -12,15 +12,66 @@
...
+        attempt += 1
+        if attempt == script_args.attempts and script_args.jobs_on_last_attempt:
+            with open("user.bazelrc", "a") as bazelrc:
+                bazelrc.write(
+                    f"build --jobs={script_args.jobs_on_last_attempt}\n"
+                )
+
+        p = subprocess.run([bazel] + bazel_args)
+
+        # If this was the last attempt, we're done. Alternately,
```
chandlerc: Unsure if this was left over from some other edit, or unfinished sentence?

If just leftover, can remove:
```suggestion
        # If this was the last attempt, we're done.
```

scripts/run_bazel.py
```diff
@@ -12,15 +12,66 @@
...
+        if attempt == script_args.attempts:
+            exit(p.returncode)
+
+        # Several error codes are reliably permanent, break immediately.
+        # `0`  -- Success.
+        # `1`  -- The build failed.
+        # `2`  -- Command line or environment problem.
+        # `3`  -- Tests failed or timed out, we don't retry at this layer
+        #         on execution timeout.
+        # `4`  -- No tests found, which should be impossible here.
```
chandlerc:
```suggestion
        # `4`  -- Test command but no tests found.
```

chandlerc (approved):
Ooooof on still more transient failures. But yeah, might as well factor it out and reuse it at this point. =]

Super minor stuff inline, submit whenever.

scripts/run_bazel.py
jonmeow: Left over from an edit, I was considering combining these.

PR #3987, opened 2024-05-24
Use the driver to link.

jonmeow:
Rather than cc_binary, try using link from #3922


jonmeow:
Note totally sure we want to do this right now, but was poking and seems to work?

chandlerc (approved):
I... don't see why not! And it gives us some CI coverage.

Issue #3860, opened 2024-04-04
What additional restrictions, if any, should be applied to functions which use positional parameters?


CJ-Johnson:
### Summary of issue:

In addition to the proposed restrictions (definition must be attached to the declaration; only one function in an enclosing scope can use positional parameters), an additional restriction was considered. That being, visibility of functions with positional parameters could be restricted to only non-public interfaces. What do the leads think about this candidate additional restriction?

See Lambdas proposal at https://github.com/carbon-language/carbon-lang/pull/3848

justzh:
See my other response for this. You need to write a trifactorial function inverse in order to close this issue.

Hira-Iftikhar-123:
**Summary of the problem with restriction on Public visibility:**
this is beneficial as the restriction could help enforce good design practices, encouraging developers to use more explicit and self-documenting parameter passing methods, such as named parameters or data structures 

**Proposed solution:**
The solution is to provide clear guidance and examples which could help developers to understand the reason of restriction and they can therefore, adhere the best practices of parameter passing.

Here are some alternative strategies or patterns that achieve the same goals without imposing such a strict limitation.

1- By encouraging the use of named parameters.
2- By promoting the use of data structures.
3- By providing clear guidelines and examples.
4- By considering gradual adoption.(by gradually applying restrictions.)

chandlerc:
My stance on this is that we should try out the minimal set of restrictions and see what happens. I feel like we can easily tighten these rules later if useful, and it seems somewhat interesting to leave the flexibility technically in the language.

Put differently, I wonder if whether to use this technique in a public API will end up being more of an HOA rule than a building code.

zygoloid:
**Leads decision:** Per @chandlerc's comment: do not try to proactively add restrictions for where functions with positional parameters can appear (eg, only in non-public interfaces or only in block scopes).

PR #3988, opened 2024-05-24
Parse parameters in name qualifiers.

zygoloid:
Parse the name of a declaration as a sequence of `NameQualifier`s -- which have a name, possibly parameters, and a trailing period -- followed by a name and possibly parameters. This prepares us for parsing declarations of members of generic classes and similar cases, but actually supporting such member redeclarations is left to a future change.

We previously required functions to have parameters, but no longer do, following the direction of #3848. Cases like namespaces that can't actually have parameters are now diagnosed in check instead of in parse.


toolchain/check/decl_name_stack.cpp
```diff
@@ -56,14 +58,24 @@ auto DeclNameStack::PushScopeAndStartName() -> void {
...
+      context_->node_stack().PopWithNodeIdIf<Parse::NodeKind::TuplePattern>();
+  auto [implicit_params_loc_id, implicit_params_id] =
+      context_->node_stack()
+          .PopWithNodeIdIf<Parse::NodeKind::ImplicitParamList>();
+  auto [loc_id, name_id] = context_->node_stack().PopNameWithNodeId();
+
+  ApplyNameQualifier(loc_id, name_id);
+
+  if (params_id || implicit_params_id) {
+    // TODO: Say which kind of declaraation we're parsing.
```
```suggestion
    // TODO: Say which kind of declaration we're parsing.
```

jonmeow (commented):
I'm not familiar with #3848, and I'm having trouble loading it because it has merge errors. Does this PR indicate that leads have in effect approved the proposal? Is there some reference for what design you're trying to implement?

zygoloid:
> I'm not familiar with #3848, and I'm having trouble loading it because it has merge errors.

I believe the PR is only intending to add [its p3848.md](https://github.com/carbon-language/carbon-lang/blob/ca4be05c9aa31ae0866ff491a32e1f12530cf6d5/proposals/p3848.md) file and the rest of the changes listed there are spurious.

> Does this PR indicate that leads have in effect approved the proposal?

We've not approved the proposal but we have consensus on the main points (including [#3860](https://github.com/carbon-language/carbon-lang/issues/3860), which is directly relevant here). But even without #3848, I think it'd still make sense to use the same parse rules (including parameters) for all names, and do any declaration-kind-specific checking in check. The main impact that #3848 has here is that this PR emits a TODO error for parameterless functions, rather than adding a proper diagnostic in check, because under #3848 they will be valid.

> Is there some reference for what design you're trying to implement?

The main purpose of this PR of this is to support declarations of members of parameterized types:

```
class C(T:! type) {
  fn F();
}
fn C(T:! type).F() {}
```

The most recent / specific proposal on that is [#3763](https://github.com/carbon-language/carbon-lang/blob/trunk/proposals/p3763.md#scope-differences). I think this syntax has been part of the intended design for generics for a while, but we don't seem to have any examples of this in /docs/design yet :(

toolchain/parse/node_kind.def
```diff
@@ -215,9 +215,24 @@ CARBON_PARSE_NODE_KIND_BRACKET(LibraryDecl, LibraryIntroducer,
 // LibrarySpecifier
 CARBON_PARSE_NODE_KIND_CHILD_COUNT(LibrarySpecifier, 1, Library)
 
+// Declaration names.
+//
+//   _repeated_: NameQualifier
+//   _external_: IdentifierName
+//   _optional_ _external_: ImplicitParamList
+//   _optional_ _external_: TuplePattern
+// _declaration name_
```
jonmeow, line 224: This comment is unclear to me: it offers a tree structure that, as far as I can tell, does not exist. Had you considered a non-tree-like representation?

toolchain/parse/state.def
```diff
@@ -227,63 +227,54 @@ CARBON_PARSE_STATE(CodeBlockFinish)
...
-// name ...        (variant is not Required)
-//     ^
+// name ...
+// ^~~~
 //   (state done)
 //
-// name ???        (variant is Required)
-//     ^
+//  ???
+// ^
```
jonmeow, line 249: What happened here?

toolchain/parse/state.def
```diff
@@ -227,63 +227,54 @@ CARBON_PARSE_STATE(CodeBlockFinish)
...
-                             Required)
+CARBON_PARSE_STATE(DeclNameAndParams)
 
 // Handles regular parameters such as `(...)` for the general declaration case.
 // Only used after implicit parameters.
 //
 // name [ ... ] ( ... )
 //             ^
 //   1. PatternListAsTuple
+//   1. DeclNameAndParamsAfterParams
```
```suggestion
//   2. DeclNameAndParamsAfterParams
```

toolchain/parse/state.def
```diff
@@ -227,63 +227,54 @@ CARBON_PARSE_STATE(CodeBlockFinish)
...
+// Handles regular parameters such as `(...)` for the general declaration case.
+// Only used after implicit parameters.
+//
+// name [ ... ] ( ... ) .
+//                      ^
+//   1. DeclNameAndParams
+//
+// name [ ... ] ( ... ) ...
+//                     ^
+//   (state done)
```
jonmeow, line 275: I'm confused, how are these two inputs different? The `.` versus `...` looks like a typo.

toolchain/parse/testdata/generics/impl/fail_out_of_line_member.carbon
```diff
@@ -84,13 +74,8 @@ fn C.(Self as Interface).F() {}
...
-// CHECK:STDOUT:         {kind: 'IdentifierName', text: '(', has_error: yes},
-// CHECK:STDOUT:       {kind: 'QualifiedName', text: '.', subtree_size: 3},
-// CHECK:STDOUT:         {kind: 'TuplePatternStart', text: '('},
-// CHECK:STDOUT:           {kind: 'IdentifierName', text: 'Self', has_error: yes},
-// CHECK:STDOUT:           {kind: 'InvalidParse', text: 'Self', has_error: yes},
-// CHECK:STDOUT:         {kind: 'BindingPattern', text: 'Self', has_error: yes, subtree_size: 3},
-// CHECK:STDOUT:       {kind: 'TuplePattern', text: ')', has_error: yes, subtree_size: 5},
-// CHECK:STDOUT:     {kind: 'FunctionDecl', text: '}', has_error: yes, subtree_size: 10},
+// CHECK:STDOUT:       {kind: 'NameQualifier', text: '.', subtree_size: 2},
+// CHECK:STDOUT:       {kind: 'InvalidParse', text: '(', has_error: yes},
```
jonmeow, line 78: This InvalidParse seems a little awkward since you document the IdentifierName as bracketing; how does the system distinguish what starts/ends where?

But also, are there other ways to cause this? What does something like `fn *.F() {}` parse like? Does that yield anything that can be identified as the "identifier"? (in particular, is there a way like this which creates an incorrectly bracketed NameQualifier?)

toolchain/parse/handle_decl_name_and_params.cpp
```diff
@@ -6,118 +6,85 @@
...
       token = state.token;
     }
-    context.emitter().Emit(token, ExpectedDeclName,
-                           context.tokens().GetKind(state.token));
+    if (state.token == *context.position()) {
+      context.emitter().Emit(token, ExpectedDeclNameAfterPeriod);
+    } else {
+      context.emitter().Emit(token, ExpectedDeclName,
+                             context.tokens().GetKind(state.token));
+    }
```
jonmeow: As you're restructuring this, can you keep the diagnostic definitions with where they're used?

toolchain/check/handle_function.cpp
```diff
@@ -211,8 +211,12 @@ static auto BuildFunctionDecl(Context& context,
     return_slot = SemIR::Function::ReturnSlot::Absent;
   }
 
-  SemIR::InstBlockId param_refs_id =
-      context.node_stack().Pop<Parse::NodeKind::TuplePattern>();
+  auto param_refs_id =
+      context.node_stack().PopIf<Parse::NodeKind::TuplePattern>();
+  if (!param_refs_id) {
+    context.TODO(node_id, "function with positional parameters");
```
jonmeow, line 217: I'm confused by this, isn't this an absence of positional params?

jonmeow:
> > I'm not familiar with #3848, and I'm having trouble loading it because it has merge errors.
> 
> I believe the PR is only intending to add [its p3848.md](https://github.com/carbon-language/carbon-lang/blob/ca4be05c9aa31ae0866ff491a32e1f12530cf6d5/proposals/p3848.md) file and the rest of the changes listed there are spurious.

Thanks -- to be sure, I really meant, it wasn't loading for me to try and find the proposal file. The link helped.

jonmeow (approved):
LGTM, although with some hesitance around bracketing.

toolchain/parse/node_kind.def
zygoloid: How would you like this to be presented?

Alternatively: how would you feel about removing all of these structure comments now that we have `typed_nodes.h` that describes the structure?

toolchain/parse/state.def
zygoloid: This is trying to say: if we're in the `DeclNameAndParams` state, and the first token isn't a name, that's an error and we're done. (We don't check there's a name next before pushing the state.)

toolchain/check/handle_function.cpp
zygoloid: Per #3848, ["positional parameters"](https://github.com/carbon-language/carbon-lang/blob/ca4be05c9aa31ae0866ff491a32e1f12530cf6d5/proposals/p3848.md#positional-parameters) are introduced by the absence of an explicit parameter list. Eg:
```
fn F => $0 < $1;
```
... has so-called positional parameters `$0` and `$1`, which are introduced because `F` doesn't have an explicit parameter list.

toolchain/parse/handle_decl_name_and_params.cpp
zygoloid: Done.

toolchain/parse/state.def
zygoloid: The first rule says: if the next token is a `.`, then consume it (because it's underlined), and continue in state `DeclNameAndParams`.

The second rule says: if the next token is anything else (per the comment at the start of the file: "`...` indicates a sequence of tokens handled by some other state"), then consume nothing and we're done.

toolchain/parse/testdata/generics/impl/fail_out_of_line_member.carbon
zygoloid: `IdentifierName` is bracketing for a `NameQualifier`, but this isn't part of a `NameQualifier`.

For `fn *.F() {}`, we fail to parse a declaration name at the `*` and stop name parsing there (this is the `???` rule you asked about in `state.def`). This eventually leads to the overall function parsing failing, because `*` isn't a valid start for a return type or a function body, and we skip to the end of the declaration. That results in:
```console
/dev/stdin:1:4: ERROR: `fn` introducer should be followed by a name.
fn *.F() {}
   ^
```
and this parse tree:
```yaml
- filename: /dev/stdin
  parse_tree: [
    {kind: 'FileStart', text: ''},
      {kind: 'FunctionIntroducer', text: 'fn'},
      {kind: 'InvalidParse', text: '*', has_error: yes},
    {kind: 'FunctionDecl', text: '}', has_error: yes, subtree_size: 3},
    {kind: 'FileEnd', text: ''},
  ]
```

toolchain/parse/state.def
jonmeow: Okay, but maybe the comment could be clearer? Maybe this is a problem with using `...` without any extra annotation, it makes `.` look like a typo.

toolchain/parse/node_kind.def
jonmeow: Removing all these comments seems fair.

toolchain/parse/state.def
jonmeow: Maybe this comment could be made clearer?

PR #3989, opened 2024-05-25
Start building and bundling LLD for linking.

chandlerc:
This is the first, minor steps in sketching out an installation tree for the driver, and putting LLD into it in an LLVM-compatible location. It then teaches the Clang runner to configure this private LLVM install in the driver, and the link command to use LLD. With that, we use our LLD rather than the system linker.

There are a couple of TODOs here that I plan to address in direct follow-ups. First, I want to extract all the installation tree management out from everything else and into a centralized place where we can document the structure and provide APIs for navigating it. Next I want to move the prelude over to use that infrastructure as well as LLD. I left these as TODOs as I wanted to make sure I had the core logic of finding LLD working first, and would prefer to checkpoint that rather than re-doing it locally.

### DIFF
--- a/toolchain/driver/BUILD
+++ b/toolchain/driver/BUILD
@@ -2,6 +2,8 @@
 # Exceptions. See /LICENSE for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+load("@bazel_skylib//rules:copy_file.bzl", "copy_file")
+load("@llvm-project//llvm:binary_alias.bzl", "binary_alias")
 load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test")
 load("//bazel/cc_toolchains:defs.bzl", "cc_env")
 load("//testing/fuzzing:rules.bzl", "cc_fuzz_test")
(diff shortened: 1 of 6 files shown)


jonmeow (commented):
Would it be reasonable to drop the `ld` reference in `bazel/carbon_rules/defs.bzl` in order to test that the built `lld` is in use?

toolchain/driver/BUILD
```diff
@@ -13,10 +15,66 @@ filegroup(
     data = glob(["testdata/**/*.carbon"]),
 )
 
+# Build rules to construct the install data tree for the Carbon toolchain.
+#
+# The directory `install` should mirror a Unix-y [FHS] style installation of
```
jonmeow, line 20: If you're going to write "FHS", please document what it stands for; it's an abbreviation I need to look up to understand.

Also, why square brackets instead of parentheses?

toolchain/driver/BUILD
```diff
@@ -13,10 +15,66 @@ filegroup(
...
+# Carbon under some prefix (`/usr`, `/usr/local`, or possibly a home directory
+# or a custom prefix). When installed, user facing binaries, which we largely
+# expect to be solely the `carbon` driver will be at `install/bin/carbon`. We
```
jonmeow, line 23: Can you elaborate how `carbon` is being put at `install/bin/carbon`? I'm not seeing changes here that would apply that.

toolchain/driver/BUILD
jonmeow, line 20: Optional suggestion, but perhaps `root` or `base` instead of `install`, given you seem to be setting this up for use within `bazel` which doesn't strictly "install"?

toolchain/driver/BUILD
```diff
@@ -13,10 +15,66 @@ filegroup(
...
+# collect data, private executables, and other internals used by the toolchain
+# in `install/lib/carbon`. While using a `lib` subdirectory isn't a perfect fit
+# for the FHS, having a single location where private data is collected is
+# significantly superior to spreading them across the system. This also matches
+# similar patterns used by Clang itself and several other language toolchains
+# and standard libraries.
```
jonmeow, line 29: This comment is seeming kind of long. Perhaps some of this is more appropriate for the PR description? Do we need all of this argument about the best approach going forward?

toolchain/driver/BUILD
```diff
@@ -13,10 +15,66 @@ filegroup(
     data = glob(["testdata/**/*.carbon"]),
 )
 
+# Build rules to construct the install data tree for the Carbon toolchain.
```
jonmeow, line 18: This seems to be a free-standing comment, tied to a couple rules in this file -- but it's not clear where those rules end, and driver-related rules start. Putting it at the top of the file like this implies it applies to _all_ rules below, particularly given the length. Should this be in its own BUILD file so that it's clearer what's for directly structure?

toolchain/driver/BUILD
```diff
@@ -13,10 +15,66 @@ filegroup(
...
+# and standard libraries.
+#
+# Code that needs data files in the toolchain should add them to paths below
+# `install/lib/carbon` here and use data dependencies on these paths so that we
+# are always testing against a realistic layout.
+
+# Copy Clang and LLVM toolchain files into a synthetic LLVM installation under
+# `install/lib/carbon/llvm` so that parts of Clang that expect to find an LLVM
+# installation at relative paths work correctly without exposing these as
+# user-facing components.
```
jonmeow, line 38: What do you mean by "without exposing these as user-facing components"?

toolchain/driver/BUILD
jonmeow: I see you move this into a separate BUILD in #3993. How much extra cost would it be to do that in this PR?

chandlerc:
FYI, going to just merge this into #3993. I didn't realize until after I had finished that just how much it cleans everything up, and as Jon is pointing out, it's probably much easier to review in that PR's form. Sorry for the noise here. I'll try to go through the comments and address all of the ones that are still relevant.

PR #3992, opened 2024-05-26
Allow GoogleTest tests to locate their executable path.

chandlerc:
This also factors out the code for doing this location from the driver to a tiny helper library.

The motivation for this is letting tests locate data files like the prelude or other files needed by the toolchain more easily. A subsequent PR will use this heavily in the Clang runner and related logic for example.

### DIFF
--- a/common/exe_path_test.cpp
+++ b/common/exe_path_test.cpp
@@ -33,6 +33,10 @@ TEST(ExePath, File) {
 }
 
 TEST(ExePath, PathLookup) {
+  // TODO: This is not likely to work well on Windows (outside of WSL). But some
+  // of that may be hidden by Bazel's test environment. Regardless, we should
+  // revisit this when we have good coverage of Windows build with something
+  // appropriate for that platform.
   std::string exe_path = FindExecutablePath("bash");
   EXPECT_NE(exe_path, "bash");
   EXPECT_TRUE(llvm::sys::fs::exists(exe_path));
--- a/common/exe_path_test.cpp
+++ b/common/exe_path_test.cpp
@@ -1,0 +1,43 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "common/exe_path.h"
+
+#include <gtest/gtest.h>
+
+#include "llvm/ADT/SmallString.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/Path.h"
+
+namespace Carbon {
+namespace {
+
+TEST(ExePath, FailureFallback) {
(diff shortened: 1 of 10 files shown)


testing/base/gtest_main.h
```diff
@@ -0,0 +1,20 @@
...
+
+#include "llvm/ADT/StringRef.h"
+
+// When using the Carbon `main` function for GoogleTest, we export some extra
+// information about the test binary that can be accessed with this header.
+
+namespace Carbon::Testing {
+
+// The executable path of the test binary.
+auto TestExePath() -> llvm::StringRef;
```
jonmeow: You might consider something like `GetExePath` or `GetTextExePath`. `TestExePath` sounds more like it's testing the exe path, not returning a value.

testing/base/gtest_main.cpp
```diff
@@ -2,11 +2,34 @@
...
 #include <gtest/gtest.h>
 
+#include <string>
+
+#include "common/check.h"
+#include "common/exe_path.h"
 #include "common/init_llvm.h"
 
+static bool after_main = false;
+static llvm::StringRef exe_path;
```
jonmeow, line 16: Minor question, but had you considered:

```suggestion
static std::string* exe_path = nullptr;
```
 
i.e., rather than having a separate bool for tracking whether exe_path is initialized, use the pointer initialization.

common/exe_path_test.cpp
```diff
@@ -0,0 +1,43 @@
...
+  std::error_code ec = llvm::sys::fs::openFileForWrite(path, fd);
+  ASSERT_TRUE(!ec) << "Error code: " << ec;
+  close(fd);
+
+  std::string exe_path = FindExecutablePath(path);
+  EXPECT_EQ(path, exe_path);
+}
+
+TEST(ExePath, PathLookup) {
+  std::string exe_path = FindExecutablePath("bash");
```
jonmeow, line 40: I'm a little wary of this test, will it work on Windows?

jonmeow (approved):
Generally LG, a few small comments but feel free to merge.

common/exe_path_test.cpp
chandlerc: Probably not well outside of WSL.

But I'd like us to do *some* test using PATH. I've added a TODO to add a different but appropriate command name for Windows when we have testing there?

If you have other ideas, happy to switch to them.

testing/base/gtest_main.cpp
chandlerc: I didn't think of it, but I think I prefer the bool. I wanted to leave open the chance that we end up switching back to `llvm::SmallString` for the storage. And more generally, just avoid coupling the type of the global to the type of the storage in `main`.

testing/base/gtest_main.h
chandlerc: Good call, switched to `GetTestExePath`.

chandlerc (commented):
Thanks for review! Merging with the suggested name. Happy to follow up with any improvements I can make to the test though.

PR #3993, opened 2024-05-27
Add utilities for managing a toolchain install, and install and use LLD.

chandlerc:
The install directory contains the BUILD logic for creating an
installable tree of data files and executables for the toolchain, and
a library to facilitate toolchain code accessing the paths to their data
within this installation.

Then adds an installation of LLD in a synthetic LLVM installation, and
teaches the Clang runner to configure this and use it for linking
instead of the system linker.

Currently, the install paths only really manage access to the LLVM
binaries installed and used by the Clang runner for linking, but
eventually other data files like the prelude and runtime libraries will
be fleshed out as well. There are TODOs for moving more things over here
such as the prelude.

One interesting aspect of this is where to put helpers like parts of
LLVM in our install. This PR suggests nesting those files under
`lib/carbon`. While using a `lib` subdirectory isn't a perfect fit for
the FHS (Filesystem Hierarchy Standard), having a single location where
private data is collected is significantly superior to spreading them
across the system. This also matches similar patterns used by Clang
itself and several other language toolchains and standard libraries.

The install directory also provides a natural place for us to build out
packaging rules to create installable packages in various formats, but
that remains future work.

### DIFF
--- a/toolchain/install/BUILD
+++ b/toolchain/install/BUILD
@@ -1,0 +1,152 @@
+# Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+# Exceptions. See /LICENSE for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+load("@llvm-project//llvm:binary_alias.bzl", "binary_alias")
+load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test")
+load("symlink_filegroup.bzl", "symlink_filegroup")
+
+package(default_visibility = ["//visibility:public"])
+
+# Build rules supporting the install data tree for the Carbon toolchain.
+#
+# The directory `prefix_root` should mirror a Unix-y FHS (Filesystem Hierarchy
+# Standard) style install of Carbon under some prefix (`/usr`, `/usr/local`, or
+# possibly a home directory or a custom prefix). When installed, user facing
+# binaries, which we largely expect to be solely the `carbon` driver will be at
+# `perfix_root/bin/carbon`. We collect data, private executables, and other
+# internals used by the toolchain in `prefix_root/lib/carbon`.
+#
+# Code that needs data files in the toolchain should add them to paths below
+# `prefix_root/lib/carbon` here and use data dependencies on these paths so that we
+# are always twesting against a realistic layout.
+
+# Copy Clang and LLVM toolchain files into a synthetic LLVM installation under
+# `prefix_root/lib/carbon/llvm` so that parts of Clang that expect to find an LLVM
+# installation at relative paths work correctly without exposing these in an
+# installed 'bin' directory where it might get added to a user's PATH.
+binary_alias(
+    name = "prefix_root/lib/carbon/llvm/bin/lld",
+    binary = "@llvm-project//lld:lld",
+)
+
+lld_bin_names = [
+    "ld.lld",
+    "ld64.lld",
+    "lld-link",
+    "wasm-ld",
+]
+
+[
+    binary_alias(
+        name = "prefix_root/lib/carbon/llvm/bin/" + bin_name,
+        binary = "@llvm-project//lld:lld",
+    )
+    for bin_name in lld_bin_names
+]
+
+filegroup(
+    name = "llvm_link_data",
+    srcs = [
+        "prefix_root/lib/carbon/llvm/bin/lld",
+    ] + [
+        "prefix_root/lib/carbon/llvm/bin/" + bin_name
+        for bin_name in lld_bin_names
+    ],
+)
+
+# All of the install data except for the user-facing binaries. This is typically
+# a reasonable data dependency for libraries and the user-facing binaries
+# without creating a cycle.
+filegroup(
+    name = "install_lib_data",
+    srcs = [
+        ":llvm_link_data",
+    ],
+)
+
+binary_alias(
+    name = "prefix_root/bin/carbon",
+    binary = "//toolchain/driver:carbon",
+)
+
+filegroup(
+    name = "install_data",
+    srcs = [
+        "prefix_root/bin/carbon",
+        ":install_lib_data",
+    ],
+)
+
+# A library for computing install paths for the toolchain. Note that this
+# library does *not* include the data itself, as that would form a dependency
+# cycle. Each part of the toolchain should add the narrow data file groups to
+# their data dependencies, and then use this library to locate them.
+cc_library(
+    name = "install_paths",
+    srcs = ["install_paths.cpp"],
+    hdrs = ["install_paths.h"],
+    deps = [
+        "//common:check",
+        "//common:command_line",
+        "//common:ostream",
+        "//common:vlog",
+        "@llvm-project//llvm:Support",
+    ],
+)
+
+# Build up some trees of data to use in testing our install detection.
+symlink_filegroup(
+    name = "test_installed_data",
+    testonly = 1,
+    srcs = ["//toolchain/install:install_data"],
+    out_prefix = "test_installed_root/",
+    remove_prefix = "toolchain/install/prefix_root/",
+)
+
+cc_binary(
+    name = "test_binary",
+    testonly = 1,
+    srcs = ["test_binary.cpp"],
+)
+
+binary_alias(
+    name = "test_direct_exec_root/test_binary",
+    testonly = 1,
+    binary = ":test_binary",
+)
+
+symlink_filegroup(
+    name = "test_direct_exec_runfiles",
+    testonly = 1,
+    srcs = [":install_data"],
+    out_prefix = "test_direct_exec_root/test_binary.runfiles/_main/",
+)
+
+filegroup(
+    name = "test_direct_exec_data",
+    testonly = 1,
+    srcs = [
+        "test_direct_exec_root/test_binary",
+        ":test_direct_exec_runfiles",
+    ],
+)
+
+cc_test(
+    name = "install_paths_test",
+    size = "small",
+    srcs = ["install_paths_test.cpp"],
+    data = [
+        ":install_data",
+        ":test_direct_exec_data",
+        ":test_installed_data",
+    ],
+    deps = [
+        ":install_paths",
+        "//common:check",
+        "//common:ostream",
+        "//testing/base:gtest_main",
+        "@googletest//:gtest",
+        "@llvm-project//llvm:Support",
+    ],
+)
--- a/toolchain/install/BUILD
+++ b/toolchain/install/BUILD
@@ -2,6 +2,7 @@
 # Exceptions. See /LICENSE for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
 load("@llvm-project//llvm:binary_alias.bzl", "binary_alias")
 load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test")
 load("symlink_filegroup.bzl", "symlink_filegroup")
@@ -10,16 +11,22 @@ package(default_visibility = ["//visibility:public"])
 
 # Build rules supporting the install data tree for the Carbon toolchain.
 #
-# The directory `prefix_root` should mirror a Unix-y FHS (Filesystem Hierarchy
-# Standard) style install of Carbon under some prefix (`/usr`, `/usr/local`, or
-# possibly a home directory or a custom prefix). When installed, user facing
-# binaries, which we largely expect to be solely the `carbon` driver will be at
-# `perfix_root/bin/carbon`. We collect data, private executables, and other
-# internals used by the toolchain in `prefix_root/lib/carbon`.
-#
-# Code that needs data files in the toolchain should add them to paths below
-# `prefix_root/lib/carbon` here and use data dependencies on these paths so that we
-# are always twesting against a realistic layout.
+# This populates a synthetic Carbon toolchain installation under the
+# `prefix_root` directory. For details on its layout, see `install_paths.h`.
+
+# A marker of a valid Carbon install. All filegroups in the install should
+# include this one.
+write_file(
+    name = "install_marker",
+    out = "prefix_root/lib/carbon/carbon_install.txt",
+    content = [
+        "// Part of the Carbon Language project, under the Apache License v2.0 with LLVM",
+        "// Exceptions. See /LICENSE for license information.",
+        "// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception",
+        "",
+        "This marks a valid Carbon install tree.",
+    ],
+)
 
 # Copy Clang and LLVM toolchain files into a synthetic LLVM installation under
 # `prefix_root/lib/carbon/llvm` so that parts of Clang that expect to find an LLVM
@@ -49,6 +56,7 @@ filegroup(
     name = "llvm_link_data",
     srcs = [
         "prefix_root/lib/carbon/llvm/bin/lld",
+        ":install_marker",
     ] + [
         "prefix_root/lib/carbon/llvm/bin/" + bin_name
         for bin_name in lld_bin_names
@@ -61,6 +69,7 @@ filegroup(
 filegroup(
     name = "install_lib_data",
     srcs = [
+        ":install_marker",
         ":llvm_link_data",
     ],
 )
@@ -75,6 +84,7 @@ filegroup(
     srcs = [
         "prefix_root/bin/carbon",
         ":install_lib_data",
+        ":install_marker",
     ],
 )
 
@@ -87,10 +97,8 @@ cc_library(
     srcs = ["install_paths.cpp"],
     hdrs = ["install_paths.h"],
     deps = [
-        "//common:check",
-        "//common:command_line",
-        "//common:ostream",
-        "//common:vlog",
+        "//common:error",
+        "@bazel_tools//tools/cpp/runfiles",
         "@llvm-project//llvm:Support",
     ],
 )
@@ -108,28 +116,7 @@ cc_binary(
     name = "test_binary",
     testonly = 1,
     srcs = ["test_binary.cpp"],
-)
-
-binary_alias(
-    name = "test_direct_exec_root/test_binary",
-    testonly = 1,
-    binary = ":test_binary",
-)
-
-symlink_filegroup(
-    name = "test_direct_exec_runfiles",
-    testonly = 1,
-    srcs = [":install_data"],
-    out_prefix = "test_direct_exec_root/test_binary.runfiles/_main/",
-)
-
-filegroup(
-    name = "test_direct_exec_data",
-    testonly = 1,
-    srcs = [
-        "test_direct_exec_root/test_binary",
-        ":test_direct_exec_runfiles",
-    ],
+    data = [":install_data"],
 )
 
 cc_test(
@@ -138,7 +125,7 @@ cc_test(
     srcs = ["install_paths_test.cpp"],
     data = [
         ":install_data",
-        ":test_direct_exec_data",
+        ":test_binary",
         ":test_installed_data",
     ],
     deps = [
@@ -146,6 +133,7 @@ cc_test(
         "//common:check",
         "//common:ostream",
         "//testing/base:gtest_main",
+        "@bazel_tools//tools/cpp/runfiles",
         "@googletest//:gtest",
         "@llvm-project//llvm:Support",
     ],
--- a/toolchain/install/BUILD
+++ b/toolchain/install/BUILD
@@ -97,6 +97,7 @@ cc_library(
     srcs = ["install_paths.cpp"],
     hdrs = ["install_paths.h"],
     deps = [
+        "//common:check",
         "//common:error",
         "@bazel_tools//tools/cpp/runfiles",
         "@llvm-project//llvm:Support",
--- a/toolchain/install/install_paths.cpp
+++ b/toolchain/install/install_paths.cpp
@@ -6,6 +6,7 @@
 
 #include <memory>
 
+#include "common/check.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/Path.h"
@@ -47,29 +48,25 @@ auto InstallPaths::MakeExeRelative(llvm::StringRef exe_path) -> InstallPaths {
 
 auto InstallPaths::MakeForBazelRunfiles(llvm::StringRef exe_path)
     -> InstallPaths {
-  InstallPaths paths;
-
   using bazel::tools::cpp::runfiles::Runfiles;
   std::string runtimes_error;
   std::unique_ptr<Runfiles> runfiles(
       Runfiles::Create(exe_path.str(), &runtimes_error));
-  if (runfiles == nullptr) {
-    paths.SetError(llvm::Twine("Failed to find runtimes tree: ") +
-                   runtimes_error);
-    return paths;
-  }
+  CARBON_CHECK(runfiles != nullptr)
+      << "Failed to find runtimes tree: " << runtimes_error;
 
   std::string relative_marker_path = (PrefixRoot.str() + MarkerPath).str();
   std::string runtimes_marker_path = runfiles->Rlocation(relative_marker_path);
 
   // Start from the marker, remove that filename, and walk up to find the
   // install prefix.
-  paths = InstallPaths(runtimes_marker_path);
+  InstallPaths paths(runtimes_marker_path);
   llvm::sys::path::remove_filename(paths.prefix_);
   llvm::sys::path::append(paths.prefix_, llvm::sys::path::Style::posix,
                           "../../");
 
   paths.CheckMarkerFile();
+  CARBON_CHECK(!paths.error()) << *paths.error();
   return paths;
 }
 
--- a/toolchain/install/install_paths.cpp
+++ b/toolchain/install/install_paths.cpp
@@ -1,0 +1,108 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "toolchain/install/install_paths.h"
+
+#include "common/check.h"
+#include "common/vlog.h"
+#include "llvm/ADT/StringRef.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/Path.h"
+
+namespace Carbon {
+
+// The suffix added to an executable name in Bazel's output to find its
+// runfiles tree.
+static constexpr llvm::StringLiteral RunfilesSuffix = ".runfiles/";
+
+// The workspace name to get to the root of our Bazel tree.
+static constexpr llvm::StringLiteral Workspace = "_main/";
+
+// The location within our Bazel output tree of the prefix_root.
+static constexpr llvm::StringLiteral PrefixRoot =
+    "toolchain/install/prefix_root/";
+
+InstallPaths::InstallPaths(llvm::StringRef exe_path,
+                           llvm::raw_ostream* vlog_stream)
+    : vlog_stream_(vlog_stream) {
+  // Map from the executable path from the executable path to an install prefix
+  // path.
+
+  // TODO: Detect a Windows executable path and use custom logic to map to the
+  // correct install prefix for that platform.
+
+  // First, see if the binary is in a `bin` directory. If so, remove it
+  // assuming this is a FHS-like install prefix. We start with this pattern both
+  // because we expect it to be the "production" pattern and so useful to
+  // involve a minimal number of filesystem accesses, and because even in
+  // testing or development builds, when we _can_ use this pattern, we prefer it
+  // to match the behavior of genuine installations as closely as possible.
+  prefix_.assign(exe_path);
+  llvm::sys::path::remove_filename(prefix_);
+  if (auto rbegin_it = llvm::sys::path::rbegin(prefix_),
+      rend_it = llvm::sys::path::rend(prefix_);
+      rbegin_it != rend_it) {
+    // Step over a trailing `.` signifying the current directory. Whether this
+    // occurs for paths is ambiguous between the documentation and the observed
+    // behavior, so this code works to be resilient.
+    if (*rbegin_it == ".") {
+      ++rbegin_it;
+    }
+    if (rbegin_it != rend_it && *rbegin_it == "bin") {
+      // Rather than trying to remove the `bin` directory, walk upwards by
+      // appending `..`. We also assume POSIX style at this point.
+      llvm::sys::path::append(prefix_, llvm::sys::path::Style::posix, "../");
+      return;
+    }
+  }
+
+  // Next check for a Bazel runfiles directory relative to the binary name. This
+  // is used for `genrule` style execution within Bazel and for direct execution
+  // of tests or other just-built binaries.
+  prefix_.assign(exe_path);
+  prefix_.append(RunfilesSuffix);
+  if (llvm::sys::fs::is_directory(prefix_)) {
+    llvm::sys::path::append(prefix_, llvm::sys::path::Style::posix, Workspace,
+                            PrefixRoot);
+    CARBON_CHECK(llvm::sys::fs::is_directory(prefix_))
+        << "Found runfiles tree but not an install prefix directory!";
+    return;
+  }
+
+  // Lastly, check for a test runtimes tree environment variable. This will
+  // match when running inside a Bazel test's runfiles tree. We check this last
+  // of all in part because that allows testing of all of the prior patterns.
+  prefix_.assign(llvm::StringRef(getenv("TEST_SRCDIR")));
+  if (llvm::sys::fs::is_directory(prefix_)) {
+    llvm::sys::path::append(prefix_, llvm::sys::path::Style::posix, Workspace,
+                            PrefixRoot);
+    CARBON_CHECK(llvm::sys::fs::is_directory(prefix_))
+        << "Found runfiles tree but not an install prefix directory!";
+    return;
+  }
+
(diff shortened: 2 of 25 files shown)


toolchain/install/installation.h
```diff
@@ -0,0 +1,74 @@
...
+// our installation to behave in a similar and compatible way.
+//
+// We detect the installation prefix by looking at the path of the executable.
+// This supports development binaries like tests, as well as any installed
+// user-facing binaries.
+//
+// Within this prefix, we expect a [FHS]-like hierarchy on Unix-y platforms:
+//
+// - `...prefix/bin/carbon` - the main CLI driver.
+// - `...prefix/lib/carbon` - private data & binaries used by the toolchain.
```
jonmeow: Should this be `prefix_root` to match the other uses?

toolchain/install/installation.h
```diff
@@ -0,0 +1,74 @@
...
+#include "llvm/ADT/StringRef.h"
+
+namespace Carbon {
+
+// Locates the toolchain installation and provides paths to various components.
+//
+// The Carbon toolchain expects to be installed into some install prefix. For
+// example, this is expected to be similar to the CMake install prefix, such as:
+//
+// - `C:/Program Files/Carbon` or similar on Windows. TODO: Confirm this.
```
jonmeow: Regarding the TODO, doesn't https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html indicate this path?

toolchain/install/installation.h
```diff
@@ -0,0 +1,74 @@
...
+#include "common/ostream.h"
+#include "llvm/ADT/SmallString.h"
+#include "llvm/ADT/StringRef.h"
+
+namespace Carbon {
+
+// Locates the toolchain installation and provides paths to various components.
+//
+// The Carbon toolchain expects to be installed into some install prefix. For
+// example, this is expected to be similar to the CMake install prefix, such as:
```
jonmeow: You should probably say one of "for example" or "such as", rather than both.

toolchain/install/installation.h
```diff
@@ -0,0 +1,74 @@
...
+// Within this prefix, we expect a [FHS]-like hierarchy on Unix-y platforms:
+//
+// - `...prefix/bin/carbon` - the main CLI driver.
+// - `...prefix/lib/carbon` - private data & binaries used by the toolchain.
+//
+// [FHS]: https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard
+//
+// TODO: Need to check the installation structure of LLVM on Windows and figure
+// out what Carbon's should be within a Windows prefix and how much of the
+// structure we can share with the Unix-y layout of the prefix.
```
jonmeow: FWIW https://github.com/llvm/llvm-project/releases/download/llvmorg-18.1.6/clang+llvm-18.1.6-x86_64-pc-windows-msvc.tar.xz looks like it has the same `bin` `lib` (and `share` `include`) split.

toolchain/install/installation.h
```diff
@@ -0,0 +1,74 @@
...
+  // Query for a specific path within the installation. Note that we want to
+  // abstract away any platform differences in the installation layout, and so
+  // while there are high-level methods here, those should primarily be used for
+  // logging or debugging. When a specific part of the installation is needed, a
+  // dedicated accessor should be added that computes the path for that
+  // component.
+  //
+  // When this class provides storage for a particular path, a `StringRef` is
+  // provided. When these transparently compute the requested path, they return
+  // a `std::string` to avoid lifetime issues.
```
jonmeow: Is this a comment on install_prefix, or about all methods? If about all methods, should it be moved up to be part of the class comment?

toolchain/install/installation.h
```diff
@@ -0,0 +1,74 @@
...
+ public:
+  // Provide the current executable's path to detect the correct installation
+  // prefix path. Optionally provide a verbose logging stream where details and
+  // any issues computing the paths will be written.
+  explicit Installation(llvm::StringRef exe_path,
+                        llvm::raw_ostream* vlog_stream = nullptr);
+
+  // Query for a specific path within the installation. Note that we want to
+  // abstract away any platform differences in the installation layout, and so
+  // while there are high-level methods here, those should primarily be used for
```
jonmeow: What's a "high-level method" in this context? I'm unsure looking at these methods what it might be referring to.

toolchain/install/installation.h
```diff
@@ -0,0 +1,74 @@
...
+// We detect the installation prefix by looking at the path of the executable.
+// This supports development binaries like tests, as well as any installed
+// user-facing binaries.
+//
+// Within this prefix, we expect a [FHS]-like hierarchy on Unix-y platforms:
+//
+// - `...prefix/bin/carbon` - the main CLI driver.
+// - `...prefix/lib/carbon` - private data & binaries used by the toolchain.
+//
+// [FHS]: https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard
```
jonmeow: Since this is a C++ comment rather than markdown, what do you think of using a parenthetical to explain the abbreviation where it's first used? I'm not even sure a link is needed, just `FHS (Filesystem Hierarchy Standard)` seems like it'd be helpful.

toolchain/install/installation.h
```diff
@@ -0,0 +1,74 @@
...
+  // provided. When these transparently compute the requested path, they return
+  // a `std::string` to avoid lifetime issues.
+
+  auto install_prefix() const -> llvm::StringRef { return install_prefix_; }
+
+  auto driver_path() const -> std::string;
+  auto llvm_install_bin_path() const -> std::string;
+
+ private:
+  llvm::SmallString<256> install_prefix_;
```
jonmeow: What's the relation between `prefix_root` and `install_prefix`? Given the different names, I'm not sure these have the same correlation that I expect -- maybe the value should have a comment?

toolchain/install/installation.cpp
```diff
@@ -0,0 +1,98 @@
...
+    }
+  }
+
+  // Next, check for a Bazel runfiles tree. This will match for development
+  // builds and direct execution of tests or other just-built binaries.
+  install_prefix_.assign(exe_path);
+  install_prefix_.append(".runfiles");
+  if (llvm::sys::fs::is_directory(install_prefix_)) {
+    llvm::sys::path::append(install_prefix_, llvm::sys::path::Style::posix,
+                            "_main/toolchain/install/prefix_root/");
```
jonmeow: Can you abstract out the `_main` references a bit further, or perhaps consolidate with the reference in driver_main.cpp? I'd like to avoid a situation where we need to chase down different ways of referring to runfiles in order to compile in different environments. In particular, if you have the explicit string ".runfiles/_main" instead of just "_main", that would be very helpful.

toolchain/install/installation.cpp
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  // Lastly, check for a test runtimes tree environment variable. This will
+  // match when running under Bazel and thus inside the runfiles tree. We check
+  // this last of all in part because that allows testing of all of the prior
+  // patterns.
+  install_prefix_.assign(llvm::StringRef(getenv("TEST_SRCDIR")));
+  if (llvm::sys::fs::is_directory(install_prefix_)) {
+    llvm::sys::path::append(install_prefix_, llvm::sys::path::Style::posix,
+                            "_main/toolchain/install/prefix_root/");
```
jonmeow: Is this necessary, given the prior check? This again comes back to the issue of supporting multiple environments -- hardcoding the runfiles paths might be somewhat necessary if we can't use the [`GetRunfiles`](https://github.com/carbon-language/carbon-lang/blob/fd9783462438f95097c60f133033f702597ac261/explorer/fuzzing/fuzzer_util.cpp#L19) API for this (in order to maintain license), although perhaps unit tests should just use that?

toolchain/install/installation.h
```diff
@@ -0,0 +1,74 @@
...
+//
+// - `...prefix/bin/carbon` - the main CLI driver.
+// - `...prefix/lib/carbon` - private data & binaries used by the toolchain.
+//
+// [FHS]: https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard
+//
+// TODO: Need to check the installation structure of LLVM on Windows and figure
+// out what Carbon's should be within a Windows prefix and how much of the
+// structure we can share with the Unix-y layout of the prefix.
+class Installation {
```
jonmeow: The naming implies to me something that performs an install. This looks like it's returning the paths to distributed tools, maybe `ToolPaths`, `ToolDirectories`, `PathHierarchy` or something like that?

toolchain/install/installation_test.cpp
```diff
@@ -0,0 +1,126 @@
...
+      << "Test binary path: " << test_binary_path;
+  test_exe_paths.push_back(test_binary_path);
+
+  // If running under Bazel and we have `TEST_SRCDIR`, also check that it works
+  // by using a nonsense executable path.
+  if (getenv("TEST_SRCDIR")) {
+    test_exe_paths.push_back("test_exe");
+  }
+
+  for (llvm::StringRef exe_path : test_exe_paths) {
```
jonmeow: Could this be a helper function rather than a `for` loop, removing test_exe_paths, particularly so that stack traces point back to the correct source? (perhaps also changing the `if`)

toolchain/install/installation_test.cpp
```diff
@@ -0,0 +1,126 @@
...
+using ::testing::StrNe;
+
+// Compute the test runfiles path, either using Bazel's `TEST_SRCDIR`
+// environment variable or the name of the test executable. Trying both of these
+// allows tests to be run directly out of the Bazel build tree, for example by a
+// debugger, and still find their runfiles.
+//
+// TODO: Extract this to a testing helper library and use it more broadly.
+static auto FindTestRunfiles() -> std::string {
+  std::string runfiles_path = getenv("TEST_SRCDIR");
```
jonmeow: Had you considered using [GetRunfiles](https://github.com/carbon-language/carbon-lang/blob/fd9783462438f95097c60f133033f702597ac261/explorer/fuzzing/fuzzer_util.cpp#L19)?

toolchain/install/test_binary.cpp
```diff
@@ -0,0 +1,5 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+auto main() -> int {}
```
jonmeow: I understand this is conforming, but maybe `return 0;`?

toolchain/install/installation.h
chandlerc: Yes, but it seemed good to confirm it at some point rather than just relying on CMake's docs.

toolchain/install/installation.h
chandlerc: I wasn't trying to exactly match, but that does make sense.

toolchain/install/installation.h
chandlerc: Sure, switched to a more casual style. I generally write comments as-if in Markdown, but didn't seem to be helping here.

toolchain/install/installation.h
chandlerc: Yep. There is also a separate an installer executable that I want to check the layout for in case it differs.

toolchain/install/installation.h
chandlerc: Yeah, not a good description. I've reworded the comment -- is that more clear?

toolchain/install/installation.h
chandlerc: A comment for all the methods. And it was confusingly reworded. Moved to the class comment and tried to improve its wording.

toolchain/install/installation.h
chandlerc: Added a comment to the accessor to specifically try to clarify.

`prefix_root` is the name of the directory in the build tree. The `install_prefix` is the detected prefix for a *specific* install. When that install is in the build tree, that will be the `prefix_root` directory. Hopefully the comment helps.

(And the method will get renamed to `prefix` when this is called `InstallPaths`.)

toolchain/install/installation.h
chandlerc: Eh, re-reading this, the comment doesn't make sense here as this is just re-illustrating what the CMake docs list. Removed the TODO.

toolchain/install/installation.h
chandlerc: Doh, yeah. Removed the "such as".

toolchain/install/installation_test.cpp
chandlerc: I either wasn't aware of that, or completely forgot about it. Thanks for bringing it up!

I tried switching the test code over to this, and it seems a bit awkward. It wants a *specific* runtime file and will give out a specific path to that runtime file. And here, we we somewhat want the *root* of the runfiles tree, in part because we're specifically testing our own implementation of navigating that tree. And I don't see a good way to avoid having our own implementation -- license as you mentioned elsewhere and just the complexity of our version seems much lower.

So I've left this as-is, but I'm not really sure what would be best here so maybe we should chat. I like the idea of using the Bazel library at least in the test code to help check things.

I guess, I could write a test that compares the paths we get from the `InstallPaths` library with the paths produced by the Bazel runfiles code for *specific* files, and that might be a nice cross check? What do you think?

toolchain/install/test_binary.cpp
chandlerc: I hadn't mostly because we don't even run the binary. =] But sure.

toolchain/install/installation_test.cpp
chandlerc: Yeah, that's a much better structure. Thanks for the suggestion, should be done now, along with a bunch of simplifications (and a fixture) to try to tie it all together more cleanly.

toolchain/install/installation.cpp
chandlerc: Regarding `driver_main.cpp`, that instance will go away in #3995 so we shouldn't end up proliferating this too much.

In terms of abstracting out, we reuse the `_main` part of the string in one place with `.runfiles` and in one place without it.

But I definitely see way to consolidate this a bunch with some common named constants that can mark what each component is and what it means. I've extracted those. Does that help?

Worried I'm not understanding what you're looking for here... if so, happy to chat and figure out what would work better for this code.

toolchain/install/installation.cpp
chandlerc: I really didn't want to have both of these either. =/

Sadly, the way Bazel executes test makes it impossible as far as I can tell to use the same logic for the following three things:

1) Running a unittest.
2) Running inside of a `genrule` or as part of a build step.
3) Running the binary directly out of `./bazel-bin/...`.

I'm not 100% certain whether `bazel run` behaves like (2) or (3) here, but it is one of them.

From what I can tell, (1) has `TEST_SRCDIR` and `RUNFILES_DIR` set, (2) has `RUNFILES_DIR` and `binary_name.runfiles/` symlink, and (3) obviously only has `binary_name.runfiles/`. For some reason (1) doesn't have `binary_name.runfiles/`, I really don't know why.

Anyways, that's how I ended up with the logic breakdown here. I really wanted to make sure that directly running `./bazel-bin/...` works, and that seemed to handle everything but tests, so I added `TEST_SRCDIR` for that case. I couldn't find a single set of logic that would handle all of these, but if I'm missing something, that would be fantastic.

Regarding `GetRunfiles`, as you mention, the license is a worry. And while it is running as *part* of the test, the actual detection code is in the "production" binary, so I don't see a way to avoid it. The API also seems quite tricky to use here, although there is probably a way. WDYT, any better ideas?

toolchain/install/installation.h
chandlerc: As discussed, going with `InstallPaths`, and I'll remove the `install_` prefix in one place and the `_path` suffixes elsewhere in the public API for this.

chandlerc (commented):
Thanks for all the feedback. I think I got all the comments here, plus a the ones in #3989. But if I missed anything, let me know.

A few of them I think I still have questions on that need to get resolved.

toolchain/install/install_paths.h
```diff
@@ -0,0 +1,78 @@
...
+#include "llvm/ADT/StringRef.h"
+
+namespace Carbon {
+
+// Locates the toolchain installation and provides paths to various components.
+//
+// The Carbon toolchain expects to be installed into some install prefix. For
+// example, this is expected to be similar to the CMake install prefix:
+//
+// - `C:/Program Files/Carbon` or similar on Windows.
```
jonmeow, line 19: BTW, are you deliberately using `/` instead of `\` here? i.e., these are not expected to be windows-formatted paths? Please explicitly document that if so.

toolchain/install/installation.cpp
jonmeow: For background, we've previously run into issues with runfiles detection in tests (addressed by #1207). That's why I'm aware of this issue, and would like to avoid creating more GetRunfiles implementations.

Regarding license, a test-only library could do the call and pass in the path. To address the single-file constraint, one option would be to use a sentinel data file to deduce the runfiles path. I understand the inconvenience, but also I would like to avoid the costs of maintaining our own implementation of GetRunfiles.

This is probably worth further discussion of details.

toolchain/install/installation_test.cpp
jonmeow: Note, see above comment about runfiles. I think we should discuss this a little further.

toolchain/install/BUILD
```diff
@@ -0,0 +1,152 @@
...
+
+package(default_visibility = ["//visibility:public"])
+
+# Build rules supporting the install data tree for the Carbon toolchain.
+#
+# The directory `prefix_root` should mirror a Unix-y FHS (Filesystem Hierarchy
+# Standard) style install of Carbon under some prefix (`/usr`, `/usr/local`, or
+# possibly a home directory or a custom prefix). When installed, user facing
+# binaries, which we largely expect to be solely the `carbon` driver will be at
+# `perfix_root/bin/carbon`. We collect data, private executables, and other
```
jonmeow:
```suggestion
# `prefix_root/bin/carbon`. We collect data, private executables, and other
```

toolchain/install/BUILD
```diff
@@ -0,0 +1,152 @@
...
+load("symlink_filegroup.bzl", "symlink_filegroup")
+
+package(default_visibility = ["//visibility:public"])
+
+# Build rules supporting the install data tree for the Carbon toolchain.
+#
+# The directory `prefix_root` should mirror a Unix-y FHS (Filesystem Hierarchy
+# Standard) style install of Carbon under some prefix (`/usr`, `/usr/local`, or
+# possibly a home directory or a custom prefix). When installed, user facing
+# binaries, which we largely expect to be solely the `carbon` driver will be at
```
jonmeow: Are you expecting related tools like formatting, tidy-ing, packaging, language server, etc to all be provided by the driver? I'd been assuming we'd have these as separate names, even if they're busyboxed together in a release.

toolchain/install/BUILD
```diff
@@ -0,0 +1,152 @@
...
+
+load("@llvm-project//llvm:binary_alias.bzl", "binary_alias")
+load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test")
+load("symlink_filegroup.bzl", "symlink_filegroup")
+
+package(default_visibility = ["//visibility:public"])
+
+# Build rules supporting the install data tree for the Carbon toolchain.
+#
+# The directory `prefix_root` should mirror a Unix-y FHS (Filesystem Hierarchy
```
jonmeow: Just to note, you could also reference install_paths.h for information on the directory structure, maybe shifting to have comments in only one place.

toolchain/install/install_paths.h
```diff
@@ -0,0 +1,78 @@
...
+// our installation to behave in a similar and compatible way.
+//
+// We detect the installation prefix by looking at the path of the executable.
+// This supports development binaries like tests, as well as any installed
+// user-facing binaries.
+//
+// Within this prefix, we expect a hierarchy on Unix-y platforms:
+//
+// - `prefix_root/bin/carbon` - the main CLI driver.
+// - `prefix_root/lib/carbon` - private data & binaries used by the toolchain.
```
jonmeow:
```suggestion
// - `prefix_root/lib/carbon/...` - private data & binaries used by the toolchain.
```

```suggestion
// - `prefix_root/lib/carbon/` - private data & binaries used by the toolchain.
```

Stray thought, to distinguish the above line from this one. Not sure if the `...` is needed if you're trying to fit this, but I think the slash at least helps.

toolchain/install/install_paths.cpp
```diff
@@ -0,0 +1,108 @@
...
+        << "Found runfiles tree but not an install prefix directory!";
+    return;
+  }
+
+  // Otherwise, we use whatever path remained from removing the filename of the
+  // executable path. This isn't a great fallback, but we don't have a lot of
+  // choices.
+  CARBON_VLOG()
+      << "Failed to detect a recognized install path, falling back to: "
+      << prefix_;
```
jonmeow: I think this fallback might actually be dangerous and miss issues, particularly given the custom runfiles implementation. Can it verify that structure seems correct?

toolchain/install/installation.cpp
jonmeow: Looking closer at the implementation... On Runfiles, EnvVars() will contain an entry whose key is RUNFILES_DIR which could be used. This looks like it's independent of how it's found, and could probably be used instead of RLocation.

Note, I think TEST_SRCDIR might be used in CreateForTest, but isn't necessary, and isn't used at all in some alternate implementations of this logic.

toolchain/install/install_paths.h
```diff
@@ -0,0 +1,78 @@
...
+// our installation to behave in a similar and compatible way.
+//
+// We detect the installation prefix by looking at the path of the executable.
+// This supports development binaries like tests, as well as any installed
+// user-facing binaries.
+//
+// Within this prefix, we expect a hierarchy on Unix-y platforms:
+//
+// - `prefix_root/bin/carbon` - the main CLI driver.
+// - `prefix_root/lib/carbon/...` - private data & binaries used by the toolchain.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
// - `prefix_root/lib/carbon/...` - private data & binaries used by the
// toolchain.
```

toolchain/install/install_paths.h
chandlerc: That's what the CMake docs use -- I don't have any opinion about the best form of `/` here.

toolchain/install/installation.cpp
chandlerc: I've added a marker file and completely re-written this code path to use Bazel's runfiles detection.

Note that this didn't work as well as I had hoped. Notably, I have to pass in the workspace `_main/` prefix into all the paths for the runfiles detection to work correctly. An because I use Bazel's runfiles detection, I can't make `.runfiles/_main` be a specific string here, it has to *just* be `_main`.... at least, as far as I can tell after trying and debugging a bunch of things. The API documentation for the Bazel APIs doesn't help me at all so I'm really just guessing and trying things here.

toolchain/install/installation_test.cpp
chandlerc: As above, and as discussed, all of this has been rewritten.

toolchain/install/BUILD
chandlerc: I'm not sure how those tools will work.

I suspect there is a huge advantage in terms of install size of we arrange for everything to be at least busyboxed together.

Whether we add symlink support or use subcommands I think is a more complex question, and I'm not sure what the answer will be in all cases.

My personal feeling is that for things where we control the user-facing CLI, avoiding the symlink and just having a subcommand will be better since we'd want to busybox it anyways. But I'm not super confident there yet, and it seems like something we should leave the door open to.

toolchain/install/BUILD
chandlerc: Done, moved to just use the header file for docs here. Good suggestion to not duplicate.

toolchain/install/install_paths.cpp
chandlerc: Ok, I've switched to verifying everything and having an error state that is exposed.

FWIW, I tried a bunch of other API designs first before landing here, none of them worked well. I wanted to use an `ErrorOr` return or an `optional` return, but both proved really challenging to make work well. Having the class itself retain error information worked *much* better. =/

chandlerc (commented):
Due to new and exciting ways download failures can manifest, struggling to get clean test runs here, but I think I've implemented what was desired in our discussion. Narrow and explicit use of different detection strategies, along with some amount of validation and error handling. The error management was awkward as we need to create drivers with bad installs currently. And moreover, don't have support in the driver for diagnosing this. So all of that is TODO -- we need diagnostics and other things to cover when we expect a non-errors install to be there and instead have an erroneous one. But the tracking of errors is in place so we should be able t add that later? This PR is already really big and taking a lot of iterations so I didn't want to add anything more to it. Worried about how much I've had to add to make all this work as-is. Anyways, hopefully this makes more sense to folks. And if not, happy to chat about what would make more sense to unblock all of this.

toolchain/install/install_paths.cpp
jonmeow: Thanks, I see that. But it looks like `error()` is only checked in install_paths_test -- am I missing something? Note, `ErrorOr` does have some burden, but that's because it's forcing checks.

If you prefer to keep returning a finished `InstallPaths` when there's a structural error, you might consider something like `CheckMarkerFile() -> ErrorOr<Success>` instead of storing an `error()` state. Really there's not much need to store the error if it's not used, and it avoids an implication that the structure is acting on invalid state.

toolchain/install/install_paths.h
```diff
@@ -0,0 +1,112 @@
...
+// that are stored in the class, and a `std::string` for any that are computed
+// on demand.
+//
+// TODO: Need to check the installation structure of LLVM on Windows and figure
+// out what Carbon's should be within a Windows prefix and how much of the
+// structure we can share with the Unix-y layout of the prefix.
+class InstallPaths {
+ public:
+  // Provide the current executable's path to detect the correct installation
+  // prefix path. This requires the toolchain to be in its installed layout.
```
jonmeow:
```suggestion
  // Provide the current executable's path to detect the correct installation
  // prefix path. This assumes the toolchain to be in its installed layout.
```

Note, if you're not validating, I think "requires" could be misleading.

toolchain/install/install_paths.cpp
jonmeow: Oh, I see, you're clearing the path, which will cause cwd to be used. But still, shouldn't more places -- at least our tests -- be validating they got a correct structure back? Do we want to support unexpected directly paths?

toolchain/install/install_paths.h
```diff
@@ -0,0 +1,112 @@
...
+// own executable path is used to locate the rest of the install. We also
+// support locating an install through Bazel's runfiles tree or through an
+// explicit path for other use cases. When locating an install, we verify it by
+// looking for the `carbon_install.txt` marker file at a specific location
+// below. The install paths object retains any error information so that the
+// driver can diagnose errors and report them as needed, but continue to
+// function minimally. No methods will crash even in an error state, they will
+// just return based on an empty install prefix.
+//
+// Within this prefix, we expect a hierarchy on Unix-y platforms:
```
jonmeow, line 40: Should the cwd fallback be noted here?

toolchain/install/install_paths.h
```diff
@@ -0,0 +1,112 @@
...
+// There are multiple ways of locating an install's prefix for different
+// situations. For command line tools distributed as part of the install, their
+// own executable path is used to locate the rest of the install. We also
+// support locating an install through Bazel's runfiles tree or through an
+// explicit path for other use cases. When locating an install, we verify it by
+// looking for the `carbon_install.txt` marker file at a specific location
+// below. The install paths object retains any error information so that the
+// driver can diagnose errors and report them as needed, but continue to
+// function minimally. No methods will crash even in an error state, they will
+// just return based on an empty install prefix.
```
jonmeow:
```suggestion
// There are multiple ways of locating an install's prefix:
//   - MakeExeRelative for command line tools in an install.
//   - MakeForBazelRunfiles for locating through Bazel's runfile tree.
//   - Make for an explicit path, for example in tests.
//
// When locating an install, we verify it by
// looking for the `carbon_install.txt` marker file at a specific location
// below. When errors occur, the install prefix is made empty, and error() can be used for diagnostics; InstallPaths remains minimally functional.
```

Suggesting as a way to break this long paragraph apart a little.

jonmeow (approved):
Approving, though I'd suggest at least having tests + fuzzer uses CHECK-fail or similar if there's an error. We should really expect valid trees in those.

toolchain/install/install_paths.h
```diff
@@ -0,0 +1,112 @@
...
+
+ private:
+  InstallPaths() : error_("No prefix provided!") {}
+  explicit InstallPaths(llvm::StringRef prefix) : prefix_(prefix) {}
+
+  auto SetError(llvm::Twine message) -> void;
+  auto CheckMarkerFile() -> void;
+
+  llvm::SmallString<256> prefix_;
+  std::optional<std::string> error_;
```
jonmeow, line 119: FWIW, I think you could also represent this as `ErrorOr<Success>`

toolchain/install/install_paths.h
```diff
@@ -0,0 +1,112 @@
...
+  // using Carbon in an environment with an unusual path to the installed files.
+  static auto Make(llvm::StringRef install_prefix) -> InstallPaths;
+
+  // Check for an error detecting the install paths correctly.
+  //
+  // An empty return means no errors encountered and the paths should work
+  // correctly.
+  //
+  // A non-empty return means there was an error, and details of the error are
+  // in the `StringRef` for inclusion in any user report.
```
jonmeow:
```suggestion
  // Check for an error detecting the install paths correctly.
  //
  // A nullopt return means no errors encountered and the paths should work
  // correctly.
  //
  // A string return means there was an error, and details of the error are
  // in the `StringRef` for inclusion in any user report.
```

Was this originally written as returning a StringRef directly, not optional?

toolchain/install/install_paths.h
```diff
@@ -0,0 +1,112 @@
...
+
+  auto driver() const -> std::string;
+  auto llvm_install_bin() const -> std::string;
+
+ private:
+  InstallPaths() : error_("No prefix provided!") {}
+  explicit InstallPaths(llvm::StringRef prefix) : prefix_(prefix) {}
+
+  auto SetError(llvm::Twine message) -> void;
+  auto CheckMarkerFile() -> void;
```
jonmeow: Would it make sense to add comments to these, to explain what they're doing?

toolchain/install/install_paths.h
chandlerc: Doh, yes. Thanks. =]

toolchain/install/install_paths.h
chandlerc: Sure.

toolchain/install/install_paths.h
chandlerc: Hmm, I tied this and it ended up being awkward... `ErrorOr` isn't copyable, and seems to somewhat want consumers to move the error out of it. Using an optional string allows just a single accessor and seems a bit simpler. Also not sure the location is really helpful from `Error` here.

toolchain/install/install_paths.cpp
chandlerc: So there are three cases...

We should definitely directly validate this in mosts tests, I'll work on adding that.

We can't (yet) validate this in fuzzers because we need to restructure the fuzzer code to support finding this. I've left TODOs about that, and wanted to get to this after moving the prelude over. But once done, that should remove the need for a fuzzer/test construction that is unvalidated I think.

The driver is a bit trickier because some invocations of the driver don't use the install. Does it make sense to unconditionally error on missing install even if we don't need it? I can see arguments in both directions and don't feel strongly here. But even if we do validate in all cases, I think we want to do that inside the driver where we have our diagnostic machinery so we can print the error using that infrastructure.

So my plan is to leave a lot of the missing validation here for this PR (for fuzzers and the driver), and then work on both getting the fuzzers to use this and getting the driver to validate it however much seems reasonable.

Does that make sense?

toolchain/install/install_paths.h
```diff
@@ -0,0 +1,110 @@
...
+// our installation to behave in a similar and compatible way.
+//
+// There are multiple ways of locating an install's prefix:
+//   - MakeExeRelative for command line tools in an install.
+//   - MakeForBazelRunfiles for locating through Bazel's runfile tree.
+//   - Make for an explicit path, for example in tests.
+//
+// When locating an install, we verify it by
+// looking for the `carbon_install.txt` marker file at a specific location
+// below. When errors occur, the install prefix is made empty, and error() can be used for diagnostics; InstallPaths remains minimally functional.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
// below. When errors occur, the install prefix is made empty, and error() can
// be used for diagnostics; InstallPaths remains minimally functional.
```

chandlerc (commented):
Updated based on review, PTAL though as there are still some discussion points where I think there are options here.

toolchain/install/install_paths.cpp
chandlerc: FWIW, this seems fine to fix forward, and so based on approval going to merge as-is. It at least covers checking in all the tests. And it'll let me empty the stack of PRs. Happy to follow up and fix more cases, and make any further API tweaks here though if there are better APIs, especially if we have a better idea of how to handle diagnosing issues here in the driver itself (the only place where it seems a bit tricky).

chandlerc:
> Updated based on review, PTAL though as there are still some discussion points where I think there are options here.

Mentioned above, but should mention in the main thread for visibility -- re-looking at this and the approval from Jon, I'm gonna merge with the tests checking this. I'll work on follow-up changes to improve fuzzing, and see how we can handle this in the driver and if we can remove the error state from the design at some point.

PR #3994, opened 2024-05-27
Add very minimal support for packaging the toolchain.

chandlerc:
This takes the installation layout and replicates it using `rules_pkg`
to build either a tarball or a zip file of the toolchain. Correctly
manages file permissions and symlinks, etc.

There are some big remaining things here:

- Figure out how we want to test this. We can add shell tests maybe?
  A bit awkward. Nicer would be to make the `//examples` tree build
  using this rather than the more native-bazel install data, however
  building these is quite slow and it seems bad to pay that cost
  constantly so dedicated testing is probably better. For now, I've
  tested these manually.

- Need to add versions to the toolchain and then thread them through
  here so they install properly as a versioned release.

But my primary goal for now is just to be able to validate that the
install tree is working outside of Bazel and this does enough for that.
The above will be longer-term things.

### DIFF
--- a/toolchain/install/BUILD
+++ b/toolchain/install/BUILD
@@ -5,6 +5,9 @@
 load("@bazel_skylib//rules:write_file.bzl", "write_file")
 load("@llvm-project//llvm:binary_alias.bzl", "binary_alias")
 load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test")
+load("@rules_pkg//pkg:mappings.bzl", "pkg_attributes", "pkg_filegroup", "pkg_files", "pkg_mklink", "strip_prefix")
+load("@rules_pkg//pkg:tar.bzl", "pkg_tar")
+load("@rules_pkg//pkg:zip.bzl", "pkg_zip")
 load("symlink_filegroup.bzl", "symlink_filegroup")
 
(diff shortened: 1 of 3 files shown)


toolchain/install/BUILD
```diff
@@ -157,3 +160,62 @@ cc_test(
...
+    ] + [
+        ":pkg_link_lld_alias_" + bin_name
+        for bin_name in lld_bin_names
+    ],
+)
+
+# TODO: We should add support for injecting a version string into both the
+# output filename and the package directory name.
+pkg_tar(
+    name = "pkg_tar",
```
jonmeow: Had you considered keeping target names distinct from rule names? e.g., something like "carbon_toolchain.txz.rule" for this and `pkg_zip`, or just removing the `pkg_` prefixes on all of these? (if just removing `pkg_` bugs you for collision risk, you could make a `pkg` subdir)

jonmeow (approved):
The related commit looks good, not merging for the stack.

toolchain/install/BUILD
chandlerc: Good call, using the `<output name>.rule` pattern, that seems nice.

chandlerc (commented):
Thanks for review. Dependency is merged, and name improvement applied. Merging this one!

PR #3995, opened 2024-05-27
Move the core standard library to the installation.

chandlerc:
This removes the `data_dir` from the driver favoring the installation
abstraction for the both locating the prelude and linking utilities.

With this, an installed toolchain should also be able to compile and
link Carbon successfully, and the build of the examples should exercise
this path almost exactly. (The only difference is using the driver
`cc_binary` directly rather than relying on the symlink from inside the
install tree.)

### DIFF
--- a/toolchain/driver/driver.cpp
+++ b/toolchain/driver/driver.cpp
@@ -32,7 +32,7 @@
 
 namespace Carbon {
 
-auto Driver::FindPreludeFiles(llvm::StringRef data_dir,
+auto Driver::FindPreludeFiles(llvm::StringRef core_package_dir,
                               llvm::raw_ostream& error_stream)
     -> llvm::SmallVector<std::string> {
   llvm::SmallVector<std::string> result;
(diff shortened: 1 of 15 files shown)


toolchain/driver/driver.cpp
```diff
@@ -32,24 +32,24 @@
...
+    llvm::SmallString<256> prelude_file(core_package_dir);
     llvm::sys::path::append(prelude_file, llvm::sys::path::Style::posix,
-                            "core/prelude.carbon");
+                            "prelude.carbon");
     result.push_back(prelude_file.str().str());
   }
 
   // Glob for <data>/core/prelude/**/*.carbon and add all the files we find.
-  llvm::SmallString<256> prelude_dir(data_dir);
+  llvm::SmallString<256> prelude_dir(core_package_dir);
```
jonmeow, line 50: Why `core_package_dir` in some spots but `core_package_path` in others? What's the difference between these values?

jonmeow (approved):
LG

toolchain/driver/driver.cpp
chandlerc: No difference. I have a mild preference for `_path` when it's actually a path string, and `_dir` for something more like a directory handle. But this code used `_dir` for `prelude_dir` so I tried to stay consistent with that in this file.

I'm happy to consolidate these names in either direction in a follow-up, can discuss which names folks would prefer because my preference is very mild here.

chandlerc (commented):
Thanks again for review, merging once the dependency finishes in the merge queue. And will follow up on any naming adjustments.

PR #4000, opened 2024-05-28
Move child count and bracketing information for parse nodes into the node kind definition.

zygoloid:
Instead of tracking the bracketing and child count information in the kind macro in `node_kind.def`, provide it to `NodeKind::Define` in `typed_nodes.h`. If a node is both bracketed and has a fixed child count, track both facts and check them both in tree verification, since it's easy to do so now.

The overall goal here is to reduce `node_kind.def` down to a simple list of names. I have a slightly different approach in mind for the token kinds.


toolchain/parse/tree.cpp
```diff
@@ -220,7 +221,18 @@ auto Tree::Verify() const -> ErrorOr<Success> {
...
+        ++children;
         if (n_impl.kind.bracket() == child_impl.kind) {
+          // If there's a bracketing node and a child count, verify the child
+          // count too.
+          if (n_impl.kind.has_child_count() &&
+              children != n_impl.kind.child_count()) {
+            return Error(llvm::formatv(
+                "NodeId #{0} is a {1} with child_count {2}, but encountered "
+                "{3} "
+                "nodes before we reached the bracketing node.",
```
jonmeow:
```suggestion
                "{3} nodes before we reached the bracketing node.",
```

toolchain/parse/tree.cpp
```diff
@@ -211,6 +211,7 @@ auto Tree::Verify() const -> ErrorOr<Success> {
 
     int subtree_size = 1;
     if (n_impl.kind.has_bracket()) {
+      int children = 0;
```
jonmeow: Maybe name child_count to match the property?

toolchain/parse/node_kind.h
```diff
@@ -119,20 +145,69 @@ class NodeKind::Definition : public NodeKind {
...
  private:
   friend class NodeKind;
 
-  constexpr Definition(NodeKind kind, NodeCategory category)
-      : NodeKind(kind), category_(category) {}
-
-  NodeCategory category_;
+  // This is factored out and non-constexpr to improve the compile-time error
+  // message if the check below fails.
+  auto MustSpecifyEitherBracketingNodeOrChildCount() {
```
jonmeow, line 173: This trick is making me look closer at the templating approach (I'm unsure why you're doing this, versus something like static_assert or another approach). The `Args` approach also feels a little odd to me (what if callers provide things multiple times? what about ordering?)

How much had you considered providing non-templated construction? I'm actually wondering if you could do something like:

```
  constexpr explicit Definition(
      NodeKind kind, NodeCategory category, BracketedByKind bracket, ChildCount child_count) {
      : NodeKind(kind), category_(category), bracket_(bracket), child_count_(child_count) {
  }
```

Then, handle overloads in Define:

```
// Note, this one would need to be defined out-of-line. The rest could be defined inline.
constexpr auto Define(
      NodeCategory category, BracketedByKind bracket, ChildCount child_count = ChildCount(0)) const -> Definition {
  return Definition(*this, category, bracket, child_count);
}

constexpr auto Define(
      NodeCategory category, ChildCount child_count) const -> Definition {
  return Define(category, *this, child_count);
}

constexpr auto Define(
      BracketedByKind bracket, ChildCount child_count = ChildCount(0)) const -> Definition {
  return Define(NodeCategory::None, bracket, child_count);
}

constexpr auto Define(
     ChildCount child_count) const -> Definition {
  return Define(NodeCategory::None, *this, child_count);
}
```

(note, the category-less overload doesn't need to be out-of-line)

The overloads aren't the best, but comparing with the current approach, I think it may still be less code overall, and possibly easier to understand due to lack of templating.

Note, I get the idea that the ApplyArg lets you add arbitrary more arguments without refactoring, but if you really want to head down that approach I'd suggest considering a parameter struct. Particularly with the `ChildCount` and `BracketedBy`, I think it might even be simpler, removing the need for BracketedByKind and ChildCount. i.e.:

```
struct DefineArgs {
  NodeCategory category = NodeCategory::None;
  std::optional<NodeKind> bracketed_by = std::nullopt;
  int child_count = -1;
};

  static constexpr auto Kind = NodeKind::BreakStatement.Define({
      .category = NodeCategory::Statement, .bracketed_by = BreakStatementStart, .child_count = 1});
```

It's slightly more verbose at the call sites, but I think not by much, and might be worth it if you're really thinking about adding more. Also, the struct approach would still enforce a consistent parameter ordering.

toolchain/parse/node_kind.h
```diff
@@ -119,20 +145,69 @@ class NodeKind::Definition : public NodeKind {
   Definition(const Definition&) = delete;
   auto operator=(const Definition&) -> Definition& = delete;
 
+  // Returns true if the node is bracketed; otherwise, child_count is used.
+  constexpr auto has_bracket() const -> bool { return bracket_ != *this; }
```
jonmeow: It looks like you're changing the semantic so that both a bracket and child_count can be provided. I can see why you're doing that, but I think this now makes the documentation unclear: "otherwise" implies to me mutually exclusive.

Please clarify comments about the relationship between the attributes. It might be worth documenting at the class level, given the scope of the relationship.

toolchain/parse/node_kind.h
```diff
@@ -119,20 +145,69 @@ class NodeKind::Definition : public NodeKind {
...
+
+  // Returns the bracketing node kind for the current node kind. Requires that
+  // has_bracket is true.
```
jonmeow, line 149: Note that with your changes to allow both, "requires" comments are now incorrect.

zygoloid: Fixed the comment on `child_count` (this one is still correct).

toolchain/parse/node_kind.h
zygoloid: Switched to an argument struct. As you say, it's a bit more verbose, but I think that's fine. (Note the `bracketed_by` now needs a `::Kind` suffix too.)

toolchain/parse/tree.cpp
zygoloid: Done.

toolchain/parse/node_kind.h
zygoloid: Done.

toolchain/parse/node_kind.h
```diff
@@ -47,6 +47,15 @@ CARBON_DEFINE_RAW_ENUM_CLASS(NodeKind, uint8_t) {
...
 // tree.
+//
+// In order to allow the children of a node to be determined without relying on
+// the subtree size field in the parse node, each node kind must have one of:
+//
+// - a bracketing node kind, which is always the kind for the first child, and
+//   is never the kind of any other child, or
+// - a fixed child count,
+//
+// or both. This is required even for nodes that are invalid.
```
jonmeow:
```suggestion
// or both. This is required even for nodes produced by an invalid parse.
```

jonmeow (approved):
LGTM, just one nit on comment phrasing, "invalid" feels ambiguous.

toolchain/parse/node_kind.h
zygoloid:
```suggestion
// or both. This is required even for nodes for which `Tree::node_has_errors` returns `true`.
```
I specifically mean nodes marked with the `has_errors` flag, so how about this?

toolchain/parse/node_kind.h
```diff
@@ -47,6 +47,15 @@ CARBON_DEFINE_RAW_ENUM_CLASS(NodeKind, uint8_t) {
...
 // tree.
+//
+// In order to allow the children of a node to be determined without relying on
+// the subtree size field in the parse node, each node kind must have one of:
+//
+// - a bracketing node kind, which is always the kind for the first child, and
+//   is never the kind of any other child, or
+// - a fixed child count,
+//
+// or both. This is required even for nodes for which `Tree::node_has_errors` returns `true`.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
// or both. This is required even for nodes for which `Tree::node_has_errors`
// returns `true`.
```

PR #4001, opened 2024-05-29
Move information on the token associated with a parse node from the `.def` file into the typed node.

zygoloid:
Instead of tracking the token associated with a parse node in the `.def` file macro, track it on the typed node instead. List the token as a field inside the node structure to show the order of the token relative to the other components of the grammar production, and to allow the token index to be accessed when the node is extracted.

Remove the corresponding information from the `.def` file, leaving behind just a list of parse node kinds in the majority of cases.

This also removes the checking of the token kind associated with a parse node in the case where the parse node has errors. Previously we had a flag on the node kind to indicate whether we should check this, but per [discord discussion](https://discord.com/channels/655572317891461132/655578254970716160/1246214418979881052), we have decided to remove this.


jonmeow:
Note, in consideration for my comments on #4000, I would suggest considering the Define comments as they apply here.

zygoloid:
> Note, in consideration for my comments on #4000, I would suggest considering the Define comments as they apply here.

They don't apply here -- this PR doesn't change `Define` now that the dependent PR has landed.

toolchain/parse/typed_nodes_test.cpp
```diff
@@ -276,6 +276,113 @@ Aggregate [^:]*: success
...
+  auto* tree = &GetTree(R"carbon(
+    var n: i32 = 0;
+  )carbon");
+  auto file = tree->ExtractFile();
+
+  ASSERT_EQ(file.decls.size(), 1);
+
+  auto n_var = tree->ExtractAs<VariableDecl>(file.decls[0]);
+  ASSERT_TRUE(n_var.has_value());
+  EXPECT_TRUE(n_var->token.index.index == 7);
```
jonmeow: This feels like a pretty strict validation on how both lex and parse work, which feels like it'll have some brittleness if we end up wanting to change things (both with pattern parsing and var, which I suspect at least patterns could change a little; maybe less so with fn below). Is this needed?

toolchain/parse/typed_nodes_test.cpp
```diff
@@ -276,6 +276,113 @@ Aggregate [^:]*: success
...
+Optional [^:]*: begin
+NodeIdForKind: ReturnType consumed
+Optional [^:]*: found
+Aggregate [^:]*: begin
+Aggregate [^:]*: begin
+Optional [^:]*: begin
+NodeIdForKind: TuplePattern consumed
+Optional [^:]*: found
+Optional [^:]*: begin
+NodeIdForKind error: wrong kind IdentifierName, expected ImplicitParamList
```
jonmeow, line 384: How much value is this test adding? I've had some frustration updating our other similar tests, that are asserting the structure of trace output. Similar to the above, but definitely I've seen more brittleness from changes in this kind of check since it's much more integrated with how the code actually functions in detail.

toolchain/parse/node_ids.h
```diff
@@ -110,6 +111,33 @@ struct NodeIdNot : public NodeId {
       : NodeId(NodeId::InvalidIndex) {}
 };
 
+// This class holds the token corresponding to a parse node, and defines the
+// expected token kind. The specified token index will always have kind K,
+// except if RequireIfInvalid is false and the node is invalid.
+template <const Lex::TokenKind& K, bool RequireIfInvalid = true>
+struct Token {
```
jonmeow, line 120: I'm concerned that calling this `Token` will lead to confusion with `Lex::TokenizedBuffer::Token`, or that `TokenIndex` is the index of a `Token`. Maybe `ExpectedToken` or `TokenIndexKind`? `ValidatedToken`?

Or structurally, this is what led to my high-level comment about token kind on the node definition instead of wrapping TokenIndex. But, another approach would be to just have a separate constexpr field for the token kind, and have the token index be instance.

Note, influencing my thoughts here is that anything operating on this needs the actual TokenIndex, so the point that code will be doing `token.index` to get a `TokenIndex` instead of an `int32_t` (`token.index.index`) feels like it has potential for confusion.

toolchain/parse/node_kind.def
```diff
@@ -9,55 +9,37 @@
...
-//       Defines a parse node for an infix operator, with the Name as token.
-//     - CARBON_PARSE_NODE_KIND_POSTFIX_OPERATOR(Name)
-//       Defines a parse node for a postfix operator, with the Name as token.
-//     - CARBON_PARSE_NODE_KIND_TOKEN_LITERAL(Name, LexTokenKind)
-//       Defines a parse node that corresponds to a token that is a single-token
-//       literal. The token is wrapped for LexTokenKinds.
-//     - CARBON_PARSE_NODE_KIND_TOKEN_MODIFIER(Name)
-//       A token-based modifier. The Name is the TokenKind, and will be appended
-//       with "Modifier" for the parse kind.
+//   Used as a fallback, or when none of the below cases apply.
```
jonmeow: The phrasing to me could be read as "fallback, aka when none of the below cases apply". That said, we have analogous phrasing in builtin_kind.def (which I'd like to keep in sync, one way or the other, for the parity in concepts):

```
//   Used as a fallback if other macros are missing. Used directly by Invalid
//   only, which is defined last.
```

Maybe here we could use something similar:

```suggestion
//   Used as a fallback if other macros are missing. Also, used directly
//   when none of the below category-oriented macros apply.
```

toolchain/parse/node_kind.cpp
```diff
@@ -37,30 +36,9 @@ CARBON_DEFINE_ENUM_CLASS_NAMES(NodeKind) = {
...
 };
 
-auto NodeKind::CheckMatchesTokenKind(Lex::TokenKind token_kind, bool has_error)
-    -> void {
-  static constexpr Lex::TokenKind TokenIfValid[] = {
-#define CARBON_IF_VALID(LexTokenKind) LexTokenKind
-#define CARBON_PARSE_NODE_KIND_WITH_TOKEN(Name, LexTokenKind) \
-  Lex::TokenKind::LexTokenKind,
+// Check that each typed node defines a `token` member.
+#define CARBON_PARSE_NODE_KIND(Name) static_assert(&Name::token != nullptr);
```
jonmeow: Does this need the nullptr comparison? FWIW I think we often don't do `!= nullptr` in conditions.

```suggestion
#define CARBON_PARSE_NODE_KIND(Name) static_assert(&Name::token);
```

toolchain/parse/tree.h
```diff
@@ -277,11 +279,18 @@ class Tree : public Printable<Tree> {
...
   template <typename T>
   auto VerifyExtractAs(NodeId node_id, ErrorBuilder* trace) const
       -> std::optional<T>;
 
+  // Sets the kind of a node. This is intended to allow putting the tree into a
+  // state where verification can fail, in order to make the failure path of
+  // `Verify` testable.
+  auto SetNodeKindForVerifyTest(NodeId node_id, NodeKind kind) -> void {
+    node_impls_[node_id.index].kind = kind;
+  }
```
jonmeow: There's the intention to have things like the parse tree be an API for tools. Would it be worth using `friend` to keep this private so that it's not part of the public API?

toolchain/parse/tree.h
```diff
@@ -103,6 +103,8 @@ class Tree : public Printable<Tree> {
     bool is_impl;
   };
 
+  struct ExtractState;
```
jonmeow: Had you considered making this a `friend` instead of a member class? As a member class, I would expect to find it in tree.cpp, rather than extract.cpp.

Also, can you document why this is here, either way? That isn't obvious to me at a glance.

toolchain/parse/extract.cpp
```diff
@@ -36,43 +36,56 @@ namespace Carbon::Parse {
 template <typename T>
 struct Extractable;
 
+struct Tree::ExtractState {
```
jonmeow: This structure seems to be uncommented -- I think it'd be good to explain what it is and how it's used.

But also, would this make sense as a class with the various functions that are passing around `ExtractState` instead as member functions? You already have several member functions here, and it doesn't feel as struct-y as a result.

jonmeow (commented):
High-level comment.

Right now, you're wrapping TokenIndex with a TokenKind that's expected by the parse node. However, we already have a parse node structure, and it seems to only be validated as part of parse node verification.

Had you considered instead tracking the TokenKind as part of the NodeKind definition? What are we getting by creating the separate verification path?

Or, maybe putting TokenIndex and TokenKind separately on the typed node object, rather than providing a wrapper for that?

Note, I comment on Token too -- I think there's some potential for confusion there, which is making me think more about alternatives.

toolchain/parse/typed_nodes_test.cpp
zygoloid: FWIW, I'd actually expect this test to be one of the least brittle ones in this file: the tests that check the verification trace output elsewhere in this file are substantially more brittle, and break frequently when the tree structure changes.

But, switched to check the token kind instead of the token index, to avoid making assumptions about the meaning of token indexes.

toolchain/parse/typed_nodes_test.cpp
zygoloid: Yeah, I've seen lots of brittleness in this file too. I think it is important that we have some tests that check that our token kind verification code is actually working -- but this test does seem to be testing too much about the output, and could easily be made less brittle. Done.

toolchain/parse/node_kind.cpp
zygoloid: To me, forming a pointer-to-member in a static assert condition looked wrong, because it clearly can't ever be null. Replaced with a requires-expression, which hopefully better captures the intent that we're checking a requirement on the type here.

toolchain/parse/tree.h
zygoloid: I don't have any particular opinion here. https://abseil.io/tips/135 suggests adding a test-only public member function instead of a friend, but I'm happy to go down the friend path if you prefer. I don't know how much that tip considered the Hyrum's law impact of making test-only functionality like this public.

toolchain/parse/tree.h
zygoloid: Ultimately, this was a member so I could access the `buffer_` member of `Tree`, but that's easily avoided. Made this into a non-member.

toolchain/parse/extract.cpp
zygoloid: Documented and made some things members. Given the way the `Extractable` specializations work out, keeping the members public seems a little simpler than moving all the remaining code into member functions and turning this into a class, but I can do that if you prefer.

zygoloid:
> Had you considered instead tracking the TokenKind as part of the NodeKind definition? What are we getting by creating the separate verification path?

For me, listing the token kind at the right place in the struct makes the typed node definitions significantly more useful and readable: these structs can then be read directly as grammar productions. For example:

```c++
struct MemberAccessExpr {
  static constexpr auto Kind = NodeKind::MemberAccessExpr.Define(
      {.category = NodeCategory::Expr, .child_count = 2});

  AnyExprId lhs;
  Token<Lex::TokenKind::Period> token;
  AnyMemberNameOrMemberExprId rhs;
};
```
corresponds to the grammar rule
```ebnf
member-access-expr = any-expr "." any-member-name-or-member-expr
```

which we can now read directly from the struct members.

Instead of having a `Token<K>` wrapping a `TokenIndex`, we could consider having separate `TokenIndex`-like types for each kind of token index that has a known token kind, in the same way we do for `NodeId`. I'm not sure that we would get enough benefit from having more strongly typed `TokenIndex`es to justify the costs there, but I'll prototype it and see how it looks.

zygoloid:
> Instead of having a `Token<K>` wrapping a `TokenIndex`, we could consider having separate `TokenIndex`-like types for each kind of token index that has a known token kind, in the same way we do for `NodeId`. I'm not sure that we would get enough benefit from having more strongly typed `TokenIndex`es to justify the costs there, but I'll prototype it and see how it looks.

[Here](https://github.com/zygoloid/carbon-lang/compare/toolchain-remove-node-def-comments-2...zygoloid:carbon-lang:toolchain-remove-node-def-comments-3?expand=1) is a prototype of how that might look.

toolchain/parse/tree.h
jonmeow: FWIW I lean down the Hyrum's law path here if we consider this a public API -- I would expect even abseil doesn't exposing its internal-test-only things as public. :)

I think the test peer approach noted in the tip might be fine. I would note the "and other clients (not just the test) might find them useful" caveat -- the intent should be that this is a public API because it's a public API, not just for our own internal tests.

toolchain/parse/extract.cpp
jonmeow: I'd really prefer a class here. For the Extractable specializations, would a `template <typename T> friend Struct Extractable<T>;` avoid any accessor overhead? (though alternately, it looks like trace and node_id might be the only things on state that are access?)

jonmeow:
FWIW, for me, the prototype reads better to me: compared with `NamedConstraintDefinitionStartId`, `Lex::CloseCurlyBraceTokenIndex` also feels more consistent style-wise than `Token<Lex::TokenKind::CloseCurlyBrace>`. The amount of code looks similar.

I agree we don't want to spend more compute on lots of strong-typing (though maybe for this area, where we're already doing it for parse nodes, it's actually okay to do it for tokens too?).

To note, for IfValid, maybe something like:

```
// A lightweight handle to a lexed token in a `TokenizedBuffer` whose kind is
// known to be `Kind`.
template <const TokenKind& K>
struct TokenIndexForKindBase : public TokenIndex {
  // NOLINTNEXTLINE(readability-identifier-naming)
  static const TokenKind& Kind;
  constexpr explicit TokenIndexForKind(TokenIndex index) : TokenIndex(index) {}
};
template <const TokenKind& K>
const TokenKind& TokenIndexForKindBase<K>::Kind = K;

// Used when the kind is always `Kind`.
template <const TokenKind& K>
struct TokenIndexForKind : TokenIndexForKindBase<K> {
  static constexpr bool IfValid = false;
};
// Used when the kind is only required to be `Kind` when the corresponding parse node doesn't have an error.
template <const TokenKind& K>
struct TokenIndexForKindIfValid : TokenIndexForKindBase<K> {
  static constexpr bool IfValid = true;
};

#define CARBON_TOKEN(TokenName) \
  using TokenName##TokenIndex = TokenIndexForKind<TokenKind::TokenName>;
  using TokenName##TokenIndexIfValid = TokenIndexForKindIfValid<TokenKind::TokenName>;
#include "toolchain/lex/token_kind.def"
```

Also, I'd also suggest doing something with `AnyToken`, not sure if you just ignored it as part of the prototype. Either renaming `AnyTokenIndex` for consistency, or just using `TokenIndex` directly to avoid the extra typedef.

toolchain/parse/tree.h
zygoloid: Done.

toolchain/parse/extract.cpp
zygoloid: Made into a class. Could probably make `Extractable` be a member class, but so far that doesn't seem necessary; using the public API seems to be working fine.

zygoloid:
How would you feel about proceeding with this PR with the `Token<K>` approach, and adding the typed `TokenIndex`es as a follow-up change? I think there's more cleanup to do before that'll be ready to go, and I'm worried about this backsliding.

jonmeow (approved):
Separating changes seems reasonable, approving.

PR #4005, opened 2024-05-30
Factor out pushing / popping of names plus parameters.

zygoloid:
Previously we did some of this in decl_name_stack and some of it in the callers of decl_name_stack. Factor out a single place to pop a name and its optional following parameters.

Part of making this behavior consistent is that we now track whether an implicit parameter list was present or absent rather than mapping an absent list to `InstBlockId::Empty`. This improves our redeclaration checking and the precision of some diagnostics.


jonmeow (approved):
Looks good!

PR #4006, opened 2024-05-30
Move errs tie to LLVM to include test infrastructure.

jonmeow:
The lack of flushing can sometimes be observed when streaming test output, particularly with --dump_output. My thought is moving it into our LLVM init is probably reasonable, given the driver's already doing this for similar reasons; this means it's used in tests through our gtest_main.cpp in addition to the driver.

### DIFF
--- a/common/init_llvm.cpp
+++ b/common/init_llvm.cpp
@@ -5,6 +5,7 @@
 #include "common/init_llvm.h"
 
 #include "llvm/Support/TargetSelect.h"
+#include "llvm/Support/raw_ostream.h"
 
 namespace Carbon {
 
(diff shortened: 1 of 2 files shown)


common/init_llvm.cpp
```diff
@@ -32,6 +33,10 @@ InitLLVM::InitLLVM(int& argc, char**& argv)
   if (InitializeTargets) {
     InitializeTargets();
   }
+
+  // Printing to stderr should flush stdout. This is most noticeable when stderr
+  // is piped to stdout.
+  llvm::errs().tie(&llvm::outs());
```
zygoloid, line 39: Looks good. Looking through the LLVM history, this was at some point intended to be the default behavior, but the change to make it so introduced a data race and got [partially reverted](https://github.com/llvm/llvm-project/commit/030897523d43e3296f69d25a71a140d9e5793c6a) (there's still a lingering comment in the header incorrectly saying that `errs()` is tied to `outs()` by default...).

zygoloid (approved):

PR #4007, opened 2024-05-30
Move some of the test information to TIP lines

jonmeow:
This has is a nice-to-have for me. Frequently I want to run a specific test, and end up digging through output to be able to copy-paste the run line. This uses TIP lines to inject the command into the file when using AUTOUPDATE.

Note, one of the reasons I want this is because "bazel test //toolchain/testing:file_test --test_output=all" has been regularly exceeding bazel's output limit for me (workaround is either opening the output file or specifying an obscure output limit flag), making it a little harder for me to get the commands. However, frequently I'm adding a file and want to iterate on it, so that's really the use case I have in mind here.


testing/file_test/file_test_base.cpp
```diff
@@ -83,30 +89,51 @@ static auto CompareFailPrefix(llvm::StringRef filename, bool success) -> void {
...
-  // If we're being run from bazel, provide some assistance for understanding
-  // and reproducing failures.
   const char* target = getenv("TEST_TARGET");
-  if (target) {
-    constexpr const char* CommandFormat =
-        "bazel {0} {1} --test_arg=--file_tests={2}";
-    test_command = llvm::formatv(CommandFormat, "test", target, test_name_);
-    // Add a crash trace entry with the run command.
-    stack_trace_entry.emplace(test_command.c_str());
+  args << "bazel " << ((mode == BazelMode::Test) ? "test" : "run") << " "
```
chandlerc, line 106: Seems a touch odd to use both `run` and `--test_arg=` for the `Dump` mode. But no strong opinion here.

testing/file_test/autoupdate.cpp
```diff
@@ -217,6 +217,22 @@ auto FileTestAutoupdater::AddRemappedNonCheckLine() -> void {
...
 
+auto FileTestAutoupdater::AddTips() -> void {
+  CARBON_CHECK(tips_.empty()) << "Should only add tips once";
+
+  tips_.reserve(4);
+  // This puts commands on a single line so that they can be easily copied.
+  tips_.emplace_back("// TIP: To test this file alone, run:");
+  tips_.emplace_back("// TIP:   " + test_command_);
+  tips_.emplace_back("// TIP: To dump output, run:");
+  tips_.emplace_back("// TIP:   " + dump_command_);
```
chandlerc, line 228: Only other thought I have here is if it makes sense to make these a bit easier to visually spot when looking at the files? I don't have many concrete ideas, and we can always revisit this, so not really suggesting a change here.

chandlerc (approved):
LG!

One question around the exact command used for `Dump`, but I'm happy either way, especially if there is a particular aspect that means we really want to do one of them.

testing/file_test/autoupdate.cpp
jonmeow: For me, I find these easy to find (typically top of the file, but at least right next to AUTOUPDATE), so I'm not quite sure what you'd be looking for.

testing/file_test/file_test_base.cpp
jonmeow: Fixing

testing/file_test/autoupdate.cpp
chandlerc: The lines just run together a bit for me skimming the file.

But this isn't important to solve here. Mostly mentioned in case you or others had ideas. Let's go ahead and land as-is.

chandlerc:
(Still LG, mashing merge button now that fix is in.)

PR #4009, opened 2024-05-30
Be a little more prescriptive about {} initialization in style.

jonmeow:
Following discussions around #3958, try to provide more specific semantics. Note we currently don't follow this everywhere, particularly in AddInst calls, but the intent is to shift. Per discussion, designated initializers are preferred when possible. And the `google-readability-casting` diagnostics are poor, but with this may primarily flag cases which should be using a different constructor syntax, so are just a rocky way to get there.

### DIFF
--- a/docs/project/cpp_style_guide.md
+++ b/docs/project/cpp_style_guide.md
@@ -135,7 +135,7 @@ these.
             how structs and tuples would be written in Carbon code.
         -   Avoid braced initialization for types that define a constructor,
             except as an initializer list
-            (`llvm::SmallVector<int> v = {0, 1};)`. Never use it with `auto`
+            (`llvm::SmallVector<int> v = {0, 1};`), `std::pair` or `std::tuple`. Never use it with `auto`
             (`auto a = {0, 1}`).
     -   Prefer parenthesized initialization (`FooType foo(10);`) in most other
         cases.
--- a/docs/project/cpp_style_guide.md
+++ b/docs/project/cpp_style_guide.md
@@ -130,8 +130,9 @@ these.
     -   Prefer braced initialization for aggregate initialization, such as
         structs, pairs, and initializer lists.
         -   Use designated initializers (`{.a = 1}`) when possible for structs,
-            but not for pairs or tuples. This is analogous to how structs and
-            tuples would be written in Carbon code.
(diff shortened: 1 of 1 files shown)


docs/project/cpp_style_guide.md
```diff
@@ -124,17 +124,27 @@ these.
...
-    -   Use `{}` initialization without the `=` only if the above options don't
+    -   Prefer braced initialization for aggregate initialization, such as
+        structs, pairs, and initializer lists.
+        -   Use designated initializers (`{.a = 1}`) when possible for structs,
+            but not for pairs or tuples. Prefer to only include the typename
+            when required to compile (`WizType{.a = 1}`). This is analogous to
+            how structs and tuples would be written in Carbon code.
+        -   Avoid braced initialization for types that define a constructor,
+            except as an initializer list
+            (`llvm::SmallVector<int> v = {0, 1};)`. Never use it with `auto`
```
zygoloid:
```suggestion
            (`llvm::SmallVector<int> v = {0, 1};`), `std::pair` or `std::tuple`. Never use it with `auto`
```

docs/project/cpp_style_guide.md
```diff
@@ -124,17 +124,27 @@ these.
...
+-   For initialization:
     -   Use assignment syntax (`=`) when initializing directly with the intended
         value (or with a braced initializer directly specifying that value).
-    -   Use the traditional constructor syntax (with parentheses) when the
-        initialization is performing some active logic, rather than simply
-        composing values together.
-    -   Use `{}` initialization without the `=` only if the above options don't
+    -   Prefer braced initialization for aggregate initialization, such as
+        structs, pairs, and initializer lists.
+        -   Use designated initializers (`{.a = 1}`) when possible for structs,
```
zygoloid, line 132: Just for posterity: I would prefer to say "Prefer to use" rather than simply "Use" here, both given that we don't currently do this and that I think it'll be needlessly verbose in some cases, so I'd prefer to leave the choice to the author and reviewer. But I believe this correctly reflects the consensus we reached :)

zygoloid (approved):

chandlerc (approved):
LGTM as well, and thanks for updating the guidance!

PR #4011, opened 2024-05-31
Switch eval.cpp struct init style.

jonmeow:
Adding designators in general consistent with #4009

With MakeConstantResult, leaning towards keeping `SemIR::BoolLiteral{...}` rather than templating `MakeConstantResult` in order to avoid having multiple `MakeConstantResult` implementations stamped out. But with `TypedInt`, removing the name since it's not necessary for compile.

Note, will be updating other files separately, just starting here since MakeConstantResult might be a bit of an edge case (with AddInst, I might go slightly more for a wrapper).


zygoloid (approved):
Given the direction you ended up going in for `AddInst`, I wonder if it'd make sense to make `MakeConstantResult` a template too?

This is a good incremental step regardless -- I'm happy for this to land as-is and for that to be considered in a later change.

PR #4012, opened 2024-05-31
Switch AddInst struct init style.

jonmeow:
Trying to conform with #4009. Changes SemIR::LocIdAndInst construction to root out struct init cases with AddInst and related functions. I'm using templating of AddInst functions in order to avoid `AddInst(loc_id, InstName{...})` and instead have `AddInst<InstName>(loc_id, {...})` with I think similar readability results. There are a couple cases where inst construction is templated and so designated initializers couldn't be used, so this may be better for those in particular due to the extra type enforcement.

This probably doesn't clean up every last case, but I was trying to get the bulk at once without bleeding over into less related changes.

### DIFF
--- a/toolchain/check/handle_impl.cpp
+++ b/toolchain/check/handle_impl.cpp
@@ -208,8 +208,8 @@ static auto BuildImplDecl(Context& context, Parse::AnyImplDeclId node_id)
   // TODO: Does lookup in an impl file need to look for a prior impl declaration
   // in the api file?
   auto impl_id = context.impls().LookupOrAdd(self_type_id, constraint_type_id);
-  auto impl_decl = SemIR::ImplDecl{impl_id, decl_block_id};
-  auto impl_decl_id = context.AddInst({node_id, impl_decl});
+  SemIR::ImplDecl impl_decl = {impl_id, decl_block_id};
+  auto impl_decl_id = context.AddInst(node_id, impl_decl);
 
   // For an `extend impl` declaration, mark the impl as extending this `impl`.
   if (context.decl_state_stack().innermost().modifier_set.HasAnyOf(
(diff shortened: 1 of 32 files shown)


toolchain/check/handle_impl.cpp
```diff
@@ -208,8 +208,8 @@ static auto BuildImplDecl(Context& context, Parse::AnyImplDeclId node_id)
   // TODO: Does lookup in an impl file need to look for a prior impl declaration
   // in the api file?
   auto impl_id = context.impls().LookupOrAdd(self_type_id, constraint_type_id);
-  auto impl_decl = SemIR::ImplDecl{impl_id, decl_block_id};
-  auto impl_decl_id = context.AddInst({node_id, impl_decl});
+  SemIR::ImplDecl impl_decl = {impl_id, decl_block_id};
```
zygoloid:
```suggestion
  SemIR::ImplDecl impl_decl = {.impl_id = impl_id, .decl_block_id = decl_block_id};
```

toolchain/sem_ir/inst.h
```diff
@@ -305,38 +305,11 @@ struct LocIdAndInst {
...
-  LocIdAndInst(decltype(InstT::Kind)::TypedNodeId node_id, InstT inst)
-      : loc_id(node_id), inst(inst) {}
-
-  // If TypedNodeId is Parse::NodeId, allow construction with a LocId.
-  // TODO: This is somewhat historical due to fetching the NodeId from insts()
-  // for things like Temporary; should we require Untyped in these cases?
-  template <typename InstT>
-    requires(std::same_as<typename decltype(InstT::Kind)::TypedNodeId,
-                          Parse::NodeId>)
-  LocIdAndInst(LocId loc_id, InstT inst) : loc_id(loc_id), inst(inst) {}
```
zygoloid: If I'm reading this right: it looks like we no longer have any compile-time enforcement that the node kind associated with an instruction matches the node kind specified in the instruction definition. Is that correct? That seems unfortunate; is it necessary?

jonmeow: Sure, I can do that approach instead. Restored.

zygoloid (approved):
I think we could remove some more of the `LocIdAndInst` constructor calls by adding a templated overload for `AddPlaceholderInst` and `AddPlaceholderInstInNoBlock`, but let's land this first.

PR #4013, opened 2024-05-31
Clean up some misc toolchain braced inits.

jonmeow:
Following up on #4012 and #4009, clean scattered cases which could be making better use of designated initializers.


jonmeow:
Depends on #4012

zygoloid (approved):

PR #4014, opened 2024-05-31
Switch init style in import_ref.cpp

jonmeow:
This file has a cluster of struct inits, so splitting apart from other edits related to #4009.


jonmeow:
Depends on #4013

jonmeow:
Minor note, rebased to drop the dependency.

zygoloid (approved):

PR #4016, opened 2024-05-31
Add typed forms of `TokenIndex` for tokens with a known kind.

zygoloid:
Instead of adding a `Token<Kind>` field to typed parse nodes that contains a `TokenIndex`, make the parse node directly hold a `TokenIndex` of a more specific kind.

### DIFF
--- a/toolchain/parse/typed_nodes.h
+++ b/toolchain/parse/typed_nodes.h
@@ -73,18 +73,16 @@ struct LeafNode {
 // `struct Aggregate { T x; U y; }`,
 //   to match children with types `T` and `U`.
 //
-// Each parse node should also have exactly one field of type `Lex::*TokenIndex`
-// that matches the token corresponding to the parse node itself. This is not a
-// child node in the parse tree, but specifies the token kind associated with
-// the parse node. The location of the field relative to the child nodes
-// indicates the location within the corresponding grammar production where the
-// token appears.
-//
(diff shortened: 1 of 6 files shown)


toolchain/parse/typed_nodes.h
```diff
@@ -72,6 +72,19 @@ struct LeafNode {
...
+// child node in the parse tree, but specifies the token kind associated with
+// the parse node. The location of the field relative to the child nodes
+// indicates the location within the corresponding grammar production where the
+// token appears.
+//
+// There should be exactly one `Lex::*TokenIndex` field in each parse node, and
+// it should always have the name `token`. In the case where a parse node can
+// correspond to any kind of token, `Lex::TokenIndex` can be used instead. This
+// should only be used when the node kind is either not used in a finished tree,
+// such as `Placeholder`, or is always invalid, such as `InvalidParse`.
```
jonmeow: This looks duplicative: both paragraphs start with a statement that each needs to have one `TokenIndex` field. The latter paragraph specifies it needs to be `token`, the former specifies it's not part of the parse tree. Should the paragraphs be unified, or one deleted?

jonmeow (approved):

toolchain/parse/typed_nodes.h
zygoloid: Yeah, there's some redundancy here. Rephrased.

PR #4018, opened 2024-06-02
Enable a couple of boring warnings.

chandlerc:
Just spotted these while looking at warnings that seem to fire on our
code are probably are things we'd fix if we saw them. None of these seem
important FWIW.

Also removes a redundant flag that is part of `-Wall`.

I have a follow-up for the high-value warning I spotted that motivated
me to look at all of this. But it's noisy so kept it as a separate PR.

### DIFF
--- a/bazel/cc_toolchains/clang_cc_toolchain_config.bzl
+++ b/bazel/cc_toolchains/clang_cc_toolchain_config.bzl
@@ -141,7 +141,8 @@ def _impl(ctx):
                             "-Wself-assign",
                             "-Wimplicit-fallthrough",
                             "-Wctad-maybe-unsupported",
-                            "-Wdelete-non-virtual-dtor",
+                            "-Wextra-semi",
+                            "-Wzero-as-null-pointer-constant",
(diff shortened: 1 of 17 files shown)


bazel/cc_toolchains/clang_cc_toolchain_config.bzl
```diff
@@ -141,10 +141,23 @@ def _impl(ctx):
                             "-Wself-assign",
                             "-Wimplicit-fallthrough",
                             "-Wctad-maybe-unsupported",
-                            "-Wdelete-non-virtual-dtor",
```
zygoloid, line 144: (For posterity: this is included in `-Wall` so had no effect.)

zygoloid (approved):

bazel/cc_toolchains/clang_cc_toolchain_config.bzl
chandlerc: Thanks, noted in the PR description.

PR #4019, opened 2024-06-02
Add `-Wmissing-prototypes` and fix issues it finds.

chandlerc:
Most of these are places where we failed to include a header file and
simply never got an error about this. The fix is to include the header
file.

Most other cases are functions that should have been marked `static` but
were not. Finding all of these was a main motivation for me enabling the
warning despite how much work it is.

One complicating factor was that we weren't including the `handle.h` for
all the state-based handler functions. While this isn't a tiny amount of
code, it is just declarations and doesn't add any extra dependencies. It
also lets us have the checking for which functions need to be `static`
and which don't. For the `parse` library I had to add the `handle.h`
header as well, I tried to match the design of it in `check`.

I have also had to work around a bug in the warning, but given the value
it seems to be providing, that seems reasonable. I've filed the bug
upstream: https://github.com/llvm/llvm-project/issues/94138

I also had to use some hacks to work around limitations of Bazel rules
that wrap `cc_library` rules and don't expose `copts`. I filed a bug for
`cc_proto_library` specifically:
~https://github.com/bazelbuild/bazel/issues/22610~ 
https://github.com/bazelbuild/bazel/issues/4446

### DIFF
--- a/explorer/fuzzing/BUILD
+++ b/explorer/fuzzing/BUILD
@@ -110,6 +110,7 @@ cc_fuzz_test(
     name = "explorer_fuzzer",
     size = "small",
     srcs = ["explorer_fuzzer.cpp"],
+    copts = ["-Wno-missing-prototypes"],
     corpus = glob(["fuzzer_corpus/*"]),
     shard_count = 8,
     tags = [
--- a/explorer/fuzzing/BUILD
+++ b/explorer/fuzzing/BUILD
@@ -110,6 +110,8 @@ cc_fuzz_test(
     name = "explorer_fuzzer",
     size = "small",
     srcs = ["explorer_fuzzer.cpp"],
+    # Uses a macro-based fuzzing library that triggers this warning. And
+    # explorer is currently feature frozen, so just disable the warning.
     copts = ["-Wno-missing-prototypes"],
     corpus = glob(["fuzzer_corpus/*"]),
     shard_count = 8,
--- a/toolchain/parse/handle.h
+++ b/toolchain/parse/handle.h
@@ -2,18 +2,17 @@
 // Exceptions. See /LICENSE for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+#ifndef CARBON_TOOLCHAIN_PARSE_HANDLE_H_
+#define CARBON_TOOLCHAIN_PARSE_HANDLE_H_
+
 #include "toolchain/parse/context.h"
-#include "toolchain/parse/handle.h"
 
 namespace Carbon::Parse {
 
-auto HandlePattern(Context& context) -> void {
-  context.PopAndDiscardState();
-  if (context.PositionKind() == Lex::TokenKind::OpenParen) {
-    context.PushState(State::PatternListAsTuple);
-  } else {
-    context.PushState(State::BindingPattern);
-  }
-}
+// Declare handlers for each parse state.
+#define CARBON_PARSE_STATE(Name) auto Handle##Name(Context& context) -> void;
+#include "toolchain/parse/state.def"
 
 }  // namespace Carbon::Parse
+
+#endif  // CARBON_TOOLCHAIN_PARSE_HANDLE_H_
--- a/toolchain/parse/parse_fuzzer.cpp
+++ b/toolchain/parse/parse_fuzzer.cpp
@@ -6,6 +6,7 @@
 #include <cstring>
 
 #include "llvm/ADT/StringRef.h"
+#include "testing/fuzzing/libfuzzer.h"
 #include "toolchain/base/value_store.h"
 #include "toolchain/diagnostics/null_diagnostics.h"
 #include "toolchain/lex/lex.h"
--- a/utils/treesitter/src/scanner.c
+++ b/utils/treesitter/src/scanner.c
@@ -10,6 +10,12 @@ enum TokenType {
   STRING,
 };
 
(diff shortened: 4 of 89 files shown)


toolchain/parse/handle.h
```diff
@@ -0,0 +1,18 @@
...
+
+#ifndef CARBON_TOOLCHAIN_PARSE_HANDLE_H_
+#define CARBON_TOOLCHAIN_PARSE_HANDLE_H_
+
+#include "toolchain/parse/context.h"
+
+namespace Carbon::Parse {
+
+// Declare handlers for each parse state.
+#define CARBON_PARSE_STATE(Name) auto Handle##Name(Context& context) -> void;
```
zygoloid, line 13: @jonmeow I think this is worth discussion to make sure we're all on the same page:

Previously, we defined these functions in the relevant `handle_*.cpp` files with no forward declaration, which I think was *probably* intentional. Adding the forward declarations means we get the substantial advantage of being able to detect missing `static`, but also means the `Handle*` functions are all in scope in all `handle_*.cpp` files, whereas we probably don't want them to be able to invoke each other.

Similarly in `check/`, where we had the header file already but didn't include it from the `handle_*.cpp` files.

I think this change is a good thing nonetheless. (It would also catch issues where the return type of a `Handle*` function is wrong.)

chandlerc: Now that you remind, me I remember this being intentional. Sorry I forgot about that.

But it is definitely something to discuss -- not at all sure what the exact balance of tradeoffs is here. I was honestly surprised how many real issues `-Wmissing-prototypes` found (completely outside of any `handle.h` stuff) where we forgot headers and didn't use `static`. So mostly sending this to see if there is a good pattern we could follow here. Not at all attached to specifically having all the `Handle*` functions in one header though, very happy if we want some different pattern or way of approaching this.

utils/treesitter/src/scanner.c
```diff
@@ -10,6 +10,10 @@ enum TokenType {
   STRING,
 };
 
+#pragma clang diagnostic push
+// The tree_sitter build uses C functions without prototypes.
+#pragma clang diagnostic ignored "-Wmissing-prototypes"
```
jonmeow, line 17: As an alternative, can you put this in copts in the BUILD file for the library?

toolchain/parse/parse_fuzzer.cpp
```diff
@@ -13,6 +13,12 @@
 
 namespace Carbon::Testing {
 
+// Declare this first to work around a bug in `-Wmissing-prototypes`:
+// https://github.com/llvm/llvm-project/issues/94138
+// NOLINTNEXTLINE: Match the documented fuzzer entry point declaration style.
+extern "C" int LLVMFuzzerTestOneInput(const unsigned char* data,
+                                      std::size_t size);
```
jonmeow: This comes up in several files, can we just make our own header with this?

toolchain/parse/handle.h
jonmeow: I think there's probably more benefit to prototype matching than avoiding the risk of recursive calls this way. I'd probably have made these functions on a class if I'd been allowed to, just to get the declaration match checking (it's a nuisance when a typo on names is only detected at link time, and I'm a little surprised this didn't seem to find any unused code where a state had been deleted but the handler left behind). Not needing to comment for `static` as much will be nice, too.

explorer/fuzzing/BUILD
```diff
@@ -110,6 +110,7 @@ cc_fuzz_test(
     name = "explorer_fuzzer",
     size = "small",
     srcs = ["explorer_fuzzer.cpp"],
+    copts = ["-Wno-missing-prototypes"],
```
jonmeow, line 115: Why this approach here, versus what's done for other fuzzers?

jonmeow (approved):

utils/treesitter/src/scanner.c
chandlerc: No, this is also part of the `tree_sitter_cc_library` that doesn't have `copts`. I'll add a comment explaining why the weird workaround.

toolchain/parse/parse_fuzzer.cpp
chandlerc: At first, avoiding a header seemed cleaner. But you're right, we use this too many places. And I have another thing to add to that header. Creating a header for our fuzzers.

explorer/fuzzing/BUILD
chandlerc: Because the explorer uses macros and such to instantiate the proto-based fuzzer. I didn't want to guess at what the right or wrong declarations would be, so I just turned off the warning. It also seemed more in line with disabling clang-tidy before spending more time trying to figure out a principled solution.

chandlerc:
PTAL and lemme know if this is good, or a different approach. Happy to also look into class methods or something else if more appropriate.

explorer/fuzzing/BUILD
jonmeow: Thanks for explaining, would it be worth adding a small comment to this effect?

jonmeow (approved):
Note there's still a build failure, maybe the changes caused a clang-tidy issue to be flagged?

But feel free to merge once that's fixed.

explorer/fuzzing/BUILD
chandlerc: Adding comment.

PR #4020, opened 2024-06-03
Rename enclosing scope to parent scope

jonmeow:
Following up on discussion from #3948, doing a general rename of "enclosing scope" to "parent scope" (and "enclosing scopes" to "ancestor scopes"). The intent is to improve understandability and collide less with C++ terminology for "enclosing scope". Note this changes most uses of "enclosing", but leaves behind a few like "enclosing function" and "enclosing block".

Note this does create some "parent class" mentions for "adapt" and "var" (the class they're within), which is maybe unfortunate, but we'd probably say "base class" if we meant inheritance so perhaps that's okay. Along the same lines, these are the only `parent_class` uses I see now, and we do have a few `base_class`.

### DIFF
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -444,7 +444,7 @@ class DeferredDefinitionWorklist {
     std::optional<DeclNameStack::SuspendedName> suspended_name;
     // Whether this scope is itself within an outer deferred definition scope.
     // If so, we'll delay processing its contents until we reach the end of the
-    // enclosing scope. For example:
+    // parent scope. For example:
     //
     // ```
     // class A {
@@ -497,10 +497,10 @@ class DeferredDefinitionWorklist {
   // Push a task to re-enter a function scope, so that functions defined within
   // it are type-checked in the right context.
   auto PushEnterDeferredDefinitionScope(Context& context) -> void {
-    bool nested = !enclosing_scopes_.empty() &&
-                  enclosing_scopes_.back().scope_index ==
+    bool nested = !ancestor_scopes_.empty() &&
+                  ancestor_scopes_.back().scope_index ==
                       context.decl_name_stack().PeekInitialScopeIndex();
-    enclosing_scopes_.push_back(
+    ancestor_scopes_.push_back(
         {.worklist_start_index = worklist_.size(),
          .scope_index = context.scope_stack().PeekIndex()});
     worklist_.push_back(
@@ -543,7 +543,7 @@ class DeferredDefinitionWorklist {
 
   // CHECK that the work list has no further work.
   auto VerifyEmpty() {
-    CARBON_CHECK(worklist_.empty() && enclosing_scopes_.empty())
+    CARBON_CHECK(worklist_.empty() && ancestor_scopes_.empty())
         << "Tasks left behind on worklist.";
   }
 
@@ -558,21 +558,21 @@ class DeferredDefinitionWorklist {
   llvm::SmallVector<Task, 0> worklist_;
 
   // A deferred definition scope that is currently still open.
-  struct EnclosingScope {
+  struct AncestorScope {
     // The index in worklist_ of the EnterDeferredDefinitionScope task.
     size_t worklist_start_index;
     // The corresponding lexical scope index.
     ScopeIndex scope_index;
   };
 
-  // The deferred definition scopes enclosing the current checking actions.
-  llvm::SmallVector<EnclosingScope> enclosing_scopes_;
+  // The deferred definition scopes for the current checking actions.
+  llvm::SmallVector<AncestorScope> ancestor_scopes_;
 };
 }  // namespace
 
 auto DeferredDefinitionWorklist::SuspendFinishedScopeAndPush(Context& context)
     -> bool {
-  auto start_index = enclosing_scopes_.pop_back_val().worklist_start_index;
+  auto start_index = ancestor_scopes_.pop_back_val().worklist_start_index;
 
   // If we've not found any deferred definitions in this scope, clean up the
   // stack.
--- a/toolchain/check/check.cpp
+++ b/toolchain/check/check.cpp
@@ -497,10 +497,10 @@ class DeferredDefinitionWorklist {
   // Push a task to re-enter a function scope, so that functions defined within
   // it are type-checked in the right context.
   auto PushEnterDeferredDefinitionScope(Context& context) -> void {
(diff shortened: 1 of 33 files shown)


toolchain/check/check.cpp
```diff
@@ -444,7 +444,7 @@ class DeferredDefinitionWorklist {
     std::optional<DeclNameStack::SuspendedName> suspended_name;
     // Whether this scope is itself within an outer deferred definition scope.
     // If so, we'll delay processing its contents until we reach the end of the
-    // enclosing scope. For example:
+    // parent scope. For example:
```
zygoloid, line 447: We mean the lexically-enclosing scope rather than necessarily the parent scope. For example:
```
class A {
  class B {
    class C;
  }
  class B.C {
    fn F() {}
  }
}
```
The scope of `A.B.C` is within an outer deferred definition scope -- namely, the scope of `A` -- which is not the parent scope (which would be `A.B`). How about:
```suggestion
    // outer scope. For example:
```

toolchain/check/context.cpp
```diff
@@ -252,10 +252,10 @@ auto Context::LookupUnqualifiedName(Parse::NodeId node_id,
                                     SemIR::NameId name_id) -> SemIR::InstId {
   // TODO: Check for shadowed lookup results.
 
-  // Find the results from enclosing lexical scopes. These will be combined with
+  // Find the results from ancestor lexical scopes. These will be combined with
   // results from non-lexical scopes such as namespaces and classes.
   auto [lexical_result, non_lexical_scopes] =
-      scope_stack().LookupInEnclosingScopes(name_id);
+      scope_stack().LookupInAcestorScopes(name_id);
```
zygoloid: Typo `Acestor`

toolchain/check/decl_name_stack.cpp
```diff
@@ -88,19 +88,19 @@ auto DeclNameStack::Suspend() -> SuspendedName {
       << "Missing call to FinishName before Suspend";
   SuspendedName result = {.name_context = decl_name_stack_.pop_back_val(),
                           .scopes = {}};
-  auto enclosing_index = result.name_context.initial_scope_index;
+  auto parent_index = result.name_context.initial_scope_index;
```
zygoloid: Maybe `initial_index`? This isn't necessarily a parent of any particular scope.

toolchain/parse/context.h
```diff
@@ -376,8 +376,7 @@ class Context {
 
   // The deferred definition indexes of functions whose definitions have begun
   // but not yet finished.
-  llvm::SmallVector<DeferredDefinitionIndex>
-      enclosing_deferred_definition_stack_;
+  llvm::SmallVector<DeferredDefinitionIndex> parent_deferred_definition_stack_;
```
zygoloid: I think "outer" or "enclosing" is clearer for purely-lexical considerations like this -- we're just talking about where the tokens appear, not a semantic relationship between scopes here. Actually, can we remove the prefix entirely:

```suggestion
  llvm::SmallVector<DeferredDefinitionIndex> deferred_definition_stack_;
```

toolchain/check/check.cpp
```diff
@@ -558,21 +558,21 @@ class DeferredDefinitionWorklist {
   llvm::SmallVector<Task, 0> worklist_;
 
   // A deferred definition scope that is currently still open.
-  struct EnclosingScope {
+  struct AncestorScope {
```
zygoloid: I think it'd make sense to call this `OpenScope` or just `Scope`. Calling it "ancestor" makes me wonder what it's an ancestor of.

toolchain/check/check.cpp
```diff
@@ -558,21 +558,21 @@ class DeferredDefinitionWorklist {
...
     // The index in worklist_ of the EnterDeferredDefinitionScope task.
     size_t worklist_start_index;
     // The corresponding lexical scope index.
     ScopeIndex scope_index;
   };
 
-  // The deferred definition scopes enclosing the current checking actions.
-  llvm::SmallVector<EnclosingScope> enclosing_scopes_;
+  // The deferred definition scopes for the current checking actions.
+  llvm::SmallVector<AncestorScope> ancestor_scopes_;
```
zygoloid: Likewise, maybe `open_scopes_` or `entered_scopes_`?

zygoloid (commented):
Generally I think this is an improvement. I think "parent" isn't working for me for a few places that are more about the parse tree and relations between tokens than about semantic parent/child relationships. I've given suggestions in those places.

toolchain/check/context.cpp
jonmeow: Done

toolchain/parse/context.h
jonmeow: Done

toolchain/check/check.cpp
jonmeow: As above

toolchain/check/check.cpp
jonmeow: How about `EnteredScope` / `entered_scopes_`? `Scope` seems a little too likely to conflict with a name elsewhere, "Open" doesn't quite feel right to me (implies namespace-as-open-scope in my head), sounds like "Entered" might be okay from below.

toolchain/check/decl_name_stack.cpp
jonmeow: Done with `scope_index` because I feel better removing the first word than the middle one, and this is just a shorthand local variable anyways.

toolchain/check/check.cpp
jonmeow: Where I have LookupInAncestorScopes, would that be better talking about entered scopes too?

zygoloid: `EnteredScope` looks good to me.

Hm. So, `LookupInAncestorScopes` performs name lookup in only the lexically-enclosing scopes that lack a `NameScopeId`, and also returns the list of ancestor scopes with a `NameScopeId` that might shadow the lexical lookup result. Maybe something like `LookupInLexicalScopes` might make sense? (`LookupInLexicalScopesAndFindNonLexicalScopes` would be more accurate but kinda long. But it only has one caller so maybe that's OK?)

zygoloid (approved):
Looks good, not merging in case you want to pick a different name for `LookupInAncestorScopes`.

toolchain/check/check.cpp
jonmeow: Going with LookupInLexicalScopes

PR #4023, opened 2024-06-03
Disable autoupdate in explorer tests.

jonmeow:
Just replacing AUTOUPDATE with NOAUTOUPDATE, and removing the autoupdate script. Tests will still run, but autoupdate may need to be fixed if significant changes are made.

I noticed this while trying to autoupdate for #4007 (because we verify that tests have been autoupdated). Autoupdate was likely broken by #3449.

Trying to run with no changes gives:

```
CHECK failure at testing/file_test/file_test_base.cpp:801: !absl::GetFlag(FLAGS_test_targets_file).empty(): Missing --test_targets_file.
```

This is because the `file_test` rule creates a file with test inputs that it runs with, which the prebuilt binary doesn't provide.

A local kludge to create a `file_test` target not using prebuilt_binary showed another error:

```
: CommandLine Error: Option ': CommandLine Error: Option 'parser_debug' registered more than once!
LLVM ERROR: inconsistency in registered CommandLine options
trace_phase' registered more than once!
LLVM ERROR: inconsistency in registered CommandLine options
```

I'm thinking here that the explorer code hasn't been in enough use, and we're seeing some rot as a consequence. Rather than trying to maintain it, I'm suggesting to go with NOAUTOUPDATE.


chandlerc (approved):
Assuming this is mechanical, LGTM.

jonmeow:
Yeah, one big search-and-replace. And then fixing where I accidentally did NOAUTOUPDATE -> NONOAUTOUPDATE

PR #4026, opened 2024-06-04
Add namespace-scoped private keyword support.

jonmeow:
Adds access to the name lookup table in name scopes. This is so that we can quickly check access during name lookup without resolving the entity itself. Does this for names in general, but does not implement handling for entity-scoped names, only namespace-scoped names (where they're essentially just not exported).

Excludes `private` names from exports. Although names should be accessible to `impl` files, that's not implemented here because we'll probably want to do it by directly copying name lookup tables.


jonmeow:
I think Richard convinced me to try a different approach.

josh11b (approved):

PR #4027, opened 2024-06-04
Switch decl_state to an arg for modifier functions

jonmeow:
With private modifiers, we'll want to start checking modifiers later, e.g. after a name conflict is detected (and potentially merged). I think we've agreed to be more explicit about whether the modifier functions are manipulating state, versus trying to keep the state on the stack a little longer (moving Pop to the end of these functions).

Removing FileScope from the stack and renaming it to DeclIntroducerStack to better reflect the usage and behavior. The FileScope mostly reflects an approach that wasn't ultimately adopted.

### DIFF
--- a/toolchain/check/BUILD
+++ b/toolchain/check/BUILD
@@ -66,8 +66,8 @@ cc_library(
     hdrs = [
         "context.h",
         "convert.h",
+        "decl_introducer_state.h",
         "decl_name_stack.h",
-        "decl_state.h",
         "diagnostic_helpers.h",
         "eval.h",
         "function.h",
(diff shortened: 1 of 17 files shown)


zygoloid (approved):
How would you feel about:
- Renaming `decl_state` to `introducer` throughout
- Removing the `State` from `DeclIntroducerState`

Happy for the answer to either to be "no" or "not in this PR" :)

jonmeow:
I'll do the `introducer` rename, but I consider `DeclIntroducerState` as state that's being incrementally built, it's just that it's near-complete when popped.

PR #4028, opened 2024-06-04
For modifiers, get the TokenKind from DeclKind instead of argument.

jonmeow:


jonmeow:
Depends on #4027

zygoloid (approved):
Nice!

The list in `DeclKindToken` makes me wonder if we should add an xmacro in `token_kind.def` for introducers at some point.

jonmeow:
I'd been considering a x-macro for the decl introducer (only), but yeah, something in tokens may make more sense and avoid the extra burden. Maybe I'll look at that separately.

PR #4029, opened 2024-06-05
Support out-of-line definitions of members of generic classes and interfaces.

zygoloid:
Check the parameters specified in a name qualifier against the parameters of the entity that the qualifier refers to.

For interfaces, this required adding minimal support for parameterized interface names.


toolchain/check/testdata/class/generic/member_out_of_line.carbon
```diff
@@ -0,0 +1,321 @@
...
+
+class A(T:! type) {
+  class B(N:! T) {
+    fn F[self: Self](a: T);
+  }
+}
+
+fn A(T:! type).B(N:! T).F[self: Self](a: T) {}
+
+// --- fail_mismatched.carbon
```
jonmeow: Would it be worth breaking this apart a little so that each error is tested separately? That'd play into the `fail_` naming ensuring each individual diagnostic occurs, versus at least one diagnostic amongst all four cases.

toolchain/check/decl_name_stack.cpp
```diff
@@ -179,24 +179,45 @@ auto DeclNameStack::LookupOrAddName(NameContext name_context,
...
-auto DeclNameStack::ApplyNameQualifier(const NameComponent& name) -> void {
-  if (name.implicit_params_id.is_valid() || name.params_id.is_valid()) {
-    context_->TODO(name.params_loc_id, "name qualifier with parameters");
-  }
+// Push a scope corresponding to a name qualifier. For example, for
+//
+//   fn Class(T:! type).F(n: i32)
+//
+// we will push the scope for `Class(T:! type)` between the scope containing the
+// declaration of `T` and the scope containing the declaration of `n`.
```
```suggestion
// Push a scope corresponding to a name qualifier. For example, for
// `fn Class(T:! type).F(n: i32)` we will push the scope for `Class(T:! type)`
// between the scope containing the declaration of `T` and the scope
// containing the declaration of `n`.
```

Optional suggestion, just thought it might be easier to read this way.

toolchain/check/decl_name_stack.cpp
```diff
@@ -215,140 +236,71 @@ auto DeclNameStack::ApplyAndLookupName(NameContext& name_context,
...
+    case NameContext::State::Error:
+      // Already in an error state, so return without examining.
+      return SemIR::NameScopeId::Invalid;
+  }
+
+  auto new_params =
+      DeclParams(name.name_loc_id, name.implicit_params_id, name.params_id);
+
+  // Find the scope corresponding to the resolved instruction.
+  CARBON_KIND_SWITCH(context_->insts().Get(name_context.resolved_inst_id)) {
```
jonmeow, line 347: Combining multiple functions like this is making it pretty long (110 LOC), is it worth thinking about how to keep it broken up a little? (partly I'm looking at it this way because it's now two switches)

You've previously suggested putting diagnostics in their own little helper functions, perhaps that would offer a few good cut points?

jonmeow (approved):
Looks good, it's nice to see ResolvedNonScope go away.

toolchain/check/testdata/class/generic/member_out_of_line.carbon
zygoloid: Done.

toolchain/check/decl_name_stack.cpp
zygoloid: Done.

jonmeow:
Looks good!

PR #4030, opened 2024-06-05
Teach fuzzers to initialize their installs from runfiles.

chandlerc:
This also allows us to default construct installs in an error state, which is useful for cases like fuzzers where we want to default construct a global, but then re-initilaize it That said, happy to consider alternative designs here.

### DIFF
--- a/toolchain/check/check_fuzzer.cpp
+++ b/toolchain/check/check_fuzzer.cpp
@@ -11,15 +11,13 @@
 
 namespace Carbon::Testing {
 
-static InstallPaths install_paths;
+static const InstallPaths* install_paths = nullptr;
 
 // NOLINTNEXTLINE(readability-non-const-parameter): External API required types.
 extern "C" auto LLVMFuzzerInitialize(int* argc, char*** argv) -> int {
-  std::string exe_path;
-  if (*argc >= 1) {
-    exe_path = FindExecutablePath((*argv)[0]);
-    install_paths = InstallPaths::MakeForBazelRunfiles(exe_path);
-  }
+  CARBON_CHECK(*argc >= 1) << "Need the `argv[0]` value to initialize!";
+  install_paths = new InstallPaths(
+      InstallPaths::MakeForBazelRunfiles(FindExecutablePath((*argv)[0])));
   return 0;
 }
 
@@ -41,7 +39,7 @@ extern "C" int LLVMFuzzerTestOneInput(const unsigned char* data,
                                        /*RequiresNullTerminator=*/false)));
 
   llvm::raw_null_ostream null_ostream;
-  Driver driver(fs, &install_paths, null_ostream, null_ostream);
+  Driver driver(fs, install_paths, null_ostream, null_ostream);
 
   // TODO: Get checking to a point where it can handle invalid parse trees
   // without crashing.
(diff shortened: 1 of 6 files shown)


toolchain/check/check_fuzzer.cpp
```diff
@@ -10,6 +11,18 @@
 
 namespace Carbon::Testing {
 
+static InstallPaths install_paths;
```
jonmeow: Is this fully conformant with static init requirements? I'm surprised you're not structuring this more as a pointer or similar POD type.

toolchain/check/check_fuzzer.cpp
```diff
@@ -10,6 +11,18 @@
...
+
+// NOLINTNEXTLINE(readability-non-const-parameter): External API required types.
+extern "C" auto LLVMFuzzerInitialize(int* argc, char*** argv) -> int {
+  std::string exe_path;
+  if (*argc >= 1) {
+    exe_path = FindExecutablePath((*argv)[0]);
```
jonmeow:
```suggestion
  if (*argc >= 1) {
    std::string exe_path = FindExecutablePath((*argv)[0]);
```

Would this make sense to be clear on the lifetime?

toolchain/driver/driver_fuzzer.cpp
```diff
@@ -15,6 +16,18 @@
 
 namespace Carbon::Testing {
 
+static InstallPaths install_paths;
+
+// NOLINTNEXTLINE(readability-non-const-parameter): External API required types.
+extern "C" auto LLVMFuzzerInitialize(int* argc, char*** argv) -> int {
+  std::string exe_path;
+  if (*argc >= 1) {
+    exe_path = FindExecutablePath((*argv)[0]);
```
jonmeow:
```suggestion
  if (*argc >= 1) {
    std::string exe_path = FindExecutablePath((*argv)[0]);
```

jonmeow (approved):
Looks good, this approach makes sense.

toolchain/install/install_paths.h
```diff
@@ -82,6 +82,9 @@ class InstallPaths {
   // using Carbon in an environment with an unusual path to the installed files.
   static auto Make(llvm::StringRef install_prefix) -> InstallPaths;
 
+  // Constructor only used in errors or globals prior to initialization.
+  InstallPaths() { SetError("No prefix provided!"); }
```
jonmeow: Should this be set up as constant initialization?
https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

chandlerc: Yeah, I was not totally satisfied with this. But I think actually your suggestion of a pointer is even better, I'm going to switch to that I think...

toolchain/check/check_fuzzer.cpp
chandlerc: No, the pointer is a better approach, thanks for suggesting. I've switched to that.

toolchain/check/check_fuzzer.cpp
chandlerc: No idea what I was thinking here, sorry. =] But this is also obviated a bit with new code pattern.

chandlerc (commented):
PTAL, new, simpler approach!

jonmeow (approved):
LGTM. :)

PR #4031, opened 2024-06-05
Add a macro for introducer tokens.

jonmeow:
Playing with the macro suggestion on #4028, replace DeclKind with some templating that asserts the in-use token is an introducer token. This allows type-safe usage of Lex::TokenKind, reducing the benefit of a separate enum while improving stdout (since now this will ostream as the keyword name).


jonmeow:
Let me know if this is going a little too far. :)

zygoloid (approved):

PR #4032, opened 2024-06-05
Start unifying let/var handling.

jonmeow:
Merges handle_let.cpp and handle_variable.cpp into a single file in order to take better advantage of commonalities.

Both still have a bunch of kludges, and there's still divergent handling in handle_binding_pattern that will yield different results. However, this fixes some things about `let` like starting to make use of global_init (imperfectly, due to how VarStorage works, I grant), and in particular adding `let` names to a name scope, not just lexical lookup.


jonmeow:
Depends on #4026 [merged]

zygoloid (commented):
Would it be possible to land the merging of `handle_let.cpp` and `handle_variable.cpp` first as a separate PR? I think that'd make this a lot easier to review, so I can just look at the differences between the old and new code rather than reviewing the whole thing effectively from scratch.

jonmeow:
> Would it be possible to land the merging of `handle_let.cpp` and `handle_variable.cpp` first as a separate PR? I think that'd make this a lot easier to review, so I can just look at the differences between the old and new code rather than reviewing the whole thing effectively from scratch.

Done, #4040, rebasing...

jonmeow:
> Done, #4040, rebasing...

[merged]

zygoloid (approved):

PR #4033, opened 2024-06-06
Improve formatting of the type of a function or generic class name.

zygoloid:
Don't format the *type of* the function `F` as `"F"`. Instead use `"<type of F>"`.

Also improve instruction naming for function and generic class name values: name the value after the function or generic class, and add a `.type` suffix to the type.


toolchain/check/testdata/var/no_prelude/decl.carbon
```diff
@@ -15,16 +15,16 @@ fn Main() {
 // CHECK:STDOUT: --- decl.carbon
 // CHECK:STDOUT:
 // CHECK:STDOUT: constants {
-// CHECK:STDOUT:   %Main: type = fn_type @Main [template]
+// CHECK:STDOUT:   %Main.type: type = fn_type @Main [template]
 // CHECK:STDOUT:   %.1: type = tuple_type () [template]
-// CHECK:STDOUT:   %struct: Main = struct_value () [template]
+// CHECK:STDOUT:   %Main: <type of Main> = struct_value () [template]
```
josh11b, line 20: I'm a bit sad that we end up with two ways to name the same thing (`%Main.type` and `<type of Main>`).

josh11b (approved):

toolchain/check/testdata/var/no_prelude/decl.carbon
zygoloid: I'll send a separate PR to address that.

PR #4034, opened 2024-06-06
More support for generic interfaces, aiming for parity with support for generic classes.

zygoloid:
This is mostly mechanically duplicating work done for generic classes to also support generic interfaces.

Also fix both generic interfaces and generic classes to support importing class and interface types with arguments from another file.


zygoloid:
Depends on #4033.

toolchain/sem_ir/inst_namer.cpp
```diff
@@ -441,6 +441,11 @@ auto InstNamer::CollectNamesInBlock(ScopeId scope_id,
         add_inst_name_id(sem_ir_.classes().Get(inst.class_id).name_id, ".type");
         continue;
       }
+      case CARBON_KIND(GenericInterfaceType inst): {
+        add_inst_name_id(sem_ir_.interfaces().Get(inst.interface_id).name_id,
```
jonmeow, line 445: FWIW, it seems like both this and the formatter could benefit from a generic GetNameId(inst) kind of function. There's already a little of that in import.cpp, which also includes the parent scope, maybe we should be thinking about moving that to be a helper in sem_ir?

I think this just adds to the discussion we've had about wondering if there should be better helpers for common features of class/function/interface, though.

(just commenting, out of scope for this change)

toolchain/check/call.cpp
```diff
@@ -42,6 +63,12 @@ auto PerformCall(Context& context, Parse::NodeId node_id,
       return PerformCallToGenericClass(context, node_id,
                                        generic_class->class_id, arg_ids);
     }
+    if (auto generic_interface =
+            context.types().TryGetAs<SemIR::GenericInterfaceType>(
+                context.insts().Get(callee_id).type_id())) {
```
jonmeow: Note here that you're doing twice (with generic_class) the equivalent of: `context.types().GetAsInst(context.insts().Get(callee_id).type_id())` -- one type lookup and two inst lookups. That seems worth refactoring to share, maybe use the CARBON_KIND_SWITCH?

toolchain/check/import_ref.cpp
```diff
@@ -371,6 +371,31 @@ class ImportRefResolver {
...
+    // Import all the values in the block.
+    auto import_block = import_ir_.inst_blocks().Get(import_block_id);
+    llvm::SmallVector<SemIR::InstId> inst_ids;
+    inst_ids.reserve(import_block.size());
+    for (auto import_inst_id : import_block) {
+      auto const_id = GetLocalConstantId(import_inst_id);
+      inst_ids.push_back(const_id.inst_id());
+    }
+
+    if (HasNewWork(initial_work)) {
```
jonmeow: We typically use HasNewWork in helpers, so suggesting a tiny comment to call it out.
```suggestion
    // Only store the block if it's complete.
    if (HasNewWork(initial_work)) {
```

Note though, not typically doing this may be less than ideal -- it seems like `GetLocalParamRefsId` should maybe do similar, rather than adding an incomplete block, for example.

jonmeow (approved):
Looks good. I think the copying here is pretty much what I'd expect.

toolchain/check/import_ref.cpp
jonmeow: Sent #4036

jonmeow: Nevermind, realized it's not needed elsewhere because constants are fetched first. They're written not to do the duplicative work, differently.

That may be something to consider here, though... Maybe it should return the vector, and only add the canonical block if everything's ready? Otherwise, you could end up canonicalizing when a later (or earlier) constant is incomplete. You may still end up with the same ID, but it's still duplicative work.

toolchain/check/eval.cpp
```diff
@@ -1009,7 +1016,16 @@ auto TryEvalInst(Context& context, SemIR::InstId inst_id, SemIR::Inst inst)
           Phase::Template);
     }
     case CARBON_KIND(SemIR::InterfaceDecl interface_decl): {
-      // TODO: Once interfaces have generic arguments, handle them.
+      // If the interface has generic arguments, we don't produce an interface
```
josh11b:
```suggestion
      // If the interface has generic parameters, we don't produce an interface
```

toolchain/sem_ir/inst_namer.cpp
zygoloid: Yeah, more work to factor out the common parts would be great.

toolchain/check/call.cpp
zygoloid: Done.

toolchain/check/import_ref.cpp
zygoloid: Changed to return the vector as suggested.

toolchain/check/eval.cpp
zygoloid: Done, and in the place I copied this from :)

toolchain/check/import_ref.cpp
```diff
@@ -371,29 +371,36 @@ class ImportRefResolver {
...
-    if (HasNewWork(initial_work)) {
+    return inst_ids;
+  }
+
+  // Gets a local canonical instruction block ID corresponding to an imported
+  // inst block whose contents were already imported, for example by
+  // GetLocalInstBlockContents.
+  auto GetLocalCanonicalInstBlockId(
+      SemIR::InstBlockId import_block_id,
+      const llvm::SmallVector<SemIR::InstId>& contents) -> SemIR::InstBlockId {
```
jonmeow: Should this be arrayref? (only thinking of it because I've seen you do it elsewhere)

```suggestion
      llvm::ArrayRef<SemIR::InstId> contents) -> SemIR::InstBlockId {
```

jonmeow (approved):
Thanks! I think the changes look good. One really tiny thing but feel free to merge.

toolchain/check/import_ref.cpp
```diff
@@ -371,6 +371,38 @@ class ImportRefResolver {
...
+
+    return inst_ids;
+  }
+
+  // Gets a local canonical instruction block ID corresponding to an imported
+  // inst block whose contents were already imported, for example by
+  // GetLocalInstBlockContents.
+  auto GetLocalCanonicalInstBlockId(
+      SemIR::InstBlockId import_block_id,
+      llvm::ArrayRef<SemIR::InstId> contents) -> SemIR::InstBlockId {
```
```suggestion
  auto GetLocalCanonicalInstBlockId(SemIR::InstBlockId import_block_id,
                                    llvm::ArrayRef<SemIR::InstId> contents)
      -> SemIR::InstBlockId {
```

PR #4035, opened 2024-06-06
Format types in SemIR using the formatting for their instruction.

zygoloid:
Don't use the pretty-printed type name, because that's intended for
diagnostics, not for a theoretically machine-readable format like SemIR.

Types are always constants, so omit the leading `constant.` on the type
instruction name.


jonmeow (approved):
This looks like a good improvement in output.

PR #4036, opened 2024-06-06
Only store imported blocks when they're complete.

jonmeow:
Thinking about this due to #4034 doing this on a canonicalized block. But it seems worth avoiding the vector copy in general.


jonmeow:
Nevermind, realizing it's already addressed here.

PR #4037, opened 2024-06-06
Get the VS Code extension working again.

zygoloid:
VSCode's JSON parser no longer accepts comments in at least textmate grammars, so remove the comments. It's unclear whether VSCode changed here or whether this hasn't worked since the comments were added.

Clean up the textmate grammar: refactor, add some missing operators and keywords, classify keywords as introducer / modifier as appropriate, fix comment grammar to require space after `//`, and generalize recognition of iN, fN, uN to recognize all such type literals.

Be more cautious in language server -- I was seeing frequent crashes for `IdentifierName` parse nodes whose token was not actually an identifier, presumably due to error recovery.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -221,24 +221,6 @@ repos:
               .*/fuzzer_corpus/.*|
               .*/testdata/.*\.golden
           )$
-      # The copyright header we use for JSON can't be represented as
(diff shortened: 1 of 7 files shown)


jonmeow (approved):
LGTM, I think we've discussed the important details here.

PR #4039, opened 2024-06-07
Refactor SemIRDiagnosticConverter out of check.cpp

jonmeow:
This is mostly just moving the code out. It also changes node_converters from a DenseMap to a SmallVector, taking advantage of CheckIRId.

### DIFF
--- a/toolchain/check/sem_ir_diagnostic_converter.h
+++ b/toolchain/check/sem_ir_diagnostic_converter.h
@@ -17,7 +17,7 @@ namespace Carbon::Check {
 class SemIRDiagnosticConverter : public DiagnosticConverter<SemIRLoc> {
  public:
   explicit SemIRDiagnosticConverter(
-      const llvm::ArrayRef<Parse::NodeLocConverter*> node_converters,
+      llvm::ArrayRef<Parse::NodeLocConverter*> node_converters,
       const SemIR::File* sem_ir)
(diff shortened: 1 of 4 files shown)


jonmeow:
Depends on #4032

toolchain/check/sem_ir_diagnostic_converter.h
```diff
@@ -0,0 +1,49 @@
...
+
+ private:
+  // Converts a node_id corresponding to a specific sem_ir to a diagnostic
+  // location.
+  auto ConvertLocInFile(const SemIR::File* sem_ir, Parse::NodeId node_id,
+                        bool token_only, ContextFnT context_fn) const
+      -> DiagnosticLoc;
+
+  // Converters for each SemIR.
+  const llvm::ArrayRef<Parse::NodeLocConverter*> node_converters_;
```
zygoloid: It seems a bit unusual to mark a non-static data member as `const`. Is this important? (All the member function are `const` so this seems like it'd make no difference other than making this class non-assignable.)

zygoloid (approved):

toolchain/check/sem_ir_diagnostic_converter.h
jonmeow: Removed, I think this was just leftover from the DenseMap pointer.

PR #4042, opened 2024-06-07
Import support for tuple, struct and integer values.

zygoloid:
These can be imported as the argument values for a generic. Also, the name of a generic is modeled as a `StructValue`.


toolchain/check/import_ref.cpp
```diff
@@ -1321,6 +1330,20 @@ class ImportRefResolver {
...
+    auto type_id = GetLocalConstantId(inst.type_id);
+    if (HasNewWork(initial_work)) {
+      return ResolveResult::Retry();
+    }
+    return {.const_id = TryEvalInst(
+                context_, SemIR::InstId::Invalid,
+                SemIR::IntLiteral{
+                    .type_id = context_.GetTypeIdForTypeConstant(type_id),
+                    .int_id = context_.ints().Add(
+                        import_ir_.ints().Get(inst.int_id))})};
```
jonmeow: These calls, as a set, sort of feel like they'd wrap better (reading a little more easily) if you used some locals for intermediate work. You could even do something like:

```suggestion
    SemIR::IntLiteral inst = {
        .type_id = context_.GetTypeIdForTypeConstant(type_id),
        .int_id = context_.ints().Add(import_ir_.ints().Get(inst.int_id))};
    return {.const_id = TryEvalInst(context_, SemIR::InstId::Invalid, inst)};
```

toolchain/sem_ir/ids.h
```diff
@@ -141,6 +141,12 @@ struct ConstantId : public IdBase, public Printable<ConstantId> {
     return InstId(Abs(index) - IndexOffset);
   }
 
+  // Returns the instruction that describes this constant value, or
+  // InstId::Invalid if this is invalid or a runtime value.
+  constexpr auto inst_id_if_valid() const -> InstId {
```
jonmeow, line 146: Feels kind of odd to be adding this for one use, although presumably you'll add more...

I should ask in particular though, why `constexpr`?

jonmeow (approved):
Looks good, please feel free to merge regardless of what you do with formatting.

toolchain/sem_ir/ids.h
zygoloid: I just made this `constexpr` because `inst_id` is, and it seemed reasonable for it to be consistent.

jonmeow: FWIW, I'd thought we were only doing constexpr when it was used that way, rather than simple consistency.

PR #4044, opened 2024-06-10
Add growth API to the new hashtables.

chandlerc:
This adds two different growth APIs. This is instead of the more conventional STL `reserve` method. One allows users that aren't trying to grow in anticipation of an *exact* count of insertions, but generally trying to size the table to the correct ballpark with a power-of-two estimate.

The other API allows pre-growing to allow a specific number of insertions to be performed without further growth. This API takes the maximum load factor and other implementation details into account.

### DIFF
--- a/common/map.h
+++ b/common/map.h
@@ -333,8 +333,8 @@ class MapBase : protected RawHashtable::BaseImpl<InputKeyT, InputValueT,
   // number based on the maximum load factor. If a specific number of insertions
   // need to be achieved without triggering growth, use the `GrowForInsertCount`
   // method.
-  auto Grow(ssize_t target_alloc_size, KeyContextT key_context = KeyContextT())
-      -> void;
+  auto GrowToAllocSize(ssize_t target_alloc_size,
+                       KeyContextT key_context = KeyContextT()) -> void;
 
   // Grow the map sufficiently to allow inserting the specified number of keys.
   auto GrowForInsertCount(ssize_t count,
@@ -551,9 +551,9 @@ MapBase<InputKeyT, InputValueT, InputKeyContextT>::Update(
 }
 
 template <typename InputKeyT, typename InputValueT, typename InputKeyContextT>
-void MapBase<InputKeyT, InputValueT, InputKeyContextT>::Grow(
+void MapBase<InputKeyT, InputValueT, InputKeyContextT>::GrowToAllocSize(
     ssize_t target_alloc_size, KeyContextT key_context) {
-  this->GrowImpl(target_alloc_size, key_context);
+  this->GrowToAllocSizeImpl(target_alloc_size, key_context);
 }
 
 template <typename InputKeyT, typename InputValueT, typename InputKeyContextT>
--- a/common/map.h
+++ b/common/map.h
@@ -325,6 +325,21 @@ class MapBase : protected RawHashtable::BaseImpl<InputKeyT, InputValueT,
              std::invocable<InsertCallbackT, LookupKeyT, void*, void*> &&
              std::invocable<UpdateCallbackT, KeyT&, ValueT&>);
 
+  // Grow the map to a specific allocation size.
+  //
+  // This will grow the map's hashtable if necessary for it to have an
+  // allocation size of `target_alloc_size` which must be a power of two. Note
+  // that this will not allow that many keys to be inserted, but a smaller
+  // number based on the maximum load factor. If a specific number of insertions
+  // need to be achieved without triggering growth, use the `GrowForInsertCount`
+  // method.
+  auto Grow(ssize_t target_alloc_size, KeyContextT key_context = KeyContextT())
+      -> void;
+
+  // Grow the map sufficiently to allow inserting the specified number of keys.
+  auto GrowForInsertCount(ssize_t count,
+                          KeyContextT key_context = KeyContextT()) -> void;
+
   // Erase a key from the map.
   template <typename LookupKeyT>
   auto Erase(LookupKeyT lookup_key, KeyContextT key_context = KeyContextT())
@@ -533,6 +548,18 @@ MapBase<InputKeyT, InputValueT, InputKeyContextT>::Update(
   insert_cb(lookup_key, static_cast<void*>(&entry->key_storage),
             static_cast<void*>(&entry->value_storage));
   return InsertKVResult(true, *entry);
+}
+
+template <typename InputKeyT, typename InputValueT, typename InputKeyContextT>
+void MapBase<InputKeyT, InputValueT, InputKeyContextT>::Grow(
+    ssize_t target_alloc_size, KeyContextT key_context) {
+  this->GrowImpl(target_alloc_size, key_context);
+}
+
+template <typename InputKeyT, typename InputValueT, typename InputKeyContextT>
+void MapBase<InputKeyT, InputValueT, InputKeyContextT>::GrowForInsertCount(
+    ssize_t count, KeyContextT key_context) {
+  this->GrowForInsertCountImpl(count, key_context);
 }
 
 template <typename InputKeyT, typename InputValueT, typename InputKeyContextT>
--- a/common/map_test.cpp
+++ b/common/map_test.cpp
@@ -250,7 +250,8 @@ TYPED_TEST(MapTest, GrowToAllocSize) {
   // No further growth triggered.
   EXPECT_EQ(storage_bytes, m.ComputeMetrics().storage_bytes);
 
-  // Get a couple of doubling based growths.
+  // Get a few doubling based growths, and at least one beyond the largest small
+  // size.
   m.GrowToAllocSize(64);
   ExpectMapElementsAre(
       m, MakeKeyValues([](int k) { return k * 100; }, llvm::seq(8, 24)));
@@ -316,14 +317,14 @@ TYPED_TEST(MapTest, GrowForInsert) {
     SCOPED_TRACE(llvm::formatv("Key: {0}", i).str());
     ASSERT_TRUE(m.Erase(i));
   }
-  m.GrowForInsertCount(1717);
+  m.GrowForInsertCount(321);
   storage_bytes = m.ComputeMetrics().storage_bytes;
-  for (int i : llvm::seq(128, 1717 + 128)) {
+  for (int i : llvm::seq(128, 321 + 128)) {
     SCOPED_TRACE(llvm::formatv("Key: {0}", i).str());
     ASSERT_TRUE(m.Insert(i, i * 100).is_inserted());
   }
   ExpectMapElementsAre(m, MakeKeyValues([](int k) { return k * 100; },
-                                        llvm::seq(128, 1717 + 128)));
+                                        llvm::seq(128, 321 + 128)));
   EXPECT_EQ(storage_bytes, m.ComputeMetrics().storage_bytes);
 }
 
--- a/common/map_test.cpp
+++ b/common/map_test.cpp
@@ -226,6 +226,92 @@ TYPED_TEST(MapTest, Conversions) {
   EXPECT_EQ(104, *cmv3[4]);
 }
 
+TYPED_TEST(MapTest, Grow) {
+  using MapT = TypeParam;
+
+  MapT m;
+  // Grow when empty. May be a no-op for some small sizes.
+  m.Grow(32);
+
+  // Add some elements that will need to be propagated through subsequent
+  // growths. Also delete some.
+  for (int i : llvm::seq(1, 24)) {
+    SCOPED_TRACE(llvm::formatv("Key: {0}", i).str());
+    ASSERT_TRUE(m.Insert(i, i * 100).is_inserted());
+  }
+  for (int i : llvm::seq(1, 8)) {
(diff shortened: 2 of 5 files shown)


common/map.h
```diff
@@ -325,6 +325,21 @@ class MapBase : protected RawHashtable::BaseImpl<InputKeyT, InputValueT,
...
 
+  // Grow the map to a specific allocation size.
+  //
+  // This will grow the map's hashtable if necessary for it to have an
+  // allocation size of `target_alloc_size` which must be a power of two. Note
+  // that this will not allow that many keys to be inserted, but a smaller
+  // number based on the maximum load factor. If a specific number of insertions
+  // need to be achieved without triggering growth, use the `GrowForInsertCount`
+  // method.
+  auto Grow(ssize_t target_alloc_size, KeyContextT key_context = KeyContextT())
```
josh11b: I feel like this should have a more-specific name (maybe `GrowToAllocation` or something), so as to express that it isn't preferred over `GrowForInsertCount`

common/raw_hashtable.h
```diff
@@ -827,6 +841,84 @@ auto BaseImpl<InputKeyT, InputValueT, InputKeyContextT>::InsertImpl(
                     "or an empty slot.";
 }
 
+template <typename InputKeyT, typename InputValueT, typename InputKeyContextT>
+[[clang::noinline]] auto
+BaseImpl<InputKeyT, InputValueT, InputKeyContextT>::GrowImpl(
+    ssize_t target_alloc_size, KeyContextT key_context) -> void {
+  CARBON_CHECK(llvm::isPowerOf2_64(target_alloc_size));
+  if (target_alloc_size < alloc_size()) {
```
josh11b:
```suggestion
  if (target_alloc_size <= alloc_size()) {
```
Or is the intent that you can call `Grow(alloc_size())` in order to clear out the tombstones? (If so, then please add that to the documented contract.)

common/raw_hashtable.h
```diff
@@ -827,6 +841,84 @@ auto BaseImpl<InputKeyT, InputValueT, InputKeyContextT>::InsertImpl(
...
+
+  // Create locals for the old state of the table.
+  ssize_t old_size = alloc_size();
+  CARBON_DCHECK(old_size > 0);
+  bool old_small = is_small();
+  Storage* old_storage = storage();
+  uint8_t* old_metadata = metadata();
+  EntryT* old_entries = entries();
+
+  // Configure for the new size allocate the new storage.
```
josh11b:
```suggestion
  // Configure for the new size and allocate the new storage.
```

common/raw_hashtable.h
```diff
@@ -1004,16 +1096,35 @@ auto BaseImpl<InputKeyT, InputValueT,
...
 }
 
-// Grow the hashtable to create space and then insert into it. Returns the
-// selected insertion entry. Never returns null. In addition to growing and
-// selecting the insertion entry, this routine updates the metadata array so
-// that this function can be directly called and the result returned from
-// `InsertImpl`.
+// Optimized routine for growing to the next alloc size.
+//
+// A particularly common and important to optimize path is growing to the next
```
josh11b:
```suggestion
// A particularly common and important-to-optimize path is growing to the next
```

common/raw_hashtable.h
```diff
@@ -1004,16 +1096,35 @@ auto BaseImpl<InputKeyT, InputValueT,
...
+// 2) The new high bit is one, the entry is at the old index plus the old
+//    size.
+//
+// 3) The entry's current index doesn't match the initial hash index because
+//    it required some amount of probing to find an empty slot.
+//
+// The design of the hash table tries to minimize how many entries fall into
+// case (3), so we expect the vast majority of entries to be in (1) or (2).
+// This lets us model growth notionally as duplicating the hash table,
+// clearing out the empty slots, and inserting any probed elements. That model
```
josh11b: Do you mean empty or tombstone/deleted slots? Or do you mean slots that were filled but are now in the other half?

common/raw_hashtable.h
```diff
@@ -1004,16 +1096,35 @@ auto BaseImpl<InputKeyT, InputValueT,
...
+//
+// 2) The new high bit is one, the entry is at the old index plus the old
+//    size.
+//
+// 3) The entry's current index doesn't match the initial hash index because
+//    it required some amount of probing to find an empty slot.
+//
+// The design of the hash table tries to minimize how many entries fall into
+// case (3), so we expect the vast majority of entries to be in (1) or (2).
+// This lets us model growth notionally as duplicating the hash table,
```
josh11b: Do you mean duplicating (making one copy) or making two copies? In particular, clarify that the original table is not being reused for "new high bit is zero" in some way.

common/raw_hashtable.h
```diff
@@ -1053,7 +1161,7 @@ BaseImpl<InputKeyT, InputValueT, InputKeyContextT>::GrowAndInsert(
       << ", size: " << old_size;
 #endif
 
-  // Compute the new size and grow the storage in place (if possible).
+  // Configure for the new size allocate the new storage.
```
josh11b:
```suggestion
  // Configure for the new size and allocate the new storage.
```

common/map_test.cpp
```diff
@@ -226,6 +226,92 @@ TYPED_TEST(MapTest, Conversions) {
   EXPECT_EQ(104, *cmv3[4]);
 }
 
+TYPED_TEST(MapTest, Grow) {
+  using MapT = TypeParam;
+
+  MapT m;
+  // Grow when empty. May be a no-op for some small sizes.
+  m.Grow(32);
```
josh11b: Is there a way to see the allocated size at this point? Even a test-only API would help.

common/map_test.cpp
```diff
@@ -226,6 +226,92 @@ TYPED_TEST(MapTest, Conversions) {
...
+    SCOPED_TRACE(llvm::formatv("Key: {0}", i).str());
+    ASSERT_TRUE(m.Insert(i, i * 100).is_inserted());
+  }
+  for (int i : llvm::seq(1, 8)) {
+    SCOPED_TRACE(llvm::formatv("Key: {0}", i).str());
+    ASSERT_TRUE(m.Erase(i));
+  }
+
+  // No-op.
+  m.Grow(16);
```
josh11b: Is there a way to see that this didn't reallocate? Possibly using `CountProbedKeys` (maybe combined with a hash function that tries to put all the keys in the same bucket) or something?

common/map_test.cpp
```diff
@@ -226,6 +226,92 @@ TYPED_TEST(MapTest, Conversions) {
...
+  ExpectMapElementsAre(
+      m, MakeKeyValues([](int k) { return k * 100; }, llvm::seq(8, 24)));
+
+  // Get a couple of doubling based growths.
+  m.Grow(64);
```
josh11b: `CountProbedKeys` would go down here, in a deterministic way depending how you manipulated the hash function.

common/map_test.cpp
```diff
@@ -226,6 +226,92 @@ TYPED_TEST(MapTest, Conversions) {
...
+  m.Grow(1024);
+  ExpectMapElementsAre(
+      m, MakeKeyValues([](int k) { return k * 100; }, llvm::seq(16, 48)));
+}
+
+TYPED_TEST(MapTest, GrowForInsert) {
+  using MapT = TypeParam;
+
+  MapT m;
+  m.GrowForInsertCount(42);
```
josh11b, line 289: Would be nice to save the allocated size after this so you could check that it doesn't grow after inserting 42 times.

common/raw_hashtable.h
chandlerc: No, your suggested change is correct. The code path below won't actually work correctly for clearing out tombstones.

If we want to support that, I'd use a different API and implementation. So not part of this PR.

common/map_test.cpp
chandlerc: I can now that there are a more useful set of metrics. I was avoiding coupling the two as I wasn't sure it was worthwhile.

Would using the metrics for this kind of thing be reasonable for you?

common/map_test.cpp
chandlerc: See above about metrics.

common/map_test.cpp
chandlerc: See above about metrics for testing that probing improves.

I can switch to the forced-collision hash if you think its important to test probing reduction here, but it seemed better to get coverage from the variety of different key/value types and small sizes.

common/map_test.cpp
chandlerc: See above about using metrics.

common/raw_hashtable.h
chandlerc: Tried to clarify, but let me know if it could be better.

common/raw_hashtable.h
chandlerc: Tried to make this more clear (it's both).

common/map.h
chandlerc: How about `GrowToAllocSize`?

common/map_test.cpp
chandlerc: Playing with it, this looks quite good. Switched to checking the growth pretty precisely. PTAL.

common/map_test.cpp
chandlerc: Should be done.

common/map_test.cpp
chandlerc: The metrics based assertions are done. Not clear if you think its important to also have the forced-collision side of this, let me know.

common/map_test.cpp
chandlerc: Done.

chandlerc (commented):
After looking at how the metrics can work here, it's pretty nice. I've updated to use them. PTAL.

common/map_test.cpp
josh11b: Great! Could you do the same for `set_test.cpp` please?

common/map_test.cpp
```diff
@@ -226,6 +226,111 @@ TYPED_TEST(MapTest, Conversions) {
...
+  }
+  m.GrowForInsertCount(42);
+  storage_bytes = m.ComputeMetrics().storage_bytes;
+  for (int i : llvm::seq(42, 84)) {
+    SCOPED_TRACE(llvm::formatv("Key: {0}", i).str());
+    ASSERT_TRUE(m.Insert(i, i * 100).is_inserted());
+  }
+  ExpectMapElementsAre(
+      m, MakeKeyValues([](int k) { return k * 100; }, llvm::seq(32, 84)));
+  EXPECT_EQ(storage_bytes, m.ComputeMetrics().storage_bytes);
```
josh11b, line 312: Could you insert some new elements here with different keys than were erased, and then check that that didn't require any growth? That would give evidence that the call to `m.GrowForInsertCount(42);` cleared out tombstones.  The test would be strongest if you ended up with 128*7/8 == 112 non-erased elements in the map.

common/map_test.cpp
josh11b: The metrics approach is fine. I suggested in another comment a test to show that tombstones were cleared out by growth. It would be nice to show `CountProbedKeys` going down when growing (since that is one of the benefits of growth and it not going down could be indicative of a bug that affects performance but not correctness), but not necessary.

common/map_test.cpp
chandlerc: The new elements inserted here already have different keys than were erased, and we do check that no growth occurs. Are you just looking for us to get closer to the load factor?

I'm not really sure how to write the test you want here, because we *don't* actually factor in clearing out tombstones when we compute growth for insert. There is a comment in the code about that being something we could consider doing in the future.

Without that we can't ask for the larger growth, and there is no guarantee that `m.GrowForInsertCount(42)` will let us insert more than `42` elements without growth, and no real contract on whether it clears out tombstones or not... It feels like this is baking in a test of an implementation detail of how growth works, which I'd rather avoid.

common/map_test.cpp
chandlerc: I feel like for both the probed keys and the tombstones, this isn't really about testing the behavior of these APIs, but of the growth routines more generally... And that's proven a bit tricky for me to do.

I tried adding the probed key count test, but we end up with zero probed keys here in some cases, and so the count doesn't change. And I left a comment in the other thread about why I'm not really sure how to test the tombstone removal without ending up over-fitting to the particular implementation.

I also tried adding a test for probed key counts decreasing with the forced hash collisions, and that actually doesn't work either -- because we force the hashes to collide, even growth doesn't remove the collision! =[

The only way I see to test this behavior realistically is to contrive a synthetic hash that has collisions below some number of bits but not above, and then notice that once we grow past that threshold, the probing decreases. But this seems much too brittle to me, and the benchmarks seem like a better place to make sure we don't have a performance problem here.

So I've tried a bunch of things here that haven't worked out. Let me know if you still see something that seems important to add coverage for.

common/map_test.cpp
chandlerc: Doh, could have sworn I did that. Sorry! =D Done.

josh11b (approved):

PR #4046, opened 2024-06-10
Collect more detailed metrics on hashtables.

chandlerc:
Previously we just looked at the raw count of probed keys. Now, we compute the average and max of both the probe _distance_ measured in the number of _groups_ probed, and the number of probe _compares_ measured in the compares required _before_ finding the matching entry.

This lets us understand the relative impact of probe-distance vs. tag collisions on a given set of benchmark keys. Some of this is motivated by considering additional optimization techniques similar to those used in Boost's table and the F14 table from Facebook/Meta.

### DIFF
--- a/common/map.h
+++ b/common/map.h
@@ -121,8 +121,8 @@ class MapView
   // benchmarking or logging of performance anomalies. The specific metrics
   // returned have no specific guarantees beyond being informative in
   // benchmarks.
-  auto GetMetrics(KeyContextT key_context = KeyContextT()) -> MetricsT {
-    return ImplT::GetMetrics(key_context);
+  auto ComputeMetrics(KeyContextT key_context = KeyContextT()) -> MetricsT {
+    return ImplT::ComputeMetricsImpl(key_context);
   }
 
  private:
@@ -231,8 +231,9 @@ class MapBase : protected RawHashtable::BaseImpl<InputKeyT, InputValueT,
   }
 
   // Convenience forwarder to the view type.
-  auto GetMetrics(KeyContextT key_context = KeyContextT()) const -> MetricsT {
-    return ViewT(*this).GetMetrics(key_context);
+  auto ComputeMetrics(KeyContextT key_context = KeyContextT()) const
+      -> MetricsT {
+    return ViewT(*this).ComputeMetrics(key_context);
   }
 
   // Insert a key and value into the map. If the key is already present, the new
--- a/common/map.h
+++ b/common/map.h
@@ -66,6 +66,7 @@ class MapView
   using KeyT = typename ImplT::KeyT;
   using ValueT = typename ImplT::ValueT;
   using KeyContextT = typename ImplT::KeyContextT;
+  using MetricsT = typename ImplT::MetricsT;
 
   // This type represents the result of lookup operations. It encodes whether
   // the lookup was a success as well as accessors for the key and value.
@@ -117,15 +118,11 @@ class MapView
     requires(std::invocable<CallbackT, KeyT&, ValueT&>);
 
   // This routine is relatively inefficient and only intended for use in
-  // benchmarking or logging of performance anomalies. The specific count
-  // returned has no specific guarantees beyond being informative in benchmarks.
-  // It counts how many of the keys in the hashtable have required probing
-  // beyond their initial group of slots.
-  //
-  // TODO: Replace with a more general metrics routine that covers other
-  // important aspects such as load factor, and average probe *distance*.
-  auto CountProbedKeys(KeyContextT key_context = KeyContextT()) -> ssize_t {
-    return ImplT::CountProbedKeys(key_context);
+  // benchmarking or logging of performance anomalies. The specific metrics
+  // returned have no specific guarantees beyond being informative in
+  // benchmarks.
+  auto GetMetrics(KeyContextT key_context = KeyContextT()) -> MetricsT {
+    return ImplT::GetMetrics(key_context);
   }
 
  private:
@@ -165,6 +162,7 @@ class MapBase : protected RawHashtable::BaseImpl<InputKeyT, InputValueT,
   using KeyContextT = typename ImplT::KeyContextT;
   using ViewT = MapView<KeyT, ValueT, KeyContextT>;
   using LookupKVResult = typename ViewT::LookupKVResult;
+  using MetricsT = typename ImplT::MetricsT;
 
   // The result type for insertion operations both indicates whether an insert
   // was needed (as opposed to finding an existing element), and provides access
@@ -233,9 +231,8 @@ class MapBase : protected RawHashtable::BaseImpl<InputKeyT, InputValueT,
   }
 
   // Convenience forwarder to the view type.
-  auto CountProbedKeys(KeyContextT key_context = KeyContextT()) const
-      -> ssize_t {
-    return ViewT(*this).CountProbedKeys(key_context);
+  auto GetMetrics(KeyContextT key_context = KeyContextT()) const -> MetricsT {
+    return ViewT(*this).GetMetrics(key_context);
   }
 
   // Insert a key and value into the map. If the key is already present, the new
--- a/common/map_benchmark.cpp
+++ b/common/map_benchmark.cpp
@@ -18,6 +18,7 @@ using RawHashtable::CarbonHashDI;
 using RawHashtable::GetKeysAndHitKeys;
 using RawHashtable::GetKeysAndMissKeys;
 using RawHashtable::HitArgs;
+using RawHashtable::ReportTableMetrics;
 using RawHashtable::SizeArgs;
 using RawHashtable::ValueToBool;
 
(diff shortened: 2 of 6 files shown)


common/map.h
```diff
@@ -117,15 +118,11 @@ class MapView
...
-  // benchmarking or logging of performance anomalies. The specific count
-  // returned has no specific guarantees beyond being informative in benchmarks.
-  // It counts how many of the keys in the hashtable have required probing
-  // beyond their initial group of slots.
-  //
-  // TODO: Replace with a more general metrics routine that covers other
-  // important aspects such as load factor, and average probe *distance*.
-  auto CountProbedKeys(KeyContextT key_context = KeyContextT()) -> ssize_t {
-    return ImplT::CountProbedKeys(key_context);
+  // benchmarking or logging of performance anomalies. The specific metrics
```
zygoloid, line 121: I wonder if it'd be worth calling out the "relatively inefficient" part in the name, eg `ComputeMetrics`.

common/raw_hashtable.h
```diff
@@ -709,26 +727,74 @@ ViewImpl<InputKeyT, InputValueT, InputKeyContextT>::ForEachEntry(
...
+
+      // For each probed key, go through the probe sequence to find both the
+      // probe distance and how many comparisons are required.
+      ssize_t distance = 0;
+      ssize_t compares = 0;
+      for (;; s.Next()) {
+        ssize_t probe_index = s.index();
+        MetadataGroup probe_g =
+            MetadataGroup::Load(local_metadata, probe_index);
+        auto metadata_matched_range = probe_g.Match(tag);
```
zygoloid: Just checking my understanding: we're computing here the number of comparisons that would be required to look up each key now (as if we iterated over every key and did a lookup with a counting comparator), not for example the number of comparisons that were performed originally when inserting the elements (which is similar but might in general be a little different, due to rehashes and tombstone reuse).

Even though we don't provide any guarantees about the metrics, a brief comment explaining each one would be useful.

common/raw_hashtable.h
```diff
@@ -709,26 +727,74 @@ ViewImpl<InputKeyT, InputValueT, InputKeyContextT>::ForEachEntry(
...
+          if (match_index >= byte_index) {
+            // Note we only count the compares that will *fail* as part of
+            // probing. The last successful compare isn't interesting, it is
+            // always needed.
+            break;
+          }
+          compares += 1;
+        }
+        break;
+      }
```
zygoloid: I find this loop structure a bit awkward, especially given that the loop body ends in `break`. Maybe we could split this into work for the groups we probed past and work we do for the group actually containing the key?
```suggestion
      for (; s.index() != group_index; s.Next()) {
        auto metadata_matched_range =
            MetadataGroup::Load(local_metadata, s.index()).Match(tag);
        auto metadata_matched_range = probe_g.Match(tag);
        compares += std::distance(metadata_matched_range.begin(),
                                  metadata_matched_range.end());
        distance += 1;
      }

      auto metadata_matched_range =
          MetadataGroup::Load(local_metadata, s.index()).Match(tag);
      CARBON_CHECK(!metadata_matched_range.empty());
      for (ssize_t match_index : metadata_matched_range) {
        if (match_index >= byte_index) {
          // Note we only count the compares that will *fail* as part of
          // probing. The last successful compare isn't interesting, it is
          // always needed.
          break;
        }
        compares += 1;
      }
```

common/raw_hashtable.h
```diff
@@ -709,26 +727,74 @@ ViewImpl<InputKeyT, InputValueT, InputKeyContextT>::ForEachEntry(
...
+          distance += 1;
+          continue;
+        }
+
+        CARBON_CHECK(!metadata_matched_range.empty());
+        for (ssize_t match_index : metadata_matched_range) {
+          if (match_index >= byte_index) {
+            // Note we only count the compares that will *fail* as part of
+            // probing. The last successful compare isn't interesting, it is
+            // always needed.
```
zygoloid: FWIW, my naïve inclination is to include the final comparison in the statistics. The reason is, if the comparison cost dominates all other factors (and successful and failed comparisons have the same cost), a reduction of X% in the average number of comparisons translates into a reduction of X% in the overall runtime if the number of comparisons includes the final comparison.

I understand you're excluding the final comparison to capture the overhead of potentially-unnecessary comparisons, though, so maybe the figure you're computing is the one that's more useful to you. If so, of course keep it as-is :) (Either way, this is a detail that'd be useful to include in a comment on the metrics so that people know how to interpret the benchmark results.)

common/raw_hashtable.h
```diff
@@ -709,26 +727,74 @@ ViewImpl<InputKeyT, InputValueT, InputKeyContextT>::ForEachEntry(
...
+      ssize_t distance = 0;
+      ssize_t compares = 0;
+      for (;; s.Next()) {
+        ssize_t probe_index = s.index();
+        MetadataGroup probe_g =
+            MetadataGroup::Load(local_metadata, probe_index);
+        auto metadata_matched_range = probe_g.Match(tag);
+        if (probe_index != group_index) {
+          compares += std::distance(metadata_matched_range.begin(),
+                                    metadata_matched_range.end());
```
zygoloid: Is it worth adding a `size` / `popcount` function to `BitIndexRange`?

common/raw_hashtable.h
```diff
@@ -709,26 +727,74 @@ ViewImpl<InputKeyT, InputValueT, InputKeyContextT>::ForEachEntry(
...
+      metrics.probe_max_distance =
+          std::max(metrics.probe_max_distance, distance);
+      metrics.probe_avg_compares += compares;
+      metrics.probe_max_compares =
+          std::max(metrics.probe_max_compares, compares);
     }
   }
-  return count;
+  metrics.probe_avg_compares /= metrics.key_count;
+  metrics.probe_avg_distance /= metrics.key_count;
```
zygoloid: Looks like we can divide by zero here.

common/set_benchmark.cpp
```diff
@@ -365,10 +365,23 @@ static void BM_SetInsertSeq(benchmark::State& state) {
...
     // because it will scale this by the number of iterations. We want to
-    // display the probe count of this benchmark *parameter*, not the probe
-    // count that resulted from the number of iterations. That means we use the
-    // normal counter API without flags.
-    state.counters["Probed"] = s.CountProbedKeys();
+    // display the metrics for this benchmark *parameter*, not what resulted
+    // from the number of iterations. That means we use the normal counter API
+    // without flags.
+    auto metrics = s.GetMetrics();
+    state.counters["P-compares"] = metrics.probe_avg_compares;
```
zygoloid: Is it worth `CHECK`ing that `metrics.key_count == keys.size()` here?

common/map_benchmark.cpp
```diff
@@ -159,6 +159,41 @@ template <typename MapT>
...
+    state.counters["Pmax-compares"] = metrics.probe_max_compares;
+    state.counters["Probed"] = metrics.probed_key_count;
+
+    state.counters["Storage"] = metrics.storage_bytes;
+    // Also compute how 'efficient' the storage is, 1.0 being zero bytes outside
+    // of key and value.
+    state.counters["Storage eff"] =
+        static_cast<double>(metrics.key_count * (sizeof(KT) + sizeof(VT))) /
+        metrics.storage_bytes;
+  }
```
zygoloid: Can we factor this out and share it between the map and set benchmarks?

common/raw_hashtable.h
chandlerc: Correct, I've added comments for all the metrics to explain this.

The way I think about this is that these are the "probing" comparisons that constitute overhead before we find the actual matching key.

common/raw_hashtable.h
chandlerc: Doh, fixed.

common/set_benchmark.cpp
chandlerc: Good idea, done.

common/map_benchmark.cpp
chandlerc: Hmm, I hadn't seen an easy way to do this, but I think I figured it out with some concept tricks. Done.

common/raw_hashtable.h
chandlerc: 🤷🏻  I think not yet, but if folks disagree I can.

common/map.h
chandlerc: Sure, renamed.

zygoloid (approved):
[Edit: data race]

LGTM

common/raw_hashtable.h
chandlerc: I've added a comment on the metrics that tries to explain the rationale for the current formulation (it's about measuring overhead of "probing"), but wanted to explicitly check that with that explanation you're happy w/ the current formulation.

chandlerc (commented):
And one discussion thread that I just wanted to make sure didn't get lost.

common/raw_hashtable.h
zygoloid: Yeah, I'm fine with it, given that this sounds like it's the more useful metric to you.

PR #4047, opened 2024-06-10
Reduce the new hashtable test times.

chandlerc:
While it was convenient once to have an immediate check while inserting,
it is indeed far too quadratic. The test was taking 30-60 seconds for
me. =[ So most of the fix here is just to stop doing the check on every
insertion for all previous elements.

There were a few other somewhat slow steps, and I tried to pull those
back as well. I don't think we lose any utility here. Now everything
runs nice and quickly. =]

### DIFF
--- a/common/map_test.cpp
+++ b/common/map_test.cpp
@@ -104,14 +104,10 @@ TYPED_TEST(MapTest, Basic) {
   for (int i : llvm::seq(2, 512)) {
     SCOPED_TRACE(llvm::formatv("Key: {0}", i).str());
     EXPECT_TRUE(m.Insert(i, i * 100).is_inserted());
-
-    // Immediately do a basic check of all elements to pin down when an
-    // insertion corrupts the rest of the table.
-    ExpectMapElementsAre(
-        m,
-        MakeKeyValues([](int k) { return k * 100 + static_cast<int>(k == 1); },
-                      llvm::seq_inclusive(1, i)));
   }
+  ExpectMapElementsAre(
+      m, MakeKeyValues([](int k) { return k * 100 + static_cast<int>(k == 1); },
+                       llvm::seq(1, 512)));
   for (int i : llvm::seq(1, 512)) {
     SCOPED_TRACE(llvm::formatv("Key: {0}", i).str());
     EXPECT_FALSE(m.Insert(i, i * 100 + 1).is_inserted());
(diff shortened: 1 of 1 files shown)


common/map_test.cpp
```diff
@@ -303,13 +299,13 @@ TYPED_TEST(MapTest, GrowForInsert) {
     SCOPED_TRACE(llvm::formatv("Key: {0}", i).str());
     ASSERT_TRUE(m.Erase(i));
   }
-  m.GrowForInsertCount(1717);
-  for (int i : llvm::seq(128, 1717 + 128)) {
+  m.GrowForInsertCount(321);
```
zygoloid: The before and after numbers here seem a bit magical / arbitrary. Where do 1717 and 321 come from?

common/map_test.cpp
```diff
@@ -568,12 +564,9 @@ TYPED_TEST(MapCollisionTest, Basic) {
...
     EXPECT_TRUE(m.Insert(i, i * 100).is_inserted());
-
-    // Immediately do a basic check of all elements to pin down when an
-    // insertion corrupts the rest of the table.
-    ExpectMapElementsAre(m, MakeKeyValues([](int k) { return k * 100; },
-                                          llvm::seq_inclusive(1, i)));
   }
+  ExpectMapElementsAre(
+      m, MakeKeyValues([](int k) { return k * 100; }, llvm::seq(1, 256)));
   EXPECT_FALSE(m.Contains(257));
```
zygoloid: This seems a bit redundant given the previous comparison. But if we're going to keep it, checking that `256` is absent would seem better (`seq(1, 256)` excludes its endpoint, so the last key we insert is 255).

zygoloid (approved):

common/map_test.cpp
chandlerc: Just arbitrary numbers that don't exactly match a power of two (or 7/8ths of a power of 2) and are larger than the current table size.

common/map_test.cpp
chandlerc: Quite redundant and yeah, the wrong edge case. I'll just remove it.

chandlerc:
I've unstacked this and am merging. It was going to have lots of merge conflicts with the growth API anyways, and this way we can get to faster test times sooner.

PR #4051, opened 2024-06-13
Remove return_type_id from Function.

zygoloid:
Instead of redundantly storing both the `return_type_id` and `return_storage_id`, where the declared return type is just the type of the return storage, store only the `return_storage_id`.

Add a convenience property to get the declared return type of the function.

In addition to avoiding storing redundant information, this is a preparatory step for an upcoming change for generics support that will make it more expensive and awkward to store `TypeId`s in places other than the type of an instruction.


josh11b (approved):
Could you update the PR description to include the motivation for this change?

PR #4052, opened 2024-06-13
Include function call and subscripting/indexing in the operator list and precedence chart

josh11b:


docs/design/expressions/README.md
```diff
@@ -302,6 +307,8 @@ Most expressions are modeled as operators:
 
 | Category   | Operator                            | Syntax    | Function                                                              |
 | ---------- | ----------------------------------- | --------- | --------------------------------------------------------------------- |
+| Call       | `()` (unary)                        | `x(...)`  | Function call: the value returned by calling the function `x`.        |
+| Call       | `[]` (unary)                        | `x[y]`    | Subscripting or indexing: returns the element `y` of `x`.             |
```
zygoloid:
```suggestion
| Call       | [`[]`](indexing.md) (unary)         | `x[y]`    | Subscripting or indexing: returns the element `y` of `x`.             |
```

docs/design/expressions/README.md
```diff
@@ -147,17 +150,19 @@ graph BT
...
     as --> pointerType
 
-    memberAccess --> top
-    pointer --> memberAccess
+    pointer --> callAndIndexing
     negation & complement & incDec --> pointer
     unary --> negation & complement
     %% Use a longer arrow here to put `not` next to `and` and `or`.
-    not -------> memberAccess
+    not -------> callAndIndexing
```
zygoloid: Looking at [the indexing design](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/expressions/indexing.md#details), I think we were a little wrong about where this goes: function call and indexing should go in the same group as `memberAccess`, not a separate group, so that `a.b().c[i].d()` and the like work.

I think we still want the change that `constType` points to that group, though.

docs/design/expressions/README.md
```diff
@@ -69,6 +69,9 @@ graph BT
                   x->(...)"]
     click memberAccess "https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/expressions/member_access.md"
```
zygoloid, line 72: Hm, is there a way we can give different links here for member access vs call vs indexing?

josh11b: Done.

docs/design/expressions/README.md
josh11b: Should be fixed now.

josh11b (commented):
Changed the background of the subgraph.

zygoloid (approved):

PR #4054, opened 2024-06-16
Introduce version and build info stamping.

chandlerc:
This adds a defined Carbon version to the Bazel build and codebase that can be used both to implement features like version checks and to report a meaningful version on the command line. This replaces a hard-coded string and a TODO in the driver.

As part of this, it adds support for defining the version in Bazel, and special build flags for overriding relevant parts such as the pre-release marker used. The exact structure and meaning of our version string, including the pre-release parts, is implemented here in line with the draft proposal:
https://docs.google.com/document/d/11S5VAPe5Pm_BZPlajWrqDDVr9qc7-7tS2VshqO0wWkk/edit?resourcekey=0-2YFC9Uvl4puuDnWlr2MmYw

This also introduces a workspace status command to the repository to extract the git commit SHA and other information when building, and the logic to stamp that into binaries as part of the version string when useful. The technique used leverages weak symbols with whole archive linking to allow a link-time override of unstamped data with stamped data in the leaf executable. This makes building with `--stamp` a reasonable default, especially for development builds. The CI system is explicitly opted out of this as there it has no benefit.

Last but not least, all of these are wired into the install rules so that we build installable packages with the version number in a conventional place in the directory and filename.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -2,6 +2,22 @@
 # Exceptions. See /LICENSE for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+# Ensure all builds have Carbon's workspace status attached. We have carefully
+# factored the stamping done by this to avoid excessive build performance impact
+# and so enable stamping with it by default. CI and systems especially dependent
+# on caching should explicitly use `--nostamp`.
+build --workspace_status_command=./scripts/workspace_status.py
+build --stamp
+
+# Provide aliases for configuring the release and pre-release version being
+# built.
+build --flag_alias=release=//bazel/version:release
+build --flag_alias=pre_release=//bazel/version:pre_release
+build --flag_alias=rc_number=//bazel/version:rc_number
+build --flag_alias=beta_number=//bazel/version:beta_number
+build --flag_alias=alpha_number=//bazel/version:alpha_number
+build --flag_alias=nightly_date=//bazel/version:nightly_date
+
 # Support running clang-tidy with:
 #   bazel build --config=clang-tidy -k //...
 # See: https://github.com/erenon/bazel_clang_tidy
--- a/.bazelrc
+++ b/.bazelrc
@@ -10,7 +10,7 @@ build --workspace_status_command=./scripts/workspace_status.py
 build --stamp
 
 # Provide aliases for configuring the release and pre-release version being
-# built.
+# built. For documentation of these flags, see //bazel/version/BUILD.
 build --flag_alias=release=//bazel/version:release
 build --flag_alias=pre_release=//bazel/version:pre_release
 build --flag_alias=rc_number=//bazel/version:rc_number
--- a/bazel/version/BUILD
+++ b/bazel/version/BUILD
@@ -28,10 +28,15 @@ bool_flag(
 
 # A `--pre_release=KIND` flag where `KIND` must be one of:
 # - `rc` -- a release candidate version.
+#    Example: `--pre_release=rc --rc_number=2`
 # - `beta` -- a beta version.
+#    Example: `--pre_release=beta --beta_number=2`
 # - `alpha` -- an alpha version.
+#    Example: `--pre_release=alpha --alpha_number=2`
 # - `nightly -- a nightly version.
+#    Example: `--pre_release=nightly --nightly_date=2024.06.17`
 # - `dev` -- the default, a development build.
+#    Example: `--pre_release=dev`
 #
 # This flag cannot be used along with `--release`, and for all but the `dev`
 # kind must be combined with one of the below flags to specify further details
@@ -48,10 +53,7 @@ string_flag(
     ],
 )
 
-# When `--pre_release=rc` is used, `--rc_number=N` set's the release candidate
-# number to `N`.
-#
-# An error if used without `--pre_release=rc`.
+# `--rc_number=N` sets the release candidate number to `N`. Requires `--pre_release=rc`.
 int_flag(
     name = "rc_number",
     build_setting_default = -1,
--- a/bazel/version/BUILD
+++ b/bazel/version/BUILD
@@ -8,11 +8,34 @@ package(default_visibility = ["//toolchain/install:__pkg__"])
 
 exports_files(["gen_tmpl.py"])
 
+# Several flags are provided for customizing the exact version used for the
+# build of Carbon. Each of these is documented here, but rather than using the
+# label-based names in Bazel invocations (`bazel build --//bazel/version:flag`)
+# we suggest using the flag aliases provided in the project's `.bazelrc` and we
+# document the flags using those aliases. The aliases match the local flag names
+# here.
+#
+# For more details on the versioning scheme used by Carbon, see:
+# - https://docs.google.com/document/d/11S5VAPe5Pm_BZPlajWrqDDVr9qc7-7tS2VshqO0wWkk/edit?resourcekey=0-2YFC9Uvl4puuDnWlr2MmYw
+#   TODO: Replace with path to the markdown once this lands.
+#
+# First, we provide a flag to enable a release version: `--release`. It is
+# disabled by default, and if enabled it must be the only version flag used.
 bool_flag(
     name = "release",
     build_setting_default = False,
 )
 
+# A `--pre_release=KIND` flag where `KIND` must be one of:
+# - `rc` -- a release candidate version.
+# - `beta` -- a beta version.
+# - `alpha` -- an alpha version.
+# - `nightly -- a nightly version.
+# - `dev` -- the default, a development build.
+#
+# This flag cannot be used along with `--release`, and for all but the `dev`
+# kind must be combined with one of the below flags to specify further details
+# of the version.
 string_flag(
     name = "pre_release",
     build_setting_default = "dev",
@@ -25,21 +48,38 @@ string_flag(
     ],
 )
 
+# When `--pre_release=rc` is used, `--rc_number=N` set's the release candidate
+# number to `N`.
+#
+# An error if used without `--pre_release=rc`.
 int_flag(
     name = "rc_number",
     build_setting_default = -1,
 )
 
+# When `--pre_release=beta` is used, `--beta_number=N` set's the beta
+# pre-release number to `N`.
+#
+# An error if used without `--pre_release=beta`.
 int_flag(
     name = "beta_number",
     build_setting_default = -1,
 )
 
+# When `--pre_release=alpha` is used, `--alpha_number=N` set's the alpha
+# pre-release number to `N`.
+#
+# An error if used without `--pre_release=alpha`.
 int_flag(
     name = "alpha_number",
     build_setting_default = -1,
 )
 
+# When `--pre_release=nightly` is used, `--nightly_date=YYYY.MM.DD` set's the
+# nightly build date. The value for this flag must be a string with the exact
+# format of `YYYY.MM.DD`.
+#
+# An error if used without `--pre_release=nightly`.
 string_flag(
     name = "nightly_date",
     build_setting_default = "",
--- a/bazel/version/BUILD
+++ b/bazel/version/BUILD
@@ -1,0 +1,55 @@
+# Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+# Exceptions. See /LICENSE for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag", "int_flag", "string_flag")
+
+package(default_visibility = ["//toolchain/install:__pkg__"])
+
+exports_files(["gen_tmpl.py"])
+
+bool_flag(
+    name = "release",
+    build_setting_default = False,
+)
+
+string_flag(
+    name = "pre_release",
+    build_setting_default = "dev",
+    values = [
+        "rc",
+        "beta",
+        "alpha",
+        "nightly",
+        "dev",
+    ],
+)
+
+int_flag(
+    name = "rc_number",
+    build_setting_default = -1,
+)
+
+int_flag(
+    name = "beta_number",
+    build_setting_default = -1,
+)
+
+int_flag(
+    name = "alpha_number",
+    build_setting_default = -1,
+)
+
+string_flag(
+    name = "nightly_date",
+    build_setting_default = "",
+)
+
+# A config setting to observe the value of the `--stamp` command line flag
+# within starlark with a macro and `select`. This is a workaround suggested for
+# a Bazel issue: https://github.com/bazelbuild/bazel/issues/11164
+config_setting(
+    name = "internal_stamp_flag_detect",
+    values = {"stamp": "1"},
+    visibility = ["//visibility:public"],
+)
--- a/bazel/version/compute_version.bzl
+++ b/bazel/version/compute_version.bzl
@@ -33,6 +33,9 @@ def compute_version(ctx):
     version = version_base
 
     # See if we need to append a pre-release suffix to the version.
+    #
+    # TODO: We should more fully check for erroneous combinations of flags here
+    # to help ensure users don't get surprising results.
     if not ctx.attr._release_flag[BuildSettingInfo].value:
         pre_release = ctx.attr._pre_release_flag[BuildSettingInfo].value
         pre_release_numbers = {
--- a/bazel/version/compute_version.bzl
+++ b/bazel/version/compute_version.bzl
@@ -1,0 +1,66 @@
+# Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+# Exceptions. See /LICENSE for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+"""Compute the version string."""
+
+load("@bazel_skylib//rules:common_settings.bzl", "BuildSettingInfo")
+load("//:version_base.bzl", "version_base")
+
+def _validate_nightly_date(date):
+    date_components = date.split(".", 2)
+    if len(date_components) != 3:
+        fail("Must provide a nightly date in 'YYYY.MM.DD' format, found '{}'.".format(date))
+    year = date_components[0]
+    if len(year) != 4 or not year.isdigit():
+        fail("The nightly date year was not a sequence of four digits.")
+    month = date_components[1]
+    if len(month) != 2 or not month.isdigit():
+        fail("The nightly date month was not a sequence of two digits.")
+    day = date_components[2]
+    if len(day) != 2 or not day.isdigit():
+        fail("The nightly date day was not a sequence of two digits.")
+
+def compute_version(ctx):
+    """Compute the version string.
+
+    Args:
+        ctx: The context for a rule computing the version.
+
+    Returns:
+        The version string.
+    """
+    version = version_base
+
+    # See if we need to append a pre-release suffix to the version.
+    if not ctx.attr._release_flag[BuildSettingInfo].value:
+        pre_release = ctx.attr._pre_release_flag[BuildSettingInfo].value
+        pre_release_numbers = {
+            "alpha": ctx.attr._alpha_number_flag[BuildSettingInfo].value,
+            "beta": ctx.attr._beta_number_flag[BuildSettingInfo].value,
+            "rc": ctx.attr._rc_number_flag[BuildSettingInfo].value,
+        }
+        if pre_release in pre_release_numbers:
+            number = pre_release_numbers[pre_release]
+            if number < 0:
+                fail("Must provide a non-negative {} number when building that pre-release.".format(pre_release))
+            version += "-{0}.{1}".format(pre_release, number)
+        elif pre_release == "nightly":
+            date = ctx.attr._nightly_date_flag[BuildSettingInfo].value
+            _validate_nightly_date(date)
+            version += "-0.nightly.{}".format(date)
+        else:
+            if pre_release != "dev":
+                fail("Invalid pre-release flag: " + pre_release)
+            version += "-0.dev"
+
+    return version
+
+VERSION_ATTRS = {
+    "_alpha_number_flag": attr.label(default = ":alpha_number"),
+    "_beta_number_flag": attr.label(default = ":beta_number"),
+    "_nightly_date_flag": attr.label(default = ":nightly_date"),
+    "_pre_release_flag": attr.label(default = ":pre_release"),
+    "_rc_number_flag": attr.label(default = ":rc_number"),
+    "_release_flag": attr.label(default = ":release"),
+}
--- a/bazel/version/compute_version.bzl
+++ b/bazel/version/compute_version.bzl
@@ -49,10 +49,10 @@ def compute_version(ctx):
             date = ctx.attr._nightly_date_flag[BuildSettingInfo].value
             _validate_nightly_date(date)
             version += "-0.nightly.{}".format(date)
-        else:
-            if pre_release != "dev":
-                fail("Invalid pre-release flag: " + pre_release)
+        elif pre_release == "dev":
             version += "-0.dev"
+        else:
+            fail("Invalid pre-release flag: " + pre_release)
 
     return version
 
--- a/bazel/version/gen_tmpl.py
+++ b/bazel/version/gen_tmpl.py
@@ -1,0 +1,115 @@
+#!/usr/bin/env python3
+
+"""Generate a file from a template, substituting the provided key/value pairs.
+
+The file format should match Python's `string.Template` substitution rules:
+- `$$` for a literal `$`
+- `$identifier` for some key `identifier` to be substituted
+- `${identifier}` when adjacent text would be interpreted as part of the
+  identifier.
+
+The keys must be strings that are valid identifiers: `[_A-Za-z][_A-Za-z0-9]*`
+
+The values may not contain newlines or any vertical whitespace.
+
+The initial key/value pairs are read from the command line using repeated
+`--substitute=KEY=DEFAULT-VALUE` flags.
+
+Updated values for those keys will be read from any files provided to the
+`--status-file` flag. This flag can be given multiple times and the values will
+be read and updated from the files in order, meaning the last file's value will
+win. New keys are never read from these files. The file format parsed is chose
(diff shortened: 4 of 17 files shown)


bazel/version/rules.bzl
```diff
@@ -0,0 +1,144 @@
...
+            inputs += [
+                ctx.info_file,
+                ctx.version_file,
+            ]
+            arguments += [
+                "--status-file=" + ctx.info_file.path,
+                "--status-file=" + ctx.version_file.path,
+            ]
+
+    ctx.actions.run(
```
jonmeow, line 98: Had you considered using [ctx.actions.expand_template](https://bazel.build/rules/lib/builtins/actions#expand_template)? skylib even provides a rule wrapper for it if you like: https://github.com/bazelbuild/bazel-skylib/blob/main/docs/expand_template_doc.md

I think that'd make it possible to remove gen_tmpl.py, but I think you'd still want most of what's here.

version_base.bzl
```diff
@@ -0,0 +1,21 @@
+# Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+# Exceptions. See /LICENSE for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+"""The base version of the Carbon repository.
+
+The Carbon SemVer version is defined here and then exported to various layers
```
jonmeow: Suggesting a small rephrasing here... For context, I'm trying to work through (a) trying to figure out how Carbon SemIR and Carbon SemVer relate (suggesting link below), and (b) "here and then" mentally parsing similar to "here and now".

Note, SemVer is not an abbreviation I was familiar with (I did a quick search to see if this was some common thing, but I could see others assuming Carbon SemVer is Carbon-specific). If you want to note it explicitly here, I'd suggest something like "This uses Semantic Versioning." as a direct statement.

```suggestion
Carbon's version is defined here for export to various layers
```

version_base.bzl
```diff
@@ -0,0 +1,21 @@
...
+version under *active development*.
+
+The current active development version is 0.0.0 -- we haven't yet made enough
+progress towards our 0.1 milestone to meaningfully bump the version to that even
+to describe the development activity. We also never plan to make even a
+non-development pre-release of 0.0.0, only nightly development pre-releases are
+expected.
+
+For more details on Carbon's versioning, see:
+TODO: Add link.
```
jonmeow:
```suggestion
  https://semver.org/
  TODO: Add Carbon-specific link.
```

bazel/version/compute_version.bzl
```diff
@@ -0,0 +1,66 @@
...
+    Args:
+        ctx: The context for a rule computing the version.
+
+    Returns:
+        The version string.
+    """
+    version = version_base
+
+    # See if we need to append a pre-release suffix to the version.
+    if not ctx.attr._release_flag[BuildSettingInfo].value:
```
jonmeow, line 39: Trying to read through this, what happens when rc, alpha, beta, etc are all set? Given the complexity of the flags, should this produce an error if flags are incorrectly used?

.bazelrc
```diff
@@ -2,6 +2,22 @@
...
 
+# Ensure all builds have Carbon's workspace status attached. We have carefully
+# factored the stamping done by this to avoid excessive build performance impact
+# and so enable stamping with it by default. CI and systems especially dependent
+# on caching should explicitly use `--nostamp`.
+build --workspace_status_command=./scripts/workspace_status.py
+build --stamp
+
+# Provide aliases for configuring the release and pre-release version being
+# built.
```
jonmeow: Should this point to documentation on the flags?

bazel/version/BUILD
```diff
@@ -0,0 +1,55 @@
...
+package(default_visibility = ["//toolchain/install:__pkg__"])
+
+exports_files(["gen_tmpl.py"])
+
+bool_flag(
+    name = "release",
+    build_setting_default = False,
+)
+
+string_flag(
```
jonmeow, line 44: Where do you document how all these flags are intended to be used?

common/version.tmpl.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <string_view>
+
+namespace Carbon {
+
+// A simplistic string-to-integer routine that is consteval for compile-time
+// extracting specific components of the version from the string form. We use
+// `std::string_view` for its broader `constexpr` API.
+static consteval auto ToInt(std::string_view str) -> int {
+  int result = 0;
+  for (;;) {
```
jonmeow:
```suggestion
  while (true) {
```

Minor, I think this is just how we're typically writing these.

common/version.tmpl.cpp
```diff
@@ -0,0 +1,51 @@
...
+  // pre-release component, but that produces the correct result here.
+  return ToInt(str.substr(0, str.find('-')));
+}
+
+// The major, minor, and patch versions are always provided and stable. They
+// don't depend on build stamping or introduce caching issues. Provide normal
+// strong definitions.
+constexpr int Version::Major = MajorVersion("$VERSION");
+constexpr int Version::Minor = MinorVersion("$VERSION");
+constexpr int Version::Patch = PatchVersion("$VERSION");
```
jonmeow, line 49: Out of curiosity, had you considered just providing these directly? i.e., where you generate the version file, just do the parse in bzl, and generate the full set of Major + Minor + Patch (or even change `version_base` to provide these versions separately)

scripts/workspace_status.py
```diff
@@ -0,0 +1,37 @@
...
+Exceptions. See /LICENSE for license information.
+SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+"""
+
+import subprocess
+
+
+def git_commit_sha() -> str:
+    return subprocess.check_output(
+        ["git", "rev-parse", "--short", "HEAD"], shell=False, text=True
```
jonmeow:
```suggestion
        ["git", "rev-parse", "--short", "HEAD"], encoding="utf-8"
```

`shell=False` is the default, and I think they're trying to move away from text (e.g., https://discuss.python.org/t/jep-400-utf-8-by-default-and-future-of-python/14246)

scripts/workspace_status.py
```diff
@@ -0,0 +1,37 @@
...
+    ).strip()
+
+
+def git_dirty_suffix() -> str:
+    status = subprocess.check_output(
+        ["git", "status", "--porcelain"], shell=False, text=True
```
jonmeow:
```suggestion
        ["git", "status", "--porcelain"], encoding="utf-8"
```

scripts/workspace_status.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python3
+
+"""Bazel `--workspace_status_command` script.
+
+This script is designed to be used in Bazel`s `--workspace_status_command` and
+generate any desirable status artifacts.
+"""
```
jonmeow, line 7: Note, to your point about making python a build dependency... I'll just note this could be shell, even letting bzl do the ".dirty" replacement, although I don't know if we're better off consistently using Python in place of shell (I mean, maybe this Python already requires something like WSL on Windows anyways: "On Windows this must be a native binary, typically an ".exe", ".bat", or a ".cmd" file." [here](https://bazel.build/docs/user-manual#workspace-status-command))

jonmeow (commented):
Generally looks good, though had you looked at bazel's expand_template support?

bazel/version/compute_version.bzl
```diff
@@ -0,0 +1,66 @@
...
+                fail("Must provide a non-negative {} number when building that pre-release.".format(pre_release))
+            version += "-{0}.{1}".format(pre_release, number)
+        elif pre_release == "nightly":
+            date = ctx.attr._nightly_date_flag[BuildSettingInfo].value
+            _validate_nightly_date(date)
+            version += "-0.nightly.{}".format(date)
+        else:
+            if pre_release != "dev":
+                fail("Invalid pre-release flag: " + pre_release)
+            version += "-0.dev"
```
jonmeow:
```suggestion
        elif pre_release == "dev":
            version += "-0.dev"
        else:
            fail("Invalid pre-release flag: " + pre_release)
```

Reduces nesting if you check this way

bazel/version/rules.bzl
chandlerc: Yeah, I had tried and it didn't work well.

It expanded keys in more places than expected for me -- so for example a `VERSION` in a comment, etc.

But more importantly, it can't read the status files (and we can't read them here), and so for the status files will require having *some* python template expander already, and it seemed cleaner to just use it everywhere here to avoid two subtly different template syntaxes.

common/version.tmpl.cpp
chandlerc: Yeah, I went back and forth on this. I can go back, I'm of two minds.

Basically, we need both the string form and integer form in C++. We only need the string form in Bazel. So we can either keep the integers all the way from Bazel into here, and we get less parsing of the version string, but more complex logic in Bazel that isn't needed on the Bazel side, more different things are expanded in the template, and it's quite opaque in the C++ code that the integers and the string match exactly.

This way, we have less logic in Bazel, fewer expanded template things, and its obvious the connection. But we have to write the constexpr parsing.

For me, the tie break was that using integer template expansions here makes the syntax trivially broken, and so editting this file is really painful. Whereas here, we can constrain the templated bits to be in string literals (or a weird attribute), and it keeps the syntax checking working which is somewhat nice.

But that's a pretty minor point. WDYT?

scripts/workspace_status.py
chandlerc: My understanding is that Bash has been specifically difficult to make work here compared to Python on Windows. And so that's part of why I went with Python. I also think it's a bit easier for us to maintain. WDYT?

bazel/version/BUILD
chandlerc: Oof, no where currently.

Where would that documentation be most discoverable do you think? Here, in `rules.bzl`, or in the `.bazelrc` where I set up aliases for them?

.bazelrc
chandlerc: See comment on the other thread -- either will add documentation here or add a pointer to that documentation here. =D

bazel/version/compute_version.bzl
chandlerc: Yeah, this could use some more error checking, I'll work on that.

chandlerc (commented):
Still working on some of the suggested improvements, but have a few questions here that would be useful while I'm working on them, so posting those responses. Thanks for the review!

bazel/version/BUILD
jonmeow: I'd suggest here, and point here from bazelrc. My thinking is it might be more discoverable in bazelrc, but bazelrc will become pretty big if we keep adding features and documentation there. (plus, some things like "string or int" is self-documenting here)

bazel/version/rules.bzl
jonmeow: I'm confused this, I don't see VERSION in a comment, maybe I'll poke at it separately... For now this seems fine though.

common/version.tmpl.cpp
jonmeow: What do you think of changing version_base.bzl to look more like:

```
version_major = 0
version_minor = 0
version_patch = 0
```

That way Bazel gets the string form with `"{0}.{1}.{2}".format(version_major, version_minor, version_patch)`, no parsing?

I'm not clear how integer template expansions break things though. What's the issue? Is it that you would need to `str(version_major)`?

scripts/workspace_status.py
```diff
@@ -0,0 +1,37 @@
...
+    return subprocess.check_output(
+        ["git", "rev-parse", "--short", "HEAD"], shell=False, text=True
+    ).strip()
+
+
+def git_dirty_suffix() -> str:
+    status = subprocess.check_output(
+        ["git", "status", "--porcelain"], shell=False, text=True
+    ).strip()
+    return ".dirty" if len(status) > 0 else ""
```
jonmeow, line 28: Please write this out as a full if/else. Maybe this is my familiarity with C++ speaking, but I think Python's ternaries are difficult to read.

common/version.tmpl.cpp
chandlerc: As I said, that was one of the things I tried. =]

The downside is that Bazel only needs the string form, so the integers will only be used for creating the substitutions for the C++ code, and there will be more substitutions there.

It removes the C++ parsing, but I'd feel tempted to keep that code in a static_assert to document that these aren't independent and the string and number match. But at that point, having them broken out in Bazel seems a bit weird.

What would help for me is to understand why this seems better? Is it getting ride of the constexpr parsing code? something else?

bazel/version/gen_tmpl.py
```diff
@@ -0,0 +1,115 @@
...
+            # Remove line endings.
+            line = line.rstrip("\r\n")
+            # Exactly matches our pattern
+            if m := re.fullmatch(_VAR_RE, line):
+                key = m.group("key")
+                if key in substitutions:
+                    value = m.group("value")
+                    if args.verbose:
+                        print(f"Parsed: '{key}': '{value}'", file=sys.stderr)
+                    substitutions.update({key: value})
```
jonmeow: What's actually in the status file? I'm wondering if this could be done without the regex:

```suggestion
            (key, value) = line.split(' ', 1)
            key = key.removeprefix("STABLE_")
            if key in substitutions:
                if args.verbose:
                    print(f"Parsed: '{key}': '{value}'", file=sys.stderr)
                substitutions.update({key: value})
```

bazel/version/gen_tmpl.py
```diff
@@ -0,0 +1,115 @@
...
+
+The initial key/value pairs are read from the command line using repeated
+`--substitute=KEY=DEFAULT-VALUE` flags.
+
+Updated values for those keys will be read from any files provided to the
+`--status-file` flag. This flag can be given multiple times and the values will
+be read and updated from the files in order, meaning the last file's value will
+win. New keys are never read from these files. The file format parsed is chose
+to be compatible with Bazel's status file format, but any files with lines
+matching the regex `^(?P<key>[_A-Za-z][_A-Za-z0-9]) (?P<value>.*)$` may be used.
```
jonmeow: This seems to be different (diverged?) from VAR_RE, would it be worth just pointing to that?

bazel/version/gen_tmpl.py
```diff
@@ -0,0 +1,115 @@
...
+        "--substitution",
+        metavar="KEY=DEFAULT-VALUE",
+        action="append",
+        help="A substitution that should be supported and its default value.",
+    )
+    parser.add_argument(
+        "--status-file",
+        metavar="FILE",
+        type=Path,
+        action="append",
```
jonmeow, line 69:
```suggestion
        action="append",
        default=[],
```

bazel/version/gen_tmpl.py
```diff
@@ -0,0 +1,115 @@
...
+
+    # Collect the supported substitutions from the command line.
+    substitutions = {}
+    for substitution_arg in args.substitution:
+        key, value = substitution_arg.split("=", maxsplit=1)
+        substitutions.update({key: value})
+
+    # Read either of the two status files provided to build up substitutions,
+    # with the stable file last so its values override any duplicates.
+    for status_file in args.status_file if args.status_file else []:
```
jonmeow:
```suggestion
    for status_file in args.status_file:
```

See related suggestion above, part of my "avoid Python ternaries" readability concern. But also here I think the default value is an easier fix.

bazel/version/gen_tmpl.py
```diff
@@ -0,0 +1,115 @@
...
+        action="append",
+        help="A file of key/value updates in Bazel's status file format.",
+    )
+    parser.add_argument("-v", "--verbose", action="store_true")
+    args = parser.parse_args()
+
+    # Collect the supported substitutions from the command line.
+    substitutions = {}
+    for substitution_arg in args.substitution:
+        key, value = substitution_arg.split("=", maxsplit=1)
```
jonmeow: Note, I'm not used to seeing `maxsplit=` explicitly specified since it's a positional parameter.
```suggestion
        key, value = substitution_arg.split("=", 1)
```

bazel/version/gen_tmpl.py
```diff
@@ -0,0 +1,115 @@
...
+The values may not contain newlines or any vertical whitespace.
+
+The initial key/value pairs are read from the command line using repeated
+`--substitute=KEY=DEFAULT-VALUE` flags.
+
+Updated values for those keys will be read from any files provided to the
+`--status-file` flag. This flag can be given multiple times and the values will
+be read and updated from the files in order, meaning the last file's value will
+win. New keys are never read from these files. The file format parsed is chose
+to be compatible with Bazel's status file format, but any files with lines
```
jonmeow: Is this file format specified anywhere?

common/version.tmpl.cpp
jonmeow: I would think it's better because it's less code overall... It gets rid of the constexpr parsing code, and replaces it with a single format string. It's also clearer in version_base that these are integers, not an arbitrary string (i.e., adding more digits "0.0.0.1" will more clearly not work)

jonmeow: Note, I'm not sure what you'd static assert on. Just don't pass the complete string to C++ at all. Have a function that does the format on request.

scripts/workspace_status.py
jonmeow: Either way, I'm not sure if we'll really get out of the WSL dependency; with WSL I don't think it makes a difference, it's only if we're trying to do Windows native (which would require eliminating genrule use too).

jonmeow:
(adding comments to gen_tmpl, sorry I hadn't looked at it originally since I'd been hoping expand_template would work)

bazel/version/compute_version.bzl
jonmeow: Note, this may not be worth treating as gating if you want to get something in today. I think documentation on the flags would be helpful, and may be good enough since not too many people should be setting these.

bazel/version/rules.bzl
chandlerc: It isn't in the current comment -- this was in a previous attempt that had a different structure.

The key thing for me is that the keys in `expand_template` from bazel are expanded undilemited. I wanted *something* to set them apart from just the string content.

That, combined with the need to read from a status file anyways, made it seem like I sohuld just use one templating system with python.

common/version.tmpl.cpp
chandlerc: I'm happy to use three integers in the `version_base.bzl` file if that seems cleaner. No strong opinion there, let me know if that would be an improvement in isolation.

Beyond that:

I think passing the integers as such into C++ still seems like the worse tradeoff to me. It's not a tiny amount of code to set up all the substitutions in starlark. And for me, a single line of starlark is more expensive than several lines of C++ code.

But you can't avoid passing the string to C++ I think... Or at least, I think that would be more complex to try to achieve. You'd need to pass:
- The 3 version numbers
- What kind of pre-release it is (if any)
- A string pre-release suffix (or both a string and a number suffix depending on the kind)

And it's really important that these are assembled into a string using the exact same format in both Starlark and C++. It seemed better to assemble all of these into a string in one place (and that has to be Starlark), and disassemble the components in another. And ideally, to not have the once-decomposed components travel across a really opaque boundary like code generation between starlark and C++.

scripts/workspace_status.py
chandlerc: I'd somewhat like to someday get to Windows native support, but agree its a long way away.

I still find the Python simple enough to prefer over Bash here?

scripts/workspace_status.py
chandlerc: Sure, I was just trying to use the language features. I somewhat agree with you about it being hard to read.

bazel/version/gen_tmpl.py
```diff
@@ -0,0 +1,115 @@
...
+    # Read either of the two status files provided to build up substitutions,
+    # with the stable file last so its values override any duplicates.
+    for status_file in args.status_file:
+        if args.verbose:
+            print(f"Reading status file: {status_file}", file=sys.stderr)
+        for line in status_file.open():
+            # Remove line endings.
+            line = line.rstrip("\r\n")
+            # Exactly matches our pattern
+            (key, value) = line.split(' ', 1)
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
            (key, value) = line.split(" ", 1)
```

bazel/version/gen_tmpl.py
chandlerc: Yes, added a link to the documentation.

bazel/version/gen_tmpl.py
chandlerc: Went with your suggestion to remove the regex, and so tried to just spell out the format here.

bazel/version/compute_version.bzl
chandlerc: I've added a TODO at least, and will follow-up with something.

chandlerc (commented):
Ok, I think I got the requested doc updates (thanks!) and a bunch of the python file improvements in. Other threads have replies. PTAL!

bazel/version/BUILD
```diff
@@ -25,21 +48,38 @@ string_flag(
     ],
 )
 
+# When `--pre_release=rc` is used, `--rc_number=N` set's the release candidate
```
jonmeow:
```suggestion
# When `--pre_release=rc` is used, `--rc_number=N` sets the release candidate
```

bazel/version/BUILD
```diff
@@ -25,21 +48,38 @@ string_flag(
...
+# number to `N`.
+#
+# An error if used without `--pre_release=rc`.
 int_flag(
     name = "rc_number",
     build_setting_default = -1,
 )
 
+# When `--pre_release=beta` is used, `--beta_number=N` set's the beta
```
jonmeow, line 62:
```suggestion
# When `--pre_release=beta` is used, `--beta_number=N` sets the beta
```

bazel/version/BUILD
```diff
@@ -25,21 +48,38 @@ string_flag(
...
+# pre-release number to `N`.
+#
+# An error if used without `--pre_release=beta`.
 int_flag(
     name = "beta_number",
     build_setting_default = -1,
 )
 
+# When `--pre_release=alpha` is used, `--alpha_number=N` set's the alpha
```
jonmeow, line 71:
```suggestion
# When `--pre_release=alpha` is used, `--alpha_number=N` sets the alpha
```

bazel/version/BUILD
```diff
@@ -25,21 +48,38 @@ string_flag(
...
+# pre-release number to `N`.
+#
+# An error if used without `--pre_release=alpha`.
 int_flag(
     name = "alpha_number",
     build_setting_default = -1,
 )
 
+# When `--pre_release=nightly` is used, `--nightly_date=YYYY.MM.DD` set's the
```
jonmeow, line 80:
```suggestion
# When `--pre_release=nightly` is used, `--nightly_date=YYYY.MM.DD` sets the
```

bazel/version/BUILD
```diff
@@ -25,21 +48,38 @@ string_flag(
     ],
 )
 
+# When `--pre_release=rc` is used, `--rc_number=N` set's the release candidate
+# number to `N`.
+#
+# An error if used without `--pre_release=rc`.
```
jonmeow: Optional, just suggesting an alternative boilerplate that's a little more terse. (not going to suggest this everywhere)

```suggestion
# `--rc_number=N` sets the release candidate number to `N`. Requires `--pre_release=rc`.
```

bazel/version/BUILD
```diff
@@ -8,11 +8,34 @@ package(default_visibility = ["//toolchain/install:__pkg__"])
...
     name = "release",
     build_setting_default = False,
 )
 
+# A `--pre_release=KIND` flag where `KIND` must be one of:
+# - `rc` -- a release candidate version.
+# - `beta` -- a beta version.
+# - `alpha` -- an alpha version.
+# - `nightly -- a nightly version.
+# - `dev` -- the default, a development build.
```
jonmeow:
```suggestion
# - `rc` -- a release candidate version.
#    Example: `--pre_release=rc --rc_number=2`
# - `beta` -- a beta version.
#    Example: `--pre_release=beta --beta_number=2`
# - `alpha` -- an alpha version.
#    Example: `--pre_release=alpha --alpha_number=2`
# - `nightly -- a nightly version.
#    Example: `--pre_release=nightly --nightly_date=2024.06.17`
# - `dev` -- the default, a development build.
#    Example: `--pre_release=dev`
```

Minor suggestion to show which require flags.

jonmeow (approved):
Thanks! I think this seems good to merge when you're happy with it.

bazel/version/rules.bzl
jonmeow: FWIW, I think the intent with `expand_template` is you delimit it more yourself (e.g., `VERSION` -> `{VERSION}`). Sort of meh, but not sure there's a good solution for the status file (which also seems a little disappointing, TBH -- but maybe there's something generic that can be done)

common/version.tmpl.cpp
jonmeow: Note, approving right now, but I still think this is something that could maybe be reduced further... I'm maybe more okay with a few boilerplate lines passing values, versus a volume of parsing/validation code. But can address that separately.

chandlerc: Thanks, and happy to talk more about simplifying here and what the options / tradeoffs are.

bazel/version/rules.bzl
chandlerc: All the examples I found with the status file roll their own source generation step. =[

What I'd really like is maybe `#embed` and having the starlark generate some content that we then copy into the status file and read. But anyways, merging for now and can revisit if there are more clean ways in the future.

chandlerc:
I messed up and didn't commit&push the fixes intended in the last comment round here. Very sorry for that. They're in https://github.com/carbon-language/carbon-lang/pull/4059 now.

bazel/version/rules.bzl
jonmeow: Mostly I'd been hoping to find some bazel module that implemented the read. ;)

PR #4055, opened 2024-06-17
Add automated nightly builds, uploaded as pre-releases.

chandlerc:
There is still a lot more we should try to do here, but this gets us
a decent start and should be enough for some of the developers working
on Carbon to use for simple cases, or tools like Compiler Explorer.

This first factors most of the complex setup for our CI workflow into
local actions that we can reuse in other workflows. It's actually
a fairly nice cleanup I think even without the specific use case of
a nightly release process. These use "composite" actions which is the
cleanest model for factoring out a subsequence of steps from a workflow.
I've tried to keep the inputs fairly minimal and easy to spot.

Next, it adds a cron-scheduled action that builds and creates a nightly
release. It also supports explicitly running the workflow, which can be
useful if the nightly release is broken in some way: a fix can be landed
and then a force run used.

Tested the `tests` workflow using our `action-test` branch:
https://github.com/carbon-language/carbon-lang/actions/runs/9545546278

Also tested the nightly release action by hacking in a `push` trigger
and using *draft* release. If you have admin rights, you can see an
example draft nightly build here:
https://github.com/carbon-language/carbon-lang/releases/tag/untagged-6a85cdce7b3a784a2e6b

Note that the GitHub releases are easily deleted, and even the tags
created in the repository can be cleaned up as needed so it should be
fairly harmless to fix this forward if needed.

### DIFF
--- a/.github/workflows/nightly-release.yaml
+++ b/.github/workflows/nightly-release.yaml
@@ -28,6 +28,9 @@ on:
     - cron: '0 2 * * *'
   # Enable manual runs for testing or manually (re-)creating a nightly release.
   workflow_dispatch:
+  # FIXME:
+  push:
+    branches: [action-test]
 
 permissions:
   contents: write # For creating and uploading to releases.
--- a/.github/workflows/nightly-release.yaml
+++ b/.github/workflows/nightly-release.yaml
@@ -52,8 +52,6 @@ jobs:
           echo "remote_cache_upload=--google_credentials=$HOME/remote_cache_key.json" \
               >> $GITHUB_ENV
 
-      - uses: ./.github/actions/build-setup-ubuntu
-
       - uses: ./.github/actions/build-setup-common
         with:
           matrix_runner: ubuntu-22.04
--- a/.github/workflows/nightly-release.yaml
+++ b/.github/workflows/nightly-release.yaml
@@ -62,10 +62,6 @@ jobs:
           echo "nightly_date=$(date '+%Y.%m.%d')" >> $GITHUB_ENV
 
       - name: Build release
-        env:
-          # 'libtool_check_unique failed to generate' workaround.
-          # https://github.com/bazelbuild/bazel/issues/14113#issuecomment-999794586
-          BAZEL_USE_CPP_ONLY_TOOLCHAIN: 1
         run: |
           ./scripts/run_bazel.py \
             --attempts=5 --jobs-on-last-attempt=4 \
--- a/.github/workflows/nightly-release.yaml
+++ b/.github/workflows/nightly-release.yaml
@@ -94,4 +94,4 @@ jobs:
             --notes 'A nightly development build of Carbon.' \
             --prerelease \
             v${{ env.release_version }} \
-            "bazel-bin/toolchain/install/carbon_toolchain-${{ env.release_version }}.tar.bz2#Linux x86-64 binaries - Ubuntu-22.04 (tar.bz2)"
+            "bazel-bin/toolchain/install/carbon_toolchain-${{ env.release_version }}.tar.bz2"
--- a/.github/workflows/nightly-release.yaml
+++ b/.github/workflows/nightly-release.yaml
@@ -28,9 +28,6 @@ on:
     - cron: '0 2 * * *'
   # Enable manual runs for testing or manually (re-)creating a nightly release.
   workflow_dispatch:
-  # FIXME:
-  push:
-    branches: [action-test]
 
 permissions:
   contents: write # For creating and uploading to releases.
@@ -95,7 +92,6 @@ jobs:
           gh release create \
             --title "Nightly build ${{ env.nightly_date }}" \
             --notes 'A nightly development build of Carbon.' \
-            -d \
             --prerelease \
             v${{ env.release_version }} \
             "bazel-bin/toolchain/install/carbon_toolchain-${{ env.release_version }}.tar.bz2#Linux x86-64 binaries - Ubuntu-22.04 (tar.bz2)"
--- a/.github/workflows/nightly-release.yaml
+++ b/.github/workflows/nightly-release.yaml
@@ -95,6 +95,7 @@ jobs:
           gh release create \
             --title "Nightly build ${{ env.nightly_date }}" \
             --notes 'A nightly development build of Carbon.' \
+            -d \
             --prerelease \
             v${{ env.release_version }} \
-            "bazel-bin/toolchain/install/carbon_toolchain-${{ env.release_version }}.tar.bz2#Linux x86-64 binaries - Ubuntu-22.04"
+            "bazel-bin/toolchain/install/carbon_toolchain-${{ env.release_version }}.tar.bz2#Linux x86-64 binaries - Ubuntu-22.04 (tar.bz2)"
--- a/.github/workflows/nightly-release.yaml
+++ b/.github/workflows/nightly-release.yaml
@@ -1,0 +1,99 @@
+# Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+# Exceptions. See /LICENSE for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# This workflow creates a GitHub "release" of a nightly build of the project.
+#
+# Note: This is just an initial rough attempt, there is a lot of future work
+# needed here. A brief summary of TODOs:
+#
+# - Configure a nice release notes template and switch to generating the title
+#   and notes instead of hard coding them.
+#
+# - Do some amount of testing prior to building and uploading the release.
+#   - Tempting to try to examine existing testing workflow, but maybe better to
+#     allow re-using any complex parts and do our own testing. That would, for
+#     example, allow us to narrow or expand the set of tests uses for
+#     pre-release testing to potentially be different from continuous testing.
+#   - Some questions around what to do in the event of a failure... error? Where
+#     does the error go? Create a draft, unpublished release instead?
+#
+# - Build artifacts for all the different OSes we have GitHub runners for rather
+#   than just x86 Linux.
+
+name: Nightly Release
+
+on:
+  schedule:
+    - cron: '0 2 * * *'
+  # Enable manual runs for testing or manually (re-)creating a nightly release.
+  workflow_dispatch:
+
+permissions:
+  contents: write # For creating and uploading to releases.
+
+jobs:
+  release:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@63c24ba6bd7ba022e95695ff85de572c04a18142 # v2.7.0
+        with:
+          egress-policy: audit
+
+      - name: Checkout branch
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - name: Set up remote cache access
+        env:
+          REMOTE_CACHE_KEY: ${{ secrets.CARBON_BUILDS_GITHUB }}
+        run: |
+          echo "$REMOTE_CACHE_KEY" | base64 -d > $HOME/remote_cache_key.json
+          echo "remote_cache_upload=--google_credentials=$HOME/remote_cache_key.json" \
+              >> $GITHUB_ENV
+
+      - uses: ./.github/actions/build-setup-ubuntu
+
+      - uses: ./.github/actions/build-setup-common
+        with:
+          matrix_runner: ubuntu-22.04
+          remote_cache_upload: ${{ env.remote_cache_upload }}
+
+      - name: Get nightly date
+        run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yaml
```diff
@@ -245,63 +88,19 @@ jobs:
...
-          build --jobs=32
+      - uses: ./.github/actions/build-setup-macos
+        if: steps.filter.outputs.has_code == 'true' && env.os == 'macos'
+        with:
+          matrix_runner: ${{ matrix.runner }}
 
-          # Avoid any cache impact from build stamping in CI.
-          build --nostamp
+      - uses: ./.github/actions/build-setup-ubuntu
+        if: steps.filter.outputs.has_code == 'true' && env.os == 'ubuntu'
```
jonmeow: How programmable are these? Can you do something like:

```
      - uses: ./.github/actions/build-setup-${{ env.os }}
        if: steps.filter.outputs.has_code == 'true'
```

(which could be used to unify macos and ubuntu)

.github/workflows/nightly-release.yaml
```diff
@@ -0,0 +1,99 @@
...
+      - name: Create the release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release create \
+            --title "Nightly build ${{ env.nightly_date }}" \
+            --notes 'A nightly development build of Carbon.' \
+            --prerelease \
+            v${{ env.release_version }} \
+            "bazel-bin/toolchain/install/carbon_toolchain-${{ env.release_version }}.tar.bz2#Linux x86-64 binaries - Ubuntu-22.04"
```
jonmeow: What's your system for choosing these release asset names? Is there a project you're using as a baseline for the format?

To detail what I'm seeing, Carbon is: "Linux x86-64 binaries - Ubuntu-22.04"; actual file: carbon_toolchain-0.0.0-0.nightly.2024.06.17.tar.bz2

Comparing with some other projects that have release assets:

- https://github.com/llvm/llvm-project/releases has: clang+llvm-18.1.7-aarch64-linux-gnu.tar.xz
- https://github.com/bazelbuild/bazel/releases has: bazel-7.2.1rc1-linux-x86_64
- https://github.com/microsoft/TypeScript/releases/tag/v5.4.5 has: typescript-5.4.5.tgz

Concretely, my suggestion would be:

- Maybe drop "toolchain" since this is probably the normal install
- Add the target platform to the filename (this especially because it seems standard for platform-specific binaries)
- Consider removing the alias altogether (at least `Ubuntu-22.04` unless there's a reason for mentioning that)

i.e., make the asset name something like "carbon-${{ env.release_version }}-linux-x86_64.tar.bz2"

.github/workflows/nightly-release.yaml
```diff
@@ -0,0 +1,99 @@
+# Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+# Exceptions. See /LICENSE for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# This workflow creates a GitHub "release" of a nightly build of the project.
```
jonmeow, line 5: Should this be accompanied by something like https://github.com/marketplace/actions/delete-releases to limit how many nightly releases remain?

.github/workflows/nightly-release.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+      - name: Extract the release version
+        run: |
+          # Make sure we can run the toolchain to get the version.
+          ./bazel-bin/toolchain/install/prefix_root/bin/carbon version
+
+          # Now stash it in a variable and export it.
+          VERSION=$( \
+            ./bazel-bin/toolchain/install/prefix_root/bin/carbon version \
+            | cut -d' ' -f5 | cut -d'+' -f1)
```
jonmeow, line 81: Maybe there should be built-in support to get just the version string? e.g., `version --simple`

.github/workflows/nightly-release.yaml
```diff
@@ -0,0 +1,99 @@
...
+          remote_cache_upload: ${{ env.remote_cache_upload }}
+
+      - name: Get nightly date
+        run: |
+          echo "nightly_date=$(date '+%Y.%m.%d')" >> $GITHUB_ENV
+
+      - name: Build release
+        env:
+          # 'libtool_check_unique failed to generate' workaround.
+          # https://github.com/bazelbuild/bazel/issues/14113#issuecomment-999794586
```
jonmeow: This looks like a macos issue; you're seeing it on ubuntu too?

.github/workflows/nightly-release.yaml
```diff
@@ -0,0 +1,99 @@
...
+          echo "$REMOTE_CACHE_KEY" | base64 -d > $HOME/remote_cache_key.json
+          echo "remote_cache_upload=--google_credentials=$HOME/remote_cache_key.json" \
+              >> $GITHUB_ENV
+
+      - uses: ./.github/actions/build-setup-ubuntu
+
+      - uses: ./.github/actions/build-setup-common
+        with:
+          matrix_runner: ubuntu-22.04
+          remote_cache_upload: ${{ env.remote_cache_upload }}
```
jonmeow: FWIW, and I'll leave this as a choice to you because it's not clearly better, but ubuntu/macos could call common directly... If you moved in the remote cache access logic (which is already duplicated here and tests.yaml), I think it'd just mean passing in matrix_runner to ubuntu explicitly (which might be good anyways?).

(depending on how far you wanted to go, maybe the checkout could be in a separate action along with the remote cache access?)

.github/workflows/nightly-release.yaml
```diff
@@ -0,0 +1,99 @@
...
+        run: |
+          echo "$REMOTE_CACHE_KEY" | base64 -d > $HOME/remote_cache_key.json
+          echo "remote_cache_upload=--google_credentials=$HOME/remote_cache_key.json" \
+              >> $GITHUB_ENV
+
+      - uses: ./.github/actions/build-setup-ubuntu
+
+      - uses: ./.github/actions/build-setup-common
+        with:
+          matrix_runner: ubuntu-22.04
```
jonmeow, line 57: Would using `matrix_runner: ${{ matrix.runner }}` be valid here even though it's not explicitly set? i.e., does GitHub have it as defaulting from runs-on?

jonmeow (approved):
I agree this is a good refactoring... I'm assuming most of the content of the copied files hasn't changed. If you think I should review closer, please create a commit that copies tests.yaml around (before the edits) so that I can see the delta in a separate commit.

.github/workflows/tests.yaml
jonmeow: (note, I could very much see this not working...)

.github/workflows/nightly-release.yaml
chandlerc: A few factors here...

Do we use a human-oriented description after the `#` or just the base filename? One benefit of a human-oriented name is that it gives us a good way to explain restrictions on where the binary artifacts will work that may not be apparent from a file name. Eventually, I'd like all of those to be embedded in the filename though, it's just not there yet today. But these will also be displayed next to source code downloads that will use human-oriented text regardless:
image
I'll at least add matching `(tar.bz2)` here


I think we do need to telegraph two restrictions on where these binaries can be used:
- Only on Linux machines, and not all Linux.
  - It needs to have glibc with at least the same version as that used on Ubuntu 22.04 (the host we build on).
- Only x86-64 machines.

The x86-64 restriction is one that I want to show up in the filename and I left a TODO about fixing this in the install code. But until then, I was trying to document it in the English alias. I suspect this is also the one where we'll start adding packages to cover the alternatives first.

The Linux and glibc version restriction is tricky. We're not building fully Linux-hermetic binaries at this point, and until we do, there is basically a "as-modern-as" requirement that is somewhat hard to document precisely in a way people understand. The clearest I came up with is "it was built on this OS version, so if a binary built for that OS works for you, you're good". Which is true, but maybe an odd way of explaining it. Happy to have better suggestions here for how we spell this. If it is something more than "linux", it may be tricky to wire through to the filename in Bazel, and so it may be something that has to be handled here in the action -- either renaming the file or adding an alias to clarify.

All this said, I'm happy to change this to whatever folks like for now. Just let me know, this isn't a case where I have any opinion or am worried about any of the alternatives really. They all seem fine.

.github/workflows/nightly-release.yaml
chandlerc: From what I can tell, there is no limit on how many we have. And it seems nice to have an archive of old releases that we can try out, for example to speed bisection, etc.

That said, not sure we need to use marketplace stuff, it's a trivial command line action to write if we want to remove or otherwise tidy up the releases.

.github/workflows/nightly-release.yaml
chandlerc: I haven't tried this, but the docs make me think that this is *only* provided with a `strategy` based job.

.github/workflows/nightly-release.yaml
chandlerc: The checkout is the only thing that *cannot* move -- that's how we find the actions.

Moving the remote cache access is tricky because it uses secrets. You can push those into actions, but you have to do so explicitly and with some care. Given that, and that only this one needs the read-only form, and how short the read-write form was, it wasn't clear this would be a benefit.

On whether to have the OS-setup ones call the common one, I went back and forth. I'm still not sure. Actually, maybe there is a factoring that is strictly better here. Tried something out, WDYT?

.github/workflows/nightly-release.yaml
chandlerc: No, I just didn't want to deviate. And I'd like to add macOS here eventually, so it seemed better to keep for now?

.github/workflows/nightly-release.yaml
chandlerc: Yeah, probably. I'd like to somewhat revisit how the version printing is wired up in the command line library more generally. The current setup doesn't work well, and I have a few different ideas for improvements.

But I didn't want to make even more changes in this iteration.

.github/workflows/tests.yaml
chandlerc: After my refactoring it's trickier. I still tried, and it didn't work. =[ But after the refacotring, this is at least contained to one place now so maybe that's OK.

chandlerc (commented):
Thanks for all the review! PTAL.

.github/workflows/nightly-release.yaml
jonmeow: To me, this is something you should just leave out. I'd at least prefer comment explaining that it's macOS specific and that you're keeping it for a speculative release.

.github/workflows/nightly-release.yaml
jonmeow: Looking through the top 10 C++ GitHub projects according to https://github.com/EvanLi/Github-Ranking/blob/master/Top100/CPP.md, half distribute assets; of those, all use filenames, and only gpt4all diverges in the naming scheme (swapping version and platform). For quick reference:

- https://github.com/electron/electron/releases/tag/v31.0.1
- https://github.com/microsoft/terminal/releases/tag/v1.20.11381.0
- https://github.com/godotengine/godot/releases/tag/4.2.2-stable
- https://github.com/opencv/opencv/releases/tag/4.10.0
- https://github.com/nomic-ai/gpt4all/releases/tag/v2.8.0
- https://github.com/protocolbuffers/protobuf/releases/tag/v27.1

So the filename naming scheme looks very common, and users should be expected be familiar with it. I expect it has not encountered readability issues, given the lack of label adoption. Maybe the lack of divergence indicates the filename structure has been working well?

Filenames do have other benefits beyond the simple consistency, such as knowing what filename to expect on download without further examination. If you still prefer a label, I'd suggest removing "binaries" or replacing it, because the library assets would probably remain part.

Note, for communicating the Ubuntu 22.04 requirement, you could add it as a suffix, such as "carbon_toolchain-0.0.0-0.nightly.2024.06.17.tar.bz2 (built on Ubuntu 22.04)". Just a thought for maintaining the filename idiom while still getting you the extra platform label.

Note though, I've approved. I don't think this is worth continuing to debate if you feel strongly.

.github/workflows/nightly-release.yaml
jonmeow: Yeah, that looks good. Thanks!

jonmeow (approved):

.github/workflows/nightly-release.yaml
chandlerc: Nice data looking at the existing projects! =D

I really didn't have a strong opinion here, I'm happy to go with just the filename and work on getting Bazel to put more information there rather than stressing about adding it in annotations.

.github/workflows/nightly-release.yaml
chandlerc: Ok, I've removed for now. Can add back if/when we have a macos builder.

PR #4056, opened 2024-06-17
Initial support for CR+LF (DOS / Windows) line endings.

chandlerc:
It turns out we can make these work with very minimal complexity because the LF is still in the right place either way. This also lets us easily support mixtures of LF and CR+LF line endings gracefully. We create the line structures around the LF bytes and then have the byte-dispatch loop notice a CR followed by an LF and skip to the LF behavior.

Rather than add the remaining complexity around supporting bare CR and LF+CR sequences (both of which are quite rare now), this just adds diagnostics when we encounter a CR byte that won't fall out of our CR+LF handling. This is a better experience for users than the alternative. We still have a TODO to handle the full complexity of vertical whitespace, but I've updated it to reflect that the common case should be handled already.

This isn't complete though: we need to add support in string literal lexing, and we need to teach the diagnostic rendering to handle the error messages above better. But those will be future PRs, this is enough to unblock folks who happen to edit a Carbon source file with notepad on Windows which seems important.


toolchain/lex/lex.cpp
```diff
@@ -708,6 +720,27 @@ auto Lexer::LexVerticalWhitespace(llvm::StringRef source_text,
...
+    LexVerticalWhitespace(source_text, position);
+    return;
+  }
+
+  CARBON_DIAGNOSTIC(UnsupportedLFCRLineEnding, Error,
+                    "The LF+CR line ending is not supported, only LF and CR+LF "
+                    "is supported.");
+  CARBON_DIAGNOSTIC(
+      UnsupportedCRLineEnding, Error,
+      "A raw CR line ending is not supported, only LF and CR+LF is supported.");
```
zygoloid:
```suggestion
                    "are supported.");
  CARBON_DIAGNOSTIC(
      UnsupportedCRLineEnding, Error,
      "A raw CR line ending is not supported, only LF and CR+LF are supported.");
```

toolchain/lex/lex.cpp
```diff
@@ -708,6 +720,27 @@ auto Lexer::LexVerticalWhitespace(llvm::StringRef source_text,
...
+  bool is_lfcr = position > 0 && source_text[position - 1] == '\n';
+  emitter_.Emit(source_text.begin() + position,
+                is_lfcr ? UnsupportedLFCRLineEnding : UnsupportedCRLineEnding);
```
zygoloid, line 741: Do we do anything to prevent the CR in the snippet from garbling the diagnostic output?

toolchain/lex/lex.cpp
```diff
@@ -708,6 +720,27 @@ auto Lexer::LexVerticalWhitespace(llvm::StringRef source_text,
...
+  const auto* line_info = current_line_info();
+  position = line_info->start + line_info->length;
```
zygoloid: I wonder if treating the `\r` like horizontal whitespace would give us better recovery than skipping to the end of the line (or to the end of the file if it exclusively uses `\r` as a line break).

zygoloid (approved):
Seems fine. Neither of the comments without a suggestion is blocking for this PR.

I look forward to receiving bug reports from people building Carbon code on MacOS 9 or RiscOS :)

toolchain/lex/lex.cpp
chandlerc: It doesn't garble it, but yeah, the diagnostics are still not great here. I mentioned this as future work in the PR description, but wanted to get the minimal functional thing out quickly. =]

toolchain/lex/lex.cpp
```diff
@@ -708,6 +720,27 @@ auto Lexer::LexVerticalWhitespace(llvm::StringRef source_text,
...
+    LexVerticalWhitespace(source_text, position);
+    return;
+  }
+
+  CARBON_DIAGNOSTIC(UnsupportedLFCRLineEnding, Error,
+                    "The LF+CR line ending is not supported, only LF and CR+LF "
+                    "are supported.");
+  CARBON_DIAGNOSTIC(
+      UnsupportedCRLineEnding, Error,
+      "A raw CR line ending is not supported, only LF and CR+LF are supported.");
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  CARBON_DIAGNOSTIC(UnsupportedCRLineEnding, Error,
                    "A raw CR line ending is not supported, only LF and CR+LF "
                    "are supported.");
```

toolchain/lex/lex.cpp
chandlerc: Yeah, thinking about it, that's going to work much better. Made that change (and simplified the test a touch).

chandlerc (commented):
Thanks!

PR #4058, opened 2024-06-17
Update bazel modules.

jonmeow:
With bazel 7.2.0, there are some dependency changes, resulting in:

```
WARNING: For repository 'bazel_skylib', the root module requires module version bazel_skylib@1.5.0, but got bazel_skylib@1.6.1 in the resolved dependency graph.
WARNING: For repository 'platforms', the root module requires module version platforms@0.0.8, but got platforms@0.0.9 in the resolved dependency graph.
```

This goes around doing some updates... protobuf's include structure has changed a little too.

PR #4063, opened 2024-06-18
Update google_benchmark and remove the patch

jonmeow:
Updating to 1.8.4 breaks the patch file, so I'm looking at solutions that don't require maintaining a patch.

Verifying this is working with `bazel build //toolchain/lex:tokenized_buffer_benchmark && strings bazel-bin/toolchain/lex/tokenized_buffer_benchmark |& grep pfm`


jonmeow:
Depends on #4058, just trying to avoid managing conflicting module changes in parallel.

chandlerc (approved):
LGTM, verified that this does in fact work for me with hardware perf counters data in a benchmark.

PR #4066, opened 2024-06-19
Add assignment support to the hashtables.

chandlerc:
This lets copy and move assignment work. While it's a bit suboptimal to do assignment with these tables, it still seems like an unreasonable burden to not allow the basics to work. Even the toolchain ended up doing this in a few places.

### DIFF
--- a/common/raw_hashtable.h
+++ b/common/raw_hashtable.h
@@ -1607,8 +1607,8 @@ auto TableImpl<InputBaseT, SmallSize>::small_storage() const -> Storage* {
   }
 }
 
-// Helper to setup the storage of a table when a specific size has already been
-// setup. If possible, uses any small storage, otherwise allocates.
+// Helper to set up the storage of a table when a specific size has already been
+// set up. If possible, uses any small storage, otherwise allocates.
 template <typename InputBaseT, ssize_t SmallSize>
 auto TableImpl<InputBaseT, SmallSize>::SetupStorage() -> void {
   CARBON_DCHECK(this->small_alloc_size() == SmallSize);
--- a/common/raw_hashtable.h
+++ b/common/raw_hashtable.h
@@ -1510,8 +1510,10 @@ TableImpl<InputBaseT, SmallSize>::TableImpl(const TableImpl& arg)
   CARBON_DCHECK(arg.small_alloc_size_ == SmallSize);
   CARBON_DCHECK(this->small_alloc_size_ == SmallSize);
 
-  SetUpStorage();
-  this->CopySlotsFrom(arg);
+  if (this->alloc_size() != 0) {
+    SetUpStorage();
+    this->CopySlotsFrom(arg);
+  }
 }
 
 template <typename InputBaseT, ssize_t SmallSize>
@@ -1650,7 +1652,8 @@ template <typename InputBaseT, ssize_t SmallSize>
 auto TableImpl<InputBaseT, SmallSize>::SetUpStorage() -> void {
   CARBON_DCHECK(this->small_alloc_size() == SmallSize);
   ssize_t local_size = this->alloc_size();
-  if (SmallSize > 0 && local_size == SmallSize) {
+  CARBON_DCHECK(local_size != 0);
+  if (local_size == SmallSize) {
     this->storage() = small_storage();
   } else {
     this->storage() = BaseT::Allocate(local_size);
--- a/common/raw_hashtable.h
+++ b/common/raw_hashtable.h
@@ -517,7 +517,10 @@ class BaseImpl {
   auto small_alloc_size() const -> ssize_t {
     return static_cast<unsigned>(small_alloc_size_);
   }
-  auto is_small() const -> bool { return alloc_size() <= small_alloc_size(); }
+  auto is_small() const -> bool {
+    CARBON_DCHECK(alloc_size() >= small_alloc_size());
+    return alloc_size() == small_alloc_size();
+  }
 
   auto Construct(Storage* small_storage) -> void;
   auto Destroy() -> void;
@@ -586,7 +589,7 @@ class TableImpl : public InputBaseT {
 
   auto small_storage() const -> Storage*;
 
-  auto SetupStorage() -> void;
+  auto SetUpStorage() -> void;
 
   [[no_unique_address]] mutable SmallStorage small_storage_;
 };
@@ -1141,9 +1144,13 @@ auto BaseImpl<InputKeyT, InputValueT, InputKeyContextT>::Destroy() -> void {
 // Copy all of the slots over from another table that is exactly the same
 // allocation size.
 //
-// This requires the current table to already have storage allocated and setup
+// This requires the current table to already have storage allocated and set up
 // but not initialized (or already cleared). It directly overwrites the storage
 // allocation of the table to match the incoming argument.
+//
+// Despite being used in construction, this shouldn't be called for a moved-from
+// `arg` -- in practice it is better for callers to handle this when setting up
+// storage.
 template <typename InputKeyT, typename InputValueT, typename InputKeyContextT>
 auto BaseImpl<InputKeyT, InputValueT, InputKeyContextT>::CopySlotsFrom(
     const BaseImpl& arg) -> void {
@@ -1172,13 +1179,26 @@ auto BaseImpl<InputKeyT, InputValueT, InputKeyContextT>::CopySlotsFrom(
 
 // Move from another table to this one.
 //
-// This requires the table to have size and growth already setup but nothing
-// else. Notably, storage should either not yet be constructed or already
-// destroyed. It both sets up the storage and handles any moving slots needed.
+// Note that the `small_storage` is *this* table's small storage pointer,
+// provided from the `TableImpl` to this `BaseImpl` method as an argument.
+//
+// Requires the table to have size and growth already set up but otherwise the
+// the table has not yet been initialized. Notably, storage should either not
+// yet be constructed or already destroyed. It both sets up the storage and
+// handles any moving slots needed.
+//
+// Note that because this is used in construction it needs to handle a
+// moved-from `arg`.
 template <typename InputKeyT, typename InputValueT, typename InputKeyContextT>
 auto BaseImpl<InputKeyT, InputValueT, InputKeyContextT>::MoveFrom(
     BaseImpl&& arg, Storage* small_storage) -> void {
   ssize_t local_size = alloc_size();
+  CARBON_DCHECK(local_size == arg.alloc_size());
+  // If `arg` is moved-from, skip the rest as the local size is all we need.
+  if (local_size == 0) {
+    return;
+  }
+
   if (arg.is_small()) {
     CARBON_DCHECK(local_size == small_alloc_size_);
     this->storage() = small_storage;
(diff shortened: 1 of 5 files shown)


common/raw_hashtable.h
```diff
@@ -1130,6 +1138,82 @@ auto BaseImpl<InputKeyT, InputValueT, InputKeyContextT>::Destroy() -> void {
...
+  memcpy(local_metadata, local_arg_metadata, local_size);
+
+  for (ssize_t group_index = 0; group_index < local_size;
+       group_index += GroupSize) {
+    auto g = MetadataGroup::Load(local_arg_metadata, group_index);
+    for (ssize_t byte_index : g.MatchPresent()) {
+      local_entries[group_index + byte_index].CopyFrom(
+          local_arg_entries[group_index + byte_index]);
+    }
+  }
```
zygoloid, line 1177: Is it worth detecting the trivially-copyable case here and doing a `memcpy` rather than copying only the present elements?

common/raw_hashtable.h
```diff
@@ -1130,6 +1138,82 @@ auto BaseImpl<InputKeyT, InputValueT, InputKeyContextT>::Destroy() -> void {
...
+}
+
+// Move from another table to this one.
+//
+// This requires the table to have size and growth already setup but nothing
+// else. Notably, storage should either not yet be constructed or already
+// destroyed. It both sets up the storage and handles any moving slots needed.
+template <typename InputKeyT, typename InputValueT, typename InputKeyContextT>
+auto BaseImpl<InputKeyT, InputValueT, InputKeyContextT>::MoveFrom(
+    BaseImpl&& arg, Storage* small_storage) -> void {
```
zygoloid, line 1194: It would have helped me as a reader for either the name of `small_storage` or the function comment to indicate that `small_storage` is the small storage of `*this`, not `arg`. (Having `small_storage` as an argument to `MoveFrom` suggested to me initially that it was part of what we were moving from.)

common/raw_hashtable.h
```diff
@@ -1130,6 +1138,82 @@ auto BaseImpl<InputKeyT, InputValueT, InputKeyContextT>::Destroy() -> void {
...
+           group_index += GroupSize) {
+        auto g = MetadataGroup::Load(local_arg_metadata, group_index);
+        for (ssize_t byte_index : g.MatchPresent()) {
+          local_entries[group_index + byte_index].MoveFrom(
+              std::move(local_arg_entries[group_index + byte_index]));
+        }
+      }
+    }
+  } else {
+    // Just point to the allocated storage.
```
zygoloid, line 1227: Would it make sense to `CARBON_DCHECK(local_size == arg.alloc_size());` here, or maybe before the `if`?

common/raw_hashtable.h
```diff
@@ -1130,6 +1138,82 @@ auto BaseImpl<InputKeyT, InputValueT, InputKeyContextT>::Destroy() -> void {
...
+    storage() = arg.storage();
+  }
+
+  // Finally, put the incoming table into a moved-from state.
+  arg.alloc_size() = 0;
+  // Replace the pointer with null to ease debugging.
+  arg.storage() = nullptr;
```
zygoloid, line 1234: Hm. `is_small()` returns `alloc_size() <= small_alloc_size()`, not `alloc_size() == small_alloc_size()`, so a moved-from hash table is small but its storage pointer doesn't point to its small storage. I found that surprising, but as far as I can tell, the hash table doesn't assume that `storage()` points to the small storage when `is_small()` is true, just that the storage doesn't need to be deallocated. So maybe that's all fine and working as intended.

A comment on `is_small` would help explain this is intentional.

common/raw_hashtable.h
```diff
@@ -1399,79 +1483,78 @@ BaseImpl<InputKeyT, InputValueT, InputKeyContextT>::GrowAndInsert(
 template <typename InputBaseT, ssize_t SmallSize>
 TableImpl<InputBaseT, SmallSize>::TableImpl(const TableImpl& arg)
     : BaseT(arg.alloc_size(), arg.growth_budget_, SmallSize) {
+  CARBON_DCHECK(arg.alloc_size() > 0) << "Cannot copy from a moved-from table!";
   CARBON_DCHECK(arg.small_alloc_size_ == SmallSize);
+  CARBON_DCHECK(!arg.is_small() || arg.alloc_size() == SmallSize);
```
zygoloid: Hm. I would not expect the copy constructor to have preconditions on its argument, so copying from an object in a valid but unspecified state seems like it ought to work -- and as far as I can tell, the standard library expects that too. (Even P2345R0, which sought to remove the requirement for most operations to work on moved-from objects that are used with the standard library, retained the requirement to be able to constructor or assign from a moved-from object.)

This is frustrating, though, since it really doesn't seem like something that will come up a lot in practice outside of buggy or inefficient code -- and I think we'd need an extra branch here to handle it. (`std::swap(x, x)` does perform a move from a moved-from object, but also seems somewhat questionable.)

common/raw_hashtable.h
```diff
@@ -1399,79 +1483,78 @@ BaseImpl<InputKeyT, InputValueT, InputKeyContextT>::GrowAndInsert(
...
+    // implementation so detect and bypass here.
+    if (&arg == this) {
+      return *this;
+    }
+    CARBON_DCHECK(arg.storage() != this->storage());
+    // The sizes match, so just clear out old entries.
+    if constexpr (!EntryT::IsTriviallyDestructible) {
+      this->view_impl_.ForEachEntry([](EntryT& entry) { entry.Destroy(); },
+                                    [](auto...) {});
     }
```
zygoloid: Do we care about the case where these destructor calls modify either `*this` or `arg`? (I think and hope the answer is no.)

common/raw_hashtable.h
```diff
@@ -1399,79 +1483,78 @@ BaseImpl<InputKeyT, InputValueT, InputKeyContextT>::GrowAndInsert(
...
-              std::move(local_arg_entries[group_index + byte_index]));
-        }
-      }
-    }
-  } else {
-    // Just point to the allocated storage.
-    this->storage() = arg.storage();
-  }
+  // Destroy and deallocate our table.
+  this->Destroy();
```
zygoloid, line 1586: As far as I can tell, the latest LWG thinking is that self move assignment is valid but leaves the object in an unspecified state. I think this is enough to make that work:
```suggestion
  this->Destroy();
  this->alloc_size() = 0;
```

common/raw_hashtable.h
```diff
@@ -1524,6 +1607,19 @@ auto TableImpl<InputBaseT, SmallSize>::small_storage() const -> Storage* {
   }
 }
 
+// Helper to setup the storage of a table when a specific size has already been
+// setup. If possible, uses any small storage, otherwise allocates.
```
zygoloid:
```suggestion
// Helper to set up the storage of a table when a specific size has already been
// set up. If possible, uses any small storage, otherwise allocates.
```

zygoloid: ("setup" is a noun; the verb is two words.) I'd also be inclined to rename the function -> `SetUpStorage.`

common/raw_hashtable.h
zygoloid: Hmm, I think I was wrong here, and self move assignment is only supposed to work for a moved from object, for which I think this code already works.

common/raw_hashtable.h
chandlerc: Probably, but I didn't want to re-work the functionality as well here. Can do that in a follow-up?

common/raw_hashtable.h
chandlerc: Added a comment. Happy to do more if you can think of a naming change that would help...

common/raw_hashtable.h
chandlerc: Done.

common/raw_hashtable.h
chandlerc: I don't think we should be calling `is_small` on a moved-from table at all.

I've switched `is_small` to be `==` and added a `DCHECK` that it is never smaller.

common/raw_hashtable.h
chandlerc: Dug and definitely need to handle moved-from throughout this. Bleh.

Added tests and reworked code.

It's a bit messy, and I'm not 100% satisfied... But I think cleaning it up requires a deeper refactoring. I have an idea for that, but as it's a bigger change I'd like to defer that for a future PR. PTAL.

common/raw_hashtable.h
chandlerc: Nope. At least, my intent is not to permit that. =]

common/raw_hashtable.h
chandlerc: I had thought that self move mustn't be permitted at all because its already gone through `std::move`.... But now that you've mentioned it, that doesn't add up at all.

Even if the standard actually does't allow other forms of self move, I'm much more comfortable defending against it. Made equivalent change with a comment explaining things and a test.

chandlerc (commented):
Thanks, PTAL

common/raw_hashtable.h
zygoloid: Absolutely.

common/raw_hashtable.h
```diff
@@ -1399,79 +1503,95 @@ BaseImpl<InputKeyT, InputValueT, InputKeyContextT>::GrowAndInsert(
...
+  // Check for completely broken objects. These invariants should be true even
+  // in a moved-from state.
+  CARBON_DCHECK(arg.alloc_size() == 0 || !arg.is_small() ||
+                arg.alloc_size() == SmallSize);
   CARBON_DCHECK(arg.small_alloc_size_ == SmallSize);
+  CARBON_DCHECK(this->small_alloc_size_ == SmallSize);
 
-  ssize_t local_size = arg.alloc_size();
+  SetUpStorage();
+  this->CopySlotsFrom(arg);
```
zygoloid:
```suggestion
  if (this->alloc_size() != 0) {
    SetUpStorage();
    this->CopySlotsFrom(arg);
  }
```
I think we want copying from a moved-from object to not allocate / copy.

common/raw_hashtable.h
```diff
@@ -1524,6 +1644,19 @@ auto TableImpl<InputBaseT, SmallSize>::small_storage() const -> Storage* {
   }
 }
 
+// Helper to set up the storage of a table when a specific size has already been
+// set up. If possible, uses any small storage, otherwise allocates.
+template <typename InputBaseT, ssize_t SmallSize>
+auto TableImpl<InputBaseT, SmallSize>::SetUpStorage() -> void {
+  CARBON_DCHECK(this->small_alloc_size() == SmallSize);
+  ssize_t local_size = this->alloc_size();
+  if (SmallSize > 0 && local_size == SmallSize) {
```
zygoloid:
```suggestion
  CARBON_DCHECK(this->small_alloc_size() == SmallSize);
  ssize_t local_size = this->alloc_size();
  CARBON_DCHECK(local_size != 0);
  if (local_size == SmallSize) {
```
I think we can't have `SmallSize == 0 && local_size == SmallSize` here, because we should never call this when `local_size == 0`.

common/raw_hashtable.h
zygoloid: Progress, not perfection :)

common/raw_hashtable.h
chandlerc: I'm... not sure how this change got lost. Sorry. I swear I had this exact code in there last time I looked. =]

common/raw_hashtable.h
chandlerc: Ah, true, thanks!

zygoloid (approved):

PR #4067, opened 2024-06-20
Merge the suffix ops into a single box in the operator precedence mermaid diagram

josh11b:
The current approach was done to work-around a limitation that we can only have a single link per box, but is unscalable. Instead have a single link to a new sections of the document that describes the box, and has multiple links.


docs/design/expressions/README.md
```diff
@@ -345,6 +337,23 @@ The binary arithmetic and bitwise operators also have
...
+    side. The right-hand side is a name.
+-   The [compound member access operators](member_access.md), `x.(...)` and
+    `x->(...)`, have an expression as their second operand, but put that
+    expression in parenthesis and so it doesn't participate in the precedence
+    considerations of its first operand.
+-   The [indexing operator](indexing.md), `x[y]`, similarly puts its second
+    operand in matching brackets.
+-   The call operator, `x(...)`, takes a comma-separated list of arguments, but
+    again puts them in parenthesis that clearly separate them for precedence
+    purposes.
```
zygoloid:
```suggestion
-   [Member access operators](member_access.md), like `x.y` and the
    dereferencing variant `x->y`, only have an expression on their left-hand
    side. The right-hand side is a name.
-   The [compound member access operators](member_access.md), `x.(...)` and
    `x->(...)`, have an expression as their second operand, but put that
    expression in parentheses and so it doesn't participate in the precedence
    considerations of its first operand.
-   The [indexing operator](indexing.md), `x[y]`, similarly puts its second
    operand in matching square brackets.
-   The call operator, `x(...)`, takes a comma-separated list of arguments, but
    again puts them in parentheses that clearly separate them for precedence
    purposes.
```

"in parenthesis" -> "in parentheses" because I'd interpret the former as meaning "as an insertion into the surrounding context" and the latter as meaning specifically "within two parenthesis tokens".

"bracket" -> "square bracket" to disambiguate for at least British English where "bracket" broadly means any kind of bracketing (parentheses, braces, square brackets, angle brackets) and narrowly means parentheses.

docs/design/expressions/README.md
```diff
@@ -345,6 +337,23 @@ The binary arithmetic and bitwise operators also have
 [compound assignment](/docs/design/assignment.md) forms. These are statements
 rather than expressions, and do not produce a value.
 
+## Suffix ops
```
zygoloid: "ops" here is a bit ambiguous -- is this "operators" or "operations"? (Or is that ambiguity intentional?) I think this is clear, but avoiding the abbreviation might be better.

zygoloid (approved):

docs/design/expressions/README.md
josh11b: Done.

PR #4069, opened 2024-06-21
Update FunctionContext::Inserter::InsertHelper for llvm changes

dklimkin:
The interface has changed upstream:

https://github.com/llvm/llvm-project/commit/80f881485accb020345ee7e1c4c3151ec55ce590

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -128,7 +128,7 @@ http_archive(
         "@carbon//bazel/llvm_project:0001_Patch_for_mallinfo2_when_using_Bazel_build_system.patch",
         "@carbon//bazel/llvm_project:0002_Added_Bazel_build_for_compiler_rt_fuzzer.patch",
     ],
-    sha256 = "4c53512522cfd625a75aa7f201e6932f18ebf3fbe8ee89e05883c3d30ca935ac",
+    sha256 = "3216d56b308d420170d6284d04560175b56fd56065a7442f32b732c606e77367",
     strip_prefix = "llvm-project-{0}".format(llvm_project_version),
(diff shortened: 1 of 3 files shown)


chandlerc (commented):
Really appreciate the fix!

We'll need to bump LLVM to the same revision in the PR to pick up the API change:
https://github.com/carbon-language/carbon-lang/blob/trunk/MODULE.bazel#L120

Do you want to do that? If not, always happy to take over the PR and get that added and merged.

dklimkin:
Thanks, exactly what I was looking for! I re-pinned now, let's see if it breaks anything else...

MODULE.bazel
```diff
@@ -116,8 +116,8 @@ bazel_dep(name = "zstd", version = "1.5.6", repo_name = "llvm_zstd")
 
 # We pin to specific upstream commits and try to track top-of-tree reasonably
 # closely rather than pinning to a specific release.
-# HEAD as of 2024-05-17.
-llvm_project_version = "a68d20e986053ec571223a9f3ead3e146a27dc82"
+# HEAD as of 2024-06-20.
+llvm_project_version = "80f881485accb020345ee7e1c4c3151ec55ce590"
```
chandlerc, line 120: Will likely need a sha256 update below and a .lock file update.

If you can run bazel it will provide the updates to the lock file automatically and tell you the sha.

=\ sorry, I know bumping the version is a bit annoying.

dklimkin: Updated SHA. I could get .lock to update with bazel.

chandlerc (approved):
Awesome, thanks!!!

PR #4070, opened 2024-06-21
Remove another hashtable iteration order dependency.

chandlerc:
Name scopes store the names in their scope in a `DenseMap`. Several places reasonably avoid depending on the iteration order by sorting the names -- they're in the formatting code path where that's a solid approach.

Unfortunately, when we're importing one scope into another, we also need to walk the entire scope and do something for each name. =[ This doesn't seem like a great place to sort things to stabilize them.

I've switched to a fairly simplistic solution of having a vector of name entries that can be iterated stably, and a separate map for lookups. I didn't use the set-of-indices trick here because it's not clear that's the right trade-off for a scope: likely a lot of small scopes here with relatively hot name lookups. And the key here isn't a large or dynamically sized thing that we're canonicalizing, it's a `NameId`. That made me lean towards duplicating the name in the hashtable for lookup and the vector for iteration.

I thought about a fancy approach of sorting the hashtable keys by their values (the indices), but that would still require a bit of copying and more code.

I also thought a bit about other optimizations, but decided to leave a comment for now -- it's not obvious to me exactly how hot this is and whether it's better served by faster lookups, being more memory dense, etc. And that might involve more of an SOA layout change or some other approach. Rather than do that here, and especially before switching hashtables, I stuck with a simple approach to address the ordering.

### DIFF
--- a/toolchain/sem_ir/name_scope.h
+++ b/toolchain/sem_ir/name_scope.h
@@ -19,6 +19,7 @@ enum class AccessKind : int8_t {
 
 struct NameScope : Printable<NameScope> {
   struct Entry {
+    NameId name_id;
     InstId inst_id;
     AccessKind access_kind;
   };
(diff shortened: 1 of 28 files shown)


toolchain/sem_ir/name_scope.h
```diff
@@ -35,24 +36,35 @@ struct NameScope : Printable<NameScope> {
...
+  // Adds a name to the scope that must not already exist.
+  auto AddRequired(Entry name_entry) -> void {
+    int index = names.size();
+    names.push_back(name_entry);
+    auto success = name_map.insert({name_entry.name_id, index}).second;
+    CARBON_CHECK(success) << "Failed to add required name: "
+                          << name_entry.name_id;
+  }
+
+  // Names in the scope. We store both an insertion-orderd vector for iterating
```
zygoloid:
```suggestion
  // Names in the scope. We store both an insertion-ordered vector for iterating
```

zygoloid (approved):
I'm a bit sad to be storing order information just to support the same-package namespace scope import use case (which won't be relevant for most name scopes), but per your comment, we can revisit if needed. Looks like this is also fixing at least one iterator invalidation bug :)

chandlerc:
> Looks like this is also fixing at least one iterator invalidation bug :)

Just for posterity -- I don't think so... but maybe in a very surprising way. Because this is a `DenseMap` and not a `SmallDenseMap`, the iterator into that map isn't invalidated by the map being moved when the vector containing scopes (and thus the map object itself) grows and reallocates.

This PR adds a *direct* use of the parent scope, which is what actually blows up here when we try to access `parent_scope->names` after injecting more scopes.

But ... yeah, that was a surprising part of this PR to debug.

toolchain/sem_ir/name_scope.h
```diff
@@ -35,24 +36,35 @@ struct NameScope : Printable<NameScope> {
...
+  // and a map from `NameId` to the index of that vector for name lookup.
+  //
+  // Optimization notes: this is somewhat memory inefficient. If this ends up
+  // either hot or a significant source of memory allocation, we should consider
+  // switching to a SOA model where the `AccessKind` is stored in a separate
+  // vector so that these can pack densely. If this ends up both cold and memory
+  // intensive, we can also switch the lookup to a set of indices into the
+  // vector rather than a map from `NameId` to index.
+  llvm::SmallVector<Entry> names;
+  llvm::DenseMap<NameId, int> name_map = llvm::DenseMap<NameId, int>();
```
jonmeow, line 67: Should this have been an int32_t? (note, mainly commenting because, are there other cases of this?)

PR #4072, opened 2024-06-22
Add a key-returning callback insert to `Set`.

chandlerc:
This turns out to be super useful now that we have the `key_context` mechanism and can do much more meaningful heterogeneous lookups, where the stored key can be *very* different from the lookup key.

### DIFF
--- a/common/set.h
+++ b/common/set.h
@@ -212,9 +212,6 @@ class SetBase
   // The point of this function is when the lookup key is _different_from the
   // stored key. However, we don't restrict it in case that blocks generic
   // usage.
-  //
-  // TODO: The `;` formatting below appears to be bugs in clang-format with
-  // concepts that should be filed upstream.
   template <typename LookupKeyT, typename KeyCallbackT>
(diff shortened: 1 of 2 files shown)


common/set.h
```diff
@@ -204,6 +204,24 @@ class SetBase
...
+  // produce a new key when no existing value is found.
+  //
+  // Example: `m.Insert(key_equivalent, [] { return real_key; });`
+  //
+  // The point of this function is when the lookup key is _different_from the
+  // stored key. However, we don't restrict it in case that blocks generic
+  // usage.
+  //
+  // TODO: The `;` formatting below appears to be bugs in clang-format with
+  // concepts that should be filed upstream.
```
zygoloid:
```suggestion
```
For whatever reason, this one seems to format OK! *shrug* Maybe it's the parens?

zygoloid (approved):
Looks great, thanks!

common/set.h
chandlerc: no idea, but yay!

PR #4075, opened 2024-06-22
Change operator precedence

josh11b:
Update the operator precedence to achieve a few goals:

-   Form operators into groups which behave similarly
-   Make the group of operators ("top-level operators") that capture everything to the right, like `if`...`then`...`else`, behave similarly to the left, so that rearranging expressions won't change how they group.
-   Add the `where` operator, used to specify constraints on facet types, to the precedence chart, to define how it interacts with other operators.
-   Make the operator precedence diagram prettier, so that it eventually can be made into a poster that Carbon programmers can hang on their walls.


proposals/p4075.md
```diff
@@ -0,0 +1,115 @@
...
+    made into a poster that Carbon programmers can hang on their walls.
+
+## Problem
+
+The `where` operator is particularly tricky:
+
+-   It is used in an `impl` declaration to specify the values of associated
+    constants (such as associated types). In that context, `impl T as I where R`
+    is interpreted conceptually as `impl T as (I where R)`. It would be nice if
+    `T as I where R` would mean the same thing in other contexts.
```
zygoloid: Maybe worth noting here: if we don't get the "would be nice" version, we want it to be invalid rather than meaning `(T as I) where R`. That is, considered in isolation, we would prefer `T as (I where R)` > invalid > `(T as I) where R`.

proposals/p4075.md
```diff
@@ -0,0 +1,129 @@
...
+-   The `where` operator will frequently be used with the binary `&` operator,
+    since that is how facet types are combined. It is desirable that
+    `I & J where R` be interpreted as `(I & J) where R`. This is expected to be
+    more common than combinging `where` and `as` outside of an `impl`
```
zygoloid:
```suggestion
    more common than combining `where` and `as` outside of an `impl`
```

proposals/p4075.md
```diff
@@ -0,0 +1,129 @@
...
+
+The `where` operator is particularly tricky:
+
+-   It is used in an `impl` declaration to specify the values of associated
+    constants (such as associated types). In that context, `impl T as I where R`
+    is interpreted conceptually as `impl T as (I where R)`. It would be nice if
+    `T as I where R` would mean the same thing in other contexts.
+-   The `where` operator will frequently be used with the binary `&` operator,
+    since that is how facet types are combined. It is desirable that
+    `I & J where R` be interpreted as `(I & J) where R`. This is expected to be
```
zygoloid: Likewise here: considered in isolation, we probably prefer `(I & J) where R` > invalid > `I & (I where R)`.

zygoloid (commented):
Looks good to me.

chandlerc (commented):
Just wanted to say I'm generally happy with this. Leaving the detail review to @zygoloid. I'm also not concerned if we discover we want to make a further tweak here based on experience, this still seems like a good improvement and bringing the `where` clause into the picture.

josh11b (commented):
Note to self: include link to open discussion on 06-13 https://docs.google.com/document/d/1s3mMCupmuSpWOFJGnvjoElcBIe2aoaysTIdyczvKX84/edit?resourcekey=0-G095Wc3sR6pW1hLJbGgE0g&tab=t.0#heading=h.p46elxrmhh8x

proposals/p4075.md
josh11b: Done.

proposals/p4075.md
josh11b: Done.

josh11b:
> Note to self: include link to open discussion on 06-13 https://docs.google.com/document/d/1s3mMCupmuSpWOFJGnvjoElcBIe2aoaysTIdyczvKX84/edit?resourcekey=0-G095Wc3sR6pW1hLJbGgE0g&tab=t.0#heading=h.p46elxrmhh8x

Done.

chandlerc (approved):
LGTM and approving for leads.

PR #4076, opened 2024-06-22
Switch to a `carbon_binary` rule with target config support.

chandlerc:
This switches from a macro that simply wraps genrules to a proper Starlark rule that runs first compile and then link actions.

Most interestingly, this uses the rule structure to allow using the Carbon toolchain built either in the target config or the exec config. While the exec config is more principled and even necessary in a cross-compile situaiton, it is dramatically less efficient when developing Carbon as all the binaries and tests outside of our examples will be built with the target config. This triggers a complete second build of the toolchain in the exec config for examples before this PR.

It is tempting to try to keep the exec config but make it not cause redundant actions, but the way Bazel sets up exec and target config makes it essentially impossible to share their artifacts. There used to be a hack in Bazel itself to force sharing but it was removed due to it violating the principled design. Instead, these rules are explicit about their intent to use the target config, much like a test would be.

I have rigged up a flag that is carefully threaded through a wrapper macro with `select`s to allow easily switching to the exec configuration in case it is desired or needed. But the the `.bazelrc` sets the default to the target config. The `BUILD` file default is the principled `exec` in case these rules are used by importing into some other Bazel workspace where we might *only* need the exec config.

The net outcome of this is shaving over 2500 actions off of a clean rebuild such as is triggered by a version bump to LLVM, including some of the very slow and expensive compiles of LLVM and Clang themselves. These would only be triggered if you built the examples so this may mostly impact our CI latency.

### DIFF
--- a/bazel/carbon_rules/defs.bzl
+++ b/bazel/carbon_rules/defs.bzl
@@ -60,7 +60,7 @@ def _carbon_binary_impl(ctx):
     )
     return [DefaultInfo(files = depset([bin]))]
 
-carbon_binary_internal = rule(
+_carbon_binary_internal = rule(
     implementation = _carbon_binary_impl,
     attrs = {
         # The exec config toolchain driver and data. These will be `None` when
--- a/bazel/carbon_rules/defs.bzl
+++ b/bazel/carbon_rules/defs.bzl
@@ -101,7 +101,7 @@ def carbon_binary(name, srcs):
       name: The name of the build target.
       srcs: List of Carbon source files to compile.
     """
-    carbon_binary_internal(
+    _carbon_binary_internal(
         name = name,
         srcs = srcs,
 
--- a/bazel/carbon_rules/defs.bzl
+++ b/bazel/carbon_rules/defs.bzl
@@ -64,7 +64,10 @@ _carbon_binary_internal = rule(
     implementation = _carbon_binary_impl,
     attrs = {
         # The exec config toolchain driver and data. These will be `None` when
-        # using the target config and populated when using the exec config.
+        # using the target config and populated when using the exec config. We
+        # have to use duplicate attributes here and below to have different
+        # `cfg` settings, as that isn't `select`-able, and we'll use `select`s
+        # when populating these.
         "internal_exec_toolchain_data": attr.label(
             cfg = "exec",
         ),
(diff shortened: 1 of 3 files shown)


bazel/carbon_rules/defs.bzl
```diff
@@ -23,32 +27,91 @@ def carbon_binary(name, srcs):
...
+    """Compiles a Carbon binary.
+
+    Args:
+      name: The name of the build target.
+      srcs: List of Carbon source files to compile.
+    """
+    carbon_binary_internal(
+        name = name,
+        srcs = srcs,
+        internal_exec_toolchain_driver = select({
```
jonmeow, line 112: Out of curiosity, why does this provide exec and target as separate flags instead of the same flag? Is it because you need the difference `cfg` value in the attr? This might be worth a comment explaining (highlighting) the `cfg` nuance.

Had you tried something like:

```
cfg = select({
  # Use the same driver as for the target platform, reducing built artifacts.
  "//bazel/carbon_rules:use_target_config_carbon_rules_config": "target",
  # Use separate host and target drivers.
  "//conditions:default": "exec",
}),
```

(if this works, I'd suggest it to reduce duplication of attrs, but I'm not sure with bazel)

jonmeow (approved):
Looks good! Not sure if you wanting zygoloid to specifically check, given discussion around this.

bazel/carbon_rules/defs.bzl
jonmeow: Unfortunately doesn't work. Still, would appreciate a comment specific to why the flags are separated.

bazel/carbon_rules/defs.bzl
```diff
@@ -23,32 +27,91 @@ def carbon_binary(name, srcs):
...
+        inputs = objs,
+        executable = toolchain_driver,
+        tools = depset(toolchain_data),
+        arguments = ["link", "--output=" + bin.path] + [o.path for o in objs],
+        mnemonic = "CarbonLink",
+        progress_message = "Linking " + bin.short_path,
+    )
+    return [DefaultInfo(files = depset([bin]))]
+
+carbon_binary_internal = rule(
```
jonmeow:
```suggestion
_carbon_binary_internal = rule(
```

(note the call would also need to be changed below)

bazel/carbon_rules/defs.bzl
chandlerc: Yeah, sadly this is the only way I could find around the `cfg` setting. Added a comment though, as it is quite mysterious.

PR #4077, opened 2024-06-23
Fix PR labeling and generate nightly build notes from them.

chandlerc:
This configures the nightly build to generate release notes and provides a template for organizing them. The organization is done through labeling of PRs and categorizing them based on those labels. I've tried to provide a rough categorization that seems reasonable for folks.

While doing this I looked at our PR labeling and found a few bugs that were preventing many labels form being attached. I've fixed those and significantly expanded the coverage of file-based labeling. I've also added code to do author-based labeling for automated PRs so those can be separated out from human PRs.

### DIFF
--- a/.github/release.yaml
+++ b/.github/release.yaml
@@ -15,19 +15,20 @@ changelog:
         - proposal
       exclude.labels:
         - bot
-    - title: 'Toolchain changes :hammer_and_wrench:'
+    - title: 'Toolchain and implementation changes :hammer_and_wrench:'
       labels:
         - toolchain
-      exclude.labels:
-        - bot
-    - title: 'Explorer changes :telescope:'
-      labels:
         - explorer
       exclude.labels:
         - bot
     - title: 'Documentation changes :memo:'
       labels:
         - documentation
+      exclude.labels:
+        - bot
+    - title: 'Utilities :triangular_ruler:'
+      labels:
+        - utilities
       exclude.labels:
         - bot
     - title: 'Infrastructure changes :building_construction:'
--- a/.github/release.yaml
+++ b/.github/release.yaml
@@ -1,0 +1,43 @@
+# Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+# Exceptions. See /LICENSE for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# Provide the configuration and categorization for generated release notes.
+
+changelog:
+  exclude:
+    labels:
+      - ignore-for-release
+
+  categories:
+    - title: Proposals Accepted and Merged 📜
+      labels:
+        - proposal
+      exclude.labels:
+        - automated
+    - title: Toolchain Changes 🛠
+      labels:
+        - toolchain
+      exclude.labels:
+        - automated
+    - title: Explorer Changes 🔭
+      labels:
+        - explorer
+      exclude.labels:
+        - automated
+    - title: Documentation Changes 🖺
+      labels:
+        - documentation
+      exclude.labels:
+        - automated
+    - title: Infrastructure Changes 🏗
+      labels:
+        - infrastructure
+      exclude.labels:
+        - automated
+    - title: Automated Updates 🤖
+      labels:
+        - automated
+    - title: Other Changes
+      labels:
+        - '*'
--- a/.github/release.yaml
+++ b/.github/release.yaml
@@ -10,34 +10,34 @@ changelog:
       - ignore-for-release
 
   categories:
-    - title: Proposals Accepted and Merged 📜
+    - title: 'Proposals accepted and merged :scroll:'
       labels:
         - proposal
       exclude.labels:
-        - automated
-    - title: Toolchain Changes 🛠
+        - bot
+    - title: 'Toolchain changes :hammer_and_wrench:'
       labels:
         - toolchain
       exclude.labels:
-        - automated
-    - title: Explorer Changes 🔭
+        - bot
+    - title: 'Explorer changes :telescope:'
       labels:
         - explorer
       exclude.labels:
-        - automated
-    - title: Documentation Changes 🖺
+        - bot
+    - title: 'Documentation changes :memo:'
       labels:
         - documentation
       exclude.labels:
-        - automated
-    - title: Infrastructure Changes 🏗
+        - bot
+    - title: 'Infrastructure changes :building_construction:'
       labels:
         - infrastructure
       exclude.labels:
-        - automated
-    - title: Automated Updates 🤖
+        - bot
+    - title: 'Automated robot PRs :robot:'
       labels:
-        - automated
-    - title: Other Changes
+        - bot
+    - title: 'Other changes'
       labels:
         - '*'
--- a/.github/workflows/label_prs.yaml
+++ b/.github/workflows/label_prs.yaml
@@ -28,9 +28,11 @@ jobs:
               - 'docs/**'
               - 'examples/**'
               - 'third_party/examples/**'
+
             explorer:
               - 'explorer/**'
               - 'installers/**'
+
             infrastructure:
               - '*.bzl'
               - '*.cfg'
@@ -44,14 +46,23 @@ jobs:
               - 'github_tools/**'
               - 'proposal/scripts/**'
               - 'scripts/**'
+
+            # Here we only want the `proposal` label when a *new* file is added
+            # directly in this directory. We use `added` and a single level glob
+            # to achieve that.
             proposal:
               - added: 'proposals/*'
+
+            # We include common, shared code into the toolchain label for
+            # convenience. Essentially, this is everything we intend to ship as
+            # part of the reference implementation of the language.
             toolchain:
               - 'common/**'
               - 'core/**'
               - 'language_server/**'
               - 'testing/**'
               - 'toolchain/**'
+
             utilities:
               - 'utils/**'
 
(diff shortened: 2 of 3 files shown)


.github/release.yaml
```diff
@@ -0,0 +1,43 @@
...
+      labels:
+        - toolchain
+      exclude.labels:
+        - automated
+    - title: Explorer Changes 🔭
+      labels:
+        - explorer
+      exclude.labels:
+        - automated
+    - title: Documentation Changes 🖺
```
jonmeow: This unicode isn't rendering for me.

Had you considered using [GitHub's supported emoji](https://github-emoji-picker.vercel.app/), maybe avoiding rendering issues that way?  e.g.,`:scroll:` in place of the unicode. Just not sure whether GitHub would handle the replacement on generation.

`:hammer_and_wrench:` is what made me choose this particular link (others missed it)
:hammer_and_wrench:

.github/workflows/label_prs.yaml
```diff
@@ -26,20 +26,36 @@ jobs:
...
               - '.*'
               - '.*/**'
+              - 'BUILD'
+              - 'MODULE.*'
+              - 'WORKSPACE'
               - 'bazel/**'
+              - 'common/**'
+              - 'github_tools/**'
+              - 'installers/**'
+              - 'language_server/**'
```
jonmeow: Before, the infrastructure cluster was really low-level stuff (project infrastructure, not code support). While common and testing could be in the middle somewhere, I think the language server and utils are very high level: they're focused on IDE support. Similarly, third_party seems like a mix (e.g. examples). Had you considered a "misc" label or similar for things like that?

.github/workflows/label_prs.yaml
```diff
@@ -55,17 +71,35 @@ jobs:
...
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           PR: ${{ github.event.pull_request.html_url }}
 
+      - id: proposal
+        if: steps.filter.outputs.proposal == 'true'
+        run: |
+          gh pr edit "${PR}" --add-label "proposal"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PR: ${{ github.event.pull_request.html_url }}
```
jonmeow, line 99: Proposals should be getting labels from the proposal_labeled and proposal_ready flows, are you seeing issues with them? I'm wary of having multiple actions trying to set labels.

.github/workflows/label_prs.yaml
```diff
@@ -55,17 +71,35 @@ jobs:
...
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           PR: ${{ github.event.pull_request.html_url }}
+
+      - id: automated
+        if:
+          contains(fromJSON('["CarbonInfraBot", "dependabot"]'),
+          github.event.pull_request.user.login)
+        run: |
+          gh pr edit "${PR}" --add-label "automated"
```
jonmeow, line 124: Minor, but maybe "bot" for a shorter label name? These will add up when we have multiple labels...

.github/release.yaml
```diff
@@ -0,0 +1,43 @@
...
+#
+# Provide the configuration and categorization for generated release notes.
+
+changelog:
+  exclude:
+    labels:
+      - ignore-for-release
+
+  categories:
+    - title: Proposals Accepted and Merged 📜
```
jonmeow: Minor, but you might consider "Proposals accepted and merged" for our documentation style (similar below)

jonmeow (commented):
This looks like a good improvement for the releases, I see I missed a bug in PR labeling.

.github/workflows/label_prs.yaml
jonmeow: Or are there non-bot automated PRs you also plan to use this for? What's the definition that you want for an "automated" PR? If I do something like clang-tidy fixing, is that automated?

.github/workflows/label_prs.yaml
chandlerc: I'm happy breaking things up more if we can come up with good category names?

I'd like to avoid "misc" as I'd like to keep things fairly descriptive, and it seems good to highlight changes to common code or "infrastructure" of any form... We don have a catch-all bucket for other changes, which I should probably move above the bot changes.

But ideally, I'd like to find more helpful descriptive categories for as much as we can. Happy to add "IDE and editor tooling" or some better phrasing for `utils/` and `language_server/` as I agree those are quite different. Any good ideas for `common/`? Seems really good to include, but not sure the right phrasing. (I'm less worried about `third_party/` as there just isn't much there, and I'll move the examples part to be documentation like the top-level examples are.)

chandlerc (commented):
(still not back to where I can hack on this, but asking a quick question below that would be useful for me when I do get a chance to circle back)

.github/workflows/label_prs.yaml
jonmeow: Perhaps a simple "third-party" label for third_party changes? For utils and language_server, I don't have a specific suggestion, because everything I come down to is essentially "misc". I mean, there's also "utils", but that's basically "misc" (although maybe "language_server" should've been under "utils" and then "utils" would make more sense)

.github/release.yaml
chandlerc: Sure, done.

.github/release.yaml
chandlerc: Sure, and thanks for the link to find all of these.

.github/workflows/label_prs.yaml
chandlerc: So, I believe that `proposal_labeled.yaml` only handles keeping updates between the draft, rfc, accepted, and declined labels consistent. And `proposal_ready.yaml` only handles a PR that has some proposal label getting marked as `ready_for_review`, and triggering the `proposal rfc` label for that. A human has to apply the `proposal` (or `proposal draft`) label first.

And those workflows don't seem the right place to do this `proposal` labeling as we really want to trigger on the *paths* here and more events. So I think this is a reasonably clean breakdown of logic based on the events that trigger workflows.

I can try to merge all the proposal label management into a single workflow if you want, but we already have 3 actions setting labels, and so I don't think this PR is making things much worse there.

.github/workflows/label_prs.yaml
chandlerc: Happy to go with `bot`. My goal was specifically to only flag fully bot-managed PRs, anything else seems too complicated to reason about.

.github/workflows/label_prs.yaml
zygoloid: How about something like this:

* `tooling`: :hammer_and_wrench: (stuff that Carbon developers use for interacting with Carbon code)

              - 'language_server/**'
              - 'utils/**'

* `infrastructure`: :gear: (stuff that we use to maintain the repository etc.)

              - 'github_tools/**'
              - 'installers/**'
              - 'proposal/scripts/**'
              - 'scripts/**'

* (not sure of the best name) `foundation` / `common` / `support`: :jigsaw: (stuff that we link into the toolchain that's not toolchain-specific)

              - 'testing/**'
              - 'common/**'
              - 'third_party/**'

(Maybe we should split `testing` and `common` out from `third_party`, based on who maintains them. I'm neutral on that.)

zygoloid: We're already using :hammer_and_wrench: for the toolchain but could switch that to :chains: maybe?

chandlerc: After more discussion on Discord, I think I've implemented the direction folks liked, PTAL!

chandlerc (commented):
Thanks for the review, PTAL!

.github/workflows/label_prs.yaml
```diff
@@ -39,20 +41,19 @@ jobs:
...
-              - 'common/**'
               - 'github_tools/**'
-              - 'installers/**'
-              - 'language_server/**'
               - 'proposal/scripts/**'
               - 'scripts/**'
-              - 'testing/**'
-              - 'third_party/**'
-              - 'utils/**'
             proposal:
```
jonmeow, line 53: Please call out why this is using "added"

.github/workflows/label_prs.yaml
```diff
@@ -39,20 +41,19 @@ jobs:
...
               - added: 'proposals/*'
```
jonmeow, line 54: Should this exclude proposals/scripts?

.github/workflows/label_prs.yaml
jonmeow: So, new_proposal.py adds the proposal label. Should PRs that don't use the new_proposal script that add files here generally be marked as proposals? That will probably include non-proposal things.

jonmeow (approved):

.github/workflows/label_prs.yaml
chandlerc: I believe it does because it only is one `*`, so this should only match added files (or new directories) directly under `proposals`.

.github/workflows/label_prs.yaml
chandlerc: Added comments.

.github/workflows/label_prs.yaml
chandlerc: As discussed, hopefully this actually matches what we want and gives coverage for proposals that aren't created with the script.

I've also added a comment to highlight how it is different.

chandlerc:
Thanks, merging. That said, if any of the comments need improving, happy to do that in a follow-up!

Issue #2110, opened 2022-08-25
Debug output/printing syntax


jonmeow:
What should debug output/printing syntax look like?

Note, I think there's a related question of "how should the prelude be implemented". However, I think there are questions about printing that are at least somewhat separate. I think there are three key pieces:

- [Rosetta](https://rosettacode.org/wiki/Print_itself) for comparison
        - C++: `std::cout` and `fprintf` are available as options.
        - Python: [`print`](https://docs.python.org/3/library/functions.html#print) prints multiple arguments with space separators and a newline. `end=''` for no newline, `sep=''` for no separator.
            - Swift: [`print`](https://developer.apple.com/documentation/swift/print(_:separator:terminator:)) is Python-like.
        - Go: [`fmt.Print`](https://pkg.go.dev/fmt#Print) prints multiple arguments with no separators and no newline. `fmt.Prinln` for a newline, `fmt.Printf` for formatted printing.
        - Rust: [`print!`](https://doc.rust-lang.org/rust-by-example/hello/print.html) does formatted printing with no newline. `println!` for a newline.
        - Kotlin: [`print`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/print.html) prints a single argument with no newline. `println` for a newline.
        - JS/TS: [`console.log`](https://developer.mozilla.org/en-US/docs/Web/API/Console/log) takes either multiple objects to print appended, or the first argument can be a substitution string and later arguments are substituted.

At present, explorer provides `Print`, as do front page [snippets](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/images/snippets.md). The design does [Console.Print](https://github.com/carbon-language/carbon-lang/tree/trunk/docs/design#hello-carbon), changing to [Carbon.Print with #2017](https://github.com/carbon-language/carbon-lang/pull/2017) based on a discussion about where the library should really be located modulo the prelude.

What I'm trying to get answered here isn't where the debug printing function belongs library-wise, but rather what the idiomatic syntax should look like (potentially desugared).

1. What should the name be?
    - Ignoring casing, is "print" or "Print" agreeable?
    - Should it be namespaced to a package and require something more like `Carbon.Print`?
    - Should it be `print` and have a keyword reservation that points at the actual symbol?
2. Should multiple arguments be allowed?
    - @chandlerc had brought up the idea of requiring a single, formatted string as a way to avoid function syntax.
        - For reference, this was partly to avoid having a function-like keyword.
[...]

PR #4078, opened 2024-06-24
Add a basic Core.Print function for ints.

jonmeow:
We'd been discussing that explorer remains necessary for print, and I was wondering if this kind of approach would be okay (we _probably_ want this to work, based on #2110, albeit with more overloads -- but I don't think there's a good way to support overloads at the moment).

```
╚╡../bazel-bin/examples/sieve
2
3
5
7
11
13
17
19
23
29
31
37
41
43
...
```


zygoloid (approved):
As a strictly short-term thing, I think this is fine. Maybe there should be a TODO somewhere to remove this once we can implement `Print` in the language instead?

jonmeow:
> As a strictly short-term thing, I think this is fine. Maybe there should be a TODO somewhere to remove this once we can implement `Print` in the language instead?

Done

PR #4079, opened 2024-06-24
Diagnose missing definitions in impl files

geoffromer:


toolchain/check/testdata/class/no_definition_in_impl_file.carbon
```diff
@@ -0,0 +1,118 @@
...
+
+// --- a.impl.carbon
+
+impl library "a";
+
+class AForwardDecl;
+
+class AForwardDecl {}
+
+// --- b.carbon
```
jonmeow: What do you think of adding two more test cases:

1. The API file has no forward declaration, the impl file has a forward declaration.
2. The API file has a forward declaration, the impl file has nothing.

I think both of these would be handled correctly; I'm only thinking about test coverage.

toolchain/check/testdata/class/no_definition_in_impl_file.carbon
```diff
@@ -0,0 +1,118 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+// TIP: To test this file alone, run:
+// TIP:   bazel test //toolchain/testing:file_test --test_arg=--file_tests=toolchain/check/testdata/class/no_definition_in_impl_file.carbon
```
jonmeow: Can more of these tests be moved into the `no_prelude` subdirectory? That would reduce dependencies if they don't require the prelude support.

toolchain/check/testdata/impl/no_definition_in_impl_file.carbon
```diff
@@ -0,0 +1,221 @@
...
+// TIP: To dump output, run:
+// TIP:   bazel run //toolchain/testing:file_test -- --dump_output --file_tests=toolchain/check/testdata/impl/no_definition_in_impl_file.carbon
+
+// --- a.carbon
+
+library "a";
+
+interface AInterface {}
+
+impl i32 as AInterface;
```
jonmeow: Since `i32` requires the prelude, perhaps this test could have fewer dependencies using something like `()` instead?

toolchain/check/testdata/class/no_definition_in_impl_file.carbon
```diff
@@ -0,0 +1,118 @@
...
+
+library "b";
+
+class BForwardDecl;
+
+// --- fail_b.impl.carbon
+
+impl library "b";
+
+import library "a";
```
jonmeow: Where you do `import library "a"` in these tests, is that required to test something? It might be helpful to name the files (a, b) after what you're trying to test, to emphasize the intent, similar to how you'd name a TEST function when using googletest.

toolchain/check/handle_function.cpp
```diff
@@ -303,6 +303,10 @@ static auto BuildFunctionDecl(Context& context,
     }
   }
 
+  if (context.is_impl_file() && !is_extern) {
```
jonmeow: Would it be worth checking `is_definition` here? i.e., only adding forward declarations? (possibly similar in the other cases, too)

toolchain/check/context.h
```diff
@@ -481,6 +488,14 @@ class Context {
   //
   // Inline 0 elements because it's expected to require heap allocation.
   llvm::SmallVector<SemIR::ConstantValueStore, 0> import_ir_constant_values_;
+
+  // True if the file being processed is an impl file.
+  bool is_impl_file_;
+
+  // Declaration instructions of entities that should have definitions by the
+  // end of the current source file. Currently only supports ClassDecl,
+  // FunctionDecl, and ImplDecl.
```
jonmeow: Looks like you have a TODO for interfaces in the implementation, maybe leave it commented in just the one place? (this comment made me look for what's missing)

toolchain/check/check.cpp
```diff
@@ -656,6 +657,43 @@ auto NodeIdTraversal::Next() -> std::optional<Parse::NodeId> {
...
 }
 
+// Emits a diagnostic for each declaration in context.definitions_required()
+// that doesn't have a definition.
+static auto DiagnoseMissingDefinitions(Context& context,
+                                       Context::DiagnosticEmitter& emitter)
+    -> void {
+  CARBON_DIAGNOSTIC(MissingDefinitionInImpl, Error,
+                    "No definition found for declaration in impl file");
+  for (SemIR::InstId decl_inst : context.definitions_required()) {
```
jonmeow: We typically suffix ID variables with "_id", can you please do similar here?
```suggestion
  for (SemIR::InstId decl_inst_id : context.definitions_required()) {
```

toolchain/check/check.cpp
```diff
@@ -656,6 +657,43 @@ auto NodeIdTraversal::Next() -> std::optional<Parse::NodeId> {
...
+        if (!context.impls().Get(impl_decl.impl_id).is_defined()) {
+          emitter.Emit(decl_inst, MissingDefinitionInImpl);
+        }
+        break;
+      }
+      // TODO: handle `interface` as well, once we can test it without
+      // triggering https://github.com/carbon-language/carbon-lang/issues/4071
+      default: {
+        CARBON_CHECK(false);  // FIXME
+      }
```
jonmeow: WDYT of being more explicit with this, such as:
```suggestion
      case CARBON_KIND(SemIR::InterfaceDecl interface_decl): {
        CARBON_FATAL() << "TODO: Add interfaces.";
      }
      default: {
        CARBON_FATAL() << "Unexpected inst in definitions_required: " << inst;
      }
```

Note that specific code would require storing the `inst` you have in the switch separately, we haven't made it available in CARBON_KIND_SWITCH (maybe should, just haven't).

toolchain/check/context.cpp
```diff
@@ -33,7 +33,7 @@ namespace Carbon::Check {
 
 Context::Context(const Lex::TokenizedBuffer& tokens, DiagnosticEmitter& emitter,
                  const Parse::Tree& parse_tree, SemIR::File& sem_ir,
-                 llvm::raw_ostream* vlog_stream)
+                 llvm::raw_ostream* vlog_stream, bool is_impl_file)
```
jonmeow: Versus taking this as a parameter, had you considered querying sem_ir? It should be `sem_ir.import_irs().Get(SemIR::ImportIRId::ApiForImpl).sem_ir != nullptr`

jonmeow (commented):
Looks good! I was considering a different approach for definitions_required, but this makes sense. Just a few suggestions for small changes.

For context, I think we could also iterate over classes, functions, etc. If it's undefined, check the declaration's LocId, and warn if is_node_id (which should only occur if it's local). However, that scales based on the number of imported entities, whereas this scales based on the number of declared (possibly forward declared, per comments) entities, which may be worthwhile for the storage cost tradeoff.

toolchain/check/check.cpp
geoffromer: Good idea; done.

(And whoops, I should have caught that FIXME before sending)

toolchain/check/context.cpp
geoffromer: Oh, cool, I didn't think of that. Done.

toolchain/check/handle_function.cpp
geoffromer: Oh, right, I had that originally but it got lost during some restructuring. Done.

`BuildImplDecl` doesn't currently have `is_definition`. Do you think it's worth plumbing that through?

toolchain/check/testdata/class/no_definition_in_impl_file.carbon
geoffromer: OK, how's this?

toolchain/check/check.cpp
```diff
@@ -664,31 +664,37 @@ static auto DiagnoseMissingDefinitions(Context& context,
...
+          emitter.Emit(decl_inst_id, MissingDefinitionInImpl);
         }
         break;
       }
+      case SemIR::InterfaceDecl::Kind: {
+        CARBON_FATAL()
+            << "TODO: Support interfaces in DiagnoseMissingDefinitions";
+      }
       // TODO: handle `interface` as well, once we can test it without
       // triggering https://github.com/carbon-language/carbon-lang/issues/4071
```
jonmeow: Should this comment be moved onto the InterfaceDecl case?

toolchain/check/handle_function.cpp
jonmeow: Yeah, I think the it's a small bool to add. I suspect it's not there right now just because merging isn't handled.

toolchain/check/testdata/class/no_definition_in_impl_file.carbon
jonmeow: Thanks, looks good.

jonmeow (approved):
Thanks, looks good!

If you want another look for is_definition in impl, let me know (and feel free not to make the change, if you don't think it makes sense in this PR). Otherwise, I think it's okay to merge once you've responded to final comments.

PR #4081, opened 2024-06-26
Create a `Generic` object to represent a generic.

zygoloid:
Build a `Generic` object for generic functions. This object tracks the generic parameters that are in scope for the generic entity. Eventually it will track other information about the generic too.

Add basic SemIR formatting support for generic functions.

### DIFF
--- a/toolchain/check/scope_stack.cpp
+++ b/toolchain/check/scope_stack.cpp
@@ -56,6 +56,16 @@ auto ScopeStack::Pop() -> void {
     CARBON_CHECK(return_scope_stack_.back().returned_var.is_valid());
     return_scope_stack_.back().returned_var = SemIR::InstId::Invalid;
   }
+
+  CARBON_CHECK(scope.next_compile_time_bind_index.index ==
+               static_cast<int32_t>(compile_time_binding_stack_.size()))
+      << "Wrong number of entries in compile-time binding stack, have "
+      << compile_time_binding_stack_.size() << ", expected "
+      << scope.next_compile_time_bind_index.index;
+  compile_time_binding_stack_.truncate(
+      scope_stack_.empty()
+          ? 0
+          : scope_stack_.back().next_compile_time_bind_index.index);
 }
 
 auto ScopeStack::PopTo(ScopeIndex index) -> void {
@@ -147,12 +157,24 @@ auto ScopeStack::SetReturnedVarOrGetExisting(SemIR::InstId inst_id)
 auto ScopeStack::Suspend() -> SuspendedScope {
   CARBON_CHECK(!scope_stack_.empty()) << "No scope to suspend";
   SuspendedScope result = {.entry = scope_stack_.pop_back_val(),
-                           .suspended_lookups = {}};
+                           .suspended_lookups = {},
+                           .compile_time_bindings = {}};
   if (result.entry.scope_id.is_valid()) {
     non_lexical_scope_stack_.pop_back();
   }
   for (auto name_id : result.entry.names) {
     result.suspended_lookups.push_back(lexical_lookup_.Suspend(name_id));
+  }
+  // Move any compile-time bindings into the suspended scope.
+  {
+    auto new_size =
+        scope_stack_.empty()
+            ? 0
+            : scope_stack_.back().next_compile_time_bind_index.index;
+    result.compile_time_bindings.assign(
+        compile_time_binding_stack_.begin() + new_size,
+        compile_time_binding_stack_.end());
+    compile_time_binding_stack_.truncate(new_size);
   }
   // This would be easy to support if we had a need, but currently we do not.
   CARBON_CHECK(!result.entry.has_returned_var)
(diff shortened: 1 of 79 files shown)


toolchain/sem_ir/file.h
```diff
@@ -208,6 +211,9 @@ class File : public Printable<File> {
   // Storage for impls.
   ImplStore impls_;
 
+  // Storage for generics.
+  ValueStore<GenericId> generics_;
```
jonmeow, line 215: Should this be printed as part of the yaml IR?

Maybe we should evaluate getting rid of the yaml given it's not really well maintained, since it's not well tested?

toolchain/sem_ir/generic.h
```diff
@@ -0,0 +1,31 @@
...
+
+#ifndef CARBON_TOOLCHAIN_SEM_IR_GENERIC_H_
+#define CARBON_TOOLCHAIN_SEM_IR_GENERIC_H_
+
+#include "toolchain/sem_ir/ids.h"
+
+namespace Carbon::SemIR {
+
+// Information for a generic entity, such as a generic class, a generic
+// interface, or generic function.
```
jonmeow, line 13: When you say generic, do you mean both checked generics and template generics, or just checked generics? I think we should be precise here.

toolchain/sem_ir/generic.h
```diff
@@ -0,0 +1,31 @@
...
+    out << "{decl: " << decl_id << ", bindings: " << bindings_id << "}";
+  }
+
+  // The following members always have values, and do not change throughout the
+  // lifetime of the generic.
+
+  // The first declaration of the generic entity.
+  InstId decl_id;
+  // A block containing the IDs of compile time bindings in this generic scope.
+  // The index in this block will match the `bind_index` of the instruction.
```
jonmeow: I'm a little uncertain here, we only have `bind_index` on BindNameInfo, is that what you mean here?

```suggestion
  // In this block, each instruction's index will match its BindNameInfo's
  // `bind_index`.
```

toolchain/check/scope_stack.h
```diff
@@ -210,6 +224,8 @@ struct ScopeStack::SuspendedScope {
   // to keep the size of a `SuspendedFunction` reasonable while avoiding heap
   // allocations most of the time.
   llvm::SmallVector<LexicalLookup::SuspendedResult, 8> suspended_lookups;
+  // The compile-time bindings that were introduced in this scope.
+  llvm::SmallVector<SemIR::InstId, 8> compile_time_bindings;
```
jonmeow: Why 8? There's a comment on suspended_lookups, but it doesn't clearly apply to this member.

If this is because we expect it to be 1:1 (or thereabout) with LexicalLookup::SuspendedResult, could they be combined? In part, I think there's a benefit to avoiding an additional vector if they're expected to be similar size; for things with a few parameters, even with no compile time bindings, it'll be smaller or equivalent just because of vector overhead.

toolchain/check/scope_stack.cpp
```diff
@@ -147,20 +157,44 @@ auto ScopeStack::SetReturnedVarOrGetExisting(SemIR::InstId inst_id)
...
+                           .suspended_lookups = {},
+                           .compile_time_bindings = {}};
   if (result.entry.scope_id.is_valid()) {
     non_lexical_scope_stack_.pop_back();
   }
   for (auto name_id : result.entry.names) {
     result.suspended_lookups.push_back(lexical_lookup_.Suspend(name_id));
   }
+  // Move any compile-time bindings into the suspended scope.
+  {
```
jonmeow: FWIW, this extra scope mostly reads weird to me -- you might consider blank lines to separate logic (instead). The separation of the comment is also an odd result.

jonmeow (commented):
I think these changes make sense, but I'd like to confirm whether "Generic" includes template generics.

toolchain/sem_ir/file.h
zygoloid: Added to YAML output.

toolchain/sem_ir/generic.h
zygoloid: I mean all generics -- the post-#2138 terminology. I've added to this comment to remind the reader of our terminology choice here.

toolchain/sem_ir/generic.h
zygoloid: Done.

toolchain/check/scope_stack.h
zygoloid: Hmm, we could almost avoid storing the compile time bindings entirely, but the list of bindings and the list of lookups can be out of sync due to error recovery. For
```
fn F(T:! type, T:! type)
```
we have two compile-time bindings but only one lookup result. I've combined these into a single vector as suggested.

toolchain/check/scope_stack.cpp
zygoloid: Done.

jonmeow (approved):
LGTM, thanks!

PR #4082, opened 2024-06-26
Treat constants with symbolic type as being symbolic.

zygoloid:
When constant evaluation produces a known non-symbolic value, treat the result as a symbolic constant anyway if the type of the value is symbolic.

We don't yet have many ways to produce a constant that has a known value but a symbolic type. The added test case is one such way: an array `[T; 0]` initialized from `()` is a symbolic constant only because its type is symbolic -- we know its value is always `()`. More ways to form such constants will be appearing soon as we start to support generics: for example, a method of a generic class has a symbolic type but a known constant value of `{}`.

When substituting into a symbolic constant, also substitute into its type.

### DIFF
--- a/toolchain/check/eval.cpp
+++ b/toolchain/check/eval.cpp
@@ -50,6 +50,11 @@ static auto GetPhase(SemIR::ConstantId constant_id) -> Phase {
   }
 }
 
+static auto GetTypePhase(Context& context, SemIR::TypeId type_id) -> Phase {
+  return context.types().GetConstantId(type_id).is_symbolic() ? Phase::Symbolic
+                                                              : Phase::Template;
+}
+
 // Returns the later of two phases.
 static auto LatestPhase(Phase a, Phase b) -> Phase {
   return static_cast<Phase>(
@@ -200,7 +205,7 @@ static auto RebuildAndValidateIfFieldsAreConstant(
   // Build a constant instruction by replacing each non-constant operand with
   // its constant value.
   auto typed_inst = inst.As<InstT>();
-  Phase phase = Phase::Template;
+  Phase phase = GetTypePhase(context, inst.type_id());
   if ((ReplaceFieldWithConstantValue(context, &typed_inst, each_field_id,
                                      &phase) &&
        ...)) {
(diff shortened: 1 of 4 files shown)


toolchain/check/eval.cpp
```diff
@@ -50,6 +50,11 @@ static auto GetPhase(SemIR::ConstantId constant_id) -> Phase {
   }
 }
 
+static auto GetTypePhase(Context& context, SemIR::TypeId type_id) -> Phase {
+  return context.types().GetConstantId(type_id).is_symbolic() ? Phase::Symbolic
+                                                              : Phase::Template;
```
jonmeow: Should this CHECK that it's constant? (i.e., that the expectation of either Symbolic or Template is correct)

(also, might consider a brief comment on the function)

toolchain/check/subst.cpp
```diff
@@ -63,7 +63,9 @@ static auto PushOperand(Context& context, Worklist& worklist,
       worklist.Push(static_cast<SemIR::InstId>(arg));
       break;
     case SemIR::IdKind::For<SemIR::TypeId>:
-      worklist.Push(context.types().GetInstId(static_cast<SemIR::TypeId>(arg)));
+      if (auto type_id = static_cast<SemIR::TypeId>(arg); type_id.is_valid()) {
```
jonmeow, line 66: Why do you use static_cast here instead of `SemIR::TypeId(arg)`? I grant the code was doing this previously, but it seems inconsistent (e.g., in ValueStore, we have `IdT id = IdT(values_.size());`, not `static_cast<IdT>(values_.size())`). I'm actually a little surprised this works with the explicit constructors.

I see you add `SemIR::TypeId(type_id)` on line 157, and `static_cast<SemIR::TypeId>(arg)` on line 106. I do think we should be consistent on this.

Similar elsewhere in this file -- a quick scan suggests this may be the only file using static_cast for Id types. If you prefer to do cleanup separately, that's fine, though I'm also fine if you were to just remove the static_casts within this PR -- it looks like there's 8 pre-existing (I don't count the `static_cast<int>`).

jonmeow (approved):
LGTM

toolchain/check/subst.cpp
zygoloid: I did this here just to be consistent with the existing code -- which I think was doing that as a probably-unnecessary attempt to avoid the clang-tidy check here. I'll send a separate PR to clean this up.

toolchain/check/eval.cpp
zygoloid: Good call -- it turns out that this was also getting called when `type_id` is invalid (specifically for `StructTypeField` instructions that don't have a `type_id`). I've cleaned this up and added a check for that case.

PR #4084, opened 2024-06-26
Prefer function-style cast instead of static_cast to convert integers to Ids.

zygoloid:
As requested in review of #4082.


jonmeow (approved):
Thanks!

PR #4085, opened 2024-06-26
Add `GenericInstance` type to represent instances of generics.

zygoloid:
Also add a corresponding value store and YAML output.

We don't create any generic instances in this change; this is just adding infrastructure for future changes.


toolchain/sem_ir/generic.h
```diff
@@ -29,6 +31,74 @@ struct Generic : public Printable<Generic> {
...
+  struct Key {
+    GenericId generic_id;
+    InstBlockId args_id;
+
+    friend auto operator==(const Key&, const Key&) -> bool = default;
+  };
+
+  // Context for hashing keys.
+  struct KeyContext {
+    llvm::ArrayRef<GenericInstance> instances;
```
jonmeow: Should this be private? It doesn't seem like it's part of the public API, so it strikes me as odd to put it first. hashtable_key_context.h's example provides the interface as a class, so would similar work here?

toolchain/sem_ir/generic.h
```diff
@@ -29,6 +31,74 @@ struct Generic : public Printable<Generic> {
   InstBlockId bindings_id;
 };
 
+// An instance of a generic entity, such as an instance of a generic function.
+// For each construct that depends on a compile-time parameter in the generic
+// entity, this contains the corresponding non-generic value. This includes
+// values for the compile-time parameters themselves.
+struct GenericInstance : Printable<GenericInstance> {
+  // Values corresponding to a region of a generic.
```
jonmeow: What's a region of a generic? I don't see the term used in the design.

zygoloid: Sorry, I'd incompletely removed this. Will be introduced in a future PR, gone for now :)

toolchain/sem_ir/generic.h
zygoloid: This is just a struct; given the limited scope I don't think it makes sense to add encapsulation here.

zygoloid: That said... looking at this again, I don't think this needs to be in the header file. Moved to the `.cpp` file. Maybe that provides the level of privateness you're looking for?

jonmeow (approved):

PR #4086, opened 2024-06-26
Build Generic objects for generic classes and interfaces.

zygoloid:
In `ClassType`s and `InterfaceType`s, track a `GenericInstanceId` for the instance rather than just the argument list.


toolchain/check/eval.cpp
```diff
@@ -15,6 +15,14 @@
 
 namespace Carbon::Check {
 
+static auto MakeGenericInstance(Context& context, SemIR::GenericId generic_id,
+                                SemIR::InstBlockId args_id)
+    -> SemIR::GenericInstanceId {
+  auto instance_id = context.generic_instances().GetOrAdd(generic_id, args_id);
+  // TODO: Perform substitution into the generic declaration if needed.
```
jonmeow, line 22: Is this function worth having a comment, particularly given the substitution note? That seems like it'll be what makes this non-trivial.

toolchain/check/eval.cpp
```diff
@@ -177,6 +185,27 @@ static auto GetConstantValue(Context& context, SemIR::TypeBlockId type_block_id,
...
+// The constant value of a generic instance is the generic instance with the
+// corresponding constant values for its arguments.
+static auto GetConstantValue(Context& context,
+                             SemIR::GenericInstanceId instance_id, Phase* phase)
+    -> SemIR::GenericInstanceId {
+  if (!instance_id.is_valid()) {
+    return SemIR::GenericInstanceId::Invalid;
+  }
+
+  auto& instance = context.generic_instances().Get(instance_id);
```
```suggestion
  const auto& instance = context.generic_instances().Get(instance_id);
```

Minor, just wary of the non-const reference.

toolchain/sem_ir/formatter.cpp
```diff
@@ -139,6 +139,11 @@ class Formatter {
     out_ << "\nclass ";
     FormatClassName(id);
 
+    if (class_info.generic_id.is_valid()) {
+      out_ << "\n    ";
+      FormatGeneric(class_info.generic_id);
+    }
```
jonmeow: Had you considered moving the `is_valid` + indentation into `FormatGeneric` (or a different shared call), to make it consistent between all calls? I see 3 places doing this in the change.

toolchain/check/import_ref.cpp
```diff
@@ -1180,6 +1206,8 @@ class ImportRefResolver {
         context_.AddPlaceholderInstInNoBlock(SemIR::LocIdAndInst(
             AddImportIRInst(import_interface.decl_id), interface_decl));
 
+    // TODO: Support for importing generics.
```
jonmeow, line 1225: Where you have these TODOs, would it be worth a `context.TODO` to produce a diagnostic? Or do you basically already have this implemented?

toolchain/check/import_ref.cpp
```diff
@@ -413,6 +413,29 @@ class ImportRefResolver {
     return context_.inst_blocks().AddCanonical(contents);
   }
 
+  auto GetLocalGenericInstanceArgs(SemIR::GenericInstanceId instance_id)
```
jonmeow, line 430: Might be good to have comments on these new functions, explaining the intent.

toolchain/check/import_ref.cpp
```diff
@@ -413,6 +413,29 @@ class ImportRefResolver {
...
+      -> llvm::SmallVector<SemIR::InstId> {
+    if (!instance_id.is_valid()) {
+      return {};
```
jonmeow, line 433: Minor, but given the vector return, do you want a single return variable for rvo?

jonmeow (approved):
Looks good, just some small comments. Let me know if you want another look before merging.

toolchain/check/eval.cpp
zygoloid: I'm relocating this and adding a comment in #4087.

toolchain/sem_ir/formatter.cpp
zygoloid: Yeah, I'd considered it. I'm on the fence -- it doesn't really seem like it should be `FormatGeneric`'s responsibility to know how the enclosing context is formatted, but it is otherwise being duplicated. Hm, I think I can see a way to handle this a bit more generically...

toolchain/check/import_ref.cpp
zygoloid: It's unnecessary in this case -- both return statements construct the `SmallVector` in place, so we get guaranteed RVO here and don't need to rely on NRVO.

toolchain/check/import_ref.cpp
zygoloid: Done.

toolchain/sem_ir/formatter.cpp
zygoloid: Done.

toolchain/check/import_ref.cpp
zygoloid: I don't have a full implementation yet, but this comes up in the prelude so adding a `context.TODO` would break every test. I'll be fixing this in the short term.

PR #4087, opened 2024-06-27
Use the correct type for `Self` in generic classes and generic interfaces

zygoloid:
In a `class C(T:! type)`, the type `Self` should be `C(T)`, not merely `C`. Similarly, in an `interface I(T:! type)`, the type of self should be `I(T)`, not merely `I`.


zygoloid:
Depends on #4086.

toolchain/check/generic.h
```diff
@@ -32,6 +32,21 @@ auto FinishGenericRedecl(Context& context, SemIR::InstId decl_id,
...
+// declaration, but not the definition, of the generic.
+//
+// `args_id` should be a canonical instruction block referring to constants.
+auto MakeGenericInstance(Context& context, SemIR::GenericId generic_id,
+                         SemIR::InstBlockId args_id)
+    -> SemIR::GenericInstanceId;
+
+// Builds the generic instance corresponding to the generic itself. For example,
+// for a generic `G(T:! type)`, this is `G(T)`.
+auto MakeUnsubstitutedGenericInstance(Context& context,
```
josh11b: "Unsubstituted" doesn't convey this to me, maybe?
```suggestion
auto MakeSelfGenericInstance(Context& context,
```
or
```suggestion
auto MakeCanonicalGenericInstance(Context& context,
```
or
```suggestion
auto MakeRepresentativeGenericInstance(Context& context,
```

toolchain/check/testdata/class/generic/self.carbon
```diff
@@ -0,0 +1,72 @@
...
+// TIP: To test this file alone, run:
+// TIP:   bazel test //toolchain/testing:file_test --test_arg=--file_tests=toolchain/check/testdata/class/generic/self.carbon
+// TIP: To dump output, run:
+// TIP:   bazel run //toolchain/testing:file_test -- --dump_output --file_tests=toolchain/check/testdata/class/generic/self.carbon
+
+class Class(T:! type) {
+  // `Self` is the same as `Class(T)` here.
+  fn F[addr self: Self*]() -> Class(T)* {
+    return self;
+  }
```
josh11b: Test it works in both directions?
```suggestion
  fn F[addr self: Self*]() -> Class(T)* {
    return self;
  }
  fn G[addr self: Class(T)*]() -> Self* {
    return self;
  }
```

toolchain/check/generic.h
zygoloid: Went with a variation of the first option and updated the generics-in-the-toolchain design doc to call this the "self instance" rather than the "unsubstituted instance".

toolchain/check/testdata/class/generic/self.carbon
zygoloid: Hm, I'm trying to test that it's the exact same type, not just that it's convertible. Maybe there's a better way to do that?

zygoloid: I've changed this a bit to try to better test the types are the same. I'm still not really happy with it, but hopefully this will do for now.

josh11b (approved):

PR #4088, opened 2024-06-27
Fix quirks in debug printing

jonmeow:
So, I noticed that ImportIRInst didn't print properly while trying to debug an issue, and that's where this started. Then I was sort of trying to figure out why we have "type_blocks" but "typeBlock", so trying to make that more consistent. "importIRInst0" felt more odd than "import_ir_inst0" which is why I'm suggesting down this particular route, but let me know if you'd prefer the reverse (but then do we also do "typeBlocks", etc, removing the consistency with the member name?)

Also starting to print more detail on import_irs, and added import_ir_insts (adjusting formatting for that too).

### DIFF
--- a/toolchain/sem_ir/ids.h
+++ b/toolchain/sem_ir/ids.h
@@ -183,7 +183,7 @@ struct BindNameId : public IdBase, public Printable<BindNameId> {
 
   using IdBase::IdBase;
   auto Print(llvm::raw_ostream& out) const -> void {
-    out << "bindName";
+    out << "bind_name";
     IdBase::Print(out);
   }
 };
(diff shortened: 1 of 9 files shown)


toolchain/sem_ir/ids.h
```diff
@@ -601,13 +601,18 @@ struct ElementIndex : public IndexBase, public Printable<ElementIndex> {
...
+struct ImportIRInstId : public IdBase, public Printable<ImportIRInstId> {
   using ValueType = ImportIRInst;
 
   // An explicitly invalid ID.
   static const ImportIRInstId Invalid;
 
   using IdBase::IdBase;
+
+  auto Print(llvm::raw_ostream& out) const -> void {
+    out << "import_ir_inst";
```
zygoloid, line 613: We seem to have no test coverage for this; can you add some?

zygoloid (approved):
Consistently using `snake_case` seems fine to me.

toolchain/sem_ir/ids.h
jonmeow: Done. Also started printing the import_ir_inst list, and adjusted formatting for that and ImportIR.

PR #4089, opened 2024-06-28
Change implicit import handling to be namespace-oriented.

jonmeow:
This refactors how the implicit import is handled in order to retain more name scope information. As a consequence, private access control works better between api files and implementation files. Note though that this will also be essential for name poisoning between the API and implementation, as discussed in #3763.

In implementing this, I ran into a couple issues with namespaces that I think point to flaws in their handling. I've fixed some and added a TODO for the biggest issue (in check.cpp line 281-288), which relates to the handling of namespaces of direct imports which are first evaluated indirectly.


toolchain/check/check.cpp
```diff
@@ -97,9 +97,35 @@ struct UnitInfo {
...
+      continue;
+    }
+    index = results.size();
+    results.push_back(
+        // TODO: For !is_local, should this use something valid that points at
+        // the API's import? That would probably require something other than a
+        // node_id, such as a LocId.
+        {.node_id = is_local ? import.names.node_id : Parse::InvalidNodeId(),
+         .sem_ir = &direct_ir,
+         .is_export = import.names.is_export});
```
zygoloid: Logically it seems like this should be `is_local && import.names.is_export`, but I suppose it makes no difference whether an import in an impl file is "exported". (No change requested but I'd be fine if you wanted to change it.)

toolchain/check/check.cpp
```diff
@@ -171,14 +198,98 @@ static auto ImportCurrentPackage(Context& context, UnitInfo& unit_info,
...
 }
 
+// Imports all other packages (excluding the current package).
+static auto ImportOtherPackages(Context& context, UnitInfo& unit_info,
+                                int total_ir_count,
+                                SemIR::TypeId namespace_type_id) -> void {
+  // For each package, the index into the API's package_imports. If this is not
+  // imported by the impl, the IdentifierId is the name of the package being
+  // imported. This is initially the size of the implementation's imports, even
+  // when there is no API, for simplicity in iteration.
```
zygoloid: If my understanding of the code below is correct, then:
```suggestion
  // For each package, the index into the API's package_imports. If this package is
  // imported by the API file, the IdentifierId is the name of the package being
  // imported; otherwise, it is invalid and the index is -1. This is initially the size of the implementation's imports, even
  // when there is no API, for simplicity in iteration.
```

toolchain/check/check.cpp
```diff
@@ -171,14 +198,98 @@ static auto ImportCurrentPackage(Context& context, UnitInfo& unit_info,
...
+      package_id = local_imports->package_id;
+      has_load_error |= local_imports->has_load_error;
+    }
+
+    // Identify the API package imports if present.
+    const UnitInfo::PackageImports* api_imports = nullptr;
+    if (api_imports_entry.second != -1) {
+      api_imports =
+          &unit_info.api_for_impl->package_imports[api_imports_entry.second];
+      package_id = api_imports_entry.first;
```
zygoloid: Is it worth `CHECK`ing that `!package_id.is_valid() || package_id == api_imports_entry.first` here?

toolchain/check/check.cpp
```diff
@@ -171,14 +198,98 @@ static auto ImportCurrentPackage(Context& context, UnitInfo& unit_info,
...
+      has_load_error |= api_imports->has_load_error;
+    }
+
+    // Do the actual import.
+    ImportLibrariesFromOtherPackage(
+        context, namespace_type_id, node_id, package_id,
```
zygoloid, line 288: I think we won't have a `node_id` here for an API-file-only import, which looks like it means we'll end up with a namespace without a location. Perhaps we could change `ImportLibrariesFromOtherPackage` to take a `LocId` instead of a `NodeId`, and create an `ImportIRInstId` here in the API-file-only case?

I think this would affect an example like:

```
// --- a.carbon
library "a";
fn A();

// --- package_a.carbon
package A;

// --- api.carbon
library "b";
import A;

// --- impl.carbon
impl library "b";
import library "a";
```

... where I think we won't have a location to use to diagnose the conflict between the API file's name A and the name A imported from "a". 

This PR is already a substantial change, so I'm fine with this not being handled as part of this PR.

toolchain/check/check.cpp
```diff
@@ -171,14 +198,98 @@ static auto ImportCurrentPackage(Context& context, UnitInfo& unit_info,
...
+        CollectTransitiveImports(local_imports, api_imports, total_ir_count),
+        has_load_error);
+  }
+}
+
 // Add imports to the root block.
+// TODO: Should this be importing all namespaces before anything else, including
+// for imports from other packages? Otherwise, name conflict resolution
+// involving something sudch as `fn F() -> Other.NS.A`, wherein `Other.NS`
```
```suggestion
// involving something such as `fn F() -> Other.NS.A`, wherein `Other.NS`
```

toolchain/check/import_ref.cpp
```diff
@@ -577,6 +577,7 @@ class ImportRefResolver {
   // unresolved constants to the work stack.
   auto GetLocalNameScopeId(SemIR::NameScopeId name_scope_id)
       -> SemIR::NameScopeId {
+    // Get the instruction that create the scope.
```
```suggestion
    // Get the instruction that created the scope.
```

toolchain/sem_ir/file.cpp
```diff
@@ -526,15 +526,16 @@ auto GetExprCategory(const File& file, InstId inst_id) -> ExprCategory {
...
-      case ImportRefUnloaded::Kind:
       case Namespace::Kind:
       case Return::Kind:
       case ReturnExpr::Kind:
       case StructTypeField::Kind:
         return ExprCategory::NotExpr;
 
-      case CARBON_KIND(ImportRefLoaded inst): {
-        auto import_ir_inst = ir->import_ir_insts().Get(inst.import_ir_inst_id);
+      case ImportRefUnloaded::Kind:
```
zygoloid, line 535: The handling of this looks correct, but I'm surprised we'd ask for the category of an `ImportRefUnloaded`. Does it make sense that that's happening? It seems a bit scary for code to be looking at the expression category, given that we won't have imported (eg) the type for this case.

zygoloid (approved):
Looks good.

toolchain/check/check.cpp
zygoloid: Also, "even when there is no API" might be clearer as "even when this file is the API file" or something?

toolchain/check/check.cpp
jonmeow: This is what the TODO in CollectDirectImports, line 115-117, is trying to cover. I think it's pretty complex though to change how import locations are tracked, e.g. touching diagnostic logic and probably adding an instruction (so that ImportIRInst has an Inst to point at), which is why I prefer a separate PR.

toolchain/check/check.cpp
jonmeow: Fixed

toolchain/check/import_ref.cpp
jonmeow: Fixed

toolchain/check/check.cpp
jonmeow: Done, added comment to ease understanding.

toolchain/check/check.cpp
jonmeow: Rewriting this comment, trying to incorporate suggestions.

toolchain/check/check.cpp

toolchain/sem_ir/file.cpp
jonmeow: This causes the error in (reduced example):

```
// --- c1.carbon

library "C1";

class C1 {}

// --- mix_current_package.carbon

library "mix_current_package";

import library "C1";

// --- mix_current_package.impl.carbon

impl library "mix_current_package";

import library "C2";

// CHECK:STDERR: mix_current_package.impl.carbon:[[@LINE+4]]:9: ERROR: Expression cannot be used as a value.
// CHECK:STDERR: var c1: C1 = {};
// CHECK:STDERR:         ^~
// CHECK:STDERR:
var c1: C1 = {};
var c2: C2 = {};
```

I spent a while digging through this, but I've forgotten the actual root cause. Unfortunately I can't investigate quickly (i.e., turn this into a CHECK) because the formatter *also* relies on being able to call GetExprCategory on ImportRefUnloaded.

This cost me the better part of a day to investigate originally, so if you're concerned I'll need to look further next week.

zygoloid: I'm not concerned enough to hold up this PR :)

I think we should make sure we understand what's happening, but let's not block on that.

zygoloid (approved):

PR #4091, opened 2024-06-28
Rename BindName to BindRuntimeName

geoffromer:
For consistency with BindSymbolicName and AnyBindName.


jonmeow:
I've commented on #toolchain: I think it would be good to discuss this more before changing it.

PR #4092, opened 2024-06-29
Track a list of dependent instructions created within a generic

zygoloid:
When checking a declaration or definition of a generic, track a list of created instructions that depend on the generic's parameters in some way, along with information on how they depend on the parameters. This will eventually be used to determine what information we need to compute when creating instances of the generic, but for now we're just building the list.

Information is tracked separately for the declaration region and the definition region of the generic, because in general these may be first provided in separate declarations, and they should be substituted into at different times.

### DIFF
--- a/toolchain/check/generic.cpp
+++ b/toolchain/check/generic.cpp
@@ -8,36 +8,60 @@
 
 namespace Carbon::Check {
 
-auto StartGenericDecl(Context& /*context*/) -> void {
-  // TODO: Start tracking the contents of this declaration.
+auto StartGenericDecl(Context& context) -> void {
+  context.generic_region_stack().Push();
 }
 
-auto StartGenericDefinition(Context& /*context*/,
+auto StartGenericDefinition(Context& context,
                             SemIR::GenericId /*generic_id*/) -> void {
-  // TODO: Start tracking the contents of this definition.
+  // Push a generic region even if we don't have a generic_id. We might still
+  // have locally-introduced generic parameters to track:
+  //
+  // fn F() {
+  //   let T:! type = i32;
+  //   var x: T;
+  // }
+  context.generic_region_stack().Push();
 }
 
 auto FinishGenericDecl(Context& context, SemIR::InstId decl_id)
     -> SemIR::GenericId {
   if (context.scope_stack().compile_time_binding_stack().empty()) {
+    CARBON_CHECK(context.generic_region_stack().PeekDependentInsts().empty())
+        << "Have dependent instructions but no compile time bindings are in "
+           "scope.";
+    context.generic_region_stack().Pop();
     return SemIR::GenericId::Invalid;
   }
 
   auto bindings_id = context.inst_blocks().Add(
       context.scope_stack().compile_time_binding_stack());
+  // TODO: Track the list of dependent instructions in this region.
+  context.generic_region_stack().Pop();
   return context.generics().Add(
       SemIR::Generic{.decl_id = decl_id, .bindings_id = bindings_id});
 }
 
-auto FinishGenericRedecl(Context& /*context*/, SemIR::InstId /*decl_id*/,
+auto FinishGenericRedecl(Context& context, SemIR::InstId /*decl_id*/,
                          SemIR::GenericId /*generic_id*/) -> void {
   // TODO: Compare contents of this declaration with the existing one on the
   // generic.
+  context.generic_region_stack().Pop();
 }
 
-auto FinishGenericDefinition(Context& /*context*/,
-                             SemIR::GenericId /*generic_id*/) -> void {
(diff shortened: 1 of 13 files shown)


toolchain/sem_ir/ids.h
```diff
@@ -557,6 +557,16 @@ struct TypeId : public IdBase, public Printable<TypeId> {
...
+  // The first index reserved for substituted types. These are types used in a
+  // generic that will be substituted for other types in each use of the
+  // generic.
+  static const int32_t FirstSubstitutedTypeIndex;
+
+  // Returns whether this is a substituted type.
+  auto is_substituted() const -> bool {
+    return index <= FirstSubstitutedTypeIndex;
+  }
+
```
jonmeow: This seems to be unused in this PR, am I missing something?

toolchain/check/generic_region_stack.h
```diff
@@ -0,0 +1,74 @@
...
+  // in some way depend on a generic parameter.
+  auto AddDependentInst(DependentInst inst) -> void;
+
+  // Returns the list of dependent instructions in the current generic region.
+  auto PeekDependentInsts() -> llvm::ArrayRef<DependentInst>;
+
+ private:
+  struct RegionInfo {
+    int32_t first_dependent_inst;
+  };
```
jonmeow: This might be worth a comment.

toolchain/check/generic_region_stack.h
```diff
@@ -0,0 +1,74 @@
...
+
+  // The current set of enclosing generic regions.
+  llvm::SmallVector<RegionInfo> regions_;
+
+  // List of symbolic constants used in any of the enclosing generic regions.
+  llvm::SmallVector<DependentInst> dependent_insts_;
```
jonmeow: If I understand correctly, this tracks the insts as a single vector, using RegionInfo to slice up sections. It looks like there's no use of spanning regions, is that correct? Why this approach versus a vector of regions with vector<DependentInst> as a per-region member? I can see there may be reasons, maybe you'll want the overlapping regions for member class types or similar, but a comment might help clarify.

I think the relationship between regions_ and dependent_insts_ would be helpful to talk about a little in particular, as the structure manipulation requires a bit of reading to understand.

toolchain/check/generic.cpp
```diff
@@ -8,36 +8,60 @@
...
-auto StartGenericDecl(Context& /*context*/) -> void {
-  // TODO: Start tracking the contents of this declaration.
+auto StartGenericDecl(Context& context) -> void {
+  context.generic_region_stack().Push();
 }
 
-auto StartGenericDefinition(Context& /*context*/,
-                            SemIR::GenericId /*generic_id*/) -> void {
-  // TODO: Start tracking the contents of this definition.
+auto StartGenericDefinition(Context& context, SemIR::GenericId /*generic_id*/)
```
jonmeow: Is there still a TODO here to use generic_id?

jonmeow (approved):
Looks good!

toolchain/sem_ir/ids.h
jonmeow: (note, partly the int32_t seems unusual, but it's hard to review without a call)
("first index" also I'd typically assume later values are >=, but maybe this is actually <=?)

toolchain/check/generic.cpp
zygoloid: Hm, actually, no -- it looks like I ended up not using this at all. Removed.

toolchain/check/generic_region_stack.h
zygoloid: Added.

toolchain/check/generic_region_stack.h
zygoloid: Added a comment. The motivation for this approach is denser storage and avoiding reallocations.

[FWIW, originally I had a second reason: for cases like

```carbon
class A(T:! type) {
  fn F(a: T) -> T;
}
```

... I was not introducing a separate generic region for the declaration of `F`, and instead was using the definition region of `A` to track the references to `T` inside the declaration of `A(T).F`. But when you start processing `F`, you don't know whether it has its own generic region or not, so the idea was that I'd track the dependent instructions in a flat list, and if `F` turns out to not be a generic function, we can just pop it from the list of regions and keep its instructions in the enclosing region.

But that turns out to not really work so well. It adds complexity in various places, and we end up wanting a separate generic region for the definition of `F` anyway, so it seems better to say that in this case, `F` is a generic function that just happens to have no more generic parameters than `A` does, so that's the direction I'm now pursuing.]

toolchain/sem_ir/ids.h
zygoloid: Thanks, this was an error splitting the PR. Changes to this file should not have been included.

PR #4094, opened 2024-06-30
A collection of hashing improvements from using hashtables.

chandlerc:
LLVM's `APInt` and `APFloat` need specialized handling to be used effectively in hashtables. We can't inject overrides into LLVM so we need to handle them in our hashing routine.

There were also problematic limits on hashing pairs and tuples. First, the unique-object-representation hashing of pairs was more restricted than tuples which was a problematic asymmetry and isn't needed. But the larger issue is that we didn't support recursively hashing when necessary. That requires a careful predicate to avoid infinite recursion but lets us handle important use cases for hashtables with a tuple as a key.

Also added support for hashing arrays that recurse in addition to arrays where we can hash the raw storage, and added overloads to redirect to common array handling from various array-like types.

Last but not least, re-worked the constraint model for hashing as raw data to not override custom hashing functions.

### DIFF
--- a/common/hashing.h
+++ b/common/hashing.h
@@ -13,6 +13,8 @@
 
 #include "common/check.h"
 #include "common/ostream.h"
+#include "llvm/ADT/APFloat.h"
+#include "llvm/ADT/APInt.h"
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/StringRef.h"
@@ -443,6 +445,11 @@ class Hasher {
 // public API. They should not be used directly by client code.
 namespace InternalHashDispatch {
 
+// Forward declare the dispatch function so we can re-do dispatch with a
+// different type if needed.
+template <typename T>
+inline auto DispatchImpl(const T& value, uint64_t seed) -> HashCode;
+
 inline auto CarbonHashValue(llvm::ArrayRef<std::byte> bytes, uint64_t seed)
     -> HashCode {
   Hasher hasher(seed);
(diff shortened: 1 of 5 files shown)


common/hashing.h
```diff
@@ -474,6 +481,63 @@ inline auto CarbonHashValue(const llvm::SmallString<Length>& value,
...
+}
+template <typename T, size_t N>
+inline auto CarbonHashValue(const T (&arg)[N], uint64_t seed) -> HashCode {
+  return DispatchImpl(llvm::ArrayRef(arg), seed);
+}
+
+inline auto CarbonHashValue(llvm::APInt value, uint64_t seed) -> HashCode {
+  Hasher hasher(seed);
+  if (LLVM_LIKELY(value.isSingleWord())) {
+    hasher.Hash(value.getBitWidth(), value.getZExtValue());
```
zygoloid, line 525: I suspect the answer is no, but is it worth treating the `getBitWidth() <= 32` case as a special case, for which we only need to hash 8 bytes rather than 12?

common/hashing.h
```diff
@@ -474,6 +481,63 @@ inline auto CarbonHashValue(const llvm::SmallString<Length>& value,
...
+  return static_cast<HashCode>(hasher);
+}
+
+// Implementation detail predicate to test for types hashable with
+// `CarbonHashValue`. Should find all of the above direct overloads as well as
+// ADL-provided overloads.
+template <typename T>
+concept HasCustomCarbonHashValue = requires(T t) {
+  { CarbonHashValue(t, static_cast<uint64_t>(0)) } -> std::same_as<HashCode>;
+};
```
zygoloid: I'm nervous about depending on this finding the overloads declared above this line but not the ones declared below -- that seems a bit subtle, as reordering the code would result in it compiling but meaning something different. (It also wouldn't work on MSVC...)

Would it be feasible to give a different name to the non-custom overloads below, so we go through two layers of customization, first for raw hashing then for custom hashing?

common/hashing.h
```diff
@@ -482,67 +546,110 @@ static_assert(std::has_unique_object_representations_v<int32_t>);
...
+
+template <typename T, typename U>
+  requires HasCustomCarbonHashValue<T> && HasCustomCarbonHashValue<U>
+inline auto CarbonHashValue(const std::pair<T, U>& value, uint64_t seed)
+    -> HashCode {
+  Hasher hasher(seed);
+  // No need to propagate the seed below as we've already incorporated it.
+  hasher.Hash(HashValue(value.first), HashValue(value.second));
+  return static_cast<HashCode>(hasher);
 }
```
zygoloid: These two overloads seem a bit limiting to me: we'll accept (for example) a pair where both elements are raw-hashable, or a pair where both elements are custom-hashable, but not a mixture.

I wonder if it'd be better for these to accept types that are hashable (in any way), excluding the cases handled by the above raw-only overloads? (Maybe `HasCustomCarbonHashValue<T>` should also be true in that case, given that it's invoking a custom function for part of the work.)

common/hashing.h
chandlerc: So far, it hasn't shown up in a profile so hard to say. And my guess is it'll be a wash given the need to stitch the two 32-bit values together, and add considerable complexity.

common/hashing.h
chandlerc: This was ... tricky. We discussed it a bit in person as well.

What I've done is sunk the raw hashing logic into `if constexpr` logic inside functions because that seems much more readable than trying to do overload resolution.

However, to make all this work robustly, we now have that logic in three places because we have three subtly different optimizations based on that logic:

1) The main dispatch case.
2) An optimized case for tuples (and pairs) to dispatch for each element.
3) An optimized case for `ArrayRef` to try and detect when we can hash the underlying storage rather than recursively hashing each element.

I couldn't see a good way to abstract out something for these three cases though as they all do somewhat subtle things with the dispatch logic.

Now that `Hasher` is doing some of this, I've renamed the method that *doesn't* do it to `HashRaw`, so that's roughly your other layer.

However, we still need a predicate on whether there is an overload. And that predicate is still technicaly order sensitive, as is the `DispatchImpl` code. I don't really see much way around that. But they are now less order sensitive than this -- they simply can't be hoisted further *up* as they must see *all* the overloads, and that should also mean they behave consistently with MSVC.

common/hashing.h
chandlerc: Agreed. I sunk all of the dispatch logic into `Hasher` itself, and taught it how to handle each entry independently now. Should even do the fast thing and avoid unnecessary recursive hashing where possible.

chandlerc (commented):
Thanks for the review. A new version that tries to do this in a bit more principled way.

common/hashing.h
```diff
@@ -233,6 +235,34 @@ class Hasher {
...
   explicit operator HashCode() const { return HashCode(buffer); }
 
+  // Incorporates a variable number of objects into the `hasher`s state.
+  //
+  // The `values` here can be anything hashable, and this routine handles
+  // recursively hashing a single value as appropriate and then in turn
+  // incorporating that. However, it is optimized for relatively small numbers
+  // of values and/or small elements. A large tree structure will be better
+  // handled by a dedicated Merkle-tree decomposition rather than the ad-hoc one
+  // provided here. This routines decomposition is mostly useful for combining N
```
zygoloid:
```suggestion
  // provided here. This routine's decomposition is mostly useful for combining N
```

common/hashing.h
```diff
@@ -474,82 +495,141 @@ inline auto CarbonHashValue(const llvm::SmallString<Length>& value,
...
-
-// C++ uses `std::nullptr_t` but unfortunately doesn't make it have a unique
-// object representation. To address that, we need a function that converts
-// `nullptr` back into a `void*` that will have a unique object representation.
-// And this needs to be done by-value as we need to build a temporary object to
-// return, which requires a separate overload rather than just using a type
-// function that could be used in parallel in the predicate below. Instead, we
-// build the predicate independently of the mapping overload, but together they
-// should produce the correct result.
+// Support types that are array like by building an `llvm::ArrayRef` out of
```
zygoloid:
```suggestion
// Support types that are array-like by building an `llvm::ArrayRef` out of
```

common/hashing.h
```diff
@@ -474,82 +495,141 @@ inline auto CarbonHashValue(const llvm::SmallString<Length>& value,
...
+// them. We can't do this by accepting any type convertible to an `ArrayRef`
+// because that type supports building a synthetic array out of any single
+// element. These need to go back through the dispatch implementation though to
+// pick up the full space of hash overloads for `T`.
```
zygoloid:
```suggestion
// element.
```
If we want to say this, I think we ought to do so on the `CarbonHashValue` overload for `ArrayRef`, but I think we can just drop it given that we don't call `DispatchImpl` either here or there.

zygoloid (approved):

common/hashing.h
```diff
@@ -233,6 +235,34 @@ class Hasher {
...
 
+  // Incorporates a variable number of objects into the `hasher`s state.
+  //
+  // The `values` here can be anything hashable, and this routine handles
+  // recursively hashing a single value as appropriate and then in turn
+  // incorporating that. However, it is optimized for relatively small numbers
+  // of values and/or small elements. A large tree structure will be better
+  // handled by a dedicated Merkle-tree decomposition rather than the ad-hoc one
+  // provided here. This routine's decomposition is mostly useful for combining N
+  // small bits of data with one recursively hashed entity.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  // provided here. This routine's decomposition is mostly useful for combining
  // N small bits of data with one recursively hashed entity.
```

CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  // provided here. This routine's decomposition is mostly useful for combining
  // N small bits of data with one recursively hashed entity.
```

PR #4095, opened 2024-06-30
Key context improvements

chandlerc:
This injects a customization point for hashtable-specific equality testing that the key context uses by default. While this is rarely needed, there are LLVM types where it is necessary and it seems a good general tool to have to avoid unnecessary complexity from custom key contexts when a simple customization of equality is all that is required.

This also adds a CRTP mixin for implementing a common pattern of key contexts where the context provides translation of some key types into another type, potentially using state. Rather than having to implement the entire key context API, code can derive from this template and simply provide a set of overloads for the types it wants to translate. Any key types used which can be passed to one of those overloads will get translated before following the same logic as the default key context. While this updates the only usage so far of this pattern, a subsequent PR will add several more users making the pattern worth abstracting here.

### DIFF
--- a/common/hashtable_key_context.h
+++ b/common/hashtable_key_context.h
@@ -5,9 +5,37 @@
 #ifndef CARBON_COMMON_HASHTABLE_KEY_CONTEXT_H_
 #define CARBON_COMMON_HASHTABLE_KEY_CONTEXT_H_
 
+#include <concepts>
+
 #include "common/hashing.h"
+#include "llvm/ADT/APFloat.h"
+#include "llvm/ADT/APInt.h"
 
 namespace Carbon {
+
+// The equality comparison used by the the hashtable key contexts in this file,
+// and suitable for using in other hashtable key contexts.
+//
+// This provides a hashtable-specific extension point to implement equality
+// comparison within a hashtable key context. By default, it will use
+// `operator==` on the LHS and RHS operands. However, types can provide a
+// dedicated customization point by implementing a free function that can be
+// found by ADL for your type called `CarbonHashtableEq` with the following
+// signature:
+//
+// ```cpp
+// auto CarbonHashtableEq(const YourType& lhs, const YourType& rhs) -> bool;
+// ```
+//
+// Any such overload will be able to override the default we provide for types
+// that can compare with `==`.
+//
+// This library also provides any customization points for LLVM or standard
+// library types either lacking `operator==` or where that operator is not
+// suitable for hashtables. For example, `llvm::APInt` and `llvm::APFloat` have
+// custom equality comparisons provided through this extension point.
+template <typename LeftT, typename RightT>
+auto HashtableEq(const LeftT& lhs, const RightT& rhs) -> bool;
 
 // Customizable context for keys in hashtables.
 //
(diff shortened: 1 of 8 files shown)


common/hashtable_key_context.h
```diff
@@ -69,16 +97,138 @@ namespace Carbon {
...
+  // namespace above and ADL-found functions within an associated namespace for
+  // either `LeftT` or `RightT`.
+  return CarbonHashtableEq(lhs, rhs);
+}
+
+}  // namespace InternalHashtableEqDispatch
+
+template <typename LeftT, typename RightT>
+inline auto HashtableEq(const LeftT& lhs, const RightT& rhs) -> bool {
+  return InternalHashtableEqDispatch::DispatchImpl(lhs, rhs);
```
zygoloid, line 185: You could remove a layer of indirection here, assuming this is the only call to `DispatchImpl`:
```suggestion
  using InternalHashtableEqDispatch::CarbonHashtableEq;
  CarbonHashtableEq(lhs, rhs);
```

common/hashtable_key_context.h
```diff
@@ -69,16 +97,138 @@ namespace Carbon {
...
+template <typename DerivedT>
+template <typename AnyKeyT>
+auto TranslatingKeyContext<DerivedT>::HashKey(const AnyKeyT& key,
+                                              uint64_t seed) const -> HashCode {
+  const DerivedT& self = *static_cast<const DerivedT*>(this);
+  if constexpr (requires { self.TranslateKey(key); }) {
+    return HashValue(self.TranslateKey(key), seed);
+  } else {
     return HashValue(key, seed);
   }
```
zygoloid, line 209: I'm a bit worried that accidentally passing in an unexpected / bogus (but hashable) type as the key type will silently "work" here. I suppose it's much less likely for the `KeyEq` code below to compile, though, so perhaps I shouldn't worry so much.

I'm wondering if we could do something to remove the risk: for example, perhaps `TranslatingKeyContext` could be given a list of key types that don't need translating, and could reject keys that aren't in the list but can't be translated. (It looks like we don't currently have an example with more than one untranslated key type, so maybe we could use `DerivedT::Key` as the untranslated type?)

zygoloid (commented):
Looks like a nice API improvement.

common/hashtable_key_context.h
chandlerc: I wanted to keep the pattern as similar to what we use for `CarbonHashValue` as I could, so I replicated the dispatch routine... And we did end up using the dispatch there.

Still prefer skipping the indirection for now?

common/hashtable_key_context.h
chandlerc: The `DefaultKeyContext` lets you just hash any type, and so it seemed strange to restrict things in this helper.

I'm also worried about listing the types being onerous because we won't get implicit conversions or other things here. If you *don't* want to translate a `ssize_t` key type, it'll be awkward to call things with integer literals because they won't promote. Even things like `nullptr` make it tricky to list all expected types.

That's part of why I used the "if the call is valid, do the call" model above.

common/hashtable_key_context.h
zygoloid: I don't feel strongly about it. (One the one hand, the two-step approach is a bit more idiomatic because it's how people seem to be encouraged to call `std::swap`. On the other hand, your current approach can be made SFINAE-friendly whereas the `std::swap` approach can't.)

If this is more consistent with what you're doing elsewhere, let's leave it alone.

common/hashtable_key_context.h
zygoloid: OK. I'm OK with moving ahead with this. Let's just keep an eye out -- if we see this kind of wrong-key-type problem coming up and being surprising, we can reconsider adding some kind of guard rail. I *think* we'll pretty consistently get a compile-time error in the key comparison rather than something mysterious where lookups fail because we use an inconsistent hash value, though.

zygoloid (approved):

PR #4097, opened 2024-07-01
Port the toolchain to use the new Carbon hashtable

chandlerc:
This works to leverage the capabilities of the hashtable as much as
possible, for example using the key context in the value stores.
However, there may still be opportunities to refactor more deeply and
use the functionality even better. Hopefully this is at least
a reasonable start and gets us a clean baseline.

On an Arm M1, this is a 15% improvement on my large lexing stress test,
but ends up a wash on my x86-64 server. This is a smaller benefit than
I expected, and it's because we're using a set-of-IDs and looking up
values with a key context for things like identifiers. This pattern has
a surprising tradeoff. The new hashtable uses significantly less memory,
a 10% peak RSS reduction just from the hashtable change. But indirecting
through the vector of values makes growing the hashtable dramatically
less cache-friendly: it causes growth to randomly access every key when
rehashing. On x86, everything gained by the faster hashtable is lost in
even slower growth. And even on Arm, this eats into the benefits.

But I have a plan to tweak how identifiers specifically work to avoid
most of the growth, and so I suspect this is the right tradeoff on the
whole. It gives us significant working set size reduction and we can
likely avoid the regressed operation (growth with rehash) in most cases
by clever reserving and if necessary by adding a hash caching layer to
the table infrastructure.

### DIFF
--- a/toolchain/base/value_store.h
+++ b/toolchain/base/value_store.h
@@ -205,7 +205,7 @@ class CanonicalValueStore {
   using ValueType = typename IdT::ValueType;
 
   // Stores a canonical copy of the value and returns an ID to reference it.
-  auto Add(const ValueType& value) -> IdT;
+  auto Add(ValueType value) -> IdT;
 
   // Returns the value for an ID.
   auto Get(IdT id) const -> const ValueType& { return values_.Get(id); }
@@ -254,8 +254,8 @@ class CanonicalValueStore<IdT>::KeyContext
 };
 
 template <typename IdT>
-auto CanonicalValueStore<IdT>::Add(const ValueType& value) -> IdT {
-  auto make_key = [&] { return IdT(values_.Add(value)); };
+auto CanonicalValueStore<IdT>::Add(ValueType value) -> IdT {
+  auto make_key = [&] { return IdT(values_.Add(std::move(value))); };
   return set_.Insert(value, make_key, KeyContext(values_.array_ref())).key();
 }
 
--- a/toolchain/base/value_store.h
+++ b/toolchain/base/value_store.h
@@ -8,11 +8,10 @@
 #include <type_traits>
 
 #include "common/check.h"
-#include "common/hashing.h"
 #include "common/ostream.h"
+#include "common/set.h"
 #include "llvm/ADT/APFloat.h"
 #include "llvm/ADT/APInt.h"
-#include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/Sequence.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/StringExtras.h"
@@ -76,26 +75,6 @@ struct FloatId : public IdBase, public Printable<FloatId> {
   }
 };
 constexpr FloatId FloatId::Invalid(FloatId::InvalidIndex);
-
-// DenseMapInfo for llvm::APFloat, for use in the canonical float value store.
-// LLVM has an implementation of this but it strangely lives in the non-public
-// header lib/IR/LLVMContextImpl.h, so we use our own.
-// TODO: Remove this once our new hash table is available.
-struct APFloatDenseMapInfo {
-  static auto getEmptyKey() -> llvm::APFloat {
-    return llvm::APFloat(llvm::APFloat::Bogus(), 1);
-  }
-  static auto getTombstoneKey() -> llvm::APFloat {
-    return llvm::APFloat(llvm::APFloat::Bogus(), 2);
-  }
-  static auto getHashValue(const llvm::APFloat& val) -> unsigned {
-    return hash_value(val);
-  }
-  static auto isEqual(const llvm::APFloat& lhs, const llvm::APFloat& rhs)
-      -> bool {
-    return lhs.bitwiseIsEqual(rhs);
-  }
-};
 
 // Corresponds to a Real value.
 struct RealId : public IdBase, public Printable<RealId> {
@@ -220,20 +199,13 @@ class ValueStore
 // to later retrieve the value.
 //
 // IdT::ValueType must represent the type being indexed.
-template <typename IdT,
-          typename DenseMapInfoT = llvm::DenseMapInfo<typename IdT::ValueType>>
+template <typename IdT>
 class CanonicalValueStore {
  public:
   using ValueType = typename IdT::ValueType;
 
   // Stores a canonical copy of the value and returns an ID to reference it.
-  auto Add(ValueType value) -> IdT {
-    auto [it, added] = map_.insert({value, IdT::Invalid});
-    if (added) {
-      it->second = values_.Add(value);
-    }
-    return it->second;
-  }
+  auto Add(const ValueType& value) -> IdT;
 
   // Returns the value for an ID.
   auto Get(IdT id) const -> const ValueType& { return values_.Get(id); }
@@ -241,7 +213,7 @@ class CanonicalValueStore {
   // Reserves space.
   auto Reserve(size_t size) -> void {
     values_.Reserve(size);
-    map_.reserve(size);
+    // map_.reserve(size);
   }
 
   // These are to support printable structures, and are not guaranteed.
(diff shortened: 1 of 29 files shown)


toolchain/sem_ir/inst.h
```diff
@@ -139,8 +139,7 @@ class Inst : public Printable<Inst> {
     } else {
       kind_ = TypedInst::Kind.AsInt();
     }
-    CARBON_CHECK(kind_ >= 0)
-        << "Negative kind values are reserved for DenseMapInfo.";
+    CARBON_CHECK(kind_ >= 0) << "Unexpected negative kind value.";
```
jonmeow: If the negative values are removed, can this just be removed? The DenseMapInfo check was added by #3979; I'm not sure how else a negative value would get here, and we don't generally validate that enum correctness.

```suggestion
```

toolchain/sem_ir/bind_name.h
```diff
@@ -33,44 +39,38 @@ inline auto CarbonHashValue(const BindNameInfo& value, uint64_t seed)
...
+    return store_->Get(id);
   }
 
  private:
-  llvm::DenseMap<BindNameInfo, BindNameId, BindNameInfoDenseMapInfo>
-      canonical_ids_;
+  const BindNameStore* store_;
 };
 
+inline auto BindNameStore::MakeCanonical(BindNameId id) -> BindNameId {
```
jonmeow, line 70: What effect does `inline` have here?

toolchain/check/scope_stack.cpp
```diff
@@ -162,20 +163,21 @@ auto ScopeStack::Suspend() -> SuspendedScope {
...
+  });
+
+  // Now we can reserve the remaining space needed.
   auto remaining_compile_time_bindings =
       scope_stack_.empty()
           ? 0
           : scope_stack_.back().next_compile_time_bind_index.index;
-
-  result.suspended_items.reserve(result.entry.names.size() +
+  result.suspended_items.reserve(result.suspended_items.size() +
```
jonmeow: Is there no way to reserve based on the size of the names anymore? I wonder if line 170 will now typically be most of the allocations (not sure that these other two will come up as often)

toolchain/check/decl_name_stack.cpp
```diff
@@ -147,12 +147,18 @@ auto DeclNameStack::AddName(NameContext name_context, SemIR::InstId target_id,
...
+                .Insert(name_context.unresolved_name_id,
+                        [&] {
+                          int index = name_scope.names.size();
+                          name_scope.names.push_back(
+                              {.name_id = name_context.unresolved_name_id,
+                               .inst_id = target_id,
+                               .access_kind = access_kind});
+                          return index;
+                        })
+                .is_inserted();
```
jonmeow: The wrapping and indentation of this is making it hard for me to read. Maybe move the lambda out-of-line, or break out `.is_inserted()`? I think if you break out `.is_inserted()` in particular (`auto result = ...; CARBON_CHECK(result.is_inserted())`, the whole wrapping will change to put `name_scope.name_map.Insert(` on a single line, etc.

toolchain/check/convert.cpp
```diff
@@ -449,8 +451,8 @@ static auto ConvertStructToStructOrClass(Context& context,
     // Find the matching source field.
     auto src_field_index = i;
     if (src_type.fields_id != dest_type.fields_id) {
-      auto src_field_it = src_field_indexes.find(dest_field.name_id);
-      if (src_field_it == src_field_indexes.end()) {
+      auto src_field_lookup = src_field_indexes.Lookup(dest_field.name_id);
+      if (!src_field_lookup) {
```
jonmeow: Optional, but looking at this, I was wondering if it'd be easier to read as:

```suggestion
      if (auto lookup = src_field_indexes.Lookup(dest_field.name_id)) {
        src_field_index = lookup.value();
      } else {
```

toolchain/check/convert.cpp
```diff
@@ -469,7 +471,7 @@ static auto ConvertStructToStructOrClass(Context& context,
         }
         return SemIR::InstId::BuiltinError;
       }
-      src_field_index = src_field_it->second;
+      src_field_index = src_field_lookup.value();
```
```suggestion
```

(see above)

toolchain/check/check.cpp
```diff
@@ -1039,26 +1039,28 @@ static auto TrackImport(
...
-    unit_info.package_imports.push_back(
-        UnitInfo::PackageImports(import.package_id, import.node_id));
-  }
+  // auto [package_imports_map_it, is_inserted] =
+  auto insert_result =
+      unit_info.package_imports_map.Insert(import.package_id, [&]() -> int32_t {
+        int32_t index = unit_info.package_imports.size();
+        unit_info.package_imports.push_back(
+            UnitInfo::PackageImports(import.package_id, import.node_id));
+        return index;
```
jonmeow: Minor, but we end up with this pattern (.size(), push_back, return index) a fair amount. Maybe a spot where we could add a helper function?

(not sure it makes sense in this PR)

toolchain/base/value_store.h
```diff
@@ -309,12 +288,51 @@ class CanonicalValueStore<StringId>
...
-  // Set inline size to 0 because these will typically be too large for the
+  class IdSetContext;
+
+  // Set inline sizes to 0 because these will typically be too large for the
   // stack, while this does make File smaller.
+  Set<StringId, /*SmallSize=*/0, IdSetContext> set_;
   llvm::SmallVector<llvm::StringRef, 0> values_;
 };
 
+class CanonicalValueStore<StringId>::IdSetContext
```
jonmeow: Why IdSetContext here but KeyContext elsewhere? Note, in part, it's not intuitive to me what the "IdSet" getting a context is, because we just call the internal set "set_", not "id_set_".

Also, you have two IdSetContexts. The unusual yet shared name makes me expect them to be identical. Could they be a shared type? The structures were separate because of the different `Add` logic, but maybe the `TranslatingKeyContext` implementation (using `ArrayRef`) is similar enough to share.

You might consider more comments here either way, this is a sizable chunk of code to not have any comment.

toolchain/base/value_store.h
```diff
@@ -309,12 +288,51 @@ class CanonicalValueStore<StringId>
...
+      .Insert(
+          value,
+          [&] {
+            auto id = static_cast<StringId>(values_.size());
+            CARBON_CHECK(id.index >= 0) << "Too many unique strings";
+            values_.push_back(value);
+            return id;
+          },
+          IdSetContext(values_))
+      .key();
```
jonmeow: Consider moving the `.key()` call to its own statement, so that clang-format wraps differently.

Note, breaking out the lambda may also benefit. 12 lines is a pretty long statement/function call. Of course, I guess I also comment elsewhere about a helper function maybe benefiting this size/push_back idiom, which would differently address the readability.

toolchain/base/value_store.h
```diff
@@ -220,28 +199,21 @@ class ValueStore
...
+  auto Add(const ValueType& value) -> IdT;
 
   // Returns the value for an ID.
   auto Get(IdT id) const -> const ValueType& { return values_.Get(id); }
 
   // Reserves space.
   auto Reserve(size_t size) -> void {
     values_.Reserve(size);
-    map_.reserve(size);
+    // map_.reserve(size);
```
jonmeow: Why is this just commented out? Should there be a TODO? Did you mean to remove this?

jonmeow (approved):
Looks good to me, just some comments about use.

Reading this code, I do wish the KeyContext object were part of map/set construction instead of per-call. Maybe there's an opportunity for a wrapper type there?

toolchain/base/value_store.h
chandlerc: Doh, meant to get back to this. Updated, and thanks for spotting.

toolchain/base/value_store.h
chandlerc: The two `IdSetContext`s is because of the specialization of `CanonicalValueStore` for `StringId`. And I think we want the same effect as that specialization: forcing the use of `llvm::StringRef` in the API, for example in the `TranslateKey` return type.

But happy to switch the type names to `KeyContext`. I wrote this one before the others and so hadn't come up with a good pattern.

toolchain/base/value_store.h
chandlerc: Switched to breaking out the lambda here and elsewhere, seems an easy way to improve readability.

toolchain/check/check.cpp
chandlerc: Yeah, might be worthwhile. But would prefer in a follow-up, as I'm not quite sure where its best to add the helper.

toolchain/check/decl_name_stack.cpp
chandlerc: Done. (Didn't *quite* get to a single line, but close.)

toolchain/check/scope_stack.cpp
chandlerc: Computing the size of the set requires walking the entire set. =/ The only way to avoid that is to have a variable that separately tracks the size. I can add such a variable if you think it's worthwhile, or remove the reserve entirely if you think that's better. I've not looked carefully at this code so I'm not sure what would be better.

toolchain/sem_ir/bind_name.h
chandlerc: It allows an out-of-line definition to be in the header file (and inlinable). This seemed likely a good inlining candidate, and so while moving it to an out-of-line definition I didn't want to also move it to the `.cpp` file.

Methods defined within the class body get an implicit `inline`, but here it is required to be explicit.

chandlerc (commented):
Thanks for the review, PTAL as there is some updated code that's a bit interesting.

toolchain/check/scope_stack.cpp
jonmeow: Given this is for scopes, my *slight* leaning is that it's worth tracking the number of entries in names here.

For context, this is the logic for suspending member entity bodies (function, class) for later evaluation. This in particular is preserving the names in lexical lookup, which might be significant in quantity.

toolchain/sem_ir/bind_name.h
jonmeow: Thanks! I hadn't realized it had an effect for an out-of-line definition in the header.

toolchain/base/value_store.h
jonmeow: FWIW, I was thinking as a type not totally dependent on the store's ValueT, such that the StringId form could pass in StringRef as the template, and others could just use the store's ValueT.

jonmeow:
Thanks! Still LGTM, note the clang-tidy failure though.

chandlerc (commented):
Thanks, landing with the clang-tidy fix (I had a stray `const &` I had added, back to a value now like it was before) and `num_names` to keep reserve the same.

PR #4099, opened 2024-07-02
Fix drop_back call in scope_stack

jonmeow:
I found this through inspection, looking at an array stack data type. Tests pass either way, not sure what a good test would be for regressions (tests do fail if the size doesn't match, but either approach gets an appropriate size). But this is followed by `truncate(remaining_compile_time_bindings)`, so it seems like `drop_back` is a better match than `drop_front`.

### DIFF
--- a/toolchain/check/scope_stack.cpp
+++ b/toolchain/check/scope_stack.cpp
@@ -179,7 +179,7 @@ auto ScopeStack::Suspend() -> SuspendedScope {
 
   // Move any compile-time bindings into the suspended scope.
   for (auto inst_id : llvm::ArrayRef(compile_time_binding_stack_)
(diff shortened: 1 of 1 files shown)


zygoloid (approved):
I spent a while trying to build a testcase for this, but I think currently it's unobservable: the only way we can observe the list of compile-time bindings is to look at how they get associated with newly-declared `Generic`s, and so to see a difference, we'd need to introduce a generic in a scope that we suspended and then resumed. And that means we'd need to declare a generic in a function body.

We don't currently support declaring classes, interfaces, or functions within a function body (the parser rejects), so there's no way to observe the bug that this PR fixes.

=> Approving without a test.

PR #4100, opened 2024-07-02
Create an array stack type for a shared use-case

jonmeow:
Based on discussion around the region handling in generic_region_stack, create a generic structure for the stack-of-vectors support. I also want to add this to InstBlockStack, but that's a little more complex due to GlobalInit, so cutting a PR here to check with review.

My work here is how I noticed #4099; I want to be sure that I'm correct about the issue, but it's the difference between being able to use PeekArray or not.

Note in scope_stack.h, I believe we could remove next_compile_time_index and make it just based on elements_size(). However, I want to verify with you before I make further changes there.


jonmeow:
Depends on #4099

chandlerc:
Shouldn't block this PR either way, but wanted to also ask whether for completely general data structures we should keep them in `//common` instead of `//toolchain/base`?

jonmeow:
> Shouldn't block this PR either way, but wanted to also ask whether for completely general data structures we should keep them in `//common` instead of `//toolchain/base`?

Done

common/array_stack.h
```diff
@@ -0,0 +1,62 @@
...
+namespace Carbon {
+
+// Provides arrays in a stack form. Only the top of the stack can have elements
+// added.
+//
+// By using a single vector for elements, the intent is that as arrays are
+// pushed and popped, the same storage will be reused. This should yield
+// efficiencies for heap allocations.
+template <typename ValueT>
+class ArrayStack {
```
chandlerc: I think it would be useful to give a bit more detailed description of the expected usage patterns, maybe with some example code?

When I first read the file I misunderstood and thought that this was an array that can be manipulated as a stack. But that's not quite accurate -- it is a stack of arrays? That said, I'm not yet sure exactly how to use it from reading the comments, or when I should be using it.

common/array_stack.h
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  // Returns the full set of values on the stack, regardless of whether any
+  // arrays are pushed.
+  auto PeekAllValues() -> llvm::ArrayRef<ValueT> { return elements_; }
+
+  // Adds a value to the top array on the stack.
+  auto PushValue(ValueT value) -> void {
+    CARBON_CHECK(!array_offsets_.empty())
+        << "Must call PushRegion before PushValue.";
```
chandlerc: I think this should now be:

```suggestion
        << "Must call PushArray before PushValue.";
```

But it's a bit hard for me to be certain until I understand the overarching abstraction a bit better.

chandlerc (commented):
(Figured I can help out w/ this review as I have no idea how to test the other change either...)

common/array_stack.h
jonmeow: Updated comments to try to explain more.

common/array_stack.h
```diff
@@ -0,0 +1,74 @@
...
+    CARBON_CHECK(!array_offsets_.empty());
+    return llvm::ArrayRef(elements_).slice(array_offsets_.back());
+  }
+
+  // Returns the full set of values on the stack, regardless of whether any
+  // arrays are pushed.
+  auto PeekAllValues() const -> llvm::ArrayRef<ValueT> { return elements_; }
+
+  // Adds a value to the top array on the stack.
+  auto PushValue(ValueT value) -> void {
```
chandlerc: Calling this `Push...` makes the API a bit harder for me tu understand because there isn't a pop and it doesn't interact with the stack-based semantic really...

WDYT about `AddToTop` or `AppendToTop`?

Also open to other words besides `Add` or `Append` here, just trying to find a way to keep it separate from the push/pop/peek done to the arrays.

Also open to other words besides `Top` -- maybe `AddToArray`?

Anyways, curious what you think might work well naming wise.

common/array_stack.h
```diff
@@ -0,0 +1,74 @@
...
+    CARBON_CHECK(!array_offsets_.empty())
+        << "Must call PushArray before PushValue.";
+    elements_.push_back(value);
+  }
+
+  // Returns the current number of values in all arrays.
+  auto elements_size() const -> size_t { return elements_.size(); }
```
chandlerc: The name `elements` seems to surface the implementation name and not really be used elsewhere.

WDYT about `all_values_size`? or `total_size` or even just `size`?

chandlerc (approved):
Ah, nice. I think I understand a bit better how this is used, neat!

I've left a few naming bikeshed questions. I'm happy for the PR to merge and rename afterward if that's easier, or to handle the bikeshedding in real-time if better, up to you. The code LG.

common/array_stack.h
jonmeow: Renamed to `all_values_size`; it seems clearer than `total_size`. I think `size` would easily be confused as being the number of arrays, or the number of values in the top array, rather than the number of values across all arrays.

common/array_stack.h
jonmeow: Renamed to `AppendToTop`

`AddToArray` doesn't work for me well because it makes me wonder _which_ array. The implicit aspect of "top" is not clear to me once you start talking about `Array`, unlike `Peek`. I'm not really wanting to say `AddToTopArray`.

PR #4101, opened 2024-07-02
Split global init out from InstBlockStack.

jonmeow:
Creates a `GlobalInit` class for storing relevant values, pulling functions off `InstBlockStack` and `Context`. Adds a `Context` pointer just so that it doesn't need to be passed in on each call (`Finalize` in particular uses several members).

Note we have several different `InstBlockStack` instances, so several copies of the relevant members were simply unused.


jonmeow:
This is pushed depending on #4100, and only the currently-last commit applies. I might merge it earlier if separately approved since the particular set of changes doesn't depend on the earlier work, it's only in my commit series due to the next step of modifying InstBlockStack for ArrayStack.

chandlerc (approved):
Very nice, LG

PR #4102, opened 2024-07-02
Store additional information for symbolic constants.

zygoloid:
When forming a `ConstantId` for a symbolic constant, add storage to track the generic in which the constant was formed and the index within that generic. These fields are not yet populated.

### DIFF
--- a/toolchain/sem_ir/constant.h
+++ b/toolchain/sem_ir/constant.h
@@ -5,7 +5,6 @@
 #ifndef CARBON_TOOLCHAIN_SEM_IR_CONSTANT_H_
 #define CARBON_TOOLCHAIN_SEM_IR_CONSTANT_H_
 
-#include "llvm/ADT/FoldingSet.h"
 #include "toolchain/sem_ir/ids.h"
 #include "toolchain/sem_ir/inst.h"
 
@@ -16,10 +15,12 @@ namespace Carbon::SemIR {
 struct SymbolicConstant {
   // The constant instruction that defines the value of this symbolic constant.
   InstId inst_id;
-  // The enclosing generic.
+  // The enclosing generic. If this is invalid, then this is an abstract
+  // symbolic constant, such as a constant instruction in the constants block,
+  // rather than one associated with a particular generic.
   GenericId generic_id;
   // The index of this symbolic constant within the generic's list of symbolic
-  // constants.
+  // constants, or invalid if `generic_id` is invalid.
   GenericInstIndex index;
 };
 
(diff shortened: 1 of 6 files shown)


toolchain/sem_ir/ids.h
```diff
@@ -155,18 +159,25 @@ struct ConstantId : public IdBase, public Printable<ConstantId> {
   // logic here. LLVM should still optimize this.
   static constexpr auto Abs(int32_t i) -> int32_t { return i > 0 ? i : -i; }
 
-  // Returns the instruction that describes this constant value, or
+  // Returns the instruction that describes this template constant value, or
   // InstId::Invalid for a runtime value. This is not part of the public
```
geoffromer: Should this say "non-template constant" instead of "runtime value"?

toolchain/sem_ir/ids.h
```diff
@@ -155,18 +159,25 @@ struct ConstantId : public IdBase, public Printable<ConstantId> {
...
-  constexpr auto inst_id() const -> InstId {
-    CARBON_CHECK(is_valid());
-    return InstId(Abs(index) - IndexOffset);
+  constexpr auto template_inst_id() const -> InstId {
+    CARBON_CHECK(is_template());
+    return InstId(index);
+  }
+
+  // Returns the symbolic constant index that describes this symbolic constant
+  // value.  Requires is_symbolic. This is not part of the public interface of
```
geoffromer: "This is not part of the public interface..." seems redundant, since it's in the `private` section. Am I misunderstanding something?

(Same goes for the previous function, although that's arguably a pre-existing problem)

toolchain/sem_ir/constant.h
```diff
@@ -11,6 +11,18 @@
 
 namespace Carbon::SemIR {
 
+// Information about a symbolic constant value. These are indexed by
+// `ConstantId`s for which `is_symbolic` is true.
+struct SymbolicConstant {
+  // The constant instruction that defines the value of this symbolic constant.
+  InstId inst_id;
+  // The enclosing generic.
```
geoffromer: Is it worth explaining that `generic_id` and `index` can be invalid, and what that means?

toolchain/sem_ir/ids.h
zygoloid: This `CHECK`-fails if it's not a template constant value; comment updated to say so.

toolchain/sem_ir/ids.h
zygoloid: Removed.

toolchain/sem_ir/constant.h
zygoloid: Done.

geoffromer (approved):

PR #4103, opened 2024-07-02
Move import refs to their own block.

jonmeow:
This executes on a TODO in AddImportRef to add instructions to their own block instead of the File block. This has an important consequence of removing a pattern from InstBlockStack that added to blocks not currently at the top, cleaning up an issue for ArrayStack. The delta here is then mostly in different formatting of the import refs, a consequence of the separation.

### DIFF
--- a/toolchain/check/import_ref.cpp
+++ b/toolchain/check/import_ref.cpp
@@ -67,9 +67,7 @@ auto AddImportRef(Context& context, SemIR::ImportIRInst import_ir_inst,
   // somewhere, because it's referenced by other instructions and needs to be
   // visible in textual IR. Adding it to the file block is arbitrary but is the
   // best place we have right now.
-  //
-  // TODO: Consider adding a dedicated block for import_refs.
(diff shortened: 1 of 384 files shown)


jonmeow:
Depends on #4101

chandlerc (approved):
Wow, I really like this... LG!!

toolchain/check/import_ref.cpp
```diff
@@ -67,9 +67,7 @@ auto AddImportRef(Context& context, SemIR::ImportIRInst import_ir_inst,
   // somewhere, because it's referenced by other instructions and needs to be
   // visible in textual IR. Adding it to the file block is arbitrary but is the
   // best place we have right now.
```
zygoloid: Should we also remove the last line of this comment?

jonmeow: The comment in context.h was copying the important bits, I'd meant to remove this. So I've cut the detail out since it's now on the vector.

PR #4104, opened 2024-07-02
Refactor InstBlockStack to use ArrayStack.

jonmeow:
The use of ArrayStack here is intended to simplify the logic, and also make better use of the inst heap allocations. Prior changes #4101 and #4103 removed the less related logic from InstBlockStack, although #4103 is the actual part that blocked using ArrayStack.

BTW, note the PrintForStackDump implementation was incorrect because it didn't apply size_. This simplification fixes the issue.

### DIFF
--- a/common/array_stack.h
+++ b/common/array_stack.h
@@ -47,6 +47,16 @@ class ArrayStack {
     return llvm::ArrayRef(values_).slice(array_offsets_.back());
   }
 
+  // Returns the array at a specific index.
+  auto PeekArrayAt(int index) const -> llvm::ArrayRef<ValueT> {
+    CARBON_CHECK(index < static_cast<int>(array_offsets_.size()));
+    auto ref = llvm::ArrayRef(values_).slice(array_offsets_[index]);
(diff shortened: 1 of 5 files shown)


jonmeow:
Depends on #4103

common/array_stack.h
```diff
@@ -47,6 +47,16 @@ class ArrayStack {
     return llvm::ArrayRef(elements_).slice(array_offsets_.back());
   }
 
+  // Returns the array at a specific index.
+  auto PeekArrayAt(int index) const -> llvm::ArrayRef<ValueT> {
+    CARBON_CHECK(index < static_cast<int>(array_offsets_.size()));
```
chandlerc: `llvm::ArrayRef` has an equivalent assert, do we need it here as well?

chandlerc (approved):
Nice!

LG with a minor nit inline.

common/array_stack.h
jonmeow: Done.

jonmeow:
Also added some tests for ArrayStack functions while I was in there.

PR #4105, opened 2024-07-02
Establish toolchain and language versioning

chandlerc:
Proposal for how Carbon version numbers work:

-   A single version across language, standard library, compiler, linker, etc.
-   Semantic Versioning (SemVer) based
-   Details of how SemVer criteria for major, minor, and patch should apply to
    Carbon
-   Details of how we will operate before 1.0 and how this connects to Carbon's
    milestones
-   Directional guidance for future work including post-1.0 versions, LTS
    versions, and standardization

### DIFF
--- a/docs/project/versioning.md
+++ b/docs/project/versioning.md
@@ -37,11 +37,13 @@ tools released by the main Carbon project.
 The scheme conforms to and is closely based on Semantic Versioning (SemVer):
 
 -   Carbon versions: `MAJOR.MINOR.PATCH`
--   `MAJOR` increments on backwards incompatible changes, including a
-    deprecation that might trigger a build-breaking warning.
--   `MINOR` increments only expected during early development with a major
-    version of `0`.
--   `PATCH` increments represent bug fixes only.
+-   Releases with backwards incompatible changes, including a
+    deprecation that might trigger a build-breaking warning,
+    increment `MAJOR` after reaching the `1.0` milestone, and
+    increment `MINOR` before then.
+-   Releases with only backwards compatible changes are not
+    expected to happen.
+-   Releases containing only bug fixes increment `PATCH`.
 -   Pre-release suffixes:
     -   `MAJOR.MINOR.PATCH-rc.N`: The N-th potentially viable candidate for a
         release.
--- a/docs/project/versioning.md
+++ b/docs/project/versioning.md
@@ -139,7 +139,7 @@ fixes can still be disruptive and so we work to hold a high bar for them:
 
 -   They must be in some way fixing a _regressions_ for users from a previous
     release, not merely a missing feature.
-    -   Can even be a regression the overall cohesion or reliability of the
+    -   Can even be a regression in the overall cohesion or reliability of the
         language or tools, which a bug in a new feature might erode.
     -   Key is that we motivate any patch fix through the lens of a regression
         fix and stabilization rather than fixing forward.
@@ -262,7 +262,7 @@ in-flight edits and all manner of other variations.
 Mechanically, we prefix the `dev` pre-release version component with `0` in the
 same way as `nightly` is prefixed to ensure effective ordering. We don't add any
 additional information to keep the mechanics of development builds simple --
-there is no easy way to extract the date of the build . The exact timestamp of
+for example, there is no easy way to extract the date of the build. The exact timestamp of
 the build may be available but doesn't participate for simplicity and minimizing
 cache impact.
 
--- a/docs/project/versioning.md
+++ b/docs/project/versioning.md
@@ -1,0 +1,269 @@
+# Toolchain and language versioning
+
+<!--
+Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+Exceptions. See /LICENSE for license information.
+SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+-->
+
+[Pull request](https://github.com/carbon-language/carbon-lang/pull/4105)
+
+<!-- toc -->
+
+## Table of contents
+
+-   [Overview](#overview)
+-   [Major version increments](#major-version-increments)
+    -   [Breaking changes](#breaking-changes)
+    -   [Exclusions to what constitutes a breaking change](#exclusions-to-what-constitutes-a-breaking-change)
+-   [Minor version increments](#minor-version-increments)
+-   [Patch version increments](#patch-version-increments)
+    -   [Examples:](#examples)
+-   [Pre-release versions](#pre-release-versions)
+    -   [Release qualification pre-releases: `MAJOR.MINOR.PATCH-rc.N`](#release-qualification-pre-releases-majorminorpatch-rcn)
+    -   [Incremental development versions: `MAJOR.MINOR.PATCH-0.{nightly,dev}.N`](#incremental-development-versions-majorminorpatch-0nightlydevn)
+        -   [Nightly pre-release versions](#nightly-pre-release-versions)
+        -   [Development pre-release versions](#development-pre-release-versions)
+-   [Relevant proposal](#relevant-proposal)
+
+<!-- tocstop -->
+
+## Overview
+
+Carbon uses a single versioning scheme across both the language itself and
+toolchain, including the standard library, compiler, linker, and all development
+tools released by the main Carbon project.
+
+The scheme conforms to and is closely based on Semantic Versioning (SemVer):
+
+-   Carbon versions: `MAJOR.MINOR.PATCH`
+-   `MAJOR` increments on backwards incompatible changes, including a
+    deprecation that might trigger a build-breaking warning.
+-   `MINOR` increments only expected during early development with a major
+    version of `0`.
+-   `PATCH` increments represent bug fixes only.
+-   Pre-release suffixes:
+    -   `MAJOR.MINOR.PATCH-rc.N`: The N-th potentially viable candidate for a
+        release.
+    -   `MAJOR.MINOR.PATCH-0.nightly.YYYY.MM.DD`: A nightly incremental
+        development build on a particular day during development of that
+        version.
+    -   `MAJOR.MINOR.PATCH-0.dev`: An interactive, incremental development build
+        on a particular day by some developer during development of that
+        version.
+
+See the sections below for the details of each aspect of these versions.
+
+## Major version increments
+
+Aligned with SemVer, the major version must increment for any _breaking change_
+to the language or any part of the toolchain.
+
+The first increment from 0 to 1 is expected to be based on achieving a desired
+milestone of feature completeness and quality to declare the language to have
+reached a stable version.
+
+Subsequent increments are expected to be done with a time-based release
+strategy. Features (or breaking changes) ready to ship will do so, and others
+will wait for the next major release. The exact cadence used is future work and
+should be determined based on discussions with customers in the ramp up to and
+after reaching the 1.0 milestone.
+
+However, just because we increment the major version for a major release and
+_can_ make breaking changes doesn't mean we _will_ or _should_. Breaking
+language changes are extraordinarily expensive for users due to the inherent
+scale of churn they impose. It is tempting to try to make non-breaking releases
+instead, but our experience with C++ language, compiler, and standard library
+updates is that truly making no breaking changes is extremely difficult and
+overly constraining. We expect most releases, especially in the early phases of
+the language to involve _some_ amount of breaking change and will simply work to
+make these as cheap to upgrade through and minimal as we can.
+
+At some future point, it is possible that Carbon will become so stable that it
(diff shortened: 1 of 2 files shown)


proposals/p4105.md
```diff
@@ -0,0 +1,280 @@
...
+        being able to sync across all of these with a single, simple versioning
+        scheme is especially important to have the tooling and ecosystem agree
+        about features of the language.
+-   [Software and language evolution](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/project/goals.md#software-and-language-evolution)
+    -   Users of Carbon need a clear model for understanding the language's
+        evolution, tracking it, and responding to significant aspects of it.
+    -   The versioning scheme needs to support a wide variety of versions built
+        in the process of evolving the language without creating confusion.
+
+## Alternatives considered
```
jonmeow, line 196: Is it worth mentioning the alpha/beta discussion? I think they were in the docs version of the proposal, but removed here. Wondering if they should be mentioned for posterity.

chandlerc: Good call, added.

chandlerc:
Ping.

docs/project/versioning.md
```diff
@@ -0,0 +1,269 @@
...
+
+<!-- tocstop -->
+
+## Overview
+
+Carbon uses a single versioning scheme across both the language itself and
+toolchain, including the standard library, compiler, linker, and all development
+tools released by the main Carbon project.
+
+The scheme conforms to and is closely based on Semantic Versioning (SemVer):
```
zygoloid: I think it'd be useful to link to https://semver.org/ here, and maybe specify a version number?). (The idea of semantic versioning predates the SemVer spec, and there are other different and incompatible forms of semantic versioning such as [libtool's system](https://www.gnu.org/software/libtool/manual/html_node/Libtool-versioning.html) so it seems good to be specific.)

docs/project/versioning.md
```diff
@@ -0,0 +1,269 @@
...
+
+-   Carbon versions: `MAJOR.MINOR.PATCH`
+-   `MAJOR` increments on backwards incompatible changes, including a
+    deprecation that might trigger a build-breaking warning.
+-   `MINOR` increments only expected during early development with a major
+    version of `0`.
+-   `PATCH` increments represent bug fixes only.
```
zygoloid: For major and patch, we say what the increments mean, whereas for minor we don't say what they mean, only that we won't do them much. If we're going to explain how to understand the version numbers rather than deferring to the SemVer spec, it would be good to say what the minor portion of the version means too.

Something like this would capture what I understood from our prior discussion:
```suggestion
-   `MAJOR` increments on backwards incompatible changes, including a
    deprecation that might trigger a build-breaking warning.
-   `MINOR` increments on backwards incompatible changes prior to
    reaching a major version of `1`, and is expected to remain at `0`
    afterwards.
-   `PATCH` increments represent bug fixes only.
```

It might be clearer to swizzle this around:
```suggestion
-   Releases with backwards incompatible changes, including a
    deprecation that might trigger a build-breaking warning,
    increment `MAJOR` after reaching the `1.0` milestone, and
    increment `MINOR` before then.
-   Releases with only backwards compatible changes are not
    expected to happen.
-   Releases containing only bug fixes increment `PATCH`.
```

docs/project/versioning.md
```diff
@@ -0,0 +1,269 @@
...
+-   Pre-release suffixes:
+    -   `MAJOR.MINOR.PATCH-rc.N`: The N-th potentially viable candidate for a
+        release.
+    -   `MAJOR.MINOR.PATCH-0.nightly.YYYY.MM.DD`: A nightly incremental
+        development build on a particular day during development of that
+        version.
+    -   `MAJOR.MINOR.PATCH-0.dev`: An interactive, incremental development build
+        on a particular day by some developer during development of that
+        version.
```
zygoloid, line 54: Given that we don't have `-alpha` and `-beta` here, it seems that we don't need the `-0.` prefix to ensure that `dev` and `nightly` are ordered before `rc`. Would it make sense to remove the `0.` part? (We could add it back in future if we add back `alpha` / `beta`.)

docs/project/versioning.md
```diff
@@ -0,0 +1,269 @@
...
+to the language or any part of the toolchain.
+
+The first increment from 0 to 1 is expected to be based on achieving a desired
+milestone of feature completeness and quality to declare the language to have
+reached a stable version.
+
+Subsequent increments are expected to be done with a time-based release
+strategy. Features (or breaking changes) ready to ship will do so, and others
+will wait for the next major release. The exact cadence used is future work and
+should be determined based on discussions with customers in the ramp up to and
```
zygoloid: Is there a better word than "customers" here, that doesn't suggest a financial relationship? Something like "stakeholders" or "affected parties" perhaps?

I'm worried that people might interpret this as our valuing the views of people who give us money over the rest of the community.

docs/project/versioning.md
```diff
@@ -0,0 +1,269 @@
...
+### Exclusions to what constitutes a breaking change
+
+Carbon excludes "reflective code" which is in some way detecting the Carbon
+version, or the presence or absence of features and as a consequence can be
+"broken" by detecting changes that are correctly designed to otherwise be
+non-breaking. We don't want adding features to be considered a breaking change
+and so exclude code that specifically detects such additions.
+
+Also excludes breaking changes to incorrect code unless it was accepted and
+functioning in some useful, and typically widespread, way despite its bugs.
```
zygoloid: This sentence is missing a subject. "Our definition also excludes [...]", or "Carbon also excludes [...]" maybe?

docs/project/versioning.md
```diff
@@ -0,0 +1,269 @@
...
+Note that we still consider restoring the _intended_ "public API" of a release
+to be a bug fix. When these bug fixes theoretically break new code in a way that
+would typically require a major version increment, they may be made with merely
+a patch version increment when they are in fact restoring our intended behavior
+for that release. However, we take the SemVer guarantees very seriously as these
+fixes can still be disruptive and so we work to hold a high bar for them:
+
+-   They must be in some way fixing a _regressions_ for users from a previous
+    release, not merely a missing feature.
+    -   Can even be a regression the overall cohesion or reliability of the
```
zygoloid:
```suggestion
    -   Can even be a regression in the overall cohesion or reliability of the
```

docs/project/versioning.md
```diff
@@ -0,0 +1,269 @@
...
+unambiguous way, and we do that with the `dev` pre-release version. Much like
+nightly versions, these are only produced as artifacts of development activities
+and are never part of any actual release process. These versions are further not
+built expected to be automated or necessarily repeatable. They may contain
+in-flight edits and all manner of other variations.
+
+Mechanically, we prefix the `dev` pre-release version component with `0` in the
+same way as `nightly` is prefixed to ensure effective ordering. We don't add any
+additional information to keep the mechanics of development builds simple --
+there is no easy way to extract the date of the build . The exact timestamp of
```
zygoloid: Maybe:
```suggestion
for example, there is no easy way to extract the date of the build. The exact timestamp of
```
(but at least remove the space before the period).

proposals/p4105.md
```diff
@@ -0,0 +1,305 @@
...
+
+Advantages:
+
+-   Carbon is at a very early stage and is a long way from needing release
+    candidates for stable releases or incrementing its major version numbers.
+-   No need to imagine distant future scenarios.
+
+Disadvantages:
+
+-   We already discuss version numbers in varyous places in the Carbon project,
```
zygoloid:
```suggestion
-   We already discuss version numbers in various places in the Carbon project,
```

proposals/p4105.md
```diff
@@ -0,0 +1,305 @@
...
+-   Establishing how the Carbon project will communicate its updates with
+    version number changes in advance of those updates makes that communication
+    more effective.
+-   Lets us telegraph our intentions and how we are thinking about releasing and
+    cadence to the community and get earlier feedback.
+
+A key to adopting the more detailed versioning plan is that we can change any
+and all of this if and when we need to. This does not lock Carbon into using
+this exact versioning scheme. We should actively collect and listen to feedback
+in this space from potential users and customers.
```
zygoloid: Certainly we should listen to feedback, but is it worth *actively* collecting it (eg, going out and soliciting information)? I don't think it's worth (for example) surveying our current and prospective users to find what they think about our versioning scheme -- if no-one proactively gives feedback I think we can assume it's working well enough.

proposals/p4105.md
```diff
@@ -0,0 +1,305 @@
...
+
+Advantages:
+
+-   Stable languages have the lowest churn costs for users, and are easier to
+    learn at scale.
+
+Disadvantages:
+
+-   We want to plan to be able to improve things and fix mistakes in the
+    language.
```
zygoloid: I think we can say something stronger: this would be in direct opposition to our language goal of supporting software and language evolution.

proposals/p4105.md
```diff
@@ -0,0 +1,305 @@
...
+    become as disruptive as smaller updates to the C++ language itself have been
+    over the years. As a consequence, while it is tempting to hope for a sharp
+    difference here we don't in practice anticipate one.
+
+### Use a custom versioning scheme rather than SemVer.
+
+Advantages:
+
+-   We don't actually use all parts of SemVer, which results in awkward unused
+    componenst of our version number.
```
zygoloid:
```suggestion
    component of our version number.
```

zygoloid (commented):
Looks good. While I have a few comments, they're mostly around the presentation rather than the rules; the only real non-cosmetic question is whether we still want the `-0.` in nightly and dev versions now that it's not necessary to get them to sort before `-alpha` and `-beta`.

docs/project/versioning.md
```diff
@@ -0,0 +1,271 @@
...
+
+The scheme conforms to and is closely based on Semantic Versioning (SemVer):
+
+-   Carbon versions: `MAJOR.MINOR.PATCH`
+-   Releases with backwards incompatible changes, including a
+    deprecation that might trigger a build-breaking warning,
+    increment `MAJOR` after reaching the `1.0` milestone, and
+    increment `MINOR` before then.
+-   Releases with only backwards compatible changes are not
+    expected to happen.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
-   Releases with backwards incompatible changes, including a deprecation that
    might trigger a build-breaking warning, increment `MAJOR` after reaching the
    `1.0` milestone, and increment `MINOR` before then.
-   Releases with only backwards compatible changes are not expected to happen.
```

docs/project/versioning.md
```diff
@@ -0,0 +1,271 @@
...
+and are never part of any actual release process. These versions are further not
+built expected to be automated or necessarily repeatable. They may contain
+in-flight edits and all manner of other variations.
+
+Mechanically, we prefix the `dev` pre-release version component with `0` in the
+same way as `nightly` is prefixed to ensure effective ordering. We don't add any
+additional information to keep the mechanics of development builds simple --
+for example, there is no easy way to extract the date of the build. The exact timestamp of
+the build may be available but doesn't participate for simplicity and minimizing
+cache impact.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
additional information to keep the mechanics of development builds simple -- for
example, there is no easy way to extract the date of the build. The exact
timestamp of the build may be available but doesn't participate for simplicity
and minimizing cache impact.
```

proposals/p4105.md
chandlerc: Agreed, done.

proposals/p4105.md
chandlerc: I like it, done.

docs/project/versioning.md
chandlerc: Done.

docs/project/versioning.md
chandlerc: Yeah, "users" is a more general and better term.

docs/project/versioning.md
chandlerc: I'm hesitant to do this... It seems nice to have `nightly` and `dev` be more explicitly separate categories from `rc` or anything else we have here.

docs/project/versioning.md
chandlerc: Done.

docs/project/versioning.md
zygoloid: OK. I don't have a strong opinion. And maybe relying on the coincidence that `dev` and `nightly` collate before `rc` is too "clever".

zygoloid (approved):
Thanks!

PR #4110, opened 2024-07-04
Build a list of dependent constants to recompute in each instance of a generic.

zygoloid:
For each generic, build a list of instructions describing the computations we need to do when resolving an instance of the generic: this is a list of the instance-specific constants and types that the generic uses. Another way of viewing this list is as a block of Carbon SemIR code that is evaluated in order to form an instance of the generic -- this is referenced in the code as the "eval block" for the generic.

For each instruction in the generic whose type or value is a symbolic constant, replace that type or constant value with a symbolic reference that says "to find the actual type or value, look at index N in the list of values for the generic instance".

For an instruction with a symbolic constant value, we can just add that instruction to our list. For an instruction with a symbolic constant type, however, we may not have a corresponding instruction computing the type within the generic and may need to build a new instruction, but will reuse one where possible. In the case where we build a new instruction, we use the existing substitution code to build the type within the eval block.

For now, this transformation is only done in the declaration region of the generic, not in the definition region. Also, we map back from the symbolic references to the underlying constant value in a few places where we will eventually need to do a lookup into a generic instance, in order to avoid regressing the tests.


toolchain/sem_ir/type.h
```diff
@@ -70,6 +70,13 @@ class TypeStore : public ValueStore<TypeId> {
     return GetAsInst(type_id).TryAs<InstT>();
   }
 
+  // Returns whether two type IDs represent the same type. This includes the
+  // case where they might be in different generics and thus might have
+  // different ConstantIds, but are still symbolically equal.
+  auto EqualAcrossDeclarations(TypeId a, TypeId b) const -> bool {
```
jonmeow: FWIW, I'd lean towards IsEqual... or AreEqual... here, similar to IsComplete (for verb phrasing)

toolchain/check/subst.h
```diff
@@ -5,6 +5,7 @@
 #ifndef CARBON_TOOLCHAIN_CHECK_SUBST_H_
 #define CARBON_TOOLCHAIN_CHECK_SUBST_H_
 
+#include "common/map.h"
```
jonmeow: I may be missing something, but why is this added? Maybe leftover while refactoring? Perhaps ironically to me, things like llvm::function_ref maybe _should_ have an include.

```suggestion
#include "llvm/ADT/STLFunctionalExtras.h"
```

toolchain/check/subst.cpp
```diff
@@ -106,8 +110,13 @@ static auto ExpandOperands(Context& context, Worklist& worklist,
 static auto PopOperand(Context& context, Worklist& worklist, SemIR::IdKind kind,
                        int32_t arg) -> int32_t {
   switch (kind) {
-    case SemIR::IdKind::For<SemIR::InstId>:
+    case SemIR::IdKind::For<SemIR::InstId>: {
+      auto inst_id = SemIR::InstId(arg);
+      if (!inst_id.is_valid()) {
```
jonmeow, line 115: I'll just note, this looks kind of weird versus line 66 in the delta. I don't know if it's worth switching formats since this is consistent with line 122 below, or you could switch both. Or don't do anything, it's probably not going to look as odd later in code, versus when I'm just seeing the delta.

toolchain/check/subst.cpp
```diff
@@ -156,8 +165,8 @@ static auto PopOperand(Context& context, Worklist& worklist, SemIR::IdKind kind,
 
 // Pops the operands of the specified instruction off the worklist and rebuilds
 // the instruction with the updated operands.
```
jonmeow: This sounds like it would unconditionally call rebuild_fn: should the documentation be adjusted?

toolchain/check/subst.h
```diff
@@ -20,6 +21,24 @@ struct Substitution {
 
 using Substitutions = llvm::ArrayRef<Substitution>;
 
+// A function that performs any needed substitution into an instruction, as part
+// of substituting into a symbolic instruction sequence. The instruction ID
+// should be updated as necessary to represent the new instruction. If the
+// function returns true, the instruction is treated as fully-substituted; if it
+// returns false, the instruction will be decomposed and substitution will be
+// performed recursively into its operands.
```
jonmeow: Most of the comments seem to be describing implementation constraints. From reading the implementation, I think this is different: is this a constraint on the caller, or is it a constraint on the implementation? If it's a constraint on the caller, perhaps it should be on SubstInst instead?

toolchain/check/subst.h
```diff
@@ -20,6 +21,24 @@ struct Substitution {
...
+using SubstInstFn = llvm::function_ref<auto(SemIR::InstId& inst_id)->bool>;
+
+// A function that rebuilds an instruction after substitution. `orig_inst_id` is
+// the instruction prior to substitution, and `new_inst` is the substituted
+// instruction. Returns the new instruction ID to use to refer to `new_inst`.
+using SubstRebuildFn = llvm::function_ref<
+    auto(SemIR::InstId orig_inst_id, SemIR::Inst new_inst)->SemIR::InstId>;
+
+// Performs substitution into `inst_id` and its operands recursively.
```
jonmeow: Maybe the flow should have some more comments here? From the comments:

subst_fn: "A function that performs any needed substitution into an instruction"
rebuild_fn: "A function that rebuilds an instruction after substitution."
SubstInst: "Performs substitution into `inst_id` and its operands recursively."

- Is `subst_fn` called on all the operands of `inst_id` unconditionally [unconditional on direct, conditional on indirect], or only recursively [conditional on both direct and indirect]?
- On line 237, "// No need to rebuild this instruction.", why do instructions that return false on subst_fn but have no operands not get rebuilt? (or, why are they allowed to return false when they don't need substitution?)
- I've also commented on `Rebuild`, because the call to `rebuild_fn` looks conditional in a way I don't see explained here. [but note since that's going to be in the cpp file, may still be worth a note here]

Here, maybe something like:

> Performs recursive substitution on instructions, starting with `inst_id`. Each instruction is passed to `subst_fn`:
> - On true, the instruction is a leaf.
> - On false, this recurses into the instruction's operands. After all operands are processed, then when [Rebuild condition], the instruction is passed to `rebuild_fn`.

But it may be worth looking at further than just that.

toolchain/check/generic.cpp
```diff
@@ -23,6 +26,124 @@ auto StartGenericDefinition(Context& context) -> void {
...
+    -> SemIR::InstBlockId {
+  context.inst_block_stack().Push();
+
+  Map<SemIR::InstId, SemIR::InstId> constants_in_generic;
+  // TODO: For the definition region, populate constants from the declaration.
+
+  // TODO: See if we can ensure that the generic region stack is unchanged
+  // throughout this work. If so, we can use a range-based loop here instead.
+  for (size_t i = 0;
+       i != context.generic_region_stack().PeekDependentInsts().size(); ++i) {
```
jonmeow:
```suggestion
  for (auto i : llvm::seq(context.generic_region_stack().PeekDependentInsts().size())) {
```

jonmeow (approved):
The implementation looks good, but I spent a lot of time examining the substitute/rebuild flow changes due to the lambdas. The flow seems fine (although may be worth considering a class with virtual functions instead of lambda parameters?), but a pass on comments seems like it could be helpful.

toolchain/check/subst.cpp
zygoloid: Done.

toolchain/check/subst.h
zygoloid: Yeah, left over from refactoring. Just removing the include since I'm also going to look at switching from lambdas to virtual functions.

toolchain/sem_ir/type.h
zygoloid: `AreEqual...` seems to work quite nicely with the value store accessors: `...types().AreEqual...`.

Done, and also renamed the corresponding function on the `constant_values` store.

toolchain/check/generic.cpp
zygoloid: Hm, yeah, while the collection could get reallocated (eg, by import ref handling), the number of instructions at this level on the stack really shouldn't change. Made this change and also added a CHECK.

toolchain/check/subst.cpp
zygoloid: Noted. I think this does look less weird in the full context of the file.

toolchain/check/subst.h
zygoloid: Changed to talk more about what the return value means and less about what `SubstInst` will do with it.

toolchain/check/subst.h
zygoloid: Did a pass over the comments.

zygoloid (commented):
> The flow seems fine (although may be worth considering a class with virtual functions instead of lambda parameters?), but a pass on comments seems like it could be helpful.

Switched to a class with virtual functions as suggested. I think that is an improvement. I didn't find a really satisfying name for the class, but I think `SubstInstCallbacks` is at least clear as to its purpose.

jonmeow (approved):
Thanks! The edits are helpful for me.

As noted in chat, in my head I'd been thinking a full class for `SubstInst` logic, not just the callbacks, but this still seems a little easier to grasp.

PR #4112, opened 2024-07-08
Add comments for instructions that lack it.

jonmeow:


toolchain/sem_ir/typed_insts.h
```diff
@@ -354,6 +372,7 @@ struct BranchWithArg {
   InstId arg_id;
 };
 
+// A builtin instruction. These are special and listed in `builtin_kind.def`.
```
josh11b: Is this an instruction that represents a built-in function? Or invokes a built-in function?

toolchain/sem_ir/typed_insts.h
```diff
@@ -731,8 +784,12 @@ struct ReturnExpr {
   InstId dest_id;
 };
 
+// Splices a block into the location where this appears. This may be an
+// expression, producing a result with a given type. For example, when
+// constructing from aggregates we may figure out which conversions are required
+// late, and splice parts together.
 struct SpliceBlock {
-  // TODO: Can we make Parse::NodeId more specific?
+  // TODO: Make Parse::NodeId more specific.
```
josh11b: I think this was an honest question -- it seems plausible that splice blocks could appear in pretty arbitrary places.

toolchain/sem_ir/typed_insts.h
jonmeow: I'm a little confused by your question, builtin_kind.def doesn't contain any functions (right now, at least). Perhaps you opened up builtin_function_kind.def instead?

toolchain/sem_ir/typed_insts.h
jonmeow: Fixed; I'd just deleted the comment, and then copy-pasted from the wrong spot.

toolchain/sem_ir/typed_insts.h
zygoloid: Maybe it'd be worth mentioning here that these are all builtin *types* specifically, other than `<error>` which can also be used as a non-type value.

toolchain/sem_ir/typed_insts.h
```diff
@@ -694,6 +742,8 @@ struct Param {
   NameId name_id;
 };
 
+// Modifies a pointee type to be a pointer. This is tracking the `*` in
+// `x: i32*`, where `pointee_id` is `i32` and `type_id` is `i32*`.
```
zygoloid:
```suggestion
// `x: i32*`, where `pointee_id` is `i32` and `type_id` is `type`.
```

toolchain/sem_ir/typed_insts.h
```diff
@@ -826,6 +893,9 @@ struct TemporaryStorage {
   TypeId type_id;
 };
 
+// Access to a tuple member. Versus `TupleIndex`, this handles indirect access,
+// such as `var tuple: (i32, i32) = (0, 1)` needing to access the `i32` values
+// for assignment.
```
zygoloid:
```suggestion
// Access to a tuple member. Versus `TupleIndex`, this handles access where
// the index was inferred rather than being specified as an expression,
// such as `var tuple: (i32, i32) = (0, 1)` needing to access the `i32` values
// for assignment.
```

toolchain/sem_ir/typed_insts.h
```diff
@@ -905,6 +981,8 @@ struct UnboundElementType {
   TypeId element_type_id;
 };
 
+// An ephemeral reference to a value. For example, when indexing an array, a
+// reference to the accessed element.
```
zygoloid:
```suggestion
// Converts from a value expression to an ephemeral reference expression, in
// the case where the value representation of the type is a pointer. For
// example, when indexing a value expression of array type, this is used to
// form a reference to the array object.
```

toolchain/sem_ir/typed_insts.h
```diff
@@ -913,6 +991,8 @@ struct ValueAsRef {
   InstId value_id;
 };
 
+// In an initializing expression, when the initializing value can be directly
+// used, tracking for the value.
```
zygoloid:
```suggestion
// Converts an initializing expression to a value expression, in the case
// where the initializing representation is the same as the value
// representation.
```

toolchain/sem_ir/typed_insts.h
jonmeow: I've added an example. I really don't want to get into documenting what's already documented in builtin_kind.def though: to me, it's opening the door for skew because this doesn't own the contents of the enum.

zygoloid (approved):

PR #4117, opened 2024-07-09
Refactor whether a function is lowered into InstKind::Define

jonmeow:
This is to remove all the FatalIfEncountered handlers in handle.cpp. They just feel like noise when reading the file. Plus it's one less bit of boilerplate to add for instructions that don't lower.

Note that I left HandleParam/HandleAddrPattern. I'd be happy to change those to just set lowered=false too, but was hesitant to given the separate logic.

Also, I'm separately considering migrating the macro logic into similar constexpr things. If I do, I might switch Define to take in a struct. But for how this particular parameter works, the overload felt reasonable, particularly since is_lowered is not used in combination with TerminatorKind.

### DIFF
--- a/toolchain/sem_ir/inst_kind.h
+++ b/toolchain/sem_ir/inst_kind.h
@@ -7,6 +7,7 @@
 
 #include <cstdint>
 
+#include "common/check.h"
 #include "common/enum_base.h"
 #include "llvm/ADT/FoldingSet.h"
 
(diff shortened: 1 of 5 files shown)


toolchain/sem_ir/inst_kind.h
```diff
@@ -82,6 +83,12 @@ class InstKind : public CARBON_ENUM_BASE(InstKind) {
       TerminatorKind terminator_kind = TerminatorKind::NotTerminator) const
       -> Definition<TypedNodeId>;
 
+  // As above, but for instructions which are not lowered; `Define(ir_name)`
+  // should be used instead of specifying `is_lowered = true`.
+  template <typename TypedNodeId>
+  constexpr auto Define(llvm::StringLiteral ir_name, bool is_lowered) const
+      -> Definition<TypedNodeId>;
```
zygoloid, line 90: I would prefer that we use an enum instead of the `bool` parameter plus parameter name comments.

zygoloid (approved):

toolchain/sem_ir/inst_kind.h
jonmeow: Is that including in #4119, you would prefer `.is_lowered = InstIsLowered::True`?

jonmeow: Per discussion, let's use #4119 to settle final state.

PR #4118, opened 2024-07-10
When performing name lookup, determine the generic instance within which the lookup result was found.

zygoloid:
Require types into which qualified lookup is performed to be completely defined. Eventually this will trigger substitution into the definition for generic types.


jonmeow (approved):
Looks good!

PR #4119, opened 2024-07-10
Refactor InstKind to move metadata from macros to the type.

jonmeow:
This adds `DefinitionInfo` for `Define`-based configuration so that parameters are optional. It also makes it easier to provide the equivalent functions on both `Definition` and `Define`.

A common pattern used here is to change from a `switch` with in-line `case`s to instead have `case`s that call an overloaded function. What's happening here is that the instruction type is used to select an overload, and if an overload is not defined, a compiler error would result. Meanwhile, clusters of overloads are being defined using `requires`-based templating, so that equivalent implementations are not copied. This addresses a limitation of a vanilla `switch` approach where it's hard to have redundant cases using conditional logic, while also getting compiler errors when adding new `InstKind` entries, which had been a significant part of why we used macros previously.

This starts hitting some odd clang-format edge cases causing `CARBON_KIND_SWITCH(inst){` (missing space), which I haven't seen before. Adding `CARBON_KIND_SWITCH` to .clang-format works around it.


jonmeow:
Depends on #4117

zygoloid (approved):

PR #4120, opened 2024-07-10
Refactor lower handlers to use overloads.

jonmeow:
Renames Lower::Handle* to Lower::LowerFunctionInst. This allows writing a templated handler for instructions, moving code out of the macro expansion and removing some of the redundancy in things like `HandleAddrOf(..., SemIR::AddrOf inst)`

### DIFF
--- a/toolchain/lower/function_context.cpp
+++ b/toolchain/lower/function_context.cpp
@@ -51,20 +51,30 @@ auto FunctionContext::LowerBlock(SemIR::InstBlockId block_id) -> void {
   }
 }
 
+// Handles typed instructions for LowerInst. We don't use HandleInst
+// overloads for the "if" cases because it's hard to get the overload behavior
+// we want when declarations for all `InstT` types are pre-declared (which is
+// hard to avoid with the macro declarations).
 template <typename InstT>
-static auto FatalErrorIfEncountered(InstT inst) -> void {
-  CARBON_FATAL()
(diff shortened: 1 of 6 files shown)


jonmeow:
Depends on #4119

toolchain/lower/function_context.cpp
```diff
@@ -51,6 +51,32 @@ auto FunctionContext::LowerBlock(SemIR::InstBlockId block_id) -> void {
   }
 }
 
+// Handles typed instructions for LowerInst. We don't use LowerFunctionInst
+// overloads for the "if" cases because it's hard to get the overload behavior
```
josh11b:
```suggestion
// overloads for instructions that are not lowered, are always constant, or are always types because it's hard to get the overload behavior
```

josh11b (approved):

toolchain/lower/function_context.cpp
jonmeow: Rephrased the comment to try to avoid talking about specific cases:

```
// Handles typed instructions for LowerInst. Many instructions lower using
// HandleInst, but others are unsupported or have trivial lowering.
//
// This only calls HandleInst for versions that should have implementations. A
// different approach would be to have the logic below implemented as HandleInst
// overloads. However, forward declarations of HandleInst exist for all `InstT`
// types, which would make getting the right overload resolution complex.
```

josh11b (approved):
(Not merging since it depends on #4119)

PR #4121, opened 2024-07-10
Rename check handlers to HandleParseNode overloads.

jonmeow:
This is for consistency with #4120. Similar to that, we can use overloads on the typed NodeId rather than individually named handlers. There isn't the same caller benefit here though, since the calls from check.cpp are already boilerplate.


chandlerc (approved):
LG!

PR #4122, opened 2024-07-10
Add an instruction to represent a use of a dependent value from a generic instance.

zygoloid:
We can't use the instruction from the generic directly, because it doesn't have the right constant value. Instead add an instruction that models the transition from the constant value in the generic to the constant value in the generic instance.

Also start associating the self generic instance with unqualified lookups that find results in an enclosing generic, so that we track the information necessary to create the new instruction.


toolchain/sem_ir/typed_insts.h
```diff
@@ -641,6 +641,18 @@ struct InitializeFrom {
   InstId dest_id;
 };
 
+// Given an instruction with a constant value that depends on a generic
+// parameter, selects an instance of that instruction with the constant value
+// corresponding to a particular instance of the generic.
+struct InstanceConstant {
```
jonmeow: Should this be specific it's for a GenericInstance? For me, "instance" could also be a class instance or similar; this is a similar problem for me when reading the IR, since we might have constants for type instances there. From the comments, I wouldn't expect this to be used for instances of arbitrary entities. Maybe there's a better name than "GenericInstanceConstant" though, perhaps just "GenericConstant"?

toolchain/sem_ir/generic.h
```diff
@@ -92,6 +107,19 @@ class GenericInstanceStore : public Yaml::Printable<GenericInstanceStore> {
...
+// generic. Note that this does not perform substitution, and will return
+// `Invalid` if the substituted constant value is not yet known.
+auto GetConstantInInstance(const File& sem_ir, GenericInstanceId instance_id,
+                           ConstantId const_id) -> ConstantId;
+
+// Gets the substituted constant value of an instruction within a specified
+// instance of a generic. Note that this does not perform substitution, and will
+// return `Invalid` if the substituted constant value is not yet known.
+auto GetConstantValueInInstance(const File& sem_ir,
+                                GenericInstanceId instance_id, InstId inst_id)
```
jonmeow, line 120: Should these mention "generic" in the name, so that it's clear from the signature an instance of _what_? e.g., GetConstantInGenericInstance

toolchain/check/eval.cpp
```diff
@@ -1121,6 +1122,11 @@ auto TryEvalInst(Context& context, SemIR::InstId inst_id, SemIR::Inst inst)
           Phase::Template);
     }
 
+    case CARBON_KIND(SemIR::InstanceConstant instance):
+      // Pull the instance-specific constant value out of the generic instance.
+      return SemIR::GetConstantValueInInstance(
+          context.sem_ir(), instance.instance_id, instance.inst_id);
```
```suggestion
    case CARBON_KIND(SemIR::InstanceConstant instance): {
      // Pull the instance-specific constant value out of the generic instance.
      return SemIR::GetConstantValueInInstance(
          context.sem_ir(), instance.instance_id, instance.inst_id);
    }
```

Minor, but while I know this macro compiles correctly without braces when it's one statement, I'd feel better with braces in case more statements get added.

toolchain/sem_ir/typed_insts.h
```diff
@@ -641,6 +641,18 @@ struct InitializeFrom {
   InstId dest_id;
 };
 
+// Given an instruction with a constant value that depends on a generic
+// parameter, selects an instance of that instruction with the constant value
```
jonmeow: Is this limited to one generic parameter, or is it generic parameters plural?
```suggestion
// Given an instruction with a constant value that depends on generic
// parameters, selects an instance of that instruction with the constant value
```

jonmeow (commented):
Looking at this, had you considered a GenericNameRef? Right now, you create 2 instructions for a NameRef that involves a generic, or 32 bytes. We could also have a GenericNameRef that stores the inst_id+instance_id in a side table, or 24 bytes per name reference (in this example, name_id + index as the 2 instruction args).

I'm wondering about this mainly because I expect references to generics to be a common pattern, so shaving the size of references might have significant impact.

Also, I think I'm missing something: why is the instance associated into an instruction _after_ name lookup, rather than the instruction in name lookup having the instance id associated? That is, I think I'd understand the current InstanceConstant approach better if it were what's in name lookup, versus being created per name lookup.

toolchain/sem_ir/typed_insts.h
zygoloid: I'd prefer to keep this as is. I think it's correct: if the value depends on multiple generic parameters, then it depends on a generic parameter (in more than one way), and while I'm not entirely sure why, I find the singular easier to read and think about. We also use the singular [in the design](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/generics/terminology.md#dependent-names).

zygoloid:
> Also, I think I'm missing something: why is the instance associated into an instruction _after_ name lookup, rather than the instruction in name lookup having the instance id associated? That is, I think I'd understand the current InstanceConstant approach better if it were what's in name lookup, versus being created per name lookup.

Per toolchain discussion on 2024-06-25, we model generic instances as a minimal overlay over the generic, rather than a separate new entity. That means that we don't create a new name lookup table / name scope per instance, and we don't create instance-specific versions of each (dependent) instruction within a generic. There is no instruction representing the version of an instruction from a generic, with the constant value / type taken from a specific instance; that's what the new `InstanceConstant` is representing.

We *could* generate the `InstanceConstant`s eagerly as part of forming the generic instance, or at least cache them on the generic instance. However, that would mean that we'd need to treat either `NameRef`s or `InstanceConstant`s that appear in a generic as a special case when forming instances of that generic, because we need to substitute into the arguments that appear in the argument list of the generic instance in the `InstanceConstant` when substituting into the generic. For example:

```
class A(T:! type) { var x: T; }
class B(T:! type) { extend base: A(T*); }
fn F[U:! type](b: B(U)) -> T* { return b.x; }
```

The generic `F` has a name reference to `x` in `A(U*)`. When forming an instance of `F`, say with `U = i32`, we want to remap that reference to `A(U*).x` into a reference to `A(i32*).x`. Normally, forming an instance just means substituting throughout the dependent instructions within `F`, so this works out if the `InstanceConstant` that picks the `A(U*)` version of `x` is within `F`. But if we store / cache the `InstanceConstant` with the `GenericInstance`, then we'll somehow need to know when transforming the `NameRef` that we need to pick a different instance of `x`.

So I think putting the reference to the generic instance with the code that uses the name `x` is the simpler choice.

> Looking at this, had you considered a GenericNameRef? Right now, you create 2 instructions for a NameRef that involves a generic, or 32 bytes. We could also have a GenericNameRef that stores the inst_id+instance_id in a side table, or 24 bytes per name reference (in this example, name_id + index as the 2 instruction args).
> 
> I'm wondering about this mainly because I expect references to generics to be a common pattern, so shaving the size of references might have significant impact.

Yeah, this may be a good thing to explore. There are some other dimensions here: for a member name reference, it'd be good for the SemIR to mention the base expression in which we accessed the member, too. For what it's worth, I think the overhead here is even a little higher than you're suggesting: there's also 4 bytes for the constant ID and 4 bytes for the node ID, so I think it's 20 bytes extra for the second instruction versus maybe 8 bytes per instruction in a side table. On the other hand, there's some maintenance and code complexity costs in the side table, plus worse locality for SemIR traversal.

I'll add a TODO to explore this.

toolchain/sem_ir/typed_insts.h
zygoloid: Renamed to `SpecificConstant` per open discussion.

toolchain/sem_ir/generic.h
zygoloid: These aren't new in this PR, just moved here from check. Will rename as part of the larger `GenericInstance` -> `Specific` renaming.

jonmeow (approved):
LGTM, thanks for applying the `Specific` changes from open discussion yesterday.

toolchain/sem_ir/typed_insts.h
```diff
@@ -821,6 +821,22 @@ struct ReturnExpr {
...
+// TODO: We only form these as the instruction referenced by a `NameRef`.
+// Consider merging an `SpecificConstant` + `NameRef` into a new form of
+// instruction in order to give a more compact representation.
+struct SpecificConstant {
+  static constexpr auto Kind = InstKind::SpecificConstant.Define<Parse::NodeId>(
+      {.ir_name = "specific_constant", .is_lowered = false});
+
+  TypeId type_id;
+  InstId inst_id;
+  GenericInstanceId instance_id;
```
jonmeow, line 837: Should `GenericInstanceId instance_id` -> `SpecificId specific_id`? Sort of a broad comment here. But, I'm just going to merge the PR since I think this is reasonable to handle separately.

toolchain/check/eval.cpp
```diff
@@ -1122,6 +1123,12 @@ auto TryEvalInst(Context& context, SemIR::InstId inst_id, SemIR::Inst inst)
           Phase::Template);
     }
 
+    case CARBON_KIND(SemIR::SpecificConstant instance): {
+      // Pull the instance-specific constant value out of the generic instance.
```
jonmeow, line 1127: Note, comments like this are part of why I was asking my other question... To me, this is now just producing out the specific's constant value. "instance-specific" reads a little awkward for me if we're trying to use "specific" as a term.

PR #4124, opened 2024-07-11
Introduce a source generator and end-to-end compile benchmarks

chandlerc:
The big addition here is a very, very rough and very early skeleton of a source code generator framework. This builds upon the lexers identifier synthesis logic, improving on its framework and wiring it up with the most rudimentary of source file generation. This is just enough to roughly replicate my "big API file" source code benchmarks.

The source generation works *very* hard to both vary the structure and content of the source as much as possible while ensuring the same *total* amount of each construct is in use, from bytes in identifiers to line breaks, parameters, etc. This lets us generate randomly structure inputs that should consistently take the exact same amount of total work to compile.

The complex identifier synthesis logic from the lexer's benchmark is moved over here and the lexer uses APIs in the source generator for identifiers. The other source synthesis in the lexer's benchmark isn't yet moved over, but should likely be slowly absorbed here as it can be refactored into a more principled and re-usable form. Some bits may stay of course if they're just too lexer-specific.

Next, this adds a simple end-to-end compile benchmark for the driver that directly and much more clearly reproduces all the measurements I've done manually up until now. It should also be easy to extend to more patterns over time as we add support to the source generator to produce those patterns.

Last but not least, I've added a tiny CLI to the source generator so that you can generate source code manually. This is especially nice for generating demo source code to actually run through the driver or look at in an editor. The CLI can also generate C++ source code which lets us do some minimal comparative benchmarking between Carbon and C++/Clang.

There are huge number of TODOs in the source generation framework. This is going to be a large ongoing effort I suspect.

There are also a bunch of rough edges I've left to try and get this out for review sooner. I've left TODOs for refactorings that really need to be done here, but hoping these can maybe be follow-ups. If not, please flag and I'll try to layer them on here.

Sample compile benchmark output, nicely showing where we are w.r.t. our goal speeds (2x behind on lex and check, 5x on parse) at least on a recent AMD server CPU:
```
------------------------------------------------------------------------------------------------------
Benchmark                                                 Time             CPU   Iterations      Lines
------------------------------------------------------------------------------------------------------
BM_CompileAPIFileDenseDecls<Phase::Lex>/256           29420 ns        29419 ns        22860 6.62847M/s
BM_CompileAPIFileDenseDecls<Phase::Lex>/1024         146130 ns       146128 ns         4840 6.69959M/s
BM_CompileAPIFileDenseDecls<Phase::Lex>/4096         601584 ns       601577 ns         1020 6.69573M/s
BM_CompileAPIFileDenseDecls<Phase::Lex>/16384       2547578 ns      2547313 ns          280   6.404M/s
BM_CompileAPIFileDenseDecls<Phase::Lex>/65536      10816591 ns     10816389 ns           80 6.05193M/s
BM_CompileAPIFileDenseDecls<Phase::Lex>/262144     52191320 ns     52189828 ns           20 5.02261M/s
BM_CompileAPIFileDenseDecls<Phase::Parse>/256        101706 ns       101698 ns         6900 1.91745M/s
BM_CompileAPIFileDenseDecls<Phase::Parse>/1024       512161 ns       512162 ns         1380  1.9115M/s
BM_CompileAPIFileDenseDecls<Phase::Parse>/4096      2078426 ns      2078430 ns          340   1.938M/s
BM_CompileAPIFileDenseDecls<Phase::Parse>/16384     8795786 ns      8795583 ns          100 1.85468M/s
BM_CompileAPIFileDenseDecls<Phase::Parse>/65536    35073596 ns     35072973 ns           20 1.86639M/s
BM_CompileAPIFileDenseDecls<Phase::Parse>/262144  151100688 ns    151097370 ns           20 1.73483M/s
BM_CompileAPIFileDenseDecls<Phase::Check>/256        957059 ns       957049 ns          740 203.751k/s
BM_CompileAPIFileDenseDecls<Phase::Check>/1024      1956134 ns      1955985 ns          360 500.515k/s
BM_CompileAPIFileDenseDecls<Phase::Check>/4096      5797864 ns      5797417 ns          120 694.792k/s
BM_CompileAPIFileDenseDecls<Phase::Check>/16384    21219608 ns     21217584 ns           40 768.843k/s
BM_CompileAPIFileDenseDecls<Phase::Check>/65536    96311116 ns     96302334 ns           20 679.734k/s
BM_CompileAPIFileDenseDecls<Phase::Check>/262144  371637963 ns    371609964 ns           20 705.387k/s
```

Lest someone think this is *bad*, the fact that we're already within 2x of our rather audacious goals makes me quite happy. =D

### DIFF
--- a/common/benchmark_main.cpp
+++ b/common/benchmark_main.cpp
@@ -28,6 +28,7 @@ auto GetBenchmarkExePath() -> llvm::StringRef {
 
 }  // namespace Carbon::Testing
 
+// TODO: Refactor this to share code with `gtest_main.cpp`.
 auto main(int orig_argc, char** orig_argv) -> int {
   // Do LLVM's initialization first, this will also transform UTF-16 to UTF-8.
   Carbon::InitLLVM init_llvm(orig_argc, orig_argv);
--- a/common/benchmark_main.cpp
+++ b/common/benchmark_main.cpp
@@ -2,16 +2,39 @@
 // Exceptions. See /LICENSE for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+#include "common/benchmark_main.h"
+
 #include <benchmark/benchmark.h>
 
+#include <string>
+
 #include "absl/flags/parse.h"
+#include "common/check.h"
+#include "common/exe_path.h"
 #include "common/init_llvm.h"
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/StringRef.h"
 
+static bool after_main = false;
+static llvm::StringRef exe_path;
+
+namespace Carbon::Testing {
+
+auto GetBenchmarkExePath() -> llvm::StringRef {
+  CARBON_CHECK(after_main)
+      << "Must not query the executable path until after `main` is entered!";
+  return exe_path;
+}
+
+}  // namespace Carbon::Testing
+
 auto main(int orig_argc, char** orig_argv) -> int {
   // Do LLVM's initialization first, this will also transform UTF-16 to UTF-8.
   Carbon::InitLLVM init_llvm(orig_argc, orig_argv);
+
+  std::string exe_path_storage = Carbon::FindExecutablePath(orig_argv[0]);
+  exe_path = exe_path_storage;
+  after_main = true;
 
   // Inject a flag to override the defaults for benchmarks. This can still be
   // disabled by user arguments.
--- a/common/benchmark_main.h
+++ b/common/benchmark_main.h
@@ -1,0 +1,22 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#ifndef CARBON_COMMON_BENCHMARK_MAIN_H_
+#define CARBON_COMMON_BENCHMARK_MAIN_H_
+
+#include "llvm/ADT/StringRef.h"
+
+// When using the Carbon `main` function for benchmarks, we export some extra
+// information about the test binary that can be accessed with this header.
+//
+// TODO: Refactor this to share code with `gtest_main.h`.
+
+namespace Carbon::Testing {
+
+// The executable path of the benchmark binary.
+auto GetBenchmarkExePath() -> llvm::StringRef;
+
+}  // namespace Carbon::Testing
+
+#endif  // CARBON_COMMON_BENCHMARK_MAIN_H_
--- a/common/benchmark_main.h
+++ b/common/benchmark_main.h
@@ -14,7 +14,7 @@
 
 namespace Carbon::Testing {
 
-// The executable path of the benchmark binary.
+// Returns the executable path of the benchmark binary.
 auto GetBenchmarkExePath() -> llvm::StringRef;
 
 }  // namespace Carbon::Testing
--- a/testing/base/source_gen.cpp
+++ b/testing/base/source_gen.cpp
@@ -55,9 +55,9 @@ static auto EstimateAvgMethodDeclLines(SourceGen::MethodDeclParams params)
 
 // Note that this should match the heuristics used when formatting.
 // TODO: See top-level TODO about line estimates and formatting.
-static auto EstimateAvgClassLines(SourceGen::ClassParams params) -> double {
-  // Blank line, comment line, and class open line.
-  double avg = 3.0;
+static auto EstimateAvgClassDefLines(SourceGen::ClassParams params) -> double {
+  // Comment line, and class open line.
+  double avg = 2.0;
 
   // One comment line and blank line per function, plus the function lines.
   avg +=
@@ -86,19 +86,28 @@ auto SourceGen::GenAPIFileDenseDecls(int target_lines, DenseDeclParams params)
   std::string source;
   llvm::raw_string_ostream os(source);
 
-  double avg_class_lines = EstimateAvgClassLines(params.class_params);
-  int num_classes = static_cast<double>(target_lines) / avg_class_lines;
-  int expected_lines = num_classes * avg_class_lines;
+  // Figure out how many classes fit in our target lines, each separated by a
+  // blank line. We need to account the comment lines below to start the file.
+  // Note that we want a blank line after our file comment block, so every class
+  // needs a blank line.
+  constexpr int NumFileCommentLines = 4;
+  double avg_class_lines = EstimateAvgClassDefLines(params.class_params);
+  CARBON_CHECK(target_lines > NumFileCommentLines + avg_class_lines)
+      << "Not enough target lines to generate a single class!";
+  int num_classes = static_cast<double>(target_lines - NumFileCommentLines) /
+                    (avg_class_lines + 1);
+  int expected_lines =
+      NumFileCommentLines + num_classes * (avg_class_lines + 1);
+
   os << "// Generated " << (!IsCpp() ? "Carbon" : "C++") << " source file.\n";
   os << llvm::formatv("// {0} target lines: {1} classes, {2} expected lines",
                       target_lines, num_classes, expected_lines)
      << "\n";
-  os << "//\n// Generating as an API file with dense declarations.\n\n";
+  os << "//\n// Generating as an API file with dense declarations.\n";
 
   auto class_gen_state = GetClassGenState(num_classes, params.class_params);
-  llvm::ListSeparator line_sep("\n");
   for ([[maybe_unused]] int i : llvm::seq(num_classes)) {
-    os << line_sep;
+    os << "\n";
     GenerateClassDef(params.class_params, class_gen_state, os);
   }
 
@@ -117,7 +126,7 @@ auto SourceGen::GetShuffledIds(int number, int min_length, int max_length,
     -> llvm::SmallVector<llvm::StringRef> {
   llvm::SmallVector<llvm::StringRef> ids =
       GetIds(number, min_length, max_length, uniform);
-  std::shuffle(ids.begin(), ids.end(), rng);
+  std::shuffle(ids.begin(), ids.end(), rng_);
   return ids;
 }
 
@@ -129,8 +138,173 @@ auto SourceGen::GetShuffledUniqueIds(int number, int min_length, int max_length,
          "lengths <= 3";
   llvm::SmallVector<llvm::StringRef> ids =
       GetUniqueIds(number, min_length, max_length, uniform);
-  std::shuffle(ids.begin(), ids.end(), rng);
+  std::shuffle(ids.begin(), ids.end(), rng_);
   return ids;
+}
+
+
+auto SourceGen::GetIds(int number, int min_length, int max_length, bool uniform)
+    -> llvm::SmallVector<llvm::StringRef> {
+  llvm::SmallVector<llvm::StringRef> ids =
+      GetIdsImpl(number, min_length, max_length, uniform,
+                 [this](int length, int length_count,
+                        llvm::SmallVectorImpl<llvm::StringRef>& dest) {
+                   auto length_ids = GetSingleLengthIds(length, length_count);
+                   dest.append(length_ids.begin(), length_ids.end());
+                 });
+
+  return ids;
+}
+
+auto SourceGen::GetUniqueIds(int number, int min_length, int max_length,
+                             bool uniform)
+    -> llvm::SmallVector<llvm::StringRef> {
+  CARBON_CHECK(min_length >= 4)
+      << "Cannot trivially guarantee enough distinct, unique identifiers for "
+         "lengths <= 3";
+  llvm::SmallVector<llvm::StringRef> ids =
+      GetIdsImpl(number, min_length, max_length, uniform,
+                 [this](int length, int length_count,
+                        llvm::SmallVectorImpl<llvm::StringRef>& dest) {
+                   AppendUniqueIdentifiers(length, length_count, dest);
+                 });
+
+  return ids;
+}
+
+auto SourceGen::GetSingleLengthIds(int length, int number)
+    -> llvm::ArrayRef<llvm::StringRef> {
+  llvm::SmallVector<llvm::StringRef>& ids =
+      ids_by_length_.Insert(length, {}).value();
+
+  if (static_cast<int>(ids.size()) < number) {
+    ids.reserve(number);
+    for (int _ : llvm::seq<int>(ids.size(), number)) {
+      auto id_storage = llvm::MutableArrayRef(reinterpret_cast<char*>(
+          storage_.Allocate(/*Size=*/length, /*Alignment=*/1)), length);
+      GenerateRandomIdentifier(id_storage);
+      llvm::StringRef new_id(id_storage.data(), length);
+      ids.push_back(new_id);
+    }
+    CARBON_CHECK(static_cast<int>(ids.size()) == number);
+  }
+  return llvm::ArrayRef(ids).slice(0, number);
+}
+
+static auto IdentifierStartChars() -> llvm::ArrayRef<char> {
+  static llvm::SmallVector<char> chars = [] {
+    llvm::SmallVector<char> chars;
+    for (char c : llvm::seq_inclusive('A', 'Z')) {
+      chars.push_back(c);
+    }
+    for (char c : llvm::seq_inclusive('a', 'z')) {
+      chars.push_back(c);
+    }
+    return chars;
+  }();
+  return chars;
+}
+
+static auto IdentifierChars() -> llvm::ArrayRef<char> {
+  static llvm::SmallVector<char> chars = [] {
+    llvm::ArrayRef<char> start_chars = IdentifierStartChars();
+    llvm::SmallVector<char> chars(start_chars.begin(), start_chars.end());
+    chars.push_back('_');
+    for (char c : llvm::seq_inclusive('0', '9')) {
+      chars.push_back(c);
+    }
+    return chars;
+  }();
+  return chars;
+}
+
+constexpr static llvm::StringRef NonCarbonCppKeywords[] = {
+    "asm", "do",     "double", "float", "int",      "long",
+    "new", "signed", "try",    "unix",  "unsigned", "xor",
+};
+
+// Returns a random identifier string of the specified length.
+//
+// Ensures this is a valid identifier, avoiding any overlapping syntaxes or
+// keywords both in Carbon and C++.
+//
+// This routine is somewhat expensive and so is useful to cache and reduce the
+// frequency of calls. However, each time it is called it computes a completely
+// new random identifier and so can be useful to eventually find a distinct
+// identifier when needed.
+auto SourceGen::GenerateRandomIdentifier(llvm::MutableArrayRef<char> id_storage) -> void {
+  llvm::ArrayRef<char> start_chars = IdentifierStartChars();
+  llvm::ArrayRef<char> chars = IdentifierChars();
+
+  auto id = llvm::StringRef(id_storage.data(), id_storage.size());
+  do {
+    id_storage[0] = start_chars[absl::Uniform<int>(rng_, 0, start_chars.size())];
+    for (int i : llvm::seq<int>(1, id_storage.size())) {
+      id_storage[i] = chars[absl::Uniform<int>(rng_, 0, chars.size())];
+    }
+  } while (
+      // TODO: Clean up and simplify this code. With some small refactorings and
+      // post-processing we should be able to make this both easier to read and
+      // less inefficient.
+      llvm::any_of(Lex::TokenKind::KeywordTokens,
+                   [id](auto token) { return id == token.fixed_spelling(); }) ||
+      llvm::is_contained(NonCarbonCppKeywords, id) ||
+      (llvm::is_contained({'i', 'u', 'f'}, id[0]) &&
+       llvm::all_of(id.substr(1),
+                    [](const char c) { return llvm::isDigit(c); })));
+}
+
+// Appends a number of unique, random identifiers with a particular length to
+// the provided destination vector.
+//
+// Uses, and when necessary grows, a cached sequence of random identifiers with
+// the specified length. Because these are cached, this is efficient to call
+// repeatedly, but will not produce a different sequence of identifiers.
+auto SourceGen::AppendUniqueIdentifiers(
+    int length, int number, llvm::SmallVectorImpl<llvm::StringRef>& dest)
+    -> void {
+  auto& [count, unique_ids] = unique_ids_by_length_.Insert(length, {}).value();
+
+  // See if we need to grow our pool of unique identifiers with the requested
+  // length.
+  if (count < number) {
+    // We'll need to insert exactly the requested new unique identifiers. All
+    // our other inserts will find an existing entry.
+    unique_ids.GrowForInsertCount(count - number);
+
+    // Generate the needed number of identifiers.
+    for ([[maybe_unused]] int i : llvm::seq<int>(count, number)) {
+      // Allocate stable storage for the identifier so we can form stable
+      // `StringRef`s to it.
+      auto id_storage = llvm::MutableArrayRef(reinterpret_cast<char*>(
+          storage_.Allocate(/*Size=*/length, /*Alignment=*/1)), length);
+      // Repeatedly generate novel identifiers of this length until we find a
+      // new unique one.
+      for (;;) {
+        GenerateRandomIdentifier(id_storage);
+        auto result =
+            unique_ids.Insert(llvm::StringRef(id_storage.data(), length));
+        if (result.is_inserted()) {
+          break;
+        }
+      }
+    }
+    count = number;
+  }
+  // Append all the identifiers directly out of the set. We make no guarantees
+  // about the relative order so we just use the non-deterministic order of the
+  // set and avoid additional storage.
+  //
+  // TODO: It's awkward the `ForEach` here can't early-exit. This just walks the
+  // whole set which is harmless if inefficient. We should add early exiting
+  // the loop support to `Set` and update this code.
+  unique_ids.ForEach([&](llvm::StringRef id) {
+    if (number > 0) {
+      dest.push_back(id);
+      --number;
+    }
+  });
+  CARBON_CHECK(number == 0);
 }
 
 // An array of the counts that should be used for each identifier length to
@@ -227,16 +401,24 @@ auto SourceGen::GetIdsImpl(int number, int min_length, int max_length,
   llvm::SmallVector<llvm::StringRef> ids;
   ids.reserve(number);
 
-  // First, compute how many identifiers of each size we'll need.
+  // First, compute the total weight of the distribution so we know how many
+  // identifiers we'll get each time we collect from it.
   int count_sum =
       uniform ? (max_length - min_length) + 1
               : Sum(llvm::ArrayRef(IdLengthCounts)
                         .slice(min_length - 1, max_length - min_length + 1));
   CARBON_CHECK(count_sum >= 1);
+
+  // Now compute how many times we'll have to sample across the distribution.
   int number_rem = number % count_sum;
+
+  // Finally, walk through each length in the distribution.
   for (int length : llvm::seq_inclusive(min_length, max_length)) {
-    // Scale this length if non-uniform.
+    // Scale how many identifiers we want of this length if computing a
+    // non-uniform distribution. For uniform, we always take one.
     int scale = uniform ? 1 : IdLengthCounts[length - 1];
+
+    // Now we can compute how many identifiers of this length to request.
     int length_count = (number * scale) / count_sum;
     if (number_rem > 0 && length_count * count_sum < number * scale) {
       ++length_count;
@@ -247,185 +429,6 @@ auto SourceGen::GetIdsImpl(int number, int min_length, int max_length,
   CARBON_CHECK(static_cast<int>(ids.size()) == number);
 
   return ids;
-}
-
-auto SourceGen::GetIds(int number, int min_length, int max_length, bool uniform)
-    -> llvm::SmallVector<llvm::StringRef> {
-  llvm::SmallVector<llvm::StringRef> ids =
-      GetIdsImpl(number, min_length, max_length, uniform,
-                 [this](int length, int length_count,
-                        llvm::SmallVectorImpl<llvm::StringRef>& dest) {
-                   auto length_ids = GetSingleLengthIds(length, length_count);
-                   dest.append(length_ids.begin(), length_ids.end());
-                 });
-
-  return ids;
-}
-
-auto SourceGen::GetUniqueIds(int number, int min_length, int max_length,
-                             bool uniform)
-    -> llvm::SmallVector<llvm::StringRef> {
-  CARBON_CHECK(min_length >= 4)
-      << "Cannot trivially guarantee enough distinct, unique identifiers for "
-         "lengths <= 3";
-  llvm::SmallVector<llvm::StringRef> ids =
-      GetIdsImpl(number, min_length, max_length, uniform,
-                 [this](int length, int length_count,
-                        llvm::SmallVectorImpl<llvm::StringRef>& dest) {
-                   AppendUniqueIdentifiers(length, length_count, dest);
-                 });
-
-  return ids;
-}
-
-auto SourceGen::GetSingleLengthIds(int length, int number)
-    -> llvm::ArrayRef<llvm::StringRef> {
-  llvm::SmallVector<llvm::StringRef>& ids =
-      ids_by_length.Insert(length, {}).value();
-
-  if (static_cast<int>(ids.size()) < number) {
-    ids.reserve(number);
-    for (int _ : llvm::seq<int>(ids.size(), number)) {
-      char* id_storage = reinterpret_cast<char*>(
-          storage.Allocate(/*Size=*/length, /*Alignment=*/1));
-      std::string new_id_tmp = GenerateRandomIdentifier(length);
-      memcpy(id_storage, new_id_tmp.data(), length);
-      llvm::StringRef new_id(id_storage, length);
-      ids.push_back(new_id);
-    }
-    CARBON_CHECK(static_cast<int>(ids.size()) == number);
-  }
-  return llvm::ArrayRef(ids).slice(0, number);
-}
-
-static auto IdentifierStartChars() -> llvm::ArrayRef<char> {
-  static llvm::SmallVector<char> chars = [] {
-    llvm::SmallVector<char> chars;
-    for (char c : llvm::seq_inclusive('A', 'Z')) {
-      chars.push_back(c);
-    }
-    for (char c : llvm::seq_inclusive('a', 'z')) {
-      chars.push_back(c);
-    }
-    return chars;
-  }();
-  return chars;
-}
-
-static auto IdentifierChars() -> llvm::ArrayRef<char> {
-  static llvm::SmallVector<char> chars = [] {
-    llvm::ArrayRef<char> start_chars = IdentifierStartChars();
-    llvm::SmallVector<char> chars(start_chars.begin(), start_chars.end());
-    chars.push_back('_');
-    for (char c : llvm::seq_inclusive('0', '9')) {
-      chars.push_back(c);
-    }
-    return chars;
-  }();
-  return chars;
-}
-
-constexpr static llvm::StringRef NonCarbonCppKeywords[] = {
-    "asm", "do",     "double", "float", "int",      "long",
-    "new", "signed", "try",    "unix",  "unsigned", "xor",
-};
-
-// Returns a random identifier string of the specified length.
-//
-// Ensures this is a valid identifier, avoiding any overlapping syntaxes or
-// keywords both in Carbon and C++.
-//
-// This routine is somewhat expensive and so is useful to cache and reduce the
-// frequency of calls. However, each time it is called it computes a completely
-// new random identifier and so can be useful to eventually find a distinct
-// identifier when needed.
-auto SourceGen::GenerateRandomIdentifier(int length) -> std::string {
-  llvm::ArrayRef<char> start_chars = IdentifierStartChars();
-  llvm::ArrayRef<char> chars = IdentifierChars();
-
-  std::string id_result;
-  llvm::raw_string_ostream os(id_result);
-  llvm::StringRef id;
-  do {
-    // Erase any prior attempts to find an identifier.
-    id_result.clear();
-    os << start_chars[absl::Uniform<int>(rng, 0, start_chars.size())];
-    for (int j : llvm::seq(1, length)) {
-      static_cast<void>(j);
-      os << chars[absl::Uniform<int>(rng, 0, chars.size())];
-    }
-    // Check if we ended up forming an integer type literal or a keyword, and
-    // try again.
-    id = llvm::StringRef(id_result);
-  } while (
-      // TODO: Clean up and simplify this code. With some small refactorings and
-      // post-processing we should be able to make this both easier to read and
-      // less inefficient.
-      llvm::any_of(Lex::TokenKind::KeywordTokens,
-                   [id](auto token) { return id == token.fixed_spelling(); }) ||
-      llvm::is_contained(NonCarbonCppKeywords, id) ||
-      (llvm::is_contained({'i', 'u', 'f'}, id[0]) &&
-       llvm::all_of(id.substr(1),
-                    [](const char c) { return llvm::isDigit(c); })));
-  CARBON_CHECK(id == id_result);
-  CARBON_CHECK(static_cast<int>(id.size()) == length);
-  return id_result;
-}
-
-// Appends a number of unique, random identifiers with a particular length to
-// the provided destination vector.
-//
-// Uses, and when necessary grows, a cached sequence of random identifiers with
-// the specified length. Because these are cached, this is efficient to call
-// repeatedly, but will not produce a different sequence of identifiers.
-auto SourceGen::AppendUniqueIdentifiers(
-    int length, int number, llvm::SmallVectorImpl<llvm::StringRef>& dest)
-    -> void {
-  auto& [count, unique_ids] = unique_ids_by_length.Insert(length, {}).value();
-
-  // See if we need to grow our pool of unique identifiers with the requested
-  // length.
-  if (count < number) {
-    // We'll need to insert exactly the requested new unique identifiers. All
-    // our other inserts will find an existing entry.
-    unique_ids.GrowForInsertCount(count - number);
-
-    // Generate the needed number of identifiers.
-    for ([[maybe_unused]] int i : llvm::seq<int>(count, number)) {
-      // Allocate stable storage for the identifier so we can form stable
-      // `StringRef`s to it.
-      char* id_storage = reinterpret_cast<char*>(
-          storage.Allocate(/*Size=*/length, /*Alignment=*/1));
-      // Repeatedly generate novel identifiers of this length until we find a
-      // new unique one.
-      for (;;) {
-        std::string new_id_tmp = GenerateRandomIdentifier(length);
-        // Copy the new identifier into our stable storage before trying to
-        // insert it into the set so the inserted string ref is stable.
-        memcpy(id_storage, new_id_tmp.data(), length);
-        llvm::StringRef new_id(id_storage, length);
-        auto result = unique_ids.Insert(new_id);
-        if (result.is_inserted()) {
-          break;
-        }
-      }
-    }
-    count = number;
-  }
-  // Append all the identifiers directly out of the set. We make no guarantees
-  // about the relative order so we just use the non-deterministic order of the
-  // set and avoid additional storage.
-  //
-  // TODO: It's awkward the `ForEach` here can't early-exit. This just walks the
-  // whole set which is harmless if inefficient. We should add early exiting
-  // the loop support to `Set` and update this code.
-  unique_ids.ForEach([&](llvm::StringRef id) {
-    if (number > 0) {
-      dest.push_back(id);
-      --number;
-    }
-  });
-  CARBON_CHECK(number == 0);
 }
 
 // Returns a shuffled sequence of integers in the range [min, max].
@@ -445,7 +448,7 @@ auto SourceGen::GetShuffledInts(int number, int min, int max)
   }
   CARBON_CHECK(static_cast<int>(ints.size()) == number);
 
-  std::shuffle(ints.begin(), ints.end(), rng);
+  std::shuffle(ints.begin(), ints.end(), rng_);
   return ints;
 }
 
@@ -534,17 +537,19 @@ class SourceGen::UniqueIdPopper {
       return data_->pop_back_val();
     }
 
-    // Out of unique elements. Pop the back and use its length to generate new
-    // identifiers until we find a unique one and return that. This ensures we
-    // continue to consume the structure and produce the same size identifiers
-    // even in the fallback. Note that fallback identifiers only live to the end
-    // of the popper.
+    // Out of unique elements. Overwrite the back, preserving its length,
+    // generating a new identifiers until we find a unique one and return that.
+    // This ensures we continue to consume the structure and produce the same
+    // size identifiers even in the fallback.
     int length = data_->pop_back_val().size();
-    fallback_ids_.push_back("");
-    std::string& fallback_id = fallback_ids_.back();
+    auto fallback_id_storage =
+        llvm::MutableArrayRef(reinterpret_cast<char*>(gen_->storage_.Allocate(
+                                  /*Size=*/length, /*Alignment=*/1)),
+                              length);
     for (;;) {
-      fallback_id = gen_->GenerateRandomIdentifier(length);
-      if (set_.Insert(llvm::StringRef(fallback_id)).is_inserted()) {
+      gen_->GenerateRandomIdentifier(fallback_id_storage);
+      auto fallback_id = llvm::StringRef(fallback_id_storage.data(), length);
+      if (set_.Insert(fallback_id).is_inserted()) {
         return fallback_id;
       }
     }
@@ -554,7 +559,6 @@ class SourceGen::UniqueIdPopper {
   SourceGen* gen_;
   llvm::SmallVectorImpl<llvm::StringRef>* data_;
   llvm::SmallVectorImpl<llvm::StringRef>::reverse_iterator it_;
-  llvm::SmallVector<std::string> fallback_ids_;
   Set<llvm::StringRef> set_;
 };
 
--- a/testing/base/source_gen.cpp
+++ b/testing/base/source_gen.cpp
@@ -22,7 +22,11 @@ auto SourceGen::Global() -> SourceGen& {
 
 SourceGen::SourceGen(Language language) : language_(language) {}
 
-// Some heuristic numbers used when formatting generated code.
+// Some heuristic numbers used when formatting generated code. These heuristics
+// are loosely based on what we expect to make Carbon code readable, and might
+// not fit as well in C++, but we use the same heuristics across languages for
+// simplicity and to make the output in different languages more directly
+// comparable.
 constexpr static int NumSingleLineFunctionParams = 3;
 constexpr static int NumSingleLineMethodParams = 2;
 constexpr static int MaxParamsPerLine = 4;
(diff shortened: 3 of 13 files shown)


toolchain/lex/tokenized_buffer_benchmark.cpp
```diff
@@ -26,180 +27,14 @@ namespace {
...
-  // Shuffle tokens so we get exactly one of each identifier but in a random
-  // order.
-  std::array<llvm::StringRef, NumTokens> tokens;
-  for (int i : llvm::seq(NumTokens)) {
-    tokens[i] = ids[i];
-  }
-  std::shuffle(tokens.begin(), tokens.end(), absl::BitGen());
-  return llvm::join(tokens, separator);
+auto RandomIdentifierSeq(int min_length = 1, int max_length = 64,
+                         bool uniform = false, llvm::StringRef separator = " ")
```
jonmeow: How are these default parameters used? I'm only seeing one call site eliding `separator`.

Also, should methods like this be `static`?

toolchain/driver/compile_benchmark_test.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+#
+# Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+# Exceptions. See /LICENSE for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+BENCHMARK="$TEST_SRCDIR/$TEST_WORKSPACE/toolchain/driver/compile_benchmark"
```
jonmeow, line 7: Can you add comments explaining why this script exists? It seems like a really thin wrapper around the benchmark, why wouldn't it just be a usage comment?

toolchain/driver/compile_benchmark.cpp
```diff
@@ -0,0 +1,124 @@
...
+#include <string>
+
+#include "common/benchmark_main.h"
+#include "testing/base/source_gen.h"
+#include "toolchain/driver/driver.h"
+
+namespace Carbon::Testing {
+namespace {
+
+constexpr ssize_t NumFiles = 20;
```
jonmeow: Please add comments to entities in this file -- I think it'd be helpful to figuring out what this file is intended for.

testing/base/source_gen_test.cpp
```diff
@@ -0,0 +1,173 @@
...
+using ::testing::ContainerEq;
+using ::testing::Contains;
+using ::testing::Eq;
+using ::testing::Ge;
+using ::testing::Le;
+using ::testing::MatchesRegex;
+using ::testing::SizeIs;
+
+template <typename T>
+static auto SumSizes(const T& range) -> ssize_t {
```
jonmeow, line 31: Can you add a comment explaining what this is doing? It looks like T is always the same type, why the template vs either `llvm::SmallVector<llvm::StringRef>` or `llvm::ArrayRef`?

testing/base/source_gen_test.cpp
```diff
@@ -0,0 +1,173 @@
...
+  return std::accumulate(
+      range.begin(), range.end(), static_cast<ssize_t>(0),
+      [](ssize_t lhs, const auto& rhs) -> ssize_t { return lhs + rhs.size(); });
```
jonmeow: So I'm trying to figure out what this is doing. Had you considered things like:

```
int sum = 0;
for (auto x : range) {
  sum += x.size();
}
return sum;
```

Or if you prefer a lambda:

```
int sum = 0;
llvm::for_each(range, [&](llvm::StringRef s) { sum += s.size(); });
return sum;
```

(note I'm suggesting int instead of ssize_t too, per https://google.github.io/styleguide/cppguide.html#Integer_Types)

testing/base/source_gen_test.cpp
```diff
@@ -0,0 +1,173 @@
...
+    auto unique2 = gen.GetShuffledUniqueIds(1000);
+    EXPECT_THAT(unique2, SizeIs(1000));
+    // Should be (at least) a different shuffle of identifiers.
+    EXPECT_THAT(unique2, Not(ContainerEq(unique)));
+    // But the sum of lengths should be identical.
+    EXPECT_THAT(SumSizes(unique2), Eq(unique_size_sum));
+  }
+
+  // Check length constraints have the desired effect.
+  unique = gen.GetShuffledUniqueIds(1000, /*min_length=*/10, /*max_length=*/20);
```
jonmeow: It's good you're marking the later parameters, but what's the first parameter?

testing/base/source_gen_test.cpp
```diff
@@ -0,0 +1,173 @@
...
+  Set<llvm::StringRef> set;
+  for (llvm::StringRef id : unique) {
+    EXPECT_THAT(id, MatchesRegex("[A-Za-z][A-Za-z0-9_]*"));
+    EXPECT_TRUE(set.Insert(id).is_inserted()) << "Colliding id: " << id;
+  }
+
+  // Check that repeated calls are different in interesting ways, but have the
+  // exact same total bytes.
+  ssize_t unique_size_sum = SumSizes(unique);
+  for ([[maybe_unused]] int i : llvm::seq(1, 10)) {
```
jonmeow: I think it'd be clearer about the number of loops if you don't pass a lower bound:

```suggestion
  for ([[maybe_unused]] int _ : llvm::seq(9)) {
```

testing/base/source_gen_test.cpp
```diff
@@ -0,0 +1,173 @@
...
+    // But the sum of lengths should be identical.
+    EXPECT_THAT(SumSizes(unique2), Eq(unique_size_sum));
+  }
+
+  // Check length constraints have the desired effect.
+  unique = gen.GetShuffledUniqueIds(1000, /*min_length=*/10, /*max_length=*/20);
+  for (llvm::StringRef id : unique) {
+    EXPECT_THAT(id.size(), Ge(10)) << "Too short id: " << id;
+    EXPECT_THAT(id.size(), Le(20)) << "Too long id: " << id;
+  }
```
jonmeow: Had you considered:

```suggestion
  EXPECT_THAT(unique, Each(SizeIs(AllOf(Ge(10), Le(20)))));
```

Or if you prefer the loop for id printing:

```suggestion
  for (llvm::StringRef id : unique) {
    EXPECT_THAT(id, SizeIs(AllOf(Ge(10), Le(20))));
  }
```

testing/base/source_gen_test.cpp
```diff
@@ -0,0 +1,173 @@
...
+  EXPECT_THAT(unique, Contains(SizeIs(13)).Times(10));
+  EXPECT_THAT(unique, Contains(SizeIs(14)).Times(10));
+  EXPECT_THAT(unique, Contains(SizeIs(15)).Times(10));
+  EXPECT_THAT(unique, Contains(SizeIs(16)).Times(10));
+  EXPECT_THAT(unique, Contains(SizeIs(17)).Times(10));
+  EXPECT_THAT(unique, Contains(SizeIs(18)).Times(10));
+  EXPECT_THAT(unique, Contains(SizeIs(19)).Times(10));
+}
+
+TEST(SourceGenTest, Ids) {
```
jonmeow: This looks very similar to the Unique test, with the one exception being the uniqueness check on line 41. Had you considered a helper function that does the common logic, just taking in something like `[&](SourceGen& gen, int n, int min_length, int max_length, bool uniform) { return gen.ShuffledIds(n, min_length, max_length, uniform);}` ?

Note my comments to the above function apply here too.

testing/base/source_gen_test.cpp
```diff
@@ -0,0 +1,173 @@
...
+  EXPECT_THAT(ids, Contains(SizeIs(13)).Times(10));
+  EXPECT_THAT(ids, Contains(SizeIs(14)).Times(10));
+  EXPECT_THAT(ids, Contains(SizeIs(15)).Times(10));
+  EXPECT_THAT(ids, Contains(SizeIs(16)).Times(10));
+  EXPECT_THAT(ids, Contains(SizeIs(17)).Times(10));
+  EXPECT_THAT(ids, Contains(SizeIs(18)).Times(10));
+  EXPECT_THAT(ids, Contains(SizeIs(19)).Times(10));
+}
+
+auto TestCompile(llvm::StringRef source) -> bool {
```
jonmeow, line 144: Would it make sense to add comments explaining what this is doing? Also, maybe it should be something like `TestCheck` since it's not a full compile? Or, just let it do the full compile?

testing/base/source_gen_test.cpp
```diff
@@ -0,0 +1,173 @@
...
+  fs.addFile("test.carbon", /*ModificationTime=*/0,
+             llvm::MemoryBuffer::getMemBuffer(source));
+  return driver.RunCommand({"compile", "--phase=check", "test.carbon"}).success;
+}
+
+TEST(SourceGenTest, GenAPIFileDenseDeclsTest) {
+  SourceGen gen;
+
+  std::string source =
+      gen.GenAPIFileDenseDecls(1000, SourceGen::DenseDeclParams{});
```
jonmeow, line 173: What do you think of commenting the 1000 parameter here, to give a hint about what it is?

testing/base/source_gen.h
```diff
@@ -0,0 +1,192 @@
...
+  auto GenAPIFileDenseDecls(int target_lines, DenseDeclParams params)
+      -> std::string;
+
+  auto GetShuffledIds(int number, int min_length = 1, int max_length = 64,
+                      bool uniform = false)
+      -> llvm::SmallVector<llvm::StringRef>;
+
+  auto GetShuffledUniqueIds(int number, int min_length = 4, int max_length = 64,
+                            bool uniform = false)
+      -> llvm::SmallVector<llvm::StringRef>;
```
jonmeow: Should these have some comments explaining what they do, and how these constants were chosen?

jonmeow (commented):
What do you think about adding more comments to this code? I'm seeing a lot that's not commented, and I think it'd be easier to review if there were more explanation of what the entities are.

chandlerc:
> What do you think about adding more comments to this code? I'm seeing a lot that's not commented, and I think it'd be easier to review if there were more explanation of what the entities are.

I just forgot to go back to that item in my own TODOs before sending this out for review, mostly in trying to get it out sooner. Sorry about that. I'll make a pass through.

testing/base/source_gen_test.cpp
chandlerc: Yeah, they're pretty similar.

That said, testing with a lambda for the actual API under test also is much less readable to me.

What about just dropping the duplication and only testing uniqueness in the unique case? The implementation is common anyways so this isn't really changing the coverage either way and seems much more readable. PTAL.

testing/base/source_gen_test.cpp
chandlerc: The first parameter is the number of IDs, which seemed pretty clear and isn't an optional parameter -- we even use it without any comment above. Adding a comment here seemed redundant.

testing/base/source_gen_test.cpp
chandlerc: Dropped the lower bound, that was a mistake anyways.

But not using `_` intentionally as GoogleTest also uses an identifier with that name and it seems too clever to use it in loops instead of just marking the loop variable as unused....

testing/base/source_gen_test.cpp
chandlerc: Because we're summing potentially many sizes, it seems important to not use `int` here. It could actually overflow, especially if the code we're testing is buggy, so I'd rather use a type that behaves more predictably. And because there is arithmetic, having the type signed seemed valuable.

I don't feel super strongly about using an explicit loop rather than an algorithm, most folks seem to prefer algorithms these days. And somewhat sadly, this appears to be C++'s "idiom" for an algorithm that does sum. I'm super tempted to create a Carbon algorithms helper library, but trying to resist... Anyways, used a for loop here.

testing/base/source_gen_test.cpp
chandlerc: Added a comment. Spelling out the types here seems like a distraction, specifically because it's awkward to avoid naming the *inner* range when using `llvm::ArrayRef` for the outer type, and the inner type could be a `llvm::StringRef` or `std::string` if we change our implementation, etc.

testing/base/source_gen.h
chandlerc: Sorry, I just forgot to add comments here, it was on my TODO before mailing. Added now.

toolchain/lex/tokenized_buffer_benchmark.cpp
chandlerc: I was just minimizing the diff from the existing code which didn't use `static` and had these defaulted as template parameters.

I've made these changes, but generally trying to not go *too* far in cleaning up this benchmark in this PR. I think the code here can use more cleanups, but just avoiding scope creep.

toolchain/driver/compile_benchmark_test.sh
chandlerc: This allows us to run the benchmarks as a test to ensure the. Without this `bazel test ...` won't run the benchmark and catch debug checks and such. I added these when adding the hashtable benchmarks as I found it very useful to avoid missing failures in one benchmark.

I've added a brief comment, but not sure how much else to really say here as it isn't clear to me what exactly isn't clear about the setup....

Also, maybe there is a cleaner way to do this, happy to shift if so (although maybe in a follow-up to do that for the hashtable benchmarks too).

testing/base/source_gen_test.cpp
chandlerc: Added a comment.

This is just checking "does the source compile", and so the implementation just uses `--phase=check` to be minimal about that. But that's somewhat an implementation detail, so didn't want to surface the phase name in the function name.

WDYT?

testing/base/source_gen_test.cpp
chandlerc: :shrug: sure.

chandlerc (commented):
I think I got to ever comment, but let me know if I missed anything.

common/benchmark_main.h
```diff
@@ -0,0 +1,22 @@
...
+#include "llvm/ADT/StringRef.h"
+
+// When using the Carbon `main` function for benchmarks, we export some extra
+// information about the test binary that can be accessed with this header.
+//
+// TODO: Refactor this to share code with `gtest_main.h`.
+
+namespace Carbon::Testing {
+
+// The executable path of the benchmark binary.
```
jonmeow:
```suggestion
// Returns the executable path of the benchmark binary.
```

[Style](https://google.github.io/styleguide/cppguide.html#Function_Comments):
> Function comments should be written with an implied subject of This function and should start with the verb phrase; for example, "Opens the file", rather than "Open the file".

testing/base/source_gen.cpp
```diff
@@ -0,0 +1,571 @@
...
+  state.member_names = GetShuffledIds(num_members, /*min_length=*/4);
+  int num_params = Sum(state.public_function_param_counts) +
+                   Sum(state.public_method_param_counts) +
+                   Sum(state.private_function_param_counts) +
+                   Sum(state.private_method_param_counts);
+  state.param_names = GetShuffledIds(num_params);
+  return state;
+}
+
+class SourceGen::UniqueIdPopper {
```
jonmeow: This class doesn't have comments in the header, nor on its members.

jonmeow (commented):
Can you take some time to walk through the source_gen files and add comments? While I see public members in source_gen.h have comments, none of the private members do (including public members of the private UniqueIdPopper), and most static/constexpr entities in the cpp file don't. Not sure if you missed this because I didn't comment on the file specifically.

common/benchmark_main.h
chandlerc: Sure, I just copied the comment in `gtest_main.h`...

testing/base/source_gen.cpp
chandlerc: Added, sorry about that.

chandlerc:
> Can you take some time to walk through the source_gen files and add comments? While I see public members in source_gen.h have comments, none of the private members do (including public members of the private UniqueIdPopper), and most static/constexpr entities in the cpp file don't. Not sure if you missed this because I didn't comment on the file specifically.

No, I think I just only looked at a few places rather than the whole file, sorry about that.

Some of the functions don't seem to need much in the way of comments, the function name and param names would just be repeated in a sentence I think... But a bunch definitely needed comments. There was a bunch of subtle stuff, etc... Anyways, everything that jumped out at me as needing a comment has one now I think, and I'm much more sure I *actually* went all the way through the file rather than just glancing at one area in the file and forgetting about the rest. If there are still things that would really benefit from a comment, it may be useful to understand a bit better what needs clarification at this point as I *hopefully* didn't miss anything for a third time.

I also spotted a few places where one comment you made applies more broadly and tried to update based on it, and cleaned up a few things that adding comments made me think better about.

common/benchmark_main.cpp
```diff
@@ -2,17 +2,40 @@
...
+#include <string>
+
 #include "absl/flags/parse.h"
+#include "common/check.h"
+#include "common/exe_path.h"
 #include "common/init_llvm.h"
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/StringRef.h"
 
+static bool after_main = false;
```
zygoloid, line 18: I found this name confusing -- I expected it to indicated whether `main` had *completed*. Maybe `entered_main`?

common/benchmark_main.cpp
```diff
@@ -2,17 +2,40 @@
...
+static llvm::StringRef exe_path;
+
+namespace Carbon::Testing {
+
+auto GetBenchmarkExePath() -> llvm::StringRef {
+  CARBON_CHECK(after_main)
+      << "Must not query the executable path until after `main` is entered!";
+  return exe_path;
```
zygoloid, line 26: Hm, do we need `after_main` at all? Would it be correct to instead `CHECK` that `!exe_path.is_empty()` here, or do we need to cope with `FindExecutablePath` producing an empty string? (If we do need it, an `optional<StringRef>` would seem clearer to me.)

common/benchmark_main.cpp
```diff
@@ -2,17 +2,40 @@
...
+
+}  // namespace Carbon::Testing
+
 auto main(int orig_argc, char** orig_argv) -> int {
   // Do LLVM's initialization first, this will also transform UTF-16 to UTF-8.
   Carbon::InitLLVM init_llvm(orig_argc, orig_argv);
 
+  std::string exe_path_storage = Carbon::FindExecutablePath(orig_argv[0]);
+  exe_path = exe_path_storage;
+  after_main = true;
```
zygoloid, line 38: Looks like we could also have problems if `GetBenchmarkExePath` is called after `main` returns. If we need to defend against calls before `main`, do we also need to defend against calls after `main`?

testing/base/source_gen.cpp
```diff
@@ -0,0 +1,676 @@
...
+auto SourceGen::Global() -> SourceGen& {
+  static SourceGen global_gen;
+  return global_gen;
+}
+
+SourceGen::SourceGen(Language language) : language_(language) {}
+
+// Some heuristic numbers used when formatting generated code.
+constexpr static int NumSingleLineFunctionParams = 3;
+constexpr static int NumSingleLineMethodParams = 2;
```
zygoloid, line 31: I'd expect this to depend on the language we're generating -- static member functions would be longer in C++ due to the `static `; methods would be longer in Carbon due to the `[self: Self]`. Or are we trying to generally get the line layout to match between languages?

testing/base/source_gen.cpp
```diff
@@ -0,0 +1,676 @@
...
+    }
+  }
+  return 1.0 + static_cast<double>(param_lines) / (params.max_params + 1);
+}
+
+// Note that this should match the heuristics used when formatting.
+// TODO: See top-level TODO about line estimates and formatting.
+static auto EstimateAvgClassLines(SourceGen::ClassParams params) -> double {
+  // Blank line, comment line, and class open line.
+  double avg = 3.0;
```
zygoloid: It seems inconsistent to include a blank line and comment in the value returned by this function but not the values returned by the other two. A comment would be useful to explain which things are covered.

It might be nice if the number returned by `EstimateAvgFooLines` was the expected number of lines added by `GenerateFoo` below -- so include the comment but not the blank line. (For that to hold, we'd also want to rename this function to `EstimateAvgClassDefLines`.)

testing/base/source_gen.cpp
```diff
@@ -0,0 +1,676 @@
...
+  return avg;
+}
+
+auto SourceGen::GenAPIFileDenseDecls(int target_lines, DenseDeclParams params)
+    -> std::string {
+  std::string source;
+  llvm::raw_string_ostream os(source);
+
+  double avg_class_lines = EstimateAvgClassLines(params.class_params);
+  int num_classes = static_cast<double>(target_lines) / avg_class_lines;
```
zygoloid: I think this should be
```suggestion
  CARBON_CHECK(target_lines >= 4);
  int num_classes = static_cast<double>(target_lines - 4) / avg_class_lines;
```
to account for the five lines of comment below, and the fact that one of the classes doesn't have its own blank line.

testing/base/source_gen.cpp
```diff
@@ -0,0 +1,676 @@
...
+                      target_lines, num_classes, expected_lines)
+     << "\n";
+  os << "//\n// Generating as an API file with dense declarations.\n\n";
+
+  auto class_gen_state = GetClassGenState(num_classes, params.class_params);
+  llvm::ListSeparator line_sep("\n");
+  for ([[maybe_unused]] int i : llvm::seq(num_classes)) {
+    os << line_sep;
+    GenerateClassDef(params.class_params, class_gen_state, os);
+  }
```
zygoloid:
```suggestion
  os << "//\n// Generating as an API file with dense declarations.\n";

  auto class_gen_state = GetClassGenState(num_classes, params.class_params);
  for ([[maybe_unused]] int i : llvm::seq(num_classes)) {
    os << '\n';
    GenerateClassDef(params.class_params, class_gen_state, os);
  }
```
Optional, but I find this a bit simpler. (It also better handles `num_classes == 0` but I think we probably don't care.)

testing/base/source_gen.cpp
```diff
@@ -0,0 +1,676 @@
...
+
+// An array of the counts that should be used for each identifier length to
+// produce our desired distribution.
+//
+// Note that the zero-based index corresponds to a 1-based length, so the count
+// for identifiers of length 1 is at index 0.
+static constexpr std::array<int, 64> IdLengthCounts = [] {
+  std::array<int, 64> id_length_counts;
+  // For non-uniform distribution, we simulate a distribution roughly based on
+  // the observed histogram of identifier lengths, but smoothed a bit and
```
zygoloid: The dip at 5 and bump at 12 have been removed entirely by this smoothing, and 10/11 look a bit over-represented. (In the data, 9 and 12 are nearly 2x as common as 10 and 11 but all four have the same probability here.) Do you think those are just measurement errors / peculiarities of LLVM, or should we be modeling them? How important is it to get this right-ish?

testing/base/source_gen.cpp
```diff
@@ -0,0 +1,676 @@
...
+// `GetUniqueIds`. Most parameters correspond to the parameters of those
+// functions. Additionally, an `AppendIds` callable is provided to implement the
+// appending operation.
+//
+// The main functionality provided here is collecting the correct number of
+// identifiers from each of the lengths in the range [min_length, max_length]
+// and either in our default representative distribution or a uniform
+// distribution.
+//
+// Note that this template must be defined prior to its use below.
```
zygoloid: It actually doesn't need to be defined first; defining it in the same file is sufficient so long as it's not `constexpr` and doesn't have a deduced return type.

testing/base/source_gen.cpp
```diff
@@ -0,0 +1,676 @@
...
+  CARBON_CHECK(count_sum >= 1);
+  int number_rem = number % count_sum;
+  for (int length : llvm::seq_inclusive(min_length, max_length)) {
+    // Scale this length if non-uniform.
+    int scale = uniform ? 1 : IdLengthCounts[length - 1];
+    int length_count = (number / count_sum) * scale;
+    if (number_rem > 0) {
+      length_count += std::min(scale, number_rem);
+      number_rem -= scale;
+    }
```
zygoloid: For small counts, this'll overallocate towards the bottom end. Eg, if I ask for 100 identifiers with lengths 4 to 64, I'll get {50, 20, 20, 10, 0, 0, 0, ...}. How about:
```suggestion
    int length_count = (number * scale) / count_sum;
    if (number_rem > 0 && length_count * count_sum < number * scale) {
      ++length_count;
      --number_rem;
    }
```
That is: get the floor of the ideal count (which will be off by less than 1) and then add one to the first `number_rem` buckets that we rounded down.

For my example, I think that'd give something more like {21, 9, 9, 9, 8, 5, 5, 5, 5, 3, 2, <19 1s>}, which seems a bit more representative. It still skews a little towards shorter identifiers, but that's probably not that unreasonable.

testing/base/source_gen.cpp
```diff
@@ -0,0 +1,676 @@
...
+  CARBON_CHECK(min_length <= max_length);
+  CARBON_CHECK(uniform || max_length <= 64)
+      << "Cannot produce a meaningful non-uniform distribution of lengths "
+         "longer than 64 as those are exceedingly rare in our observed data "
+         "sets.";
+
+  llvm::SmallVector<llvm::StringRef> ids;
+  ids.reserve(number);
+
+  // First, compute how many identifiers of each size we'll need.
```
zygoloid: I found this comment confusing. Describing `count_sum` as something more like the total weight of the distribution would have been more useful to me.

testing/base/source_gen.cpp
```diff
@@ -0,0 +1,676 @@
...
+}
+
+auto SourceGen::GetSingleLengthIds(int length, int number)
+    -> llvm::ArrayRef<llvm::StringRef> {
+  llvm::SmallVector<llvm::StringRef>& ids =
+      ids_by_length.Insert(length, {}).value();
+
+  if (static_cast<int>(ids.size()) < number) {
+    ids.reserve(number);
+    for ([[maybe_unused]] int i : llvm::seq<int>(ids.size(), number)) {
```
zygoloid:
```suggestion
    for (int _ : llvm::seq<int>(ids.size(), number)) {
```
Try C++23 Today! :)

(Clang suppresses the `-Wunused` warning for variables named `_` in all language modes in preparation for C++23's special variables-named-`_` feature.)

There are a few more instances elsewhere in the PR if you're OK with pre-adopting this.

testing/base/source_gen.cpp
```diff
@@ -0,0 +1,676 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "testing/base/source_gen.h"
+
+#include <numeric>
```
zygoloid, line 7:
```suggestion
#include <cstring>
#include <numeric>
```

testing/base/source_gen.cpp
```diff
@@ -0,0 +1,676 @@
...
+  llvm::SmallVector<llvm::StringRef>& ids =
+      ids_by_length.Insert(length, {}).value();
+
+  if (static_cast<int>(ids.size()) < number) {
+    ids.reserve(number);
+    for ([[maybe_unused]] int i : llvm::seq<int>(ids.size(), number)) {
+      char* id_storage = reinterpret_cast<char*>(
+          storage.Allocate(/*Size=*/length, /*Alignment=*/1));
+      std::string new_id_tmp = GenerateRandomIdentifier(length);
+      memcpy(id_storage, new_id_tmp.data(), length);
```
zygoloid:
```suggestion
      std::memcpy(id_storage, new_id_tmp.data(), length);
```

testing/base/source_gen.cpp
```diff
@@ -0,0 +1,676 @@
...
+// Returns a random identifier string of the specified length.
+//
+// Ensures this is a valid identifier, avoiding any overlapping syntaxes or
+// keywords both in Carbon and C++.
+//
+// This routine is somewhat expensive and so is useful to cache and reduce the
+// frequency of calls. However, each time it is called it computes a completely
+// new random identifier and so can be useful to eventually find a distinct
+// identifier when needed.
+auto SourceGen::GenerateRandomIdentifier(int length) -> std::string {
```
zygoloid: Would it make sense to make this instead take a `MutableArrayRef<char>` of the desired length and fill it in? Most callers allocate separate storage and copy the result at the moment, so there is unnecessary allocation and copying happening here.

testing/base/source_gen.cpp
```diff
@@ -0,0 +1,676 @@
...
+      dest.push_back(id);
+      --number;
+    }
+  });
+  CARBON_CHECK(number == 0);
+}
+
+// Returns a shuffled sequence of integers in the range [min, max].
+//
+// These order of the returned integers is random, but the distribution is
```
zygoloid:
```suggestion
// The order of the returned integers is random, but the distribution is
```

testing/base/source_gen.cpp
```diff
@@ -0,0 +1,676 @@
...
+// computed by appending numbers in round robin over the range up to the
+// requested number.
```
zygoloid: I didn't understand what this comment meant until I read the algorithm. Maybe we could rephrase, eg something like:
```suggestion
// The order of the returned integers is random, but each integer in the range appears the same number of times in the result, with the number of appearances rounded up for lower numbers and rounded down for higher numbers in order to exactly produce `number` results.
```

testing/base/source_gen.cpp
```diff
@@ -0,0 +1,676 @@
...
+  state.member_names = GetShuffledIds(num_members, /*min_length=*/4);
+  int num_params = Sum(state.public_function_param_counts) +
+                   Sum(state.public_method_param_counts) +
+                   Sum(state.private_function_param_counts) +
+                   Sum(state.private_method_param_counts);
+  state.param_names = GetShuffledIds(num_params);
+  return state;
+}
+
+// A helper to pop series of unique IDs off a sequence of random IDs that may
```
zygoloid: I think you're using "IDs" here to mean "identifiers". I found that confusing given that it's not an abbreviation we use elsewhere and we use `Id` elsewhere to mean a numeric ID. I think using "identifier" in comments and either `Ident` or `Identifier` in code would help. (You already have some `Identifier` usage here in addition to the `Id` usage.)

testing/base/source_gen.h
```diff
@@ -0,0 +1,227 @@
...
+// set of long-term goals:
+//
+// - Generate interesting patterns and structures of code that have emerged as
+//   toolchain performance bottlenecks in practice in C++ codebases.
+// - Generate code that includes most Carbon language features (and whatever
+//   reasonable C++ analogs could be used for comparative purposes):
+//   - Functions
+//   - Classes with class functions, methods, and fields
+//   - Interfaces
+//   - Nested and unnested impls
```
zygoloid, line 33:
```suggestion
//   - Checked generics and templates
//   - Nested and unnested impls
```
maybe?

testing/base/source_gen.h
```diff
@@ -0,0 +1,227 @@
...
+    // other than class definitions.
+
+    // Parameters used when generating class definitions.
+    ClassParams class_params = {};
+  };
+
+  // Access a global instance of this type for use in benchmarks, tests, or
+  // other places where sharing a common instance is useful. Note that there is
+  // nothing thread safe about this instance or type.
+  static auto Global() -> SourceGen&;
```
zygoloid:
```suggestion
  // Access a global instance of this type to generate Carbon code for benchmarks, tests, or
  // other places where sharing a common instance is useful. Note that there is
  // nothing thread safe about this instance or type.
  static auto Global() -> SourceGen&;
```
Seems important to point out this global instance generates Carbon not C++.

testing/base/source_gen.h
```diff
@@ -0,0 +1,227 @@
...
+  auto GetShuffledUniqueIds(int number, int min_length = 4, int max_length = 64,
+                            bool uniform = false)
+      -> llvm::SmallVector<llvm::StringRef>;
+
+  // Returns a collection of un-shuffled identifiers, otherwise the same as
+  // `GetShuffledIds`.
+  //
+  // Usually, benchmarks should use the shuffled version. However, this is
+  // useful when there is already a post-processing step to shuffle things as it
+  // is *dramatically* more efficient, especially in debug builds .
```
zygoloid:
```suggestion
  // is *dramatically* more efficient, especially in debug builds.
```

testing/base/source_gen_test.cpp
```diff
@@ -0,0 +1,152 @@
...
+    // But the sum of lengths should be identical.
+    EXPECT_THAT(SumSizes(ids2), Eq(ids_size_sum));
+  }
+
+  // Check length constraints have the desired effect.
+  ids = gen.GetShuffledIds(1000, /*min_length=*/10, /*max_length=*/20);
+  EXPECT_THAT(ids, Each(SizeIs(AllOf(Ge(10), Le(20)))));
+
+  // Check that uniform id length results in exact coverage of each possible
+  // length for an easy case.
```
zygoloid: I think it'd be useful to also test the case with a remainder.

testing/base/source_gen_test.cpp
```diff
@@ -0,0 +1,152 @@
...
+  EXPECT_THAT(ids, Contains(SizeIs(14)).Times(10));
+  EXPECT_THAT(ids, Contains(SizeIs(15)).Times(10));
+  EXPECT_THAT(ids, Contains(SizeIs(16)).Times(10));
+  EXPECT_THAT(ids, Contains(SizeIs(17)).Times(10));
+  EXPECT_THAT(ids, Contains(SizeIs(18)).Times(10));
+  EXPECT_THAT(ids, Contains(SizeIs(19)).Times(10));
+}
+
+// Largely covered by `Ids`, but need to check for uniqueness specifically.
+TEST(SourceGenTest, UniqueIds) {
```
zygoloid: I think it'd be good to test the length distribution and rounding behavior here.

testing/base/source_gen_test.cpp
```diff
@@ -0,0 +1,152 @@
...
+  return driver.RunCommand({"compile", "--phase=check", "test.carbon"}).success;
+}
+
+TEST(SourceGenTest, GenAPIFileDenseDeclsTest) {
+  SourceGen gen;
+
+  std::string source =
+      gen.GenAPIFileDenseDecls(1000, SourceGen::DenseDeclParams{});
+  // Should be within 10% of the requested line count.
+  EXPECT_THAT(source, Contains('\n').Times(AllOf(Ge(900), Le(1100))));
```
zygoloid: Do we really need to allow 10% each way? Given that we're forcing the sample distribution to exactly match the specified distribution (up to rounding various things to an integer), I think we should always get a lot closer than this.

testing/base/source_gen_test.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+  // Make sure we generated valid Carbon code.
+  EXPECT_TRUE(TestCompile(source));
+}
+
+TEST(SourceGenTest, GenAPIFileDenseDeclsCppTest) {
+  SourceGen gen(SourceGen::Language::Cpp);
+
+  // Generate a 1000-line file which is enough to have a reasonably accurate
+  // line count estimate and gave a few classes.
```
zygoloid:
```suggestion
  // line count estimate and have a few classes.
```

zygoloid (commented):
@jonmeow asked me to take a look at this.

common/benchmark_main.cpp
chandlerc: Happy to re-work this and all the other things in this code, but this is copied directly from the unittest version, and there is a TODO to try to refactor and share the code.

It seems a bit weird to only fix it here when it's copied, but also seems weird to make this PR larger to fix in both places...

My preference is just to address comments here in the PR that factors things together, but lemme know what you'd prefer.

common/benchmark_main.cpp
chandlerc: See above.

common/benchmark_main.cpp
chandlerc: See above.

testing/base/source_gen.h
```diff
@@ -0,0 +1,228 @@
...
+    // TODO: Add more parameters to control generating top-level constructs
+    // other than class definitions.
+
+    // Parameters used when generating class definitions.
+    ClassParams class_params = {};
+  };
+
+  // Access a global instance of this type to generate Carbon code for benchmarks, tests, or
+  // other places where sharing a common instance is useful. Note that there is
+  // nothing thread safe about this instance or type.
```
CarbonInfraBot, line 108: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  // Access a global instance of this type to generate Carbon code for
  // benchmarks, tests, or other places where sharing a common instance is
  // useful. Note that there is nothing thread safe about this instance or type.
```

testing/base/source_gen.cpp
chandlerc: Just to be clear, all of the formatting is really provisional as just a really rough hack to try and not have everything on a single line... I didn't put a lot of effort into making it good, was mostly just roughing _something_ out to start with...

Long term, I'd like to come up with a way to actually use a target line length and such to manage this and something closer to actually using an automated formatting strategy, but haven't come up with a good way yet.

All that said, my initial idea was to focus the line estimation (and formatting) on Carbon alone, and just somewhat ignore where those heuristics end up not quite matching C++. I don't really expect the exact line count / line width to be a big factor in compile time, but it'll be important to have exactly the same number of constructs and such with both languages to make comparative benchmarks meaningful.

So I think to the extent this stays long term, it'll probably be Carbon specific if that makes sense?

testing/base/source_gen.h
```diff
@@ -0,0 +1,240 @@
...
+                            llvm::SmallVectorImpl<llvm::StringRef>& param_names,
+                            llvm::raw_ostream& os) -> void;
+  auto GenerateClassDef(const ClassParams& params, ClassGenState& state,
+                        llvm::raw_ostream& os) -> void;
+
+  absl::BitGen rng;
+  llvm::BumpPtrAllocator storage;
+
+  Map<int, llvm::SmallVector<llvm::StringRef>> ids_by_length;
+  Map<int, std::pair<int, Set<llvm::StringRef>>> unique_ids_by_length;
```
jonmeow: Should these have underscore suffixes?

jonmeow (commented):
Drive-by comment. Note I'm assuming zygoloid will finish the review, so I've marked my comments resolved without looking closely.

testing/base/source_gen_main.cpp
```diff
@@ -0,0 +1,112 @@
...
+  output->flush();
+  return true;
+}
+
+}  // namespace
+}  // namespace Carbon::Testing
+
+auto main(int argc, char** argv) -> int {
+  // Do LLVM's initialization first, this will also transform UTF-16 to UTF-8.
+  Carbon::InitLLVM init_llvm(argc, argv);
```
jonmeow, line 108: You might consider using [SetWorkingDirForBazel](https://github.com/carbon-language/carbon-lang/blob/trunk/common/bazel_working_dir.h) so that `bazel run :source_gen --output=foo.carbon` puts `foo.carbon` in the working directory that `bazel` is run from, instead of the working directory used inside `bazel run`.

(noticed while trying to use this)

testing/base/source_gen.cpp
chandlerc: Agreed and done for all of these (I think). This also corrects the estimation by not including a blank line for the first class.

testing/base/source_gen.cpp
chandlerc: Somewhat tweaked after changing the way we handle the estimation as suggested, and added a named constant for this. Should address the core idea though.

I've also made the check a bit more strict to ensure we get at least one class.

testing/base/source_gen.cpp
chandlerc: Did something roughly similar as part of getting the estimation correct.

I use a string literal rather than a character literal though as I find that a bit clearer when streaming.

testing/base/source_gen.cpp
chandlerc: > The dip at 5 and bump at 12 have been removed entirely by this smoothing, and 10/11 look a bit over-represented. (In the data, 9 and 12 are nearly 2x as common as 10 and 11 but all four have the same probability here.) Do you think those are just measurement errors / peculiarities of LLVM, or should we be modeling them?

I'm not sure, the sample size is a bit small. But I also haven't investigated much because...

> How important is it to get this right-ish?

I suspect not very. I think it's important to have *very roughly* the right shape, but I doubt the details of this matter much.

But maybe more importantly, all of this is just refactored out of the old lexer benchmark. I've not tried to change or update anything here. Happy to revisit the distribution though if there is a better one. Even if it's not about being "right", the key is to have a distribution that matches what we want to optimize... And there we want "everything" to be fast, but to have a rough weighting that matches reality I think?

testing/base/source_gen.cpp
chandlerc: ... Really? Huh. Sure, moved!

testing/base/source_gen.cpp
```diff
@@ -0,0 +1,674 @@
...
+      dest.push_back(id);
+      --number;
+    }
+  });
+  CARBON_CHECK(number == 0);
+}
+
+// Returns a shuffled sequence of integers in the range [min, max].
+//
+// The order of the returned integers is random, but each integer in the range appears the same number of times in the result, with the number of appearances rounded up for lower numbers and rounded down for higher numbers in order to exactly produce `number` results.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
// The order of the returned integers is random, but each integer in the range
// appears the same number of times in the result, with the number of
// appearances rounded up for lower numbers and rounded down for higher numbers
// in order to exactly produce `number` results.
```

testing/base/source_gen.cpp
chandlerc: Rewrote this and other comments here to try to be much more helpful. Sorry about that.

testing/base/source_gen.cpp
chandlerc: We have quite a bit of use without `std::`. We also don't use `std::` in other analogous cases like `uint8_t` and friends, so it seems odd to use it here.

I generally don't think the `std::` helps readability or portability here, and so I lean towards omitting it consistently. Happy to fix the places where we *do* use it to be more consistent if desired...

That said, not using `memcpy` here any more.

testing/base/source_gen.cpp
chandlerc: Sure! I thought it would make the code quite a bit more complicated, but yeah, as it turns out this always works well. And it nudged me to simplify the unique popper's fallback contract an more. Good idea!

testing/base/source_gen.h
chandlerc: Doh, yes. Sorry I missed that.

testing/base/source_gen_main.cpp
chandlerc: Good idea, done!

testing/base/source_gen.cpp
chandlerc: Not using `memcpy` any more.

testing/base/source_gen.cpp
chandlerc: I tried this, and the math doesn't seem to check out.

Specifically, we don't actually get the right number of identifiers with the code as suggested (the check below fails). I've tried several things, and none of them seem to reliably produce what you're suggesting here, but probably just haven't hit on the right model.

I feel like for this to work we'd need to introduce some kind of *clamping* of the distribution. I'm not opposed, but I don't have the math for it, and it seems like this could be a follow-up if you're interested? I don't really have a strong use case for making the small number of identifiers that much more representative TBH, so unclear this is worthwhile.

testing/base/source_gen_test.cpp
chandlerc: Good idea, done.

testing/base/source_gen_test.cpp
chandlerc: Curious what test you would like for that? (And it seems like that should be in the non-unique test?)

I can copy and paste the distribution from the implementation into the test, and check that it matches, but that seemed like it would add a lot of fragility without much added value.

I could test small N-counts, but as you point out the behavior there isn't necessarily "good" or something that seems like we should hold as a thing that doesn't regress.

I've added a test that with a reasonably large N-count we get at least one identifier in each length bucket.

testing/base/source_gen_test.cpp
chandlerc: It didn't seem especially important to be that close, and for small files, it is tricky. I was worried about being fragile. For example, the current logic doesn't end up within 1%. (979 lines is one that I just got.)

I've switched it to 5% -- could get to 1% with a larger file size, but curious what you think here?

chandlerc (commented):
Sorry it's been slow, I keep being interrupted working on this. =/

I had somewhat rushed to get it out for review because folks seemed to really want it even if in a rough shape. But it seems like that's not actually working well. I've suggested fixing forward for a lot of things here, but maybe that approach just isn't appealing? If so, maybe good to chat about that to figure out what makes sense.

As a concrete case -- some of the things with several comments are really pre-existing issues that are showing up again. If fixing forward doesn't make sense, then I'll need to work on fixing those in separate PRs and rebase this on top of them. Want to understand what folks would prefer as a strategy here.

Also, I still have to do the 'id' -> 'identifier' renaming. I think I got everything else, but that one I'll just need to spend some time going through and renaming. It makes sense, but figured I'd send out the other changes and ask the questions above first as its already been a while.

common/benchmark_main.cpp
zygoloid: I'm happy to defer handling the comments in this file to another PR and deal with them at the same time as we improve the other file. Maybe for now just repeating the TODO in this file would help.

testing/base/source_gen.cpp
zygoloid: That makes sense. How about adding something to the comment like:

```
// Some heuristic numbers used when formatting generated code. These heuristics
// are loosely based on what we expect to make Carbon code readable, and might not
// fit as well in C++, but we use the same heuristics across languages for
// simplicity and to make the output in different languages more directly comparable.
```

testing/base/source_gen.cpp
```diff
@@ -0,0 +1,690 @@
...
+    // non-uniform distribution. For uniform, we always take one.
+    int scale = uniform ? 1 : IdLengthCounts[length - 1];
+
+    // Now we can compute how many identifiers of this length to request.
+    int length_count = (number / count_sum) * scale;
+    if (number_rem > 0) {
+      int rem_adjustment = std::min(scale, number_rem);
+      length_count += rem_adjustment;
+      number_rem -= rem_adjustment;
+    }
```
zygoloid, line 436: I think that
```suggestion
    int length_count = (scale * number) / count_sum;
    if (number_rem > 0) {
      length_count++;
      number_rem--;
    }
```
would give a better approximation of the distribution (with an error of at most one in each bucket's size, rather than at most `IdLengthCounts[i]`).

With the code as-is, if you ask for 120 identifiers with your heuristic distribution, then `count_sum` is something like 350, `number / count_sum` is 0, and you'll end up with 40 of size 1, 40 of size 2, and 40 of size 3, despite size 4 being the largest in the distribution. Whereas with the modified version, you should get around 14 of size 1, 2, and 3, and around 18 of size 4. (Ballpark numbers, I didn't actually work out the correct `count_sum` figure.)

testing/base/source_gen_test.cpp
```diff
@@ -0,0 +1,171 @@
...
+  EXPECT_THAT(ids, Contains(SizeIs(10)).Times(10));
+  EXPECT_THAT(ids, Contains(SizeIs(11)).Times(10));
+  EXPECT_THAT(ids, Contains(SizeIs(12)).Times(10));
+  EXPECT_THAT(ids, Contains(SizeIs(13)).Times(10));
+  EXPECT_THAT(ids, Contains(SizeIs(14)).Times(10));
+  EXPECT_THAT(ids, Contains(SizeIs(15)).Times(10));
+  EXPECT_THAT(ids, Contains(SizeIs(16)).Times(10));
+  EXPECT_THAT(ids, Contains(SizeIs(17)).Times(9));
+  EXPECT_THAT(ids, Contains(SizeIs(18)).Times(9));
+  EXPECT_THAT(ids, Contains(SizeIs(19)).Times(9));
```
zygoloid: Are these tests for the uniform distribution in the right test function?

testing/base/source_gen_test.cpp
zygoloid: Yes, I think I meant a non-unique test :)

One simple thing you could maybe test is that identifiers of length 4 are the most common length, which I think ought to hold for N >= 4 or so (with the adjustment I suggested earlier), and shouldn't be too much of a change detector test since I don't think we'd expect that property to change as we tune the distribution.

testing/base/source_gen.cpp
chandlerc: So, I tried to reply to the original form of this comment here:

https://github.com/carbon-language/carbon-lang/pull/4124#discussion_r1708920131

I re-checked, and the issue still seems to be there.

common/benchmark_main.cpp
chandlerc: Done.

testing/base/source_gen.cpp
chandlerc: Done.

testing/base/source_gen_test.cpp
chandlerc: I... think so?

There are two axes: unique vs. non-unique, and uniform distribution vs. non-uniform distribution.

I have a function split out for the uniqueness tests. I guess can split the two distributions into separate functions if useful?

testing/base/source_gen_test.cpp
chandlerc: Tried to add something like this...

chandlerc (commented):
Still need to do the great renaming, but a quick update with some of the other fixes.

testing/base/source_gen.cpp
zygoloid: Sorry, didn't see your previous comment for Github Reasons.

Hm, yeah, I see the problem. It's not easy to predict the amount of rounding error in advance with my approach. We could do it in two passes, but yeah, probably doesn't matter enough to bother given we primarily expect to generate large examples. This is good enough for an initial implementation.

chandlerc:
Ok, all the renaming of `ID` stuff is I think done. At least, my regex-ing in `source_gen*` seems to be clean.

testing/base/source_gen.cpp
```diff
@@ -386,35 +389,35 @@ static auto Sum(const T& range) -> int {
...
 // The main functionality provided here is collecting the correct number of
 // identifiers from each of the lengths in the range [min_length, max_length]
 // and either in our default representative distribution or a uniform
 // distribution.
 //
 // Note that this template must be defined prior to its use below.
-template <typename AppendIds>
-auto SourceGen::GetIdsImpl(int number, int min_length, int max_length,
-                           bool uniform, AppendIds append_ids)
+template <typename AppendFunc>
```
jonmeow: Very minor, but I think we're generally using `Fn` instead of `Func` as an abbreviation, consistent with Carbon.

But, and sorry if I'm missing context here, but had you considered something like `llvm::function_ref<int, int, llvm::SmallVector<llvm::StringRef>&>`?

testing/base/source_gen.cpp
```diff
@@ -125,75 +125,76 @@ auto SourceGen::GenAPIFileDenseDecls(int target_lines, DenseDeclParams params)
...
-                 [this](int length, int length_count,
-                        llvm::SmallVectorImpl<llvm::StringRef>& dest) {
-                   AppendUniqueIdentifiers(length, length_count, dest);
-                 });
-
-  return ids;
+  llvm::SmallVector<llvm::StringRef> idents =
+      GetIdentifiersImpl(number, min_length, max_length, uniform,
+                         [this](int length, int length_count,
+                                llvm::SmallVectorImpl<llvm::StringRef>& dest) {
```
jonmeow, line 172: Why SmallVectorImpl instead of SmallVector? (actually, I see this in a few spots)

jonmeow (approved):
LGTM. Note, I'm trying to focus on the delta since zygoloid's last review. Feel free to merge, just a few small spelling questions.

testing/base/source_gen.cpp
chandlerc: Good call on naming, but even better call on using `function_ref`, switched to that.

But using `SmallVectorImpl` to avoid encoding the small size (as above).

testing/base/source_gen.cpp
chandlerc: This is the idiom of passing a small vector reference for appending without encoding a specific small size.

chandlerc (commented):
PTAL!

PR #4125, opened 2024-07-12
Use verbose formatting of instructions on crash messages.

jonmeow:
Changes crash messages to start printing verbose forms of instructions, rather than just the ID. Fixes some indentation issues with stacks. Also switches unexpected inst formatting, because now there are lots, and it'd be helpful to know where they are.

This uses a pimpl pattern for Formatter due to the number of member functions on Formatter. Maybe we should refactor that, but this didn't feel like a good place to do so.

Note, I have two concerns about this change... to note them here, to make sure others are considering them when evaluating the implementation:

1. Some instructions are very verbose to print, as evidenced by the fn_decl printing (which includes function params) or scope printing (which includes scope members).
    - I'm not sure whether there's a way to simply reduce this, as it seems essential to the requested printing of instructions.
    - Long-term, we may at least want to limit the number of lines printed here. However, I've already spent a fair amount of time here and I think it's in a good state to evaluate.
2. Increased complexity in the crash handler may result in crash messages failing to generate.
    - For example, a crash in Formatter (and its deps, such as InstNamer or location handling) prevents a stack from being printed. I'm pretty sure I've written crashes in Formatter before.

Here's an example crash snippet (generated by adding a crash inside `return` handling) before:

```
2.	NodeStack:
	0.	FunctionDefinitionStart -> function2
	1.	ReturnStatementStart -> no value
	2.	IntLiteral -> inst+26
inst_block_stack_:
	0.	block<invalid>	{inst+0, inst+1, inst+2, inst+23}
	1.	block9	{inst+26}
param_and_arg_refs_stack:
args_type_info_stack_:
```

And after:

```
2.	Check::Context
          NodeStack:
            0. FunctionDefinitionStart: function2
            1. ReturnStatementStart: no value
            2. IntLiteral:
              unexpected.inst+26.loc12_10: i32 = int_literal 0 [template = constants.%.2]
          inst_block_stack_:
            0. block<invalid> {
                package: <namespace> = namespace [template] {
                  .Core = unexpected.inst+2
                  .F = unexpected.inst+23.loc11_22
                }
                unexpected.inst+1 = import Core
                unexpected.inst+2: <namespace> = namespace unexpected.inst+1, [template] {}
                unexpected.inst+23.loc11_22: %F.type = fn_decl @F [template = constants.%F] {
                  unexpected.inst+9.loc11_9: init type = call constants.%Bool() [template = bool]
                  unexpected.inst+10.loc11_9: type = value_of_initializer unexpected.inst+9.loc11_9 [template = bool]
                  unexpected.inst+11.loc11_9: type = converted unexpected.inst+9.loc11_9, unexpected.inst+10.loc11_9 [template = bool]
                  unexpected.inst+12.loc11_6: bool = param b
                  @F.%b: bool = bind_name b, unexpected.inst+12.loc11_6
                  unexpected.inst+19.loc11_18: init type = call constants.%Int32() [template = i32]
                  unexpected.inst+20.loc11_18: type = value_of_initializer unexpected.inst+19.loc11_18 [template = i32]
                  unexpected.inst+21.loc11_18: type = converted unexpected.inst+19.loc11_18, unexpected.inst+20.loc11_18 [template = i32]
                  @F.%return: ref i32 = var <return slot>
                }
              }
            1. block9 {
                unexpected.inst+26.loc12_10: i32 = int_literal 0 [template = constants.%.2]
              }
          param_and_arg_refs_stack:
          args_type_info_stack_:
```


toolchain/sem_ir/formatter.cpp
```diff
@@ -22,70 +22,56 @@
...
-    // blocks. For example, there may be branching in the initializer of a
-    // global or a type expression.
-    if (auto block_id = sem_ir_.top_inst_block_id(); block_id.is_valid()) {
-      llvm::SaveAndRestore file_scope(scope_, InstNamer::ScopeId::File);
-      FormatCodeBlock(block_id);
+    constexpr int MaxPrinted = 10;
+    if (block.size() > MaxPrinted) {
+      Indent();
+      *out_ << "... skipping " << (block.size() - MaxPrinted) << " insts ...\n";
+      block = block.take_back(MaxPrinted);
```
zygoloid: Maybe
```suggestion
      *out_ << "... skipping " << (block.size() - MaxPrinted + 1) << " insts ...\n";
      block = block.take_back(MaxPrinted - 1);
```
so we never print more than `MaxPrinted` lines, including the "skipping" line. (It bugs me when tools say "skipping 1 thing", and could have used the same amount of vertical space to actually print the thing they skipped.)

toolchain/sem_ir/formatter.cpp
```diff
@@ -952,10 +987,26 @@ class Formatter {
...
+auto Formatter::PrintPartialTrailingCodeBlock(
+    llvm::ArrayRef<SemIR::InstId> block, int indent, llvm::raw_ostream& out)
+    -> void {
+  formatter_->PrintPartialTrailingCodeBlock(block, indent, out);
+}
+
+auto Formatter::PrintInst(SemIR::InstId inst_id, int indent,
+                          llvm::raw_ostream& out) -> void {
+  formatter_->PrintInst(inst_id, indent, out);
 }
```
zygoloid: Instead of the pimpl approach and saving/restoring members, perhaps we could create an instance of `FormatterImpl` in each of these `Formatter::Print` functions? We don't actually want to persist any state between them other than the `InstNamer` (which the `Formatter` could still own).

toolchain/sem_ir/inst_namer.cpp
```diff
@@ -152,7 +152,15 @@ auto InstNamer::GetNameFor(ScopeId scope_id, InstId inst_id) const
...
-    llvm::raw_string_ostream(str) << "<unexpected instref " << inst_id << ">";
+    llvm::raw_string_ostream str_stream(str);
+    str_stream << "unexpected." << inst_id;
+    auto loc_id = sem_ir_.insts().GetLocId(inst_id);
+    // TODO: Consider handling inst_id cases.
+    if (loc_id.is_node_id()) {
+      auto token = parse_tree_.node_token(loc_id.node_id());
+      str_stream << ".loc" << tokenized_buffer_.GetLineNumber(token) << "_"
+                 << tokenized_buffer_.GetColumnNumber(token);
+    }
```
zygoloid: Including the location here seems nice. That said, I would prefer to keep the angle brackets here, both for consistency with other things we print this way, and to make it clearer that this output is not valid SemIR (while we may not have plans to parse SemIR ourselves, other tools might -- it's also nice to make it easier to see in test output when the test produced these bad outputs).

toolchain/sem_ir/formatter.cpp
jonmeow: Done, with a slightly different approach: I think it's cleaner to do this with a single edit, in the `if`.

toolchain/sem_ir/inst_namer.cpp
jonmeow: Done

toolchain/sem_ir/formatter.cpp
jonmeow: Done

jonmeow:
Note, per side-discussion about the risks, we'll try this out. People will be asked to comment to #toolchain if they use the more detailed info. If nobody's using it, we may back it out (at least in part)

toolchain/sem_ir/formatter.h
```diff
@@ -5,16 +5,39 @@
 #ifndef CARBON_TOOLCHAIN_SEM_IR_FORMATTER_H_
 #define CARBON_TOOLCHAIN_SEM_IR_FORMATTER_H_
 
+#include <memory>
+
```
zygoloid: I think this is now unused.

zygoloid (approved):

toolchain/sem_ir/formatter.h
jonmeow: Fixed

PR #4126, opened 2024-07-12
Adjust driver references to use the install version.

jonmeow:
Also makes an alias so that this is easier to find. Verified that running the alias still finds prelude files.

Note the actual target is printed when building (although a symlink is also created, using that symlink confuses file-finding).

```
╚╡bazel build :carbon
...
Target //toolchain/install:prefix_root/bin/carbon up-to-date:
  bazel-bin/toolchain/install/prefix_root/bin/carbon
```

### DIFF
--- a/toolchain/BUILD
+++ b/toolchain/BUILD
@@ -1,0 +1,10 @@
+# Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+# Exceptions. See /LICENSE for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# A convenience target for running the toolchain with the full prelude
+# available.
+alias(
+    name = "carbon",
+    actual = "//toolchain/install:prefix_root/bin/carbon",
+)
--- a/toolchain/BUILD
+++ b/toolchain/BUILD
@@ -5,6 +5,6 @@
 # A convenience target for running the toolchain with the full prelude
 # available.
 alias(
-    name = "carbon",
+    name = "toolchain",
     actual = "//toolchain/install:prefix_root/bin/carbon",
 )
--- a/README.md
+++ b/README.md
@@ -311,7 +311,7 @@ Carbon:
 
 ```shell
 # Build and run the toolchain's help to get documentation on the command line.
-$ bazel run //toolchain:carbon -- help
+$ bazel run //toolchain -- help
 ```
 
 For complete instructions, including installing dependencies on various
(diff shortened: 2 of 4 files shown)


toolchain/BUILD
```diff
@@ -0,0 +1,10 @@
+# Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+# Exceptions. See /LICENSE for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# A convenience target for running the toolchain with the full prelude
+# available.
+alias(
+    name = "carbon",
+    actual = "//toolchain/install:prefix_root/bin/carbon",
```
zygoloid, line 9: Does this cause the core libraries to get built and installed into the install prefix? I may just be missing it, but I don't see a suitable dependency in `toolchain/install/BUILD`.

toolchain/BUILD
```diff
@@ -0,0 +1,10 @@
+# Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+# Exceptions. See /LICENSE for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# A convenience target for running the toolchain with the full prelude
+# available.
+alias(
+    name = "carbon",
```
zygoloid: Did you consider naming this `toolchain` instead, to support simply `bazel run //toolchain`? I don't feel at all strongly about this, but the shorter name seems slightly nicer. But perhaps using the intended final name of the binary here is a good thing?

zygoloid (approved):
Great idea, thanks!

toolchain/BUILD
jonmeow: ╚╡bazel query 'somepath(//toolchain/install:prefix_root/bin/carbon, //toolchain/install:core_data)'
//toolchain/install:prefix_root/bin/carbon
//toolchain/driver:carbon
//toolchain/driver:driver
//toolchain/install:install_lib_data
//toolchain/install:core_data

toolchain/BUILD
jonmeow: Done.

I was hesitant similar to how it was "//toolchain/driver:carbon" instead of either "//toolchain/driver" or "//toolchain/carbon", but don't care strongly.

toolchain/BUILD
jonmeow: I totally grant that I don't like this kind of circular dependency.

PR #4127, opened 2024-07-12
Refactor tool fetching for easier updates.

jonmeow:
Adds scripts/calculate_release_shas.py to print the versions, and updates tool versions. Consolidates target-determinator logic with the other tool logic.

This is also intended to make it easier to add more tools.


chandlerc (approved):
LG

PR #4128, opened 2024-07-12
Compute specific constant values.

zygoloid:
When forming a specific (previously called a generic instance), evaluate the eval block of the generic to determine the values of any constants used in that specific. The majority of the work here is updating eval.cpp so that it can use the results of prior evaluations in the same block when computing later values.

Include the computed results in the formatted SemIR output.

### DIFF
--- a/toolchain/check/eval.cpp
+++ b/toolchain/check/eval.cpp
@@ -18,6 +18,108 @@
 namespace Carbon::Check {
 
 namespace {
+// Information about an eval block of a generic instance that we are currently
+// building.
+struct GenericInstanceEvalBlockInfo {
+  // The region within the generic instance whose eval block we are building.
+  SemIR::GenericInstIndex::Region region;
+  // The work-in-progress contents of the eval block.
+  llvm::ArrayRef<SemIR::InstId> values;
+};
+
+// Information about the context within which we are performing evaluation.
+struct EvalContext {
+  auto GetInContext(SemIR::ConstantId const_id) -> SemIR::ConstantId {
+    if (!const_id.is_symbolic()) {
+      return const_id;
+    }
+
+    // While resolving a generic instance, map from previous instructions in the
+    // eval block into their evaluated values. These values won't be present on
+    // the instance itself yet, so `GetConstantInInstance` won't be able to find
+    // them.
+    if (instance_eval_block_info) {
+      const auto& symbolic_info =
+          context.constant_values().GetSymbolicConstant(const_id);
+      if (symbolic_info.index.is_valid() &&
+          symbolic_info.generic_id ==
+              context.generic_instances().Get(instance_id).generic_id &&
+          symbolic_info.index.region() == instance_eval_block_info->region) {
+        auto inst_id =
+            instance_eval_block_info->values[symbolic_info.index.index()];
+        CARBON_CHECK(inst_id.is_valid())
+            << "Forward reference in eval block: index "
+            << symbolic_info.index.index() << " referenced before evaluation";
+        return context.constant_values().Get(inst_id);
+      }
+    }
+
+    // Map from an instance-specific constant value to the canonical value.
+    return GetConstantInInstance(context.sem_ir(), instance_id, const_id);
+  }
+
+  // Gets the constant value of the specified instruction in this context.
+  auto GetConstantValue(SemIR::InstId inst_id) -> SemIR::ConstantId {
+    return GetInContext(context.constant_values().Get(inst_id));
+  }
+
+  // Gets the constant value of the specified type in this context.
+  auto GetConstantValue(SemIR::TypeId type_id) -> SemIR::ConstantId {
+    return GetInContext(context.types().GetConstantId(type_id));
+  }
+
+  // Gets the instruction describing the constant value of the specified type in
+  // this context.
+  template<typename IdT>
+  auto GetConstantValueAsInst(IdT id) -> SemIR::Inst {
+    return insts().Get(
+        context.constant_values().GetInstId(GetConstantValue(id)));
+  }
+
+  auto ints() -> CanonicalValueStore<IntId>& { return sem_ir().ints(); }
+  auto floats() -> FloatValueStore& { return sem_ir().floats(); }
+  auto bind_names() -> SemIR::BindNameStore& {
+    return sem_ir().bind_names();
+  }
+  auto functions() -> const ValueStore<SemIR::FunctionId>& {
+    return sem_ir().functions();
+  }
+  auto classes() -> const ValueStore<SemIR::ClassId>& {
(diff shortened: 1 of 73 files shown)


toolchain/check/eval.cpp
```diff
@@ -17,6 +17,104 @@
...
+struct SpecificEvalInfo {
+  // The region within the specific whose eval block we are building.
+  SemIR::GenericInstIndex::Region region;
+  // The work-in-progress contents of the eval block.
+  llvm::ArrayRef<SemIR::InstId> values;
+};
+
+// Information about the context within which we are performing evaluation.
+struct EvalContext {
+  auto GetInContext(SemIR::ConstantId const_id) -> SemIR::ConstantId {
```
jonmeow, line 35: Comment? It's not clear to me how this differs from the others, and it looks like a public API (though, if this is class-ified, should it be private?)

toolchain/check/eval.cpp
```diff
@@ -17,6 +17,104 @@
...
+// Information about an eval block of a specific that we are currently building.
+struct SpecificEvalInfo {
+  // The region within the specific whose eval block we are building.
+  SemIR::GenericInstIndex::Region region;
+  // The work-in-progress contents of the eval block.
+  llvm::ArrayRef<SemIR::InstId> values;
+};
+
+// Information about the context within which we are performing evaluation.
+struct EvalContext {
```
jonmeow, line 30: Noting how much of this is functions, should it be a class? Examining member uses, it looks like `context` and `specific_id` are part of the API, but `file` and `specific_eval_info` are not.

Though, thinking about this from a different perspective, had you considered making the static functions in eval.cpp be members of this class? I think we use the `Context` object elsewhere because we have a lot of different files we want to compose access from, but here, we have a scoped use-case. It'd also avoid the need to pass `EvalContext` as a parameter.

Without a `class`, `file` and `sem_ir()` in particular look like an odd match to me: why an accessor to return a public member?

toolchain/check/eval.cpp
```diff
@@ -17,6 +17,104 @@
...
+
+  auto sem_ir() -> SemIR::File& { return file; }
+
+  auto emitter() -> Context::DiagnosticEmitter& { return context.emitter(); }
+
+  // The type-checking context in which we're performing evaluation.
+  Context& context;
+  // A cached reference to the file, to avoid a double indirection when
+  // accessing its value stores.
+  SemIR::File& file = context.sem_ir();
```
jonmeow: To note, this feels odd to me: it's unusual, and seems like it'd have a negligible effect on performance. Maybe I'm thinking about that wrong though?

The flipside is that all the internal `sem_ir()` calls look odd; if you keep this, maybe change them to reference the cached reference directly?

toolchain/check/eval.cpp
```diff
@@ -1272,4 +1401,48 @@ auto TryEvalInst(Context& context, SemIR::InstId inst_id, SemIR::Inst inst)
...
+
+  SpecificEvalInfo eval_info = {
+      .region = region,
+      .values = result,
+  };
+  EvalContext eval_context = {
+      .context = context,
+      .specific_id = specific_id,
+      .specific_eval_info = &eval_info,
+  };
```
jonmeow: FWIW, the extra struct from SpecificEvalInfo had made me expect it'd be replaced and/or separately used, but this is the only assignment I see other than `nullptr`. Given the way EvalContext is constructed, had you considered moving its members over, or perhaps making it `std::optional` instead of `*`?

toolchain/sem_ir/formatter.cpp
```diff
@@ -327,6 +331,55 @@ class Formatter {
...
+    const auto& generic = sem_ir_.generics().Get(specific.generic_id);
+
+    OpenBrace();
+    FormatSpecificRegion(generic, specific,
+                         GenericInstIndex::Region::Declaration, "declaration");
+    FormatSpecificRegion(generic, specific,
+                         GenericInstIndex::Region::Definition, "definition");
+    CloseBrace();
+
+    out_ << '\n';
```
```suggestion
    out_ << " ";

    const auto& generic = sem_ir_.generics().Get(specific.generic_id);

    OpenBrace();
    FormatSpecificRegion(generic, specific,
                         GenericInstIndex::Region::Declaration, "declaration");
    FormatSpecificRegion(generic, specific,
                         GenericInstIndex::Region::Definition, "definition");
    CloseBrace();

    out_ << "\n";
```

Minor consistency suggestion... note line 358 does `"\n"`, I don't think there'll be much performance impact either way.

toolchain/check/eval.cpp
```diff
@@ -17,6 +17,104 @@
...
+    return GetInContext(context.types().GetConstantId(type_id));
+  }
+
+  // Gets the instruction describing the constant value of the specified type in
+  // this context.
+  template <typename IdT>
+  auto GetConstantValueAsInst(IdT id) -> SemIR::Inst {
+    return insts().Get(
+        context.constant_values().GetInstId(GetConstantValue(id)));
+  }
```
jonmeow: This call confuses me a little: should it be taking a TypeId as parameter? It looks like the templated version isn't used.

(or will more callers be added? for one caller, maybe this function should be inlined)

jonmeow (approved):
Looks good... Please go ahead and merge when you're comfortable. While I'm interested in your thoughts about class-ifying eval.cpp, I expect you may want to split it out (if at all).

toolchain/check/eval.cpp
zygoloid: My intent is to make this into a class, but I'm trying to keep this PR minimal. Will send a follow-up with that change.

toolchain/check/eval.cpp
zygoloid: Removed for now; we can re-add if we have evidence of this being an issue.

toolchain/check/eval.cpp
zygoloid: Added a comment.

toolchain/check/eval.cpp
zygoloid: I've ungeneralized it. I'd prefer to not inline it; I found that having a name for this operation made it a lot easier to follow what the code was doing, and I am anticipating more callers.

toolchain/check/eval.cpp
zygoloid: Originally I was concerned about this being large (containing a `SmallVector`) and I didn't want to include it in `EvalContext` for that reason, but that ended up not being a problem. Turned into a `std::optional`.

Bonniemarie216 (approved):

jonmeow:
@Bonniemarie216 I see you approved a lot of different changes. Please don't do that anymore, because it's not contributing to the review, but it does create notifications for people who are monitoring repository activity. Thanks!

Issue #3896, opened 2024-04-17
Getting-Started package issue


keys-i:
### Description of the bug:

bazel is not a package of Ubuntu

### Link to documentation:

https://github.com/carbon-language/carbon-lang

### What should it say instead?

bazel-bootstrap \

### Any other information, logs, or outputs that you want to share?

I tried running this on Ubuntu 23.10 which told me to use bazel-bootstrap

jonmeow:
Thanks. Note that bazel-bootstrap is not the same, and will conflict with a correct bazel install [both try to install /usr/bin/bazel]. We should fix notes, but in the meantime bazel's install instructions are at https://bazel.build/install/ubuntu#install-on-ubuntu; I'd suggest either setting up bazel's apt repo ([commands here](https://bazel.build/install/ubuntu#add-dis-uri)), or downloading bazelisk.

jonmeow:
Instructions updated by #4129

PR #4129, opened 2024-07-12
Add a run_bazelisk wrapper script for linux.

### LINKED ISSUE #3071: Can't find bazel package on ubuntu 22.04
### Description of the bug:

In the documentation is said:

These commands should help set up a development environment on your machine.

### Debian or Ubuntu

```
# Update apt.
sudo apt update

# Install tools.
sudo apt install \
  bazel \
  clang \
  gh \
  libc++-dev \
  lld \
  python3 \
  zlib1g-dev

# Install pre-commit.
pip3 install pre-commit

# Set up git.
# If you don't already have a fork:
gh repo fork --clone carbon-language/carbon-lang
cd carbon-lang
pre-commit install

# Run tests.
bazel test //...:all
```
but actually bazel cannot be installed out of the box in ubuntu 22.04.

```
user@pc:~$ sudo apt install bazel
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
E: Unable to locate package bazel
```

Furthermore python3-pip should be added to the list of packages to install because is needed for the command
`pip3 install pre-commit`

### Link to documentation:

https://github.com/carbon-language/carbon-lang/blob/trunk/docs/project/contribution_tools.md

Improve the documentation to list the exact commands to follow to install bazel

_No response_

jonmeow:
I was considering making a "/bazel" alias for this, but I'm really on the fence about whether that's a good choice. However, I do think this is good to have to simplify installs.

Fixes #3071 and #3896

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,7 +288,6 @@ sudo apt update
 
 # Install tools.
 sudo apt install \
-  bazel \
   clang \
   libc++-dev \
   libc++abi-dev \
@@ -303,7 +302,7 @@ Then you can build and run the explorer:
 
 ```shell
 # Build and run the explorer.
-$ bazel run //explorer -- ./explorer/testdata/print/format_only.carbon
+$ ./scripts/run_bazelisk.py run //explorer -- ./explorer/testdata/print/format_only.carbon
 ```
 
 And you can try out our toolchain which has a very early-stage compiler for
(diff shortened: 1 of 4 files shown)


chandlerc (approved):
Nice, and as you say in the change description, seems like a good compromise -- let folks directly install bazelisk if they can/want-to, but have this as a fallback that is reliable and doesn't require any extra steps.

PR #4131, opened 2024-07-14
Build an evaluation block for the definition region of a generic.

zygoloid:


zygoloid:
Depends on #4128.

jonmeow (approved):
Looks good. Not merging due to dependency.

PR #4138, opened 2024-07-16
Rebuild the type of constants during evaluation.

zygoloid:
When evaluating in a generic context, a constant with a symbolic type might evaluate to a constant with a concrete type (or a more specific symbolic type). This can't actually happen yet given the current state of the toolchain, as far as I can determine, so this is more just a refactoring for now, but will be relied upon by future generics work.

### DIFF
--- a/toolchain/check/eval.cpp
+++ b/toolchain/check/eval.cpp
@@ -222,7 +222,7 @@ static auto GetConstantValue(EvalContext& eval_context, SemIR::InstId inst_id,
   return eval_context.context.constant_values().GetInstId(const_id);
 }
 
-// Gets the type corresponding to the specified type in this evaluation context.
+// Given a type which may refer to a generic parameter, returns the corresponding type in the evaluation context.
 static auto GetConstantValue(EvalContext& eval_context, SemIR::TypeId type_id,
                              Phase* phase) -> SemIR::TypeId {
   auto const_id = eval_context.GetConstantValue(type_id);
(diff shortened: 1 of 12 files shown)


toolchain/check/eval.cpp
```diff
@@ -227,12 +222,12 @@ static auto GetConstantValue(EvalContext& eval_context, SemIR::InstId inst_id,
   return eval_context.context.constant_values().GetInstId(const_id);
 }
 
-// A type is always constant, but we still need to extract its phase.
+// Gets the type corresponding to the specified type in this evaluation context.
```
jonmeow: I'm not sure whether "type" in the above is referring to the input type_id or the returned type_id (I could read it either way for the pair). Maybe:

```suggestion
// Given a type which may refer to a generic parameter, returns the corresponding type in the evaluation context.
```

jonmeow (approved):
LGTM

toolchain/check/eval.cpp
```diff
@@ -227,12 +222,12 @@ static auto GetConstantValue(EvalContext& eval_context, SemIR::InstId inst_id,
   return eval_context.context.constant_values().GetInstId(const_id);
 }
 
-// A type is always constant, but we still need to extract its phase.
+// Given a type which may refer to a generic parameter, returns the corresponding type in the evaluation context.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
// Given a type which may refer to a generic parameter, returns the
// corresponding type in the evaluation context.
```

PR #4139, opened 2024-07-16
Substitute into generic class and interface definitions when we require them to be fully defined.

zygoloid:
When referring to a constant within a specific, such as a field of a generic class, use that specific version of the constant's value.

### DIFF
--- a/toolchain/check/decl_name_stack.cpp
+++ b/toolchain/check/decl_name_stack.cpp
@@ -206,7 +206,7 @@ static auto PushNameQualifierScope(Context& context,
   // parameter scope.
   context.scope_stack().PopIfEmpty();
 
-  // When declaring a member of a generic, resolve the specific self.
+  // When declaring a member of a generic, resolve the self specific.
   if (instance_id.is_valid()) {
     ResolveSpecificDefinition(context, instance_id);
   }
--- a/toolchain/check/decl_name_stack.cpp
+++ b/toolchain/check/decl_name_stack.cpp
@@ -206,6 +206,11 @@ static auto PushNameQualifierScope(Context& context,
   // parameter scope.
   context.scope_stack().PopIfEmpty();
 
+  // When declaring a member of a generic, resolve the specific self.
+  if (instance_id.is_valid()) {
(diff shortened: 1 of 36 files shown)


zygoloid:
Depends on #4138.

toolchain/check/decl_name_stack.cpp
```diff
@@ -206,6 +206,11 @@ static auto PushNameQualifierScope(Context& context,
   // parameter scope.
   context.scope_stack().PopIfEmpty();
 
+  // When declaring a member of a generic, resolve the specific self.
```
```suggestion
  // When declaring a member of a generic, resolve the specific's self.
```

toolchain/check/generic.h
```diff
@@ -46,6 +46,12 @@ auto MakeGenericInstance(Context& context, SemIR::GenericId generic_id,
 auto MakeGenericSelfInstance(Context& context, SemIR::GenericId generic_id)
     -> SemIR::GenericInstanceId;
 
+// Attempts to resolve the definition of the given specific, by evaluating the
+// eval block of the corresponding generic. Returns false if a definition is
+// not available.
```
jonmeow: As part of resolving, does this modify anything? That may be important to comment on -- I don't quite know what "resolve" means in this context.

jonmeow (approved):
LGTM

toolchain/check/decl_name_stack.cpp
zygoloid: What I mean here is the specific that is the generic's self -- the thing we used to call the "self instance". Maybe
```suggestion
  // When declaring a member of a generic, resolve the self specific.
```
would work?

toolchain/check/generic.h
zygoloid: Attempted to clarify.

toolchain/check/decl_name_stack.cpp
jonmeow: Yeah, I'd just sub "instance" with "specific" for how you'd phrase it. Maybe `resolve Self's specific.` if that makes sense.

PR #4141, opened 2024-07-17
Mark some leaf classes `final` to suppress `-Wnon-virtual-dtor`.

zygoloid:


jonmeow (approved):
I think we're still discussing what to do with this, but it seems easy to roll back if we decide we want a different approach.

PR #4149, opened 2024-07-19
Implement syntactic merge checks for parameters.

jonmeow:
Note this isn't implementing checking through imports. The parse node there is harder to access through the context, so would require examining the entity in order to get the import declaration, to get at the ImportIR. We also don't have a parse tree attached in that case, and would need to add one to SemIR::File. But I believe we do want to add that, so it's explicitly a TODO.

Note GetTokenText re-lexes literal values, so there's a bit of potential overhead there. Not sure if we want a more efficient manner for comparing in cases like this.

### DIFF
--- a/toolchain/check/merge.cpp
+++ b/toolchain/check/merge.cpp
@@ -268,9 +268,77 @@ static auto CheckRedeclParams(Context& context, SemIRLoc new_decl_loc,
   return true;
 }
 
+// Returns true if the two nodes represent the same syntax.
+static auto IsNodeSyntaxEqual(Context& context, Parse::NodeId new_node_id,
+                              Parse::NodeId prev_node_id) -> bool {
+  if (context.parse_tree().node_kind(new_node_id) !=
+      context.parse_tree().node_kind(prev_node_id)) {
+    return false;
+  }
+
+  // TODO: Should there be a trivial way to check if we need to check spellings?
+  // Identifiers and literals need their text checked, but keywords and
+  // operators shouldn't.
+  auto new_spelling = context.tokens().GetTokenText(
+      context.parse_tree().node_token(new_node_id));
+  auto prev_spelling = context.tokens().GetTokenText(
+      context.parse_tree().node_token(prev_node_id));
+  return new_spelling == prev_spelling;
+}
+
+// Returns false if redeclaration parameter syntax doesn't match.
+static auto CheckRedeclParamSyntax(Context& context,
+                                   Parse::NodeId new_first_param_node_id,
+                                   Parse::NodeId new_last_param_node_id,
+                                   Parse::NodeId prev_first_param_node_id,
+                                   Parse::NodeId prev_last_param_node_id)
+    -> bool {
+  // Parse nodes may not always be available to compare.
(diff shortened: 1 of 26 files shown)


josh11b:
Please find another reviewer, as I'll be on vacation for the next two weeks.

toolchain/check/merge.cpp
```diff
@@ -268,22 +268,99 @@ static auto CheckRedeclParams(Context& context, SemIRLoc new_decl_loc,
...
+// Returns true if the two nodes represent the same syntax.
+static auto IsNodeSyntaxEqual(Context& context, Parse::NodeId new_node_id,
+                              Parse::NodeId prev_node_id) -> bool {
+  if (context.parse_tree().node_kind(new_node_id) !=
+      context.parse_tree().node_kind(prev_node_id)) {
+    return false;
+  }
+
+  // TODO: Should there be a trivial way to check if we need to check spellings?
+  // Identifiers and literals need their text checked, but keywords and
```
zygoloid: (It'd be nice to just compare the `IdentifierId`s for identifiers where possible, though we'd need to check for `r#` somehow and still presumably need string comparisons for cross-file cases. But not in this PR!)

toolchain/check/merge.cpp
```diff
@@ -268,22 +268,99 @@ static auto CheckRedeclParams(Context& context, SemIRLoc new_decl_loc,
...
+         "prev_first_param_node_id.is_valid";
+
+  auto new_range = context.parse_tree().postorder(new_first_param_node_id,
+                                                  new_last_param_node_id);
+  auto prev_range = context.parse_tree().postorder(prev_first_param_node_id,
+                                                   prev_last_param_node_id);
+
+  // zip is using the shortest range. If they differ in length, there should be
+  // some difference inside the range, or the semantics should have a mismatch
+  // (such as a different parameter count). As a consequence, we don't
```
zygoloid: I think we don't need to rely on a semantic mismatch here -- parse nodes are self-delimiting (we can reconstruct the tree structure from the sequence of nodes) so if the nodes match then they have the same tree structure, and because one of them is a complete tree, the other must be too.
```suggestion
  // some difference inside the range. As a consequence, we don't
```

zygoloid (approved):
Thanks!

Another edge case that would be useful to explicitly test (mostly because we've had discussions about using a different rule) is grouping parentheses differing between declarations.

toolchain/check/merge.cpp
jonmeow: Adjusted, because I think it's important to note _why_ there should be a difference. Right now, that's actually because the `)` is part of the range.

toolchain/check/merge.cpp
jonmeow: Adjusted

jonmeow:
> Another edge case that would be useful to explicitly test (mostly because we've had discussions about using a different rule) is grouping parentheses differing between declarations.

Added test

Also, made me realize I wasn't testing deduced params explicitly, so added that too.

jonmeow:
Per discussion, added a test for `r#` identifiers (because you'd mentioned them in a comment), and made them a TODO

zygoloid (approved):

PR #4153, opened 2024-07-22
Improve namespace handling in imports.

jonmeow:
This implements a few closely related features:

- Starts merging namespaces discovered inside imports.
- Stores results of cross-package name lookup as an entry inside the scope.
  - Note this is particularly visible with `i32`.
- Moves more of the imported instructions to the import scope.

Note this is primarily for executing the namespace TODO in check.cpp, which is removed here. `testdata/namespace/merging_with_indirections.carbon` tests key behavior.


jonmeow:
Depends on #4154 [2024-07-23: dependencies are merged]

toolchain/check/import.cpp
```diff
@@ -421,4 +440,147 @@ auto ImportLibrariesFromOtherPackage(Context& context,
   }
 }
 
+// Looks up a name in an imported scope. An `identifier` is provided if
+// `name_id` corresponds to an identifier in the current file; otherwise,
+// `name_id` is file-agnostic and can be used directly.
```
```suggestion
// Looks up a name in a scope imported from another package. An `identifier` is provided if
// `name_id` corresponds to an identifier in the current file; otherwise,
// `name_id` is file-agnostic and can be used directly.
```
I think this is only handling the cross-package case (due to the access check).

toolchain/check/import.cpp
```diff
@@ -421,4 +440,147 @@ auto ImportLibrariesFromOtherPackage(Context& context,
...
+          import_ir.sem_ir, import_scope_entry->inst_id);
+      continue;
+    }
+
+    // Add the first result found.
+    result_id = AddScopedImportRef(
+        context, scope_id, context.name_scopes().Get(scope_id), name_id,
+        {.ir_id = import_ir_id, .inst_id = import_scope_entry->inst_id},
+        SemIR::AccessKind::Public);
+    LoadImportRef(context, result_id);
```
zygoloid: This looks correct but a little subtle: in particular, the fact that we're guaranteed to not add an `ImportRef` for a namespace is non-obvious from the control flow. I wonder we can reorder it a bit to make that clearer -- for example:

```suggestion
    // Add the first result found.
    if (!result_id.is_valid()) {
      // If the imported instruction is a namespace, we add it directly instead of
      // as an ImportRef.
      if (auto import_ns = import_inst.TryAs<SemIR::Namespace>()) {
        result_id = AddNamespaceFromOtherPackage(context, import_ir_id,
                                                 import_scope_entry->inst_id,
                                                 *import_ns, scope_id, name_id);
      } else {
        result_id = AddScopedImportRef(
            context, scope_id, context.name_scopes().Get(scope_id), name_id,
            {.ir_id = import_ir_id, .inst_id = import_scope_entry->inst_id},
            SemIR::AccessKind::Public);
        LoadImportRef(context, result_id);
      }
      continue;
    }

    // When namespaces collide between files, merge lookup in the scopes.
    if (auto import_ns = import_inst.TryAs<SemIR::Namespace>()) {
      if (auto ns = context.insts().TryGetAs<SemIR::Namespace>(result_id)) {
        auto& name_scope = context.name_scopes().Get(ns->name_scope_id);
        name_scope.import_ir_scopes.push_back(
            {import_ir_id, import_ns->name_scope_id});
        continue;
      }
    }

    // When there's a name collision, they need to either be the same canonical
    // instruction, or we'll diagnose.
    if (!canonical_result_inst) {
      canonical_result_inst =
          GetCanonicalImportIRInst(context, &context.sem_ir(), result_id);
    }
    VerifySameCanonicalImportIRInst(
        context, result_id, *canonical_result_inst, import_ir_id,
        import_ir.sem_ir, import_scope_entry->inst_id);
```

toolchain/check/testdata/namespace/merging_with_indirections.carbon
```diff
@@ -0,0 +1,225 @@
...
+// CHECK:STDOUT:
+// CHECK:STDOUT: file {
+// CHECK:STDOUT:   package: <namespace> = namespace [template] {
+// CHECK:STDOUT:     .NS1 = imports.%NS1
+// CHECK:STDOUT:     .Core = imports.%Core
+// CHECK:STDOUT:     .F = %F.decl
+// CHECK:STDOUT:   }
+// CHECK:STDOUT:   %Core.import = import Core
+// CHECK:STDOUT:   %default.import = import <invalid>
+// CHECK:STDOUT:   %.loc5: <namespace> = namespace [template] {}
```
zygoloid, line 120: Just an observation, not requesting a change in this PR:

It looks like we create a vestigial namespace instruction here for `namespace NS1` that's not connected to the corresponding name scope. I think it'd be better for this to refer to the same name scope as the namespace instruction in the imports block, like a redeclaration of (eg) a class would.

zygoloid (approved):

toolchain/check/testdata/namespace/merging_with_indirections.carbon
jonmeow: I think namespaces are different though, in that multiple libraries can declare the same namespace. It's not obvious to me which should own it, if there's an issue. Maybe there's better to do here, but a redeclaration of a class just doesn't feel like the right model.

PR #4154, opened 2024-07-22
Use the package/library name in ImportIRId formatting.

jonmeow:
Also adds import_ir_scope to namespace formatting. I'd done this as an aid for #4153, and am splitting it out.

### DIFF
--- a/toolchain/sem_ir/file.h
+++ b/toolchain/sem_ir/file.h
@@ -175,7 +175,7 @@ class File : public Printable<File> {
   auto filename() const -> llvm::StringRef { return filename_; }
 
  private:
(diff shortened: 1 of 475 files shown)


toolchain/sem_ir/file.h
```diff
@@ -172,10 +175,18 @@ class File : public Printable<File> {
   auto filename() const -> llvm::StringRef { return filename_; }
 
  private:
+  // True if there parts of the IR may be invalid.
```
geoffromer:
```suggestion
  // True if parts of the IR may be invalid.
```

geoffromer (approved):

PR #4155, opened 2024-07-22
Rebuild all constants in the eval block.

zygoloid:
Instead of reusing instructions from the generic entity in the eval block, rebuild constants in the same way we rebuild types. The previous attempt to not rebuild these constants assumed that every constant used in a generic would be built in that generic, and not referenced directly or referenced from some enclosing scope, which isn't true in practice and is a fragile assumption in any case.

We could add back some reuse of instructions from the generic -- if we happen to see the right instruction to build a constant, we could opportunistically reuse it -- but given the complexity added by doing so, I'm not pursuing that here.

Now that the eval block for a generic consists of instructions uniquely owned by that generic, rather than often being shared with another entity, include the generic in the formatted SemIR output. I'm using the same scope name for the generic object itself as for the parameterized class / function / interface, because there are very frequently references between them and this keeps the IR simpler and more readable, and avoids needing to invent a second name for the scope.


toolchain/sem_ir/inst_namer.h
```diff
@@ -164,6 +173,7 @@ class InstNamer {
   std::vector<std::pair<ScopeId, Namespace::Name>> insts;
   std::vector<std::pair<ScopeId, Namespace::Name>> labels;
   std::vector<Scope> scopes;
+  std::vector<ScopeId> generic_scopes;
```
jonmeow: Should this have a `_` suffix? The others are also missing one, I remember thinking about this during recent formatter changes and I'd meant to clean it up, but it slipped my mind. I understand if you don't want to rename the existing ones in this PR, but other fields already have a suffix `_` so it doesn't strike me as simple consistency.

Also, should it have comments? I think it might help newcomers understand the behavior.

toolchain/sem_ir/formatter.cpp
```diff
@@ -212,14 +201,11 @@ class FormatterImpl {
 
   auto FormatInterface(InterfaceId id) -> void {
     const Interface& interface_info = sem_ir_.interfaces().Get(id);
+    FormatGeneric(interface_info.generic_id);
```
jonmeow: Just a thought, but I've been mulling commonality in this code... if FormatInterfaceName etc were renamed to FormatName (or FormatIdAsName?) and relied on overload resolution, maybe a chunk of these functions could be moved into a template for code sharing?

(not for this PR, just wondering about your thoughts -- I've also been considering doing things like translating FormatInst to a switch instead of overloads)

toolchain/check/testdata/interface/no_prelude/fail_add_member_outside_definition.carbon
```diff
@@ -79,13 +79,17 @@ interface Outer {
...
 // CHECK:STDOUT:
-// CHECK:STDOUT: interface @Inner
-// CHECK:STDOUT:     generic [@Outer.%Self: %.3] {
-// CHECK:STDOUT:   %Self: %.4 = bind_symbolic_name Self 1 [symbolic = %Self (constants.%Self.3)]
+// CHECK:STDOUT: generic @Inner(@Outer.%Self: %.3) {
+// CHECK:STDOUT: !definition:
+// CHECK:STDOUT:   %Self.2: %.4 = bind_symbolic_name Self 1 [symbolic = %Self.2 (constants.%Self.3)]
+// CHECK:STDOUT: }
+// CHECK:STDOUT:
+// CHECK:STDOUT: interface @Inner {
```
jonmeow: Formatting-wise, these look disconnected to me. Is that intentional? Perhaps the `interface` could be moved into the generic at an indent, or add some other connection other than the `@Inner` name match?

jonmeow (commented):
Although you note using the same name for the generic and declaration-specific (e.g., fn) entry is deliberate, I'm finding it a little difficult to read. The lack of nesting makes them look like they're declaring entities with separate names (at least, as I read IR), and something like `@GetValue.%.1` now seems ambiguous as to which body I should be looking into.

Had you considered further nesting declarations, e.g., although I mention putting the `fn` inside the `generic`, maybe you could alternately have `!generic:` as an indented block inside the `fn`, perhaps indenting it? Or perhaps indenting `fn ...` inside the `generic ...`? Or, just name `generic @GetAddr` to `generic @GetAddr.generic` / `@GetAddr.params` to produce a distinct name from `fn @GetAddr`.

Also, why is `%.1` so vague about its source? But that looks pre-existing (e.g., `fn @GetAddr[addr @Class.%self.loc12_19.3: @Class.%.loc12_29 (%.2)]() -> %.3`)

toolchain/sem_ir/inst_namer.h
zygoloid: Added comments and trailing `_`s to all these members.

toolchain/sem_ir/formatter.cpp
zygoloid: Turns out that this is directly useful for the new approach to generic formatting, so I went ahead and did it in this PR.

toolchain/check/testdata/interface/no_prelude/fail_add_member_outside_definition.carbon
zygoloid: Done.

jonmeow (approved):
Thanks! I think this looks better.

PR #4156, opened 2024-07-23
Support for calling non-generic methods in a specific class.

zygoloid:
Use the specific parameter types for checking, and the specific return
type as the type of the call.


zygoloid:
Depends on #4155.

toolchain/check/scope_stack.h
```diff
@@ -88,6 +88,14 @@ class ScopeStack {
   // there is no such instruction, such as for a block scope.
   auto PeekInstId() const -> SemIR::InstId { return Peek().scope_inst_id; }
 
+  // Returns the specific associated with the innermost enclosing scope that is
+  // associated with a specific. This will generally be the self specific of the
+  // innermost enclosing generic, as there is no way to enter any other specific
+  // scope.
+  auto PeekSpecificId() const -> SemIR::GenericInstanceId {
+    return Peek().instance_id;
+  }
```
jonmeow, line 97: Would it be worth just exposing `Peek()`, rather than per-field accessors (noting this is the 4th)?

toolchain/check/testdata/class/generic/subst_into_decl.carbon
```diff
@@ -0,0 +1,216 @@
...
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+// TIP: To test this file alone, run:
+// TIP:   bazel test //toolchain/testing:file_test --test_arg=--file_tests=toolchain/check/testdata/class/generic/subst_into_decl.carbon
+// TIP: To dump output, run:
+// TIP:   bazel run //toolchain/testing:file_test -- --dump_output --file_tests=toolchain/check/testdata/class/generic/subst_into_decl.carbon
+
+class Class(T:! type) {
```
jonmeow: Compared with fail_todo_use.carbon, what do you think of combining the tests into the same file? They seem similar, although the other is exercising a little bit that apparently isn't implemented yet.

toolchain/sem_ir/function.cpp
```diff
@@ -35,14 +38,18 @@ auto GetCalleeFunction(const File& sem_ir, InstId callee_id) -> CalleeFunction {
...
-auto Function::declared_return_type(const File& file) const -> TypeId {
+auto Function::declared_return_type(const File& file,
+                                    GenericInstanceId specific_id) const
+    -> TypeId {
   if (!return_storage_id.is_valid()) {
     return TypeId::Invalid;
   }
-  return file.insts().Get(return_storage_id).type_id();
+  return GetTypeInInstance(file, specific_id,
+                           file.insts().Get(return_storage_id).type_id());
```
jonmeow, line 52: To note, the accessor naming is starting to feel awkward here. This call seems non-trivial.

jonmeow (approved):
Looks good!

toolchain/check/scope_stack.h
zygoloid: My inclination is to keep `ScopeStackEntry` as a private implementation detail -- I'm not sure we want to expose everything in it.

toolchain/check/testdata/class/generic/subst_into_decl.carbon
zygoloid: Refactored into a test for just member access and a test for just initialization. I also noticed I didn't have any coverage for member access naming a static member function, which seemingly doesn't work yet. Added test coverage for that as a TODO.

toolchain/sem_ir/function.cpp
zygoloid: Renamed away from accessor style.

zygoloid:
@jonmeow The return slot change is a little non-trivial, please can you take another look? (I want to make some more significant changes there, but I'm trying to make a more minimal change for now to keep the fuzzer happy.)

toolchain/check/function.cpp
```diff
@@ -94,18 +99,23 @@ auto CheckFunctionReturnType(Context& context, SemIRLoc loc,
...
   } else if (SemIR::GetInitRepr(context.sem_ir(), return_type_id)
                  .has_return_slot()) {
-    function.return_slot = SemIR::Function::ReturnSlot::Present;
+    result = SemIR::Function::ReturnSlot::Present;
   } else {
-    function.return_slot = SemIR::Function::ReturnSlot::Absent;
+    result = SemIR::Function::ReturnSlot::Absent;
+  }
+  if (!specific_id.is_valid()) {
+    function.return_slot = result;
```
jonmeow, line 116: Should we have return slot caching for specifics?

jonmeow (approved):
Generally looks good, but can you document the new return slot behavior on Function? It looks like it's now either (a) cached for a non-generic function, or (b) computed on the fly for a generic function unless entirely absent. However, `has_return_slot` and `return_slot` don't really explain the behavior (`has_return_slot` talks about (a) but ignores (b)). [a TODO might be fine, or just leave it alone if you're already working on that and don't want a TODO to immediately remove]

toolchain/check/function.cpp
zygoloid: I'm not sure the caching is even worthwhile for non-generic functions. I don't think we look at the return slot information for a function frequently enough for this to matter, and it's pretty cheap anyway (mostly just a hash table lookup to find if the type is complete). The functional effect of the caching is that we only diagnose an incomplete return type once, but perhaps we can do something more narrowly targeted if we think that's important.

Anyway, that's for a future PR I think :)

PR #4161, opened 2024-07-24
Move common entity fields to a 'base' struct.

jonmeow:
I'd considered moving DeclParams uses over, but when handling qualified names, there's a parse node instead of an instruction. I did try to unify a couple other uses though, including adding MergeDefinition. I expect `interface` will use a little more once it's more completely implemented, but maybe I'm wrong about that.


jonmeow:
Per discussion, restructured from a `base` field to instead use multiple inheritance.

toolchain/sem_ir/entity_with_params_base.h
```diff
@@ -0,0 +1,91 @@
...
+//   - Allows named initialization, such as:
+//     `{{.name_id = ...}, {.function_field = ...}}`
+//   - Makes `entity.name_id` access work.
+//   - Allows passing a `EntityWithParamsBase*` when only common fields are
+//     needed.
+//   - Does all this in a way that's vanilla C++.
+struct EntityWithParamsBase {
+  auto PrintBaseFields(llvm::raw_ostream& out) const -> void {
+    out << "name: " << name_id << ", parent_scope: " << parent_scope_id;
+  }
```
zygoloid, line 40: It seems a bit weird to mention the "Base" in the name here. Maybe we could just call this `PrintFields`, shadow it in the derived class when necessary, and have `Printable<T>` call `PrintFields` if it exists:

```c++
struct EntityWithParamsBase {
  auto PrintFields(llvm::raw_ostream& out) const -> void { ... }
  // ...
}

struct Function : EntityWithParamsBase, FunctionFields, Printable<Function> {
  auto PrintFields(llvm::raw_ostream& out) const -> void {
    EntityWithParamsBase::PrintFields(out);
    out << ...function fields...
  }
};
```

But I'm happy to leave that to later given that this is already a large PR.

zygoloid (approved):

PR #4163, opened 2024-07-24
Remove caching of return slot usage.

zygoloid:
The caching isn't buying us much, and is adding complexity and divergence between the codepaths for generic and non-generic functions.

This means we no longer suppress diagnostics for the second or subsequent time we call a function with an incomplete return type. If we want to add that back, it might be worth considering moving the suppression to `TryToCompleteType` and only diagnosing that a type is incomplete once, regardless of why we're requiring it to be complete.


toolchain/sem_ir/function.h
```diff
@@ -15,38 +15,42 @@ namespace Carbon::SemIR {
...
+  struct ReturnInfo {
+    // Builds return information for a given declared return type.
+    static auto ForType(const File& file, TypeId type_id) -> ReturnInfo;
+
+    // Returns whether the return information could be fully computed.
+    auto is_valid() const -> bool { return slot != ReturnSlot::Incomplete; }
+
+    // Returns whether the function has a return slot. Can only be called for
+    // valid return info.
+    auto has_return_slot() const -> bool {
```
jonmeow: FWIW, you might consider either `ReturnSlot slot` -> `ReturnSlot return_slot` or `has_return_slot` -> `has_slot`

toolchain/lower/file_context.cpp
```diff
@@ -136,17 +137,21 @@ auto FileContext::BuildFunctionDecl(SemIR::FunctionId function_id)
...
+  // lowering it if it's needed.
+
+  // TODO: Pass in a specific ID for generic functions.
+  const auto specific_id = SemIR::GenericInstanceId::Invalid;
 
-  const bool has_return_slot = function.has_return_slot();
+  const auto return_info = function.GetReturnInfo(sem_ir(), specific_id);
+  CARBON_CHECK(return_info.is_valid()) << "Should not lower invalid functions.";
+
+  const bool has_return_slot = return_info.has_return_slot();
```
jonmeow: FWIW, I'd lean towards eliminating the `has_return_slot` local in spots like this. I'm not sure the locals in this file were needed before (by contrast, `bool has_return_slot = GetInitRepr(sem_ir_, inst.type_id).has_return_slot();` makes sense). But when you already have the `return_info` object, also having `return_info.has_return_slot()` separate feels like it's just being verbose.

toolchain/check/function.cpp
```diff
@@ -73,49 +73,29 @@ auto CheckFunctionTypeMatches(Context& context,
...
+    -> SemIR::Function::ReturnInfo {
+  auto return_info = function.GetReturnInfo(context.sem_ir(), specific_id);
 
-  auto return_type_id =
-      function.GetDeclaredReturnType(context.sem_ir(), specific_id);
-  CARBON_CHECK(return_type_id.is_valid())
-      << "Have return storage but no return type.";
+  // If we couldn't determine the return information due to the return type
+  // being incomplete, try to complete it now.
+  if (return_info.slot == SemIR::Function::ReturnSlot::Incomplete) {
```
jonmeow: Had you considered restructuring this for early returns, in order to trim nested scopes? e.g.:

```suggestion
  if (return_info.slot.is_valid()) {
    return return_info;
  }
```

If you do this, might want to adjust comments appropriately. Also, line 94 could return directly.

But either way, probably use `is_valid` here if you're adding it.

toolchain/lower/file_context.cpp
```diff
@@ -136,17 +137,21 @@ auto FileContext::BuildFunctionDecl(SemIR::FunctionId function_id)
...
-  }
+  // TODO: Consider tracking whether the function has been used, and only
+  // lowering it if it's needed.
+
+  // TODO: Pass in a specific ID for generic functions.
+  const auto specific_id = SemIR::GenericInstanceId::Invalid;
 
-  const bool has_return_slot = function.has_return_slot();
+  const auto return_info = function.GetReturnInfo(sem_ir(), specific_id);
+  CARBON_CHECK(return_info.is_valid()) << "Should not lower invalid functions.";
```
jonmeow, line 147: Isn't this duplicating the check in `has_return_slot`?

toolchain/sem_ir/function.h
```diff
@@ -15,38 +15,42 @@ namespace Carbon::SemIR {
...
-    // Computing whether the function should have a return slot failed, for
-    // example because the return type was incomplete.
-    Error
+    // Computing whether the function should have a return slot failed because
+    // the return type was incomplete.
+    Incomplete,
+  };
+
+  // Information about how a function returns its return value.
+  struct ReturnInfo {
```
jonmeow: FWIW, had you considered:

SemIR::Function::ReturnInfo -> SemIR::ReturnSlot
SemIR::Function::ReturnSlot -> SemIR::ReturnSlot::Kind

The Kind enum probably won't be used much outside (I do see the switches), but maybe naming this `ReturnSlot` would be helpful? And while the nesting of `ReturnSlot` in `Function` made sense when it was for `Function::return_slot`, it seems odder to me when `Function::ReturnSlot` is for `ReturnInfo::slot` versus having it be `ReturnInfo::ReturnSlot` (or as in the example above, `ReturnSlot::Kind`).

jonmeow (approved):

toolchain/lower/file_context.cpp
zygoloid: Yes, but it produces a more useful check failure message.

toolchain/check/function.cpp
zygoloid: I think that structure would be awkward if there's ever a second reason why we can't use a type as a return type. Which there is: we can't use abstract class types as return types either. I should add a diagnostic for that... will send follow-up PR.

toolchain/sem_ir/function.h
zygoloid: Done.

toolchain/lower/file_context.cpp
zygoloid: Done.

toolchain/sem_ir/function.h
zygoloid: I think I can remove `ReturnSlot` entirely. Will investigate in a separate PR.

zygoloid: For now I've moved these two types out of `Function`.

Bonniemarie216 (approved):

PR #4164, opened 2024-07-24
fix: `clangd: -32001: invalid AST`

brymer-meneses:
Ignore the `-march=*` in `compile_commands.json` which causes issues in Neovim on Apple M2. This fix is from this [comment](https://github.com/clangd/clangd/issues/1582#issuecomment-1500031372)

### DIFF
--- a/.clangd
+++ b/.clangd
@@ -3,4 +3,5 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 CompileFlags:
+  # Workaround for https://github.com/clangd/clangd/issues/1582
   Remove: [-march=*]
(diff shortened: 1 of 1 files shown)


.clangd
```diff
@@ -0,0 +1,6 @@
+# Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+# Exceptions. See /LICENSE for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+CompileFlags:
+  Remove: [-march=*]
```
jonmeow, line 7:
```suggestion
  # Workaround for https://github.com/clangd/clangd/issues/1582
  Remove: [-march=*]
```

jonmeow (approved):
Thanks! I think this is good to add, just want a note so we remember why it's here (particularly if the issue is fixed)

.clangd
brymer-meneses: Sure, I have just committed this change.

PR #4165, opened 2024-07-24
Rename "generic instance" to "specific" throughout the toolchain.

zygoloid:
As discussed in toolchain meeting, we want to avoid overloading the meaning of "instance", and "specific" was the best name we found. It's a little unorthodox and inventive, but hopefully over time will become as unsurprising as the term "generic" is.


jonmeow (approved):
LGTM, looks comprehensive to me. Thanks!

Bonniemarie216 (approved):

PR #4171, opened 2024-07-25
Refactor some commonality in formatter.

jonmeow:
Also tries to add comments for things. Note, I'm trying to improve understandability here, so if you don't think this is helping I can undo things.

### DIFF
--- a/toolchain/sem_ir/formatter.cpp
+++ b/toolchain/sem_ir/formatter.cpp
@@ -36,8 +36,9 @@ class FormatterImpl {
   auto Format() -> void {
     out_ << "--- " << sem_ir_.filename() << "\n\n";
 
-    FormatConstants();
-    FormatImportRefs();
+    FormatScope(InstNamer::ScopeId::Constants, sem_ir_.constants().array_ref());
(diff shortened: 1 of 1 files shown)


toolchain/sem_ir/formatter.cpp
```diff
@@ -456,27 +452,34 @@ class FormatterImpl {
...
 
+  // Prints instructions for a code block.
   auto FormatCodeBlock(llvm::ArrayRef<InstId> block) -> void {
     for (const InstId inst_id : block) {
       FormatInst(inst_id);
     }
   }
 
+  // Prints a code block with braces, intended to be used trailing after other
+  // content on the same line (may wrap lines).
```
geoffromer: What do you mean by "may wrap lines"? I think of that as meaning inserting line breaks to try to maintain a particular line width, but I didn't think the formatter ever did that.

geoffromer (approved):

toolchain/sem_ir/formatter.cpp
jonmeow: Just, may add line breaks. Not necessarily wrapping to a specific width. Versus printing all content on the same line.

PR #4173, opened 2024-07-26
Add initial parsing for 'extern library'

jonmeow:


toolchain/parse/node_kind.def
```diff
@@ -286,12 +286,14 @@ CARBON_PARSE_NODE_KIND(StructComma)
 CARBON_PARSE_NODE_KIND(StructLiteral)
 CARBON_PARSE_NODE_KIND(StructTypeLiteral)
 
+CARBON_PARSE_NODE_KIND(ExternLibraryModifier)
+CARBON_PARSE_NODE_KIND(ExternTokenModifier)
```
geoffromer: Optional: this name makes it sound like it's a modifier for tokens, rather than a modifier that is a token. How about something like `BareExternModifier`?

geoffromer (approved):

toolchain/parse/node_kind.def
jonmeow: I think "token" is consistent with the name used below, `CARBON_PARSE_NODE_KIND_TOKEN_MODIFIER`

I also don't think "Bare" carries meaning, it's not a term we use elsewhere.

jonmeow: Going a different direction, renaming to "ExternModifier" + "ExternModifierWithLibrary" since the former keeps the same name.

geoffromer: LGTM

Bonniemarie216 (approved):

PR #4178, opened 2024-07-31
Fix incorrect `value_id` and location in imported `BindSymbolicName`.

zygoloid:
Instead of updating the `value_id` on the canonical constant `BindSymbolicName` to refer to some particular instance of that constant, create a new instruction, and attach the proper location to it.


jonmeow (approved):
Looks good.

PR #4179, opened 2024-07-31
Import support for generics and specifics

zygoloid:
Import generics and specifics when they are referenced by imported entities.

When importing a generic, we import the symbolic constants required by its eval block, and then rebuild the eval block itself given the list of constants it needs to compute. This is likely a bit less efficient than directly importing the contents of the eval block, but avoids needing to either extend the importer code to be able to import the instructions that can appear in the eval block or extend the evaluator to cope with instructions from a different `SemIR::File`.

Importing a symbolic constant is unaffected, and does not yet preserve the associated generic and index within that generic, so uses of a generic from an imported IR still don't pick up values from the specific, but the improved functionality can be seen in the changes to the SemIR in the testcases.


toolchain/check/generic.cpp
```diff
@@ -231,6 +241,41 @@ static auto MakeGenericEvalBlock(Context& context, SemIR::GenericId generic_id,
...
+
+  Map<SemIR::InstId, SemIR::InstId> constants_in_generic;
+
+  // For the definition region, populate constants from the declaration.
+  if (region == SemIR::GenericInstIndex::Region::Definition) {
+    auto decl_eval_block = context.inst_blocks().Get(
+        context.generics().Get(generic_id).decl_block_id);
+    for (auto inst_id : decl_eval_block) {
+      constants_in_generic.Insert(
+          context.constant_values().GetConstantInstId(inst_id), inst_id);
```
jonmeow: Should this be asserting success of the insert? If not, maybe add a comment noting they may be duplicated?

toolchain/check/generic.cpp
```diff
@@ -231,6 +241,41 @@ static auto MakeGenericEvalBlock(Context& context, SemIR::GenericId generic_id,
...
+    -> SemIR::InstBlockId {
+  context.inst_block_stack().Push();
+
+  Map<SemIR::InstId, SemIR::InstId> constants_in_generic;
+
+  // For the definition region, populate constants from the declaration.
+  if (region == SemIR::GenericInstIndex::Region::Definition) {
+    auto decl_eval_block = context.inst_blocks().Get(
+        context.generics().Get(generic_id).decl_block_id);
+    for (auto inst_id : decl_eval_block) {
```
jonmeow:
```suggestion
    constants_in_generic.GrowForInsertCount(decl_eval_block.size());
    for (auto inst_id : decl_eval_block) {
```

toolchain/check/import_ref.cpp
```diff
@@ -409,42 +437,154 @@ class ImportRefResolver {
...
+  auto GetLocalGenericData(SemIR::GenericId generic_id) -> GenericData {
+    if (!generic_id.is_valid()) {
+      return GenericData();
     }
-    return GetLocalInstBlockContents(
-        import_ir_.specifics().Get(specific_id).args_id);
+
+    const auto& generic = import_ir_.generics().Get(generic_id);
+
+    GenericData result = {
```
jonmeow: You might consider shifting the construction to guarantee RVO, as discussed.

toolchain/check/import_ref.cpp
```diff
@@ -409,42 +437,154 @@ class ImportRefResolver {
...
     if (!specific_id.is_valid()) {
-      return SemIR::SpecificId::Invalid;
+      return {.generic_const_id = SemIR::ConstantId::Invalid, .args = {}};
     }
     const auto& specific = import_ir_.specifics().Get(specific_id);
-    // TODO: Import the generic.
-    auto generic_id = SemIR::GenericId::Invalid;
-    auto args_id = GetLocalCanonicalInstBlockId(specific.args_id, args);
-    // TODO: Also import the specific.
+    return {
```
jonmeow, line 550: Similar RVO comment

toolchain/check/import_ref.cpp
```diff
@@ -409,42 +437,154 @@ class ImportRefResolver {
...
+        .args = GetLocalInstBlockContents(specific.args_id),
+        .decl_block = GetLocalInstBlockContents(specific.decl_block_id),
+        .definition_block =
+            GetLocalInstBlockContents(specific.definition_block_id),
+    };
+  }
+
+  // Gets a local specific whose data was already imported by
+  // GetLocalSpecificData. Does not add any new work.
+  auto GetLocalSpecific(SemIR::SpecificId import_specific_id,
```
jonmeow: You might consider "GetOrAddLocalSpecific" to reflect that it's not simply fetching data, plus the GetOrAdd call.

toolchain/check/import_ref.cpp
```diff
@@ -409,42 +437,154 @@ class ImportRefResolver {
...
+    // Form a corresponding local specific ID.
+    const auto& import_specific =
+        import_ir_.specifics().Get(import_specific_id);
+    auto generic_id = GetLocalGenericId(data.generic_const_id);
+    auto args_id =
+        GetLocalCanonicalInstBlockId(import_specific.args_id, data.args);
+
+    // Populate the specific. Note that we might get data from multiple
+    // different import IRs, so only import data we don't already have.
+    auto specific_id = context_.specifics().GetOrAdd(generic_id, args_id);
```
jonmeow, line 576: Should this specific_id be returned instead of calling GetOrAdd again below? If not, can you add some comments explaining what's going on?

Separately, if specific_id is pre-existing, is it still necessary to do the sets below? i.e., are the `is_valid` checks mainly because `GetOrAdd` doesn't include a signal for whether the import was already done, or because of forward declarations being an issue? (particularly `decl_block_id` being conditional)

toolchain/check/import_ref.cpp
```diff
@@ -409,42 +437,154 @@ class ImportRefResolver {
...
-  // Gets a local argument list corresponding to the arguments of an imported
-  // specific.
-  auto GetLocalSpecificArgs(SemIR::SpecificId specific_id)
-      -> llvm::SmallVector<SemIR::InstId> {
-    if (!specific_id.is_valid()) {
-      return {};
+  // Gets a local version of the data associated with a generic.
+  auto GetLocalGenericData(SemIR::GenericId generic_id) -> GenericData {
+    if (!generic_id.is_valid()) {
+      return GenericData();
```
jonmeow, line 457: FWIW, had you considered using std::optional here and for GetLocalSpecificData? I know we often don't, but kind of wonder if it'd actually be better in these cases, particularly due to the `SmallVector`s

jonmeow (approved):
Awesome! Some comments, but I'm glad to see this working and fairly cleanly.

toolchain/check/import_ref.cpp
```diff
@@ -409,42 +437,154 @@ class ImportRefResolver {
     return context_.inst_blocks().AddCanonical(contents);
   }
 
-  // Gets a local version of an imported generic.
-  auto GetLocalGeneric(SemIR::GenericId generic_id) -> SemIR::GenericId {
+  // Gets an incomplete local version of an imported generic. Most fields are
+  // set in the second pass.
+  auto GetIncompleteLocalGeneric(SemIR::InstId decl_id,
```
jonmeow: FWIW, in other spots we do MakeIncompleteClass. "Get" doesn't imply creation for me.

toolchain/check/generic.cpp
zygoloid: Done, also factored this out since we do the same thing in the previous function.

toolchain/check/generic.cpp
zygoloid: Done. I also added a small size for the map to further reduce allocations, with a TODO to work out a better / less arbitrary one, and I added another `GrowForInsertCount` before the second loop below, because that will also add entries to the map.

toolchain/check/import_ref.cpp

toolchain/check/import_ref.cpp
zygoloid: This one is already in a form that guarantees RVO.

toolchain/check/import_ref.cpp

toolchain/check/import_ref.cpp
zygoloid: Changed to return `specific_id` (just an oversight through refactoring I think).

Setting the definition region conditionally is necessary because we might import the same specific from two IRs, where one of them has a definition and the other does not. The declaration region could be set only the first time, but I think it's reasonable to be doing the check like this, given that it mirrors what we do for the definition.

(I think we might also want to form specifics without even resolving the declaration region at some point. But for now we always resolve the declaration immediately when forming the specific. Conversely, I think we might want to consider lazily importing the specific regions at some point.)

toolchain/check/import_ref.cpp
zygoloid: I hadn't considered that. Thinking about it now, I agree it might avoid some unnecessary work setting up empty `SmallVectors`, but it would mean we no longer have guaranteed RVO.

toolchain/check/import_ref.cpp

PR #4180, opened 2024-08-01
When importing symbolic constants and types, also import the associated generic and index.

zygoloid:
A symbolic constant has an instruction to compute the constant value, as well as potentially also having a generic ID and an index within that generic to indicate where corresponding values can be found in a specific. Import those pieces of information when importing such a constant.

We try to import the generic before we start the main work of importing the constant, and retry the import process if importing the generic adds work to the worklist. This means that the first time we import anything within a generic, we can now perform three passes calling `TryResolveInst` instead of two, but the first pass is very lightweight and only looks up and adds a single instruction, so the added overhead of the extra pass should be minimal.

To avoid introducing cycles when importing a generic function, make the import of a function declaration build the new `Function`, `FunctionDecl`, and `FunctionType` in the first pass, like classes and interfaces do.


toolchain/check/import_ref.cpp
```diff
@@ -131,30 +131,34 @@ auto VerifySameCanonicalImportIRInst(Context& context, SemIR::InstId prev_id,
...
 //    we'll make sure to run it again later.
 //
-// TryResolveInst/TryResolveTypedInst can complete in one call for a given
-// instruction, but should always complete within two calls. However, due to the
-// chance of a second call, it's important to reserve all expensive logic until
-// it's been established that input constants are available; this in particular
-// includes GetTypeIdForTypeConstant calls which do a hash table lookup.
+// TryResolveInst can complete in one call for a given instruction, but should
+// always complete within three calls, and TryResolveTypedInst should always
+// complete within two calls. However, due to the chance of a second call, it's
```
```suggestion
// complete within two calls. However, due to the chance of a later call, it's
```

toolchain/check/import_ref.cpp
```diff
@@ -847,6 +883,39 @@ class ImportRefResolver {
...
   auto TryResolveInst(SemIR::InstId inst_id, SemIR::ConstantId const_id)
       -> ResolveResult {
+    auto inst_const_id = import_ir_.constant_values().Get(inst_id);
+    if (!inst_const_id.is_valid() || !inst_const_id.is_symbolic() ||
+        const_id.is_valid()) {
+      return TryResolveInstCanonical(inst_id, const_id);
+    }
+
+    // Try to import the generic, and retry if it's not ready yet. Note that if
+    // this retries, we can require three passes to import an instruction.
```
jonmeow, line 897: I'm having trouble seeing what the three passes are. Is it because:

1) generic_id needs work
2) TryResolveInstCanonical needs work
3) TryResolveInstCanonical succeeds

Maybe this could be detailed in comments?

Had you considered doing something to interweave TryResolveInstCanonical with the symbolic_const load, in order to avoid the third pass? i.e., moving HasNewWork down:

```
    const auto& symbolic_const =
        import_ir_.constant_values().GetSymbolicConstant(inst_const_id);
    auto initial_work = work_stack_.size();
    auto generic_const_id = GetLocalConstantId(symbolic_const.generic_id);

    // Import the constant and rebuild the symbolic constant data.
    auto result = TryResolveInstCanonical(inst_id, const_id);
      return ResolveResult::Retry();
    }
    if (result.const_id.is_valid()) {
```

(** not actually expecting this to work, but maybe something similar)

jonmeow (approved):

toolchain/check/import_ref.cpp
zygoloid: Clarified in the changes for the other comment.

toolchain/check/import_ref.cpp
zygoloid: Extended the comment to try to make this clearer.

Yes, I considered trying to arrange things so we only ever retry once. The main sticking point is that we don't want to do the expensive work in `TryResolveInstCanonical`  / `TryResolveTypedInst` twice in the case where *it* doesn't add new work, but `TryResolveInst` does add new work.

I think we could get back to doing two passes if we stop computing `initial_work` locally, and instead compute it at the start of each iteration of `Resolve`. We'd still need to address another problem: when `TryResolveTypedInst` returns a `const_id`, we need to preserve that from one call to the next, but need to *not* set that as the constant ID of the instruction when resolving an instruction with a generic constant value (otherwise other resolve steps might pick up and use that incorrect value).

I can look at that if you think it's important to avoid the third pass, but I'd prefer to not do so in this PR. (I think it's *probably* not worth adding much complexity to address this, given that the extra pass added here should be very cheap -- it's only resolving a single constant. But I can take a look and see how complex it is to remove the extra pass.)

zygoloid: Ah, here's why we may need three passes:

In order to break cycles, we need a retry after we build the initial constant value of a class / interface / function and before we import the rest of the declaration. But we can't build that initial constant value until we have imported at least the generic ID. So, in general, we need:

- one pass to find we need to import the generic, import it, and try again
  - then we resolve the generic declaration
- one pass to build the initial constant value (including the generic ID)
  - then we resolve the constants needed to build the full value of the class / function / interface
- one pass to fill in the rest of the details that aren't in the initial constant value

Maybe we can rearrange the logic to avoid the extra pass almost all of the time, though. I think I see how to do that.

jonmeow: FWIW, I think it'd be worth it because I'd expect generic imports to be common. But to be sure, I'd approved because it seemed okay as-is.

zygoloid: Done in #4182. With that, we only have three passes in the case where an instruction that retries with a constant value (`ClassDecl`, `InterfaceDecl`, `FunctionDecl`) also has a symbolic constant value because it's nested in another generic. In all other cases, we'll finish in one or two passes.

PR #4182, opened 2024-08-01
Perform an extra pass to import a generic for a symbolic constant less often.

zygoloid:
Instead of always forcing an extra pass when we need to import a generic
ID for a generic that isn't already imported, attempt to import the
rest of the instruction in the same pass. There are then three
possibilities:

- The instruction needs a retry anyway to form its constant value, and
  we avoid an extra pass.
- The instruction produces its constant value on the first pass but
  still needs a retry. In this case, the handler for that instruction
  is expected to retry itself, before building its constant value. The
  third pass in this case can't be avoided.
- The instruction succeeds on its first pass. We still need an extra
  pass; track the constant produced by resolution separately.


zygoloid:
Depends on #4180

toolchain/check/import_ref.cpp
```diff
@@ -1788,6 +1803,7 @@ class ImportRefResolver {
   SemIR::ImportIRId import_ir_id_;
   const SemIR::File& import_ir_;
   llvm::SmallVector<Work> work_stack_;
+  std::size_t initial_work_ = 0;
```
jonmeow: Where are the semantics of this documented?

Also, [style](https://google.github.io/styleguide/cppguide.html#Integer_Types):

> Prefer to omit the std:: prefix for these types, as the extra 5 characters do not merit the added clutter.

toolchain/check/import_ref.cpp
```diff
@@ -885,27 +903,40 @@ class ImportRefResolver {
...
-    if (result.const_id.is_valid()) {
+    auto result = abstract_const_id.is_valid()
+                      ? ResolveResult{.const_id = abstract_const_id}
+                      : TryResolveInstCanonical(inst_id, const_id);
+
+    if (need_to_import_generic) {
+      if (result.const_id.is_valid()) {
+        CARBON_CHECK(result.state == ResolveResult::Finished)
+            << "Should not retry with a constant result with if we already "
+               "added new work";
```
jonmeow: I'm having trouble reading this, "with ... with if"

toolchain/check/import_ref.cpp
```diff
@@ -194,23 +198,30 @@ class ImportRefResolver {
...
-                   existing.const_id == new_const_id)
-          << "Constant value changed in second pass.";
-      if (!existing.const_id.is_valid()) {
-        SetResolvedConstId(work.inst_id, existing.indirect_insts, new_const_id);
+      initial_work_ = work_stack_.size();
+      auto [new_const_id, state] = TryResolveInst(
+          work.inst_id, existing.const_id, work.abstract_const_id);
+      CARBON_CHECK((state != ResolveResult::Finished) == HasNewWork());
+
+      if (state == ResolveResult::NeedToRebuildConstant) {
```
jonmeow: To note, previously the idea had been that the steps were separated out in the comments above. So to understand "Step 2", I should look under "2. Resolve the instruction". With these changes, I'm not clear that's still the case: it looks like the rebuild behavior has been broken out (actually, I'm not sure exactly where to look to understand the NeedToRebuildConstant behavior, it's not mapping well for me in the comments). You might consider shifting how you're placing comments, maybe putting them closer to where they execute.

jonmeow (dismissed):
The implementation seems fine, but it may be worth thinking a little about how comments are being structured.

zygoloid:
I think this is still subtly wrong.

Suppose we import instruction A, which we can fully import without another pass, except that its constant value is in generic G that we've not imported yet. We'll add G to the work list, which now contains [{A, RebuildConstant, abstract_const_value}, {G}]. Then suppose that importing G causes us to add A back to the work list, and G imports enough to produce a forward declaration. The work list is now [{A, RebuildConstant, abstract_const_value}, {G, Retry}, {A}]. But now when we try to import A again, we won't have the abstract_const_value, so we end up resolving A again, which means that whatever SemIR entities it creates will be created twice.

jonmeow:
To be sure, I don't have a particular answer here. I think the constant would need to be associated with, and queried from, the instruction to determine the current state. I'm not sure whether storing that association would be a net benefit, or if there's some other approach to take. If you want to close the PR that seems fine, particularly since we could consider trade-offs once things are more complete for benchmarking comparisons.

zygoloid:
Revised approach: for a generic constant, always import the generic in the first pass. Update all the `TryResolve` functions to ensure they retry if there is new work before they produce their constant, *even if* they don't add any new work themselves. (They nearly all already checked for new work.)

toolchain/check/import_ref.cpp

The style guide is wrong here -- omitting the `std::` prefix is incorrect if we include `<cstddef>` rather than `<stddef.h>` (which style also recommends) -- but we already do this elsewhere so one more won't hurt :)

toolchain/check/import_ref.cpp
zygoloid: Resolving, this code no longer exists.

toolchain/check/import_ref.cpp
zygoloid: I've tried to make the code comments match the algorithm description a bit better.

toolchain/check/import_ref.cpp
```diff
@@ -125,44 +125,58 @@ auto VerifySameCanonicalImportIRInst(Context& context, SemIR::InstId prev_id,
...
-//    - For a symbolic constant within a generic, find the generic itself. If it
-//      needs to be imported, return Retry() to import the generic before we
-//      import the constant.
-//    - For instructions that can be forward declared, if we don't already have
-//      a constant value from a previous attempt at resolution, start by making
-//      a forward declared constant value to address circular references.
-//    - Gather all input constants.
+//    - If we've processed this work item before, then we now process it again,
+//      because it didn't complete last time, even though we have a constant
+//      value already.
```
```suggestion
//    - If we've processed this work item before, then we now process it again.
//      It didn't complete last time, even though we have a constant value
//      already.
```

Suggesting to have fewer commas per sentence.

toolchain/check/import_ref.cpp
```diff
@@ -125,44 +125,58 @@ auto VerifySameCanonicalImportIRInst(Context& context, SemIR::InstId prev_id,
...
+//
+// 2. Resolve the instruction (TryResolveInst/TryResolveTypedInst), in up to
+//    three passes. First pass:
```
jonmeow: Had you considered structuring this more like:

```
- First pass:
  - ...
  - ...
- Second pass:
  - ...
```

toolchain/check/import_ref.cpp
```diff
@@ -125,44 +125,58 @@ auto VerifySameCanonicalImportIRInst(Context& context, SemIR::InstId prev_id,
...
+//    - If HasNewWork() reports that any work was added, then return
+//      Retry(constant_value): this instruction needs another call to complete.
+//      Gather the now-resolved constants and continue to the next step once the
+//      retry happens.
 //
-// - TryResolveInst can retry once if the generic is not yet loaded.
-// - TryResolveTypedInst can retry once if its inputs are not yet loaded.
-// - The third call should succeed.
+//    Third pass:
+//    - After this point, the constant value for the instruction is already set,
```
jonmeow: What does "after this point" mean? After the third pass? After the second pass?

```suggestion
//    - After the second pass, the constant value for the instruction is already set,
```

toolchain/check/import_ref.cpp
jonmeow: I wonder, is  "pass" the right phrasing here? I think these *can* be a pass, but maybe sub-step phrasing would work better. e.g.:

```suggestion
// 2. Resolve the instruction (TryResolveInst/TryResolveTypedInst). This is done in three sub-steps, each of which can result end in a retry. Whenever one does end in a retry, it will be rerun but always proceed to the next sub-step (up to three passes total, one per sub-step). First sub-step:
```

(note, other pass -> sub-step remaining would probably be appropriate)

toolchain/check/import_ref.cpp
```diff
@@ -125,44 +125,58 @@ auto VerifySameCanonicalImportIRInst(Context& context, SemIR::InstId prev_id,
...
+//    the work item to track what work still needs to be done.
 //
-// Due to the chance of a second call to TryResolveTypedInst, it's important to
-// reserve all expensive logic until it's been established that input constants
-// are available.
+// TryResolveInst can complete in one call for a given instruction. It should
+// always complete within three calls, although it is expected to be rare for
+// three calls to be needed. Due to the chance of a second or third call to
+// TryResolveInst, it's important to reserve all expensive logic until it's
+// been established that input constants are available.
```
jonmeow: Had you considered moving this under step 2 docs, since that's mostly about TryResolveInst behavior?

jonmeow (approved):
LG, glad you thought of a fix. :)

jonmeow (approved):
LG, glad you thought of a fix. :)

jonmeow (approved):
LG, glad you thought of a fix. :)

jonmeow (commented):
LG, glad you thought of a fix. :)

jonmeow (commented):
LG, glad you thought of a fix. :)

toolchain/check/import_ref.cpp
```diff
@@ -125,44 +125,58 @@ auto VerifySameCanonicalImportIRInst(Context& context, SemIR::InstId prev_id,
...
+//    - If HasNewWork() reports that any work was added, then return
+//      Retry(constant_value): this instruction needs another call to complete.
+//      Gather the now-resolved constants and continue to the next step once the
+//      retry happens.
 //
-// - TryResolveInst can retry once if the generic is not yet loaded.
-// - TryResolveTypedInst can retry once if its inputs are not yet loaded.
-// - The third call should succeed.
+//    Third pass:
+//    - After the second pass, the constant value for the instruction is already set,
```
```suggestion
//    - After the second pass, the constant value for the instruction is already
//    set,
```

toolchain/check/import_ref.cpp
zygoloid: I think for the purpose of cross-referencing this from other comments it's useful to use a different word than "step" / "sub-step" (given that we also cross-reference the steps). I've gone with "phase" for now.

toolchain/check/import_ref.cpp
zygoloid: Done, merged into the suggested new text.

toolchain/check/import_ref.cpp

PR #4184, opened 2024-08-02
Initial rough support for deducing generic arguments in a call to a generic function.

zygoloid:


toolchain/check/deduce.cpp
```diff
@@ -0,0 +1,196 @@
...
+#include "toolchain/base/kind_switch.h"
+#include "toolchain/check/context.h"
+#include "toolchain/check/generic.h"
+#include "toolchain/check/subst.h"
+#include "toolchain/sem_ir/typed_insts.h"
+
+namespace Carbon::Check {
+
+namespace {
+struct DeductionWorklist {
```
geoffromer: I think it would help to have a brief comment on this struct, in particular highlighting that this is basically just `deductions` plus some convenience methods for adding items to it. If it's semantically important that `deductions` be processed in stack order, that would be worth mentioning too.

toolchain/check/deduce.cpp
```diff
@@ -0,0 +1,196 @@
...
+      -> void {
+    deductions.push_back(
+        {.param = param, .arg = arg, .needs_substitution = needs_substitution});
+  }
+
+  auto AddBlock(llvm::ArrayRef<SemIR::InstId> params,
+                llvm::ArrayRef<SemIR::InstId> args, bool needs_substitution)
+      -> void {
+    if (params.size() != args.size()) {
+      return;
```
geoffromer, line 47: It surprises me that this isn't an error; it might be worth having a comment to explain it.

toolchain/check/deduce.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+namespace {
+struct DeductionWorklist {
+  auto Add(SemIR::InstId param, SemIR::InstId arg, bool needs_substitution)
+      -> void {
+    deductions.push_back(
+        {.param = param, .arg = arg, .needs_substitution = needs_substitution});
+  }
+
+  auto AddBlock(llvm::ArrayRef<SemIR::InstId> params,
```
geoffromer: Optional: I'd find `AddAll` a slightly more self-explanatory name, because it conveys that we're not somehow adding the block as a unit, we're adding each element individually.

toolchain/check/deduce.cpp
```diff
@@ -0,0 +1,196 @@
...
+                                SemIR::InstBlockId implicit_params_id,
+                                SemIR::InstBlockId params_id,
+                                SemIR::InstId self_id,
+                                llvm::ArrayRef<SemIR::InstId> arg_ids)
+    -> SemIR::SpecificId {
+  DeductionWorklist worklist = {.context = context};
+
+  // TODO: Perform deduction for type of self
+  static_cast<void>(implicit_params_id);
+  static_cast<void>(self_id);
```
geoffromer: Optional: would `[[maybe_unused]]` work instead?

toolchain/check/deduce.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+  // Copy any outer generic arguments from the specified instance and prepare to
+  // substitute them into the function declaration.
+  llvm::SmallVector<SemIR::InstId> results;
+  llvm::SmallVector<Substitution> substitutions;
+  if (enclosing_specific_id.is_valid()) {
```
geoffromer, line 97: If it's not valid, does that mean there was an earlier error and we're proceeding on a best-effort basis, or does it just mean there was no enclosing specific? If the latter, should `enclosing_specific_id` be a `std::optional`? I haven't gotten a handle on when we wrap ID types in `std::optional` vs when we rely on the invalid ID.

toolchain/check/deduce.cpp
```diff
@@ -0,0 +1,196 @@
...
+    -> SemIR::SpecificId {
+  DeductionWorklist worklist = {.context = context};
+
+  // TODO: Perform deduction for type of self
+  static_cast<void>(implicit_params_id);
+  static_cast<void>(self_id);
+
+  // Copy any outer generic arguments from the specified instance and prepare to
+  // substitute them into the function declaration.
+  llvm::SmallVector<SemIR::InstId> results;
```
geoffromer: The name `results` is vague, and makes it a little harder for me to follow the logic. Would something like `generic_arg_ids` work instead?

toolchain/check/deduce.cpp
geoffromer:
```suggestion
  llvm::SmallVector<SemIR::InstId> results;
  llvm::SmallVector<Substitution> substitutions;

  // Copy any outer generic arguments from the specified instance and prepare to
  // substitute them into the function declaration.
  if (enclosing_specific_id.is_valid()) {
```

toolchain/check/deduce.cpp
```diff
@@ -0,0 +1,196 @@
...
+  }
+  auto first_deduced_index = SemIR::CompileTimeBindIndex(results.size());
+
+  worklist.AddBlock(params_id, arg_ids, /*needs_substitution=*/true);
+
+  results.resize(context.inst_blocks()
+                     .Get(context.generics().Get(generic_id).bindings_id)
+                     .size(),
+                 SemIR::InstId::Invalid);
+
```
geoffromer:
```suggestion
  results.resize(context.inst_blocks()
                     .Get(context.generics().Get(generic_id).bindings_id)
                     .size(),
                 SemIR::InstId::Invalid);

  worklist.AddBlock(params_id, arg_ids, /*needs_substitution=*/true);

```

(This is very optional, I just find it slightly clearer to group together the code that's setting up `results` for the main loop)

toolchain/check/deduce.cpp
```diff
@@ -0,0 +1,196 @@
...
+      default:
+        break;
+    }
+  }
+
+  // Check we deduced an argument value for every parameter.
+  for (auto [i, deduced_arg_id] : llvm::enumerate(results)) {
+    if (!deduced_arg_id.is_valid()) {
+      auto binding_id = context.inst_blocks().Get(
```
geoffromer, line 198:
```suggestion
      CARBON_CHECK(i >= first_deduced_index);
      auto binding_id = context.inst_blocks().Get(
```

This diagnostic seems like it would be confusing if it referred to a parameter of the enclosing generic, and presumably that's not supposed to happen anyway.

toolchain/check/testdata/function/generic/redeclare.carbon
```diff
@@ -32,14 +32,10 @@ fn F(T:! type, U:! type) -> T*;
 // CHECK:STDERR: ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 // CHECK:STDERR:
 fn F(T:! type, U:! type) -> U* {
-  // CHECK:STDERR: fail_different_return_type.carbon:[[@LINE+11]]:3: ERROR: Cannot implicitly convert from `T*` to `U*`.
-  // CHECK:STDERR:   return F(T);
-  // CHECK:STDERR:   ^~~~~~~~~~~~
-  // CHECK:STDERR:
-  // CHECK:STDERR: fail_different_return_type.carbon:[[@LINE+7]]:10: ERROR: 1 argument(s) passed to function expecting 2 argument(s).
+  // CHECK:STDERR: fail_different_return_type.carbon:[[@LINE+7]]:10: ERROR: Cannot deduce value for generic parameter `T`.
```
geoffromer, line 37: This seems like a regression: the old errors make sense given how the compiler is resolving `F`, but the new error does not. Is this an artifact of the earlier redeclaration error, or would something like this have the same problem?

```
fn G(T:! type, U:! type) -> T*;
fn F(T:! type, U:! type) -> U* {
  return G(T);
}
```

toolchain/check/testdata/function/generic/deduce.carbon
```diff
@@ -0,0 +1,630 @@
...
+//
+// AUTOUPDATE
+// TIP: To test this file alone, run:
+// TIP:   bazel test //toolchain/testing:file_test --test_arg=--file_tests=toolchain/check/testdata/function/generic/deduce.carbon
+// TIP: To dump output, run:
+// TIP:   bazel run //toolchain/testing:file_test -- --dump_output --file_tests=toolchain/check/testdata/function/generic/deduce.carbon
+
+// --- deduce_explicit.carbon
+
+library "fail_todo_deduce_explicit";
```
geoffromer:
```suggestion
library "deduce_explicit";
```

toolchain/check/testdata/function/generic/deduce.carbon
```diff
@@ -0,0 +1,630 @@
...
+  // CHECK:STDERR: ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+  // CHECK:STDERR:
+  StructParam({.a = 1, .b = 2});
+}
+
+// --- fail_deduce_incomplete.carbon
+
+library "fail_deduce_incomplete";
+
+fn ImplicitNotDeducible[T:! type, U:! type](x: T) -> U;
```
geoffromer, line 95: This is probably future work, but it would really be nice if we could diagnose that `U` isn't deducible for any possible usage, without having to see a usage.

geoffromer (commented):
FYI I haven't reviewed the generated SemIR yet; I'm [looking for guidance](https://discord.com/channels/655572317891461132/655578254970716160/1269002893000704112) about how best to do that.

Also FYI, I'm on vacation next week, so if we don't finish this today you may need to find another reviewer.

toolchain/check/testdata/function/generic/deduce.carbon
```diff
@@ -0,0 +1,630 @@
...
+// --- deduce_explicit.carbon
+
+library "fail_todo_deduce_explicit";
+
+fn ExplicitGenericParam(T:! type) -> T*;
+
+fn CallExplicitGenericParam() -> i32* {
+  return ExplicitGenericParam(i32);
+}
+
```
geoffromer, line 20: Would it be worthwhile to have a test that covers passing a symbolic rather than template constant to a generic parameter?

toolchain/check/testdata/class/generic/method_deduce.carbon
```diff
@@ -0,0 +1,253 @@
...
+// CHECK:STDOUT:
+// CHECK:STDOUT:   class {
+// CHECK:STDOUT:     %Get.decl: @Class.%Get.type (%Get.type.1) = fn_decl @Get [symbolic = %Get (constants.%Get.1)] {
+// CHECK:STDOUT:       %U.loc15_10.1: type = param U
+// CHECK:STDOUT:       %U.loc15_10.2: type = bind_symbolic_name U 1, %U.loc15_10.1 [symbolic = @Get.%U.1 (constants.%U)]
+// CHECK:STDOUT:       %T.ref.loc15: type = name_ref T, file.%T.loc14_13.2 [symbolic = @Get.%T (constants.%T)]
+// CHECK:STDOUT:       %U.ref.loc15: type = name_ref U, %U.loc15_10.2 [symbolic = @Get.%U.1 (constants.%U)]
+// CHECK:STDOUT:       %.loc15_28.1: %.3 = tuple_literal (%T.ref.loc15, %U.ref.loc15)
+// CHECK:STDOUT:       %.loc15_28.2: type = converted %.loc15_28.1, constants.%.4 [symbolic = @Get.%.1 (constants.%.4)]
+// CHECK:STDOUT:       %return.var.loc15: ref @Get.%.1 (%.4) = var <return slot>
```
geoffromer, line 146: Should this `%.4` be `constants.%.4`? If not, I can't find what it's referring to.

geoffromer (commented):
(Finished initial review)

toolchain/check/deduce.cpp
zygoloid: Done. Your comment caused me to realize that I do actually care about the processing order a little bit; changed to process in depth-first order and turned this into a class given that it's now maintaining an invariant.

toolchain/check/deduce.cpp
zygoloid: Added comment.

toolchain/check/deduce.cpp
zygoloid: Done.

toolchain/check/deduce.cpp
zygoloid: Done.

toolchain/check/deduce.cpp
zygoloid: All our `Id` types have an `::Invalid` value that represents "absent", as if there's an implicit `optional` wrapping them all. We use the invalid state to mean "none" pretty heavily.

(For what it's worth, I'm not sure "valid" / "invalid" is the best terminology here, because it has nothing to do with program validity, and we have a separate `Error` value for some of our `Id` types as a placeholder for "couldn't form this value due to an error". Might be worth changing this to `::Absent` and `is_present()` or similar.)

toolchain/check/deduce.cpp
zygoloid: Renamed to `result_arg_ids`. I find it useful for this to capture that it's the output of the algorithm, given that we may also have input generic args from the `enclosing_specific_id`. Could go with `result_generic_arg_ids` if that works better for you, though it's a bit long.

toolchain/check/deduce.cpp
zygoloid: Done. For me this communicates the intent less well (the initial value of `results` and `substitutions` are being computed by the process described by the comment, as if they were return values of that process), but I can see that being subjective. Happy to defer to reviewer's preference :)

toolchain/check/deduce.cpp
zygoloid: Thanks, I agree that's clearer. Done.

toolchain/check/deduce.cpp
zygoloid: Yeah, good point. Changed to only check the deduced arguments.

toolchain/check/testdata/function/generic/redeclare.carbon
zygoloid: This is a consequence of our performing deduction before we try converting the call arguments, and deduction just immediately failing if the arity doesn't match. I'm planning on revisiting this to improve the diagnostic; I've added a TODO to this test for now.

toolchain/check/testdata/function/generic/deduce.carbon
zygoloid: Added a TODO.

toolchain/check/testdata/function/generic/deduce.carbon
zygoloid: Sure, added.

toolchain/check/testdata/class/generic/method_deduce.carbon
zygoloid: Names in SemIR formatted types are by default in the `constants` scope, because type constants are almost always in that scope. (This can be confusing, but it does substantially reduce the verbosity. Maybe we should reconsider, perhaps using a shorter name for `constants`?)

geoffromer: I don't have a strong opinion about how to spell names that refer to constants, but what really threw me off here was the fact that the previous line uses the longer form `constants.%.4`. Is there some significance to that? If not, it would be nice to fix the inconsistency one way or the other, but presumably that's out of scope for this PR.

geoffromer (approved):

PR #4185, opened 2024-08-02
Support 'bazel run //examples:sieve'

jonmeow:
#4076 changed the rule setup and incidentally stopped supporting `bazel run`. This should make `bazel run` work again.


chandlerc (approved):
LG

PR #4186, opened 2024-08-03
change tuple indexing

brymer-meneses:
this changes the tuple index from `tuple[0]` to `tuple.0` in accordance with the accepted propsal https://github.com/carbon-language/carbon-lang/pull/3646


toolchain/check/testdata/index/fail_tuple_large_index.carbon
```diff
@@ -10,15 +10,15 @@
 
 var a: (i32,) = (12,);
 var b: (i32,) = a;
-// CHECK:STDERR: fail_tuple_large_index.carbon:[[@LINE+4]]:14: ERROR: Tuple element index `1` is past the end of type `(i32,)`.
+// CHECK:STDERR: fail_tuple_large_index.carbon:[[@LINE+4]]:14: ERROR: Type `(i32,)` does not support indexing.
 // CHECK:STDERR: var c: i32 = b[1];
 // CHECK:STDERR:              ^~~~
 // CHECK:STDERR:
 var c: i32 = b[1];
```
josh11b: I'm surprised this wasn't changed to `b.1` like other instances

toolchain/check/handle_name.cpp
```diff
@@ -22,6 +23,15 @@ auto HandleParseNode(Context& context, Parse::MemberAccessExprId node_id)
...
     context.node_stack().Push(node_id, member_id);
+  } else if (context.node_stack().PeekIs<Parse::NodeKind::IntLiteral>()) {
+    auto index_inst_id = context.node_stack().PopExpr();
+    auto tuple_inst_id = context.node_stack().PopExpr();
+
+    auto tuple_value_inst_id =
+        PerformTupleIndex(context, node_id, tuple_inst_id, index_inst_id);
+
+    context.node_stack().Push(node_id, tuple_value_inst_id);
+
```
josh11b: For consistency with this code and other Carbon code.
```suggestion
```

toolchain/check/handle_name.cpp
```diff
@@ -52,6 +62,16 @@ auto HandleParseNode(Context& context, Parse::PointerMemberAccessExprId node_id)
...
+  } else if (context.node_stack().PeekIs<Parse::NodeKind::IntLiteral>()) {
+    auto index_inst_id = context.node_stack().PopExpr();
+    auto tuple_pointer_inst_id = context.node_stack().PopExpr();
+    auto tuple_inst_id = PerformPointerDereference(
+        context, node_id, tuple_pointer_inst_id, diagnose_not_pointer);
+    auto tuple_value_inst_id =
+        PerformTupleIndex(context, node_id, tuple_inst_id, index_inst_id);
+
+    context.node_stack().Push(node_id, tuple_value_inst_id);
+
```
josh11b:
```suggestion
```

toolchain/check/member_access.cpp
```diff
@@ -391,7 +391,6 @@ auto PerformMemberAccess(Context& context, Parse::NodeId node_id,
 
   return member_id;
 }
```
josh11b: Should not delete the newline that was there.

```suggestion
}

```

toolchain/check/testdata/index/index_not_literal.carbon
```diff
@@ -9,7 +9,10 @@
 // TIP:   bazel run //toolchain/testing:file_test -- --dump_output --file_tests=toolchain/check/testdata/index/index_not_literal.carbon
 
 var a: (i32, i32) = (12, 34);
-var b: i32 = a[{.index = 1}.index];
+// CHECK:STDERR: index_not_literal.carbon:[[@LINE+3]]:14: ERROR: Member name of type `i32` in compound member access is not an instance member or an interface member.
+// CHECK:STDERR: var b: i32 = a.({.index = 1}.index);
+// CHECK:STDERR:              ^~~~~~~~~~~~~~~~~~~~~~
+var b: i32 = a.({.index = 1}.index);
```
josh11b: This seems like a regression? Shouldn't a constant integer value be accepted in `a.(___)` as long as its constant value is in range?

toolchain/lower/testdata/index/tuple_return_value_access.carbon
```diff
@@ -11,34 +11,8 @@
 fn F() -> (i32, i32) { return (12, 24); }
 
 fn Run() {
+  // CHECK:STDERR: tuple_return_value_access.carbon:[[@LINE+3]]:16: ERROR: Type `(i32, i32)` does not support indexing.
+  // CHECK:STDERR:   var t: i32 = F()[1];
+  // CHECK:STDERR:                ^~~~~~
   var t: i32 = F()[1];
```
josh11b: Why was this not changed?
```suggestion
  var t: i32 = F().1;
```

toolchain/lower/testdata/let/tuple.carbon
```diff
@@ -12,64 +12,8 @@ fn F() -> i32 {
   var a: (i32, i32, i32) = (1, 2, 3);
   var b: (i32, i32) = (4, 5);
   let c: ((i32, i32, i32), (i32, i32)) = (a, b);
-  return c[1][1];
+  // CHECK:STDERR: tuple.carbon:[[@LINE+3]]:10: ERROR: Member name of type `i32` in compound member access is not an instance member or an interface member.
+  // CHECK:STDERR:   return c.1.(1);
+  // CHECK:STDERR:          ^~~~~~~
+  return c.1.(1);
```
josh11b: If the previous code worked, shouldn't this work too?

toolchain/parse/handle_period.cpp
```diff
@@ -23,13 +23,16 @@ static auto HandlePeriodOrArrow(Context& context, NodeKind node_kind,
...
+  } else if (context.ConsumeAndAddLeafNodeIf(Lex::TokenKind::IntLiteral,
+                                             NodeKind::IntLiteral)) {
     // OK, `.base`.
   } else if (paren_state != State::Invalid &&
              context.PositionIs(Lex::TokenKind::OpenParen)) {
     state.state = paren_state;
     context.PushState(state);
     context.PushState(State::OnlyParenExpr);
     return;
+
```
josh11b:
```suggestion
```

toolchain/check/member_access.cpp
```diff
@@ -429,4 +428,81 @@ auto PerformCompoundMemberAccess(Context& context, Parse::NodeId node_id,
...
+                                                {.type_id = element_type_id,
+                                                 .tuple_id = tuple_inst_id,
+                                                 .index_id = index_inst_id});
+    }
+    default: {
+      if (tuple_type_id != SemIR::TypeId::Error) {
+        CARBON_DIAGNOSTIC(
+            TupleWrongIndexSyntax, Error,
+            "Type `{0}` does not support array indexing. Did you mean tuple.0?",
+            SemIR::TypeId);
```
josh11b: Should add a test that verifies that this diagnostic is being produced correctly. Also, isn't this the opposite case, where you are using tuple indexing on a non-tuple value?

toolchain/check/member_access.cpp
```diff
@@ -429,4 +428,81 @@ auto PerformCompoundMemberAccess(Context& context, Parse::NodeId node_id,
...
+      index_inst_id = ConvertToValueOfType(
+          context, index_node_id, index_inst_id,
+          context.GetBuiltinType(SemIR::BuiltinInstKind::IntType));
+      auto index_const_id = context.constant_values().Get(index_inst_id);
+      if (index_const_id == SemIR::ConstantId::Error) {
+        index_inst_id = SemIR::InstId::BuiltinError;
+      } else if (!index_const_id.is_template()) {
+        // TODO: Decide what to do if the index is a symbolic constant.
+        CARBON_DIAGNOSTIC(TupleIndexNotConstant, Error,
+                          "Tuple index must be a constant.");
```
josh11b: Is this error still produced? Looks like the test that used to trigger this case no longer does.

toolchain/check/member_access.cpp
```diff
@@ -429,4 +428,81 @@ auto PerformCompoundMemberAccess(Context& context, Parse::NodeId node_id,
...
+        context.emitter().Emit(node_id, TupleIndexNotConstant);
+        index_inst_id = SemIR::InstId::BuiltinError;
+
```
josh11b:
```suggestion
```

toolchain/check/member_access.cpp
```diff
@@ -391,7 +391,6 @@ auto PerformMemberAccess(Context& context, Parse::NodeId node_id,
 
   return member_id;
 }
-
 auto PerformCompoundMemberAccess(Context& context, Parse::NodeId node_id,
```
josh11b: Shouldn't this be changed to support expressions that evaluate to a constant integer? This is where I would expect things like `t.(1)` to be supported.

josh11b (commented):
Some initial feedback. I'll look again once these are addressed.

toolchain/check/member_access.cpp
brymer-meneses: > Also, isn't this the opposite case, where you are using tuple indexing on a non-tuple value?

My bad, for some reason I was thinking of `tuple[0]`. Should we add a note telling the user the right way to index a tuple?

toolchain/check/member_access.cpp
```diff
@@ -410,6 +409,8 @@ auto PerformCompoundMemberAccess(Context& context, Parse::NodeId node_id,
           member.type_id())) {
     member_id = PerformImplLookup(context, node_id, base_type_const_id,
                                   *assoc_type, member_id);
+  } else if (context.types().Is<SemIR::TupleType>(base_type_id)) {
```
brymer-meneses: I'm not sure if this is the right way to  disambiguate between other stuff and `tuple.(indexConstant)`.

toolchain/check/member_access.cpp
brymer-meneses: this should be resolved by now

toolchain/sem_ir/typed_insts.h
```diff
@@ -967,10 +968,10 @@ struct TupleAccess {
   ElementIndex index;
 };
 
-// Access to a tuple member by index, such as `tuple[index]`.
+// Access to a tuple member by index, such as `tuple.0`.
 struct TupleIndex {
-  static constexpr auto Kind = InstKind::TupleIndex.Define<Parse::IndexExprId>(
-      {.ir_name = "tuple_index"});
+  static constexpr auto Kind =
+      InstKind::TupleIndex.Define<Parse::NodeId>({.ir_name = "tuple_index"});
```
brymer-meneses, line 975: This is also a hack. I'm not sure if this is okay.

toolchain/check/member_access.cpp
brymer-meneses: Yea, this fails, and results to a compiler crash. 

```
fn Main() {
  var non_tuple: [i32; 1] = (1,);
  var first: i32 = non_tuple.0;
}
```

EDIT: this is an unrelated issue, and i think i just fixed this https://github.com/carbon-language/carbon-lang/pull/4186/commits/133ad706654cf9d58bccee0df0145e5b73df40f5.

toolchain/check/member_access.cpp
brymer-meneses: this should be resolved by now

brymer-meneses:
I think everything should be resolved by now, except the parts wherein I am confused on what to do.

brymer-meneses:
Also, apologies if I'm unnecessarily pinging you by re-requesting a review even if you said that you'd review this once everything gets resolved.

toolchain/check/testdata/tuples/fail_non_tuple_index.carbon
```diff
@@ -0,0 +1,87 @@
...
+// TIP:   bazel test //toolchain/testing:file_test --test_arg=--file_tests=toolchain/check/testdata/tuples/fail_non_tuple_index.carbon
+// TIP: To dump output, run:
+// TIP:   bazel run //toolchain/testing:file_test -- --dump_output --file_tests=toolchain/check/testdata/tuples/fail_non_tuple_index.carbon
+
+fn Main() {
+  var non_tuple: [i32; 1] = (1,);
+  // CHECK:STDERR: fail_non_tuple_index.carbon:[[@LINE+3]]:20: ERROR: Type `[i32; 1]` does not support tuple indexing. Only tuples can be indexed that way.
+  // CHECK:STDERR:   var first: i32 = non_tuple.0;
+  // CHECK:STDERR:                    ^~~~~~~~~~~
+  var first: i32 = non_tuple.0;
```
jonmeow: It looks like the other tuple indexing tests are under `testdata/index`. Might that be a better place for this test?

toolchain/lower/testdata/index/array_element_access.carbon
```diff
@@ -13,7 +13,7 @@ fn B() -> [i32; 2] { return (1, 2); }
 
 fn Run() {
   var a: [i32; 2] = A();
-  var b: i32 = A()[0];
+  var b: i32 = A().0;
```
jonmeow: This should be focused on testing array indexing. I see this is a tuple, so how about:

```suggestion
  var b: i32 = 1;
```

This way, there's still a non-constant index for `c`.

toolchain/parse/testdata/basics/fail_invalid_designators.carbon
```diff
@@ -15,14 +15,10 @@ fn F() {
...
-  // CHECK:STDERR: fail_invalid_designators.carbon:[[@LINE+4]]:5: ERROR: Expected identifier after `.`.
+  // CHECK:STDERR: fail_invalid_designators.carbon:[[@LINE+3]]:5: ERROR: Expected identifier after `.`.
   // CHECK:STDERR:   a.fn;
   // CHECK:STDERR:     ^~
-  // CHECK:STDERR:
   a.fn;
-  // CHECK:STDERR: fail_invalid_designators.carbon:[[@LINE+3]]:5: ERROR: Expected identifier after `.`.
-  // CHECK:STDERR:   a.42;
-  // CHECK:STDERR:     ^~
   a.42;
```
jonmeow: This test is for invalid designators, and is expected to fail. As a consequence, if this started failing, the test would still basically pass.

What do you think of separating `a.42` out to a separate parse test, so that it's tested that this parses without an error? e.g., adding a test under parse/testdata/index for the tuple index case.

toolchain/check/testdata/index/fail_negative_indexing.carbon
```diff
@@ -9,10 +9,10 @@
...
 
 var a: (i32, i32) = (12, 6);
-// CHECK:STDERR: fail_negative_indexing.carbon:[[@LINE+3]]:16: ERROR: Cannot access member of interface Negate in type i32 that does not implement that interface.
-// CHECK:STDERR: var b: i32 = a[-10];
-// CHECK:STDERR:                ^~~
-var b: i32 = a[-10];
+// CHECK:STDERR: fail_negative_indexing.carbon:[[@LINE+3]]:17: ERROR: Cannot access member of interface Negate in type i32 that does not implement that interface.
+// CHECK:STDERR: var b: i32 = a.(-10);
+// CHECK:STDERR:                 ^~~
+var b: i32 = a.(-10);
```
jonmeow: It looks like several of these tests weren't expressly for tuples (versus ones that have array or tuple in the name, such as fail_tuple_index_error). I expect they were intended to cover indexing in general. I think that made sense when both used the same codepath, but now that tuple indexing is splitting out from array indexing more completely, should we have tests that array indexing has corresponding errors?

toolchain/parse/handle_period.cpp
```diff
@@ -23,6 +23,8 @@ static auto HandlePeriodOrArrow(Context& context, NodeKind node_kind,
     // OK, `.` identifier.
   } else if (context.ConsumeAndAddLeafNodeIf(Lex::TokenKind::Base,
                                              NodeKind::BaseName)) {
+  } else if (context.ConsumeAndAddLeafNodeIf(Lex::TokenKind::IntLiteral,
+                                             NodeKind::IntLiteral)) {
     // OK, `.base`.
```
jonmeow, line 29: This comment looks like it should be associated with the "Base" case above. Maybe it'd be good to add a similar comment for IntLiteral, e.g.:

```
// OK, `.42`.
```

toolchain/parse/typed_nodes.h
```diff
@@ -1110,6 +1110,9 @@ struct StructTypeLiteral {
   Lex::CloseCurlyBraceTokenIndex token;
 };
 
+using IntLiteral = LeafNode<NodeKind::IntLiteral, Lex::IntLiteralTokenIndex,
+                            NodeCategory::Expr | NodeCategory::IntConst>;
```
jonmeow: We've been trying to cluster node kinds in this file. This section is really for struct-related nodes. Can you move this to the "Literals, operators, and modifiers", around line 925, maybe putting it after the `#include` there (before `extern`)? That's where other similar literals are also defined.

toolchain/sem_ir/typed_insts.h
```diff
@@ -967,10 +968,10 @@ struct TupleAccess {
   ElementIndex index;
 };
 
-// Access to a tuple member by index, such as `tuple[index]`.
+// Access to a tuple member by index, such as `tuple.0`.
 struct TupleIndex {
-  static constexpr auto Kind = InstKind::TupleIndex.Define<Parse::IndexExprId>(
-      {.ir_name = "tuple_index"});
+  static constexpr auto Kind =
```
jonmeow, line 974:
```suggestion
  // TODO: Make Parse::NodeId more specific.
  static constexpr auto Kind =
```

toolchain/sem_ir/typed_insts.h
jonmeow: This should be fine, added a suggestion for a TODO similar to what we have elsewhere (ArrayIndex is similar).

toolchain/check/member_access.h
```diff
@@ -23,6 +23,10 @@ auto PerformCompoundMemberAccess(Context& context, Parse::NodeId node_id,
                                  SemIR::InstId base_id,
                                  SemIR::InstId member_expr_id) -> SemIR::InstId;
 
+auto PerformTupleIndex(Context& context, Parse::NodeId node_id,
```
jonmeow, line 28: We try to keep comments on functions. Here's a suggestion similar to other functions in this file:

```suggestion
// Creates SemIR to perform a tuple index with base expression `tuple_inst_id` and
// index expression `index_inst_id`. Returns the result of the access.
auto PerformTupleIndex(Context& context, Parse::NodeId node_id,
```

toolchain/check/handle_name.cpp
```diff
@@ -22,6 +23,14 @@ auto HandleParseNode(Context& context, Parse::MemberAccessExprId node_id)
     auto member_id =
         PerformCompoundMemberAccess(context, node_id, base_id, member_expr_id);
     context.node_stack().Push(node_id, member_id);
+  } else if (context.node_stack().PeekIs<Parse::NodeKind::IntLiteral>()) {
```
jonmeow: Here and below, I think the repeated `PeekIs` (for both ParenExpr and IntLiteral) is going to lead to the node kind being queried. To avoid the duplicate logic, can you store the valid of the kind and compare? i.e., something like:

```
auto node_kind = context.node_stack().PeekNodeKind();
if (node_kind == Parse::NodeKind::ParenExpr) {
  ...
} else if (node_kind == Parse::NodeKind::IntLiteral) {
```

toolchain/check/member_access.cpp
jonmeow: I understand the desire to use `types().Is` here, but I'm going to suggest making use of the already-fetched type constant:

```suggestion
  } else if (context.constant_values().GetAsInst(base_type_const_id).Is<SemIR::TupleType>()) {
```

toolchain/check/member_access.cpp
```diff
@@ -429,4 +430,74 @@ auto PerformCompoundMemberAccess(Context& context, Parse::NodeId node_id,
...
+                                               .index_id = index_inst_id});
+  } else {
+    CARBON_DIAGNOSTIC(TupleIndexOnANonTupleType, Error,
+                      "Type `{0}` does not support tuple indexing. Only "
+                      "tuples can be indexed that way.",
+                      SemIR::TypeId);
+    context.emitter().Emit(node_id, TupleIndexOnANonTupleType, tuple_type_id);
+  }
+
+  return SemIR::InstId::BuiltinError;
```
jonmeow: We tend to favor early returns to reduce nesting. Here I think you could do something like:

```suggestion
  auto tuple_type =
      context.types().TryGetAs<SemIR::TupleType>(tuple_type_id));
  if (!tuple_type) {
    CARBON_DIAGNOSTIC(TupleIndexOnANonTupleType, Error,
                      "Type `{0}` does not support tuple indexing. Only "
                      "tuples can be indexed that way.",
                      SemIR::TypeId);
    context.emitter().Emit(node_id, TupleIndexOnANonTupleType, tuple_type_id);
    return SemIR::InstId::BuiltinError;
  }
  
  SemIR::TypeId element_type_id = SemIR::TypeId::Error;
  auto index_node_id = context.insts().GetLocId(index_inst_id);
  index_inst_id = ConvertToValueOfType(
      context, index_node_id, index_inst_id,
      context.GetBuiltinType(SemIR::BuiltinInstKind::IntType));
  auto index_const_id = context.constant_values().Get(index_inst_id);
  if (index_const_id == SemIR::ConstantId::Error) {
    index_inst_id = SemIR::InstId::BuiltinError;
  } else if (!index_const_id.is_template()) {
    // TODO: Decide what to do if the index is a symbolic constant.
    CARBON_DIAGNOSTIC(TupleIndexNotConstant, Error,
                      "Tuple index must be a constant.");
    context.emitter().Emit(node_id, TupleIndexNotConstant);
    index_inst_id = SemIR::InstId::BuiltinError;
  } else {
    auto index_literal = context.insts().GetAs<SemIR::IntLiteral>(
        context.constant_values().GetInstId(index_const_id));
    auto type_block = context.type_blocks().Get(tuple_type->elements_id);
    if (const auto* index_val = ValidateTupleIndex(
            context, node_id, tuple_inst, index_literal, type_block.size())) {
      element_type_id = type_block[index_val->getZExtValue()];
    } else {
      index_inst_id = SemIR::InstId::BuiltinError;
    }
  }

  return context.AddInst<SemIR::TupleIndex>(node_id,
                                            {.type_id = element_type_id,
                                             .tuple_id = tuple_inst_id,
                                             .index_id = index_inst_id});  
```

Note, this also couple the `BuiltinError` return with the diagnostic, and I think it helps make it clearer that the other path always has a valid return.

jonmeow (commented):
Thanks, this looks really good! Josh had asked me to take a look, and I think the logic looks correct. I did see some spots though where I think it may be helpful to adjust API use, just subtle toolchain things.

jonmeow:
As a sidenote (not something I'm requesting for this PR), I'm thinking we may want to split up the `index` directory to reflect the logic split. Maybe leaving the `IndexWith`-y things in `index`, and moving the tuple indexing out to something like `tuples/access`. (which, is now making me notice that odd plural of "tuples": #4200)

This is partly because I feel odd seeing so many files with either `array` or `tuple` in the name, it seems like it'd be nice to have that implicit from the directory name. You can do this here if you want, but I'm happy to pick it up after you're merged since it feels like scope creep here.

toolchain/check/member_access.cpp
josh11b: That would be great, but I wouldn't say it is required.

brymer-meneses:
> You can do this here if you want, but I'm happy to pick it up after you're merged since it feels like scope creep here.

I'll happily do this in a separate PR, as I do not want to clutter this PR by syncing to trunk (to have https://github.com/carbon-language/carbon-lang/pull/4200).

toolchain/check/testdata/tuples/fail_non_tuple_index.carbon
brymer-meneses: I didn't notice that there was a similar test named `toolchain/check/testdata/fail_non_tuple_access`. I think I'll just merge this with that.

toolchain/check/member_access.cpp
brymer-meneses: It seems like the method `GetAsInst` does not exist in `ConstantValueStore`. I can try implementing it, I just need pointers on how.

brymer-meneses:
On a related note, do you think we should rename tuple index with tuple access across the repo? That would mean renaming a lot of things ( `SemIR`, `typed_nodes.h`) and stuff.

jonmeow:
> On a related note, do you think we should rename tuple index with tuple access across the repo? That would mean renaming a lot of things ( `SemIR`, `typed_nodes.h`) and stuff.

Asking for thoughts [on discord](https://discord.com/channels/655572317891461132/655578254970716160/1271195835975204946). Maybe best to split out any rename.

jonmeow:
Per discussion, we'll want to consolidate `TupleIndex` into the already-existing `TupleAccess`. They're essentially equivalent, and that'd align with how `StructAccess` works: using the same instruction for both implicit and explicit access. [ed: note this too might be better in a different PR, but I'm not sure how big the delta would be]

toolchain/check/member_access.cpp
brymer-meneses: @jonmeow sorry for the ping, i think you might have overlooked this

toolchain/check/testdata/tuples/fail_non_tuple_index.carbon
jonmeow: Had you meant to remove this?

toolchain/parse/testdata/index/tuple_index.carbon
```diff
@@ -0,0 +1,42 @@
...
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+// TIP: To test this file alone, run:
+// TIP:   bazel test //toolchain/testing:file_test --test_arg=--file_tests=toolchain/parse/testdata/tuple/index.carbon
+// TIP: To dump output, run:
+// TIP:   bazel run //toolchain/testing:file_test -- --dump_output --file_tests=toolchain/parse/testdata/tuple/index.carbon
+
+let tuple: (i32, i32) = (5, 5);
```
jonmeow:
```suggestion
```

This is really small (I'm mainly mentioning it because of fail_non_tuple_index.carbon), but I think it's helpful to both emphasize the `tuple.0` parse tree, and as a hint that this line isn't needed (name lookup doesn't occur during parse). So the parse tree below ends up a lot shorter, making it easier to see the syntax under test.

jonmeow (approved):
Thanks for all the changes! Basically looks good, but I think you might've missed deleting the fail_non_tuple_index.carbon. Since there's fail_non_tuple_access (as you observed), can you delete that and I'll merge?

jonmeow:
> Thanks for all the changes! Basically looks good, but I think you might've missed deleting the fail_non_tuple_index.carbon. Since there's fail_non_tuple_access (as you observed), can you delete that and I'll merge?

Actually, it looks like things have gotten slightly out of date due to a LLVM IR change (#4212 looks like the source of conflicts). I'm happy to have you merge/autoupdate, or I can take care of that (let me know how you'd prefer).

brymer-meneses:
I think, I might have messed things up when trying to sync.

PR #1526, opened 2022-07-21
Documentation website

emillaine:
Closes #1492. Adds a website that is auto-generated from the markdown files in the repo, using [Docusaurus](https://docusaurus.io/docs). This will make it easier to browse the docs, by adding a navigation sidebar as well as a table-of-contents sidebar, mobile support, and other things. The site can be autodeployed to https://carbon-language.github.io/ (or to a custom domain using a CNAME file), and is currently configured accordingly (deployment instructions in [website/README.md](https://github.com/carbon-language/carbon-lang/pull/1526/files#diff-0eb684c3ea7be1b3de6b0b158c7059da05dff87877c4c40d32ea3a8eaef7a75eR43-R53)).

- The front page is generated from the repository root readme.
- The top navigation items Design, Project, Guides, and Spec corresponds to the respective directories under `docs/`.
- Footer has links to GitHub repo, Discord, and Code of Conduct.

Preview: https://emlai.github.io/

To do:
- [ ] Render [mermaid diagrams](https://github.com/carbon-language/carbon-lang/pull/1526#issuecomment-1227782614). 
- [ ] Before or after merging, for someone with admin access to the organization: Create carbon-language.github.io repo, enable GitHub Pages for it, and configure deploy keys ([instructions](https://github.com/carbon-language/carbon-lang/pull/1526/files#diff-0eb684c3ea7be1b3de6b0b158c7059da05dff87877c4c40d32ea3a8eaef7a75eR43-R53)). You can ping me (emlai#9471) on Discord if you need help.

Follow-up improvements (non-blocking tasks that are not necessary for merging this PR):
- [ ] Mention in docs that they get auto-deployed to the website when a change is pushed to trunk.
- [ ] Add site-wide search such as [Algolia](https://docsearch.algolia.com/apply/) which will scrape the site and host the search index in their own database. Requires that the site is deployed first before setting up.
- [ ] Fix broken links and enable Docusaurus link validation ([src](https://github.com/carbon-language/carbon-lang/pull/1526/files#diff-bc4913d026ae51bef2bc8b7a1c040fdf52b999d99e6fa40d87b379f98a60e0eaR16)), blocked by Docusaurus issue: https://github.com/facebook/docusaurus/issues/7829
- [ ] Fix description meta tags, currently `<!--`: `<meta data-rh="true" name="description" content="<!--">`

## How to test

Go to https://emlai.github.io/ and check that links, navigation, etc. works, and everything looks ok.

Issue #1492, opened 2022-07-21
Need a website for documentation


yadunandanbhat:
Can we get a simple readthedocs type of website for the carbon's documentation. Because its hard to navigate in github and the readability is not very good.

Its easy to convert the existing readme to documentation website using tools like mkdocs, kramdown, jekyll etc!

emillaine:
I made a simple script for converting the markdown files under `docs/` to HTML using `pandoc`: https://github.com/emlai/carbon-lang/commit/200eeb3de40303e11aa0a278008eadae71bd031c

You can see the result here: https://emlai.github.io/

It's still missing styles and a navigation bar, but the links seem to be working correctly.

Nike682631:
I'd be happy to work on this issue if it's open for anyone

emillaine:
I updated [my branch](https://github.com/emlai/carbon-lang/tree/website) to use mkdocs after I got it to handle absolute markdown links correctly. It adds a theme, a navbar, and a search, so it looks much better now: https://emlai.github.io/

There's still stuff to do:

- It would be better to show the repository root readme on the front page.
- Create a GitHub action for auto-deploying the site.
- Organize the code snippets in a better way. Currently they're under `Images > Front page snippets for Carbon`.
- Replace the logo with the Carbon logo.
- Customize the theme to make it look less stock and more "Carbony".
- etc.

@Nike682631 If you want to help, we can sync on the Discord!

Nike682631:
@emlai sure, what's your discord?

emillaine:
@Nike682631 emlai#9471

yadunandanbhat:
great work @emlai !!!!

seanmrnda:
We need a book, similar to rust book.

joaodematejr:
recommendation to be studied => https://docusaurus.io/docs

seanmrnda:
Best so far, convertible to pdf.

https://github.com/rust-lang/mdbook

jonmeow:
If you're working on a GitHub pages PR, one thing we've been cautious of is developers interpreting a documentation website as implicit encouragement to use Carbon in production. However, hopefully it's clear to most that Carbon isn't there.

To address that concern, can you maybe add a reminder at the top that Carbon is experimental and not intended for use in applications, linking to https://github.com/carbon-language/carbon-lang/blob/trunk/docs/project/faq.md#how-soon-can-we-use-carbon

emillaine:
[...]

PR #4189, opened 2024-08-05
Build a website.

jonmeow:
Demo site: https://jonmeow.carbon-lang.dev/

I'm trying to keep work under the `/website` subdirectory so that the misc files don't interfere with unrelated views of the repository. The `prebuild.py` script does some work to move things around and add frontmatter, helping the jekyll generation.

I'm using the "just-the-docs" theme because I think it's a decent match for what we want, and getting jekyll up and running with it wasn't too difficult. Note #1526 proposed using Docusaurus; I started out there, but was having trouble getting it working with newer versions. The plugins in particular I got stuck trying to make work, which sent me looking for options that we could have working with less customization. I do lean towards jekyll though, because it's what GH uses so hopefully we can get a more consistent experience.

Having a website has been approved for a while under #1492, but hasn't been a priority. I'm mainly doing this because I want to just be able to point people to carbon-lang.dev and have easy links that way.


jonmeow:
NB, https://www.carbon-lang.dev/ should be configured, I *think* this will be effective once committed.

.github/workflows/gh_pages_ci.yaml
```diff
@@ -0,0 +1,41 @@
...
+        run: ./website/prebuild.py
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@a6e6f86333f0a2523ece813039b8b4be04560854 # v1.190.0
+        with:
+          # Runs 'bundle install' and caches installed gems automatically.
+          bundler-cache: true
+          # Increment this number if you need to re-download cached gems.
+          cache-version: 0
+      - name: Build with Jekyll
+        run: bundle exec jekyll build
```
zygoloid, line 41: Mostly just thinking out loud here in case this sparks an idea:

It looks like this takes ~30s. That seems a bit wasteful for PRs that don't change anything about documentation (which I'd expect to be the common case), so if there's an easy way to turn this off for changes that obviously don't affect documentation that might be worth pursuing. But maybe it's not enough to be worth worrying about -- this is certainly a lot less compute resource than we're spending on running unit tests -- and in any case we don't need that in the first iteration.

website/prebuild.py
```diff
@@ -0,0 +1,165 @@
...
+
+    # Use an object for a reference.
+    nav_order = [0]
+
+    # Returns an incrementing value for ordering.
+    def next(nav_order: list[int]) -> int:
+        nav_order[0] += 1
+        return nav_order[0]
+
+    label_root_file("README.md", "README", next(nav_order))
```
zygoloid:
```suggestion
    label_root_file("README.md", "Home", next(nav_order))
```
or similar? Calling the front page "README" in the navbar seems surprising.

website/prebuild.py
```diff
@@ -0,0 +1,165 @@
...
+    label_root_file("CONTRIBUTING.md", "Contributing", next(nav_order))
+    label_subdir("docs/design", next(nav_order), True)
+    label_subdir("docs/guides", next(nav_order))
+    label_subdir("docs/project", next(nav_order), True)
+    label_subdir("docs/spec", next(nav_order))
+    label_subdir("toolchain", next(nav_order))
+    label_subdir("explorer", next(nav_order))
```
zygoloid: I wonder if it'd be worth nesting these under an `Implementations` section. Or alternatively, hiding the explorer docs entirely.

website/prebuild.py
```diff
@@ -0,0 +1,165 @@
...
+    label_subdir("testing", next(nav_order))
+    label_subdir("utils", next(nav_order))
+    label_subdir("proposals", next(nav_order))
```
zygoloid, line 186: Can we hide the files in subdirectories of "proposals"? There are three non-proposal files from subdirectories showing up here right now.

website/prebuild.py
```diff
@@ -0,0 +1,165 @@
...
+        child_content = child.read_text()
+        child_title = get_title(child, child_content)
+        child_nav_order = None
+
+        if subdir_str == "proposals":
+            # Use proposal numbers as part of the title and ordering.
+            m = re.match(r"p(\d+).md", child.name)
+            if m:
+                child_title = f"#{m[1]}: {child_title}"
+                child_nav_order = int(m[1])
```
zygoloid:
```suggestion
        if subdir_str == "proposals":
            # Use proposal numbers as part of the title and ordering.
            # Skip files that aren't proposals.
            m = re.match(r"p(\d+).md", child.name)
            if !m:
                continue
            child_title = f"#{m[1]}: {child_title}"
            child_nav_order = int(m[1])
```

website/prebuild.py
zygoloid: Added suggestion.

website/prebuild.py
```diff
@@ -0,0 +1,165 @@
...
+def get_title(f: Path, content: str) -> str:
+    """Returns a file's title according to markdown.
+
+    Replacements are for YAML compatibility in `add_frontmatter`.
+    """
+    m = re.search("^# (.*)$", content, re.MULTILINE)
+    assert m, f
+    title = m[1]
+    title = title.replace("\\", "\\\\")
+    title = title.replace('"', '\\"')
```
zygoloid, line 39: It'd be good to also convert `` ` ``s appearing in page titles into `<code>` ... `</code>`; we have a few of those in the design.

zygoloid (commented):
The mermaid diagrams are unreadable in dark theme. I wonder if there's some way we can theme mermaid so it uses better colors in dark theme. Alternatively, emlai's website had a light mode / dark mode toggle -- I wonder if we could add something similar easily. But I think this shouldn't be a blocking concern; we can iterate on this.

.github/workflows/gh_pages_ci.yaml
jonmeow: My sense was that we'd more likely regret something like https://jonmeow.carbon-lang.dev/toolchain/check/check.cpp being out-of-date than regret the compute cost. The documentation does link source files.

website/prebuild.py
jonmeow: Done

website/prebuild.py
jonmeow: Done

website/prebuild.py
jonmeow: Ack

website/prebuild.py
jonmeow: I'm hesitant to hide the explorer stuff until it's well and truly gone. Adding an "Implementation"

.github/workflows/gh_pages_ci.yaml
zygoloid: That seems like a great rationale, thanks. And yeah, the compute cost is relatively minor compared to our other checks. It might be the long pole in changes that *don't* touch source code, but those changes are the ones that really need this checking anyway!

website/prebuild.py
jonmeow: Done

jonmeow (commented):
Note, also added some overflow-y css for the nav bar, which was bugging me.

zygoloid:
Looks good to me, I also asked @wolffg to take a quick look.

brymer-meneses:
> Alternatively, emlai's website had a light mode / dark mode toggle -- I wonder if we could add something similar easily. But I think this shouldn't be a blocking concern; we can iterate on this.

We can make use of something like [vitepress](https://vitepress.dev/) to achieve this. It also uses markdown so there would be minimal configuration needed, but I'm not sure about the added complexity of Node and stuff.

jonmeow:
> > Alternatively, emlai's website had a light mode / dark mode toggle -- I wonder if we could add something similar easily. But I think this shouldn't be a blocking concern; we can iterate on this.

To be sure, in case you overlooked this in the PR description, I was having trouble getting emlai's website up-to-date for dependencies.

> We can make use of something like [vitepress](https://vitepress.dev/) to achieve this. It also uses markdown so there would be minimal configuration needed, but I'm not sure about the added complexity of Node and stuff.

A quick look at vitepress (installing it, trying some configuration) suggests it's going to require work similar to Hugo/Docusaurus.

If someone wants to pick up getting everything working great, let me know and I can set this PR aside. I'm just looking for something that looks good and is easy for me to maintain, without spending much more time on it. I'm seeing if we can have something simple for now, leaving more work for later.

p0358:
The contrast feels way too low for comfortable reading, the text should be way more black in the light theme. (it seems to fail the WCAG AAA guidelines for *minimum* contrast ratio too)

jonmeow:
> The contrast feels way too low for comfortable reading, the text should be way more black in the light theme. (it seems to fail the WCAG AAA guidelines for _minimum_ contrast ratio too)

Can you please provide specific suggestions (e.g., proposed CSS) about the changes you are suggesting? I'm not familiar with WCAG AAA guidelines or how to identify/fix the issue you're referring to.

To add, I'm aware of specifically comments in the code blocks not having enough contrast. But AFAIK the rest is fine for contrast.

That said, this PR is currently paused while I try to figure out whether I can actually proceed with this (there is discussion occurring not on this thread), so fixing contrast is something I'd prioritize more once/if I confirm I can proceed.

jonmeow:
Per discussion with @wolffg and @chandlerc:
- Fix the comment color.
- Remove the "Carbon is not ready for use." javascript, it's feeling redundant/slow to people.
- Need to switch to docs.carbon-lang.dev

chandlerc:
Just to summarize a few more things from the discussion (but let me know if I misunderstood any of it)...

- We do want to make sure that if we're publishing a website it has good accessibility.
- But our goal is just to quickly get technical docs to render on a website, so we're hoping to avoid sinking too much time into debugging stuff and looking for hopefully quick/easy fixes.
- Ultimately, this isn't going to be a real or full Carbon website -- the project isn't at that stage yet. This is just about having a website for our technical docs rendered in a way that can be navigated a bit more easily than the markdown preview on GitHub.

-Chandler

jonmeow:
Note, I've done a few more edits to adjust text colors. I believe the css issues specific to the this change have been addressed. (Specifically, [lighthouse](https://developer.chrome.com/docs/lighthouse/overview) complains about things, but not contrast)

website/_includes/nav_footer_custom.html
```diff
@@ -0,0 +1,16 @@
...
+<!--
+Used from:
+https://github.com/just-the-docs/just-the-docs/blob/main/_includes/components/sidebar.html
+-->
+<footer class="site-footer">
+  This site uses
+  <a href="https://github.com/just-the-docs/just-the-docs">Just the Docs</a>.
+  <br />
+  Hosted on GitHub Pages.
+</footer>
```
chandlerc: Could we include copyright and license info here? seems like that would have it show up on every page. Unsure if there is a custom-footer option for the pages themselves...

website/_config.yml
```diff
@@ -0,0 +1,36 @@
+# Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+# Exceptions. See /LICENSE for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+title: Carbon Language
```
chandlerc: I'm guessing this is where the top-left corner "Carbon Language" comes from?

Wondering if we could say "Carbon Language Documentation" without it ruining the layout of the page... Or in some other way make it more clear that this isn't intended to be the long-term website for the project or language, but a build and render of our current documentation.

I can try getting this to render locally and play with some options if useful.

chandlerc (commented):
Couple of small tweaks based on our discussion.

p0358:
Thank you, the website appears way more readable now

website/_includes/nav_footer_custom.html
jonmeow: There's support for custom footers (https://just-the-docs.github.io/just-the-docs/docs/customization/#custom-footer). That could actually be used instead of this; I only put it here because of the default. I worry about making long footers though, I'll check with you separately about that.

jonmeow: Note, pushed with this, though still feels like a long footer to me.

website/_config.yml
jonmeow: Done.

Note, switched capitalization for https://developers.google.com/style/capitalization#capitalization-in-titles-and-headings

website/_includes/footer_custom.html
```diff
@@ -0,0 +1,31 @@
...
+    <a
+      href="https://github.com/carbon-language/carbon-lang/blob/trunk/{{ page.path }}"
+      id="edit-this-page"
+      >View on GitHub</a
+    >
+  </p>
+
+  <p class="text-small">
+    Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+    Exceptions. See <a href="/LICENSE">/LICENSE</a> for license information.
```
chandlerc: Maybe a wording tweak and link to GitHub?

```suggestion
    Part of the Carbon Language project, licensed under the
    <a href="https://github.com/carbon-language/carbon-lang/blob/trunk/LICENSE">Apache License v2.0 with LLVM
    Exceptions<a>.
```

website/README.md
```diff
@@ -0,0 +1,21 @@
...
+<!--
+Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+Exceptions. See /LICENSE for license information.
+SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+-->
+
+Website generation is done by
+[gh_pages_deploy.yaml](/.github/workflows/gh_pages_deploy.yaml). It runs
+`prebuild.py`, which prepares files for website generation, then builds the
+website using Jekyll, configured through `_config.yml`.
```
chandlerc, line 33: Just hoping to document a bit the intended scope of the current effort, mostly to avoid confusion from external folks maybe looking at this part of the project...

```suggestion
Currently, the main website for Carbon is the GitHub project page as this
is still an experiment and in early stages of development.

However, we do maintain a documentation site where our project's documentation is rendered and more easily navigated. The documentation site generation is done by
[gh_pages_deploy.yaml](/.github/workflows/gh_pages_deploy.yaml). It runs
`prebuild.py`, which prepares files for website generation, then builds the
website using Jekyll, configured through `_config.yml`.
```

Happy to iterate on better phrasing though.

website/_config.yml
```diff
@@ -0,0 +1,25 @@
+# Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+# Exceptions. See /LICENSE for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+title: Carbon language documentation
```
chandlerc: While I understand you wanted to follow the guidance on capitalization here, I think this should still capitalize `Language` as that's part of the proper name "Carbon Language"?

(We also capitalize it in other places such as the title text of the main readme.)

```suggestion
title: Carbon Language documentation
```

chandlerc (commented):
Generally, I think the code here is looking reasonable. Couple more suggestions inline.

Curious, should the URL be mentioned anywhere?

jonmeow:
> Generally, I think the code here is looking reasonable. Couple more suggestions inline.
> 
> Curious, should the URL be mentioned anywhere?

If you mean "in general", I think we might want to add a link on GitHub once it's functional. (note there's a spot for this in the top right of https://github.com/carbon-language/carbon-lang/)

If you mean "why doesn't the site have the FQDN in files", I think that's because in order to test webservers, you need relative links, without a FQDN. Bear in mind we aren't building a host, just the HTML.

website/README.md
jonmeow: I've added a paragraph along these lines. Note though, I really want to keep the "how does this work" section focused on that.

jonmeow: Note, if there are more comments, maybe we could meet to run through details and get things resolved?

jonmeow:
> > Curious, should the URL be mentioned anywhere?
> 
> If you mean "in general", I think we might want to add a link on GitHub once it's functional. (note there's a spot for this in the top right of https://github.com/carbon-language/carbon-lang/)
> 
> If you mean "why doesn't the site have the FQDN in files", I think that's because in order to test webservers, you need relative links, without a FQDN. Bear in mind we aren't building a host, just the HTML.

Maybe you mean "how is this configured"... added README notes, probably good to have written down.

website/README.md
chandlerc: Nope, this was one of the only comments I had, thanks.

chandlerc:
> > > Curious, should the URL be mentioned anywhere?
> > 
> > 
> > If you mean "in general", I think we might want to add a link on GitHub once it's functional. (note there's a spot for this in the top right of [carbon-language/carbon-lang](https://github.com/carbon-language/carbon-lang/))
> > If you mean "why doesn't the site have the FQDN in files", I think that's because in order to test webservers, you need relative links, without a FQDN. Bear in mind we aren't building a host, just the HTML.
> 
> Maybe you mean "how is this configured"... added README notes, probably good to have written down.

Yeah, this was basically what led to my question, and the docs make sense, thanks!

chandlerc (approved):
Approving for leads, ship it and thanks.

PR #4190, opened 2024-08-05
Build a generic representing the parameterization of the body of an interface by its implicit `Self` parameter.

zygoloid:
This is distinct from the generic representing the interface itself, which may or may not exist but doesn't have the `Self` parameter.

For formatting purposes, I've merged the inner generic into the interface definition rather than adding a separate nested `generic` block. That seemed like the most logical option given that an interface definition is always generic and doesn't have parameters of its own, and this avoids an extra level of indentation for every interface. I'm also using the same SemIR scope name for the two generics and the interface itself, to minimize scope qualifications.

This is in preparation for using specifics of the generic with self for checking `impl`s and uses of the interface rather than performing a from-scratch substitution each time.

### DIFF
--- a/toolchain/check/context.cpp
+++ b/toolchain/check/context.cpp
@@ -1014,10 +1014,6 @@ auto Context::TryToDefineType(
       builder.Emit();
       return false;
     }
-
-    if (interface->specific_id.is_valid()) {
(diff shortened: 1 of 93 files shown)


zygoloid:
Converted to draft: I'm not confident this is the right direction.

PR #4191, opened 2024-08-06
Distinguish between whether an entity has its own parameter lists and whether it is generic.

zygoloid:
It's actually possible to get into all four combinations of having parameter lists versus being generic:

- An entity nested within a generic, such as a member class, can be generic even if it has no parameters.

- As a corner case, an entity with an *empty* parameter list has parameter lists, but isn't a generic because it doesn't have any generic parameters.


jonmeow (approved):
Looks good!

Bonniemarie216 (approved):

Bonniemarie216 (approved):

PR #4192, opened 2024-08-06
Update workflows for security hardening.

jonmeow:
Also a small pass on workflow names.

Note, I'm a little concerned that the test/nightly release/pre-commit endpoints may be fragile. At the same time, it's also where it may be most useful, to prevent network access by arbitrary test code. I think this is imperfect, but maybe we can try it out and see if it's much of an issue.

Note, the discord wiki action is currently broken, this should fix it.

### DIFF
--- a/.github/workflows/discord_wiki.yaml
+++ b/.github/workflows/discord_wiki.yaml
@@ -2,18 +2,20 @@
 # Exceptions. See /LICENSE for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-name: Wiki Changed Discord Notification
+name: Discord Wiki Change Notifications
 
 on: gollum
 
-permissions: none
+# Minimum permissions.
+permissions:
+  contents: read
 
 jobs:
   notify:
     runs-on: ubuntu-latest
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@63c24ba6bd7ba022e95695ff85de572c04a18142 # v2.7.0
+        uses: step-security/harden-runner@17d0e2bd7d51742c71671bd19fa12bdc9d40a3d6 # v2.8.1
         with:
           egress-policy: audit
 
--- a/.github/workflows/nightly_release.yaml
+++ b/.github/workflows/nightly_release.yaml
@@ -37,9 +37,30 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@63c24ba6bd7ba022e95695ff85de572c04a18142 # v2.7.0
+        uses: step-security/harden-runner@17d0e2bd7d51742c71671bd19fa12bdc9d40a3d6 # v2.8.1
         with:
-          egress-policy: audit
+          egress-policy: block
+          # When adding endpoints, see README.md.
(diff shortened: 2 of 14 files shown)


.github/workflows/nightly_release.yaml
```diff
@@ -37,9 +37,30 @@ jobs:
...
+            oauth2.googleapis.com:443
+            objects.githubusercontent.com:443
+            phoenixnap.dl.sourceforge.net:443
+            pilotfiber.dl.sourceforge.net:443
+            psychz.dl.sourceforge.net:443
+            releases.bazel.build:443
+            sourceforge.net:443
+            storage.googleapis.com:443
+            uploads.github.com:443
+            versaweb.dl.sourceforge.net:443
```
zygoloid: I don't know how to audit these `*.dl.sourceforge.net` domains. Can we blanket permit anything of the form `*.dl.sourceforge.net:443`? (If not, how do we know which ones are safe?)

.github/workflows/discord_wiki.yaml
```diff
@@ -2,18 +2,20 @@
...
 
 jobs:
   notify:
     runs-on: ubuntu-latest
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@63c24ba6bd7ba022e95695ff85de572c04a18142 # v2.7.0
+        uses: step-security/harden-runner@17d0e2bd7d51742c71671bd19fa12bdc9d40a3d6 # v2.8.1
         with:
           egress-policy: audit
```
zygoloid, line 20: Can we narrow this?

.github/workflows/sync_repos.yaml
```diff
@@ -25,7 +25,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@63c24ba6bd7ba022e95695ff85de572c04a18142 # v2.7.0
+        uses: step-security/harden-runner@17d0e2bd7d51742c71671bd19fa12bdc9d40a3d6 # v2.8.1
         with:
           egress-policy: audit
```
zygoloid, line 30: Can we narrow this?

.github/workflows/discord_wiki.yaml
jonmeow: Probably, but the workflow won't run right now (see the action history and the permissions change). I'm not confident about what it'll access without seeing.

.github/workflows/sync_repos.yaml
jonmeow: I don't see it at https://app.stepsecurity.io/github/carbon-language/carbon-lang/actions/runs

.github/workflows/discord_wiki.yaml
zygoloid: Sure. Progress over perfection, we can revisit these.

.github/workflows/nightly_release.yaml
jonmeow: Seems to have worked, that's the best I can say because the syntax is badly undocumented. pre-commit uses the in-progress workflow file.

I'd also tried `|` instead of `>` to workaround formatting, but that didn't work.

zygoloid (approved):

PR #4196, opened 2024-08-07
Fix importing of generic types.

zygoloid:
Ensure we don't lose the symbolic constant value by mapping through to the underlying inst ID.

### DIFF
--- a/toolchain/check/import_ref.cpp
+++ b/toolchain/check/import_ref.cpp
@@ -665,14 +665,14 @@ class ImportRefResolver {
       // If the parameter is a symbolic binding, build the BindSymbolicName
       // constant.
       auto bind_id = inst_id;
-      if (auto addr =
-              import_ir_.insts().TryGetAs<SemIR::AddrPattern>(bind_id)) {
+      auto bind_inst = import_ir_.insts().Get(bind_id);
+      if (auto addr = bind_inst.TryAs<SemIR::AddrPattern>()) {
         bind_id = addr->inner_id;
+        bind_inst = import_ir_.insts().Get(bind_id);
       }
-      auto bind_const_id =
-          import_ir_.insts().Is<SemIR::BindSymbolicName>(bind_id)
-              ? GetLocalConstantId(bind_id)
-              : SemIR::ConstantId::Invalid;
+      auto bind_const_id = bind_inst.Is<SemIR::BindSymbolicName>()
+                               ? GetLocalConstantId(bind_id)
+                               : SemIR::ConstantId::Invalid;
       param_data.push_back(
           {.type_const_id = type_const_id, .bind_const_id = bind_const_id});
     }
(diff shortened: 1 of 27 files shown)


toolchain/check/import_ref.cpp
```diff
@@ -651,24 +669,28 @@ class ImportRefResolver {
               import_ir_.insts().TryGetAs<SemIR::AddrPattern>(bind_id)) {
         bind_id = addr->inner_id;
       }
-      GetLocalConstantId(bind_id);
+      auto bind_const_id =
+          import_ir_.insts().Is<SemIR::BindSymbolicName>(bind_id)
```
jonmeow: I'll note, this looks slightly redundant with the above TryGetAs. If the original bind_id isn't an AddrPattern, it'll fetch the same inst twice. I think that's the common case, too.

You might consider something like:

```
auto bind_inst = import_ir_.insts().Get(bind_id);
if (auto addr = bind_inst.TryAs<SemIR::AddrPattern>()) {
  bind_id = addr->inner_id;
  bind_inst = import_ir_.insts().Get(bind_id);
}
auto bind_const_id = bind_inst.Is<SemIR::BindSymbolicName>()
```

Although maybe this will be in the noise...

toolchain/check/import_ref.cpp
```diff
@@ -712,14 +735,17 @@ class ImportRefResolver {
...
             // We already imported a constant value for this symbolic binding.
-            // We can reuse most of it.
-            auto new_bind_inst_id = GetLocalConstantInstId(bind_id);
+            // We can reuse most of it, but update the value to point to our
+            // specific parameter, and preserve the constant value.
             auto new_bind_inst =
                 context_.insts().GetAs<SemIR::BindSymbolicName>(
-                    new_bind_inst_id);
+                    context_.constant_values().GetInstId(
+                        param_data.bind_const_id));
```
jonmeow, line 743: Can bind_const_id ever point at an error? i.e., somewhat concerned that this might fail on an import of an interface with an invalid parameter. How tested are invalid generic imports?

jonmeow (approved):
LG, though might want to double-check test cases

toolchain/check/import_ref.cpp

I was curious whether this would make any difference, so I took a look at the assembly. Looks like `insts().Get` is generally not getting inlined (probably because the `CHECK` makes it quite large), so my conclusion is... yes, this probably helps, but I'm sad that it's needed. Maybe we should use a `DCHECK` in `ValueStore::Get`?

toolchain/check/import_ref.cpp
zygoloid: If we get here then in the imported IR we had a `BindSymbolicName`, and the `bind_const_id` is the result of importing that `BindSymbolicName`, which can't produce an error. We only support a few very specific instruction patterns here, and hit a fatal error a few lines below if we see anything not matching those patterns (unchanged in this PR).

To your broader point, I don't think we have good checking for imports of invalid generics. Is that something we're generally testing and expecting to work currently? I can see tooling use cases where it'd be important, but it's not something I've really been thinking about.

jonmeow: The important thing to me is really just that check shouldn't crash. Imports of invalid generics shouldn't cause a crash. I've added tests for some various invalid import cases. I don't think you've added tests for that as part of adding generics.

Issue #4145, opened 2024-07-17
Check usage for verbose formatting of instructions in crashes


jonmeow:
#4125 added verbose formatting of instructions in crashes. If it's not used, or it gets in the way (a crash during a pretty printing) we might want to roll it back. This issue is to help track whether it's in use.

For example, this is the `%<unexpected>` formatted SemIR in the example crash output:

```
2.	Check::Context
          NodeStack:
(elided)
            2. IntLiteral:
              %<unexpected>.inst+26.loc12_10: i32 = int_literal 0 [template = constants.%.2]
```

Please comment here if you either:

1) Have a toolchain crash where you find the instruction formatting useful (it'd particularly help to explain why formatted was helpful)
    - Also if you have a crash where you expected it to be helpful, but it wasn't.
2) Have a toolchain crash which doesn't produce a stack trace, instead doing something like a `SIGSEGV` after argument printing and before `Check::Context` printing (which is symptomatic of a crash during the formatting, and could be worked around by just commenting it out temporarily).

jonmeow:
Trying to make some formatter changes, and I'm having to disable this in order to get stack traces already.

jonmeow:
In #4198, I got a stack trace from Formatter while trying to debug an interface crash. I'm trying to make the formatter more robust as a consequence. Note what I saw was actually a little quirky, with *both* stack traces and *both* CHECK messages, I think because of how CHECK works. But then I was trying to figure out what exactly was going on, it's pretty misleading to have the crash in formatter.

```
╚╡bazel run //toolchain -- compile --phase=check test.carbon
INFO: Invocation ID: 5ea2342f-85c0-4e36-898a-d37cba1c5e35
INFO: Analyzed target //toolchain:toolchain (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //toolchain/install:prefix_root/bin/carbon up-to-date:
  bazel-bin/toolchain/install/prefix_root/bin/carbon
INFO: Elapsed time: 0.320s, Critical Path: 0.07s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Running command line: bazel-bin/toolchain/install/prefix_root/bin/carbon compile '--phase=check' test.carbon
Please report issues to https://github.com/carbon-language/carbon-lang/issues and include the crash backtrace.
Stack dump:
[...]

PR #4198, opened 2024-08-07
Fix handling of interface redefinitions.

jonmeow:
The prior code crashed when trying to find the function's `Self` parameter, I believe. `fail_redefine_with_dependents.carbon` handles this case. It wasn't caught by the prior case because the `F` didn't have any dependent parameters.

Note this also ran into a formatter crash, with invalid constants. I'm fixing that here, but will also note it on #4145 (the crash in FinishGenericDecl was muddled by a crash in Formatter code).

### DIFF
--- a/toolchain/check/handle_interface.cpp
+++ b/toolchain/check/handle_interface.cpp
@@ -136,7 +136,8 @@ auto HandleParseNode(Context& context,
   auto& interface_info = context.interfaces().Get(interface_id);
 
   // Track that this declaration is the definition.
-  CARBON_CHECK(!interface_info.is_defined());
+  CARBON_CHECK(!interface_info.is_defined())
+      << "Can't merge with defined interfaces.";
   interface_info.definition_id = interface_decl_id;
   interface_info.scope_id =
       context.name_scopes().Add(interface_decl_id, SemIR::NameId::Invalid,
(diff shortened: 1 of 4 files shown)


toolchain/check/handle_interface.cpp
```diff
@@ -60,19 +61,37 @@ static auto BuildInterfaceDecl(Context& context,
             context.insts().Get(existing_id).TryAs<SemIR::InterfaceDecl>()) {
       // TODO: Implement full redeclaration checking. See `MergeClassDecl`. For
       // now we just check the generic parameters match.
```
zygoloid:
```suggestion
```

toolchain/sem_ir/formatter.cpp
```diff
@@ -1027,13 +1028,13 @@ class FormatterImpl {
...
         out_ << " (";
-        FormatName(sem_ir_.constant_values().GetInstId(id));
+        FormatName(sem_ir_.constant_values().GetInstIdIfValid(id));
         out_ << ")";
         return;
       }
     }
 
-    FormatName(sem_ir_.constant_values().GetInstId(id));
+    FormatName(sem_ir_.constant_values().GetInstIdIfValid(id));
```
zygoloid: These changes look redundant: we already checked for an invalid `id` at the start of this function.

zygoloid (approved):
Thanks!

toolchain/sem_ir/formatter.cpp
jonmeow: Undone here, I admit I didn't look closely on the crash.

toolchain/check/handle_interface.cpp
jonmeow: Done

zygoloid (approved):

PR #4201, opened 2024-08-07
Add indentation to yaml test failure output to make it easier to read.

zygoloid:


toolchain/testing/yaml_test_helpers.h
```diff
@@ -58,6 +58,19 @@
 
 namespace Carbon::Testing::Yaml {
 
+// Adds the specified indentation before each newline in the given string.
+inline auto IndentString(std::string_view str, std::string_view indent = "  ")
```
jonmeow: Is the "indent" parameter needed?

(if so, you might consider llvm::StringRef for broader consistency, although I know some testing stuff uses string_view)

toolchain/testing/yaml_test_helpers.h
```diff
@@ -58,6 +58,19 @@
...
+    -> std::string {
+  std::string result;
+  for (auto c : str) {
+    result.push_back(c);
+    if (c == '\n') {
+      result += indent;
+    }
+  }
+  return result;
```
jonmeow:
```suggestion
  return absl::StrReplaceAll(str, {{"\n", "\n  "}});
```

I was considering suggesting `result.reserve(str.size())`, but maybe you can just include `absl/strings/str_replace.h` and let it handle the complexities? (also probably doing better than a char-by-char loop)

jonmeow (approved):

PR #4202, opened 2024-08-08
Defer resolving the eval blocks and value blocks of generics and specifics until we've finished other resolution work.

zygoloid:
This avoids import cycles, and reduces the number of temporary vectors we build (and potentially throw away on retry). Import the self specific when importing a generic, now that there's no risk that will introduce cycles.

Note that we could take the same approach to import classes, interfaces, and so on, instead of the current third phase of resolution for those instructions, but in this PR I'm just addressing the import cycle I'm currently seeing in a work-in-progress PR.

### DIFF
--- a/toolchain/check/import_ref.cpp
+++ b/toolchain/check/import_ref.cpp
@@ -1800,6 +1800,8 @@ class ImportRefResolver {
 
   // Perform any work that we deferred until the end of the main Resolve loop.
   auto PerformPendingWork() -> void {
+    // Note that the individual Finish steps can add new pending work, so keep
+    // going until we have no more work to do.
     while (!pending_generics_.empty() || !pending_specifics_.empty()) {
       while (!pending_generics_.empty()) {
         FinishPendingGeneric(pending_generics_.pop_back_val());
@@ -1844,24 +1846,26 @@ class ImportRefResolver {
   auto FinishPendingGeneric(PendingGeneric pending) -> void {
     const auto& import_generic = import_ir_.generics().Get(pending.import_id);
 
-    // Note that the generics list can be reallocated by ResolveLocalEvalBlock
-    // importing more generics.
+    // Don't store this between calls: the generics list can be reallocated by
+    // ResolveLocalEvalBlock importing more generics.
+    auto local_generic = [&]() -> auto& {
+      return context_.generics().Get(pending.local_id);
+    };
+
     auto decl_block_id =
         ResolveLocalEvalBlock(import_generic, pending.local_id,
                               SemIR::GenericInstIndex::Region::Declaration);
-    auto* local_generic = &context_.generics().Get(pending.local_id);
-    local_generic->decl_block_id = decl_block_id;
+    local_generic().decl_block_id = decl_block_id;
 
     auto self_specific_id = MakeSelfSpecific(context_, pending.local_id);
-    local_generic->self_specific_id = self_specific_id;
+    local_generic().self_specific_id = self_specific_id;
     pending_specifics_.push_back({.import_id = import_generic.self_specific_id,
                                   .local_id = self_specific_id});
 
     auto definition_block_id =
         ResolveLocalEvalBlock(import_generic, pending.local_id,
                               SemIR::GenericInstIndex::Region::Definition);
-    local_generic = &context_.generics().Get(pending.local_id);
-    local_generic->definition_block_id = definition_block_id;
+    local_generic().definition_block_id = definition_block_id;
   }
 
   // Resolves and returns a local inst block of constant instructions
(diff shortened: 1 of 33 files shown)


toolchain/check/import_ref.cpp
```diff
@@ -1811,6 +1798,106 @@ class ImportRefResolver {
          .element_type_id = context_.GetTypeIdForTypeConstant(elem_const_id)});
   }
 
+  // Perform any work that we deferred until the end of the main Resolve loop.
+  auto PerformPendingWork() -> void {
+    while (!pending_generics_.empty() || !pending_specifics_.empty()) {
```
jonmeow, line 1805: This is weird, can you comment it. :)

toolchain/check/import_ref.cpp
```diff
@@ -1811,6 +1798,106 @@ class ImportRefResolver {
...
+    auto self_specific_id = MakeSelfSpecific(context_, pending.local_id);
+    local_generic->self_specific_id = self_specific_id;
+    pending_specifics_.push_back({.import_id = import_generic.self_specific_id,
+                                  .local_id = self_specific_id});
+
+    auto definition_block_id =
+        ResolveLocalEvalBlock(import_generic, pending.local_id,
+                              SemIR::GenericInstIndex::Region::Definition);
+    local_generic = &context_.generics().Get(pending.local_id);
+    local_generic->definition_block_id = definition_block_id;
```
jonmeow: The use of a pointer (and reassignment) looked a little odd to me, maybe:

```suggestion
    auto& local_generic = context_.generics().Get(pending.local_id);
    local_generic.decl_block_id = decl_block_id;

    auto self_specific_id = MakeSelfSpecific(context_, pending.local_id);
    local_generic.self_specific_id = self_specific_id;
    pending_specifics_.push_back({.import_id = import_generic.self_specific_id,
                                  .local_id = self_specific_id});

    // This may invalidate local_generic.
    auto definition_block_id =
        ResolveLocalEvalBlock(import_generic, pending.local_id,
                              SemIR::GenericInstIndex::Region::Definition);
    context_.generics().Get(pending.local_id).definition_block_id = definition_block_id;
```

toolchain/check/import_ref.cpp
```diff
@@ -1811,6 +1798,106 @@ class ImportRefResolver {
...
+      local_specific = &context_.specifics().Get(pending.local_id);
+      local_specific->decl_block_id = decl_block_id;
+    }
+
+    if (!local_specific->definition_block_id.is_valid() &&
+        import_specific.definition_block_id.is_valid()) {
+      auto definition_block_id =
+          ResolveLocalInstBlock(import_specific.definition_block_id);
+      local_specific = &context_.specifics().Get(pending.local_id);
+      local_specific->definition_block_id = definition_block_id;
```
jonmeow: Similar to the above, I'd suggest only reusing the same variable when it's significant (which for line 1892, it is, but that doesn't look true here).

```suggestion
      context_.specifics().Get(pending.local_id).definition_block_id = definition_block_id;
```

jonmeow (approved):
LG, just a few minor things that had me looking closer. Per separate discussion, I'd also missed in the PR description the mention that `Self` is now imported, explaining the test changes.

toolchain/check/import_ref.cpp
jonmeow: Note, suggesting this mainly because I find it easier to figure out what the source is when it's assigned in fewer places.

toolchain/check/import_ref.cpp
zygoloid: I started with something like this, but it seemed error-prone to me and harder to systematically get right. Changed to a different approach, what do you think?

toolchain/check/import_ref.cpp
```diff
@@ -1844,24 +1846,26 @@ class ImportRefResolver {
   auto FinishPendingGeneric(PendingGeneric pending) -> void {
     const auto& import_generic = import_ir_.generics().Get(pending.import_id);
 
-    // Note that the generics list can be reallocated by ResolveLocalEvalBlock
-    // importing more generics.
+    // Don't store this between calls: the generics list can be reallocated by
+    // ResolveLocalEvalBlock importing more generics.
+    auto local_generic = [&]() -> auto& {
+      return context_.generics().Get(pending.local_id);
+    };
```
jonmeow: Maybe move onto the type as a function, so calls look like `pending.Get(context)`? (lambdas always make me hesitant)

jonmeow (approved):

toolchain/check/import_ref.cpp
jonmeow: Not, discussed offline, just inlining the fetches

PR #4203, opened 2024-08-08
Basic support for implementing and using a parameterized interface.

zygoloid:
The main change here is to form a specific when checking an interface
function against an impl function, instead of just substituting the
`Self` type.

### DIFF
--- a/toolchain/check/function.h
+++ b/toolchain/check/function.h
@@ -27,13 +27,13 @@ struct SuspendedFunction {
 };
 
 // Checks that `new_function` has the same parameter types and return type as
-// `prev_function`, applying the specified set of substitutions to the
-// previous function. Prints a suitable diagnostic and returns false if not.
-auto CheckFunctionTypeMatches(Context& context,
-                              const SemIR::Function& new_function,
-                              const SemIR::Function& prev_function,
-                              Substitutions substitutions, bool check_syntax)
-    -> bool;
+// `prev_function`, or possibly a specific version of `prev_function`. Prints a
+// suitable diagnostic and returns false if not.
+auto CheckFunctionTypeMatches(
+    Context& context, const SemIR::Function& new_function,
+    const SemIR::Function& prev_function,
+    SemIR::SpecificId prev_specific_id = SemIR::SpecificId::Invalid,
+    bool check_syntax = true) -> bool;
 
 // Checks that the return type of the specified function is complete, issuing an
 // error if not. This computes the return slot usage for the function if
--- a/toolchain/check/function.h
+++ b/toolchain/check/function.h
@@ -27,8 +27,8 @@ struct SuspendedFunction {
 };
 
 // Checks that `new_function` has the same parameter types and return type as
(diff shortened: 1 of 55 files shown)


zygoloid:
Depends on #4202.

toolchain/check/function.h
```diff
@@ -27,13 +27,13 @@ struct SuspendedFunction {
...
 
 // Checks that `new_function` has the same parameter types and return type as
-// `prev_function`, applying the specified set of substitutions to the
-// previous function. Prints a suitable diagnostic and returns false if not.
-auto CheckFunctionTypeMatches(Context& context,
-                              const SemIR::Function& new_function,
-                              const SemIR::Function& prev_function,
-                              Substitutions substitutions, bool check_syntax)
-    -> bool;
+// `prev_function`, or possibly a specific version of `prev_function`. Prints a
```
josh11b: ..., or if `prev_specific_id` is specified, then the specific version of `prev_function`.

toolchain/check/testdata/impl/fail_impl_bad_type.carbon
```diff
@@ -21,7 +21,8 @@ impl true as I {}
 // CHECK:STDOUT:   %.1: type = interface_type @I [template]
 // CHECK:STDOUT:   %Self: %.1 = bind_symbolic_name Self 0 [symbolic]
 // CHECK:STDOUT:   %.2: bool = bool_literal true [template]
-// CHECK:STDOUT:   %.3: <witness> = interface_witness () [template]
+// CHECK:STDOUT:   %.3: type = tuple_type () [template]
```
josh11b, line 24: This was added by this PR, but doesn't seem to be referenced by anything and doesn't seem to correspond to anything in the source code. Is there some mistake?

josh11b (approved):
Independent of this PR, could you document the format of a `specific` block in the SemIR, what it means, what the `!definition:` label marks, and its relationship with the `generic` block using the same name in https://docs.google.com/document/d/1RRYMm42osyqhI2LyjrjockYCutQ5dOf8Abu50kTrkX0/edit?pli=1&resourcekey=0-kHyqOESbOHmzZphUbtLrTw&tab=t.0#heading=h.xlsjzlpp9x0k ?

The SemIR is pretty difficult to understand to new eyes at the moment. It would be good to spend some effort to describe more of what is going on. Things like what are the arguments to `fn_type` (sometimes there is a comma?), what the meaning of `[symbolic]` versus `[template]` is, what does `=>` signify, and so on. Right now the section in the toolchain architecture is quite short and doesn't cover much.

toolchain/check/testdata/impl/fail_impl_bad_type.carbon
zygoloid: I tracked this down -- this is the value representation of the interface type `%.1`. The mapping from complete types to their value representations isn't something we currently display in formatted SemIR. Perhaps it should be.

josh11b: Either that or be given a more informative name?

PR #4205, opened 2024-08-08
add syntax highlighting for `x-macro` files

brymer-meneses:
This tells Github to detect files having the extension .def as C++ files, which ensures that these files get syntax highlighted prpperly.

I'm not sure whether more files like these are present throughout the repo. FWIW, it can also be configured to do other things, such as detect generated, vendored files and exclude these from the Github stats.

See https://github.com/github-linguist/linguist/blob/master/docs/overrides.md.


jonmeow (approved):
Cool, thanks!

PR #4209, opened 2024-08-09
Use `As` and `ImplicitAs` interfaces for conversions.

zygoloid:
Add these interfaces to the core library. For now, they're two separate interfaces because we don't yet support one interface extending another.

This collapses a lot of the layering in check: for example, the call building logic depends on implicit conversions, conversions now depend on the overloaded operator machinery, and that machinery depends on building calls.

In passing, improve the diagnostics for failing to find a name required from the prelude. Also convert all the transitively-called code from `NodeId` to `LocId` given the latter is what the conversion machinery has available.


toolchain/sem_ir/typed_insts.h
```diff
@@ -717,7 +717,7 @@ struct InterfaceWitness {
 // Accesses an element of an interface witness by index.
 struct InterfaceWitnessAccess {
   static constexpr auto Kind =
-      InstKind::InterfaceWitnessAccess.Define<Parse::InvalidNodeId>(
+      InstKind::InterfaceWitnessAccess.Define<Parse::NodeId>(
```
jonmeow: Sorry I've been missing this before, but adding the node id really makes this stand out. Looking at how this is used, it's only to pass in a LocId, not a NodeId, so I don't think one should be added. For context, there's a TODO on the relevant LocIdAndInst constructor:

```
// TODO: This is somewhat historical due to fetching the NodeId from insts()
// for things like Temporary; should we require Untyped in these cases?
```

Note "Untyped" is out-of-date, this comment should instead say "ReusingLoc".

What does it mean to associate a NodeId with this, which node is it expected to correspond to? Had you considered calling "ReusingLoc" instead of adding a NodeId?

toolchain/check/call.cpp
```diff
@@ -58,30 +56,29 @@ static auto PerformCallToGenericInterface(Context& context,
 
   // Convert the arguments to match the parameters.
   auto converted_args_id = ConvertCallArgs(
-      context, node_id, /*self_id=*/SemIR::InstId::Invalid, arg_ids,
+      context, loc_id, /*self_id=*/SemIR::InstId::Invalid, arg_ids,
       /*return_storage_id=*/SemIR::InstId::Invalid, interface_info,
       specific_id);
-  return context.AddInst<SemIR::Call>(node_id,
+  return context.AddInst<SemIR::Call>(loc_id,
```
jonmeow, line 62: In these cases (I also see line 36, and more below), where you're changing from a node_id to a loc_id, should this actually be calling ReusingLoc? (also see my other comment on typed_nodes, it's part of why I'm looking closer)

toolchain/check/operator.h
```diff
@@ -13,19 +13,29 @@ namespace Carbon::Check {
...
   llvm::StringLiteral interface_name;
+  llvm::ArrayRef<SemIR::InstId> interface_args = {};
   llvm::StringLiteral op_name = "Op";
 };
 
 // Checks and builds SemIR for a unary operator expression. For example,
-// `$operand` or `operand$`.
-auto BuildUnaryOperator(Context& context, Parse::AnyExprId node_id, Operator op,
-                        SemIR::InstId operand_id) -> SemIR::InstId;
+// `$operand` or `operand$`. If specified, `missing_impl_diagnoser` is used to
```
jonmeow: I'm not quite sure what you're trying to get at here, I'm used to seeing `$operand` for shell expansions, and I'm not aware of it being valid in Carbon. [I think it's been used in the past as a placeholder, probably that's why `$`, but I'm not sure that's going to make sense to newcomers.] Maybe we can use `*`?
```suggestion
// `*operand` or `operand*`. If specified, `missing_impl_diagnoser` is used to
```

toolchain/check/operator.h
```diff
@@ -13,19 +13,29 @@ namespace Carbon::Check {
...
+    SemIR::InstId operand_id,
+    std::optional<Context::Diagnoser> missing_impl_diagnoser = std::nullopt)
+    -> SemIR::InstId;
 
 // Checks and builds SemIR for a binary operator expression. For example,
-// `lhs_id $ rhs_id`.
-auto BuildBinaryOperator(Context& context, Parse::AnyExprId node_id,
-                         Operator op, SemIR::InstId lhs_id,
-                         SemIR::InstId rhs_id) -> SemIR::InstId;
+// `lhs_id $ rhs_id`. If specified, `missing_impl_diagnoser` is used to build a
```
jonmeow:
```suggestion
// `lhs_id * rhs_id`. If specified, `missing_impl_diagnoser` is used to build a
```

toolchain/check/context.h
```diff
@@ -50,6 +50,7 @@ class Context {
  public:
   using DiagnosticEmitter = Carbon::DiagnosticEmitter<SemIRLoc>;
   using DiagnosticBuilder = DiagnosticEmitter::DiagnosticBuilder;
+  using Diagnoser = llvm::function_ref<auto()->Context::DiagnosticBuilder>;
```
jonmeow: The way the alias hides the return type changes how I read the "diagnoser" variables. I expect a "Diagnoser" function to fully build and emit the diagnostic (and where I saw comments about this elsewhere, "is used to build" doesn't clarify my misunderstanding). Maybe "BuildDiagnostic" or "MakeDiagnosticBuilder" to reflect that this is doing the build, but not doing the full diagnosing (with emitting)?

Also, this feels a little less like a thin alias versus the other "using" statements, maybe this should have a comment explaining its use?

toolchain/check/context.cpp
```diff
@@ -395,18 +395,18 @@ static auto GetCorePackage(Context& context, SemIRLoc loc)
...
-  // We expect it to be a namespace.
-  if (auto namespace_inst =
-          context.insts().TryGetAs<SemIR::Namespace>(core_inst_id)) {
-    return namespace_inst->name_scope_id;
-  }
-  // TODO: This should really diagnose the name issue.
-  context.DiagnoseNameNotFound(loc, core_name_id);
+  CARBON_DIAGNOSTIC(CoreNotFound, Error,
+                    "Package `Core` implicitly referenced here but not found.",
+                    SemIR::NameId);
```
jonmeow: Should there be a separate diagnostic for the not-a-namespace case?

toolchain/check/context.cpp
```diff
@@ -395,18 +395,18 @@ static auto GetCorePackage(Context& context, SemIRLoc loc)
...
 
-  // We expect it to be a namespace.
-  if (auto namespace_inst =
-          context.insts().TryGetAs<SemIR::Namespace>(core_inst_id)) {
-    return namespace_inst->name_scope_id;
-  }
-  // TODO: This should really diagnose the name issue.
-  context.DiagnoseNameNotFound(loc, core_name_id);
+  CARBON_DIAGNOSTIC(CoreNotFound, Error,
+                    "Package `Core` implicitly referenced here but not found.",
```
```suggestion
                    "Package `Core` implicitly referenced here, but not found.",
```

toolchain/check/context.cpp
```diff
@@ -421,7 +421,11 @@ auto Context::LookupNameInCore(SemIRLoc loc, llvm::StringRef name)
   auto inst_id = LookupNameInExactScope(loc, name_id, core_package_id,
                                         name_scopes().Get(core_package_id));
   if (!inst_id.is_valid()) {
-    DiagnoseNameNotFound(loc, name_id);
+    CARBON_DIAGNOSTIC(
+        CoreNameNotFound, Error,
+        "Name `Core.{0}` implicitly referenced here but not found.",
```
```suggestion
        "Name `Core.{0}` implicitly referenced here, but not found.",
```

toolchain/check/member_access.cpp
```diff
@@ -155,23 +155,33 @@ static auto LookupInterfaceWitness(Context& context,
...
+          .Emit();
+    } else {
+      CARBON_DIAGNOSTIC(MissingImplInMemberAccess, Error,
+                        "Cannot access member of interface `{0}` in type `{1}` "
+                        "that does not implement that interface.",
+                        SemIR::NameId, std::string);
+      context.emitter().Emit(loc_id, MissingImplInMemberAccess,
+                             interface.name_id,
+                             context.sem_ir().StringifyType(type_const_id));
+    }
```
jonmeow: Can these use TypeId directly?

```suggestion
    if (missing_impl_diagnoser) {
      CARBON_DIAGNOSTIC(MissingImplInMemberAccessNote, Note,
                        "Type `{1}` does not implement interface `{0}`.",
                        SemIR::NameId, SemIR::TypeId);
      (*missing_impl_diagnoser)()
          .Note(loc_id, MissingImplInMemberAccessNote, interface.name_id,
               type_const_id)
          .Emit();
    } else {
      CARBON_DIAGNOSTIC(MissingImplInMemberAccess, Error,
                        "Cannot access member of interface `{0}` in type `{1}` "
                        "that does not implement that interface.",
                        SemIR::NameId, SemIR::TypeId);
      context.emitter().Emit(loc_id, MissingImplInMemberAccess,
                             interface.name_id,
                             type_const_id);
    }
```

jonmeow (approved):
Generally looks good, although there's a lot more reuse of locations, which got me to look at the TODO that'd hamper that. #4211 for the relevant changes.

I'm approving, if you want to merge before heading out and leave the locations for me to clean up, that's fine by me. Though if you object to the location stuff, please comment on #4211 before going. :)

toolchain/check/operator.h
chandlerc: FWIW, in C++ land, specifically WG21, `$` is commonly used as a placeholder symbol when describing unary- and binary-operator syntaxes that apply to multiple different operator symbols.

Not saying it's a great idiom, but that's what I suspect is going on here. Not sure what the best syntax is though.

toolchain/sem_ir/typed_insts.h
zygoloid: The `NodeId` is the parse node corresponding to the instruction. `InvalidNodeId` means there isn't one, plain `NodeId` means it could be a parse node of any kind. In this PR we start passing a location in when constructing an `InterfaceWitnessAction`, so there is now an associated parse node, so I'm using `NodeId` to mean "any parse node" rather than `InvalidNodeId` to mean "no parse node".

toolchain/check/call.cpp
zygoloid: Maybe we should have a `CHECK` or `DCHECK` that the `loc_id` refers to a node of the expected type (either directly or via import)?

toolchain/check/operator.h
zygoloid: Yeah, this is `$` as in "some arbitrary operator". But in any case this is pre-existing. I'd be fine if the example were changed to use `*` instead.

toolchain/check/context.cpp
zygoloid: My thinking was: if lookup for `Core` finds something that's not even a namespace, then it definitely didn't find a package named `Core`, so the same diagnostic would still make sense.

toolchain/check/context.cpp
zygoloid: These aren't independent clauses -- the "not found" is implicitly referring to the subject "package `Core`" -- so I think the "but" should not be preceded by a comma.

toolchain/check/member_access.cpp
zygoloid: Done. (`type_const_id` is a `ConstantId` not a `TypeId`, so the suggestion doesn't quite work.)

toolchain/check/call.cpp
jonmeow: There's really no reason to do that for a Parse::NodeId entry. That is to say, every location -- even invalid -- is an accepted value for NodeId.

toolchain/sem_ir/typed_insts.h
jonmeow: Let's discuss when you're back. As I note on #4211, I don't agree with this interpretation, I don't think it reflects how the IR is working. That is to say, my understanding is that instructions that currently have InvalidNodeId have nodes associated for diagnostics, with the exception of BuiltinInst. Even import refs should trace back to a node id, albeit indirectly.

toolchain/check/context.cpp
jonmeow: I think there's a phrasing problem here; I have trouble reading it. Can I suggest instead something like:

Could not resolve implicit reference to package `Core`.

toolchain/sem_ir/typed_insts.h
jonmeow: Maybe I'm misunderstanding the existence of NodeId on some of these, actually, due to what's happening with constants and NoLoc. I'd not digested all of that, sorry. But still, we do have other instructions which use InvalidNodeId and get locations, particularly imports. Anyways, maybe #4211 is a better place to discuss.

toolchain/check/call.cpp
jonmeow: Note too, my original comment is really about trying to better track how LocId is constructed.

zygoloid:
I've run out of time to work on this; if someone wants to take over while I'm on vacation that'd be great. The remaining issue is that at least one test crashes, but only when built with `-c opt`. 

The crash is a `CHECK` failure:

```
CHECK failure at ./toolchain/base/value_store.h:174: id.index >= 0: instInvalidParseSubtree
```

coming from within a call to `Convert`. But that doesn't make any sense to me: `instInvalidParseSubtree` is meaningless, `InvalidParseSubtree` is a kind of `Node`, not a kind of `Inst`. Also, why does this only happen in an opt build? (The check message doesn't appear to end up in the log file unless a `flush` call is added to the crash handling machinery; I've pushed a change to do that.)

toolchain/check/context.cpp
jonmeow: Or just drop "here"

toolchain/check/context.cpp
jonmeow: Should it proactively identify that it's looking at package then, and not just an arbitrary namespace named `Core`? You could do that by ensuring the namespace has import IRs.

toolchain/check/context.cpp
zygoloid: Hm, I think I'm starting to see the (or at least an) issue: the "but" can be (mis)read as binding tightly to the "here" (as it would in "implicitly referenced here but not there") rather than to the "implicitly referenced" (as it would in "implicitly referenced but not found"). Your original suggestion of adding a comma seems like a reasonable way to fix that. While I think it's not strictly grammatically correct, our diagnostic messages are not written in formal English anyway. And looking through Clang's diagnostics, it uses "blah here, but [...]" with a dependent clause after the comma in various places too.

(I'm a bit hesitant to adopt your suggested "Could not resolve ..." phrasing -- I think we should prefer diagnostics to state what's wrong with the code, not how we discovered the problem, so "[you] referenced a thing that doesn't exist" seems better to me than "[I] could not resolve a reference to the thing" to me.)

toolchain/check/context.cpp
zygoloid: I'm on the fence. If someone has turned off the implicit prelude import and defined a local namespace named `Core`, it seems fine to me for that to work. But maybe that's not something we want to support, I'm not sure. Either way this seems like something it'd be good to have a test for, if we don't already.

But... this behavior is not being changed in this PR. I'm only adding a more precise diagnostic for the existing failures in this PR, not changing which cases get diagnosed.

toolchain/check/call.cpp
zygoloid: I think this was addressed by #4264.

toolchain/check/context.h
zygoloid: I'm not coming up with a great name for this. Things like `BuildDiagnosticBuilder` or `MakeDiagnosticBuilder` or `DiagnosticBuilderBuilder` seem clunky. `BuildDiagnostic` seems easily confused with `DiagnosticBuilder` (and it seems weird that `BuildDiagnostic` would return a `DiagnosticBuilder` not a `Diagnostic`). Something like `StartDiagnosing` or `InitDiagnostic` could maybe work, but those names don't really spark joy. Will ask in #naming.

common/check_internal.cpp
```diff
@@ -23,6 +23,8 @@ ExitingStream::~ExitingStream() {
 
 auto ExitingStream::Done() -> void {
   buffer_ << "\n";
+  buffer_.flush();
```
jonmeow, line 26: How did this come up for the change?

toolchain/check/context.cpp
jonmeow: > But... this behavior is not being changed in this PR.

You're removing a TODO. Maybe it should stay. :)

> If someone has turned off the implicit prelude import and defined a local namespace named Core, it seems fine to me for that to work.

FWIW, it's a confusing approach to me. I would've thought either:

- When outside `Core`, we use standard name lookup (allowing any scope named `Core`)
- We diagnose when `Core` isn't a package, in order to minimize the chance of user error.

As it stands, this looks like an edge-case feature where people can disable the prelude and, if they do, the prelude can be either a package or a namespace but not other scoped entities. Maybe this should be a leads question about what `Core` is allowed to be?

(whoops, forgot I wasn't drafting review responses)

toolchain/check/convert.cpp
```diff
@@ -958,25 +959,31 @@ auto Convert(Context& context, SemIR::LocId loc_id, SemIR::InstId expr_id,
...
-    context.emitter()
-        .Build(loc_id,
-               target.kind == ConversionTarget::ExplicitAs
-                   ? ExplicitAsConversionFailure
-                   : ImplicitAsConversionFailure,
-               expr.type_id(), target.type_id)
-        .Emit();
-    return SemIR::InstId::BuiltinError;
+    SemIR::InstId interface_args[] = {
+        context.types().GetInstId(target.type_id)};
```
jonmeow, line 966: IIUC, you're declaring this as an array so that the `ArrayRef` on `Operator` doesn't have lifetime issues. Should this have comments? My first thought was actually that you should just inline the array being passed.

Note, I'm not sure if this is more of a hesitance over `interface_args` being an arrayref and also being used for indirect passing. I assume though that `interface_args` will be used more often in the future, with more sizes, and that a separate `SmallVector` would be undesirable for copies.

Maybe renaming `interface_args` to `interface_args_ref` would help hint at the lifetime?

toolchain/check/context.cpp
```diff
@@ -449,7 +449,11 @@ auto Context::LookupNameInCore(SemIRLoc loc, llvm::StringRef name)
   auto inst_id = LookupNameInExactScope(loc, name_id, core_package_id,
                                         name_scopes().Get(core_package_id));
   if (!inst_id.is_valid()) {
-    DiagnoseNameNotFound(loc, name_id);
+    CARBON_DIAGNOSTIC(
+        CoreNameNotFound, Error,
+        "Name `Core.{0}` implicitly referenced here, but not found.",
+        SemIR::NameId);
+    emitter_->Emit(loc, CoreNameNotFound, name_id);
```
jonmeow: I'm trying to find this new diagnostic in a test file, is it untested? I understand you're not changing the control flow, but if you're changing the diagnostic maybe that should be tested?

jonmeow (commented):
Giving this a fresh look because it's been a bit, sorry if I've touched on things that were sitting around for a bit.

common/check_internal.cpp
zygoloid: A lifetime issue with the array of interface arguments caused the tests to fail with opt builds but not with dbg or fastbuild, and the opt builds also seemed to lose the end of the check failure diagnostic if I didn't explicitly flush the stream.

toolchain/check/context.cpp
zygoloid: Added the TODO back and moved out to #4276.

toolchain/check/context.cpp
zygoloid: Added tests in #4276.

toolchain/check/context.h
zygoloid: Renamed to `BuildDiagnosticFn` per open discussion.

toolchain/check/convert.cpp
zygoloid: Renamed.

jonmeow (approved):
Looks good! Glad to get this in. :)

PR #4210, opened 2024-08-09
Add `require` node(s) to the parse tree

gysddn:
Added new nodes to the parse tree to handle require declarations

`Require` is a new bracketed parse node with the following structure:

Require
  -Impls (New node for `expr impls expr` statements)
    - AnyExpr
    - AnyExpr 
  -RequireIntroducer (Bracket node)

Two test cases added for the usage in the interfaces.


toolchain/parse/testdata/generics/interface/require_name.carbon
```diff
@@ -0,0 +1,41 @@
...
+//
+// AUTOUPDATE
+// TIP: To test this file alone, run:
+// TIP:   bazel test //toolchain/testing:file_test --test_arg=--file_tests=toolchain/parse/testdata/generics/interface/require_name.carbon
+// TIP: To dump output, run:
+// TIP:   bazel run //toolchain/testing:file_test -- --dump_output --file_tests=toolchain/parse/testdata/generics/interface/require_name.carbon
+
+interface Foo {
+  fn Foo1();
+}
```
jonmeow: Syntax that's not part of what's being tested isn't needed in these files. Name lookup doesn't occur.

toolchain/parse/testdata/generics/interface/require_name.carbon
```diff
@@ -0,0 +1,41 @@
...
+
+interface Bar {
+  require i32 impls Foo;
+}
```
jonmeow: Can you add tests of incorrect syntax? There are a lot of possible typos here, and we should make sure error handling is working.

Just as a few examples:

```
// Missing components.
require;
require impls Foo;
require i32;
require i32 impls;

// Extra expressions.
require i32 i32 impls Foo;
require i32 impls Foo Foo;
```

Plus, what happens when there's a missing semicolon?

toolchain/parse/testdata/generics/interface/require_name.carbon
```diff
@@ -0,0 +1,41 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+// TIP: To test this file alone, run:
```
jonmeow: We've shifted towards using split-file tests (`// --- name.carbon`) for testing of a single feature, these single-test files are increasingly rare. For an example, see https://github.com/carbon-language/carbon-lang/blob/trunk/toolchain/parse/testdata/packages/import/export.carbon

You could probably use this to put `require_name` and `require_self` tests in the same file together, along with tests of incorrect syntax.

jonmeow (commented):
@josh11b I'm trying to figure out how to review this. Looking at the generics design, I see [examples of its use](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/generics/details.md#interface-requiring-other-interfaces-revisited), but I'm having trouble finding what the full syntax is. Is the syntax in the design somewhere that I'm missing? Some of the examples include `where`, does this overlap with the `where` parsing you've been looking at?

@gysddn FYI, the items on [Toolchain tasks](https://docs.google.com/document/d/1O2rcDO70BamTERrKhJRLODarNCQOHQWUweNdSfLw-So/edit?resourcekey=0-gWJkSd-tlpCmk2lok1UcPg&tab=t.0) are things that are closer to our development focus. I'm going to need some time to figure out what the syntax is intended to be before I can really review, but I've left some comments regarding testing since those are more general comments not specific to this change. It's important to be testing both success cases and failure cases.

josh11b:
> @josh11b I'm trying to figure out how to review this. Looking at the generics design, I see [examples of its use](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/generics/details.md#interface-requiring-other-interfaces-revisited), but I'm having trouble finding what the full syntax is. Is the syntax in the design somewhere that I'm missing? Some of the examples include `where`, does this overlap with the `where` parsing you've been looking at?

The argument to `require` is supposed to match the argument to `where`, which I recently was working on -- see #4075 . The arguments are described in https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/generics/details.md#where-constraints -- though the description is spread out across that whole section, without a short summary of just the syntax.

One thing to note, though, that in discussions we've decided that the argument to a `where` clause is not an expression, and so it has its own distinct interpretation of symbols like `==` that appear in ordinary expressions.

jonmeow:
> The argument to `require` is supposed to match the argument to `where`, which I recently was working on -- see #4075 . The arguments are described in https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/generics/details.md#where-constraints -- though the description is spread out across that whole section, without a short summary of just the syntax.

So to be sure I understand correctly, `require` actually has a single argument, and that argument is the same as an argument to `where`? i.e., the `impls` just happens to be in all the examples?

(if that understanding is correct, probably this change should be reverted since josh11b is already working on `where` syntax, and this change isn't going to provide the right support)

josh11b:
> > The argument to `require` is supposed to match the argument to `where`, which I recently was working on -- see #4075 . The arguments are described in https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/generics/details.md#where-constraints -- though the description is spread out across that whole section, without a short summary of just the syntax.
> 
> So to be sure I understand correctly, `require` actually has a single argument, and that argument is the same as an argument to `where`? i.e., the `impls` just happens to be in all the examples?
> 
> (if that understanding is correct, probably this change should be reverted since josh11b is already working on `where` syntax, and this change isn't going to provide the right support)

https://github.com/carbon-language/carbon-lang/blob/trunk/proposals/p2760.md#proposal

> For now, only the `impls` forms of `where` clauses are permitted after `require`.

Eventually my expectation is that we would accept any kind of `where` clause after `require` when parsing, and then add restrictions in the check stage; but for now it is totally fine to only accept `impls` after `require` in parse.

gysddn:
> So to be sure I understand correctly, `require` actually has a single argument, and that argument is the same as an argument to `where`? i.e., the `impls` just happens to be in all the examples?

Yeah, I didn't really realize this while making these changes.
If the syntax will be provided by @josh11b changes, then I can update this PR or just make another one.

jonmeow:
To be sure we're all on the same page:

1. Josh is fine with the basic syntax support being added.
    - That is, `require <expr> impls <expr>;`.
2. Per my comments, edge cases and incorrect syntax must be tested.
    - The parse tree must be balanced even for parse trees that contain error nodes.

toolchain/parse/typed_nodes.h
```diff
@@ -410,6 +410,28 @@ struct BuiltinFunctionDefinition {
...
+      {.category = NodeCategory::Decl, .child_count = 2});
+  AnyExprId lhs;
+  Lex::ImplsTokenIndex token;
+  AnyExprId rhs;
+};
+
+struct Require {
+  static constexpr auto Kind =
+      NodeKind::Require.Define({.category = NodeCategory::Decl,
+                                .bracketed_by = RequireIntroducer::Kind});
```
jonmeow: Since this is set up to require `impls`, it should have a child count of 2 instead of just bracketing.

gysddn: Should I remove `RequireIntroducer` altogether and just have the tree like this:

```
- Require
  - Impls
    - Expr
    - Expr
```

jonmeow: `RequireIntroducer` should remain. For valid trees, we require every token to have a parse node (in some rare cases, two parse nodes).

gysddn: Updated

toolchain/parse/handle_require.cpp
```diff
@@ -0,0 +1,55 @@
...
+  context.PushState(State::RequireFinish);
+  context.PushState(State::Impls);
+  context.PushState(State::Expr);
+}
+
+auto HandleImpls(Context& context) -> void {
+  auto state = context.PopState();
+
+  if (auto impls = context.ConsumeIf(Lex::TokenKind::Impls)) {
+    // context.AddNode(NodeKind::ImplsTypeAs, *impls, state.has_error);
```
jonmeow:
```suggestion
```

Was this leftover from work? Should it be removed?

toolchain/parse/state.def
```diff
@@ -310,6 +310,11 @@ CARBON_PARSE_STATE(DeclNameAndParamsAfterParams)
 //   1. Alias
 //   2. DeclScopeLoop
 //
+// require ...
+// ^~~~~~~
+//   1. Require
+//   2. DeclScopeLoop
+//
```
jonmeow: Looking at surrounding entries, we're adding them lexically (alias, base, choice, etc). Can you please match that style, placing after `package`?

toolchain/parse/state.def
```diff
@@ -741,6 +746,40 @@ CARBON_PARSE_STATE(NamespaceFinish)
...
+// expression.
+// impls ...
+// ^~~~~
+//
+//   1. Expr
+//   2. ImplsFinish
+//
+CARBON_PARSE_STATE(Impls)
+
+// Finishes impls processing
```
```suggestion
// Finishes `impls` processing.
//
```

As above. Also, fix punctuation.

toolchain/parse/state.def
```diff
@@ -741,6 +746,40 @@ CARBON_PARSE_STATE(NamespaceFinish)
...
+// require ...
+//        ^
+//   1. Expr
+//   2. Impls
+//   3. RequireFinish
+//
+CARBON_PARSE_STATE(Require)
+
+// Handles the processing of `impls` in a `require` declaration after the type
+// expression.
```
jonmeow, line 759:
```suggestion
// expression.
//
```

Similar to other comments, let's put a blank line separating comment from example.

toolchain/parse/state.def
```diff
@@ -741,6 +746,40 @@ CARBON_PARSE_STATE(NamespaceFinish)
...
+
+// Finishes impls processing
+// impls type_expression ...
+//                      ^
+//   (state done)
+//
+
+CARBON_PARSE_STATE(ImplsFinish)
+
+// Finishes `require` processing
```
```suggestion
// Finishes `require` processing.
//
```
As above.

toolchain/parse/testdata/generics/interface/require_name.carbon
jonmeow: It looks like you've added most of these, but is there a test case with a missing semicolon?

e.g.:

```
interface Foo {
  require i32 impls Bar
}
```

I expect this to work fine, but having tests is important to both catch unexpected results and prevent regressions.

toolchain/parse/state.def
```diff
@@ -741,6 +746,40 @@ CARBON_PARSE_STATE(NamespaceFinish)
...
 CARBON_PARSE_STATE(Package)
 
+// Handles `require` after the introducer.
+//
+// require ...
+//        ^
+//   1. Expr
+//   2. Impls
+//   3. RequireFinish
+//
```
```suggestion
```

We typically don't put blank lines at the end of a comment block.

toolchain/parse/state.def
```diff
@@ -741,6 +746,40 @@ CARBON_PARSE_STATE(NamespaceFinish)
...
+CARBON_PARSE_STATE(Require)
+
+// Handles the processing of `impls` in a `require` declaration after the type
+// expression.
+// impls ...
+// ^~~~~
+//
+//   1. Expr
+//   2. ImplsFinish
+//
```
jonmeow: The syntax you're documenting here is for the error case, but you have it for the expression case. Let's document both paths.

Also note this suggestion is fixing extra blank lines.

```suggestion
// expr impls ...
//      ^~~~~
//   1. Expr
//   2. ImplsFinish
//
// expr ???
//     ^
//   (state done)
```

toolchain/parse/state.def
```diff
@@ -741,6 +746,40 @@ CARBON_PARSE_STATE(NamespaceFinish)
...
+CARBON_PARSE_STATE(Impls)
+
+// Finishes impls processing
+// impls type_expression ...
+//                      ^
```
```suggestion
// expr impls expr
//                ^
```

Let's include the `expr` before `impls` here.

toolchain/parse/state.def
```diff
@@ -741,6 +746,40 @@ CARBON_PARSE_STATE(NamespaceFinish)
...
+//   (state done)
+//
+
```
```suggestion
```

toolchain/parse/state.def
```diff
@@ -741,6 +746,40 @@ CARBON_PARSE_STATE(NamespaceFinish)
...
+//   1. Expr
+//   2. ImplsFinish
+//
+CARBON_PARSE_STATE(Impls)
+
+// Finishes impls processing
+// impls type_expression ...
+//                      ^
+//   (state done)
+//
```
```suggestion
// expr impls expr
//                ^
//   (state done)
```

Similar to `impls`, I think the syntax example can be more specific. When there is no following syntax handled, we omit the trailing "...". Also, since parsing only checks for an expression, let's only say "expr".

jonmeow (commented):
Thanks! I think the behavior looks good.

I'm suggesting changes to the state.def comments in order to try and match the file's style better. If I'm misreading something there, please let me know.

toolchain/parse/state.def
gysddn: Updated the comments

toolchain/parse/handle_require.cpp
gysddn: Oh yeah, sorry about that 😅

PR #4211, opened 2024-08-09
Address the LocIdAndInst::ReusingLoc TODO

jonmeow:
The TODO for switching to ReusingLoc (previously Untyped) had been there for a while, so I'm trying to address it here. The intent had been to be clearer about when the construction is validated, particularly so that we aren't accidentally accepting an incorrect NodeId. Note, this does fix an incorrect use of InvalidNodeId where NoLoc should've been called.

Since this changes the semantics of when `Parse::NodeId` is helpful in `typed_nodes.h`, I'm doing a pass to either refine or switch to `Parse::InvalidNodeId` where it compiles. I think most remaining `Parse::NodeId` examples are things we _should_ be able to refine with a little more work (versus before where `Parse::NodeId` also indicated `LocId` construction might be used).

I'm also changing context.h to use `requires` that match what `LocIdAndInst` has, I think it makes the diagnostics a little better. And note I do add an overload for `ImportIRInstId`, also matching `LocIdAndInst`, and widely used for import refs.

### DIFF
--- a/toolchain/check/context.h
+++ b/toolchain/check/context.h
@@ -67,7 +67,7 @@ class Context {
   // Adds an instruction to the current block, returning the produced ID.
   auto AddInst(SemIR::LocIdAndInst loc_id_and_inst) -> SemIR::InstId;
 
-  // Convenience for AddInst on specific instruction types.
+  // Convenience for AddInst with typed nodes.
   template <typename InstT>
     requires(SemIR::Internal::HasNodeId<InstT>)
   auto AddInst(decltype(InstT::Kind)::TypedNodeId node_id, InstT inst)
@@ -75,7 +75,8 @@ class Context {
     return AddInst(SemIR::LocIdAndInst(node_id, inst));
   }
 
-  // Convenience for AddInst on specific instruction types.
+  // Convenience for AddInst when reusing a location, which any instruction can
+  // do.
   template <typename InstT>
   auto AddInstReusingLoc(SemIR::LocId loc_id, InstT inst) -> SemIR::InstId {
     return AddInst(SemIR::LocIdAndInst::ReusingLoc<InstT>(loc_id, inst));
@@ -85,7 +86,7 @@ class Context {
   // rarely.
   auto AddInstInNoBlock(SemIR::LocIdAndInst loc_id_and_inst) -> SemIR::InstId;
 
-  // Convenience for AddInstInNoBlock on specific instruction types.
+  // Convenience for AddInstInNoBlock with typed nodes.
   template <typename InstT>
     requires(SemIR::Internal::HasNodeId<InstT>)
   auto AddInstInNoBlock(decltype(InstT::Kind)::TypedNodeId node_id, InstT inst)
--- a/toolchain/check/context.h
+++ b/toolchain/check/context.h
@@ -68,9 +68,17 @@ class Context {
   auto AddInst(SemIR::LocIdAndInst loc_id_and_inst) -> SemIR::InstId;
 
   // Convenience for AddInst on specific instruction types.
-  template <typename InstT, typename LocT>
-  auto AddInst(LocT loc_id, InstT inst) -> SemIR::InstId {
-    return AddInst(SemIR::LocIdAndInst(loc_id, inst));
+  template <typename InstT>
+    requires(SemIR::Internal::HasNodeId<InstT>)
+  auto AddInst(decltype(InstT::Kind)::TypedNodeId node_id, InstT inst)
+      -> SemIR::InstId {
+    return AddInst(SemIR::LocIdAndInst(node_id, inst));
+  }
+
+  // Convenience for AddInst on specific instruction types.
+  template <typename InstT>
+  auto AddInstReusingLoc(SemIR::LocId loc_id, InstT inst) -> SemIR::InstId {
+    return AddInst(SemIR::LocIdAndInst::ReusingLoc<InstT>(loc_id, inst));
   }
 
   // Adds an instruction in no block, returning the produced ID. Should be used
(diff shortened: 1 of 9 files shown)


jonmeow:
Note an alternative here is to just stop validating so much, removing `ReusingLoc` and just provide the `LocId` constructor.

I'm hoping @zygoloid has time to review before vanishing, but @josh11b I thought you might make a good backup since you added a lot of the NodeId restrictions.

toolchain/sem_ir/typed_insts.h
```diff
@@ -666,9 +669,9 @@ struct ImportRefLoaded {
 // `src_id`, by performing a final copy from source to destination, for types
 // whose initialization is not in-place.
 struct InitializeFrom {
-  // TODO: Make Parse::NodeId more specific.
-  static constexpr auto Kind = InstKind::InitializeFrom.Define<Parse::NodeId>(
-      {.ir_name = "initialize_from"});
+  static constexpr auto Kind =
+      InstKind::InitializeFrom.Define<Parse::InvalidNodeId>(
```
zygoloid: As I recall, the point of the typed node ID parameter here was to guarantee that the node ID for this instruction will always be of the given type. If we specify it as being `InvalidNodeId`, I think that means the node ID for this instruction is always invalid, that is, it has no associated parse node. That doesn't seem to be the case for `InitializeFrom` -- instead, it looks like it can be associated with any parse node, so `NodeId` seems right to me.

jonmeow: I could do that, but we don't built it using a node ID. `Convert` is constructing it with a `LocId`, and looking, I don't ever see a parse node directly used for a `Convert` call. So is the preference for, when it's never constructed with a NodeId, to still support constructing with an arbitrary NodeId just-in-case? Maybe that's something I can just put in a comment, to reflect that it's not validated? (fundamentally, I think this is true of all InvalidNodeId entries: they typically have locations, we just don't associate NodeIds)

toolchain/sem_ir/inst.h
```diff
@@ -305,11 +305,15 @@ inline auto operator<<(llvm::raw_ostream& out, TypedInst inst)
 // Associates a LocId and Inst in order to provide type-checking that the
 // TypedNodeId corresponds to the InstT.
 struct LocIdAndInst {
+  // Used when there is no associated location. Note, we should generally do our
+  // best to associate a location for diagnostics.
```
geoffromer:
```suggestion
  // Constructs a LocIdAndInst with no associated location. Note, we should generally do our
  // best to associate a location for diagnostics.
```

I find function comments much clearer when they focus primarily on the function's contract, i.e. what it does or what it returns, rather than how it's used.

toolchain/sem_ir/inst.h
```diff
@@ -305,11 +305,15 @@ inline auto operator<<(llvm::raw_ostream& out, TypedInst inst)
...
   template <typename InstT>
   static auto NoLoc(InstT inst) -> LocIdAndInst {
     return LocIdAndInst(LocId::Invalid, inst, /*is_untyped=*/true);
   }
 
+  // Used when a location comes from some other place, and can't trivially be
+  // validated to match the InstT's expected node ID.
```
geoffromer: Would something like this be accurate?

```suggestion
  // Constructs a LocIdAndInst that reuses the location associated with some other inst,
  // typically because `inst` doesn't have an explicit representation in the parse tree.
```

If so, I would find that a lot clearer.

toolchain/check/convert.cpp
```diff
@@ -150,14 +151,14 @@ static auto MakeElementAccessInst(Context& context, SemIR::LocId loc_id,
     // TODO: Add a new instruction kind for indexing an array at a constant
     // index so that we don't need an integer literal instruction here, and
     // remove this special case.
-    auto index_id = block.template AddInst<SemIR::IntLiteral>(
+    auto index_id = block.template AddInstReusingLoc<SemIR::IntLiteral>(
```
geoffromer, line 154: Is this actually reusing a location, or is it just using `ReusingLoc` in order to bypass typechecking? If the latter, maybe we should rename `ReusingLoc` to something like `UntypedLoc`? And/or maybe we should have a TODO to plumb the typed parse node into this function so that it can use `AddInst`?

toolchain/sem_ir/typed_insts.h
```diff
@@ -666,7 +669,9 @@ struct ImportRefLoaded {
 // `src_id`, by performing a final copy from source to destination, for types
 // whose initialization is not in-place.
 struct InitializeFrom {
-  // TODO: Make Parse::NodeId more specific.
+  // Note this Parse::NodeId is unused. InitializeFrom is only constructed by
+  // reusing locations.
+  // TODO: Figure out if there's a better way to handle this case.
```
geoffromer, line 674: How is this different from all the types that use `InvalidNodeId`, e.g. `Temporary`?

jonmeow: I'm not sure what semantic is intended here. See my comments to zygoloid above. But I think it's something we can handle separately from this PR.

toolchain/check/convert.cpp
jonmeow: The `loc_id` here is what's passed in. I don't know if a typed node could be plumbed, I'm guessing no due to the semantics of element access (e.g., the index could be the result of an expression). Note too the TODO above is to get rid of this, replacing with indexing at a constant index.

geoffromer (approved):
(I'm not merging because I can't tell whether @zygoloid's concerns have been addressed)

toolchain/check/context.h
```diff
@@ -68,19 +68,36 @@ class Context {
...
-  auto AddInst(LocT loc_id, InstT inst) -> SemIR::InstId {
-    return AddInst(SemIR::LocIdAndInst(loc_id, inst));
+  template <typename InstT>
+    requires(SemIR::Internal::HasNodeId<InstT>)
+  auto AddInst(decltype(InstT::Kind)::TypedNodeId node_id, InstT inst)
+      -> SemIR::InstId {
+    return AddInst(SemIR::LocIdAndInst(node_id, inst));
+  }
+
+  // Convenience for AddInst on specific instruction types.
```
chandlerc: Maybe update the comment to mention the distinction of reusing the location and the lack of requires?

chandlerc:
> (I'm not merging because I can't tell whether @zygoloid's concerns have been addressed)

FWIW, while I'm not sure they have either, I do think they've been captured in a TODO and we should be able to revisit this in the future. I don't think we need to block this PR, especially with vacation.

chandlerc (approved):
FYI, just noticed one thing in-line in passing. LGTM, merge whenever.

toolchain/check/context.h
jonmeow: Adjusted

geoffromer:
> I don't think we need to block this PR, especially with vacation.

Just to be clear, I didn't intend to block this (hence my approval), I just intended to let @jonmeow make the decision about whether it was ready to merge.

PR #4212, opened 2024-08-12
Update to a recent LLVM version and the changes to LLVM IR.

josh11b:


jonmeow (approved):
Thanks for the update!

PR #4215, opened 2024-08-13
Error on non-constant parameters to a type.

jonmeow:
At present, this is a crash bug. I don't know whether this is the best fix, but I figure it'll work until zygoloid has a chance to look.


chandlerc (approved):
LGTM

chandlerc:
Seems pretty straightforward so grabbed this review while I was at the summit. Marking for merge.

PR #4216, opened 2024-08-14
Refactor testing exe path and benchmark main handling.

chandlerc:
Consolidates both main libraries into `//testing/base`, and factors out the exe path handling for benchmarks and unit tests into a common library to remove duplication. Refactors how that logic is managed to be cleaner and avoid a confusing bool that came up in code review.

Updates all the tests and benchmarks that use these. I still need to update other benchmarks to use the same main, but I wanted to keep this PR somewhat minimal.

This also fixes a bug noticed in passing that the compilation benchmark didn't have the required dependency on the benchmark library itself, just the benchmark main library.


testing/base/testing_exe_path.h
```diff
@@ -2,8 +2,8 @@
 // Exceptions. See /LICENSE for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-#ifndef CARBON_TESTING_BASE_GTEST_MAIN_H_
-#define CARBON_TESTING_BASE_GTEST_MAIN_H_
+#ifndef CARBON_TESTING_BASE_TESTING_EXE_PATH_H_
```
jonmeow: Maybe rename this file to just `exe_path` since `testing` is already in the directory structure?

I mean, we do have common:exe_path, but I don't think that the name mirroring is a problem.

testing/base/BUILD
```diff
@@ -104,3 +110,27 @@ cc_test(
         "@googletest//:gtest",
     ],
 )
+
+cc_library(
+    name = "testing_exe_path",
+    testonly = 1,
+    srcs = ["testing_exe_path.cpp"],
```
jonmeow: Why isn't testing_exe_path.cpp in this change?

testing/base/testing_exe_path.h
chandlerc: The name mirroring is why I had the testing...

It also does something quite different -- this even depends on `//common:exe_path`... Is there maybe a better but different name?

testing/base/BUILD
chandlerc: Because I forgot to `git add` it. Doh!

jonmeow (approved):
Per discussion, maybe global_exe_path

testing/base/testing_exe_path.h
chandlerc: Going with `global_exe_path` which telegraphs the fact that this is really a global storage.

PR #4217, opened 2024-08-14
Adjust check's node formatting in crash output.

jonmeow:
I'm trying to make the line of code more clearly nested in crash output (the way it is, I sometimes forget about it). Also, `Check::HandleFunctionDecl` is the old naming scheme, it's now all `Check::HandleParseNode`, so I'm replacing that.

Before:

```
3.	extern_library_owner.carbon:6:1: Check::HandleFunctionDecl
extern fn F();
^~~~~~~~~~~~~~
 #0 0x0000564c0057ef1d llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) ...
```

After:

```
3.	extern_library_owner.carbon:6:1: checking FunctionDecl
          extern fn F();
          ^~~~~~~~~~~~~~
 #0 0x00005629029ffd9d llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) ...
```


chandlerc (approved):
LG

PR #4218, opened 2024-08-15
change tuple index

brymer-meneses:
This changes the tuple index from tuple[0] to tuple.0 in accordance with the accepted propsal https://github.com/carbon-language/carbon-lang/pull/3646

I messed up syncing to trunk on my original PR https://github.com/carbon-language/carbon-lang/pull/4186, that's why I'm starting on a blank state. Please let me know if I missed incorporating a change from my prior PR.


jonmeow (approved):
Thanks! Looks right here, I'm going to merge.

jonmeow:
Sorry, can you run autoupdate again? Looks like there's still some small difference in the lowering output, a line of code change (sometimes these take a couple autoupdate runs to get fixed, when autoupdate causes the line numbers to change)

brymer-meneses:
Sorry, about that. I think that should be resolved by now.

PR #4219, opened 2024-08-15
Factor library names into their own ID structure.

jonmeow:
This supports distinguishing between unset, Default, and "incorrect but already diagnosed, do not use" for `extern library` logic.


chandlerc (approved):
LG

PR #4220, opened 2024-08-15
Implement 'extern library' support for functions.

jonmeow:
Support for types (particularly classes) is left as a TODO.

There's also an issue I'm observing with a "define in impl" test, but this is probably an issue with resolving the prior declaration which is imported indirectly. The PR was already feeling big, so I'm choosing to cut here.

Note, this does not implement the rule "The owning library's API file must import the `extern` declaration, and must also contain a declaration."

### DIFF
--- a/toolchain/check/handle_modifier.cpp
+++ b/toolchain/check/handle_modifier.cpp
@@ -34,9 +34,10 @@ static auto DiagnoseNotAllowedWith(Context& context, Parse::NodeId first_node,
       .Emit();
 }
 
-// Handles the main modifier keyword (either alone, such as `private`, or the
-// first in a complex modifier, such as `extern library ...`). If valid, adds it
-// to the modifier set and returns true. Otherwise, diagnoses and returns false.
+// Handles the keyword that starts a modifier. This may a standalone keyword,
+// such as `private`, or the first in a complex modifier, such as `extern` in
+// `extern library ...`. If valid, adds it to the modifier set and returns true.
+// Otherwise, diagnoses and returns false.
 static auto HandleModifier(Context& context, Parse::NodeId node_id,
                            KeywordModifierSet keyword) -> bool {
   auto& s = context.decl_introducer_state_stack().innermost();
--- a/toolchain/check/handle_modifier.cpp
+++ b/toolchain/check/handle_modifier.cpp
@@ -34,6 +34,8 @@ static auto DiagnoseNotAllowedWith(Context& context, Parse::NodeId first_node,
       .Emit();
 }
 
+// Handles the main modifier keyword. If valid, adds it to the modifier set and
+// returns true. Otherwise, diagnoses and returns false.
 static auto HandleModifier(Context& context, Parse::NodeId node_id,
                            KeywordModifierSet keyword) -> bool {
   auto& s = context.decl_introducer_state_stack().innermost();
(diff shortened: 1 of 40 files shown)


jonmeow:
Depends on #4219, split off LibraryNameId to try to make this smaller.

toolchain/check/testdata/class/fail_import_misuses.carbon
```diff
@@ -23,7 +23,7 @@ library "b";
 
 import library "a";
 
-// CHECK:STDERR: fail_b.carbon:[[@LINE+10]]:1: ERROR: Only one library can declare `class Empty` without `extern`.
+// CHECK:STDERR: fail_b.carbon:[[@LINE+10]]:1: ERROR: Redeclaration of `class Empty` is redundant.
```
geoffromer, line 26: This seems like a mild regression, if I understand the rules correctly: this isn't merely redundant, it's contradictory: both libraries are asserting that they own the same name. Is that right?

If so, I don't think this is blocking, but it might be worth having a TODO somewhere.

toolchain/check/handle_modifier.cpp
```diff
@@ -34,6 +34,8 @@ static auto DiagnoseNotAllowedWith(Context& context, Parse::NodeId first_node,
       .Emit();
 }
 
+// Handles the main modifier keyword. If valid, adds it to the modifier set and
```
geoffromer: What's the _main_ modifier keyword?

toolchain/check/testdata/class/fail_import_misuses.carbon
jonmeow: I've rephrased the TODO in handle_class.cpp, line 167, to try to be clearer that `extern` for classes still needs work.

Regarding this diagnostic, it stems from incomplete implementation (both before and after) so I think we shouldn't anchor on it.

toolchain/check/handle_modifier.cpp
jonmeow: I've tried to give a couple more examples of what this handles for modifiers with multiple keywords.

toolchain/check/handle_modifier.cpp
```diff
@@ -34,6 +34,9 @@ static auto DiagnoseNotAllowedWith(Context& context, Parse::NodeId first_node,
       .Emit();
 }
 
+// Handles the main modifier keyword (either alone, such as `private`, or the
+// first in a complex modifier, such as `extern library ...`). If valid, adds it
+// to the modifier set and returns true. Otherwise, diagnoses and returns false.
```
geoffromer: This sounds to me like it's saying that in `private abstract fn Foo()`, this function handles the `private` but not the `abstract`, which doesn't seem right. After talking to josh11b I think I understand what this is trying to say; assuming this is correct, I would find it much clearer:

```suggestion
// Handles a modifier keyword, which may be a complete modifier (such as `private`), or the
// first token of a complex modifier (such as `extern library ...`). If valid, adds it
// to the modifier set and returns true. Otherwise, diagnoses and returns false.
```

jonmeow: In the `extern library` modifier, `library` is also a keyword. I would think a keyword that helps form a modifier is a "modifier keyword". However, this is not called for the "library" modifier keyword. So I think saying this handles "a modifier keyword" could be misleading.

I've tried a different phrasing, does that work for you?

geoffromer (approved):

PR #4221, opened 2024-08-16
Initial support for binding patterns in SemIR

geoffromer:
Introduces the `BindingPattern` and `SymbolicBindingPattern` insts, and a separate stack of pattern blocks that they are emitted into. The intent is to generate the corresponding pattern-matching insts (like `BindName`) from them in a separate pass, but that is deferred to future PRs.

See [here](https://docs.google.com/document/d/1U_vQH17V893J9aF1LJXUnFYBNSs2MjKl4bJPaWCB2zo/edit?usp=sharing&resourcekey=0-w0xGYZ0An31Kpz-wvzSXwQ) for the design this is based on, but note that during review we have chosen to deviate from that design by putting the patterns in separate blocks, and omitting the "forward references" from a `BindingPattern` to its corresponding `BindName`. This in turn necessitates having separate inst kinds for symbolic and non-symbolic binding patterns.


toolchain/sem_ir/inst_namer.cpp
```diff
@@ -416,6 +416,11 @@ auto InstNamer::CollectNamesInBlock(ScopeId scope_id,
             sem_ir_.entity_names().Get(inst.entity_name_id).name_id);
         continue;
       }
+      case CARBON_KIND(BindingPattern inst): {
+        // See comment on AddrPattern
```
jonmeow:
```suggestion
        // See comment on AddrPattern.
```

Please try to keep punctuation on comments.

That said, is this referring to the TODO? Should this be a TODO as well?

```suggestion
        // TODO: See comment on AddrPattern.
```

toolchain/sem_ir/inst_kind.def
```diff
@@ -75,6 +75,7 @@ CARBON_SEM_IR_INST_KIND(Param)
 CARBON_SEM_IR_INST_KIND(PointerType)
 CARBON_SEM_IR_INST_KIND(ReturnExpr)
 CARBON_SEM_IR_INST_KIND(Return)
+CARBON_SEM_IR_INST_KIND(BindingPattern)
```
jonmeow: Things like this are lexically sorted, can you please add using lexical ordering?

Note, this comes up in multiple files, e.g. file.cpp and handle.cpp. Please look through this change and fix to follow lexical ordering when surrounding code is presently lexically ordered.

toolchain/sem_ir/function.cpp
```diff
@@ -46,11 +46,21 @@ auto Function::GetParamFromParamRefId(const File& sem_ir, InstId param_ref_id)
...
   if (auto addr_pattern = ref.TryAs<SemIR::AddrPattern>()) {
     param_ref_id = addr_pattern->inner_id;
     ref = sem_ir.insts().Get(param_ref_id);
   }
 
+  if (auto binding_pattern = ref.TryAs<SemIR::BindingPattern>()) {
+    param_ref_id = binding_pattern->bind_inst_id;
+    ref = sem_ir.insts().Get(param_ref_id);
+  }
+
```
jonmeow: Why do you unwrap binding patterns twice here?

toolchain/check/eval.cpp
```diff
@@ -1401,6 +1401,14 @@ auto TryEvalInstInContext(EvalContext& eval_context, SemIR::InstId inst_id,
       return MakeConstantResult(eval_context.context(), bind, Phase::Symbolic);
     }
 
+    // TODO: Does it make sense to evaluate both BindSymbolicName and
+    // BindingPattern?
```
jonmeow: In general, instructions should have a constant associated -- the only exception I'm aware of is per line 1474, instructions that are either always non-constants or not used in expressions. I think you can probably observe the consequence for BindSymbolicName if you just make it return non-constant; I'd expect it symbolic constants to stop working.

toolchain/check/handle_pattern_list.cpp
```diff
@@ -44,7 +44,17 @@ auto HandleParseNode(Context& context, Parse::TuplePatternId node_id) -> bool {
   // make use of it.
   auto refs_id = context.param_and_arg_refs_stack().EndAndPop(
       Parse::NodeKind::TuplePatternStart);
-  context.node_stack().Push(node_id, refs_id);
+  // TODO: do this at full-pattern level.
+  std::vector<SemIR::InstId> inner_param_insts;
```
jonmeow:
```suggestion
  llvm::SmallVector<SemIR::InstId> inner_param_insts;
```

We're generally using LLVM types where they exist.

toolchain/check/handle_pattern_list.cpp
```diff
@@ -44,7 +44,17 @@ auto HandleParseNode(Context& context, Parse::TuplePatternId node_id) -> bool {
...
+  for (SemIR::InstId inst_id : context.inst_blocks().Get(refs_id)) {
```
jonmeow:
```suggestion
  auto refs_block = context.inst_blocks().Get(refs_id);
  inner_param_insts.reserve(refs_block.size());
  for (SemIR::InstId inst_id : refs_block) {
```

Maybe reserve the block size, since it seems to be known?

toolchain/check/handle_pattern_list.cpp
jonmeow: Since this is adding substantial logic, would you please add a comment explaining what you're doing here? Why does it unwrap the block instead of leaving that for the consumer?

toolchain/check/eval.cpp
```diff
@@ -1401,6 +1401,14 @@ auto TryEvalInstInContext(EvalContext& eval_context, SemIR::InstId inst_id,
...
+    case CARBON_KIND(SemIR::BindingPattern pattern): {
+      auto bind_inst = eval_context.insts().Get(pattern.bind_inst_id);
+      return TryEvalInstInContext(eval_context, pattern.bind_inst_id,
+                                  bind_inst);
```
jonmeow: This looks a little odd to me, we don't call TryEvalInstInContext like this anywhere else. The call may result in a duplicate evaluation of `bind_inst_id`; had you tried `eval_context.GetConstantValue` similar to other handlers?

(see also the comment on in handle_binding_pattern.cpp regarding placeholders)

toolchain/check/handle_binding_pattern.cpp
```diff
@@ -64,6 +64,13 @@ static auto HandleAnyBindingPattern(Context& context, Parse::NodeId node_id,
   auto push_bind_name = [&](SemIR::InstId bind_id) {
     context.node_stack().Push(node_id, bind_id);
     if (is_generic && !is_associated_constant) {
+      // TODO: this is a temporary workaround until bind_id has a more
+      // consistent kind.
+      if (auto binding_pattern =
+              context.insts().TryGetAs<SemIR::BindingPattern>(bind_id);
+          binding_pattern.has_value()) {
```
jonmeow:
```suggestion
              context.insts().TryGetAs<SemIR::BindingPattern>(bind_id)) {
```

Does this work as an implicit boolean test? I think that's what we're generally doing with things that return an optional.

toolchain/lower/handle.cpp
```diff
@@ -213,6 +213,11 @@ auto HandleInst(FunctionContext& context, SemIR::InstId /*inst_id*/,
   }
 }
 
+auto HandleInst(FunctionContext& /*context*/, SemIR::InstId /*inst_id*/,
+                SemIR::BindingPattern /*inst*/) -> void {
+  // Patterns are not lowered.
+}
```
jonmeow: Instead of adding a stub function, what happens if you set `.is_lowered = false` on the instruction?

toolchain/sem_ir/formatter.cpp
```diff
@@ -855,6 +855,10 @@ class FormatterImpl {
     out_ << "}";
   }
 
+  auto FormatInstructionRHS(BindingPattern inst) -> void {
+    FormatArgs(inst.entity_name_id);
+  }
```
jonmeow: Why doesn't this use the default implementation?

toolchain/check/testdata/class/generic/fail_todo_use.carbon
```diff
@@ -0,0 +1,212 @@
...
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+// TIP: To test this file alone, run:
+// TIP:   bazel test //toolchain/testing:file_test --test_arg=--file_tests=toolchain/check/testdata/class/generic/fail_todo_use.carbon
+// TIP: To dump output, run:
+// TIP:   bazel run //toolchain/testing:file_test -- --dump_output --file_tests=toolchain/check/testdata/class/generic/fail_todo_use.carbon
+
+class Class(T:! type) {
```
jonmeow: It looks like you're adding this file, how does it tie into the change? It also looks pretty similar to `init.carbon`

toolchain/check/import_ref.cpp
```diff
@@ -684,11 +690,20 @@ class ImportRefResolver {
       auto inst = import_ir_.insts().Get(ref_id);
       auto addr_inst = inst.TryAs<SemIR::AddrPattern>();
 
+      auto pattern_id = ref_id;
       auto bind_id = ref_id;
       auto param_id = ref_id;
```
jonmeow: Since you're adding another of these, what do you think of moving the declaration bind_id to be right before the `binding_pattern` declaration, and the `param_id` declaration to be right before the `bind_inst` declaration? i.e., only declare the ID when it can hold a distinct value from the prior ID. I think that'd get rid of some of the `bind_id = pattern_id` reassignments.

toolchain/check/handle_binding_pattern.cpp
```diff
@@ -178,8 +192,16 @@ static auto HandleAnyBindingPattern(Context& context, Parse::NodeId node_id,
       // formed its initializer.
       // TODO: For general pattern parsing, we'll need to create a block to hold
       // the `let` pattern before we see the initializer.
-      auto bind_id = context.AddPlaceholderInstInNoBlock(
-          make_bind_name(cast_type_id, SemIR::InstId::Invalid));
+      SemIR::LocIdAndInst bind_name =
+          make_bind_name(cast_type_id, SemIR::InstId::Invalid);
+      auto bind_id = context.AddPlaceholderInstInNoBlock(bind_name);
```
jonmeow: The `AddInst` call below creates a constant for `BindingPattern` which relies on the constant from this. The reason `AddPlaceholderInstInNoBlock` is updated by `ReplaceInstBeforeConstantUse` is because we can't use it for constants before it's replaced. As a consequence, this code is going to lead to errors.

Also note, the `AddInst` below is going to add the `BindingPattern` to a block in an incorrect order. e.g., look at the ordering of binding_pattern and bind_symbolic_name in toolchain/check/testdata/interface/no_prelude/generic_binding_after_assoc_const.carbon

Rather than trying to add the `BindingPattern` here, I suggest looking at how the binding is added to a block.

toolchain/lower/file_context.cpp
```diff
@@ -297,6 +297,10 @@ auto FileContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
       bind_name_id = addr->inner_id;
     }
     auto bind_name = sem_ir().insts().Get(bind_name_id);
+    if (auto pattern = bind_name.TryAs<SemIR::BindingPattern>()) {
+      bind_name_id = pattern->bind_inst_id;
+      bind_name = sem_ir().insts().Get(bind_name_id);
+    }
```
jonmeow: The `addr` check above is doing the same thing, why does this look so different?

jonmeow (commented):
To be sure I understand correctly, is this change supposed to only be adding `BindingPattern` without really using it? Can you please use the PR description to elaborate a little on what's being implemented here?

toolchain/sem_ir/typed_insts.h
```diff
@@ -302,6 +302,17 @@ struct BindValue {
...
 };
 
+// Represents a binding pattern value with name `entity_name_id`.
+// `bind_inst_id` is the instruction which actually binds the value.
+struct BindingPattern {
+  static constexpr auto Kind = InstKind::BindingPattern.Define<Parse::NodeId>(
+      {.ir_name = "binding_pattern"});
+
+  TypeId type_id;
+  EntityNameId entity_name_id;
```
jonmeow, line 317: Sorry, one more question. It looks like each place this is set, it's actually coming from `bind_inst_id`. Why is it provided in both places?

toolchain/check/eval.cpp
geoffromer: Ah, OK. And `BindingPattern` isn't used in expressions, so I've added it to the block below.

toolchain/check/handle_binding_pattern.cpp
geoffromer: > The `AddInst` call below creates a constant for `BindingPattern` which relies on the constant from this. The reason `AddPlaceholderInstInNoBlock` is updated by `ReplaceInstBeforeConstantUse` is because we can't use it for constants before it's replaced. As a consequence, this code is going to lead to errors.

IIUC this should be fixed now that evaluation of `BindingPattern` doesn't evaluate `bind_inst_id` anymore.

> Also note, the `AddInst` below is going to add the `BindingPattern` to a block in an incorrect order. e.g., look at the ordering of binding_pattern and bind_symbolic_name in toolchain/check/testdata/interface/no_prelude/generic_binding_after_assoc_const.carbon
>
> Rather than trying to add the `BindingPattern` here, I suggest looking at how the binding is added to a block.

I believe the order I have is correct: as per the [design](https://docs.google.com/document/d/1U_vQH17V893J9aF1LJXUnFYBNSs2MjKl4bJPaWCB2zo/edit?usp=sharing&resourcekey=0-w0xGYZ0An31Kpz-wvzSXwQ), the pattern IR comes before the match IR.

toolchain/check/import_ref.cpp
geoffromer: Good idea. I also found a couple other simplifications.

toolchain/lower/file_context.cpp
geoffromer: No reason, other than my newbie flailing. Fixed.

toolchain/lower/handle.cpp
geoffromer: I get this:

> FATAL failure at toolchain/lower/function_context.cpp:65: Encountered an instruction that isn't expected to lower. It's possible that logic needs to be changed in order to stop showing this instruction in lowered contexts. Instruction: {kind: BindingPattern, arg0: entity_name2, arg1: inst+74, type: type(inst+4)}

toolchain/sem_ir/formatter.cpp
geoffromer: I'm not sure anymore; this had no effect. I need it now, though, to work around the removal of `entity_name_id`.

toolchain/sem_ir/function.cpp
geoffromer: Possibly a rebase gone awry? Fixed.

toolchain/sem_ir/inst_kind.def
geoffromer: Done. Let me know if I missed anything.

toolchain/check/testdata/class/generic/fail_todo_use.carbon
geoffromer: Sorry, this was a false start that I overlooked amidst all the other testdata edits. Removed.

toolchain/sem_ir/typed_insts.h
geoffromer: It looks like this field isn't actually used, so I've removed it.

toolchain/check/handle_pattern_list.cpp
geoffromer: Yeah, unwrapping in the consumer is cleaner. I've fixed that and added comments.

geoffromer (commented):
> To be sure I understand correctly, is this change supposed to only be adding `BindingPattern` without really using it?

This PR "really uses it" in the way that it's designed to be used. In particular, we traverse the `BindingPattern`s in order to emit the corresponding `BindName`s.

> Can you please use the PR description to elaborate a little on what's being implemented here?

How's this?

toolchain/sem_ir/formatter.cpp
jonmeow: Sorry, I must be missing something. Even with the new changes, can you elaborate why the default behavior is incorrect?

toolchain/check/handle_binding_pattern.cpp
jonmeow: Let me try to dig a little more into the ordering issue.

Stepping back to the design, what I see is:

```
%x.patt: pattern = binding_pattern `x`, i32
%y.patt: pattern = expr_pattern %y.expected in {
  %GetY: <function> = name_ref @GetY
  %y.expected: i32 = call %GetY()
}
```

So what I'd expect is that `BindingPattern`, under that model, looks like:

```
TypeId type_id;
EntityNameId entity_name_id;
```

Should this be breaking further from the current bind model? e.g., perhaps this should break further from the current binding model?

---

For comparison, the current state is:

```
TypeId type_id;
InstId bind_inst_id;
```

The issue with `bind_inst_id` here is that it refers to a _later_ instruction. So if you think about this in terms of an IR, the execution of this instruction depends on a later instruction.

Note too the ordering isn't always like that. e.g., here's a snippet for `Param`:\

```
// CHECK:STDOUT:     %a.loc11_8.1: i32 = param a
// CHECK:STDOUT:     %.loc11_8: i32 = binding_pattern a, @Add.%a
```

So I think there is inconsistent ordering going on here. I also see you're partly solving this by declaring all binding patterns non-constant. Should we ever be able to constant-evaluate binding patterns? e.g., I'd wonder how that affects something like:

```
fn F(x:! i32) -> bool {
  match (x) {
    case 1 => { return true; }
    default => { return false; }
  }
}
```

In this example, `case 1` is runtime.

jonmeow: Note, it's also possible @zygoloid might have separate thoughts here. I'm at least partly trying to pull apart how I'm reading the SemIR there.

geoffromer: > So what I'd expect is that `BindingPattern`, under that model, looks like:
> 
> ```
> TypeId type_id;
> EntityNameId entity_name_id;
> ```

Conceptually I agree, but that runs into the issue discussed [here](https://docs.google.com/document/d/1U_vQH17V893J9aF1LJXUnFYBNSs2MjKl4bJPaWCB2zo/edit?resourcekey=0-w0xGYZ0An31Kpz-wvzSXwQ&tab=t.0#heading=h.gnlt773jqu04): in order to handle patterns like `(n: i32, n)`, we need to add the name to lookup immediately, so we need to allocate the inst that it resolves to immediately. That's why I'm creating the `bind_name` inst right now, even though I'm not adding it to a block until later. I've added some comments to try to clarify.

> Should this be breaking further from the current bind model? e.g., perhaps this should break further from the current binding model?

I don't understand the current binding model well enough to say; I'm just deferring to the design doc.

> For comparison, the current state is:
> 
> ```
> TypeId type_id;
> InstId bind_inst_id;
> ```
> 
> The issue with `bind_inst_id` here is that it refers to a _later_ instruction. So if you think about this in terms of an IR, the execution of this instruction depends on a later instruction.

`binding_pattern` insts aren't ever really executed, though, they're just traversed during the check phase in order to generate the pattern-match IR. `bind_inst_id` isn't an execution dependency of the `binding_pattern`, it's part of the pattern-match IR that we've precomputed and cached.

> 
> Note too the ordering isn't always like that. e.g., here's a snippet for `Param`:\
> 
> ```
> // CHECK:STDOUT:     %a.loc11_8.1: i32 = param a
> // CHECK:STDOUT:     %.loc11_8: i32 = binding_pattern a, @Add.%a
> ```
> 
> So I think there is inconsistent ordering going on here.

Good point. I have a fix for that later in my stack of commits, but it would widen the scope of this review even further, so I've fixed it here with a temporary hack to treat it like `bind_name`.

> I also see you're partly solving this by declaring all binding patterns non-constant. Should we ever be able to constant-evaluate binding patterns?

My intent was to declare that all binding patterns are _non-expressions_. Did I do that wrong?

toolchain/sem_ir/formatter.cpp
geoffromer: Consider e.g. this line from `toolchain/check/testdata/tuple/two_elements.carbon`:

```
// CHECK:STDOUT: %.loc11_5: %.3 = binding_pattern v, @__global_init.%v
```

With the default formatter, it would instead be:

```
// CHECK:STDOUT: %.loc11_5: %.3 = binding_pattern @__global_init.%v
```

As you pointed out, conceptually a `binding_pattern` consists of a type and a name (in this case `v`), so it seems desirable for the pretty-printed form to include that name. If anything, arguably we should omit the `@__global_init.%v` since it's not really an operand of the inst (and as you pointed out, it's a forward reference to an inst that will come later).

toolchain/check/handle_binding_pattern.cpp
jonmeow: Sorry, I think I'm somewhat slow at this just because I haven't really thought through the pattern design. Maybe a VC would be helpful to discuss this? I think I have a different understanding of the design, and it may be most helpful to talk through. Would you like to meet this week, or wait for zygoloid to be back since he's more familiar with how patterns work?

Let me ask though, the current design talks about putting `binding_pattern` in one block (a pattern block), and using that to run a match, resulting in `bind_name` in the statement block. In the design's model, `bind_name` is not associated cross-block. What's your plan here? Had you considered starting with that separation, rather than trying to blend it into the current setup and separate later?

There may be a few consequences of using a separate block to consider, for how it would affect this change:

1. The instruction would not be in a statement block, and so should not be seen by lowering.
    - The stub in lowering is only necessary because it's being added to a lowered block.
2. Because this is a separate block, things such as `n` in `(n: i32, n)` would probably want to resolve to the `BindingPattern` rather than a `BindName`.
    - AFAIU this behaves similar to a nested scope, adding lexical name lookup entries that are not accessible later.
3. In the design, there's no clear connection between the `BindingPattern` and a `BindName`.
   - Instead, the pattern is "matched" and results in the `br` code.

Note I'm also realizing looking again at:

```
%x.patt: pattern = binding_pattern `x`, i32
```

I keep getting the structure of `BindingPattern` wrong. To detail how you'd get this SemIR output:

```
// In builtin_inst_kind.def:
CARBON_SEM_IR_BUILTIN_INST_KIND(PatternType, "pattern")

// In typed_insts.h:
struct BindingPattern {
  // The type_id is always BuiltinPatternType.
  TypeId type_id;
  EntityNameId entity_name_id;
  TypeId bind_type_id;
}
```

i.e.:

```
%x.patt: pattern = binding_pattern `x`, i32
                   ^~~~~~~~~~~~~~~ kind == BindingPattern
         ^~~~~~~ type_id == BuiltinPatternType
                                   ^~~ entity_name_id == `x`
                                        ^~~ bind_type_id == i32
```

Is this helpful?

> `binding_pattern` insts aren't ever really executed, though, they're just traversed during the check phase in order to generate the pattern-match IR. `bind_inst_id` isn't an execution dependency of the `binding_pattern`, it's part of the pattern-match IR that we've precomputed and cached.

Are you thinking of execution in terms of lowering to LLVM IR? SemIR may only be re-processed for that today, but for example, we will probably also use SemIR for unused code analysis (stripping out unreferenced instructions) and evaluation of constexpr functions, just as a couple examples. The ability to evaluate SemIR using information accumulation seems important to maintain.

> My intent was to declare that all binding patterns are _non-expressions_. Did I do that wrong?

Where the comment says "These cases are either not expressions or not constant," are you expecting that was written to include patterns? I might just be reading this differently: since the toolchain hasn't implemented actual pattern matching, it simply wasn't considered by the comment; the comment doesn't speak about patterns, positive or negative. Nor does the pattern matching design doc talk about constants.

I'm not sure what the plan for constant evaluation of patterns would be; maybe a TODO? I don't expect it to be resolved by this PR. But, since there's not a design that gets into the deep details, I'm suspicious constants will be needed -- I'd call it an open question.

geoffromer: > Sorry, I think I'm somewhat slow at this just because I haven't really thought through the pattern design. Maybe a VC would be helpful to discuss this? I think I have a different understanding of the design, and it may be most helpful to talk through. Would you like to meet this week, or wait for zygoloid to be back since he's more familiar with how patterns work?

A VC sounds like a good idea, but my availability for that is uncertain at the moment (I'm WFH covering for a sick co-parent). I'll ping you once I know what my situation is, and in the meantime I'll respond here to what I can.

> Let me ask though, the current design talks about putting `binding_pattern` in one block (a pattern block), and using that to run a match, resulting in `bind_name` in the statement block. In the design's model, `bind_name` is not associated cross-block. What's your plan here? Had you considered starting with that separation, rather than trying to blend it into the current setup and separate later?

Here again, I'm somewhat blindly following the design doc:

>>>TODO: Should the pattern block be a separate inst block, or should it just be a series of insts that appear in the block where the pattern match occurs? The separate block seems "tidier", but it's not clear that it's necessary or adds value.
>>> * For now, let's just put these _pattern instructions directly in the context where the pattern matching is performed.

That's obviously not very definitive, so I'm certainly open to having a separate pattern block, but that's where I started from.

> 2. Because this is a separate block, things such as `n` in `(n: i32, n)` would probably want to resolve to the `BindingPattern` rather than a `BindName`.
>    * AFAIU this behaves similar to a nested scope, adding lexical name lookup entries that are not accessible later.

Even if the pattern insts are a separate block, expressions within a pattern would not be evaluated as part of that block; they need to be evaluated as part of pattern-matching, not as part of pattern definition (in the design doc, that's represented by modeling expression patterns with `splice_block`s). And in the pattern-matching context, we want `n` to refer to the inst that will actually produce the value of `n`, and that's a `BindName`.

> 3. In the design, there's no clear connection between the `BindingPattern` and a `BindName`.
>    
>    * Instead, the pattern is "matched" and results in the `br` code.

Yes, that matches my understanding, but I think that's true regardless of whether they're in separate blocks.

> Note I'm also realizing looking again at:
> 
> ```
> %x.patt: pattern = binding_pattern `x`, i32
> ```
> 
> I keep getting the structure of `BindingPattern` wrong. To detail how you'd get this SemIR output:
> 
> ```
> // In builtin_inst_kind.def:
> CARBON_SEM_IR_BUILTIN_INST_KIND(PatternType, "pattern")
> 
> // In typed_insts.h:
> struct BindingPattern {
>   // The type_id is always BuiltinPatternType.
>   TypeId type_id;
>   EntityNameId entity_name_id;
>   TypeId bind_type_id;
> }
> ```
> 
> i.e.:
> 
> ```
> %x.patt: pattern = binding_pattern `x`, i32
>                    ^~~~~~~~~~~~~~~ kind == BindingPattern
>          ^~~~~~~ type_id == BuiltinPatternType
>                                    ^~~ entity_name_id == `x`
>                                         ^~~ bind_type_id == i32
> ```

I think what you're getting at here is that I deviated from the design doc by having `type_id` represent the type of the binding (and omitting `bind_type_id`), rather than making it always be `BuiltinPatternType`. My intuition was that when we we start doing nontrivial type-checking of complex patterns, it's going to be much simpler to have the relevant types in the same place they are for all the other inst kinds, so I wanted to see if I could make that work. So far I haven't run into any problems with this approach, but it should be relatively easy to change course if that turns out to be necessary.

> > `binding_pattern` insts aren't ever really executed, though, they're just traversed during the check phase in order to generate the pattern-match IR. `bind_inst_id` isn't an execution dependency of the `binding_pattern`, it's part of the pattern-match IR that we've precomputed and cached.
> 
> Are you thinking of execution in terms of lowering to LLVM IR? SemIR may only be re-processed for that today, but for example, we will probably also use SemIR for unused code analysis (stripping out unreferenced instructions) and evaluation of constexpr functions, just as a couple examples. The ability to evaluate SemIR using information accumulation seems important to maintain.
> 
> > My intent was to declare that all binding patterns are _non-expressions_. Did I do that wrong?
> 
> Where the comment says "These cases are either not expressions or not constant," are you expecting that was written to include patterns? I might just be reading this differently: since the toolchain hasn't implemented actual pattern matching, it simply wasn't considered by the comment; the comment doesn't speak about patterns, positive or negative. Nor does the pattern matching design doc talk about constants.
> 
> I'm not sure what the plan for constant evaluation of patterns would be; maybe a TODO? I don't expect it to be resolved by this PR. But, since there's not a design that gets into the deep details, I'm suspicious constants will be needed -- I'd call it an open question.

Throughout above comments, I'm unclear on what it would mean to "evaluate" a pattern, apart from matching it against a scrutinee. This is probably a good topic for the VC.

jonmeow: > > Sorry, I think I'm somewhat slow at this just because I haven't really thought through the pattern design. Maybe a VC would be helpful to discuss this? I think I have a different understanding of the design, and it may be most helpful to talk through. Would you like to meet this week, or wait for zygoloid to be back since he's more familiar with how patterns work?
> 
> A VC sounds like a good idea, but my availability for that is uncertain at the moment (I'm WFH covering for a sick co-parent). I'll ping you once I know what my situation is, and in the meantime I'll respond here to what I can.

Okay. I'm going to pause review here pending a VC.

geoffromer: I just realized the edit you're commenting on doesn't belong in this PR; the `LetIntroducer` case was supposed to be unchanged. I've fixed that now. I don't think that really affects the issues we're discussing in this thread, because they apply just as much to the `TuplePatternStart` case, but please let me know if I've misunderstood that. Otherwise, I'll continue to work on the previously-discussed changes.

geoffromer: OK, as discussed elsewhere I've revised this to drop the forward references.

toolchain/sem_ir/inst_namer.cpp
```diff
@@ -416,6 +416,13 @@ auto InstNamer::CollectNamesInBlock(ScopeId scope_id,
             sem_ir_.entity_names().Get(inst.entity_name_id).name_id);
         continue;
       }
+      case BindingPattern::Kind:
+      case SymbolicBindingPattern::Kind: {
+        auto inst = untyped_inst.As<AnyBindingPattern>();
+        add_inst_name_id(
+            sem_ir_.entity_names().Get(inst.entity_name_id).name_id, ".patt");
```
geoffromer, line 429: Note: the "`.patt`" never appears in the textual IR output. The only effect of omitting that parameter here is that the instruction numbering changes slightly. I can't tell whether any of that is working as intended.

geoffromer (commented):
I'm particularly looking for feedback on the textual IR changes, because I'm not familiar enough with the conventions to tell if what I have is good, or how to improve them if not.

toolchain/check/context.cpp
```diff
@@ -147,6 +148,13 @@ auto Context::AddPlaceholderInst(SemIR::LocIdAndInst loc_id_and_inst)
   return inst_id;
 }
 
+auto Context::AddPatternInst(SemIR::LocIdAndInst loc_id_and_inst)
+    -> SemIR::InstId {
+  auto inst_id = sem_ir().insts().AddInNoBlock(loc_id_and_inst);
```
jonmeow: Had you considered calling either Context::AddInstInNoBlock or Context::AddPlaceholderInstInNoBlock here? Note both of those provide a little extra logic (for VLOG and constant), which is why they're called from AddInst/AddPlaceholderInst. Normally we'd want a constant associated, which would be AddInstInNoBlock.

toolchain/check/handle_alias.cpp
```diff
@@ -16,11 +16,15 @@ auto HandleParseNode(Context& context, Parse::AliasIntroducerId /*node_id*/)
...
   return true;
 }
 
-auto HandleParseNode(Context& /*context*/,
-                     Parse::AliasInitializerId /*node_id*/) -> bool {
+auto HandleParseNode(Context& context, Parse::AliasInitializerId /*node_id*/)
+    -> bool {
+  // TODO: when/if parameterized aliases are supported, this must be
+  // attached to the `BindAlias` inst.
+  (void)context.pattern_block_stack().Pop();
```
jonmeow:
```suggestion
  context.pattern_block_stack().PopAndDiscard();
```

Generally I don't think we write `(void)` except where required by the compiler for an unused variable/parameter. If you're trying to indicate that there's a value being discarded, maybe the `PopAndDiscard` call would capture that?

(note, this same comment would apply elsewhere that you're doing this)

toolchain/sem_ir/typed_insts.h
```diff
@@ -302,6 +302,36 @@ struct BindValue {
   InstId value_id;
 };
 
+// Common representation for various `*binding_pattern` nodes.
+struct AnyBindingPattern {
+  // TODO: Also handle BindTemplateName once it exists.
```
jonmeow: Should this TODO be updated?

toolchain/sem_ir/inst_namer.cpp
jonmeow: Looking at an example, I think this is related to:

`%.loc11_8: i32 = binding_pattern a`

I would expect that should be named. My guess would be that the block isn't getting touched. This is probably worth some debugging about why the assignment isn't taking, e.g. by figuring out one of the InstIds for a BindingPattern and then looking at naming calls related to it. (I'd probably just insert logs in the get and set code for starters, maybe logging for every BindingPattern instruction in a test where there's only one)

Do you want help figuring this out, or is this something you'd like to figure out what's going on?

toolchain/sem_ir/inst_namer.cpp
```diff
@@ -438,7 +445,9 @@ auto InstNamer::CollectNamesInBlock(ScopeId scope_id,
       }
       case CARBON_KIND(ClassDecl inst): {
         add_inst_name_id(sem_ir_.classes().Get(inst.class_id).name_id, ".decl");
-        CollectNamesInBlock(scope_id, inst.decl_block_id);
+        DeclInfo decl = sem_ir_.decls().Get(inst.decl_id);
+        CollectNamesInBlock(scope_id, decl.pattern_block_id);
+        CollectNamesInBlock(scope_id, decl.decl_block_id);
```
jonmeow: Looking at the copies, had you considered a helper function for this?

toolchain/sem_ir/decl.h
```diff
@@ -0,0 +1,23 @@
...
+namespace Carbon::SemIR {
+
+// Information about a declaration.
+struct DeclInfo {
+  // The pattern block, containing the pattern insts for the parameter pattern.
+  InstBlockId pattern_block_id;
+  // The declaration block, containing the declaration's parameters and their
+  // types.
+  InstBlockId decl_block_id;
+};
```
jonmeow: What had you considered as alternatives before the DeclInfo approach? A couple I'm wondering about are:

- Adding an instruction to `decl_block_id` which just declares that there's a pattern
  - e.g. `struct PatternBlock { InstBlockId block_id; }`
  - i.e., as a generic approach that could apply to any block that adds a pattern.
- Putting `pattern_block_id` on `EntityWithParams`
    - i.e., tracking it alongside parameters that it's providing a pattern for

To explain why I'm thinking about this, adding more ValueStores feels a little incremental complexity for understanding the toolchain. `decl_block_id` was added to instructions because it fit into the instruction and was mainly for formatting. However, as part of splitting out to a new structure, it's no longer taking advantage of free space in the declaration's instruction.

Maybe it'd help for me to understand, what's the access pattern you expect for `pattern_block_id`?

If only kept for the printed IR, the `PatternBlock` instruction approach might be good: it keeps an association that you could get formatted instructions for, and also provides an approach that would work for arbitrary constructs. The overhead is marginally higher (essentially 12 more bytes), but more flexible beyond the current set of declarations. I expect this approach would need the `PatternBlock` to be skipped for lowering, but it'd just be the one instruction.

If you need to go back to the pattern block from the decl instruction later during checking, then `EntityWithParams` might be good: we would probably need it to be associated with the params themselves. `EntityWithParams` is already shared where you're using `DeclInfo`, meaning overhead is marginally lower (essentially 4 fewer bytes).

jonmeow (commented):
FYI, I'm still looking at this today, but wanted to publish comments in case you want to discuss something during the toolchain meeting. Overall I think the direction this PR is taking seems a good start, though there are still some details that may be good to discuss (particularly see DeclInfo).

toolchain/check/handle_class.cpp
```diff
@@ -41,6 +41,8 @@ auto HandleParseNode(Context& context, Parse::ClassIntroducerId node_id)
   context.decl_name_stack().PushScopeAndStartName();
   // This class is potentially generic.
   StartGenericDecl(context);
+  // Start a new pattern block for the signature.
+  context.pattern_block_stack().Push();
```
jonmeow, line 47: Just a note, but something else to consider with storage might be whether we want the parameters and implicit parameters to each have their own pattern block. A single block is fine for now, I'm not asking for a change in this PR, but it might be something that affects the implementation more later.

toolchain/check/pattern_block_stack.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include "common/array_stack.h"
+#include "toolchain/sem_ir/ids.h"
+
+namespace Carbon::Check {
+
+class Context;
+
+// Stack of blocks containing pattern insts.
+class PatternBlockStack {
```
jonmeow: Had you reconsidered reusing InstBlockStack? There's a lot of logic there that might apply here too: avoiding allocation for empty blocks, debug logging, printing support for stack traces, and so on.

jonmeow (commented):
Actually, I guess I was almost done looking through, just a couple more comments.

toolchain/check/context.cpp
geoffromer: I considered and rejected it, but my reasons are no longer valid (if they ever were), and in fact I'm now realizing this is probably the cause of the bug I've been failing to track down all morning. So, thanks for pointing this out! Fixed.

toolchain/sem_ir/inst_namer.cpp
geoffromer: Ah, I found the problem: I used `break` instead of `continue` on the next line.

toolchain/check/pattern_block_stack.h
geoffromer: As with `Context::AddInstInNoBlock` I considered and rejected using it, for reasons that no longer make sense. Fixed.

toolchain/sem_ir/inst_namer.cpp
geoffromer: I had, but didn't get around to it. It's moot now.

toolchain/sem_ir/decl.h
geoffromer: I hadn't considered those options. I've added the pattern block to `EntityWithParams`, although note that `Impl` doesn't derive from `EntityWithParams` (yet?), so I had to also add it there. This feels a little awkward to me, because the pattern block is conceptually more closely tied to the decl block than to the rest of the entity, but that may be preferable to adding another `ValueStore`.

I don't really expect the toolchain to access `pattern_block_id` except when rendering the IR, but part of the rationale for introducing the pattern block and attaching it to some "owner" is because other tools (e.g. static analysis) might want to access it, and that will be much easier if it's a named member of some object rather than an inst they have to scan for. I also don't think the `PatternBlock` approach will be as flexible as you say: in other uses of patterns (e.g. `let` declarations), there's no block we can add it to while preserving the structural connection between the pattern insts and the pattern-match insts, so it will have to be a named member in those cases.

Side note: this changes the representation of chains of function-style qualifiers like `Class(T:! type).F(n: T)` (see `toolchain/check/testdata/class/generic/member_out_of_line.carbon` for an example), but I think both forms are probably wrong, and I think the pre-existing representation of the decl is probably wrong too, so I'd rather punt on that whole issue for now.

toolchain/check/name_component.h
```diff
@@ -33,6 +33,9 @@ struct NameComponent {
   // The explicit parameter list.
   Parse::NodeId params_loc_id;
   SemIR::InstBlockId params_id;
+
+  // The pattern block.
+  SemIR::InstBlockId pattern_block_id;
 };
 
 // Pop a name component from the node stack.
```
jonmeow: Would it make sense to document the pattern block stack pop here now?

toolchain/sem_ir/decl.h
jonmeow: > Side note: this changes the representation of chains of function-style qualifiers like `Class(T:! type).F(n: T)` (see `toolchain/check/testdata/class/generic/member_out_of_line.carbon` for an example), but I think both forms are probably wrong, and I think the pre-existing representation of the decl is probably wrong too, so I'd rather punt on that whole issue for now.

I'm not seeing the issue, can you elaborate?

That said, I have an extended comment here, but let me ask two questions:

1. Why do declarations always have a pattern block, even declarations that never have a pattern (such as alias and namespace)?
2. Considering programmatic access for tools, how would a tool disambiguate between explicit and implicit parameter patterns?

I'm asking this way because I think they're intertwined in how they'd be addressed; let me give some detail on both:

1. Why do declarations always have a pattern block, even declarations that never have a pattern (such as alias and namespace)?

In addition to `alias` and similar, consider `impl`. I see patterns in the example `testdata/impl/impl_forall.carbon`, but not in the general case. Here's an example:

```
impl forall [T:! type] T as Simple {
  fn F() {}
}

...

// CHECK:STDOUT:   impl_decl @impl {
// CHECK:STDOUT:     %T.patt: type = symbolic_binding_pattern T 0
// CHECK:STDOUT:   } {
```

Had you considered only creating a pattern block when there's a `forall`, avoiding the creation when there is none? In particular, `forall` uses an implicit parameter list, which connects me to the second question.

Sharing `PopNameComponent` is fine, but perhaps `PopNameComponent` should have an option not to pop a pattern block?

2. Considering programmatic access for tools, how would a tool disambiguate between explicit and implicit parameter patterns?

Right now, we do have decl_block_id merging a few different things. However, as you note, it's not really intended for inspection. Should parameter patterns be separated similarly to how parameter references are?

Note, I think this intertwines with the former question. If pattern blocks were split, one option might be to use TuplePattern and ImplicitParamList start/end nodes to push/pop pattern blocks. It looks like the start nodes don't have anything attached at present, so with a little work they might be useable as channels to store block IDs too.

To the extent that we might want different behavior between tuple patterns and explicit parameter lists, we currently don't support tuple patterns in check except as parameter lists. But, maybe these should be using distinct parse nodes instead of the same, or it may also get to a related question of how nested patterns should be represented/stored (maybe a nested tuple pattern actually needs this same behavior?).

jonmeow: > note that `Impl` doesn't derive from `EntityWithParams` (yet?),

Note, https://github.com/carbon-language/carbon-lang/pull/4336 is changing this.

geoffromer: > > Side note: this changes the representation of chains of function-style qualifiers like `Class(T:! type).F(n: T)` (see `toolchain/check/testdata/class/generic/member_out_of_line.carbon` for an example), but I think both forms are probably wrong, and I think the pre-existing representation of the decl is probably wrong too, so I'd rather punt on that whole issue for now.
> 
> I'm not seeing the issue, can you elaborate?

See `%F.decl` in that file: after my revisions last week (specifically [this one](https://github.com/carbon-language/carbon-lang/commit/ed374dcd7f3aad4d6c2b7acf092ae931d9c5742a#diff-5d84bb40fe60cd6c7ee3ddd37fcb93c1d9027118b6fa20f1896f539301b1404bR149)), there's no longer a `symbolic_binding_pattern T` in the pattern block. By contrast, there _is_ a `bind_symbolic_name T` in the decl block.

I'm going to take your other questions in reverse order:

> 2. Considering programmatic access for tools, how would a tool disambiguate between explicit and implicit parameter patterns?

My plan is to extend `NameComponent` and `EntityWithParams` with two new `InstBlockId` members that relate to the pattern block in much the same way that `implicit_params_id` and `params_id` currently relate to the decl block:
- They don't "own" the insts they contain, but refer to insts in the pattern block.
- They have one inst per parameter.
- Each inst is the "root" inst for the corresponding parameter: all the pattern insts relating to that parameter are reachable from it.
This isn't just for future SemIR consumers; `check` itself is going to need it once we shift to generating the pattern-match SemIR from the pattern SemIR.

I already have that implemented, but it seems out of scope for this PR.

> 1. Why do declarations always have a pattern block, even declarations that never have a pattern (such as alias and namespace)?

The basic problem is that the AST doesn't give us a reliable context-free signal that a full-pattern is starting (or ending), so we have to push a pattern block whenever one _might_ be about to start, and pop once the full-pattern (if any) has definitely ended. That applies even to `alias` and `namespace` because the AST allows those to be parameterized.

You're right that we could avoid that problem if we put the implicit and explicit parameter lists in separate pattern blocks, and used separate AST node kinds for parameter lists and other tuple patterns. However, the consensus [on Discord](https://discord.com/channels/655572317891461132/655578254970716160/1280995814805737513) seemed to be to have one pattern block for each full-pattern, and to me that seems like the more semantically faithful representation.

Furthermore, I don't know if that change would really gain us much. My strong suspicion is that if a SemIR consumer cares about the distinction between implicit and explicit parameters, they probably care about which parameter a given inst is associated with. That means they aren't going to want to work with the pattern block directly, because it's too unstructured -- they're going to want to work with the one-inst-per-parameter blocks that I mentioned earlier, and traverse from there. That's certainly been my experience so far with consuming the pattern IR in toolchain code.

geoffromer:
I've added the TODOs we agreed on in the toolchain discussion yesterday. I wasn't quite sure where to put the TODO to embed more information in the parse tree, so I went with kind of a saturation approach.

I believe that's all the changes we agreed on, so this is ready for review.

jonmeow (approved):

geoffromer:
Note: the contents of #4333 were merged into this PR after review, owing to a git usage error on my part. My apologies for the confusion.

PR #3797, opened 2024-03-19
Raw identifier syntax

jonmeow:
We want to support legacy identifiers that overlap with new keywords (for example, `base`). This is being called "raw identifier syntax" using `r#<identifier>`, and is based on [Rust](https://doc.rust-lang.org/reference/identifiers.html).

Note this proposal is derived from [Proposal #17: Lexical conventions](https://github.com/carbon-language/carbon-lang/pull/17).

Co-authored-by: zygoloid <richard@metafoo.co.uk>

PR #4223, opened 2024-08-16
Add docs for raw identifier syntax.

jonmeow:
Feature approved in #3797

### DIFF
--- a/docs/design/lexical_conventions/words.md
+++ b/docs/design/lexical_conventions/words.md
@@ -101,7 +101,7 @@ The following words are interpreted as keywords:
 
 A raw identifier looks like `r#<identifier>`. This can be used for identifiers
 which have the same spelling as keywords; for example, `r#impl`. It can help
-when using C++ code with identifiers that are a keyword in Carbon.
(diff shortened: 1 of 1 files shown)


docs/design/lexical_conventions/words.md
```diff
@@ -96,12 +97,28 @@ The following words are interpreted as keywords:
 -   `where`
 -   `while`
 
+### Raw identifiers
+
+A raw identifier looks like `r#<identifier>`. This can be used for identifiers
+which have the same spelling as keywords; for example, `r#impl`. It can help
+when using C++ code with identifiers that are a keyword in Carbon.
```
chandlerc: Maybe:
```suggestion
which have the same spelling as keywords; for example, `r#impl`. It can help
when using C++ code with an identifier that is a keyword in Carbon.
```

Or:
```suggestion
which have the same spelling as keywords; for example, `r#impl`. It can help
when using C++ code with identifiers that are keywords in Carbon.
```

chandlerc (approved):
LGTM

PR #4224, opened 2024-08-16
Document the export keyword

jonmeow:
This was approved in #3938


docs/design/code_and_name_organization/README.md
```diff
@@ -615,6 +616,41 @@ import library "Shapes";
...
 ```
 
+#### Exporting imported names
+
+The `export` keyword supports exporting names that come from imported libraries.
+This can be used to create an API file that provides contents from multiple
+other API files.
+
+`export` can be used on either a full import, as a modifier to the `import`
+keyword, or a specific entity, with `export <name>` declarations.
```
chandlerc: I misread this as three alternatives at first... Maybe:

```suggestion
`export` can be used either as a modifier to the `import`
keyword to export the entire imported library, or in an `export <name>` declaration to export a specific entity.
```

Or maybe another phrasing that works better for you/

chandlerc (approved):
LGTM

PR #4225, opened 2024-08-16
Add rudimentry debug info metadata emission

dwblaikie:
This adds just the debug info metadata for Compilation Units (the top level container of debug info) - but without anything in them, LLVM won't emit them at all, so while this is testable at the IR level, it isn't observable at the object level until more debug info is added.

A couple of starting points in this patch:
* A flag (`--debug-info`, seems to match the naming/style of other flags in the carbon driver, though this is different from the naming conventions of clang/gcc) that enables debug info when lowering. Open to other names/approaches (on by default?  historically debug info's been to large/expensive to do this, so sticking with that precedent for now).
* Enabling that flag by default in the lowering tests - I do find the churn on golden tests a bit rough, and adding more features to all the tests means more churn, but it seems consistent with the approach so far - keep an eye on this and perhaps revisit this if the churn gets too annoying


toolchain/lower/file_context.h
```diff
@@ -101,6 +107,9 @@ class FileContext {
   // Maps constants to their lowered values.
   // We resize this directly to the (often large) correct size.
   llvm::SmallVector<llvm::Constant*, 0> constants_;
+
+  // Specify whether to include debug info metadata in the generated LLVM IR.
+  bool include_debug_info_;
```
jonmeow: WDYT of moving this up in the member list, maybe before `sem_ir` so that the member order and constructor parameter order correspond a little more?

toolchain/lower/file_context.cpp
```diff
@@ -66,6 +70,22 @@ auto FileContext::Run() -> std::unique_ptr<llvm::Module> {
   return std::move(llvm_module_);
 }
 
+auto FileContext::BuildCompileUnit(llvm::StringRef module_name) -> void {
+  if (!include_debug_info_)
```
```suggestion
  if (!include_debug_info_) {
```

toolchain/lower/file_context.cpp
```diff
@@ -66,6 +70,22 @@
...
+    return;
+  // FIXME: Include directory path in the cu_file.
```
```suggestion
  }
  // FIXME: Include directory path in the cu_file.
```

toolchain/lower/file_context.h
```diff
@@ -76,6 +80,8 @@ class FileContext {
   // State for building the LLVM IR.
   llvm::LLVMContext* llvm_context_;
   std::unique_ptr<llvm::Module> llvm_module_;
+  llvm::DIBuilder di_builder_;
+  llvm::DICompileUnit *di_compile_unit_;
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  llvm::DICompileUnit* di_compile_unit_;
```

toolchain/lower/file_context.cpp
```diff
@@ -66,6 +70,22 @@ auto FileContext::Run() -> std::unique_ptr<llvm::Module> {
...
+  llvm::DIFile* cu_file = di_builder_.createFile(module_name, "");
```
jonmeow: Pardon my unfamiliarity, but why "cu" instead of "di"?

toolchain/lower/file_context.cpp
```diff
@@ -20,16 +20,20 @@
...
       llvm_module_(std::make_unique<llvm::Module>(module_name, llvm_context)),
+      di_builder_(*llvm_module_),
       sem_ir_(&sem_ir),
       inst_namer_(inst_namer),
-      vlog_stream_(vlog_stream) {
+      vlog_stream_(vlog_stream),
+      include_debug_info_(include_debug_info) {
   CARBON_CHECK(!sem_ir.has_errors())
       << "Generating LLVM IR from invalid SemIR::File is unsupported.";
+  BuildCompileUnit(module_name);
```
jonmeow: Maybe rename this to `BuildDICompileUnit`? I was having trouble seeing where `di_compile_unit_` was initialized -- but maybe it'd be even better to have the method be static and return the `DICompileUnit*` instead of void, allowing initializing as `di_compile_unit_(BuildDICompileUnit(module_name))`?

toolchain/lower/file_context.cpp
```diff
@@ -66,6 +70,22 @@ auto FileContext::Run() -> std::unique_ptr<llvm::Module> {
...
+  if (!include_debug_info_)
+    return;
+  // FIXME: Include directory path in the cu_file.
+  llvm::DIFile* cu_file = di_builder_.createFile(module_name, "");
+  // FIXME: Introduce a new language code for Carbon. C works well for now since
+  // it's something debuggers will already know/have support for at least.
+  // Probably have to bump to C++ at some point for virtual functions,
+  // templates, etc.
+  di_compile_unit_ = di_builder_.createCompileUnit(
+      llvm::dwarf::DW_LANG_C, cu_file, "carbon", false, "", 0, "");
```
jonmeow: Is there value in labelling the last 4 parameters? I think it may make it easier for someone like me to understand the parameters without going to docs. That said, it looks like the last defaults to empty, so maybe can be removed.

```suggestion
      llvm::dwarf::DW_LANG_C, cu_file, "carbon", /*isOptimized=*/false, /*Flags=*/"", /*RV=*/0);
```

jonmeow (commented):
Looks good! Thank you for the contributions. Just a few style comments.

Are you using [pre-commit](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/project/contribution_tools.md#running-pre-commit) (`pre-commit install`)? I think it would've automatically fixed this by running `clang-format` for you.

toolchain/lower/file_context.cpp
dwblaikie: Ah - because it's the file for the compilation unit (cu) - I'll expand out that here. These domain-specific acronyms make more sense in other parts of LLVM/Clang that are more focussed just on debug info, but here where it's mixed with non-debug-info concepts, being more explicit is probably good.

toolchain/lower/file_context.h
dwblaikie: Sure - though ended up removing it once `BuildDICompileUnit` was a standalone helper, and we can rely on the CU being non-null as the signifier of whether debug info is being emitted when later patches add other debug info emission further along the lowering pipeline.

toolchain/lower/file_context.cpp
dwblaikie: Sure, renamed to BuildDICompileUnit.
It'd take a few more parameters to make it static (`include_debug_info_` (though we could move that out into the call site), `di_builder_`, and `llvm_module_`)) - I'll give it a go and you can see whether it feels better that way.

dwblaikie:
> Are you using [pre-commit](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/project/contribution_tools.md#running-pre-commit) (`pre-commit install`)? I think it would've automatically fixed this by running `clang-format` for you.

Ah, nope - missed that, sorry. Thanks for the pointer/reminder. Think I managed to run it on some of my stuff that's already committed (with `pre-commit run --all-files`) to the branch - it caught one formatting error, but otherwise seems clean

toolchain/lower/file_context.cpp
jonmeow: This looks good, I think one nice thing about static is it's harder to accidentally use uninitialized members.

jonmeow (approved):
Thanks, glad to have debug info being added!

chandlerc:
> Mostly this'll be a place to discuss the flag naming/functionality, and testing strategy. Should we have debug info on in all lowering tests? Should there be some separate directory just for debug info lowering test coverage? I'm open to ideas/suggestions/requests.

Not a comment on the patch, but maybe update the description before merging to summarize what your suggested starting point is for these questions, and what motivates that starting point?

(FWIW, not disagreeing with the current behavior, mostly looking to have a record of it.)

dwblaikie:
> > Mostly this'll be a place to discuss the flag naming/functionality, and testing strategy. Should we have debug info on in all lowering tests? Should there be some separate directory just for debug info lowering test coverage? I'm open to ideas/suggestions/requests.
> 
> Not a comment on the patch, but maybe update the description before merging to summarize what your suggested starting point is for these questions, and what motivates that starting point?
> 
> (FWIW, not disagreeing with the current behavior, mostly looking to have a record of it.)

Updated - hopefully that's something along the lines of what you're looking for.

jonmeow (approved):

PR #4226, opened 2024-08-17
move tuple testdata from `index/` to `tuple/access`

brymer-meneses:


brymer-meneses:
As per the discussion here https://github.com/carbon-language/carbon-lang/pull/4186#issuecomment-2274333669. Let me know if I missed anything.

jonmeow (approved):
Thanks for the cleanup! This all looks good to me.

PR #4227, opened 2024-08-19
Modify toolchain/autoupdate_testdata.py to use scripts/scripts_utils.py

hamphet:
The current `toolchain/autoupdate_testdata.py` script assumes the correct bazel version is already installed. This change uses `scripts/scripts_utils.py` to fallback to `bazelisk`.

### DIFF
--- a/toolchain/autoupdate_testdata.py
+++ b/toolchain/autoupdate_testdata.py
@@ -15,12 +15,13 @@ from pathlib import Path
 
 
 def main() -> None:
+    bazel = str(Path(__file__).parents[1] / "scripts" / "run_bazel.py")
     # Use the most recently used build mode, or `fastbuild` if missing
     # `bazel-bin`.
     build_mode = "fastbuild"
     workspace = subprocess.check_output(
         [
-            "bazel",
+            bazel,
             "info",
             "workspace",
             "--ui_event_filters=stdout",
@@ -37,7 +38,7 @@ def main() -> None:
             exit(f"Build mode not found in `bazel-bin` symlink: {link}")
 
     argv = [
-        "bazel",
+        bazel,
         "run",
         "-c",
         build_mode,


toolchain/autoupdate_testdata.py
```diff
@@ -11,16 +11,23 @@
 import re
 import subprocess
 import sys
+import os
 from pathlib import Path
 
+directory = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
+sys.path.append(os.path.join(directory))
```
jonmeow: WDYT of using pathlib for this:

```suggestion
sys.path.append(str(Path(__file__).parents[1]))
```

What I'm finding is that `__file__` is always absolute [as of python 3.9](https://docs.python.org/3/whatsnew/3.9.html#other-language-changes), so doesn't need `realpath`

toolchain/autoupdate_testdata.py
```diff
@@ -11,16 +11,23 @@
 import re
 import subprocess
 import sys
+import os
```
jonmeow:
```suggestion
```

(see below)

jonmeow (approved):
Thanks! Sorry about the inconvenience with this.

jonmeow:
Is the `__init__.py` necessary? I think it may be causing the issue for `pre-commit` execution.

toolchain/autoupdate_testdata.py
hamphet: Ack.

hamphet:
Thanks for the suggestions, looks much better!

toolchain/autoupdate_testdata.py
```diff
@@ -13,14 +13,19 @@
 import sys
 from pathlib import Path
 
+sys.path.append(str(Path(__file__).parents[1]))
+import scripts.scripts_utils  # noqa: E402
+
```
jonmeow:
```suggestion
```

toolchain/autoupdate_testdata.py
```diff
@@ -13,14 +13,19 @@
...
 
 def main() -> None:
+    scripts.scripts_utils.chdir_repo_root()
+    bazel = scripts.scripts_utils.locate_bazel()
```
jonmeow:
```suggestion
    bazel = str(Path(__file__).parents[1] / "scripts" / "run_bazel.py")
```

jonmeow (commented):
mypy isn't doing great here, and `carbon-lang` is a nuisance for a directory name; can you use `run_bazel.py` for this?

toolchain/autoupdate_testdata.py
hamphet: Done.

PR #4228, opened 2024-08-19
Lower global variables

hamphet:


toolchain/lower/file_context.h
```diff
@@ -73,6 +76,8 @@ class FileContext {
   // the caller.
   auto BuildType(SemIR::InstId inst_id) -> llvm::Type*;
 
+  auto BuildGlobalVariableDecl(SemIR::InstId inst_id) -> llvm::GlobalVariable*;
```
jonmeow: We're generally trying to have brief comments on functions, how about something similar to the above:

```suggestion
  // Builds the global for the given instruction, which should then be cached by the caller.
  auto BuildGlobalVariableDecl(SemIR::InstId inst_id) -> llvm::GlobalVariable*;
```

toolchain/lower/file_context.cpp
```diff
@@ -50,7 +51,17 @@ auto FileContext::Run() -> std::unique_ptr<llvm::Module> {
...
+  // Lower global variable declarations.
+  for (auto inst_id :
+       sem_ir().inst_blocks().Get(sem_ir().top_inst_block_id())) {
+    // Only `VarStorage` indicates global variable declaration in the
+    // top instruction block.
+    if (auto inst = sem_ir().insts().Get(inst_id);
+        inst.kind() != SemIR::InstKind::VarStorage) {
+      continue;
+    }
+    global_variables_.Insert(inst_id, BuildGlobalVariableDecl(inst_id));
```
jonmeow: What do you think of changing `BuildGlobalVariableDecl` to take a `SemIR::VarStorage` instead of an `inst_id`?

We have helpers like `insts().Is` and `insts().TryGetAs`, so you could write this like:

```suggestion
    if (auto var = sem_ir().insts().TryGetAs<SemIR::VarStorage>(inst_id)) {
      global_variables_.Insert(inst_id, BuildGlobalVariableDecl(*var));
    }
```

Note a subtle effect of that is it means the array lookup for `inst_id` only occurs once, since `BuildGlobalVariableDecl` directly reuses the `Get` result from here.

toolchain/lower/file_context.cpp
```diff
@@ -463,4 +479,17 @@ auto FileContext::BuildType(SemIR::InstId inst_id) -> llvm::Type* {
   }
 }
 
+auto FileContext::BuildGlobalVariableDecl(SemIR::InstId inst_id)
+    -> llvm::GlobalVariable* {
+  auto var_storage = sem_ir().insts().Get(inst_id).As<SemIR::VarStorage>();
+  auto mangled_name =
```
jonmeow, line 483:
```suggestion
  // TODO: Mangle name.
  auto mangled_name =
```

I'm not exactly sure what the samantics for `GetAsStringIfIdentifier` will be once name mangling is actually added, but seems worth leaving a breadcrumb here.

toolchain/lower/file_context.cpp
```diff
@@ -50,7 +51,17 @@ auto FileContext::Run() -> std::unique_ptr<llvm::Module> {
     functions_[i] = BuildFunctionDecl(SemIR::FunctionId(i));
   }
 
-  // TODO: Lower global variable declarations.
+  // Lower global variable declarations.
+  for (auto inst_id :
+       sem_ir().inst_blocks().Get(sem_ir().top_inst_block_id())) {
+    // Only `VarStorage` indicates global variable declaration in the
```
```suggestion
    // Only `VarStorage` indicates a global variable declaration in the
```

toolchain/lower/file_context.cpp
```diff
@@ -463,4 +479,17 @@ auto FileContext::BuildType(SemIR::InstId inst_id) -> llvm::Type* {
   }
 }
 
+auto FileContext::BuildGlobalVariableDecl(SemIR::InstId inst_id)
+    -> llvm::GlobalVariable* {
+  auto var_storage = sem_ir().insts().Get(inst_id).As<SemIR::VarStorage>();
```
jonmeow: Note I'm suggesting passing in the `VarStorage` above, but I'll note for future reference this could also be written as `GetAs<SemIR::VarStorage>(inst_id);`

jonmeow (commented):
Thank you! The tests look good to me. I have a suggestion on `insts` calls, but that's the main thing.

toolchain/lower/file_context.h
hamphet: Done.

toolchain/lower/file_context.cpp
hamphet: Thanks! Done.

toolchain/lower/file_context.cpp
hamphet: Done.

toolchain/lower/file_context.cpp
hamphet: Ack. Thanks!

jonmeow (approved):
Thanks, changes all look great!

toolchain/lower/file_context.cpp
hamphet: Done.

PR #4230, opened 2024-08-19
Add documentation for entity declaration design work

jonmeow:
Trying to pull in key elements of #3762, #3763, and #3980 (decl matching and `extern`, essentially). These aren't specific to any particular declaration type, but are common to entities, so suggesting a new doc oriented on that.

There's probably more that could be said here, I'm just focused on getting the recent formal discussion mirrored into the design.

### DIFF
--- a/docs/design/declaring_entities.md
+++ b/docs/design/declaring_entities.md
@@ -66,37 +66,35 @@ In order to determine whether two redeclarations refer to the same entity, we
 apply the rules:
 
 -   Two named declarations _declare the same entity_ if they have the same scope
-    and the same name.
--   When two named declarations declare the same entity, the second is said to be a
-    _redeclaration_, including the case where the first declaration is imported.
--   Two declarations _differ_ if the sequence of tokens in the declaration
-    following the introducer keyword and the optional scope, up to the semicolon
-    or open brace, is different, except for `unused` modifiers on parameters.
--   The program is invalid if it contains two owning declarations of the same
-    entity that differ.
-    -   The non-owned `extern library` declarations will only use semantic
-        matching for redeclarations, not syntactic matching.
+    and the same name. This includes imported declarations.
+-   When two named declarations declare the same entity, the second is said to
+    be a _redeclaration_.
+-   Two owned declarations _differ_ if they don't syntactically match.
+    -   The non-owned `extern library` declarations only require semantic
+        matching.
+-   The program is invalid if it contains two declarations of the same entity
+    that differ.
 
 ```carbon
 class A {
+  // This function will be redeclared in order to provide a definition.
   fn F(n: i32);
-  fn G(n: i32);
 }
+
+// ✅ Valid: The declaration matches syntactically.
+fn A.F(n: i32) {}
 
 // ❌ Invalid: The parameter name differs.
 fn A.F(m: i32) {}
 
 // ❌ Invalid: The parameter type differs syntactically.
-fn A.G(n: (i32)) {}
-
-// ✅ Valid: only difference is the scope is explicit.
-fn A.F(n: i32) {} 
+fn A.F(n: (i32)) {}
 ```
 
 ### Details
 
 TODO: Figure out what details to pull from
-[#3762](https://github.com/carbon-language/carbon-lang/pull/3763) and
+[#3762](https://github.com/carbon-language/carbon-lang/pull/3762) and
 [#3763](https://github.com/carbon-language/carbon-lang/pull/3763).
 
 ## `extern` and `extern library`
@@ -106,10 +104,11 @@ There are two forms of the `extern` modifier:
 -   On an owning declaration, `extern` limits access to the definition.
     -   The entity must be directly imported in order to use of the definition.
     -   An `extern library` declaration is optional.
--   On a non-owning declaration, `extern library` provides a tool for breaking
-    dependency cycles.
+-   On a non-owning declaration, `extern library` allows references to an entity
+    without depending on the owning library.
     -   The library name indicates where the entity is defined.
-    -
+    -   This may be used to break dependency cycles between libraries, which
+        could be used to improve build parallelization.
 
 For example, a use of both might look like:
 
@@ -154,9 +153,10 @@ extern fn MyClassFactory(val: i32) -> MyClass* {
 
 ### Effect on indirect imports
 
-Indirect imports won't see the definition of an `extern` entity. We expect this to
-primarily affect return types of functions. If an incomplete type is encountered
-this way, it can be resolved by directly importing the definition. For example:
+Indirect imports won't see the definition of an `extern` entity. We expect this
+to primarily affect return types of functions. If an incomplete type is
+encountered this way, it can be resolved by directly importing the definition.
+For example:
 
 ```
 library "type";
--- a/docs/design/declaring_entities.md
+++ b/docs/design/declaring_entities.md
@@ -14,6 +14,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 -   [Matching redeclarations of an entity](#matching-redeclarations-of-an-entity)
     -   [Details](#details)
 -   [`extern` and `extern library`](#extern-and-extern-library)
+    -   [Valid scopes for `extern`](#valid-scopes-for-extern)
     -   [Effect on indirect imports](#effect-on-indirect-imports)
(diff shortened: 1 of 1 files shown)


docs/design/declaring_entities.md
```diff
@@ -0,0 +1,233 @@
...
+```
+library "MyLibrary";
+
+fn DoSomething();
+```
+
+```
+impl library "MyLibrary";
+
+fn DoSomething {
```
josh11b:
```suggestion
fn DoSomething() {
```

docs/design/declaring_entities.md
```diff
@@ -0,0 +1,233 @@
...
+
+In order to determine whether two redeclarations refer to the same entity, we
+apply the rules:
+
+-   Two named declarations _declare the same entity_ if they have the same scope
+    and the same name.
+-   One declaration _redeclares_ another if they declare the same entity and the
+    second declaration appears after the first, including the case where the
+    first declaration is imported. In this case, the second declaration is said
+    to be a _redeclaration_ of the first.
```
josh11b: This seems longer than necessary. How about

```suggestion
-   When two named declarations declare the same entity, the second is said to be a
    _redeclaration_, including the case where the first declaration is imported.
```

Is there something else that this doesn't capture that you were trying to convey?

docs/design/declaring_entities.md
```diff
@@ -0,0 +1,233 @@
...
+class A {
+  fn F(n: i32);
+  fn G(n: i32);
+}
+
+// ❌ Invalid: The parameter name differs.
+fn A.F(m: i32) {}
+
+// ❌ Invalid: The parameter type differs syntactically.
+fn A.G(n: (i32)) {}
```
josh11b: Would be nice to show more examples.

```suggestion
fn A.G(n: (i32)) {}

// ✅ Valid: only difference is the scope is explicit.
fn A.F(n: i32) {} 
```

Possibly also an example showing an extern library declaration with semantic but not syntactic matching?

docs/design/declaring_entities.md
```diff
@@ -0,0 +1,233 @@
...
+
+## `extern` and `extern library`
+
+There are two forms of the `extern` modifier:
+
+-   On an owning declaration, `extern` limits access to the definition.
+    -   The entity must be directly imported in order to use of the definition.
+    -   An `extern library` declaration is optional.
+-   On a non-owning declaration, `extern library` provides a tool for breaking
+    dependency cycles.
```
josh11b: I believe this is also used to improve build scalability?

docs/design/declaring_entities.md
```diff
@@ -0,0 +1,233 @@
...
+    -   The library name indicates where the entity is defined.
+    -
```
josh11b: Unfinished?

docs/design/declaring_entities.md
```diff
@@ -0,0 +1,233 @@
...
+  var c: MyClass* = new MyClass();
+  c->val = val;
+  return c;
+}
+```
+
+### Effect on indirect imports
+
+Indirect imports won't see the definition of an entity. We expect this to
+primarily effect return types of functions. If an incomplete type is encountered
```
josh11b:
```suggestion
primarily affect return types of functions. If an incomplete type is encountered
```

docs/design/declaring_entities.md
```diff
@@ -0,0 +1,233 @@
...
+extern fn MyClassFactory(val: i32) -> MyClass* {
+  var c: MyClass* = new MyClass();
+  c->val = val;
+  return c;
+}
+```
+
+### Effect on indirect imports
+
+Indirect imports won't see the definition of an entity. We expect this to
```
josh11b:
```suggestion
Indirect imports won't see the definition of an `extern` entity. We expect this to
```

docs/design/declaring_entities.md
jonmeow: To be sure, this was directly copied from #3763, but I think what you're saying is getting at what zygoloid was trying to say. I'm moving the comment about imports to the first bullet though, since it's not specific to the first declaration.

docs/design/declaring_entities.md
jonmeow: Added your example, but rephrasing why this is valid. Also just dropping the switch to `G`, it feels odd when there's a second `F` example and I'm hesitant to add an `H`.

This is currently a short summary, and `extern library` is a peripheral feature, so I think it should probably be kept for details.

I see why you think this isn't syntactic matching -- but I think this is just a problem with how syntactic matching was summarized in #3762, it wasn't getting into detail for qualified declarations. That is, for each scope and the entity itself (each of `A` and `F(n: i32)` here), the entity name and parameters should syntactically match; as you point out, it's inaccurate to say "following the introducer keyword".

Rather than trying to address that here, I'm eliding that for details of syntactic versus semantic matching (the original proposal has a section for this, but it didn't feel like it should be copy-pasted).

docs/design/declaring_entities.md
jonmeow: Removed

docs/design/declaring_entities.md
jonmeow: Sure, added a note. More importantly, rephrased this to better capture what it actually _does_.

docs/design/declaring_entities.md
```diff
@@ -66,36 +67,35 @@ In order to determine whether two redeclarations refer to the same entity, we
...
-    -   The non-owned `extern library` declarations will only use semantic
-        matching for redeclarations, not syntactic matching.
+    and the same name. This includes imported declarations.
+-   When two named declarations declare the same entity, the second is said to
+    be a _redeclaration_.
+-   Two owned declarations _differ_ if they don't syntactically match.
+    -   The non-owned `extern library` declarations only require semantic
+        matching.
+-   The program is invalid if it contains two declarations of the same entity
+    that differ.
```
josh11b: I think we need the definition of "differ" to include non-owned `extern library` declarations for the last rule here, which isn't specific to owned declarations. Perhaps something like:
```suggestion
-   Two owned declarations _differ_ if they don't syntactically match.
    -   Otherwise, if one is a non-owned `extern library` declaration, declarations
        differ if they don't match semantically.
-   The program is invalid if it contains two declarations of the same entity
    that differ.
```

docs/design/declaring_entities.md
```diff
@@ -105,10 +105,11 @@ There are two forms of the `extern` modifier:
...
     -   The entity must be directly imported in order to use of the definition.
     -   An `extern library` declaration is optional.
--   On a non-owning declaration, `extern library` provides a tool for breaking
-    dependency cycles.
+-   On a non-owning declaration, `extern library` allows references to an entity
+    without depending on the owning library.
     -   The library name indicates where the entity is defined.
-    -
+    -   This may be used to break dependency cycles between libraries, which
+        could be used to improve build parallelization.
```
josh11b: It seems like this might be alternatively be used to improve build scalability by reducing dependencies. This is independent of breaking dependency cycles, and isn't specifically about build parallelization.

```suggestion
    -   This may be used to break dependency cycles between libraries, which
        could be used to improve build parallelization.
    -   This may be used to reduce transitive dependencies, improving build
        scalability.
 ```

Comment (jonmeow) [reply docs/design/declaring_entities.md]: How about this rephrasing... (to copy here)

```
    -   This can be used to improve build performance, such as by splitting out
        a declaration in order to reduce a library's dependencies.
```

Note, trying to step aside from specific ways it might improve build performance because I want these bullets to focus more on what the feature does, versus how it might be used.

josh11b: I'm fine with the new text, but don't you want something about that it can allow you to break cycles too?

josh11b (approved):

docs/design/declaring_entities.md
jonmeow: To me, that sort of feels like details where if we really want to explain how to use the feature this way, we should have more examples of refactorings to do.

jonmeow: i.e., I'm trying to figure out the right line to draw between proposal advocating for adoption of the feature, design trying to explain what it is, and user guide explaining how to apply it in anger.

josh11b: I feel like the design shouldn't just say what it is, but also the why, and the purpose of the feature is definitely in scope.

PR #4232, opened 2024-08-21
Enable debug info by default

dwblaikie:
Discussed in the toolchain meeting today - we'd like to try having this on by default and see if the cost isn't too high.

The nodebug test is a bit verbose, because it doesn't have the `--exclude-dump-file-prefix` that test_file would usually add. Is there a nicer way I could write this test to verify that --no-debug-info does what it's meant to?


jonmeow:
Note, I removed the description boilerplate from the top comment.

Note nodebug isn't just verbose, it includes your local machine paths; it's host-dependent. Maybe add something like `%{core_package_dir}` to `file_test` support so that you can add `--exclude-dump-file-prefix=%{core_package_dir}` to `ARGS`? (it seems fine if you want to leave that as a TODO, but the test file would then need to be removed from the PR)

dwblaikie:
> Note, I removed the description boilerplate from the top comment.

Ah, thanks for that! I'll figure out the description template, etc eventually/keep an eye out for that in the future.

> Note nodebug isn't just verbose, it includes your local machine paths; it's host-dependent. 

Oh, right - good call/catch. Thanks!

> Maybe add something like `%{core_package_dir}` to `file_test` support so that you can add `--exclude-dump-file-prefix=%{core_package_dir}` to `ARGS`? (it seems fine if you want to leave that as a TODO, but the test file would then need to be removed from the PR)

Looking into that. I'm a bit of a testing pedant, so would prefer not to commit the change untested, and happy to frontload the test infrastructure improvements necessary to ensure coverage.

Ended up that I needed to add generic functionality to file_test_base (since it errors on unknown characters after %) - so allowed derived classes to expose a collection of replacements, and used those...

testing/file_test/file_test_base.h
```diff
@@ -86,6 +87,8 @@ class FileTestBase : public testing::Test {
   // Returns default arguments. Only called when a file doesn't set ARGS.
   virtual auto GetDefaultArgs() -> llvm::SmallVector<std::string> = 0;
 
+  virtual auto GetReplacements() -> llvm::StringMap<std::string> { return {}; }
```
jonmeow: Maybe `GetArgReplacements`, mirroring `DoArgReplacements` and contrasting with `GetLineNumberReplacements`?

Also, please add a comment explaining the function's semantics.

josh11b (approved):

jonmeow (commented):
A couple more things to change:

1. testing/file_test/README.md is trying to document syntax, can you add a note about this under where it says "Supported replacements within arguments"?

2. In testing/file_test/testdata, can you add a test similar to `args.carbon`, with something like `ARGS: abc %{replacement} %s`?

    - In `testing/file_test/file_test_base_test.cpp`, if you add the new test file where `args.carbon` is (the early return), you can validate that the replacement is correctly round tripping.
    - Note this directory also has an autoupdate script (a more trivial one than toolchain's)

jonmeow:
> 2. In testing/file_test/testdata, can you add a test similar to `args.carbon`, with something like `ARGS: abc %{replacement} %s`?

Or really, even just add `%{replacement}` to `args.carbon`. It doesn't need to be a separate file, it can just be added to the existing test.

dwblaikie:
> A couple more things to change:
> 
> 1. testing/file_test/README.md is trying to document syntax, can you add a note about this under where it says "Supported replacements within arguments"?
> 2. In testing/file_test/testdata, can you add a test similar to `args.carbon`, with something like `ARGS: abc %{replacement} %s`?
>    
>    * In `testing/file_test/file_test_base_test.cpp`, if you add the new test file where `args.carbon` is (the early return), you can validate that the replacement is correctly round tripping.
>    * Note this directory also has an autoupdate script (a more trivial one than toolchain's)

Ah, for sure - thanks for the pointers. Done!

testing/file_test/file_test_base.h
dwblaikie: Done both.

jonmeow:
Just a reminder for the README.md edit, in case it was missed. :)

dwblaikie:
> Just a reminder for the README.md edit, in case it was missed. :)

Ah, thought I'd done that - sorry I missed it. Added some now.

jonmeow (approved):
Thanks! All looks good here.

PR #4233, opened 2024-08-21
Add function debug info descriptions

dwblaikie:
Still doesn't have line tables, so of limited value (at least now this'll be enough that LLVM really generates debug info into the resulting object file (whereas with only the compilation unit metadata, LLVM will consider it empty and avoid emitting any of it)) - but another step along the path.

This also doesn't attach the right source location to the functions - I'll do that in a follow-up change because I think it'll require the majority of the refactoring between driver and check to extract the essential functionality sem_ir_diagnostic_converter, I think, to allow retrieving source locations during lowering.

### DIFF
--- a/toolchain/lower/file_context.cpp
+++ b/toolchain/lower/file_context.cpp
@@ -289,7 +289,7 @@ auto FileContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
     return;
   }
 
-  llvm_function->setSubprogram(BuildSubprogram(function, llvm_function));
+  llvm_function->setSubprogram(BuildDISubprogram(function, llvm_function));
 
   FunctionContext function_lowering(*this, llvm_function, vlog_stream_);
 
@@ -362,7 +362,7 @@ auto FileContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
   }
 }
 
-auto FileContext::BuildSubprogram(
+auto FileContext::BuildDISubprogram(
     const Carbon::SemIR::Function& /*semir_function*/,
     const llvm::Function* llvm_function) -> llvm::DISubprogram* {
   if (!di_compile_unit_) {
--- a/toolchain/lower/file_context.cpp
+++ b/toolchain/lower/file_context.cpp
@@ -379,6 +379,7 @@ auto FileContext::BuildDISubprogram(const Carbon::SemIR::Function& /*function*/,
       /*ScopeLine=*/0, llvm::DINode::FlagZero,
       llvm::DISubprogram::SPFlagDefinition);
 }
+
 static auto BuildTypeForInst(FileContext& context, SemIR::ArrayType inst)
     -> llvm::Type* {
   return llvm::ArrayType::get(
--- a/toolchain/lower/file_context.cpp
+++ b/toolchain/lower/file_context.cpp
@@ -362,17 +362,22 @@ auto FileContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
   }
 }
 
-auto FileContext::BuildDISubprogram(
-    const Carbon::SemIR::Function& /*semir_function*/,
-    const llvm::Function* llvm_function) -> llvm::DISubprogram* {
+auto FileContext::BuildDISubprogram(const Carbon::SemIR::Function& /*function*/,
+                                    const llvm::Function* llvm_function)
+    -> llvm::DISubprogram* {
   if (!di_compile_unit_) {
     return nullptr;
   }
+  // FIXME: Add more details here, including mangled name, real subroutine type
+  // (once type information is built), etc.
   return di_builder_.createFunction(
-      di_compile_unit_, llvm_function->getName(), "", nullptr, 0,
+      di_compile_unit_, llvm_function->getName(), /*LinkageName=*/"",
+      /*File=*/nullptr,
+      /*LineNo=*/0,
       di_builder_.createSubroutineType(
           di_builder_.getOrCreateTypeArray(std::nullopt)),
-      0, llvm::DINode::FlagZero, llvm::DISubprogram::SPFlagDefinition);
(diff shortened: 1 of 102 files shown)


toolchain/lower/file_context.h
```diff
@@ -78,6 +78,10 @@ class FileContext {
   // declaration with no definition, does nothing.
   auto BuildFunctionDefinition(SemIR::FunctionId function_id) -> void;
 
+  auto BuildSubprogram(const Carbon::SemIR::Function& function,
```
jonmeow: Please add a short comment explaining call semantics.

toolchain/lower/file_context.cpp
```diff
@@ -360,6 +362,18 @@ auto FileContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
...
+    const Carbon::SemIR::Function& /*semir_function*/,
+    const llvm::Function* llvm_function) -> llvm::DISubprogram* {
+  if (!di_compile_unit_) {
+    return nullptr;
+  }
+  return di_builder_.createFunction(
+      di_compile_unit_, llvm_function->getName(), "", nullptr, 0,
+      di_builder_.createSubroutineType(
+          di_builder_.getOrCreateTypeArray(std::nullopt)),
+      0, llvm::DINode::FlagZero, llvm::DISubprogram::SPFlagDefinition);
```
jonmeow: Can you just quickly add `/*...=*/` parameter comments for the `""`, `nullptr`, and `0` parameters? i.e., so that it's clearer on a skim what this is zeroing-out.

toolchain/lower/file_context.cpp
```diff
@@ -360,6 +362,18 @@ auto FileContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
   }
 }
 
+auto FileContext::BuildSubprogram(
+    const Carbon::SemIR::Function& /*semir_function*/,
+    const llvm::Function* llvm_function) -> llvm::DISubprogram* {
+  if (!di_compile_unit_) {
+    return nullptr;
+  }
+  return di_builder_.createFunction(
```
jonmeow, line 373: I see the mention of adding line numbers in the PR description, but if there are things more long-term, like if the LinkageName should be a mangled name, are those things we'd want a TODO for?

(note, I'm not sure whether we'll really track a "ScopeLine")

toolchain/lower/file_context.cpp
```diff
@@ -360,6 +362,18 @@ auto FileContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
   }
 }
 
+auto FileContext::BuildSubprogram(
+    const Carbon::SemIR::Function& /*semir_function*/,
```
```suggestion
    const Carbon::SemIR::Function& /*function*/,
```

Match the header? (clang-tidy might catch this, if you're watching in vscode)

jonmeow (commented):
This seems like a good step, but just want to be sure I'm understanding the API and plans correctly.

toolchain/lower/file_context.h
dwblaikie: Done

toolchain/lower/file_context.cpp
dwblaikie: Done

toolchain/lower/file_context.cpp
dwblaikie: Added a general TODO for things we should fill in, mentioning some examples.

The ScopeLine I think is the `{` of the function, useful for some profile-related functionality (changing all line numbers in samples to be relative to the scope line to improve their resilience to unrelated changes in a file)

toolchain/lower/file_context.cpp
dwblaikie: Done

toolchain/lower/file_context.cpp
```diff
@@ -362,17 +362,22 @@ auto FileContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
   }
 }
 
-auto FileContext::BuildSubprogram(
-    const Carbon::SemIR::Function& /*semir_function*/,
-    const llvm::Function* llvm_function) -> llvm::DISubprogram* {
+auto FileContext::BuildDISubprogram(const Carbon::SemIR::Function& /*function*/,
```
```suggestion
auto FileContext::BuildDISubprogram(const SemIR::Function& /*function*/,
```

toolchain/lower/file_context.h
```diff
@@ -78,8 +78,9 @@ class FileContext {
   // declaration with no definition, does nothing.
   auto BuildFunctionDefinition(SemIR::FunctionId function_id) -> void;
 
-  auto BuildSubprogram(const Carbon::SemIR::Function& function,
-                       const llvm::Function* llvm_function)
+  // Build the DISubprogram metadata for the given function.
+  auto BuildDISubprogram(const Carbon::SemIR::Function& function,
```
jonmeow:
```suggestion
  auto BuildDISubprogram(const SemIR::Function& function,
```

jonmeow (approved):
Looks good, although on re-read I noticed the `Carbon::` namespacing which I assume is a typo.

toolchain/lower/file_context.cpp
```diff
@@ -360,6 +362,23 @@ auto FileContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
...
+  // (once type information is built), etc.
+  return di_builder_.createFunction(
+      di_compile_unit_, llvm_function->getName(), /*LinkageName=*/"",
+      /*File=*/nullptr,
+      /*LineNo=*/0,
+      di_builder_.createSubroutineType(
+          di_builder_.getOrCreateTypeArray(std::nullopt)),
+      /*ScopeLine=*/0, llvm::DINode::FlagZero,
+      llvm::DISubprogram::SPFlagDefinition);
+}
```
jonmeow, line 381:
```suggestion
}

```

PR #4234, opened 2024-08-21
Minor updates to support newer Clang versions.

chandlerc:
A flag was renamed after LLVM 18, and a warning caught a couple more trivially fixed issues.


jonmeow (approved):
Thanks!

PR #4235, opened 2024-08-21
Switch python syntax to use Path.parents

jonmeow:
I think `paths.parents[2]` is easier to read than `path.parent.parent.parent`, just applying uniformly. I'd noticed this while looking at #4227

Also remove a couple `resolve()` calls that shouldn't be necessary since `__file__` is absolute (elsewhere in the same files, `resolve()` is used to resolve potentially relative paths)


chandlerc (approved):
LG

Bonniemarie216 (approved):

PR #4237, opened 2024-08-22
Report invalid parse on `var x: i32 = {.7 = 8}`

brymer-meneses:


toolchain/parse/handle_period.cpp
```diff
@@ -18,6 +18,18 @@ static auto HandlePeriodOrArrow(Context& context, NodeKind node_kind,
   auto dot = context.ConsumeChecked(is_arrow ? Lex::TokenKind::MinusGreater
                                              : Lex::TokenKind::Period);
 
+  // Integer literals are invalid struct field designator
+  if (node_kind == NodeKind::StructFieldDesignator &&
+      context.PositionIs(Lex::TokenKind::IntLiteral)) {
```
jonmeow: Thanks for finding this! Looking at the code below, what if on line 39 you change:

```
} else if (context.ConsumeAndAddLeafNodeIf(Lex::TokenKind::IntLiteral,
```

to:

```
} else if (node_kind != NodeKind::StructFieldDesignator && context.ConsumeAndAddLeafNodeIf(Lex::TokenKind::IntLiteral,
```

Does that yield the similar output, removing the new diagnostic?

jonmeow (approved):
Thanks again for finding this!

PR #4238, opened 2024-08-22
Remove some explicit (Mutable)ArrayRef constructions

dwblaikie:
Rely on implicit conversion in call sites and initialization.

Removing the explicit conversions is only code simplication.
Moving from `auto x = Y(z)` to `Y x = z;` helps ensure that only implicit constructors/conversions are happening (whereas the prior syntax allows explicit conversions) which can help with readability since implicit conversions are generally "less complex"/risky/attention-requiring.


jonmeow (approved):
Thanks! This all looks better to me.

PR #4242, opened 2024-08-22
Move toolchain architecture to markdown

jonmeow:
Note I'm mostly trying to capture [the docs](https://docs.google.com/document/d/1RRYMm42osyqhI2LyjrjockYCutQ5dOf8Abu50kTrkX0/edit?resourcekey=0-kHyqOESbOHmzZphUbtLrTw&tab=t.0) as they exist today, not fixing issues with the docs. I think the doc itself hasn't changed much lately (i.e., for months). Trying to organize it a little better though, particularly so that it shows up reasonably when looking in github or the website.


toolchain/docs/parse.md
```diff
@@ -0,0 +1,912 @@
...
+
+The parse tree's storage layout is in postorder. For example, given the code:
+
+```cpp
+fn foo() -> f64 {
+  return 42;
+}
+```
+
+The node order is (with indentation to indicate nesting):
```
eyelash, line 100: The following code block is missing the indentation from the original Google Doc.

toolchain/docs/diagnostics.md
```diff
@@ -0,0 +1,222 @@
...
+
+## Diagnostic message style guide
+
+In order to provide a consistent experience, Carbon diagnostics should be
+written in the following style:
+
+-   Start diagnostics with a capital letter or quoted code, and end them with a
+    period.
+
+-   Quoted code should be enclosed in backticks, for example: `"{0} is bad."`
```
eyelash: The backticks are missing here.

jonmeow: Thanks, fixed.

toolchain/docs/parse.md
jonmeow: Fixed. Looks like this was getting auto-formatted away. Other examples are fine because they're not fully valid yaml.

toolchain/docs/adding_features.md
```diff
@@ -0,0 +1,413 @@
...
+`check/handle_*.cpp` file.
+
+If the resulting SemIR needs a new instruction:
+
+-   add a new kind to [sem_ir/inst_kind.def](/toolchain/sem_ir/inst_kind.def)
+    -   Add a `CARBON_SEM_IR_INST_KIND(NewInstKindName)` line in alphabetical
+        order
+-   a new struct definition to
+    [sem_ir/typed_insts.h](/toolchain/sem_ir/typed_insts.h), with (italics
+    highlight what changes):
```
geoffromer: The example code here is missing.

toolchain/docs/adding_features.md
```diff
@@ -0,0 +1,413 @@
...
+transitions. If several states should share handling, name them
+`FeatureAsVariant`.
+
+Adding a state requires adding a `Handle<name>` function in an appropriate
+`parse/handle_*.cpp` file, possibly a new file. The macros are used to generate
+declarations in the header, so only extra helper functions should be added
+there. Every state handler pops the state from the stack before any other
+processing.
+
+## Typed parse node metadata implementation
```
geoffromer:
```suggestion
### Typed parse node metadata implementation
```

toolchain/docs/adding_features.md
```diff
@@ -0,0 +1,413 @@
...
+
+Adding an instruction will also require a handler in the Lower step.
+
+Most new instructions will automatically be formatted reasonably by the SemIR
+formatter.
+
+If the resulting SemIR needs a new built-in, add it to
+[builtin_inst_kind.def](/toolchain/sem_ir/builtin_inst_kind.def).
+
+## SemIR typed instruction metadata implementation
```
geoffromer:
```suggestion
### SemIR typed instruction metadata implementation
```

toolchain/docs/adding_features.md
```diff
@@ -0,0 +1,413 @@
...
+    [sem_ir/typed_insts_test.cpp](/toolchain/sem_ir/typed_insts_test.cpp)
+
+## Lower
+
+Each SemIR instruction requires adding a `Handle<kind>` function in a
+`lower/handle_*.cpp` file.
+
+## Tests and debugging
+
+## Running tests
```
geoffromer:
```suggestion
### Running tests
```

toolchain/docs/adding_features.md
```diff
@@ -0,0 +1,413 @@
...
+        console.
+    -   This form will often be most useful when iterating over a specific test.
+-   `bazel run //toolchain/parse:testdata/basics/empty.carbon.verbose`
+    -   Similar to the previous command, but with the `-v` flag implied.
+-   `bazel run //toolchain/driver:carbon -- compile --phase=parse --dump-parse-tree toolchain/parse/testdata/basics/empty.carbon`
+    -   Explicitly runs `carbon` with the provided arguments.
+-   `bazel-bin/toolchain/driver/carbon compile --phase=parse --dump-parse-tree toolchain/parse/testdata/basics/empty.carbon`
+    -   Similar to the previous command, but without using `bazel`.
+
+## Updating tests
```
geoffromer:
```suggestion
### Updating tests
```

toolchain/docs/adding_features.md
```diff
@@ -0,0 +1,413 @@
...
+
+The `toolchain/autoupdate_testdata.py` script can be used to update output. It
+invokes the `file_test` autoupdate support. See
+[testing/file_test/README.md](/testing/file_test/README.md) for file syntax.
+
+### Reviewing test deltas
```
geoffromer:
```suggestion
#### Reviewing test deltas
```

toolchain/docs/adding_features.md
```diff
@@ -0,0 +1,413 @@
...
+
+Using `autoupdate_testdata.py` can be useful to produce deltas during the
+development process because it allows `git status` and `git diff` to be used to
+examine what changed.
+
+## Verbose output
```
geoffromer:
```suggestion
### Verbose output
```

toolchain/docs/adding_features.md
```diff
@@ -0,0 +1,413 @@
...
+
+The `-v` flag can be passed to trace state, and should be specified before the
+subcommand name: `carbon -v compile ...`. `CARBON_VLOG` is used to print output
+in this mode. There is currently no control over the degree of verbosity.
+
+## Stack traces
```
geoffromer:
```suggestion
### Stack traces
```

toolchain/docs/check.md
```diff
@@ -0,0 +1,608 @@
...
+    -   [ExprCategory::NotExpression](#exprcategorynotexpression)
+    -   [ExprCategory::Value](#exprcategoryvalue)
+    -   [ExprCategory::DurableReference and ExprCategory::EphemeralReference](#exprcategorydurablereference-and-exprcategoryephemeralreference)
+    -   [ExprCategory::Initializing](#exprcategoryinitializing)
+    -   [ExprCategory::Mixed](#exprcategorymixed)
+-   [Value bindings](#value-bindings)
+-   [Handling Parse::Tree errors (not yet implemented)](#handling-parsetree-errors-not-yet-implemented)
+
+<!-- tocstop -->
+
```
geoffromer, line 41: I think it would be helpful to keep the introductory paragraph here, even though it's technically redundant with README.md, because it makes this doc self-contained. Same goes for the other docs that were factored out this way.

toolchain/docs/check.md
```diff
@@ -0,0 +1,608 @@
...
+A `SemIR::InstBlock` can represent a code block. However, it can also be created
+when a series of instructions needs to be closely associated, such as a
+parameter list.
+
+A `SemIR::Builtin` represents a language built-in, such as the unconstrained
+facet type `type`. We will also have built-in functions which would need to form
+the implementation of some library types, such as `i32`. Built-ins are in a
+stable index across `SemIR` instances.
+
+## Parameters and arguments
```
geoffromer:
```suggestion
### Parameters and arguments
```

toolchain/docs/check.md
```diff
@@ -0,0 +1,608 @@
...
+first will contain the full IR, while the second will contain references to the
+last instruction for each parameter or argument. The references block will have
+a size equal to the number of parameters or arguments, allowing for quick size
+comparisons and indexed access.
+
+## SemIR textual format
+
+There are two textual ways to view `SemIR`.
+
+## Raw form
```
geoffromer:
```suggestion
### Raw form
```

toolchain/docs/check.md
```diff
@@ -0,0 +1,608 @@
...
+
+The raw form of SemIR shows the details of the representation, such as numeric
+instruction and block IDs. The representation is intended to very closely match
+the `SemIR::File` and `SemIR::Inst` representations. This can be useful when
+debugging low-level issues with the `SemIR` representation.
+
+The driver will print this when passed `--dump-raw-sem-ir`.
+
+## Formatted IR
```
geoffromer:
```suggestion
### Formatted IR
```

toolchain/docs/check.md
```diff
@@ -0,0 +1,608 @@
...
+The driver will print this when passed `--dump-sem-ir`.
+
+Unlike the raw form, certain representational choices in the `SemIR` data may
+not be visible in this form. However, it is intended to be possible to parse the
+`SemIR` output and form an equivalent – but not necessarily identical – `SemIR`
+representation, although no such parser currently exists.
+
+As an example, given the program:
+
+```cpp
```
geoffromer: I think we're using `carbon` for Carbon code blocks, even though it doesn't do anything yet.

toolchain/docs/check.md
```diff
@@ -0,0 +1,608 @@
...
+
+Names in formatted IR are all invented by the formatter, and generally are of
+the form `<base_name>[.loc<line>[_<col>[.<counter>]]]` where `<line>` and
+`<col>` describe the location of the instruction, and `<counter>` is used as a
+disambiguator if multiple instructions appear at the same location. Trailing
+name components are only included if they are necessary to disambiguate the
+name. `<base_name>` is a guessed good name for the instruction, often derived
+from source-level identifiers, and is empty if no guess was made.
+
+### Instructions
```
geoffromer:
```suggestion
#### Instructions
```

toolchain/docs/check.md
```diff
@@ -0,0 +1,608 @@
...
+    collapsed into a single
+    `if %cond br !label1 else if ... else br !labelN(%arg)` line.
+-   A struct type, formed by a sequence of `StructTypeField` instructions
+    followed by a `StructType` instruction, is collapsed into a single
+    `struct_type{.field1: %value1, ..., .fieldN: %valueN}` line.
+
+These exceptions may be found in
+[toolchain/sem_ir/formatter.cpp](/toolchain/sem_ir/formatter.cpp).
+
+### Top-level entities
```
geoffromer:
```suggestion
#### Top-level entities
```

toolchain/docs/check.md
```diff
@@ -0,0 +1,608 @@
...
+[check/context.h](/toolchain/check/context.h), which stores things in a
+collection of stacks. The common pattern is that the children of a node are
+processed first. They produce information that is then consumed when processing
+the parent node.
+
+One example of this pattern is expressions. Each subexpression outputs SemIR
+instructions to compute the value of that subexpression to the current
+instruction block, added to the top of the
+
+`InstBlockStack` stored in the `Context` object. It leaves an instruction id on
```
geoffromer: Accidental paragraph break

toolchain/docs/check.md
```diff
@@ -0,0 +1,608 @@
...
+In all three cases, the introducer node `ReturnStatementStart` pushes an entry
+on the [node stack](#node-stack) with just the parse node and no id, called a
+_solo parse node_. The handler for the parent `ReturnStatement` node can pop and
+process entries from the node stack until it finds that solo parse node from
+`ReturnStatementStart` that indicates it is done.
+
+Another pattern that arises is state is set up by an introducer node, updated by
+its siblings, and then consumed by the bracketing parent node. FIXME: example
+
+## Node stack
```
geoffromer:
```suggestion
### Node stack
```

toolchain/docs/check.md
```diff
@@ -0,0 +1,608 @@
...
+the parent node knows how many nodes it must pop from the stack based either on
+knowing the fixed number of children for that node kind or popping nodes until
+it reaches a bracketing node. The arity or bracketing node kind for each parent
+node is documented in [parse/node_kind.def](/toolchain/parse/node_kind.def).
+
+When each `Parse::Node` is evaluated, the SemIR for it is typically immediately
+generated as `SemIR::Inst`s. To help generate the IR to an appropriate context,
+scopes have separate `SemIR::InstBlock`s.
+
+## Delayed evaluation (not yet implemented)
```
geoffromer:
```suggestion
### Delayed evaluation (not yet implemented)
```

toolchain/docs/check.md
```diff
@@ -0,0 +1,608 @@
...
+Sometimes, nodes will need to have delayed evaluation; for example, an inline
+definition of a class member function needs to be evaluated after the class is
+fully declared. The `SemIR::Inst`s cannot be immediately generated because they
+may include name references to the class. We're likely to store a reference to
+the relevant `Parse::Node` for each definition for re-evaluation after the class
+scope completes. This means that nodes in a definition would be traversed twice,
+once while determining that they're inline and without full checking or IR
+generation, then again with full checking and IR generation.
+
+## Templates (not yet implemented)
```
geoffromer:
```suggestion
### Templates (not yet implemented)
```

toolchain/docs/check.md
```diff
@@ -0,0 +1,608 @@
...
+
+Templates need to have partial semantic checking when declared, but can't be
+fully implemented before they're instantiated against a specific type.
+
+We are likely to generate a partial IR for templates, allowing for checking with
+the incomplete information in the IR. Instantiation will likely use that IR and
+fill in the missing information, but it could also reevaluate the original
+`Parse::Node`s with the known template state.
+
+## Rewrites
```
geoffromer:
```suggestion
### Rewrites
```

toolchain/docs/check.md
```diff
@@ -0,0 +1,608 @@
...
+instructions to be created separately from the node block stack, and merge those
+instructions into the current block once we have decided on their contents.
+
+## Types
+
+Type expressions are treated like any other expression, and are modeled as
+`SemIR::Inst`s. The types computed by type expressions are deduplicated,
+resulting in a canonical `SemIR::TypeId` for each distinct type.
+
+## Type printing (not yet implemented)
```
geoffromer:
```suggestion
### Type printing (not yet implemented)
```

toolchain/docs/check.md
```diff
@@ -0,0 +1,608 @@
...
+### ExprCategory::Mixed
+
+This instruction represents a language construct that doesn't have a single
+expression category. This is used for struct and tuple literals, where the
+elements of the literal can have different expression categories. Instructions
+with a mixed expression category are treated as a special case in conversion,
+which recurses into the elements of those instructions before performing
+conversions.
+
+## Value bindings
```
geoffromer:
```suggestion
### Value bindings
```

toolchain/docs/check.md
```diff
@@ -0,0 +1,608 @@
...
+
+# Lower
+
+Lowering takes the SemIR and produces LLVM IR. At present, this is done in a
+single pass, although it's possible we may need to do a second pass so that we
+can first generate type information for function arguments.
+
+Lowering is done per `SemIR::InstBlock`. This minimizes changes to the
+`IRBuilder` insertion point, something that is both expensive and potentially
+fragile.
```
geoffromer: This was presumably supposed to go in README.md, not here.

toolchain/docs/parse.md
```diff
@@ -0,0 +1,917 @@
...
+```carbon
+var x: i32 = y + 1;
+```
+
+Lexing creates distinct tokens for each syntactic element, which will form the
+basis of the parse tree:
+
+<!--
+These charts are using a hidden "root" node in order to adjust alignment.
+We're trying to keep the "lexed" tokens separate from "parse tree" nodes.
```
geoffromer: I find these figures pretty hard to follow, compared to the ones in the original doc. Maybe we should keep them as SVG until someone has time to make Mermaid render them a little more like the originals?

toolchain/docs/parse.md
```diff
@@ -0,0 +1,917 @@
...
+    node9 --- node1 & node3 & node5 & node7
+    node9 ~~~ root
+```
+
+In storage, this tree will be flat and in postorder. Because the order hasn't
+changed much from the original code, we can do the reordering for postorder with
+a minimal number of nodes being delayed for later output: it will be linear with
+respect to the depth of the parse tree.
+
+```mermaid
```
geoffromer: This figure especially suffers from the translation to Mermaid, because you lose the way the nodes line up across the three levels.

toolchain/docs/parse.md
```diff
@@ -0,0 +1,917 @@
...
+prevent the parent state from emitting redundant diagnostics; for example, an
+invalid expression might have more invalid tokens following it, and the parent
+might skip those without emitting diagnostics.
+
+## How is this accomplished?
+
+The specific approach to producing the desired tree depends on the kind of
+grammar rule being implemented, as well as the desired output tree structure.
+
+## Introducer
```
geoffromer:
```suggestion
### Introducer
```

toolchain/docs/parse.md
```diff
@@ -0,0 +1,917 @@
...
+context.AddNode(NodeKind::ForStatement, state.token, state.subtree_start,
+                state.has_error);
+```
+
+If this state is for an element of a scope like the statements in a code block,
+most introducer tokens indicate that the current state should be repeated, to
+handle the next statement, but some other token, like a close curly brace (`}`)
+means that the state should be exited.
+
+## Optional modifiers before an introducer
```
geoffromer:
```suggestion
### Optional modifiers before an introducer
```

toolchain/docs/parse.md
```diff
@@ -0,0 +1,917 @@
...
+    ```cpp
+    state = context.PopState();
+    context.AddNode(NodeKind::___, context.Consume(),
+                    state.subtree_start, state.has_error);
+    ```
+
+    -   This `state.subtree_start` will mark everything since the bracketing
+        introducer node as the children of this node.
+
+## Something required in context
```
geoffromer:
```suggestion
### Something required in context
```

toolchain/docs/parse.md
```diff
@@ -0,0 +1,917 @@
...
+
+FIXME
+
+Example: name after introducer
+[parse/handle_decl_name_and_params.cpp](/toolchain/parse/handle_decl_name_and_params.cpp)
+
+Example: `[<implicit parameter list>]` after `impl forall`
+[parse/handle_impl.cpp](/toolchain/parse/handle_impl.cpp)
+
+## Optional clauses
```
geoffromer:
```suggestion
### Optional clauses
```

toolchain/docs/parse.md
```diff
@@ -0,0 +1,917 @@
...
+
+### Case 1: introducer to optional clause is used as parent node
```
geoffromer:
```suggestion
#### Case 1: introducer to optional clause is used as parent node
```

toolchain/docs/parse.md
```diff
@@ -0,0 +1,917 @@
...
+The `->` token is saved by `context.PushState(`...`)`, so it is available as
+`state.token` when calling
+`context.AddNode(NodeKind::ReturnType, state.token,`...`)` later in
+`HandleFunctionReturnTypeFinish`.
+
+Also see how the optional initializer is handled on `var`, treating the `=` as
+its introducer in `HandleVarAfterPattern` and `HandleVarInitializer` in
+[parse/handle_var.cpp](/toolchain/parse/handle_var.cpp).
+
+### Case 2: parent node is required token after optional clause, with different parent node kinds for different options
```
geoffromer:
```suggestion
#### Case 2: parent node is required token after optional clause, with different parent node kinds for different options
```

toolchain/docs/parse.md
```diff
@@ -0,0 +1,917 @@
...
+-   Error from no `as` present but the type expression was okay -> create a new
+    error.
+-   There was error from the child type expression and no `as` present -> no new
+    diagnostic, we suppress errors once one is emitted until we can recover.
+
+If there is no `as` token, we don't output either a `TypeImplAs` or a
+`DefaultSelfImplAs` node, as required by the parent node, so in those cases we
+mark the parent as having an error.
+
+### Case 3: optional sibling
```
geoffromer:
```suggestion
#### Case 3: optional sibling
```

toolchain/docs/parse.md
```diff
@@ -0,0 +1,917 @@
...
+  context.PushState(State::ImplBeforeAs);
+  context.PushStateForExpr(PrecedenceGroup::ForImplAs());
+}
+```
+
+and then `HandleImplBeforeAs` follows
+[the "something required in context" pattern](#something-required-in-context) to
+deal with the `as` that follows when the type expression is present.
+
+## Operators
```
geoffromer:
```suggestion
### Operators
```

toolchain/docs/parse.md
```diff
@@ -0,0 +1,917 @@
...
+        introducer node as the children of this node.
+
+## Something required in context
+
+FIXME
+
+Example: name after introducer
+[parse/handle_decl_name_and_params.cpp](/toolchain/parse/handle_decl_name_and_params.cpp)
+
+Example: `[<implicit parameter list>]` after `impl forall`
```
geoffromer:
```suggestion
Example: `[` <implicit parameter list> `]` after `impl forall`
```

jonmeow: `<>` are HTML characters, so I was choosing to make this code formatted instead of `&lt;` which seemed less readable for people who look at the md file instead of the HTML. That is, do you find it more readable as-is, or as:

```suggestion
Example: `[` &lt;implicit parameter list> `]` after `impl forall`
```

toolchain/docs/parse.md
jonmeow: My preference would be to switch towards Mermaid for maintainability -- I think the current is pretty good. It will also work better with GitHub: for example, the rendering should adjust based on GitHub's light and dark modes, whereas an svg cannot. I understand docs is better for some things, but I do think it's worse for others.

Note the svg's in the main README have also been the source of several complaints over time. I don't expect the svg's I'm keeping in this PR to do better, but I actually wonder if I can remove and/or simplify those in order as part of making the relevant sections more helpful.

toolchain/docs/parse.md
jonmeow: Thanks, I understand. Do the first and third line up for you, though? The rest I did as best I could (I spent a while tinkering with the diagrams), but happy to take suggested mermaid edits if you have ideas for how to improve.

toolchain/docs/check.md
jonmeow: I prefer to avoid redundancy due to the likely skew over time from putting identical content in multiple files. How about instead this approach which drops the summaries from README.md?

Also, I noticed "lower" was in the wrong file, so fixing that, while also giving the driver and lex their own files. They could probably have more detail anyways.

toolchain/docs/check.md
jonmeow: I'd fixed a few of these (the doc was using cpp highlighting for a lot), but missed here and have fixed now.

toolchain/docs/check.md
jonmeow: Fixed

toolchain/docs/check.md
jonmeow: Thanks, moved to lower.md

toolchain/docs/adding_features.md
jonmeow: Fixed. Dropped the italics note since that's infeasible to retain, adding a comment where I thought the intent was unclear.

toolchain/docs/parse.md
geoffromer: I think I'd ideally prefer something like this:

```suggestion
Example: "`[` _implicit parameter list_ `]`" after `impl forall`
```

If that's too much scope drift for this PR, I think what you have now is better than the `&lt;` version.

toolchain/docs/parse.md
geoffromer: Yeah, I understand the maintainability concerns, and themeability is certainly nice, but none of that matters unless the figures are successfully communicating the ideas, and right now I don't know if they are. At least for me, what made the original figures effective was that they worked like frames of animation: items only change position within the frame when the figure is depicting a state change, and the frames are small enough that I can keep several of them on screen and visually scan between them to see the changes. I think it was also crucial that the depiction of the tree made the postorder structure explicit -- for example, on line 361 the main point we make about the following figure is that the `;` is bracketed by `var`, but that isn't visible the new figure unless the `var` is on the far left and the `;` is on the far right. Without those things, I don't know if I would have understood what the figures were trying to tell me.

Maybe we should get a second opinion on this?

toolchain/docs/parse.md
geoffromer: Yes, the first and third lines line up. I don't know Mermaid, but from what little I've seen, I'm not optimistic that it can give us the level of visual control that we would need to fix these problems, which is why I suggested switching to SVG.

jonmeow:
@zygo

jonmeow:
@zygoloid Maybe you can give your thoughts on the parse charts, at https://github.com/carbon-language/carbon-lang/pull/4242#discussion_r1736967802? Geoffrey is asking to go back to svg, I think the mermaid charts are good enough to understand what's going on, maybe you can weigh in.

toolchain/docs/parse.md
jonmeow: Sure, adding @zygoloid. A lot of people are out with the long weekend.

jonmeow: BTW, here's some edits that layer the frames more.

jonmeow: Note, three different approaches:

- [Original](https://github.com/carbon-language/carbon-lang/blob/a6b3e2ad797e2992610d2c7db55b989981d9ef2d/toolchain/docs/parse.md#visual-example)
- [More overlaid](https://github.com/carbon-language/carbon-lang/blob/4b560db502cf720c25488506297970acbfa9d4a8/toolchain/docs/parse.md#visual-example)
- [With some subtle final changes](https://github.com/carbon-language/carbon-lang/blob/9fb8c55db0d87af3449fb459d15dab209749bac6/toolchain/docs/parse.md#visual-example)

jonmeow: Per discussion, ascii art

jonmeow:
> @zygoloid Maybe you can give your thoughts on the parse charts, at [#4242 (comment)](https://github.com/carbon-language/carbon-lang/pull/4242#discussion_r1736967802)? Geoffrey is asking to go back to svg, I think the mermaid charts are good enough to understand what's going on, maybe you can weigh in.

Just to note, we ended up discussing this in open discussion on Thursday. @chandlerc's suggested the ascii graphs.

toolchain/docs/parse.md
```diff
@@ -0,0 +1,758 @@
...
+                                                    | ; |
+                                                    +---+
+```
+
+**Flattened for storage**:
+
+```ascii
++-----+ +---+ +-----+ +---+ +---+ +---+ +---+ +---+ +---+
+| var | | x | | i32 | | : | | = | | y | | 1 | | + | | ; |
++-----+ +---+ +-----+ +---+ +---+ +---+ +---+ +---+ +---+
```
geoffromer: I think it might work better to integrate the labels into the ASCII art. As it is, the boundaries between ASCII blocks (which are visible in the rendered form as well as the markdown) make these feel like three separate figures, and push the eye away from seeing the vertical alignment which is one of the key points of the figure.

```suggestion
```ascii
Tokens:
+-----+ +---+ +---+ +-----+ +---+ +---+ +---+ +---+ +---+
| var | | x | | : | | i32 | | = | | y | | + | | 1 | | ; |
+-----+ +---+ +---+ +-----+ +---+ +---+ +---+ +---+ +---+

Parse tree:
        +---+ +-----+             +---+ +---+
        | x | | i32 |             | y | | 1 |
        +---+ +-----+             +---+ +---+
          |      |                  |     |
          +------+------+           +-----+-----+
                        |                       |
+-----+               +---+ +---+             +---+
| var |               | : | | = |             | + |
+-----+               +---+ +---+             +---+
   |                    |     |                 |
   +--------------------+-----+-----------------+-----+
                                                      |
                                                    +---+
                                                    | ; |
                                                    +---+
Flattened for storage:
+-----+ +---+ +-----+ +---+ +---+ +---+ +---+ +---+ +---+
| var | | x | | i32 | | : | | = | | y | | 1 | | + | | ; |
+-----+ +---+ +-----+ +---+ +---+ +---+ +---+ +---+ +---+
```

I think we should add those embedded captions to the other figures as well (see previous suggestions), so that there's clearer visual continuity between the figures, as per my earlier description of them as frames of animation.

Optional: if it works, it might be worthwhile to make these `<pre>` blocks instead of markdown code blocks, in order to put `<b></b>` tags around the captions.

toolchain/docs/parse.md
```diff
@@ -0,0 +1,758 @@
...
+
+```carbon
+var x: i32 = y + 1;
+```
+
+Lexing creates distinct tokens for each syntactic element, which will form the
+basis of the parse tree:
+
+```ascii
++-----+ +---+ +---+ +-----+ +---+ +---+ +---+ +---+ +---+
```
geoffromer:
```suggestion
Tokens:
+-----+ +---+ +---+ +-----+ +---+ +---+ +---+ +---+ +---+
```

(See comment on line 327 for the rationale for this and the following suggestions)

toolchain/docs/parse.md
```diff
@@ -0,0 +1,758 @@
...
+| var | | x | | : | | i32 | | = | | y | | + | | 1 | | ; |
++-----+ +---+ +---+ +-----+ +---+ +---+ +---+ +---+ +---+
+```
+
+First the `var` keyword is used as a "bracketing" node (VariableIntroducer).
+When this is seen in a postorder traversal, it tells us to expect the basics of
+a variable declaration structure.
+
+```ascii
+        +---+ +---+ +-----+ +---+ +---+ +---+ +---+ +---+
```
geoffromer:
```suggestion
Tokens:
        +---+ +---+ +-----+ +---+ +---+ +---+ +---+ +---+
```

toolchain/docs/parse.md
```diff
@@ -0,0 +1,758 @@
...
+        | x | | : | | i32 | | = | | y | | + | | 1 | | ; |
+        +---+ +---+ +-----+ +---+ +---+ +---+ +---+ +---+
+```
+
+```ascii
```
geoffromer:
```suggestion

Parse tree:
```

toolchain/docs/parse.md
```diff
@@ -0,0 +1,758 @@
...
++-----+
+```
+
+Next, we can consider the pattern binding. Here, `x` is the identifier and `i32`
+is the type expression. The `:` provides a parent node that must always contain
+two children, the name and type expression. Because it always has two direct
+children, it doesn't need to be bracketed.
+
+```ascii
+                            +---+ +---+ +---+ +---+ +---+
```
geoffromer:
```suggestion
Tokens:
                            +---+ +---+ +---+ +---+ +---+
```

toolchain/docs/parse.md
```diff
@@ -0,0 +1,758 @@
...
+                            | = | | y | | + | | 1 | | ; |
+                            +---+ +---+ +---+ +---+ +---+
+```
+
+```ascii
```
geoffromer:
```suggestion

Parse tree:
```

toolchain/docs/parse.md
```diff
@@ -0,0 +1,758 @@
...
+indicate the transition from binding to assignment expression, which is
+important for expression parsing during checking.
+
+```ascii
+                                  +---+ +---+ +---+ +---+
+                                  | y | | + | | 1 | | ; |
+                                  +---+ +---+ +---+ +---+
+```
+
+```ascii
```
geoffromer:
```suggestion

Parse tree:
```

toolchain/docs/parse.md
```diff
@@ -0,0 +1,758 @@
...
+| var |               | : |
++-----+               +---+
+```
+
+We use the `=` as a separator (instead of a node with children like `:`) to help
+indicate the transition from binding to assignment expression, which is
+important for expression parsing during checking.
+
+```ascii
+                                  +---+ +---+ +---+ +---+
```
geoffromer:
```suggestion
Tokens:
                                  +---+ +---+ +---+ +---+
```

toolchain/docs/parse.md
```diff
@@ -0,0 +1,758 @@
...
++-----+               +---+ +---+
+| var |               | : | | = |
++-----+               +---+ +---+
+```
+
+The expression is a subtree with `+` as the parent, and the two operands as
+child nodes.
+
+```ascii
+                                                    +---+
```
geoffromer:
```suggestion
Tokens:
                                                    +---+
```

toolchain/docs/parse.md
```diff
@@ -0,0 +1,758 @@
...
+                                                    | ; |
+                                                    +---+
+```
+
+```ascii
```
geoffromer:
```suggestion

Parse tree:
```

toolchain/docs/parse.md
```diff
@@ -0,0 +1,758 @@
...
+| var |               | : | | = |             | + |
++-----+               +---+ +---+             +---+
+```
+
+Finally, the `;` is used as the "root" of the variable declaration. It's
+explicitly tracked as the `;` for a variable declaration so that it's
+unambiguously bracketed by `var`.
+
+```ascii
+        +---+ +-----+             +---+ +---+
```
geoffromer:
```suggestion
Tokens:




Parse tree:
        +---+ +-----+             +---+ +---+
```

toolchain/docs/parse.md
jonmeow: I think I'd like to hear more opinions on this. For me, I find it hard to read so much ascii and get a distinction. The switch in font and the boundaries of the sections is intended to reflect that these are distinctly separate storage locations, even if they are conceptually related.

jonmeow: That is, the separation you are bothered by is separation which I am trying to make easy to see. So I think we're at cross purposes here.

toolchain/docs/parse.md
jonmeow: Note, I was omitting labels in [ed: earlier] charts because they are not in the toolchain architecture doc.

eyelash:
Not sure what exactly was discussed in open discussion (can't find it in the minutes) but using [box-drawing characters](https://en.wikipedia.org/wiki/Box-drawing_characters) would probably make these ASCII diagrams easier to read.

For example:

<pre>
        ┌───┐ ┌─────┐             ┌───┐ ┌───┐
        │ x │ │ i32 │             │ y │ │ 1 │
        └─┬─┘ └──┬──┘             └─┬─┘ └─┬─┘
          │      │                  │     │
          └──────┴──────┐           └─────┴─────┐
                        │                       │
┌─────┐               ┌─┴─┐ ┌───┐             ┌─┴─┐
│ var │               │ : │ │ = │             │ + │
└──┬──┘               └─┬─┘ └─┬─┘             └─┬─┘
   │                    │     │                 │
   └────────────────────┴─────┴─────────────────┴─────┐
                                                      │
                                                    ┌─┴─┐
                                                    │ ; │
                                                    └───┘
</pre>

jonmeow:
> Not sure what exactly was discussed in open discussion (can't find it in the minutes) but using [box-drawing characters](https://en.wikipedia.org/wiki/Box-drawing_characters) would probably make these ASCII diagrams easier to read.
> 
> For example:
> 
>         ┌───┐ ┌─────┐             ┌───┐ ┌───┐
>         │ x │ │ i32 │             │ y │ │ 1 │
>         └─┬─┘ └──┬──┘             └─┬─┘ └─┬─┘
>           │      │                  │     │
>           └──────┴──────┐           └─────┴─────┐
>                         │                       │
> ┌─────┐               ┌─┴─┐ ┌───┐             ┌─┴─┐
> │ var │               │ : │ │ = │             │ + │
> └──┬──┘               └─┬─┘ └─┬─┘             └─┬─┘
>    │                    │     │                 │
>    └────────────────────┴─────┴─────────────────┴─────┐
>                                                       │
>                                                     ┌─┴─┐
>                                                     │ ; │
>                                                     └───┘

This kind of character set change would require a chunk of time to just rewrite the graphs, and would be harder to update versus the basic ascii currently in use. It might make sense to consider separately, but given this review is jammed on specific style of these graphs, I want to be careful about investing lots of time into aesthetics that might be undone.

toolchain/docs/parse.md
jonmeow: Updated per discussion, including making nodes equal width.

geoffromer (approved):

PR #4243, opened 2024-08-22
 Add filename and line number to function debug info metadata

dwblaikie:
Refactors a bunch of the SemIRDiagnosticConverter to be able to use that from Lower to access source locations there to use in debug info.

I assume some of this is a bit jank/would need to be fixed/improved in the future - like the context functor that's passed into ConvertLoc? (not totally clear what that's for/what the debug info will be missing out on in its absence, I could throw a FIXME in there if you like)


toolchain/lower/file_context.cpp
```diff
@@ -32,11 +33,13 @@ FileContext::FileContext(llvm::LLVMContext& llvm_context,
           include_debug_info
               ? BuildDICompileUnit(module_name, *llvm_module_, di_builder_)
               : nullptr),
+      converter_(converter),
       sem_ir_(&sem_ir),
       inst_namer_(inst_namer),
       vlog_stream_(vlog_stream) {
   CARBON_CHECK(!sem_ir.has_errors())
       << "Generating LLVM IR from invalid SemIR::File is unsupported.";
+  (void)converter_;
```
```suggestion
```

jonmeow (commented):
This seems good, one small edit.

jonmeow (approved):

PR #4244, opened 2024-08-23
Teach source generation to reference more interesting types.

chandlerc:
This teaches our source generation tool to create interesting type references. This include both referencing a weighted distribution of explicitly specified types, and referencing types that are being defined in the generated file.

Generating more interesting explicit types will exercise more of Carbon's prelude, but because C++ doesn't have an automatic prelude with fundamental types like `int64_t` or tuples, we include some minimal headers when generating the C++ analog. This likely makes the comparison more fair rather than less fair as Carbon's toolchain isn't processing just the generated source, but also its prelude.

The current set of fixed types is based primarily on the set of types that the toolchain currently implements and a set that seems reasonably interesting to exercise for compile time performance. We want to try to cover things that should be optimized in the toolchain, even if a single source file might not typically hit all of them.

The weights of everything are completely arbitrary, based on intuition and some hand inspection of some random source files. There is also an intentional bias towards non-zero coverage and so the tail is much larger than it should be in reality. The result is that the weights more reflect the _priority_ of optimizing compile time than the _observed_ distribution in practice. We can refine the weighting scheme in the future though, potentially with multiple modes to separate coverage from maximally representative weights, etc. The goal is just to have a starting point.

The scheme for referencing the defined types requires some care and complexity to avoid referencing types before they are defined while still referencing all of the types defined and ensuring the number of references is stable even as the order is randomized to avoid fixed patterns in the source code.

All of this also triggered some minor refactoring of the state used to generate class definitions in the source generator. There are probably some good follow-on refactoring opportunities, but I'd prefer to leave those to future work.

I don't have any tests here because most of how this is observable is already tested -- the existing tests ensure the file sizes remain consistent and that the generated code is compiled correctly. But if folks have any ideas of useful tests here, happy to add them.

### DIFF
--- a/testing/base/source_gen.cpp
+++ b/testing/base/source_gen.cpp
@@ -163,8 +163,7 @@ auto SourceGen::ClassGenState::GetValidTypeName() -> llvm::StringRef {
 // references to use throughout those classes.
 //
 // We combine a list of fixed types in the `type_use_params` with the list of
-// class names that will be defined to form the spelling of all
-// the referenced types. The `type_use_params` provides weights for each fixed
+// class names that will be defined to form the spelling of all the referenced types. The `type_use_params` provides weights for each fixed
 // type as well as an overall weight for referencing class names that are being
 // declared. We build a set of type references so that its histogram will
 // roughly match these weights.
--- a/testing/base/source_gen.cpp
+++ b/testing/base/source_gen.cpp
@@ -28,8 +28,8 @@ constexpr static int MinMemberNameLength = 4;
 
 // The shuffled state used to generate some number of classes.
 //
-// This state encodes everything used to generate class definitions. The state will be
-// consumed until empty.
+// This state encodes everything used to generate class definitions. The state
+// will be consumed until empty.
 //
 // Detailed comments for out-of-line methods are on their definitions.
 class SourceGen::ClassGenState {
@@ -219,7 +219,8 @@ auto SourceGen::ClassGenState::BuildClassAndTypeNames(
   // not overshoot and may end up with a remainder.
   for (const auto& fixed_type_weight : type_use_params.fixed_type_weights) {
     // For very small `num_types`, we may exhaust the fixed types. For
-    // simplicity, we don't try to scale the weights down, we just cut off early.
+    // simplicity, we don't try to scale the weights down, we just cut off
+    // early.
     if (static_cast<int>(type_names_.size()) == num_types) {
       break;
     }
@@ -859,7 +860,8 @@ auto SourceGen::GenerateClassDef(const ClassParams& params,
   }
 
   // Mark this class as now a valid type now that field type names have been
-  // collected. We can reference this class from functions and methods within the definition.
+  // collected. We can reference this class from functions and methods within
+  // the definition.
   state.AddValidTypeName(name);
 
   UniqueIdentifierPopper unique_member_names(*this, state.member_names());
--- a/testing/base/source_gen.cpp
+++ b/testing/base/source_gen.cpp
@@ -163,10 +163,11 @@ auto SourceGen::ClassGenState::GetValidTypeName() -> llvm::StringRef {
 // references to use throughout those classes.
 //
 // We combine a list of fixed types in the `type_use_params` with the list of
-// class names that will be defined to form the spelling of all the referenced types. The `type_use_params` provides weights for each fixed
-// type as well as an overall weight for referencing class names that are being
-// declared. We build a set of type references so that its histogram will
-// roughly match these weights.
+// class names that will be defined to form the spelling of all the referenced
+// types. The `type_use_params` provides weights for each fixed type as well as
+// an overall weight for referencing class names that are being declared. We
+// build a set of type references so that its histogram will roughly match these
+// weights.
 //
 // For each of the fixed types, `type_use_params` provides a spelling for both
 // Carbon and C++.
--- a/testing/base/source_gen.cpp
+++ b/testing/base/source_gen.cpp
@@ -160,24 +160,25 @@ auto SourceGen::ClassGenState::GetValidTypeName() -> llvm::StringRef {
 }
 
 // We want to synthesize shuffled type names from the class names that will be
-// declared, but mixing in `i32` and other builtin types. A somewhat hack-y
-// approach, but currently the strategy is to map every type shorter than the
-// `MinClassNameLength` to one of our builtin types and the longer names to
-// class names in the state.
+// declared, but mixing in `i32` and other builtin types.
 //
-// We use non-unique random identifiers to get a sequence of lengths, and map
-// each identifier to one of our desired types as above. This allows repetition
-// as the identifiers repeat. We can also create a larger pool of builtin type
-// spellings eventually and use the different short identifier spellings to
-// select from this pool.
+// We combine a list of fixed types in the `type_use_params` with the list of
+// (un-shuffled) class names that will be defined to form the spelling of all
+// the referenced types. The `type_use_params` provides weights for each fixed
+// type as well as an overall weight for referencing class names that are being
+// declared. We build a set of type references so that its histogram will
+// roughly match these weights.
 //
-// When searching for a given length, we allow wrapping around the class names.
-// This may map several random identifiers to the same declared class name.
+// For the fixed types, we assume each type has a spelling that will be valid
+// provided by params for both Carbon and C++.
 //
-// TODO: Variation in how many times a name is repeated may introduce some
-// unnecessary noise, we should consider a fixed distribution of duplication
-// rather than just accepting whatever occurs randomly. Currently this only
-// stabilizes the total text size resulting from referencing prior type names.
+// For referencing declared class names, we use the un-shuffled sequence, and
+// evenly distribute our references across it to the extent possible.
+//
+// Once all of the references are formed, we randomly shuffle the entire
+// sequence to provide an unpredictable order of reference, and also shuffle the
+// declared class names now that we're not sampling from that list to build the
+// references.
 auto SourceGen::ClassGenState::BuildClassAndTypeNames(
     SourceGen& gen, int num_classes, int num_types,
     const TypeUseParams& type_use_params) -> void {
--- a/testing/base/source_gen.cpp
+++ b/testing/base/source_gen.cpp
@@ -163,22 +163,20 @@ auto SourceGen::ClassGenState::GetValidTypeName() -> llvm::StringRef {
 // declared, but mixing in `i32` and other builtin types.
 //
 // We combine a list of fixed types in the `type_use_params` with the list of
-// (un-shuffled) class names that will be defined to form the spelling of all
+// class names that will be defined to form the spelling of all
 // the referenced types. The `type_use_params` provides weights for each fixed
 // type as well as an overall weight for referencing class names that are being
 // declared. We build a set of type references so that its histogram will
 // roughly match these weights.
 //
-// For the fixed types, we assume each type has a spelling that will be valid
-// provided by params for both Carbon and C++.
+// For each of the fixed types, `type_use_params` provides a spelling for both Carbon and C++.
 //
-// For referencing declared class names, we use the un-shuffled sequence, and
-// evenly distribute our references across it to the extent possible.
+// We distribute our references to declared class names evenly to the extent possible.
 //
-// Once all of the references are formed, we randomly shuffle the entire
-// sequence to provide an unpredictable order of reference, and also shuffle the
-// declared class names now that we're not sampling from that list to build the
-// references.
+// Before all the references are formed, the class names are kept their original unshuffled order.
+// This ensures that any uneven sampling of names is done deterministically. At the end, we
+// randomly shuffle the sequences of both the declared class names and type references
+// to provide an unpredictable order in the generated output.
 auto SourceGen::ClassGenState::BuildClassAndTypeNames(
     SourceGen& gen, int num_classes, int num_types,
     const TypeUseParams& type_use_params) -> void {
--- a/testing/base/source_gen.cpp
+++ b/testing/base/source_gen.cpp
@@ -28,8 +28,7 @@ constexpr static int MinMemberNameLength = 4;
 
 // The shuffled state used to generate some number of classes.
 //
-// This state encodes all the shuffled entropy used for generating a number of
-// class definitions. While generating definitions, the state here will be
+// This state encodes everything used to generate class definitions. The state will be
 // consumed until empty.
 //
 // Detailed comments for out-of-line methods are on their definitions.
@@ -209,7 +208,7 @@ auto SourceGen::ClassGenState::BuildClassAndTypeNames(
   // Now append the remainder number of class names. This is where the class
   // names being un-shuffled is essential. We're going to have one extra
   // reference to some fraction of the class names and we want that to be a
-  // stable fraction.
+  // stable subset.
   type_names_.append(class_names_.begin(),
                      class_names_.begin() + (num_declared_types % num_classes));
   CARBON_CHECK(static_cast<int>(type_names_.size()) == num_declared_types);
@@ -218,8 +217,8 @@ auto SourceGen::ClassGenState::BuildClassAndTypeNames(
   // type. This isn't exact however, so we'll both clamp the number of copies to
   // not overshoot and may end up with a remainder.
   for (const auto& fixed_type_weight : type_use_params.fixed_type_weights) {
-    // For very small numbers of types we may exhaust the fixed types. We don't
-    // try to scale the weights down for simplicity, we just cut off early.
+    // For very small `num_types`, we may exhaust the fixed types. For
+    // simplicity, we don't try to scale the weights down, we just cut off early.
     if (static_cast<int>(type_names_.size()) == num_types) {
       break;
     }
@@ -235,9 +234,7 @@ auto SourceGen::ClassGenState::BuildClassAndTypeNames(
   // through the fixed types without any weighting. With reasonably large
   // numbers of types this won't distort the distribution in an interesting way
   // and is simpler than trying to scale the distribution down.
-  for ([[maybe_unused]] auto _ :
-       llvm::seq(1 + (num_types - type_names_.size()) /
-                         type_use_params.fixed_type_weights.size())) {
+  while (static_cast<int>(type_names_.size()) < num_types) {
     for (const auto& fixed_type_weight :
          llvm::ArrayRef(type_use_params.fixed_type_weights)
              .take_front(num_types - type_names_.size())) {
@@ -852,9 +849,8 @@ auto SourceGen::GenerateClassDef(const ClassParams& params,
     os << " public:\n";
   }
 
-  // Collect field types before any others as those can't be the class we're
-  // currently declaring. We do this before inserting that type into the valid
-  // set.
+  // Field types can't be the class we're currently declaring. We enforce this
+  // by collecting them before inserting that type into the valid set.
   llvm::SmallVector<llvm::StringRef> field_type_names;
   field_type_names.reserve(params.private_field_decls);
   for ([[maybe_unused]] int _ : llvm::seq(params.private_field_decls)) {
@@ -862,7 +858,7 @@ auto SourceGen::GenerateClassDef(const ClassParams& params,
   }
 
   // Mark this class as now a valid type now that field type names have been
-  // collected. We can self-reference from other types within the definition.
+  // collected. We can reference this class from functions and methods within the definition.
   state.AddValidTypeName(name);
 
   UniqueIdentifierPopper unique_member_names(*this, state.member_names());
--- a/testing/base/source_gen.cpp
+++ b/testing/base/source_gen.cpp
@@ -169,14 +169,17 @@ auto SourceGen::ClassGenState::GetValidTypeName() -> llvm::StringRef {
 // declared. We build a set of type references so that its histogram will
 // roughly match these weights.
 //
-// For each of the fixed types, `type_use_params` provides a spelling for both Carbon and C++.
+// For each of the fixed types, `type_use_params` provides a spelling for both
+// Carbon and C++.
 //
-// We distribute our references to declared class names evenly to the extent possible.
+// We distribute our references to declared class names evenly to the extent
+// possible.
 //
-// Before all the references are formed, the class names are kept their original unshuffled order.
-// This ensures that any uneven sampling of names is done deterministically. At the end, we
-// randomly shuffle the sequences of both the declared class names and type references
-// to provide an unpredictable order in the generated output.
+// Before all the references are formed, the class names are kept their original
+// unshuffled order. This ensures that any uneven sampling of names is done
+// deterministically. At the end, we randomly shuffle the sequences of both the
+// declared class names and type references to provide an unpredictable order in
+// the generated output.
 auto SourceGen::ClassGenState::BuildClassAndTypeNames(
     SourceGen& gen, int num_classes, int num_types,
     const TypeUseParams& type_use_params) -> void {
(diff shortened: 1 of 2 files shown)


testing/base/source_gen.cpp
```diff
@@ -22,6 +22,237 @@ auto SourceGen::Global() -> SourceGen& {
...
+  auto member_names() -> llvm::SmallVectorImpl<llvm::StringRef>& {
+    return member_names_;
+  }
+  auto param_names() -> llvm::SmallVectorImpl<llvm::StringRef>& {
+    return param_names_;
+  }
+
+  auto type_names() -> llvm::SmallVectorImpl<llvm::StringRef>& {
+    return type_names_;
+  }
```
josh11b, line 66: These accessors aren't clearly adding value -- why not just make these members public and access them directly? Particularly since this class is local to this file.

testing/base/source_gen.cpp
```diff
@@ -22,6 +22,237 @@ auto SourceGen::Global() -> SourceGen& {
...
 SourceGen::SourceGen(Language language) : language_(language) {}
 
+// Heuristic numbers used in synthesizing various identifier sequences.
+constexpr static int MinClassNameLength = 5;
+constexpr static int MinMemberNameLength = 4;
+
+// The shuffled state used to generate some number of classes.
+//
+// This state encodes all the shuffled entropy used for generating a number of
+// class definitions. While generating definitions, the state here will be
```
josh11b:
```suggestion
// This state encodes everything used to generate class definitions. The state will be
```

testing/base/source_gen.cpp
```diff
@@ -22,6 +22,237 @@ auto SourceGen::Global() -> SourceGen& {
...
+  // class names until there is some remainder of names needed.
+  int num_declared_types =
+      num_types * type_use_params.declared_types_weight / type_weight_sum;
+  for ([[maybe_unused]] auto _ : llvm::seq(num_declared_types / num_classes)) {
+    type_names_.append(class_names_.begin(), class_names_.end());
+  }
+  // Now append the remainder number of class names. This is where the class
+  // names being un-shuffled is essential. We're going to have one extra
+  // reference to some fraction of the class names and we want that to be a
+  // stable fraction.
```
josh11b: This change is to emphasize it is the names that are stable, not the number represented by the fraction.

```suggestion
  // stable subset.
```

testing/base/source_gen.cpp
```diff
@@ -22,6 +22,237 @@ auto SourceGen::Global() -> SourceGen& {
...
+// declared, but mixing in `i32` and other builtin types. A somewhat hack-y
+// approach, but currently the strategy is to map every type shorter than the
+// `MinClassNameLength` to one of our builtin types and the longer names to
+// class names in the state.
+//
+// We use non-unique random identifiers to get a sequence of lengths, and map
+// each identifier to one of our desired types as above. This allows repetition
+// as the identifiers repeat. We can also create a larger pool of builtin type
+// spellings eventually and use the different short identifier spellings to
+// select from this pool.
```
josh11b: Is this still true? It doesn't seem to match the code.

testing/base/source_gen.h
```diff
@@ -94,13 +96,78 @@ class SourceGen {
...
+  // These govern what distribution of types are used for function parameters,
+  // returns, and fields.
+  //
+  // Mainly these provide a coarse histogram of weights to shape the
+  // distribution of different type options, and try to fit that as closely as
+  // possible.
+  //
+  // We also control the percentage of functions with a return type (as opposed
+  // to void), and the percentage of declared type references that should be
+  // pointers as opposed to values.
```
josh11b: Is this accurate? I don't see it in the code. In fact I see instead a comment above:

```
  // TODO: Add heuristic for how many functions have return types.
```

testing/base/source_gen.cpp
```diff
@@ -22,6 +22,237 @@ auto SourceGen::Global() -> SourceGen& {
...
+  type_names_.append(class_names_.begin(),
+                     class_names_.begin() + (num_declared_types % num_classes));
+  CARBON_CHECK(static_cast<int>(type_names_.size()) == num_declared_types);
+
+  // Use each fixed type weight to append the expected number of copies of that
+  // type. This isn't exact however, so we'll both clamp the number of copies to
+  // not overshoot and may end up with a remainder.
+  for (const auto& fixed_type_weight : type_use_params.fixed_type_weights) {
+    // For very small numbers of types we may exhaust the fixed types. We don't
+    // try to scale the weights down for simplicity, we just cut off early.
```
josh11b:
```suggestion
    // For very small `num_types`, we may exhaust the fixed types. For
    // simplicity, we don't try to scale the weights down, we just cut off early.
```

testing/base/source_gen.cpp
```diff
@@ -22,6 +22,237 @@ auto SourceGen::Global() -> SourceGen& {
...
+    if (static_cast<int>(type_names_.size()) == num_types) {
+      break;
+    }
```
josh11b: Why have this code? It seems like the `std::min` below would handle that case fine.

testing/base/source_gen.cpp
```diff
@@ -22,6 +22,237 @@ auto SourceGen::Global() -> SourceGen& {
...
+    int num_fixed_type =
+        std::min<int>(num_types - type_names_.size(),
```
josh11b: Is this `std::min` needed in practice? It seems like with all the rounding down that it wouldn't be needed.

testing/base/source_gen.cpp
```diff
@@ -22,6 +22,237 @@ auto SourceGen::Global() -> SourceGen& {
...
+       llvm::seq(1 + (num_types - type_names_.size()) /
+                         type_use_params.fixed_type_weights.size())) {
+    for (const auto& fixed_type_weight :
+         llvm::ArrayRef(type_use_params.fixed_type_weights)
+             .take_front(num_types - type_names_.size())) {
+      type_names_.push_back(gen.IsCpp() ? fixed_type_weight.cpp_spelling
+                                        : fixed_type_weight.carbon_spelling);
+    }
+  }
+  CARBON_CHECK(static_cast<int>(type_names_.size()) == num_types);
```
josh11b: This code seems like it might be more complex and fragile than necessary?
```suggestion
  while (static_cast<int>(type_names_.size()) < num_types) {
    for (const auto& fixed_type_weight :
         llvm::ArrayRef(type_use_params.fixed_type_weights)
             .take_front(num_types - type_names_.size())) {
      type_names_.push_back(gen.IsCpp() ? fixed_type_weight.cpp_spelling
                                        : fixed_type_weight.carbon_spelling);
    }
  }
  CARBON_CHECK(static_cast<int>(type_names_.size()) == num_types);
```

testing/base/source_gen.cpp
```diff
@@ -638,27 +845,44 @@ auto SourceGen::GenerateFunctionDecl(
...
-  os << "class " << state.class_names.pop_back_val() << " {\n";
+  os << "class " << name << " {\n";
   if (IsCpp()) {
     os << " public:\n";
   }
 
-  UniqueIdentifierPopper unique_member_names(*this, state.member_names);
+  // Collect field types before any others as those can't be the class we're
+  // currently declaring. We do this before inserting that type into the valid
+  // set.
```
josh11b:
```suggestion
  // Field types can't be the class we're currently declaring. We enforce this
  // by collecting them before inserting that type into the valid set.
```

testing/base/source_gen.cpp
```diff
@@ -638,27 +845,44 @@ auto SourceGen::GenerateFunctionDecl(
...
+  llvm::SmallVector<llvm::StringRef> field_type_names;
+  field_type_names.reserve(params.private_field_decls);
+  for ([[maybe_unused]] int _ : llvm::seq(params.private_field_decls)) {
+    field_type_names.push_back(state.GetValidTypeName());
+  }
+
+  // Mark this class as now a valid type now that field type names have been
+  // collected. We can self-reference from other types within the definition.
```
josh11b:
```suggestion
  // collected. We can reference this class from functions and methods within the definition.
```

testing/base/source_gen.cpp
```diff
@@ -669,24 +893,28 @@ auto SourceGen::GenerateClassDef(const ClassParams& params,
...
+    GenerateFunctionDecl(
+        unique_member_names.Pop(), /*is_private=*/true,
+        /*is_method=*/true, state.private_method_param_counts().pop_back_val(),
+        /*indent=*/"  ", state.param_names(),
+        [&] { return state.GetValidTypeName(); }, os);
   }
   os << line_sep;
-  for ([[maybe_unused]] int _ : llvm::seq(params.private_field_decls)) {
+  for (llvm::StringRef type_name : field_type_names) {
     if (!IsCpp()) {
```
josh11b, line 904: Perhaps swap this around so you don't have a negated condition?

testing/base/source_gen.h
```diff
@@ -163,8 +230,9 @@ class SourceGen {
   // `GetShuffledIdentifiers`.
   //
   // Usually, benchmarks should use the shuffled version. However, this is
-  // useful when there is already a post-processing step to shuffle things as it
-  // is *dramatically* more efficient, especially in debug builds.
+  // useful when the identifiers will be used in multiple different ways and
+  // only then shuffled, or if there is already a post-processing step to
+  // shuffle things which is very expensive in debug builds.
```
josh11b:
```suggestion
  // useful when deterministic access to the identifiers is needed to avoid
  // introducing noise, or if there is already a post-processing step to shuffle
  // things, since shuffling is very expensive in debug builds.
```

testing/base/source_gen.h
```diff
@@ -173,7 +241,9 @@ class SourceGen {
   // as `GetShuffledUniqueIdentifiers`.
   //
   // Usually, benchmarks should use the shuffled version. However, this is
-  // useful when there is already a post-processing step to shuffle things.
+  // useful when the identifiers will be used in multiple different ways and
+  // only then shuffled, or if there is already a post-processing step to
+  // shuffle things which is very expensive in debug builds.
```
josh11b:
```suggestion
```suggestion
  // useful when deterministic access to the identifiers is needed to avoid
  // introducing noise, or if there is already a post-processing step to shuffle
  // things, since shuffling is very expensive in debug builds.
``````

File: testing/base/source_gen.cpp (hunk 1 above)
Comment (chandlerc) [reply testing/base/source_gen.cpp:66]: Because my understanding of the style guide precludes that. If I'm wrong, I'd be delighted for these to be public. @jonmeow

File: testing/base/source_gen.cpp (hunk 4 above)
Comment (chandlerc) [reply testing/base/source_gen.cpp]: Doh, sorry, forgot to update this comment after that approach didn't work out and I pursued a better approach.

File: testing/base/source_gen.h (hunk 5 above)
Comment (chandlerc) [reply testing/base/source_gen.h]: Just forgot to remove this when adding the TODO, sorry.

File: testing/base/source_gen.cpp [hunk 15]
```diff
@@ -22,6 +22,234 @@ auto SourceGen::Global() -> SourceGen& {
...
 SourceGen::SourceGen(Language language) : language_(language) {}
 
+// Heuristic numbers used in synthesizing various identifier sequences.
+constexpr static int MinClassNameLength = 5;
+constexpr static int MinMemberNameLength = 4;
+
+// The shuffled state used to generate some number of classes.
+//
+// This state encodes everything used to generate class definitions. The state will be
+// consumed until empty.
```
CarbonInfraBot, line 32: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
// This state encodes everything used to generate class definitions. The state
// will be consumed until empty.
```

testing/base/source_gen.cpp
```diff
@@ -22,6 +22,234 @@
...
+  type_names_.append(class_names_.begin(),
+                     class_names_.begin() + (num_declared_types % num_classes));
+  CARBON_CHECK(static_cast<int>(type_names_.size()) == num_declared_types);
+
+  // Use each fixed type weight to append the expected number of copies of that
+  // type. This isn't exact however, so we'll both clamp the number of copies to
+  // not overshoot and may end up with a remainder.
+  for (const auto& fixed_type_weight : type_use_params.fixed_type_weights) {
+    // For very small `num_types`, we may exhaust the fixed types. For
+    // simplicity, we don't try to scale the weights down, we just cut off early.
```
CarbonInfraBot, line 221: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
    // simplicity, we don't try to scale the weights down, we just cut off
    // early.
```

testing/base/source_gen.cpp
```diff
@@ -638,27 +842,43 @@
...
+  // Field types can't be the class we're currently declaring. We enforce this
+  // by collecting them before inserting that type into the valid set.
+  llvm::SmallVector<llvm::StringRef> field_type_names;
+  field_type_names.reserve(params.private_field_decls);
+  for ([[maybe_unused]] int _ : llvm::seq(params.private_field_decls)) {
+    field_type_names.push_back(state.GetValidTypeName());
+  }
+
+  // Mark this class as now a valid type now that field type names have been
+  // collected. We can reference this class from functions and methods within the definition.
```
CarbonInfraBot, line 861: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  // collected. We can reference this class from functions and methods within
  // the definition.
```

testing/base/source_gen.h
```diff
@@ -173,7 +241,10 @@ class SourceGen {
...
   //
   // Usually, benchmarks should use the shuffled version. However, this is
-  // useful when there is already a post-processing step to shuffle things.
+```suggestion
+  // useful when deterministic access to the identifiers is needed to avoid
+  // introducing noise, or if there is already a post-processing step to shuffle
+  // things, since shuffling is very expensive in debug builds.
   auto GetUniqueIdentifiers(int number, int min_length = 1, int max_length = 64,
                             bool uniform = false)
       -> llvm::SmallVector<llvm::StringRef>;
```
CarbonInfraBot, line 250: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  ```suggestion
      // useful when deterministic access to the identifiers is needed to avoid
      // introducing noise, or if there is already a post-processing step to
      // shuffle things, since shuffling is very expensive in debug builds.
      auto
      GetUniqueIdentifiers(int number, int min_length = 1, int max_length = 64,
                           bool uniform = false)
          -> llvm::SmallVector<llvm::StringRef>;
```

testing/base/source_gen.cpp
chandlerc: I could swear this was needed, but I can't reproduce it (even without the min) so removing.

testing/base/source_gen.cpp
chandlerc: This makes sense to me, but somehow I needed it in the past.

But now, it seems to work fine without this or the if above? Great I guess!

testing/base/source_gen.cpp
chandlerc: There are a bunch of negative conditions and positive ones mixed....

Can I address this one way or the other in a follow-up PR? Can chat to see which direction works best.

chandlerc (commented):
Thanks, PTAL!

testing/base/source_gen.cpp
josh11b: Fine for a follow up. I noticed it in this PR, but it is a pre-existing condition.

testing/base/source_gen.cpp
```diff
@@ -161,24 +160,25 @@ auto SourceGen::ClassGenState::GetValidTypeName() -> llvm::StringRef {
...
+// (un-shuffled) class names that will be defined to form the spelling of all
+// the referenced types. The `type_use_params` provides weights for each fixed
+// type as well as an overall weight for referencing class names that are being
+// declared. We build a set of type references so that its histogram will
+// roughly match these weights.
 //
-// When searching for a given length, we allow wrapping around the class names.
-// This may map several random identifiers to the same declared class name.
+// For the fixed types, we assume each type has a spelling that will be valid
+// provided by params for both Carbon and C++.
```
josh11b: Not sure what you are trying to say with this paragraph. Perhaps this?
```suggestion
// For each of the fixed types, `type_use_params` provides a spelling for both Carbon and C++.
```

testing/base/source_gen.cpp
```diff
@@ -161,24 +160,25 @@ auto SourceGen::ClassGenState::GetValidTypeName() -> llvm::StringRef {
...
-// class names in the state.
+// declared, but mixing in `i32` and other builtin types.
 //
-// We use non-unique random identifiers to get a sequence of lengths, and map
-// each identifier to one of our desired types as above. This allows repetition
-// as the identifiers repeat. We can also create a larger pool of builtin type
-// spellings eventually and use the different short identifier spellings to
-// select from this pool.
+// We combine a list of fixed types in the `type_use_params` with the list of
+// (un-shuffled) class names that will be defined to form the spelling of all
```
josh11b: The references to shuffling are a bit spread out -- I think it would be clearer to explain the strategy in one place.
```suggestion
// class names that will be defined to form the spelling of all
```

testing/base/source_gen.cpp
```diff
@@ -161,24 +160,25 @@ auto SourceGen::ClassGenState::GetValidTypeName() -> llvm::StringRef {
...
-// This may map several random identifiers to the same declared class name.
+// For the fixed types, we assume each type has a spelling that will be valid
+// provided by params for both Carbon and C++.
 //
-// TODO: Variation in how many times a name is repeated may introduce some
-// unnecessary noise, we should consider a fixed distribution of duplication
-// rather than just accepting whatever occurs randomly. Currently this only
-// stabilizes the total text size resulting from referencing prior type names.
+// For referencing declared class names, we use the un-shuffled sequence, and
+// evenly distribute our references across it to the extent possible.
```
josh11b:
```suggestion
// We distribute our references to declared class names evenly to the extent possible.
```

testing/base/source_gen.cpp
```diff
@@ -161,24 +160,25 @@ auto SourceGen::ClassGenState::GetValidTypeName() -> llvm::StringRef {
 }
 
 // We want to synthesize shuffled type names from the class names that will be
```
josh11b: This also generates the shuffled class names themselves. Perhaps include that too? Maybe also make it clear in this intro that this other list of shuffled type names will be used for type references?

testing/base/source_gen.cpp
```diff
@@ -161,24 +160,25 @@ auto SourceGen::ClassGenState::GetValidTypeName() -> llvm::StringRef {
...
-// unnecessary noise, we should consider a fixed distribution of duplication
-// rather than just accepting whatever occurs randomly. Currently this only
-// stabilizes the total text size resulting from referencing prior type names.
+// For referencing declared class names, we use the un-shuffled sequence, and
+// evenly distribute our references across it to the extent possible.
+//
+// Once all of the references are formed, we randomly shuffle the entire
+// sequence to provide an unpredictable order of reference, and also shuffle the
+// declared class names now that we're not sampling from that list to build the
+// references.
```
josh11b:
```suggestion
// Before all the references are formed, the class names are kept their original unshuffled order.
// This ensures that any uneven sampling of names is done deterministically. At the end, we
// randomly shuffle the sequences of both the declared class names and type references
// to provide an unpredictable order in the generated output.
```

testing/base/source_gen.cpp
```diff
@@ -22,6 +22,225 @@ auto SourceGen::Global() -> SourceGen& {
...
+// declared, but mixing in `i32` and other builtin types.
+//
+// We combine a list of fixed types in the `type_use_params` with the list of
+// class names that will be defined to form the spelling of all
+// the referenced types. The `type_use_params` provides weights for each fixed
+// type as well as an overall weight for referencing class names that are being
+// declared. We build a set of type references so that its histogram will
+// roughly match these weights.
+//
+// For each of the fixed types, `type_use_params` provides a spelling for both Carbon and C++.
```
CarbonInfraBot, line 172: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
// For each of the fixed types, `type_use_params` provides a spelling for both
// Carbon and C++.
```

testing/base/source_gen.cpp
```diff
@@ -22,6 +22,225 @@
...
+//
+// We distribute our references to declared class names evenly to the extent possible.
```
CarbonInfraBot, line 174: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
// We distribute our references to declared class names evenly to the extent
// possible.
```

testing/base/source_gen.cpp
```diff
@@ -22,6 +22,225 @@
...
+//
+// Before all the references are formed, the class names are kept their original unshuffled order.
+// This ensures that any uneven sampling of names is done deterministically. At the end, we
+// randomly shuffle the sequences of both the declared class names and type references
+// to provide an unpredictable order in the generated output.
```
CarbonInfraBot, line 179: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
// Before all the references are formed, the class names are kept their original
// unshuffled order. This ensures that any uneven sampling of names is done
// deterministically. At the end, we randomly shuffle the sequences of both the
// declared class names and type references to provide an unpredictable order in
// the generated output.
```

testing/base/source_gen.cpp
chandlerc: Yeah, tried a better initial sentence.

testing/base/source_gen.cpp
```diff
@@ -159,26 +159,27 @@ auto SourceGen::ClassGenState::GetValidTypeName() -> llvm::StringRef {
...
 
-// We want to synthesize shuffled type names from the class names that will be
-// declared, but mixing in `i32` and other builtin types.
+// Build both the class names this file will declare and a list of type
+// references to use throughout those classes.
 //
 // We combine a list of fixed types in the `type_use_params` with the list of
-// (un-shuffled) class names that will be defined to form the spelling of all
+// class names that will be defined to form the spelling of all
 // the referenced types. The `type_use_params` provides weights for each fixed
```
josh11b: Re-flow the text please
```suggestion
// class names that will be defined to form the spelling of all the referenced types. The `type_use_params` provides weights for each fixed
```

josh11b (approved):

testing/base/source_gen.cpp
```diff
@@ -22,6 +22,227 @@ auto SourceGen::Global() -> SourceGen& {
...
+}
+
+// Build both the class names this file will declare and a list of type
+// references to use throughout those classes.
+//
+// We combine a list of fixed types in the `type_use_params` with the list of
+// class names that will be defined to form the spelling of all the referenced types. The `type_use_params` provides weights for each fixed
+// type as well as an overall weight for referencing class names that are being
+// declared. We build a set of type references so that its histogram will
+// roughly match these weights.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
// class names that will be defined to form the spelling of all the referenced
// types. The `type_use_params` provides weights for each fixed type as well as
// an overall weight for referencing class names that are being declared. We
// build a set of type references so that its histogram will roughly match these
// weights.
```

chandlerc:
Thanks, merging! Will follow up on the accessors vs. public members and fix forward if needed.

PR #4245, opened 2024-08-23
Collection of minor tweaks to get approx. 10-15% compile time

chandlerc:
Most of these are about enabling inlining, in a couple of cases moving code to a header and throughout switching to `CARBON_DCHECK`. The code size of `CARBON_CHECK` seems to make inliing quite unreliable. I'm going to think about whether there are ways to improve this, but a reasonably small number of these seem worth switching for now to get some compile time savings.

Also moves VLOG out of the hot path which helps a bit as well.

All combined, this net a bit over 10%, although it varies a bit exactly how much. We're now pretty consistently over 800k lines/second for check in the compilation benchmark for files >=4k lines, which makes me happy. That's remarkably close to our original target.

Not really planning to keep optimizing here, just was glancing at the profile and many of these stood out to me and were easy to fix.

### DIFF
--- a/common/vlog.h
+++ b/common/vlog.h
@@ -13,9 +13,10 @@ namespace Carbon {
 //
 // For example:
 //   CARBON_VLOG() << "Verbose message";
-#define CARBON_VLOG()                 \
-  (vlog_stream_ == nullptr) ? (void)0 \
-                            : CARBON_VLOG_INTERNAL_STREAM(vlog_stream_)
+#define CARBON_VLOG()                             \
+  __builtin_expect(vlog_stream_ == nullptr, true) \
+      ? (void)0                                   \
+      : CARBON_VLOG_INTERNAL_STREAM(vlog_stream_)
 
 }  // namespace Carbon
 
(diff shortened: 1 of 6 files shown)


common/vlog.h
```diff
@@ -13,9 +13,10 @@ namespace Carbon {
 //
 // For example:
 //   CARBON_VLOG() << "Verbose message";
-#define CARBON_VLOG()                 \
-  (vlog_stream_ == nullptr) ? (void)0 \
-                            : CARBON_VLOG_INTERNAL_STREAM(vlog_stream_)
+#define CARBON_VLOG()                             \
+  __builtin_expect(vlog_stream_ == nullptr, true) \
```
jonmeow, line 17: Should CARBON_CHECK have a similar `__builtin_expect((__VA_ARGS__), false)` ?

jonmeow (approved):

common/vlog.h
chandlerc: The `noreturn` provides some help there, but yeah, we can probably do more. Will send another PR.

PR #4248, opened 2024-08-26
Enforce `private` and `protected` access modifiers for class member access

brymer-meneses:
Print diagnostics for invalid class member access. This doesn't take into account compound member access.

### DIFF
--- a/toolchain/check/context.cpp
+++ b/toolchain/check/context.cpp
@@ -371,7 +371,7 @@ static auto DiagnoseInvalidQualifiedNameAccess(Context& context, SemIRLoc loc,
   // TODO: Support passing AccessKind to diagnostics.
   CARBON_DIAGNOSTIC(ClassInvalidMemberAccess, Error,
                     "Cannot access {0} member `{1}` of type `{2}`.",
-                    std::string_view, SemIR::NameId, SemIR::TypeId);
+                    llvm::StringLiteral, SemIR::NameId, SemIR::TypeId);
   CARBON_DIAGNOSTIC(ClassMemberDefinition, Note,
                     "The {0} member `{1}` is defined here.",
                     llvm::StringLiteral, SemIR::NameId);
--- a/toolchain/check/context.cpp
+++ b/toolchain/check/context.cpp
@@ -4,6 +4,7 @@
 
 #include "toolchain/check/context.h"
 
+#include <optional>
 #include <string>
 #include <utility>
 
@@ -292,7 +293,8 @@ auto Context::LookupNameInDecl(SemIR::LocId loc_id, SemIR::NameId name_id,
     //    // Error, no `F` in `B`.
     //    fn B.F() {}
     return LookupNameInExactScope(loc_id, name_id, scope_id,
-                                  name_scopes().Get(scope_id));
+                                  name_scopes().Get(scope_id))
+        .first;
   }
 }
 
@@ -334,26 +336,115 @@ auto Context::LookupUnqualifiedName(Parse::NodeId node_id,
 auto Context::LookupNameInExactScope(SemIRLoc loc, SemIR::NameId name_id,
                                      SemIR::NameScopeId scope_id,
                                      const SemIR::NameScope& scope)
-    -> SemIR::InstId {
+    -> std::pair<SemIR::InstId, SemIR::AccessKind> {
   if (auto lookup = scope.name_map.Lookup(name_id)) {
-    auto inst_id = scope.names[lookup.value()].inst_id;
-    LoadImportRef(*this, inst_id);
-    return inst_id;
+    auto entry = scope.names[lookup.value()];
+    LoadImportRef(*this, entry.inst_id);
+    return {entry.inst_id, entry.access_kind};
   }
+
   if (!scope.import_ir_scopes.empty()) {
-    return ImportNameFromOtherPackage(*this, loc, scope_id,
-                                      scope.import_ir_scopes, name_id);
+    // TODO: Enforce other access modifiers for imports.
+    return {ImportNameFromOtherPackage(*this, loc, scope_id,
+                                       scope.import_ir_scopes, name_id),
+            SemIR::AccessKind::Public};
   }
-  return SemIR::InstId::Invalid;
+  return {SemIR::InstId::Invalid, SemIR::AccessKind::Public};
 }
 
+// Prints diagnostics on invalid qualified name access.
+static auto DiagnoseInvalidQualifiedNameAccess(Context& context, SemIRLoc loc,
+                                               SemIR::InstId scope_result_id,
+                                               SemIR::NameId name_id,
+                                               SemIR::AccessKind access_kind,
+                                               bool is_parent_access,
+                                               AccessInfo access_info) -> void {
+  auto class_type = context.insts().TryGetAs<SemIR::ClassType>(
+      context.constant_values().GetInstId(access_info.constant_id));
+  if (!class_type) {
+    return;
+  }
+
+  // TODO: Support scoped entities other than just classes.
+  auto class_info = context.classes().Get(class_type->class_id);
+
+  // TODO: Support passing AccessKind to diagnostics.
+  CARBON_DIAGNOSTIC(ClassInvalidMemberAccess, Error,
+                    "Cannot access {0} member `{1}` of type `{2}`.",
+                    std::string_view, SemIR::NameId, SemIR::TypeId);
+  CARBON_DIAGNOSTIC(ClassMemberDefinition, Note,
+                    "The {0} member `{1}` is defined here.",
+                    llvm::StringLiteral, SemIR::NameId);
+
+  auto parent_type_id = class_info.self_type_id;
+  auto access_desc = access_kind == SemIR::AccessKind::Private
+                         ? llvm::StringLiteral("private")
+                         : llvm::StringLiteral("protected");
+
+  if (access_kind == SemIR::AccessKind::Private && is_parent_access) {
+    if (auto base_decl = context.insts().TryGetAsIfValid<SemIR::BaseDecl>(
+            class_info.base_id)) {
+      parent_type_id = base_decl->base_type_id;
+    } else if (auto adapt_decl =
+                   context.insts().TryGetAsIfValid<SemIR::AdaptDecl>(
+                       class_info.adapt_id)) {
+      parent_type_id = adapt_decl->adapted_type_id;
+    } else {
+      CARBON_FATAL() << "Expected parent for parent access";
+    }
+  }
+
+  context.emitter()
+      .Build(loc, ClassInvalidMemberAccess, access_desc, name_id,
+             parent_type_id)
+      .Note(scope_result_id, ClassMemberDefinition, access_desc, name_id)
+      .Emit();
+}
+
+// Returns whether the access is prohibited by the access modifiers.
+static auto IsAccessProhibited(std::optional<AccessInfo> access_info,
+                               SemIR::AccessKind access_kind,
+                               bool is_parent_access) -> bool {
+  if (!access_info) {
+    return false;
+  }
+
+  switch (access_kind) {
+    case SemIR::AccessKind::Public:
+      return false;
+    case SemIR::AccessKind::Protected:
+      return access_info->highest_allowed_access == SemIR::AccessKind::Public;
+    case SemIR::AccessKind::Private:
+      return access_info->highest_allowed_access !=
+                 SemIR::AccessKind::Private ||
+             is_parent_access;
+  }
+}
+
+// Information regarding a prohibited access.
+struct ProhibitedAccessInfo {
+  // The resulting inst of the lookup.
+  SemIR::InstId scope_result_id;
+  // The access kind of the lookup.
+  SemIR::AccessKind access_kind;
+  // If the lookup is from an extended scope. For example, if this is a base
+  // class member access from a class that extends it.
+  bool is_parent_access;
+};
+
 auto Context::LookupQualifiedName(SemIRLoc loc, SemIR::NameId name_id,
-                                  LookupScope scope, bool required)
+                                  LookupScope scope, bool required,
+                                  std::optional<AccessInfo> access_info)
     -> LookupResult {
   llvm::SmallVector<LookupScope> scopes = {scope};
+
+  // TODO: Support reporting of multiple prohibited access.
+  llvm::SmallVector<ProhibitedAccessInfo> prohibited_accesses;
+
   LookupResult result = {.specific_id = SemIR::SpecificId::Invalid,
                          .inst_id = SemIR::InstId::Invalid};
   bool has_error = false;
+  bool is_parent_access = false;
 
   // Walk this scope and, if nothing is found here, the scopes it extends.
   while (!scopes.empty()) {
@@ -361,10 +452,25 @@ auto Context::LookupQualifiedName(SemIRLoc loc, SemIR::NameId name_id,
     const auto& name_scope = name_scopes().Get(scope_id);
     has_error |= name_scope.has_error;
 
-    auto scope_result_id =
+    auto [scope_result_id, access_kind] =
         LookupNameInExactScope(loc, name_id, scope_id, name_scope);
-    if (!scope_result_id.is_valid()) {
-      // Nothing found in this scope: also look in its extended scopes.
+
+    auto is_access_prohibited =
+        IsAccessProhibited(access_info, access_kind, is_parent_access);
+
+    // Keep track of prohibited accesses, this will be useful for reporting
+    // multiple prohibited accesses if we can't find a suitable lookup.
+    if (is_access_prohibited) {
+      prohibited_accesses.push_back({
+          .scope_result_id = scope_result_id,
+          .access_kind = access_kind,
+          .is_parent_access = is_parent_access,
+      });
+    }
+
+    if (!scope_result_id.is_valid() || is_access_prohibited) {
+      // If nothing is found in this scope or if we encountered an invalid
+      // access, look in its extended scopes.
       auto extended = name_scope.extended_scopes;
       scopes.reserve(scopes.size() + extended.size());
       for (auto extended_id : llvm::reverse(extended)) {
@@ -373,6 +479,7 @@ auto Context::LookupQualifiedName(SemIRLoc loc, SemIR::NameId name_id,
         scopes.push_back({.name_scope_id = extended_id,
                           .specific_id = SemIR::SpecificId::Invalid});
       }
+      is_parent_access |= !extended.empty();
       continue;
     }
 
@@ -397,8 +504,22 @@ auto Context::LookupQualifiedName(SemIRLoc loc, SemIR::NameId name_id,
 
   if (required && !result.inst_id.is_valid()) {
     if (!has_error) {
-      DiagnoseNameNotFound(loc, name_id);
+      if (prohibited_accesses.empty()) {
+        DiagnoseNameNotFound(loc, name_id);
+      } else {
+        //  TODO: We should report multiple prohibited accesses in case we don't
+        //  find a valid lookup. Reporting the last one should suffice for now.
+        auto [scope_result_id, access_kind, is_parent_access] =
+            prohibited_accesses.back();
+
+        // Note, `access_info` is guaranteed to have a value here, since
+        // `prohibited_accesses` is non-empty.
+        DiagnoseInvalidQualifiedNameAccess(*this, loc, scope_result_id, name_id,
+                                           access_kind, is_parent_access,
+                                           *access_info);
+      }
     }
+
     return {.specific_id = SemIR::SpecificId::Invalid,
             .inst_id = SemIR::InstId::BuiltinError};
   }
@@ -420,7 +541,7 @@ static auto GetCorePackage(Context& context, SemIRLoc loc)
   auto core_name_id = SemIR::NameId::ForIdentifier(core_ident_id);
 
   // Look up `package.Core`.
-  auto core_inst_id = context.LookupNameInExactScope(
+  auto [core_inst_id, _] = context.LookupNameInExactScope(
       loc, core_name_id, SemIR::NameScopeId::Package,
       context.name_scopes().Get(SemIR::NameScopeId::Package));
   if (core_inst_id.is_valid()) {
@@ -447,8 +568,8 @@ auto Context::LookupNameInCore(SemIRLoc loc, llvm::StringRef name)
   }
 
   auto name_id = SemIR::NameId::ForIdentifier(identifiers().Add(name));
-  auto inst_id = LookupNameInExactScope(loc, name_id, core_package_id,
-                                        name_scopes().Get(core_package_id));
+  auto [inst_id, _] = LookupNameInExactScope(
+      loc, name_id, core_package_id, name_scopes().Get(core_package_id));
   if (!inst_id.is_valid()) {
     CARBON_DIAGNOSTIC(
         CoreNameNotFound, Error,
--- a/toolchain/check/context.h
+++ b/toolchain/check/context.h
@@ -24,6 +24,7 @@
 #include "toolchain/sem_ir/ids.h"
 #include "toolchain/sem_ir/import_ir.h"
 #include "toolchain/sem_ir/inst.h"
+#include "toolchain/sem_ir/name_scope.h"
 #include "toolchain/sem_ir/typed_insts.h"
 
 namespace Carbon::Check {
@@ -44,6 +45,16 @@ struct LookupResult {
   SemIR::SpecificId specific_id;
   // The declaration that was found by name lookup.
   SemIR::InstId inst_id;
+};
+
+// Information about an access.
+struct AccessInfo {
+  // The constant being accessed.
+  SemIR::ConstantId constant_id;
+
+  // The highest allowed access for a lookup. For example, `Protected` allows
+  // access to `Public` and `Protected` names, but not `Private`.
+  SemIR::AccessKind highest_allowed_access;
 };
 
 // Context and shared functionality for semantics handlers.
@@ -173,12 +184,14 @@ class Context {
   // instruction if the name is not found.
   auto LookupNameInExactScope(SemIRLoc loc, SemIR::NameId name_id,
                               SemIR::NameScopeId scope_id,
-                              const SemIR::NameScope& scope) -> SemIR::InstId;
+                              const SemIR::NameScope& scope)
+      -> std::pair<SemIR::InstId, SemIR::AccessKind>;
 
   // Performs a qualified name lookup in a specified scope and in scopes that
   // it extends, returning the referenced instruction.
   auto LookupQualifiedName(SemIRLoc loc, SemIR::NameId name_id,
-                           LookupScope scope, bool required = true)
+                           LookupScope scope, bool required = true,
+                           std::optional<AccessInfo> access_info = std::nullopt)
       -> LookupResult;
 
   // Returns the instruction corresponding to a name in the core package, or
--- a/toolchain/check/impl.cpp
+++ b/toolchain/check/impl.cpp
@@ -163,7 +163,7 @@ static auto BuildInterfaceWitness(
           CARBON_FATAL() << "Unexpected type: " << type_inst;
         }
         auto& fn = context.functions().Get(fn_type->function_id);
-        auto impl_decl_id = context.LookupNameInExactScope(
+        auto [impl_decl_id, _] = context.LookupNameInExactScope(
             decl_id, fn.name_id, impl.scope_id, impl_scope);
         if (impl_decl_id.is_valid()) {
           used_decl_ids.push_back(impl_decl_id);
--- a/toolchain/check/member_access.cpp
+++ b/toolchain/check/member_access.cpp
@@ -4,6 +4,8 @@
 
 #include "toolchain/check/member_access.h"
 
+#include <optional>
+
 #include "llvm/ADT/STLExtras.h"
 #include "toolchain/base/kind_switch.h"
 #include "toolchain/check/context.h"
@@ -13,6 +15,7 @@
 #include "toolchain/sem_ir/generic.h"
 #include "toolchain/sem_ir/ids.h"
 #include "toolchain/sem_ir/inst.h"
+#include "toolchain/sem_ir/name_scope.h"
 #include "toolchain/sem_ir/typed_insts.h"
 
 namespace Carbon::Check {
@@ -96,6 +99,82 @@ static auto IsInstanceMethod(const SemIR::File& sem_ir,
   }
 
   return false;
+}
+
+// Returns the FunctionId of the current function if it exists.
+static auto GetCurrentFunction(Context& context)
+    -> std::optional<SemIR::FunctionId> {
+  if (context.return_scope_stack().empty()) {
+    return std::nullopt;
+  }
+
+  return context.insts()
+      .GetAs<SemIR::FunctionDecl>(context.return_scope_stack().back().decl_id)
+      .function_id;
+}
+
+// Returns the highest allowed access. For example, if this returns `Protected`
+// then only `Public` and `Protected` accesses are allowed--not `Private`.
+static auto GetHighestAllowedAccess(Context& context, SemIRLoc loc,
+                                    SemIR::ConstantId name_scope_const_id)
+    -> SemIR::AccessKind {
+  // TODO: Maybe use LookupUnqualifiedName for `Self` to support things like
+  // `var x: Self.ParentProtectedType`?
+  auto current_function = GetCurrentFunction(context);
+  // If `current_function` is a `nullopt` then we're accessing from a global
+  // variable.
+  if (!current_function) {
+    return SemIR::AccessKind::Public;
+  }
+
+  auto scope_id = context.functions().Get(*current_function).parent_scope_id;
+  if (!scope_id.is_valid()) {
+    return SemIR::AccessKind::Public;
+  }
+  auto scope = context.name_scopes().Get(scope_id);
+
+  // Lookup the inst for `Self` in the parent scope of the current function.
+  auto [self_type_inst_id, _] = context.LookupNameInExactScope(
+      loc, SemIR::NameId::SelfType, scope_id, scope);
+  if (!self_type_inst_id.is_valid()) {
+    return SemIR::AccessKind::Public;
+  }
+
+  // TODO: Support other types for `Self`.
+  auto self_class_type =
+      context.insts().TryGetAs<SemIR::ClassType>(self_type_inst_id);
+  if (!self_class_type) {
+    return SemIR::AccessKind::Public;
+  }
+
+  auto self_class_info = context.classes().Get(self_class_type->class_id);
+
+  // TODO: Support other types.
+  if (auto class_type = context.insts().TryGetAs<SemIR::ClassType>(
+          context.constant_values().GetInstId(name_scope_const_id))) {
+    auto class_info = context.classes().Get(class_type->class_id);
+
+    if (self_class_info.self_type_id == class_info.self_type_id) {
+      return SemIR::AccessKind::Private;
+    }
+
+    // If the type_id of `Self` does not match with the one we're currently
+    // accessing. Try checking if this class is of the parent type of `Self`.
+    if (auto base_decl = context.insts().TryGetAsIfValid<SemIR::BaseDecl>(
+            self_class_info.base_id)) {
+      if (base_decl->base_type_id == class_info.self_type_id) {
+        return SemIR::AccessKind::Protected;
+      }
+    } else if (auto adapt_decl =
+                   context.insts().TryGetAsIfValid<SemIR::AdaptDecl>(
+                       self_class_info.adapt_id)) {
+      if (adapt_decl->adapted_type_id == class_info.self_type_id) {
+        return SemIR::AccessKind::Protected;
+      }
+    }
+  }
+
+  return SemIR::AccessKind::Public;
 }
 
 // Returns whether `scope` is a scope for which impl lookup should be performed
@@ -224,7 +303,18 @@ static auto LookupMemberNameInScope(Context& context, SemIR::LocId loc_id,
   LookupResult result = {.specific_id = SemIR::SpecificId::Invalid,
                          .inst_id = SemIR::InstId::BuiltinError};
   if (lookup_scope.name_scope_id.is_valid()) {
-    result = context.LookupQualifiedName(loc_id, name_id, lookup_scope);
+    AccessInfo access_info = {
+        .constant_id = name_scope_const_id,
+        .highest_allowed_access =
+            GetHighestAllowedAccess(context, loc_id, name_scope_const_id),
+    };
+
+    result = context.LookupQualifiedName(loc_id, name_id, lookup_scope,
+                                         /*required=*/true, access_info);
+
+    if (!result.inst_id.is_valid()) {
+      return SemIR::InstId::BuiltinError;
+    }
   }
 
   // TODO: This duplicates the work that HandleNameAsExpr does. Factor this out.
--- a/toolchain/check/member_access.cpp
+++ b/toolchain/check/member_access.cpp
@@ -158,8 +158,8 @@ static auto GetHighestAllowedAccess(Context& context, SemIRLoc loc,
       return SemIR::AccessKind::Private;
     }
 
-    // If the type_id of `Self` does not match with the one we're currently
-    // accessing. Try checking if this class is of the parent type of `Self`.
+    // If the `type_id` of `Self` does not match with the one we're currently
+    // accessing, try checking if this class is of the parent type of `Self`.
     if (auto base_decl = context.insts().TryGetAsIfValid<SemIR::BaseDecl>(
             self_class_info.base_id)) {
       if (base_decl->base_type_id == class_info.self_type_id) {
--- a/toolchain/check/testdata/class/access_modifers.carbon
+++ b/toolchain/check/testdata/class/access_modifers.carbon
@@ -1,0 +1,648 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+// TIP: To test this file alone, run:
+// TIP:   bazel test //toolchain/testing:file_test --test_arg=--file_tests=toolchain/check/testdata/class/access_modifers.carbon
+// TIP: To dump output, run:
+// TIP:   bazel run //toolchain/testing:file_test -- --dump_output --file_tests=toolchain/check/testdata/class/access_modifers.carbon
+
+// --- fail_private_field_access.carbon
+
+library "private_field_access";
+
+class Circle {
+  private var radius: i32;
+  private let SOME_INTERNAL_CONSTANT: i32 = 5;
+
+  private fn SomeInternalFunction() -> i32 {
+    return 0;
+  }
+
+  fn Make() -> Self {
+    return {.radius = 5};
+  }
+}
+
+fn Run() {
+  let circle: Circle = Circle.Make();
+  // CHECK:STDERR: fail_private_field_access.carbon:[[@LINE+7]]:21: ERROR: Cannot access private member `radius` of type `Circle`.
+  // CHECK:STDERR:   let radius: i32 = circle.radius;
+  // CHECK:STDERR:                     ^~~~~~~~~~~~~
+  // CHECK:STDERR: fail_private_field_access.carbon:[[@LINE-17]]:15: The private member `radius` is defined here.
+  // CHECK:STDERR:   private var radius: i32;
+  // CHECK:STDERR:               ^~~~~~~~~~~
+  // CHECK:STDERR:
+  let radius: i32 = circle.radius;
+  // CHECK:STDERR: fail_private_field_access.carbon:[[@LINE+7]]:3: ERROR: Cannot access private member `radius` of type `Circle`.
+  // CHECK:STDERR:   circle.radius = 5;
+  // CHECK:STDERR:   ^~~~~~~~~~~~~
+  // CHECK:STDERR: fail_private_field_access.carbon:[[@LINE-25]]:15: The private member `radius` is defined here.
+  // CHECK:STDERR:   private var radius: i32;
+  // CHECK:STDERR:               ^~~~~~~~~~~
+  // CHECK:STDERR:
+  circle.radius = 5;
+  // CHECK:STDERR: fail_private_field_access.carbon:[[@LINE+7]]:3: ERROR: Cannot access private member `SOME_INTERNAL_CONSTANT` of type `Circle`.
+  // CHECK:STDERR:   circle.SOME_INTERNAL_CONSTANT;
+  // CHECK:STDERR:   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+  // CHECK:STDERR: fail_private_field_access.carbon:[[@LINE-32]]:15: The private member `SOME_INTERNAL_CONSTANT` is defined here.
+  // CHECK:STDERR:   private let SOME_INTERNAL_CONSTANT: i32 = 5;
+  // CHECK:STDERR:               ^~~~~~~~~~~~~~~~~~~~~~
+  // CHECK:STDERR:
+  circle.SOME_INTERNAL_CONSTANT;
+
+  // CHECK:STDERR: fail_private_field_access.carbon:[[@LINE+7]]:3: ERROR: Cannot access private member `SomeInternalFunction` of type `Circle`.
+  // CHECK:STDERR:   circle.SomeInternalFunction();
+  // CHECK:STDERR:   ^~~~~~~~~~~~~~~~~~~~~~~~~~~
+  // CHECK:STDERR: fail_private_field_access.carbon:[[@LINE-39]]:3: The private member `SomeInternalFunction` is defined here.
+  // CHECK:STDERR:   private fn SomeInternalFunction() -> i32 {
+  // CHECK:STDERR:   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+  // CHECK:STDERR:
+  circle.SomeInternalFunction();
+}
+
+// --- fail_protected_field_access.carbon
+
+library "protected_field_access";
+
+class A {
+  protected var x: i32;
+}
+
+fn Run() {
+  // CHECK:STDERR: fail_protected_field_access.carbon:[[@LINE+7]]:16: ERROR: Cannot access protected member `x` of type `A`.
+  // CHECK:STDERR:   let x: i32 = A.x;
+  // CHECK:STDERR:                ^~~
+  // CHECK:STDERR: fail_protected_field_access.carbon:[[@LINE-7]]:17: The protected member `x` is defined here.
+  // CHECK:STDERR:   protected var x: i32;
+  // CHECK:STDERR:                 ^~~~~~
+  // CHECK:STDERR:
+  let x: i32 = A.x;
+}
+
+// --- instance_private_field_access_on_self.carbon
+
+library "instance_private_field_access_on_self";
+
+class Circle {
+  private var radius: i32;
+
+  fn GetRadius[self: Self]() -> i32 {
+    return self.radius;
+  }
+
+  private fn SomeInternalFunction() -> i32 {
+    return 0;
+  }
+
+  fn Compute[self: Self]() -> i32 {
+    return self.SomeInternalFunction();
+  }
+}
+
+// --- public_global_access.carbon
+
+library "public_global_access";
+
+class A {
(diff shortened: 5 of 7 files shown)


toolchain/check/member_access.cpp
```diff
@@ -98,6 +101,39 @@ static auto IsInstanceMethod(const SemIR::File& sem_ir,
   return false;
 }
 
+// Returns whether the current function is an instance method of a class, that
+// is, whether it has an implicit `self` parameter and it's type
+// matches the type of class_self_type_id.
+static auto IsCurrentFunctionAnInstanceMethod(Context& context,
```
brymer-meneses: I feel conflicted on the name of this function, I think it can be named better, but I'm not sure what to.

brymer-meneses:
I'm also not sure whether it's best to split the test file into multiple files for this feature.

jonmeow (commented):
I'm focusing for the moment on the way name lookup is done in `DiagnoseInvalidMemberAccess`'s `get_member_info` versus `Context::LookupQualifiedName` (called by `LookupMemberNameInScope`).

Hash table lookup is typically going to be a little more expensive, because it's indirect. Sometimes it's unavoidable, but had you considered either implementing as part of `LookupQualifiedName`, or making more direct use of its results in order to avoid extra lookup?

Note, `Context::LookupQualifiedName` currently can't have multiple extended scopes, but it may end up having ambiguities which are non-ambiguous when accounting for access control.

That said, I think already the separate implementations create a risk of divergent name lookup results. In particular, how about a test such as:

```
base class A {
  fn F();
}

base class B {
  extend base: A;
  private fn F() -> i32;
}

base class C {
  extend base: B;
  fn G[self: Self]() -> () { return self.F(); }
}
```

Here, `F()` should resolve to `A::F` and compile successfully. The difference in return types should help identify if `DiagnoseInvalidMemberAccess` returns success while `LookupQualifiedName` continues to return `B::F`.

brymer-meneses:
> had you considered either implementing as part of LookupQualifiedName, or making more direct use of its results in order to avoid extra lookup?
 
`access_kind` is not exposed through `LookupQualifiedName` so I had to expose most of the logic to another function. I'm not sure if this is the right approach because of the code duplication.

brymer-meneses:
> Here, `F()` should resolve to `A::F` and compile successfully. The difference in return types should help identify if `DiagnoseInvalidMemberAccess` returns success while `LookupQualifiedName` continues to return `B::F`.

So running through the tester, I get: 
```
base class A {
  fn F();
}

base class B {
  extend base: A;
  private fn F() -> i32;
}

base class C {
  extend base: B;
  // CHECK:STDERR: lookup_private_access.carbon:[[@LINE+3]]:37: ERROR: Cannot access inherited private field `F` of type `C` inherited from `B`.
  // CHECK:STDERR:   fn G[self: Self]() -> () { return self.F(); }
  // CHECK:STDERR:                                     ^~~~~~
  fn G[self: Self]() -> () { return self.F(); }
}
```

I assume it has something to do with `GetEntryInLookupScope` which is mostly from `LookupQualifiedName`

jonmeow:
To be sure, what I'm trying to say is that I think this kind of access control check probably belongs in `LookupQualifiedName`, not separate from it. The current `GetEntryInLookupScope` approach is problematic because it creates a separate implementation of name lookup, which carries two problems:

- It risks divergence from the `LookupQualifiedName` implementation.
    - (this is what the example code is showing)
- It will repeat substantial work, even on valid code.
    - (In theory, we're more okay with doing extra work for invalid code, e.g. to improve diagnostics -- but valid code should be efficient)

My suggestion, to offer a little detail would be:

- Set aside the current work for a moment. It's useful for understanding the things which will come up, but may not make sense to use directly.
- Add an argument to LookupQualifiedName to indicate whether it should be allowing protected/private access. When allowed:
    - On the first scope, LookupQualifiedName would allow access to private members.
    - On extended scopes, LookupQualifiedName would only allow access to protected members.
    - (when not allowed, only public is allowed)
- LookupQualifiedName could track if it had seen inaccessible names, so that instead of DiagnoseNotFound it could instead diagnose the access restriction.

There's a bit more to flesh out there, but to be sure, member_access.cpp may only need a change to indicate to LookupQualifiedName which kind of access enforcement it requires. Rather than doing separate enforcement, it would be integrated into the existing lookup code.

Let me know if there's a little more I can clarify here, sorry it took some time to sort out my thoughts here.

brymer-meneses:
I think I have implemented all your suggestions. What do you think about printing a diagnostic for `self.base.x`? Should this PR handle that too?

toolchain/check/testdata/impl/no_prelude/import_self.carbon
```diff
@@ -27,6 +27,9 @@ impl () as Add {
 }
 
 fn F(x: (), y: ()) -> () {
+  // CHECK:STDERR: b.carbon:[[@LINE+3]]:13: ERROR: Cannot implicitly convert from `<associated <type of Op> in Add>` to `i32`.
+  // CHECK:STDERR:   return x.(Add.Op)(y);
```
brymer-meneses: I'm not sure what's causing this regression.

toolchain/check/context.h
```diff
@@ -45,6 +46,15 @@ struct LookupResult {
   SemIR::InstId inst_id;
 };
 
+// Information about an access
+struct AccessInfo {
+  // the instruction being accessed
+  SemIR::InstId base_id;
+
+  bool allow_private_access;
+  bool allow_protected_access;
```
jonmeow: Had you considered reusing the AccessKind enum here, rather than providing separate booleans?

```suggestion
  SemIR::AccessKind allowed_access;
```

toolchain/check/context.h
```diff
@@ -45,6 +46,15 @@ struct LookupResult {
   SemIR::InstId inst_id;
 };
 
+// Information about an access
+struct AccessInfo {
+  // the instruction being accessed
```
```suggestion
// Information about an access.
struct AccessInfo {
  // The instruction being accessed.
```

Punctuation style

toolchain/check/context.cpp
```diff
@@ -306,37 +383,66 @@ auto Context::LookupUnqualifiedName(Parse::NodeId node_id,
...
-    return inst_id;
+    auto entry = scope.names[lookup.value()];
+    LoadImportRef(*this, entry.inst_id);
+    return {entry.inst_id, entry.access_kind};
   }
+
   if (!scope.import_ir_scopes.empty()) {
-    return ImportNameFromOtherPackage(*this, loc, scope_id,
-                                      scope.import_ir_scopes, name_id);
+    // TODO: Enforce other access modifiers for imports
```
```suggestion
    // TODO: Enforce other access modifiers for imports.
```

[Punctuation style](https://google.github.io/styleguide/cppguide.html#Punctuation,_Spelling_and_Grammar) is small, but it can help to just add periods at the end of comments.

Note, I'll be making similar suggestions in other parts of this change, just noting it in particular here as a reference.

toolchain/diagnostics/diagnostic_kind.def
```diff
@@ -336,6 +336,14 @@ CARBON_DIAGNOSTIC_KIND(ModifierNotAllowedWith)
 CARBON_DIAGNOSTIC_KIND(ModifierMustAppearBefore)
 CARBON_DIAGNOSTIC_KIND(ModifierPrevious)
 
+// Access modifiers
```
```suggestion
// Access modifiers.
```

toolchain/sem_ir/class.h
```diff
@@ -70,6 +70,16 @@ struct Class : public EntityWithParamsBase,
   // Determines whether this class has been fully defined. This is false until
   // we reach the `}` of the class definition.
   auto is_defined() const -> bool { return object_repr_id.is_valid(); }
+
+  // Returns the inst id of the class this extends. This can be either a base
+  // class declaration or an adapted type declaration
```
jonmeow:
```suggestion
  // Returns the inst id of the class this extends. This can be either a base
  // class declaration or an adapted type declaration.
```

toolchain/check/member_access.cpp
```diff
@@ -207,14 +228,38 @@ static auto PerformImplLookup(Context& context, Parse::NodeId node_id,
...
+
+        if (is_access_from_instance_method) {
+          access_info.allow_private_access = true;
+          access_info.allow_protected_access = true;
+        }
+      }
+    }
+
+    result = context.LookupQualifiedName(node_id, name_id, lookup_scope,
+                                         /*required*/ true, access_info);
```
jonmeow:
```suggestion
                                         /*required=*/true, access_info);
```

The `=` here is significant for clang-format, and changes the spacing.

toolchain/check/testdata/class/access_modifers.carbon
```diff
@@ -0,0 +1,952 @@
...
+  // CHECK:STDERR:   private fn SomeInternalFunction() -> i32 {
+  // CHECK:STDERR:   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+  // CHECK:STDERR:
+  // CHECK:STDERR: fail_inherited_private_field_access.carbon: ERROR: Main//default previously provided by `fail_private_field_access.carbon`.
+  // CHECK:STDERR:
+  circle.SomeInternalFunction();
+}
+
+// --- fail_inherited_private_field_access.carbon
+
```
jonmeow:
```suggestion
// --- fail_inherited_private_field_access.carbon

library "inherited_private_field_access";

```

toolchain/check/testdata/class/access_modifers.carbon
```diff
@@ -0,0 +1,952 @@
...
+//
+// AUTOUPDATE
+// TIP: To test this file alone, run:
+// TIP:   bazel test //toolchain/testing:file_test --test_arg=--file_tests=toolchain/check/testdata/class/access_modifers.carbon
+// TIP: To dump output, run:
+// TIP:   bazel run //toolchain/testing:file_test -- --dump_output --file_tests=toolchain/check/testdata/class/access_modifers.carbon
+
+// --- fail_private_field_access.carbon
+
+
```
jonmeow:
```suggestion
// --- fail_private_field_access.carbon

library "private_field_access";

```

In the test, you have warnings like:

```
ERROR: Main//default previously provided by `fail_private_field_access.carbon`.
```

This error occurs because we're essentially just compiling all the files in a single invocation, and all the files are trying to provide the same, default library. Providing a `library` directive fixes it; we typically just name libraries after the filename.

toolchain/check/testdata/class/access_modifers.carbon
```diff
@@ -0,0 +1,952 @@
...
+    // CHECK:STDERR:
+    // CHECK:STDERR: fail_protected_field_access.carbon: ERROR: Main//default previously provided by `fail_private_field_access.carbon`.
+    // CHECK:STDERR:
+    return self.y;
+  }
+}
+
+// --- fail_protected_field_access.carbon
+
+
```
jonmeow:
```suggestion
// --- fail_protected_field_access.carbon

library "protected_field_access";

```

toolchain/check/testdata/class/access_modifers.carbon
```diff
@@ -0,0 +1,952 @@
...
+  // CHECK:STDERR:
+  // CHECK:STDERR: instance_protected_field_access.carbon: ERROR: Main//default previously provided by `fail_private_field_access.carbon`.
+  // CHECK:STDERR:
+  // CHECK:STDERR: shadowing_access.carbon: ERROR: Main//default previously provided by `fail_private_field_access.carbon`.
+  let y: i32 = circle.y;
+}
+
+
+// --- instance_private_field_access.carbon
+
```
jonmeow:
```suggestion
// --- instance_private_field_access.carbon

library "instance_private_field_access";

```

toolchain/check/testdata/class/access_modifers.carbon
```diff
@@ -0,0 +1,952 @@
...
+    return 0;
+  }
+
+  fn Compute[self: Self]() -> i32 {
+    return self.SomeInternalFunction();
+  }
+}
+
+// --- instance_protected_field_access.carbon
+
```
jonmeow:
```suggestion
// --- instance_protected_field_access.carbon

library "instance_protected_field_access";

```

toolchain/check/testdata/class/access_modifers.carbon
```diff
@@ -0,0 +1,952 @@
...
+  extend base: Shape;
+  private var radius: i32;
+
+  fn GetPosition[self: Self]() -> (i32, i32) {
+    return (self.x, self.y);
+  }
+}
+
+// --- shadowing_access.carbon
+
```
jonmeow:
```suggestion
// --- shadowing_access.carbon

library "shadowing_access";

```

toolchain/check/testdata/class/access_modifers.carbon
```diff
@@ -0,0 +1,952 @@
...
+class Circle {
+  private var radius: i32;
+  private let SOME_INTERNAL_CONSTANT: i32 = 5;
+
+  private fn SomeInternalFunction() -> i32 {
+    return 0;
+  }
+
+  fn Make() -> Self {
+    return { .radius = 5 };
```
jonmeow:
```suggestion
    return {.radius = 5};
```

I'll be making a number of suggestions like this to get the file to align a little more with where we expect style to be.

toolchain/check/testdata/class/access_modifers.carbon
```diff
@@ -0,0 +1,952 @@
...
+  }
+
```
jonmeow:
```suggestion
```

(style)

toolchain/check/testdata/class/access_modifers.carbon
```diff
@@ -0,0 +1,952 @@
...
+// --- fail_inherited_private_field_access.carbon
+
+base class Shape {
+  private var x: i32;
+  private var y: i32;
+}
+
+class Square {
+  extend base: Shape;
+  private var radius: i32;
```
jonmeow:
```suggestion
```

This looks unused.

toolchain/check/testdata/class/access_modifers.carbon
```diff
@@ -0,0 +1,952 @@
...
+  // CHECK:STDERR:
+  // CHECK:STDERR: fail_inherited_private_field_access.carbon: ERROR: Main//default previously provided by `fail_private_field_access.carbon`.
+  // CHECK:STDERR:
+  circle.SomeInternalFunction();
+}
+
+// --- fail_inherited_private_field_access.carbon
+
+base class Shape {
+  private var x: i32;
```
jonmeow:
```suggestion
```

This looks unused. In tests, I think it'd be helpful to focus on what's being tested -- it'll reduce the amount of reading and trying to figure out whether there's some potential interference you're validating (which I don't think is the case here)

toolchain/check/testdata/class/access_modifers.carbon
```diff
@@ -0,0 +1,952 @@
...
+  fn SomeFunc[self: Self]() -> i32{
+    // CHECK:STDERR: fail_protected_field_access.carbon:[[@LINE+7]]:12: ERROR: Cannot access protected member `INTERNAL_CONSTANT` of type `Internal`
+    // CHECK:STDERR:     return self.internal.INTERNAL_CONSTANT;
+    // CHECK:STDERR:            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    // CHECK:STDERR: fail_protected_field_access.carbon:[[@LINE-17]]:17: Protected member `INTERNAL_CONSTANT` defined here.`
+    // CHECK:STDERR:   protected let INTERNAL_CONSTANT: i32 = 5;
+    // CHECK:STDERR:                 ^~~~~~~~~~~~~~~~~
+    // CHECK:STDERR:
+    return self.internal.INTERNAL_CONSTANT;
+  }
```
jonmeow: What do you think of separating this out to a different test (a separate `// ---` split)? i.e., so that tests related to the inheritance errors are separate from tests related to member variables.

toolchain/check/testdata/class/access_modifers.carbon
```diff
@@ -0,0 +1,952 @@
...
+
+}
+
+base class Internal {
+  protected let INTERNAL_CONSTANT: i32 = 5;
+}
+
+class Circle {
+  extend base: Shape;
+  private var radius: i32;
```
jonmeow:
```suggestion
```

This looks unused, can it be removed?

toolchain/check/testdata/class/access_modifers.carbon
```diff
@@ -0,0 +1,952 @@
...
+    // CHECK:STDERR: fail_protected_field_access.carbon:[[@LINE+7]]:12: ERROR: Cannot access protected member `INTERNAL_CONSTANT` of type `Internal`
+    // CHECK:STDERR:     return self.internal.INTERNAL_CONSTANT;
+    // CHECK:STDERR:            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    // CHECK:STDERR: fail_protected_field_access.carbon:[[@LINE-17]]:17: Protected member `INTERNAL_CONSTANT` defined here.`
+    // CHECK:STDERR:   protected let INTERNAL_CONSTANT: i32 = 5;
+    // CHECK:STDERR:                 ^~~~~~~~~~~~~~~~~
+    // CHECK:STDERR:
+    return self.internal.INTERNAL_CONSTANT;
+  }
+
```
jonmeow:
```suggestion
```

(style)

toolchain/check/testdata/class/access_modifers.carbon
```diff
@@ -0,0 +1,952 @@
...
+}
+
+class Circle {
+  extend base: Shape;
+  private var radius: i32;
+  private var internal: Internal;
+
+
+  fn Make() -> Self {
+    return { .radius = 5, .base = {.x = 5, .y= 5}, .internal = {} };
```
jonmeow:
```suggestion
    return {.radius = 5, .base = {.x = 5, .y = 5}, .internal = {}};
```

(style)

toolchain/check/testdata/class/access_modifers.carbon
```diff
@@ -0,0 +1,952 @@
...
+base class Internal {
+  protected let INTERNAL_CONSTANT: i32 = 5;
+}
+
+class Circle {
+  extend base: Shape;
+  private var radius: i32;
+  private var internal: Internal;
+
+
```
jonmeow:
```suggestion
```

(style)

toolchain/check/testdata/class/access_modifers.carbon
```diff
@@ -0,0 +1,952 @@
...
+  }
+}
+
+// --- fail_protected_field_access.carbon
+
+
+base class Shape {
+  protected var x: i32;
+  protected var y: i32;
+
```
jonmeow:
```suggestion
```

(style)

toolchain/check/testdata/class/access_modifers.carbon
```diff
@@ -0,0 +1,952 @@
...
+  // CHECK:STDERR: fail_protected_field_access.carbon:[[@LINE-44]]:17: Protected member `y` defined here.`
+  // CHECK:STDERR:   protected var y: i32;
+  // CHECK:STDERR:                 ^~~~~~
+  // CHECK:STDERR:
+  // CHECK:STDERR: instance_private_field_access.carbon: ERROR: Main//default previously provided by `fail_private_field_access.carbon`.
+  // CHECK:STDERR:
+  // CHECK:STDERR: instance_protected_field_access.carbon: ERROR: Main//default previously provided by `fail_private_field_access.carbon`.
+  // CHECK:STDERR:
+  // CHECK:STDERR: shadowing_access.carbon: ERROR: Main//default previously provided by `fail_private_field_access.carbon`.
+  let y: i32 = circle.y;
```
jonmeow: What's distinct about `circle.x` versus `circle.y` here? That is, this looks a lot like it's repeating the same error; is distinct behavior being tested, or can we remove one?

toolchain/check/member_access.cpp
```diff
@@ -207,14 +228,38 @@ static auto PerformImplLookup(Context& context, Parse::NodeId node_id,
...
+
+    if (auto class_type = context.insts().TryGetAs<SemIR::ClassType>(
+            context.constant_values().GetInstId(name_scope_const_id))) {
+      auto class_inst = context.classes().Get(class_type->class_id);
+
+      if (auto current_function = GetCurrentFunction(context)) {
+        auto is_access_from_instance_method = IsInstanceMethod(
+            context.sem_ir(), *current_function, class_inst.self_type_id);
+
+        if (is_access_from_instance_method) {
```
jonmeow: Is this trying to determine whether the current function is a member of the type being accessed? The use of `is_access_from_instance_method` I think will have a different consequence: access _from_ non-member methods, even to private methods on the current type, looks like it will be blocked. Sometimes a "TODO" comment is good enough for things that don't work, but since you're modifying `IsInstanceMethod` to support this, I think it's worth a look to see if we can get better results without that edit.

It's possible comparing scopes will help. I think we do want something like `ParentType.X` to work. So something avoiding the `IsInstanceMethod` call that yields a similar result might be:

```suggestion
    // Allow restricted access when accessing a member of the current type.
    // TODO: This should probably be more specific to `self`.
    // TODO: Support entities other than classes.
    if (auto class_type = context.insts().TryGetAs<SemIR::ClassType>(
            context.constant_values().GetInstId(name_scope_const_id))) {
      auto class_scope_id = context.classes().Get(class_type->class_id).scope_id;

      // TODO: Support contexts other than functions, e.g.: `fn F() -> Self.Type`.
      if (auto fn_id = GetCurrentFunction(context)) {
        auto fn_scope_id = context.functions().Get(*fn_id).name_scope_id;
        
        if (class_scope_id == fn_scope_id) {
```

It's probably also worth adding tests of this behavior -- regardless of whether they pass or not (prefixing with `fail_todo_` is something we'll do for tests that currently fail, but _should_ pass).

In the below examples, I think `G` would be marked as non-instance and then excluded from private/protected access. However, I think they should succeed (except for the one `fail_` example).

```
// --- noninstance_private_on_self.carbon

library "noninstance_private_on_self";

class C {
  private fn F() {}
  fn G() { Self.F(); }
}

// --- noninstance_protected_on_self.carbon

library "noninstance_protected_on_self";

class C {
  protected fn F() {}
  fn G() { Self.F(); }
}

// --- fail_noninstance_private_on_parent.carbon

library "noninstance_private_on_parent";

base class B {
  private fn F() {}
}

class C {
  extend base: B;
  fn G() { Self.F(); }
}

// --- noninstance_protected_on_parent.carbon

library "noninstance_protected_on_parent";

base class B {
  protected fn F() {}
}

class C {
  extend base: B;
  fn G() { Self.F(); }
}
```

toolchain/check/member_access.cpp
```diff
@@ -417,15 +462,18 @@ auto PerformCompoundMemberAccess(Context& context, Parse::NodeId node_id,
   auto member_id = member_expr_id;
   auto member = context.insts().Get(member_id);
 
+  // If this is a tuple access
+  if (context.insts().Is<SemIR::TupleType>(
+          context.constant_values().GetInstId(base_type_const_id))) {
+    return PerformTupleIndex(context, node_id, base_id, member_expr_id);
+  }
```
jonmeow: Why is this moved? It doesn't look like it should be affected by the access control edits.

toolchain/check/member_access.cpp
```diff
@@ -332,8 +377,8 @@ static auto ValidateTupleIndex(Context& context, Parse::NodeId node_id,
 auto PerformMemberAccess(Context& context, Parse::NodeId node_id,
                          SemIR::InstId base_id, SemIR::NameId name_id)
     -> SemIR::InstId {
-  // If the base is a name scope, such as a class or namespace, perform lookup
-  // into that scope.
+  // If the base is a name scope, such as a class or namespace, perform
+  // lookup into that scope.
```
jonmeow: Why is this changed? If there's no particular reason, I think we should keep the prior state -- it looks like it was wrapped correctly to 80 chars.

toolchain/check/member_access.cpp
```diff
@@ -83,21 +86,39 @@ static auto GetClassElementIndex(Context& context, SemIR::InstId element_id)
...
     }
   }
 
   return false;
 }
 
+// Returns the FunctionId of the current function if it exists.
+static auto GetCurrentFunction(Context& context)
+    -> std::optional<SemIR::FunctionId> {
+  auto return_scope_stack = context.return_scope_stack();
```
jonmeow: I think we typically avoid doing `auto` captures for `context` members. Can you just write it directly below, e.g. `context.return_scope_stack().empty()`?

toolchain/check/impl.cpp
```diff
@@ -163,7 +163,7 @@ static auto BuildInterfaceWitness(
           CARBON_FATAL() << "Unexpected type: " << type_inst;
         }
         auto& fn = context.functions().Get(fn_type->function_id);
-        auto impl_decl_id = context.LookupNameInExactScope(
+        auto [impl_decl_id, access_kind] = context.LookupNameInExactScope(
```
jonmeow:
```suggestion
        auto [impl_decl_id, _] = context.LookupNameInExactScope(
```

Idiomatically, unused fields are given the `_` name to indicate they're discarded.

toolchain/check/context.cpp
```diff
@@ -306,37 +383,66 @@ auto Context::LookupUnqualifiedName(Parse::NodeId node_id,
...
+      return true;
+    }
+
+    if (access_kind == SemIR::AccessKind::Protected &&
+        access_info->allow_protected_access) {
+      return true;
+    }
+
+    return false;
+  };
```
jonmeow: This is decently sized, and only called in one spot. Generally I think we should be cautious about lambdas, so that the containing function's flow control is easy to read. Here, a couple options are:

1) Move out to a static helper function.
2) Move it inline, replacing the lambda call.

toolchain/check/context.cpp
```diff
@@ -306,37 +383,66 @@ auto Context::LookupUnqualifiedName(Parse::NodeId node_id,
...
+        access_info->allow_private_access && !is_parent_access) {
+      return true;
+    }
+
+    if (access_kind == SemIR::AccessKind::Protected &&
+        access_info->allow_protected_access) {
+      return true;
+    }
+
+    return false;
```
jonmeow: Had you considered writing this as a `switch(access_kind)` instead of individually checking `access_kind` values?

e.g.:

```suggestion
    if (!access_info) {
      return true;
    }
    switch (access_kind) {
      case SemIR::AccessKind::Public:
        return true;
      case SemIR::AccessKind::Protected:
        return access_info->allow_protected_access;
      case SemIR::AccessKind::Private:
        return access_info->allow_private_access && !is_parent_access;
    }
```

toolchain/check/context.cpp
```diff
@@ -306,37 +383,66 @@ auto Context::LookupUnqualifiedName(Parse::NodeId node_id,
...
     has_error |= name_scope.has_error;
 
-    auto scope_result_id =
+    auto [scope_result_id, access_kind] =
         LookupNameInExactScope(node_id, name_id, scope_id, name_scope);
-    if (!scope_result_id.is_valid()) {
-      // Nothing found in this scope: also look in its extended scopes.
+
+    auto is_access_not_allowed =
+        !is_access_allowed(access_kind, is_parent_access);
```
jonmeow: How about aligning both of these names? That is, either have both be "allowed" or "disallowed", rather than switching the boolean. I think that'd help readability by avoiding a logic inversion, making it clearer that this is just storing the result.

toolchain/check/context.cpp
```diff
@@ -345,6 +451,17 @@ auto Context::LookupQualifiedName(Parse::NodeId node_id, SemIR::NameId name_id,
         scopes.push_back({.name_scope_id = extended_id,
                           .specific_id = SemIR::SpecificId::Invalid});
       }
+      is_parent_access |= !extended.empty();
+
+      // If this is the last scope and we still haven't found a valid
+      // lookup, diagnose an invalid access.
+      if (scopes.empty() && is_access_not_allowed) {
+        DiagnoseInvalidAccess(node_id, /*member_id*/ scope_result_id, name_id,
```
```suggestion
        DiagnoseInvalidAccess(node_id, /*member_id=*/scope_result_id, name_id,
```

I note this elsewhere too, but the `=` is significant for clang-format.

toolchain/check/context.cpp
```diff
@@ -345,6 +451,17 @@ auto Context::LookupQualifiedName(Parse::NodeId node_id, SemIR::NameId name_id,
         scopes.push_back({.name_scope_id = extended_id,
                           .specific_id = SemIR::SpecificId::Invalid});
       }
+      is_parent_access |= !extended.empty();
+
+      // If this is the last scope and we still haven't found a valid
+      // lookup, diagnose an invalid access.
+      if (scopes.empty() && is_access_not_allowed) {
```
jonmeow: This should only run after all scopes have been checked, right? Rather than putting this code here, what changes if you put this outside the `while` loop? What if it's specifically where `DiagnoseNameNotFound` is called, also taking advantage of the `has_error` check that's already there?

Note you will probably need to track the entities which access was denied to, rather than assuming they are in the final scope. I think it's worth this change because I think the implementation is going to change again. For example:

```
base class A {}

base class B {
  extend base: A;
  private fn F[self: Self]();
}

class C {
  extend base: B;
  fn G[self: Self]() { self.F(); }
}

File: toolchain/check/testdata/class/access_modifers.carbon [hunk 36]
```diff
@@ -0,0 +1,952 @@
...
+  private var x: i32;
+  private var y: i32;
+}
+
+class Square {
+  extend base: Shape;
+  private var radius: i32;
+
+  fn GetPosition[self: Self]() -> i32 {
+    // CHECK:STDERR: fail_inherited_private_field_access.carbon:[[@LINE+9]]:12: ERROR: Cannot access inherited private member `y` of type `Square` inherited from `Shape`.
```
jonmeow: I think that calling `y` a member of `Square` is misleading. Maybe we should just say `` private member `y` of type `Shape` `` for now, and work on better phrasing later?

toolchain/check/context.cpp
```diff
@@ -187,6 +187,82 @@ auto Context::DiagnoseNameNotFound(SemIRLoc loc, SemIR::NameId name_id)
   emitter_->Emit(loc, NameNotFound, name_id);
 }
 
+auto Context::DiagnoseInvalidAccess(
```
jonmeow: We'll make things a member of `Context` when they're called from other files. Since this is only called locally, maybe it can be a `static` function in the cpp file?

You might consider also moving the implementation next to `LookupQualifiedName` to make it easier to find, since that's likely to be the only caller.

jonmeow: Would a name like `DiagnoseInvalidQualifiedNameAccess` work for you? In particular, this isn't generally diagnosing invalid access, it's specific to that scenario.

toolchain/check/context.cpp
```diff
@@ -187,6 +187,82 @@ auto Context::DiagnoseNameNotFound(SemIRLoc loc, SemIR::NameId name_id)
...
+
+  auto inst = insts().Get(access_info->base_id);
+  auto inst_type_id = inst.type_id();
+
+  if (auto class_type = types().TryGetAs<SemIR::ClassType>(inst_type_id)) {
+    auto class_inst = classes().Get(class_type->class_id);
+    auto is_access_from_instance_method = access_info->allow_protected_access &&
+                                          access_info->allow_private_access;
+
+    CARBON_DCHECK(class_inst.is_defined());
```
```suggestion
    CARBON_CHECK(class_inst.is_defined());
```

We're only using DCHECK where a small performance cost still significantly affects performance. When diagnosing invalid code, there isn't much reason to avoid CHECK -- invalid code is expected to perform worse for compilation.

toolchain/check/context.cpp
```diff
@@ -187,6 +187,82 @@ auto Context::DiagnoseNameNotFound(SemIRLoc loc, SemIR::NameId name_id)
...
+    SemIR::AccessKind access_kind, bool is_parent_access,
+    std::optional<AccessInfo> access_info) -> void {
+  if (!access_info) {
+    return;
+  }
+
+  auto inst = insts().Get(access_info->base_id);
+  auto inst_type_id = inst.type_id();
+
+  if (auto class_type = types().TryGetAs<SemIR::ClassType>(inst_type_id)) {
```
jonmeow: Rather than nesting the rest of the function, I think it'd be better to return early if this isn't supported.

```suggestion
  // TODO: Support scoped entities other than just classes.
  if (auto class_type = types().TryGetAs<SemIR::ClassType>(inst_type_id)) {
```

I assume you're deliberately not supporting other entities, but this should be called out as a TODO.

toolchain/check/context.cpp
```diff
@@ -187,6 +187,82 @@ auto Context::DiagnoseNameNotFound(SemIRLoc loc, SemIR::NameId name_id)
...
+      if (!is_access_from_instance_method) {
+        CARBON_DIAGNOSTIC(ClassPrivateMemberAccess, Error,
+                          "Cannot access private member `{0}` of type `{1}`",
+                          SemIR::NameId, SemIR::TypeId);
+        emitter()
+            .Build(node_id, ClassPrivateMemberAccess, name_id, inst_type_id)
+            .Note(member_id, ClassPrivateMemberDefinition, name_id)
+            .Emit();
+        return;
+      }
```
jonmeow: These two diagnostics are very similar. Can you make "private" versus "protected" a parameter to a single diagnostic that handles both cases?

toolchain/check/context.cpp
```diff
@@ -187,6 +187,82 @@ auto Context::DiagnoseNameNotFound(SemIRLoc loc, SemIR::NameId name_id)
...
+
+    CARBON_DCHECK(class_inst.is_defined());
+
+    if (access_kind == SemIR::AccessKind::Protected) {
+      if (!is_access_from_instance_method) {
+        CARBON_DIAGNOSTIC(ClassProtectedMemberAccess, Error,
+                          "Cannot access protected member `{0}` of type `{1}`",
+                          SemIR::NameId, SemIR::TypeId);
+        CARBON_DIAGNOSTIC(ClassProtectedMemberDefinition, Note,
+                          "Protected member `{0}` defined here.`",
```
```suggestion
                          "Protected member `{0}` defined here.",
```

toolchain/check/context.cpp
```diff
@@ -187,6 +187,82 @@ auto Context::DiagnoseNameNotFound(SemIRLoc loc, SemIR::NameId name_id)
...
+            .Build(node_id, ClassProtectedMemberAccess, name_id, inst_type_id)
+            .Note(member_id, ClassProtectedMemberDefinition, name_id)
+            .Emit();
+        return;
+      }
+    }
+
+    if (access_kind == SemIR::AccessKind::Private) {
+      CARBON_DIAGNOSTIC(ClassPrivateMemberDefinition, Note,
+                        "Private member `{0}` defined here.`", SemIR::NameId);
```
```suggestion
                        "Private member `{0}` defined here.", SemIR::NameId);
```

toolchain/check/context.cpp
```diff
@@ -187,6 +187,82 @@ auto Context::DiagnoseNameNotFound(SemIRLoc loc, SemIR::NameId name_id)
...
+auto Context::DiagnoseInvalidAccess(
+    Parse::NodeId node_id, SemIR::InstId member_id, SemIR::NameId name_id,
+    SemIR::AccessKind access_kind, bool is_parent_access,
+    std::optional<AccessInfo> access_info) -> void {
+  if (!access_info) {
+    return;
+  }
+
+  auto inst = insts().Get(access_info->base_id);
+  auto inst_type_id = inst.type_id();
```
jonmeow: Had you considered calling `inst.type_id()` instead of adding this local variable?

toolchain/check/context.cpp
```diff
@@ -187,6 +187,82 @@ auto Context::DiagnoseNameNotFound(SemIRLoc loc, SemIR::NameId name_id)
...
+
+  if (auto class_type = types().TryGetAs<SemIR::ClassType>(inst_type_id)) {
+    auto class_inst = classes().Get(class_type->class_id);
```
jonmeow: This is the `Class` object, which is not an instruction ("inst"). We typically call things like this "class_info".

toolchain/check/context.cpp
```diff
@@ -187,6 +187,82 @@ auto Context::DiagnoseNameNotFound(SemIRLoc loc, SemIR::NameId name_id)
...
+        auto parent_id = class_inst.parent_id();
+        auto parent_type_id = SemIR::TypeId::Invalid;
+        if (auto base_decl = insts().TryGetAs<SemIR::BaseDecl>(parent_id)) {
+          parent_type_id = base_decl->base_type_id;
+        }
+        if (auto adapt_decl = insts().TryGetAs<SemIR::AdaptDecl>(parent_id)) {
+          parent_type_id = adapt_decl->adapted_type_id;
+        }
+
+        CARBON_DIAGNOSTIC(
```
```suggestion
        // TODO: Use a different diagnostic for "adapts" versus "extends".
        CARBON_DIAGNOSTIC(
```

toolchain/check/context.cpp
```diff
@@ -187,6 +187,82 @@ auto Context::DiagnoseNameNotFound(SemIRLoc loc, SemIR::NameId name_id)
...
+
+      if (is_parent_access && is_access_from_instance_method) {
+        auto parent_id = class_inst.parent_id();
+        auto parent_type_id = SemIR::TypeId::Invalid;
+        if (auto base_decl = insts().TryGetAs<SemIR::BaseDecl>(parent_id)) {
+          parent_type_id = base_decl->base_type_id;
+        }
+        if (auto adapt_decl = insts().TryGetAs<SemIR::AdaptDecl>(parent_id)) {
+          parent_type_id = adapt_decl->adapted_type_id;
+        }
```
jonmeow: It had seemed like `parent_id` was being added in order to transparently check both, but then you need to individually compare `BaseDecl` and `AdaptDecl`. Have you considered removing `parent_id`, and instead querying both here, doing the appropriate cast for each?

jonmeow (commented):
Thanks for the changes! I have a lot of comments, but I want to be clear, I do think this is a big improvement. I just tend to add a lot of comments for style things (a chunk of these, for example, are just small style things in the test file that happen to stick out to me).

I want to highlight two that are more structural:

- Moving the diagnostic call to where `DiagnoseNameNotFound` is found.
    - [comment here](https://github.com/carbon-language/carbon-lang/pull/4248#discussion_r1744237369)
    - I think this will shift how you track skipped calls, and also may lead to more changes to DiagnoseInvalidAccess.
- Avoiding querying whether the current function is an instance function before determining what to do with it.
    - [comment here](https://github.com/carbon-language/carbon-lang/pull/4248#discussion_r1744202933)

I'd suggest starting with these two comments, because they may still shift your approach a little.

toolchain/check/context.h
brymer-meneses: I considered doing this. But, I'm not sure how `allowed_access` can contain information if the access is allowed for _both_ protected and private, or only needing private. 

I thought of turning `AccessKind` into a flag enum, similar to `NodeCategory`, but I felt conflicted since `AccessKind` is originally used to indicate an exclusive access in `Entry.` 

Or perhaps do you mean, like `highest_allowed_access`, so if the access is allowed for `private`, it's allowed for `protected` as well?

toolchain/check/member_access.cpp
brymer-meneses: Sorry for the oversight, I felt like editing this because the `else if` doesn't really make sense, and might introduce a subtle bug. 

This is small, that I was not sure if making another PR for this is warranted. But I'll be happy to do that if you want.

toolchain/check/member_access.cpp
brymer-meneses: Sorry about that, this was caused by `clang-format,` on my ide. I'm not really sure why it doesn't format as expected and seems to format other things as well. But `precommit` seems to revert most of the unnecessary changes, for some reason this wasn't.

toolchain/check/context.cpp
```diff
@@ -306,37 +308,155 @@ auto Context::LookupUnqualifiedName(Parse::NodeId node_id,
...
+  std::optional<SemIR::ClassType> class_type = std::nullopt;
+
+  auto inst_constant_id = context.constant_values().Get(access_info->base_id);
+  if (!inst_constant_id.is_constant()) {
+    class_type = context.types().TryGetAs<SemIR::ClassType>(
+        context.insts().Get(access_info->base_id).type_id());
+  } else {
+    class_type = context.insts().TryGetAs<SemIR::ClassType>(
+        context.constant_values().GetInstId(inst_constant_id));
+  }
```
brymer-meneses: I feel like this can be written more cleanly.

toolchain/check/member_access.cpp
jonmeow: Have you tried reverting this change? It looks to me like this is probably responsible for the regression in `import_self.carbon` output. I think the `else` was significant.

A small comment change here might be okay, but maybe a separate PR is justified due to the logic change, to help give confidence that it doesn't affect tests?

toolchain/check/testdata/impl/no_prelude/import_self.carbon
jonmeow: https://github.com/carbon-language/carbon-lang/pull/4248/files#r1744204989

toolchain/check/context.h
```diff
@@ -45,6 +46,25 @@ struct LookupResult {
...
+  //
+  // - If this is set to `AccessKind::Private` then this access will be
+  // allowed for `AccessKind::Private`, `AccessKind::Protected` and
+  // `AccessKind::Public`.
+  //
+  // - If this is set to `AccessKind::Protected` then this will be
+  // allowed for both `AccessKind::Protected` and `AccessKind::Public`.
+  //
+  // - If this is set to `AccessKind::Public` then this will only be
+  // allowed for `AccessKind::Public`.
```
```suggestion
  // The highest allowed access for a lookup. For example, `Protected` allows
  // access to `Public` and `Protected` names, but not `Private`.
```

The current comment seems a little long, does this shorter example work for you?

toolchain/check/context.cpp
```diff
@@ -306,37 +309,146 @@ auto Context::LookupUnqualifiedName(Parse::NodeId node_id,
...
+  if (!access_info) {
+    return false;
+  }
+
+  switch (access_kind) {
+    case SemIR::AccessKind::Public:
+      return false;
+    case SemIR::AccessKind::Protected:
+      return access_info->highest_allowed_access !=
+             SemIR::AccessKind::Protected;
```
jonmeow: IIUC you're still making changes, but I'll flag != versus < here. Looking at the tests in https://github.com/carbon-language/carbon-lang/pull/4248#discussion_r1744202933, `noninstance_protected_on_self` might catch this.

toolchain/check/member_access.cpp
```diff
@@ -98,6 +102,79 @@ static auto IsInstanceMethod(const SemIR::File& sem_ir,
...
+
+  return context.insts()
+      .GetAs<SemIR::FunctionDecl>(context.return_scope_stack().back().decl_id)
+      .function_id;
+}
+
+static auto GetHighestAllowedAccess(Context& context, Parse::NodeId node_id,
+                                    SemIR::ConstantId name_scope_const_id)
+    -> SemIR::AccessKind {
+  auto current_function = GetCurrentFunction(context);
```
jonmeow, line 123:
```suggestion
  // TODO: Maybe use LookupUnqualifiedName for `Self` to support things like
  // `var x: Self.ParentProtectedType`?
  auto current_function = GetCurrentFunction(context);
```

Just a note. I think it's fine to only support functions for now.

jonmeow (commented):
NB, just a few comments since it looks like you're still addressing old comments, but I wanted to point out a couple small things.

toolchain/check/context.h
brymer-meneses: Sounds good.

toolchain/check/member_access.cpp
brymer-meneses: sounds good.

toolchain/check/context.cpp
brymer-meneses: It doesn't seem to be the case?

brymer-meneses: Oh, I see the problem now. Thanks!

toolchain/check/context.cpp
```diff
@@ -4,7 +4,9 @@
 
 #include "toolchain/check/context.h"
 
+#include <optional>
 #include <string>
+#include <string_view>
```
jonmeow: We use llvm::StringRef instead of std::string_view (in general, we are preferring LLVM types over std types where feasible). Note, I have particular suggestions below.

```suggestion
```

toolchain/check/member_access.cpp
```diff
@@ -98,6 +101,84 @@ static auto IsInstanceMethod(const SemIR::File& sem_ir,
...
+    } else if (auto adapt_decl =
+                   context.insts().TryGetAsIfValid<SemIR::AdaptDecl>(
+                       self_class_info.adapt_id)) {
+      is_protected_access =
+          adapt_decl->adapted_type_id == class_info.self_type_id;
+    }
+
+    if (is_protected_access) {
+      return SemIR::AccessKind::Protected;
+    }
```
jonmeow: I'd typically write "bool" instead of "auto", but what do you think of directly returning here?

```suggestion
    if (auto base_decl = context.insts().TryGetAsIfValid<SemIR::BaseDecl>(
            self_class_info.base_id)) {
      if (base_decl->base_type_id == class_info.self_type_id) {
        return SemIR::AccessKind::Protected;
      }
    } else if (auto adapt_decl =
                   context.insts().TryGetAsIfValid<SemIR::AdaptDecl>(
                       self_class_info.adapt_id)) {
      if (adapt_decl->adapted_type_id == class_info.self_type_id) {
        return SemIR::AccessKind::Protected;
      }
    }
```

toolchain/check/testdata/class/access_modifers.carbon
```diff
@@ -0,0 +1,659 @@
...
+
+// --- public_global_access.carbon
+
+library "public_global_access";
+
+class A {
+  let x: i32 = 5;
+
+  private let internal: i32 = 10;
+  let y: i32 = internal;
```
jonmeow:
```suggestion
```

This test's name is "public global access". Should this private test be removed? It looks similar to tests elsewhere, too.

toolchain/check/testdata/class/inheritance_access.carbon
```diff
@@ -0,0 +1,1014 @@
...
+library "instance_protected_field_access";
+
+base class Shape {
+  protected var x: i32;
+  protected var y: i32;
+}
+
+class Circle {
+  extend base: Shape;
+  private var radius: i32;
```
jonmeow:
```suggestion
```

Unused.

toolchain/check/testdata/class/inheritance_access.carbon
```diff
@@ -0,0 +1,1014 @@
...
+  fn G() -> i32 {
+    return A.SOME_CONSTANT;
+  }
+
+  fn H() -> i32 {
+    return A.SomeProtectedFunction();
+  }
+}
+
+
```
jonmeow:
```suggestion

```

Try to only put one line between files and code.

toolchain/check/context.cpp
```diff
@@ -306,37 +309,138 @@ auto Context::LookupUnqualifiedName(Parse::NodeId node_id,
...
+  CARBON_DIAGNOSTIC(ClassInvalidMemberAccess, Error,
+                    "Cannot access {0} member `{1}` of type `{2}`.",
+                    std::string_view, SemIR::NameId, SemIR::TypeId);
+  CARBON_DIAGNOSTIC(ClassMemberDefinition, Note,
+                    "The {0} member `{1}` is defined here.", std::string_view,
+                    SemIR::NameId);
+
+  auto parent_type_id = class_info.self_type_id;
+  std::string_view access_desc =
+      access_kind == SemIR::AccessKind::Private ? "private" : "protected";
```
```suggestion
  auto access_desc =
      access_kind == SemIR::AccessKind::Private ? llvm::StringLiteral("private") : llvm::StringLiteral("protected");
```

StringLiteral has constexpr support, so it's preferred when working with constant expressions such as this/

toolchain/check/context.cpp
```diff
@@ -306,37 +309,138 @@ auto Context::LookupUnqualifiedName(Parse::NodeId node_id,
...
+
 auto Context::LookupQualifiedName(Parse::NodeId node_id, SemIR::NameId name_id,
-                                  LookupScope scope, bool required)
+                                  LookupScope scope, bool required,
+                                  std::optional<AccessInfo> access_info)
     -> LookupResult {
   llvm::SmallVector<LookupScope> scopes = {scope};
+
+  // TODO: Support reporting of multiple prohibited access.
+  llvm::SmallVector<ProhibitedAccessInfo> prohibited_accesses = {};
```
```suggestion
  llvm::SmallVector<ProhibitedAccessInfo> prohibited_accesses;
```

There's no need to explicitly initialize as empty.

toolchain/check/context.cpp
```diff
@@ -306,37 +309,138 @@ auto Context::LookupUnqualifiedName(Parse::NodeId node_id,
...
+            class_info.base_id)) {
+      parent_type_id = base_decl->base_type_id;
+    }
+    if (auto adapt_decl = context.insts().TryGetAsIfValid<SemIR::AdaptDecl>(
+            class_info.adapt_id)) {
+      parent_type_id = adapt_decl->adapted_type_id;
+    }
+
+    // Make sure that `parent_type_id` is replaced if this is a parent access.
+    CARBON_CHECK(parent_type_id != class_info.self_type_id);
```
```suggestion
    if (auto base_decl = context.insts().TryGetAsIfValid<SemIR::BaseDecl>(
            class_info.base_id)) {
      parent_type_id = base_decl->base_type_id;
    } else if (auto adapt_decl = context.insts().TryGetAsIfValid<SemIR::AdaptDecl>(
            class_info.adapt_id)) {
      parent_type_id = adapt_decl->adapted_type_id;
    } else {
      CARBON_FATAL() << "Expected parent for parent access";
    }
```

I think it'd help here to use `else` to error directly.

toolchain/check/context.cpp
```diff
@@ -368,9 +473,24 @@ auto Context::LookupQualifiedName(Parse::NodeId node_id, SemIR::NameId name_id,
...
   if (required && !result.inst_id.is_valid()) {
-    if (!has_error) {
+    if (has_error) {
+      return {.specific_id = SemIR::SpecificId::Invalid,
+              .inst_id = SemIR::InstId::BuiltinError};
+    }
+
+    if (prohibited_accesses.empty()) {
       DiagnoseNameNotFound(node_id, name_id);
+    } else if (access_info.has_value()) {
```
jonmeow: Does this need to be checked, given prohibited_accesses is non-empty? i.e.:

```suggestion
    } else {
```

toolchain/check/context.cpp
```diff
@@ -368,9 +473,24 @@ auto Context::LookupQualifiedName(Parse::NodeId node_id, SemIR::NameId name_id,
   }
 
   if (required && !result.inst_id.is_valid()) {
-    if (!has_error) {
+    if (has_error) {
+      return {.specific_id = SemIR::SpecificId::Invalid,
+              .inst_id = SemIR::InstId::BuiltinError};
+    }
```
jonmeow: What do you think of restoring the `if (!has_error)` flow? Previously, it mean that the return value was not duplicated. I think that was probably helpful, in that in emphasized the return value was not associated with a specific diagnostic; it's also a somewhat complex return, so only having one copy is also kind of nice.

toolchain/check/context.cpp
```diff
@@ -306,37 +309,138 @@ auto Context::LookupUnqualifiedName(Parse::NodeId node_id,
...
+
+  // TODO: Support scoped entities other than just classes.
+  auto class_info = context.classes().Get(class_type->class_id);
+
+  CARBON_DIAGNOSTIC(ClassInvalidMemberAccess, Error,
+                    "Cannot access {0} member `{1}` of type `{2}`.",
+                    std::string_view, SemIR::NameId, SemIR::TypeId);
+  CARBON_DIAGNOSTIC(ClassMemberDefinition, Note,
+                    "The {0} member `{1}` is defined here.", std::string_view,
+                    SemIR::NameId);
```
```suggestion
  // TODO: Support passing AccessKind to diagnostics.
  CARBON_DIAGNOSTIC(ClassInvalidMemberAccess, Error,
                    "Cannot access {0} member `{1}` of type `{2}`.",
                    llvm::StringLiteral, SemIR::NameId, SemIR::TypeId);
  CARBON_DIAGNOSTIC(ClassMemberDefinition, Note,
                    "The {0} member `{1}` is defined here.", llvm::StringLiteral,
                    SemIR::NameId);
```

We should probably support AccessKind directly, but per my above comment about string_view, StringLiteral is okay for now.

jonmeow (commented):
Thanks! Just a few last things.

Since this is almost ready to merge, can you also update against trunk? Looking at GitHub, there are a few files with merge conflicts.

toolchain/check/context.cpp
jonmeow: It looks like you've retained a string_view use, can you please take another look?

jonmeow (approved):
This looks basically ready to merge, but I see one string_view use got missed. Can you please fix the argument to ClassInvalidMemberAccess?

toolchain/check/context.cpp
brymer-meneses: Done, sorry about that.

jonmeow (approved):
Thanks! This is great. I'll kick off a merge. :)

PR #4249, opened 2024-08-26
Fix crash when disabling debug info

dwblaikie:
The source line debug info generation assumed that the function would have debug info. Check for a non-null di_subprogram_ to ensure we are emitting debug info for the function.

Rather than checking the di_builder_ - this way if we implement `nodebug` function attributes, it'll fall out naturally (by creating a null di_subprogram_) rather than having to come back and change this from "is debug info enabled" to "is debug info enabled for this function" later on.


jonmeow (approved):
Good catch!

PR #4246, opened 2024-08-23
Getting commit access

jonmeow:
Establish a process for getting commit access. We will:

-   Grant access based on a developer's commit history.
    -   Someone with commit access should nominate, and a contributor may ask.
    -   A lead will approve nominations. Only one lead is needed.
-   Remove commit access once someone is idle for 6 months.
    -   "Idle" means no significant project activity on any of GitHub, Discord,
        or in meetings.
    -   Access removed due to being idle will be restored on request.

PR #4250, opened 2024-08-26
Update group documentation

jonmeow:
I'm trying to use this just to document the de facto state. This is linked to #4246, but I'm trying to keep the two arranged to allow independent merges.


chandlerc (approved):
LG, matches current state so seems good to land as is.

PR #4251, opened 2024-08-26
debug-info: Add checks for instructions without source locations

dwblaikie:
I think this is what you were looking for, @jonmeow? (I was trying to figure out how to tersely phrase the check to be "this isn't supported here yet, and not clear if it needs to be supported here - but if this assert is hit, then we should add support here?" - open to wording suggestions)

(this will fail testing due to #4249 - so shouldn't be submitted before that, though they aren't textually/patch-ually dependent on each other)

### DIFF
--- a/toolchain/lower/function_context.cpp
+++ b/toolchain/lower/function_context.cpp
@@ -98,6 +98,11 @@ auto FunctionContext::LowerInst(SemIR::InstId inst_id) -> void {
   CARBON_VLOG() << "Lowering " << inst_id << ": " << inst << "\n";
   builder_.getInserter().SetCurrentInstId(inst_id);
   auto loc = file_context_->GetDiagnosticLoc(inst_id);
+  CARBON_CHECK(loc.filename == di_subprogram_->getFile()->getFilename())
+      << "Instructions located in a different file from their enclosing "
+         "function aren't handled yet";
+  CARBON_CHECK(loc.line_number >= 1 && loc.column_number >= 1)
+      << "Instructions without line locations aren't handled yet";
   builder_.SetCurrentDebugLocation(
       llvm::DILocation::get(builder_.getContext(), loc.line_number,
                             loc.column_number, di_subprogram_));


jonmeow:
@dwblaikie Can you please fix your commit message to not @ me? I will already get notifications from PRs, I don't need notifications from individual commits too. i.e., please fix: https://github.com/dwblaikie/carbon-lang/commit/8134dfb13167427b05dbdaf4a9da691db4625b24

I would generally encourage using PR comments to @ people rather than commits, since the PR is where discussion will occur.

jonmeow:
BTW, per josh11b at https://github.com/carbon-language/carbon-lang/pull/4247#discussion_r1731711169, I think we can't just add these CHECKs and the line number issue I was noting is already a problem. There's more going on here so I'm poking around.

toolchain/lower/function_context.cpp
```diff
@@ -98,6 +98,11 @@ auto FunctionContext::LowerInst(SemIR::InstId inst_id) -> void {
   CARBON_VLOG() << "Lowering " << inst_id << ": " << inst << "\n";
   builder_.getInserter().SetCurrentInstId(inst_id);
   auto loc = file_context_->GetDiagnosticLoc(inst_id);
+  CARBON_CHECK(loc.filename == di_subprogram_->getFile()->getFilename())
+      << "Instructions located in a different file from their enclosing "
+         "function aren't handled yet";
+  CARBON_CHECK(loc.line_number >= 1 && loc.column_number >= 1)
+      << "Instructions without line locations aren't handled yet";
```
jonmeow, line 105: FYI, I think this isn't catching the existing issues because it's not used everywhere GetDiagnosticLoc is.

jonmeow:
FYI, #4252 with a different approach on the line numbers, which looks like it fixes the current issues.

toolchain/lower/function_context.cpp
jonmeow: To be sure, what I was noting here was that file_context.cpp also queries ConvertLoc. Since it won't use this code, the function missing a line number could be missed. That is, this would miss:

```
// CHECK:STDOUT: !4 = distinct !DISubprogram(name: "__global_init", scope: null, file: !3, line: 4294967295, type: !5, spFlags: DISPFlagDefinition, unit: !2)
```

#4252 fixes this.

dwblaikie:
> FYI, #4252 with a different approach on the line numbers, which looks like it fixes the current issues.

Works for me - I'll abandon/delete this change. Thanks!

dwblaikie:
> @dwblaikie Can you please fix your commit message to not @ me? I will already get notifications from PRs, I don't need notifications from individual commits too. i.e., please fix: [dwblaikie@8134dfb](https://github.com/dwblaikie/carbon-lang/commit/8134dfb13167427b05dbdaf4a9da691db4625b24)
> 
> I would generally encourage using PR comments to @ people rather than commits, since the PR is where discussion will occur.

Ah, yeah - sorry, old habit from LLVM where commit message -> description in Phabricator I guess, without the extra layer of commits nested within a review. Will try to avoid writing review descriptions in commits in the future.

jonmeow:
> Ah, yeah - sorry, old habit from LLVM where commit message -> description in Phabricator I guess, without the extra layer of commits nested within a review. Will try to avoid writing review descriptions in commits in the future.

Thanks! Other aspect of this with GitHub, I would've also gotten a notification when the commit was merged. I really wish GH gave me a way of narrowly disabling that.

PR #4252, opened 2024-08-26
Handle unknown lines in DebugInfo

jonmeow:
Mainly, changes the default from -1 to 0 in DiagnosticLoc, still trying to keep reusing that. Nothing except for the lowered output is affected, so I think this is fine.

Also, have lowering consistently call GetDiagnosticLoc.

Pulls in one of the CHECKs suggested from #4251 

Co-authored-by: David Blaikie <dblaikie@gmail.com>


jonmeow:
FYI @dwblaikie

toolchain/diagnostics/diagnostic.h
```diff
@@ -58,10 +58,11 @@ struct DiagnosticLoc {
   llvm::StringRef filename;
   // A reference to the line of the error.
   llvm::StringRef line;
-  // 1-based line number.
-  int32_t line_number = -1;
+  // 1-based line number. This defaults to 0 because LLVM's DebugInfo uses that
+  // for unknown values, and we're sharing the logic.
+  int32_t line_number = 0;
```
dwblaikie: I certainly don't mind mapping -1 -> 0 in the debug info lowering, rather than sharing this representational detail.

Though independently of that, I'd favor 0 as the invalid value (& maybe use an unsigned integer here, though that's probably more contentious) if it's not an otherwise valid value - then it's clear there's only one invalid value (0) and all other values are valid. Rather than wondering if 0 encodes something different from -1 (or -2), etc.

dwblaikie (approved):
(not sure what the conventions are for approval - since the author has commit rights, whether you just wanted my domain-specific review, or if I should leave approval to someone with commit rights/some other authority?)

Change looks OK to me (one minor side comment/thought).

The output changes look OK/consistent with how Clang generates debug info source locations for global constructors. (the global ctor function itself has no line, but the instructions inside it do point to the relevant parts of the initializing subexpressions)

dwblaikie (approved):
(thought I picked the approve button, but guess got sidetracked by my own rambling considerations in the previous message)

ooh, I did pick approve, but it doesn't highlight green because of my permissions - then I didn't notice the subtle grey checkmark. Don't mind me.

toolchain/diagnostics/diagnostic.h
jonmeow: FWIW, I'm really tempted to split logic just to retain -1.

The result of `-1` as `line: 4294967295` gave a strong hint as to which code was both receiving and failing to handle the invalid value. I don't think `0` will do as well, just because it's a more reasonable unsigned value, and blends in better. Also, `-1` is more familiar to me as a common, invalid value.

With `-1`, comments can still address other values, like just saying that `0` is unused.

dwblaikie: I'm good with that - though I'd like it to be documented that the invalid value is -1, positive values are expected, and nothing else is expected. And having checks == -1 exactly, rather than <= 0 - the latter makes me a bit uncomfortable "what values are there?". -1 is clearer for me to read.

jonmeow: Per discussion, switched to -1 and renamed GetDiagnosticLoc to GetLocForDI

dwblaikie (approved):
Looks good to me - thanks!

PR #4253, opened 2024-08-26
DebugInfo: Add the name as-written, separate from the mangled name

dwblaikie:
There's not much mangling happening yet - but Run -> main (and some overloading numbering happening, maybe LLVM is doing that 'helpfully' under the hood?) is enough to demonstrate this improvement/fix.

Ah, here it is:
```
#0  llvm::ValueSymbolTable::makeUniqueName (this=0x50287fe5b6c0, V=0x50287fe827e8, UniqueName="F") at external/_main~llvm_project~llvm-project/llvm/lib/IR/ValueSymbolTable.cpp:45
#1  0x000055555c40f964 in llvm::ValueSymbolTable::reinsertValue (this=0x50287fe5b6c0, V=0x50287fe827e8) at external/_main~llvm_project~llvm-project/llvm/lib/IR/ValueSymbolTable.cpp:100
#2  0x000055555c2a91df in llvm::SymbolTableListTraits<llvm::Function>::addNodeToList (this=0x50287fd16f18, V=0x50287fe827e8) at external/_main~llvm_project~llvm-project/llvm/lib/IR/SymbolTableListTraitsImpl.h:75
#3  0x000055555c2a90e5 in llvm::iplist_impl<llvm::simple_ilist<llvm::Function>, llvm::SymbolTableListTraits<llvm::Function> >::insert (this=0x50287fd16f18, where=..., New=0x50287fe827e8)
    at external/_main~llvm_project~llvm-project/llvm/include/llvm/ADT/ilist.h:166
#4  0x000055555c27fef2 in llvm::iplist_impl<llvm::simple_ilist<llvm::Function>, llvm::SymbolTableListTraits<llvm::Function> >::push_back (this=0x50287fd16f18, val=0x50287fe827e8) at external/_main~llvm_project~llvm-project/llvm/include/llvm/ADT/ilist.h:250
#5  0x000055555c27faeb in llvm::Function::Function (this=0x50287fe827e8, Ty=0x50287fd43058, Linkage=llvm::GlobalValue::ExternalLinkage, AddrSpace=0, name="F", ParentModule=0x50287fd16f00) at external/_main~llvm_project~llvm-project/llvm/lib/IR/Function.cpp:521
#6  0x0000555559441f95 in llvm::Function::Create (Ty=0x50287fd43058, Linkage=llvm::GlobalValue::ExternalLinkage, AddrSpace=0, N="F", M=0x50287fd16f00) at external/_main~llvm_project~llvm-project/llvm/include/llvm/IR/Function.h:175
#7  0x000055555c27ebac in llvm::Function::Create (Ty=0x50287fd43058, Linkage=llvm::GlobalValue::ExternalLinkage, N="F", M=...) at external/_main~llvm_project~llvm-project/llvm/lib/IR/Function.cpp:398
#8  0x0000555558ce7bb5 in Carbon::Lower::FileContext::BuildFunctionDecl (this=0x7fffffffc438, function_id=...) at toolchain/lower/file_context.cpp:257
```

That's where LLVM decides to make a new name (name.number) when asked to create a new global with the same name as an existing global.

It's not a valid mangling scheme - since the name won't be stable between different compilations, but it is enough to make single-compilation code build/run for now.


toolchain/lower/file_context.cpp
```diff
@@ -373,10 +373,18 @@ auto FileContext::BuildDISubprogram(const SemIR::Function& function,
   auto loc = converter_.ConvertLoc(
       function.definition_id,
       [](DiagnosticLoc, const Internal::DiagnosticBase<>&) {});
+  llvm::StringRef name;
+  if (function.name_id.is_valid()) {
+    name = *sem_ir().names().GetAsStringIfIdentifier(function.name_id);
```
jonmeow: BuildFunctionDecl does similar, had you considered refactoring to use the same name in both places?

dwblaikie: Like refactor out a helper function to call from both places, or pass the name, retrieved in `BuildFunctionDecl` along through the decl to the def building somehow and then down into `BuildDISubprogram`?

It seemed short enough it wasn't necessarily worth a helper function - but I'm certainly open to it, if that's what you had in mind?

Sharing the value from `BuildFunctionDecl` into the def building code and into the BuildDISubprogram seemed a bit more difficult (because the decl might be built some time before/pretty distant from the definition, I assumed?)

jonmeow: Sorry, I think I misunderstood exactly what you were trying to do with this name. I'm realizing two things:

1) You're trying to _avoid_ using a mangled name here, right? Why does this compare with `linkage_name`, versus unconditionally setting the value? Maybe adding comments might be helpful?

2) Have you looked at GetFormatted instead of GetAsStringIfIdentifier?

    - In particular, can you add a test with `fn r#self() {}`? I think we haven't really tested with raw identifiers yet, I'm wondering if it's going to show here.

dwblaikie: 1. Added comment in 0eaa1ecddea8f51aaf50b72dd7c1b8c11bb5a771 - we only want the linkage name if it differs from the source name.

2. Thanks for the offline discussion - adjusted the existing name query/checking to be a bit clearer in 4c9021cbc8f9f50790d72f04aedc25aa70e7291e and added test coverage for raw identifiers in functions (which seems to come out naturally from the identifier handling - so isn't dependent on this patch or other changes) in #4257

jonmeow: > 1. Added comment in [0eaa1ec](https://github.com/carbon-language/carbon-lang/commit/0eaa1ecddea8f51aaf50b72dd7c1b8c11bb5a771) - we only want the linkage name if it differs from the source name.

Per discussion, let's get rid of the conditional since it's mainly going to affect "extern C"-style things, which are atypical in Carbon and probably okay to store a name for. For a while, we'll just have name mangling and nothing will match.

dwblaikie: done

toolchain/lower/file_context.cpp
```diff
@@ -374,14 +374,11 @@ auto FileContext::BuildDISubprogram(const SemIR::Function& function,
       function.definition_id,
       [](DiagnosticLoc, const Internal::DiagnosticBase<>&) {});
   llvm::StringRef name;
-  if (function.name_id.is_valid()) {
-    name = *sem_ir().names().GetAsStringIfIdentifier(function.name_id);
+  if (auto maybe_name =
+          sem_ir().names().GetAsStringIfIdentifier(function.name_id)) {
+    name = *maybe_name;
   }
```
jonmeow: I'd thought you were going to make this CHECK it had a name, i.e., something like:

```suggestion
  auto name = *sem_ir().names().GetAsStringIfIdentifier(function.name_id);
  CARBON_CHECK(name) << "Unexpected special name for function: "
                     << function.name_id;
```

dwblaikie:
Hold off on this got this a bit messed up with some mangler prototyping leaking in here - don't mind me while I juggle git some more...

toolchain/lower/file_context.cpp
dwblaikie: Ah, right - yep, done.

toolchain/lower/file_context.cpp
```diff
@@ -373,10 +372,15 @@ auto FileContext::BuildDISubprogram(const SemIR::Function& function,
   auto loc = converter_.ConvertLoc(
       function.definition_id,
       [](DiagnosticLoc, const Internal::DiagnosticBase<>&) {});
-  // FIXME: Add more details here, including mangled name, real subroutine type
-  // (once type information is built), etc.
+  auto opt_name = sem_ir().names().GetAsStringIfIdentifier(function.name_id);
+  CARBON_CHECK(opt_name) << "Unexpected special name for function: "
+                         << function.name_id;
+  llvm::StringRef name = *opt_name;
+  llvm::StringRef linkage_name = llvm_function->getName();
```
jonmeow: Is it still worth having separate variables for these?

dwblaikie: Oh, rather than rolling the expressions into their use? Sure - done!

jonmeow (approved):
Thanks! Sorry this took a bit to work through.

Note the mangler files are still in the PR, so you'll get one more comment: opt_name -> name? (I assume you were doing opt_name due to the StringRef name, but in this PR it contrasts with name used for the other GetAsStringIfIdentifier)

dwblaikie:
> Thanks! Sorry this took a bit to work through.

All good - appreciate your patience.

> Note the mangler files are still in the PR,

Oh, yeah, wrangling git... OK, force push to remove the commit that added that in.

> so you'll get one more comment: opt_name -> name? (I assume you were doing opt_name due to the StringRef name, but in this PR it contrasts with name used for the other GetAsStringIfIdentifier)

Done.

jonmeow (approved):
Thanks! FYI, I think this lost the semi-related change to CHECK in the other name handling, but since you're looking at name mangling already I think it makes sense to just move forward here and deal with that separately.

PR #4254, opened 2024-08-27
Parsing of designators like `.x` or `.Self`

josh11b:
These appear in `where` clauses, as in:

```
U:! InterfaceB where .C = Vector(.D)
V:! type where Vector(.Self) impls Sortable
```

`.Self` can additionally appear in type expressions in a binding pattern such as `T:! InterfaceA(.Self)`.


toolchain/parse/handle_expr.cpp
```diff
@@ -169,6 +169,42 @@ auto HandleExprInPostfix(Context& context) -> void {
       context.PushState(state);
       break;
     }
+    case Lex::TokenKind::Period: {
```
jonmeow, line 172: Please add tests of this in normal expressions, in check testdata. Even with the TODOs, I think it'd help improve confidence that valid parse trees can't cause crashes in the check phase (and eventually, will need to be handled).

e.g., as a few examples:

```
// --- fail_designated_return.carbon

fn Foo() -> () {
  var x: ();
  return .x;
}

// --- fail_unknown_designator.carbon

fn Foo() -> () {
  return .undef;
}

// --- fail_self_designator.carbon

class C {
  fn F() -> Self { return .Self; }
}

// --- fail_self_designator_in_type.carbon

class C {
  fn F() -> .Self { return Self; }
}
```

(note, `()` so that this fits into the `no_prelude` model)

toolchain/parse/handle_expr.cpp
```diff
@@ -169,6 +169,42 @@ auto HandleExprInPostfix(Context& context) -> void {
...
+      if (context.ConsumeAndAddLeafNodeIf(Lex::TokenKind::Identifier,
+                                          NodeKind::IdentifierName)) {
+        // OK, `.` identifier.
+      } else if (context.ConsumeAndAddLeafNodeIf(
+                     Lex::TokenKind::SelfTypeIdentifier, NodeKind::DotSelf)) {
+        // OK, `.Self`.
+      } else {
+        // TODO: Custom diagnostic for `.1`? Or treat it as a designator for a
+        // tuple element?
+        CARBON_DIAGNOSTIC(ExpectedIdentifierOrSelfAfterDot, Error,
```
jonmeow: Had you considered calling it a "Period" instead of a "Dot" in your changes, mirroring the token name? I see "Dot" in "DotOrArrow", but that looks like the exception and could probably be fixed to use "Period" consistent with other code.

toolchain/parse/typed_nodes.h
```diff
@@ -1016,6 +1016,21 @@ struct IfExprElse {
   AnyExprId else_result;
 };
 
+// The `Self` in `.Self`
+// FIXME: SelfDesignator?
```
jonmeow: Had you considered something like `SelfTypeName`, mirroring `SelfTypeIdentifier` and `BaseName` handling?

There's also the similar `SelfTypeNameExpr`, although that evaluates, which you don't want here (not sure if there's a good thing to rename `SelfTypeNameExpr` to, to better represent the evaluation aspect).

Note, elsewhere I believe we're doing SelfType or SelfValue to disambiguate between "Self" and "self"

toolchain/parse/typed_nodes.h
```diff
@@ -1016,6 +1016,21 @@ struct IfExprElse {
...
+
+// `.Member` or `.Self` in an expression context, used in `where` and `require`
+// clauses.
+// TODO: Do we want to support `.1`, a designator for accessing a tuple member?
+struct DesignatorExpr {
+  static constexpr auto Kind = NodeKind::DesignatorExpr.Define(
+      {.category = NodeCategory::Expr, .child_count = 1});
+
+  Lex::PeriodTokenIndex token;
+  NodeIdOneOf<IdentifierName, DotSelf> designator;
```
jonmeow: Sorry, pedantic terminology question: is this the designator, or designatee?

toolchain/parse/handle_expr.cpp
```diff
@@ -169,6 +169,42 @@ auto HandleExprInPostfix(Context& context) -> void {
...
+      auto period = context.Consume();
+      if (context.ConsumeAndAddLeafNodeIf(Lex::TokenKind::Identifier,
+                                          NodeKind::IdentifierName)) {
+        // OK, `.` identifier.
+      } else if (context.ConsumeAndAddLeafNodeIf(
+                     Lex::TokenKind::SelfTypeIdentifier, NodeKind::DotSelf)) {
+        // OK, `.Self`.
+      } else {
+        // TODO: Custom diagnostic for `.1`? Or treat it as a designator for a
+        // tuple element?
```
jonmeow: FWIW I'd expect either the current diagnostic is fine, or `.1` should be supported. Note you already have the TODO on `DesignatorExpr` though, so not sure you need the TODO here.

toolchain/parse/handle_expr.cpp
```diff
@@ -169,6 +169,42 @@ auto HandleExprInPostfix(Context& context) -> void {
...
+                              /*has_error=*/true);
+        } else if (context.PositionIs(Lex::TokenKind::IntLiteral)) {
+          context.AddLeafNode(NodeKind::InvalidParse, context.Consume(),
+                              /*has_error=*/true);
+        } else {
+          context.AddLeafNode(NodeKind::InvalidParse, *context.position(),
+                              /*has_error=*/true);
+          // Indicate the error to the parent state so that it can avoid
+          // producing more errors.
+          context.ReturnErrorOnState();
```
jonmeow: Why is ReturnError called on this path, but not the other error paths? Is there more information that you could add to the comment?

toolchain/parse/testdata/where_expr/designators.carbon
```diff
@@ -0,0 +1,85 @@
...
+  // CHECK:STDERR:    ^
+  // CHECK:STDERR:
+  .;
+
+  // CHECK:STDERR: fail_designator.carbon:[[@LINE+3]]:4: ERROR: Expected identifier or `Self` after `.`.
+  // CHECK:STDERR:   .*;
+  // CHECK:STDERR:    ^
+  .*;
+}
+
```
jonmeow, line 57:
```suggestion

// --- fail_designator_eof.carbon

var x: i32 = .
```

We should probably check when the next token is EOF.

toolchain/parse/testdata/where_expr/designators.carbon
```diff
@@ -0,0 +1,85 @@
...
+  // CHECK:STDERR:   .;
+  // CHECK:STDERR:    ^
+  // CHECK:STDERR:
+  .;
+
+  // CHECK:STDERR: fail_designator.carbon:[[@LINE+3]]:4: ERROR: Expected identifier or `Self` after `.`.
+  // CHECK:STDERR:   .*;
+  // CHECK:STDERR:    ^
+  .*;
+}
```
jonmeow: Had you considered giving each of these their own split file? Because the file's "fail" check only happens at that level, the presence of a diagnostic is only validated once (one of these could start passing, and after an autoupdate, the test would still pass). By giving each their own split, the fact that each issues a diagnostic would be tested.

toolchain/parse/handle_expr.cpp
```diff
@@ -169,6 +169,42 @@ auto HandleExprInPostfix(Context& context) -> void {
       context.PushState(state);
       break;
     }
+    case Lex::TokenKind::Period: {
+      // `.Member` or `.Self`
```
jonmeow: For readability, could this comment be rephrased as a sentence describing what's important about these?

jonmeow (commented):
The logic here looks good. I'm still suggesting a few more tests to cover edge cases though, particularly for check behavior. The TODOs should handle it, but it's nice to be sure.

toolchain/parse/testdata/where_expr/designators.carbon
josh11b: Done.

toolchain/parse/handle_expr.cpp
josh11b: Done.

toolchain/parse/handle_expr.cpp
josh11b: Removed TODO.

toolchain/parse/handle_expr.cpp
josh11b: Added more to the comment.

toolchain/parse/handle_expr.cpp
josh11b: Added.

toolchain/parse/handle_expr.cpp
josh11b: Done.

toolchain/parse/typed_nodes.h
josh11b: Used `SelfTypeName`.

toolchain/parse/typed_nodes.h
josh11b: Switched to `name`.

jonmeow (approved):
Thanks, looks good!

There's a conflict in diagnostic_kind.def, but feel free to merge after resolving.

PR #4258, opened 2024-08-27
Fix indirect import handling for functions.

jonmeow:
The particular test this focused on is indirect_two_file in toolchain/check/testdata/function/definition/no_prelude/extern_library.carbon.

This removes `parent_scope_id_for_new_inst` because I think it's returning unhelpful results. The use was at the root of incorrect results for the indirect import chain. `name_id_for_new_inst` is actually wrapping a union, so it's more important.

The merging of `is_extern` and `first_owning_decl_id` in `handle_function.cpp` feels like it's less correct with the changes that've been made to `extern`. This ripples in tests, because the error recovery shifts.

### DIFF
--- a/toolchain/check/handle_function.cpp
+++ b/toolchain/check/handle_function.cpp
@@ -97,7 +97,6 @@ static auto MergeFunctionRedecl(Context& context, SemIRLoc new_loc,
     prev_function.MergeDefinition(new_function);
     prev_function.return_storage_id = new_function.return_storage_id;
   }
-  // The new function might have return slot information if it was imported.
   if ((prev_import_ir_id.is_valid() && !new_is_import)) {
     ReplacePrevInstForMerge(context, new_function.parent_scope_id,
                             prev_function.name_id,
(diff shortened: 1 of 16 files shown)


toolchain/check/handle_function.cpp
```diff
@@ -87,18 +88,18 @@ static auto MergeFunctionRedecl(Context& context, SemIRLoc new_loc,
...
+  if (!prev_function.first_owning_decl_id.is_valid()) {
+    prev_function.first_owning_decl_id = new_function.first_owning_decl_id;
+  }
   if (new_is_definition) {
     // Track the signature from the definition, so that IDs in the body
     // match IDs in the signature.
     prev_function.MergeDefinition(new_function);
     prev_function.return_storage_id = new_function.return_storage_id;
   }
   // The new function might have return slot information if it was imported.
```
zygoloid:
```suggestion
```
I think this comment got left behind when some other code got deleted -- it doesn't seem to make sense here.

zygoloid (approved):

PR #4261, opened 2024-08-27
Fix parse to use the error tracking consumer for has_errors_.

### LINKED ISSUE #4259: Recovered parse errors should fail compilation
### Description of the bug:

Some errors, such as missing or additional whitespace next to an operator, are recovered by the compiler. In this case, the compiler outputs a diagnostic, but no parse nodes are marked as having an error, and so the compilation succeeds.

This problem can be seen in any of the files starting with `recover_` in the
[`toolchain/parse/testdata/operators/` directory](https://github.com/carbon-language/carbon-lang/tree/trunk/toolchain/parse/testdata/operators).

 Those tests should only pass with a filename (or split filename) starting with `fail_`. We need some way to indicate an error in the parse output (like I believe exists for the lex output) even when there are no nodes with an error in the output parse tree.

### What actually happened?

Those files have diagnostics with a `CHECK:STDERR` prefix, but the parse stage succeeds.

jonmeow:
This is how we are setting has_errors_ in other stages; this should only make parse consistent.

Fixes #4259

### DIFF
--- a/toolchain/parse/parse.cpp
+++ b/toolchain/parse/parse.cpp
@@ -20,7 +20,8 @@ auto HandleInvalid(Context& context) -> void {
 auto Parse(Lex::TokenizedBuffer& tokens, DiagnosticConsumer& consumer,
            llvm::raw_ostream* vlog_stream) -> Tree {
   Lex::TokenDiagnosticConverter converter(&tokens);
-  Lex::TokenDiagnosticEmitter emitter(converter, consumer);
+  ErrorTrackingDiagnosticConsumer err_tracker(consumer);
+  Lex::TokenDiagnosticEmitter emitter(converter, err_tracker);
 
   // Delegate to the parser.
   Tree tree(tokens);
@@ -44,6 +45,7 @@ auto Parse(Lex::TokenizedBuffer& tokens, DiagnosticConsumer& consumer,
   }
 
   context.AddLeafNode(NodeKind::FileEnd, *context.position());
+  tree.set_has_errors(err_tracker.seen_error());
 
   if (auto verify = tree.Verify(); !verify.ok()) {
     // TODO: This is temporarily printing to stderr directly during development.
--- a/toolchain/parse/tree.h
+++ b/toolchain/parse/tree.h
@@ -103,7 +103,6 @@ class Tree : public Printable<Tree> {
     node_impls_.reserve(tokens_->expected_parse_tree_size());
   }
 
-  // Tests whether there are any errors in the parse tree.
   auto has_errors() const -> bool { return has_errors_; }
 
   auto set_has_errors(bool has_errors) -> void { has_errors_ = has_errors; }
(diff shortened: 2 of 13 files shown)


jonmeow:
Depends on #4260 (merged)

toolchain/parse/parse.cpp
```diff
@@ -44,6 +45,7 @@ auto Parse(Lex::TokenizedBuffer& tokens, DiagnosticConsumer& consumer,
   }
 
   context.AddLeafNode(NodeKind::FileEnd, *context.position());
+  tree.set_has_errors(err_tracker.seen_error());
```
geoffromer, line 48: Checking my understanding here:
- On the tree as a whole, `has_errors` indicates whether the input was syntactically invalid.
- On individual parse nodes, `has_error` indicates whether the node satisfies its kind's invariants, and may be false even if it corresponds to a syntactically invalid input.

If that's right, we should probably rename `has_error` on nodes to something like `well_formed`, but that's out of scope for this PR.

toolchain/parse/tree.h
```diff
@@ -106,6 +106,8 @@ class Tree : public Printable<Tree> {
   // Tests whether there are any errors in the parse tree.
```
geoffromer: While we're here I think it would be helpful to clarify what exactly constitutes an error in this context, e.g.

```
// Returns true if this tree represents a syntactically invalid input.
```

(Sorry, GH won't let me suggest an edit here)

toolchain/parse/tree.h
```diff
@@ -106,6 +106,8 @@ class Tree : public Printable<Tree> {
...
   auto has_errors() const -> bool { return has_errors_; }
 
+  auto set_has_errors(bool has_errors) -> void { has_errors_ = has_errors; }
```
geoffromer, line 108: Might it make sense to call this method `Finalize`, to make explicit that it signifies the point at which `*this` becomes immutable, and should only be called once?

toolchain/parse/parse.cpp
jonmeow: Maybe I'd do a separate pass at refactoring both, but I actually think a node's `has_error` is the more correct one. The node has an error in its immediate structure (non-transitive from children), typically because parse encountered a problem it couldn't recover from.

OTOH, `has_errors` can include any diagnostic that should prevent lowering. That goes beyond invalid syntax (and `Error` diagnostics). It's probably more the one that should be renamed, but Lex and Check both use the same name, so similar scope issue.

toolchain/parse/tree.h
jonmeow: For contrast with SemIR, we have SemIR::File::set_has_errors, and Check::Context::Finalize. Here, I'm adding Parse::Tree::Set_has_errors, but it didn't seem worth adding a Parse::Context::Finalize (Check::Context::Finalize also doesn't call SemIR::File::set_has_errors).

toolchain/parse/tree.h
jonmeow: I've removed this comment, and worked on the comment on has_errors_ (I think we typically don't have comments on accessors like this, and I'd rather only have one comment to update). Trying to also reflect better my comments above about the flag's semantic.

geoffromer (approved):

PR #4264, opened 2024-08-28
Enforce that the parse node for an instruction has the kind specified in the instruction definition

zygoloid:
Remove `ReusingLoc` and add enforcement that even for imported locations, the kind of the parse node for an instruction matches the kind specified in the instruction definition.

Change the node kind for a few instructions to `NodeId`:

- A couple of instructions had a typed node but could be created implicitly with any node as part of a builtin implicit conversion. This happened for `AddrOf`, `ArrayIndex`, and `Deref`.
- A bunch of instructions had `InvalidNodeId` as their associated parse node kind but were actually always created with a location.


toolchain/sem_ir/inst.h
```diff
@@ -305,19 +305,19 @@ inline auto operator<<(llvm::raw_ostream& out, TypedInst inst)
...
-  // Constructs a LocIdAndInst that reuses the location associated with some
-  // other inst, typically because `inst` doesn't have an explicit
-  // representation in the parse tree.
-  template <typename InstT>
-  static auto ReusingLoc(LocId loc_id, InstT inst) -> LocIdAndInst {
-    return LocIdAndInst(loc_id, inst, /*is_untyped=*/true);
+  // Unsafely form a pair of a location and an instruction. Used in the cases
+  // where we can't statically enforce the type matches.
+  static auto UncheckedLoc(LocId loc_id, Inst inst) -> LocIdAndInst {
+    return LocIdAndInst(loc_id, inst, /*is_unchecked=*/true);
```
jonmeow, line 320: I'm only seeing this called from `MakeImportedLocAndInst` and `GetWithLocId`. Am I missing more general `LocId` uses? Had you considered keeping the `ImportIRInstId`-specific constructor in order to prevent use with non-import use-cases?

toolchain/sem_ir/inst.h
```diff
@@ -338,8 +339,8 @@ struct LocIdAndInst {
   // Expose the internal constructor for GetWithLocId.
   friend class InstStore;
```
jonmeow: Acknowledging my above comment would moot this, noting you currently have `GetWithLocId` using `UncheckedLoc`, is this `friend` still necessary?

toolchain/sem_ir/typed_insts.h
```diff
@@ -55,10 +55,8 @@ struct AdaptDecl {
 
 // The `&` address-of operator, as in `&lvalue`.
 struct AddrOf {
-  static constexpr auto Kind =
-      InstKind::AddrOf.Define<Parse::PrefixOperatorAmpId>(
-          {.ir_name = "addr_of",
-           .constant_kind = InstConstantKind::Conditional});
+  static constexpr auto Kind = InstKind::AddrOf.Define<Parse::NodeId>(
```
jonmeow, line 60: For the cases where you're switching to NodeId in order to allow implicit construction, should those have some other path so that they allow differentiating between "here's the expected node in regular code" from "here's a manual creation"? Perhaps the implicit creation is a use-case for UncheckedLoc?

Put differently, if we add metaprogramming as a way to generate code structures (i.e., macro-style metaprogramming versus generic-style metaprogramming), how do we stop from having every instruction end up with NodeId here (since the NodeId would be in the metaprogram)?

toolchain/check/member_access.cpp
```diff
@@ -196,11 +196,10 @@ static auto PerformImplLookup(Context& context, Parse::NodeId node_id,
...
 
-  return context.AddInst(
-      SemIR::LocIdAndInst::NoLoc<SemIR::InterfaceWitnessAccess>(
-          {.type_id = subst_type_id,
-           .witness_id = witness_id,
-           .index = assoc_entity->index}));
+  return context.AddInst<SemIR::InterfaceWitnessAccess>(
+      node_id, {.type_id = subst_type_id,
+                .witness_id = witness_id,
+                .index = assoc_entity->index});
```
jonmeow: Is the change in associated nodes something that could be separated out so that it's easier to review testdata changes from the rest? i.e., it might be nice to review as one change that adds locations (here and InterfaceWitness?) and affects testdata, and one that just changes how the locations are passed, if that works out. Is there any other code change that affects testdata?

toolchain/check/context.cpp
```diff
@@ -118,6 +118,38 @@ auto Context::FinishInst(SemIR::InstId inst_id, SemIR::Inst inst) -> void {
...
+}
+
+auto Context::CheckCompatibleImportedNodeKind(
+    SemIR::ImportIRInstId imported_loc_id, SemIR::InstKind kind) -> void {
+  auto& import_ir_inst = import_ir_insts().Get(imported_loc_id);
+  auto imported_kind = import_irs()
+                           .Get(import_ir_inst.ir_id)
+                           .sem_ir->insts()
+                           .Get(import_ir_inst.inst_id)
+                           .kind();
```
jonmeow: Is this something that could be broken apart to improve wrapping? e.g.:

```suggestion
  const auto* import_ir = import_irs().Get(import_ir_inst.ir_id).sem_ir;
  auto imported_kind = import_ir->insts().Get(import_ir_inst.inst_id).kind();
```

toolchain/check/context.cpp
```diff
@@ -118,6 +118,38 @@ auto Context::FinishInst(SemIR::InstId inst_id, SemIR::Inst inst) -> void {
   }
 }
 
+// Returns whether a parse node associated with an instruction of kind
+// `src_kind` is known to be usable as the location of an instruction of kind
+// `target_kind`. This may have false negatives for cases that don't occur in
+// practice.
```
jonmeow: By "don't occur in practice", is this just noting how it's really only handling equality and the `ImportDecl` special-case? If so, maybe this could instead say something like:

```suggestion
// `target_kind`. This may have false negatives when InstKinds don't match,
// but that only occurs for ImportDecl/Namespace in practice.
```

A different option you might consider is renaming this to something like "IsCompatibleImportedInstKind", then it's not a false negative, it's just that ImportDecl/Namespace is the only case where we twiddle instruction kinds IIUC.

toolchain/check/context.h
```diff
@@ -68,37 +69,34 @@ class Context {
...
+  // Returns a LocIdAndInst for an instruction with an imported location. Checks
+  // that the imported location is compatible with the kind of instruction being
+  // created.
   template <typename InstT>
-  auto AddInstReusingLoc(SemIR::LocId loc_id, InstT inst) -> SemIR::InstId {
-    return AddInst(SemIR::LocIdAndInst::ReusingLoc<InstT>(loc_id, inst));
+    requires SemIR::Internal::HasNodeId<InstT>
+  auto MakeImportedLocAndInst(SemIR::ImportIRInstId imported_loc_id, InstT inst)
+      -> SemIR::LocIdAndInst {
+    if constexpr (!SemIR::Internal::HasUntypedNodeId<InstT>) {
```
jonmeow, line 85: Out of curiosity, what's an example isntruction that gets an ImportRef and has an untyped node ids? I'm just kind of wondering, because in my head I'm thinking the entities we import would generally have typed node ids, but I expect I'm just forgetting something here.

jonmeow (commented):
I think the changes here generally make sense, but can the testdata-affecting changes could be split out? Are those sufficiently isolated from the LocId changes?

toolchain/sem_ir/typed_insts.h
zygoloid: I don't think I understand what you mean by "manual creation" versus "regular code" here.

The purpose of having a specific typed node associated with an instruction was as a guarantee to consumers of that instruction that the corresponding node will always be of the specified kind, so if we make changes that mean that we can't guarantee a particular node kind, we would need to change the kind to something more general, and perhaps to `NodeId`. If we have to do that on all instructions, then we should remove the typed node type entirely as it would serve no purpose.

For metaprogramming in particular: I think it's sufficiently unclear how we're going to approach metaprogramming that it's hard to speculate about what impact it might have. Maybe we'll want a new kind of `LocId` for locations produced by metaprograms, and maybe we will be able to map back to a node of the right kind in the metaprogram (depending on how we design the metaprogramming facilities).

toolchain/check/member_access.cpp
zygoloid: Done.

toolchain/check/context.h
zygoloid: I think `ImportRefUnloaded` and `ImportRefLoaded` are the only ones.

toolchain/sem_ir/inst.h
zygoloid: Those are the only two callers right now. I can rename this to something like `UncheckedLocUnsafe` if you think this name isn't enough to discourage people from calling this function.

I want to discourage both the import cases and the non-import cases, so I don't think having an `ImportIRInstId`-specific version really helps in that sense. I could maybe do something to befriend both callers, but it's awkward because one of them is in check rather than semir.

toolchain/sem_ir/inst.h
zygoloid: Removed.

toolchain/check/context.cpp
zygoloid: Renamed.

toolchain/sem_ir/typed_insts.h
jonmeow: I mean really "created implicitly in handlers based on parse tree nodes" versus "created based on valid code". i.e., you mention the possibility of a new kind of `LocId` for metaprograms: what if you think of desugaring implicit conversions the same way? (note, I don't intend for this to block the PR: at most, maybe a note about the typical node, to make it easier to put back)

Note, maybe this is a little tangential, and I don't mean to block the PR with it, but I think the typed nodes on instructions is at an awkward point for me, neither used often enough that it's reliably there, nor ambiguous enough that I feel confident to say we can just remove it. This PR is putting more weight on the latter side though, where maybe it's not worth the work to maintain it for the nodes that _do_ have it.

toolchain/check/context.h
jonmeow: Then FWIW, I'll suggest (because it's more explicit, and draws a line in the sand for typed nodes):

```suggestion
    // All other imported instructions should have typed nodes.
    // TODO: Maybe support InstT::Is<AnyImportRef>?
    if constexpr (InstT::Kind != SemIR::ImportRefUnloaded::Kind &&
                  InstT::Kind != SemIR::ImportRefLoaded::Kind) {
```

But understood if you prefer to keep it as-is.

toolchain/sem_ir/inst.h
jonmeow: When you say discouraging imports, do you just mean direct calls for imports, instead pushing them through context?

Had you considered something like moving as much of the logic in context.cpp for `HasCompatibleImportedNodeKind` over here, just fetching the kind on call? i.e., yielding something like `ImportedLoc(ImportIRInstId import_ir_inst_id, InstKind imported_kind, InstT inst) -> LocIdAndInst`.

jonmeow (approved):
Thanks for the changes. I think the remaining notes we can iterate on separately. Not going to merge in case you want the `HasUntypedNodeId` suggestion, but to be clear, feel free to decline that and merge.

toolchain/sem_ir/inst.h
zygoloid: > When you say discouraging imports, do you just mean direct calls for imports, instead pushing them through context?

Yes, exactly.

If we wanted to do the checking here, we'd also need to pass the `SemIR::File` in, and that feels like a layering issue to me -- this type seems like a lower layer than `File`. Also, the instruction and location are public members anyway, so we wouldn't be preventing them from mismatching, unless we also made other changes to encapsulate those fields (maybe we should do that, but it's a larger change than I wanted to make here).

Maybe a middle ground would be to move the function that safely builds an imported `LocIdAndInst` from `Check::Context` to `SemIR::File`.

toolchain/sem_ir/typed_insts.h
zygoloid: Added a comment for the two cases where we now have a less precise parse node saying what the common case is -- that seems useful information to retain, even though we have no guarantee that the parse node will actually be the one we mention.

I'm not sure that the distinction you're describing between implicitly vs explicitly created instructions is one that we should spend a lot of effort modeling in SemIR, given the decision in #3833 that SemIR should represent a desugared form of a program as if the desugared form were written directly. It might make sense to track more information about how / why we created a given instruction if it's useful for diagnostic purposes to be able to point to a different location -- and I think that might justify doing something specific for metaprograms -- but it's not clear to me that tracking which form of desugaring produced an instruction in general would pay off. Maybe we can look more into this if we find we're producing confusing diagnostics.

(And yeah, I share your concern that the typed node information might not really be pulling its weight in general.)

toolchain/check/context.h
zygoloid: I think I'd prefer to keep it as is: I think it's at least slightly happenstance that we have typed nodes for all imported nodes right now. For the most part we're importing declarations where we have a good reason to have typed nodes, but for things like function parameter patterns, I could easily imagine ending up with an instruction with an untyped node, and I think that'd be fine and we shouldn't reject it here.

PR #4267, opened 2024-08-29
Rudimentary name mangling support

dwblaikie:
This seems to be enough to avoid naming collisions for functions in any of the current test cases (verified by asserting that the name of the `llvm::Function` matches the name passed to create it - not triggering LLVM's numbering that happens when names collide)

It currently implements mangling for namespace scopes, class scopes, and impls.
Nothing generic is mangled yet - haven't looked at how that works, though evidently it's not covered by existing testing, I guess.

Follow-up change will document the current mangling algorithm in `toolchain/docs/lower.md`


toolchain/lower/mangler.h
```diff
@@ -0,0 +1,28 @@
...
+  explicit Mangler(FileContext& file_context) : file_context_(file_context) {}
+  auto Mangle(SemIR::FunctionId function_id) -> std::string;
+
+ private:
+  auto MangleInverseQualifiedNameScope(bool first_name_component,
+                                       llvm::raw_ostream& os,
+                                       SemIR::NameScopeId name_scope_id)
+      -> void;
+  FileContext& file_context_;
+};
```
jonmeow:
```suggestion

class Mangler {
 public:
  explicit Mangler(FileContext& file_context) : file_context_(file_context) {}
  
  auto Mangle(SemIR::FunctionId function_id) -> std::string;

 private:
  auto MangleInverseQualifiedNameScope(bool first_name_component,
                                       llvm::raw_ostream& os,
                                       SemIR::NameScopeId name_scope_id)
            -> void;
            
  FileContext& file_context_;
};

```

Vertical whitespace for:

- Separating the class from the wrapping namespace
- Generally spacing out constructors, functions, and member variables.
    - Generally I'd suggest vertical whitespace except between things like a chain of accessors or variables that don't need comments. (with comments, it can be nice to have a blank line before each comment starts)

toolchain/lower/mangler.h
```diff
@@ -0,0 +1,28 @@
...
+#ifndef CARBON_TOOLCHAIN_LOWER_MANGLER_H_
+#define CARBON_TOOLCHAIN_LOWER_MANGLER_H_
+
+#include <string>
+
+#include "toolchain/lower/file_context.h"
+#include "toolchain/sem_ir/ids.h"
+
+namespace Carbon::Lower {
+class Mangler {
```
jonmeow, line 18: Can you add comments to these?

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,91 @@
...
+  auto types = sem_ir.types();
+  auto names = sem_ir.names();
+  const auto& function = sem_ir.functions().Get(function_id);
+  if (SemIR::IsEntryPoint(sem_ir, function_id)) {
+    // TODO: Add an implicit `return 0` if `Run` doesn't return `i32`.
+    return "main";
+  }
+  std::string result;
+  llvm::raw_string_ostream os(result);
+  os << "_C";
```
jonmeow, line 106: Sorry, I was kind of hoping @chandlerc would review and handle this in particular. I know we'd discussed `_C` as an option, but actually, considering [swift mangling](https://github.com/swiftlang/swift/blob/main/docs/ABI/Mangling.rst) uses `$s`, should we consider a similar `$c`? I think one of the concerns with `_C` might've been too high a risk of conflicting with a C-based name, but maybe using a special character in mangling is a good way to avoid that problem entirely? Or maybe `c.` if we think we can use `.` in names?

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,91 @@
...
+        break;
+    }
+    name_scope_id = parent.parent_scope_id;
+  }
+}
+
+auto Mangler::Mangle(SemIR::FunctionId function_id) -> std::string {
+  auto& sem_ir = file_context_.sem_ir();
+  auto types = sem_ir.types();
+  auto names = sem_ir.names();
```
jonmeow: These feel a little odd to me as local variables, I don't think we generally do that. (`types` looks unused here) Maybe you could inline them, but since you have the class maybe just add a member helper like `sem_ir() { return file_context_.sem_ir(); }`? That'd mirror what we do in a lot of spots, and would end up looking a lot like what you're doing here.

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,91 @@
...
+#include "toolchain/sem_ir/entry_point.h"
+
+namespace Carbon::Lower {
+
+auto Mangler::MangleInverseQualifiedNameScope(bool first_name_component,
+                                              llvm::raw_ostream& os,
+                                              SemIR::NameScopeId name_scope_id)
+    -> void {
+  auto& sem_ir = file_context_.sem_ir();
+  auto names = sem_ir.names();
```
jonmeow: Please check clang-tidy:
https://github.com/carbon-language/carbon-lang/actions/runs/10624260039/job/29485428614

Note, clang-tidy should show up in vscode if clangd is set up. But you can also run it manually (see the .bazelrc)

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,91 @@
...
+    CARBON_KIND_SWITCH(sem_ir.insts().Get(parent.inst_id)) {
+      case CARBON_KIND(SemIR::ImplDecl impl_decl): {
+        const auto& impl = sem_ir.impls().Get(impl_decl.impl_id);
+        if (auto opt_class_self =
+                types.TryGetAs<SemIR::ClassType>(impl.self_id)) {
+          MangleInverseQualifiedNameScope(
+              true, os,
+              sem_ir.classes().Get(opt_class_self->class_id).scope_id);
+        } else {
+          auto builtin_self = types.GetAs<SemIR::BuiltinInst>(impl.self_id);
```
jonmeow: Had you considered modelling this as another `CARBON_KIND_SWITCH` or `Get` + `TryAs` + `As` instead of `TryGetAs` + `GetAs`? In particular, the type store has some indirection that is probably repeated in the `Get`.

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,91 @@
...
+      }
+      case CARBON_KIND(SemIR::ClassDecl class_decl): {
+        auto& class_node = sem_ir.classes().Get(class_decl.class_id);
+        os << names.GetFormatted(class_node.name_id);
+        break;
+      }
+      case SemIR::Namespace::Kind: {
+        auto name = names.GetAsStringIfIdentifier(parent.name_id);
+        CARBON_CHECK(name) << "Unexpected special name for namespace: "
+                           << parent.name_id;
```
jonmeow: After our whole discussion, I was thinking you'd do `GetAsStringIfIdentifier` everywhere, were you running into an issue or did I misunderstand?

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,91 @@
...
+        auto opt_interface_constraint =
+            types.GetAs<SemIR::InterfaceType>(impl.constraint_id);
+        os << names.GetFormatted(sem_ir.interfaces()
+                                     .Get(opt_interface_constraint.interface_id)
+                                     .name_id);
+        return;
+      }
+      case CARBON_KIND(SemIR::ClassDecl class_decl): {
+        auto& class_node = sem_ir.classes().Get(class_decl.class_id);
+        os << names.GetFormatted(class_node.name_id);
```
jonmeow: I think we typically append "_info" for `class` disambiguation (similarly `ns` for namespace). `_node` has a different meaning, since we do have nodes in parse.

```suggestion
        auto& class_info = sem_ir.classes().Get(class_decl.class_id);
        os << names.GetFormatted(class_info.name_id);
```

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,91 @@
...
+                                              SemIR::NameScopeId name_scope_id)
+    -> void {
+  auto& sem_ir = file_context_.sem_ir();
+  auto names = sem_ir.names();
+  auto types = sem_ir.types();
+  while (name_scope_id.is_valid()) {
+    const auto& parent = sem_ir.name_scopes().Get(name_scope_id);
+    if (parent.inst_id == SemIR::InstId::PackageNamespace) {
+      break;
+    }
```
jonmeow: I think this is similar and avoids a lookup for the package namespace:

```suggestion
  while (name_scope_id.is_valid() && name_scope_id != SemIR::NameScopeId::Package) {
    const auto& parent = sem_ir.name_scopes().Get(name_scope_id);
```

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,91 @@
...
+    if (!first_name_component) {
+      os << '.';
+    }
+    first_name_component = false;
+    CARBON_KIND_SWITCH(sem_ir.insts().Get(parent.inst_id)) {
+      case CARBON_KIND(SemIR::ImplDecl impl_decl): {
+        const auto& impl = sem_ir.impls().Get(impl_decl.impl_id);
+        if (auto opt_class_self =
+                types.TryGetAs<SemIR::ClassType>(impl.self_id)) {
+          MangleInverseQualifiedNameScope(
```
jonmeow: We're avoiding recursive functions in order to avoid stack depth issues. Can you build an iterative approach for this?

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,91 @@
...
+              true, os,
+              sem_ir.classes().Get(opt_class_self->class_id).scope_id);
+        } else {
+          auto builtin_self = types.GetAs<SemIR::BuiltinInst>(impl.self_id);
+          os << builtin_self.builtin_inst_kind.label();
+        }
+        os << ':';
+        auto opt_interface_constraint =
```
jonmeow: What's the "opt_" mean here?

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,91 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "toolchain/lower/mangler.h"
+
+#include "toolchain/base/kind_switch.h"
+#include "toolchain/sem_ir/entry_point.h"
+
+namespace Carbon::Lower {
+
+auto Mangler::MangleInverseQualifiedNameScope(bool first_name_component,
```
jonmeow: For my own understanding, why inverse instead of forward?

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,91 @@
...
+                                              llvm::raw_ostream& os,
+                                              SemIR::NameScopeId name_scope_id)
+    -> void {
+  auto& sem_ir = file_context_.sem_ir();
+  auto names = sem_ir.names();
+  auto types = sem_ir.types();
+  while (name_scope_id.is_valid()) {
+    const auto& parent = sem_ir.name_scopes().Get(name_scope_id);
```
jonmeow: Had you considered "name_scope" instead of "parent"? What's this the parent of?

toolchain/lower/file_context.cpp
```diff
@@ -238,17 +239,8 @@ auto FileContext::BuildFunctionDecl(SemIR::FunctionId function_id)
           ? return_type
           : llvm::Type::getVoidTy(llvm_context());
 
-  std::string mangled_name;
-  if (SemIR::IsEntryPoint(sem_ir(), function_id)) {
-    // TODO: Add an implicit `return 0` if `Run` doesn't return `i32`.
```
jonmeow, line 243: I wouldn't expect the name mangler to add a return. Maybe this TODO should remain here? Perhaps this code should even keep the `IsEntryPoint` call so that doing so is smooth?

jonmeow (commented):
Thanks, I'm glad to see the start on mangling! There are some bits that make sense to me, some questions about how mangling should really work.

For documentation, my sense would be that it's a design choice. If you'd like to write a proposal for that (with alternatives, etc), let me know and I'll help walk you through that. Otherwise, maybe just create a Google Doc explaining it for now? Ideally I'd say to update `toolchain/docs/lower.md`, but unfortunately #4242 is still in review.

toolchain/lower/mangler.cpp
dwblaikie: Optional - not sure if there's another convention here. I'd prefer something at the declaration that indicates the optionality (either using an explicit type instead of `auto` or something in the name like `opt_`)

toolchain/lower/file_context.cpp
dwblaikie: Rephrased the comment and applied it to the return type computation in the previous statement/just above this line.

The fix isn't really in the name mangling, it's in choosing the return type and in the code generation for the function body. So it seems a bit more accurate having the comment apply to the return type computation - though it's still a fair way from where the other part of the fix has to happen in the code generation.

toolchain/lower/mangler.cpp
dwblaikie: Ah, sure - it's the parent of the previous entity (the loop is looping through parents) - but, yeah, might as well name it to match the name_scope_id.

toolchain/lower/mangler.cpp
dwblaikie: Added a comment explaining this - I think someone, maybe @zygoloid, suggested that mangling in inverse order may reduce the incidence of names having a common prefix which might slow down uniqueness tests/comparisons (since the comparison would always have to walk the common prefixes before finding something unique).

toolchain/lower/mangler.cpp
dwblaikie: Sure - added some helpers.

toolchain/lower/mangler.cpp
dwblaikie: Think I got those addressed.

I am not using vscode at the moment, but really should give it a go - and running it with bazel is a bit tricky since the project isn't clalng-tidy clean at the moment.

jonmeow: I believe the project is clang-tidy clean under LLVM 16 -- we are running it on every commit at that version. If you're having issues, fixes are welcome, but also using LLVM 16 should be clean.

toolchain/lower/mangler.cpp
dwblaikie: Ah, sure, will do.

toolchain/lower/mangler.cpp
dwblaikie: I'm open to other ideas - The "_<upper>" form seems to be dominant, though (LLVM's demangler supports _Z for C++, _R for Rust, and _D for the D programming language).

Just looking at LLVM's demangler makes the fair point that a `.` prefix would be problematic (not that that's what you were suggesting) - because of the leading `.` in ELF symbol names, so the demangler strips those off by default.

Work backwards from _Z and use _Y (or perhaps _X, because X is cool)?

Though I'm really OK with whatever everyone else wants - and wouldn't totally object to using something following swift's pattern. So if $c is best for you, happy to switch to that.

toolchain/lower/mangler.h
dwblaikie: Sure

toolchain/lower/mangler.cpp
jonmeow: I think `GetAs` doesn't return an optional.

Note, typically we just don't remark on it. But I think it may not make sense here.

toolchain/lower/mangler.cpp
dwblaikie: Sure thing - think I've got that sorted, and basically meant figuring out mangling qualified interface names in too, so that's added/working.

toolchain/lower/mangler.cpp
dwblaikie: Yeah, think I just got a bit confused after stumbling over the NameScope::name_id is only valid for namespaces, and other name_ids you have to go looking in the underlying entity - and then I was looking for how to render those entities' names and came across other APIs and copied those, unsure if/how they were load bearing.

But thanks for the prompt - zooming out again, with the code tidied up a bit, I've been able to use that consistently everywhere, I think. (though not able to share many call sites/deduplicate much code, since the name_ids all come from different places, etc)

toolchain/lower/mangler.cpp
dwblaikie: ah, yep - didn't realize the bazel build used my host tools like that - I'd assumed it was more hermetic. got my local situation version matched so that'll be easier (until I forget to update my PATHs after a reboot between LLVM and Carbon development... )

toolchain/lower/mangler.cpp
dwblaikie: ah, yeah, must've been some lingering thing. Fixed/removed the prefix.

dwblaikie (commented):
OK - sorry for the drip feed of responses earlier. Working on my github review workflow some more.

I think I've addressed all the comments and things are a bit tidier now.

toolchain/lower/mangler.cpp
jonmeow: We can separate this out if you want to start with _C as probably good.

I was actually thinking further, and if `C.` or `C$` were an option (upper or lower case), particularly using a non-identifier character as separators instead of prefix, it might both (a) make the name easy to read and (b) offer a name that couldn't conflict with a C name.

But I'm realizing the difference between `_C` and `$c` isn't actually that big for us: we should be suffixing all mangled names with the package scope (e.g., `_CFoo.Main`), which should already avoid name conflicts with C names if we're using `.` separators. So if you don't think `C<char>` is a good choice, let's stick with `_C` for cross-language consistency.

toolchain/lower/file_context.cpp
```diff
@@ -233,22 +234,14 @@ auto FileContext::BuildFunctionDecl(SemIR::FunctionId function_id)
 
   // Compute the return type to use for the LLVM function. If the initializing
   // representation doesn't produce a value, set the return type to void.
+  // TODO: Remap return type to i32 if `Run` doesn't return a value.
```
```suggestion
  // TODO: For the `Run` entry point, remap return type to i32 if it doesn't return a value.
```

Note, I'm really just trying to capture that `Run` means the entry point. I think that was previously apparent next to the `IsEntryPoint` before, but I think it's not as obvious now.

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,99 @@
...
+        CARBON_FATAL() << "Attempting to mangle unsupported SemIR.";
+        break;
+    }
+  }
+}
+
+auto Mangler::Mangle(SemIR::FunctionId function_id) -> std::string {
+  const auto& function = sem_ir().functions().Get(function_id);
+  if (SemIR::IsEntryPoint(sem_ir(), function_id)) {
+    // TODO: Add an implicit `return 0` if `Run` doesn't return `i32`.
```
jonmeow:
```suggestion
```

toolchain/lower/mangler.h
```diff
@@ -0,0 +1,48 @@
...
+  // specified `FileContext`.
+  explicit Mangler(FileContext& file_context) : file_context_(file_context) {}
+
+  // Produce a deterministically unique mangled name for the function specified
+  // by `function_id`.
+  auto Mangle(SemIR::FunctionId function_id) -> std::string;
+
+ private:
+  // Mangle this qualified name with inner scope first, working outwards. This
+  // may reduce the incidence of common prefixes in the name mangling. (ie:
```
jonmeow:
```suggestion
  // may reduce the incidence of common prefixes in the name mangling. (i.e.:
```

toolchain/lower/mangler.h
```diff
@@ -0,0 +1,48 @@
...
+#include <string>
+
+#include "toolchain/lower/file_context.h"
+#include "toolchain/sem_ir/ids.h"
+
+namespace Carbon::Lower {
+
+// A class for producing mangled (deterministically unique, at least partially
+// human readable) names for externally referenceable entities such as extern
+// functions.
```
jonmeow:
```suggestion
// human readable) names for externally referenceable entities such as
// functions.
```

"extern" has what I think is a slightly different meaning in Carbon than in C++, would it make sense to omit it here?

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,99 @@
...
+                                              SemIR::NameScopeId name_scope_id)
+    -> void {
+  struct NameEntry {
+    SemIR::NameScopeId name_id;
+    char prefix;
+  };
+  llvm::SmallVector<NameEntry> names_to_render;
+  auto add_scope = [&](SemIR::NameScopeId name_scope_id, char prefix) {
+    if (name_scope_id.is_valid() &&
+        name_scope_id != SemIR::NameScopeId::Package) {
```
jonmeow: Sorry, I was thinking about this a little more and I'm realizing the package name needs to be part of mangled names. That is, in tests, `@_CF` should be `@_CF.Main` because it's valid to have `F()` in different packages.

Would you prefer this to be a TODO, or to address it here? `sem_ir().package_id()` is Invalid for `Main` and would be used directly otherwise. At present, NameScope::is_closed_import indicates a package import, at which point the parent wouldn't be traversed (`_COp.Number:Negate.Core` is an example showing the `Core` package, in tests at present because `Core` is imported).

To be sure, the implication I'm describing is that given:

```
// --- foo.carbon
package Foo;
fn F() {}
// --- bar.carbon
package Bar;
import Foo;
fn G() { return Foo.F(); }
```

You will presently produce `@_CF` for foo.carbon, and `@_CF.Foo` for bar.carbon.

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+#include "toolchain/base/kind_switch.h"
+#include "toolchain/sem_ir/entry_point.h"
+
+namespace Carbon::Lower {
+
+auto Mangler::MangleInverseQualifiedNameScope(llvm::raw_ostream& os,
+                                              SemIR::NameScopeId name_scope_id)
+    -> void {
+  struct NameEntry {
```
jonmeow, line 16: I think it'd be helpful to have some implementation comments. For example:

```suggestion
  // Maintain a stack of names for delayed rendering of interface impls.
  struct NameEntry {
```

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,99 @@
...
+    SemIR::NameScopeId name_id;
```
jonmeow:
```suggestion
    SemIR::NameScopeId name_scope_id;
```

We do also have a `NameId` structure -- I think we tend to call these either `scope_id` or `name_scope_id`

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,99 @@
...
+    }
+    add_scope(name_scope.parent_scope_id, '.');
+    CARBON_KIND_SWITCH(sem_ir().insts().Get(name_scope.inst_id)) {
+      case CARBON_KIND(SemIR::ImplDecl impl_decl): {
+        const auto& impl = sem_ir().impls().Get(impl_decl.impl_id);
+
+        auto interface_type =
+            types().GetAs<SemIR::InterfaceType>(impl.constraint_id);
+        auto interface = sem_ir().interfaces().Get(interface_type.interface_id);
+        names_to_render.push_back({interface.scope_id, ':'});
```
jonmeow: Can you use `add_scope` where you currently use `names_to_render.push_back`? Mainly thinking for consistency, one way other the other.

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+auto Mangler::MangleInverseQualifiedNameScope(llvm::raw_ostream& os,
+                                              SemIR::NameScopeId name_scope_id)
+    -> void {
+  struct NameEntry {
+    SemIR::NameScopeId name_id;
+    char prefix;
+  };
+  llvm::SmallVector<NameEntry> names_to_render;
+  auto add_scope = [&](SemIR::NameScopeId name_scope_id, char prefix) {
```
jonmeow: This is a stack, right? Maybe "push" instead of "add"?

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,99 @@
...
+    if (name_scope_id.is_valid() &&
```
jonmeow: Is name_scope_id ever invalid here?

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include "toolchain/sem_ir/entry_point.h"
+
+namespace Carbon::Lower {
+
+auto Mangler::MangleInverseQualifiedNameScope(llvm::raw_ostream& os,
+                                              SemIR::NameScopeId name_scope_id)
+    -> void {
+  struct NameEntry {
+    SemIR::NameScopeId name_id;
+    char prefix;
```
jonmeow, line 23: Can you document the intended semantics of `prefix`? It looks like `:`, `.`, and `\0` are intended values here -- using `\0` in particular seems very subtle since you seem to be relying on implicit 0-init for that.

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,99 @@
...
+  };
+  llvm::SmallVector<NameEntry> names_to_render;
+  auto add_scope = [&](SemIR::NameScopeId name_scope_id, char prefix) {
+    if (name_scope_id.is_valid() &&
+        name_scope_id != SemIR::NameScopeId::Package) {
+      names_to_render.push_back({name_scope_id, prefix});
```
jonmeow: Please use designated initializers for structs. We've gone in that direction for [struct style](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/project/cpp_style_guide.md) because it matches what we'd do in Carbon, and avoids some initialization nuances of C++.

Note, I'd also be agnostic if you switch to a `pair` instead of a struct, since we call out an exception for that as tuple-ish.

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,99 @@
...
+      os << prefix;
+    }
+    add_scope(name_scope.parent_scope_id, '.');
+    CARBON_KIND_SWITCH(sem_ir().insts().Get(name_scope.inst_id)) {
+      case CARBON_KIND(SemIR::ImplDecl impl_decl): {
+        const auto& impl = sem_ir().impls().Get(impl_decl.impl_id);
+
+        auto interface_type =
+            types().GetAs<SemIR::InterfaceType>(impl.constraint_id);
+        auto interface = sem_ir().interfaces().Get(interface_type.interface_id);
```
jonmeow:
```suggestion
        const auto& interface = sem_ir().interfaces().Get(interface_type.interface_id);
```

I'm guessing this is just an oversight since it's similar to `impl` above, which you do `const auto&` on.

jonmeow (commented):
Thanks, the approach is looking good. I did realize we need to handle the package differently, but I'm fine splitting that out as a TODO if you'd rather just finish here. I think the necessary changes won't significantly affect the current logic.

BTW, might be worth adding a brief TODO about the generic/impl mangling you note in the PR description, but I'll leave that to you.

And toolchain/docs/lower.md is now merged for edits, but that can also be done separately.

toolchain/lower/mangler.cpp
dwblaikie: Ah, yeah, meant to leave a FIXME in there somewhere - but thanks for the pointers and examples explaining how the package naming works - so I've implemented that

toolchain/lower/mangler.cpp
dwblaikie: Nope - I might've got confused from some versions of this, or some code that I was "inspired" by.

Removed.

toolchain/lower/mangler.cpp
dwblaikie: Ended up removing this helper since the conditional stuff went away (between invalid cases not actually happening, and the package cases being supported/happening in the main loop instead)

toolchain/lower/mangler.cpp
dwblaikie: Ah, sure - documented the 3 cases (& also removed dependence on implicit 0-init, FWIW) in the comment. Though not sure if it'd be better to document the cases at their usage, and leave the comment here generically describing any prefix and the null-means-no-prefix case.

toolchain/lower/mangler.cpp
dwblaikie: Sure thing, done (kept with a named struct and added designated initializers).

toolchain/lower/mangler.cpp
dwblaikie: I had been using direct push_back when the extra conditionals (valid/package) weren't needed, but now that those conditionals aren't needed at all - I removed the helper entirely and left these all as raw push_back, because that seemed still sufficiently legible to not warrant the extra lambda indirection to me, at least.

dwblaikie (commented):
> Thanks, the approach is looking good. I did realize we need to handle the package differently, but I'm fine splitting that out as a TODO if you'd rather just finish here. I think the necessary changes won't significantly affect the current logic.

Ended up adding package support along the way.

> BTW, might be worth adding a brief TODO about the generic/impl mangling you note in the PR description, but I'll leave that to you.

Ended up implementing the impl qualifier mangling in some of the review iterations, so that's done. I'll leave a FIXME for generic support.

> And toolchain/docs/lower.md is now merged for edits, but that can also be done separately.

Mentioned that in the description, to be done in a follow-up.

dwblaikie:
(with the fix for the mangling ending too early with package support and impls - now `bazel check //...` with a patch to assert an llvm::Function's name matches the name requested (ie: testing that we never request duplicate names and trigger LLVM's name deduplication support) doesn't fail - which is something :) Not sure if it's worth including that assertion in trunk... I guess it probably is?)

toolchain/lower/testdata/impl/extend_impl.carbon
```diff
@@ -31,20 +31,20 @@ fn InstanceAccess(a: A) -> i32 {
 // CHECK:STDOUT: ; ModuleID = 'extend_impl.carbon'
 // CHECK:STDOUT: source_filename = "extend_impl.carbon"
 // CHECK:STDOUT:
-// CHECK:STDOUT: define i32 @F() !dbg !4 {
+// CHECK:STDOUT: define i32 @"_CF.A.Main:I.Main.A.Main"() !dbg !4 {
```
zygoloid: This mangling looks a little strange -- specifically repeating the `.A.Main` part twice. It also looks like the trailing portion might be ambiguous. How should we read this?

toolchain/lower/mangler.cpp
```diff
@@ -12,40 +12,52 @@ namespace Carbon::Lower {
...
     -> void {
+  // Maintain a stack of names for delayed rendering of interface impls.
   struct NameEntry {
-    SemIR::NameScopeId name_id;
+    SemIR::NameScopeId name_scope_id;
+
+    // The prefix emitted before this name component. If '\0', no prefix will be
+    // emitted.
+    // * namespace components are separated by '.'
+    // * the two components of an interface are separated by ':'
```
jonmeow:
```suggestion
    // - Namespace components are separated by '.'.
    // - The two components of an interface are separated by ':'.
```

Minor capitalization/punctuation. The '-' is just consistency, we've been using '-' (including md files, where prettier reformats '*' to '-' automatically)

toolchain/lower/mangler.cpp
jonmeow: I'm fine with this, it fits with top-down reading, and people can always cross-reference uses.

jonmeow (approved):
I think this looks good now. For zygoloid's comment, if you can address it easily here feel free and I can take another look, but I think it'd also make sense to merge this and work in a separate PR since it'll have more narrow impact. (note, when `impl.self_id` is a class, 3 scopes get added to names_to_render, maybe 1 should be cut)

toolchain/lower/testdata/impl/extend_impl.carbon
dwblaikie: Ah, yep - that's a bug.

Fixed.

The issue was that it was mangling in the parent name scopes, including the parent name scope of wherever the impl was defined - which, if I understand correctly, isn't relevant. (an impl is uniquely defined by its interface and constraint, regardless of the scope in which that impl is defined - ie: two impls of the same interface+constraint, written in different namespaces, are invalid/do collide. (well, I guess there's no syntax to define an impl in a namespace, so it only happens to collide between, I guess, two files both containing an impl or a single file containing an inline impl inside a class definition, and one outside, which fails with:
```
basic.carbon:11:1: ERROR: Redefinition of `impl B as A`.
impl B as A {
^~~~~~~~~~~~~
basic.carbon:6:3: Previous definition was here.
  extend impl as A {
  ^~~~~~~~~~~~~~~~~~
```

Fixed this by adding the lexical parent scope after the switch, and having the impl handling `continue` resulting in the parent scope addition being skipped.

Now there shouldn't be any trailing portion after an impl, it should be `<function name>.<self name>:<constraint name>`.

dwblaikie:
Very much messed this PR state up... trying to figure out how to fix it :/

toolchain/lower/file_context.cpp
```diff
@@ -233,22 +234,14 @@ auto FileContext::BuildFunctionDecl(SemIR::FunctionId function_id)
 
   // Compute the return type to use for the LLVM function. If the initializing
   // representation doesn't produce a value, set the return type to void.
+  // TODO: For the `Run` entry point, remap return type to i32 if it doesn't return a value.
```
CarbonInfraBot, line 237: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  // TODO: For the `Run` entry point, remap return type to i32 if it doesn't
  // return a value.
```

```suggestion
  // TODO: For the `Run` entry point, remap return type to i32 if it doesn't
  // return a value.
```

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,118 @@
...
+    // emitted.
+    // - Namespace components are separated by '.'.
+    // - The two components of an interface are separated by ':'.
+    char prefix;
+  };
+  llvm::SmallVector<NameEntry> names_to_render;
+  names_to_render.push_back({.name_scope_id = name_scope_id, .prefix = '.'});
+  while (!names_to_render.empty()) {
+    auto [name_scope_id, prefix] = names_to_render.back();
+    names_to_render.pop_back();
```
jonmeow:
```suggestion
    auto [name_scope_id, prefix] = names_to_render.pop_back_val();
```

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,118 @@
...
+        os << sem_ir().identifiers().Get(package_id);
+      } else {
+        os << "Main";
+      }
+      continue;
+    }
+    const auto& name_scope = sem_ir().name_scopes().Get(name_scope_id);
+    CARBON_KIND_SWITCH(sem_ir().insts().Get(name_scope.inst_id)) {
+      case CARBON_KIND(SemIR::ImplDecl impl_decl): {
+        names_to_render.clear();
```
jonmeow: Is this clear needed? Looks like you moved the push to the end, so maybe it's not needed anymore?

Note I have in mind something like:

```
class C {
  extend impl as I {
    fn F() {
      class C { extend impl as I { ... } }
    }
  }
}
```

Which seems like a terrible idea, but at the same time, valid code.

```suggestion
```

jonmeow (approved):
Still LGTM, just pointing out a couple small things.

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,117 @@
...
+        os << sem_ir().identifiers().Get(package_id);
+      } else {
+        os << "Main";
+      }
+      continue;
+    }
+    const auto& name_scope = sem_ir().name_scopes().Get(name_scope_id);
+    CARBON_KIND_SWITCH(sem_ir().insts().Get(name_scope.inst_id)) {
+      case CARBON_KIND(SemIR::ImplDecl impl_decl): {
+
```
CarbonInfraBot, line 43: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
```

toolchain/lower/mangler.cpp
```diff
@@ -0,0 +1,116 @@
...
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
```

toolchain/lower/mangler.cpp
dwblaikie: Removed the `clear()` call - thanks for the catch.

Though the example code doesn't compile at the moment (maybe local classes aren't supported yet?
```
basic.carbon:7:7: ERROR: Expected expression.
      class C {
      ^~~~~
```

Though once supported, the mangling code will need to be enhanced to handle intermediate function names, not just at the top level.

Without considering local types, I was thinking about nested types and when dependent types get added to interfaces/impls, that might present similar problems depending on exactly how they work. Then we might have a case where an impl is a scope, though I guess it'd still be the top level scope... hmm, but might've bifurcated earlier on - at that point we might want some way to reference existing entities in the mangling, like Itanium...

jonmeow: Yeah, not yet supported.

jonmeow: I couldn't think of a case with existing code either, due to the restrictions on `impl as`

PR #4268, opened 2024-08-30
Refactor FindPreludeFiles into InstallPaths

jonmeow:
From the driver's perspective, `FindPreludeFiles` is closely tied to `compile`. This makes it difficult to refactor commands without affecting the test dependencies on `FindPreludeFiles`. `InstallPaths` seems like a decent home since it is responsible for the install structure.

I'm switching to an `Error` return to allow callers to choose how to handle it (e.g., in file tests, we typically don't want the direct error stream).

### DIFF
--- a/toolchain/install/install_paths.cpp
+++ b/toolchain/install/install_paths.cpp
@@ -76,6 +76,47 @@ auto InstallPaths::Make(llvm::StringRef install_prefix) -> InstallPaths {
   return paths;
 }
 
+auto InstallPaths::FindPreludeFiles() const
+    -> ErrorOr<llvm::SmallVector<std::string>> {
+  // This is structured to avoid a vector copy on success.
+  ErrorOr<llvm::SmallVector<std::string>> result =
+      llvm::SmallVector<std::string>();
+
+  std::string dir = core_package();
+
+  // Include <data>/core/prelude.carbon, which is the entry point into the
+  // prelude.
+  {
+    llvm::SmallString<256> prelude_file(dir);
+    llvm::sys::path::append(prelude_file, llvm::sys::path::Style::posix,
+                            "prelude.carbon");
+    result->push_back(prelude_file.str().str());
+  }
+
+  // Glob for <data>/core/prelude/**/*.carbon and add all the files we find.
+  // TODO: This should be using the vfs instead of actual filesystem.
+  llvm::SmallString<256> prelude_dir(dir);
+  llvm::sys::path::append(prelude_dir, llvm::sys::path::Style::posix,
+                          "prelude");
+  std::error_code ec;
+  for (llvm::sys::fs::recursive_directory_iterator prelude_files_it(
+           prelude_dir, ec, /*follow_symlinks=*/false);
+       prelude_files_it != llvm::sys::fs::recursive_directory_iterator();
+       prelude_files_it.increment(ec)) {
+    if (ec) {
+      result = ErrorBuilder() << "Could not find prelude: " << ec.message();
+      return result;
+    }
+
+    auto prelude_file = prelude_files_it->path();
+    if (llvm::sys::path::extension(prelude_file) == ".carbon") {
+      result->push_back(prelude_file);
+    }
+  }
+
+  return result;
+}
+
 auto InstallPaths::SetError(llvm::Twine message) -> void {
   // Use an empty prefix on error as that should use the working directory which
   // is the least likely problematic.
--- a/toolchain/install/install_paths.cpp
+++ b/toolchain/install/install_paths.cpp
@@ -94,7 +94,6 @@ auto InstallPaths::FindPreludeFiles() const
   }
 
   // Glob for <data>/core/prelude/**/*.carbon and add all the files we find.
-  // TODO: This should be using the vfs instead of actual filesystem.
   llvm::SmallString<256> prelude_dir(dir);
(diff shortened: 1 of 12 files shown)


jonmeow:
I'll note, and I'm trying to decide what to do with this, but right now the `InstallPaths` is typically created with the real filesystem, then the `Driver` uses a **separate** `InMemoryFileSystem`. This mostly works because paths are mirrored. However, it also means (for example) that even though `InstallPaths` validates the marker file for the real filesystem, the driver won't see one.

toolchain/install/install_paths.cpp
```diff
@@ -76,6 +76,47 @@ auto InstallPaths::Make(llvm::StringRef install_prefix) -> InstallPaths {
...
+  // prelude.
+  {
+    llvm::SmallString<256> prelude_file(dir);
+    llvm::sys::path::append(prelude_file, llvm::sys::path::Style::posix,
+                            "prelude.carbon");
+    result->push_back(prelude_file.str().str());
+  }
+
+  // Glob for <data>/core/prelude/**/*.carbon and add all the files we find.
+  // TODO: This should be using the vfs instead of actual filesystem.
```
zygoloid: Does this TODO still make sense? `InstallPaths`, as it stands, doesn't make use of the VFS at all, so if this functionality fits here (which I think it does) it seems logical that it wouldn't do so either.
```suggestion
```
Perhaps a good model here would be that (if needed) we construct a fake `InstallPaths` for testing, for which we specify a list of prelude files on construction. But for this PR, I think what you have here is fine.

toolchain/install/install_paths_test_helpers.h
```diff
@@ -0,0 +1,19 @@
...
+#define CARBON_TOOLCHAIN_INSTALL_INSTALL_PATHS_TEST_HELPERS_H_
+
+#include "llvm/Support/VirtualFileSystem.h"
+#include "toolchain/install/install_paths.h"
+
+namespace Carbon::Testing {
+
+// Prepares the VFS with prelude files from the real filesystem.
+auto CopyPreludeFiles(InstallPaths install_paths,
+                      llvm::vfs::InMemoryFileSystem* vfs) -> void;
```
zygoloid: For me it'd be clearer to name this `AddPreludeFiles` or `AddPreludeFilesToVfs` or something. This name sounds superficially like it would perform an actual on-disk file copy.

zygoloid (approved):
LGTM. Both comments are optional; feel free to merge as-is if you prefer.

toolchain/install/install_paths.cpp
jonmeow: Let me try to give an example of why I think this should be a TODO (you say "still", but I'm trying to add it).

We allow passing a vfs into the driver. Presumably we want to support, for example, people running the toolchain completely in-memory. However, because InstallPaths requires using the real filesystem, it means that some things will use the vfs, and some things won't.

For example, the toolchain is detected at a real filesystem path. That is used to get prelude paths. The prelude paths are read from the vfs, which consequently must be identical to the real filesystem paths.

A different approach, of course, would be to consistently use the real filesystem for things that come from `InstallPaths`. I suspect though that we would want to go in the other direction. Maybe I'm misunderstanding how others lean for vfs use?

toolchain/install/install_paths_test_helpers.h
jonmeow: How about `CopyPreludeFilesToVfs`?

toolchain/install/install_paths.cpp
zygoloid: My thinking is that, while it seems good to have a TODO, it doesn't belong *here*. This class is all about finding things in the real file system. But in [testing/base/source_gen_test.cpp](https://github.com/carbon-language/carbon-lang/pull/4268/files#diff-7b8e6f9beadffd5a28d8b509c5e531f6808c9d90b5718af1b72b8ceaf5d25577), in `TestCompile`, we're passing an `InstallPaths` object that describes paths relative to the real file system (as all `InstallPaths` objects do today) and a VFS to the `Driver` constructor, and I think *that* is the thing that's wrong / should have a TODO, because that's where we're mixing up real file system and VFS stuff. So I'd be happier if the TODO lived there rather than here. But as noted, I'm happy for you to merge this as-is, with or without the TODO here.

As a fix for the TODO, I think a clean approach here would be to make `InstallPaths` more of an abstraction with a "real file system" implementation (this one), and a "fake file system for tests" implementation that uses the list of prelude file names that it copied into the VFS as the result of its `FindPreludeFiles` call. (But ideally not in this PR!)

toolchain/install/install_paths_test_helpers.h
zygoloid: Sure, that's clear enough that it's not touching the disk.

toolchain/install/install_paths.cpp
jonmeow: Moved the comment to the class level in the header, does that make sense? (in part to keep it in a central spot).

Note, I think we may want to support VFS for non-test Driver invocations.

toolchain/install/install_paths_test_helpers.h
jonmeow: Switched to `AddPreludeFilesToVfs`

toolchain/install/install_paths.cpp
zygoloid: Thanks, that looks great.

zygoloid (approved):

PR #4269, opened 2024-09-03
Shrink the lexer's token location and line data structures.

chandlerc:
First, this replaces the separate line index and column index in the token information with a single 32-bit byte offset of the token. This is then used to compute line and column numbers with a binary search of the line structure and then using that to compute the column within the line. In practice, this is _much_ more efficient:

- Smaller token data structure. This will hopefully combine with a subsequent optimization PR that shrinks the token data structure still further.
- Fewer stores to form each token's information in the tight hot loop of the lexer.
- Less state to maintain while lexing, fewer computations while lexing.

We only have to search to build the line and column information off the hot lexing path, and so this ends up being a significant win and shrinks some of the more significant data structures.

Second, this shrinks the line start to a 32-bit integer and removes the line length. Our source buffer already ensures we only have 2 GiB of source with a nice diagnostic. I've just added a check to help document this in the lexer. The line length can be avoided in all of the cases it was being used, largely by looking at the next line's start and working from there. This also precipitated cleaning up some code that dated from when lines were only built during lexing rather than being pre-built, which resulted in nice simplifications.

With this PR, I think it makes sense to re-name a bunch of methods on `TokenizedBuffer`, but to an extent that was already needed as these methods somewhat predate the more pervasive style conventions. I avoided that here to keep this PR focused on the implementation change, I'll create a subsequent PR to update the API to both better nomenclature and remove deviations from our conventions.

There may also be a way to de-duplicate the binary search in the diagnostic location conversion and the main line accessor binary search, but it wasn't obvious to me that it would be a net savings, so left it alone for now.

The performance impact of this varies quite a bit...

The lexer's benchmark improves pretty consistent across the board on both x86 and Arm. For x86, where I have nice comparison tools, it appears 3% to 20% faster depending on the specific pattern. For Arm server CPUs at least it seems a much smaller but still an improvement.

The overall compilation benchmarks however don't improve much with these changes alone on x86. Significant reduction in instruction count required for lexing, but the overall performance is bottlenecked elsewhere in the overall compilation it seems. However, on Arm, despite the more modest gains in special cases of lexing, this shows fairly consistent 1-2% improvements in overall lexing performance on our compilation benchmark. And the expectaiton is these improvements will compound with subsequent work to further compact our representation.

### DIFF
--- a/toolchain/lex/lex.cpp
+++ b/toolchain/lex/lex.cpp
@@ -101,8 +101,8 @@ class [[clang::internal_linkage]] Lexer {
   auto next_line() -> LineIndex { return LineIndex(line_index_ + 1); }
 
   auto next_line_info() -> TokenizedBuffer::LineInfo* {
-    CARBON_DCHECK(line_index_ + 1 <
-                  static_cast<ssize_t>(buffer_.line_infos_.size()));
+    CARBON_CHECK(line_index_ + 1 <
+                 static_cast<ssize_t>(buffer_.line_infos_.size()));
     return &buffer_.line_infos_[line_index_ + 1];
   }
 
--- a/toolchain/lex/lex.cpp
+++ b/toolchain/lex/lex.cpp
@@ -959,6 +959,7 @@ auto Lexer::LexNumericLiteral(llvm::StringRef source_text, ssize_t& position)
     return LexError(source_text, position);
   }
 
+  // Capture the position before we step past the token.
   int32_t byte_offset = position;
   int token_size = literal->text().size();
   position += token_size;
@@ -1000,6 +1001,7 @@ auto Lexer::LexStringLiteral(llvm::StringRef source_text, ssize_t& position)
     return LexError(source_text, position);
   }
 
+  // Capture the position before we step past the token.
   int32_t byte_offset = position;
   int string_column = byte_offset - current_line_info()->start;
   ssize_t literal_size = literal->text().size();
@@ -1157,6 +1159,7 @@ auto Lexer::LexKeywordOrIdentifier(llvm::StringRef source_text,
   CARBON_CHECK(
       IsIdStartByteTable[static_cast<unsigned char>(source_text[position])]);
 
+  // Capture the position before we step past the token.
   int32_t byte_offset = position;
 
   // Take the valid characters off the front of the source buffer.
--- a/toolchain/lex/lex.cpp
+++ b/toolchain/lex/lex.cpp
@@ -5,6 +5,7 @@
 #include "toolchain/lex/lex.h"
 
 #include <array>
+#include <limits>
 
 #include "common/check.h"
 #include "common/variant_helpers.h"
@@ -97,9 +98,12 @@ class [[clang::internal_linkage]] Lexer {
     return &buffer_.line_infos_[line_index_];
   }
 
-  auto ComputeColumn(ssize_t position) -> int {
-    CARBON_DCHECK(position >= current_line_info()->start);
-    return position - current_line_info()->start;
+  auto next_line() -> LineIndex { return LineIndex(line_index_ + 1); }
+
+  auto next_line_info() -> TokenizedBuffer::LineInfo* {
+    CARBON_DCHECK(line_index_ + 1 <
+                  static_cast<ssize_t>(buffer_.line_infos_.size()));
+    return &buffer_.line_infos_[line_index_ + 1];
   }
 
   auto NoteWhitespace() -> void {
@@ -142,7 +146,8 @@ class [[clang::internal_linkage]] Lexer {
 
   // Given a word that has already been lexed, determine whether it is a type
   // literal and if so form the corresponding token.
-  auto LexWordAsTypeLiteralToken(llvm::StringRef word, int column) -> LexResult;
+  auto LexWordAsTypeLiteralToken(llvm::StringRef word, int32_t byte_offset)
+      -> LexResult;
 
   auto LexKeywordOrIdentifier(llvm::StringRef source_text, ssize_t& position)
       -> LexResult;
@@ -661,6 +666,10 @@ static auto EstimateUpperBoundOnNumIdentifiers(int line_count) -> int {
 auto Lexer::Lex() && -> TokenizedBuffer {
   llvm::StringRef source_text = buffer_.source_->text();
 
+  // Enforced by the source buffer, but something we heavily rely on throughout
+  // the lexer.
+  CARBON_CHECK(source_text.size() < std::numeric_limits<int32_t>::max());
+
   // First build up our line data structures.
   MakeLines(source_text);
 
@@ -717,18 +726,18 @@ auto Lexer::MakeLines(llvm::StringRef source_text) -> void {
   while (const char* nl = reinterpret_cast<const char*>(
              memchr(&text[start], '\n', size - start))) {
     ssize_t nl_index = nl - text;
-    buffer_.AddLine(TokenizedBuffer::LineInfo(start, nl_index - start));
+    buffer_.AddLine(TokenizedBuffer::LineInfo(start));
     start = nl_index + 1;
   }
   // The last line ends at the end of the file.
-  buffer_.AddLine(TokenizedBuffer::LineInfo(start, size - start));
+  buffer_.AddLine(TokenizedBuffer::LineInfo(start));
 
   // If the last line wasn't empty, the file ends with an unterminated line.
   // Add an extra blank line so that we never need to handle the special case
   // of being on the last line inside the lexer and needing to not increment
   // to the next line.
   if (start != size) {
-    buffer_.AddLine(TokenizedBuffer::LineInfo(size, 0));
+    buffer_.AddLine(TokenizedBuffer::LineInfo(size));
   }
 
   // Now that all the infos are allocated, get a fresh pointer to the first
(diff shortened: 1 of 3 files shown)


toolchain/lex/lex.cpp
```diff
@@ -97,9 +98,12 @@ class [[clang::internal_linkage]] Lexer {
...
   }
 
-  auto ComputeColumn(ssize_t position) -> int {
-    CARBON_DCHECK(position >= current_line_info()->start);
-    return position - current_line_info()->start;
+  auto next_line() -> LineIndex { return LineIndex(line_index_ + 1); }
+
+  auto next_line_info() -> TokenizedBuffer::LineInfo* {
+    CARBON_DCHECK(line_index_ + 1 <
+                  static_cast<ssize_t>(buffer_.line_infos_.size()));
```
jonmeow: Regarding CHECK versus DCHECK, were you observing this as a performance issue? It looks like it's only called inside multi-line string literals, so I wouldn't have expected it to be a bottleneck.

Note, the prior ComputeColumn that you might've copied the DCHECK from had more callers -- it looks like you're using this differently.

toolchain/lex/lex.cpp
```diff
@@ -1041,9 +1044,8 @@ auto Lexer::LexOneCharSymbolToken(llvm::StringRef source_text, TokenKind kind,
       << "' instead of the spelling '" << kind.fixed_spelling()
       << "' of the incoming token kind '" << kind << "'";
 
-  TokenIndex token = buffer_.AddToken({.kind = kind,
-                                       .token_line = current_line(),
-                                       .column = ComputeColumn(position)});
+  TokenIndex token = buffer_.AddToken(
+      {.kind = kind, .byte_offset = static_cast<int32_t>(position)});
```
jonmeow, line 1050: Minor, but where you have `byte_offset = static_cast`, is the cast itself needed? (you have this in a few spots)

toolchain/lex/lex.cpp
```diff
@@ -1158,7 +1157,7 @@ auto Lexer::LexKeywordOrIdentifier(llvm::StringRef source_text,
   CARBON_CHECK(
       IsIdStartByteTable[static_cast<unsigned char>(source_text[position])]);
 
-  int column = ComputeColumn(position);
+  int32_t byte_offset = position;
```
jonmeow, line 1163: Where you do this, would it be worth a comment reminding that `position` is modified later?

Maybe we should rename `position`, like `byte_cursor` might be clearer? (note, not asking for a rename here, just mulling whether it's worth it given the `byte_offset` name)

toolchain/lex/tokenized_buffer.h
```diff
@@ -322,6 +307,7 @@ class TokenizedBuffer : public Printable<TokenizedBuffer> {
                            SourceBuffer& source)
       : value_stores_(&value_stores), source_(&source) {}
 
+  auto FindLineIndexImpl(int32_t offset) const -> LineIndex;
```
jonmeow: I see a `GetLine`, which this could be an `Impl` for as in `GetLineImpl`, but I don't see a `FindLineIndex` for this to be an `Impl` of. Maybe one of:

- `FindLineIndex`
- `GetLineImpl`
- `GetLineForOffset`

toolchain/lex/tokenized_buffer.h
```diff
@@ -292,23 +289,11 @@ class TokenizedBuffer : public Printable<TokenizedBuffer> {
...
-    // The length will always be assigned later. Indent may be assigned if
-    // non-zero.
-    explicit LineInfo(int64_t start)
-        : start(start),
-          length(static_cast<int32_t>(llvm::StringRef::npos)),
-          indent(0) {}
-
-    explicit LineInfo(int64_t start, int32_t length)
-        : start(start), length(length), indent(0) {}
+    explicit LineInfo(int64_t start) : start(start), indent(0) {}
```
jonmeow: Why int64_t here?
```suggestion
    explicit LineInfo(int32_t start) : start(start), indent(0) {}
```

toolchain/lex/tokenized_buffer.h
jonmeow: Why just `offset` here but `byte_offset` elsewhere? Comments could also help explain what this is an offset into.

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -355,40 +373,32 @@ auto TokenizedBuffer::SourceBufferDiagnosticConverter::ConvertLoc(
...
-    // Look for the next newline since we don't know the length. We can start at
-    // the column because prior newlines will have been lexed.
-    auto end_newline_pos = line.find('\n', column_number);
-    if (end_newline_pos != llvm::StringRef::npos) {
-      line = line.take_front(end_newline_pos);
-    }
-  }
+  // Grab the line from the buffer, either by slicing from this line to the next
+  // minus the newline, or if on the last line from the start to the end of the
+  // buffer.
```
jonmeow: Small edit, missing comma:

```suggestion
  // Grab the line from the buffer, either by slicing from this line to the next
  // minus the newline, or if on the last line, from the start to the end of the
  // buffer.
```

Larger edit, splitting out a sentence:

```suggestion
  // Grab the line from the buffer by slicing from this line to the next
  // minus the newline. When on the last line, instead use the start to the end of the
  // buffer.
```

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -313,6 +311,26 @@ auto TokenizedBuffer::PrintToken(llvm::raw_ostream& output_stream,
   output_stream << " },";
 }
 
+auto TokenizedBuffer::FindLineIndexImpl(int32_t offset) const -> LineIndex {
+  CARBON_DCHECK(!line_infos_.empty());
+  const auto* line_it =
+      std::partition_point(line_infos_.begin(), line_infos_.end(),
+                           [offset](const LineInfo& line_info) {
```
jonmeow: Is the reference here deliberate/required? This is a small struct, so IIUC it's more efficient if we can not have it:

```suggestion
                           [offset](LineInfo line_info) {
```

jonmeow (approved):
Looks good! I think this also ends up being a nice simplification for code.

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -313,6 +311,26 @@ auto TokenizedBuffer::PrintToken(llvm::raw_ostream& output_stream,
   output_stream << " },";
 }
 
+auto TokenizedBuffer::FindLineIndexImpl(int32_t offset) const -> LineIndex {
+  CARBON_DCHECK(!line_infos_.empty());
+  const auto* line_it =
+      std::partition_point(line_infos_.begin(), line_infos_.end(),
+                           [offset](LineInfo line_info) {
+                             return line_info.start <= offset;
+                           });
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  const auto* line_it = std::partition_point(
      line_infos_.begin(), line_infos_.end(),
      [offset](LineInfo line_info) { return line_info.start <= offset; });
```

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -355,40 +373,32 @@
...
-    // Look for the next newline since we don't know the length. We can start at
-    // the column because prior newlines will have been lexed.
-    auto end_newline_pos = line.find('\n', column_number);
-    if (end_newline_pos != llvm::StringRef::npos) {
-      line = line.take_front(end_newline_pos);
-    }
-  }
+  // Grab the line from the buffer by slicing from this line to the next
+  // minus the newline. When on the last line, instead use the start to the end of the
+  // buffer.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  // minus the newline. When on the last line, instead use the start to the end
  // of the buffer.
```

toolchain/lex/tokenized_buffer.h
chandlerc: Went with `FindLineIndex`.

chandlerc: No reason, switched to `byte_offset`. Comment still useful?

toolchain/lex/lex.cpp
chandlerc: I have this in a few places, does it seem worth a comment in all? just here?

toolchain/lex/lex.cpp
chandlerc: Yes, because of the `{}`s, we get a warning

toolchain/lex/lex.cpp
chandlerc: Moved back to CHECK, I don't think this was important for performance.

toolchain/lex/tokenized_buffer.h
jonmeow: Thanks!

I still appreciate comments on functions, but understand if you're going to decline.

toolchain/lex/lex.cpp
jonmeow: I'm not sure how many. What I'm trying to figure out, particular with a `byte_cursor` rename, is if a different name is a better approach than just a comment.

If you don't think it's helpful enough, we can avoid a comment.

chandlerc: No, I think the comment makes sense. I've put it on all of them. And can follow up with renames, there are a bunch of renamings that I think will make sense, but want to flush the stack of optimizations first for merge simplicity.

toolchain/lex/tokenized_buffer.h
chandlerc: Comment added (to the definition since this is supposed to be an implementation detail function). Happy to add / move as useful in follow up.

chandlerc (commented):
Thanks again, I think all the review is addressed so merging!

PR #4270, opened 2024-09-03
Bit-pack the lexer's token info

chandlerc:
This makes each token info consist of 8 bytes of data:
- 1 byte of the kind
- 1 bit for whitespace tracking
- 23 bits of payload
- 32 bits for byte offset in the file

This builds directly on representing the location of the token as
a single 32-bit offset, now compressing the rest of the data into
a single 32-bit bitfield.

This adds some implementation limits: we can no longer lex more than
2^23 tokens in a single source file. Nor can we have more than 2^23
string literals, integer literals, real literals, or identifiers. Only
the first of these is even close to an issue, and even then seems
unlikely to ever be a problem in practice.

The memory efficiency here is great and the motivating goal. But to make
this work well, we also need to streamline how we create the tokens.
Otherwise, all the bit fiddling can end up erasing our gains. This PR
adds a number of APIs to manage creating and accessing the now
significantly more complex storage of token infos to try and help with
this.

One big change required to simplify the writes here is to switch from
computing whether a token has trailing space after-the-fact to
pre-computing whether a token will have leading space. That lets us have
the leading space information available immediately when forming the
token, and avoids doing a single bit flip afterward.

Another change that helps with this representation is to minimize the
updating of groups after-the-fact. The code now tries to set the opening
index directly when creating the closing token and only updates the
opening group afterward. Because of the bit packing, this is a reduction
of 0.5% of dynamic instructions in the compile benchmark, and has
dramatic improvements for the grouping symbol focused benchmarks.

All combined, this is a significant improvement on the lexer-focused
benchmarks despite the added complexity, and a significant win on our
compile time benchmarks due to both the lexer improvements and
downstream memory density improvements: 5-12% reduction in lex time,
growing larger as files get larger. About a 4.5% reduction in parse
time, and even a 1-2% reduction in total check time. =D


toolchain/lex/lex.cpp
```diff
@@ -93,19 +97,61 @@ class [[clang::internal_linkage]] Lexer {
...
 
-  auto NoteWhitespace() -> void {
-    buffer_.token_infos_.back().has_trailing_space = true;
+  auto next_line_info() -> LineInfo* {
+    CARBON_DCHECK(line_index_ + 1 <
+                  static_cast<ssize_t>(buffer_.line_infos_.size()));
+    return &buffer_.line_infos_[line_index_ + 1];
   }
 
+  auto NoteWhitespace() -> void;
```
jonmeow: I think this function was easier to understand when the implementation was here, maybe it should have a comment now? Perhaps `set_has_leading_whitespace` would be a better name?

Though, with the one line implementation still in this file, why is the implementation now out-of-line?

toolchain/lex/lex.cpp
```diff
@@ -93,19 +97,61 @@ class [[clang::internal_linkage]] Lexer {
...
+
+  auto AddLexedToken(TokenInfo info) -> TokenIndex;
```
jonmeow: This is also clearing some state, maybe it should have a comment?

toolchain/lex/tokenized_buffer.h
```diff
@@ -260,55 +260,101 @@ class TokenizedBuffer : public Printable<TokenizedBuffer> {
...
-      int32_t error_length;
-    };
+    unsigned has_leading_space_ : 1;
+
+    // Payload, typically representing an index into a kind-dependent array.
+    unsigned token_payload_ : PayloadBits;
+
+    // Zero-based byte offset of the token within the file.
+    int32_t byte_offset_;
   };
```
jonmeow, line 415: Is it worth adding a `static_assert(sizeof(TokenInfo) == 8)` to ensure the packing is WAI?

toolchain/lex/lex.cpp
```diff
@@ -745,6 +793,59 @@ auto Lexer::MakeLines(llvm::StringRef source_text) -> void {
   line_index_ = 0;
 }
 
+auto Lexer::NoteWhitespace() -> void { has_leading_space_ = true; }
+
+auto Lexer::AddLexedToken(TokenInfo info) -> TokenIndex {
+  auto token = buffer_.AddToken(info);
+  has_leading_space_ = false;
+  return token;
```
jonmeow: Had you considered reordering:

```suggestion
  has_leading_space_ = false;
  return buffer_.AddToken(info);
```

toolchain/lex/lex.cpp
```diff
@@ -173,6 +219,8 @@ class [[clang::internal_linkage]] Lexer {
 
   ssize_t line_index_;
 
+  bool has_leading_space_ = false;
```
jonmeow, line 209: Is it worth starting to add comments about the way these are used? What made you choose "leading" here, other than that previously there was "trailing" whitespace on a token?

toolchain/lex/tokenized_buffer.h
```diff
@@ -260,32 +260,93 @@ class TokenizedBuffer : public Printable<TokenizedBuffer> {
...
+
+    auto byte_offset() const -> int32_t { return byte_offset_; }
+
+    TokenInfo(TokenKind kind, bool has_leading_space, int32_t byte_offset)
+        : kind_(kind.AsInt()),
+          has_leading_space_(has_leading_space),
+          byte_offset_(byte_offset) {}
+
+    TokenInfo(TokenKind kind, bool has_leading_space, int payload,
+              int32_t byte_offset)
```
jonmeow: Why do you provide different constructors for these? i.e., had you considered a `static constexpr TokenInfo::NoPayload = 0` placeholder? Could the first constructor be delegated to the second (`: TokenInfo(kind, has_leading_space, 0, byte_offset)`)?

toolchain/lex/tokenized_buffer.h
```diff
@@ -260,32 +260,93 @@ class TokenizedBuffer : public Printable<TokenizedBuffer> {
...
+    }
+
+   private:
+    static constexpr int PayloadBits = 23;
+
+    unsigned kind_ : sizeof(TokenKind) * 8;
 
     // Whether the token has trailing whitespace.
-    bool has_trailing_space = false;
+    unsigned has_leading_space_ : 1;
```
jonmeow: Why do you use `unsigned` for these instead of `TokenKind::RawEnumType` and `bool`? Note, I think `RawEnumType` allows `TokenKind::FromInt` -> `TokenKind::Make` and `kind_(kind.AsInt())` -> `kind_(kind)`

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -291,20 +298,20 @@ auto TokenizedBuffer::PrintToken(llvm::raw_ostream& output_stream,
...
       break;
   }
 
-  if (token_info.has_trailing_space) {
-    output_stream << ", has_trailing_space: true";
+  if (token_info.has_leading_space()) {
+    output_stream << ", has_leading_space: true";
   }
-  if (token_info.is_recovery) {
+  if (!recovery_tokens_.empty() && recovery_tokens_[token_index]) {
```
jonmeow: Can this use the helper?

```suggestion
  if (IsRecoveryToken(token)) {
```

toolchain/lex/tokenized_buffer.h
```diff
@@ -260,32 +260,93 @@ class TokenizedBuffer : public Printable<TokenizedBuffer> {
...
+      CARBON_DCHECK(kind().is_closing_symbol());
+      return TokenIndex(token_payload_);
+    }
+    auto set_opening_token_index(TokenIndex opening_index) -> void {
+      CARBON_DCHECK(kind().is_closing_symbol());
+      CARBON_DCHECK(opening_index.index < (2 << PayloadBits));
+      token_payload_ = opening_index.index;
+    }
+
+    auto error_length() const -> int {
```
jonmeow, line 352: Is the `int` deliberate?

```suggestion
    auto error_length() const -> int32_t {
```

toolchain/lex/tokenized_buffer.h
```diff
@@ -260,32 +260,93 @@ class TokenizedBuffer : public Printable<TokenizedBuffer> {
...
+          has_leading_space_(has_leading_space),
+          byte_offset_(byte_offset) {}
+
+    TokenInfo(TokenKind kind, bool has_leading_space, int payload,
+              int32_t byte_offset)
+        : kind_(kind.AsInt()),
+          has_leading_space_(has_leading_space),
+          token_payload_(payload),
+          byte_offset_(byte_offset) {
+      CARBON_DCHECK(payload >= 0 && payload < (2 << PayloadBits));
```
jonmeow: I was really looking hard for a negative value check on payload, not sure if a note here would've caught my attention, but maybe?

```suggestion
      CARBON_DCHECK(payload >= 0 && payload < (2 << PayloadBits)) << "Payload won't fit into unsigned bit pack: " << payload;
```

toolchain/lex/lex.cpp
```diff
@@ -1322,44 +1425,55 @@ class Lexer::ErrorRecoveryBuffer {
...
     CARBON_CHECK(insert_before.index > 0)
         << "Cannot insert before the start of file token.";
+    CARBON_CHECK(insert_before.index <
+                 static_cast<int>(buffer_.token_infos_.size()))
+        << "Cannot insert after the end of file token.";
     CARBON_CHECK(new_tokens_.empty() ||
                  new_tokens_.back().first <= insert_before)
         << "Insertions performed out of order.";
 
+    // If the token we're inserting before had leading whitespace, mark the
```
jonmeow: Sorry, I'm mentally trying to figure out if "The token we're inserting before" means "inserting previously" (i.e., the token being inserted previously had leading whitespace) or "the next token". Would it work for you to refer directly to the parameter:

```suggestion
    // If the `insert_before` token has leading whitespace, mark the
```

But if you prefer, I think the tense of "has" versus "had" is also load bearing:

```suggestion
    // If the token we're inserting before has leading whitespace, mark the
```

toolchain/lex/tokenized_buffer.h
```diff
@@ -260,32 +260,93 @@ class TokenizedBuffer : public Printable<TokenizedBuffer> {
...
+    }
+
+    auto byte_offset() const -> int32_t { return byte_offset_; }
+
+    TokenInfo(TokenKind kind, bool has_leading_space, int32_t byte_offset)
+        : kind_(kind.AsInt()),
+          has_leading_space_(has_leading_space),
+          byte_offset_(byte_offset) {}
+
+    TokenInfo(TokenKind kind, bool has_leading_space, int payload,
```
jonmeow, line 394: Had you considered making `payload` either an `int32_t` to match the caller use-case or `unsigned` to match the storage type?

toolchain/lex/lex.cpp
```diff
@@ -1322,44 +1425,55 @@ class Lexer::ErrorRecoveryBuffer {
...
 
   // Replace the given token with an error token. We do this immediately,
   // because we don't benefit from buffering it.
   auto ReplaceWithError(TokenIndex token) -> void {
     auto& token_info = buffer_.GetTokenInfo(token);
-    token_info.error_length = buffer_.GetTokenText(token).size();
-    token_info.kind = TokenKind::Error;
+    int error_length = buffer_.GetTokenText(token).size();
+    bool has_leading_space = token_info.has_leading_space();
+    int32_t byte_offset = token_info.byte_offset();
```
jonmeow: Had you considered inlining these two variables? It feels like reducing the amount to read might be helpful here.

Though, comparing with `set_closing_token`, had you considered adding `set_error(int32_t error_length)` as an API?

toolchain/lex/lex.cpp
```diff
@@ -745,6 +793,59 @@ auto Lexer::MakeLines(llvm::StringRef source_text) -> void {
...
+      kind != TokenKind::IntLiteral && kind != TokenKind::RealLiteral);
+  return AddLexedToken(TokenInfo(kind, has_leading_space_, byte_offset));
+}
+
+template <TokenKind::RawEnumType Kind>
+  requires(Kind == TokenKind::Identifier)
+auto Lexer::LexToken(IdentifierId id, int32_t byte_offset) -> TokenIndex {
+  return AddLexedToken(TokenInfo(TokenKind::Identifier, has_leading_space_,
+                                 id.index, byte_offset));
+}
```
jonmeow: What made you choose to define these out-of-line rather than inline? They seem generally short, with more syntax for the templating than the body.

toolchain/lex/lex.cpp
```diff
@@ -745,6 +793,59 @@ auto Lexer::MakeLines(llvm::StringRef source_text) -> void {
...
+auto Lexer::AddLexedToken(TokenInfo info) -> TokenIndex {
+  auto token = buffer_.AddToken(info);
+  has_leading_space_ = false;
+  return token;
+}
+
+auto Lexer::LexToken(TokenKind kind, int32_t byte_offset) -> TokenIndex {
+  CARBON_DCHECK(
+      kind != TokenKind::Identifier && kind != TokenKind::StringLiteral &&
+      kind != TokenKind::IntLiteral && kind != TokenKind::RealLiteral);
```
jonmeow: Why are type literals allowed here, when there's a specific overload for them? What about `TokenKind::Error` too?

That is, how is this specific list chosen?

toolchain/lex/lex.cpp
```diff
@@ -745,6 +793,59 @@ auto Lexer::MakeLines(llvm::StringRef source_text) -> void {
...
+  return AddLexedToken(TokenInfo(kind, has_leading_space_, byte_offset));
+}
+
+template <TokenKind::RawEnumType Kind>
+  requires(Kind == TokenKind::Identifier)
+auto Lexer::LexToken(IdentifierId id, int32_t byte_offset) -> TokenIndex {
+  return AddLexedToken(TokenInfo(TokenKind::Identifier, has_leading_space_,
```
jonmeow: In these overloads, had you considered using `TokenKind::Make(Kind)` everywhere rather than only in the IntLiteral-plus overload? I think there might be some value to consistency with that, also making it clear how the `Kind` parameter is used.

Alternately, had you considered removing the template parameter because overload selection could rely on the `IdentifierId` and similar? Or, versus `LexToken<TokenKind::Identifier>`, perhaps `LexIdentifierToken` and similar? (perhaps even `LexNumericToken` for the multi-kind one below)

toolchain/lex/tokenized_buffer.h
```diff
@@ -260,55 +260,101 @@ class TokenizedBuffer : public Printable<TokenizedBuffer> {
...
+              int32_t byte_offset)
+        : kind_(kind.AsInt()),
+          has_leading_space_(has_leading_space),
+          token_payload_(payload),
+          byte_offset_(byte_offset) {
+      CARBON_DCHECK(payload >= 0 && payload < (2 << PayloadBits));
+    }
+
+   private:
+    static constexpr int PayloadBits = 23;
```
geoffromer: I don't know if this is going to be enough. 2^23 is about 8 million, but I've found a `.proto` file whose generated C++ header contains over 44 million whitespace-separated words (which is likely an undercount of the number of C++ tokens).

toolchain/lex/lex.cpp
```diff
@@ -1044,41 +1130,67 @@ auto Lexer::LexOneCharSymbolToken(llvm::StringRef source_text, TokenKind kind,
...
+      << "' instead of the spelling '" << kind.fixed_spelling()
+      << "' of the incoming token kind '" << kind << "'";
+
+  int32_t byte_offset = position;
+  ++position;
+
+  // Lex the opening symbol with a zero closing index. We'll update this
+  // later if we successfully match it to a closing symbol.
+  TokenIndex token =
+      AddLexedToken(TokenInfo(kind, has_leading_space_, 0, byte_offset));
```
jonmeow: What made you choose to add the `Lexer::LexToken` overloads instead of having these provided as constructors on `TokenInfo` and writing `AddLexedToken(TokenInfo(...))`? This particularly feels odd to me since, in some spots, code is doing `AddLexedToken(TokenInfo(...))` directly (e.g., `LexOpeningSymbolToken`). Why don't the calls align?

To note there are a few paths above -- I think you are providing multiple ways to handle the differences:

- Overloads of `LexToken`, which pass the appropriate payload setup to `TokenInfo`.
- Two `TokenInfo` constructors, one if there's no payload.
- Calling `AddLexedToken` directly with an explicit `0` for no payload.

Particularly with the last, maybe you were iterating and missed it? I'd suggest choosing between:

- One `TokenInfo` constructor, have `LexToken` choose the payload (including `0`).
- Multiple `TokenInfo` constructors, remove `LexToken`, only do `AddLexedToken(TokenInfo(...))`

I have a slight leaning towards the latter because you couldn't have code "accidentally" do `AddLexedToken(TokenInfo(...))` itself (it would be the right way to do it), but `LexToken` is less typing.

jonmeow (commented):
This generally looks good, although I have some questions about the `AddLexedToken` + `TokenInfo` constructor setup.

toolchain/lex/tokenized_buffer.h
```diff
@@ -260,55 +260,101 @@ class TokenizedBuffer : public Printable<TokenizedBuffer> {
...
+
+    unsigned kind_ : sizeof(TokenKind) * 8;
```
geoffromer: It seems like the performance-tuning in this PR is based on the assumption that `sizeof(TokenKind)` will be 1. If so, it might be clearer and safer to do something like this:

```suggestion
    unsigned kind_ : 8;
    static_assert(sizeof(TokenKind) == 1);
```

toolchain/lex/tokenized_buffer.h
```diff
@@ -260,55 +260,101 @@ class TokenizedBuffer : public Printable<TokenizedBuffer> {
     int indent;
   };
 
-  struct TokenInfo {
-    TokenKind kind;
+  class TokenInfo {
+   public:
+    auto kind() const -> TokenKind { return TokenKind::FromInt(kind_); }
+    auto has_leading_space() const -> bool { return has_leading_space_; }
+
```
geoffromer: This type is somewhat harder to navigate in its new form. Something like this seems like it might help:

```suggestion
  // Stores lexical information about a token: its kind, whether it has a leading
  // space, its byte offset in the file, and a kind-dependent payload.
  class TokenInfo {
   public:
    auto kind() const -> TokenKind { return TokenKind::FromInt(kind_); }
    auto has_leading_space() const -> bool { return has_leading_space_; }
    auto byte_offset() const -> int32_t { return byte_offset_; }
    
    // Accessors/mutators for the payload. At most one payload accessor is valid
    // for a given kind, along with its corresponding mutator (if any).
```

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -293,26 +298,46 @@ auto TokenizedBuffer::PrintToken(llvm::raw_ostream& output_stream,
...
     output_stream << ", recovery: true";
   }
 
   output_stream << " },";
 }
 
+auto TokenizedBuffer::FindLineIndexImpl(int32_t offset) const -> LineIndex {
```
geoffromer: The `Impl` seems unnecessary; it makes me look for the corresponding `FindLineIndex`, which doesn't exist. How about calling this `GetLineIndex` or `GetLineIndexAtOffset`?

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -293,26 +298,46 @@ auto TokenizedBuffer::PrintToken(llvm::raw_ostream& output_stream,
...
+  const auto* line_it =
+      std::partition_point(line_infos_.begin(), line_infos_.end(),
+                           [offset](const LineInfo& line_info) {
+                             return line_info.start <= offset;
+                           });
+  --line_it;
+
+  // If this isn't the first line but it starts past the end of the source, then
+  // this is a synthetic line added for simplicity of lexing. Step back one
+  // further to find the last non-synthetic line.
```
geoffromer: This seems like it should be mentioned in the comment on `line_infos_` and/or `LineInfo`.

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -355,40 +380,32 @@ auto TokenizedBuffer::SourceBufferDiagnosticConverter::ConvertLoc(
...
-  // Find the first line starting after the given location. Note that we can't
-  // inspect `line.length` here because it is not necessarily correct for the
-  // final line during lexing (but will be correct later for the parse tree).
-  const auto* line_it = std::partition_point(
+  // Find the first line starting after the given location.
+  const auto* next_line_it = std::partition_point(
       buffer_->line_infos_.begin(), buffer_->line_infos_.end(),
       [offset](const LineInfo& line) { return line.start <= offset; });
 
   // Step back one line to find the line containing the given position.
```
geoffromer: Do you also need to deal with the issue of synthetic lines here, like in `FindLineIndexImpl`?

toolchain/lex/tokenized_buffer.cpp
```diff
@@ -355,40 +380,32 @@ auto TokenizedBuffer::SourceBufferDiagnosticConverter::ConvertLoc(
...
-    // Look for the next newline since we don't know the length. We can start at
-    // the column because prior newlines will have been lexed.
-    auto end_newline_pos = line.find('\n', column_number);
-    if (end_newline_pos != llvm::StringRef::npos) {
-      line = line.take_front(end_newline_pos);
-    }
-  }
+  // Grab the line from the buffer, either by slicing from this line to the next
+  // minus the newline, or if on the last line from the start to the end of the
+  // buffer.
```
geoffromer:
```suggestion
  // minus the newline, or if on the last line from the start of the line to the
  // end of the buffer.
```

toolchain/lex/lex.cpp
```diff
@@ -93,19 +97,61 @@ class [[clang::internal_linkage]] Lexer {
...
+  auto AddLexedToken(TokenInfo info) -> TokenIndex;
+
+  // A collection of overloads that provide type-checked APIs for lexing
+  // specific kinds of tokens and providing the appropriate payload data for
+  // each of them.
+  //
+  // The base overload allows a runtime token kind for any kind that doesn't
+  // require a payload at all. Token kinds that require a payload are an error
+  // to use with this overload.
+  auto LexToken(TokenKind kind, int32_t byte_offset) -> TokenIndex;
```
geoffromer: I think the first overload should be separated from the rest, because it has distinctly different semantics (e.g. the meaning of the first parameter). Among other things, I think we can document the remaining overloads more clearly if we're not documenting the first one at the same time:

```suggestion
  // Lexes a token with the given kind at the given offset. `kind` must not be
  // a kind that requires a payload.
  auto LexToken(TokenKind kind, int32_t byte_offset) -> TokenIndex;

  // Lexes a token with the given kind at the given offset, with the given payload.
  // `Payload` must be the payload type associated with `Kind`.
  template <TokenKind::RawEnumType Kind, typename Payload>
  auto LexTokenWithPayload(Payload payload, int32_t byte_offset) = delete;
```

If you make this change, I think you can probably drop the comments from the overloads below. It might also be worthwhile to express the overloads as specializations, e.g:

```cpp
  template <>
  auto LexToken<TokenKind::Identifier>(IdentifierId id, int32_t byte_offset) -> TokenIndex;
```

toolchain/lex/lex.cpp
```diff
@@ -140,9 +186,14 @@ class [[clang::internal_linkage]] Lexer {
   auto LexSymbolToken(llvm::StringRef source_text, ssize_t& position)
       -> LexResult;
 
+  template <TokenKind::RawEnumType Kind>
+  auto LexTypeLiteralToken(llvm::StringRef suffix, int32_t byte_offset)
```
geoffromer: I think it would be helpful for this function to have a comment, especially to clarify how this differs from `LexWordAsTypeLiteralToken`

toolchain/lex/lex.cpp
chandlerc: I was trying to move all the non-getters to consistently put their definitions in the same place.

I think most of them are out-of-line to reduce indentation, and I'd rather the consistency I think?

Happy to add a comment though.

toolchain/lex/lex.cpp
chandlerc: Yeah, this one badly needs it, sorry. Added.

toolchain/lex/lex.cpp
chandlerc: Yes, went back and forth on this. I guess let's go with your suggested ordering.

toolchain/lex/lex.cpp
chandlerc: The list just got stale, it should preclude those too. Added.

toolchain/lex/lex.cpp
chandlerc: I really want to minimize the mutating API on `TokenInfo`, because it's brittle and can be slow... I think we want the API to somewhat strongly encourage one-shot construction. That's why I've been trying to remove all the setters. The closing token is the only one that somewhat fundamentally can't be removed.

I could add the error setter I guess, as that is at least not on the critical path.

As for why I didn't inline the variable, because the `token_info` is on the LHS of the assignment, I find it confusingly like self-assignment to have both accessors on the RHS and the variable on the LHS below. Extracting variables makes it clear what the sequencing here is.

But if we move to an error mutating API, that doesn't matter I guess.

toolchain/lex/tokenized_buffer.cpp
chandlerc: This is part of the other PR, and was renamed there.

toolchain/lex/tokenized_buffer.cpp
chandlerc: This too is part of the other PR -- the concern didn't come up, but I can send a follow-up if you want?

(There are comments about this, but in the `MakeLines` routine of the lexer, not on the data structure itself.)

toolchain/lex/tokenized_buffer.cpp
chandlerc: This is part of the previous PR, and also isn't changed by the PR, so I didn't want to update it without having a test case or reason to change it. The synthetic line has been there for a long time, so this code seems to be working as-is. But the above actually needs to handle it.

toolchain/lex/tokenized_buffer.cpp
chandlerc: (Another from the other PR)

toolchain/lex/tokenized_buffer.h
chandlerc: Will add comments (but somewhat different ones), PTAL

toolchain/lex/tokenized_buffer.h
chandlerc: This isn't stored once we extract it, so was trying to revert to non-bit-sized types.

toolchain/lex/tokenized_buffer.h
chandlerc: I think some of the callers just have an arbitrary `int`.

Generally, I think `int32_t` is best when we're actually storing it in memory, or *about* to store it in memory and it gets the types to match exactly.

But here, the types can't match exactly, so I went with a generic signed integer type on the API surface.

The `unsigned` should be an implementation detail of the bitfield where we don't want to store a sign bit.

toolchain/lex/tokenized_buffer.h
chandlerc: The goal is to leave the payload uninitialized as it should *not* be accessed when the info was constructed in this way. My hope is that this will allow, for example, MSan to poison those bits.

There was some hacky code that I've removed in the lexer that used an explicit 0 payload, but I think that was incorrect and I've updated the code to more strictly use the overload when they don't have a meaningful payload.

toolchain/lex/tokenized_buffer.h
chandlerc: Sure, done!

toolchain/lex/tokenized_buffer.h
chandlerc: I've added a static assert below that should cover this as well.

toolchain/lex/lex.cpp
chandlerc: While this is an interesting approach, I think all of this is just too much complexity for what its worth.

All I really want to handle here is injecting the has_leading_space_ parameter in a common place, and delegating to `AddLexedToken`.

I've re-done this code to not waste time on all the type checked versions. It turns out that for the vast majority, there was only one caller anyways so we were trading a bug here for a bug there which seems like not a win.

toolchain/lex/lex.cpp
chandlerc: Removed this function now that I'm not excessively trying to track the kind as a type here.

toolchain/lex/lex.cpp
chandlerc: Yeah, done.

toolchain/lex/lex.cpp
chandlerc: See longer comment below on this vs. constructors. Maybe best ot focus discussion there.

toolchain/lex/lex.cpp
chandlerc: Just trying to have fewer places to look for definitions. The only non-accessor defined inline previously was `NoteWhitespace`.

toolchain/lex/lex.cpp
chandlerc: So, the real reason I wanted to have `LexToken` is to try to have one place where `has_leading_space_` needs to be remembered and passed in to the token construction, as that is literally the same for every token (unlike byte offset for example).

The secondary thing I like is that it concentrates more of the logic in the lexer rather than in the tokenized buffer that is more about *consuming* tokens.

But I agree that the complex overloads aren't helping here. *And*, as you noticed here, I failed to use them consistently. I've tried to simplify this down to just a payload overload and a non-payload overload. There is only one caller for most of these, and so I think the type checking we're getting isn't really that valuable.

I've also switched this code to use them correctly rather than directly constructing things here. This part of the PR was a bit rushed, sorry about that. Should be much cleaner now.

WDYT?

toolchain/lex/lex.cpp
chandlerc: To be clear, I've switched this to a `ResetToError` method. It still does the constructor call in its body so that it gets the payload size checking without duplicating code.

toolchain/lex/tokenized_buffer.h
chandlerc: Words outside of comments? wow. Can you send me a reference to this?

My generated 256k line file, which is *huge*, still isn't that big.

This would be a file that *at it's smallest* is 150mb of source code text... And much more likely those words average out to 3-4 bytes per word, so we're talking about more like 250mb of source code text.... Maybe we could insist on generated code splitting into multiple files at that point?

toolchain/lex/tokenized_buffer.h
chandlerc: Maybe recent versions of C++ or C++ compilers have fixed this, but I have an old habit of making bitfields unsigned to avoid any implementation oddities around sign extension. My memory is that `bool` specifically broke things in LLVM when used in a bitfield.

The thing I found most uniformly consistent in behavior was to declare *everything* as `unsigned` and manually manage the bits going in and out and then transform them into proper types.

If this is no longer an issue, happy to adjust though. @zygoloid might recall the specific issue that caused us to start following this practice in LLVM code.

chandlerc (commented):
Thanks for all the comments! I think I responded to all of them.

Also have updated to be on top of the trunk with the base PR merged in, so you should be able to view this more normally as an unstacked PR, sorry for the confusion there.

toolchain/lex/lex.cpp
jonmeow: I think in code we're generally only putting class member definitions out-of-line when they're too long to fit well. Under that model, I'd lean towards consistency being inline.

For reference, I'd taken the movement with Lex* as being because most of those are large bodies, rather than indentation-motivated. Above, current_line_info and next_line_info similarly have short definitions inline. I wouldn't view it as an issue to find definitions when they're inline; it's mainly an issue finding out-of-line definitions when they're in multiple locations (e.g., template things out-of-line in a header but non-template things out-of-line in a cpp file).

Note, I don't consider this preference blocking if you prefer it as-is.

toolchain/lex/lex.cpp
```diff
@@ -967,29 +1014,24 @@ auto Lexer::LexNumericLiteral(llvm::StringRef source_text, ssize_t& position)
...
-        return token;
+        return LexToken(
+            TokenKind::RealLiteral,
+            buffer_.value_stores_->reals()
+                .Add(Real{.mantissa = value.mantissa,
+                          .exponent = value.exponent,
+                          .is_decimal =
+                              (value.radix == NumericLiteral::Radix::Decimal)})
+                .index,
+            byte_offset);
```
jonmeow:
```suggestion
        auto real_id = buffer_.value_stores_->reals().Add(Real{
            .mantissa = value.mantissa,
            .exponent = value.exponent,
            .is_decimal = (value.radix == NumericLiteral::Radix::Decimal)});
        return LexToken(TokenKind::RealLiteral, real_id.index, byte_offset);
```

Small suggestion, I think separating the add may make this easier to read.

toolchain/lex/tokenized_buffer.h
jonmeow: You understand uninitialized state better than me, but my impression would be that in opt mode we wouldn't get MSan; it would just be uninitialized value. By contrast, a NoPayload of (in particular, new idea) `2 << PayloadBits - 1` is something you could DCHECK on, and in opt mode might yield more reliable errors.

Given your intent, perhaps it should be commented why no explicit payload is set?

toolchain/lex/tokenized_buffer.h
jonmeow: Looking around, in [subst.cpp](https://github.com/carbon-language/carbon-lang/blob/trunk/toolchain/check/subst.cpp#L21-L27) zygoloid added typed bit-packed values, so agreed it'd be good to figure out.

Note, I would still expect the accessors because the `RawEnumType` is unwieldy to work with directly.

toolchain/lex/tokenized_buffer.h
geoffromer: "Words" is as defined by `wc`, so it includes comment words. And yes, it's over 850 Mb.

toolchain/lex/lex.cpp
jonmeow: SGTM

toolchain/lex/lex.cpp
jonmeow: This looks good, thanks!

jonmeow (approved):
Thanks for the changes! For my part, I'm good with the change, with a few small comments that you can choose whether to make changes. Approving, assuming you'll address geoffromer's questions before merging.

toolchain/lex/lex.cpp
```diff
@@ -94,21 +97,38 @@ class [[clang::internal_linkage]] Lexer {
...
+  auto NoteWhitespace() -> void;
+
+  // Add a lexed token to the tokenized buffer, and reset any token-specific
+  // state tracked in the lexer for the next token.
+  auto AddLexedToken(TokenInfo info) -> TokenIndex;
+
+  // An overload for lexing a token.Builds the correctly encoded token info,
+  // adds it to the tokenized buffer and returns the token index.
+  //
+  // This overload is for tokens without a payload.
```
geoffromer:
```suggestion
  // Lexes a token with no payload: builds the correctly encoded token info,
  // adds it to the tokenized buffer and returns the token index.
```

Per the style guide, function comments should start with a verb phrase with "this function" as the implied subject. It also seems helpful (and more concise) to frontload the payload aspect a little more, but I don't feel strongly about that part.

toolchain/lex/lex.cpp
```diff
@@ -94,21 +97,38 @@ class [[clang::internal_linkage]] Lexer {
...
+  auto LexToken(TokenKind kind, int32_t byte_offset) -> TokenIndex;
+
+  // An overload for lexing a token.Builds the correctly encoded token info,
+  // adds it to the tokenized buffer and returns the token index.
+  //
+  // This overload is for tokens with a payload.
```
geoffromer:
```suggestion
  // Lexes a token with a payload: builds the correctly encoded token info,
  // adds it to the tokenized buffer and returns the token index.
```

toolchain/lex/lex.cpp
```diff
@@ -94,21 +97,38 @@ class [[clang::internal_linkage]] Lexer {
...
+  auto LexToken(TokenKind kind, int token_payload, int32_t byte_offset)
```
geoffromer: It seems like it would be better to put `token_payload` last. When overload sets have different numbers of arguments, the shorter parameter list is usually a prefix of the longer one, as if the overload set was a single function with defaulted parameters. It seems especially important for the parameter order to be unsurprising in this case, since the caller generally won't get a type error if they get it wrong.

toolchain/lex/tokenized_buffer.h
geoffromer: The class comment is great, but I think it would still be helpful to move `byte_offset` up, and add a comment that collectively documents the payload accessors and mutators (especially their preconditions). That keeps `byte_offset` from getting lost at the bottom, and avoids the need to add a comment to each payload method (which the style guide arguably calls for, but I don't think would be helpful here).

geoffromer (commented):
Gah, sorry, I saw your warning that this was a stacked PR, but then failed to set my diff view properly. I've resolved the comments that concern the previous PR.

toolchain/lex/tokenized_buffer.h
chandlerc: My problem with reordering the accessors is that as-is they follow the order of storage in tthe struct. And I think that order is important for something whose entire business is careful layout in memory.

And moreover, as I indicated in the class comment, `byte_offset` is importantly different from all the others, and might be usefully separated out of this struct and into some parallel structure. That also motivated me to keep the others grouped together.

Is there another change that would help the byte offset not be lost for you, besides mingling the order? (I'm having trouble coming up with ideas myself I'm afraid, the byte offset doesn't get lost for me.)

toolchain/lex/lex.cpp
chandlerc: The downside is that the payload is tightly coupled with the kind, while the byte offset is much less interesting and useful to think about separately.

On the balance, I find the current structure more readable, if not perfect.

But if this is a big issue for folks, we don't need these to be overloads -- would it help to have a name to distinguish them?

toolchain/lex/lex.cpp
chandlerc: Eh, sure, moved these inline.

toolchain/lex/tokenized_buffer.h
geoffromer: I can think of a few other options:
- Giving the payload names a common prefix, e.g. `payload_ident_id`, `payload_string_literal_id`. That doesn't help with `byte_offset` getting visually lost, but it at least makes clear which operations are payload operations (and hence makes clear the scope of the collective payload comment). This may not be a net win for the callsites, though.
- Add a method comment on `byte_offset`, as well as the collective payload comment (and then probably comment the other two unconditional accessors, for consistency). That way the comment on `byte_offset` visually flags the end of the payload methods, and calls attention to the fact that there's another method after it. This is arguably what the style guide calls for, but it's not consistent with the toolchain's convention of putting comments on the private members and leaving the accessors uncommented.
- Move the method bodies out of line, and use whitespace to convey the grouping:
   ```cpp
      auto kind() const -> TokenKind;
      auto has_leading_space() const -> bool;

      // Accessors/mutators for the payload. At most one payload accessor is valid
      // for a given kind, along with its corresponding mutator (if any).
      auto ident_id() const -> IdentifierId;
      auto set_ident_id(IdentifierId ident_id) -> void;
      auto string_literal_id() const -> StringLiteralValueId;
      ...
      auto error_length() const;

      auto byte_offset() const -> int32_t;
   ```
   But that's a mixed bag because right now the method bodies act as implicit documentation of which kinds they can be called with, so arguably this should be paired with more explicit documentation of the kind/payload mapping, maybe at the class level.

These all have drawbacks, as does the status quo, but it's probably not worth sinking more time into this question. I'm fine with however you want to deal with it.

toolchain/lex/lex.cpp
geoffromer: I think a separate name will help somewhat, but it won't fully solve the problem unless the name is so different that the reader doesn't think of them as related (which seems undesirable).

I see your point about the problem with putting the payload last. To me it seems like the lesser evil, but it's your call.

geoffromer (approved):

toolchain/lex/tokenized_buffer.h
chandlerc: I like the idea of just adding comments to the accessors.

I've moved and significantly expanded the comments from the fields, and switched the limited field comments to talk about the *storage* peculiarities only, and not as much about the semantic meaning when accessed. I think its a net improvement, bet let me know.

toolchain/lex/tokenized_buffer.h
chandlerc: Added a comment here about the uninitialized memory and sanitizer behavior. (And sorry I missed this comment earlier.)

toolchain/lex/tokenized_buffer.h
chandlerc: (For others, Geoff was able to point me at the specific example.)

While the proposed scheme would preclude this, I think that's a reasonable implementation limit to impose in order to make essentially all other code more efficient.

The code in question is generated code for protocol buffers, where the `.proto` file itself is *also* generated and nearly 9mb long.

I think it would be entirely reasonable to say this isn't supported and insist that users desiring something of this form generate multiple files.

That said, if we in practice find this to be a difficult limitation, I think we can easily remove the limit in question on the number of tokens: instead of storing a token index directly here, we can store an index into a table of open/close pairs -- we can even shard that into separate tables for each kind of grouping symbol. This would allow `2^22 - 1` tokens with inline grouping indices, and if more tokens are found, fall back to separate tables for each kind, and only run out at `2^22 - 1` pairs of a single grouping token. That would be a file with 4mb of curlies or parens alone, and in a single file with no `#include`s. And if we want to go still further, we could begin dropping the cross-reference index at that point and degrade diagnostics instead.

Overall, this seems like a plausible limit to try to hold initially and one we could credibly relax in tactical ways if specific problems arise that we can't reasonably ask users to work around.

toolchain/lex/tokenized_buffer.h
chandlerc: Hmm, ok, that seems reasonable evidence that this works well enough these days. Switched to use slightly better types, and removed the added behavior in token kind that we don't need now. =] Nice!

I kept `unsigned` on the payload because the plan is to use the extra bit and not store negative payload indices, so it seems a reasonable type.

(also, sorry for losing this comment in the discussion)

toolchain/lex/lex.cpp
chandlerc: Did the rename, I think I prefer sticking with that for now.

jonmeow (approved):
I think this change is looking good. I'm not merging just in case geoffromer has more comments.

geoffromer (approved):

PR #4271, opened 2024-09-03
Add TODO breadcrumb from `for` design to semantics proposal

dwblaikie:


jonmeow (approved):
Sure, good to note we're missing a proposal here. :)

PR #4272, opened 2024-09-03
Add SemIR support for virtual functions

dwblaikie:
I guess this technically would also allow code to pass check that hasn't before, and that isn't covered by tests (since it's masked by other failures in the tests that already test this functionality) - should I add another test/add some code to a valid test case?

Also, this'll miscompile in lowering, since there's no support there yet - should I do anything about that to make lowering fail in some way? Or is it acceptable that some things just silently mis-lower? (I could add a currently-miscompiling test case too, to demonstrate this? (not sure if the autogenerated tests leave space for comments that would explain that the currently-tested behavior is incorrect?))

Is the addition to EntityWithParamsBase suitable? of course not all functions can be virtual, so it's a wasted bit at the moment for all those cases (though it's free, since it's bitpacked - but as we want to add more bits in there it might not be a scalable solution)?


toolchain/check/handle_function.cpp
```diff
@@ -194,7 +194,10 @@ static auto BuildFunctionDecl(Context& context,
   DiagnoseModifiers(context, introducer, is_definition, parent_scope_inst_id,
                     parent_scope_inst);
   bool is_extern = introducer.modifier_set.HasAnyOf(KeywordModifierSet::Extern);
-  if (introducer.modifier_set.HasAnyOf(KeywordModifierSet::Method)) {
+  bool is_virtual =
+      introducer.modifier_set.HasAnyOf(KeywordModifierSet::Virtual);
+  if (!is_virtual &&
+      introducer.modifier_set.HasAnyOf(KeywordModifierSet::Method)) {
```
josh11b: I think this only works because the method modifiers are mutually exclusive? Otherwise this would accept `virtual` along with an unsupported method modifier. Maybe worth a comment.

toolchain/check/testdata/class/fail_todo_modifiers.carbon
```diff
@@ -42,10 +38,6 @@ abstract class Abstract {
   // CHECK:STDERR:
   abstract fn J();
 
-  // CHECK:STDERR: fail_todo_modifiers.carbon:[[@LINE+4]]:3: ERROR: Semantics TODO: `method modifier`.
-  // CHECK:STDERR:   virtual fn K();
-  // CHECK:STDERR:   ^~~~~~~
-  // CHECK:STDERR:
   virtual fn K();
```
jonmeow: If these are now working, they shouldn't be in a fail test anymore, and should instead be tested to succeed.

toolchain/check/testdata/class/fail_todo_modifiers.carbon
```diff
@@ -42,10 +38,6 @@ abstract class Abstract {
   // CHECK:STDERR:
   abstract fn J();
```
jonmeow: Maybe also track `abstract` and `impl`, which are both related to `virtual`?

toolchain/sem_ir/entity_with_params_base.h
```diff
@@ -93,7 +93,9 @@ struct EntityWithParamsBase {
   // A block containing a single reference instruction per parameter.
   InstBlockId param_refs_id;
   // True if declarations are `extern`.
-  bool is_extern;
+  bool is_extern : 1;
+  // Tru eif the declarations are `virtual`.
```
jonmeow:
```suggestion
  // True if the declarations are `virtual`.
```

jonmeow:
> I guess this technically would also allow code to pass check that hasn't before, and that isn't covered by tests (since it's masked by other failures in the tests that already test this functionality) - should I add another test/add some code to a valid test case?

Yes, I think there should be tests. Note though, I've commented where this is affecting tests -- those should be cleaned up.

> Also, this'll miscompile in lowering, since there's no support there yet - should I do anything about that to make lowering fail in some way? Or is it acceptable that some things just silently mis-lower? (I could add a currently-miscompiling test case too, to demonstrate this? (not sure if the autogenerated tests leave space for comments that would explain that the currently-tested behavior is incorrect?))

This is fine: code which passes check is not currently assumed to compile correctly (and could crash in lowering).

We're not fuzzing lowering right now because of this.

> Is the addition to EntityWithParamsBase suitable? of course not all functions can be virtual, so it's a wasted bit at the moment for all those cases (though it's free, since it's bitpacked - but as we want to add more bits in there it might not be a scalable solution)?

I think these belong on Function, not EntityWithParamsBase. You're correct not all functions can be virtual, but not all entities are functions and I think that's more significant. I would be reticent to put entity-specific features on EntityWithParamsBase.

Note I don't think we've done any explicit bitpacking in SemIR (at least, not on these entities). It may be something we want to adjust, but right now we're prioritizing readability over saving bytes.

josh11b (commented):
Seems like you have some changes to make based on @jonmeow 's comments, so I'm pausing my review.

toolchain/check/handle_function.cpp
dwblaikie: Oh, right - thanks!

Now that I've implemented @jonmeow's suggestion of also handling abstract and impl, there's nothing left to diagnose here so I've removed this block entirely.

toolchain/check/testdata/class/fail_todo_modifiers.carbon
dwblaikie: Sure, done

toolchain/check/testdata/class/fail_todo_modifiers.carbon
dwblaikie: Done (& renamed the remaining test, because it no longer fails, and since it's specific to access modifiers I made the name more specific to that too)

dwblaikie (commented):
> > Is the addition to EntityWithParamsBase suitable? of course not all functions can be virtual, so it's a wasted bit at the moment for all those cases (though it's free, since it's bitpacked - but as we want to add more bits in there it might not be a scalable solution)?
> 
> I think these belong on Function, not EntityWithParamsBase. You're correct not all functions can be virtual, but not all entities are functions and I think that's more significant. I would be reticent to put entity-specific features on EntityWithParamsBase.

Ah, right, makes sense - done that.

> Note I don't think we've done any explicit bitpacking in SemIR (at least, not on these entities). It may be something we want to adjust, but right now we're prioritizing readability over saving bytes.

Ah, sure - I've removed the bit field stuff then.

dwblaikie:
Oh, I had one other question - is mapping these language features directly to SemIR the right choice? Or should we consider a more semantic description of these properties? (like could all 3 of these properties map to "virtual"? I guess then we wouldn't know the difference between abstract and a virtual function that's defined externally (if that's even possible in Carbon? I assume so) - could have two properties, then, virtual and abstract (& perhaps all abstract things are also marked virtual))?

I guess probably not - but thought it was worth asking/understanding better what semantic representation SemIR is going for.

toolchain/sem_ir/function.h
```diff
@@ -23,6 +23,18 @@ struct FunctionFields {
...
+  // implementation, but can be overridden in a derived class.
+  bool is_virtual;
+
+  // Is this function declaration is abstract - having no implementation, and
+  // must be implemented in a concrete derived class.
+  bool is_abstract;
+
+  // Is this function declaration an implementation - overriding an abstract or
+  // virtual function declared in a base class.
+  bool is_impl;
```
jonmeow: These are mutually exclusive, had you considered an enum?

toolchain/check/testdata/class/dynamic_function_modifiers.carbon
```diff
@@ -0,0 +1,101 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// AUTOUPDATE
+// TIP: To test this file alone, run:
+// TIP:   bazel test //toolchain/testing:file_test --test_arg=--file_tests=toolchain/check/testdata/class/dynamic_function_modifiers.carbon
```
jonmeow: Maybe "virtual_function_modifiers" (or even just "virtual_modifiers")? Although these are one way to get dynamic dispatch, https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/classes.md#virtual-methods calls these "virtual methods" and "virtual override keywords"

toolchain/check/testdata/class/dynamic_function_modifiers.carbon
```diff
@@ -0,0 +1,101 @@
...
+// TIP: To dump output, run:
+// TIP:   bazel run //toolchain/testing:file_test -- --dump_output --file_tests=toolchain/check/testdata/class/dynamic_function_modifiers.carbon
+
+
+base class Base {
+
```
jonmeow:
```suggestion
```

Minor style

toolchain/check/testdata/class/dynamic_function_modifiers.carbon
```diff
@@ -0,0 +1,101 @@
...
+  virtual fn H();
+
+  impl fn I();
+}
+
+abstract class Abstract {
+
```
jonmeow:
```suggestion
```

jonmeow:
> Oh, I had one other question - is mapping these language features directly to SemIR the right choice? Or should we consider a more semantic description of these properties? (like could all 3 of these properties map to "virtual"? I guess then we wouldn't know the difference between abstract and a virtual function that's defined externally (if that's even possible in Carbon? I assume so) - could have two properties, then, virtual and abstract (& perhaps all abstract things are also marked virtual))?
> 
> I guess probably not - but thought it was worth asking/understanding better what semantic representation SemIR is going for.

This is essentially what I'd expect, with the `enum` comment. Note though, I would generally encourage a little more implementation work to see how changes interact -- this is just a straightforward change to me, and some things (e.g. printing of "virtual" in formatter) seem like they might mirror a need to do similar "virtual"/"impl"/"abstract" stringification in diagnostics, which might yield a slightly different approach.

Regarding mapping all to "virtual", each of the keywords implies different semantics, so they do need to be tracked separately.

> could have two properties, then, virtual and abstract (& perhaps all abstract things are also marked virtual))?

For example, it is invalid to declare a final class that inherits from a class without providing `impl` for all the `abstract` functions. That is not true of `virtual` functions either way: `virtual` provides a root of a virtual call, and shadowed `virtual` functions are just that (shadowed, not `impl`s).

toolchain/check/testdata/class/dynamic_function_modifiers.carbon
dwblaikie: Ah, sure - done.

toolchain/sem_ir/function.h
dwblaikie: Done - open to haggling over naming, of course.

dwblaikie:
> > Oh, I had one other question - is mapping these language features directly to SemIR the right choice? Or should we consider a more semantic description of these properties? (like could all 3 of these properties map to "virtual"? I guess then we wouldn't know the difference between abstract and a virtual function that's defined externally (if that's even possible in Carbon? I assume so) - could have two properties, then, virtual and abstract (& perhaps all abstract things are also marked virtual))?
> > I guess probably not - but thought it was worth asking/understanding better what semantic representation SemIR is going for.
> 
> This is essentially what I'd expect, with the `enum` comment. Note though, I would generally encourage a little more implementation work to see how changes interact -- this is just a straightforward change to me, and some things (e.g. printing of "virtual" in formatter) seem like they might mirror a need to do similar "virtual"/"impl"/"abstract" stringification in diagnostics, which might yield a slightly different approach.

Sorry, I'm not quite following here - you mean: Good to go with what we have now, not try to overdesign it without experience. Or you'd encourage doing more implementation work/prototyping before a patch like this goes in, so we're heading in a more informed direction?

> Regarding mapping all to "virtual", each of the keywords implies different semantics, so they do need to be tracked separately.
> 
> > could have two properties, then, virtual and abstract (& perhaps all abstract things are also marked virtual))?
> 
> For example, it is invalid to declare a final class that inherits from a class without providing `impl` for all the `abstract` functions. That is not true of `virtual` functions either way: `virtual` provides a root of a virtual call, and shadowed `virtual` functions are just that (shadowed, not `impl`s).

Sorry, I should've used more words: I figured those invariants would be checked in `check`, and then maybe SemIR would implement the resulting semantics without the need to maintain enough information to diagnose those things - just enough to implement the requested semantics.

But I guess it's probably convenient to have all 3 markers for lowering, if not technically necessary (C++ being a proof of concept that it's not /necessary/ and can be derived from a rougher description). Being able to tell which functions introduce new vtable slots (virtual), those without implementations (so vtable slot can be initialized with null/skip the relocation/symbol reference), and overrides.

toolchain/sem_ir/function.h
```diff
@@ -23,6 +23,12 @@ struct FunctionFields {
   // always present if the function has a declared return type.
   InstId return_storage_id;
 
+  enum class VirtualModifier { None, Virtual, Abstract, Impl };
```
jonmeow: Sorry, can you move this to the top (for [declaration order styles](https://google.github.io/styleguide/cppguide.html#Declaration_Order), types/enums come before  instance members)

toolchain/check/handle_function.cpp
```diff
@@ -194,9 +194,17 @@ static auto BuildFunctionDecl(Context& context,
...
+
+  if (introducer.modifier_set.HasAnyOf(KeywordModifierSet::Virtual)) {
+    virtual_modifier = SemIR::FunctionFields::VirtualModifier::Virtual;
+  }
+  if (introducer.modifier_set.HasAnyOf(KeywordModifierSet::Abstract)) {
+    virtual_modifier = SemIR::FunctionFields::VirtualModifier::Abstract;
+  }
+  if (introducer.modifier_set.HasAnyOf(KeywordModifierSet::Impl)) {
+    virtual_modifier = SemIR::FunctionFields::VirtualModifier::Impl;
   }
```
jonmeow, line 208: Note, not suggesting a change here, but thinking about this...

Compare what you're doing here with the `class` code:
https://github.com/carbon-language/carbon-lang/blob/trunk/toolchain/check/handle_class.cpp#L203-L208

Maybe we should have a helper or something on `KeywordModifierSet` to help switch modifiers to an appropriate enum? It feels like something that we could maybe do with a variadic template...

jonmeow (approved):
Basically LG here, but one small style request (and one mulling comment, just mentioning the thought here)

Feel free to merge after adjusting the struct order

toolchain/sem_ir/function.h
dwblaikie: Ah, sure - done in https://github.com/carbon-language/carbon-lang/pull/4272/commits/a2d80a6bffeee49e89fdcccfdee812c012b505e2

toolchain/check/handle_function.cpp
dwblaikie: Oh, yeah - I can look into that.

dwblaikie: Sent https://github.com/carbon-language/carbon-lang/pull/4290

Aside: The `handle_class.cpp` case uses a non-scoped enum (`enum`, not `enum class`) and references the entities via the derived type (`Class::Enumerator` rather than `ClassFields::Enumerator`) - I guess the latter's clearly an improvement over the code I've written here for the function case, but should we standardize on the enum or enum class here? In this particular case an unscoped enum seems OK?

PR #4275, opened 2024-09-04
Parse support for `where` operator

josh11b:
Includes support for the `impls`, `=`, and `==` requirement operators to the right of a `where`, but `and` to allow multiple requirements is still a TODO.


toolchain/parse/handle_requirement.cpp
```diff
@@ -0,0 +1,84 @@
...
+      break;
+    }
+    case Lex::TokenKind::EqualEqual: {
+      break;
+    }
+    default: {
+      if (!state.has_error) {
+        CARBON_DIAGNOSTIC(
+            ExpectedRequirementOperator, Error,
+            "requirement should use `impls`, `=`, or `==` operator.");
```
zygoloid:
```suggestion
            "Requirement should use `impls`, `=`, or `==` operator.");
```

toolchain/parse/handle_requirement.cpp
```diff
@@ -0,0 +1,84 @@
...
+      // HandleRequirementOperator on one of the three requirement operator
+      // tokens.
+      CARBON_FATAL() << "Unexpected token kind for requirement operator: "
+                     << token_kind;
+      return;
+  }
+}
+
+auto HandleWhereFinish(Context& context) -> void {
+  auto state = context.PopState();
```
zygoloid, line 76: Maybe add a `TODO` here to handle an `and`?

toolchain/parse/handle_requirement.h
```diff
@@ -0,0 +1,18 @@
...
+#ifndef CARBON_TOOLCHAIN_PARSE_HANDLE_REQUIREMENT_H_
+#define CARBON_TOOLCHAIN_PARSE_HANDLE_REQUIREMENT_H_
+
+#include "toolchain/parse/context.h"
+
+namespace Carbon::Parse {
+
+// Push state at the beginning of a requirement expression after a `where` or
+// `require` token.
+auto BeginRequirement(Context& context) -> void;
```
zygoloid: Our convention is that the only entry points into the `handle_*.cpp` files are the `Handle*` functions, with any logic beyond that being split out into separate files. I think in other similar cases, we have added a state instead of a function call, and doing so here would be more consistent.

toolchain/parse/precedence.cpp
```diff
@@ -29,6 +29,8 @@ enum PrecedenceLevel : int8_t {
   // Type formation.
   TypePrefix,
   TypePostfix,
+  // `where` keyword
```
zygoloid:
```suggestion
  // `where` keyword.
```

toolchain/parse/typed_nodes.h
```diff
@@ -1032,6 +1034,51 @@ struct DesignatorExpr {
...
+
+struct RequirementImpls {
+  static constexpr auto Kind = NodeKind::RequirementImpls.Define(
+      {.category = NodeCategory::Requirement, .child_count = 2});
+  AnyExprId lhs;
+  Lex::ImplsTokenIndex token;
+  AnyExprId rhs;
+};
+
+// virtual node
```
zygoloid: Seems like a good idea to use the same commenting style here and in `ShortCircuit{And,Or}Operand` when mentioning that this is a virtual node.

toolchain/parse/typed_nodes.h
```diff
@@ -1032,6 +1034,51 @@ struct DesignatorExpr {
   NodeIdOneOf<IdentifierName, SelfTypeName> name;
 };
 
+// FIXME: RequirementRewrite?
+struct RequirementAssign {
+  static constexpr auto Kind = NodeKind::RequirementAssign.Define(
+      {.category = NodeCategory::Requirement, .child_count = 2});
+  AnyExprId lhs;
```
zygoloid, line 1041: Should we syntactically require the LHS to be a `DesignatorExpr`? (I'm fine for that to be future work rather than something done in this PR.)

toolchain/parse/typed_nodes.h
```diff
@@ -1032,6 +1034,51 @@ struct DesignatorExpr {
...
+struct RequirementImpls {
+  static constexpr auto Kind = NodeKind::RequirementImpls.Define(
+      {.category = NodeCategory::Requirement, .child_count = 2});
+  AnyExprId lhs;
+  Lex::ImplsTokenIndex token;
+  AnyExprId rhs;
+};
+
+// virtual node
+// FIXME: WhereType? WhereOperand?
```
zygoloid: `WhereOperand` would be consistent with what we do for `and` / `or`.

toolchain/parse/typed_nodes.h
```diff
@@ -1032,6 +1034,51 @@ struct DesignatorExpr {
   NodeIdOneOf<IdentifierName, SelfTypeName> name;
 };
 
+// FIXME: RequirementRewrite?
```
zygoloid: I note this is a FIXME not a TODO. Do you want the naming question here addressed before you merge this?

I lean towards using syntactic rather than semantic names for parse tree nodes, so maybe`RequirementEqual` and `RequirementEqualEqual`, matching the `TokenKind` names for the operators.

zygoloid (commented):
Looks good.

toolchain/parse/handle_requirement.cpp
josh11b: Since I can see the future https://github.com/carbon-language/carbon-lang/compare/trunk...josh11b:carbon-lang:and?expand=1 , I'm putting the `// TODO` a few lines higher, since that is the place that will get changed.

toolchain/parse/handle_requirement.h
josh11b: That seems very heavy for the 2 lines that function performs. Would it be okay to just inline it? Or include the body in the .h file? It will be called from another place in handle_requirement.cpp once `and` support is added.

toolchain/parse/typed_nodes.h
josh11b: I did want to tackle the naming as part of this PR. I've switched to those syntactic names.

toolchain/parse/typed_nodes.h
josh11b: I was thinking that should be done in check. I think the alternative would require a lot of special handling in parse. I assume the decision comes down to error handling, which I assumed would be easier in check.

toolchain/parse/typed_nodes.h
josh11b: I've made that change, but I didn't really understand those comments. They talk about how the `and`/`or` token is owned by one of the nodes, but nothing seems to distinguish which node is virtual and which owns the token.

toolchain/parse/typed_nodes.h
josh11b: Done.

toolchain/parse/handle_requirement.h
josh11b: Created a state after discussing offline.

toolchain/parse/typed_nodes.h
zygoloid: Yeah, I think the only thing that distinguishes which node "owns" the token is these comments :) And it seems to be an arbitrary choice that doesn't really impact anything -- I think we could equally say that the nodes share the token. But I think if there were a difference, it'd be: if a tool wants to map from a token to a parse node, the "non-virtual" one is the one they should find, which I think means the non-virtual node should always be a parent of the virtual node, so that they can both be discovered by tooling. Maybe that should be the rule we use: virtual parse nodes for a token should always be children of the non-virtual parse node.

toolchain/parse/typed_nodes.h
zygoloid: If this is a syntactic restriction (that is, the syntax must be exactly `where .Foo = <expr>`, then I think that belongs in parse. If it's a semantic requirement, and we would allow (for example) `where (.Self.(SelfType.Foo)) = <expr>`, or any other expression that symbolically evaluates to the name of an associated constant in `.Self`, then I think that belongs in check. (I would expect that the only way for check to reliably distinguish these two examples would be by looking at the parse tree nodes after the fact, and that's something that I think check should ideally not be doing.)

As it stands, I think the rule is syntactic: [the design](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/generics/details.md#rewrite-constraints) says "In a rewrite constraint, the left-hand operand of `=` must be a `.` followed by the name of an associated constant." I'm not tied to that, and I think it'd be fine to switch to a semantic rule if we want.

josh11b: Added TODO

zygoloid (approved):

josh11b:
Improved the tests, and now fixing up after merging.

PR #4276, opened 2024-09-04
Improve diagnostics for the case where some or all of the prelude is missing.

zygoloid:


zygoloid:
Factored out of #4209.

toolchain/check/testdata/packages/no_prelude/missing_prelude.carbon
```diff
@@ -0,0 +1,253 @@
...
+
+// --- prelude_as_namespace.carbon
+
+// TODO: Decide whether we want to accept this.
+namespace Core;
+
+fn Core.Int32() -> {} { return {}; }
+
+var n: {} = i32;
+
```
jonmeow, line 73:
```suggestion

// --- fail_prelude_as_class.carbon

// TODO: Decide whether we want to accept this.
class Core {
  fn Int32() -> {} { return {}; }
}

var n: {} = i32;

```

toolchain/check/testdata/packages/no_prelude/missing_prelude.carbon
```diff
@@ -0,0 +1,253 @@
...
+// --- prelude_use_in_prelude.carbon
+
+package Core library "prelude_use_in_prelude";
+
+// Core is not an imported package here.
+
+fn Int32() -> () { return (); }
+
+var n: () = i32;
```
jonmeow, line 61: Other tests use `{}`, so suggesting for consistency.

```suggestion
fn Int32() -> {} { return {}; }

var n: {} = i32;
```

jonmeow (approved):
Thanks!

toolchain/check/testdata/packages/no_prelude/missing_prelude.carbon
zygoloid: I was using a different type here to ensure we were picking the definition from this file, not from the other library.

PR #4277, opened 2024-09-05
Fix use-after-free bug in call argument conversion.

zygoloid:
Conversion can trigger new entities to be imported, which can invalidate the reference it holds to an EntityWithParamBase. Instead of holding such a reference, pull the information we need out of the entity early and only pass that into ConvertCallArgs.

I couldn't find a good standalone way to test this, but this fixes the test failure we otherwise see on MacOS after #4209, so it will be tested once that PR lands.


jonmeow (approved):
Nice catch!

PR #4278, opened 2024-09-05
Make the 'library' lines in tests use a substitution.

jonmeow:
This opens the door for replacing all `library ...` lines in toolchain test files with `library "[[@TEST_NAME]]";`. That's technically more typing in a lot of cases, but OTOH means we can just do some copy-paste boilerplate and stop carefully writing library names.

Also cleans up the test setup, because it's getting messy. I'm trying to make it easier to see the divisions of tests and the output associated with them. StringSwitch looked like a way to do this, with a few edits to make it work nicely.

### DIFF
--- a/testing/file_test/file_test_base.cpp
+++ b/testing/file_test/file_test_base.cpp
@@ -443,7 +443,7 @@ struct SplitState {
   auto has_splits() const -> bool { return file_index > 0; }
 
   auto add_content(llvm::StringRef line) -> void {
-    content.append(line);
+    content.append(line.str());
     content.append("\n");
   }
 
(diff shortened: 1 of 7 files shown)


testing/file_test/file_test_base.cpp
```diff
@@ -462,13 +462,62 @@ struct SplitState {
...
+  llvm::StringRef test_name = filename;
+  if (auto last_slash = test_name.rfind("/");
+      last_slash != llvm::StringRef::npos) {
+    test_name = test_name.substr(last_slash + 1);
+  }
+  if (auto ext_dot = test_name.find("."); ext_dot != llvm::StringRef::npos) {
+    test_name = test_name.substr(0, ext_dot);
+  }
+  while (test_name.consume_front("fail_") || test_name.consume_front("todo_")) {
+  }
```
chandlerc: Bit surprised we want to support repeated `fail_` prefixes? Maybe just:

```suggestion
  test_name.consume_front("todo_");
  test_name.consume_front("fail_");
```

Or even:
```suggestion
  test_name.consume_front("todo_");
  test_name.consume_front("fail_");
  test_name.consume_front("todo_");
```

testing/file_test/file_test_base.cpp
```diff
@@ -462,13 +462,62 @@ struct SplitState {
...
+
+  while (keyword_pos != std::string::npos) {
+    static constexpr llvm::StringLiteral TestName = "[[@TEST_NAME]]";
+    auto keyword = llvm::StringRef(*content).substr(keyword_pos);
+    if (keyword.starts_with(TestName)) {
+      keyword = TestName;
```
chandlerc: Curious why this is assigned here, it doesn't seem to be used?

chandlerc (commented):
Nice, some fairly minor comments inline.

@zygoloid - also curious if you think this will be a help with boiler plate by making it (much) more regular?

testing/file_test/file_test_base.cpp
```diff
@@ -462,13 +462,62 @@ struct SplitState {
   int file_index = 0;
 };
 
+// Replaces the content keywords.
+//
+// TEST_NAME is the only content keyword at present, but we do validate that
+// other names
```
zygoloid: Missing end of sentence?

testing/file_test/file_test_base.cpp
zygoloid: IIUC we only recognize `fail_` if it's at the start, so
```suggestion
  test_name.consume_front("fail_");
  test_name.consume_front("todo_");
```

zygoloid (commented):
> @zygoloid - also curious if you think this will be a help with boiler plate by making it (much) more regular?

I think it'll make copy-pasting files within a split-file test easier. I think this would be a win (but a small one).

testing/file_test/file_test_base.cpp
jonmeow: We have both fail_todo_ and todo_fail_. So I've dropped the loop on request, but we should still handle both variations.

jonmeow: And to be clear, these mean different things:

TODO, currently passes, should fail
TODO, currently fails, should succeed

testing/file_test/file_test_base.cpp
jonmeow: Cleaned up

testing/file_test/file_test_base.cpp
jonmeow: Done

zygoloid: Has this change been pushed yet? (I don't see the updated file here.)

testing/file_test/file_test_base.cpp
zygoloid: New approach works for me.

testing/file_test/file_test_base.cpp
jonmeow: re-pushed

zygoloid (approved):

PR #4283, opened 2024-09-08
Switch `CARBON_VLOG` to support a format string API.

chandlerc:
The goal is to replace our stream operator APIs with format string APIs that can be made to have much less impact on inlining and other optimizations of the performance critical path through the code.

Several experiments show that the most compact representation we can arrange for is one that calls an uninlined function and passes a minimal number of arguments to it. It doesn't help to do any work to minimize the arguments such as building a lambda -- the cost of extra code to merge the arguments is likely to outweigh the benefit.

Initial experiments showed that switching a hot but uninlined function to this new API enabled inlining and the subsequent performance improvement.

This also adds a 'TemplateString` utility that allows using a string literal as a template parameter. This is useful to remove the format string itself from the arguments passed to the function by passing it as a template argument instead.

Currently, support is left in place for both APIs because with `CARBON_VLOG` we can detect whether or not any message was provided expecting a format string. This should allow incrementally migrating code to this API. I've added some test coverage in this PR, but I'll separate out any switching of parts of the codebase over.

The goal is to eventually replace all the usages and remove the streaming support entirely.

This PR doesn't update `CARBON_CHECK` in the same way because it is substantially more complex to switch. I have a few experimental PRs looking at that and will discuss how best to approach this with the specific challenges check presents separately. But the goal is for all of the macro-based output APIs to move to format strings rather than streams.

PR #4285, opened 2024-09-09
Switch `CARBON_CHECK` to a format string API

chandlerc:
This switches `DCHECK` and `FATAL` as well.

The goal is to reduce the code size impact of these assertions so that
we can keep more of them enabled. Currently, the largest cost I see from
`CHECK` is not the actual check or the cold code itself, but actually
the failure to inline trivial functions due to the presence of the cold
code. This means that our goal isn't to reduce apparent code size in the
final binary but the LLVM IR cost assessed for these routines in the
inliner, which closely correlates with code size but is a bit different.

As discussed in #4283, experimentation shows that a single function call
with a minimal number of arguments is the lowest cost model for these.
This is easily achieved with a format-string API that internally uses
`llvm::formatv`. This PR is essentially the `CHECK` version of #4283.

However, the check macros are substantially harder to make work with
both format strings and streaming because they also take a condition.
Also, unexpectedly, I was very successful at devising a regular
expression based automated rewrite from the streaming to the format
string form with only low 10s of manual fixes. This includes compacting
strings broken up across lines, etc. Given how well that went, I've
prepared this PR which just directly switches to the format string API
and migrate everything to use it.

One nice side-effect is that the format string approach ends up greatly
simplifying the implementation here as well.

This is ... *shockingly* effective. Parsing speeds up by more than 3%
with just this change. And checking speeds up by **8%** with this change
alone:
```
BM_CompileAPIFileDenseDecls<Phase::Parse>/256      86.3µs ± 1%  82.9µs ± 1%  -3.94%  (p=0.000 n=17+19)
BM_CompileAPIFileDenseDecls<Phase::Parse>/1024      431µs ± 1%   415µs ± 1%  -3.76%  (p=0.000 n=18+19)
BM_CompileAPIFileDenseDecls<Phase::Parse>/4096     1.77ms ± 1%  1.71ms ± 1%  -3.18%  (p=0.000 n=18+19)
BM_CompileAPIFileDenseDecls<Phase::Parse>/16384    7.44ms ± 1%  7.17ms ± 2%  -3.56%  (p=0.000 n=18+20)
BM_CompileAPIFileDenseDecls<Phase::Parse>/65536    30.7ms ± 1%  29.7ms ± 1%  -3.15%  (p=0.000 n=18+20)
BM_CompileAPIFileDenseDecls<Phase::Parse>/262144    131ms ± 1%   127ms ± 1%  -2.81%  (p=0.000 n=18+18)
BM_CompileAPIFileDenseDecls<Phase::Check>/256       878µs ± 2%   800µs ± 1%  -8.91%  (p=0.000 n=19+20)
BM_CompileAPIFileDenseDecls<Phase::Check>/1024     1.88ms ± 2%  1.72ms ± 1%  -8.56%  (p=0.000 n=19+20)
BM_CompileAPIFileDenseDecls<Phase::Check>/4096     5.78ms ± 2%  5.28ms ± 1%  -8.70%  (p=0.000 n=20+18)
BM_CompileAPIFileDenseDecls<Phase::Check>/16384    21.9ms ± 1%  20.1ms ± 1%  -8.02%  (p=0.000 n=18+20)
BM_CompileAPIFileDenseDecls<Phase::Check>/65536    90.4ms ± 2%  83.1ms ± 1%  -8.04%  (p=0.000 n=19+20)
BM_CompileAPIFileDenseDecls<Phase::Check>/262144    381ms ± 2%   352ms ± 1%  -7.79%  (p=0.000 n=19+19)
```


chandlerc:
Also, to note, I generated almost all of the edits here with regular expressions -- I'm happy to re-do them, so if there are changes that folks would like to suggest, that shouldn't really be a problem.

I used the `sd` tool: https://github.com/chmln/sd

And here are some of the relevant regexs:
```console
# Tries to find `CARBON_D?CHECK` statements with just a string
# message. This proved useful to do prior to any with arguments
# below, even though it is in some ways a generalization of that
# regular expression. See that comment for the limitations.
> sd -f ms 'CARBON_(D?CHECK)\(((?:[^()]|\((?:[^()]|\([^()]*\))*\))+)\)\s*<<\s+"((?:[^"]+|"\s*(?:<< )?")+)";' 'CARBON_$1($2, "$3");' (rg -l 'CARBON_D?CHECK')

# Tries to find `CARBON_D?CHECK` statement with one non-string
# parameter and optional prefix and suffix string parameters and
# rewrite them into format string form accepting one argument.
# Only handles a limited degree of parenthesis and complexity in
# the condition and a limited degree of template argument nesting
# in parameters. The tail can be copy/pasted to create 2-argument,
# 3-argument, and more forms. I found never found a 5-argument.
> sd -f ms 'CARBON_(D?CHECK)\(((?:[^()]|\((?:[^()]|\([^()]*\))*\))+)\)(?:\s*<< "((?:[^"]+|"\s*(?:<< )?")+)")?\s*<< ((?:[^"<;]*(?:[^"<;&&[:^space:]]|<[^<>]*>))+)(?:\s*<< "((?:[^"]+|"\s*(?:<< )?")+)")?;' 'CARBON_$1($2, "$3{0}$5", $4);' (rg -l 'CARBON_D?CHECK')

# Finds string literals continued across lines, both with streams
# in between and through string literal concatenation, and merges
# one such line break away. Run repeatedly to replace long runs.
> sd -f ms '(CARBON_D?CHECK\([^;"]+"[^"]*)"\n\s*(?:<< )?"' '$1' (rg -l 'CARBON_D?CHECK')
```

I also used the relevant variations of these for `CARBON_FATAL`, etc.

There were several manual fixes required, but quite manageable. 10 maybe across the whole codebase.

common/check.h
```diff
@@ -14,30 +14,35 @@ namespace Carbon {
...
-#define CARBON_CHECK(...)                                                   \
-  (__VA_ARGS__) ? (void)0                                                   \
-                : CARBON_CHECK_INTERNAL_STREAM()                            \
-                      << "CHECK failure at " << __FILE__ << ":" << __LINE__ \
-                      << ": " #__VA_ARGS__                                  \
-                      << Carbon::Internal::ExitingStream::AddSeparator()
+//   CARBON_CHECK(is_valid, "Data is not valid!");
+#define CARBON_CHECK(condition, ...) \
+  (condition) ? (void)0              \
+              : CARBON_INTERNAL_CHECK(condition __VA_OPT__(, ) __VA_ARGS__)
```
zygoloid: This no longer supports commas in `condition`, which might be worth mentioning in the comment here. It looks (from a brief non-exhaustive skip) like this wasn't actually a problem for any of our existing checks, though?
```suggestion
//   CARBON_CHECK(is_valid, "Data is not valid!");
//
// The condition must be parenthesized if it contains top-level commas, for
// example in a template argument list:
//   CARBON_CHECK((inst.IsOneOf<Call, TupleLiteral>()),
//                "Unexpected inst {0}", inst);
#define CARBON_CHECK(condition, ...) \
  (condition) ? (void)0              \
              : CARBON_INTERNAL_CHECK(condition __VA_OPT__(, ) __VA_ARGS__)
```

zygoloid (commented):
I've not reviewed the migration commit, but the macro changes commit LGTM.

common/check.h
chandlerc: There were a few in `//explorer` that I had to add parentheses for, but generally was very rare. Added comment.

chandlerc:
I think this is ready for review -- or to discuss how to break up the review of the migration if its too much?

common/check.h
```diff
@@ -14,30 +14,40 @@ namespace Carbon {
...
+      ? (void)0                       \
+      : CARBON_INTERNAL_DEAD_DCHECK(condition __VA_OPT__(, ) __VA_ARGS__)
 #endif
 
-// This is similar to CHECK, but is unconditional. Writing CARBON_FATAL() is
-// clearer than CARBON_CHECK(false) because it avoids confusion about control
-// flow.
+// This is similar to CHECK, but is unconditional. Writing
+// `CARBON_FATAL("message")` is clearer than `CARBON_CHECK(false, "message")
+// because it avoids confusion about control flow.
```
josh11b, line 46: This seems to be mixing "clarity" (human understandability and nice error message) with compiler optimization ("confusion about control flow")?

jonmeow:
Build failure:

```
toolchain/parse/handle_requirement.cpp:68:22: error: invalid operands to binary expression ('void' and 'const char[49]')
      CARBON_FATAL() << "Unexpected token kind for requirement operator: "
      ~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

chandlerc:
> Build failure:
> 
> ```
> toolchain/parse/handle_requirement.cpp:68:22: error: invalid operands to binary expression ('void' and 'const char[49]')
>       CARBON_FATAL() << "Unexpected token kind for requirement operator: "
>       ~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```

Yeah, some new code is all. Updated and fixed.

common/check_internal.cpp
```diff
@@ -14,22 +14,20 @@ static auto PrintAfterStackTrace(void* str) -> void {
...
-  // Register another signal handler to print the buffered message. This is
-  // because we want it at the bottom of output, after LLVM's builtin stack
-  // output, rather than the top.
+auto CheckFailImpl(const char* kind, const char* file, int line,
+                   const char* condition_str, llvm::StringRef extra_message)
+    -> void {
+  // Render the final check string here.
+  std::string message = llvm::formatv(
+      "{0} failure at {1}:{2}{3}{4}{5}{6}\n", kind, file, line,
+      llvm::StringRef(condition_str).empty() ? "" : ": ", condition_str,
```
jonmeow, line 23: Noting the StringRef cast, why doesn't this just use StringRef parameters? I assume you have a reason, but looking at template_string.h and check_internal.h, I don't see a comment explaining the preference for `const char*` over `llvm::StringRef` for parameters. Can you add something?

common/check_internal.h
```diff
@@ -5,73 +5,128 @@
...
+}
+#else
+// Use a template parameter for the format string in optimized builds. This will
+// result in essentially every check generating a separate function, but makes
+// the impact on the caller slightly smaller.
+template <TemplateString Kind, TemplateString File, int Line,
+          TemplateString ConditionStr, TemplateString FormatStr, typename... Ts>
+[[noreturn, gnu::cold, clang::noinline, clang::preserve_most]] auto CheckFail(
+    Ts&&... values) -> void {
+  if constexpr (llvm::StringRef(FormatStr).empty() == 0) {
```
josh11b: This doesn't match the other case and seems suspicious:
```suggestion
  if constexpr (sizeof...(values) == 0) {
```
but it does seem like there is a special case for no args below that creates an empty format string in that case (but I think that is just a misleading comment?), so maybe this should be?
```suggestion
  if constexpr (llvm::StringRef(FormatStr).empty()) {
```

It isn't clear to me we are getting the right message when there are check failures in these two cases.

Would it be worth using runtime arguments in the case where there are no format arguments? Your comment above doesn't explain why template arguments are used in optimized builds despite introducing many instantiations.

common/check_internal.h
```diff
@@ -5,73 +5,128 @@
...
+    CheckFailImpl(Kind.c_str(), File.c_str(), Line, ConditionStr.c_str(),
+                  FormatStr.c_str());
```
josh11b: Known to be empty in this case?

toolchain/check/handle_struct.cpp
```diff
@@ -125,8 +125,8 @@ auto HandleParseNode(Context& context, Parse::StructTypeLiteralId node_id)
   context.node_stack()
       .PopAndDiscardSoloNodeId<Parse::NodeKind::StructTypeLiteralStart>();
 
-  CARBON_CHECK(refs_id != SemIR::InstBlockId::Empty)
-      << "{} is handled by StructLiteral.";
+  CARBON_CHECK(refs_id != SemIR::InstBlockId::Empty,
+               "{} is handled by StructLiteral.");
```
jonmeow: Should this be escaped for formatv?

jonmeow (approved):
Approving. zygoloid said he'd looked at the macros, I glanced but I've focused on the migration. I saw one issue, but generally looks clean, so I think it makes sense to approve and let you merge if you're ready (particularly to avoid more skew).

I'm realizing though, a couple risks with formatv:

- There's a spot where we pass in the full sem_ir. Probably not a big memory risk, but generating the intermediate string is at least a little risky.
- The stringent formatv checking has me concerned on a different end... Because it enforces arguments, but doesn't check at compile time, and we can't really unit test CHECK-failures, we risk having typos in these that are uncaught.
    - This mainly makes me wonder if we should build our own format function, particularly to validate parameters at compile time. It doesn't feel like it should be blocking, given the other concerns being addressed here.

common/check_internal.h
```diff
@@ -5,73 +5,128 @@
...
+// Implements the fatal macro.
+//
+// Forms a message prefix for the failure that includes its source location,
+// then adds any format string and formatted arguments.
+#define CARBON_INTERNAL_FATAL(...)                 \
+  CARBON_INTERNAL_CHECK_IMPL##__VA_OPT__(_FORMAT)( \
+      "FATAL", __FILE__, __LINE__, "" __VA_OPT__(, ) __VA_ARGS__)
+
+#ifdef NDEBUG
+// For optimized builds we have a dead check that we want to potentially "use"
```
josh11b: Clarify that this is dead because it is for DCHECK, which is only checking in debug builds.

common/check_internal.h
```diff
@@ -5,73 +5,128 @@
...
+// arguments, but otherwise have the minimal overhead. We avoid forming
+// interesting format strings here so that we don't have to repeatedly
+// instantiate the `Check` function above. This format string would be an error
+// if actually used, but as this is dead that shouldn't matter.
```
josh11b:
```suggestion
// if actually used.
```

common/check_internal.h
```diff
@@ -5,73 +5,128 @@
...
+#ifndef NDEBUG
+// Pass the format string as a runtime parameter in debug builds.
+#define CARBON_INTERNAL_CHECK_IMPL_CALL(kind, file, line, condition_str, \
+                                        format_str, ...)                 \
+  Carbon::Internal::CheckFail(kind, file, line, condition_str,           \
+                              format_str __VA_OPT__(, ) __VA_ARGS__)
+#else
+// Pass the format string as a template parameter in optimized builds. Note that
+// the template arguments include commas and so require extra parentheses to use
+// within a macro.
```
josh11b: Personally I expect lots of extra parentheses when writing macros, and would just add them in all cases, even the debug case, without any specific comment.

common/check_internal.h
```diff
@@ -5,73 +5,128 @@
...
+                                          format_str, ...)                 \
+  CARBON_INTERNAL_CHECK_IMPL_CALL(kind, file, line, condition_str,         \
+                                  format_str __VA_OPT__(, ) __VA_ARGS__)
+
+// Implements the failure of a check.
+//
+// Forms a message prefix for the failure that includes its source location and
+// the failing condition, then adds any format string and formatted arguments.
+#define CARBON_INTERNAL_CHECK(condition, ...)      \
+  CARBON_INTERNAL_CHECK_IMPL##__VA_OPT__(_FORMAT)( \
```
josh11b, line 83: Heavy use of `__VA_OPT__` is making this code significantly harder to follow.

common/check_internal.h
```diff
@@ -5,73 +5,128 @@
...
+// the template arguments include commas and so require extra parentheses to use
+// within a macro.
+#define CARBON_INTERNAL_CHECK_IMPL_CALL(kind, file, line, condition_str,     \
+                                        format_str, ...)                     \
+  (Carbon::Internal::CheckFail<kind, file, line, condition_str, format_str>( \
+      __VA_ARGS__))
+#endif
+
+// Calls `CheckFail` with the provided message plus a newline and with no
+// parameters. This implements check messages without any formatted values.
```
josh11b: This isn't about whether there are parameters/formatted values, but rather no "provided message"/format string at all?

common/check_test.cpp
```diff
@@ -26,32 +26,32 @@ TEST(CheckTest, CheckTrueCallbackNotUsed) {
...
-  CARBON_CHECK(true) << callback();
+  CARBON_CHECK(true, "{0}", callback());
   EXPECT_FALSE(called);
 }
 
 TEST(CheckTest, CheckFalseMessage) {
-  ASSERT_DEATH({ CARBON_CHECK(false) << "msg"; },
+  ASSERT_DEATH({ CARBON_CHECK(false, "msg"); },
                "\nCHECK failure at common/check_test.cpp:.+: false: msg\n");
 }
```
josh11b, line 37:
```suggestion

TEST(CheckTest, CheckFalseFormattedMessage) {
  const char msg[] = "msg";
  std::string str = "str";
  int i = 1;
  ASSERT_DEATH({ CARBON_CHECK(false, "{0} {1} {2} {3}", msg, str, i, 0); },
               "\nCHECK failure at common/check_test.cpp:.+: false: msg str 1 0\n");
}

```

josh11b (commented):
Here is some initial feedback before I look at all the callsites.

chandlerc:
Note, I had a bad git push earlier and so a weird temporary version I was experimenting with got put back in place. Sorry for that. I think the version now is the one I intended and working on the last batches of comments.

common/check_internal.cpp
chandlerc: Added a comment -- the goal is to reduce codesize of the caller by using C strings.

toolchain/check/handle_struct.cpp
chandlerc: Weirdly, this works as is... because we avoid `formatv` when there are no arguments.

But your comment is worrisome, because that seems like an unfortunate implementation detail to leak in this way. I'm not really sure what the best approach is here, and interested in ideas.

jonmeow: Can it just pass to formatv irrespective?

common/check.h
chandlerc: Not sure -- this comment text was already present, not something I was adding. Happy to have suggestions for better wording though.

common/check_internal.h
chandlerc: The second one is definitely the right one here, but for surprising reasons. I've clarified the comment there.

There are comments about why we try to use compile time parameters here, and this code was simplified by getting back to the correct version. Sorry you saw the messy one.

common/check_internal.h
chandlerc: Sure, can pass an explicit empty string literal here.

common/check_internal.h
chandlerc: Some has been reduced, but yeah. I don't love the `__VA_OPT__` but I don't know how to remove any of the ones left.

toolchain/check/handle_struct.cpp
jonmeow: Note, if you're concerned about codesize of that, the version with no arguments could do just a thin non-templated wrapper of formatv. Just to force that the argument string gets the same treatment.

But it's probably still going to be missed, unless we get something with compile-time validation.

(either way, I don't view this as critical for this PR, but maybe worth a comment/todo)

common/check_internal.h
chandlerc: Done.

common/check_internal.cpp
jonmeow: Where should I be looking? I don't see anything in https://github.com/carbon-language/carbon-lang/pull/4285/commits/04dc8068872c3328642f77727ce02723d627cca2

common/check_internal.h
```diff
@@ -14,6 +14,11 @@ namespace Carbon::Internal {
 //
 // This is out-of-line and will arrange to stop the program, print any debugging
 // information and this string.
+//
+// This API uses `const char*` C string arguments rather than `llvm::StringRef`
+// because that lets the code size of calling it be smaller as it only needs to
```
jonmeow:
```suggestion
// because that lets the code size of calling it be smaller: it only needs to
```

Huh, so even though there's 5 arguments, it materializes all of that as a single pointer? Or do you mean more like "materialize each as a pointer argument" instead of "materialize a single pointer argument"?

common/check_internal.h
```diff
@@ -5,73 +5,105 @@
...
+// materialize a single pointer argument. The runtime cost of re-computing the
+// size should be minimal.
+[[noreturn]] auto CheckFailImpl(const char* kind, const char* file, int line,
+                                const char* condition_str,
+                                llvm::StringRef extra_message) -> void;
+
+// Prints a check failure, including rendering any user-provided message based
+// on its format string.
```
josh11b:
```suggestion
// Prints a check failure, including rendering any user-provided message using
// a format string.
```

common/check_internal.h
chandlerc: No, I mean that if we make these `llvm::StringRef`s, then it has to materialize both a pointer and a size far that argument intsead of just a pointer.

chandlerc: can add "for each" to the sentence?

chandlerc: Tried a better comment to really spell out what is different about the first arguments from the last. Sorry if that was the source of confusion and I didn't address it at first.

common/check_internal.h
chandlerc: I think all of this was fixed by getting the right version instead of the weird hacks I was playing with.

toolchain/check/handle_struct.cpp
chandlerc: So, it actually *is* passed to `formatv` always, I misremembered. So I think we're fine here.

common/check_internal.cpp
chandlerc: I think this is now in `check_internal.h` on line 18?

common/check_internal.h
jonmeow: Ah, so this is just that sizeof(const char*) < sizeof(StringRef), less argument data to pass, okay.

jonmeow: Am I awful to ask if `int line` should be `int32_t line`, or `const char* extra_message`?

chandlerc: The size of int won't matter here, which is why I just used the general `int`.

Could change extra message but it would actually make it a tiny bit more constraining to call. Seemed better to use the general type there as only one.

common/check_test.cpp
```diff
@@ -26,32 +26,40 @@ TEST(CheckTest, CheckTrueCallbackNotUsed) {
...
 
+TEST(CheckTest, CheckFalseFormattedMessage) {
+  const char msg[] = "msg";
+  std::string str = "str";
+  int i = 1;
+  ASSERT_DEATH({ CARBON_CHECK(false, "{0} {1} {2} {3}", msg, str, i, 0); },
+               "\nCHECK failure at common/check_test.cpp:.+: false: msg str 1 0\n");
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  ASSERT_DEATH(
      { CARBON_CHECK(false, "{0} {1} {2} {3}", msg, str, i, 0); },
      "\nCHECK failure at common/check_test.cpp:.+: false: msg str 1 0\n");
```

josh11b (approved):

common/check_internal.h
josh11b: Missed?

chandlerc: Ah, sorry, forgot in trying to restore the good version to git. Fixed now.

common/check_internal.h
chandlerc: Going to merge while this is clean, but happy to follow up on this if useful.

PR #4287, opened 2024-09-09
Manually free up disk space on macOS

chandlerc:
This should mitigate the effects of a GitHub regression that reduced space on these runners:
https://github.com/actions/runner-images/issues/10511

After this, we're mostly fine, but builds that happen to compile enough of the codebase can bump past it. With this PR we have over 50 GiB of space which is more than we need.

See a test run here: https://github.com/carbon-language/carbon-lang/actions/runs/10780743574

### DIFF
--- a/.github/actions/build-setup-macos/action.yml
+++ b/.github/actions/build-setup-macos/action.yml
@@ -18,9 +18,6 @@ runs:
         xcrun simctl delete all
         sudo rm -rf ~/Library/Developer/CoreSimulator/Caches/*
 
-        echo '*** Remove older Xcode versions'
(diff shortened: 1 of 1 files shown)


zygoloid (commented):
LGTM

josh11b (approved):

chandlerc:
First run worked, and freed up 80GiB, but took 5 minutes just deleting stuff. I'm trying a more minimal version now.

chandlerc:
Ok, now just removing the iOS simulator which is plenty of space and super fast compared to removing versions of Xcode.

zygoloid (approved):

yasali (approved):

PR #4290, opened 2024-09-10
Add KeywordModifierSet helper for conversion to (likely SemIR) enums

dwblaikie:
(based on https://github.com/carbon-language/carbon-lang/pull/4272#discussion_r1751001345)

Could haggle over the name "ToEnum" probably avoids the debate over "enumeration" (the type being returned) v "enumerator" (the value being returned)

### DIFF
--- a/toolchain/check/handle_class.cpp
+++ b/toolchain/check/handle_class.cpp
@@ -200,12 +200,10 @@ static auto BuildClassDecl(Context& context, Parse::AnyClassDeclId node_id,
   if (introducer.extern_library.is_valid()) {
     context.TODO(node_id, "extern library");
   }
-  auto inheritance_kind =
-      introducer.modifier_set.HasAnyOf(KeywordModifierSet::Abstract)
-          ? SemIR::Class::Abstract
-      : introducer.modifier_set.HasAnyOf(KeywordModifierSet::Base)
-          ? SemIR::Class::Base
-          : SemIR::Class::Final;
+  SemIR::ClassFields::InheritanceKind inheritance_kind =
+      introducer.modifier_set.ToEnumerator(SemIR::Class::Final)
+          .Case(KeywordModifierSet::Abstract, SemIR::Class::Abstract)
+          .Case(KeywordModifierSet::Base, SemIR::Class::Base);
 
   auto decl_block_id = context.inst_block_stack().Pop();
 
--- a/toolchain/check/handle_class.cpp
+++ b/toolchain/check/handle_class.cpp
@@ -200,10 +200,11 @@ static auto BuildClassDecl(Context& context, Parse::AnyClassDeclId node_id,
   if (introducer.extern_library.is_valid()) {
     context.TODO(node_id, "extern library");
   }
-  SemIR::Class::InheritanceKind inheritance_kind =
-      introducer.modifier_set.ToEnum(SemIR::Class::Final)
+  auto inheritance_kind =
+      introducer.modifier_set.ToEnum<SemIR::Class::InheritanceKind>()
           .Case(KeywordModifierSet::Abstract, SemIR::Class::Abstract)
-          .Case(KeywordModifierSet::Base, SemIR::Class::Base);
+          .Case(KeywordModifierSet::Base, SemIR::Class::Base)
+          .Default(SemIR::Class::Final);
 
   auto decl_block_id = context.inst_block_stack().Pop();
 
--- a/toolchain/check/handle_class.cpp
+++ b/toolchain/check/handle_class.cpp
@@ -201,7 +201,7 @@ static auto BuildClassDecl(Context& context, Parse::AnyClassDeclId node_id,
     context.TODO(node_id, "extern library");
   }
   SemIR::Class::InheritanceKind inheritance_kind =
-      introducer.modifier_set.ToEnumerator(SemIR::Class::Final)
+      introducer.modifier_set.ToEnum(SemIR::Class::Final)
           .Case(KeywordModifierSet::Abstract, SemIR::Class::Abstract)
           .Case(KeywordModifierSet::Base, SemIR::Class::Base);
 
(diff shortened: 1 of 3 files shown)


toolchain/check/handle_class.cpp
```diff
@@ -200,12 +200,10 @@ static auto BuildClassDecl(Context& context, Parse::AnyClassDeclId node_id,
   if (introducer.extern_library.is_valid()) {
     context.TODO(node_id, "extern library");
   }
-  auto inheritance_kind =
-      introducer.modifier_set.HasAnyOf(KeywordModifierSet::Abstract)
-          ? SemIR::Class::Abstract
-      : introducer.modifier_set.HasAnyOf(KeywordModifierSet::Base)
-          ? SemIR::Class::Base
-          : SemIR::Class::Final;
+  SemIR::ClassFields::InheritanceKind inheritance_kind =
```
jonmeow:
```suggestion
  SemIR::Class::InheritanceKind inheritance_kind =
```

I think this would work, and feels a little more consistent with `SemIR::Class` below (which, I think the shorter bit helps)

Note, part of this will be that the `Class` versus `ClassFields` split is relatively new, but I generally don't think we should refer to the `FooFields` types directly (it's an abstraction working around C++, trying to hide C++ behaviors)

toolchain/check/handle_function.cpp
```diff
@@ -195,17 +195,14 @@ static auto BuildFunctionDecl(Context& context,
                     parent_scope_inst);
   bool is_extern = introducer.modifier_set.HasAnyOf(KeywordModifierSet::Extern);
   SemIR::FunctionFields::VirtualModifier virtual_modifier =
```
jonmeow: Note, comparing with the above, maybe SemIR::Function instead of SemIR::FunctionFields here too.

toolchain/check/keyword_modifier_set.h
```diff
@@ -70,10 +70,38 @@ class KeywordModifierSet {
...
   // Returns true if there's a non-empty set intersection.
-  constexpr auto HasAnyOf(KeywordModifierSet other) -> bool {
+  constexpr auto HasAnyOf(KeywordModifierSet other) const -> bool {
     return set_ & other.set_;
   }
 
+  // Return a builder that implicitly converts to the specified enumeration type
+  // mapping the cases passed to the `Case`s specified. For example:
+  //   ```
+  //   SomeEnum e = set.ToEnumerator(SomeEnum::Default)
```
jonmeow: I prefer Enum over Enumerator for brevity, but had you considered an API like StringSwitch:

`ToEnum<...>().Case(...).Case(...).Default(...)`

This is mainly to make it clearer what the default is. It'd mean making result_ optional and replacing `operator T()` with `auto Default(...) -> T`

I think this doesn't even increase verbosity much for call sites because:

```
SemIR::FunctionFields::VirtualModifier virtual_modifier = modifier_set.ToEnum(...)
```

becomes:

```
auto virtual_modifier = modifier_set.ToEnum<SemIR::FunctionFields::VirtualModifier>(...)
```

(if you're really not a fan of this, you might consider renaming `ToEnumerator` to `ToEnumWithDefault` to make it explicit how the argument is used)

toolchain/check/keyword_modifier_set.h
```diff
@@ -70,10 +70,38 @@ class KeywordModifierSet {
...
+     public:
+      Converter(const KeywordModifierSet& set, T default_value)
+          : set_(set), result_(default_value) {}
+      auto Case(RawEnumType raw_enumerator, T result) -> Converter& {
+        if (set_.HasAnyOf(raw_enumerator)) {
+          result_ = result;
+        }
+        return *this;
+      }
+      operator T() { return result_; }
```
jonmeow:
```suggestion
          : set_(set), result_(default_value) {}

      auto Case(RawEnumType raw_enumerator, T result) -> Converter& {
        if (set_.HasAnyOf(raw_enumerator)) {
          result_ = result;
        }
        return *this;
      }

      operator T() { return result_; }
```

Just whitespace to make it easier to see the start/end.

toolchain/check/keyword_modifier_set.h
```diff
@@ -70,10 +70,38 @@ class KeywordModifierSet {
...
 
+  // Return a builder that implicitly converts to the specified enumeration type
+  // mapping the cases passed to the `Case`s specified. For example:
+  //   ```
+  //   SomeEnum e = set.ToEnumerator(SomeEnum::Default)
+  //                    .Case(KeywordModifierSet::A, SomeEnum::A)
+  //                    .Case(KeywordModifierSet::B, SomeEnum::B);
+  //   ```
+  template <typename T>
+  auto ToEnumerator(T default_value) {
```
jonmeow:
```suggestion
  auto ToEnumerator(T default_value) -> auto {
```

I checked with zygoloid and he's okay with the nested scope, but suggested `-> auto` here to make it clear this is a deliberate return and not forgotten.

jonmeow:
I'm happy to have this, now we can bikeshed over API. 😉

toolchain/check/keyword_modifier_set.h
dwblaikie: Changed to the shorter `ToEnum` in ae262bdd31b9b16978f98461efe99d215a9ce0bd

I liked not having to explicitly specify the type, though, yeah - if it means using auto on the resulting variable the tradeoff is a bit of a wash. And not having to have an implicit conversion is nice - having Default be the terminating operation makes it clearer I think.

Switched to this in 1b6d5cdf84b44d63851aeccfc497aa637571dd00

toolchain/check/handle_function.cpp
dwblaikie: Done in a58d56003f56db16ce257712dbbd7335485a6e1d

toolchain/check/keyword_modifier_set.h
```diff
@@ -74,19 +76,19 @@ class KeywordModifierSet {
     return set_ & other.set_;
   }
 
-  // Return a builder that implicitly converts to the specified enumeration type
-  // mapping the cases passed to the `Case`s specified. For example:
+  // Return a builder that returns the new enumeration type once a series of
+  // mapping `Case`s and a final `Default` are provided:
```
jonmeow:
```suggestion
  // mapping `Case`s and a final `Default` are provided. For example:
```

FWIW I think your previous separate "For example" read better.

jonmeow (approved):
Looks great, thanks!

PR #4297, opened 2024-09-11
add additional testdata for access checks

brymer-meneses:
I just realized that https://github.com/carbon-language/carbon-lang/pull/4248 should now correctly enforce compound member access.  This change adds tests for this functionality.

### DIFF
--- a/toolchain/check/testdata/class/inheritance_access.carbon
+++ b/toolchain/check/testdata/class/inheritance_access.carbon
@@ -198,11 +198,12 @@ base class A {
 }
 
 class B {
+  extend base: A;
   fn F[self: Self]() {
     // CHECK:STDERR: fail_compound_member_access.carbon:[[@LINE+6]]:11: ERROR: Cannot access private member `x` of type `A`.
     // CHECK:STDERR:     self.(A.x);
     // CHECK:STDERR:           ^~~
-    // CHECK:STDERR: fail_compound_member_access.carbon:[[@LINE-8]]:15: The private member `x` is defined here.
+    // CHECK:STDERR: fail_compound_member_access.carbon:[[@LINE-9]]:15: The private member `x` is defined here.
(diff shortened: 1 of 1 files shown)


toolchain/check/testdata/class/inheritance_access.carbon
```diff
@@ -178,16 +178,53 @@ class B {
...
 
+// --- fail_compound_member_access.carbon
+
+library "[[@TEST_NAME]]";
+
+base class A {
+  private var x: i32;
+}
+
+class B {
```
geoffromer, line 200: Should this have `extend base: A;`?

Given that it doesn't, it's a little surprising that the error is about `x` having the wrong access control, rather than about `A` not being a base class of `B` or something like that, but we don't need to fix that right now.

brymer-meneses: > Should this have extend base: A;?

Yea, I think you're right.

toolchain/check/testdata/class/inheritance_access.carbon
```diff
@@ -178,16 +178,53 @@ class B {
...
+    // CHECK:STDERR: fail_compound_member_access.carbon:[[@LINE-8]]:15: The private member `x` is defined here.
+    // CHECK:STDERR:   private var x: i32;
+    // CHECK:STDERR:               ^~~~~~
+    self.(A.x);
+  }
+}
+
+// --- inherited_compound_member_access.carbon
+
+library "inherited_compound_member_access";
```
jonmeow:
```suggestion
library "[[@TEST_NAME]]";
```

jonmeow: Note, I was only looking for issues with library directives due to the recent change.

geoffromer (approved):

PR #4299, opened 2024-09-11
Break out driver environment info into its own type.

jonmeow:
I want to split commands out so that we don't keep piling onto driver (particularly as I'm eyeing clang-related commands). This extracts out the DriverEnv so that it can be easily shared, with the CompilationUnit as an example.

CARBON_VLOG_TO is to remove the vlog_stream_ requirement of CARBON_VLOG.

### DIFF
--- a/common/vlog.h
+++ b/common/vlog.h
@@ -26,9 +26,11 @@ template <TemplateString FormatStr, typename... Ts>
 
 }  // namespace Carbon::Internal
 
-// Logs when verbose logging is enabled (vlog_stream_ is non-null).
+// Logs when verbose logging is enabled. CARBON_VLOG_TO uses a provided stream;
+// CARBON_VLOG requires a member named `vlog_stream_`.
 //
 // For example:
+//   CARBON_VLOG_TO(vlog_stream, "Verbose message: {0}", "extra information");
 //   CARBON_VLOG("Verbose message: {0}", "extra information");
 //
 // The first argument must be a string literal format string valid for passing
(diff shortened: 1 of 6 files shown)


common/vlog.h
```diff
@@ -42,10 +44,13 @@ template <TemplateString FormatStr, typename... Ts>
 // However, the streaming syntax has higher overhead and can inhibit inlining.
 // Code should prefer the format string form, and eventually when all code has
 // migrated the streaming interface will be removed.
-#define CARBON_VLOG(FormatStr, ...)                                          \
-  __builtin_expect(vlog_stream_ == nullptr, true)                            \
-      ? (void)0                                                              \
-      : Carbon::Internal::VLogImpl<"" FormatStr>(vlog_stream_ __VA_OPT__(, ) \
+#define CARBON_VLOG_TO(Stream, FormatStr, ...)                         \
+  __builtin_expect(Stream == nullptr, true)                            \
```
josh11b, line 48:
```suggestion
  __builtin_expect((Stream) == nullptr, true)                          \
```

toolchain/driver/driver.cpp
```diff
@@ -789,7 +792,7 @@ class Driver::CompilationUnit {
            !input_filename_.starts_with(options_.exclude_dump_file_prefix);
   }
 
-  Driver* driver_;
+  DriverEnv& driver_env_;
```
josh11b: I would prefer fields not to have a reference type, particularly non-const. Can this be made into a pointer instead? I see that some other fields are references too -- what is the motivation?

jonmeow: Generally I think there's a mix in the toolchain, tending more to references than pointers. One constraint of references is that they're implicitly non-null.

jonmeow: Perhaps as a stronger argument, three lines down there's a &. This type should be self-consistent.

https://discord.com/channels/655572317891461132/821113559755784242/1283516297686286377 regarding this.

jonmeow: Trying to read a possible conclusion from current discussion, updated to `DriverEnv*`

josh11b (approved):

PR #4301, opened 2024-09-11
Convert CommandLine member references to pointers

jonmeow:
This follows up on a [style question](https://discord.com/channels/655572317891461132/821113559755784242/1283516297686286377) about whether to prefer reference members or pointers. This PR converts to pointers as a demonstration of that style choice.

Note, I'm trying to update constructors to match use of `*` based on whether they keep a reference. I'm removing a few `const&` uses where no reference was kept (i.e., it was just copied, and didn't seem worth a move).

I'm changing `AddArgImpl` to return an `Arg*` because it just gets passed to a constructor, seems simpler this way.

### DIFF
--- a/common/command_line.cpp
+++ b/common/command_line.cpp
@@ -83,6 +83,7 @@ Command::Command(CommandInfo info, Command* parent)
 
 class MetaPrinter {
  public:
+  // `out` must not be null.
   explicit MetaPrinter(llvm::raw_ostream* out) : out_(out) {}
 
   // Registers this meta printer with a command through the provided builder.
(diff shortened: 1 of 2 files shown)


josh11b (approved):

jonmeow:
Discussion seems to be leaning towards more passing by reference, so I've changed that here.

josh11b (approved):
Approving since this matches what was discussed in #style, even though I prefer the all pointer version.

chandlerc (commented):
Just wanted to say -- this also LGTM as is as an incremental step. e can fix forward with whatever subsequent style tweaks come up. Or happy if you'd rather update this PR based on the latest discussions.

jonmeow:
That's fine, given the prior commit already had pointers I think it's easier to unroll to that here. I've also add "must not be null" comments per your style request.

josh11b (approved):

PR #4302, opened 2024-09-11
Rework how we check calls to support deduced implicit parameters

zygoloid:
Instead of the `call` instruction having a block with one argument per explicit argument, preceded optionally by `self` and followed optionally by a return slot, change the `call` to store only the *runtime* arguments. Store an index on the runtime parameters to make it easier to determine the correspondence between arguments and parameters in a call. Compile-time parameters, whether implicit or explicit, are no longer included in the call argument list. Instead, they're tracked only in the `specific_id` on the callee.

For calls to generic classes and generic interfaces, it no longer makes sense to form a `call` instruction, given that the entirety of the result is determined by the `specific_id`, which is now formed when checking the call. Instead, the `call` instruction now only models function calls, and not calls to other kinds of parameterized entity names, and we create a `class_type` or `interface_type` instead of a `call` instruction to model these kinds of calls. Notionally the model here is that we're following the #3720 approach for calls, but for now we inline the `Call.Op` function when forming SemIR.

We now also track the enclosing specific for a generic class or generic interface that appears within an enclosing generic. This is necessary in order for deduction of the inner generic parameters to not get confused by the outer generic parameters  being absent.

In order to not regress diagnostics, the template argument deduction mechanism has been extended to specify the name of the parameter we're deducing against when possible, and call arity mismatch errors are now diagnosed before performing deduction rather than afterwards.


zygoloid:
Let me know if you'd prefer that I try to split some smaller parts out of this for review purposes -- there's probably a few pieces here that could be landed separately.

toolchain/sem_ir/formatter.cpp
```diff
@@ -829,6 +837,14 @@ class FormatterImpl {
     }
   }
 
+  auto FormatInstRHS(GenericInterfaceType inst) -> void {
+    if (inst.enclosing_specific_id.is_valid()) {
```
jonmeow, line 841: Minor suggestion, comparing with FunctionType, ClassType, and InterfaceType, maybe add a helper:

```
// Shared formatting for entities that have an optional specific.
template<typename EntityId>
auto FormatInstRHSForEntity(EntityId entity_id, SpecificId specific_id) -> void {
  if (specific_id.is_valid()) {
    FormatArgs(entity_id, specific_id);
  } else {
    FormatArgs(entity_id);
  }
}
```

toolchain/lower/file_context.cpp
```diff
@@ -213,10 +213,13 @@ auto FileContext::BuildFunctionDecl(SemIR::FunctionId function_id)
...
   for (auto param_ref_id :
        llvm::concat<const SemIR::InstId>(implicit_param_refs, param_refs)) {
-    auto param_type_id =
-        SemIR::Function::GetParamFromParamRefId(sem_ir(), param_ref_id)
-            .second.type_id;
-    switch (auto value_rep = SemIR::ValueRepr::ForType(sem_ir(), param_type_id);
+    auto [param_id, param] =
+        SemIR::Function::GetParamFromParamRefId(sem_ir(), param_ref_id);
+    if (!param.index.is_valid()) {
+      // Parameter is not passed at runtime.
```
jonmeow: Do you think the comment would still be helpful if the name were `param.runtime_index`?

Partly noting due to the identical comment on 316. It strikes me a related option might be, had you considered starting to add accessors to the typed instructions, e.g. `param.is_runtime()` or `is_lowered`?

Note, I'm also fine leaving this as-is if you prefer the comment, the code is clear.

toolchain/check/call.cpp
```diff
@@ -9,60 +9,100 @@
...
 
+// Information about a callee in a checked call to an entity.
+struct CalleeInfo {
+  // Whether the call appears to be valid.
+  bool is_valid;
+  // The specific version of the callee to invoke.
+  SemIR::SpecificId specific_id;
+};
+
+static auto ResolveCalleeInCall(Context& context, SemIR::LocId loc_id,
```
jonmeow, line 29: This looks like a pretty complex signature, is it worth a comment? What's the semantic of self_id?

toolchain/check/call.cpp
```diff
@@ -9,60 +9,100 @@
...
+
+  // Perform argument deduction.
+  auto specific_id = SemIR::SpecificId::Invalid;
+  if (entity_generic_id.is_valid()) {
+    specific_id = DeduceGenericCallArguments(
+        context, loc_id, entity_generic_id, enclosing_specific_id,
+        callee_info.implicit_param_refs_id, callee_info.param_refs_id, self_id,
+        arg_ids);
+    if (!specific_id.is_valid()) {
+      return {.is_valid = false, .specific_id = SemIR::SpecificId::Invalid};
```
jonmeow: Had you considered a `std::optional<SpecificId>` return, with `nullopt` for invalid calls?

toolchain/check/handle_function.cpp
```diff
@@ -65,6 +65,51 @@ static auto DiagnoseModifiers(Context& context, DeclIntroducerState& introducer,
...
+    if (bind_name) {
+      param_id = bind_name->value_id;
+      param = context.insts().Get(param_id);
+    }
+
+    auto param_inst = param.TryAs<SemIR::Param>();
+    if (!param_inst) {
+      // Once we support more generalized patterns we will need to diagnose
+      // parameters with unsupported patterns.
+      context.TODO(param_id, "unexpected syntax for parameter");
```
jonmeow, line 96: There aren't any tests hitting this case, can there be, or is the code just not there yet?

toolchain/check/import_ref.cpp
```diff
@@ -1266,14 +1269,31 @@ class ImportRefResolver {
...
+      auto import_specific_id = SemIR::SpecificId::Invalid;
+      if (auto import_generic_class_type =
+              import_ir_.types().TryGetAs<SemIR::GenericClassType>(
+                  inst.type_id)) {
+        import_specific_id = import_generic_class_type->enclosing_specific_id;
+      }
+      auto specific_data = GetLocalSpecificData(import_specific_id);
+      if (HasNewWork()) {
+        // This is the end of the first phase. Don't make a new class yet if
+        // we already have new work.
```
jonmeow, line 1283: This is marked as the end of the first phase, but so is line 1289. Is this actually becoming four phases? Should the check on line 1288 just be removed, comparing with the interface version?

I'll be approving assuming 1288 should just be removed, please let me know if I'm misunderstanding.

(note, this inconsistency makes me wonder if there's a good way to factor towards more code sharing)

jonmeow (approved):
This makes sense to me. There's just one inconsistency in import_ref.cpp that I'm trying to figure out, but otherwise looks good. Approving per my comment there.

toolchain/sem_ir/formatter.cpp
zygoloid: Yeah, we have a bunch of instructions like this. I'll send a follow-up PR for this.

toolchain/lower/file_context.cpp
zygoloid: Renamed to `runtime_index`.

toolchain/check/call.cpp
zygoloid: Commented.

toolchain/check/call.cpp
zygoloid: Done.

toolchain/check/handle_function.cpp
zygoloid: The code to parse more complex patterns isn't there yet; I've added a test so we should cover this once we get the parsing support.

toolchain/check/import_ref.cpp
zygoloid: Thanks, the second check is redundant and it looks like I forgot to delete it. Deleted now matching the other cases.

Yeah, I think it'd be good to look into refactoring function / class / interface handling.

PR #4305, opened 2024-09-12
Switch to ctx.label.package for rule, and adjust relative path handling

jonmeow:
build_file_path works with bazel but I'd missed it's [deprecated](https://bazel.build/rules/lib/builtins/ctx#build_file_path).

Relative path handling also could use some improvements.

Both of these are really to support non-standard environments, essentially.


chandlerc (approved):
LGTM

josh11b (approved):

PR #4306, opened 2024-09-12
Assert that the llvm::Function is created with the same name as requested

dwblaikie:
This will catch some cases of bugs in the name mangling logic - if within a single file we incorrectly mangle two distinct entities to the same name, llvm::Function will assign a new name to the second copy showing one of the two entities should have a distinct name/is missing something in their mangling.

### DIFF
--- a/toolchain/lower/file_context.cpp
+++ b/toolchain/lower/file_context.cpp
@@ -250,7 +250,7 @@ auto FileContext::BuildFunctionDecl(SemIR::FunctionId function_id)
                              mangled_name, llvm_module());
 
   CARBON_CHECK(llvm_function->getName() == mangled_name,
-               "Do not rely on llvm::Function overload renaming");
+               "Mangled name collision: {0}", mangled_name);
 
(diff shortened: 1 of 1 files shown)


toolchain/lower/file_context.cpp
```diff
@@ -249,6 +249,9 @@ auto FileContext::BuildFunctionDecl(SemIR::FunctionId function_id)
       llvm::Function::Create(function_type, llvm::Function::ExternalLinkage,
                              mangled_name, llvm_module());
 
+  CARBON_CHECK(llvm_function->getName() == mangled_name,
+               "Do not rely on llvm::Function overload renaming");
```
zygoloid: I think it'd be useful to include `mangled_name` in the diagnostic here.

zygoloid (approved):

toolchain/lower/file_context.cpp
dwblaikie: Oh, didn't know CARBON_CHECK could have a parameterized message. Done that

PR #4310, opened 2024-09-13
Style notes on passing and storing object addresses

jonmeow:
[Context](https://discord.com/channels/655572317891461132/821113559755784242/1283516297686286377). Some relevant Google C++ style is at [Inputs and Outputs](https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs). #4301 is an example application of the style.

### DIFF
--- a/docs/project/cpp_style_guide.md
+++ b/docs/project/cpp_style_guide.md
@@ -147,9 +147,13 @@ these.
         although these guidelines differ slightly.
 -   Always mark constructors `explicit` unless there's a specific reason to
     support implicit or `{}` initialization.
--   When storing unowned objects as members, prefer to pass by reference and
-    store as a pointer. For example, `Foo* foo_;` and
-    `Bar(Foo& foo) : foo_(&foo)`.
+-   When passing a reference to an unowned object as an argument, use a reference
+    unless one of the following cases applies:
+    -    If it is optional, use a pointer and document that it may be null.
+    -    If it is captured and must outlive the call expression itself, use a
+         pointer an document that it must not be null (unless it is also optional).
+-   When storing a reference to an unowned object as a member, prefer to store as a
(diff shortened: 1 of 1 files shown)


zygoloid (commented):
LGTM

docs/project/cpp_style_guide.md
```diff
@@ -147,6 +147,9 @@ these.
         although these guidelines differ slightly.
 -   Always mark constructors `explicit` unless there's a specific reason to
     support implicit or `{}` initialization.
+-   When storing unowned objects as members, prefer to pass by reference and
+    store as a pointer. For example, `Foo* foo_;` and
+    `Bar(Foo& foo) : foo_(&foo)`.
```
chandlerc: Based on further discussion, let's keep the storing part but add some more explicit guidance for passing. This is allowed and *slightly* encouraged in the Google style guide, but I think we can be a bit more direct and explicit for Carbon:

```suggestion
-   When passing a reference to an unowned object as an argument, use a reference
    unless one of the following cases applies:
    -    If it is optional, use a pointer and document that it may be null.
    -    If it is captured and must outlive the call expression itself, use a
         pointer an document that it must not be null (unless it is also optional).
-   When storing a reference to an unowned object as a member, prefer to store as a
    pointer. For example, `Foo* foo_;` and `Bar(Foo* foo) : foo_(foo)`.
```

chandlerc:
I think the thread is coming to consensus now, and suggested an update to match what I think I'm seeing there... but also maybe I'm misreading so good to use the suggested change to make sure we're all on the same page.

docs/project/cpp_style_guide.md
jonmeow: Pulled, and made some edits (including "when passing/storing a reference to an unowned object" -> "when passing/storing an object's address" in order to avoid different meanings of "reference", cleaning up the example)

chandlerc (approved):
LGTM, nice rewording too. =]

PR #4312, opened 2024-09-13
Apply `is_closed_import` to imported namespaces

dwblaikie:
Spin off from https://github.com/carbon-language/carbon-lang/pull/4294#discussion_r1752916451

### DIFF
--- a/toolchain/check/import.cpp
+++ b/toolchain/check/import.cpp
@@ -80,6 +80,14 @@ static auto CopyNameFromImportIR(Context& context,
   return import_name_id;
 }
 
+namespace {
+struct NamespaceResult {
+  SemIR::NameScopeId name_scope_id;
+  SemIR::InstId inst_id;
+  bool is_duplicate_of_namespace_in_current_package;
+};
+}  // namespace
+
 // Adds a namespace to the IR. The bool on return is true if there was a name
 // conflict. diagnose_duplicate_namespace is used when handling a cross-package
 // import, where an existing namespace is in the current package and the new
@@ -89,7 +97,7 @@ static auto AddNamespace(Context& context, SemIR::TypeId namespace_type_id,
                          SemIR::NameScopeId parent_scope_id,
                          bool diagnose_duplicate_namespace,
                          llvm::function_ref<SemIR::InstId()> make_import_id)
-    -> std::tuple<SemIR::NameScopeId, SemIR::InstId, bool> {
+    -> NamespaceResult {
   auto* parent_scope = &context.name_scopes().Get(parent_scope_id);
   auto insert_result =
       parent_scope->name_map.Insert(name_id, parent_scope->names.size());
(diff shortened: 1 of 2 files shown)


toolchain/check/import.cpp
```diff
@@ -165,7 +165,8 @@ static auto CopySingleNameScopeFromImportIR(
     Map<SemIR::NameScopeId, SemIR::NameScopeId>* copied_namespaces,
     SemIR::ImportIRId ir_id, SemIR::InstId import_inst_id,
     SemIR::NameScopeId import_scope_id, SemIR::NameScopeId parent_scope_id,
-    SemIR::NameId name_id) -> std::pair<SemIR::NameScopeId, SemIR::InstId> {
+    SemIR::NameId name_id)
+    -> std::tuple<SemIR::NameScopeId, SemIR::InstId, bool> {
```
zygoloid: It'd be useful to update the function comment to describe the new part of the return type. Also, maybe it's worth using a struct for this type? It looks like the same type is now returned from here and from `AddNamespace`.

toolchain/check/import.cpp
```diff
@@ -500,12 +497,14 @@ static auto AddNamespaceFromOtherPackage(Context& context,
...
-  context.name_scopes()
-      .Get(new_scope_id)
-      .import_ir_scopes.push_back({import_ir_id, import_ns.name_scope_id});
+  auto [new_scope_id, namespace_inst_id, is_duplicate] =
+      CopySingleNameScopeFromImportIR(
+          context, namespace_type_id, /*copied_namespaces=*/nullptr,
+          import_ir_id, import_inst_id, import_ns.name_scope_id,
+          parent_scope_id, name_id);
+  auto& scope = context.name_scopes().Get(new_scope_id);
+  scope.is_closed_import = !is_duplicate;
```
zygoloid: I found this a bit confusing; I wonder if better naming for `is_duplicate` would help. Something like `is_duplicate_of_namespace_in_current_package` would be clearer as to why it matters here, but maybe there's a shorter way of expressing the same thing? (`found_in_current_package`?)

zygoloid (commented):
Thanks, seems reasonable.

toolchain/check/import.cpp
dwblaikie: Updated comment and introduced a named return type (`NamespaceResult` - could be more verbose (`NamespaceInsertionResult`? `AddNamespaceResult` seemed ambiguous (could be a verb "add (namespace result)" or noun "(add namespace) result") so I shied away from that).

toolchain/check/import.cpp
dwblaikie: Went with the verbose name - it's only written 3 times (now that there's a named type, which reduces some of the mentions) so being especially long doesn't seem too bad

zygoloid: Works for me, thanks.

toolchain/check/import.cpp
zygoloid: Yeah, `NamespaceResult` seems fine given that it's scoped to this file.

toolchain/check/import.cpp
```diff
@@ -500,13 +509,16 @@ static auto AddNamespaceFromOtherPackage(Context& context,
...
       context.GetBuiltinType(SemIR::BuiltinInstKind::NamespaceType);
-  auto [new_scope_id, namespace_inst_id] = CopySingleNameScopeFromImportIR(
-      context, namespace_type_id, /*copied_namespaces=*/nullptr, import_ir_id,
-      import_inst_id, import_ns.name_scope_id, parent_scope_id, name_id);
-  context.name_scopes()
-      .Get(new_scope_id)
-      .import_ir_scopes.push_back({import_ir_id, import_ns.name_scope_id});
-  return namespace_inst_id;
+  NamespaceResult
+      result =  // auto [new_scope_id, namespace_inst_id, is_duplicate] =
```
```suggestion
  NamespaceResult result =
```

zygoloid (approved):

toolchain/check/import.cpp
```diff
@@ -500,13 +509,15 @@ static auto AddNamespaceFromOtherPackage(Context& context,
...
-      import_inst_id, import_ns.name_scope_id, parent_scope_id, name_id);
-  context.name_scopes()
-      .Get(new_scope_id)
-      .import_ir_scopes.push_back({import_ir_id, import_ns.name_scope_id});
-  return namespace_inst_id;
+  NamespaceResult result =
+      CopySingleNameScopeFromImportIR(
+          context, namespace_type_id, /*copied_namespaces=*/nullptr,
+          import_ir_id, import_inst_id, import_ns.name_scope_id,
+          parent_scope_id, name_id);
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  NamespaceResult result = CopySingleNameScopeFromImportIR(
      context, namespace_type_id, /*copied_namespaces=*/nullptr, import_ir_id,
      import_inst_id, import_ns.name_scope_id, parent_scope_id, name_id);
```

PR #4315, opened 2024-09-13
Omit a trailing `SpecificId` by default when formatting instruction arguments.

zygoloid:


jonmeow (approved):
Feels a little like magic, but maybe that's okay.

PR #4317, opened 2024-09-15
Improve access checking code

brymer-meneses:
This change accomplishes the TODOs for access checking. More specifically it,
- makes `SemIR::AccessKind` formattable using `llvm::formatv`.
- makes use of `LookupUnqualifiedName` to find `Self`.

### DIFF
--- a/toolchain/check/context.cpp
+++ b/toolchain/check/context.cpp
@@ -299,7 +299,8 @@ auto Context::LookupNameInDecl(SemIR::LocId loc_id, SemIR::NameId name_id,
 }
 
 auto Context::LookupUnqualifiedName(Parse::NodeId node_id,
-                                    SemIR::NameId name_id) -> LookupResult {
+                                    SemIR::NameId name_id, bool required)
+    -> LookupResult {
   // TODO: Check for shadowed lookup results.
 
   // Find the results from ancestor lexical scopes. These will be combined with
@@ -327,8 +328,11 @@ auto Context::LookupUnqualifiedName(Parse::NodeId node_id,
             .inst_id = lexical_result};
   }
 
-  // We didn't find anything at all.
-  DiagnoseNameNotFound(node_id, name_id);
+  if (required) {
+    // We didn't find anything at all.
+    DiagnoseNameNotFound(node_id, name_id);
+  }
+
   return {.specific_id = SemIR::SpecificId::Invalid,
           .inst_id = SemIR::InstId::BuiltinError};
 }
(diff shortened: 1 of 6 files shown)


toolchain/sem_ir/name_scope.h
```diff
@@ -6,16 +6,46 @@
...
 #include "toolchain/sem_ir/inst.h"
 
 namespace Carbon::SemIR {
 
 // Access control for an entity.
-enum class AccessKind : int8_t {
-  Public,
-  Protected,
-  Private,
+class AccessKind : public Printable<AccessKind> {
```
jonmeow: In general, what we're trying to do with enum type wrappers is add more functionality than just an enum can provide. e.g., associating both names and additional data directly with the enum value. We might want something lighter weight when we don't need all the features.

For example, had you considered providing a `format_provider`? We're doing this elsewhere for enums used with diagnostics; here's an example: https://github.com/carbon-language/carbon-lang/blob/trunk/toolchain/parse/context.cpp#L31-L49

toolchain/check/context.cpp
```diff
@@ -327,8 +328,11 @@ auto Context::LookupUnqualifiedName(Parse::NodeId node_id,
             .inst_id = lexical_result};
   }
 
-  // We didn't find anything at all.
-  DiagnoseNameNotFound(node_id, name_id);
+  if (required) {
+    // We didn't find anything at all.
```
```suggestion
  // We didn't find anything at all.
  if (required) {
```

I think the comment is explaining the current state, regardless of the value of `required`, so probably should remain outside the `if`.

toolchain/check/testdata/class/access_modifers.carbon
```diff
@@ -143,19 +143,12 @@ library "[[@TEST_NAME]]";
 
 class A {
   private let internal: i32 = 10;
-  // CHECK:STDERR: fail_todo_global_self_access.carbon:[[@LINE+13]]:16: ERROR: Member access into incomplete class `A`.
+  // CHECK:STDERR: fail_todo_global_self_access.carbon:[[@LINE+6]]:16: ERROR: Member access into incomplete class `A`.
```
jonmeow: This test probably doesn't need to be a todo anymore (i.e., `fail_todo` -> `fail`). But you might want to add a test that should pass:

```
library "[[@TEST_NAME]]";

class A {
  private fn F() {}
  private fn G() { return Self.F(); }
}
```

jonmeow (approved):
Looks good, thanks!

brymer-meneses:
Seems like this was failing because of some line changes, sorry about that.

jonmeow:
No worries!

PR #4318, opened 2024-09-16
Use `TupleAccess` instead of `TupleIndex`

brymer-meneses:
This change removes the `TupleIndex` instruction, and instead consolidate it with the `TupleAccess` instruction, per this [discussion](https://discord.com/channels/655572317891461132/655578254970716160/1271195835975204946). This change, in turn removes `AnyAggregateIndex`.


toolchain/check/member_access.cpp
```diff
@@ -570,29 +571,32 @@ auto PerformTupleIndex(Context& context, SemIR::LocId loc_id,
...
-    index_inst_id = SemIR::InstId::BuiltinError;
+    return SemIR::InstId::BuiltinError;
   } else if (!index_const_id.is_template()) {
     // TODO: Decide what to do if the index is a symbolic constant.
     CARBON_DIAGNOSTIC(TupleIndexNotConstant, Error,
                       "Tuple index must be a constant.");
     context.emitter().Emit(loc_id, TupleIndexNotConstant);
-    index_inst_id = SemIR::InstId::BuiltinError;
-  } else {
-    auto index_literal = context.insts().GetAs<SemIR::IntLiteral>(
```
brymer-meneses, line 581: This change is necessary since without early returning a `SemIR::InstId::BuiltinError` I hit this https://github.com/carbon-language/carbon-lang/blob/06344aeb7cd22c3a8b32cc7b48d49a57f8bcbafa/toolchain/check/eval.cpp#L468 on this test https://github.com/carbon-language/carbon-lang/blob/06344aeb7cd22c3a8b32cc7b48d49a57f8bcbafa/toolchain/check/testdata/var/no_prelude/fail_init_type_mismatch.carbon#L15
Which is a weird bug, since it only happens when `--no-prelude-import` is passed, and seems to be okay without.

toolchain/sem_ir/typed_insts.h
```diff
@@ -102,8 +102,7 @@ struct AnyAggregateAccess {
 // Common representation for aggregate index nodes, which access an element
 // determined by evaluating an expression.
 struct AnyAggregateIndex {
-  static constexpr InstKind Kinds[] = {InstKind::ArrayIndex,
-                                       InstKind::TupleIndex};
+  static constexpr InstKind Kinds[] = {InstKind::ArrayIndex};
```
geoffromer: If only an `ArrayIndex` can be an `AnyAggregateIndex`we should be able to drop `AnyAggregateIndex` and use `ArrayIndex` instead.

toolchain/check/member_access.cpp
```diff
@@ -546,12 +546,13 @@ auto PerformCompoundMemberAccess(
...
-auto PerformTupleIndex(Context& context, SemIR::LocId loc_id,
-                       SemIR::InstId tuple_inst_id, SemIR::InstId index_inst_id)
-    -> SemIR::InstId {
+auto PerformTupleAccess(Context& context, SemIR::LocId loc_id,
+                        SemIR::InstId tuple_inst_id,
+                        SemIR::InstId index_inst_id) -> SemIR::InstId {
   tuple_inst_id = ConvertToValueOrRefExpr(context, tuple_inst_id);
   auto tuple_inst = context.insts().Get(tuple_inst_id);
   auto tuple_type_id = tuple_inst.type_id();
+  auto tuple_index = SemIR::ElementIndex::Invalid;
```
geoffromer: Does this need to be declared up here, or could we declare it down on line 594 where it's initialized?

toolchain/check/member_access.cpp
geoffromer: I think the root issue here is that all the existing code that works with `ElementIndex` values isn't prepared to handle invalid values (since until now they weren't possible). So without this change, this function was emitting an invalid `ElementIndex`, and then some code somewhere tried to use it as if it were a valid index. And I think the fix you have here is the right one: make sure an invalid `ElementIndex` never leaves this function.

toolchain/sem_ir/ids.h
```diff
@@ -710,12 +710,17 @@ constexpr TypeBlockId TypeBlockId::Invalid = TypeBlockId(InvalidIndex);
 // An index for element access, for structs, tuples, and classes.
 struct ElementIndex : public IndexBase, public Printable<ElementIndex> {
   using IndexBase::IndexBase;
+
+  static const ElementIndex Invalid;
```
geoffromer: As noted on my other comment, the existing code that handles `ElementIndex` values isn't prepared to handle an invalid one, so I think we should avoid giving this type an invalid value in the first place.

geoffromer (approved):

PR #4319, opened 2024-09-16
Remove `[[clang::preserve_most]]`

chandlerc:
These appear to be causing some subtle misinteractions with MSan that we don't understand, and may be a compiler bug. =/ Fortunately, they weren't essential to the performance gains so just remove them for now. When benchmarked on an x86 server, where I would expect this to be more important due to relatively few named registers, the performance change appears to be either an improvement or in the noise.

Huge credit to Jon for tracking down that this is related to the MSan issues.


jonmeow (approved):
Thanks for checking the performance effect!

> misinteractions with MSan

This investigation actually started with an opt crash (not msan), particularly with clang near head. It's just that we can observe it with both msan and opt.

PR #4320, opened 2024-09-17
Support initialization of specific classes from struct literals

zygoloid:
Add support for initializing types like `GenericClass(i32)` from a struct literal. A new kind of instruction, `complete_type_witness`, is added to the class definition to track the object representation type so that it's visible to the generics machinery. Accesses to the object representation of a class have all been updated to pass in the class's `SpecificId` so that the types of the fields of the specific class are used instead of the types of the fields of the generic class in places that look at the object representation -- primarily class initialization.


zygoloid:
Most of the test changes are caused by the change in representation. For the functionality changes, see `toolchain/check/testdata/class/generic/init.carbon`.

toolchain/check/handle_class.cpp
```diff
@@ -561,55 +561,89 @@ auto HandleParseNode(Context& context, Parse::BaseDeclId node_id) -> bool {
...
+// returns a corresponding complete type witness instruction.
+static auto CheckCompleteAdapterClassType(Context& context,
+                                          Parse::NodeId node_id,
+                                          SemIR::ClassId class_id,
+                                          SemIR::InstBlockId fields_id)
+    -> SemIR::InstId {
+  const auto& class_info = context.classes().Get(class_id);
+  if (class_info.base_id.is_valid()) {
+    CARBON_DIAGNOSTIC(AdaptWithBase, Error,
+                      "Adapter cannot have a base class.");
```
jonmeow, line 574: Getting to the diagnostic phrasing discussion, should this be "Adapter has a base class"?

Note I don't have a strong opinion, just asking to try to get a similar voice as we add diagnostics.

Oh, but more seriously... you can probably remove the capital and period now. The related PR is still pending but I think it'd make sense to stop writing those into new diagnostics.

toolchain/check/handle_class.cpp
```diff
@@ -561,55 +561,89 @@ auto HandleParseNode(Context& context, Parse::BaseDeclId node_id) -> bool {
...
+    CARBON_DIAGNOSTIC(AdaptBaseHere, Note, "`base` declaration is here.");
```
jonmeow, line 575: FWIW with this and AdaptFieldsHere,  had you considered "AdaptWithBaseHere", using the main error as a prefix to tie the note more closely together? (or, AdaptBase/AdaptFields, dropping "With")

jonmeow (approved):
Looks good, mainly a minor question if you want to start writing new diagnostics using #4321 since it seems we have consensus there.

zygoloid:
Per in-person discussion: the diagnostic stuff here was just moved to a different function, not added or renamed, so I'm going to leave those comments for a separate PR.

PR #4321, opened 2024-09-17
Iterate on diagnostic structure

jonmeow:
Part of this is also a plan to change "ERROR:" -> "error:" in diagnostics. I think I'm the odd one out on sentence casing. The rest is mostly trying to figure out advice that I think we can live with.

Note I'm not immediately planning to rewrite all diagnostics (unless maybe there's a simple regex). I'm more trying to redirect style a little to where preferences lie, particularly for new diagnostics.

### DIFF
--- a/toolchain/docs/diagnostics.md
+++ b/toolchain/docs/diagnostics.md
@@ -165,10 +165,10 @@ methods for formatting arguments:
     [llvm::formatv](https://llvm.org/doxygen/FormatVariadic_8h_source.html)
     support.
     -   This includes `char` and integer types (`int`, `int32_t`, and so on).
-    -   String types also use this and can be added as needed, but are
-        discouraged in favor of other types.
-        -   `llvm::StringLiteral` should be used when working with literals; use
-            `std::string` when allocations are required.
+    -   String types can be added as needed, but are discouraged in favor of
+        other types.
+        -   Use `llvm::StringLiteral` where appropriate; use `std::string` when
+            allocations are required.
         -   `llvm::StringRef` is disallowed due to lifetime issues.
 -   `llvm::format_provider<...>` specializations.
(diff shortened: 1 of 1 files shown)


zygoloid (commented):
Thanks, these guidelines look reasonable to me.

chandlerc (commented):
LGTM as well, leaving it for Dave in case he has comments here.

dwblaikie:
Any chance we could include a reference to clang's diagnostic style guide as a baseline+exceptions/clarificaitons for Carbon? (somewhat like carbon's style guide basing on the google style guide) since the existence of these rules is at least partially motivated by compatibility both for the ecosystem in general, and for cases in the future where a single compiler invocation may produce both clang and carbon diagnostics together.

jonmeow:
> Any chance we could include a reference to clang's diagnostic style guide as a baseline+exceptions/clarificaitons for Carbon? (somewhat like carbon's style guide basing on the google style guide) since the existence of these rules is at least partially motivated by compatibility both for the ecosystem in general, and for cases in the future where a single compiler invocation may produce both clang and carbon diagnostics together.

I've added a note that we're trying to be similar. I wasn't taking a "baseline+exceptions" because of a couple reasons:

- While we're trying to be similar to Clang for intermingled diagnostics, we're also trying to stand on our own as a programming language. I want developers to be able to read our documentation and understand what to do, without looking at Clang's documentation (which also differs for the basic `%0` versus `{0}` in examples).

- A fair amount of the prior phrasing feels to me like it's trying to address things similar to what Clang's diagnostic wording is, but in a slightly different way, and so I'm not quite sure how to phrase that as "+exceptions" without rewriting most of either or both.

dwblaikie (approved):

PR #4322, opened 2024-09-18
Add a clang subcommand.

jonmeow:
This makes something like `bazel run :toolchain -- clang -- -c test.cpp` work, because that can be run in-process. Note that `bazel run :toolchain -- clang -- test.cpp` still requires subprocessing, and does not work.

Note, the vision here is that we are trying to align how clang and carbon compile c++ code. This is work towards intertwining command execution.

### DIFF
--- a/toolchain/driver/clang_runner.cpp
+++ b/toolchain/driver/clang_runner.cpp
@@ -62,7 +62,6 @@ auto ClangRunner::Run(llvm::ArrayRef<llvm::StringRef> args) -> bool {
     maybe_v_arg = v_arg_storage;
   }
 
-  CARBON_CHECK(!args.empty());
   CARBON_VLOG("Running Clang driver with arguments: \n");
 
   // Render the arguments into null-terminated C-strings for use by the Clang
--- a/toolchain/driver/clang_runner.cpp
+++ b/toolchain/driver/clang_runner.cpp
@@ -35,8 +35,9 @@
 // While not in a header, this is the API used by llvm-driver.cpp for
 // busyboxing.
 //
-// NOLINTNEXTLINE
-int clang_main(int Argc, char** Argv, const llvm::ToolContext& ToolContext);
+// NOLINTNEXTLINE(readability-identifier-naming)
+auto clang_main(int Argc, char** Argv, const llvm::ToolContext& ToolContext)
+    -> int;
 
 namespace Carbon {
 
--- a/toolchain/driver/clang_runner.cpp
+++ b/toolchain/driver/clang_runner.cpp
@@ -126,7 +126,8 @@ auto ClangRunner::Run(llvm::ArrayRef<llvm::StringRef> args) -> bool {
 
   // When there's only one command being run, this will run it in-process.
   // However, a `clang` invocation may cause multiple `cc1` invocations, which
-  // still subprocess.
+  // still subprocess. See `InProcess` comment at:
+  // https://github.com/llvm/llvm-project/blob/86ce8e4504c06ecc3cc42f002ad4eb05cac10925/clang/lib/Driver/Job.cpp#L411-L413
   //
   // TODO: It would be nice to find a way to set up the driver's understanding
   // of the executable name in a way that causes the multiple `cc1` invocations
--- a/toolchain/driver/clang_runner.cpp
+++ b/toolchain/driver/clang_runner.cpp
@@ -23,10 +23,20 @@
 #include "llvm/ADT/StringRef.h"
 #include "llvm/IR/LLVMContext.h"
 #include "llvm/Support/FileSystem.h"
+#include "llvm/Support/LLVMDriver.h"
 #include "llvm/Support/Path.h"
 #include "llvm/Support/Program.h"
 #include "llvm/Support/VirtualFileSystem.h"
 #include "llvm/TargetParser/Host.h"
+
+// Defined in:
+// https://github.com/llvm/llvm-project/blob/main/clang/tools/driver/driver.cpp
+//
+// While not in a header, this is the API used by llvm-driver.cpp for
+// busyboxing.
+//
+// NOLINTNEXTLINE
+int clang_main(int Argc, char** Argv, const llvm::ToolContext& ToolContext);
 
 namespace Carbon {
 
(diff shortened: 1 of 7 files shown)


jonmeow:
Note, I'm sending this to you as-is because I want to make sure the subprocessing behavior aligns with your expectations. I believe this is getting equivalent behavior as to clang:

```
╚╡strace -fe trace=clone,fork,execve clang -c test.cpp
execve("/usr/lib/llvm-16/bin/clang", ["clang", "-c", "test.cpp"], 0x7ffc84ae67e0 /* 41 vars */) = 0
+++ exited with 0 +++
```

```
╚╡strace -fe trace=clone,fork,execve clang test.cpp
execve("/usr/lib/llvm-16/bin/clang", ["clang", "test.cpp"], 0x7ffd23f946f8 /* 41 vars */) = 0
strace: Process 1074218 attached
[pid 1074218] execve("/usr/lib/llvm-16/bin/clang", ["/usr/lib/llvm-16/bin/clang", "-cc1", "-triple", "x86_64-pc-linux-gnu", "-emit-obj", "-mrelax-all", "-disable-free", "-clear-ast-before-backend", "-disable-llvm-verifier", "-discard-value-names", "-main-file-name", "test.cpp", "-mrelocation-model", "pic", "-pic-level", "2", "-pic-is-pie", "-mframe-pointer=all", "-fmath-errno", "-ffp-contract=on", "-fno-rounding-math", "-mconstructor-aliases", "-funwind-tables=2", "-target-cpu", "x86-64", "-tune-cpu", "generic", "-mllvm", "-treat-scalable-fixed-error-as-w"..., "-debugger-tuning=gdb", "-fcoverage-compilation-dir=/usr/"..., "-resource-dir", ...], 0x7fffd11b9750 /* 41 vars */) = 0
[pid 1074218] +++ exited with 0 +++
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=1074218, si_uid=18553, si_status=0, si_utime=2 /* 0.02 s */, si_stime=2 /* 0.02 s */} ---
strace: Process 1074219 attached
[pid 1074219] execve("/usr/bin/ld", ["/usr/bin/ld", "-pie", "--hash-style=gnu", "--build-id", "--eh-frame-hdr", "-m", "elf_x86_64", "-dynamic-linker", "/lib64/ld-linux-x86-64.so.2", "-o", "a.out", "/lib/x86_64-linux-gnu/Scrt1.o", "/lib/x86_64-linux-gnu/crti.o", "/usr/lib/gcc/x86_64-linux-gnu/14"..., "-L/usr/lib/gcc/x86_64-linux-gnu/"..., "-L/usr/lib/gcc/x86_64-linux-gnu/"..., "-L/lib/x86_64-linux-gnu", "-L/lib/../lib64", "-L/usr/lib/x86_64-linux-gnu", "-L/usr/lib/../lib64", "-L/usr/lib/llvm-16/bin/../lib", "-L/lib", "-L/usr/lib", "/tmp/test-46b600.o", "-lgcc", "--as-needed", "-lgcc_s", "--no-as-needed", "-lc", "-lgcc", "--as-needed", "-lgcc_s", ...], 0x7fffd11b9750 /* 41 vars */) = 0
[pid 1074219] +++ exited with 0 +++
```

toolchain/driver/clang_runner.cpp
```diff
@@ -113,9 +124,15 @@ auto ClangRunner::Run(llvm::ArrayRef<llvm::StringRef> args) -> bool {
   driver.Dir = installation_->llvm_install_bin();
   CARBON_VLOG("Setting bin directory to: {0}\n", driver.Dir);
 
-  // TODO: Directly run in-process rather than using a subprocess. This is both
-  // more efficient and makes debugging (much) easier. Needs code like:
-  // driver.CC1Main = [](llvm::SmallVectorImpl<const char*>& argv) {};
+  // When there's only one command being run, this will run it in-process.
+  // However, a `clang` invocation may cause multiple `cc1` invocations, which
+  // still subprocess.
```
chandlerc: This makes sense to me.

I have an idea of another step we may want to eventually try to take here, but emphasis on "may" (not at all sure) and "eventually". =] Suggesting as a TODO in case you want to capture it and/or tweak the wording in the code. But also happy to just leave for discussion or the future.

```suggestion
  // When there's only one command being run, this will run it in-process.
  // However, a `clang` invocation may cause multiple `cc1` invocations, which
  // still subprocess.
  //
  // TODO: It would be nice to find a way to set up the driver's understanding
  // of the executable name in a way that causes the multiple `cc1` invocations
  // to actually result in `carbon clang -- ...` invocations (even if as
  // subprocesses). This may dovetail with having symlinks that redirect to a
  // busybox of LLD as well, and having even the subprocesses consistently run
  // the Carbon install toolchain and not a system toolchain whenever possible.
```

Maybe this is already part of what you're thinking about with busyboxing the linker too, feel free to skip the TODO if already part of your larger plans.

chandlerc (commented):
Yeah, all of this matches my expectation.

Worth adding a small test that checks the whole `-c` flow maybe? Also left an optional TODO idea, but that's not blocking or especially important.

toolchain/driver/clang_runner.cpp
jonmeow: Added. Also yes, that's what I've been working towards with busyboxing and why I was examining the subprocess behavior of ClangRunner.

chandlerc:
FWIW, I think this LGTM other than this question:
 
> Worth adding a small test that checks the whole `-c` flow maybe?

If we can't test this usefully yet, I'm OK with that but maybe worth mentioning in the PR. Was hoping we could at least get a trivial `int test() { return 0; }` kind of test through `-c` compilation with this sufficiently portably.

jonmeow:
Sorry, I'd missed this comment. It seems to be difficult because ClangRunner doesn't forward the VFS. Also, the `link` command seems to be untested (other than the invocation via sieve) -- for similar reasons?

I can work on fixing the missed VFS forwarding, it looks feasible. It's non-trivial to add though, so will be a bit. If you're okay with it, I'd prefer to break apart issues.

jonmeow:
Note, can't even write a trivial test using the FileTest infrastructure because the Driver pipes output straight to stderr, which is a nuisance for autoupdate.

jonmeow:
Actually, added a couple tests alongside the link one in clang_runner_test

chandlerc (approved):
That test looks perfect and about what I'd expected to be able to do given the limits you point out.

Nice to figure out how to thread VFS stuff through some day maybe, but definitely not here, was just hoping for something quick to make sure the basic code doesn't fall over. =D

PR #4329, opened 2024-09-19
`where` check stage, step 1: designators

josh11b:
Right now, there is no checking of `where` requirements. The result of a where expression is just the type on the left-hand side. It does now introduce `.Self` so that it is available in expressions on the right-hand side, in addition to designators corresponding to the members of type on the left-hand side. Note, though, that diagnostics could still be improved significantly.


toolchain/check/node_stack.h
```diff
@@ -444,10 +444,13 @@ class NodeStack {
         case Parse::NodeKind::ImplDefinitionStart:
           return Id::KindFor<SemIR::ImplId>();
         case Parse::NodeKind::SelfValueName:
+        case Parse::NodeKind::SelfTypeName:
```
jonmeow:
```suggestion
        case Parse::NodeKind::SelfTypeName:
        case Parse::NodeKind::SelfValueName:
```

Lexical ordering

toolchain/check/handle_where.cpp
```diff
@@ -8,41 +8,78 @@
...
+  context.node_stack().PopExpr();
+  context.node_stack().PopExpr();
+  return true;
 }
 
-auto HandleParseNode(Context& context, Parse::RequirementAndId node_id)
+auto HandleParseNode(Context& /*context*/, Parse::RequirementAndId /*node_id*/)
     -> bool {
-  return context.TODO(node_id, "HandleRequirementAnd");
+  // Nothing to do
```
jonmeow:
```suggestion
  // Nothing to do.
```

[Punctuation style](https://google.github.io/styleguide/cppguide.html#Punctuation,_Spelling_and_Grammar)

josh11b:
FYI, there are a lot of FIXME comments in this PR. I was hoping to get feedback so I could resolve them before merging.

toolchain/check/handle_where.cpp
```diff
@@ -8,41 +8,78 @@
...
+  context.scope_stack().Push();
+  // Introduce `.Self` as a symbolic binding. Its type is the value of the
+  // expression to the left of `where`, so `MyInterface` in the example above.
+  // Because there is no equivalent non-symbolic value, we use `Invalid` as
+  // the `value_id` on the `BindSymbolicName`.
+  auto entity_name_id = context.entity_names().Add(
+      {.name_id = SemIR::NameId::PeriodSelf,
+       .parent_scope_id = context.decl_name_stack().PeekParentScopeId(),
+       .bind_index = context.scope_stack().AddCompileTimeBinding()});
+  // FIXME: should this have a location associated with `node_id`?
```
jonmeow: I think current code doesn't generally associate locations for `Self` because it's implicitly generated rather than declared; there wouldn't be anything to refer back to other than the main declaration (`interface I;`). I think you could use `self_node` though (maybe better than `node_id`, the `where` operand?). I'm not sure whether it'll come up in diagnostics either way, but there might be subtly different results.

toolchain/check/handle_name.cpp
```diff
@@ -168,6 +173,30 @@ auto HandleParseNode(Context& context, Parse::NameQualifierId /*node_id*/)
   return true;
 }
 
+auto HandleParseNode(Context& context, Parse::DesignatorExprId node_id)
+    -> bool {
+  SemIR::NameId name_id = context.node_stack().PopName();
+
+  // TODO: set `required` to `false` and check for not-found to give a clearer
+  // error when .Self is not found.
+  // FIXME: should this be a call to HandleNameAsExpr instead?
```
jonmeow: That might be the easiest way to get the constant & NameRef structure (I do think you want that, though @zygoloid would probably be faster at coming up with test cases). I don't know if you want a helper as noted below to avoid a push/pop in the member access case, I think either approach might be fine.

toolchain/check/handle_where.cpp
```diff
@@ -8,41 +8,78 @@
...
     -> bool {
-  return context.TODO(node_id, "HandleRequirementImpls");
+  // TODO: Implement
+  context.node_stack().PopExpr();
+  context.node_stack().PopExpr();
+  return true;
 }
 
-auto HandleParseNode(Context& context, Parse::RequirementAndId node_id)
+auto HandleParseNode(Context& /*context*/, Parse::RequirementAndId /*node_id*/)
```
jonmeow, line 70: Are there tests of all the handers that you're enabling here? I see some = and ==, but no `and` or `impls` tests. Even with TODOs, I think it would be good to have tests to verify you're not adding crash bugs (`return context.TODO` is `return false`, which stops checking and prevents crashes).

jonmeow (commented):
Generally makes sense -- focusing comments on FIXMEs per request, but noting a couple minor style things that I thought would probably be final.

toolchain/check/handle_where.cpp
josh11b: `node_id` has the location of the `where` keyword (with the type to the left as its child), which is where the `.Self` name was introduced. `self_node` is the type to the left of `where`, which is relevant but the connection is less direct.

toolchain/check/handle_where.cpp
josh11b: Added some more testing.

toolchain/check/handle_name.cpp
zygoloid: In principle, yes, this should be using `HandleNameAsExpr` I think, though I don't think it really makes much difference right now:

The work to map to a value in a specific is probably not going to be very important here -- that mostly affects cases where we find an unqualified lookup result in an extended scope such as a base class, and I don't think we expect to find `.Self` in such a scope -- but it wouldn't be wrong to do it, and it'd be more correct under maintenance / future language changes.

Building a `NameRef` would be nice (for consistency and for tooling if nothing else), though `.Self` is a bit weird compared to other names we might reference because it's not explicitly declared. I think I'd still be slightly on the side of including a `NameRef` though -- and more generally calling `HandleNameAsExpr`.

In terms of dealing with the undesired push, maybe the cleanest thing would be to move the existing push out to the callers. It'd be repeated three times, but I think it's most obvious what effect a handler has on the stack if it directly does all of its pushes and pops, so that seems OK to me.

toolchain/check/handle_where.cpp
jonmeow: `node_id` might have a more direction connection because it informs the value of `.Self` for the RHS, but how do you anticipate the location being used for diagnostics or debug line numbers?

For example, in my head if we have a declaration like:

```
impl forall [U:! type, T:! AddWith(U) where .Self impls AssignWith(.Self.Result)]
     T as AddAssignWith(U) {
```

I could see generating a diagnostic like:

```
error: `Result` is not a member of `.Self`
impl forall [U:! type, T:! AddWith(U) where .Self impls AssignWith(.Self.Result)]
                                                                         ^~~~~~
note: `.Self` defined as `AddWith(U)` here
impl forall [U:! type, T:! AddWith(U) where .Self impls AssignWith(.Self.Result)]
                           ^~~~~~~~~~
```

IIUC, that would be using `self_node` to directly get `AddWith(U)`? Can you give an example of how you'd expect `node_id` to be used if it were in the location?

jonmeow: I should note too, maybe we'd just say "`Result` is not a member of `AddWith(U)`" and never reference the origin of that type. That's a possibility too, and I think part of why we end up not using `Self` locations in most cases. Then having no location isn't a big issue, although maybe we should still be trying to do something.

toolchain/check/handle_name.cpp
jonmeow: > Building a `NameRef` would be nice (for consistency and for tooling if nothing else), though `.Self` is a bit weird compared to other names we might reference because it's not explicitly declared.

I think `self` and `Self` are similar, and using `HandleNameAsExpr`.

https://github.com/carbon-language/carbon-lang/blob/trunk/toolchain/check/handle_name.cpp#L150-L163

josh11b: Should I only call `HandleNameAsExpr` in the `name_id == SemIR::NameId::SelfType` / `.Self` branch of the `if` and do something else in the `.Member` branch?

josh11b: In particular, do we want to have a `.Member` produce a name reference to `.Self`?

josh11b: I've updated the PR so that it calls `HandleNameAsExpr` in both branches of the `if`, and so always produces a name reference to `.Self`. I added a note in the `.Member` case to explain the error when `.Selef` is not found in the current scope.

toolchain/check/handle_where.cpp
josh11b: My thought was that the error would be like what you wrote, but also include `where`.

```
note: `.Self` defined as `AddWith(U)` here
impl forall [U:! type, T:! AddWith(U) where .Self impls AssignWith(.Self.Result)]
                           ^~~~~~~~~~~~~~~~
```

I think any of the three options is reasonable though.

josh11b: Thinking about this further, I think there is another example we should consider:
```
fn F[T:! AddWith(.Self)](x: T, y: T)
```
In this case the `.Self` is introduced by the `:!`. Furthermore, the current idea is to mark the `T` node as being part of the compile-time binding since the `:!` is output at the end of the expression (unless we were to make it a virtual node too). So I think we should go with the no-node approach for now, since that is all I can easily support with this other case.

toolchain/check/handle_name.cpp
```diff
@@ -168,6 +178,37 @@ auto HandleParseNode(Context& context, Parse::NameQualifierId /*node_id*/)
...
+  } else {
+    // Otherwise this is `.Member`, so look up `.Self` and then `Member` in
+    // `.Self`.
+    SemIR::InstId period_self_id = SemIR::InstId::Invalid;
+    {
+      DiagnosticAnnotationScope annotate_diagnostics(
+          &context.emitter(), [&](auto& builder) {
+            CARBON_DIAGNOSTIC(
+                NoPeriodSelfForDesignator, Note,
+                "Designator may only be used when `.Self` is in scope.");
```
jonmeow: Minimal style change, considering See #4321:

```suggestion
                "designator may only be used when `.Self` is in scope");
```

But, had you considered changing HandleNameAsExpr to pass `/*required=*/false` to `LookupUnqualifiedName` and emitting as a full diagnostic here, rather than an annotation?

Note, a similar diagnostic might be `CoreNameNotFound` which uses: ``name `Core.{0}` implicitly referenced here, but not found``... So if you had control over the diagnostic, maybe you could just replace the current ``name `.Self` not found`` with a more similar ``name `.Self` implicitly referenced by designated expression, but not found``.

jonmeow (commented):
Looks good, just a question about the diagnostic format.

toolchain/check/handle_name.cpp
josh11b: Made the style change, and added a TODO to make the more invasive change to the diagnostic.

jonmeow (approved):

PR #4330, opened 2024-09-20
Add a `note: ` prefix to all notes.

zygoloid:
Omit the `note: ` prefix and the snippet from the "in import" note that precedes a diagnostic.

This makes our diagnostic output more closely match that of Clang and GCC.


toolchain/diagnostics/diagnostic_consumer.cpp
```diff
@@ -27,10 +27,15 @@ auto StreamDiagnosticConsumer::HandleDiagnostic(Diagnostic diagnostic) -> void {
...
+        *stream_ << "note: ";
+        break;
+      case DiagnosticLevel::Context:
         break;
     }
     *stream_ << message.format_fn(message) << "\n";
-    message.loc.FormatSnippet(*stream_);
+    if (message.level != DiagnosticLevel::Context) {
+      message.loc.FormatSnippet(*stream_);
+    }
```
chandlerc: I'm OK starting here, but I think we should add a TODO or do some follow-up work as a consequence.

At the moment, the `in import` message doesn't contain the name of the imported or importing library or any other information. The name of the imported library however is visible in the snippet before this change and removed after. So it somewhat looks like a regression.

However, I don't think relying on the snippet for this information is the right call, so I do think this is the right direction. My guess is we want to revisit the contextual diagnostics to tweak their output at some point, but happy for that to be in the future.

chandlerc (approved):
LGTM as I think this is a good initial step, but see inline -- I think it exposes some additional work we should probably make a note of on the import diagnostics. With the snippets, it wasn't as obviously needed, but seems more important now.

toolchain/diagnostics/diagnostic.h
```diff
@@ -18,8 +18,12 @@
 namespace Carbon {
 
 enum class DiagnosticLevel : int8_t {
-  // A note, not indicating an error on its own, but possibly providing context
-  // for an error.
+  // Information about the location of another diagnostic, showing how we
+  // reached that location. This is currently only used for the "in import"
+  // message.
+  Context,
```
jonmeow: We have a lot of things named "Context" in the toolchain, and the uses of this look like `, Context,`. What do you think of "PrefixNote", "Origin", or something like that?

toolchain/diagnostics/diagnostic_consumer.cpp
jonmeow: FWIW, my take would be that while the message could provide a library name, shouldn't we still associate a line of code when possible? i.e., if a developer is trying to figure out why they're getting the library, wouldn't that be helpful?

chandlerc: I think the location is still there, as part of the message prefix? It's just the snippet being elided.

jonmeow: Sure then, the snippet. I find snippets useful, not sure why we'd elide them. If this is what you both want though, can you please add comments explaining the choice?

zygoloid: Comment added.

toolchain/diagnostics/diagnostic.h
zygoloid: Renamed to `LocationInfo`, does that work better for you?

toolchain/diagnostics/diagnostic_consumer.cpp
chandlerc: FWIW, I don't feel strongly about the snippets either way. I somewhat see both sides -- they both seem useful and a bit verbose. I guess matching Clang and GCC tilts me very, very slightly towards eliding them, but I think it'd be fine to revisit that if this is a big regression Jon. Or revisit it by exploring other ways to capture some of the value.

But maybe that's OK to fix-forward given the `note: ` churn that this PR already has?

chandlerc:
(All of this still LG, but good to wait for Jon before landing)

toolchain/diagnostics/diagnostic.h
jonmeow: Seems good enough.

toolchain/diagnostics/diagnostic_consumer.cpp
jonmeow: I'm in disagreement, but I think I'm in the minority.

jonmeow (approved):

PR #4331, opened 2024-09-20
Improve install_paths handling for relative paths.

jonmeow:
Because install_paths is not presently validated, and it's resolved after the `SetWorkingDirForBazel` call, if a relative path is used with bazel then it would fail silently. This starts making the driver share install path errors, and starts changing how `//toolchain` launches `carbon`.

Note the implementation is still brittle and will break with symlinks. That's something I plan to address as part of busyboxing.

### DIFF
--- a/toolchain/install/BUILD
+++ b/toolchain/install/BUILD
@@ -234,7 +234,7 @@ pkg_tar_and_test(
 
 # Support `bazel run` on specific binaries.
 run_tool(
-    name = "carbon.bin",
+    name = "run_carbon",
     data = [":install_data"],
     tool = "prefix_root/bin/carbon",
 )
--- a/toolchain/install/BUILD
+++ b/toolchain/install/BUILD
@@ -7,6 +7,7 @@ load("@llvm-project//llvm:binary_alias.bzl", "binary_alias")
 load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test")
 load("@rules_pkg//pkg:mappings.bzl", "pkg_attributes", "pkg_filegroup", "pkg_files", "pkg_mklink", "strip_prefix")
 load("pkg_helpers.bzl", "pkg_naming_variables", "pkg_tar_and_test")
+load("run_tool.bzl", "run_tool")
 load("symlink_filegroup.bzl", "symlink_filegroup")
 
 package(default_visibility = ["//visibility:public"])
@@ -229,4 +230,11 @@ pkg_tar_and_test(
         ":install_data",
     ],
     test_install_marker = ":install_marker",
+)
+
+# Support `bazel run` on specific binaries.
+run_tool(
+    name = "carbon.bin",
+    data = [":install_data"],
+    tool = "prefix_root/bin/carbon",
 )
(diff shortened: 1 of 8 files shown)


toolchain/install/BUILD
```diff
@@ -230,3 +231,10 @@ pkg_tar_and_test(
     ],
     test_install_marker = ":install_marker",
 )
+
+# Support `bazel run` on specific binaries.
+run_tool(
+    name = "carbon.bin",
```
chandlerc: I expected `carbon.bin` to name a binary target (as in, a file, or some data).

WDYT about:
```suggestion
    name = "run_carbon",
```

toolchain/install/install_paths.h
```diff
@@ -83,8 +83,9 @@ class InstallPaths {
   // fails for any reason, it will `CARBON_CHECK` fail with the error message.
   static auto MakeForBazelRunfiles(llvm::StringRef exe_path) -> InstallPaths;
 
-  // Provide an explicit install paths prefix. This is useful for testing or for
-  // using Carbon in an environment with an unusual path to the installed files.
+  // Provide an explicit install paths prefix, which must be absolute. This is
+  // useful for testing or for using Carbon in an environment with an unusual
+  // path to the installed files.
```
chandlerc, line 88: Could you also document on the class that instances use absolute paths to be resilient to working directory changes once constructed? It's an interesting property, and not one I had thought about.

chandlerc (commented):
Nice catch and nice fix on the tool running!

Some minor suggestions inline.

Also, just to write down what was in my head reviewing this -- I'm a bit sad at needing absolute paths. I tried to figure out why and it is mostly a weird thing: when working with a toolchain and some code in the root of some subtree, and using relative full paths to everything, it seems nice to not surface the working directory path in things like error messages and such. That "shouldn't" be relevant.

But as this PR shows, it *is* relevant if users ever change their working directory after creating the install paths construct! And it seems really nice and principled to make the install paths idempotent or durable to such changes. So I end up really liking the behavior this provides even though I started off very hesitant.

I think the thing I liked about relative paths is actually not about them being relevant, but more about them being minimal and not surfacing extraneous details. And if we ever really care about this, there are better ways to get there and a better time to get there: when we're actually printing or serializing something. That's the point where we could try to compute useful working directory and/or a minimal relative path. But it does seem much more principled to compute the absolute path here in order to not be brittle in the face of working directory changes later on.

toolchain/install/install_paths.h
jonmeow: Done. (note, the above paragraph is just a re-wrap)

toolchain/install/BUILD
jonmeow: Done

toolchain/install/install_paths.h
jonmeow: Actually, moved the comment to `prefix()` if you're okay with that. Just trying to avoid adding to the already-long class comment.

I imagine the re-wrap spot will still be apparent.

chandlerc: On one hand, I like the comment on prefix.

I still somewhat feel like it might be relevant to have a class level comment because this impacts when you should call some of the factory functions...

But not really blocking here.

chandlerc (approved):
LG

toolchain/install/install_paths.h
jonmeow: Moved to the class. I'll send you a separate change with suggested cleanups for the class comment.

PR #4332, opened 2024-09-20
Refactor install structure to make changes easier.

jonmeow:
Instead of separately constructing file information for `filegroup` and `pkg_filegroup`, this instead creates a single structure which is used to generate both. Additionally, I'm unifying the `llvm_link_data` and `install_lib_data` targets (though the `llvm_link_data` target is problematic for busyboxing, I don't think it can keep working as it does right now).

Note I'm also stopping reuse of llvm's binary_alias. We need to be able to symlink non-binary files, so I'm going to just share logic there. (plus, I admit I find the name "binary_alias" confusing since it's not an alias in bazel terms)

### DIFF
--- a/toolchain/install/install_filegroups.bzl
+++ b/toolchain/install/install_filegroups.bzl
@@ -1,0 +1,128 @@
+# Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+# Exceptions. See /LICENSE for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+"""Rules for constructing install information."""
+
+load("@rules_pkg//pkg:mappings.bzl", "pkg_attributes", "pkg_filegroup", "pkg_files", "pkg_mklink", "strip_prefix")
+load("symlink_helpers.bzl", "symlink_file", "symlink_filegroup")
+
+def install_filegroup(name, filegroup_target):
+    """Adds a filegroup for install.
+
+    Used in the `install_dirs` dict.
+
+    Args:
+      name: The base directory for the filegroup.
+      filegroup_target: A relative path for the symlink.
+    """
+    return {
+        "filegroup": filegroup_target,
+        "for_driver": False,
+        "name": name,
+    }
+
+def install_symlink(name, symlink_to):
+    """Adds a symlink for install.
+
+    Used in the `install_dirs` dict.
+
+    Args:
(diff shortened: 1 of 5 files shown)


jonmeow:
Depends on #4331

jonmeow:
Note, I grant the different starlark here is subjective. If you find the previous version more understandable, let me know, but also happy to try to improve understandability here.

toolchain/install/install_filegroups.bzl
```diff
@@ -0,0 +1,132 @@
...
+load("symlink_helpers.bzl", "symlink_file", "symlink_filegroup")
+
+def install_filegroup(name, filegroup_target):
+    """Adds a filegroup for install.
+
+    Used in the `install_dirs` dict.
+
+    Args:
+      name: The base directory for the filegroup.
+      filegroup_target: A relative path for the symlink.
```
chandlerc: Not clear what "symlink" this is a target for?

Or was this supposed to read "A relative path for the filegroup"? If so, not sure I understand what that means.

chandlerc: Having finish reading, I think this is just a copy/paste. I'm pretty sure I understand how this is used, maybe:

```suggestion
      filegroup_target: The bazel filegroup target to install.
```

toolchain/install/install_filegroups.bzl
```diff
@@ -0,0 +1,132 @@
...
+      name: The filename to use.
+      symlink_to: A relative path for the symlink.
+    """
+    return {
+        "for_driver": False,
+        "name": name,
+        "symlink": symlink_to,
+    }
+
+def install_target(name, target, executable = False, for_driver = False):
```
chandlerc: What do you think about s/`for_driver`/`is_driver`/ ? That would read a bit better to me I think...

```suggestion
def install_target(name, target, executable = False, is_driver = False):
```

chandlerc (commented):
Generally happy with this approach, and with the starlark. I agree it's nicer to just have a single cohesive structure. Some fairly minor comments inline.

toolchain/install/install_filegroups.bzl
jonmeow: Done

jonmeow: Done

chandlerc (approved):
LG!

PR #4336, opened 2024-09-24
Basic support for declaring generic impls.

zygoloid:
Update impl handling to more closely match other kinds of declaration, including support for declaring and defining gneeric `impl`s.

When of performing redeclaration lookup for impls by looking for the self and constraint type, produce a list of impls rather than a single impl because it's possible for there to be multiple impls with different deduced parameters but the same self type and constraint. Following #3763, only consider impls to be redeclarations if they're spelled the same, not just if they have the same self and constraint types.

No support yet for impl selection to deduce the arguments of a generic impl.


toolchain/sem_ir/impl.h
```diff
@@ -50,20 +35,44 @@ struct Impl : public Printable<Impl> {
...
-    return result.value();
+  // TODO: Switch to something like TinyPtrVector. We expect it to be rare for
+  // there to be more than one ImplId per bucket.
+  using IdVector = llvm::SmallVector<ImplId, 1>;
+
+  // Looks up the list of impls with this self type and constraint. This only
+  // includes impls from the current file and its API file.
+  auto LookupBucket(TypeId self_id, TypeId constraint_id) -> IdVector& {
+    return lookup_
+        .Insert(std::pair{self_id, constraint_id}, [] { return IdVector(); })
```
jonmeow: Sorry, this is making me stumble a bit on reading. Maybe a pass on comments/names would help? To detail where I'd appreciate edits...

The `Lookup` name of this function feels like a mismatch with the `Insert` call. The mutation, adding an empty list, could invalidate references but isn't noted in the comments.

Also, with "bucket" being added to the name, this doesn't seem to be referred to as a "bucket" elsewhere [in this file in particular]. There's "IdVector" and "lookup_", although maybe those can be renamed to reflect the "bucket" name?

Or is "lookup bucket" supposed to be a full noun, versus verb + noun? The statement "looks up" makes me read "lookup" as a verb, but in import_ref.cpp you say "add this to impl lookup" which seems more noun-like. If you intend a noun, maybe something like `GetOrAddLookupBucket` with a comment "Returns the lookup bucket for impls ..." (replacing "Looks up the list of impls ...")?

toolchain/check/handle_impl.cpp
```diff
@@ -206,22 +259,48 @@ static auto BuildImplDecl(Context& context, Parse::AnyImplDeclId node_id,
...
-  static_cast<void>(params_id);
+  // Add the impl declaration.
+  SemIR::ImplDecl impl_decl = {.impl_id = SemIR::ImplId::Invalid,
+                               .decl_block_id = decl_block_id};
+  auto impl_decl_id =
+      context.AddPlaceholderInst(SemIR::LocIdAndInst(node_id, impl_decl));
+
+  SemIR::Impl impl_info = {
+      name_context.MakeEntityWithParamsBase(name, impl_decl_id,
+                                            /* is_extern=*/false,
```
```suggestion
                                            /*is_extern=*/false,
```

I'm confused that clang-format didn't do this

toolchain/check/handle_impl.cpp
```diff
@@ -206,22 +259,48 @@ static auto BuildImplDecl(Context& context, Parse::AnyImplDeclId node_id,
...
+      context.impls().LookupBucket(impl_info.self_id, impl_info.constraint_id);
+  for (auto prev_impl_id : impl_lookup) {
+    if (MergeImplRedecl(context, impl_info, prev_impl_id)) {
+      impl_decl.impl_id = prev_impl_id;
+      break;
+    }
+  }
+
+  bool is_new_impl = !impl_decl.impl_id.is_valid();
+  if (is_new_impl) {
```
```suggestion
  // Create a new impl if this isn't a valid redeclaration.
  if (!impl_decl.impl_id.is_valid()) {
```

Comparing with similar logic in other entities (class/function/interface), with `is_new_class`, the variable is reused. Here, this is the only use of `is_new_impl`; the declaration made me look, which is why I'm suggesting eliding it. But then I expect the variable name is somewhat documenting, so suggesting the comment similar to what we have elsewhere.

toolchain/check/handle_impl.cpp
```diff
@@ -206,22 +259,48 @@ static auto BuildImplDecl(Context& context, Parse::AnyImplDeclId node_id,
...
 
   // Add the impl declaration.
   // TODO: Does lookup in an impl file need to look for a prior impl declaration
   // in the api file?
-  auto impl_id = context.impls().LookupOrAdd(self_type_id, constraint_type_id);
-  SemIR::ImplDecl impl_decl = {.impl_id = impl_id,
-                               .decl_block_id = decl_block_id};
-  auto impl_decl_id = context.AddInst(node_id, impl_decl);
+  auto& impl_lookup =
+      context.impls().LookupBucket(impl_info.self_id, impl_info.constraint_id);
```
```suggestion
  auto& impl_bucket =
      context.impls().LookupBucket(impl_info.self_id, impl_info.constraint_id);
```

For me, "lookup" is a verb, "bucket" is the noun for the returned object.

toolchain/check/import_ref.cpp
```diff
@@ -1568,6 +1570,131 @@ class ImportRefResolver {
         context_.types().GetConstantId(interface_val.type_id()));
   }
 
+  // Forms a constant ID that represents the given instruction ID that denotes
+  // an impl declaration. Impl declarations don't have a constant value, but for
+  // the purposes of importing we pretend that they do.
+  //
+  // TODO: Consider alternative options here. We could make impl declarations
+  // actually evaluate to themselves to avoid this hack.
```
jonmeow, line 1588: I'm fine with this as a TODO, but I would've thought impl decls should be constant. Maybe it hasn't come up just because we mostly talk about them in terms of the constraints they satisfy? I wonder if metaprogramming would end up forcing this.

toolchain/check/import_ref.cpp
```diff
@@ -1568,6 +1570,131 @@ class ImportRefResolver {
...
+
+      // Push a block so that we can add scoped instructions to it.
+      context_.inst_block_stack().Push();
+      AddNameScopeImportRefs(import_scope, new_scope);
+      new_impl.body_block_id = context_.inst_block_stack().Pop();
+    }
+  }
+
+  auto TryResolveTypedInst(SemIR::ImplDecl inst,
+                           SemIR::ConstantId impl_const_id) -> ResolveResult {
```
jonmeow, line 1649: Where should I be looking for tests of what's changing here? I'm not seeing more things being tested in the modified `import.carbon` files, and maybe we don't have any tests of `forall` on import boundaries?

What about out-of-line definitions of functions in an impl? (or is that just not supported yet, even though there are some notes around line 1620?)

toolchain/check/import_ref.cpp
```diff
@@ -1568,6 +1570,131 @@ class ImportRefResolver {
...
+          new_impl.first_owning_decl_id, SemIR::NameId::Invalid,
+          new_impl.parent_scope_id);
+      auto& new_scope = context_.name_scopes().Get(new_impl.scope_id);
+      const auto& import_scope =
+          import_ir_.name_scopes().Get(import_impl.scope_id);
+
+      // Push a block so that we can add scoped instructions to it.
+      context_.inst_block_stack().Push();
+      AddNameScopeImportRefs(import_scope, new_scope);
+      new_impl.body_block_id = context_.inst_block_stack().Pop();
```
jonmeow, line 1644: I think I'm missing something, what's actually added to the body block? I see there are similar calls a couple other places, but `AddNameScopeImportRefs` uses `AddImportRef` which uses `AddPlaceholderInstInNoBlock` -- I believe the import refs are being tracked separately, not as part of `inst_block_stack`.

toolchain/check/import_ref.cpp
```diff
@@ -1568,6 +1570,131 @@ class ImportRefResolver {
...
+
+    if (HasNewWork()) {
+      return Retry(impl_const_id);
+    }
+
+    auto& new_impl = context_.impls().Get(impl_id);
+    new_impl.parent_scope_id = parent_scope_id;
+    new_impl.implicit_param_refs_id = GetLocalParamRefsId(
+        import_impl.implicit_param_refs_id, implicit_param_const_ids);
+    CARBON_CHECK(!new_impl.param_refs_id.is_valid());
```
```suggestion
    CARBON_CHECK(!import_impl.param_refs_id.is_valid() && !new_impl.param_refs_id.is_valid());
```

FWIW, my leaning would be to assume that the new id is invalid because it hasn't been set and, if checking anything, check that there was nothing to import. Maybe just check both?

jonmeow (commented):
Generally makes sense. I'm trying to avoid asking for things you're intending as a TODO here (particularly around redeclarations).

toolchain/sem_ir/impl.h
zygoloid: I was thinking of "look up" as a verb, but yeah, that's not a good match for what it does. Renamed to `GetOrAddLookupBucket` as suggested.

toolchain/check/import_ref.cpp
zygoloid: It's not clear to me. `impl` declaration instructions currently don't have a type or a value, and there's no way to refer to them, so this hasn't come up before and I suspect probably won't come up in other contexts. But it'd definitely be convenient here for them to evaluate to themselves. I think we should give that a go; I'm planning on trying that in a follow-up PR.

toolchain/check/import_ref.cpp
zygoloid: Yeah, all the other places that call `AddNameScopeImportRefs` do this. It does look like it's not necessary any more (if it ever was); I'll see if I can remove all of those after this lands.

toolchain/check/import_ref.cpp
```diff
@@ -1568,6 +1570,131 @@ class ImportRefResolver {
...
+
+    if (HasNewWork()) {
+      return Retry(impl_const_id);
+    }
+
+    auto& new_impl = context_.impls().Get(impl_id);
+    new_impl.parent_scope_id = parent_scope_id;
+    new_impl.implicit_param_refs_id = GetLocalParamRefsId(
+        import_impl.implicit_param_refs_id, implicit_param_const_ids);
+    CARBON_CHECK(!import_impl.param_refs_id.is_valid() && !new_impl.param_refs_id.is_valid());
```
```suggestion
    CARBON_CHECK(!import_impl.param_refs_id.is_valid() &&
                 !new_impl.param_refs_id.is_valid());
```

toolchain/check/import_ref.cpp
zygoloid: Added check/testdata/impl/no_prelude/import_generic.carbon to test this. In doing so, I found a couple of things that didn't work; fixed. (Though I think the way we're importing generic eval blocks is a bit broken in general and I should have another look at that; I've patched it up enough to get the test passing at least.)

There's currently no support for naming an `impl` as a `NameComponent`, so it's not possible to define a `default fn` of an `impl` out of line yet. (We also don't support `default fn` in general.) So I don't think that can be tested right now. :(

toolchain/check/handle_impl.cpp
zygoloid: The reference could also be invalidated by the call to `FinishGenericDecl` triggering an import. Changed to redo the lookup on insert to avoid that.

zygoloid: Renamed to `lookup_bucket`.

toolchain/sem_ir/impl.h
jonmeow: Thanks! FWIW, maybe still `IdVector` -> `LookupBucket` (that is, even given the TODO I assume you'd still have `using IdVector = TinyIdVector<ImplId>;`, so calling it the `LookupBucket` seems consistent)

toolchain/sem_ir/impl.h
```diff
@@ -50,20 +35,46 @@ struct Impl : public Printable<Impl> {
...
-    });
-    return result.value();
+  // TODO: Switch to something like TinyPtrVector. We expect it to be rare for
+  // there to be more than one ImplId per bucket.
+  using IdVector = llvm::SmallVector<ImplId, 1>;
+
+  // Returns the lookup bucket containing the list of impls with this self type
+  // and constraint, or adds a new bucket if this is the first time we've seen
+  // an impl of this kind. The lookup bucket only includes impls from the
+  // current file and its API file.
```
jonmeow:
```suggestion
  // an impl of this kind. Callers must only append impls from the
  // current file and its API file.
```

Suggesting this partly because the phrasing implies the lookup bucket is complete ("only includes"), rather than modified.

toolchain/check/import_ref.cpp
```diff
@@ -1643,8 +1652,14 @@ class ImportRefResolver {
 
     SemIR::ImplId impl_id = SemIR::ImplId::Invalid;
     if (!impl_const_id.is_valid()) {
-      // For impl loading, we can always start at the second phase. Create a
-      // forward declaration of the impl for any recursive references.
+      if (HasNewWork()) {
+        // This is the end of the first phase. Don't make a new impl yet if we
+        // already have new work.
+        return Retry();
+      }
```
jonmeow, line 1660: When is there new work here? I thought work could only be added by TryResolveTypedInst, and I'm not seeing what in the code above would result in new work. For example, comparing to other similar entity logic, this doesn't resolve a specific before checking HasNewWork.

toolchain/check/import_ref.cpp
```diff
@@ -2009,9 +2025,14 @@ class ImportRefResolver {
...
-        FinishPendingGeneric(pending_generics_.pop_back_val());
+      // Process generics in the order that we added them because a later
+      // generic might refer to an earlier one.
+      // TODO: Import the generic eval block rather than calling
+      // RebuildGenericEvalBlock to rebuild it so that order doesn't matter.
+      for (size_t i = 0; i != pending_generics_.size(); ++i) {
+        FinishPendingGeneric(pending_generics_[i]);
       }
+      pending_generics_.clear();
       while (!pending_specifics_.empty()) {
```
jonmeow, line 2041: Would it be worth updating the specific loop to match? According to FinishPendingSpecific, it can add more specifics, so maybe a similar issue could arise?

(or if you prefer, maybe it's just worth it to have both loops read the same?)

toolchain/sem_ir/impl.h

toolchain/check/import_ref.cpp
zygoloid: In principle, new work can be added by `TryResolveInst` before we even get here. (This would happen when resolving a generic constant that refers to an impl, which is actually not possible because we never form a generic constant that refers to an impl, but this code shouldn't be assuming that.)

toolchain/sem_ir/impl.h
zygoloid: Rephrased to clarify that this is a constraint on the caller.

toolchain/check/import_ref.cpp
zygoloid: There is no corresponding issue for specifics, because importing them doesn't call into any code that assumes the SemIR is in a valid state, and the approach used for generics is less efficient (it might allocate more memory, for example).

I've extended the comment to try to explain a bit better why generics need more delicate handling than specifics here.

toolchain/check/import_ref.cpp
jonmeow: Maybe this should `CARBON_CHECK(!HasNewWork())` instead?

toolchain/check/import_ref.cpp
jonmeow: Okay, although couldn't the current loop also be less efficient since it's a backwards iteration and the `pop_back_val` is a repeated resize instead of `clear()` once?

jonmeow (approved):
Looks good. I'd prefer to `CARBON_CHECK(!HasNewWork())` if that works for you, but either way I think this is in good shape to merge.

toolchain/check/import_ref.cpp
zygoloid: It could be. My expectation is that the extra allocation cost would outweigh all the other factors, and the loop body contains enough work that the resize and reverse iteration won't be above the noise floor.

toolchain/check/import_ref.cpp
zygoloid: It could be, but I think that couples this code unnecessarily to the assumption that we don't do any extra work before we get here, so I'd prefer this as-is.

PR #4337, opened 2024-09-24
Flush pending diagnostics on crash.

jonmeow:
This risks diagnsotic formatting crashing, but I think we more frequently see cases where it'd be interesting to know what diagnostics were being delayed as part of the default sorting.

### DIFF
--- a/toolchain/driver/compile_subcommand.cpp
+++ b/toolchain/driver/compile_subcommand.cpp
@@ -666,13 +666,14 @@ auto CompileSubcommand::Run(DriverEnv& driver_env) -> DriverResult {
   });
 
   PrettyStackTraceFunction flush_on_crash([&](llvm::raw_ostream& out) {
-    // When crashing, flush diagnostics. If sorting diagnostics, they go to the
-    // crash stream; if streaming, the original stream is flushed.
+    // When crashing, flush diagnostics. If sorting diagnostics, they can be
+    // redirected to the crash stream; if streaming, the original stream is
+    // flushed.
     // TODO: Eventually we'll want to limit the count.
     if (options_.stream_errors) {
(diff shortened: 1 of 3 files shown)


josh11b (approved):

toolchain/driver/compile_subcommand.cpp
```diff
@@ -660,6 +665,18 @@ auto CompileSubcommand::Run(DriverEnv& driver_env) -> DriverResult {
     stream_consumer.Flush();
   });
 
+  PrettyStackTraceFunction flush_on_crash([&](llvm::raw_ostream& out) {
+    // When crashing, print pending diagnostics to the crash stream.
+    // TODO: Eventually we'll want to limit the count.
+    out << "Pending diagnostics\n";
```
zygoloid: Should we skip printing this if diagnostic streaming is enabled? (And maybe if there are no diagnostics?)

jonmeow: PrettyStackTrace requires a [stack allocation](https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/Support/PrettyStackTrace.h#L47-L51). A unique_ptr sort of works, but only if you immediately allocate it (or never allocate it) so that [construction and] destruction order is still stack-like. I believe it's infeasible to skip printing here (without removing the PrettyStackTraceFunction altogether): you end up with a number and malformatted stack trace output. We can print something different, but we cannot choose to skip printing based on later state.

I've added a slightly different label if streaming is enabled. We could skip both printing and flushing for streamed errors (i.e., just don't allocate a PrettyStackTraceFunction in this case), but my leaning would be that it's potentially helpful and more consistent to flush in both modes.

josh11b (approved):

PR #4338, opened 2024-09-24
Disallow compile time bindings where they aren't clearly supported.

jonmeow:
This is resolving a fuzz-discovered crash related to function suspends and compile time bind indices. Although the crash originally came from clearly invalid syntax (missing the `=` inside a `class` decl), the syntax with a value should also be valid but has the same crash.

This approach disallows compile-time bindings in contexts that can create ambiguous results, particularly class declarations. These are an issue because a suspended function can have let declarations after it. I'm allowing them in function bodies and interface scopes.


zygoloid (commented):
I think this is only hiding the issue, but we're still in an inconsistent state. If we have

```
class C {
  fn F(T:! type) {}
  var U:! type;
}
```
then we would assign the index 0 to both `T` and `U` when we come to handle the definition of `F`. Setting the next index to 2 in this case doesn't fix the broken numbering of existing generic parameters -- the binding at position i in the stack should have a compile time binding index of i, and it won't, so our invariants are still broken.

I think we should either not assign an index to `U` in the first place, or pop it from the scope at the end of that field declaration.

jonmeow:
Note, if you prefer, I think I could also only disallow in class and impl scopes. Those both should have suspended function issues. Package scope I think I can actually make work under the current approach.

toolchain/check/handle_binding_pattern.cpp
```diff
@@ -200,7 +202,21 @@ auto HandleParseNode(Context& context, Parse::BindingPatternId node_id)
...
+  bool is_generic = true;
+  if (context.decl_introducer_state_stack().innermost().kind ==
+      Lex::TokenKind::Let) {
+    auto scope_inst = context.insts().Get(context.scope_stack().PeekInstId());
+    if (!scope_inst.Is<SemIR::InterfaceDecl>() &&
+        !scope_inst.Is<SemIR::FunctionDecl>()) {
+      CARBON_DIAGNOSTIC(LetCompileTimeBindingUnsupported, Error,
+                        "`let` compile time bindings require function or "
+                        "interface (may be supported later)");
+      context.emitter().Emit(node_id, LetCompileTimeBindingUnsupported);
```
zygoloid: Maybe this should be a `context.TODO` since the generics design does allow compile-time bindings in other scopes.
```suggestion
      context.TODO(node_id,
                   "`let` compile time binding outside function or interface");
```

toolchain/check/testdata/let/no_prelude/import.carbon
```diff
@@ -8,65 +8,77 @@
...
 
-// --- import_other.carbon
+// --- fail_import_other.carbon
 
 import Other;
 
+// CHECK:STDERR: fail_import_other.carbon:[[@LINE+3]]:5: error: `let` compile time bindings require function or interface (may be supported later)
+// CHECK:STDERR: let b:! () = Other.a;
+// CHECK:STDERR:     ^~~~~~
 let b:! () = Other.a;
```
zygoloid: Can we change these from `:!` to `:` bindings? It doesn't look like this test was trying to test compile time bindings in particular.

zygoloid (approved):

toolchain/check/testdata/let/no_prelude/import.carbon
jonmeow: Done

PR #4339, opened 2024-09-25
Add support for capturing console output to FileTest.

jonmeow:
One of the things that ClangRunnerTest is doing is capturing stderr/stdout because clang prints to it directly. This adds support for that to FileTest.

I'm renaming the current `capture_output` field to `dump_output` because the name is ambiguous after this change, and the flag is already named `--dump_output`. It's still not great, but at least it's more distinct.

Note ClangRunner still doesn't use the vfs; that still needs work. I'm just moving the NoArgs test over as a trivial test of the functionality.

### DIFF
--- a/testing/file_test/README.md
+++ b/testing/file_test/README.md
@@ -149,6 +149,19 @@ Supported comment markers are:
     is responsible for providing default arguments.
 
 -   ```
+    // SET-CAPTURE-CONSOLE-OUTPUT
+    ```
+
+    By default, stderr and stdout are expected to be piped through provided
+    streams. Adding this causes the test's own stderr and stdout to be captured
+    and added as well.
+
+    This should be avoided because weare partly ensuring that streams are an
+    API, but is helpful when wrapping Clang, where stderr is used directly.
+
+    SET-CAPTURE-CONSOLE-OUTPUT can be specified at most once.
+
+-   ```
     // SET-CHECK-SUBSET
     ```
 
--- a/testing/file_test/README.md
+++ b/testing/file_test/README.md
@@ -156,7 +156,7 @@ Supported comment markers are:
     streams. Adding this causes the test's own stderr and stdout to be captured
     and added as well.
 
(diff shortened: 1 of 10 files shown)


testing/file_test/README.md
```diff
@@ -148,6 +148,19 @@ Supported comment markers are:
...
 
+-   ```
+    // SET-CAPTURE-CONSOLE-OUTPUT
+    ```
+
+    By default, stderr and stdout are expected to be piped through provided
+    streams. Adding this causes the test's own stderr and stdout to be captured
+    and added as well.
+
+    This should be avoided because weare partly ensuring that streams are an
```
geoffromer:
```suggestion
    This should be avoided because we are partly ensuring that streams are an
```

testing/file_test/file_test_base.cpp
```diff
@@ -316,16 +324,36 @@ auto FileTestBase::ProcessTestFileAndRun(TestContext& context)
   llvm::PrettyStackTraceProgram stack_trace_entry(
       test_argv_for_stack_trace.size() - 1, test_argv_for_stack_trace.data());
 
+  // Conditionally capture console output. We use a scope exit to ensure the
+  // captures terminate even on run failures.
+  std::unique_ptr<std::unique_lock<std::mutex>> console_lock;
```
geoffromer: Would it work to use `unique_lock` directly, without the `unique_ptr` wrapper? `unique_lock` is default-constructible and move-assignable, which seems like all you need here.

geoffromer (approved):

testing/file_test/README.md
jonmeow: Done

testing/file_test/file_test_base.cpp
jonmeow: Done, missed that.

geoffromer (commented):
LGTM, but it looks like a test is failing.

PR #4340, opened 2024-09-25
Push a generic region when handling a `where` expression.

zygoloid:
This fixes a crash in the case where a `where` expression appears outside of any generic.


zygoloid:
I don't know whether this is the right long-term approach, but #4336 is currently hitting this after merging with trunk and this change unblocks it.

josh11b (approved):

PR #4341, opened 2024-09-25
Refactor InstallPaths API and comments a little.

jonmeow:
Stemming from #4331, trying to break apart InstallPaths class comments into three parts:

- Construction semantics, staying in the class comment
  - Trying to refer to methods with more detailed  documentation.
- Install prefix contents, now on `prefix_`
- Install structure, consolidating on `install_dirs`

For code refactoring, `driver()` and `prefix()` were only used by the install paths test. Rather than having a comment not to use `prefix()`, this instead extracts it out to a TestPeer model (which we have elsewhere with `TypedNodesTestPeer`, thus my choice in approaches).

### DIFF
--- a/toolchain/install/install_paths.h
+++ b/toolchain/install/install_paths.h
@@ -14,48 +14,23 @@ namespace Carbon {
 
 // Locates the toolchain installation and provides paths to various components.
 //
-// The Carbon toolchain expects to be installed into some install prefix. For
-// example, this is expected to be similar to the CMake install prefix:
+// The Carbon toolchain expects to be installed into some install prefix; see
+// `prefix_` for details. When locating an install, we verify it with
+// `CheckMarkerFile`. When errors occur, `SetError` makes `error()`
+// available for diagnostics and clears the install prefix (leaving things
+// minimally functional).
 //
-// - `C:/Program Files/Carbon` or similar on Windows.
-// - `/usr` or `/usr/local` on Linux and most BSDs.
-// - `/opt/homebrew` or similar on macOS with Homebrew.
-// - `bazel-bin/some/bazel/target.runfiles/_main/toolchain/install/prefix_root`
-//   for unit tests and just-built binaries during development.
+// The factory methods locate the install prefix based on their use-case:
 //
-// See https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html
-// for more details. While we don't build the toolchain with CMake, we expect
-// our installation to behave in a similar and compatible way.
-//
-// There are multiple ways of locating an install's prefix:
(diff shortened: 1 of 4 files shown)


toolchain/install/install_paths.h
```diff
@@ -103,35 +78,51 @@ class InstallPaths {
...
 
+  // The computed installation prefix. This will be an absolute path. We keep an
+  // absolute path for when the command line uses a relative path
+  // (`./bin/carbon`) and the working directory changes after initialization
+  // (for example, to Bazel's working directory). In the event of an error, this
+  // will be the empty string.
+  //
+  // When run from bazel (for example, in unit tests or development binaries)
+  // this will look like:
+  // `bazel-bin/some/bazel/target.runfiles/_main/toolchain/install/prefix_root`
```
josh11b, line 110: The `prefix_root` at the end looks a little suspicious to me. Is that right?

josh11b (approved):

toolchain/install/install_paths_test.cpp
```diff
@@ -16,6 +16,14 @@
 #include "tools/cpp/runfiles/runfiles.h"
 
 namespace Carbon {
+
+class InstallPathsTestPeer {
```
chandlerc, line 20: (not for this PR, maybe for a follow-up)

Is "test peer" a well established term? I couldn't find it when searching, and it was new to me. And the term "peer" is specific enough that I somewhat expect it to have a fairly specific meaning.

If its not that established, I'm still perfectly happy with the pattern itself to allow testing -- but maybe `...TestHelper` as a name to be a little more generic? But as I said, not this PR, a follow-up to maybe rename both if folks like that.

chandlerc (approved):
LG, nice cleanup!

toolchain/install/install_paths.h
chandlerc: We use an explicit `prefix_root` for all the things installed in the package to keep them distinct from the code and other things in the `install` package for managing an installation.

toolchain/install/install_paths_test.cpp
jonmeow: `Peer` specifically comes from https://abseil.io/tips/135

jonmeow: Note, left to my own devices, I'd just make InstallPathsTest itself the friend... (I would mainly expect peers/helpers for something we use in multiple places, not just once)

jonmeow: (I'm merging, expecting we'll come up with some unified solution for InstallPathsTestPeer, TypedNodesTestPeer, and InstTestHelper)

PR #4342, opened 2024-09-25
Don't substitute into the targeted instructions of an associated constant.

zygoloid:
When an instruction makes an absolute reference to another instruction, such as when `assoc_const` refers to the declaration of the associated constant in an interface, substitution into that instruction should not substitute into the referenced instruction.

Mark the corresponding `InstId` fields in the typed instructions as being absolute by giving them a distinct ID type that `Subst` doesn't substitute into. This formation of unnecessarily complicated SemIR that could in some cases lead to a CHECK failure when printing formatted SemIR because the same instruction ends up in multiple scopes.

### DIFF
--- a/toolchain/sem_ir/ids.h
+++ b/toolchain/sem_ir/ids.h
@@ -94,7 +94,8 @@ constexpr InstId InstId::Invalid = InstId(InvalidIndex);
 class AbsoluteInstId : public InstId {
  public:
   // Implicitly converts from the base class.
-  explicit(false) constexpr AbsoluteInstId(InstId inst_id) : InstId(inst_id) {}
+  // NOLINTNEXTLINE(google-explicit-constructor)
+  constexpr AbsoluteInstId(InstId inst_id) : InstId(inst_id) {}
 
   using InstId::InstId;
 };
--- a/toolchain/sem_ir/ids.h
+++ b/toolchain/sem_ir/ids.h
@@ -86,6 +86,19 @@ constexpr InstId InstId::Invalid = InstId(InvalidIndex);
       InstId::ForBuiltin(BuiltinInstKind::Name);
 #include "toolchain/sem_ir/builtin_inst_kind.def"
 
+// An ID of an instruction that is referenced absolutely within a typed
+// instruction. This means that the instruction always represents the ID of a
+// global entity that is independent of the current context, and operations like
+// substitution into the enclosing instruction should not substitute into fields
+// with this type.
+class AbsoluteInstId : public InstId {
+ public:
+  // Implicitly converts from the base class.
+  explicit(false) constexpr AbsoluteInstId(InstId inst_id) : InstId(inst_id) {}
+
+  using InstId::InstId;
+};
+
 // The package namespace will be the instruction after builtins.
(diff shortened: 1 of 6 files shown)


zygoloid:
I've split this PR into three commits: the first adds the test and shows the old behavior (with the CHECK disabled), the second fixes the bug, and the third updates the test to show new new behavior (and re-enables the CHECK). This should hopefully give a better idea of what this PR is fixing.

zygoloid:
#4336 is blocked on this after merge because the check added in #4333 (landed in #4221) starts firing on one of the #4336 tests due to this bug.

toolchain/sem_ir/ids.h
```diff
@@ -86,6 +86,20 @@ constexpr InstId InstId::Invalid = InstId(InvalidIndex);
       InstId::ForBuiltin(BuiltinInstKind::Name);
 #include "toolchain/sem_ir/builtin_inst_kind.def"
 
+// An ID of an instruction that is referenced absolutely within a typed
```
geoffromer: The phrasing of this comment makes it sound kind of like "absoluteness" is a property of the inst itself, but based on the code, I gather it's actually a property of the ID. In other words, in principle you could have an `AbsoluteInstId` and an `InstId` that refer to the same inst. Is that right?

toolchain/sem_ir/ids.h
```diff
@@ -86,6 +86,20 @@ constexpr InstId InstId::Invalid = InstId(InvalidIndex);
...
+// instruction. This means that the instruction always represents the ID of a
+// global entity that is independent of the current context, and operations like
+// substitution into the enclosing instruction should not substitute into fields
+// with this type.
+class AbsoluteInstId : public InstId {
```
geoffromer, line 98: This means we inherit the base class `Print`, but might we want the difference to be visible in the printed output?

toolchain/sem_ir/ids.h
```diff
@@ -86,6 +86,20 @@ constexpr InstId InstId::Invalid = InstId(InvalidIndex);
...
+ public:
+  // Implicitly converts from the base class.
+  // NOLINTNEXTLINE(google-explicit-constructor)
+  constexpr AbsoluteInstId(InstId inst_id) : InstId(inst_id) {}
```
geoffromer, line 103: The implicit conversion makes me a little nervous, especially since the conversion in the other direction is implicit as well. What would we lose if it was explicit?

toolchain/sem_ir/ids.h
zygoloid: I don't think we do; we're trying to attach an extra bit of information onto the typed inst field, not change how the field behaves.

toolchain/sem_ir/ids.h
zygoloid: It'd just introduce noise into all the places that construct typed insts, explicitly performing a conversion they don't care about.

toolchain/sem_ir/ids.h
zygoloid: Yes, that's right. I've updated the comment to try to make this clearer.

geoffromer (approved):

PR #4346, opened 2024-09-26
Fix TryEvalInstInContext to be static

jonmeow:
Also noticed `MakeIntTypeResult` should be static.


zygoloid (approved):
Looks good. I thought we had `-Wmissing-prototypes` enabled to catch things like this?

PR #4348, opened 2024-09-26
Update instructions for adding a SemIR instruction.

josh11b:


toolchain/docs/adding_features.md
```diff
@@ -261,13 +268,38 @@ If the resulting SemIR needs a new instruction:
     };
     ```
 
-Adding an instruction will also require a handler in the Lower step.
+    -   [`sem_ir/inst_kind.h`](/toolchain/sem_ir/inst_kind.h) documents the
+        different options when defining a new instruction, as well as their
+        defaults, see `InstKind::DefinitionInfo`.
+    -   If an instruction doesn't have a `type_id`, it won't produce a value,
+        and so won't be given a label when formatted, and can't be the argument
```
jonmeow: "won't be given a label when formatted"

What do you mean by "label"? My best guess would be an instruction name, but my understanding was that instructions always have names (even just %.loc).

toolchain/docs/adding_features.md
```diff
@@ -261,13 +268,38 @@ If the resulting SemIR needs a new instruction:
...
+        of another instruction.
```
jonmeow:
```suggestion
        and so won't be given a label when formatted, and can't be used for an expression operand.
```

Instruction arguments don't broadly need to be typed, only when there's an expression at play. Maybe this should be more specific?

toolchain/docs/adding_features.md
```diff
@@ -261,13 +268,38 @@ If the resulting SemIR needs a new instruction:
...
+    -   If an instruction always produces a type, so it sets
+        `.is_type = InstIsType::Always`, then its `type_id` field will be set to
+        `SemIR::TypeId::TypeType`.
```
jonmeow: `.is_type` indicates whether the `type_id` is set to `TypeType` (`Never`/`Maybe`/`Always`), but it doesn't cause the value of `type_id` to be set; instead, it mainly adds safeguards and changes some lowering logic.

What's the nuance you're trying to get at here? Above there's the reference to inst_kind.h, which has more detail.

toolchain/docs/adding_features.md
```diff
@@ -261,13 +268,38 @@ If the resulting SemIR needs a new instruction:
...
+    -   If an instruction produces a value, but not one that can be used in an
+        expression, you can make a new builtin type for its output. For example,
```
jonmeow: I thought namespaces had types because they could show up in expressions as part of name lookup, but never have values. Also, aren't bound methods and witnesses used as part of expressions?

To be sure, I would expect most people shouldn't be adding builtin inst kinds, as noted by the comments there. They're definitely an edge case.

toolchain/docs/adding_features.md
```diff
@@ -232,25 +233,31 @@ Note: this is broadly similar to
...
-            // Optional: , TerminatorKind::KindOfTerminator
+        static constexpr auto Kind =
+            // `Parse::SomeId` should be some type from `parse/node_ids.h`,
+            // specifying the kind of parse nodes for this instruction.
+            // - `Parse::NodeId` if it can be any kind of parse node.
+            // - `Parse::InvalidNodeId` if no associated parse node.
+            InstKind::NewInstKindName.Define<Parse::SomeId>(
+               // the name used in textual IR
+              {.ir_name = "new_inst_kind_name"}
+               // Other parameters have defaults
```
jonmeow:
```suggestion
                // The name used in textual IR:
                {.ir_name = "new_inst_kind_name"}
                // Other parameters have defaults.
```

Since you're editing this, maybe we should fix punctuation... but also, why the uneven indent?

toolchain/docs/adding_features.md
```diff
@@ -232,25 +233,31 @@ Note: this is broadly similar to
...
     struct NewInstKindName {
-        static constexpr auto Kind = InstKind::NewInstKindName.Define(
-            // the name used in textual IR
-            "new_inst_kind_name"
-            // Optional: , TerminatorKind::KindOfTerminator
+        static constexpr auto Kind =
+            // `Parse::SomeId` should be some type from `parse/node_ids.h`,
+            // specifying the kind of parse nodes for this instruction.
+            // - `Parse::NodeId` if it can be any kind of parse node.
+            // - `Parse::InvalidNodeId` if no associated parse node.
```
jonmeow:
```suggestion
            // `Parse::SomeId` should be one of:
            // - A node ID from `parse/node_ids.h`,
            //   specifying the kind of parse nodes for this instruction.
            // - `Parse::NodeId` if it can be any kind of parse node.
            // - `Parse::InvalidNodeId` if no associated parse node.
```

Note partly suggesting to change this to read as a list of alternatives, but also to say "node ID" instead of "type", particularly since "type" has a different meaning on an Inst.

But also, if someone is just joining and looks at `parse/node_ids.h`, I'm concerned they may not be able to figure out what to put here -- although that technically defines the node IDs, they aren't actually listed there. Maybe something like "The new kind added to `inst_kind.def` above, suffixed by `Id`" might be easier for someone in the common case.

toolchain/docs/adding_features.md
```diff
@@ -290,6 +290,11 @@ updated. The updates needed, can depend on whether the instruction produces a
 type. Look to the comments on those functions for instructions on what is
 needed.
 
+If the instruction owns references other instructions, add a case to
```
jonmeow: What does "owns references" mean? Do you mean if an instruction has an InstBlockId?

jonmeow (commented):
I see you're still making changes to this PR. I assume you won't be done with it soon, I'll note I go on vacation tomorrow so you might want to find another reviewer to finish.

toolchain/docs/adding_features.md
josh11b: I switched to say name. Instructions without type_ids do get a name, but if it is used I think it indicates that there is an error.

josh11b:
@zygoloid Could you help me figure out the right changes to make to the text to address these comments?

toolchain/docs/adding_features.md
josh11b: Updated the text to add more detail.

toolchain/docs/adding_features.md
josh11b: These are references to parse node kinds not instruction node kinds, but I think I get your point and added some text.

toolchain/docs/adding_features.md
josh11b: I've deleted this bullet and updated the comment before `type_id` above, and added a different bullet saying types are not required to be the argument to an instruction.

toolchain/docs/adding_features.md
josh11b: I agree it isn't common, but
a) if it does come up, you'd be happy to know how to do it
b) existing instructions you might read about use this feature at times
c) it isn't clear what criteria is being used to decide when to make a new builtin type instead of an instruction without a type id
d) I personally would like the deeper understanding that includes edge cases rather than just recipes for common cases.

toolchain/docs/adding_features.md
josh11b: I've deleted this bullet and updated the comment before `type_id` above.

toolchain/docs/adding_features.md
josh11b: I've revised the text to hopefully make my point more clearly.

toolchain/docs/adding_features.md
```diff
@@ -232,42 +233,108 @@ Note: this is broadly similar to
...
-Adding an instruction will also require a handler in the Lower step.
+    -   [`sem_ir/inst_kind.h`](/toolchain/sem_ir/inst_kind.h) documents the
+        different options when defining a new instruction, as well as their
+        defaults, see `InstKind::DefinitionInfo`.
+    -   If an instruction always produces a type:
+
+        -   set `.is_type = InstIsType::Always` in its `Kind` definition;
+        -   when constructing instructions of this kind, pass
+            `SemIR::TypeId::TypeType` in as the value of the `type_id` field, as
+            in:
```
jonmeow: When I look at https://developers.google.com/style/lists#types-of-lists, I see examples like:

> Here's a sequence of steps to follow:
> 
> 1. Open the box.
> 2. Remove the bobcat from the box.
> 3. Feed the bobcat.

To reflect that structure here, I suggest:

```suggestion
        -   Set `.is_type = InstIsType::Always` in its `Kind` definition.
        -   When constructing instructions of this kind, pass
            `SemIR::TypeId::TypeType` in as the value of the `type_id` field, as
            in:
```

Let me know if I'm missing something in the style guide, though.

toolchain/docs/adding_features.md
```diff
@@ -232,42 +233,108 @@ Note: this is broadly similar to
...
+            ```
+            SemIR::InstId inst_id = context.AddInst<SemIR::NewInstKindName>(
+                node_id, {.type_id = SemIR::TypeId::TypeType, ...});
+            ```
+
+    -   If an instruction produces a value used in an expression, but not one
+        that has an ordinary type, you can make a new builtin type for its
+        output. This is rare, but used, for example, when an expression
+        implicitly uses a value as part of SemIR evaluation or as part of
+        desugaring. We have builtin types for bound methods, namespaces,
```
jonmeow:
```suggestion
    -   Although most instructions have distinct types represented by instructions like `ClassType`, we also have builtin types for cases where types don't need to be distinct per-entity. This is rare, but used, for example, when an expression
        implicitly uses a value as part of SemIR evaluation or as part of
        desugaring. We have builtin types for bound methods, namespaces,
```

Do also bear in mind that `type` is also a builtin type.

toolchain/docs/adding_features.md
```diff
@@ -232,42 +233,108 @@ Note: this is broadly similar to
...
+type. Look to the comments on those functions for instructions on what is
+needed.
+
+Instructions won't be given a name unless
+[`InstNamer::CollectNamesInBlock](/toolchain/sem_ir/inst_namer.cpp) is called on
+the `InstBlockId` they are a member of. As of this writing,
+`InstNamer::CollectNamesInBlock` should only be called once per `InstBlockId`.
+To accomplish this, there should be one instruction kind that "owns" the
+instruction block, and will have a case in `InstNamer::CollectNamesInBlock` that
+visits the `InstBlockId` recursively. That instruction kind will typically use
```
jonmeow:
```suggestion
visits the `InstBlockId`. That instruction kind will typically use
```

Recursion is invariably an oversight. That should be fixed; added a note to toolchain tasks.

toolchain/docs/adding_features.md
```diff
@@ -232,42 +233,106 @@ Note: this is broadly similar to
...
+
+            ```
+            SemIR::InstId inst_id = context.AddInst<SemIR::NewInstKindName>(
+                node_id, {.type_id = SemIR::TypeId::TypeType, ...});
+            ```
+
+    -   Although most instructions have distinct types represented by instructions like `ClassType`, we also have builtin types for cases where types don't need to be distinct per-entity. This is rare, but used, for example, when an expression
+        implicitly uses a value as part of SemIR evaluation or as part of
+        desugaring. We have builtin types for bound methods, namespaces,
+        witnesses, among others. These are defined in
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
    -   Although most instructions have distinct types represented by
        instructions like `ClassType`, we also have builtin types for cases
        where types don't need to be distinct per-entity. This is rare, but
        used, for example, when an expression implicitly uses a value as part of
        SemIR evaluation or as part of desugaring. We have builtin types for
        bound methods, namespaces, witnesses, among others. These are defined in
```

josh11b (commented):
I have applied your suggestions

jonmeow (approved):
Looks good.

PR #4350, opened 2024-09-26
Reject abstract function definitions

dwblaikie:
Not sure about error recovery options - can/should we drop the definition as a means of recovery when building the SemIR? I guess probably not, so I guess this change is about right.

Phrasing of the error message I'm certainly open to.


toolchain/check/handle_function.cpp
```diff
@@ -440,6 +440,12 @@ auto HandleParseNode(Context& context, Parse::FunctionDefinitionId node_id)
   auto& function = context.functions().Get(function_id);
   FinishGenericDefinition(context, function.generic_id);
 
+  if (function.virtual_modifier == SemIR::Function::VirtualModifier::Abstract) {
+    CARBON_DIAGNOSTIC(DefinedAbstractFunction, Error,
+                      "`abstract` function with definition");
+    context.emitter().Emit(TokenOnly(node_id), DefinedAbstractFunction);
```
zygoloid: This will associate the diagnostic with the `}`; I think it might be a bit nicer to check this in `BuildFunctionDecl` instead and point the diagnostic at the `{` (which is the token corresponding to the `node_id` there).

zygoloid (commented):
Minor thought about the diagnostic location, but otherwise LGTM

toolchain/check/handle_function.cpp
dwblaikie: Oh, sure - I did realize that an out of line definition might have different issues (since the attribute wouldn't be on the out of line definition, but on the original declaration) so tested that and did have some trouble.

I implemented a possible fix for this - but wouldn't mind a second set of eyes. Is that the right way to go about it? Looking up the `SemIR::Function` based on the `FunctionDecl`'s `function_id` - I guess that finds some kind of canonical `Function` that seems to have the right modifiers, at least, so the diagnosis works.

(I had tried moving the logic into `CheckMethodModifiersOnFunction` but that wasn't triggered for the out of line definition case - though did produce a somewhat/maybe nicer diagnostic that highlighted the `abstract` keyword (and maybe I could've figured out a way to have it also note the `{` I think - since it looks like the infrastructure there has some way to pass an extra location) - but probably not worth figuring out how to also highlight the `abstract` keyword in the working implementation at the moment?)

toolchain/check/handle_function.cpp
```diff
@@ -297,6 +297,16 @@ static auto BuildFunctionDecl(Context& context,
   // Write the function ID into the FunctionDecl.
   context.ReplaceInstBeforeConstantUse(decl_id, function_decl);
 
+  // Diagnose `abstract` function with definition using the canonical Function's
+  // modifiers.
+  if (is_definition &&
+      context.functions().Get(function_decl.function_id).virtual_modifier ==
+          SemIR::Function::VirtualModifier::Abstract) {
+    CARBON_DIAGNOSTIC(DefinedAbstractFunction, Error,
+                      "`abstract` function with definition");
```
zygoloid: I'd find this slightly more readable as "definition of `abstract` function" but no strong preference. (I think the "with" phrasing is closer to what we use elsewhere.)

zygoloid (approved):
LG, thanks! Feel free to merge once you're happy with the commit message.

PR #4351, opened 2024-09-27
Use more compact storage for impl lookup buckets.

zygoloid:
If the bucket is of size zero or one, which is expected to be the common
case, then store it directly. For the remaining cases, use a side table
of lookup buckets.

### DIFF
--- a/toolchain/sem_ir/impl.h
+++ b/toolchain/sem_ir/impl.h
@@ -60,9 +60,10 @@ class ImplStore {
  private:
   // An ID of either a single impl or a lookup bucket.
   class ImplOrLookupBucketId : public IdBase {
-   public:
-    using IdBase::IdBase;
+   private:
+    explicit constexpr ImplOrLookupBucketId(int index) : IdBase(index) {}
 
+   public:
     // An explicitly invalid ID, corresponding to to ImplId::Invalid.
     static const ImplOrLookupBucketId Invalid;
 
--- a/toolchain/sem_ir/impl.h
+++ b/toolchain/sem_ir/impl.h
@@ -189,6 +189,7 @@ class ImplStore {
  private:
   ValueStore<ImplId> values_;
   Map<std::pair<TypeId, TypeId>, ImplOrLookupBucketId> lookup_;
+  // Buckets with at least 2 entries, which will be rare; see LookupBucketRef.
   llvm::SmallVector<llvm::SmallVector<ImplId, 2>> lookup_buckets_;
 };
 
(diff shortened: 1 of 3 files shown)


zygoloid:
Depends on #4336. Only review the final commit here.

toolchain/sem_ir/impl.h
```diff
@@ -57,22 +57,110 @@ struct Impl : public EntityWithParamsBase,
 // A collection of `Impl`s, which can be accessed by the self type and
 // constraint implemented.
 class ImplStore {
+ private:
+  // An ID of either a single impl or a lookup bucket.
+  class ImplOrLookupBucketId : public IdBase {
+   public:
+    using IdBase::IdBase;
```
jonmeow: Had you considered making the constructor private in order to emphasize use of the factory functions?

toolchain/sem_ir/impl.h
```diff
@@ -57,22 +57,110 @@ struct Impl : public EntityWithParamsBase,
 // A collection of `Impl`s, which can be accessed by the self type and
 // constraint implemented.
 class ImplStore {
+ private:
+  // An ID of either a single impl or a lookup bucket.
+  class ImplOrLookupBucketId : public IdBase {
```
jonmeow, line 62: Abstractly, we have a couple other things which have TinyPtrVector-like TODOs. Particularly NameScope's extended_scopes had seemed similar, but this approach won't work for it (I think it'd need to own the vector). I think this approach is fine, but I wanted to note this because I don't think this approach would work directly in other use-cases; I'm not sure precisely where the trade-offs stand to try to get something generally applicable.

toolchain/sem_ir/impl.h
```diff
@@ -100,9 +188,13 @@ class ImplStore {
 
  private:
   ValueStore<ImplId> values_;
-  Map<std::pair<TypeId, TypeId>, LookupBucket> lookup_;
+  Map<std::pair<TypeId, TypeId>, ImplOrLookupBucketId> lookup_;
+  llvm::SmallVector<llvm::SmallVector<ImplId, 2>> lookup_buckets_;
```
jonmeow, line 194:
```suggestion
  // Buckets with at least 2 entries, which will be rare; see LookupBucketRef.
  llvm::SmallVector<llvm::SmallVector<ImplId, 2>> lookup_buckets_;
```

jonmeow (approved):
Nice! Leaving it to you to figure out merging.

toolchain/sem_ir/impl.h

PR #4352, opened 2024-09-27
Make ImplDecls evaluate to themselves.

zygoloid:
Change ImplDecls so they evaluate to themselves in general, rather than
having a special case in import handling that pretends that they do.


zygoloid:
Depends on #4336; only review the last commit in this PR.

jonmeow (approved):
Thanks! Approving and leaving it to you to figure out merges.

PR #4353, opened 2024-09-27
Fix invocation issues with clang runner

jonmeow:
`ToolContext` should be explicitly initialized.

`-c` can still require a valid, writable `-o` path.


toolchain/driver/clang_runner.cpp
```diff
@@ -135,7 +135,10 @@ auto ClangRunner::Run(llvm::ArrayRef<llvm::StringRef> args) -> bool {
   // busybox of LLD as well, and having even the subprocesses consistently run
   // the Carbon install toolchain and not a system toolchain whenever possible.
   driver.CC1Main = [](llvm::SmallVectorImpl<const char*>& argv) -> int {
-    llvm::ToolContext tool_context;
+    // TODO: Try to a better path for argv[0] (maybe in the LLVM install paths).
+    // This works for now.
```
chandlerc:
```suggestion
    // TODO: Try to use a better path for argv[0] (maybe in the LLVM install
    // paths). This works for now.
```

chandlerc (approved):
LGTM

PR #4355, opened 2024-10-01
When converting an expression to type `type`, retain the resulting instruction as well as the `TypeId`.

zygoloid:
The `TypeId` is lossy, as it represents only the canonical type, and not the specific computation that produced it.


chandlerc (approved):
Nice, LG! (just helping out with reviews)

PR #4358, opened 2024-10-01
Add comma between the arguments to the `bind_symbolic_name` instruction

josh11b:


chandlerc (approved):
LG

toolchain/check/testdata/alias/no_prelude/fail_params.carbon
```diff
@@ -29,7 +29,7 @@ alias A(T:! type) = T*;
 // CHECK:STDOUT:     .A = %A
 // CHECK:STDOUT:   }
 // CHECK:STDOUT:   %T.param: type = param T, runtime_param<invalid>
-// CHECK:STDOUT:   %T: type = bind_symbolic_name T 0, %T.param [symbolic = constants.%T]
+// CHECK:STDOUT:   %T: type = bind_symbolic_name T, 0, %T.param [symbolic = constants.%T]
```
josh11b, line 32: One hesitation here is that there are actually two optional parameters. The prior formatting allowed those to be distinguished. After this change, they can be distinguished by whether they are an integer or not, but we could potentially do something to make it more explicit.

chandlerc: No strong opinion about doing additional formatting tweaks, but this already seems like a strict improvement. The absence of a comma seems too subtle to really be a helpful way to represent the two optional aspect you describe.

PR #4363, opened 2024-10-03
Fix our default `fastbuild` Bazel config

chandlerc:
Causes the default development (`fastbuild`) Bazel build and test to both use ASan, minimal optimizations, and produce good backtraces with source locations.

This also fixes all of the non-host configs that were deeply broken for the latest releases of Bazel going back quite some time.

The intent of our our Bazel toolchain config was for a minimally optimized, minimal debug info, and ASan + UBSan configuration to be the `fastbuild`, or the default development build of the project. This matches its inclusion of asserts, etc.

At some point quite some time ago, all of this stopped working. Bazel no longer has a `nonhost` feature. This was disabled a long time ago, briefly argued to be re-enabled, but has persistently been removed. However, since then the host and non-host features have been separated including the compilation mode, and so none of that is needed now. Instead, we can use a much simpler and more principled approach to all of the feature configuration now which this PR implements.

However, we added TCMalloc _after_ all of the ASan stuff became broken, and so we never saw that it is fundamentally incompatible with ASan. So this PR also reworks how TCMalloc is used to only apply to `-c opt` builds on Linux, and it also explicitly disables it when using `--config=asan`. I've not found a convenient way to tie the malloc library choice to a toolchain feature in Bazel, so this relies on the config being used rather than the feature in isolation.

Last but not least, with this change `fastbuild` creates substantially larger output and so this change also passes several new flags to reduce the size costs. One is a general improvement from outlining ASan instrumentation. The others are in a special feature as they reduce the error message quality for two of the more expensive UBSan checks in favor of small generated code size. Keeping these last two separate allows disabling this locally if needed to debug a failure.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -86,10 +86,15 @@ build --strip=never
 # Enable Abseil for GoogleTest.
 build --define=absl=1
 
+# Enable TCMalloc on Linux in optimized builds.
+build --custom_malloc=//bazel/malloc:tcmalloc_if_linux_opt
+
 # Configuration for enabling Address Sanitizer. Note that this is enabled by
 # default for fastbuild. The config is provided to enable ASan even in
-# optimized or other build configurations.
+# optimized or other build configurations. Note that ASan and TCMalloc are
+# incompatible so this explicitly forces the system malloc.
 build:asan --features=asan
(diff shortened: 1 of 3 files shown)


chandlerc:
FYI, this should be ready to go! =]

josh11b (approved):

PR #4364, opened 2024-10-03
`where` check stage, step 3: some type checking

josh11b:
With this, we now check:

* The left argument to `where` is a facet type
* The right argument of a rewrite (`=`) requirement converts to the type of the left argument.
* The left argument of an `impls` requirement is a type and the right argument is a facet type.

No checking is done for `==` constraints yet.

In addition, make the "is facet type" query into its own function and fix some comments noticed as part of this change.

This change reveals that accessing the members of a facet, like `.Self`, isn't doing the right thing, and will have to be fixed in a follow-on PR. Some tests have been adjusted or disabled as a result.


toolchain/check/testdata/impl/fail_todo_impl_assoc_const.carbon
```diff
@@ -10,27 +10,51 @@
 
 interface I { let T:! type; }
 
-// CHECK:STDERR: fail_todo_impl_assoc_const.carbon:[[@LINE+3]]:1: error: semantics TODO: `impl of interface with associated constant`
+// CHECK:STDERR: fail_todo_impl_assoc_const.carbon:[[@LINE+10]]:1: error: semantics TODO: `impl of interface with associated constant`
 // CHECK:STDERR: impl bool as I where .T = bool {}
 // CHECK:STDERR: ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+// CHECK:STDERR:
+// CHECK:STDERR: fail_todo_impl_assoc_const.carbon:[[@LINE+6]]:27: error: cannot implicitly convert from `type` to `<associated type in I>`
```
geoffromer, line 17: Would it be straightforward to have the TODO diagnostic above suppress these diagnostics, e.g. by checking for the `BuiltinError` witness that's produced when the TODO diagnostic is emitted?

toolchain/check/testdata/where_expr/constraints.carbon
```diff
@@ -19,23 +19,101 @@ interface I {
...
+
+// --- fail_todo_equal_constraint.carbon
+
+library "[[@TEST_NAME]]";
+
+interface N {
+  let P:! type;
+}
+
+// CHECK:STDERR: fail_todo_equal_constraint.carbon:[[@LINE+7]]:27: error: cannot implicitly convert from `{}` to `<associated type in N>`
```
geoffromer, line 36: I assume this is intended to eventually compile without errors. Would it be straightforward to get a TODO diagnostic here in the meantime? It's not a big deal, but if someone stumbles across this error during development, letting them know this is just a missing feature could save them a wild goose chase.

toolchain/check/testdata/where_expr/constraints.carbon
```diff
@@ -19,23 +19,101 @@ interface I {
...
+fn AssociatedTypeImpls(W:! K where .Associated impls M);
+
+// --- fail_todo_check_rewrite_constraints.carbon
 
 library "[[@TEST_NAME]]";
 
 import library "state_constraints";
 
 // TODO: Should fail since `2` can't be converted to the type of `I.Member`,
 // but constraints are not yet type checked.
```
geoffromer: Has this TODO been resolved? If so, should we take `todo` out of the filename?

toolchain/check/testdata/where_expr/constraints.carbon
```diff
@@ -19,23 +19,101 @@ interface I {
...
+fn RewriteTypeMismatch(X:! I where .Member = 2);
+
+// --- fail_left_of_impls_non_type.carbon
+
+library "[[@TEST_NAME]]";
+
+// CHECK:STDERR: fail_left_of_impls_non_type.carbon:[[@LINE+7]]:32: error: cannot implicitly convert from `i32` to `type`
+// CHECK:STDERR: fn NonTypeImpls(U:! type where 7 impls type);
+// CHECK:STDERR:                                ^
+// CHECK:STDERR: fail_left_of_impls_non_type.carbon:[[@LINE+4]]:32: note: type `i32` does not implement interface `ImplicitAs`
```
geoffromer, line 90: This may be out of scope, but should this say `ImplicitAs(type)`?

toolchain/check/testdata/where_expr/designator.carbon
```diff
@@ -4,9 +4,9 @@
 //
 // AUTOUPDATE
 // TIP: To test this file alone, run:
-// TIP:   bazel test //toolchain/testing:file_test --test_arg=--file_tests=toolchain/check/testdata/where_expr/no_prelude/designator.carbon
+// TIP:   bazel test //toolchain/testing:file_test --test_arg=--file_tests=toolchain/check/testdata/where_expr/designator.carbon
```
geoffromer, line 7: Why move this out of `no_prelude`? This seems like it would still work without the prelude.

josh11b: The added code does an implicit conversion to type `type`, which involves the `ImplicitAs` interface defined in the prelude.

toolchain/check/testdata/where_expr/constraints.carbon
josh11b: Well, it now tries to do the type checking, but the bug in the member access machinery means it doesn't get the right result, so there is definitely still some "todo" left. I've removed `todo` from the filename, and updated the comment to clarify what still remains.

toolchain/check/testdata/where_expr/constraints.carbon
josh11b: Yes to both it being out of scope and should say `ImplicitAs(type)`.

toolchain/check/testdata/where_expr/constraints.carbon
josh11b: Unfortunately, it is not straightforward for me. There is a bug in the implementation of member access for facet types. @zygoloid and I started looking at how to fix it, but it turned out to be a bigger project. I think we are going to need to do some more foundational work in the representation of facet values and facet types first. I am not aware of an easy way to detect this situation at the moment. You can see how far we got in https://github.com/carbon-language/carbon-lang/pull/4371.

toolchain/check/testdata/where_expr/constraints.carbon
geoffromer: I don't see those changes.

josh11b: Sorry, I fogot to push them. They are there now.

geoffromer:
Looks like you missed one comment.

toolchain/check/testdata/impl/fail_todo_impl_assoc_const.carbon
josh11b: I don't think so, since I think they are actually generated in the opposite order.

geoffromer (approved):

geoffromer:
(Not merging due to conflict)

PR #4366, opened 2024-10-03
Change how to get info for a parameter

josh11b:
Updates `SemIR::Function::GetParamFromParamRefId` to return more information in the form of a new `ParamInfo` struct. This struct has a method for getting the `NameId` from the name binding instruction. The callers previously got it from the `Param` instruction, but the plan is for that instruction to no longer be associated with a name.


toolchain/sem_ir/function.h
```diff
@@ -65,10 +65,18 @@ struct Function : public EntityWithParamsBase,
...
 
   // Given a parameter reference instruction from `param_refs_id` or
-  // `implicit_param_refs_id`, returns the corresponding `Param` instruction
-  // and its ID.
+  // `implicit_param_refs_id`, returns a `ParamInfo` value with the
+  // corresponding instruction, its ID, and the name binding, if present.
+  struct ParamInfo {
+    InstId inst_id;
+    std::optional<AnyBindName> bind_name;
+    Param inst;
```
geoffromer: It seems cleaner and less surprising to have `inst_id` right next to `inst`.

```suggestion
    Param inst;
    std::optional<AnyBindName> bind_name;
```

toolchain/check/convert.cpp
```diff
@@ -1207,18 +1207,18 @@ auto ConvertCallArgs(Context& context, SemIR::LocId call_loc_id,
   for (auto implicit_param_id : implicit_param_refs) {
     auto addr_pattern =
         context.insts().TryGetAs<SemIR::AddrPattern>(implicit_param_id);
-    auto [param_id, param] = SemIR::Function::GetParamFromParamRefId(
-        context.sem_ir(), implicit_param_id);
-    if (param.name_id == SemIR::NameId::SelfValue) {
+    auto param = SemIR::Function::GetParamFromParamRefId(context.sem_ir(),
```
geoffromer: Here and elsewhere, it seems like this should be named `param_info` and/or its type should be explicit; otherwise it's too easy to mistakenly assume the type is `Param`.

josh11b: Done.

geoffromer (approved):

PR #4367, opened 2024-10-03
Remove redundant optional wrapping llvm::function_ref

dwblaikie:
llvm::function_ref (like std::unique_ptr, for instance) already has a null/empty state, so use that to avoid confusion/duplication of empty states between optional and the nested function_refs.

### DIFF
--- a/toolchain/check/context.h
+++ b/toolchain/check/context.h
@@ -301,9 +301,8 @@ class Context {
   // If the type is not complete, `diagnoser` is invoked to diagnose the issue,
   // if a `diagnoser` is provided. The builder it returns will be annotated to
   // describe the reason why the type is not complete.
-  auto TryToCompleteType(
-      SemIR::TypeId type_id,
-      std::optional<BuildDiagnosticFn> diagnoser = std::nullopt) -> bool;
+  auto TryToCompleteType(SemIR::TypeId type_id,
+                         BuildDiagnosticFn diagnoser = nullptr) -> bool;
 
   // Attempts to complete and define the type `type_id`. Returns `true` if the
   // type is defined, or `false` if no definition is available. A defined type
@@ -311,9 +310,8 @@ class Context {
   //
   // This is the same as `TryToCompleteType` except for interfaces, which are
   // complete before they are fully defined.
-  auto TryToDefineType(
-      SemIR::TypeId type_id,
-      std::optional<BuildDiagnosticFn> diagnoser = std::nullopt) -> bool;
+  auto TryToDefineType(SemIR::TypeId type_id,
+                       BuildDiagnosticFn diagnoser = nullptr) -> bool;
 
   // Returns the type `type_id` as a complete type, or produces an incomplete
   // type error and returns an error type. This is a convenience wrapper around
--- a/toolchain/check/context.h
+++ b/toolchain/check/context.h
@@ -316,8 +316,8 @@ class Context {
   // Returns the type `type_id` as a complete type, or produces an incomplete
   // type error and returns an error type. This is a convenience wrapper around
   // TryToCompleteType.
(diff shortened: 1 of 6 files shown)


zygoloid (commented):
Yeah, having two different null states here seems confusing, and we may as well use the existing efficient null state.

That said: we have uses of `BuildDiagnosticFn` where the diagnoser is not optional. It looks like there's only one such case at the moment -- `Context::AsCompleteType` -- but I expect we'll end up with roughly as many mandatory uses of this type as optional ones. I think we need *some* way to keep track of which ones are mandatory and which ones are optional. Any ideas? Maybe two different typedefs for the same type?

dwblaikie:
> we have uses of `BuildDiagnosticFn` where the diagnoser is not optional. It looks like there's only one such case at the moment -- `Context::AsCompleteType` -- but I expect we'll end up with roughly as many mandatory uses of this type as optional ones. I think we need _some_ way to keep track of which ones are mandatory and which ones are optional. Any ideas? Maybe two different typedefs for the same type?

hrm - fair question. :/ 

Yeah, distinct typedef's probably a start. I do like type checking or at least assertion-y things, though, so possibly some NonNull wrapper type? Equally this is C++ and we probably have lots of other invariants that aren't checked or enshrined in the type system (which std::unique_ptr uses have a non-null guarantee? which Insts are known-valid (it was another one I was thinking of cleaning up, from looking around at optional usage, seemed there were a few `optional<Inst>` and I wondered if they could be invalid `Inst` instead? (or do they represent a genuine second state where valid, invalid, and nullopt are all possible/communicate different things))

zygoloid (commented):
As a convention, how about:

- If a parameter has a default argument, it's clearly optional, so no additional documentation is necessary.
- Otherwise, add a comment (or where it makes sense, a default argument) for optional function_ref parameters.
- In the remaining cases, the default assumption is that the argument is mandatory.

Does that work?

dwblaikie:
> As a convention, how about:
> 
> * If a parameter has a default argument, it's clearly optional, so no additional documentation is necessary.
> * Otherwise, add a comment (or where it makes sense, a default argument) for optional function_ref parameters.
> * In the remaining cases, the default assumption is that the argument is mandatory.
> 
> Does that work?

Sure - the net effect, I think, is that everything fits into the "has a default argument" case. One case you mentioned in person could use a default argument for one of its two callers, so I made that change. The only other case of missing a default argument was `AsCompleteType` which no caller passes a null diagnoser, but it's documented as a wrapper around `TryToCompleteType` which does have a default argument - so for consistency I added the default argument to `AsCompleteType` too.

toolchain/check/context.h
```diff
@@ -301,25 +301,23 @@ class Context {
...
+  auto TryToDefineType(SemIR::TypeId type_id,
+                       BuildDiagnosticFn diagnoser = nullptr) -> bool;
 
   // Returns the type `type_id` as a complete type, or produces an incomplete
   // type error and returns an error type. This is a convenience wrapper around
   // TryToCompleteType.
-  auto AsCompleteType(SemIR::TypeId type_id, BuildDiagnosticFn diagnoser)
-      -> SemIR::TypeId {
+  auto AsCompleteType(SemIR::TypeId type_id,
+                      BuildDiagnosticFn diagnoser = nullptr) -> SemIR::TypeId {
```
zygoloid: This should not have a default argument -- this diagnoser is intentionally not optional. If it were omitted, we'd return `Error` here without actually emitting an error diagnostic, which would be a bug.

dwblaikie: Ah, fair. Thanks for the catch/explanation. Fixed.

dwblaikie: (oops, forgot to push, done that now... )

zygoloid (approved):

PR #4368, opened 2024-10-03
Track the instruction used to name the type and constraint in an impl.

zygoloid:
This is necessary in order to have access to the specific versions of their constant values in a generic impl.

Stub out impl deduction.

### DIFF
--- a/toolchain/check/deduce.cpp
+++ b/toolchain/check/deduce.cpp
@@ -10,6 +10,7 @@
 #include "toolchain/check/generic.h"
 #include "toolchain/check/subst.h"
 #include "toolchain/sem_ir/ids.h"
+#include "toolchain/sem_ir/impl.h"
 #include "toolchain/sem_ir/typed_insts.h"
 
 namespace Carbon::Check {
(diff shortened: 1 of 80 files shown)


toolchain/check/deduce.cpp
```diff
@@ -238,4 +239,17 @@ auto DeduceGenericCallArguments(
...
 }
 
+// Deduces the impl arguments to use in a use of a parameterized impl. Returns
+// `Invalid` if deduction fails.
+auto DeduceImplArguments(Context& context, const SemIR::Impl& impl,
+                         SemIR::ConstantId self_id,
+                         SemIR::ConstantId constraint_id) -> SemIR::SpecificId {
+  CARBON_CHECK(impl.generic_id.is_valid(),
+               "Performing deduction for non-generic impl");
+  static_cast<void>(context);
```
josh11b, line 250: Perhaps a TODO here to indicate this is just a stub?

toolchain/check/handle_impl.cpp
```diff
@@ -329,13 +338,14 @@ auto HandleParseNode(Context& context, Parse::ImplDefinitionStartId node_id)
...
+                      "redefinition of `impl {0} as {1}`", std::string,
+                      std::string);
     CARBON_DIAGNOSTIC(ImplPreviousDefinition, Note,
                       "previous definition was here");
     context.emitter()
-        .Build(node_id, ImplRedefinition, impl_info.self_id,
-               impl_info.constraint_id)
+        .Build(node_id, ImplRedefinition,
+               context.sem_ir().StringifyTypeExpr(impl_info.self_id),
+               context.sem_ir().StringifyTypeExpr(impl_info.constraint_id))
```
josh11b, line 351: Could you add a comment here explaining why you need to go through StringifyTypeExpr, and maybe a TODO for things to be done better?

josh11b (commented):
Can we make a test showing that we now report the correct source location for a type where we couldn't before?

(Probably not in this PR.) Some of the new references are to instructions with bare 'loc' names. It would help readability if we could get descriptive instruction names in more cases.

toolchain/check/import_ref.cpp
```diff
@@ -70,6 +71,25 @@ auto AddImportRef(Context& context, SemIR::ImportIRInst import_ir_inst,
   return import_ref_id;
 }
 
+static auto AddLoadedImportRef(Context& context,
```
josh11b, line 78: Please add a comment saying why to use this function instead of e.g. `AddImportRef`, or how it is different

josh11b (approved):

PR #4370, opened 2024-10-04
Expose indexing as a language interface

brymer-meneses:
This PR makes it so that types can implement the `IndexWith` interface so that they can provide their custom indexing behavior.


brymer-meneses:
I think this should be ready for review. The bulk of the test happens at `toolchain/check/testdata/operators/overloaded/index.carbon`, most of the testdata changes are just importing `Core//prelude/operators/index`.

toolchain/check/handle_index.cpp
```diff
@@ -16,17 +21,93 @@ auto HandleParseNode(Context& /*context*/, Parse::IndexExprStartId /*node_id*/)
...
+    CARBON_DIAGNOSTIC(TypeNotIndexable, Error,
+                      "type `{0}` does not support indexing", SemIR::TypeId);
+    context.emitter().Emit(node_id, TypeNotIndexable, operand_type_id);
+    return SemIR::InstId::BuiltinError;
+  }
+
+  CARBON_CHECK(args.size() == 2,
+               "IndexWith should have two generic constraints");
+
+  auto op = Operator{
```
jonmeow:
```suggestion
  Operator op{
```

There was a debate about [initialization style](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/project/cpp_style_guide.md)... I think structs are a case for "Braced initialization without `=`"

toolchain/check/handle_index.cpp
```diff
@@ -16,17 +21,93 @@ auto HandleParseNode(Context& /*context*/, Parse::IndexExprStartId /*node_id*/)
...
+// the `SubscriptType` and the `ElementType`, if the class does not implement
+// the said interface this returns an empty array reference.
+static auto GetIndexWithArgs(Context& context, Parse::NodeId node_id,
+                             SemIR::TypeId self_id)
+    -> llvm::ArrayRef<SemIR::InstId> {
+  auto index_with_interface =
+      context.types().GetAs<SemIR::GenericInterfaceType>(
+          context.insts()
+              .GetAs<SemIR::StructValue>(
+                  context.LookupNameInCore(node_id, "IndexWith"))
```
jonmeow: This should probably verify that it finds a well-formed `IndexWith`. I think there are some crashes if it doesn't.

e.g., you can test this with a test in a `no_prelude` directory:

```
// --- wrong_index_with.carbon

library "[[@TEST_NAME]]";

namespace Core;
class Core.IndexWith {}
fn F() { 0[1]; }

// --- missing_index_with.carbon

library "[[@TEST_NAME]]";

fn F() { 0[1]; }
```

As a fix, you can do something like:

```
auto index_with_inst = context.insts().TryGetAsIfValid<SemIR::StructValue>(context.LookupNameInCore(node_id, "IndexWith"));
if (!index_with_inst) {
  return {};
}
auto index_with_interface =
    context.types().TryGetAs<SemIR::GenericInterfaceType>(index_with_inst->type_id);
if (!index_with_interface) {
  return {};
}
```

toolchain/check/handle_index.cpp
```diff
@@ -16,17 +21,93 @@ auto HandleParseNode(Context& /*context*/, Parse::IndexExprStartId /*node_id*/)
   return true;
 }
 
+// Returns the argument values of the `IndexWith` interface, this corresponds to
+// the `SubscriptType` and the `ElementType`, if the class does not implement
+// the said interface this returns an empty array reference.
```
jonmeow:
```suggestion
// Returns the argument values of the `IndexWith` interface. Arguments correspond to
// the `SubscriptType` and the `ElementType`. If the class does not implement
// the said interface, this returns an empty array reference.
```

toolchain/check/handle_index.cpp
```diff
@@ -16,17 +21,93 @@ auto HandleParseNode(Context& /*context*/, Parse::IndexExprStartId /*node_id*/)
...
+static auto GetIndexWithArgs(Context& context, Parse::NodeId node_id,
```
jonmeow, line 31:
```suggestion
// TODO: Switch to using an associated type instead of a parameter for the `ElementType`.
static auto GetIndexWithArgs(Context& context, Parse::NodeId node_id,
```

We should capture the need to switch to an associated type somewhere, maybe here because it affects how it'll be called?

toolchain/check/handle_index.cpp
```diff
@@ -16,17 +21,93 @@ auto HandleParseNode(Context& /*context*/, Parse::IndexExprStartId /*node_id*/)
...
+        context.specifics().Get(interface_type->specific_id).args_id);
+  }
+
+  return {};
+}
+
+// Creates SemIR to perform an index with base expression `operand_inst_id` and
+// `operand_type_id` for types that are not an array. This entails, checking if
+// the base expression implements the `IndexWith` interface, if so uses the `At`
+// associative method while printing a diagnostic if it isn't.
```
jonmeow:
```suggestion
// Performs an index with base expression `operand_inst_id` and
// `operand_type_id` for types that are not an array. This checks if
// the base expression implements the `IndexWith` interface; if so, uses the `At`
// associative method, otherwise prints a diagnostic.
```

toolchain/check/handle_index.cpp
```diff
@@ -16,17 +21,93 @@ auto HandleParseNode(Context& /*context*/, Parse::IndexExprStartId /*node_id*/)
...
+  // an error.
+  if (args.empty()) {
+    CARBON_DIAGNOSTIC(TypeNotIndexable, Error,
+                      "type `{0}` does not support indexing", SemIR::TypeId);
+    context.emitter().Emit(node_id, TypeNotIndexable, operand_type_id);
+    return SemIR::InstId::BuiltinError;
+  }
+
+  CARBON_CHECK(args.size() == 2,
+               "IndexWith should have two generic constraints");
```
jonmeow: Similar to the above, can you add a test like:

```
// --- wrong_arg_count.carbon

library "[[@TEST_NAME]]";

namespace Core;
interface Core.IndexWith(SubscriptType:! type) {
  fn At[self: Self](subscript: SubscriptType) -> ();
}

impl () as Core.IndexWith(()) {
  fn At[self: Self](subscript: ()) -> () {
    return ();
  }
}

fn F() { ()[()]; }
```

It's sometimes important to avoid using CHECK when Carbon code could then cause a compiler crash.

jonmeow (commented):
Thanks! Most of my comments are for crashes on bad inputs. Checking with zygoloid I think we're also getting close to having associated type support, but since it's not ready yet, this seems good for short-term support -- we should just have some TODOs.

toolchain/check/handle_index.cpp
brymer-meneses: shouldn't this be allowed?

toolchain/check/handle_index.cpp
brymer-meneses: this should be a fail right?

toolchain/check/handle_index.cpp
brymer-meneses: Is a diagnostic more appropriate here then?

jonmeow: Yes... You can compare with similar for the other built-in interfaces, I was surprised this wasn't catching the same validation but I think it's due to how you're doing the lookup.

jonmeow:
Thanks for the changes! It looks though like there are still test failures on toolchain/check/testdata/class/fail_redeclaration_introducer.carbon -- where did that file come from? It appears to be unrelated to the indexing change, should it be removed?

brymer-meneses:
I'm also not sure, perhaps i messed up syncing to trunk again--sorry about that!

jonmeow (approved):
Thanks!

brymer-meneses:
Not really sure why are the tests failing :<<

brymer-meneses:
My bad, forgot to add `fail_` on tests that are expected to fail

PR #4373, opened 2024-10-04
User Snippets and file icon for carbon and microsoft C/C++ configurations

BhJaipal:
Adds User Snippets and file icon for VSCode and C/C++ configuration for microsoft C/C++ intellisence


utils/vscode/carbon.code-snippets
```diff
@@ -0,0 +1,55 @@
+{
+  "package": {
+    "scope": "carbon",
+    "prefix": "package",
+    "body": "package $fileNameWithoutExt",
```
geoffromer: The filename might be a good default for the library name (although I'm not sure about that), but it's definitely not a good default for the package name. The rule of thumb is that a package [should be roughly the same size as a git repository](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/code_and_name_organization/README.md#sizing-packages-and-libraries), so it should be extremely rare for different files in the same directory to have different packages.

utils/vscode/carbon.code-snippets
```diff
@@ -0,0 +1,55 @@
...
+  },
+  "Main Run function": {
+    "scope": "carbon",
+    "prefix": "Run",
+    "body": ["fn Run() -> i32 {", "\t$1", "\treturn 0;", "}"]
+  },
+  "New Function": {
+    "scope": "carbon",
+    "prefix": "fn",
+    "body": ["fn $1($2) ->  $3 {", "\t$4", "}"]
```
geoffromer:
```suggestion
    "body": ["fn $1($2) -> $3 {", "\t$4", "}"]
```

geoffromer: This snippet seems like it could be kind of inconvenient if the user doesn't want the function to return anything (which will probably be pretty common).

utils/vscode/carbon.code-snippets
```diff
@@ -0,0 +1,55 @@
...
+  },
+  "Else if statement": {
+    "scope": "carbon",
+    "prefix": "elif",
+    "body": ["else if ($1) {", "\t$2", "}"]
+  },
+  "For each loop with for": {
+    "scope": "carbon",
+    "prefix": "foreach",
+    "body": ["for (var $1: $2 in $3) {", "\t$4", "}"]
```
geoffromer: This probably shouldn't have `var`. It isn't necessary unless you want to mutate the variable in the loop body, which probably isn't the common case.

utils/vscode/carbon.code-snippets
```diff
@@ -0,0 +1,55 @@
...
+  "Class declaration": {
+    "scope": "carbon",
+    "prefix": "class",
+    "body": [
+      "class $1 {",
+      "\tvar $2;",
+      "\tfn Make() -> Self {",
+      "\t\t$3",
+      "\t}",
+      "}"
```
geoffromer: Class bodies don't seem like a good fit for snippets, because they don't have a fixed structure. In particular, a class body doesn't necessarily consist of a public data member followed by a `Make` function. In fact, that particular pattern seems likely to be fairly uncommon, because `Make` functions are most useful when the data members are private.

utils/vscode/favicon.png
geoffromer, line 1: We're intentionally [avoiding having a Carbon logo](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/project/faq.md#why-isnt-there-a-carbon-language-logo) for now, so we probably shouldn't have a favicon.

utils/vscode/package.json
```diff
@@ -3,21 +3,27 @@
   "displayName": "carbon-lang",
   "description": "Carbon Language support",
   "version": "0.0.2",
+  "publisher": "carbon-language",
```
geoffromer, line 6: If I understand correctly, this field is supposed to be a unique identifier that's registered with Visual Studio Code Marketplace, and belongs to the Azure DevOps organization publishing the extension. I don't think we're ready to set up that kind of account-management infrastructure, so for now I think we should continue to leave this field out, and not publish the extension in the marketplace.

utils/vscode/package.json
```diff
@@ -33,8 +39,19 @@
...
+      {
+        "language": "carbon",
+        "path": "./carbon.code-snippets"
+      }
     ]
   },
+  "keywords": [
+    "carbon",
+    "snippet",
+    "snippets"
```
geoffromer: We're apparently limited to 5 keywords, so I don't know if we should spend 2 of them on "snippet". That said, there's probably not much point in having this field at all until we publish the extension.

utils/vscode/carbon.code-snippets
BhJaipal: So, I should use $1 instead of fileNameWithoutExt?

utils/vscode/carbon.code-snippets
BhJaipal: > This snippet seems like it could be kind of inconvenient if the user doesn't want the function to return anything (which will probably be pretty common).

If they don't want, they can remove the line it but I think having a default return in snippet would be nice

utils/vscode/carbon.code-snippets
BhJaipal: Should I use constant instead? I couldn't actually find constant keyword
So I used it

utils/vscode/carbon.code-snippets
BhJaipal: Ok, I will probably remove it then

utils/vscode/package.json
BhJaipal: Ok, I will remove one

utils/vscode/package.json
BhJaipal: Without publisher we can't actually create vsix package, I will change it if you wish

utils/vscode/favicon.png
BhJaipal: Ok, I will remove it

BhJaipal:
@geoffromer Can you review this latest commit?

BhJaipal:
@jonmeow please review

utils/nvim/setup.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
geoffromer, line 1: This change doesn't seem related to user snippets. Am I missing something?

Same question about `utils/treesitter/.gitignore`.

utils/vscode/carbon.code-snippets
```diff
@@ -2,54 +2,67 @@
   "package": {
     "scope": "carbon",
     "prefix": "package",
-    "body": "package $fileNameWithoutExt",
+    "body": "impl library ${1:$fileNameWithoutExt}",
```
geoffromer, line 5: I think we should still have `package`, especially since that's the name and the prefix for the snippet; we just shouldn't provide a default. We could always have a separate `library` snippet for the case where the user doesn't want to specify a package.

It's probably also better to leave out `impl`, because every library has an API file but not every library has an `impl` file, so leaving it out will be the right guess slightly more often. It would be ideal if we could infer this from the filename (`impl` files should always have a `.impl.carbon` extension), but I don't know if that's feasible.

Finally, the library name should have quotes around it.

```suggestion
    "body": "package $1 library \"${2:$fileNameWithoutExt}\"",
```

utils/vscode/carbon.code-snippets
```diff
@@ -2,54 +2,67 @@
...
+      "\t$2",
+      "}"
+    ]
   },
   "For each loop with for": {
     "scope": "carbon",
     "prefix": "foreach",
-    "body": ["for (var $1: $2 in $3) {", "\t$4", "}"]
+    "body": [
+      "for (const $1: $2 in $3) {",
```
geoffromer, line 49: Bindings are constant by default, so we don't need any keyword here (and `const` isn't valid here):

```suggestion
      "for ($1: $2 in $3) {",
```

utils/vscode/package.json
```diff
@@ -14,16 +14,13 @@
     "Programming Languages",
     "Snippets"
   ],
+  "type": "commonjs",
```
geoffromer, line 17: I'm having trouble finding documentation for this field. What does it mean?

utils/vscode/package.json
```diff
@@ -33,16 +36,22 @@
...
   },
+  "keywords": [
+    "carbon",
+    "snippets"
+  ],
   "scripts": {
     "package": "mkdir -p out && vsce package -o out/carbon.vsix"
   },
-  "devDependencies": {
-    "@vscode/vsce": "^2.27.0"
```
geoffromer, line 42: Why drop this dependency? Are you sure it's not needed?

utils/vscode/dist/extension.js
geoffromer, line 1: This doesn't seem right. At least judging from [this doc](https://code.visualstudio.com/api/working-with-extensions/bundling-extension), the convention is that `src` contains the source code for the extension, and then the contents of `dist` are generated from the source code as part of bundling. As such, I'm not sure `dist/extension.js` should even be checked in to source control, but in any event we certainly shouldn't be removing `src/extension.js`.

utils/vscode/package-lock.json
geoffromer, line 1: Why remove this file?

utils/vscode/carbon.code-snippets
```diff
@@ -0,0 +1,68 @@
...
+  },
+  "Main Run function": {
+    "scope": "carbon",
+    "prefix": "Run",
+    "body": [
+      "fn Run() -> i32 {",
+      "\t$1",
+      "\treturn 0;",
+      "}"
+    ]
```
CarbonInfraBot, line 16: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
    "body": ["fn Run() -> i32 {", "\t$1", "\treturn 0;", "}"]
```

utils/vscode/carbon.code-snippets
```diff
@@ -0,0 +1,68 @@
...
+  },
+  "New Function": {
+    "scope": "carbon",
+    "prefix": "fn",
+    "body": [
+      "fn $1($2) {",
+      "\t$3",
+      "}"
+    ]
```
CarbonInfraBot, line 25: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
    "body": ["fn $1($2) {", "\t$3", "}"]
```

utils/vscode/carbon.code-snippets
```diff
@@ -0,0 +1,68 @@
...
+  },
+  "If statement": {
+    "scope": "carbon",
+    "prefix": "if",
+    "body": [
+      "if ($1) {",
+      "\t$2",
+      "}"
+    ]
```
CarbonInfraBot, line 34: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
    "body": ["if ($1) {", "\t$2", "}"]
```

utils/vscode/carbon.code-snippets
```diff
@@ -0,0 +1,68 @@
...
+  },
+  "Else if statement": {
+    "scope": "carbon",
+    "prefix": "elif",
+    "body": [
+      "else if ($1) {",
+      "\t$2",
+      "}"
+    ]
```
CarbonInfraBot, line 43: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
    "body": ["else if ($1) {", "\t$2", "}"]
```

utils/vscode/carbon.code-snippets
```diff
@@ -0,0 +1,68 @@
...
+  },
+  "For each loop with for": {
+    "scope": "carbon",
+    "prefix": "foreach",
+    "body": [
+      "for (const $1: $2 in $3) {",
+      "\t$4",
+      "}"
+    ]
```
CarbonInfraBot, line 52: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
    "body": ["for (const $1: $2 in $3) {", "\t$4", "}"]
```

utils/vscode/carbon.code-snippets
```diff
@@ -0,0 +1,68 @@
...
+  },
+  "Match case": {
+    "scope": "carbon",
+    "prefix": "match",
+    "body": [
+      "match ($1) {",
+      "\tcase ($2) => {$3}",
+      "\tdefault => {$4}"
+    ]
```
CarbonInfraBot, line 61: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
    "body": ["match ($1) {", "\tcase ($2) => {$3}", "\tdefault => {$4}"]
```

utils/vscode/carbon.code-snippets
```diff
@@ -0,0 +1,68 @@
...
+  },
+  "Extends Class": {
+    "scope": "carbon",
+    "prefix": "extend",
+    "body": "extend base: $1;"
+  }
+}
```
CarbonInfraBot, line 68: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
}
```

utils/nvim/setup.sh
BhJaipal: With `/bin/sh` operations like `set` doesn't work

utils/vscode/carbon.code-snippets
BhJaipal: Ok, I will do it

utils/vscode/package.json
BhJaipal: CommonJS:
 ```js
let http = require("http");
```

Module/ ESM:
```js
import http from "http";
```

utils/vscode/package.json
BhJaipal: I will fix in next commit

utils/vscode/dist/extension.js
BhJaipal: I actually think TypeScript would be much helpful and better for creating extension and later we can compile it to js and create vsix package

PR #4374, opened 2024-10-04
Improve infrastructure for formatting types in diagnostics.

zygoloid:
Instead of stringifying types in the caller in some cases, add new types to represent:

- `InstIdAsType`: an `InstId` diagnostic argument that represents a type expression that should be included in the diagnostic
- `InstIdAsTypeOfExpr`: an `InstId` diagnostic argument that represents an expression whose type should be included in the diagnostic

For these cases, we can produce more user-friendly descriptions of a type than we can with a canonicalized `TypeId`. Add comments to discourage using `TypeId` diagnostic arguments when one of the above can be used, and move over existing uses where it's straightforward to do so.

Move type stringification code to its own files and out of `SemIR::File` to make `File` smaller and to further discourage the direct use of the stringification logic.

Also update type printing to include the `` ` `` delimiters surrounding the type. The intent is that we will eventually want to include other information when formatting a type, like Clang does when printing a typedef (`'string' (aka 'std::basic_string<char>')`), and such formatting requires that the diagnostic machinery produces the `` ` ``s itself.

There are a couple of cases where we really want to format valid Carbon type syntax directly into a diagnostic, rather than an `aka` or similar, because the diagnostic text includes part of the type itself, for example: ``"consider using `partial {0}`"``. For such cases, a `Raw` form of the diagnostic argument types is added: `TypeIdAsRawType` and `InstIdAsRawType`. In principle we could instead use ``"consider using `partial {0:raw}`"``, but our diagnostic machinery isn't set up for that.


toolchain/sem_ir/stringify_type.h
```diff
@@ -0,0 +1,21 @@
...
+#define CARBON_TOOLCHAIN_SEM_IR_STRINGIFY_TYPE_H_
+
+#include "toolchain/sem_ir/file.h"
+#include "toolchain/sem_ir/ids.h"
+
+namespace Carbon::SemIR {
+
+// Produces a string version of a type expression. Generally, this should not be
+// called directly. To format a string into a diagnostic, use a diagnostic
+// parameter of type InstIdAsType or InstIdAsTypeOfExpr where possible, or of
```
josh11b:
```suggestion
// parameter of type `InstIdAsType`, `InstIdAsRawType`, or `InstIdAsTypeOfExpr` where possible, or of
```

toolchain/check/diagnostic_helpers.h
```diff
@@ -50,6 +50,58 @@ struct TypedInt {
   llvm::APInt value;
 };
 
+// A type value for rendering in a diagnostic without enclosing "`"s. See
+// InstIdAsRawType for details.
```
josh11b:
```suggestion
// `InstIdAsRawType` for details.
```

toolchain/check/diagnostic_helpers.h
```diff
@@ -50,6 +50,58 @@ struct TypedInt {
   llvm::APInt value;
 };
 
+// A type value for rendering in a diagnostic without enclosing "`"s. See
```
josh11b: Can you please add a reference to this file and/or these types to https://github.com/carbon-language/carbon-lang/blob/trunk/toolchain/docs/diagnostics.md#diagnostic-parameter-types?

toolchain/check/diagnostic_helpers.h
```diff
@@ -50,6 +50,58 @@ struct TypedInt {
...
+
+  SemIR::InstId inst_id;
+};
+
+// An expression whose type should be rendered in a diagnostic, including
+// enclosing "`"s. This should be used instead of TypeId as a diagnostic
+// argument wherever possible, because we should eventually be able to produce a
+// sugared type name in this case, whereas a TypeId will render as a canonical
+// type.
+struct InstIdAsTypeOfExpr {
```
josh11b: I feel like we should put the ones you will commonly use and/or are recommended first, and the ones most similar together. Maybe this order?
```
InstIdAsTypeOfExpr
InstIdAsType
InstIdAsRawType
TypeIdAsRawType
```
That way you could, for example, document what `Raw` means once.

josh11b: `InstIdAsTypeOfExpr` didn't convey to me what this did. Maybe `TypeOfInstId`?

toolchain/check/diagnostic_helpers.h
```diff
@@ -50,6 +50,58 @@ struct TypedInt {
...
+struct InstIdAsType {
+  using DiagnosticType = DiagnosticTypeInfo<std::string>;
+
+  // NOLINTNEXTLINE(google-explicit-constructor)
+  InstIdAsType(SemIR::InstId inst_id) : inst_id(inst_id) {}
+
+  SemIR::InstId inst_id;
+};
+
+// A type expression, for rendering in a diagnostic. Like InstIdAsType, but the
```
josh11b:
```suggestion
// A type expression, for rendering in a diagnostic. Like `InstIdAsType`, but the
```

toolchain/check/diagnostic_helpers.h
```diff
@@ -50,6 +50,58 @@ struct TypedInt {
...
+  // NOLINTNEXTLINE(google-explicit-constructor)
+  InstIdAsRawType(SemIR::InstId inst_id) : inst_id(inst_id) {}
+
+  SemIR::InstId inst_id;
+};
+
+// An expression whose type should be rendered in a diagnostic, including
+// enclosing "`"s. This should be used instead of TypeId as a diagnostic
+// argument wherever possible, because we should eventually be able to produce a
+// sugared type name in this case, whereas a TypeId will render as a canonical
```
josh11b:
```suggestion
// sugared type name in this case, whereas a `TypeId` will render as a canonical
```

toolchain/sem_ir/stringify_type.h
```diff
@@ -0,0 +1,21 @@
...
+// type TypeId if you don't have an expression describing the type.
```
josh11b:
```suggestion
// type `TypeId` or `TypeIdAsRawType` if you don't have an expression describing the type.
```

toolchain/check/diagnostic_helpers.h
```diff
@@ -50,6 +50,58 @@ struct TypedInt {
...
+  using DiagnosticType = DiagnosticTypeInfo<std::string>;
+
+  // NOLINTNEXTLINE(google-explicit-constructor)
+  InstIdAsRawType(SemIR::InstId inst_id) : inst_id(inst_id) {}
+
+  SemIR::InstId inst_id;
+};
+
+// An expression whose type should be rendered in a diagnostic, including
+// enclosing "`"s. This should be used instead of TypeId as a diagnostic
```
josh11b:
```suggestion
// enclosing "`"s. This should be used instead of `TypeId` as a diagnostic
```

toolchain/check/diagnostic_helpers.h
```diff
@@ -50,6 +50,58 @@ struct TypedInt {
...
+  InstIdAsType(SemIR::InstId inst_id) : inst_id(inst_id) {}
+
+  SemIR::InstId inst_id;
+};
+
+// A type expression, for rendering in a diagnostic. Like `InstIdAsType`, but the
+// type will not be enclosed in "`"s. Once we start including "aka" types for
+// sugared types, such annotations will also not be included for raw types. This
+// is intended for cases where the type is part of a larger syntactic construct
+// in a diagnostic, such as "redefinition of `impl {0} as {1}`".
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
// A type expression, for rendering in a diagnostic. Like `InstIdAsType`, but
// the type will not be enclosed in "`"s. Once we start including "aka" types
// for sugared types, such annotations will also not be included for raw types.
// This is intended for cases where the type is part of a larger syntactic
// construct in a diagnostic, such as "redefinition of `impl {0} as {1}`".
```

toolchain/sem_ir/stringify_type.h
```diff
@@ -0,0 +1,21 @@
...
+
+#include "toolchain/sem_ir/file.h"
+#include "toolchain/sem_ir/ids.h"
+
+namespace Carbon::SemIR {
+
+// Produces a string version of a type expression. Generally, this should not be
+// called directly. To format a string into a diagnostic, use a diagnostic
+// parameter of type `InstIdAsType`, `InstIdAsRawType`, or `InstIdAsTypeOfExpr` where possible, or of
+// type `TypeId` or `TypeIdAsRawType` if you don't have an expression describing the type.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
// parameter of type `InstIdAsType`, `InstIdAsRawType`, or `InstIdAsTypeOfExpr`
// where possible, or of type `TypeId` or `TypeIdAsRawType` if you don't have an
// expression describing the type.
```

toolchain/check/diagnostic_helpers.h
zygoloid: Done.

I suppose that also opens the door for things like `TypeOfInstIdAsRawType`, which we currently don't have an obvious name for (I think `InstIdAsRawTypeOfExpr` doesn't work).

zygoloid: Done.

toolchain/check/diagnostic_helpers.h
zygoloid: Done.

josh11b (approved):

josh11b:
Thanks! I really appreciate that new documentation in particular.

PR #4376, opened 2024-10-07
Switch to LLD on macOS.

chandlerc:
This gives us a consistent linker with Linux, and allows us to use `--start-lib`/`--end-lib` flags that remove the need to actually build archive files. The result of skipping archive files should significantly shrink the macOS build output as well as making it a bit faster.

I've checked and the other flags we use for linking on Linux are either unnecessary (clang-rt is already used for example) or incompatible (there is no static libc++ linking AFAICT on macOS).


chandlerc:
:sigh: never mind, LLVM-16 seems to have bugs in LLD linking on ARM macOS. This will have to wait until we can raise the minimum version of LLVM on macOS.

PR #4377, opened 2024-10-07
Add link to video for C++Now variadics talk

josh11b:


eyelash (commented):
I think you changed the wrong link there. You changed the one for CppNorth instead of the one for C++Now.

josh11b:
> I think you changed the wrong link there. You changed the one for CppNorth instead of the one for C++Now.

My mistake! Thank you for catching that!

chandlerc (approved):
LG, thanks both!

PR #4378, opened 2024-10-07
Disallow `abstract` or `base` on class declarations (that are not definitions)

dwblaikie:
Per: [p3762](https://docs.carbon-lang.dev/proposals/p3762.html#modifier-keywords:~:text=Other%20class%2C%20impl%2C%20and%20interface%20modifiers%20%28abstract%2C%20base%2C%20final%29%20exist%20only%20on%20the%20definition%2C%20not%20on%20the%20forward%20declaration): "Other class, impl, and interface modifiers (`abstract`, `base`, `final`) exist only on the definition, not on the forward declaration."


dwblaikie:
Hmm, in the process of testing this further, I came across existing test coverage that verified a different rule (that the base/abstract specifiers had to match between redeclarations) and linked to a still-open issue: #3384 

Looks like #3762 was merged in March 2024, #3384 was filed in Nov 2023, but @josh11b mentioned in https://github.com/carbon-language/carbon-lang/issues/3384#issuecomment-2123417694 that there was still ongoing discussion about how this should be addressed?

So... I provide this patch as a foray into discussing how this is meant to work/what bits of the spec/approved proposals are authoritative with respect to answering this question.

josh11b:
> Hmm, in the process of testing this further, I came across existing test coverage that verified a different rule (that the base/abstract specifiers had to match between redeclarations) and linked to a still-open issue: #3384
> 
> Looks like #3762 was merged in March 2024, #3384 was filed in Nov 2023, but @josh11b mentioned in [#3384 (comment)](https://github.com/carbon-language/carbon-lang/issues/3384#issuecomment-2123417694) that there was still ongoing discussion about how this should be addressed?
> 
> So... I provide this patch as a foray into discussing how this is meant to work/what bits of the spec/approved proposals are authoritative with respect to answering this question.

It looks like #3762 and [that discussion](https://docs.google.com/document/d/1s3mMCupmuSpWOFJGnvjoElcBIe2aoaysTIdyczvKX84/edit?pli=1&resourcekey=0-G095Wc3sR6pW1hLJbGgE0g&tab=t.0#heading=h.bu7djkos4xo) ultimately agree that for `abstract` and `base`, it should only be on the class definition, not other declarations. This also agrees with the last thing written in #3384 , but we should poke the leads to accept that resolution or say otherwise. But in general I would say a proposal accepted after an issue was posed takes precedence.

toolchain/check/testdata/class/fail_abstract.carbon
```diff
@@ -20,16 +20,22 @@ class Derived {
...
   // CHECK:STDERR:                   ^~~~~~~~
+  // CHECK:STDERR:
   return {.base = {.a = 1}, .d = 7};
 }
 
 fn Access(d: Derived) -> (i32, i32) {
   return (d.d, d.base.a);
 }
 
+// CHECK:STDERR: fail_abstract.carbon:[[@LINE+3]]:1: error: `abstract` not allowed on `class` declaration
```
josh11b: It would be nice (but not required) to have the error say this modifier should only be on the definition.

toolchain/check/testdata/class/fail_modifiers.carbon
```diff
@@ -8,6 +8,10 @@
 // TIP: To dump output, run:
 // TIP:   bazel run //toolchain/testing:file_test -- --dump_output --file_tests=toolchain/check/testdata/class/fail_modifiers.carbon
 
+// CHECK:STDERR: fail_modifiers.carbon:[[@LINE+11]]:9: error: `abstract` not allowed on `class` declaration
+// CHECK:STDERR: private abstract private class DuplicatePrivate;
```
josh11b: Feels better to have separate tests for "abstract not allowed on declaration" from "modifier repeated" and so on. Perhaps make these tests use a class definition (`{}` instead of `;`), now that those are different? (similarly other tests)

josh11b:
> > Hmm, in the process of testing this further, I came across existing test coverage that verified a different rule (that the base/abstract specifiers had to match between redeclarations) and linked to a still-open issue: #3384
> > Looks like #3762 was merged in March 2024, #3384 was filed in Nov 2023, but @josh11b mentioned in [#3384 (comment)](https://github.com/carbon-language/carbon-lang/issues/3384#issuecomment-2123417694) that there was still ongoing discussion about how this should be addressed?
> > So... I provide this patch as a foray into discussing how this is meant to work/what bits of the spec/approved proposals are authoritative with respect to answering this question.
> 
> It looks like #3762 and [that discussion](https://docs.google.com/document/d/1s3mMCupmuSpWOFJGnvjoElcBIe2aoaysTIdyczvKX84/edit?pli=1&resourcekey=0-G095Wc3sR6pW1hLJbGgE0g&tab=t.0#heading=h.bu7djkos4xo) ultimately agree that for `abstract` and `base`, it should only be on the class definition, not other declarations. This also agrees with the last thing written in #3384 , but we should poke the leads to accept that resolution or say otherwise. But in general I would say a proposal accepted after an issue was posed takes precedence.

Update: #3384 has now been resolved.

toolchain/check/testdata/class/fail_modifiers.carbon
dwblaikie: Agreed & have updated the test in that direction - though wasn't sure how much you were thinking we should go (like - revisiting existing untouched test cases into more distinct/isolated tests? Happy to do that, but haven't done it yet).

toolchain/check/testdata/class/fail_abstract.carbon
dwblaikie: Yeah, it crossed my mind when developing this, but the generic functionality for disallowed modifiers was rather convenient & adding this special case felt like it'd be pretty ugly.

I've implemented it/updated this patch with that change just for the exploration - maybe there's nicer ways to do it (with a callback?) or perhaps it feels worth the complexity/inelegance. Open to ideas/refactorings/revertings/renamings/etc.

toolchain/check/handle_class.cpp
```diff
@@ -181,10 +181,13 @@ static auto BuildClassDecl(Context& context, Parse::AnyClassDeclId node_id,
...
-                       KeywordModifierSet::Access | KeywordModifierSet::Extern |
-                           (is_definition ? KeywordModifierSet::Class
-                                          : KeywordModifierSet::None));
+  auto accepted_modifiers =
+      KeywordModifierSet::Access | KeywordModifierSet::Extern;
+  auto invalid_unless_definition_modifiers = KeywordModifierSet::None;
+  (is_definition ? accepted_modifiers : invalid_unless_definition_modifiers) |=
+      KeywordModifierSet::Class;
+  LimitModifiersOnDecl(context, introducer, accepted_modifiers,
+                       invalid_unless_definition_modifiers);
```
josh11b: I find this a bit hard to read. I wonder if it would be simpler to introduce another function for this case? Maybe something like:

```suggestion
  LimitModifiersOnDecl(context, introducer,
                       KeywordModifierSet::Access | KeywordModifierSet::Extern |
                       KeywordModifierSet::Class);
  if (!is_definition) {
    LimitModifiersNotDefinition(context, introducer,
                         KeywordModifierSet::Access | KeywordModifierSet::Extern);
  }
```

The only difference between the two would be the diagnostic issued, so they could mostly share implementation.

toolchain/check/testdata/class/fail_abstract.carbon
```diff
@@ -20,16 +20,22 @@ class Derived {
...
 }
 
 fn Access(d: Derived) -> (i32, i32) {
   return (d.d, d.base.a);
 }
 
+// CHECK:STDERR: fail_abstract.carbon:[[@LINE+3]]:1: error: `abstract` not allowed on `class` declaration, only definition
+// CHECK:STDERR: abstract class AbstractDecl;
+// CHECK:STDERR: ^~~~~~~~
+abstract class AbstractDecl;
```
josh11b: This should be in another file split. The ideal here is that each should-fail test is in its own split, so accidentally succeeding is considered an error. In this case, the previous test should really be a "fail_todo" test, that is expected to succeed when the feature is implemented.

toolchain/check/testdata/class/fail_base_bad_type.carbon
```diff
@@ -99,6 +99,10 @@ fn ConvertToBadBaseStruct(p: DeriveFromStruct*) -> {.a: i32, .b: i32}* { return
 // It would be OK to reject this if we start actually looking in the struct type.
 fn AccessMemberWithInvalidBaseStruct(p: DeriveFromStruct*) -> i32 { return (*p).n; }
 
+// CHECK:STDERR: fail_base_bad_type.carbon:[[@LINE+4]]:1: error: `base` not allowed on `class` declaration, only definition
+// CHECK:STDERR: base class Incomplete;
+// CHECK:STDERR: ^~~~
+// CHECK:STDERR:
```
josh11b, line 124: Again a case where the tests should be in separate splits. Many of these files predate us supporting splits. This is also a case where "`base` only allowed on definitions" should be separated from other tests about incomplete types.

toolchain/check/handle_class.cpp
dwblaikie: Thanks for the suggestion, yeah - that does simplify things greatly! (done)

toolchain/check/testdata/class/fail_base_bad_type.carbon
dwblaikie: Made some splits here too.

toolchain/check/testdata/class/fail_abstract.carbon
dwblaikie: Ah, thanks - sure, had a go at adding some splits.

josh11b (approved):
Thanks, particularly for cleaning up the tests you were touching!

PR #4379, opened 2024-10-07
Handle block scopes in compile-time binding check

geoffromer:
This fixes a fuzzer-found crash when a compile-time binding occurs inside a block (which is represented by an invalid inst ID).

### DIFF
--- a/toolchain/check/handle_binding_pattern.cpp
+++ b/toolchain/check/handle_binding_pattern.cpp
@@ -219,10 +219,20 @@ auto HandleParseNode(Context& context,
   bool is_generic = true;
   if (context.decl_introducer_state_stack().innermost().kind ==
       Lex::TokenKind::Let) {
-    if (context.GetCurrentScopeAs<SemIR::ClassDecl>() ||
-        context.GetCurrentScopeAs<SemIR::ImplDecl>()) {
-      context.TODO(node_id, "`let` compile time binding in class or impl");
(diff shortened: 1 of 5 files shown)


zygoloid (commented):
It's not captured in the review for #4338, but we intentionally chose to use a small list of allowed contexts here because other contexts are either not known to work yet or known to not work yet.

For example, given
```carbon
let T:! type = {};
fn F() {}
fn G() { F(); }
```
... we produce a bogus error on the call to `F()` saying we can't deduce a value for `T` (because we expect all generic parameters in scope in the declaration of `F` to have deduced arguments), and
```carbon
fn F(U:! type);
let T:! type = {};
fn F(U:! type) {}
fn G() { F({}); }
```
... crashes due to a mismatch between the two declarations of `F`.

So while I think this change gets us closer to matching the design, I don't think the rest of the toolchain is ready for us to go this far yet. Maybe for now we can check for and reject the case where `PeekInstId` returns `InstId::Invalid`? Or perhaps we can also reject `let` bindings in a namespace scope?

geoffromer:
Let's try accepting the case where the inst ID is invalid. That seems brittle, but hopefully it can tide us over until we can figure out how to properly support this kind of query.

zygoloid (approved):
I'm happy to expand this and see if the fuzzer manages to find ways to break it :)

PR #4380, opened 2024-10-08
Basic support for argument deduction in generic `impl`s.

zygoloid:
Refactor the current function call deduction logic to make it reusable. Call into it from `impl` deduction. Also build a generic region for the definition portion of a generic `impl` and substitute into it before accessing the witness in a specific `impl`.

This is enough to get simple uses of generic `impl`s to work. The main blocker for more complex cases is that we have very little support for non-trivial deduction, so while we can deduce `forall [T:! type] T as I`, we can't deduce `forall [T:! type] C as I(T)` yet.


toolchain/check/deduce.cpp
```diff
@@ -75,6 +75,60 @@ class DeductionWorklist {
...
+    // Initialize the deduced arguments to Invalid.
+    result_arg_ids_.resize(
+        context.inst_blocks()
+            .Get(context.generics().Get(generic_id_).bindings_id)
+            .size(),
+        SemIR::InstId::Invalid);
+  }
+
+  auto context() const -> Context& { return *context_; }
+  auto worklist() -> DeductionWorklist& { return worklist_; }
```
geoffromer: Might it make sense to remove this accessor and expose `Add` and `AddAll` methods instead? That seems like it would make the abstraction boundary much clearer: clients set up the work to be done, then call `Deduce` to do the work. As it stands, the API raises the possibility that clients might sometimes process items from the worklist themselves.

toolchain/check/deduce.cpp
```diff
@@ -75,6 +75,60 @@ class DeductionWorklist {
...
+
+  // Adds the arguments from the given specific as known enclosing generic
+  // arguments that will not be deduced.
+  auto AddEnclosingArgs(SemIR::SpecificId enclosing_specific_id) -> void;
```
geoffromer: Might it make sense to make `enclosing_specific_id` a constructor parameter, and make this an implementation detail of the constructor? That would structurally guarantee that this work is done before `Deduce` (which I believe is currently an undocumented requirement), and also make it less surprising that `enclosing_specific_id` is allowed to be invalid.

toolchain/check/deduce.cpp
```diff
@@ -75,6 +75,60 @@ class DeductionWorklist {
...
+  // argument list.
+  auto MakeSpecific() -> SemIR::SpecificId;
+
+ private:
+  Context* context_;
+  SemIR::LocId loc_id_;
+  SemIR::GenericId generic_id_;
+  bool diagnose_;
+
+  DeductionWorklist worklist_ = DeductionWorklist(*context_);
```
geoffromer: Could this go in the constructor init list instead?

toolchain/check/deduce.cpp
```diff
@@ -84,121 +138,111 @@ static auto NoteGenericHere(Context& context, SemIR::GenericId generic_id,
...
-    llvm::ArrayRef<SemIR::InstId> arg_ids) -> SemIR::SpecificId {
-  DeductionWorklist worklist(context);
-
-  llvm::SmallVector<SemIR::InstId> result_arg_ids;
-  llvm::SmallVector<Substitution> substitutions;
-
+auto DeductionContext::AddEnclosingArgs(SemIR::SpecificId enclosing_specific_id)
+    -> void {
+  if (!enclosing_specific_id.is_valid()) {
+    return;
```
geoffromer: It might make the logical structure a little clearer if we set `first_deduced_index_` to 0 here, rather than in the default member initializer; that way this method is unconditionally responsible for initializing `first_deduced_index_`.

toolchain/check/deduce.cpp
```diff
@@ -75,6 +75,60 @@ class DeductionWorklist {
...
+  // deduction succeeded.
+  auto Deduce() -> bool;
+
+  // Returns whether every generic parameter has a corresponding deduced generic
+  // argument. If not, issues a suitable diagnostic.
+  auto CheckDeductionIsComplete() -> bool;
+
+  // Forms a specific corresponding to the deduced generic with the deduced
+  // argument list.
+  auto MakeSpecific() -> SemIR::SpecificId;
```
geoffromer, line 115: I'm guessing this can't be called before `Deduce`? If so, it would probably be better to document that fact.

toolchain/check/deduce.cpp
zygoloid: Sure, done. I'd arranged it this way because `impl`s don't have an enclosing specific, but I suppose it's still meaningful to ask for one in all cases.

toolchain/check/deduce.cpp
zygoloid: If you prefer. Done.

toolchain/check/deduce.cpp
zygoloid: The type doesn't have a default constructor. But after inlining this into the constructor, it's at least all done in the same function now.

toolchain/check/deduce.cpp
zygoloid: Added comment.

toolchain/check/deduce.cpp
zygoloid: Changed to add forwarding functions. I've made these templates to handle the various overloads and to prepare for adding more overloads in my next PR.

geoffromer (approved):

PR #4381, opened 2024-10-08
Disallow creating instances of abstract classes

dwblaikie:
A good first-pass, at least. (abstract adapters are rejected with this change, though pending further language design discussion)


toolchain/check/context.h
```diff
@@ -333,11 +340,12 @@ class Context {
   // Returns the type `type_id` as a complete type, or produces an incomplete
   // type error and returns an error type. This is a convenience wrapper around
   // TryToCompleteType. `diagnoser` must not be null.
-  auto AsCompleteType(SemIR::TypeId type_id, BuildDiagnosticFn diagnoser)
+  auto AsCompleteType(SemIR::TypeId type_id, BuildDiagnosticFn diagnoser,
+                      BuildDiagnosticFn abstract_diagnoser)
```
zygoloid: Should we include a default argument for `abstract_diagnoser`? I think it's optional here, right?

toolchain/check/handle_binding_pattern.cpp
```diff
@@ -99,16 +99,28 @@ static auto HandleAnyBindingPattern(Context& context, Parse::NodeId node_id,
...
+                cast_type_id);
+          },
+          [&] {
+            CARBON_DIAGNOSTIC(AbstractTypeInVarDecl, Error,
+                              "{0} has abstract type `{1}`",
+                              llvm::StringLiteral, SemIR::TypeId);
+            return context.emitter().Build(
+                type_node, AbstractTypeInVarDecl,
+                parent_class_decl ? llvm::StringLiteral("Field")
+                                  : llvm::StringLiteral("Variable"),
```
zygoloid:
```suggestion
                parent_class_decl ? llvm::StringLiteral("field")
                                  : llvm::StringLiteral("variable"),
                cast_type_id);
          },
          [&] {
            CARBON_DIAGNOSTIC(AbstractTypeInVarDecl, Error,
                              "{0} has abstract type `{1}`",
                              llvm::StringLiteral, SemIR::TypeId);
            return context.emitter().Build(
                type_node, AbstractTypeInVarDecl,
                parent_class_decl ? llvm::StringLiteral("field")
                                  : llvm::StringLiteral("variable"),
```
Let's fix the case of these while we're here.

toolchain/check/handle_class.cpp
```diff
@@ -393,13 +393,22 @@ auto HandleParseNode(Context& context, Parse::AdaptDeclId node_id) -> bool {
...
+        return context.emitter().Build(node_id, IncompleteTypeInAdaptDecl,
+                                       adapted_type_id);
+      },
+      [&] {
+        CARBON_DIAGNOSTIC(AbstractTypeInAdaptDecl, Error,
+                          "adapted type `{0}` is an abstract type",
+                          SemIR::TypeId);
+        return context.emitter().Build(node_id, AbstractTypeInAdaptDecl,
+                                       adapted_type_id);
+      });
```
zygoloid: It's not clear to me whether this should be valid. Let's reject for now, but I've filed #4387 to consider this.

toolchain/check/testdata/class/fail_abstract.carbon
```diff
@@ -20,16 +70,59 @@ class Derived {
 
 fn Make() -> Derived {
   // TODO: This should be valid, and should construct an instance of `partial Abstract` as the base.
-  // CHECK:STDERR: fail_abstract.carbon:[[@LINE+3]]:19: error: cannot construct instance of abstract class; consider using `partial Abstract` instead
+  // CHECK:STDERR: fail_abstract.carbon:[[@LINE+7]]:10: error: initialization of abstract type `Abstract`
   // CHECK:STDERR:   return {.base = {.a = 1}, .d = 7};
-  // CHECK:STDERR:                   ^~~~~~~~
```
zygoloid: Is the old diagnostic here still reachable? If not, we should remove it.

toolchain/check/convert.cpp
```diff
@@ -937,24 +937,46 @@ auto Convert(Context& context, SemIR::LocId loc_id, SemIR::InstId expr_id,
...
+          [&] {
+            CARBON_DIAGNOSTIC(AbstractTypeInInit, Error,
+                              "initialization of abstract type `{0}`",
+                              SemIR::TypeId);
+            CARBON_DIAGNOSTIC(AbstractTypeInValueConversion, Error,
+                              "forming value of abstract type `{0}`",
+                              SemIR::TypeId);
+            CARBON_DIAGNOSTIC(AbstractTypeInConversion, Error,
+                              "invalid use of abstract type `{0}`",
+                              SemIR::TypeId);
```
zygoloid: I think we don't want to diagnose in these two cases, only when initializing.

toolchain/check/handle_binding_pattern.cpp
```diff
@@ -185,13 +197,22 @@ static auto HandleAnyBindingPattern(Context& context, Parse::NodeId node_id,
...
+            return context.emitter().Build(type_node, IncompleteTypeInLetDecl,
+                                           cast_type_id);
+          },
+          [&] {
+            CARBON_DIAGNOSTIC(AbstractTypeInLetDecl, Error,
+                              "`let` binding has abstract type `{0}`",
+                              SemIR::TypeId);
+            return context.emitter().Build(type_node, AbstractTypeInLetDecl,
+                                           cast_type_id);
+          });
```
zygoloid: I think this case should be permitted, as a `let` binding is forming a value not an object.

toolchain/check/handle_function.cpp
```diff
@@ -364,15 +364,26 @@ static auto HandleFunctionDefinitionAfterSignature(
...
+        },
+        [&] {
+          CARBON_DIAGNOSTIC(
+              AbstractTypeInFunctionParam, Error,
+              "parameter has abstract type `{0}` in function definition",
+              SemIR::TypeId);
+          return context.emitter().Build(param_info.inst_id,
+                                         AbstractTypeInFunctionParam,
+                                         param_info.inst.type_id);
+        });
```
zygoloid: I think this case should be permitted because a parameter binding is a value binding (by default, and we don't support the `var` modifier yet).

zygoloid (commented):
The intended model is that you cannot create objects of abstract class types, but you can have *values* of abstract class types -- the value representation in that case is a pointer to the instance, so it doesn't need the class to be non-abstract. So I think we would want to remove some of these checks; I'll point out which ones.

toolchain/check/context.h
dwblaikie: Yep! (allows updating one of the callers that was already passing nullptr)

toolchain/check/convert.cpp
dwblaikie: Done (this also included adding null `DiagnosticBuilder` - open to riffing on the API (currently just uses the default ctor to create a null DiagnosticBuilder, but could use a named ctor of some kind - and has explicit bool conversion which could be some more explicitly named function))

toolchain/check/handle_binding_pattern.cpp
dwblaikie: Done - thanks for the help understanding the language design here.

toolchain/check/handle_function.cpp
dwblaikie: Done

toolchain/check/testdata/class/fail_abstract.carbon
dwblaikie: Seems unreachable - removed.

dwblaikie (commented):
Handled feedback, and did a more major pass on the test coverage, splitting things (I feel like maybe I /should/ be writing those tests differently to, for example, share the definition of the abstract class rather than redefining it in even split - though it's so short importing it doesn't seem much more legible either - totally open to suggestions there).

Did manage to get test coverage on the return type check (at a call site to, rather than a definition of the abstract-returning function).

toolchain/diagnostics/diagnostic_emitter.h
```diff
@@ -55,6 +55,7 @@ class DiagnosticEmitter {
     DiagnosticBuilder(DiagnosticBuilder&&) noexcept = default;
     auto operator=(DiagnosticBuilder&&) noexcept
         -> DiagnosticBuilder& = default;
+    DiagnosticBuilder() : emitter_(nullptr) {}
```
zygoloid: A brief comment here would be handy.

toolchain/check/convert.cpp
```diff
@@ -963,19 +955,14 @@ auto Convert(Context& context, SemIR::LocId loc_id, SemIR::InstId expr_id,
...
-                : target.kind == ConversionTarget::Value
-                    ? AbstractTypeInValueConversion
-                    : AbstractTypeInConversion,
-                target.type_id);
+            if (!target.is_initializer()) {
+              return DiagnosticEmitter<SemIRLoc>::DiagnosticBuilder();
+            }
+            // return // add support for creating a null DiagnosticBuilder and
+            //  return one here. Include null-testability and use that
+            //  to skip adding notes in the caller if it's null.
```
zygoloid:
```suggestion
```
Looks like this has been addressed already.

toolchain/check/convert.cpp
zygoloid: I think a named constructor would be nice. Maybe even a method on `DiagnosticEmitter`, say `BuildSuppressed` or something, so you don't need to repeat the template argument.

zygoloid (approved):
Thanks, looks good.

PR #4382, opened 2024-10-08
Improve `-c dbg`, enabling Split DWARF and other enhancements

chandlerc:
This should substantially reduce the total build size of `-c dbg` builds, and especially improve cache hits during incremental development. I'm seeing over 50% reduction total on Linux in just the raw size of a complete debug build. Even on macOS where we can't use split DWARF there are substantial reductions.

Note that LLDB and GDB want slightly different flags to have the best experience with split debug information, and so the build and documentation have been updated to enable LLDB's flags by default but provide clear instructions for switching to GDB's flags, and they are structured so that this can be done persistently for an individual developer.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -116,8 +116,9 @@ build --enable_platform_specific_config
 build:linux --define=pfm=1
 
 # Enable split debug info on Linux, which is significantly more space efficient
-# and should work well with modern debuggers. Note that this is not currently
-# supported on macOS.
+# and should work well with modern debuggers. Note that this is Linux specific
+# as macOS has its own approach that is always partially but not completely
+# split.
 #
 # Note: if using GDB, see documentation to get that working:
 # https://docs.carbon-lang.dev/docs/project/contribution_tools.html#debugging-with-gdb-instead-of-lldb
--- a/.bazelrc
+++ b/.bazelrc
@@ -109,9 +109,19 @@ build --test_env=ASAN_SYMBOLIZER_PATH
 # `en_US`.
 build --action_env=LANG=en_US.UTF-8
 
-# Enable libpfm for google_benchmark on Linux only.
+# Allow per-platform configuration.
 build --enable_platform_specific_config
+
+# Enable libpfm for google_benchmark on Linux only.
 build:linux --define=pfm=1
+
+# Enable split debug info on Linux, which is significantly more space efficient
+# and should work well with modern debuggers. Note that this is not currently
+# supported on macOS.
+#
+# Note: if using GDB, see documentation to get that working:
+# https://docs.carbon-lang.dev/docs/project/contribution_tools.html#debugging-with-gdb-instead-of-lldb
+build:linux --fission=yes
 
 # Disables `actions.declare_symlink`. Done for cross-environment support.
 build --allow_unresolved_symlinks=false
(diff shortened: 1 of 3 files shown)


chandlerc:
FYI, this also enables compressed debug information in fastbuild, which actually is a pretty huge reduction in build artifact size there. =]

.bazelrc
```diff
@@ -109,10 +109,20 @@ build --test_env=ASAN_SYMBOLIZER_PATH
...
-# Enable libpfm for google_benchmark on Linux only.
+# Allow per-platform configuration.
 build --enable_platform_specific_config
+
+# Enable libpfm for google_benchmark on Linux only.
 build:linux --define=pfm=1
 
+# Enable split debug info on Linux, which is significantly more space efficient
+# and should work well with modern debuggers. Note that this is not currently
+# supported on macOS.
```
dwblaikie: FWIW the MacOS debug info model looks a lot like Split DWARF already - I wouldn't expect they'll ever support actual Split DWARF because they've got a thing that works for them.
(The MacOS model moves more debug info out of the linked executable (including the line table, for instance) - but leaves it in the objects (there's a way to use Split DWARF that does this too - so you only have .o files, but the debug info in them still doesn't get linked in) - this has the build artifact size benefits, but does still mean you have to have all those debug info bytes available to the linker, even if it ignores them - not as good for distributed builds)

bazel/cc_toolchains/clang_cc_toolchain_config.bzl
```diff
@@ -361,31 +358,104 @@ def _impl(ctx):
...
             flag_set(
                 actions = codegen_compile_actions,
                 flag_groups = ([
                     flag_group(
-                        flags = ["-g"],
+                        flags = ["-glldb"],
+                    ),
+                    flag_group(
+                        flags = ["-gpubnames"],
+                        expand_if_available = "per_object_debug_info_file",
```
dwblaikie: This enables the index only when using split DWARF - it's probably worth enabling even if we aren't using split DWARF?

docs/project/contribution_tools.md
```diff
@@ -293,14 +298,32 @@ example:
...
+
+### Debugging with GDB instead of LLDB
+
+If you prefer using GDB, you may want to pass some extra flags to the build:
+
+```shell
+bazel build -c dbg --features=-lldb_flags --features=gdb_flags //toolchain
+```
+
+Or you can add them to your `user.bazelrc`, they are designed to be safe to pass
```
dwblaikie, line 318: Nice :)

dwblaikie (approved):
Couple of minor optional points, but overall this looks great - nice to have the workarounds/opt-outs and common errors (re: DWARFv5 unsupported in a debugger) documented!

bazel/cc_toolchains/clang_cc_toolchain_config.bzl
chandlerc: Sure, done. (Didn't know it was used w/o split.)

.bazelrc
chandlerc: Tweaked the comment. The fact that debug info remains in object files is still unfortunate because it means it gets copied into `.a` files from what I can see even though it doesn't make it all the way into the executable. Eventually we can maybe skip building `.a`s when we can use LLD on macOS.

Anyways, unlikely to change so made it more clear that it's just a Linux-specific option.

chandlerc:
Thanks, for review! Merging.

> nice to have the workarounds/opt-outs and common errors (re: DWARFv5 unsupported in a debugger) documented!

Most of this was already there, I just moved it around.

.bazelrc
dwblaikie: Oh, yeah - .a files. Yep, avoiding those would help on MacOS, and still help a bit on Linux too - fewer unique bytes on rebuild, etc.

PR #4383, opened 2024-10-08
Add skeletal format subcommand.

jonmeow:
This extracts out the SourceBuffer handling of `-` in order to trivially share it.

Note this still has a number of TODOs, it's just setting up the essential subcommand infrastructure, with some tests demonstrating that it at least does something.


jonmeow:
Part 1 of 4 (each using the last/eponymous commit)

1: #4383
2: #4384
3: #4385
4: #4386

toolchain/driver/format_subcommand.cpp
```diff
@@ -0,0 +1,98 @@
...
+  DriverResult result = {.success = true};
+  if (options_.input_filenames.size() > 1 &&
+      !options_.output_filename.empty()) {
+    driver_env.error_stream << "error: cannot format multiple input files when "
+                               "--output is set\n";
+    result.success = false;
+    return result;
+  }
+
+  auto per_file_error = [&]() {
```
chandlerc: Maybe:
```suggestion
  auto mark_per_file_error = [&]() {
```

toolchain/driver/format_subcommand.cpp
```diff
@@ -0,0 +1,98 @@
...
+    result.per_file_success.back().second = false;
+  };
+
+  StreamDiagnosticConsumer consumer(driver_env.error_stream);
+  for (auto& f : options_.input_filenames) {
+    // Push a result, which we'll update on failure.
+    result.per_file_success.push_back({f.str(), true});
+
+    // TODO: Consider refactoring this for sharing with compile.
+    auto source = SourceBuffer::MakeFromFileOrStdin(driver_env.fs, f, consumer);
```
chandlerc, line 74: Does it make sense to check for `-` here when there are multiple inputs? Or does the driver environment already handle that?

chandlerc: (or a TODO, I understand this is just a framework)

toolchain/driver/format_subcommand.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include "common/command_line.h"
+#include "toolchain/driver/driver_env.h"
+#include "toolchain/driver/driver_subcommand.h"
+
+namespace Carbon {
+
+// Options for the format subcommand.
+//
+// See the implementation of `Build` for documentation on members.
```
chandlerc, line 16: Maybe a TODO here or on the `Build` command for adding line-range flags and formatting support?

chandlerc (approved):
Generally LG, some fairly minor comments inline (TODOs, a naming tweak).

Feel free to land with fixes.

toolchain/driver/format_subcommand.cpp
jonmeow: Done

toolchain/driver/format_subcommand.cpp
jonmeow: Added a TODO. I grant the behavior for the user if they specify `-` multiple times may be odd (content in the first, empty for later, I'm expecting).

I think the driver environment doesn't handle this. Maybe it should, given that `compile` would have that issue too? Might mean factoring out this function wasn't worthwhile on the whole.

toolchain/driver/format_subcommand.h
jonmeow: I'm putting this on the `Format` call... later PRs are putting a few TODOs there already, so it seems like a good central spot to find missing features.

PR #4384, opened 2024-10-08
Add AUTOUPDATE-SPLIT to help with format tests.

jonmeow:
Since formatting covers comments, and the CHECK lines are in comments, it can create recursive behaviors. This introduces AUTOUPDATE-SPLIT as a way of formally designating a split to exclusively be used for autoupdate output.

### DIFF
--- a/testing/file_test/autoupdate.cpp
+++ b/testing/file_test/autoupdate.cpp
@@ -347,8 +347,9 @@ auto FileTestAutoupdater::Run(bool dry_run) -> bool {
     ++non_check_line_;
   }
 
-  // This will result in all check lines (and only check lines) being
-  // added to the split.
+  // When autoupdate_split_ was true, this will result in all check lines (and
+  // only check lines) being added to the split by FinishFile. We don't use
+  // autoupdate_split_ past this point.
   autoupdate_split_ = false;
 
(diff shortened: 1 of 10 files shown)


jonmeow:
Part 2 of 4 (each using the last/eponymous commit)

1: #4383
2: #4384
3: #4385
4: #4386

testing/file_test/autoupdate.cpp
```diff
@@ -341,6 +347,10 @@ auto FileTestAutoupdater::Run(bool dry_run) -> bool {
     ++non_check_line_;
   }
 
+  // This will result in all check lines (and only check lines) being
+  // added to the split.
```
geoffromer: Only if `autoupdate_split_` was initially true, right?

Would it work to put this inside the loop above, right before the `break`? It seems like it might be clearer in that context.

jonmeow: > Only if `autoupdate_split_` was initially true, right?

Modifying comment.

> Would it work to put this inside the loop above, right before the `break`? It seems like it might be clearer in that context.

It's mainly to control the behavior of FinishFile, so I lean towards it being adjacent to the affected function call(s)...

geoffromer (approved):

PR #4385, opened 2024-10-08
Add tracking of lexed comments, with skeletal formatting.

jonmeow:
In order to format comments, it's helpful if they're tracked. This tracks them separately from tokens in order to avoid interfering with parse; it'd be inconvenient if comment tokens could show up in arbitrary locations, albeit possible to support.

This additionally extracts out the TokenIterator support into a template in order to generally have it available for IndexBase types. I'm only adding it for CommentInfo, not sure if we'll want it elsewhere, but this structure still felt like a good fit.

### DIFF
--- a/toolchain/lex/tokenized_buffer.h
+++ b/toolchain/lex/tokenized_buffer.h
@@ -163,7 +163,7 @@ class TokenizedBuffer : public Printable<TokenizedBuffer> {
   auto GetPrevLine(LineIndex line) const -> LineIndex;
 
   // Returns true if the token comes after the comment.
-  auto IsBeforeComment(TokenIndex token, CommentIndex comment_index) const
+  auto IsAfterComment(TokenIndex token, CommentIndex comment_index) const
       -> bool;
 
   // Returns the comment's full text range.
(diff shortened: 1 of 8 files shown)


jonmeow:
[ed: previous PRs merged]

Part 3 of 4 (each using the last/eponymous commit)

1: #4383
2: #4384
3: #4385
4: #4386

toolchain/lex/tokenized_buffer.h
```diff
@@ -179,6 +162,13 @@ class TokenizedBuffer : public Printable<TokenizedBuffer> {
   // Returns the previous line handle.
   auto GetPrevLine(LineIndex line) const -> LineIndex;
 
+  // Returns true if the token comes after the comment.
+  auto IsBeforeComment(TokenIndex token, CommentIndex comment_index) const
+      -> bool;
```
chandlerc: I read `buffer.IsBeforeComment(token, comment_index)` as querying whether "`token` is before `comment_index`". Could we make this implement that API? Or be named `IsAfterComment` so that it reads "`token` is after `comment`"?

chandlerc (commented):
Really nice. And sorry about earlier, I think I misclicked and didn't realize how isolated this PR already was in the stack.

One interesting comment inline on the API here.

toolchain/lex/tokenized_buffer.h
jonmeow: Fixed to be `IsAfterComment`, that was an error in naming.

Note, I've also removed the base PRs so that this can be merged.

chandlerc (approved):
LG!

PR #4386, opened 2024-10-08
Implement some token-based formatting structure.

jonmeow:
Here I'm trying to add some simple formatting based on the token kind, aiming mainly to keep the implementation short for now.

This approach won't generalize to arbitrary structures (e.g., it doesn't discern between braces for a function body and a struct literal). I think we'll probably want to build a parse tree and associate parse kinds with tokens in order to format, additionally doing something less linear. But my essential goal at present is to just get a proof-of-concept that the basics can yield something that looks okay.


jonmeow:
[ed: dependencies now merged]

Part 4 of 4 (each using the last/eponymous commit)

1: #4383
2: #4384
3: #4385
4: #4386

toolchain/format/format.cpp
```diff
@@ -5,10 +5,59 @@
...
+    out.indent(indent);
+    line_state = LineState::Indented;
+  }
+}
+
+static auto AddWhitespace(llvm::raw_ostream& out, LineState& line_state,
+                          int indent) -> void {
+  if (line_state == LineState::WantsSpace) {
+    out << " ";
+    line_state = LineState::Indented;
```
geoffromer: This state transition would have been clearer to me if `Indented` were called something like `WantsText`. Would that make sense?

toolchain/format/format.cpp
```diff
@@ -5,10 +5,59 @@
...
+  // There is no output for the current line.
+  Empty,
+  // The current line has been indented, and may have more text.
+  Indented,
+  // If more output is added to the current line, add a space first.
+  WantsSpace,
+};
+
+// Adds a newline when needed.
+static auto AddNewline(llvm::raw_ostream& out, LineState& line_state) -> void {
```
geoffromer: These `Add` functions (and maybe `Format`) seem like they want to be members of a class that has `out`, `line_state`, and `indent` as data members. Is there any reason you're avoiding that?

toolchain/format/format.cpp
jonmeow: Renamed to `HasContent`.

Note that `WantsSpace` indicates that a space needs to be added if more content is added. If the `Wants` is interpreted similarly, `WantsText` would then imply that text should be added if more content is added, which would be incorrect.

toolchain/format/format.cpp
jonmeow: Done. We may want to switch to the `Context` model other areas use, but it should be simple enough to switch.

jonmeow:
Note, rebased since dependent changes are merged.

toolchain/format/BUILD
```diff
@@ -13,11 +13,16 @@ filegroup(
 
 cc_library(
     name = "format",
-    srcs = ["format.cpp"],
+    srcs = [
+        "format.cpp",
+        "formatter.cpp",
+        "formatter.h",
```
geoffromer, line 19: [No change needed]

FWIW I think it would have been fine to do all of this in format.cpp, but this way is fine too.

toolchain/format/formatter.h
```diff
@@ -0,0 +1,75 @@
...
+  // See class comments.
+  auto Run() -> bool;
+
+ private:
+  // Tracks the status of the current line of output.
+  enum class LineState : uint8_t {
+    // There is no output for the current line.
+    Empty,
+    // The current line has been indented, and may have text.
+    HasContent,
```
geoffromer: This still feels kind of confusing, since `WantsSpace` also indicates that the current line has been indented and has content, right? The only difference is whether any additional content should be preceded by a space. How about `WantsNoSpace`?

toolchain/format/formatter.cpp
```diff
@@ -0,0 +1,110 @@
...
+      case Lex::TokenKind::FileStart:
+        break;
+
+      case Lex::TokenKind::FileEnd:
+        AddNewline();
+        break;
+
+      case Lex::TokenKind::OpenCurlyBrace:
+        AddWhitespace();
+        *out_ << "{";
```
geoffromer, line 48: It seems like this should be followed by `line_state_ = LineState::HasContent;`, or at least a comment indicating that we're intentionally carrying over the `HasContent` state from `AddWhitespace`, and the same goes for the other places where we write characters to the output without modifying `line_state_`. That would make the state-machine structure clearer: we write some text to the current line, and then we update `line_state_` to keep it in sync with the actual state of the line.

toolchain/format/BUILD
jonmeow: I wanted a separate header for the class.

toolchain/format/formatter.cpp
jonmeow: It sounds like you're suggesting documenting function semantics at each call site, instead of the function itself. I think it's better to aim for central, rather than duplicated, documentation.

How about this: I've adjusted function names to try to better reflect semantics: RequireEmptyLine/PrepareForSpacedContent/PrepareForPackedContent. Does that work for you?

toolchain/format/formatter.h
jonmeow: I was starting with WantsNoSpace, but reflecting the below changes, how about HasSeparator/NeedsSeparator?

toolchain/format/formatter.cpp
jonmeow: By the way, if you want the explicit notes that line_state is modified, I can always revert back to the original (non-class) structure.

geoffromer (approved):

PR #4388, opened 2024-10-08
Generate parameter pattern-match IR from pattern IR

geoffromer:
Also propagate the pattern IR along with the pattern-match IR, and use it where appropriate.

Strictly speaking, some parts of the pattern-match IR are allocated eagerly, while traversing the pattern's parse tree, but they still aren't actually emitted until we traverse the associated pattern insts.


jonmeow:
I don't expect that we both need to review this, I'm going to let zygoloid handle it if he wants. :)

toolchain/check/deduce.cpp
```diff
@@ -170,12 +170,21 @@ auto DeduceGenericCallArguments(
       needs_substitution = false;
     }
 
-    CARBON_KIND_SWITCH(context.insts().Get(context.constant_values().GetInstId(
-                           param_const_id))) {
+    auto untyped_inst = context.insts().Get(
+        context.constant_values().GetInstId(param_const_id));
+    CARBON_KIND_SWITCH(untyped_inst) {
```
zygoloid: Looks like this will need to be merged with the recent deduction changes.

toolchain/check/deduce.cpp
```diff
@@ -170,12 +170,21 @@ auto DeduceGenericCallArguments(
...
+      case SemIR::InstKind::SymbolicBindingPattern:
+      case SemIR::InstKind::BindSymbolicName: {
+        auto entity_name_id = SemIR::EntityNameId::Invalid;
+        if (auto bind = untyped_inst.TryAs<SemIR::SymbolicBindingPattern>()) {
+          entity_name_id = bind->entity_name_id;
+        } else {
+          entity_name_id =
+              untyped_inst.As<SemIR::BindSymbolicName>().entity_name_id;
+        }
+        auto& entity_name = context.entity_names().Get(entity_name_id);
```
zygoloid: Is it feasible to use an `Any`-style typed inst here? (If this is just a temporary measure until we convert more things to use patterns then it's probably not worth it, but if you expect that we'll be matching both forms long-term, then that might be nicer.)

toolchain/check/eval.cpp
```diff
@@ -1355,6 +1355,26 @@ static auto TryEvalInstInContext(EvalContext& eval_context,
     case SemIR::ValueAsRef::Kind:
       break;
 
+    case CARBON_KIND(SemIR::SymbolicBindingPattern bind): {
```
zygoloid, line 1363: What does the constant value of a pattern represent? Where is this used?

toolchain/check/generic.cpp
```diff
@@ -425,35 +431,32 @@ auto ResolveSpecificDefinition(Context& context, SemIR::SpecificId specific_id)
...
-static auto ReplaceInstructionWithError(Context& context,
-                                        SemIR::InstId& inst_id) -> void {
-  inst_id = context.AddInstInNoBlock<SemIR::Param>(
-      context.insts().GetLocId(inst_id),
-      {.type_id = SemIR::TypeId::Error,
-       .name_id = SemIR::NameId::Base,
-       .runtime_index = SemIR::RuntimeParamIndex::Invalid});
+static auto ReplaceInstructionWithError(SemIR::InstId& inst_id) -> void {
+  inst_id = SemIR::InstId::BuiltinError;
 }
```
zygoloid: I don't think this function is pulling its weight any more; the callers would be clearer with this inlined.

toolchain/check/handle_function.cpp
```diff
@@ -75,18 +75,35 @@ static auto CheckFunctionSignature(Context& context,
...
+  for (auto [param_id, param_pattern_id] : llvm::zip(
+           llvm::concat<const SemIR::InstId>(
+               context.inst_blocks().GetOrEmpty(
+                   name_and_params.implicit_params_id),
+               context.inst_blocks().GetOrEmpty(name_and_params.params_id)),
+           llvm::concat<const SemIR::InstId>(
+               context.inst_blocks().GetOrEmpty(
+                   name_and_params.implicit_param_patterns_id),
+               context.inst_blocks().GetOrEmpty(
+                   name_and_params.param_patterns_id)))) {
```
zygoloid: I wonder if it'd be more efficient to `zip_equal` first and then `concat`, so that the loop logic doesn't need to deal with the two implicit parameter lists ending at different times.

toolchain/check/import_ref.cpp
```diff
@@ -718,76 +744,147 @@ class ImportRefResolver {
...
+    if (!param_patterns_id.is_valid() ||
+        param_patterns_id == SemIR::InstBlockId::Empty) {
+      return param_patterns_id;
+    }
+    const auto& param_patterns =
+        import_ir_.inst_blocks().Get(param_patterns_id);
+    llvm::SmallVector<SemIR::InstId> new_patterns;
+    for (auto param_id : param_patterns) {
+      // Figure out the param structure. This echoes
+      // Function::GetParamFromParamRefId.
```
zygoloid: Should this mention the pattern function instead of the param function?

toolchain/check/import_ref.cpp
```diff
@@ -718,76 +744,147 @@ class ImportRefResolver {
...
+      // TODO: Consider a different parameter handling to simplify import logic.
```
zygoloid: I see this TODO is copy-pasted, but do you have a different approach in mind here? If not, it's probably not worth keeping it here.

toolchain/check/import_ref.cpp
```diff
@@ -718,76 +744,147 @@ class ImportRefResolver {
...
+      }
+
+      auto param_pattern =
+          import_ir_.insts().GetAs<SemIR::ParamPattern>(param_pattern_id);
+
+      auto binding_id = param_pattern.subpattern_id;
+      auto binding =
+          import_ir_.insts().GetAs<SemIR::AnyBindingPattern>(binding_id);
+
+      // Rebuild the param instruction.
```
```suggestion
      // Rebuild the pattern.
```

toolchain/check/import_ref.cpp
```diff
@@ -1032,6 +1136,10 @@ class ImportRefResolver {
       case CARBON_KIND(SemIR::BindAlias inst): {
         return TryResolveTypedInst(inst);
       }
+      case SemIR::BindingPattern::Kind: {
+        // TODO: Should we be resolving BindingPatterns at all?
+        return ResolveAsConstant(SemIR::ConstantId::NotConstant);
+      }
```
zygoloid: Is this actually reached? Could we fall through to the default `TODO` call instead?

toolchain/check/import_ref.cpp
```diff
@@ -1090,6 +1198,9 @@ class ImportRefResolver {
       case CARBON_KIND(SemIR::PointerType inst): {
         return TryResolveTypedInst(inst);
       }
+      case CARBON_KIND(SemIR::SymbolicBindingPattern inst): {
+        return TryResolveTypedInst(inst);
+      }
```
zygoloid, line 1198: (Mirroring question from before:) do we / why do we need a constant value for a symbolic binding pattern, and what does it represent?

toolchain/check/pattern_match.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+ private:
+  llvm::SmallVector<WorkItem> stack_;
+
+  SemIR::InstId result_;
+
+  SemIR::SpecificId callee_specific_id_;
+};
+
+auto ProcessPattern(Context& context, MatchContext& match,
```
zygoloid: A comment on this function would be helpful; it's not obvious to me a priori what kind of processing is being performed here.

toolchain/check/pattern_match.cpp
```diff
@@ -0,0 +1,193 @@
...
+                    MatchContext::WorkItem entry) {
```
zygoloid:
```suggestion
                    MatchContext::WorkItem entry) -> void {
```

toolchain/check/pattern_match.cpp
```diff
@@ -0,0 +1,193 @@
...
+  auto pattern = context.insts().GetWithLocId(entry.pattern_id);
+  CARBON_KIND_SWITCH(pattern.inst) {
+    case SemIR::BindingPattern::Kind:
+    case SemIR::SymbolicBindingPattern::Kind: {
+      CARBON_CHECK(!entry.scrutinee_id.is_valid());
+      auto binding_pattern = pattern.inst.As<SemIR::AnyBindingPattern>();
+      auto bind_name = context.insts().GetAs<SemIR::AnyBindName>(
+          binding_pattern.bind_name_id);
+      context.inst_block_stack().AddInstId(bind_name.value_id);
```
zygoloid, line 113: Do we need to handle the case where `value_id` is `InstId::Invalid`? (I would actually expect it to always be invalid at this point -- it seems like we would find a value for the `BindName` as a consequence of pattern matching against some initial value.)

If this can be valid, how do we know that we have only one instruction to add here?

toolchain/lower/file_context.cpp
```diff
@@ -512,7 +511,8 @@ static auto BuildTypeForInst(FileContext& context, InstT /*inst*/)
 // Treat non-monomorphized symbolic types as opaque.
 template <typename InstT>
   requires(InstT::Kind.template IsAnyOf<SemIR::BindSymbolicName,
-                                        SemIR::InterfaceWitnessAccess>())
+                                        SemIR::InterfaceWitnessAccess,
+                                        SemIR::SymbolicBindingPattern>())
```
zygoloid: Can `SymbolicBindingPattern` be a type?

toolchain/check/subst.cpp
```diff
@@ -262,16 +262,23 @@ class SubstConstantCallbacks final : public SubstInstCallbacks {
...
-    if (!bind) {
+    auto entity_name_id = SemIR::EntityNameId::Invalid;
+    if (auto bind =
+            context_.insts().TryGetAs<SemIR::BindSymbolicName>(inst_id)) {
+      entity_name_id = bind->entity_name_id;
+    } else if (auto bind =
+                   context_.insts().TryGetAs<SemIR::SymbolicBindingPattern>(
+                       inst_id)) {
+      entity_name_id = bind->entity_name_id;
+    } else {
```
zygoloid, line 273: (As I've commented in other places, I wasn't expecting to see `SymbolicBindingPattern` turning up as part of a constant value and being replaced by the value of its binding.)

toolchain/check/pattern_match.cpp
```diff
@@ -0,0 +1,193 @@
...
+      context.inst_block_stack().AddInstId(binding_pattern.bind_name_id);
+      match.Finish(binding_pattern.bind_name_id);
+      break;
+    }
+    case CARBON_KIND(SemIR::AddrPattern addr_pattern): {
+      if (!entry.scrutinee_id.is_valid()) {
+        // We're still on the caller side of the pattern, so we traverse without
+        // emitting any insts.
```
zygoloid: This `if` condition + comment is a bit confusing, because the `if` condition is "are we matching on the callee side" and the comment seems to say we're on the caller side. I've figured out what you mean, but to make this clearer to future readers, how would you feel about something like:
```suggestion
      if (!entry.scrutinee_id.is_valid()) {
        // We're emitting match IR for the callee but we're still on the caller side of the pattern,
        // so we traverse without emitting any insts.
```

I also think it might be clearer to add a `bool` to `MatchContext` to indicate whether we're emitting match IR for the caller or callee side, rather than tracking it on the `WorkItem` (which suggests it might change between work items). Maybe a CHECK at the start of this function that we have a scrutinee iff we're on the caller side would be good.

toolchain/check/pattern_match.cpp
```diff
@@ -0,0 +1,193 @@
...
+        match.AddWork({.pattern_id = addr_pattern.inner_id,
+                       .scrutinee_id = SemIR::InstId::Invalid});
+        break;
+      }
+      auto scrutinee_ref_id =
```
zygoloid, line 128: I've not looked, but is there any work we could share between this and the handling of unary `&`? If so, maybe add a TODO to unify them?

toolchain/check/pattern_match.cpp
```diff
@@ -0,0 +1,193 @@
...
+  auto AddWork(WorkItem work_item) -> void {
+    CARBON_CHECK(!result_.is_valid());
+    stack_.push_back(work_item);
+  }
+
+  // Returns the next work item to process.
+  auto NextWorkItem() -> WorkItem { return stack_.pop_back_val(); }
+
+  // Sets the result of this pattern matching operation. Must not be called
+  // when there is still pending work, except to report an error.
```
zygoloid, line 67: Just noting for the future (not requesting a change in this PR): I think this single-result model isn't going to be what we want once we start supporting patterns that do not contain exactly one binding. (On the caller side, I think we want to add zero or more values to the argument list in the call, one for each `ParamPattern` we reach; on the callee side, I think we probably want to build up a value for the binding and set that value as the `bind_name`'s `value_id` once we reach it. Or something like that.)

But progress > perfection :)

toolchain/check/handle_binding_pattern.cpp
```diff
@@ -241,21 +255,29 @@ auto HandleParseNode(Context& context,
...
+      auto addr_pattern_id = context.AddPatternInst<SemIR::AddrPattern>(
+          node_id,
+          {.type_id = pointer_type->pointee_id, .inner_id = param_pattern_id});
+      context.node_stack().Push(node_id, addr_pattern_id);
+    } else {
+      CARBON_DIAGNOSTIC(
+          AddrOnNonPointerType, Error,
+          "`addr` can only be applied to a binding with a pointer type");
+      context.emitter().Emit(node_id, AddrOnNonPointerType);
+      context.node_stack().Push(node_id, param_pattern_id);
```
zygoloid, line 286: I didn't find a corresponding rule in the design, I just found [this](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/classes.md#:~:text=The%20pattern%20%27addr%20self%3A%20type%27%20means%20%22first%20take%20the%20address%20of%20the%20argument%2C%20which%20must%20be%20an%20l%2Dvalue%2C%20and%20then%20match%20pattern%20%27self%3A%20type%27%20against%20it%22.) which seems to suggest that `addr self: TypeThatSelfPtrImplicitlyConvertsTo` would be OK.

Do we need to add this rule to the implementation? If so, should we change the design to match?

(It looks like we use this assumption to form the type of the pattern, but what do we use that type for? Is it meaningful, or could we get the same result by using a placeholder `BuiltinPatternType` for all patterns? I think in general we don't know what type a pattern or subpattern is matching until we see the type of the scrutinee.)

zygoloid (commented):
This looks like a good step, thanks!

I think my main question here is around the constant evaluation and constant value of `SymbolicPatternBinding`. It looks like this was following what we do for `BindSymbolicName`, but it's not clear to me that `SymbolicPatternBinding` needs the same kind of treatment, or should ever appear as (part of) a symbolic constant value.

(Still TODO on my side: review the `testdata` changes to check for unexpected deltas.)

toolchain/check/deduce.cpp
geoffromer: Did you want me to do that now, or wait for the end of the review?

toolchain/check/deduce.cpp
geoffromer: My hope is that this is a temporary measure, so I've added a TODO. I don't yet have a definite plan for resolving it (see [this Discord discussion](https://discord.com/channels/655572317891461132/655578254970716160/1285378842541948980)), but `BindSymbolicName` and `SymbolicBindingPattern` conceptually have very little in common, so I'd rather not factor out an inst category for them. Notably, as far as I can tell the code in `subst.cpp` is the only other place that would benefit from such an `Any` inst.

toolchain/check/eval.cpp
geoffromer: The constant value of a pattern represents the result of symbolically evaluating the pattern as an expression, so in particular the constant value of a `SymbolicBindingPattern` represents the same thing as the constant value of the corresponding `BindSymbolicName`. This is used in `DeduceGenericCallArguments` -- see also the [Discord discussion](https://discord.com/channels/655572317891461132/655578254970716160/1285378842541948980) I linked in my reply above.

toolchain/check/handle_binding_pattern.cpp
geoffromer: Yeah, you're right that the type of an `AddrPattern` isn't actually used. However, it's not the case that pattern types in general aren't used. At least for now, I think `*BindingPattern`s need to have `type_id`s corresponding to their declared types, because they're used as values of those types.

More importantly, we need to be able to assign a type to a `Param` inst. Right now those are generated from `*BindingPattern`s, but that's temporary. My plan is to generate them from the corresponding `ParamPattern`s, for simplicity and for consistency with the rest of the design. That means a `ParamPattern` needs to have a type, which is computed from the type of its child, so its child needs to have a type, and so on. This isn't a problem for `AddrPattern` because it's always matched by the callee, so it's never a descendant of a `ParamPattern`.

So rather than a catch-all `BuiltinPatternType`, I've introduced a `BuiltinAutoType` to act as the type of patterns that have a deduced type, like `addr` patterns, and of course `auto` bindings. What do you think?

toolchain/check/handle_function.cpp
geoffromer: Yes, this should've been `zip_equal`. Fixed.

I can't quite tell if you're suggesting to also reverse the order of composition, making this `concat(zip_equal, zip_equal)` instead of `zip_equal(concat, concat)`. If so, that's not as straightforward as you'd hope: `concat` needs its inputs to be at least forward iterators, but `zip` produces only input iterators, and the mismatch leads to stack-use-after-return errors.

toolchain/check/import_ref.cpp
geoffromer: The only thing that comes to mind is treating each pattern inst as a separate worklist item. That would be more flexible and probably easier to understand, but might be somewhat less efficient.

toolchain/check/import_ref.cpp
geoffromer: See my response to the other question.

toolchain/check/pattern_match.cpp
geoffromer: Yes, this aspect of `MatchContext` is definitely a placeholder.

toolchain/check/pattern_match.cpp
geoffromer: Done, and also gave it a more meaningful name.

toolchain/check/pattern_match.cpp
geoffromer: By construction, the `value_id` of a `BindName` is always the corresponding `Param` inst (see `handle_binding_pattern.cpp`). Of course, that only works because we don't support any non-function-call pattern matching yet. In a follow-up PR I want to emit the `Param` while processing the `ParamPattern`, and then pass it in here as the scrutinee, so this code will be responsible for updating `bind_name.value_id` to refer to the scrutinee. I've added a TODO for that.

toolchain/check/pattern_match.cpp
geoffromer: I don't think there's a meaningful amount of shareable code here. Unary `&` doesn't have the `ConvertToValueOrRefExpr` call because it doesn't allow initializing expressions. It has a `switch` on expression category, but the diagnostic messages are different, and the success/failure structure is different because unary `&` doesn't allow ephemeral refs. Even the inst allocation is different: unary `&` uses `AddInstAndPush`, but we use `AddInst`.

toolchain/check/pattern_match.cpp
geoffromer: OK, how's this?

> I also think it might be clearer to add a `bool` to `MatchContext` to indicate whether we're emitting match IR for the caller or callee side, rather than tracking it on the `WorkItem` (which suggests it might change between work items).

Good point, but I made it an `enum` because we eventually want to support non-function-call pattern matches as well.

> Maybe a CHECK at the start of this function that we have a scrutinee iff we're on the caller side would be good.

I don't intend that assertion to hold in the long run, so I'd rather not assert it now. For example, as noted on the previous comment, I intend for the `*BindingPattern` case to have a scrutinee eventually (hopefully the PR after this one). The invariant that I'm aiming for is that we have a scrutinee iff we should emit match insts, but we aren't quite there yet, so I've marked that as a TODO. In any event, that `CHECK` will need to be applied to the individual cases, because there's no top-level representation of "should emit match insts".

toolchain/lower/file_context.cpp
geoffromer: Yes. I think this another instance of the point discussed elsewhere, about using patterns as values.

toolchain/check/deduce.cpp
zygoloid: It's probably somewhat easier to review if you wait, but I can make it work :) I bring it up now because I think the merge might not be trivial.

toolchain/check/eval.cpp
zygoloid: Maybe I'm misunderstanding -- I though the conclusion of [this discussion](https://discord.com/channels/655572317891461132/655578254970716160/1289359153243885659) was that we'd have the name of the binding always resolve to the `Bind*Name` instruction, not to the `BindingPattern`.

toolchain/check/handle_function.cpp
zygoloid: Yes, my suggestion was to flip the order, so it'd be `concat(zip_equal, zip_equal)`. The current loop structure seems likely to create an inefficient loop, because each iteration we need to check separately whether each of the two implicit lists has ended. If `concat(zip_equal, zip_equal)` doesn't work, maybe there's another option.

Playing around with it a bit, maybe something like:
```c++
  for (auto [params_id, patterns_id] :
       {std::pair{name_and_params.implicit_params_id,
                  name_and_params.implicit_param_patterns_id},
        std::pair{name_and_params.params_id,
                  name_and_params.param_patterns_id}}) {
    for (auto [param_id, pattern_id] : llvm::zip_equal(params_id, patterns_id)) {
```
That also seems a bit more readable to me.

toolchain/check/pattern_match.cpp
zygoloid: Ah, I see, for a parameter name binding this makes more sense. Thanks.

toolchain/check/deduce.cpp
zygoloid: OK. Regarding the TODO: while I do agree that function call deduction should not be operating on pattern match IR, I don't think that means we won't see a `BindSymbolicName`. For example, in a pattern like

```carbon
fn F[T:! type](x: T);
```

when deducing the `BindingPattern` for `x: T` against an argument, we deduce the expression `T` against the type of the argument, which means we're deducing a `BindSymbolicName`. So I think we may find we end up needing to deduce against both.

But yeah, I think it makes sense to say that we don't expect many places to want to treat the two of them as the same thing, and deduction here is a special exception. (I don't even think deduction wants the same behavior for the two of them here -- values inferred from `SymbolicBindingPattern`s should be substituted into later parameters, whereas values inferred from `BindSymbolicName`s should be still be deduced if we see them again later. But that's an issue for the future! For now it's just good to be able to tell these cases apart.)

toolchain/check/handle_binding_pattern.cpp
zygoloid: Hm, I see. Yeah, I think I was wrong to think that we don't need types for patterns until we see the scrutinee. As you say, for parameters we really do need a type, and indeed we presumably want a rule for parameters that disallows any case where the type is not known -- or where matching against an argument of a different type would do something that's not equivalent to converting the argument to the parameter type and then matching it.

That means that we can support binding patterns, `var` patterns, tuple patterns, and struct patterns in function parameter lists, but not expression patterns, the designator form of alternative patterns, nor `auto` deduction. That sounds right to me, and has a lot of overlap with not wanting refutable patterns in function parameters (which is mostly the same but also rules out the fully-qualified form of alternative patterns but would allow `auto`).

OK. So with this new understanding, I think I agree, something like `BuiltinAutoType` makes sense.

Aside: I also wonder whether we'll want a distinct `ExprCategory` for patterns -- rather than treating them as simply `NotExpr`s with a type, classifying them as `ExprCategory::Pattern` might allow code to more easily reason about them. But let's defer that until we have an actual use for that information. Also, maybe modeling this separately from the `ExprCategory` is better, to avoid adding churn to the `switch`es on `ExprCategory`.

toolchain/lower/file_context.cpp
zygoloid: I think this case is a little different. Even if `SymboilcBindingPattern` can be used as a value, it shouldn't be showing up here: it can't be an instruction referred to by a `TypeId`, because it constant evaluates to a `BindSymbolicName` not to a `SymbolicBindingPattern`. I think its `InstKind` should classify it as `InstIsType::Never` because it's never the canonical definition of a type.

toolchain/check/deduce.cpp
geoffromer: OK, I'd prefer to merge at the end anyway, so let's do that.

toolchain/check/deduce.cpp
geoffromer: OK, so should I drop the TODO?

toolchain/check/eval.cpp
geoffromer: That's my understanding as well, but how is this code inconsistent with that plan?

toolchain/lower/file_context.cpp
geoffromer: OK, that seems to work.

toolchain/check/handle_binding_pattern.cpp
geoffromer: OK, so is this resolved?

toolchain/check/deduce.cpp
zygoloid: I think so, yes. We're already passing in the pattern in call.cpp and deducing against it here, so I think this is actually done already with this PR :)

toolchain/check/eval.cpp
zygoloid: Aha, I see. I had missed [this comment](https://discord.com/channels/655572317891461132/655578254970716160/1285628379202064425).

I was expecting that the end result of the decision was that we would never use `SymbolicBindingPattern` as a value or evaluate it as an expression. I think we should try to reach that point, but I also think it'd be OK to not do so in this PR. However, it looks like we've got a fair few changes here supporting constant evaluation of `SymbolicBindingPattern`s, and it might be easier (and closer to what I think the end goal would be) to special-case `SymbolicBindingPattern`s in deduction more directly.

Right now, deduction bails out when it reaches an instruction that doesn't have a symbolic constant value simply because it's a quick way to detect a case where there's nothing to deduce inside, but that check becomes incorrect once we're deducing against patterns. You could just remove it, or alternatively check whether either there's a symbolic constant value or the parameter is a pattern instruction.

geoffromer: (Sorry, I should've linked more directly to the relevant part of the discussion.)

As discussed in the meeting, updating deduction to not require this looks a bit too complicated to do in this PR, so I've added a TODO.

zygoloid (approved):
Looks good, hopefully the merge isn't too painful.

geoffromer:
> Looks good, hopefully the merge isn't too painful.

One of the least painful merges I've done recently, in fact :-)

PR #4389, opened 2024-10-08
Add support for deduction of most kinds of type constant.

zygoloid:
This adds deduction in all the cases where we can match the instruction
fields of the parameter against the corresponding instruction fields of
the argument. This handles all current type constants except for struct
types, for which we would want to match by field name.

### DIFF
--- a/toolchain/check/deduce.cpp
+++ b/toolchain/check/deduce.cpp
@@ -39,11 +39,31 @@ class DeductionWorklist {
         {.param = param, .arg = arg, .needs_substitution = needs_substitution});
   }
 
+  // Adds a single (param, arg) type deduction.
+  auto Add(SemIR::TypeId param, SemIR::TypeId arg, bool needs_substitution)
+      -> void {
+    Add(context_.types().GetInstId(param), context_.types().GetInstId(arg),
+        needs_substitution);
+  }
+
+  // Adds a single (param, arg) deduction of a specific.
+  auto Add(SemIR::SpecificId param, SemIR::SpecificId arg,
+           bool needs_substitution) -> void {
+    auto& param_specific = context_.specifics().Get(param);
+    auto& arg_specific = context_.specifics().Get(arg);
+    if (param_specific.generic_id != arg_specific.generic_id) {
+      // TODO: Decide whether to error on this or just treat the specific as
+      // non-deduced. For now we treat it as non-deduced.
+      return;
+    }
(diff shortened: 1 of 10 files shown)


zygoloid:
Depends on #4380.

toolchain/check/deduce.cpp
```diff
@@ -39,11 +39,31 @@ class DeductionWorklist {
...
+    AddAll(param_specific.args_id, arg_specific.args_id, needs_substitution);
+  }
+
   // Adds a list of (param, arg) deductions. These are added in reverse order so
   // they are popped in forward order.
-  auto AddAll(llvm::ArrayRef<SemIR::InstId> params,
-              llvm::ArrayRef<SemIR::InstId> args, bool needs_substitution)
-      -> void {
+  template <typename InnerId>
+  auto AddAll(llvm::ArrayRef<InnerId> params, llvm::ArrayRef<InnerId> args,
```
jonmeow:
```suggestion
  template <typename ElementId>
  auto AddAll(llvm::ArrayRef<ElementId> params, llvm::ArrayRef<ElementId> args,
```

Minor, but maybe ElementId instead?

toolchain/check/deduce.cpp
```diff
@@ -208,65 +266,99 @@ auto DeductionContext::Deduce() -> bool {
...
+      case SemIR::TupleValue::Kind: {
+        auto arg_inst = context().insts().Get(arg_id);
+        if (arg_inst.kind() != param_inst.kind()) {
+          break;
+        }
+        auto [kind0, kind1] = param_inst.ArgKinds();
+        worklist_.AddInstArg(kind0, param_inst.arg0(), arg_inst.arg0(),
+                             needs_substitution);
+        worklist_.AddInstArg(kind1, param_inst.arg1(), arg_inst.arg1(),
+                             needs_substitution);
```
jonmeow, line 338: Is it worth at all conditioning this based on the # of args to the inst? Or if the arg is invalid?

jonmeow (approved):
Basically LGTM, feel free to merge when ready.

toolchain/check/deduce.cpp
zygoloid: We could have different case blocks for different numbers of arguments, but I think it's *probably* not worth the complexity and tighter coupling. The switch in `AddInstArg` immediately catches the case of no corresponding argument, so we should deal with that case quickly.

So far, we don't have any cases where the arg can be present but invalid. I guess we can handle that if/when it starts coming up.

PR #4393, opened 2024-10-09
Assign locations to instructions in a generic eval block.

zygoloid:
The locations point to the first instruction in the generic that needed the relevant constant value or type.

For now, this must makes the formatted SemIR a bit more useful, but in the future it will also provide locations for diagnostics caused by monomorphization failure.


jonmeow (approved):
LG

PR #4394, opened 2024-10-09
Update conduct team member list

CelineausBerlin:
deleted @flysand7 from the conduct team member list after they left the team

### DIFF
--- a/docs/project/teams/conduct_team.md
+++ b/docs/project/teams/conduct_team.md
@@ -21,7 +21,6 @@ The current members of this list are:
 -   Céline Dedaj (@celineausberlin on Discord and GitHub)
 -   Christopher Di Bella (@cjdb.work on Discord, @cjdb on GitHub)
 -   Lexi Bromfield (@lexinadia on Discord and @lexi-nadia on GitHub)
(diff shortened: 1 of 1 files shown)


jonmeow:
Noting #4392 missed this, had you considered changing one location (either or) to refer to the other, eliminating one copy?

CelineausBerlin:
Good point! I'll look into it. Thank you!

CelineausBerlin:
After touching base with the conduct team, we want to avoid duplicate information, indeed: 
The [conduct team page](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/project/teams/conduct_team.md) should contain the team breakdown, not the [CoC page](https://github.com/carbon-language/carbon-lang/blob/trunk/CODE_OF_CONDUCT.md). 
So, before we modify the latter, let's update the former (with #4394). 

Then, when this update is done, we'll go and update the [CoC file](https://github.com/carbon-language/carbon-lang/blob/trunk/CODE_OF_CONDUCT.md) so it makes a reference to the conduct team page.

Does that resolve the issue?

zygoloid (approved):
Sounds reasonable to me.

PR #4395, opened 2024-10-09
Track and resolve the specific callee in a call to a generic function

zygoloid:
Add a new `specific_function` instruction that represents a generic function plus its deduced argument list as a callee in a function call. The new instruction can only appear as the immediate operand of a call instruction, so we give it a builtin placeholder type.

At the end of each file, require definitions for all specific functions used in that file. Resolve the generic with the argument list to produce those specific function definitions as needed, and diagnose if the generic doesn't have a definition available.

A few tests are updated in cases where they declared and used generic functions but didn't previously provide a function definition.


toolchain/sem_ir/typed_insts.h
```diff
@@ -910,6 +910,27 @@ struct SpecificConstant {
...
 
+// A specific instance of a generic function. This represents the callee in a
+// call instruction that is calling a generic function, where the specific
+// arguments of the function have been deduced.
+//
+// TODO: This value corresponds to the `(FunctionType as Call(...)).Op` function
+// in the overloaded calls design. Eventually we should represent it more
+// directly as a member of the `Call` interface.
+struct SpecificFunction {
+  static constexpr auto Kind = InstKind::SpecificFunction.Define<Parse::NodeId>(
```
geoffromer, line 921: `Define<Parse::NodeID>` usually seems to be accompanied by a TODO to make the node type more specific. Should that be done here?

toolchain/check/call.cpp
```diff
@@ -141,10 +143,20 @@ auto PerformCall(Context& context, SemIR::LocId loc_id, SemIR::InstId callee_id,
   // for the call.
   auto callee_specific_id = ResolveCalleeInCall(
       context, loc_id, callable, "function", callable.generic_id,
-      callee_function.specific_id, callee_function.self_id, arg_ids);
+      callee_function.enclosing_specific_id, callee_function.self_id, arg_ids);
   if (!callee_specific_id) {
     return SemIR::InstId::BuiltinError;
   }
+  if (callee_specific_id->is_valid()) {
```
geoffromer, line 150: What does it mean for `callee_specific_id` to be invalid, as opposed to null?

toolchain/check/testdata/function/generic/undefined.carbon
```diff
@@ -0,0 +1,446 @@
...
+  // CHECK:STDERR:
+  return Undefined(0);
+}
+
+// --- fail_call_monomorphization_error.carbon
+
+library "[[@TEST_NAME]]";
+
+fn ErrorIfNIsZero(N:! i32) {
+  // CHECK:STDERR: fail_call_monomorphization_error.carbon:[[@LINE+3]]:10: error: integer type width of 0 is not positive
```
geoffromer: This isn't a template generic, so we shouldn't be getting monomorphization errors, right? Should there be a TODO here for that?

Also, what does this case have to do with the rest of this file, or the rest of this PR?

toolchain/check/call.cpp
zygoloid: An invalid `SpecificId` generally means "not generic". The nullopt return value here is used to indicate that we're recovering from an error.

toolchain/check/testdata/function/generic/undefined.carbon
zygoloid: This is a "generics calling templates" error case, which can lead to monomorphization / instantiation errors. Notionally the error happens in some templated region inside the `Core.Int` generic but the outcome is the same -- we reject if the generic is used with bad values for `N`.

The relevance to this PR is that this error demonstrates that we're actually resolving the generic definition in response to the generic function being used. The error message is a tool by which the work we're doing is visible, that's less subtle than the produced SemIR being correct. But yeah, the relevance to this file is ... questionable :) I'll split this out to a separate file.

zygoloid: Done and added comment to try to explain.

toolchain/sem_ir/typed_insts.h
zygoloid: No -- the `SpecificFunction` has the location of the callee in the call expression, which is an arbitrary expression, so we can't associate it with a more specific parse node.

toolchain/check/call.cpp
geoffromer: OK, this is a bit of a drive-by fix, but could you document that on `ResolveCalleeInCall`?

geoffromer (approved):

toolchain/check/call.cpp
zygoloid: Done.

PR #4396, opened 2024-10-09
Refactor compile-related tests to share construction.

jonmeow:
Note in particular that this fixes an issue where SharedValueStore had been shared across files, when they should be per-file. This is only visible when doing multiple compilations in a single test, which was rare before.

This also moves these tests into the Testing namespace. My memory of the various namespacing changes is that we'd generally agreed to have tests in Testing so that we'd see SemIR:: and similar, same as we would in a lot of the implementation.

### DIFF
--- a/toolchain/lex/tokenized_buffer_test.cpp
+++ b/toolchain/lex/tokenized_buffer_test.cpp
@@ -23,15 +23,14 @@
 namespace Carbon::Lex {
 namespace {
 
+using ::Carbon::Testing::ExpectedToken;
+using ::Carbon::Testing::IsSingleDiagnostic;
+using ::Carbon::Testing::TestRawOstream;
 using ::testing::_;
 using ::testing::ElementsAre;
 using ::testing::Eq;
 using ::testing::HasSubstr;
 using ::testing::Pair;
-
-using ::Carbon::Testing::ExpectedToken;
-using ::Carbon::Testing::IsSingleDiagnostic;
-using ::Carbon::Testing::TestRawOstream;
 
 namespace Yaml = ::Carbon::Testing::Yaml;
 
--- a/toolchain/lex/tokenized_buffer_test.cpp
+++ b/toolchain/lex/tokenized_buffer_test.cpp
@@ -17,14 +17,12 @@
 #include "toolchain/diagnostics/mocks.h"
 #include "toolchain/lex/lex.h"
 #include "toolchain/lex/tokenized_buffer_test_helpers.h"
+#include "toolchain/testing/compile_helper.h"
 #include "toolchain/testing/yaml_test_helpers.h"
 
-namespace Carbon::Lex {
+namespace Carbon::Testing {
 namespace {
 
-using ::Carbon::Testing::ExpectedToken;
-using ::Carbon::Testing::IsSingleDiagnostic;
-using ::Carbon::Testing::TestRawOstream;
 using ::testing::_;
 using ::testing::ElementsAre;
 using ::testing::Eq;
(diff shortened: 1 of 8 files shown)


toolchain/testing/compile_helper.h
```diff
@@ -0,0 +1,93 @@
...
+#include "toolchain/diagnostics/diagnostic_consumer.h"
+#include "toolchain/lex/lex.h"
+#include "toolchain/parse/parse.h"
+#include "toolchain/parse/tree_and_subtrees.h"
+#include "toolchain/source/source_buffer.h"
+
+namespace Carbon::Testing {
+
+// A test helper for compile-related functionality.
+class CompileHelper {
```
chandlerc, line 20: Would it be OK to add a source file and moving the inline definitions out-of-line?

toolchain/lex/tokenized_buffer_test.cpp
```diff
@@ -17,14 +17,12 @@
 #include "toolchain/diagnostics/mocks.h"
 #include "toolchain/lex/lex.h"
 #include "toolchain/lex/tokenized_buffer_test_helpers.h"
+#include "toolchain/testing/compile_helper.h"
 #include "toolchain/testing/yaml_test_helpers.h"
 
-namespace Carbon::Lex {
+namespace Carbon::Testing {
```
chandlerc: For some reason, when we switched to using the `Lex` sub-namespace, I thought we had ended up switching how tests work instead of putting them in the `Testing` namespace. But I can't find such a discussion, only the discussion suggesting putting the tests in `Testing` as you've switched to here.

Anyways, I'm happy either way, only mentioning it in case there *was* such a discussion, and I just can't find it but you or someone else remember it.

toolchain/parse/typed_nodes_test.cpp
```diff
@@ -21,160 +22,137 @@ namespace Carbon::Parse {
...
-    tree->SetNodeKindForTesting(node_id, kind);
+  static auto SetNodeKind(const Tree& tree, NodeId node_id, NodeKind kind)
+      -> void {
+    const_cast<Tree&>(tree).SetNodeKindForTesting(node_id, kind);
   }
 };
 
+}  // namespace Carbon::Parse
+
+namespace Carbon::Testing {
```
chandlerc: The need for this does make me somewhat prefer tests in the component namespace rather than the `Testing` namespace, and using-declarations for the `Testing` names... But very weakly.

chandlerc (approved):
Really nice helper!

I had remembered reversing the `Testing` namespace decision, but I can't find anything in Discord... 🤷🏻 (see inline comment, not blocking)

I'd like to move the implementation of the helper to a source file if we could though. LGTM to land with that.

toolchain/lex/tokenized_buffer_test.cpp
jonmeow: I've tracked down discussion [here](https://discord.com/channels/655572317891461132/655578254970716160/1141153794097422457), via https://github.com/carbon-language/carbon-lang/pull/3170/files#r1310772837. I'm updating #3070 to mention this.

jonmeow: But now I see #3244, I guess there was a little more. Maybe the tests still in Carbon::Testing are actually the mistake?

toolchain/parse/typed_nodes_test.cpp
jonmeow: Undid this part of the change

toolchain/lex/tokenized_buffer_test.cpp
jonmeow: Undid this part of the change

toolchain/testing/compile_helper.h
jonmeow: Done

jonmeow:
I'm going to go ahead and merge since I have more work depending on this, but please let me know if you see any issues with the changes (particularly mistakes with NS edits) and I'll make another PR to fix.

PR #4397, opened 2024-10-10
Stitch together adjacent comments using the indent.

jonmeow:
This is improving the comment production to produce fewer distinct comments.

At present, comment processing uses strict prefix matching. It either expects `// ` (with a space) for valid comments, or just `//` (without a space) for invalid comments that lacked the space.

As a consequence, the following would be three comments:

```
// Comment 1
//
//
// Comment 4
```

This is because a 3-character prefix is used for valid comments. The prefix switches between lines 1 and 2, and again between lines 3 and 4, each resulting in a separate comment.

For contrast, this is one comment because only a 2-character prefix is used:

```
//Comment 1
//
//
//Comment 4
```

That's because all lines lack a suffix space.

Additionally, with SIMD 16-byte boundaries, further splits can occur if processing needs to transition to non-SIMD.

Here, I'm trying to just address all of this by stitching together adjacent comments. Since a lexed comment starts at the `//` excluding the indent, the delta from the prior comment must be precisely the indent.

### DIFF
--- a/toolchain/lex/lex.cpp
+++ b/toolchain/lex/lex.cpp
@@ -200,8 +200,6 @@ class [[clang::internal_linkage]] Lexer {
  private:
   class ErrorRecoveryBuffer;
 
-  auto AddLexedComment() -> void;
-
   TokenizedBuffer buffer_;
 
   ssize_t line_index_;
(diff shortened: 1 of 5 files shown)


toolchain/lex/lex.cpp
```diff
@@ -858,6 +860,75 @@ auto Lexer::LexCommentOrSlash(llvm::StringRef source_text, ssize_t& position)
...
 
+// Lexes for a comment prefix using SIMD support. Returns true if the end of
+// lines with the provided prefix has been found. Returns false if it's unable
+// to finish lexing (but may be partially lexed).
+//
+// TODO: We should extend this to 32-byte SIMD on platforms with support.
+static auto LexCommentPrefixInSimd(llvm::StringRef source_text,
+                                   ssize_t& position, ssize_t prefix_size,
+                                   ssize_t first_line_start,
+                                   llvm::function_ref<void()> skip_to_next_line)
```
chandlerc: Could you split out the refactoring into a separate commit or PR?

Because this PR *does* contain a behavior change, it makes it harder for me to review this section. The code here is fairly complex, and there seem to be three things I need to look for:

1) Extracting it to a separate function
2) Changing how the different architectures are grouped in the code
3) (possibly) some behavior change to support the stitching below.

And it's hard to review all three without the incremental diff for (2) or at least (3).

FWIW, I find the original structure of the code easier to read -- the duplication of the loop condition and call to `skip_to_next_line`, for me, are outweighed by being able to see the complete loop for each architecture.

toolchain/lex/tokenized_buffer.h
```diff
@@ -457,6 +459,19 @@ class TokenizedBuffer : public Printable<TokenizedBuffer> {
   auto PrintToken(llvm::raw_ostream& output_stream, TokenIndex token,
                   PrintWidths widths) const -> void;
 
+  // Adds a comment. This uses the indent to potential stitch together two
+  // adjacent comments.
```
chandlerc:
```suggestion
  // Adds a comment. This uses the indent to potentially stitch together two
  // adjacent comments.
```

toolchain/lex/tokenized_buffer_test.cpp
```diff
@@ -1106,6 +1107,113 @@ TEST_F(LexerTest, DiagnosticUnrecognizedChar) {
...
+                           .str();
+
+  auto& buffer = compile_helper_.GetTokenizedBuffer(source);
+  EXPECT_TRUE(buffer.has_errors());
+
+  EXPECT_THAT(buffer.comments_size(), Eq(5));
+  for (int i :
+       llvm::seq(std::min(static_cast<int>(buffer.comments_size()), 5))) {
+    EXPECT_THAT(buffer.GetCommentText(CommentIndex(i)).str(),
+                testing::StrEq(Comments[i]));
```
chandlerc: Maybe use the size of the array?

```suggestion
  EXPECT_THAT(buffer.comments_size(), Eq(std::size(Comments)));
  for (int i :
       llvm::seq(std::min<int>(buffer.comments_size(), std::size(Comments)))) {
    EXPECT_THAT(buffer.GetCommentText(CommentIndex(i)).str(),
                testing::StrEq(Comments[i]));
```

Not sure I got the `std::min` type trick quite right...

toolchain/lex/lex.cpp
jonmeow: Reverted. (not sure I'll bother with a separate PR, the TODO had seemed small while I was touching the code, but I think duplication makes it less appealing to me to fix)

chandlerc (approved):
LGTM, the stitching is really nice!

PR #4399, opened 2024-10-10
Initial, very rough lowering for calls to specific functions and specific function declarations.

zygoloid:
Generate declarations of specific functions on demand. Definitions are not emitted yet, and I'm using a temporary, known-broken scheme for name mangling.

### DIFF
--- a/toolchain/lower/file_context.h
+++ b/toolchain/lower/file_context.h
@@ -49,22 +49,7 @@ class FileContext {
 
   // Gets a or creates callable's function. Returns nullptr for a builtin.
   auto GetOrCreateFunction(SemIR::FunctionId function_id,
-                           SemIR::SpecificId specific_id) -> llvm::Function* {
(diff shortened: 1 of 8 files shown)


toolchain/lower/file_context.h
```diff
@@ -47,6 +47,25 @@ class FileContext {
     return functions_[function_id.index];
   }
 
+  // Gets a or creates callable's function. Returns nullptr for a builtin.
+  auto GetOrCreateFunction(SemIR::FunctionId function_id,
+                           SemIR::SpecificId specific_id) -> llvm::Function* {
```
jonmeow: This seems long enough that maybe it belongs in the cpp (the implementation)

jonmeow (approved):
LG

toolchain/lower/file_context.h
zygoloid: Done.

PR #4400, opened 2024-10-11
Update CODE_OF_CONDUCT.md to delete duplicate info

CelineausBerlin:
Deleted conduct team member list to _avoid duplicate information_: 
the up-to-date member list is already _on the conduct team page_, which is _referenced_ in this very section.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -145,15 +145,9 @@ any member of the conduct team directly.
 
 ## Conduct team
 
-The conduct team can be emailed at conduct@carbon-lang.dev, and currently has
-the following members:
+The conduct team can be emailed at conduct@carbon-lang.dev.
 
--   Allison Poppe (@acpoppe on Discord and GitHub)
--   Céline Dedaj (@celineausberlin on Discord and GitHub)
--   Christopher Di Bella (@cjdb.work on Discord, @cjdb on GitHub)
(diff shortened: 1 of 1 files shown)


chandlerc:
A looong time ago when first working on Codes-of-Conduct (for LLVM at the time), I head read in a few places that it was specifically helpful for them to directly state who was on the conduct team rather than putting it behind a link. But I don't remember the full context, and at least some of it was more a reaction to CoCs that made it very obfuscated who exactly would receive reports, not about there being something literally important about it being on a single page.

So at least for myself, I find making the more durable statement that there is an explicit conduct team and linking to that to be (again, personally) better.

But CC-ing @KateGregory to get other perspectives here and also because she's been involved in CoCs longer than I have and so might remember relevant context here.

KateGregory:
> But CC-ing @KateGregory to get other perspectives here and also because she's been involved in CoCs longer than I have and so might remember relevant context here.

It's important to me that people can see who is on the team. I don't have a problem if you need to click from the code of conduct to see who is on the team. There are a lot of conferences where no matter how I click I can't find out who any of the people are on the CoC team. I think the issue of a stale page because information is in multiple places is a good enough reason to have the team on a separate page from the actual code.

CelineausBerlin:
Does the proposed change resolve this then?
(taking the member list out of the CoC's page and keeping it only on the conduct team's page)

PR #4407, opened 2024-10-14
Add vtable pointers to class layout

dwblaikie:
A small step to virtual functions - adding vtable pointers to the layout, but not initializing or otherwise using them at this stage.

A few open design questions I'd love feedback on:

* Is this the right/good enough SemIR representation for now? This patch adds a `is_dynamic` attribute to `SemIR::Class` and populates/flags it based on the flag of the base class, or if any virtual function is declared in the class (or, at least that's my intent). Some other options include:
  * Each `Class` could store a `ClassId` (or `TypeId`?) of the (possibly indirect, possibly self) base class that is the first one that is dynamic/has a vtable pointer
  * Could make the property narrower, like `has vtable pointer` and have it `true` only on the type that introduces the vtable - then derived classes would have to walk their base classes to check if they're the one that needs to define the vtable pointer or not
* Should the vtable be the first element in the type? If there's a non-dynamic base type, we could have a layout that's `{<non-dynamic base type>, vtable ptr, <derived members>}`? Derived types would still be able to uniquely identify where their vtable pointer is just fine... - and the vtable pointer is, in a sense, a member of that intermediate type, so it does seem a bit strange to force it to the front - but I guess it's probably more efficient in some ways?

Open to any other suggestions/advice/thoughts on the direction, etc.


zygoloid:
> * Is this the right/good enough SemIR representation for now? This patch adds a `is_dynamic` attribute to `SemIR::Class` and populates/flags it based on the flag of the base class, or if any virtual function is declared in the class (or, at least that's my intent).

I think storing this flag is reasonable; it seems like a property of the class that we will want to query relatively frequently. One other question we will frequently need to answer is "where is the vptr?" -- eg, when forming a virtual call or initializing an instance of the class -- so perhaps it'd make sense to also store some information about that, when we get to the point of wanting to access the vptr.

> * Should the vtable be the first element in the type? If there's a non-dynamic base type, we could have a layout that's `{<non-dynamic base type>, vtable ptr, <derived members>}`? Derived types would still be able to uniquely identify where their vtable pointer is just fine... - and the vtable pointer is, in a sense, a member of that intermediate type, so it does seem a bit strange to force it to the front - but I guess it's probably more efficient in some ways?

I think we've said that we want the derived <-> base conversion to be a no-op, which would force the vptr to not be at the start of the object in this case. However, that would force a significant ABI difference from C++, which might be motivation to reconsider and accept that derived <-> base conversions might add an offset. @josh11b may have further thoughts.

How hard would this be to revisit in future?

toolchain/check/handle_class.cpp
```diff
@@ -542,6 +542,13 @@ auto HandleParseNode(Context& context, Parse::BaseDeclId node_id) -> bool {
        .index = SemIR::ElementIndex(
            context.args_type_info_stack().PeekCurrentBlockContents().size())});
 
+  if (base_info.type_id != SemIR::TypeId::Error) {
+    auto base_class_info = context.classes().Get(
+        context.types().GetAs<SemIR::ClassType>(base_info.type_id).class_id);
+    CARBON_CHECK(!class_info.is_dynamic);
+    class_info.is_dynamic = base_class_info.is_dynamic;
```
zygoloid: What happens for 
```carbon
base class A {
  virtual fn F[addr self: Self*]();
}
class B {
  virtual fn G[addr self: Self*]();
  extend base: A;
}
```
or similarly if a class attempts to derive from a dynamic base class twice? During error recovery I'm concerned that we could end up here with `class_info.is_dynamic` already set to `true`. Could we `|=` the base class flag instead of `CHECK`-failing?

toolchain/check/handle_class.cpp
```diff
@@ -651,10 +658,37 @@ static auto CheckCompleteClassType(Context& context, Parse::NodeId node_id,
     return CheckCompleteAdapterClassType(context, node_id, class_id, fields_id);
   }
 
+  bool defining_vtable_ptr = class_info.is_dynamic;
+  if (class_info.base_id.is_valid()) {
+    auto base_info = context.insts().GetAs<SemIR::BaseDecl>(class_info.base_id);
+    if (auto* base_class_info = TryGetAsClass(context, base_info.base_type_id);
```
zygoloid, line 665:
```suggestion
    // TODO: If the base class is template dependent, we will need to decide
    // whether to add a vptr as part of instantiation.
    if (auto* base_class_info = TryGetAsClass(context, base_info.base_type_id);
```

dwblaikie:
> > * Is this the right/good enough SemIR representation for now? This patch adds a `is_dynamic` attribute to `SemIR::Class` and populates/flags it based on the flag of the base class, or if any virtual function is declared in the class (or, at least that's my intent).
> 
> I think storing this flag is reasonable; it seems like a property of the class that we will want to query relatively frequently. One other question we will frequently need to answer is "where is the vptr?" -- eg, when forming a virtual call or initializing an instance of the class -- so perhaps it'd make sense to also store some information about that, when we get to the point of wanting to access the vptr.

Fair enough - yeah, can add more/change how we're doing this later. But sounds like you're on board with this as a first pass.

> > * Should the vtable be the first element in the type? If there's a non-dynamic base type, we could have a layout that's `{<non-dynamic base type>, vtable ptr, <derived members>}`? Derived types would still be able to uniquely identify where their vtable pointer is just fine... - and the vtable pointer is, in a sense, a member of that intermediate type, so it does seem a bit strange to force it to the front - but I guess it's probably more efficient in some ways?
> 
> I think we've said that we want the derived <-> base conversion to be a no-op, which would force the vptr to not be at the start of the object in this case. However, that would force a significant ABI difference from C++, which might be motivation to reconsider and accept that derived <-> base conversions might add an offset. @josh11b may have further thoughts.

Oh, fair point... I'll see about making the change now to put the vptr after the base (I expect this'll be easy enough, and can flip back/forth in this review if folks have opinions).

> How hard would this be to revisit in future?

Seems easy enough - don't have a stable ABI, so I think we can move it around as we see fit.

Mostly I'm asking pretty short-sightedly about design questions, just whatever questions/answers you folks want answered now/set us in the direction you'd like to head.

toolchain/check/handle_class.cpp
```diff
@@ -651,10 +658,37 @@ static auto CheckCompleteClassType(Context& context, Parse::NodeId node_id,
...
+          context.insts().GetAs<SemIR::StructTypeField>(fields.front());
+      object_rep_fields.push_back(
+          context.AddInstInNoBlock<SemIR::StructTypeField>(
+              Parse::NodeId::Invalid,
+              {.name_id = SemIR::NameId::Vptr,
+               .field_type_id = context.GetPointerType(field.field_type_id)}));
+    }
+  }
+
+  object_rep_fields.append(fields.begin(), fields.end());
```
zygoloid: Perhaps...
```suggestion
    object_rep_fields.append(fields.begin(), fields.end());
  }
```
given that you don't use `object_rep_fields` in the `!defining_vtable_ptr` case.

toolchain/check/handle_class.cpp
```diff
@@ -651,10 +658,37 @@ static auto CheckCompleteClassType(Context& context, Parse::NodeId node_id,
...
+  if (class_info.base_id.is_valid()) {
+    auto base_info = context.insts().GetAs<SemIR::BaseDecl>(class_info.base_id);
+    if (auto* base_class_info = TryGetAsClass(context, base_info.base_type_id);
+        base_class_info && base_class_info->is_dynamic) {
+      defining_vtable_ptr = false;
+    }
+  }
+
+  llvm::SmallVector<SemIR::InstId> object_rep_fields;
+  auto fields = context.inst_blocks().Get(fields_id);
```
zygoloid: In the caller of this function, we call `context.args_type_info_stack().Pop();` which will be the point at which the `InstBlock` is allocated. In the case where we define a vtable pointer, we never use that `InstBlock` for anything. I wonder if with some refactoring we could avoid allocating it at all? (Maybe, move the pop from the caller into this function, and add a mechanism to `InstBlockStack` to allow the vptr to be inserted in the right place before we build the block.)

toolchain/check/handle_class.cpp
```diff
@@ -651,10 +658,37 @@ static auto CheckCompleteClassType(Context& context, Parse::NodeId node_id,
...
+  if (defining_vtable_ptr) {
+    if (!fields.empty()) {
```
zygoloid: Why do we only add a vptr if we already have at least one field?

toolchain/check/handle_class.cpp
```diff
@@ -651,10 +658,37 @@ static auto CheckCompleteClassType(Context& context, Parse::NodeId node_id,
...
+      auto field =
+          context.insts().GetAs<SemIR::StructTypeField>(fields.front());
+      object_rep_fields.push_back(
+          context.AddInstInNoBlock<SemIR::StructTypeField>(
+              Parse::NodeId::Invalid,
+              {.name_id = SemIR::NameId::Vptr,
+               .field_type_id = context.GetPointerType(field.field_type_id)}));
```
zygoloid: Should this be using `VtableType`?

zygoloid (commented):
I'd like to see some `check` tests here -- just to see the object representation type in the SemIR for some simple cases. (Maybe also a `fail_todo_something` case for initializing a class with a vptr, to show that it fails -- I'd expect a "missing initializer for <vptr> field" error for now.)

zygoloid:
> I'd like to see some check tests here -- just to see the object representation type in the SemIR for some simple cases.

Looks like we do actually already have coverage of this part -- as you were :)

toolchain/sem_ir/name.cpp
```diff
@@ -25,6 +25,8 @@ static auto GetSpecialName(NameId name_id, bool for_ir) -> llvm::StringRef {
       return "package";
     case NameId::Base.index:
       return "base";
+    case NameId::Vptr.index:
+      return "vptr";
```
zygoloid: `vptr` isn't a keyword, so this might be misleading if we ever use this to print Carbon-like syntax. Maybe:
```suggestion
      return for_ir ? "vptr" : "<vptr>";
```

toolchain/check/handle_class.cpp
dwblaikie: Oh, yeah for sure - that's a broader question: What's the sort of project ethos around `CHECK`-failing, given the state of incompleteness of the project at the moment?

My very rough intent here was that we could `CHECK`-fail here, and eventually add the diagnostics in to catch the cases that would let you reach these `CHECK`-fails.

But error recovery is also a thing - some parts of the diagnostic infrastructure (like the stuff for modifiers) seems to modify (modifying the modifiers! \o/) the state so it's valid going forward - stripping off invalid modifiers at the point where they're checked for validity.

But, yeah, for the purpose of making this code less opinionated about prior error recovery - changed to `|=` and added test coverage (which should change in behavior/diagnose, but will be there to demonstrate that it continues not to `CHECK`-fail)

dwblaikie:
> (Maybe also a `fail_todo_something` case for initializing a class with a vptr, to show that it fails -- I'd expect a "missing initializer for field" error for now.)

Yep, on the money. Test added.

toolchain/check/handle_class.cpp
dwblaikie: Sure - done. (& I think that's the last of the current feedback addressed)

common/array_stack.h
```diff
@@ -67,6 +67,13 @@ class ArrayStack {
     values_.push_back(value);
   }
 
+  // Prepends a value to the top array on the stack.
+  auto PrependToTop(ValueT value) -> void {
+    CARBON_CHECK(!array_offsets_.empty(),
+                 "Must call PushArray before PushValue.");
```
zygoloid:
```suggestion
                 "Must call PushArray before PrependToTop.");
```
Can you also fix this in `AppendToTop` while you're here? :)

toolchain/check/handle_class.cpp
zygoloid: Sounds good, thanks!

In general: we (ideally) don't want it to be possible for input to trigger a CHECK fail, in part because our fuzzer will eventually find such cases. In cases where support for something is missing / broken, one tool we have available is to use `context.TODO` to report a diagnostic and then return `false` from the `Handle` function (which will cause us to bail out of the overall checking process).

common/array_stack.h
dwblaikie: Ah, yep - done!

zygoloid (approved):
Thank you, looks great!

PR #4410, opened 2024-10-15
Switch back to the system llvm-symbolizer.

jonmeow:
Undoes most of #4347, because of [performance complaints](https://discord.com/channels/655572317891461132/707150492370862090/1295527235133898772). With a 30-ish frame stack trace and `-c dbg`, my installed `llvm-symbolizer` still seems slow (~6s), but the hermetic `llvm-symbolizer` adds ~4s (i.e., ~10s total). I don't think we can easily force the hermetic version to build in opt configuration, so I'm backing it out.

### DIFF
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -129,7 +129,7 @@
   "moduleExtensions": {
     "//bazel/cc_toolchains:clang_configuration.bzl%clang_toolchain_extension": {
       "general": {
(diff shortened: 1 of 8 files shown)


geoffromer (approved):
This looks good to me, but should we consider raising the test timeout instead? I don't so much mind the waiting; the real problem is the test timeout eating the stack trace altogether.

jonmeow:
My concern is that this affects arbitrary tests, and we would need to raise timeouts across the board to some unknown value (we'd only find out on crashes). Having obscure issues turn into tech support seems undesirable.

PR #4411, opened 2024-10-16
Implement basic bool and int formatting for diagnostics

jonmeow:
Note, this supports plurals, but doesn't apply it anywhere. I'm mainly doing that to demonstrate the approach regarding syntax. See format_providers.h for details.


toolchain/check/merge.cpp
```diff
@@ -6,11 +6,35 @@
...
+template <>
+struct llvm::format_provider<Carbon::Check::ParamKind> {
+  using ParamKind = Carbon::Check::ParamKind;
+  static void format(const ParamKind& kind, raw_ostream& out,
+                     StringRef /*style*/) {
+    if (kind == ParamKind::Implicit) {
+      out << "implicit ";
+    }
+  }
+};
```
zygoloid: This doesn't look like it would support things like diagnostic translation well; could we use something like `FormatBool` instead?

toolchain/diagnostics/format_providers.h
```diff
@@ -0,0 +1,135 @@
...
+// Selects a formatv string based on the value.
+//
+// When used, the true and false outputs are separated by a `|`.
+//
+// For example, `{0:true|false}` would yield standard bool formatting.
+//
+// If needed, the _full_ style string can be wrapped with `'` in order to
+// preserve prefix or suffix whitespace (which is stripped by formatv). For
+// example, `{0:' true | false '}` retains whitespace which would be dropped
+// before `true` and after `false`.
```
zygoloid, line 23: Seems like a reasonable workaround. But the need to do this makes me sad.

toolchain/diagnostics/format_providers.h
```diff
@@ -0,0 +1,135 @@
...
+ public:
+  int value;
+};
+
+}  // namespace Carbon
+
+// See FormatBool.
+template <>
+struct llvm::format_provider<Carbon::FormatBool> {
+  static void format(const Carbon::FormatBool& wrapper, raw_ostream& out,
```
zygoloid: These `format` functions seem a bit long to include in the header, consider moving them to the `.cpp` file.

toolchain/diagnostics/format_providers.h
```diff
@@ -0,0 +1,135 @@
...
+//
+// When used, the true and false outputs are separated by a `|`.
+//
+// For example, `{0:true|false}` would yield standard bool formatting.
+//
+// If needed, the _full_ style string can be wrapped with `'` in order to
+// preserve prefix or suffix whitespace (which is stripped by formatv). For
+// example, `{0:' true | false '}` retains whitespace which would be dropped
+// before `true` and after `false`.
+struct FormatBool {
```
zygoloid: For the existing diagnostic helpers, the naming convention in use is `<thing being formatted>As<how to format it>` (eg `InstIdAsType`, `TypeIdAsRawType`). Would it make sense to be consistent with that? Eg, `BoolAsSelect`, `IntAsSelect`

I understand the desire to namespace these a bit, especially since they're effectively global (in namespace `Carbon`), but I guess `FormatInt` doesn't convey to me that it's selecting a string format based on the int value. Maybe `FormatIntAsSelect` if you still want the namespacing?

toolchain/diagnostics/format_providers.h
```diff
@@ -0,0 +1,135 @@
...
+ public:
+  bool value;
```
zygoloid, line 28: For the diagnostic helpers in check, we provide an implicit conversion from the wrapped type to make the diagnostic emission cleaner. I think I'd prefer that over `{.value = blah}`.

toolchain/parse/context.cpp
```diff
@@ -309,13 +279,16 @@ auto Context::DiagnoseOperatorFixity(OperatorFixity fixity) -> void {
...
     if (!IsLexicallyValidInfixOperator()) {
       CARBON_DIAGNOSTIC(BinaryOperatorRequiresWhitespace, Error,
-                        "whitespace missing {0} binary operator", RelativeLoc);
-      emitter_->Emit(*position_, BinaryOperatorRequiresWhitespace,
-                     tokens().HasLeadingWhitespace(*position_)
-                         ? RelativeLoc::After
-                         : (tokens().HasTrailingWhitespace(*position_)
-                                ? RelativeLoc::Before
-                                : RelativeLoc::Around));
+                        "whitespace missing {0:=-1:before|=0:around|=1:after} "
```
zygoloid, line 282: This slightly makes me want
```suggestion
                        "whitespace missing {0:<0:before|=0:around|>0:after} "
```
but it's probably not worth adding for this one use.

toolchain/diagnostics/format_providers.h
```diff
@@ -0,0 +1,135 @@
...
+    // Remove wrapping quotes if present.
+    if (style.starts_with('\'') && style.ends_with('\'')) {
+      style = style.drop_front().drop_back();
+    }
+
+    auto sep = style.find('|');
+    CARBON_CHECK(
+        sep != llvm::StringRef::npos,
+        "FormatBool requires a `|` separating true and false results: `{0}`",
+        style);
```
zygoloid: Would it make sense to also check there isn't a second `|`?

zygoloid (commented):
Thank you for working on this!

zygoloid:
On the subject of plurals, something that comes up a lot in Clang is the desire to do this kind of thing: "{0} apple{0:=1|:s}" -- notably, `{0}` is formatted both as an integer and as a string selection. I wonder if we should do the same here -- if the format specifier is empty, then format the integer as an integer. Otherwise I think for the plural cases we end up passing the same integer twice.

toolchain/diagnostics/format_providers.h
jonmeow: `FormatIntAsSelect` seems really long. `BoolAsSelect` it is.

toolchain/diagnostics/format_providers.h
jonmeow: 🤷 

As discussed, we can revisit and see what std::format does whenever we're able to switch.

toolchain/diagnostics/format_providers.h
jonmeow: Done

toolchain/parse/context.cpp
jonmeow: FWIW I would still lean towards the specific values, since they're set below. I just felt -1/0/1 was easier to mentally map here than 0/1/2, but I could switch if desired

toolchain/diagnostics/format_providers.h
jonmeow: Done

toolchain/diagnostics/format_providers.h
jonmeow: Done

toolchain/check/merge.cpp
jonmeow: Done, it's just very copy-paste right now.

I'm keeping the enum for explicitness.

jonmeow (commented):
Added support for repeat/non-styled use (PluralExample test)

toolchain/diagnostics/format_providers.h
dwblaikie: Looks like std::format passes through the string with leading/trailing whitespace to the formatter.

So I'm happy to make the LLVM change to avoid trimming this so it works as desired. I'll follow-up here with the upstream PR/commit soon.

zygoloid (approved):

toolchain/diagnostics/format_providers.h
jonmeow: We can update the implementation once LLVM has support; I'd rather not block this PR on that, to minimize merge conflicts.

dwblaikie: Posted upstream change here: https://github.com/llvm/llvm-project/pull/112625

PR #4414, opened 2024-10-16
Superficial support for `Core.BigInt` type

zygoloid:
Add a `Core.BigInt` type and a corresponding builtin type in the toolchain. See [corresponding section of the design](https://docs.carbon-lang.dev/docs/design/expressions/literals.html#defined-types).

So far this type is not used for anything, and there is no way to create an instance of it.


jonmeow:
Can you note in the PR description why you're doing this? I assume it's support for something, but it's non-obvious to me.

core/prelude/types.carbon
```diff
@@ -11,6 +11,7 @@ export import library "prelude/types/bool";
 // noise in the toolchain tests.
 // import library "prelude/types/i32";
 
+fn BigInt() -> type = "bigint.make_type";
```
jonmeow:
```suggestion
fn BigInt() -> type = "big_int.make_type";
```

Suggesting for the underscore between words.

zygoloid:
> Can you note in the PR description why you're doing this? I assume it's support for something, but it's non-obvious to me.

Added a link to the design. Did you want more than that? (Is your question why at all, or why now?)

jonmeow (approved):
LGTM, though would appreciate more information in the PR description before merging. [ed: I see this is now addressed]

core/prelude/types.carbon
zygoloid: Done throughout.

PR #4415, opened 2024-10-16
Don't deduce values for explicitly-specified generic bindings.

zygoloid:
Distinguish between deduction against a symbolic binding pattern and deduction against a symbolic binding name. In the former case, the value is being explicitly specified and must be constant. In the latter case we encountered a use of the binding name as a subexpression, and should deduce against it if it's not explicitly specified.


toolchain/check/deduce.cpp
```diff
@@ -246,54 +270,68 @@ auto DeductionContext::Deduce() -> bool {
...
+        auto index = entity_name.bind_index;
+        if (!index.is_valid()) {
+          break;
+        }
+        CARBON_CHECK(
+            index >= first_deduced_index_ &&
+                static_cast<size_t>(index.index) < result_arg_ids_.size(),
+            "Unexpected index {0} for symbolic binding pattern; "
+            "expected to be in range [{1}, {2})",
+            first_deduced_index_.index, result_arg_ids_.size());
```
geoffromer: It looks like the format string wants 3 arguments, but I only see 2.

toolchain/check/deduce.cpp
```diff
@@ -246,54 +270,68 @@ auto DeductionContext::Deduce() -> bool {
...
-      case SemIR::InstKind::BindSymbolicName: {
-        auto entity_name_id = SemIR::EntityNameId::Invalid;
-        if (auto bind = param_inst.TryAs<SemIR::SymbolicBindingPattern>()) {
-          entity_name_id = bind->entity_name_id;
-        } else {
-          entity_name_id =
-              param_inst.As<SemIR::BindSymbolicName>().entity_name_id;
+      // Deducing a symbolic binding pattern from an argument deduces the
+      // binding as having that constant value. For example, deducing
+      // `(T:! type)` against `(i32)` deduces `T` to be `i32`. The argument is
```
geoffromer: This seems like an example of initialization rather than deduction per se. Is that always true for this case? If so, I think it would help to spell that out a little more explicitly.

toolchain/check/testdata/struct/import.carbon
```diff
@@ -30,9 +30,13 @@ var c: C({.a = 1, .b = 2}) = F();
 // --- fail_bad_type.impl.carbon
 
 impl package Implicit;
-// CHECK:STDERR: fail_bad_type.impl.carbon:[[@LINE+4]]:14: error: missing value for field `a` in struct initialization
+// CHECK:STDERR: fail_bad_type.impl.carbon:[[@LINE+8]]:14: error: missing value for field `a` in struct initialization
 // CHECK:STDERR: var c_bad: C({.c = 1, .d = 2}) = F();
 // CHECK:STDERR:              ^~~~~~~~~~~~~~~~
+// CHECK:STDERR: fail_bad_type.impl.carbon:[[@LINE-4]]:6: in import
```
geoffromer, line 36: This is out of scope for the current PR, but I wonder if we should omit this message on notes; the phrasing in this context is confusing, and it doesn't seem like it adds anything.

toolchain/check/deduce.cpp
zygoloid: Thanks, fixed, and tested manually by forcing the CHECK to fail.

toolchain/check/deduce.cpp
zygoloid: Yes, it is, changed the comment to try to make that clearer.

toolchain/check/testdata/struct/import.carbon
zygoloid: Yeah, that might be an improvement.

geoffromer (approved):

PR #4416, opened 2024-10-16
Remove uses of StringLiteral in format strings.

jonmeow:
Building on #4411, avoid using StringLiteral in format strings. This includes a diagnostic check to prevent regressions (which is also how I gathered issues).

Note, I haven't looked at `std::string` uses yet, but we might need things like that to be able to pass strings in code back to the user. StringLiteral though means that it's literally written down in the toolchain, at which point it should probably be written in the format string instead of separately.

### DIFF
--- a/toolchain/check/call.cpp
+++ b/toolchain/check/call.cpp
@@ -20,7 +20,7 @@
 namespace Carbon::Check {
 
 namespace {
-// Entity kinds for the diagnostic. Converted to an int for a select.
+// Entity kinds, for diagnostics. Converted to an int for a select.
 enum class EntityKind : uint8_t {
   Function,
   GenericClass,
--- a/toolchain/check/call.cpp
+++ b/toolchain/check/call.cpp
@@ -22,9 +22,9 @@ namespace Carbon::Check {
 namespace {
 // Entity kinds, for diagnostics. Converted to an int for a select.
 enum class EntityKind : uint8_t {
-  Function,
-  GenericClass,
-  GenericInterface,
+  Function = 0,
+  GenericClass = 1,
+  GenericInterface = 2,
 };
 }  // namespace
 
--- a/toolchain/check/call.cpp
+++ b/toolchain/check/call.cpp
@@ -19,6 +19,15 @@
 
 namespace Carbon::Check {
 
+namespace {
+// Entity kinds for the diagnostic. Converted to an int for a select.
+enum class EntityKind : uint8_t {
+  Function,
+  GenericClass,
+  GenericInterface,
+};
+}  // namespace
+
 // Resolves the callee expression in a call to a specific callee, or diagnoses
 // if no specific callee can be identified. This verifies the arity of the
 // callee and determines any compile-time arguments, but doesn't check that the
@@ -31,7 +40,7 @@ namespace Carbon::Check {
 // callee is not generic, or `nullopt` if an error has been diagnosed.
 static auto ResolveCalleeInCall(Context& context, SemIR::LocId loc_id,
                                 const SemIR::EntityWithParamsBase& entity,
-                                llvm::StringLiteral entity_kind_for_diagnostic,
+                                EntityKind entity_kind_for_diagnostic,
                                 SemIR::GenericId entity_generic_id,
                                 SemIR::SpecificId enclosing_specific_id,
                                 SemIR::InstId self_id,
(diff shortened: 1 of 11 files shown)


toolchain/check/call.cpp
```diff
@@ -18,6 +19,15 @@
 
 namespace Carbon::Check {
 
+namespace {
+// Entity kinds for the diagnostic. Converted to an int for a select.
```
geoffromer:
```suggestion
// Entity kinds, for diagnostics. Converted to an int for a select.
```

toolchain/check/call.cpp
```diff
@@ -18,6 +19,15 @@
...
+enum class EntityKind : uint8_t {
+  Function,
+  GenericClass,
+  GenericInterface,
```
geoffromer:
```suggestion
  Function = 0,
  GenericClass = 1,
  GenericInterface = 2,
```

Since the integer values are significant, we should probably declare them explicitly.

toolchain/check/call.cpp
```diff
@@ -42,16 +52,20 @@ static auto ResolveCalleeInCall(Context& context, SemIR::LocId loc_id,
   auto params = context.inst_blocks().GetOrEmpty(callee_info.param_refs_id);
   if (arg_ids.size() != params.size()) {
     CARBON_DIAGNOSTIC(CallArgCountMismatch, Error,
-                      "{0} argument(s) passed to {1} expecting "
-                      "{2} argument(s).",
-                      int, llvm::StringLiteral, int);
-    CARBON_DIAGNOSTIC(InCallToEntity, Note, "calling {0} declared here",
-                      llvm::StringLiteral);
+                      "{0} argument(s) passed to "
+                      "{1:=0:function|=1:generic class|=2:generic interface}"
```
geoffromer, line 56: In this case, I'm not sure I agree with the thesis that the string "should probably be written in the format string instead of separately". This change makes it possible to spot bugs in the diagnostic's phrasing by local inspection, but the tradeoff is that it makes it noticeably harder for me to read the format string as a whole, and it introduces a new point of failure: the int -> kind mapping in the format string might not match the kind -> int mapping in the `enum` (which can't be spotted by local inspection).

If you want to make sure phrasing problems can be spotted locally, I'd suggest that we stick with the previous format string, but map the enum to a string literal locally. That way all the components of the message are available locally, and there's no potentially error-prone indirection through integers.

jonmeow:
NB, rebased due to conflicting changes in #4423, figured you probably hadn't looked at this so probably easier to merge now.

toolchain/diagnostics/diagnostic.h
```diff
@@ -119,8 +119,10 @@ struct DiagnosticBase {
   explicit constexpr DiagnosticBase(DiagnosticKind kind, DiagnosticLevel level,
                                     llvm::StringLiteral format)
       : Kind(kind), Level(level), Format(format) {
-    static_assert((... && !std::is_same_v<Args, llvm::StringRef>),
-                  "Use std::string or llvm::StringLiteral for diagnostics to "
+    static_assert((... && !(std::is_same_v<Args, llvm::StringRef> ||
+                            std::is_same_v<Args, llvm::StringLiteral>)),
+                  "For diagnostics, use a format provider (see "
+                  "toolchain/diagnostics/format_providers.h) or std::string to "
                   "avoid lifetime issues.");
```
geoffromer: String literals don't have lifetime issues, so we probably need a separate error message for that.

toolchain/check/call.cpp
jonmeow: This isn't a phrasing issue; it's one of translatability. For example, let's say someone wants to translate diagnostics to French. If "generic" is part of the hardcoded inputs, how does someone provide a translation for it?

In this model, someone just needs to provide an appropriate format string, and there should be sufficient context that they can adjust ordering as needed.

To be clear, it's not the ability to spot bugs. It's avoiding fundamentally English-centric design.

jonmeow: Note too, we'd still need to add the ability to replace [ed: format] strings (that's just not here right now). But the intent here is to stop [ed: or at least reduce] building up technical debt in terms of diagnostics that would be incompatible with translation.

toolchain/diagnostics/diagnostic.h
jonmeow: Changed the error message to just point at the diagnostic parameter type advice.

And fixed the StringLiteral note there.

toolchain/check/call.cpp
geoffromer: I see; that makes sense. I'm still concerned about the risk of bugs from the indirection through integers, and I feel like we can address that in a way that's consistent with translatability, but I'm happy to treat that as future work.

toolchain/docs/diagnostics.md
```diff
@@ -167,9 +167,11 @@ methods for formatting arguments:
     -   This includes `char` and integer types (`int`, `int32_t`, and so on).
     -   String types can be added as needed, but stringifying values using the
         methods noted below is preferred.
-        -   Use `llvm::StringLiteral` where appropriate; use `std::string` when
-            allocations are required.
+        -   Use `std::string` when allocations are required.
         -   `llvm::StringRef` is disallowed due to lifetime issues.
+        -   `llvm::StringLiteral` is disallowed because format providers such as
+            `BoolAsSelect` should work in cases where a `StringLiteral` could be
+            used.
```
geoffromer:
```suggestion
            used, and because string literal parameters tend to make
            the resulting diagnostics hard to translate.
```

When someone's blocked from using a string literal, I think they're more likely to choose the right alternative if they understand the translatability goal.

geoffromer (approved):

PR #4418, opened 2024-10-17
Refactor modifier formatting to remove string passing.

jonmeow:
I'm taking the approach of making DiagnosticBase an API so that we can pass similar diagnostics as parameters. An alternative would be to do the function_ref approach we've done elsewhere, but these felt more boilerplate to me.

Note I'm also modifying messages here. Let me know if you'd like different changes and/or just keeping current formatting (keeping current formatting would also allow removing some of the templating I've added, but it felt helpful putting explicit tokens where possible). But also, things like "`protected` not allowed on `interface` declaration at file scope" were part of the phrasing issue, I think.


toolchain/check/modifiers.cpp
```diff
@@ -50,28 +77,49 @@ auto ForbidModifiersOnDecl(Context& context, DeclIntroducerState& introducer,
...
+                    Lex::TokenKind);
+  ForbidModifiersOnDecl(context, ModifierNotAllowedOnDeclaration, introducer,
+                        ~allowed);
+}
+
+auto LimitModifiersOnNotDefinition(Context& context,
+                                   DeclIntroducerState& introducer,
+                                   KeywordModifierSet allowed) -> void {
+  CARBON_DIAGNOSTIC(ModifierOnlyAllowedOnDefinition, Error,
+                    "`{0}` not allowed on `{1}` declaration, only definition",
```
zygoloid: I'm a bit uncomfortable about implying that a definition isn't a kind of declaration. How would you feel about
```suggestion
                    "`{0}` not allowed on `{1}` forward declaration, only definition",
```
or maybe
```suggestion
                    "`{0}` not allowed on `{1}` non-defining declaration",
```
That said, this is entirely a drive-by (the phrasing here is unchanged from what we had before) so I'm happy for this PR to go ahead as-is and we can maybe consider this rephrasing separately.

zygoloid (approved):

PR #4420, opened 2024-10-17
More focused diagnostic notes for parameters

geoffromer:


geoffromer:
This is motivated by some planned refactoring in `convert.cpp`, but IMO the resulting diagnostics also have better ergonomics, so it's a win/win.

zygoloid (approved):

PR #4421, opened 2024-10-17
Remove a couple std::string uses in diagnostics.

jonmeow:
We can't completely remove std::string from diagnostics because it's probably better to provide a string than something like a StringLiteralValueId or NameId (because those may be opaque for someone trying to present the diagnostic). So this change is just playing whackamole on another couple things that could easily use the new format providers.


jonmeow:
(shifting reviewers to try to balance load)

geoffromer (approved):

PR #4422, opened 2024-10-17
Restructure handling of runtime parameters

geoffromer:
- Generate runtime indices as part of pattern matching, rather than as a separate postprocessing/rewriting step.
- In contexts where runtime parameters aren't permitted, avoid emitting insts for them to begin with, rather than trying to detect the problem and rewrite the IR to remove them later on.


toolchain/check/pattern_match.cpp
```diff
@@ -55,22 +56,57 @@ class MatchContext {
...
+  // Allocates an InstBlock containing the inner parameters recorded by
+  // RecordInnerParam since the last call to this function (if any), and returns
+  // its ID. Valid only during callee pattern matching.
+  auto ConsumeInnerParams(Context& context) -> SemIR::InstBlockId {
+    CARBON_CHECK(stack_.empty());
+    CARBON_CHECK(kind_ == MatchKind::Callee);
+    auto block_id = context.inst_blocks().Add(inner_param_insts_);
+    inner_param_insts_.clear();
+    return block_id;
+  }
```
zygoloid: If I'm reading the code below correctly, what you're calling "inner parameters" here are the bindings in the pattern. Is that right? If so, would it make sense to use that terminology here, or are you planning for this to change to instead be a list of parameters? (Either way, I think it'd be good for the comment on `RecordInnerParam` to clarify, perhaps with an example.)

toolchain/sem_ir/ids.h
```diff
@@ -281,17 +281,26 @@ struct RuntimeParamIndex : public IndexBase,
...
+    } else {
+      IndexBase::Print(out);
+    }
   }
 };
 
 constexpr RuntimeParamIndex RuntimeParamIndex::Invalid =
     RuntimeParamIndex(InvalidIndex);
 
+constexpr RuntimeParamIndex RuntimeParamIndex::Unknown = RuntimeParamIndex(-2);
```
zygoloid: Elsewhere we fairly consistently use `InvalidIndex - 1` for this kind of thing rather than assuming that `InvalidIndex` has a particular value. (I think we do end up assuming that in practice in various places but it seems good to be consistent.)
```suggestion
constexpr RuntimeParamIndex RuntimeParamIndex::Unknown =
    RuntimeParamIndex(InvalidIndex - 1);
```

zygoloid (approved):

toolchain/check/pattern_match.cpp
geoffromer: Yeah, I think I was way overthinking this; in practice an inner parameter is always a name binding and vice versa, so there's no need for a separate term. Is this clear enough now?

toolchain/check/pattern_match.cpp
```diff
@@ -141,13 +141,14 @@ class MatchContext {
...
   if (entry.pattern_id == SemIR::InstId::BuiltinError) {
-    match.RecordInnerParam(SemIR::InstId::BuiltinError);
+    match.RecordBindName(SemIR::InstId::BuiltinError);
     return;
   }
   auto pattern = context.insts().GetWithLocId(entry.pattern_id);
   CARBON_KIND_SWITCH(pattern.inst) {
     case SemIR::BindingPattern::Kind:
     case SemIR::SymbolicBindingPattern::Kind: {
+      CARBON_CHECK(match.kind() == MatchKind::Callee);
```
geoffromer: Note I also added this check, which was left out of the PR by accident.

zygoloid (approved):

toolchain/check/pattern_match.cpp
zygoloid: Yeah, that works for me. Thanks!

toolchain/sem_ir/file.cpp
```diff
@@ -291,6 +291,12 @@ auto GetExprCategory(const File& file, InstId inst_id) -> ExprCategory {
       }
 
       case CARBON_KIND(BindName inst): {
+        // TODO: don't rely on value_id for expression category, since it may
```
geoffromer, line 294: @zygoloid please take another look. I had to add this pretty gross hack to work around a crash bug that was exposed by a new test case in `check/testdata/class/self.carbon`.

(The changes in the other files are just routine merging as far as I know.)

zygoloid: I see. I guess we would want the `BindName` inst to track whether it's in a `var` context or not, maybe with two different inst kinds, but this seems like an acceptable short-term workaround.

zygoloid (approved):

PR #4425, opened 2024-10-17
Support printing the diagnostic kind for verification.

jonmeow:
This is to help identify which diagnostics we're actually using.

Note that driver/testdata still has tests which don't pass this flag, and so continue to test the kind-less (default) behavior.


dwblaikie (approved):
Looks good to me

PR #4426, opened 2024-10-17
Add tests to catch untested diagnostics.

jonmeow:
Use the diagnostic kind printing in #4425 to catch when we have diagnostics with no tests.

This merges a couple other use-cases of filegroup manifests into a common rule.

Note I do add a few tests for things, and also some things are _actually_ unit tested (just not in the file_test structure). But I stopped when I realized that dealing with merge conflicts is going to be a pain. I might end up reverting test changes (as part of merge conflict resolution) and doing narrow test additions in a separate PR, after both this and #4425 are merged.


dwblaikie (approved):
Looks plausible, though I'm admittedly not quite following some of the details, like the unexpected matches & I'm not so fluent with bazel build files - so might be worth a second set of eyes, but your call.

PR #4428, opened 2024-10-17
Split out clang-tidy to not run in merge

jonmeow:
Because clang-tidy is slow (and I'm not sure we can make it really fast), trying to run it slightly less. Also, I noticed we can shave a few minutes by disabling apt removal without losing too much free space.

Note that since this removes the old clang-tidy, I'll need to change the branch protections before merging.


.github/actions/build-setup-ubuntu/action.yml
```diff
@@ -18,6 +18,9 @@ runs:
         android: true
         dotnet: true
         haskell: true
+        # Without this, we save about 22GB in seconds. It adds minutes and only
```
zygoloid: I found this comment confusing; does this setting affect the saving of 22GB? If not, could we instead just mention what this does, eg "Enabling this costs several minutes and only saves about 4GB."

.github/actions/test-setup/action.yml
```diff
@@ -0,0 +1,132 @@
...
+    - name: Compute impacted pull request targets (for push)
+      if: steps.filter.outputs.has_code == 'true' && github.event_name == 'push'
+      shell: bash
+      env:
+        TARGETS_FILE: ${{ inputs.targets_file }}
+      run: |
+        echo "//..." >$TARGETS_FILE
+
+    # Compute the set of possible rules impacted by this change using
+    # Bazel-based diffing. This lets PRs and the merge queue have a much more
```
zygoloid, line 102: Is "and the merge queue" accurate? I thought the merge queue used the "push" event, so it'll hit the previous case

.github/workflows/clang_tidy.yaml
```diff
@@ -0,0 +1,80 @@
...
+        if: steps.test-setup.outputs.has_code == 'true'
+        env:
+          TARGETS_FILE: ${{ runner.temp }}/targets
+        run: |
+          ./scripts/run_bazel.py \
+            --attempts=5 \
+            build --config=clang-tidy -k \
+            --target_pattern_file=$TARGETS_FILE
+
+      # See "Disk space before build".
```
zygoloid:
```suggestion
      # See "Disk space before build" in the test setup action.
```
or something?

.github/workflows/tests.yaml
```diff
@@ -55,121 +51,24 @@ jobs:
...
             storage.googleapis.com:443
 
-      # Checkout the pull request head or the branch.
-      - name: Checkout pull request
-        if: github.event_name == 'pull_request'
-        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
-        with:
-          ref: ${{ github.event.pull_request.head.sha }}
-
       - name: Checkout branch
```
zygoloid: Do we not need to specify the SHA for a PR? Was that simply redundant?

.github/workflows/tests.yaml
```diff
@@ -184,23 +83,7 @@ jobs:
...
-          'clang-tidy'
-        env:
-          TARGETS_FILE: ${{ runner.temp }}/targets
-        run: |
-          ./scripts/run_bazel.py \
-            --attempts=5 \
-            build --config=clang-tidy -k \
-            --target_pattern_file=$TARGETS_FILE
-
       # See "Disk space before build".
```
zygoloid: Same here.

zygoloid (approved):
LG

.github/workflows/tests.yaml
jonmeow: The checkout with a sha was just moved to test-setup (i.e., choosing to checkout again in order to reduce duplication). But actually, looking at the history and behavior, I think this is just cargo culting -- it was added by #603, which used pull_request_target, which we don't anymore. Everything seems fine without it, so removing.

For reference (with a small code edit):
- tidy on push: https://github.com/carbon-language/carbon-lang/actions/runs/11410891835/job/31754107266
- test on push: https://github.com/carbon-language/carbon-lang/actions/runs/11410891832/job/31754107221
- tidy on PR: https://github.com/carbon-language/carbon-lang/actions/runs/11410892118/job/31754108277
- test on PR: https://github.com/carbon-language/carbon-lang/actions/runs/11410892127/job/31754119765

Note there, I'm primarily verifying that test execution all looks correct, caching is working, etc.

.github/actions/build-setup-ubuntu/action.yml
jonmeow: Rephrased comments.

.github/actions/test-setup/action.yml
jonmeow: I'm pretty sure merge queue uses `merge_group`; the list of events is at https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows. At the top of tests.yaml, there's:

```
on:
  push:
    branches: [trunk, action-test]
  pull_request:
  merge_group:
```

Since this is `github.event_name != 'push'`, that leaves `pull_request` and `merge_group`.

Also note that `base_sha` is calculated differently based on the source:

```
          base_sha:
            ${{ github.event_name == 'pull_request' &&
            github.event.pull_request.base.sha ||
            github.event.merge_group.base_sha }}
```

Note, you can also compare with `if: ${{ github.event_name == 'merge_group' }}` in https://github.com/orgs/community/discussions/43988#discussioncomment-9667914

.github/workflows/clang_tidy.yaml
jonmeow: Done

.github/workflows/tests.yaml
jonmeow: Done

jonmeow:
Since I've made a few edits here, not sure if you want to look again.

PR #4429, opened 2024-10-19
Issue a diagnostic if we try to parse a source file that is too large.

zygoloid:
Previously in an optimized build we'd produce bogus tokens, such as tokens with incorrect IdentifierIds, and in a debug build we would try to CHECK-fail -- but actually wouldn't, because we're incorrectly checking for `2 << bits` instead of `1 << bits`. I hit this while I was trying to do some profiling and was seeing some very strange diagnostics.

The diagnostic is pointed at the first token that is beyond the limit to help people determine where to split their files.


toolchain/lex/tokenized_buffer_test.cpp
```diff
@@ -1107,6 +1107,20 @@ TEST_F(LexerTest, DiagnosticUnrecognizedChar) {
...
+  Testing::MockDiagnosticConsumer consumer;
+  static constexpr size_t NumLines = 10'000'000;
+  std::string input;
+  input.reserve(NumLines * 3);
+  for (int _ [[maybe_unused]] : llvm::seq(NumLines)) {
+    input += "{}\n";
+  }
+  EXPECT_CALL(consumer, HandleDiagnostic(IsSingleDiagnostic(
+                            DiagnosticKind::TooManyTokens,
+                            DiagnosticLevel::Error, 0x400000, 2, _)));
```
jonmeow: `0x400000` seems opaque in its origin (I'm looking back and forth between files and it _seems_ correct to assume it's PayloadBits, but it's really making me think, particularly due to the switch from bit shift to hex). Had you considered either refactoring the `1 << TokenizedBuffer::TokenInfo::PayloadBits` so that it's shared here, or not asserting the specific value? e.g., there could be a `TokenizedBuffer::MaxTokens`.

toolchain/lex/tokenized_buffer.h
```diff
@@ -396,8 +396,10 @@ class TokenizedBuffer : public Printable<TokenizedBuffer> {
           has_leading_space_(has_leading_space),
           token_payload_(payload),
           byte_offset_(byte_offset) {
-      CARBON_DCHECK(payload >= 0 && payload < (2 << PayloadBits),
-                    "Payload won't fit into unsigned bit pack: {0}", payload);
+      // Note that we don't check the payload fits in `token_payload_` here.
+      // Payloads are typically ID types for which we create at most one per
+      // token, and in `Lexer::LexFileEnd` we diagnose if the file has too many
+      // tokens to fit in `token_payload_`.
```
jonmeow: Had you considered having the above comments refer to this, since you provide more detail here? Moving it to either `LexFileEnd` or `token_payload_` might make it easier to cross-reference.

Another related thought would be to migrate the logic to something like `TokenizedBuffer::Verify`, similar to `Parse::Tree::Verify`, as another way to give an easy cross-reference (I don't think it's complex enough to have complexity motivate a refactor).

toolchain/lex/tokenized_buffer_test.cpp
```diff
@@ -1107,6 +1107,20 @@ TEST_F(LexerTest, DiagnosticUnrecognizedChar) {
   compile_helper_.GetTokenizedBuffer("\b", &consumer);
 }
 
+TEST_F(LexerTest, DiagnosticFileTooLarge) {
+  Testing::MockDiagnosticConsumer consumer;
+  static constexpr size_t NumLines = 10'000'000;
+  std::string input;
+  input.reserve(NumLines * 3);
+  for (int _ [[maybe_unused]] : llvm::seq(NumLines)) {
```
jonmeow:
```suggestion
  for ([[maybe_unused]] int _ : llvm::seq(NumLines)) {
```

I assume order doesn't matter to the compiler, but glancing at other locations, I think we're doing this order.

toolchain/lex/tokenized_buffer.h
```diff
@@ -306,7 +306,7 @@ class TokenizedBuffer : public Printable<TokenizedBuffer> {
     }
     auto set_ident_id(IdentifierId ident_id) -> void {
       CARBON_DCHECK(kind() == TokenKind::Identifier);
-      CARBON_DCHECK(ident_id.index < (2 << PayloadBits));
+      // Overflow is checked in `Lexer::LexFileEnd`.
```
jonmeow: Is it worth noting somewhere that this relies on no identifiers being added to the shared value store outside this specific tokenized buffer?

i.e., this should work because the identifier count is limited by the token count. However, if there's ever a second source of identifiers, this breaks.

jonmeow (approved):

toolchain/lex/tokenized_buffer_test.cpp
zygoloid: It's actually one less than `PayloadBits` because we have two tokens per line. My thinking was that I wanted the test to fail if someone changed `PayloadBits` so that their attention would be drawn to the fact that they're modifying the maximum number of tokens supported (which is otherwise very implicit here). But I think adding a `MaxTokens` would deal with this nicely; I'll do that.

Also I just realized the `2` here is off by one if we want to point at the first "bad" token, because we need to leave room for the end-of-file token.

jonmeow:
Note, failing test is that this should be added to IsUntestedDiagnostic in emitted_diagnostics_test.cpp

toolchain/lex/tokenized_buffer_test.cpp
zygoloid: Done.

toolchain/lex/tokenized_buffer.h
zygoloid: Moved comment to `token_payload_`. I've also factored out the finalization steps into a separate function on the lexer.

toolchain/lex/tokenized_buffer.h
zygoloid: Added a comment to `Lexer::Finalize`.

toolchain/lex/lex.cpp
```diff
@@ -1342,11 +1349,29 @@ auto Lexer::LexFileEnd(llvm::StringRef source_text, ssize_t position) -> void {
...
     DiagnoseAndFixMismatchedBrackets();
   }
+
+  // Reject source files with so many tokens that we may have exceeded the
+  // number of bits in `token_payload_`.
+  //
+  // Note that we rely on this check also catching the case where there are too
+  // many identifiers to fit an `IdentifierId` into a `token_payload_`, and
+  // likewise for `IntId` and so on. If we start adding any of those IDs prior
+  // to lexing, we may need to also limit the number of identifiers etc. here.
```
jonmeow:
```suggestion
  // to lexing, we may need to also limit the number of IDs here.
```

The period "etc." throws my reading, maybe we can back-reference "any of those IDs"?

toolchain/lex/tokenized_buffer.h
```diff
@@ -395,18 +396,22 @@ class TokenizedBuffer : public Printable<TokenizedBuffer> {
...
-    }
+          byte_offset_(byte_offset) {}
 
     // A bitfield that encodes the token's kind, the leading space flag, and the
     // remaining bits in a payload. These are encoded together as a bitfield for
     // density and because these are the hottest fields of tokens for consumers
     // after lexing.
     TokenKind::RawEnumType kind_ : sizeof(TokenKind) * 8;
     bool has_leading_space_ : 1;
+    // Payload values are typically ID types for which we create at most one per
```
jonmeow, line 406:
```suggestion

    // Payload values are typically ID types for which we create at most one per
```

Whitespace suggestion due to the long comment.

toolchain/lex/tokenized_buffer.h
```diff
@@ -395,18 +396,22 @@ class TokenizedBuffer : public Printable<TokenizedBuffer> {
...
+    // token, so we ensure that `token_payload_` is large enough to fit any
+    // token index. Stores to this field may overflow, but we produce an error
+    // in `Lexer::LexFileEnd` if the file has more than `MaxTokens` tokens, so
+    // this value never overflows if lexing succeeds.
     unsigned token_payload_ : PayloadBits;
+    static_assert(MaxTokens <= 1 << PayloadBits,
+                  "Not enough payload bits to store a token index");
```
jonmeow, line 415: What made you choose this approach versus relocating PayloadBits to make MaxTokens directly calculated based on it?

jonmeow (commented):
LG

toolchain/lex/tokenized_buffer.h
```diff
@@ -395,18 +396,22 @@ class TokenizedBuffer : public Printable<TokenizedBuffer> {
...
     // A bitfield that encodes the token's kind, the leading space flag, and the
     // remaining bits in a payload. These are encoded together as a bitfield for
     // density and because these are the hottest fields of tokens for consumers
     // after lexing.
     TokenKind::RawEnumType kind_ : sizeof(TokenKind) * 8;
     bool has_leading_space_ : 1;
+    // Payload values are typically ID types for which we create at most one per
+    // token, so we ensure that `token_payload_` is large enough to fit any
+    // token index. Stores to this field may overflow, but we produce an error
+    // in `Lexer::LexFileEnd` if the file has more than `MaxTokens` tokens, so
```
jonmeow:
```suggestion
    // in `Lexer::Finalize` if the file has more than `MaxTokens` tokens, so
```

toolchain/lex/lex.cpp
zygoloid: Maybe even more explicitly:
```suggestion
  // to lexing, we may need to also limit the number of those IDs here.
```

toolchain/lex/tokenized_buffer.h
zygoloid: This makes the previous comment that's trying to talk about all three bit-fields less clear.

What do you think about moving this comment before all the bit-fields, with the other one?

toolchain/lex/tokenized_buffer.h
zygoloid: I wanted an explicit value for `MaxTokens` so that anyone changing that value knows they're changing the token limit, not just some representation detail of the tokenized buffer.

PR #4430, opened 2024-10-20
Speed up type literal lexing and make it more strict.

chandlerc:
This rejects type literals with more digits than we can lex without APInt's help, and using a custom diagnostic. This is a pretty arbitrary implementation limit, I'm wide open to even more strict rules here.

Despite no special casing and a very simplistic approach, by not using APInt this completely eliminates the lexing overhead for `i32` in the generated compilation benchmark where that specific type literal is very common. We see a 10% improvement in lexing there:
```
BM_CompileAPIFileDenseDecls<Phase::Lex>/256        39.0µs ± 4%  34.8µs ± 2%  -10.86%  (p=0.000 n=19+20)
BM_CompileAPIFileDenseDecls<Phase::Lex>/1024        180µs ± 1%   158µs ± 2%  -12.22%  (p=0.000 n=18+20)
BM_CompileAPIFileDenseDecls<Phase::Lex>/4096        731µs ± 2%   641µs ± 1%  -12.31%  (p=0.000 n=18+20)
BM_CompileAPIFileDenseDecls<Phase::Lex>/16384      3.20ms ± 2%  2.86ms ± 2%  -10.47%  (p=0.000 n=18+19)
BM_CompileAPIFileDenseDecls<Phase::Lex>/65536      13.8ms ± 1%  12.4ms ± 2%   -9.78%  (p=0.000 n=18+19)
BM_CompileAPIFileDenseDecls<Phase::Lex>/262144     64.0ms ± 2%  58.4ms ± 2%   -8.70%  (p=0.000 n=19+18)
```

This starts to fix a TODO in the diagnostic for these by giving a reasonably good diagnostic about a very large type literal. However, in practice it regresses the diagnostics because error tokens produce noisy extraneous diagnostics from parse and check currently. Leaving the TODO there, and I have a follow-up PR to start improving the extraneous diagnostics.


toolchain/lex/tokenized_buffer_test.cpp
```diff
@@ -1028,10 +1028,10 @@ TEST_F(LexerTest, TypeLiteralTooManyDigits) {
   code.append(Count, '9');
 
   Testing::MockDiagnosticConsumer consumer;
-  EXPECT_CALL(consumer,
-              HandleDiagnostic(IsSingleDiagnostic(
-                  DiagnosticKind::TooManyDigits, DiagnosticLevel::Error, 1, 2,
```
jonmeow, line 1033: It looks like you're removing the test of `TooManyDigits`, but not the diagnostic itself (i.e., nothing else is actually making sure it sees this specific diagnostic). Can you please keep testing it explicitly?

toolchain/lex/lex.cpp
```diff
@@ -1179,17 +1175,43 @@ auto Lexer::LexWordAsTypeLiteralToken(llvm::StringRef word, int32_t byte_offset)
       return LexResult::NoMatch();
   };
 
+  // No leading zeros allowed.
+  if ('1' > word[1] || word[1] > '9') {
+    return LexResult::NoMatch();
+  }
+
   llvm::StringRef suffix = word.substr(1);
-  if (!CanLexInt(emitter_, suffix)) {
```
jonmeow, line 1190: Looking at other callers for CanLexInt, should string literals be sharing the logic? i.e., should CanLexInt be factored into just numeric literals?

toolchain/lex/lex.cpp
```diff
@@ -1179,17 +1175,43 @@ auto Lexer::LexWordAsTypeLiteralToken(llvm::StringRef word, int32_t byte_offset)
...
+  constexpr ssize_t DigitLimit = std::numeric_limits<int64_t>::digits10;
```
jonmeow: You use `int64_t` here but `suffix_value` is `ssize_t`. Should the same type be used?

jonmeow: Can you test this with a boundary value, such as `9` repeated for the number of digits? Is that handled right now? i.e., something that fits within the digit count for `ssize_t`, but will overflow a `ssize_t` if lexed.

toolchain/lex/lex.cpp
```diff
@@ -1179,17 +1175,43 @@ auto Lexer::LexWordAsTypeLiteralToken(llvm::StringRef word, int32_t byte_offset)
...
+
+  // It's tempting to do something more clever because we know the length ahead
+  // of time, but we expect these to be short (1-3 digits) and profiling doesn't
+  // show the loop as hot in the short cases.
+  ssize_t suffix_value = suffix[0] - '0';
+  for (char c : suffix.drop_front()) {
+    if (!IsDecimalDigit(c)) {
+      return LexResult::NoMatch();
+    }
+    suffix_value = suffix_value * 10 + (c - '0');
```
jonmeow: I'm trying to see if there's some pre-existing implementation we could use, maybe `std::from_chars(suffix.begin(), suffix.end(), suffix_value);`?

```suggestion
  // It's tempting to do something more clever because we know the length ahead
  // of time, but we expect these to be short (1-3 digits) and profiling doesn't
  // show the loop as hot in the short cases.
  if (std::from_chars(suffix.begin(), suffix.end(), suffix_value) != suffix.end()) {
    return LexResult::NoMatch();
```

toolchain/lex/tokenized_buffer_test.cpp
jonmeow: Eh, maybe I'll just need to look at this anyways, judging by #4426 already not seeing a test for TooManyDigits.

chandlerc: I think it's still covered by `numeric_literal_test.cpp`, but as you noted, not file test yet. I think this is still ok in this PR?

toolchain/lex/lex.cpp
chandlerc: The API here is a bit awkward, but found a way to make it work.

toolchain/lex/lex.cpp
chandlerc: Yeah, that was an accident. I've restructured the code so it constructively uses the right type.

chandlerc: Good idea, added a test that adds 9s successively until it is rejected, and checks the value round trips each time prior to that. I also tried to condense the test a bit since I was adding more code there.

The digit count is specifically the digits that round trip, so it's already the conservative value.

toolchain/lex/lex.cpp
chandlerc: Maybe? I'm happy to look at those next?

chandlerc (commented):
Thanks, PTAL!

toolchain/lex/tokenized_buffer_test.cpp
jonmeow: Yeah, that's fine.

toolchain/lex/lex.cpp
jonmeow: SG

jonmeow: Actually, thinking about this further... maybe you should look at it here, because you're taking a shared function and implementing a specialized version, rather than continuing sharing?

jonmeow (approved):
Approving, because I expect you're going to be of a different mind about the importance of sharing of code.

toolchain/lex/lex.cpp
chandlerc: As discussed, happy to look at this, but would prefer in a follow-up.

chandlerc:
> Approving, because I expect you're going to be of a different mind about the importance of sharing of code.

Thanks, will look at the other things in follow-ups!

chandlerc:
:sigh: Looks like `std::from_chars` isn't available on Clang 16's libc++ -- OK for me to go back to the loop?

chandlerc:
I've restored the for loop for now, but maybe there is another approach? Nothing really simple comes to mind, and the loop is at least *somewhat* simple...

FWIW, it appears to be a bit faster to use the loop on x86. (About the same on Arm for me.)

PR #4432, opened 2024-10-21
Model the return slot as an output parameter

geoffromer:
Also fix `Param` insts to have meaningful names in pretty-printing, to help clarify relationship with return slot.


toolchain/sem_ir/typed_insts.h
```diff
@@ -833,6 +833,10 @@ struct Param {
 
   TypeId type_id;
   RuntimeParamIndex runtime_index;
+
+  // A name to associate with this Param in pretty-printed IR. This is not
+  // necessarily unique, or even valid, and has no semantic significance.
+  NameId pretty_name;
```
josh11b:
```suggestion
  NameId pretty_name_id;
```

toolchain/check/call.cpp
```diff
@@ -175,13 +175,13 @@ auto PerformCall(Context& context, SemIR::LocId loc_id, SemIR::InstId callee_id,
   }
 
   // If there is a return slot, build storage for the result.
-  SemIR::InstId return_storage_id = SemIR::InstId::Invalid;
+  SemIR::InstId return_slot_id = SemIR::InstId::Invalid;
```
zygoloid: Maybe `return_slot_arg_id`, given that this is the argument that goes in the return slot, not the parameter.

toolchain/sem_ir/typed_insts.h
```diff
@@ -881,6 +885,28 @@ struct ReturnExpr {
...
 
+// The return slot of a function declaration, as exposed in the function body.
+// This acts as an output parameter, analogous to `BindName` for input
+// parameters.
+struct ReturnSlot {
+  static constexpr auto Kind =
+      InstKind::ReturnSlot.Define<Parse::NodeId>({.ir_name = "return_slot"});
+
+  TypeId type_id;
+  InstId value_id;
```
josh11b: Comment? I'm guessing `type_id` matches the return type of the function (though it might be good to separately associate the specific instruction representing that expression in the source for diagnostics). What is the value_id?

toolchain/sem_ir/inst_namer.cpp
```diff
@@ -530,9 +530,9 @@ auto InstNamer::CollectNamesInBlock(ScopeId scope_id,
         add_inst_name_id(sem_ir_.name_scopes().Get(inst.name_scope_id).name_id);
         continue;
       }
-      case InstKind::Param: {
+      case CARBON_KIND(Param inst): {
         // TODO: Find a way to use the name of the enclosing bind inst here.
```
josh11b: Is this TODO now addressed?

toolchain/check/name_component.h
```diff
@@ -36,12 +36,18 @@ struct NameComponent {
   SemIR::InstBlockId params_id;
   SemIR::InstBlockId param_patterns_id;
 
+  // The return slot.
+  SemIR::InstId return_slot_pattern_id;
+  SemIR::InstId return_slot_id;
```
zygoloid: It seems a bit strange to include this in `NameComponent` given that it never appears in a name qualifier. I would have expected the return type to be a separate part of function syntax that appears after the name, rather than being treated as part of the name itself. Is it important that we handle this here rather than only in logic that's specific to function declarations? If so, have you thought about how we would model `fn F() => expr;`?

toolchain/sem_ir/typed_insts.h
zygoloid: Have you considered including the `InstId` of the pattern here instead of this name? It seems like we can in general reconstruct this from the pattern, and storing the pattern would likely be more generally useful.

toolchain/check/handle_function.cpp
```diff
@@ -44,9 +44,13 @@ auto HandleParseNode(Context& context, Parse::ReturnTypeId node_id) -> bool {
...
   auto type_id = ExprAsType(context, type_node_id, type_inst_id).type_id;
-  // TODO: Use a dedicated instruction rather than VarStorage here.
-  context.AddInstAndPush<SemIR::VarStorage>(
-      node_id, {.type_id = type_id, .name_id = SemIR::NameId::ReturnSlot});
+  auto return_slot_id = context.AddPatternInst<SemIR::ReturnSlotPattern>(
+      node_id, {.type_id = type_id});
+  auto param_pattern_id = context.AddPatternInst<SemIR::ParamPattern>(
+      node_id, {.type_id = type_id,
+                .subpattern_id = return_slot_id,
+                .runtime_index = SemIR::RuntimeParamIndex::Unknown});
```
zygoloid, line 54: Looking at the generated SemIR, it looks like the `Param` inst we create for this pattern is treated as a value expression. That doesn't seem correct to me. Is there some way we can fix that? (Are we going to need a second instruction kind for an `OutParam` or something like that?)

toolchain/check/call.cpp
geoffromer: Done, here and in a few other places.

toolchain/check/handle_function.cpp
geoffromer: That's a good point. I see three plausible options for fixing this:
1. Add a second kind of inst (probably for both `Param` and `ParamPattern`).
2. Don't have separate parameter insts for the return slot, and just use `OutParam` and `OutParamPattern` directly.
3. Give `Param` the ID of its `ParamPattern` (or the subpattern thereof?) as you suggest elsewhere, and update `GetExprCategory` to traverse into that.

1 would complicate the invariant that all calling-convention parameters are `Param`s, and that the caller and callee sides of the pattern are always separated by `ParamPattern`, but that's probably manageable with some more `Any` inst categories.

2 has the same problem, and also breaks the invariant that param insts are only referenced within the decl block (i.e. the calling-convention-facing insts are always separate from the function-body-facing insts). I don't know if that matters in practice, but I've found it useful for understanding the roles of the various parameter-related insts.

I'm worried that 3 will be confusing, because it's much more powerful than we need. I found `Param` much harder to understand when it had a `name_id` member, until I realized that it had no semantic significance, which is why I've been so careful to make that fact explicit now that I'm bringing that member back. Giving it the ID of the `ParamPattern` would make that problem much worse, because an `InstId` is vastly more powerful than a `NameId`. We wouldn't even be able to promise that it's not semantically significant; we could try to promise that it's only significant for purposes of determining the `ExprCategory`, but I think that's going to be a hard line to hold.

I'm ambivalent between 1 and 2, but I prefer either of them over 3. Do you have a preference?

Also, do you think we need a separate `ExprCategory` enumerator for this, e.g. `InitializingRef`? `DurableRef` is closer than `Value`, but it still seems a little sketchy because it's normally not correct to try to initialize a durable reference expression. This decision would be a little clearer if this kind of expression _only_ permitted initialization, but in order to support `returned var` I think it has to permit ordinary reads and writes as well.

toolchain/check/name_component.h
geoffromer: > It seems a bit strange to include this in `NameComponent` given that it never appears in a name qualifier. I would have expected the return type to be a separate part of function syntax that appears after the name, rather than being treated as part of the name itself.

A name qualifier never has a return type, but IIUC it never has implicit parameters either, and yet `NameComponent` includes implicit parameters. I don't see `NameComponent` as a representation of a name qualifier; I see it as a representation of anything that consists of a name and a full-pattern. The premise of this PR is that the return type of a function is part of the full-pattern of the function signature, so it follows that the return type is part of the `NameComponent`.

> Is it important that we handle this here rather than only in logic that's specific to function declarations?

It may not be important now, but in the future I think it's likely that we'll want to have a type that can represent any full-pattern, whether or not it's prefixed with a name. Right now that's easy: it's just a matter of factoring out a struct consisting of everything below `name_id` (or maybe below `last_param_node_id`). If the return slot members are in a different struct, that becomes difficult if not impossible.

> If so, have you thought about how we would model `fn F() => expr;`?

I expect we'd model it the same way we model `fn F() -> auto { return expr; }`. If your concern is about the `auto`, that does raise some challenges, because naively it creates a circular dependency between typechecking the body and typechecking the signature. However, that doesn't seem like a problem with `NameComponent` in particular; it's a problem with modeling the return slot as a parameter at all.

toolchain/sem_ir/typed_insts.h
geoffromer: See my response to your comment on `handle_function.cpp`, where I call this option 3.

toolchain/sem_ir/typed_insts.h
geoffromer: OK, how's this?

> (though it might be good to separately associate the specific instruction representing that expression in the source for diagnostics)

Do you see any places we can use that right now? If not, I'd rather wait until we have a need for it.

toolchain/check/name_component.h
zygoloid: > A name qualifier never has a return type, but IIUC it never has implicit parameters either, and yet `NameComponent` includes implicit parameters.

A `NameComponent` can include implicit parameters when used as a qualifier:

```carbon
class C[T:! type](N:! T) {
  fn F();
}
fn C[T:! type](N:! T).F() {}
```

> It may not be important now, but in the future I think it's likely that we'll want to have a type that can represent any full-pattern, whether or not it's prefixed with a name. Right now that's easy: it's just a matter of factoring out a struct consisting of everything below name_id (or maybe below last_param_node_id). If the return slot members are in a different struct, that becomes difficult if not impossible.

I see. For right now, it seems that the place where this comes up is `PopNameComponent`, which is calling `CalleePatternMatch` itself and handling the return pattern and return slot parameter in addition to the declared parameters. Doing that all together seems logical given that it's effectively computing the calling convention of the function (in the case where the name is for a function). I wonder whether we could factor that process a bit differently so we only deal with return types along a function-specific codepath (eg, maybe add a `PopFunctionSignature` that handles the return type, and returns a `FunctionSignature` containing a `NameComponent` and a return type).

But in any case: while this representation seems to me to be conflating two similar but different things, I'm OK with us using it for now and revisiting this once the broader pattern matching support you mentioned is built. Please add a TODO to reconsider including the return information in `NameComponent`.

toolchain/check/handle_function.cpp
zygoloid: [We also discussed this in the toolchain meeting today; I wrote up the comment below prior to that meeting.]

For option (2), I think I'd want something specific to `return` in the instruction name, given that we don't have the return slot instruction any more. But I think I like separating the return parameter from the return slot, at least in theory, and that might open up a way to represent things like a return value that's partly returned in memory and partly returned through registers, or for which we pass multiple parameters for locations of parts of the return value. So I think that leads me to prefer option (1).

Additionally I think we'll want something similar to distinguish `var` parameters from `let` parameters, because the same expr category decision is going to be needed there too. So I think we will end up with an `AnyParam` category anyway. In principle we could use the same instruction kind for both `var` parameters and return slot parameters, but in practice I think we should distinguish them. (Also, they're not the same, I suppose: for a type that's returned in registers, a `RefParam` should be lowered as a pointer parameter but an `OutParam` should not add a parameter to the LLVM function at all.)

geoffromer: OK, I've added a TODO to implement option 1.

toolchain/check/name_component.h
geoffromer: OK, I've added a TODO.

zygoloid (commented):
LGTM, leaving for @josh11b to approve once he's happy.

toolchain/check/return.cpp
```diff
@@ -42,7 +42,7 @@ static auto NoteReturnType(Context::DiagnosticBuilder& diag,
                            SemIR::TypeId return_type_id) {
   CARBON_DIAGNOSTIC(ReturnTypeHereNote, Note, "return type of function is {0}",
                     SemIR::TypeId);
-  diag.Note(function.return_storage_id, ReturnTypeHereNote, return_type_id);
+  diag.Note(function.return_slot_id, ReturnTypeHereNote, return_type_id);
```
josh11b: Here is an example where it would be better to use an instruction id instead of a TypeId, so that we could match the spelling of the type the user wrote. We have been using TypeIds previously, but the goal is to switch.

geoffromer: OK, given an `InstId` representing the type expression the user wrote, how do I obtain the spelling the user wrote? Is there an existing example of the new pattern that I can follow to update this code?

zygoloid: Use the type `InstIdAsType` as the type of the diagnostic argument; search for that type name for examples.

geoffromer: OK, done.

josh11b (approved):

PR #4434, opened 2024-10-22
Delete `Function::ParamInfo::GetNameId`

josh11b:
No longer used as of #4422 .


geoffromer (approved):
Huh, I thought this was still in use. Nice!

PR #4438, opened 2024-10-23
Add data_runfiles to manifests

jonmeow:
This is trying to help systems that don't use --incompatible_always_include_files_in_data


chandlerc (approved):
FWIW, LGTM -- but leaving in case Josh / Jon have more context.

PR #4439, opened 2024-10-23
Pass the manifest path by flag.

jonmeow:
This is just a cleanup to remove the hardcoded path, since it's not really necessary in context.


chandlerc (approved):
Nice, much simpler!

PR #4440, opened 2024-10-23
Fix use of runfiles

jonmeow:
I was looking at the documentation again, and I realized that while this works, I think the intention is that `runfiles` is set _instead_ of both fields (I was a little confused by the error I was getting when setting `runfiles` _with_ `default_runfiles`)

I believe I've verified this works in necessary situations.


chandlerc (approved):
Huh... but, sure? LG!

PR #4444, opened 2024-10-25
Small refactoring to Extract for compile time.

jonmeow:
AFAICT https://github.com/carbon-language/carbon-lang/pull/4363 made builds of extract.cpp go from ~15s to ~35s. I'm not sure how to really improve on this, short of adding boilerplate to the types in order to reduce template use (e.g., instead of using struct reflection to return fields, we could have something that directly returns fields). But, this switch to `MaybeTrace` seems to be about a 20% build time improvement (down to ~30s), with `noinline` accounting for a part of that.


josh11b:
https://en.m.wikipedia.org/wiki/Four_fours

josh11b (approved):

PR #4447, opened 2024-10-25
Don't eagerly import all impls.

zygoloid:
Instead, eagerly import only the impls from the api file corresponding to the current file, if any, because we need those for impl redeclaration lookup. For all other cases, load only the impls in libraries that are referenced as part of an impl lookup query.


josh11b (approved):

josh11b:
Looks like there is a test fuzzer failure?

PR #4449, opened 2024-10-28
Try working around xcrun failures

jonmeow:
I expect the change here will try rerunning a couple times, then fail if it's a permanent failure. 

I've seen similar failures, and believe this is flaky. Here's the specific example that caused me to try a workaround:
https://github.com/carbon-language/carbon-lang/actions/runs/11560505812/job/32177497296

I don't see a difference in the runner information between failing and passing runs, which might've indicated a canary. That's why I'm doing this as a trivial retry.

### DIFF
--- a/.github/actions/build-setup-macos/action.yml
+++ b/.github/actions/build-setup-macos/action.yml
@@ -14,8 +14,13 @@ runs:
     - name: Free up disk space
       shell: bash
       run: |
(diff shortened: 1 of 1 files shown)


zygoloid (approved):
Sure, let's give this a go. I'm not confident this will solve the problem but it can't hurt at least.

jonmeow:
Yep, I'm not sure it'll solve it either, but figured it occurs often enough to justify trying something. :)

PR #4450, opened 2024-10-28
Switch to using `Core.BigInt` as the type of the size of a type literal.

zygoloid:
This removes one of the few ways in which `i32` is special and gets us closer to removing it as a special case.

### DIFF
--- a/toolchain/check/handle_literal.cpp
+++ b/toolchain/check/handle_literal.cpp
@@ -49,6 +49,20 @@ static auto MakeI32Literal(Context& context, Parse::NodeId node_id,
        .int_id = context.ints().Add(i32_val)});
 }
 
+// Forms an IntLiteral instruction with type `BigInt` for a given literal
+// integer value, which is assumed to be unsigned.
+static auto MakeBigIntLiteral(Context& context, Parse::NodeId node_id,
+                              IntId int_id) -> SemIR::InstId {
+  // TODO: `IntId`s with different bit-widths are considered different values
+  // here. Decide how we want to canonicalize these. For now this is only used
+  // by type literals, so we rely on the lexer picking some consistent rule.
+  auto val = context.ints().Get(int_id);
+  return context.AddInst<SemIR::IntLiteral>(
+      node_id,
+      {.type_id = context.GetBuiltinType(SemIR::BuiltinInstKind::BigIntType),
+       .int_id = int_id});
+}
+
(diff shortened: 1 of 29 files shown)


toolchain/check/handle_literal.cpp
```diff
@@ -49,6 +49,20 @@ static auto MakeI32Literal(Context& context, Parse::NodeId node_id,
...
 }
 
+// Forms an IntLiteral instruction with type `BigInt` for a given literal
+// integer value, which is assumed to be unsigned.
+static auto MakeBigIntLiteral(Context& context, Parse::NodeId node_id,
+                              IntId int_id) -> SemIR::InstId {
+  // TODO: `IntId`s with different bit-widths are considered different values
+  // here. Decide how we want to canonicalize these. For now this is only used
+  // by type literals, so we rely on the lexer picking some consistent rule.
+  auto val = context.ints().Get(int_id);
```
jonmeow: What's the `Get` for? `val` is unused...

toolchain/check/handle_literal.cpp
```diff
@@ -49,6 +49,20 @@ static auto MakeI32Literal(Context& context, Parse::NodeId node_id,
        .int_id = context.ints().Add(i32_val)});
 }
 
+// Forms an IntLiteral instruction with type `BigInt` for a given literal
+// integer value, which is assumed to be unsigned.
+static auto MakeBigIntLiteral(Context& context, Parse::NodeId node_id,
+                              IntId int_id) -> SemIR::InstId {
+  // TODO: `IntId`s with different bit-widths are considered different values
+  // here. Decide how we want to canonicalize these. For now this is only used
+  // by type literals, so we rely on the lexer picking some consistent rule.
```
jonmeow, line 58: I think Chandler was looking at this separately, you might want to talk with him. A TODO here might be lost, because this may be more of a TODO for the numeric literal lexer.

jonmeow (approved):
Basically looks good, but might want to remove things from `MakeBigIntLiteral` before merging.

toolchain/check/handle_literal.cpp
zygoloid:
```suggestion
```
It's just left behind from when I was doing some canonicalization here.

toolchain/check/handle_literal.cpp
zygoloid: I couldn't find another good place for this TODO to live. I'll try to remember it once Chandler's change lands.

PR #4452, opened 2024-10-29
Remove CalleeParamsInfo

jonmeow:
I'm seeing three issues with CalleeParamsInfo:

1. Although ResolveCalleeInCall had been extracted out and the EntityWithParamsBase could be used directly, that had not been cleaned up.
2. implicit_param_refs_id is unused; param_refs_id was only used by ResolveCalleeInCall. The factoring as a struct seems to be obscuring what's used and what isn't (creating unnecessary copies).
3. On #4446, CalleeParamsInfo seemed to be obfuscating what ConvertCallArgs actually worked with (a Function, not a generic entity).

I'm thinking that just removing CalleeParamsInfo is the best resolution here, it looks like it's tripping people up more than it's helping.

Note, I think Function used to be named Callable, which was where the "callable" name originally came from (I might be wrong about this). But "function" seems clearer about the type now.


jonmeow:
Note, I was going back and forth about passing parameters individually versus just passing the `Function` as a whole, my leaning is towards the `Function` partly based on the way that'll resolve conflicts with #4446 (essentially, making `return_slot_patterns_id` part of the things implicitly passed)

jonmeow:
This PR seems to have slipped through the cracks, maybe @geoffromer can review?

geoffromer (approved):

PR #4453, opened 2024-10-29
Elide prelude components in the IR formatter.

jonmeow:
This is in particular to avoid churn from changes such as #4370. I think the import list can be helpful (particularly to understand what the library is aware of), but it's a different trade-off for the prelude package due to the implicit imports.

### DIFF
--- a/toolchain/sem_ir/formatter.cpp
+++ b/toolchain/sem_ir/formatter.cpp
@@ -542,11 +542,24 @@ class FormatterImpl {
       out_ << "extend " << extended_scope_id << "\n";
     }
 
+    // This is used to cluster all "Core//prelude/..." imports, but not
+    // "Core//prelude" itself. This avoids unrelated churn in test files when we
+    // add or remove an unused prelude file, but is intended to still show the
+    // existence of indirect imports.
+    bool has_prelude_components = false;
     for (auto [import_ir_id, unused] : scope.import_ir_scopes) {
+      auto label = GetImportIRLabel(import_ir_id);
+      if (label.starts_with("Core//prelude/")) {
+        if (has_prelude_components) {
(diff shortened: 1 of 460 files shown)


toolchain/sem_ir/formatter.cpp
```diff
@@ -948,23 +961,10 @@ class FormatterImpl {
...
-    }
-    out_ << "//";
-    CARBON_CHECK(import_ir.library_id().is_valid());
-    if (import_ir.library_id() == LibraryNameId::Default) {
-      out_ << "default";
-    } else {
-      out_ << import_ir.string_literal_values().Get(
-          import_ir.library_id().AsStringLiteralValueId());
+      out_ << id;
     }
```
chandlerc, line 968: Optional idea, I'm happy with either form though:

```suggestion
    out_ << (id.is_valid() ? GetImportIRLabel(id) : id);
```

chandlerc (approved):
Makes sense, especially for tests where the prelude is *so* noisy.

I wonder if we may want this to work the other way by default outside of file tests, but no reason to worry about that today. LG.

toolchain/sem_ir/formatter.cpp
jonmeow:
```
toolchain/sem_ir/formatter.cpp:964:28: error: incompatible operand types ('std::string' (aka 'basic_string<char>') and 'ImportIRId')
    out_ << (id.is_valid() ? GetImportIRLabel(id) : id);
                           ^ ~~~~~~~~~~~~~~~~~~~~   ~~
```

chandlerc: :sigh: right, C++ makes that tricky. Oh well (and thanks).

PR #4454, opened 2024-10-29
Add a framework for typed InstIds.

jonmeow:
I'm switching `Temporary::storage_id` to a `TemporaryStorageInstId` as an example of this. Other than that, I'm not sure whether there are obvious low hanging fruit to change -- I figured starting to add the capability would be helpful towards use in future code.

Note, if you don't think this is more broadly helpful, I can abort this, but I think this is something you'd brought up so just trying  to provide it while feeling a little blocked elsewhere.

### DIFF
--- a/toolchain/sem_ir/id_kind.h
+++ b/toolchain/sem_ir/id_kind.h
@@ -115,8 +115,6 @@ class TypeEnum {
 
 // An enum of all the ID types used as instruction operands.
 using IdKind = TypeEnum<
-    // From sem_ir/builtin_inst_kind.h.
-    BuiltinInstKind,
     // From base/value_store.h.
     IntId, RealId, FloatId, StringLiteralValueId,
     // From sem_ir/id.h.
@@ -124,7 +122,12 @@ using IdKind = TypeEnum<
     CompileTimeBindIndex, RuntimeParamIndex, FunctionId, ClassId, InterfaceId,
     ImplId, GenericId, SpecificId, ImportIRId, ImportIRInstId, LocId, BoolValue,
     IntKind, NameId, NameScopeId, InstBlockId, TypeId, TypeBlockId,
-    ElementIndex, LibraryNameId, FloatKind>;
+    ElementIndex, LibraryNameId, FloatKind,
+// Subtyped InstId variants.
+#define CARBON_SEM_IR_INST_KIND(Name) Name##InstId,
+#include "toolchain/sem_ir/inst_kind.def"
+    // From sem_ir/builtin_inst_kind.h.
+    BuiltinInstKind>;
 
 }  // namespace Carbon::SemIR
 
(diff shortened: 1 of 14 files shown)


toolchain/sem_ir/id_kind.h
```diff
@@ -115,16 +116,19 @@ class TypeEnum {
...
     // From sem_ir/id.h.
     InstId, AbsoluteInstId, MatchingInstId, ConstantId, EntityNameId,
     CompileTimeBindIndex, RuntimeParamIndex, FunctionId, ClassId, InterfaceId,
     ImplId, GenericId, SpecificId, ImportIRId, ImportIRInstId, LocId, BoolValue,
     IntKind, NameId, NameScopeId, InstBlockId, TypeId, TypeBlockId,
-    ElementIndex, LibraryNameId, FloatKind>;
+    ElementIndex, LibraryNameId, FloatKind,
+// Subtyped InstId variants.
+#define CARBON_SEM_IR_INST_KIND(Name) Name##InstId,
+#include "toolchain/sem_ir/inst_kind.def"
```
zygoloid, line 129: I'm concerned by this part: right now we have a few places where we're uniformly processing `InstId` operands of instructions by looking at the `IdKind`, and after this change those places will no longer visit the `storage_id` field on `Temporary` ([example](https://github.com/carbon-language/carbon-lang/blob/85f6bf32b5a42fcd5d0ac5faa4e4780a68787710/toolchain/check/subst.cpp#L113), [example](https://github.com/carbon-language/carbon-lang/blob/85f6bf32b5a42fcd5d0ac5faa4e4780a68787710/toolchain/check/deduce.cpp#L105)). Maybe we could map all the typed `InstId` kinds to `IdKind::For<InstId>`?

zygoloid (commented):
I think this could be useful. Do you know how much this affects compile times? (It looks like it shouldn't be much to me.)

jonmeow:
> I think this could be useful. Do you know how much this affects compile times? (It looks like it shouldn't be much to me.)

Right now it doesn't affect much. Maybe I'm a little biased due to typed node IDs and extract.cpp performance, but I think the biggest risk for compile performance is if it sees broad use in typed insts. Whereas right now that's not an issue, the number of kinds they deal with right now is small. If these start seeing broad use in signatures, I would expect that the things dealing with typed inst reflection will slow down (due to less reuse), and overall we could see a significant slowdown in compile times. However, it's possibly premature to make assumptions in that direction.

jonmeow:
Per discussion, closing this PR. We went through current uses of `As` and `GetAs`, and it didn't look like there were enough spots where a `TypedInstId` could simplify things. We may look into specialized `StructTypeField` storage,  but that wouldn't have been helped here unless we had typed `InstBlockId` fetching too.

toolchain/sem_ir/id_kind.h
jonmeow: PS: I was concerned that collapsing to `IdKind::For<InstId>` would affect the ability of `NodeStack` (and possibly others) to propagate a `TypedInstId`. This is what led the discussion below, where there's maybe not actually enough value, closing the PR.

PR #4456, opened 2024-10-30
Set `.python-version` to 3.10 for pyenv users

samestep:
The contribution docs for macOS say to install Python 3.10:

https://github.com/carbon-language/carbon-lang/blob/957599b2ab036d01b325aa6e82edea0fce7b4c53/docs/project/contribution_tools.md#L118-L124

For people who are using [pyenv](https://github.com/pyenv/pyenv) instead of Homebrew Python, this PR adds a `.python-version` file to specify Python 3.10 for pyenv to use.

However, I also see that later in the same document, the docs only say that Python >=3.9 is required:

https://github.com/carbon-language/carbon-lang/blob/957599b2ab036d01b325aa6e82edea0fce7b4c53/docs/project/contribution_tools.md?plain=1#L162

So in that case, feel free to just close this PR. I also saw that #778 specifically moved Carbon away from pyenv, so if pyenv is discouraged in general, also feel free to just close this.


jonmeow (approved):
LG -- Sorry, for some reason I'd thought someone else got this PR.

jonmeow:
Sorry, I see the pre-commit issue. In .pre-commit-config.yml, around line 215, I'd suggest inserting an exception, basically:

```diff
               .github/pull_request_template.md|
+              .python-version|
               compile_flags.txt|
```

samestep:
I just added the copyright text instead, guessing that works too?

jonmeow:
> I just added the copyright text instead, guessing that works too?

See the error -- Python's strict about file contents.

samestep:
Huh, weird... I guess `actions/setup-python` is buggy, since [pyenv says this should be fine](https://github.com/pyenv/pyenv/tree/v2.4.17?tab=readme-ov-file#understanding-python-version-selection):

> You can also specify multiple versions in a `.python-version` file by hand, separated by newlines. Lines starting with a `#` are ignored.

In any case, I've now implemented your suggestion. Sorry for taking up so much of your time on this!

jonmeow:
> In any case, I've now implemented your suggestion. Sorry for taking up so much of your time on this!

No problem, it's really not much time and I should've reviewed last week. FWIW, I wasn't actually looking at what _should_ be supported.

PR #4457, opened 2024-10-30
Model return slot as parameter in lowering

geoffromer:


zygoloid (commented):
I think there's a multiple-ownership issue here: `splice_block` owns its block, but the decl block is already owned by the `function_decl` instruction. In practice this is resulting in the decl block being printed twice in SemIR.

Given that different `function_decl`s for the same function will have different decl blocks, I don't think we want to just transfer ownership from the `function_decl` to the `Function` entity. I think either of these options could be reasonable:

- Instead of `splice_block`, use a different instruction that invokes another block without splicing it. We'd need some way of naming the block owned by the `function_decl` -- maybe the instruction could name the `function_decl` instruction itself?
- Omit the `splice_block` entirely and instead make lowering first lower the decl block of the definition and then the function body.

I don't think I have a strong preference between these, but the latter option would avoid the need to add a new kind of special-purpose instruction.

toolchain/lower/file_context.cpp
```diff
@@ -324,41 +329,59 @@ auto FileContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
...
-          .has_return_slot()) {
-    function_lowering.SetLocal(function.return_slot_id,
-                               llvm_function->getArg(param_index));
-    ++param_index;
-  }
+  // The SemIR calling-convention parameters of the function, in order of
+  // runtime index. This is a transitional step toward generating this list
+  // in the check phase, which is why we're using the runtime index order
+  // even though it's less convenient for this usage.
+  llvm::SmallVector<SemIR::InstId> calling_convention_param_ids;
```
zygoloid, line 336: Can we `reserve` this to the right size in advance? (Or even, `resize` it to the right size and index by the runtime index when adding entries.)

toolchain/lower/file_context.cpp
```diff
@@ -324,41 +329,59 @@ auto FileContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
...
-  if (SemIR::ReturnTypeInfo::ForFunction(sem_ir(), function, specific_id)
-          .has_return_slot()) {
-    function_lowering.SetLocal(function.return_slot_id,
-                               llvm_function->getArg(param_index));
-    ++param_index;
-  }
+  // The SemIR calling-convention parameters of the function, in order of
+  // runtime index. This is a transitional step toward generating this list
+  // in the check phase, which is why we're using the runtime index order
+  // even though it's less convenient for this usage.
```
zygoloid, line 335: You don't seem to really be making use of the fact that this is in SemIR calling convention order. If you instead built a list in LLVM IR calling convention order, I think the `lower_param` code could instead be a loop over that array. Perhaps you could factor out the code in `BuildFunctionDecl` that constructs `param_inst_ids`, since that's then the list that we'd want here.

toolchain/lower/file_context.cpp
geoffromer: I've added a `reserve` call with an upper bound. Precomputing the exact size is annoying because we'd need to traverse the params twice, first to figure out which ones are runtime params and then to actually build the vector. Do you think that would be worthwhile?

toolchain/lower/file_context.cpp
geoffromer: The next PR is going to build `calling_convention_param_ids` in `CalleePatternMatch`, and propagate that through instead of `param_refs` and `implicit_param_refs`. At that point it would be difficult and unnatural to arrange for `calling_convention_param_ids` to be in LLVM order. Consequently, I'm intentionally creating it in SemIR order now, in order to validate the code below that manages the transition from SemIR order to LLVM order.

geoffromer (commented):
> * Omit the `splice_block` entirely and instead make lowering first lower the decl block of the definition and then the function body.

I initially tried something like that, but got stuck on how to give the lowered decl block a suitable terminator. I just tried again, and I think I found a way to make it work without too much mess.

toolchain/lower/file_context.cpp
zygoloid: OK, thanks for explaining, that makes sense :)

toolchain/lower/file_context.cpp
zygoloid: If this is becoming an `InstBlock` in a subsequent PR anyway, I don't think it's worth spending any more time on optimizing it.

toolchain/lower/file_context.cpp
```diff
@@ -371,13 +410,12 @@ auto FileContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
     function_lowering.LowerBlock(block_id);
   }
```
zygoloid, line 424: This unconditionally creates an entry block separate from the function body, which seems a little wasteful. Perhaps instead, we could allow the decl block to fall through into the entry block:
```c++
  for (auto block_id : llvm::concat<const SemIR::InstBlockId>(
           singleton_decl_block, body_block_ids)) {
    CARBON_VLOG("Lowering {0}\n", block_id);
    // Reuse the decl block as the first block of the body if possible. We assume that
    // the first body block is never a branch target.
    if (block_id != body_block_ids.front() || !function_lowering.builder().GetInsertBlock()) {
      auto* llvm_block = function_lowering.GetBlock(block_id);
      // Keep the LLVM blocks in lexical order.
      llvm_block->moveBefore(llvm_function->end());
      function_lowering.builder().SetInsertPoint(llvm_block);
    }
    function_lowering.LowerBlock(block_id);
  }
```

geoffromer: Hmm. The code I deleted just below here suggests that we can't assume the first body block is never a branch target, but how about this?

zygoloid: Thanks, that seems like it strikes a good balance.

toolchain/lower/file_context.cpp
```diff
@@ -397,25 +397,45 @@ auto FileContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
...
+  // If the decl block is empty, reuse it as the first body block. We don't do
+  // this when the decl block is non-empty so that any branches back to the
+  // first body block don't also re-execute the decl.
+  if (llvm_decl_block->empty()) {
+    CARBON_CHECK(function_lowering.TryToReuseBlock(body_block_ids.front(),
+                                                   llvm_decl_block));
+  } else {
+    function_lowering.builder().SetInsertPoint(llvm_decl_block);
+    function_lowering.builder().CreateBr(
+        function_lowering.GetBlock(body_block_ids.front()));
```
zygoloid: Maybe
```suggestion
  // If the decl block is empty, reuse it as the first body block. We don't do
  // this when the decl block is non-empty so that any branches back to the
  // first body block don't also re-execute the decl.
  llvm::BasicBlock* block = context.builder().GetInsertBlock();
  if (block->empty() &&
      function_lowering.TryToReuseBlock(body_block_ids.front(),
                                        block))) {
    // Reuse this block as the first block of the function body.
  } else {
    function_lowering.builder().CreateBr(
        function_lowering.GetBlock(body_block_ids.front()));
```
to make this more similar to the other uses of `TryToReuseBlock` and handle the possibility that we somehow generate multiple LLVM basic blocks while lowering the decl block. (We shouldn't need to `SetInsertionPoint` because it should still be in the `llvm_decl_block`.) I wonder if we should actually factor out a function to do this block-reuse or branch process -- we seem to do the same thing [three](https://github.com/carbon-language/carbon-lang/blob/954441c358d1b7df5515ab2728953bce0eb690d4/toolchain/lower/handle.cpp#L104) [times](https://github.com/carbon-language/carbon-lang/blob/954441c358d1b7df5515ab2728953bce0eb690d4/toolchain/lower/handle.cpp#L135) now.

zygoloid (approved):

toolchain/lower/file_context.cpp
```diff
@@ -324,51 +329,103 @@ auto FileContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
...
+  };
+
+  // If the decl block is empty, reuse it as the first body block. We don't do
+  // this when the decl block is non-empty so that any branches back to the
+  // first body block don't also re-execute the decl.
+  llvm::BasicBlock* block = context.builder().GetInsertBlock();
+  if (block->empty() &&
+      function_lowering.TryToReuseBlock(body_block_ids.front(),
+                                        block))) {
+    // Reuse this block as the first block of the function body.
```
```suggestion
      function_lowering.TryToReuseBlock(body_block_ids.front(), block)) {) {
      // Reuse this block as the first block of the function body.
    }
```

toolchain/lower/file_context.cpp
geoffromer: Done, with some tweaks. I assume removing the first `lower_block` call was an accident?

zygoloid (approved):

toolchain/lower/file_context.cpp
zygoloid: Yes, oops, I just meant to remove the variable not the initializer.

PR #4458, opened 2024-10-30
Switch tar verification to a manifest comparison.

jonmeow:
This removes the install marker-relative path checks, and replaces it with bidirectional verification: previously, files in the tar file had to be in install data, but there was no check that files in install data were all in the tar.

### DIFF
--- a/toolchain/install/toolchain_tar_test.py
+++ b/toolchain/install/toolchain_tar_test.py
@@ -38,6 +38,7 @@ def main() -> None:
     )
     args = parser.parse_args()
 
+    # Gather install data files.
     with open(args.install_data_manifest) as manifest:
         # Remove everything up to and including `prefix_root`.
         install_files = set(
@@ -48,8 +49,7 @@ def main() -> None:
         )
     assert len(install_files), f"`{args.install_data_manifest}` is empty."
 
-    # First check that every file and directory in the tar file exists in our
-    # prefix root, and build a set of those paths.
+    # Gather tar files.
     with tarfile.open(args.tar_file) as tar:
         # Remove the first path component.
         tar_files = set(
--- a/toolchain/install/toolchain_tar_test.py
+++ b/toolchain/install/toolchain_tar_test.py
@@ -9,9 +9,19 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 """
 
 import argparse
-import sys
 from pathlib import Path
+import re
 import tarfile
+
+
+def expect_empty_set(filename: str, file_set: set[str]) -> bool:
+    """Prints and returns false when the set has entries."""
+    if file_set:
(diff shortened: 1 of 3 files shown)


toolchain/install/toolchain_tar_test.py
```diff
@@ -22,42 +32,42 @@ def main() -> None:
...
-            sys.exit(
-                "ERROR: File `{0}` is not in tar file.".format(relative_path)
-            )
+    # Verify file sets match.
+    tar_okay = expect_empty_set(args.tar_file, tar_files - install_files)
+    install_okay = expect_empty_set(
+        args.install_data_manifest, install_files - tar_files
+    )
+    if not (tar_okay and install_okay):
+        exit("error: tar and install data did not match.")
```
chandlerc: Is there a trivial function to print the difference here? Not worth lots of code, mostly wondering if Python already has this.

chandlerc (approved):
Very nice! Glad to see the TODO go away. Totally optional comment inline.

toolchain/install/toolchain_tar_test.py
jonmeow: Can you please clarify what you're looking for versus what `expect_empty_set` is printing?

chandlerc: Sorry, I missed that `expect_empty_set` actually printed it out above, for some reason I thought it was just verifying empty with an expect framework. Anyways, change still LG.

chandlerc (approved):

toolchain/install/toolchain_tar_test.py
jonmeow: Thinking about this further, switched to `unittest` which has better support for this, with less code. To give you an idea:

```
======================================================================
FAIL: test_tar (__main__.ToolchainTarTest.test_tar)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/google/home/jperkins/.cache/bazel/_bazel_jperkins/85deb7d9d96f7e0e80b42618a55969d7/sandbox/linux-sandbox/10463/execroot/_main/bazel-out/k8-fastbuild/bin/toolchain/install/carbon_toolchain_tar_test.runfiles/_main/toolchain/install/toolchain_tar_test.py", line 47, in test_tar
    self.assertSetEqual(install_files, tar_files)
AssertionError: Items in the first set but not the second:
'lib/carbon/carbon_install.txt'
'lib/carbon/carbon-busybox'
'lib/carbon/llvm/bin/lld'
Items in the second set but not the first:
'lld'
'carbon-busybox'
'carbon_install.txt'

----------------------------------------------------------------------
```

chandlerc: Wow, very nice! Thanks!

chandlerc (approved):
(and LG with new unittest module based code)

PR #4459, opened 2024-10-30
Refactor run_tool

jonmeow:
Removes the python script, should get equivalent results in the current setup.


chandlerc (approved):
oooo, nice! Didn't realize how easy that would be in starlark! LG

PR #4460, opened 2024-10-30
Add facet type values and an instruction that produces them

josh11b:
Still to do:
* Represent facet type values in a canonical form
* Produce & consume facet type values instead of interface values
* `type` should be associated with a canonical facet type value
* Support `&` on facet type values
* Type check and enforce requirements in facet types


zygoloid (approved):

toolchain/check/handle_where.cpp
```diff
@@ -79,6 +79,7 @@ auto HandleParseNode(Context& context, Parse::RequirementEqualEqualId node_id)
   auto rhs = context.node_stack().PopExpr();
   auto lhs = context.node_stack().PopExpr();
   // TODO: type check lhs and rhs are comparable
+  // TODO: require that at least one side uses a designator
```
jonmeow, line 82: Just a reminder, as I was seeing several of these, but it'd be helpful if you could adhere to sentence capitalization and punctuation in comments. I believe this includes TODOs.

https://google.github.io/styleguide/cppguide.html#Punctuation,_Spelling_and_Grammar

PR #4462, opened 2024-10-31
Switch eval of ArrayIndex to use CARBON_KIND

jonmeow:
Just a minor cleanup.


chandlerc (approved):
LG (just grabbing some easy code reviews to save load)

PR #4463, opened 2024-10-31
Don't create a compile time binding after CompileTimeBindingInVarDecl

jonmeow:
Related to #4461, more generally try not to produce a compile-time binding when the code shouldn't be able to do so.


chandlerc (approved):
LG

PR #4464, opened 2024-10-31
Fix lowering of `specific_function`s referring to methods.

zygoloid:
In this case, the callee may be non-constant because it includes a reference to `self`, so we need to be able to lower a non-constant `specific_function`.


jonmeow (approved):
LG

PR #4467, opened 2024-10-31
Fix lowering of a conversion from a type with a pointer value representation to a type with a copy value representation.

zygoloid:
We previously generated a `value_bind` instruction of the wrong type, resulting in lowering building bad LLVM IR.

Also fix another issue exposed by this change, where we would import constants without marking their types as complete, and then crash in lowering while trying to lower them. This happens in particular for the `FunctionType`s of functions in `ImplDecl`s. Address this by skipping lowering for constants with incomplete types.


jonmeow (approved):
LG

PR #4469, opened 2024-11-01
Move the language server into toolchain's busybox.

jonmeow:
Removes the separate language server binary; I'm not sure we need to provide it. Instead, `carbon language-server` is added as a subcommand.

Moves //language_server to //toolchain/language_server. Splits into a trivial language_server.h, and a substantive server.h. I wasn't sure about a better name, but wanted the split similar to check/check.h, lex/lex.h, etc. At the same time, the class is probably going to be a little big so not a good fit to through into just a cpp file.

This fixes some style issues with the language server class, but generally I'm trying to not address things here in order to keep it simpler.


toolchain/language_server/server.h
```diff
@@ -38,46 +39,48 @@ class LanguageServer : public clang::clangd::Transport::MessageHandler,
...
-  void OnDocumentSymbol(
+  auto OnDocumentSymbol(
       clang::clangd::DocumentSymbolParams const& params,
-      clang::clangd::Callback<std::vector<clang::clangd::DocumentSymbol>> cb);
+      clang::clangd::Callback<std::vector<clang::clangd::DocumentSymbol>> cb)
+      -> void;
+
+  const std::unique_ptr<clang::clangd::Transport> transport_;
+  // content of files managed by the language client.
+  std::unordered_map<std::string, std::string> files_;
```
zygoloid, line 81: Looks like you removed the include of `<unordered_map>` but kept the map. Can that be fixed somehow? (Also we should include `<memory>` for `unique_ptr`.)

toolchain/language_server/server.h
```diff
@@ -38,46 +39,48 @@ class LanguageServer : public clang::clangd::Transport::MessageHandler,
...
-                    clang::clangd::Callback<llvm::json::Object> cb);
+  auto OnInitialize(clang::clangd::NoParams const& client_capabilities,
+                    clang::clangd::Callback<llvm::json::Object> cb) -> void;
 
   // Code outline
-  void OnDocumentSymbol(
+  auto OnDocumentSymbol(
       clang::clangd::DocumentSymbolParams const& params,
-      clang::clangd::Callback<std::vector<clang::clangd::DocumentSymbol>> cb);
+      clang::clangd::Callback<std::vector<clang::clangd::DocumentSymbol>> cb)
```
zygoloid, line 76: Do we need to include `<vector>` for this?

zygoloid (approved):

toolchain/language_server/server.h
jonmeow: Added includes

toolchain/language_server/server.h
jonmeow: Added includes

PR #4470, opened 2024-11-01
Refactor struct literal parse nodes.

jonmeow:
Split StructComma into StructLiteralComma and StructTypeLiteralComma in order to easily differentiate handling (remains the same in this PR).

Add "Literal" to StructField and StructTypeField because it feels inconsistent versus the other non-shared things. StructFieldDesignator remains shared between value literals and type literals.

Note I probably would've made StructFieldDesignator non-shared too, but that'd require either a lookahead of 2 (to see the separator`) or a writeback after parsing the separator, neither of which felt especially crucial for this, when what I'm really trying to do is split type literal handling a little further.


chandlerc (approved):
LG!

PR #4471, opened 2024-11-01
Adjust some build troubleshooting notes

jonmeow:
Came up due to [libc++ install issues](https://discord.com/channels/655572317891461132/655577725347561492/1302023663155023905)

We've discussed clang version verification, and adding that as long as I'm in here. The more significant bit is the libc++ check, which if it's not installed should fail like:

```
(tons of output)
ignoring nonexistent directory "/include"
#include "..." search starts here:
#include <...> search starts here:
 /usr/lib/llvm-16/lib/clang/16/include
 /usr/local/include
 /usr/include/x86_64-linux-gnu
 /usr/include
End of search list.
/usr/local/google/home/jperkins/.cache/bazel/_bazel_jperkins/85deb7d9d96f7e0e80b42618a55969d7/external/_main~clang_toolchain_extension~bazel_cc_toolchain/_temp:6:2: error: "No libc++ install found!"
#error "No libc++ install found!"
 ^
1 error generated.
ERROR: Analysis of target '//toolchain:toolchain' failed; build aborted: Analysis failed
INFO: Elapsed time: 0.265s, Critical Path: 0.08s
INFO: 1 process: 1 internal.
ERROR: Build did NOT complete successfully
```

pre-commit runs bazel, and GitHub runners have an old clang by default (caught by the new check), so I'm installing here for a consistent version.

### DIFF
--- a/bazel/cc_toolchains/clang_configuration.bzl
+++ b/bazel/cc_toolchains/clang_configuration.bzl
@@ -107,8 +107,8 @@ def _compute_bsd_sysroot(repository_ctx):
 # File content used when computing search paths. This additionally verifies that
 # libc++ is installed.
 _CLANG_INCLUDE_FILE_CONTENT = """
-#if __has_include(<utility>)
-#include <utility>
+#if __has_include(<version>)
+#include <version>
 #endif
 #ifndef _LIBCPP_STD_VER
 #error "No libc++ install found!"
--- a/bazel/cc_toolchains/clang_configuration.bzl
+++ b/bazel/cc_toolchains/clang_configuration.bzl
@@ -121,7 +121,7 @@ def _compute_clang_cpp_include_search_paths(repository_ctx, clang, sysroot):
     Returns the resulting paths as a list of strings.
     """
 
-    # Create file for Clang to use as input.
+    # Create a file for Clang to use as input.
     repository_ctx.file("_temp", _CLANG_INCLUDE_FILE_CONTENT)
     input_file = repository_ctx.path("_temp")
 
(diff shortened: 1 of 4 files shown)


docs/project/contribution_tools.md
```diff
@@ -255,6 +254,12 @@ work reliably include:
 
 ## Troubleshooting build issues
 
+### `bazel clean`
+
+Changes to packages installed on your system may not be noticed by `bazel`. This
+includes things such as changing LLVM versions, or installing libc++. Running
+`bazel clean` should force cached state to be rebuilt.
```
danakj, line 261: Can we leave a breadcrumb of the exact error here, so a search will turn it up as well?

Suggestion:

As one example, Bazel may raise "absolute path inclusion(s) found in rule" and "includes the following non-builtin files with absolute paths" as an error for libc++ includes if the location of the standard library header files changes.

jonmeow: I'm now tackling the libc++ issue down a different route, just validation during repo setup (also clang versions that we'd discussed in the past)

bazel/cc_toolchains/clang_configuration.bzl
```diff
@@ -100,24 +104,26 @@ def _compute_bsd_sysroot(repository_ctx):
         return sysroot_path.realpath
     return default
 
+# File content used when computing search paths. This additionally verifies that
+# libc++ is installed.
+_CLANG_INCLUDE_FILE_CONTENT = """
+#if __has_include(<utility>)
+#include <utility>
```
chandlerc: As suggested by @zygoloid -- can probably even more cleanly use:

```suggestion
#if __has_include(<version>)
#include <version>
```

chandlerc (approved):
LGTM, but maybe good to double check that with the precise & functional testing of `libc++` now this has enough info in the troubleshooting from @danakj's perspective before merging?

danakj (approved):
LGTM, thanks!

PR #4474, opened 2024-11-02
Refactor subcommand addition for sharing.

jonmeow:
Trying to standardize the setup a little more.

### DIFF
--- a/toolchain/driver/driver_subcommand.h
+++ b/toolchain/driver/driver_subcommand.h
@@ -5,6 +5,7 @@
 #ifndef CARBON_TOOLCHAIN_DRIVER_DRIVER_SUBCOMMAND_H_
 #define CARBON_TOOLCHAIN_DRIVER_DRIVER_SUBCOMMAND_H_
 
+#include "common/command_line.h"
 #include "common/ostream.h"
 #include "llvm/Support/VirtualFileSystem.h"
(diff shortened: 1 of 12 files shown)


toolchain/driver/driver_subcommand.h
```diff
@@ -25,7 +26,28 @@ struct DriverResult {
 // A subcommand for the driver.
 class DriverSubcommand {
  public:
+  explicit DriverSubcommand(CommandLine::CommandInfo info) : info_(info) {}
+
+  // Adds the subcommand to the main command, assigning `selected_command` when
+  // the subcommand is in use.
+  auto AddSubcommand(CommandLine::CommandBuilder& b,
```
chandlerc: Calling this method `AddSubcommand` ends up a bit confusing for me because the relationship between `this` and `b` are reversed here vs. the analogous method on `b`.

Maybe a slightly different name would make the relationship more clear?
`RegisterWith` or `AddTo`?

Happy for that to be in this PR or a follow-up if that's easier for you.

chandlerc (approved):
LG with one comment inline.

toolchain/driver/driver_subcommand.h
jonmeow: Going with `AddTo`

PR #4475, opened 2024-11-02
Rename `IntLiteral` to `IntValue`.

zygoloid:
This instruction represents integer values, whether they come from literals or calculations, so it the old name is inaccurate. I also plan to rename `BigInt` to `IntLiteral` based on recent discussion and this change aims to avoid confusion stemming from the same name being used for two different things.

I'm not renaming `FloatLiteral` because recent discussion suggests we may want distinct `FloatLiteral` versus `FloatValue` representations in SemIR.


chandlerc (approved):
Spot checked, and looks good.

LGTM for merging as it seems easy to just do another rename even if we decide a different name would be better, and this unblocks subsequent renames / avoids merge conflicts.

PR #4476, opened 2024-11-02
Rename `BigInt` to `IntLiteral`.

zygoloid:
In preparation for changing the type of integer literals to `IntLiteral`.


chandlerc (approved):
Spot checked the mechanical changes and they seem good.

PR #4478, opened 2024-11-04
Correct typo in p0144 Numeric literal semantics

danakj:
This was already corrected in the design at
https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/expressions/literals.md#implicit-conversions

The text means to say "rejected" but says "represented".


danakj:
https://github.com/carbon-language/carbon-lang/blob/26d7717d606e59cb0cce7a324962c146b469c4d0/proposals/p0144.md?plain=1#L139-L140

I was wondering these meant to say IntLiteral() and FloatLiteral() as well?

Edit: Answering my own question. The [design for Defined Types](https://github.com/carbon-language/carbon-lang/blob/trunk/docs/design/expressions/literals.md#defined-types) says the same and provides a link. It means the Int(N)/Float(N) builtin types which have aliases like i32 etc, as in runtime values instead of literals.

zygoloid (approved):

danakj:
Thanks, could you also approve the workflows, they seem to be separate?

PR #4479, opened 2024-11-04
Remove param_refs and implicit_param_refs

geoffromer:
This introduces `calling_convention_param_ids`, a single block that consolidates all the information that was being used by consumers of `param_refs` and `implicit_param_refs`, in a form that's easier to produce and typically easier to consume.

See also [this Discord discussion](https://discord.com/channels/655572317891461132/655578254970716160/1300545448909738125) regarding the decision to keep the return slot last in the SemIR calling convention, even though it goes first in the LLVM calling convention.


toolchain/check/convert.h
```diff
@@ -97,18 +97,14 @@ auto ConvertForExplicitAs(Context& context, Parse::NodeId as_node,
 struct CalleeParamsInfo {
```
jonmeow: Just to note, #4452 is now merged

toolchain/sem_ir/entity_with_params_base.h
```diff
@@ -95,21 +94,21 @@ struct EntityWithParamsBase {
...
   // pattern insts pertaining to that parameter.
   InstBlockId param_patterns_id;
+  // A block containing, for each calling convention parameter (including the
+  // return slot), a reference to the instruction in the function's declaration
+  // block that represents that parameter. These instructions will be in the
+  // `AnyParam` category. This is not populated on imported entities, because it
+  // is relevant only for the function definition.
+  // TODO: can this be moved to `Function`, since it is not applicable to other
+  // kinds of entities?
+  InstBlockId calling_convention_params_id;
```
jonmeow: @zygoloid Can you just confirm that this is what you're expecting?

toolchain/sem_ir/entity_with_params_base.h
```diff
@@ -95,21 +94,21 @@ struct EntityWithParamsBase {
...
-  InstBlockId param_refs_id;
-  // A block containing, for each explicit parameter, a reference to the
   // instruction in the entity's pattern block that depends on all other
   // pattern insts pertaining to that parameter.
   InstBlockId param_patterns_id;
+  // A block containing, for each calling convention parameter (including the
+  // return slot), a reference to the instruction in the function's declaration
+  // block that represents that parameter. These instructions will be in the
+  // `AnyParam` category. This is not populated on imported entities, because it
+  // is relevant only for the function definition.
```
jonmeow: What's the order of return slot, parameters, and implicit parameters?

toolchain/lower/file_context.cpp
```diff
@@ -377,12 +351,17 @@ auto FileContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
...
+    sequential_param_ids = sequential_param_ids.drop_back();
+  }
 
   // The LLVM calling convention has the return slot first rather than last.
-  if (has_return_slot) {
+  // Note that this queries whether there is a return slot at the LLVM level,
+  // whereas `function.return_slot_id.is_valid()` queries whether there is a
+  // return slot at the return slot at the SemIR level.
+  if (SemIR::ReturnTypeInfo::ForFunction(sem_ir(), function, specific_id)
+          .has_return_slot()) {
```
jonmeow: It's interesting to me that these two `if` conditions are separate. Are there cases where:

1. `return_slot_id.is_valid()` returns false, but `has_return_slot` returns true?
2. `has_return_slot` returns false, but `return_slot_id.is_valid()` returns true?

If one case exists and the other does not, perhaps the `if` should be nested?

Given the addition of `calling_convention_param_ids`, maybe you could replace `sequential_param_ids` and write this as something like:

```
if (return_slot_id.is_valid()) {
  // If necessary, you could write `if (has_return_slot)` here
  lower_param(calling_convention_param_ids.back());
  calling_convention_param_ids = calling_convention_param_ids.drop_back();
}
```

toolchain/check/pattern_match.h
```diff
@@ -11,14 +11,13 @@
 namespace Carbon::Check {
 
 // The pattern-match counterparts of the patterns passed to CalleePatternMatch.
+// FIXME rename, recomment.
```
jonmeow: Was this missed?

```suggestion
```

toolchain/check/pattern_match.cpp
```diff
@@ -290,39 +295,38 @@ auto CalleePatternMatch(Context& context,
...
   }
 
-  if (return_slot_pattern_id.is_valid()) {
-    match.AddWork({.pattern_id = return_slot_pattern_id,
-                   .scrutinee_id = SemIR::InstId::Invalid});
-    CARBON_CHECK(match.DoWork(context) == SemIR::InstBlockId::Empty);
+  if (!return_slot_pattern_id.is_valid() && !param_patterns_id.is_valid() &&
+      !implicit_param_patterns_id.is_valid()) {
+    return {.calling_convention_params_id = SemIR::InstBlockId::Invalid,
+            .return_slot_id = SemIR::InstId::Invalid};
```
jonmeow: Had you considered checking this at the top, before the `MatchContext` construction?

toolchain/check/pattern_match.cpp
```diff
@@ -277,6 +281,7 @@ auto MatchContext::EmitPatternMatch(Context& context,
           pattern.loc_id, {.type_id = return_slot_pattern.type_id,
                            .type_inst_id = return_slot_pattern.type_inst_id,
                            .storage_id = entry.scrutinee_id});
+      results_.push_back(entry.scrutinee_id);
```
jonmeow, line 287: For my own understanding, why is this pushed now but not before?

toolchain/check/handle_impl.cpp
```diff
@@ -205,14 +205,17 @@ static auto PopImplIntroducerAndParamsAsNameComponent(
...
-    parameter_blocks =
+    auto parameter_blocks =
         CalleePatternMatch(context, *implicit_param_patterns_id,
                            SemIR::InstBlockId::Invalid, SemIR::InstId::Invalid);
+    CARBON_CHECK(parameter_blocks.calling_convention_params_id ==
+                     SemIR::InstBlockId::Empty,
+                 "{0} {1}", parameter_blocks.calling_convention_params_id,
+                 context.inst_blocks()
+                     .Get(parameter_blocks.calling_convention_params_id)
+                     .size());
```
jonmeow: Would it be worth elaborating on this CHECK message with explanatory text? What's an example look like in a failure?

(or, if this was temporary for debugging an issue, should it be removed?)

toolchain/check/handle_function.cpp
```diff
@@ -342,26 +340,28 @@ static auto HandleFunctionDefinitionAfterSignature(
...
   // Check the parameter types are complete.
-  for (auto param_ref_id : llvm::concat<const SemIR::InstId>(
-           context.inst_blocks().GetOrEmpty(function.implicit_param_refs_id),
-           context.inst_blocks().GetOrEmpty(function.param_refs_id))) {
+  for (auto param_ref_id : params_to_complete) {
     if (param_ref_id == SemIR::InstId::BuiltinError) {
       continue;
     }
-    auto param_info =
-        SemIR::Function::GetParamFromParamRefId(context.sem_ir(), param_ref_id);
```
jonmeow, line 352: I see you removing several references to `GetParamFromParamRefId`, is it still in use?

jonmeow (commented):
Should the PR description have some more explanation, like a link to the discussion about this and a mention of what it's being replaced with?

toolchain/check/convert.h
geoffromer: I expect the merge conflicts to be pretty straightforward, but I can do the merge during the review instead of the end if you like.

toolchain/check/handle_function.cpp
geoffromer: Oh, good catch! No, it's not; removed now.

toolchain/check/handle_impl.cpp
geoffromer: Yeah, the message is left over from debugging. I've removed it.

toolchain/check/pattern_match.cpp
geoffromer: Because we're now computing the calling-convention parameter list, and the return slot is part of the calling convention.

Previously we were computing it separately (see line 318 of the status quo) and exposing it separately (as `return_slot_id` in places like `Function` and `NameComponent`). With this PR, we're computing it together with the rest of the calling convention parameters, and exposing it both as a `return_slot_id` member and in the parameter list, as a transitional step toward removing the `return_slot_id` members. That cleanup will happen in the next PR, which as a byproduct will remove `return_slot_id_` from this class, making this line the only way this case produces output.

toolchain/check/pattern_match.cpp
geoffromer: Yeah, that's a bit cleaner. Done.

toolchain/check/pattern_match.h
geoffromer: Oh, whoops, yes.

I've updated the comment, but if it's OK with you I think it'll be simpler to keep the name unchanged for now, even though it's not a great fit, because the next PR is going to eliminate this struct altogether.

toolchain/lower/file_context.cpp
geoffromer: > It's interesting to me that these two `if` conditions are separate. Are there cases where:
> 
> 1. `return_slot_id.is_valid()` returns false, but `has_return_slot` returns true?
> 2. `has_return_slot` returns false, but `return_slot_id.is_valid()` returns true?
> 
> If one case exists and the other does not, perhaps the `if` should be nested?

I believe 2 is possible but 1 is not, so I've nested the conditions.

> 
> Given the addition of `calling_convention_param_ids`, maybe you could replace `sequential_param_ids` and write this as something like:
> 
> ```
> if (return_slot_id.is_valid()) {
>   // If necessary, you could write `if (has_return_slot)` here
>   lower_param(calling_convention_param_ids.back());
>   calling_convention_param_ids = calling_convention_param_ids.drop_back();
> }
> ```

I considered that, but decided not to because it could be confusing to have a variable named `calling_convention_param_ids` that doesn't actually contain the full list of calling convention params. I could do it the other way around by dropping `calling_convention_param_ids` and using `sequential_param_ids` everywhere, but that has the same problem: `sequential_param_ids` wouldn't be in sequential order until the end of this branch (and the code within the branch would be relying on that fact).

geoffromer (commented):
> Should the PR description have some more explanation, like a link to the discussion about this and a mention of what it's being replaced with?

Good point, I've fleshed out the description, but what discussion are you referring to?

toolchain/check/pattern_match.h
jonmeow: That's fine, I just think we prefer TODO for anything being checked in, and when I see FIXME I read it as "meant to fix prior to review"

toolchain/check/pattern_match.cpp
jonmeow: Thanks!

toolchain/lower/file_context.cpp
```diff
@@ -348,21 +348,21 @@ auto FileContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
...
     sequential_param_ids = sequential_param_ids.drop_back();
+    // The LLVM calling convention has the return slot first rather than last.
+    // Note that this queries whether there is a return slot at the LLVM level,
+    // whereas `function.return_slot_id.is_valid()` queries whether there is a
+    // return slot at the SemIR level.
+    if (SemIR::ReturnTypeInfo::ForFunction(sem_ir(), function, specific_id)
+            .has_return_slot()) {
+      lower_param(calling_convention_param_ids.back());
+    }
   }
```
```suggestion
  llvm::ArrayRef<SemIR::InstId> sequential_param_ids;
  if (function.return_slot_id.is_valid()) {
    // The LLVM calling convention has the return slot first rather than last.
    // Note that this queries whether there is a return slot at the LLVM level,
    // whereas `function.return_slot_id.is_valid()` queries whether there is a
    // return slot at the SemIR level.
    if (SemIR::ReturnTypeInfo::ForFunction(sem_ir(), function, specific_id)
            .has_return_slot()) {
      lower_param(calling_convention_param_ids.back());
    }
    sequential_param_ids = calling_convention_param_ids.drop_back();
  } else {
    sequential_param_ids = calling_convention_param_ids;
  }
```

Optional suggestion, was just wondering if you might find this flow easier to read.

toolchain/check/pattern_match.cpp
jonmeow: Why do you still construct `MatchContext` before the comparison?

jonmeow:
> > Should the PR description have some more explanation, like a link to the discussion about this and a mention of what it's being replaced with?
> 
> Good point, I've fleshed out the description, but what discussion are you referring to?

https://discord.com/channels/655572317891461132/655578254970716160/1300545448909738125

toolchain/sem_ir/entity_with_params_base.h
```diff
@@ -95,21 +94,23 @@ struct EntityWithParamsBase {
...
-  // pattern-match insts pertaining to that parameter.
-  InstBlockId param_refs_id;
-  // A block containing, for each explicit parameter, a reference to the
   // instruction in the entity's pattern block that depends on all other
   // pattern insts pertaining to that parameter.
   InstBlockId param_patterns_id;
+  // A block containing, for each calling convention parameter, a reference to
+  // the instruction in the function's declaration block that represents that
+  // parameter. These parameters appear in declaration order: `self` (if
+  // present), then the explicit runtime parameters, then the return slot (which
```
jonmeow: Right now, the only implicit parameter the toolchain supports is `self`. However, once more are supported, won't those become part of the calling convention? Also, are you intending to exclude non-runtime explicit parameters (do we have those?)?

You might consider something like:

```suggestion
  // parameter. Parameters are in declaration order: implicit parameters, explicit parameters, then the return slot (which
```

toolchain/check/name_component.h
```diff
@@ -28,14 +28,16 @@ struct NameComponent {
...
   Parse::NodeId implicit_params_loc_id;
-  SemIR::InstBlockId implicit_params_id;
   SemIR::InstBlockId implicit_param_patterns_id;
 
   // The explicit parameter list.
   Parse::NodeId params_loc_id;
-  SemIR::InstBlockId params_id;
   SemIR::InstBlockId param_patterns_id;
 
+  // The calling convention parameters of the function. These will all be
```
jonmeow:
```suggestion
  // The calling convention parameters of the entity. These will all be
```

Won't this be set for all parameterized entities, not only functions?

toolchain/sem_ir/entity_with_params_base.h
jonmeow: Just to note, one thing I'd particularly want to be sure of is that you both understand what "calling convention parameter" is, that it's a term I'm just not familiar with. To me, that sounds distinctly qualified from "implicit" or "explicit", not a union. If you're still considering names, I would suggest something like "full parameters" or "signature parameters" for this.

jonmeow (commented):
Just to note, I'm basically fine with this, but would like to make sure zygoloid is happy with this, particularly the naming (which throws me).

toolchain/check/name_component.h
geoffromer: I believe functions are the only entities that have a calling convention (because they're the only entities that can be called with runtime arguments).

toolchain/sem_ir/entity_with_params_base.h
geoffromer: I'm intending to exclude all non-runtime parameters, because they aren't part of the calling convention (i.e. they aren't passed as part of the function call). Unless I'm missing something, `self` is the only non-compile-time parameter that can appear in the implicit parameter list.

toolchain/check/pattern_match.h
geoffromer: Oh, absolutely, the FIXME was just an oversight.

toolchain/check/pattern_match.cpp
geoffromer: No good reason. Fixed.

toolchain/sem_ir/entity_with_params_base.h
geoffromer: See `pattern_match.h` for the definition of that term (such as it is). Do you think it would make sense to move that here?

toolchain/check/name_component.h
zygoloid: Would it be reasonable to add a TODO similar to the one for the return slot fields here?

I think it'd also be useful to specify more explicitly what you mean by "calling convention parameters" here. Maybe overall something like:

> The calling convention parameters of this entity, if it is a function. These are the parameters that correspond to the arguments in a `Call` instruction that calls this function. These will all be instances of `AnyParam`.

I also wonder if a shorter name like `call_params_id` or maybe `call_inst_params_id` might capture the same meaning. ("Calling convention" sounds to me like it might be talking about ABI, which I think is separate.)

zygoloid: Hm, or maybe this should reference `EntityWithParamsBase` for the full explanation of this field. I think the TODO and the change from "the function" to "the entity if it is a function" would still be useful here, though.

toolchain/sem_ir/entity_with_params_base.h
```diff
@@ -95,21 +94,23 @@ struct EntityWithParamsBase {
...
   InstBlockId implicit_param_patterns_id;
   // A block containing, for each explicit parameter, a reference to the
-  // instruction in the entity's declaration block that depends on all other
-  // pattern-match insts pertaining to that parameter.
-  InstBlockId param_refs_id;
-  // A block containing, for each explicit parameter, a reference to the
   // instruction in the entity's pattern block that depends on all other
   // pattern insts pertaining to that parameter.
   InstBlockId param_patterns_id;
+  // A block containing, for each calling convention parameter, a reference to
```
zygoloid: I think it'd be useful to explain what you mean by "calling convention parameter" here: that is, this list is the parameters corresponding to the arguments in a `Call` instruction.

toolchain/sem_ir/entity_with_params_base.h
jonmeow: I think this would be the more central location to document a a term. However, I've also brought this up on [#naming](https://discord.com/channels/655572317891461132/963846118964350976/1305655302947082262).

toolchain/check/name_component.h
geoffromer: Done, and renamed to `call_params_id` as discussed at today's meeting.

toolchain/sem_ir/entity_with_params_base.h
geoffromer: Done.

toolchain/sem_ir/entity_with_params_base.h
```diff
@@ -95,21 +94,27 @@ struct EntityWithParamsBase {
...
-  // pattern-match insts pertaining to that parameter.
-  InstBlockId param_refs_id;
-  // A block containing, for each explicit parameter, a reference to the
   // instruction in the entity's pattern block that depends on all other
   // pattern insts pertaining to that parameter.
   InstBlockId param_patterns_id;
+  // If this entity is a function, this block consists of references to the
+  // `AnyParam` insts that represent the function's call parameters. The "call
+  // parameters" are the parameters corresponding to the arguments that are
+  // passed to a `Call` inst, so they include `self` (if applicable) and the
```
jonmeow: Why does this say "call parameters" instead of "`Call`"? Had you considered something like:

```suggestion
  // `AnyParam` insts that represent the function's `Call` parameters. They include `self` (if applicable) and the
```

toolchain/sem_ir/entity_with_params_base.h
```diff
@@ -95,21 +94,27 @@ struct EntityWithParamsBase {
...
+  // return slot, but do not include compile-time parameters.
+  //
+  // The parameters appear in declaration order: `self` (if present), then the
+  // explicit runtime parameters, then the return slot (which is "declared" by
+  // the function's return type declaration). This is not populated on imported
```
jonmeow: You note the things which are included twice here, had you considered removing one of the mentions?

toolchain/lower/file_context.cpp
```diff
@@ -374,16 +348,22 @@ auto FileContext::BuildFunctionDefinition(SemIR::FunctionId function_id)
...
-  llvm::ArrayRef<SemIR::InstId> sequential_param_ids =
-      calling_convention_param_ids;
-
-  // The LLVM calling convention has the return slot first rather than last.
-  if (has_return_slot) {
-    lower_param(calling_convention_param_ids.back());
-
-    sequential_param_ids = sequential_param_ids.drop_back();
+  // The subset of call_param_ids that is in sequential order.
+  llvm::ArrayRef<SemIR::InstId> sequential_param_ids;
```
jonmeow: In the new setup, isn't `call_param_ids` a sequential block? Could you explain what "sequential order" means in this context, or maybe the naming should be updated?

jonmeow (commented):
At a high-level, I think we were particularly discussing "`Call` parameters", i.e. referring to the specific entity, rather than "call parameters". I think that makes a difference in documentation, with one example below. What do you think of changing how you're referring to that?

toolchain/sem_ir/entity_with_params_base.h
geoffromer: I want to be a little more explicit about the fact that I'm defining this as a term of art, particularly so that it can be reused in places like `name_component.h`. It hadn't occurred to me to say "`Call`"  instead of "call", but I suppose that works; done.

toolchain/sem_ir/entity_with_params_base.h
geoffromer: I thought about it, but I couldn't find a way to do it without creating other problems. When I'm defining "`Call` parameters" it seems important to give examples of how it differs from other meanings of "parameters", and at some point I need to specify the parameter order fairly precisely since all the code that uses this field needs to agree on that, but combining the two forces me to front-load the ordering specification, which seems like it's dumping too much information on the reader all at once.

The `self` example in the first paragraph wasn't adding much, so I've at least removed that, but I'm not sure what else to do.

toolchain/lower/file_context.cpp
geoffromer: I've rephrased the comment: these are the parameters that are already in the order that the LLVM calling convention expects, so we can just lower them in a loop. I'm having trouble coming up with a better name to capture that idea, but I'm open to suggestions.

toolchain/sem_ir/ids.h
```diff
@@ -276,6 +276,8 @@ constexpr CompileTimeBindIndex CompileTimeBindIndex::Invalid =
 // arguments passed to them at runtime. In a `call` instruction, a runtime
 // argument will have the position in the argument list corresponding to its
 // runtime parameter index.
+// TODO: Rename this to CallParamIndex, for consistency with the "call
```
```suggestion
// TODO: Rename this to CallParamIndex, for consistency with the "`Call`
```

toolchain/check/pattern_match.h
```diff
@@ -30,25 +29,23 @@ struct ParameterBlocks {
...
-// the corresponding semantic placeholders that they bind to.
+// See EntityWithParamsBase::call_params for a discussion of "call parameters"
 
 // Emits the pattern-match IR for the declaration of a parameterized entity with
 // the given implicit and explicit parameter patterns, and the given return slot
 // pattern (any of which may be invalid if not applicable). This IR performs the
 // callee side of pattern matching, starting at the `ParamPattern` insts, and
-// matching them against the corresponding calling-convention parameters.
+// matching them against the corresponding call parameters (see
+// entity_with_params_base.h for the definition of that term).
```
jonmeow: Missed this, maybe?

```suggestion
// matching them against the corresponding `Call` parameters.
```

toolchain/check/pattern_match.h
```diff
@@ -30,25 +29,23 @@ struct ParameterBlocks {
 // are matched by the callee, and pattern insts that have a `ParamPattern`
 // as a descendant are matched by the caller.
 //
-// "Calling convention arguments" are the values actually passed from caller to
-// callee at the semantic IR level, and "calling convention parameters" are
-// the corresponding semantic placeholders that they bind to.
+// See EntityWithParamsBase::call_params for a discussion of "call parameters"
```
jonmeow: Is this carrying its weight if you fix this to be `Call` parameters? Note too, you had a similar reference below. I could comment about punctuation here, but maybe it should just be removed?

(can't make a suggested edit for this since I'd suggest also removing the blank comment line above)

toolchain/check/pattern_match.h
```diff
@@ -10,15 +10,14 @@
 
 namespace Carbon::Check {
 
-// The pattern-match counterparts of the patterns passed to CalleePatternMatch.
+// The outputs of CalleePatternMatch.
+// TODO: rename or remove this struct.
```
jonmeow:
```suggestion
// TODO: Rename or remove this struct.
```

Since we've been talking about TODO comments a lot, I'll note all the style examples at https://google.github.io/styleguide/cppguide.html#TODO_Comments use the same sentence capitalization; it probably makes sense to be consistent.

toolchain/check/pattern_match.h
```diff
@@ -10,15 +10,14 @@
...
-  // The implicit parameter list.
-  SemIR::InstBlockId implicit_params_id;
-
-  // The explicit parameter list.
-  SemIR::InstBlockId params_id;
+  // The call parameters of the function.
+  SemIR::InstBlockId call_params_id;
 
   // The return slot.
+  // TODO: drop this and just use the last element of above?
```
jonmeow:
```suggestion
  // TODO: Drop this and just use the last element of above?
```

(see style comment above)

toolchain/check/handle_impl.cpp
```diff
@@ -205,14 +205,12 @@ static auto PopImplIntroducerAndParamsAsNameComponent(
...
-      .implicit_params_id = SemIR::InstBlockId::Invalid,
-      .params_id = SemIR::InstBlockId::Invalid,
-      .return_slot_id = SemIR::InstId::Invalid};
   if (implicit_param_patterns_id) {
-    parameter_blocks =
+    auto parameter_blocks =
         CalleePatternMatch(context, *implicit_param_patterns_id,
                            SemIR::InstBlockId::Invalid, SemIR::InstId::Invalid);
+    CARBON_CHECK(parameter_blocks.call_params_id == SemIR::InstBlockId::Empty);
+    CARBON_CHECK(parameter_blocks.return_slot_id == SemIR::InstId::Invalid);
```
jonmeow, line 208: Can you elaborate what's going on here? Previously, the result of `parameter_blocks.implicit_params_id` was retained. What's the effect of running it when the result is discarded?

toolchain/sem_ir/entity_with_params_base.h
```diff
@@ -95,21 +94,27 @@ struct EntityWithParamsBase {
...
+  // "`Call` parameters" are the parameters corresponding to the arguments that
+  // are passed to a `Call` inst, so they do not include compile-time
+  // parameters, but they do include the return slot.
+  //
+  // The parameters appear in declaration order: `self` (if present), then the
+  // explicit runtime parameters, then the return slot (which is "declared" by
+  // the function's return type declaration). This is not populated on imported
+  // functions, because it is relevant only for a function definition.
+  //
+  // TODO: can this be moved to `Function`, since it is not applicable to other
```
jonmeow:
```suggestion
  // TODO: Can this be moved to `Function`, since it is not applicable to other
```

Capitalization for style (https://google.github.io/styleguide/cppguide.html#TODO_Comments)

toolchain/sem_ir/entity_with_params_base.h
jonmeow: I think extra documentation can help, but I want to focus on this:

> this block consists of references to the `AnyParam` insts that represent the function's `Call` parameters. The "`Call` parameters" are the parameters corresponding to the arguments that are passed to a `Call` inst

Isn't this just saying that the `Call` parameters are parameters to the `Call` instruction? I end up a little uncertain because, to me, when I see something that looks repetitive, I tend to assume that I'm actually misreading and missing a nuance. As a consequence, the repetition actually makes this _harder_ for me to read, not easier. Maybe there's some more terse way of stating this, such as "`Call` inst parameters" if you're concerned about non-instruction interpretations of just "`Call`"?

Although, this also raises another question. Maybe this comment should be on `Call` so that when we say "`Call` parameters" the answer is found on `Call` without needing to say "look at `EntityWithParamsBase`"? Sorry about this, since I think this has already moved around a little.

toolchain/check/handle_impl.cpp
geoffromer: The primary job of `CalleePatternMatch` is to emit the pattern-match IR, and we still want to do that here, even though we no longer use the return value.

toolchain/check/pattern_match.h
geoffromer: This is no longer needed because it's redundant with the citation below.

toolchain/check/pattern_match.h
geoffromer: I've fixed the spelling of "call", but kept the citation, because I don't think "`Call` parameters" is sufficiently self-explanatory (see the comment thread in `entity_with_params_base.h`).

toolchain/sem_ir/entity_with_params_base.h
geoffromer: > Isn't this just saying that the `Call` parameters are parameters to the `Call` instruction?

I wouldn't put it that way, no. `EntityWithParamsBase` represents part of the declaration of a callable entity; it doesn't represent a particular call to that entity, so there's no `Call` inst in "scope" here. By the same token, `Call` doesn't have parameters, it has arguments. Does that make sense?

> Although, this also raises another question. Maybe this comment should be on `Call` so that when we say "`Call` parameters" the answer is found on `Call` without needing to say "look at `EntityWithParamsBase`"? Sorry about this, since I think this has already moved around a little.

That could work, but I think it would be awkward, for the same reason: the `Call` inst represents a particular call to a callable entity, so I think it would feel like a non sequitur to talk about how that callable entity represents its parameters in its own IR.

(While looking at this I noticed that the `Call` documentation is out of date, so I fixed that.)

toolchain/check/handle_impl.cpp
jonmeow: Thanks, what do you think of adding a comment here to explain the code? I've taken a stab above to try to clarify what I'm requesting.

toolchain/check/handle_impl.cpp
```diff
@@ -205,14 +205,12 @@ static auto PopImplIntroducerAndParamsAsNameComponent(
   auto [implicit_params_loc_id, implicit_param_patterns_id] =
       context.node_stack().PopWithNodeIdIf<Parse::NodeKind::ImplForall>();
 
-  ParameterBlocks parameter_blocks{
-      .implicit_params_id = SemIR::InstBlockId::Invalid,
-      .params_id = SemIR::InstBlockId::Invalid,
-      .return_slot_id = SemIR::InstId::Invalid};
   if (implicit_param_patterns_id) {
-    parameter_blocks =
+    auto parameter_blocks =
```
jonmeow, line 204:
```suggestion
    // Emit the `forall` match. We don't need to store references because it's compile-time only.
    auto parameter_blocks =
```

toolchain/check/pattern_match.h
```diff
@@ -10,15 +10,14 @@
...
-// The pattern-match counterparts of the patterns passed to CalleePatternMatch.
+// The outputs of CalleePatternMatch.
+// TODO: Rename or remove this struct.
 struct ParameterBlocks {
-  // The implicit parameter list.
-  SemIR::InstBlockId implicit_params_id;
-
-  // The explicit parameter list.
-  SemIR::InstBlockId params_id;
+  // The call parameters of the function.
```
jonmeow: Sorry, I missed this last time.
```suggestion
  // The `Call` parameters of the function.
```

toolchain/sem_ir/entity_with_params_base.h
jonmeow: > > Isn't this just saying that the `Call` parameters are parameters to the `Call` instruction?
> 
> I wouldn't put it that way, no. `EntityWithParamsBase` represents part of the declaration of a callable entity; it doesn't represent a particular call to that entity, so there's no `Call` inst in "scope" here. By the same token, `Call` doesn't have parameters, it has arguments. Does that make sense?

FWIW, here's a different phrasing that I don't know if it'd work for you:

```
`Call` parameters correspond to `Call::args_id`, but in parameter form.
```

I'll leave it to you whether to adopt something like that.

> > Although, this also raises another question. Maybe this comment should be on `Call` so that when we say "`Call` parameters" the answer is found on `Call` without needing to say "look at `EntityWithParamsBase`"? Sorry about this, since I think this has already moved around a little.
> 
> That could work, but I think it would be awkward, for the same reason: the `Call` inst represents a particular call to a callable entity, so I think it would feel like a non sequitur to talk about how that callable entity represents its parameters in its own IR.
> 
> (While looking at this I noticed that the `Call` documentation is out of date, so I fixed that.)

To be clear here, what I really mean is instead of having things like "(see entity_with_params_base.h for the definition of that term)." in `pattern_match.h`, maybe having it on `Call` would mean that people seeing "`Call` parameters" and wondering what that means would be able to find it on `Call` without the explicit reference. I'm not meaning to argue the detail in this half of the comment.

jonmeow (approved):
LGTM. Some lingering comments, including one more "call parameters" I noticed, but generally I think this is fine to merge. To be clear, feel free to merge regardless of how you resolve remaining comments. Sorry about the long review here.

toolchain/sem_ir/typed_insts.h
```diff
@@ -431,10 +431,9 @@ struct Call {
 
   TypeId type_id;
   InstId callee_id;
-  // The arguments block contains IDs for the following arguments, in order:
-  //  - The argument for each implicit parameter.
-  //  - The argument for each explicit parameter.
-  //  - The argument for the return slot, if present.
+  // The arguments block contains IDs for the arguments corresponding to each
+  // runtime parameter, in lexical order of the parameter declarations, followed
```
jonmeow:
```suggestion
  // Runtime arguments in lexical order of the parameter declarations, followed
```

toolchain/check/handle_impl.cpp
geoffromer: Done, with somewhat different wording.

toolchain/sem_ir/entity_with_params_base.h
geoffromer: > FWIW, here's a different phrasing that I don't know if it'd work for you:
> 
> ```
> `Call` parameters correspond to `Call::args_id`, but in parameter form.
> ```
> 
> I'll leave it to you whether to adopt something like that.

I'm afraid the "in parameter form" part doesn't work for me -- they don't just have different forms, they have different content, because they represent categorically different entities.

That said, I don't want to just shoot down your suggestions; this concept is kind of subtle, so it's important for the explanation to be clear, and evidently it isn't. If you have any thoughts about how to fix it, I'm happy to iterate more in a follow-up PR.

PR #4487, opened 2024-11-05
Canonicalize away bit width and embed small integers into `IntId`s

chandlerc:
The first change here is to canonicalize away bit width when tracking
integers in our shared value store. This lets us have a more definitive
model of "what is the mathematical value". It also frees us to use more
efficient bit widths when available, such as bits inside the ID itself.

For canonicalizing, we try to minimize the width adjustments and
maximize the use of the SSO in APInt, and so we never shrink belowe
64-bits and grow in multiples of the word bit width in the
implementation. We also canonicalize to the signed 2s compliment
representation so we can represent negative numbers in an intuitive way.

The canonicalizing requires getting the bit width out of the type and
adjusting to it within the toolchain when doing any kind of math, and
this PR updates various places to do that, as well as adding some
convenience APIs to assist.

Then we take advantage of the canonical form and embed small integers
into the ID itself rather than allocating storage for them and
referencing them with an index. This is especially helpful for the
pervasive small integers such as the sizes of types, arrays, etc. Those
no longer require indirection at all. Various short-cut APIs to take
advantage of this have also been added.

This PR improves lexing by about 5% when there are lots of `i32` types.


toolchain/base/int_store.h
```diff
@@ -0,0 +1,186 @@
...
+// different bit width `APInt` than was used when adding. There are also
+// optimized paths for adding integer values representable using native integer
+// types.
+//
+// Because the integers in the store are canonicalized without a specific bit
+// width there are helper functions to coerce them to a specific desired bit
+// width for use.
+//
+// This leverages a significant optimization for small integer values -- rather
+// than canonicalizing and making unique them in a `ValueStore`, they are
```
danakj:
```suggestion
// than canonicalizing and making them unique in a `ValueStore`, they are
```

toolchain/base/int_store.h
```diff
@@ -0,0 +1,186 @@
...
+// directly embedded in the `IntId` itself. Only larger integers are store in an
```
danakj:
```suggestion
// directly embedded in the `IntId` itself. Only larger integers are stored in an
```

toolchain/base/int_store.cpp
```diff
@@ -0,0 +1,60 @@
...
+  return IntId::MakeIndexOrInvalid(ap_id.index);
+}
+
+auto IntStore::OutputYaml() const -> Yaml::OutputMapping {
+  return values_.OutputYaml();
+}
+
+auto IntStore::CollectMemUsage(MemUsage& mem_usage, llvm::StringRef label) const
+    -> void {
+  values_.CollectMemUsage(mem_usage, label);
```
danakj: Most code calls Collect() and expects that to forward to CollectMemUsage or otherwise:
```suggestion
  mem_usage.Collect(std::string(label), values_);
```

toolchain/base/value_ids.h
```diff
@@ -41,20 +50,147 @@ class Real : public Printable<Real> {
...
+  //
+  // We represent this as a signed TokenIdBits-bit 2s compliment integer. The
+  // sign extension from TokenIdBits to a register size can be folded into the
+  // shift used to extract from compressed bitfield storage.
+  //
+  // We then divide the smallest 1/4th of the space to indices, and the larger
+  // 3/4ths to embedded values. For 23-bits total this still gives us 2 million
+  // unique integers larger than the embedded ones, which would be difficult to
+  // fill without exceeding the number of tokens we can lex (8 million). For
+  // non-token based integers, the indics can continue downward to the 32-bit
```
danakj:
```suggestion
  // non-token based integers, the indices can continue downward to the 32-bit
```

toolchain/check/eval.cpp
```diff
@@ -681,9 +681,58 @@ static auto PerformBuiltinBinaryIntOp(Context& context, SemIRLoc loc,
...
   }
 
-  bool overflow = false;
+  auto [lhs_is_signed, lhs_bit_width_id] =
+      context.sem_ir().GetIntTypeInfo(lhs.type_id);
+  llvm::APInt lhs_val = context.ints().GetAtWidth(lhs.int_id, lhs_bit_width_id);
+
   llvm::APInt result_val;
+
+  // First handle switch, which can directly use the canonical RHS and doesn't
```
danakj: Did you mean "shift"?
```suggestion
  // First handle shift, which can directly use the canonical RHS and doesn't
```

danakj (commented):
Reading through to try wrap my head around everything, noticed a few inconsequential things along the way.

toolchain/base/int_store.h
```diff
@@ -0,0 +1,186 @@
...
+// types.
+//
+// Because the integers in the store are canonicalized without a specific bit
+// width there are helper functions to coerce them to a specific desired bit
+// width for use.
+//
+// This leverages a significant optimization for small integer values -- rather
+// than canonicalizing and making them unique in a `ValueStore`, they are
+// directly embedded in the `IntId` itself. Only larger integers are stored in an
+// array of `APInt` values and represented as an index in the ID.
```
CarbonInfraBot, line 40: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
// directly embedded in the `IntId` itself. Only larger integers are stored in
// an array of `APInt` values and represented as an index in the ID.
```

jonmeow:
> This PR improves lexing by about 5% when there are lots of `i32` types.

What percentage of tokens/bytes being `i32` results in 5% lex improvement? Can you give a little more context for this?

toolchain/sem_ir/type.h
```diff
@@ -99,6 +99,10 @@ class TypeStore : public Yaml::Printable<TypeStore> {
   }
 
   // Determines whether the given type is a signed integer type.
+  //
+  // TODO: When we don't have a builtin int type mixed with actual `IntType`
+  // instructions, clients should directly query the `IntType` instruction to
+  // compute this information.
```
jonmeow, line 105: I'm unsure about this, even with your changes there are still a half dozen spots that call this, maybe it's worth keeping the helper for caller simplicity even if BuiltinIntType doesn't need to be handled?

Or, maybe they should be using GetIntTypeInfo to consolidate?

toolchain/sem_ir/file.h
```diff
@@ -70,6 +70,30 @@ class File : public Printable<File> {
     return types().GetAs<PointerType>(pointer_id).pointee_id;
   }
 
+  struct IntTypeInfo {
+    bool is_signed;
+    IntId bit_width = IntId::Invalid;
+  };
+
```
jonmeow: Should this be following declaration order?

Also, I'll make my usual comment request, even if you don't want to write more than "Return type for GetIntTypeInfo".

toolchain/sem_ir/file.h
```diff
@@ -70,6 +70,30 @@ class File : public Printable<File> {
...
+  // Compute the core integer type information from a type ID.
+  //
+  // TODO: When we don't have a builtin int type mixed with actual `IntType`
+  // instructions, clients should directly query the `IntType` instruction to
+  // compute this information.
+  auto GetIntTypeInfo(TypeId int_type_id) const -> IntTypeInfo {
+    auto inst_id = types().GetInstId(int_type_id);
+    if (inst_id == InstId::BuiltinIntType) {
+      return {true, ints().LookupSigned(llvm::APInt(/*numBits=*/64, 32))};
```
jonmeow:
```suggestion
      return {true, ints().LookupSigned(llvm::APInt(/*numBits=*/64, /*val=*/32))};
```

I don't know what's what here without looking it up, and the "32" is also for bit width, so commenting both parameters would help me.

Alternately, you have a private constant for the 64-bit width on your store, had you considered making that public and used by callers? (mainly tests TBH, so I'm not sure this is really valuable) I think the name could instead get you to:

```suggestion
      return {true, ints().LookupSigned(llvm::APInt(IntStore::MinAPWidth, 32))};
```

jonmeow:
```suggestion
      return {.is_signed = true, .bit_width = ints().LookupSigned(llvm::APInt(/*numBits=*/64, 32))};
```

I think we'd agreed to use designated initializers for struct types (https://github.com/carbon-language/carbon-lang/blob/trunk/docs/project/cpp_style_guide.md#syntax-and-formatting)

toolchain/sem_ir/file.h
```diff
@@ -70,6 +70,30 @@ class File : public Printable<File> {
...
+    }
+    auto int_type = insts().TryGetAs<IntType>(inst_id);
+    CARBON_CHECK(int_type,
+                 "Integer type ID associated with an unknown instruction: {0}",
+                 insts().Get(inst_id));
+    auto bit_width_inst = insts().TryGetAs<IntValue>(int_type->bit_width_id);
+    return {int_type->int_kind.is_signed(),
+            bit_width_inst ? bit_width_inst->int_id : IntId::Invalid};
```
jonmeow:
```suggestion
    return {.is_signed = int_type->int_kind.is_signed(),
            .bit_width = bit_width_inst ? bit_width_inst->int_id : IntId::Invalid};
```

As above

toolchain/sem_ir/file.h
```diff
@@ -70,6 +70,30 @@ class File : public Printable<File> {
...
+  // compute this information.
+  auto GetIntTypeInfo(TypeId int_type_id) const -> IntTypeInfo {
+    auto inst_id = types().GetInstId(int_type_id);
+    if (inst_id == InstId::BuiltinIntType) {
+      return {true, ints().LookupSigned(llvm::APInt(/*numBits=*/64, 32))};
+    }
+    auto int_type = insts().TryGetAs<IntType>(inst_id);
+    CARBON_CHECK(int_type,
+                 "Integer type ID associated with an unknown instruction: {0}",
+                 insts().Get(inst_id));
```
jonmeow:
```suggestion
    auto int_type = insts().GetAs<IntType>(inst_id);
```

Why does this use `TryGetAs` instead of `GetAs`? That'll produce a similar failure, printing the kind of both for comparison.

toolchain/sem_ir/file.h
```diff
@@ -70,6 +70,30 @@ class File : public Printable<File> {
     return types().GetAs<PointerType>(pointer_id).pointee_id;
   }
 
+  struct IntTypeInfo {
+    bool is_signed;
+    IntId bit_width = IntId::Invalid;
+  };
+
+  // Compute the core integer type information from a type ID.
```
jonmeow: Should this document the return value, particularly when it's not an integer?

toolchain/lower/constant.cpp
```diff
@@ -210,13 +210,18 @@ static auto EmitAsConstant(ConstantContext& context, SemIR::IntValue inst)
...
+  if (!int_type) {
     auto* struct_type = llvm::dyn_cast<llvm::StructType>(type);
     CARBON_CHECK(struct_type && struct_type->getNumElements() == 0);
     return llvm::ConstantStruct::get(struct_type);
   }
 
-  return llvm::ConstantInt::get(type, context.sem_ir().ints().Get(inst.int_id));
+  auto val = context.sem_ir().ints().Get(inst.int_id);
+  int bit_width = int_type->getBitWidth();
+  bool is_signed = context.sem_ir().GetIntTypeInfo(inst.type_id).is_signed;
```
jonmeow, line 222: To the comment in type.h about IsSignedInt maybe being helpful, this could be:

```suggestion
  bool is_signed = context.sem_ir().types().IsSignedInt(inst.type_id);
```

This may even be positive now, since the bit width of `inst.type_id` isn't used here, you're getting it from `int_type`

toolchain/check/import_ref.cpp
```diff
@@ -2105,7 +2105,8 @@ class ImportRefResolver {
 
     return ResolveAs<SemIR::IntValue>(
         {.type_id = context_.GetTypeIdForTypeConstant(type_id),
-         .int_id = context_.ints().Add(import_ir_.ints().Get(inst.int_id))});
+         .int_id =
+             context_.ints().AddSigned(import_ir_.ints().Get(inst.int_id))});
```
jonmeow: Is this at all worth specializing for small values in order to eliminate the APInt step?

toolchain/check/handle_literal.cpp
```diff
@@ -46,7 +46,7 @@ static auto MakeI32Literal(Context& context, Parse::NodeId node_id,
   return context.AddInst<SemIR::IntValue>(
       node_id,
       {.type_id = context.GetBuiltinType(SemIR::BuiltinInstKind::IntType),
-       .int_id = context.ints().Add(i32_val)});
+       .int_id = context.ints().AddUnsigned(i32_val)});
```
jonmeow: Since you're mostly dealing with signed values in the store, had you considered changing spots like this to use a signed APInt? IIUC here that's just `zextOrTrunc(32);` -> `sextOrTrunc(32);`

(also interesting here since the builtin IntType is marked as signed)

Looking a little further, if this and numeric literal were changed to create signed APInts, would that eliminate the need for unsigned APInt logic? i.e., would that be worth it to not be juggling unsigned and signed APInts, or is there a performance cost there? (note, even if you think that's useful in general, it might be enough work to be left to a TODO)

toolchain/base/int_store_test.cpp
```diff
@@ -0,0 +1,150 @@
...
+};
+
+namespace {
+
+using ::testing::Eq;
+using ::testing::Not;
+
+static constexpr int MinAPWidth = IntStoreTestPeer::MinAPWidth;
+
+TEST(IntStore, Basic) {
```
jonmeow, line 32: MinAPWidth is 64, but the tests are with 32-bit values. Would it make sense to add some more 64-bit value tests? e.g.:

```
Basic:
std::numeric_limits<int64_t>::max()
std::numeric_limits<int64_t>::min()

Signed:
llvm::APInt max64(64, std::numeric_limits<int64_t>::max(), /*isSigned=*/true);
llvm::APInt min64(64, std::numeric_limits<int64_t>::min(), /*isSigned=*/true);
```

Also, maybe some 128-bit values should be tested?

toolchain/base/int_store.h
```diff
@@ -0,0 +1,186 @@
...
+    // Fallback for larger values.
+    return AddUnsignedLarge(std::move(value));
+  }
+
+  // Returns the value for an ID.
+  //
+  // This will always be a signed `APInt` with a canonical bit width for the
+  // specific integer value in question.
+  auto Get(IntId id) const -> llvm::APInt {
+    if (id.is_value()) [[likely]] {
```
jonmeow, line 278: We're currently using `LLVM_LIKELY`/`LLVM_UNLIKELY`, is this equivalent? Are you proposing to switch given the attribute is available?

toolchain/base/int_store.h
```diff
@@ -0,0 +1,186 @@
...
+    // First try directly making this into an ID.
+    if (IntId id = IntId::TryMakeSignedValue(value); id.is_valid()) [[likely]] {
+      return id;
+    }
+
+    // Fallback for larger values.
+    return AddSignedLarge(std::move(value));
+  }
+
+  // Stores a canonical copy of an unsigned value and returns its ID.
```
jonmeow: And here:

```suggestion
  // Returns the ID corresponding to an equivalent signed integer value, storing an `APInt` if necessary to represent it.
```

toolchain/base/int_store.h
```diff
@@ -0,0 +1,186 @@
...
+// width there are helper functions to coerce them to a specific desired bit
+// width for use.
+//
+// This leverages a significant optimization for small integer values -- rather
+// than canonicalizing and making them unique in a `ValueStore`, they are
+// directly embedded in the `IntId` itself. Only larger integers are stored in
+// an array of `APInt` values and represented as an index in the ID.
+class IntStore {
+ public:
+  // Adds an integer value representable in a host `int64_t` to the store.
```
jonmeow: What's a "host" `int64_t`?

Also, isn't the comment at the end more precise, that this only stores a value if it's needed?

toolchain/base/int_store.h
```diff
@@ -0,0 +1,186 @@
...
+    // First try directly making this into an ID.
+    if (IntId id = IntId::TryMakeValue(value); id.is_valid()) [[likely]] {
+      return id;
+    }
+
+    // Fallback for larger values.
+    return AddLarge(value);
+  }
+
+  // Stores a canonical copy of a signed value and returns its ID.
```
jonmeow: Isn't this only conditionally going to store a copy? Maybe similar to the above:

```suggestion
  // Returns the ID corresponding to this integer value, storing an `APInt` if
  // necessary to represent it.
```

toolchain/base/int_store.h
```diff
@@ -0,0 +1,186 @@
+// Part of the Carbon Language project, under the Apache License v2.0 with LLVM
+// Exceptions. See /LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#ifndef CARBON_TOOLCHAIN_BASE_INT_STORE_H_
```
jonmeow, line 5: Organizationally, do you think it'd help if we had toolchain/base/int.h with both IntStore and IntId? We do _somewhat_ similar in sem_ir, with things like type.h (just noting that one since it's in the PR)

toolchain/base/value_ids.h
```diff
@@ -14,6 +17,12 @@
 
 namespace Carbon {
 
+// Valid IDs which are associated with tokens during lexing need to fit into a
+// compressed storage space, which may influence the specific formulation of the
+// ID. Note that there may still be IDs either not associated with tokens or
+// computed after lexing outside of this range.
+constexpr int TokenIdBits = 23;
```
jonmeow: Very minor, but maybe IntId::TokenIdBits unless you expect other uses?

toolchain/base/value_ids.h
```diff
@@ -41,20 +50,147 @@ class Real : public Printable<Real> {
...
+  // expect invalid tokens to be *error* tokens and not need to represent an
+  // invalid integer.
+  static constexpr int TokenIdBitsShift = 32 - TokenIdBits;
+  static constexpr int32_t MaxValue =
+      std::numeric_limits<int32_t>::max() >> TokenIdBitsShift;
+  static constexpr int32_t ZeroIndexId = std::numeric_limits<int32_t>::min() >>
+                                         (TokenIdBitsShift + 1);
+  static constexpr int32_t MinValue = ZeroIndexId + 1;
+  static constexpr int32_t InvalidId = std::numeric_limits<int32_t>::min();
+  static constexpr int32_t InvalidIndex = ZeroIndexId - InvalidId;
```
jonmeow: Is it worth documenting how these constants are used? Without reading the code, to me they're just opaque names. e.g. it takes me a lot of time to figure out how ZeroIndexId is actually being used.

toolchain/base/int_store.h
```diff
@@ -0,0 +1,186 @@
...
+
+ private:
+  friend struct Testing::IntStoreTestPeer;
+
+  struct APIntId : IdBase, Printable<APIntId> {
+    using ValueType = llvm::APInt;
+    static const APIntId Invalid;
+    using IdBase::IdBase;
+    auto Print(llvm::raw_ostream& out) const -> void {
+      out << "ap-int";
```
jonmeow:
```suggestion
      out << "ap_int";
```

Looking at check/ids.h, we have things like `check_ir` (underscore) rather than hyphenation.

toolchain/base/int_store.h
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  // Returns the value for an ID adjusted to the bit width specified with
+  // another integer ID.
+  //
+  // This simply looks up the width integer ID, and then calls the above
+  // `GetAtWidth` overload using the value found for it. See that overload for
+  // more details.
+  auto GetAtWidth(IntId id, IntId bit_width_id) const -> llvm::APInt {
+    const llvm::APInt& bit_width = Get(bit_width_id);
```
jonmeow:
```suggestion
    const llvm::APInt bit_width = Get(bit_width_id);
```

Isn't the `&` incorrect here?

toolchain/base/int_store.h
```diff
@@ -0,0 +1,186 @@
...
+    CARBON_CHECK(bit_width.isStrictlyPositive() &&
+                     bit_width.isSignedIntN(sizeof(int) * 8),
```
jonmeow:
```suggestion
                     bit_width.isSignedIntN(MinAPWidth),
```

Why the calculation here versus the constant? Is it equivalent?

toolchain/base/int_store.h
```diff
@@ -0,0 +1,186 @@
...
+  // Returns the ID corresponding to this integer value, storing an `APInt` if
+  // necessary to represent it.
+  auto Add(int64_t value) -> IntId {
+    // First try directly making this into an ID.
+    if (IntId id = IntId::TryMakeValue(value); id.is_valid()) [[likely]] {
+      return id;
+    }
+
+    // Fallback for larger values.
+    return AddLarge(value);
```
jonmeow: Why do you keep these functions inline, but there are 1- and 2-line functions in the cpp file?

toolchain/base/value_ids.h
```diff
@@ -41,20 +50,147 @@ class Real : public Printable<Real> {
...
+class IntId : public Printable<IntId> {
+ public:
   using ValueType = llvm::APInt;
+
   static const IntId Invalid;
-  using IdBase::IdBase;
+
+  static auto MakeIndexOrInvalid(int index) -> IntId {
+    CARBON_DCHECK(index >= 0 && index <= InvalidIndex);
+    return IntId(ZeroIndexId - index);
```
jonmeow: Is there validation that this doesn't produce incorrect values? Is it possible to have a unit test that tries making too many unique integers, to check for graceful failure?

toolchain/base/value_ids.h
```diff
@@ -41,20 +50,147 @@ class Real : public Printable<Real> {
...
+    if (MinValue <= value && value <= MaxValue) {
+      return IntId(value);
+    }
+
+    return Invalid;
+  }
+
+  // Tries to make a signed APInt into an embedded value in the ID, and if
+  // unable to do that returns the `Invalid` ID.
+  static auto TryMakeSignedValue(llvm::APInt value) -> IntId {
```
jonmeow: FWIW, since you'd asked organizational comments, it might be worth moving these Make functions to IntStore (if the result is more compact)... for example, I'm having to flip back and forth between files in order to understand how IntStore::AddSigned works, and that might've been something that could be in one spot.

toolchain/base/int_store_test.cpp
jonmeow, line 32: Is it worth some is_index / is_value checks on results?

toolchain/base/int_store_test.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+  EXPECT_THAT(ints.Get(id1), Eq(1));
+  EXPECT_THAT(ints.Get(id2), Eq(2));
+  EXPECT_THAT(ints.Get(id3), Eq(999'999'999'999));
+}
+
+TEST(IntStore, APSigned) {
+  IntStore ints;
+
+  llvm::APInt one_ap(MinAPWidth, 1, /*isSigned=*/true);
```
jonmeow: I don't know whether you want to spend time on this, but I wonder if you could do something like:

```
struct TestValue {
  llvm::StringLiteral label;
  llvm::APInt value;
  testing::Matcher<llvm::APInt> matcher;
  // Set after initialization.
  IntId id = IntId::Invalid;
}

auto entries[] = {
  {"one_ap", llvm::APInt(MinAPWidth, 1, /*isSigned=*/true), Eq<llvm::APInt>(1)},
  ...
}

for (auto [i, entry] : llvm::enumerate(entries)) {
  SCOPED_TRACE(entry.label);
  entry.id = ints.AddSigned(entry.value);
  ASSERT_TRUE(entry.id.is_valid());
  for (int j : llvm::seq(i)) {
    EXPECT_THAT(entry.id, Not(Eq(entry[j].id)));
  }
}

for (auto entry : entries) {
  SCOPED_TRACE(entry.label);
  EXPECT_THAT(ints.Get(entry.id), entry.matcher);
}
```

Though not sure whether you like this idea or not -- do what you want with it, I was just thinking about it while trying to figure out which entries were being validated because "biggest_small_ap" has no correspondence with "ids[...]" for me.

jonmeow (commented):
Generally LG, sorry about my usual spread of comments. High level I think the IntId and IntStore changes look pretty much like what I'd expected after discussions, I'm glad for the noted performance improvements.

toolchain/base/int_store.h
chandlerc: I just noticed that we have standard attributes now. Happy to either switch to LLVM ones until we can move the rest of the code, or move the rest of the code in a follow-up.

toolchain/base/int_store.h
chandlerc: FWIW, it is correct, but extremely misleading. Your suggestion is better.

The const reference binds and life-time extends the temporary returned by `Get`. Which is what we want, but a very subtle way of getting there.

toolchain/base/value_ids.h
chandlerc: We pack other IDs into the token payload, and so there is definitely the potential to want to do something similar as we're doing for integers with other literals, particularly floating point.

But the magnitude of the problem is much smaller, and so maybe I'm imagining too much in the future. I'm happy to sink this back into `IntId` for now.

toolchain/base/int_store.h
```diff
@@ -0,0 +1,187 @@
...
+    // First try directly making this into an ID.
+    if (IntId id = IntId::TryMakeSignedValue(value); id.is_valid()) [[likely]] {
+      return id;
+    }
+
+    // Fallback for larger values.
+    return AddSignedLarge(std::move(value));
+  }
+
+  // Returns the ID corresponding to an equivalent signed integer value, storing an `APInt` if necessary to represent it.
```
CarbonInfraBot, line 74: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  // Returns the ID corresponding to an equivalent signed integer value, storing
  // an `APInt` if necessary to represent it.
```

toolchain/base/int_store.h
```diff
@@ -0,0 +1,187 @@
...
+  // another integer ID.
+  //
+  // This simply looks up the width integer ID, and then calls the above
+  // `GetAtWidth` overload using the value found for it. See that overload for
+  // more details.
+  auto GetAtWidth(IntId id, IntId bit_width_id) const -> llvm::APInt {
+    const llvm::APInt bit_width = Get(bit_width_id);
+    CARBON_CHECK(bit_width.isStrictlyPositive() &&
+                     bit_width.isSignedIntN(MinAPWidth),
+                 "Invalid bit width value: {0}", bit_width);
```
CarbonInfraBot, line 120: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
    CARBON_CHECK(
        bit_width.isStrictlyPositive() && bit_width.isSignedIntN(MinAPWidth),
        "Invalid bit width value: {0}", bit_width);
```

toolchain/sem_ir/file.h
```diff
@@ -70,6 +70,27 @@ class File : public Printable<File> {
...
+  //
+  // TODO: When we don't have a builtin int type mixed with actual `IntType`
+  // instructions, clients should directly query the `IntType` instruction to
+  // compute this information.
+  auto GetIntTypeInfo(TypeId int_type_id) const -> IntTypeInfo {
+    auto inst_id = types().GetInstId(int_type_id);
+    if (inst_id == InstId::BuiltinIntType) {
+      return {.is_signed = true, .bit_width = ints().LookupSigned(llvm::APInt(/*numBits=*/64, 32))};
+    }
```
CarbonInfraBot, line 87: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
      return {
          .is_signed = true,
          .bit_width = ints().LookupSigned(llvm::APInt(/*numBits=*/64, 32))};
```

toolchain/sem_ir/file.h
```diff
@@ -70,6 +70,27 @@
...
+    auto int_type = insts().GetAs<IntType>(inst_id);
+    auto bit_width_inst = insts().TryGetAs<IntValue>(int_type->bit_width_id);
+    return {.is_signed = int_type->int_kind.is_signed(),
+            .bit_width = bit_width_inst ? bit_width_inst->int_id : IntId::Invalid};
+  }
```
CarbonInfraBot, line 92: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
    return {
        .is_signed = int_type->int_kind.is_signed(),
        .bit_width = bit_width_inst ? bit_width_inst->int_id : IntId::Invalid};
```

toolchain/base/int_store.h
chandlerc: Yeah, good call and switched to just start with the second paragraph (tweaked a touched).

toolchain/base/int_store.h
chandlerc: Sorry about that, this is far from obvious. I only switched to this after some profiling showed surprising code size. Added some comments on the declarations of these helpers. Lemme know if they clarify enough? Also if there is a better place to put them?

toolchain/base/int_store_test.cpp
chandlerc: Good call, added tests for more of the API surface.

chandlerc: The 12-nines should be above 32-bits?

But still, added a bunch of test cases including these.

I had some larger integers in bit width, but wasn't multiplying by enough to really make the bits count. I've fixed that and so now both the 128-bit and 512-bit tests shouldn't work with any smaller power of two bit width.

toolchain/base/int_store_test.cpp
chandlerc: I agree that tracking things through the array isn't awesome. I was struggling to come up with something to minimize how much code we end up with here, but not sure I'm happy with it.

Some of this is that I think checking for collisions with IDs is somewhat an unhelpful test in the first place. Testing that we round trip through the store is really all we need here -- if the IDs collide, that will fail but in a way that's more obvious.

I've reworked the test with something similar to your suggestion if not *quite* the same to try and get similar effect. WDYT?

toolchain/base/value_ids.h
chandlerc: Hmm...

I don't think the unit test is easy to do here, as we don't even have the token payload size limitation, and so we can have a *lot* of unique integers. Should be 2 billion - 8 million or something, and each needs its own APInt.

But one thing that made me happy about the logic here is that we actually compute the ID from `InvalidIndex` (which is the largest value of `index` allowed) in a `constexpr` context below. And that should ensure that this subtraction doesn't hit UB provided the assert above it holds, and produces the expected ID value even for the largest value. And for the smallest of `0`, its pretty easy to analyze.

toolchain/base/value_ids.h
chandlerc: As discussed, merged into one file.

Once there, I moved these all to be private helper functions in `IntStore`.

I actually tried inlining most of them, but it felt slightly awkward. We end up wanting both `Add...` and `Lookup...` code paths in the store I think, at least for generality. And these helpers are useful to extract and make common between those.

I actually added another `Lookup` to simplify one of the places where we unnecessarily were forming an APInt. Currently there aren't a lot of `Lookup` calls, but it seems like an important API from a library design perspective so I didn't want to fully remove them.

That said, happy to revisit or discuss if there is a cleaner way to structure this... not super confident in the exact result I ended up with.

toolchain/sem_ir/file.h
chandlerc: I'm trying to avoid just duplicating the names of the struct fields, not sure what other documentation to provide.

I added at least one maybe more relevant part to the comment, but unsure what else you'd find helps readability here. Happy to add something you suggest?

toolchain/sem_ir/file.h
chandlerc: Ok, moved, and added the comment that this is the return for the method. Not sure what else to document that doesn't just repeat what the names already seem to document, but if you have specific things you'd like to see in a comment, happy to add them.

toolchain/check/import_ref.cpp
chandlerc: Good idea, done.

toolchain/check/handle_literal.cpp
chandlerc: This code path didn't get updated enough, all of this should have been simplified with this PR to just pass through the ID after verifying that the value fits into an `i32`. The extending and creating a new ID all stemmed from when there was implicit bit width in the integer IDs themselves. The new code should be more clear.

That said, I have thought about removing `AddUnsigned` and forcing the lexer to form the unsigned APInt, but I'm worried that would add cost due to needing a wider APInt ealier in the process.

Because we want to canonicalize the bit width inside the store, I didn't want clients to do any unnecessary resizing if possible, and the cleanest way I see to do that is to let them directly add an unsigned APInt if that's what they have.

toolchain/sem_ir/type.h
chandlerc: My hope was to migrate them to directly querying the `IntType` instruction when that becomes the canonical source of information, and to add helpers in or around that typed instruction.

But I'd also be fine with them moving to `GetIntTypeInfo` initially.

I can do this in this PR if you want, I just didn't want to make this PR larger.

I tried to change the body of this function to call `GetIntTypeInfo`, but can't because of layering. And I can't put `GetIntTypeInfo` here because of layering. (At least, I think that's the issue I remember hitting.)

toolchain/lower/constant.cpp
chandlerc: Because I was touching this code anyways, I tried to use `GetIntTypeInfo` consistently rather than adding more calls to `IsSignedInt`.

I'm happy to iterate on what code we want here. I had expected us to want to get rid of this entirely and be able to get `is_signed` from the `int_type` eventually. I can leave a TODO to that effect? (I just didn't want to get into improving helpers on `IntType` in this PR.)

toolchain/base/int_store.h
chandlerc: SGTM. I'll do the rename from `int_store.h` to `int.h` last to preserve review threads as much as I can.

toolchain/base/value_ids.h
chandlerc: Sure, I've tried to improve the comment structure a bit.

What I ended up doing is lifting the big comment block up to the `IntId` class since its now in its own file, and added dedicated comments for each constant to try and explain the interesting parts of it.

Let me know if I missed anything you would find useful or any of the new comments aren't reading well.

chandlerc (commented):
Thanks for the detailed comments, I think I've gotten to them all, but let me know if I missed anything!

toolchain/sem_ir/file.h
chandlerc: Simplified all of this by directly adding a `Lookup` that takes `int64_t` so we can use much simpler code. Does that work for you?

chandlerc (commented):
Doh, missed replying to one thread it seems, but found it now and replied below. (The code change was already in, just lost the thread.)

chandlerc:
> > This PR improves lexing by about 5% when there are lots of `i32` types.
> 
> What percentage of tokens/bytes being `i32` results in 5% lex improvement? Can you give a little more context for this?

This is just in the compile_benchmark for the lex phase, using the generated source there:
```
BM_CompileAPIFileDenseDecls<Phase::Lex>/256       36.2µs ± 2%  35.8µs ± 3%  -1.09%  (p=0.003 n=20+19)
BM_CompileAPIFileDenseDecls<Phase::Lex>/1024       163µs ± 1%   159µs ± 1%  -2.48%  (p=0.000 n=19+18)
BM_CompileAPIFileDenseDecls<Phase::Lex>/4096       660µs ± 1%   640µs ± 1%  -3.13%  (p=0.000 n=20+19)
BM_CompileAPIFileDenseDecls<Phase::Lex>/16384     2.97ms ± 2%  2.82ms ± 1%  -5.07%  (p=0.000 n=20+20)
BM_CompileAPIFileDenseDecls<Phase::Lex>/65536     12.8ms ± 1%  12.2ms ± 1%  -4.42%  (p=0.000 n=20+19)
BM_CompileAPIFileDenseDecls<Phase::Lex>/262144    58.8ms ± 1%  57.2ms ± 2%  -2.73%  (p=0.000 n=19+20)
```

Seems to fluctuate a bit between 2% and 5%. The 1% for the smallest file is because we spend more time in setup/teardown.

The % of tokens that are `i32` in this test is 4.6% -- not tiny, but also not huge.

toolchain/sem_ir/file.h
```diff
@@ -70,6 +77,25 @@ class File : public Printable<File> {
     return types().GetAs<PointerType>(pointer_id).pointee_id;
   }
 
+  // Returns integer type information from a type ID. Abstracts away the
+  // difference between an `IntType` instruction defined type and a builtin
+  // instruction defined type.
```
jonmeow:
```suggestion
  // instruction defined type. Uses IntId::Invalid for types that have an
  // invalid width.
```

toolchain/base/int_store.h
```diff
@@ -0,0 +1,422 @@
...
+  // allocated `APInt` for storage. Creating or manipulating that storage is
+  // only a few lines of code, but it ends up expensive and a lot of code so we
+  // move these out-of-line.
+  auto AddLarge(int64_t value) -> IntId;
+  auto AddSignedLarge(llvm::APInt value) -> IntId;
+  auto AddUnsignedLarge(llvm::APInt value) -> IntId;
+  auto LookupLarge(int64_t value) const -> IntId;
+  auto LookupSignedLarge(llvm::APInt value) const -> IntId;
+
+  CanonicalValueStore<APIntId> values_;
```
jonmeow, line 420:
```suggestion
  // Stores values which don't fit in an IntId. These are always signed.
  CanonicalValueStore<APIntId> values_;
```

toolchain/base/int_store.h
```diff
@@ -0,0 +1,422 @@
...
+  auto size() const -> size_t { return values_.size(); }
+
+  // Collects the memory usage of the separately stored integers.
+  auto CollectMemUsage(MemUsage& mem_usage, llvm::StringRef label) const
+      -> void;
+
+ private:
+  friend struct Testing::IntStoreTestPeer;
+
+  struct APIntId : IdBase, Printable<APIntId> {
```
jonmeow, line 352:
```suggestion
  // Used for `values_`; tracked using `IntId`'s index range.
  struct APIntId : IdBase, Printable<APIntId> {
```

toolchain/base/int_store.h
```diff
@@ -0,0 +1,422 @@
...
+  // Converts an ID to an index. Requires that `is_index()` is true.
+  //
+  // Note that this does *not* require that the ID is valid. An invalid ID will
+  // turn into an invalid index.
+  constexpr auto AsIndex() const -> int {
+    CARBON_DCHECK(is_index());
+    return ZeroIndexId - id_;
+  }
+
+  constexpr auto AsTokenPayload() const -> uint32_t {
```
jonmeow, line 115:
```suggestion
  // Returns the ID formatted as a lex token payload.
  constexpr auto AsTokenPayload() const -> uint32_t {
```

toolchain/base/int_store.h
jonmeow: My thought is we've generally agreed to use C++ attribute forms so that seems the better choice. I don't think it makes sense to switch this code if the rest changes.

toolchain/check/handle_literal.cpp
jonmeow: Yes, the new code addresses my question here.

I still wonder about the lexing numeric literal case, but that can be addressed separately. My naive belief is that the widening to 64-bit wouldn't have much effect, or even might be an improvement by eliminating a conversion. But, numeric literal parsing could also possible be improved by returning an IntIds instead of APInt after these changes.

toolchain/base/int_store.h
```diff
@@ -0,0 +1,422 @@
...
+//
+// This stores integers as abstract, signed mathematical integers. The bit width
+// of specific `APInt` values, either as inputs or outputs, is disregarded for
+// the purpose of canonicalization and the returned integer may use a very
+// different bit width `APInt` than was used when adding. There are also
+// optimized paths for adding integer values representable using native integer
+// types.
+//
+// Because the integers in the store are canonicalized without a specific bit
+// width there are helper functions to coerce them to a specific desired bit
```
jonmeow:
```suggestion
// width, there are helper functions to coerce them to a specific desired bit
```

To comment, I think it's usually "Because X, Y". The comma is because you've switched around what would directly be written as "Y because X". (Because of reversed clause order, the comma.)

toolchain/base/int_store.h
```diff
@@ -0,0 +1,422 @@
...
+// A canonicalizing value store with deep optimizations for integers.
+//
+// This stores integers as abstract, signed mathematical integers. The bit width
+// of specific `APInt` values, either as inputs or outputs, is disregarded for
+// the purpose of canonicalization and the returned integer may use a very
+// different bit width `APInt` than was used when adding. There are also
+// optimized paths for adding integer values representable using native integer
+// types.
+//
+// Because the integers in the store are canonicalized without a specific bit
```
jonmeow:
```suggestion
// Because the integers in the store are canonicalized with only a minimum bit
```

toolchain/base/value_ids.h
jonmeow: More focused on lex, that's a lower limit of 2 million right? Is that feasible to test, like with a string of long integers one after the other?

I think 2B may be infeasible to reach until we get metaprogramming.

Note, fine to not address this in this PR, but I do lean towards that we should test lex thresholds given the low-ish limits.

toolchain/sem_ir/file.h
```diff
@@ -34,6 +35,12 @@ namespace Carbon::SemIR {
 // Provides semantic analysis on a Parse::Tree.
 class File : public Printable<File> {
  public:
+  // Used to return information about an integer type in `GetIntTypeInfo`.
+  struct IntTypeInfo {
+    bool is_signed;
+    IntId bit_width = IntId::Invalid;
```
jonmeow:
```suggestion
    IntId bit_width;
```

Looks like this default is unused, suggesting removal.

toolchain/base/int_store.h
```diff
@@ -0,0 +1,422 @@
...
+  static constexpr int32_t MaxValue =
+      std::numeric_limits<int32_t>::max() >> TokenIdBitsShift;
+
+  // The ID value that represents an index of `0`. This is the first ID value
+  // representing an index, and all indices are `<=` to this.
+  //
+  // Because this is the first index ID, and we encoded indices as successive
+  // negative numbers counting downwards, we can both use a comparison with
+  // this ID to distinguish value and index IDs, and to compute the actual index
+  // from the ID. The computation of an index in fact is just a subtraction:
```
jonmeow: I'm having trouble reading this due to the commas. What do you think of:

```suggestion
  // ZeroIndexId is the first index ID, and we encode indices as successive
  // negative numbers counting downwards. The setup allows us to both use a comparison with
  // this ID to distinguish value and index IDs, and to compute the actual index
  // from the ID.
  //
  // The computation of an index in fact is just a subtraction:
```

toolchain/base/int_store.h
```diff
@@ -0,0 +1,422 @@
...
+  // Canonicalize an incoming signed APInt to the correct bit width.
+  static auto CanonicalizeSigned(llvm::APInt value) -> llvm::APInt;
+
+  // Canonicalize an incoming unsigned APInt to the correct bit width.
+  static auto CanonicalizeUnsigned(llvm::APInt value) -> llvm::APInt;
+
+  // Helper functions for handling values that are large enough to require an
+  // allocated `APInt` for storage. Creating or manipulating that storage is
+  // only a few lines of code, but it ends up expensive and a lot of code so we
+  // move these out-of-line.
```
jonmeow:
```suggestion
  // only a few lines of code, but we move these out-of-line because the generated code is big and harms performance for the non-`Large` common case.
```

Suggesting a slightly different comment due to discussion.

toolchain/base/int_store_test.cpp
jonmeow: Looks good, thanks!

toolchain/base/int_store.h
```diff
@@ -0,0 +1,422 @@
...
+
+  // Document the specific values of some of these constants to help visualize
+  // how the bit patterns map from the above computations.
+  //
+  // Each bit is either `T` for part of the token or `P` as part
+  // of the available payload that we use for the ID:
+  //
+  // clang-format off: visualizing bit positions
+  //
+  //                           0bTTTT'TTTT'TPPP'PPPP'PPPP'PPPP'PPPP'PPPP
```
jonmeow: Nuanced thing, the `clang-format` throws me a little, maybe one of:

```suggestion
  // clang-format off: visualizing bit positions
  //
  // Each bit is either `T` for part of the token or `P` as part
  // of the available payload that we use for the ID:
  //
  //                           0bTTTT'TTTT'TPPP'PPPP'PPPP'PPPP'PPPP'PPPP
```

```suggestion
  // Each bit is tagged either `T` for part of the token or `P` as part
  // of the available payload that we use for the ID.
  //
  // clang-format off: visualizing bit positions
  //
  //                           0bTTTT'TTTT'TPPP'PPPP'PPPP'PPPP'PPPP'PPPP
```

jonmeow (approved):
This looks good. I think my comments are pretty small, except for the one "lex file with 2M ints" test suggestion which I'm happy to split out. So feel free to merge when you've had a chance to go through remaining stuff.

toolchain/base/int_store.h
```diff
@@ -0,0 +1,426 @@
...
+  static constexpr int32_t MaxValue =
+      std::numeric_limits<int32_t>::max() >> TokenIdBitsShift;
+
+  // The ID value that represents an index of `0`. This is the first ID value
+  // representing an index, and all indices are `<=` to this.
+  //
+  // ZeroIndexId is the first index ID, and we encode indices as successive
+  // negative numbers counting downwards. The setup allows us to both use a comparison with
+  // this ID to distinguish value and index IDs, and to compute the actual index
+  // from the ID.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  // negative numbers counting downwards. The setup allows us to both use a
  // comparison with this ID to distinguish value and index IDs, and to compute
  // the actual index from the ID.
```

toolchain/base/int_store.h
```diff
@@ -0,0 +1,426 @@
...
+
+  // Canonicalize an incoming signed APInt to the correct bit width.
+  static auto CanonicalizeSigned(llvm::APInt value) -> llvm::APInt;
+
+  // Canonicalize an incoming unsigned APInt to the correct bit width.
+  static auto CanonicalizeUnsigned(llvm::APInt value) -> llvm::APInt;
+
+  // Helper functions for handling values that are large enough to require an
+  // allocated `APInt` for storage. Creating or manipulating that storage is
+  // only a few lines of code, but we move these out-of-line because the generated code is big and harms performance for the non-`Large` common case.
```
CarbonInfraBot: **[diff]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>
```suggestion
  // only a few lines of code, but we move these out-of-line because the
  // generated code is big and harms performance for the non-`Large` common
  // case.
```

toolchain/base/value_ids.h
chandlerc: Sure, happy to look at testing the lexer limit in a follow-up.

toolchain/base/int_store.h
chandlerc: Will do.

toolchain/check/handle_literal.cpp
chandlerc: Yeah... There are a lot of potential improvements inside the numeric literal lexing at this point.

- Fast path small integers w/o forming an APInt at all
- As you suggest, avoid moving them around and just use the IntId

But leaving that for the future...

chandlerc:
> This looks good. I think my comments are pretty small, except for the one "lex file with 2M ints" test suggestion which I'm happy to split out. So feel free to merge when you've had a chance to go through remaining stuff.

Thanks! All the comment fixes applied, some responses inline but all agreeing. I'll merge when ready.

PR #4490, opened 2024-11-05
Skip vptr when performing object initialization

dwblaikie:
The cehck for vptr could be done differently (is this class dynamic and
its base class non-dynamic), and if we change the layout (to put the
vptr after any base) then this code will break (could add an assertion
that the vptr isn't present apart from at the start of the field list if
that'd seem worthwhile).

There's still later issues with lowering (if this patch causes crashes
in lowering, do they result in fuzz failures/need to be avoided before
this change is submitted?)


toolchain/check/convert.cpp
```diff
@@ -449,14 +455,18 @@ static auto ConvertStructToStructOrClass(Context& context,
...
           ? SemIR::CopyOnWriteInstBlock(sem_ir, literal_elems_id)
           : SemIR::CopyOnWriteInstBlock(
                 sem_ir, SemIR::CopyOnWriteInstBlock::UninitializedBlock{
-                            src_elem_fields.size()});
+                            dest_elem_fields.size()});
   for (auto [i, dest_field_id] : llvm::enumerate(dest_elem_fields)) {
     auto dest_field =
         sem_ir.insts().GetAs<SemIR::StructTypeField>(dest_field_id);
+    if (dest_field.name_id == SemIR::NameId::Vptr) {
+      new_block.Set(i, SemIR::InstId::BuiltinError);
```
zygoloid, line 468: Please add a TODO comment here to put a real vtable pointer in the vptr slot.

toolchain/check/convert.cpp
```diff
@@ -392,6 +392,12 @@ static auto ConvertStructToStructOrClass(Context& context,
   auto& sem_ir = context.sem_ir();
   auto src_elem_fields = sem_ir.inst_blocks().Get(src_type.fields_id);
   auto dest_elem_fields = sem_ir.inst_blocks().Get(dest_type.fields_id);
+  bool dest_has_vptr =
+      !dest_elem_fields.empty() &&
+      sem_ir.insts()
+              .GetAs<SemIR::StructTypeField>(dest_elem_fields.front())
+              .name_id == SemIR::NameId::Vptr;
```
zygoloid, line 399: This might be worth factoring out into a helper function. Let's keep an eye on this and see if we want to do it in more places.

zygoloid (approved):
Looks good, thanks!

PR #4492, opened 2024-11-05
Convert StructTypeField to a specific type.

jonmeow:
This converts `StructTypeField` from an instruction to a dedicated type, with its own store. This had originated from discussing how `.GetAs<SemIR::StructTypeField>` was more prevalent than for other instructions, but is probably more interesting for the storage savings (16 bytes StructTypeField + 4 byte LocId + 4 byte InstId -> 8 byte StructTypeField).

Due to the different structure, these now have their own stack during construction, reducing (but not eliminating) `args_type_info_stack_` use-cases.

The test changes of different InstIds is expected because structs and classes generate fewer instructions now. Other than that, results should remain the same.

I'm generally trying to avoid unrelated cleanup here due to the PR size, though I did scrutinize the `VerifyOnFinish` calls, adding one and commenting others (putting them in member order because that's how I was checking what was verified and what wasn't).


toolchain/check/testdata/struct/fail_duplicate_name.carbon
```diff
@@ -8,48 +8,48 @@
...
+// CHECK:STDERR:                  ^~~~~~~
 // CHECK:STDERR:
 let v: {.a: i32, .a: i32} = {.a = 1};
 
-// CHECK:STDERR: fail_duplicate_name.carbon:[[@LINE+7]]:26: error: duplicated field name `def` in struct literal [StructNameDuplicate]
+// CHECK:STDERR: fail_duplicate_name.carbon:[[@LINE+7]]:32: error: duplicated field name `def` in struct literal [StructNameDuplicate]
 // CHECK:STDERR: let w: i32 = {.def = 1, .def = 2}.def;
-// CHECK:STDERR:                          ^~~
-// CHECK:STDERR: fail_duplicate_name.carbon:[[@LINE+4]]:16: note: field with the same name here [StructNamePrevious]
+// CHECK:STDERR:                                ^
```
josh11b: This location information doesn't seem to be as good as before.

josh11b (commented):
Nice improvement!

I was expecting you would have to add handling for `StructTypeFieldsId` to check/deduce.cpp, but it looks like the corresponding test is a TODO, see
https://github.com/carbon-language/carbon-lang/blob/trunk/toolchain/check/testdata/function/generic/deduce.carbon#L84

I think this change makes fixing that test pretty straightforward. If you were feeling ambitious, you could fix it in this PR or a follow-up.

jonmeow:
> Nice improvement!
> 
> I was expecting you would have to add handling for `StructTypeFieldsId` to check/deduce.cpp, but it looks like the corresponding test is a TODO, see https://github.com/carbon-language/carbon-lang/blob/trunk/toolchain/check/testdata/function/generic/deduce.carbon#L84
> 
> I think this change makes fixing that test pretty straightforward. If you were feeling ambitious, you could fix it in this PR or a follow-up.

Thanks for pointing this out, but this PR is already very big and complex, and I don't think fixing an unrelated TODO should be part of it.

toolchain/check/testdata/struct/fail_duplicate_name.carbon
jonmeow: Fixed, should be back to what it was. Probably a better approach for transferring names too.

josh11b (commented):
I have recently touched deduce.cpp, I'd be happy to make that change since I'm familiar with what is called for.

josh11b (approved):

PR #4494, opened 2024-11-05
Remove forward references from binding patterns

geoffromer:
This is primarily to free up space in the BindingPattern insts, but as a side effect it moves the link between BindingPattern and its BindName out of the SemIR, and into a transient data structure in Context.


geoffromer:
This PR seems to fix a pre-existing bug where we were materializing redundant copies of `symbolic_binding_pattern` insts, but for the life of me I can't explain where the bug is or how this PR could possibly have fixed it.

toolchain/check/handle_binding_pattern.cpp
```diff
@@ -219,18 +219,19 @@ static auto HandleAnyBindingPattern(Context& context, Parse::NodeId node_id,
         context.AddNameToLookup(name_id, bind_id);
         auto entity_name_id =
             context.insts().GetAs<SemIR::AnyBindName>(bind_id).entity_name_id;
+        CARBON_CHECK(context.bind_name_cache()
+                         .Insert(entity_name_id, bind_id)
+                         .is_inserted());
```
zygoloid: IIRC we've said in the past that we don't want side-effects in `CHECK` conditions because, while they're always evaluated now, we might revisit that choice in the future. (As a reader I also tend to skip over `CHECK`s when working out what the code does.)
```suggestion
        bool inserted = context.bind_name_cache()
                         .Insert(entity_name_id, bind_id)
                         .is_inserted();
        CARBON_CHECK(inserted);
```

toolchain/check/pattern_match.cpp
```diff
@@ -151,14 +151,15 @@ auto MatchContext::EmitPatternMatch(Context& context,
     case SemIR::SymbolicBindingPattern::Kind: {
       CARBON_CHECK(kind_ == MatchKind::Callee);
       auto binding_pattern = pattern.inst.As<SemIR::AnyBindingPattern>();
-      auto bind_name = context.insts().GetAs<SemIR::AnyBindName>(
-          binding_pattern.bind_name_id);
+      auto bind_name_id = context.bind_name_cache()
+                              .Lookup(binding_pattern.entity_name_id)
+                              .value();
```
zygoloid: Does it make sense to remove the entry from the cache here?

zygoloid (approved):
For the weird bug fix: it looks like we were previously including the `bind_name_id` in the canonical identity of a `SymbolicBindingPattern`, so we got different constant values for different `bind_name_id`s -- and you can't see the difference because the formatter was not printing the `MatchingInstId` field.

toolchain/check/pattern_match.cpp
geoffromer: I think removing it outright could mask bugs, by allowing it to be re-inserted (which should never happen), but we can at least invalidate the entry. Done.

PR #4497, opened 2024-11-06
Don't create instructions for implicit constants. 

zygoloid:
When an instruction is created as part of an implicit call to an
interface member, we generated a bunch of constants for naming the
interface, finding the corresponding specific, accessing its member
function, and so on. This led to significant bloat in SemIR.

Instead, we now track whether an instruction is created implicitly in
its location, and where relevant, we use the constant value of the
instruction directly instead of storing a new `Inst`.

This doesn't reduce the amount of work we need to do, but does make the
representation in SemIR smaller and more readable.


zygoloid:
I'm not entirely sure whether this is a good direction, but it seems like it should be easy to back out if we decide we don't like it.

josh11b (approved):

PR #4499, opened 2024-11-07
Replace `InterfaceType` with `FacetType`

josh11b:
This does a few things:
* Replaces the single `TypeId` in the `FacetTypeInfo` struct with a vector of `InterfaceId`, `SpecificId` pairs (sorted in id order) representing the set of interface requirements of the facet type. This will later be used to support facet types with multiple interface requirements (as in `I & J` or `I where .Self impls J`).
* Replace `InterfaceType` instructions (used as the type of an `InterfaceDecl` instruction) with `FacetType` instructions (introduced in #4460) with a (newly introduced) `FacetTypeFromInterface()` function.
* Replace code that consumed `InterfaceType` values with code that consumed `FaceType` values. I've generally left the assumption in the code that it is dealing with a single interface, using the (newly introduced) `FacetTypeInfo::TryAsSingleInterface`, and producing an error otherwise. There isn't yet support for the `&` operator or `where .Self impls`, so this is generally a good assumption for now, except you can get a facet type with no associated interfaces from a `type where`... expression. In some cases, the facet type value is pulled from the evaluation of an `InterfaceDecl` instruction, where the single interface assumption will hold permanently.
* Some related cleans up: nicer stringification and formatting of facet types, suppression of some errors when there already was an error.

There is still a lot left to do, including:
* Type `type` should be a facet type with a reserved id, replacing the built-in instruction.
* Code using `TryAsSingleInterface` should generally be upgraded to handle more than (or less than) one interface. Name lookup should be particularly exciting.
* Operator `&` should be defined on facet types, unioning their interface and other requirements.
* Requirements from a `where` clause don't do anything yet.
* Impls and impl lookup need to resolve facet types, and do things like determine if all the associated constants are given values.


toolchain/sem_ir/facet_type_info.h
```diff
@@ -13,20 +15,72 @@ struct FacetTypeInfo : Printable<FacetTypeInfo> {
   // TODO: Need to switch to a processed, canonical form, that can support facet
   // type equality as defined by
   // https://github.com/carbon-language/carbon-lang/issues/2409.
-  TypeId base_facet_type_id;
+
+  // TODO: Replace these vectors with an array allocated in an
+  // `llvm::BumpPtrAllocator`.
+
+  // `Impls` holds the interfaces this facet type requires.
+  struct Impls {
```
geoffromer: Would it be accurate to say that this represents an `impls` constraint, or a `.Self impls` constraint? If so, I think it would help to say that. Also, it seems like this only holds one interface?

Instead or in addition, it might make sense to call this `ImplsConstraint` instead of `Impls`. Both changes will help steer readers away from interpreting `Impls` as the plural of `Impl`.

toolchain/sem_ir/formatter.cpp
```diff
@@ -929,11 +929,27 @@ class FormatterImpl {
...
+          out_ << ", ";
+          FormatName(impls.specific_id);
+        }
+      }
+    }
+
     if (info.requirement_block_id.is_valid()) {
       // TODO: Include specifics.
-      out_ << "+requirements";
+      out_ << " where...";
```
geoffromer: Might it make sense to make this something like "where TODO", to make the output a little more self-explanatory?

toolchain/check/context.cpp
```diff
@@ -1190,7 +1188,18 @@ auto Context::TryToDefineType(SemIR::TypeId type_id,
...
   }
 
-  if (auto interface = types().TryGetAs<SemIR::InterfaceType>(type_id)) {
+  if (auto facet_type = types().TryGetAs<SemIR::FacetType>(type_id)) {
+    const auto& facet_type_info =
+        sem_ir().facet_types().Get(facet_type->facet_type_id);
+    auto interface = facet_type_info.TryAsSingleInterface();
+    if (!interface) {
+      auto builder = diagnoser();
+      CARBON_DIAGNOSTIC(SingleInterfaceFacetTypeOnly, Note,
```
geoffromer, line 1194: Would `context.TODO` work here instead?

toolchain/check/context.cpp
```diff
@@ -1221,6 +1230,17 @@ auto Context::GetTypeIdForTypeConstant(SemIR::ConstantId constant_id)
...
 }
 
+auto Context::FacetTypeFromInterface(SemIR::InterfaceId interface_id,
+                                     SemIR::SpecificId specific_id)
+    -> SemIR::FacetType {
+  llvm::SmallVector<SemIR::FacetTypeInfo::Impls> impls;
+  impls.emplace_back(interface_id, specific_id);
+  SemIR::FacetTypeId facet_type_id =
+      sem_ir().facet_types().Add(SemIR::FacetTypeInfo{
+          .impls = impls, .requirement_block_id = SemIR::InstBlockId::Invalid});
```
geoffromer: Would something like this work instead? If so, I think I might find it a little clearer because it's more declarative.

```suggestion
  SemIR::FacetTypeId facet_type_id =
      sem_ir().facet_types().Add(SemIR::FacetTypeInfo{
          .impls = {{interface_id, specific_id}}, .requirement_block_id = SemIR::InstBlockId::Invalid});
```

toolchain/check/eval.cpp
```diff
@@ -1506,15 +1505,28 @@ static auto TryEvalInstInContext(EvalContext& eval_context,
...
-      SemIR::InstBlockId requirement_block_id = typed_inst.requirements_id;
-      // TODO: Process & canonicalize requirements.
-      return MakeFacetTypeResult(eval_context.context(), base_facet_type_id,
-                                 requirement_block_id, phase);
+      SemIR::Inst base_facet_inst =
+          eval_context.GetConstantValueAsInst(base_facet_type_id);
+      SemIR::FacetTypeInfo info = {.requirement_block_id =
+                                       SemIR::InstBlockId::Invalid};
+      // `where` provides that the base facet is an error, `type`, or a facet
+      // type.
```
geoffromer, line 1516: Then shouldn't there be a `type` case below?

toolchain/check/import_ref.cpp
```diff
@@ -965,12 +965,20 @@ class ImportRefResolver {
       case CARBON_KIND(SemIR::ClassType inst): {
         return context_.classes().Get(inst.class_id).scope_id;
       }
+      case CARBON_KIND(SemIR::FacetType inst): {
+        const SemIR::FacetTypeInfo& facet_type_info =
+            context_.sem_ir().facet_types().Get(inst.facet_type_id);
+        // This is specifically the facet type produce by an interface
```
geoffromer:
```suggestion
        // This is specifically the facet type produced by an interface
```

toolchain/check/import_ref.cpp
```diff
@@ -2055,33 +2066,55 @@ class ImportRefResolver {
...
+        const SemIR::FacetTypeInfo& new_facet_type_info =
+            context_.sem_ir().facet_types().Get(facet_type->facet_type_id);
+        impls.append(new_facet_type_info.impls);
+      } else {
+        auto generic_interface_type =
+            context_.types().GetAs<SemIR::GenericInterfaceType>(
+                interface_const_inst.type_id());
+        auto specific_id =
+            GetOrAddLocalSpecific(interface.specific_id, specific_data);
+        impls.emplace_back(generic_interface_type.interface_id, specific_id);
```
geoffromer: It's [generally better](https://abseil.io/tips/112) to use `push_back` when it works.

```suggestion
        impls.push_back({generic_interface_type.interface_id, specific_id});
```

toolchain/check/member_access.cpp
```diff
@@ -55,10 +55,15 @@ static auto GetAsLookupScope(Context& context, SemIR::LocId loc_id,
...
-                       .specific_id = base_as_interface->specific_id};
+    const auto& facet_type_info =
+        context.sem_ir().facet_types().Get(base_as_facet_type->facet_type_id);
+    auto base_as_interface = facet_type_info.TryAsSingleInterface();
+    if (base_as_interface) {
+      auto& interface_info =
+          context.interfaces().Get(base_as_interface->interface_id);
+      return LookupScope{.name_scope_id = interface_info.scope_id,
+                         .specific_id = base_as_interface->specific_id};
+    }
```
geoffromer, line 66: Do you intend to eventually support multiple interfaces here? If so, should there be a `context.TODO` or at least a TODO comment?

toolchain/check/testdata/impl/fail_impl_bad_interface.carbon
```diff
@@ -8,19 +8,39 @@
 // TIP: To dump output, run:
 // TIP:   bazel run //toolchain/testing:file_test -- --dump_output --file_tests=toolchain/check/testdata/impl/fail_impl_bad_interface.carbon
 
-// CHECK:STDERR: fail_impl_bad_interface.carbon:[[@LINE+10]]:1: error: semantics TODO: `impl as non-interface` [SemanticsTodo]
```
geoffromer, line 11: FWIW, losing this diagnostic seems like a regression to me, since it's the clearest explanation of what's wrong with this code, but we can presumably circle back to that later.

toolchain/check/testdata/impl/fail_impl_bad_interface.carbon
```diff
@@ -8,19 +8,39 @@
...
+
+library "[[@TEST_NAME]]";
+
+// CHECK:STDERR: fail_impl_as_type.carbon:[[@LINE+4]]:1: error: impl as non-facet-type [ImplAsNonFacetType]
+// CHECK:STDERR: impl bool as type {}
+// CHECK:STDERR: ^~~~~~~~~~~~~~~~~~~
+// CHECK:STDERR:
+impl bool as type {}
+
+// --- fail_impl_as_type_where.carbon
```
geoffromer, line 34: We often seem to name files `fail_todo_...` or `todo_fail...` when the failure is expected, but a bug, and `file_test` has support for that.

toolchain/check/context.cpp
josh11b: I considered that, but there was an issue with getting the location, which is currently hidden inside the `diagnoser` closure, so I opted into adding a note to it instead. This particular case I have replacement code that doesn't use `TryAsSingleInterface` and doesn't need this diagnostic, if you'd like me to backport that change.

toolchain/check/eval.cpp
josh11b: The `else` case CARBON_CHECKs that it is `type`.

toolchain/check/member_access.cpp
josh11b: I have a TODO on `TryAsSingleInterface`, wasn't clear to that they were also needed on the calls to it.

toolchain/sem_ir/facet_type_info.h
josh11b: I renamed the class and the vector, but there already is a TODO to extend this beyond `.Self impls`. Is that enough or would you like me to do more here?

toolchain/check/testdata/impl/fail_impl_bad_interface.carbon
josh11b: I don't honestly expect this particular test to pass even when the code is improved to handle more than one interface, since "impl as <something with no interface requirements>" should likely not be allowed in general. I added a comment.

toolchain/check/testdata/impl/fail_impl_bad_interface.carbon
josh11b: The diagnostic was not particularly targeted before. It would print this anytime there was any error in the expression after the `as`.

toolchain/sem_ir/formatter.cpp
josh11b: Sure, done.

geoffromer (approved):

PR #4500, opened 2024-11-07
Support deduction of the types of struct fields

josh11b:
Follow-on to #4492 .

### DIFF
--- a/toolchain/check/deduce.cpp
+++ b/toolchain/check/deduce.cpp
@@ -80,6 +80,28 @@ class DeductionWorklist {
     AddAll(context_.inst_blocks().Get(params), args, needs_substitution);
   }
 
+  auto AddAll(SemIR::StructTypeFieldsId params, SemIR::StructTypeFieldsId args,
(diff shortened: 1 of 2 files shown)


toolchain/check/deduce.cpp
```diff
@@ -80,6 +80,28 @@ class DeductionWorklist {
...
+      // as non-deduced. For now we treat it as non-deduced.
+      return;
+    }
+    // Don't do deduction unless the names match in order.
+    // TODO: Support reordering of names.
+    for (auto [param, arg] : llvm::zip_equal(param_fields, arg_fields)) {
+      if (param.name_id != arg.name_id) {
+        return;
+      }
+    }
```
jonmeow, line 98: How are each of these tested? I see where it's tested when the struct is consistent, but what about the various invalid cases?

josh11b: Added tests.

jonmeow (approved):
Thanks! Looks good.

PR #4501, opened 2024-11-07
Preserve the is_dynamic property of classes when importing them

dwblaikie:
The test update shows a class derived from an imported base class with a vptr, and without this change the derived class got its own vptr, with this change the derived class can see the base is dynamic, so the derived doesn't need to add a vptr and can rely on the base class's vptr instead.


dwblaikie:
The test update shows a class derived from an imported base class with a vptr, and without this change the derived class got its own vptr, with this change the derived class can see the base is dynamic, so the derived doesn't need to add a vptr and can rely on the base class's vptr instead.

jonmeow (approved):
LG, though should your comment in the conversation be moved into the PR description so that it's used as the commit message on merge?

PR #4502, opened 2024-11-07
Make empty ids for all block types

jonmeow:
Adding empty IDs to all block types. This is primarily for TypeBlockId::Empty, which allows me to disambiguate empty tuple types, which I think might yield a SemIR readability improvement (PR forthcoming). I'm splitting PRs so that the limited test impact here is clear.


dwblaikie (approved):
SGTM

PR #4503, opened 2024-11-07
Name empty tuples distinctly in SemIR.

jonmeow:
Building on #4502, give empty tuple values a distinct name, and also explicitly name tuple types (previously values but not types were named).


jonmeow:
To be clear, depends on #4502

dwblaikie (approved):
Looks good

PR #4504, opened 2024-11-07
Test `class` with multiple `extend`

josh11b:
Adds coverage for an existing diagnostic. Code appears to work without modification.


josh11b:
It is complaining that a test that I thought was succeeding because of no diagnostics is actually failing (and does not start with `fail_`). I'm investigating.

jonmeow:
You just need `library "[[@TEST_NAME]]"`

josh11b:
> library "[[@TEST_NAME]]"

Thanks, should be fixed now.

josh11b:
I just realized some comments need to be updated too.

josh11b:
> I just realized some comments need to be updated too.

Done now.

jonmeow (approved):

PR #4505, opened 2024-11-08
Change the bazel-out structure to avoid busybox symlinks.

jonmeow:
As described in symlink_helpers.bzl, copied here for visibility:

Symlinking busybox things needs special logic.

This is because Bazel doesn't cache the actual symlink, resulting in essentially resolved symlinks being produced in place of the expected tool. As a consequence, we can't rely on the symlink name when dealing with busybox entries.

An example repro of this using a local build cache is:

    bazel build //toolchain
    bazel clean
    bazel build //toolchain

We could in theory get reasonable behavior with `ctx.actions.declare_symlink`, but that's disallowed in our `.bazelrc` for cross-environment compatibility.

The particular approach here uses the Python script as a launching pad so that the busybox still receives an appropriate location in argv[0], allowing it to find other files in the lib directory. Arguments are inserted to get equivalent behavior as if symlink resolution had occurred.

The underlying bug is noted at: https://github.com/bazelbuild/bazel/issues/23620

### DIFF
--- a/toolchain/install/symlink_helpers.bzl
+++ b/toolchain/install/symlink_helpers.bzl
@@ -4,36 +4,66 @@
 
 """Rules for symlinking in ways that assist install_filegroups."""
 
-def _symlink_filegroup_impl(ctx):
-    prefix = ctx.attr.out_prefix
+_SYMLINK_BUSYBOX_TMPL = """#!/usr/bin/env python3
 
-    outputs = []
-    for f in ctx.files.srcs:
-        # We normalize the path to be package-relative in order to ensure
-        # consistent paths across possible repositories.
-        relative_path = f.short_path.removeprefix(f.owner.package)
+from pathlib import Path
+import os
+import sys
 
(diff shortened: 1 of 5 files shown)


toolchain/install/symlink_helpers.bzl
```diff
@@ -75,3 +105,35 @@ symlink_file = rule(
         "symlink_label": attr.label(allow_single_file = True),
     },
 )
+
```
geoffromer, line 108: Did you make any changes below here? I ask because I don't see any changes compared to the code that was previously at the top of the file, but if there aren't any, it's bizarre that the diff is treating it as new code.

geoffromer (approved):

toolchain/install/symlink_helpers.bzl
jonmeow: No, this was just shuffling as I was trying to figure out organization.

PR #4507, opened 2024-11-11
Include `specific_id` in `extended_scopes`, fixing some uses of extending a generic

josh11b:


zygoloid (commented):
I think tracking a `SpecificId` in a `NameScopeId` won't be sufficient when that name scope is itself generic. For example, given:
```
class A(T:! type) { var x: T; }
class B(T:! type) { extend base: A(T); }
```
... the `NameScope` for `B` will list class `A` with the `SpecificId` for `A(T)`, and when we do a name lookup into `B(i32)` we won't have a good way to find the `SpecificId` for `A(i32)`.

I think one reasonable way to deal with this would be for the list of extended scopes to be a list of `ConstantId`s, not a list of `NameScopeId`s. So for the above case, we'd list that the `NameScope` for `B` extends the `ClassType` `A(T)`. When looking through  the list of extended scopes, we could then get the constant value of that `ConstantId` in the current specific (eg, in `B(i32)` in the example), map the resulting `ConstantId` to a `NameScopeId` and `SpecificId` pair, and continue lookup in the extended scope.

josh11b:
This has been replaced by #4524 .

PR #4508, opened 2024-11-11
Remove some single-interface restrictions from some uses of facet types

josh11b:


zygoloid (approved):
Looks good.

Can you add some test coverage that demonstrates we're checking that each interface in a multi-interface facet type is checked for completeness as appropriate?

josh11b:
> Can you add some test coverage that demonstrates we're checking that each interface in a multi-interface facet type is checked for completeness as appropriate?

We currently have no way to construct a multi-interface facet type.

